PR #445, opened 2015-08-14
Const tut

adelbertc:


docs/src/main/tut/const.md
```diff
@@ -6,4 +6,278 @@ source: "https://github.com/non/cats/blob/master/data/src/main/scala/cats/data/C
...
+```
+
+The `const` function takes two arguments and simply returns the first argument, ignoring the second.
+
+```scala
+final case class Const[A, B](getConst: A)
+```
+
+The `Const` data type takes two type parameters, but only ever stores a value of the first type paramter.
+Because it never actually uses the second type parameter, it is referred to as a "phantom type."
```
fthomas: Small nitpick: put the dot after the double quote in "phantom type."

non: Agreed. This goes against the punctuation rules that Americans are taught, but conforms to British usage (and the convention will help us avoid other ambiguities in our documentation).

https://en.wikipedia.org/wiki/Quotation_marks_in_English#Order_of_punctuation

fthomas: Interesting, I've never seen this one. As a programmer the American style just looks like a syntax error :-)

non: I've always thought so too -- it reminds me of the bad old days of `<b>broken <i>html</b> tags</i>`.

stew:
:+1:

ceedubs:
Fantastic. You are on a roll, @adelbertc!

PR #447, opened 2015-08-14
this traversal method is undocumented and unused

stew:
Can find any place we use this, if we aren't using it, should we just remove it?

To me it doesn't look like it adds much value on its own.

### DIFF
--- a/core/src/main/scala/cats/Traverse.scala
+++ b/core/src/main/scala/cats/Traverse.scala
@@ -45,11 +45,4 @@ def traverseU[A, GB](fa: F[A])(f: A => GB)(implicit U: Unapply[Applicative, GB])
 
   override def map[A, B](fa: F[A])(f: A => B): F[B] =
     traverse[Id, A, B](fa)(f)
-
-  def traversal[G[_]: Applicative]: Traversal[G] =
-    new Traversal[G]
-
-  class Traversal[G[_]: Applicative] {
-    def run[A, B](fa: F[A])(f: A => G[B]): G[F[B]] = traverse[G, A, B](fa)(f)
-  }
 }


non:
:+1:

ceedubs:
:+1:

adelbertc:
Only use case I can think of off the top of my head was to specify the `G[_]` to aid the type checker (`traverse` has issues inferring, say, `Validated[String, ?]`) but given we have `traverseU`...

```scala
// Can't figure out `Applicative[Validated[String, ?]]`
scala> Traverse[List].traverse(List(1, 2, 3))(Validated.valid[String, Int])
<console>:20: error: no type parameters for method traverse: (f: Int => G[B])(implicit evidence$1: cats.Applicative[G])G[List[B]] exist so that it can be applied to arguments (Int => cats.data.Validated[String,Int])
 --- because ---
argument expression's type is not compatible with formal parameter type;
 found   : Int => cats.data.Validated[String,Int]
 required: Int => ?G[?B]

       Traverse[List].traverse(List(1, 2, 3))(Validated.valid[String, Int])
                              ^
<console>:20: error: type mismatch;
 found   : Int => cats.data.Validated[String,Int]
 required: Int => G[B]
       Traverse[List].traverse(List(1, 2, 3))(Validated.valid[String, Int])
                                                             ^
<console>:20: error: could not find implicit value for evidence parameter of type cats.Applicative[G]
       Traverse[List].traverse(List(1, 2, 3))(Validated.valid[String, Int])
                                             ^

// We just need to specify the `Validated[String, ?]` bit without needing the `A` or `B`
scala> Traverse[List].traversal[Validated[String, ?]].run(List(1, 2, 3))(Validated.valid[String, Int])
res3: cats.data.Validated[String,List[Int]] = Valid(List(1, 2, 3))

// Or we can use `traverseU`...
scala> Traverse[List].traverseU(List(1, 2, 3))(Validated.valid[String, Int])
res4: cats.data.Validated[String,List[Int]] = Valid(List(1, 2, 3))

// Behavior if we didn't have `traverseU` or `traversal`
scala> Traverse[List].traverse[Validated[String, ?], Int, Int](List(1, 2, 3))(Validated.valid[String, Int])
res5: cats.data.Validated[String,List[Int]] = Valid(List(1, 2, 3))
```

PR #448, opened 2015-08-14
Provide some documentation for Semigroup

philwills:
Even though Semigroup is defined in Algebra, it seems useful to provide some documentation for it here, as cats provides an alias.


philwills:
A lot of this is lightly adapted from the SemigroupK docs. 

I don't claim to have any deep background in category theory, so all corrections and feedback would be most welcome.

non:
Looks good to me. Thanks! :+1:

PR #449, opened 2015-08-15
Fix #442: Link to .html instead of .md files in docs

### LINKED ISSUE #442: Links in Applicative doc point to .md not .html
The links on page http://non.github.io/cats//tut/applicative.html point to .md files instead of .html.

For example
Applicative functors are a simple extension of the [Apply functor](http://non.github.io/cats//tut/apply.md) which adds a single method, pure:

fthomas:


stew:
:+1:

Issue #155, opened 2015-02-09
Have Travis CI push doc pages?


ceedubs:
Thanks to the excellent work of @stew and @tpolecat, we now have fancy documentation at http://non.github.io/cats/. It may be tough for us to remember to keep it up to date. Would it be possible/desired to have the Travis CI job for the master branch automatically publish those pages?

I'm not sure if this is tough from a permissions perspective. Also I don't know if there might be an issue with a really noisy git history for the pages that gets bulky when you pull the project?

travisbrown:
I think it's perfectly reasonable to purge the gh-pages history occasionally if it gets too bulky, and it's not too hard to set up Travis to publish the site automatically.

The only concern I'd have about this is that when you have Travis doing something that relies on the central repo's `GH_TOKEN`, that means any builds on forks are going to fail. I'm sure there's some way to script your way out of this, but I've run into the issue on other projects, and if there's a nice solution I haven't seen it.

non:
It sounds like @InTheNow is going to look into this.

@travisbrown do you feel strongly that we shouldn't set it up from Travis?

travisbrown:
No, going ahead sounds good to me!—at a glance it looks like it shouldn't be too hard to set this up with `deploy` or `after_success` to get around my concern.

rossabaker:
http4s publishes from Travis.  We call this [bash script](https://github.com/http4s/http4s/blob/v0.7.0/bin/travis-publish) from after_success.  I don't think that's a drop-in solution for cats, but it's a starting point.

InTheNow:
@rossabaker thanks for the head start link and @travisbrown for the tip :)

rossabaker:
The trickiest thing I encountered implementing this is to not prune docs from prior releases.  Help may be on the way with sbt/sbt-ghpages#10, but until then, I had to resort to [copying code out of the plugin](https://github.com/http4s/http4s/blob/059e0381628eb205b2c440c9146955582374c69a/project/Http4sGhPages.scala).

The API version `TaskKey[(Int, Int)]`, which reflects the current version minus the patch number.  I use that to specify the api subdirectory to git clean, and leave the remainder of api unscathed.  (This is also a useful key should we want to get into the business of MiMa checking for binary compatibility.)

PR #450, opened 2015-08-15
I hit these problems trying to get started with cats during Stew's SBTB talk

SethTisue:
please sanity-check the `catsJVM/publishLocal` part, I think it's right,
but I did a bunch of fumbling around before hitting on it and it's possible
it isn't enough by itself

### DIFF
--- a/_tut/freemonad.md
+++ b/_tut/freemonad.md
@@ -101,7 +101,7 @@ provide a `Functor` instance for `KVStoreA[_]`.
 ### Import Free in your `build.sbt`
 
 ```scala
-libraryDependencies += "cats" %% "cats-free" % "0.1.0-SNAPSHOT"
+libraryDependencies += "cats" %% "cats-free" % "0.1.3-SNAPSHOT"
 ```
 
 ### Free your ADT
--- a/index.md
+++ b/index.md
@@ -22,7 +22,7 @@ Cats has not yet published artifacts, so in order to use Cats you will have to g
 
 Then in your project, add to your build.sbt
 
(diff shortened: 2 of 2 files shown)


SethTisue:
oh, hmm, this should maybe not be accepted. in the readme on the master branch, it recommends using "0.1.2" rather than building from source and doing publishLocal.

non:
I'm going to close this, since it seems like the README is currently providing correct instructions. Does that make sense @SethTisue?

SethTisue:
But http://non.github.io/cats/ is currently providing incorrect instructions. And there's a meta-issue, that there are two different sets of getting started instructions at all. I'm just not sure what the right fix is, from the maintainers' perspective.

adelbertc:
Yeah the site is outdated, we haven't pushed a new version yet. The source is up-to-date though: https://github.com/non/cats/blob/master/docs/src/site/index.md

fthomas:
Am I alone in thinking that the Free Monad tut should not contain a Cats version number? Unless we have some automatism that updates the version number prior to a release, it probably will be out of date most of the time.

adelbertc:
Agreed with @fthomas

ceedubs:
@SethTisue thank you for creating this. Getting started should be very straightforward and easy, so issues like this are of great importance. It sounds like your code changes are no longer needed, but I didn't want to close your PR until I was sure we had the meta issues you brought up tracked.

There is #155 for the docs not being automatically published.

I just created #479 for the duplication between the readme and the doc page.

I think those two cover the current issues described in this PR, so I'm going to close it. Feel free to reopen it if you think there's more to it.

Thanks again!

PR #451, opened 2015-08-16
Xor tut

adelbertc:


non:
Hey @adelbertc this looks great. As usual I have some minor copy-editing things I noticed.

docs/src/main/tut/xor.md
```diff
@@ -7,3 +7,338 @@ scaladoc: "#cats.data.Xor"
 ---
 # Xor
 
+In day to day programming, it is fairly common to find ourselves writing functions that
```
non: I think we write this as "day-to-day"

docs/src/main/tut/xor.md
```diff
@@ -7,3 +7,338 @@ scaladoc: "#cats.data.Xor"
...
+can fail. For instance, querying a service may result in a connection issue, or some
+unexpected JSON response.
+
+To communicate these errors it has become common practice to throw exceptions. However,
+exceptions are not tracked in any way, shape, or form by the Scala compiler. To see
+if a function throws an exception, and moreover what kind of exception it throws, we
```
non: Maybe simplify this a bit?

"To see what kind of exceptions (if any) a function may throw, ..."

docs/src/main/tut/xor.md
```diff
@@ -7,3 +7,338 @@ scaladoc: "#cats.data.Xor"
...
+the constituent functions can be thrown. Moreover, they may throw the same kind of exception
+(e.g. `IllegalArgumentException`) and thus it gets tricky tracking exactly where that
+exception came from.
+
+How then do we communicate an error? By making it explicit in the data type we return.
+
+## Xor
+
+### Why not `Either`
+`Xor` is very similar to `scala.util.Either` - in fact, they are isomorphic.
```
non: Maybe a quick gloss on what this means?

"... in fact they are *isomorphic* (that is, any `Either` value can be rewritten as an `Xor` value, and vice versa)."

docs/src/main/tut/xor.md
```diff
@@ -7,3 +7,338 @@ scaladoc: "#cats.data.Xor"
...
+object Xor {
+  final case class Left[+A](a: A) extends Xor[A, Nothing]
+  final case class Right[+B](b: B) extends Xor[Nothing, B]
+}
+```
+
+Just like `Either`, it has two type parameters. Instances of `Xor` either hold a value
+of one type parameter, or the other. Why then does it exist at all?
+
+Taking a look at `Either`, we notice immediately there is no `flatMap` or `map` method on
```
non: Maybe "...we notice it lacks `map` and `flatMap` methods." ?

docs/src/main/tut/xor.md
```diff
@@ -7,3 +7,338 @@ scaladoc: "#cats.data.Xor"
...
+it. Instead, we must call `Either#right` (or `Either#left`) and then call `flatMap` or `map`
+on the result of that. The result of calling `Either#right` is a `RightProjection`, the name
+indicating what direction the `Either` is "biased" in. `RightProjection` is "right-biased", so
+calling `flatMap` or `map` on a `RightProjection` acts on the right type-parameter.
```
non: Rather than immediately introducing the word bias, it might be useful to explain the problem directly, and then introduce the term bias to summarize it.

"In order to map over an `Either[A, B]` value, we have to state which side we want to map over. For example, if we want to map `Either[A, B]` to `Either[A, B]` we would need to map over the right side. This can be accomplished by using the `.right` method, which returns a `RightProjection` instance.

...

We say that `RightProjection` is right-biased, since its `.map` and `.flatMap` operations act on the value on the right side."

Does this make sense? Maybe it's too verbose? I just feel like we need to provide a bit more context around the term.

docs/src/main/tut/xor.md
```diff
@@ -7,3 +7,338 @@ scaladoc: "#cats.data.Xor"
...
+val e2: Either[String, Int] = Left("hello")
+e2.right.map(_ + 1)
+```
+
+Note the return types are themselves back to `Either`, so if we want to make more calls to
+`flatMap` or `map` then we again must call `right` or `left`.
+
+More often than not we want to just bias towards one side and call it a day - by convention,
+the right side is most often chosen. This is the primary difference between `Xor` and `Either` -
+it is right-biased by default. `Xor` also has some more convenient methods on it, but the most
```
non: Maybe instead say "`Xor` is right-biased". I think the referent of "it" is possibly confusing, and I would leave out the "by default" since it might also be a bit misleading (it might imply that bias is a matter of configuration).

docs/src/main/tut/xor.md
```diff
@@ -7,3 +7,338 @@ scaladoc: "#cats.data.Xor"
...
+import cats.data.Xor
+
+val xor1: Xor[String, Int] = Xor.right(5)
+xor1.map(_ + 1)
+
+val xor2: Xor[String, Int] = Xor.left("hello")
+xor2.map(_ + 1)
+```
+
+Because `Xor` is right-biased by default, it is easy to define a `Monad` instance for it (you
```
non: s/easy/possible/

I think it's good to try to stay away from saying the task is "easy" (easy to whom?) and also, the key is that its bias makes the behavior of a monad instance constant with the behavior of the type itself (unlike `Either` whose monad would introduce bias that is only used in a generic context).

docs/src/main/tut/xor.md
```diff
@@ -7,3 +7,338 @@ scaladoc: "#cats.data.Xor"
...
+    else 1.0 / i
+
+  def stringify(d: Double): String = d.toString
+}
+```
+
+Instead, let's make the fact that some of our functions can fail explicit in the return type.
+
+```tut
+object Xory {
```
non: Really nit-picky suggestion, use `ExceptionStyle` and `XorStyle` instead. I worry that the `Xory` name might not obviously read as you intend, i.e. "xor-y" versus "exception-y"

docs/src/main/tut/xor.md
```diff
@@ -7,3 +7,338 @@ scaladoc: "#cats.data.Xor"
...
+  case Xor.Right(s) => println(s"Got reciprocal: ${s}")
+}
+```
+
+Not bad - if we leave out any of those clauses the compiler will yell at us, as it should. However,
+note the `Xor.Left(_)` clause - the compiler will complain if we leave that out because it knows
+that given the type `Xor[Exception, String]`, there can be inhabitants of `Xor.Left` that are not
+`NumberFormatException` or `IllegalArgumentException`. However, we "know" by inspection of the source
+that those will be the only exceptions thrown, so it seems strange to have to account for other exceptions.
+This implies that our choice of type is not strong enough.
```
non: Maybe rephrase as "This implies that we could do better." ?

docs/src/main/tut/xor.md
```diff
@@ -7,3 +7,338 @@ scaladoc: "#cats.data.Xor"
...
+  def stringify(d: Double): String = d.toString
+
+  def magic(s: String): Xor[Error, String] =
+    parse(s).flatMap(reciprocal).map(stringify)
+}
+```
+
+For our little module, we enumerate any and all errors that can occur. Then, instead of using
+exception classes as error values, we use one of the enumerated cases. Now when we pattern
+match, we get much nicer matching.
```
non: Maybe just another quick aside that since `Error` is sealed, no third-party can add additional subtypes which we might fail to handle. You mentioned sealed ADTs earlier, so maybe this is redundant, but it seems like an important point.

non:
Again, I just want to echo that this is a really great tutorial. :100:

adelbertc:
Thanks for the careful review @non ! 

Question: I'm thinking about lumping `XorT` into the same tutorial as `Xor`, I'd imagine a separate `XorT` tutorial would involve a lot of the same things as that of `Xor`, with the addition of mentioning that there's an `F[_]` to take care of with the number things you can do varying depending on what constraints exist on `F[_]`.

non:
@adelbertc That makes a lot of sense. I'd be happy to group them together if that makes sense.

That said, I'm also happy to merge this now and then add `XorT` in the future (either to this tutorial or its own).

non:
:+1: I think this is ready to merge.

adelbertc:
@non Let's merge now and I'll add `XorT` later (unless someone beats me to it :-) )

PR #454, opened 2015-08-16
Improves miscellaneous items in documentation

feynmanliang:
Documentation improvements I made while going through cats' documentation for the first time. 

 * Adds many markup changes for inline code and references to other sections
 * Revises explanation text to be more accessible for newcomers (hopefully!)
 * Modifies explanation and code example comparing `Semigroup` vs `SemigroupK`

### DIFF
--- a/docs/src/main/tut/apply.md
+++ b/docs/src/main/tut/apply.md
@@ -7,7 +7,7 @@ scaladoc: "#cats.Apply"
 ---
 # Apply
 
-`Apply` extends the [`Functor`](functor.html) type class (which features the familiar `map`
+`Apply` extends the [Functor](functor.html) type class (which features the familiar `map`
 function) with a new function `ap`. The `ap` function is similar to `map`
 in that we are transforming a value in a context (a context being the `F` in `F[A]`;
 a context can be `Option`, `List` or `Future` for example).
--- a/docs/src/main/tut/apply.md
+++ b/docs/src/main/tut/apply.md
@@ -7,7 +7,7 @@ scaladoc: "#cats.Apply"
 ---
 # Apply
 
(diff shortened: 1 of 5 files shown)


non:
Thanks! These seem like great changes. :+1:

non:
(I'm leaving this open for now in case others have suggestions, but IMO it is ready to be merged.)

docs/src/main/tut/apply.md
```diff
@@ -7,7 +7,7 @@ scaladoc: "#cats.Apply"
 ---
 # Apply
 
-`Apply` extends the [`Functor`](functor.html) type class (which features the familiar `map`
+`Apply` extends the [Functor](functor.html) type class (which features the familiar `map`
```
fthomas: We just merged #459 which added the backticks to links. We should adopt one style and use that throughout the docs. What is the rationale for removing the backticks here?

fthomas:
LGTM, thanks!

I left a comment about using backticks in links. It is ok for me to address this in another PR.

docs/src/main/tut/apply.md
feynmanliang: I wanted everything to be consistent and the predominant style seemed to be excluding backticks within links. It is also less obvious that the generated doc is a link since the code's grey text-color style takes precedence over the link's blue text-color style.

Given #459 merged, I am happy to revise this PR to make consistent.

feynmanliang:
@fthomas Merged master to incorporate #459 and went through the docs to make sure code inside links all have backticks.

adelbertc:
:+1:

fthomas:
Thanks for the updates, @feynmanliang. Wrt to the color of links, maybe we should change the CSS such that the color of links has precedence over the color of "backticked" code.

PR #455, opened 2015-08-17
Add Stream, StreamT, Nondeterminism, etc.

non:
This work is a start toward fleshing out some important
functionality that is missing in Cats right now.

Here's what this commit does:

 * Adds speculative 'cats-task' subproject
 * Adds `cats.data.Stream[A]` (concrete stream type)
 * Adds `cats.data.StreamT[F, A]` (monad transformer)
 * Adds `cats.task.Nondeterminism[F]` (nondeterminism monad)
 * Adds `Nondeterminism[Future]` constructor
 * Adds `.pointEval(a)` to `Applicative[F]`
 * Adds `.memoize` method to `Eval[A]`
 * Renames optionT.scala to OptionT.scala
 * Adds `CatsProps` for ScalaTest/ScalaCheck integration

Here is a short list of known issues, or things we need to figure out:

 * Audit existing code to use `.pointEval` where appropriate
 * Autogenerate arity functions for `Nondeterminism[F]`
 * Consider adding more methods to `Nondeterminism[F]`
 * Add tests for `Eval[A]#memoize`
 * Improve `Stream[A]`/`StreamT[F, A]` docs
 * Avoid `cats.std.future._` / `cats.task.std.future._` clashes
 * Flesh out `cats.task` package with more types


non:
This PR should definitely not be merged as-is. I just wanted to give folks a chance to see what I had been working on, and weigh-in if this is a good decision or not.

Since I ended up basing `Nondeterminism[F]` on `StreamT[F, A]` I decided to put them in the same PR. However, I'd be happy to submit a PR that just included `Stream[A]` and `StreamT[F, A]` if folks would prefer to be a bit more atomic about it.

tpolecat:
Wow that didn't take long.

core/src/main/scala/cats/Applicative.scala
```diff
@@ -13,13 +13,21 @@ import simulacrum.typeclass
...
    * Applicative[Option].pure(10) = Some(10)
    */
   def pure[A](x: A): F[A]
 
+  /**
+   * `pureEval` lifts any value into the Applicative Functor.
+   *
+   * This variant supports optional laziness.
+   */
+  def pureEval[A](x: Eval[A]): F[A] = pure(x.value)
```
adelbertc: Making the default delegate to `pure` would negate any laziness properties of the `Eval` instance we pass in right? Oh - the intention would be to override as necessary, say when defining it for a `Task`-y thing?

core/src/main/scala/cats/data/Stream.scala
```diff
@@ -0,0 +1,713 @@
+package cats
+package data
+
+import cats.syntax.order._
+import scala.reflect.ClassTag
+
+import scala.annotation.tailrec
+import scala.collection.mutable
+
+sealed abstract class Stream[A] { lhs =>
```
adelbertc: Do we want to possible name this something else to avoid clashing with the std lib, which I think has `Stream` imported by default?

core/src/main/scala/cats/Applicative.scala
non: Right exactly. Particular monad instances could do something different here.

core/src/main/scala/cats/data/Stream.scala
non: Yeah, I'm definitely open to it if we can think of something good. I'd like to ensure that if we choose `Xyz` we also use `XyzT` for the transformer.

xuwei-k:
https://github.com/xuwei-k/Cats-StreamT/commit/0f960476def03cf6f9a67

StreamT does not satisfy laws

non:
Aha! Thanks for catching that. I'll fix it.

InTheNow:
`taskJS/test` fails for the reasons that #458 fails - eg `Promise.tryAwait`

task/src/main/scala/cats/task/Nondeterminism.scala
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * Nondeterministic monad.
+ */
+@typeclass(excludeParents=List("NondeterminismFunctions"))
+trait Nondeterminism[F[_]] extends Monad[F] with NondeterminismFunctions[F] {
+
+  type PF[-A, +B] = PartialFunction[A, B]
+
+  def arrange[A](fas: List[F[A]]): StreamT[F, A]
```
julien-truffaut: can we generalise `List` ? Maybe `Foldable`?

non: Sure, `[L: Foldable]` definitely makes sense here.

core/src/main/scala/cats/data/Streaming.scala
```diff
@@ -0,0 +1,789 @@
...
+import scala.collection.mutable
+
+sealed abstract class Streaming[A] { lhs =>
+
+  import Streaming.{Empty, Next, This}
+
+  /**
+   * The stream's catamorphism.
+   *
+   * This method allows the stream to be transformed to an abtirary
```
adelbertc: Editing mishap?

non:
Just to comment here: there is an additional issue which @xuwei-k's test caught -- `flatMap` is created an extra `Next(Empty())` node in place of `Empty()`. Structurally this means the results differ, though they ultimately contain the same elements. I'm going to revise the design to remove this problem.

core/src/main/scala/cats/data/Streaming.scala
```diff
@@ -0,0 +1,789 @@
...
+   *  - Next(tail): a deferred stream.
+   *
+   * This represents a lazy, possibly infinite stream of values.
+   * Eval[_] is used to represent possible laziness (via Now, Later,
+   * and Always). The head of `This` is eager -- a lazy head can be
+   * represented using `Next(Always(...))` or `Next(Later(...))`.
+   */
+  case class Empty[A]() extends Streaming[A]
+  case class Next[A](next: Eval[Streaming[A]]) extends Streaming[A]
+  case class This[A](a: A, tail: Eval[Streaming[A]]) extends Streaming[A]
```
ceedubs: I think all of these case classes should be `final`.

core/src/main/scala/cats/data/StreamingT.scala
```diff
@@ -0,0 +1,320 @@
...
+   *  - Next(tail): a deferred stream.
+   *
+   * This represents a lazy, possibly infinite stream of values.
+   * Eval[_] is used to represent possible laziness (via Now, Later,
+   * and Always). The head of `This` is eager -- a lazy head can be
+   * represented using `Next(Always(...))` or `Next(Later(...))`.
+   */
+  case class Empty[F[_], A]() extends StreamingT[F, A]
+  case class Next[F[_], A](next: F[StreamingT[F, A]]) extends StreamingT[F, A]
+  case class This[F[_], A](a: A, tail: F[StreamingT[F, A]]) extends StreamingT[F, A]
```
ceedubs: These classes should also be `final`.

adelbertc:
Probably worth putting a note somewhere discussing what `Streaming` can do that `Stream` cannot - it's a question I got a couple times when my colleagues saw me using `scalaz.EphemeralStream`. Perhaps a brief comment in the Scaladoc and later on we can write a more in-depth one as a tut on the website or on the Typelevel blog?

xuwei-k:
Maybe, still does not satisfy monad laws (associativity law)

https://github.com/xuwei-k/Cats-StreamT/commit/17bcf4f8910e77c2fb1264077dfdeae7191fdf8b
https://travis-ci.org/xuwei-k/Cats-StreamT/builds/76778801#L347

xuwei-k:
counter example https://github.com/xuwei-k/cats/commit/2f2dfcac37ecb3da73dd9

core/src/main/scala/cats/data/Streaming.scala
```diff
@@ -0,0 +1,829 @@
...
+    this match {
+      case Empty() => Next(rhs)
+      case Next(lt) => Next(lt.map(_ concat rhs))
+      case This(a, lt) => This(a, lt.map(_ concat rhs))
+    }
+
+  /**
+   * Lazily zip two streams together.
+   *
+   * The lenght of the result will be the shorter of the two
```
coltfred: Typo in `length`

non:
@xuwei-k Thanks for all your help. With these changes your counter-example passes, and the scalaprops tests passed for me as well.

All: I haven't sealed the `Streaming[A]` ADT yet, since pattern matching on that may be useful to implement tail-recursive methods. However, since in this case it's possible to construct differently-shaped ADTs which are equivalent, I would be happy to seal those to preserve referential transparency there too.

xuwei-k:
https://github.com/xuwei-k/cats/commit/fc088061a4ee7fd3816794d192

non:
@xuwei-k Thanks, I'll pull in that commit and take a look. If you needed proof that Scalaprops is way better than the currently released ScalaCheck I think you have it! :wink:

non:
I see what the problem is -- I defined `Eq[StreamT[F, A]]` in a very brittle way, and in this case it was Scalaprops Shapeless-based generators that did a better job than the one I wrote.

I've fixed the bug (and confirmed that Scalaprops passes on my branch). Now I'm going to try to improve my own generator and see if I can get it failing with Cats' own tests.

Thanks so much for your help @xuwei-k!

core/src/main/scala/cats/data/Streaming.scala
```diff
@@ -0,0 +1,953 @@
...
+  def unfold[A](o: Option[A])(f: A => Option[A]): Streaming[A] =
+    o match {
+      case None => Empty()
+      case Some(a) => This(a, Always(unfold(f(a))(f)))
+    }
+
+  /**
+   * An empty loop, will wait forever if evaluated.
+   */
+  def godot: Streaming[Nothing] =
```
ceedubs: Is there a use-case for this?

core/src/main/scala/cats/data/package.scala
```diff
@@ -1,9 +1,11 @@
 package cats
 
+import scala.collection.immutable.{Stream => SStream}
```
ceedubs: This seems unnecessary since the cats version is called `Streaming` as opposed to `Stream`. I'm wondering if this is just here because you might have started out calling the cats version `Stream`?

laws/shared/src/main/scala/cats/laws/discipline/Arbitrary.scala
```diff
@@ -44,6 +44,16 @@ object arbitrary {
       A.arbitrary.map(a => Eval.later(a)),
       A.arbitrary.map(a => Eval.always(a))))
 
+  import cats.data.{Streaming, StreamingT}
+
+  implicit def streamingArbitrary[A](implicit A: Arbitrary[A]): Arbitrary[Streaming[A]] =
+    Arbitrary(Gen.listOf(A.arbitrary).map(Streaming.fromList))
```
ceedubs: It would be nice to generate some lazy values too, but with the way scalacheck implements `() => A`, I don't know that we could do anything useful here.

ceedubs:
I have heard at least one (but I think multiple) scalaz contributors say that they think `Nondeterminism` was kind of a failed experiment. I'm not sure I know the reason they say that, but it makes me question whether we want to introduce it to cats. @tpolecat were you a person who said that?

I know that my experience with `Nondeterminism` is that every time I've used it I've had a very specific type in mind (Task), so it didn't really feel like an abstraction. And sometimes I've been tempted to fake a `Nondeterminism` instance for a deterministic constructor like `Id` to reuse code. This makes me wonder if the functionality provided by `Nondeterminism` could be achieved through something like instances for a `Par` (parallel) tag or something. For example if we have a value class `Par`, can we just have an `[a]Applicative[Par[Task[A]]]` handle the nondeterministic behavior?


This PR looks great to me (though there are still some pending inline comments). I just thought I would bring `Nondeterminism` up for discussion.

core/src/main/scala/cats/Eval.scala
```diff
@@ -100,7 +109,9 @@ sealed abstract class Eval[A] { self =>
  * This type should be used when an A value is already in hand, or
  * when the computation to produce an A value is pure and very fast.
  */
-case class Now[A](value: A) extends Eval[A]
+case class Now[A](value: A) extends Eval[A] {
```
ceedubs: Not part of your PR, but I just noticed that the `Eval` constructors should probably be final.

core/src/main/scala/cats/data/Streaming.scala
non: Probably not. We should probably remove this, I was having a bit too much fun! :cat:

core/src/main/scala/cats/data/package.scala
non: Yes, indeed. I'll revert this. Good eye!

rossabaker:
I hereby cosign [@ceedbubs' last comment](https://github.com/non/cats/pull/455#issuecomment-134567505) with citations for the claim of Nondeterminism's disavowal:
* https://github.com/scalaz/scalaz-stream/issues/321#issuecomment-76804744
* https://github.com/scalaz/scalaz-stream/issues/383#issuecomment-114976484

laws/shared/src/main/scala/cats/laws/discipline/Arbitrary.scala
```diff
@@ -52,4 +52,27 @@ object arbitrary {
...
+  //   const(StreamingT.empty[F, A]))
+  //
+  // Howver, getting this right with Scalacheck (and avoiding SOEs) is
+  // somewhat fiddly, so this will have to do for now.
+  //
+  // The max possible size of a StreamingT instance (n) will result in
+  // instances of up to n^3 in length when testing flatMap
+  // composition. The current value (8) could result in streams of up
+  // to 512 elements in length. Thus, since F may not be stack-safe,
+  // we want to keep n relatively small.
```
xuwei-k: workaround in scalaz https://github.com/scalaz/scalaz/commit/450ed685ac52ca6fe63

core/src/main/scala/cats/data/Streaming.scala
ceedubs: :)

xuwei-k:
https://github.com/xuwei-k/cats/commit/70f46ed60a5efa5c267b5d

adelbertc:
Seems `Nondeterminism` is controversial so in the interest of moving other awesomeness (`Streaming`) forward, perhaps split it out? It also seems to be related to `Task`-y things which I believe @stew is working on, so splitting it out could open the doors for pushing `Task` (or `Comp` I think he's calling it) and talking about how it relates.

non:
Great. I will do that later tonight. I'll also try to run Scalaprops a ton of times and see if I can generate any counter examples to our laws.

(I'm also eager to see if @xuwei-k will demolish my most recent effort as quickly as all the others! :wink:)

laws/shared/src/main/scala/cats/laws/discipline/Arbitrary.scala
```diff
@@ -52,4 +52,27 @@ object arbitrary {
...
+    Arbitrary(Gen.listOf(A.arbitrary).map(Streaming.fromList(_)))
+
+  // TODO: it would be better to do this recursively, i.e. more like:
+  //
+  // Gen.oneOf(
+  //   for { a <- arbitrary[A]; s <- arbitrary[F[StreamingT[F, A]]] } yield cons(a, s),
+  //   for { s <- arbitrary[F[StreamingT[F, A]]] } yield wait(s),
+  //   const(StreamingT.empty[F, A]))
+  //
+  // Howver, getting this right with Scalacheck (and avoiding SOEs) is
```
ceedubs: minor typo (However)

ceedubs:
I kicked off a new Travis build because the 2.11 run failed on Scalacheck data generation, which I assume was a fluke.

https://github.com/non/cats/pull/455/files#r37494538 hasn't been addressed yet, but other than that, this looks good to me if @xuwei-k can't break it :). Thanks @non!

non:
One final idea: I was thinking of renaming the ADT nodes from `Empty`, `This`, and `Next` into `Empty`, `Wait`, and `Cons`. What do you all think?

The nice thing is that those names fit the methods used to construct them on the companion, and may be a bit more evocative than the current choices.

ceedubs:
> I was thinking of renaming the ADT nodes from Empty, This, and Next into Empty, Wait, and Cons

I'm assuming that those two lists don't line up and that you meant `This` => `Cons` and `Next` => `Wait`, right? If so, I think that would be a good change.

core/src/main/scala/cats/data/Streaming.scala
```diff
@@ -0,0 +1,947 @@
...
+ * thought of as a collection, with two key differences:
+ *
+ *  1. It may be infinite; it does not necessarily have a finite
+ *     length. For this reason, there is no `.length` method.
+ *
+ *  2. It may be lazy. In other words, the entire stream may not be in
+ *     memory. In this case, each "step" of the stream has
+ *     instructions for producing the next step.
+ *
+ * Streams are not necessarily lazy: they use `Eval[Stream[A]]` to
```
fthomas: Here and in following sentences should `Stream[A]` be `Streaming[A]`?

core/src/main/scala/cats/data/Streaming.scala
```diff
@@ -0,0 +1,947 @@
...
+  /**
+   * Zip the items of the stream with their position.
+   *
+   * Indices start at 0, so
+   *
+   *   Streaming(x, y, z).zipWithIndex
+   *
+   * lazily produces:
+   *
+   *   Streaing((x, 0), (y, 1), (z, 2))
```
fthomas: s/Streaing/Streaming/

core/src/main/scala/cats/data/Streaming.scala
non: Yes, good catch.

non:
This last two commits handle the ADT renaming, and also fix the two typos that @fthomas noticed.

non:
One final possible issue is that `StreamingT[F, A]` is not necessarily  stack-safe (unlike `Streaming[A]`, which is). I tried to do a good job of documenting that `Streaming[F, A]` is dependent on `F`s stack safety.

An alternate strategy would be to trampoline `StreamingT` as well. This would require major changes to the existing code, and might do redundant trampolining when `F` is also trampolined. However, it would make the type safer to work with.

What do you all think?

non:
(Of course, we could always merge this now and consider trampolining in the future.)

ceedubs:
I think it makes sense to depend on `F`'s stack safety. It would be an unnecessary performance hit to double-trampoline if `F` is already stack-safe. People can wrap `Trampoline` around `F` if necessary. The current behavior aligns with `StateT`, which depends on `F` being stack-safe (`State` is an alias for `StateT` where `F` is `Trampoline`).

ceedubs:
:+1:

adelbertc:
Agree w/ @ceedubs , I think it's reasonable to expect `StreamingT` to depend on `F`'s stack safety, same as `StateT` (and `Kleisli` I think).

Awesome PR, :+1: from me. Since it's pretty big I'll let someone else merge for good measure :-)

core/src/main/scala/cats/data/Streaming.scala
```diff
@@ -0,0 +1,947 @@
...
+   * takes only those elements that fulfill the predicate `f`. Once an
+   * element is found which does not fulfill the predicate, no further
+   * elements will be returned.
+   *
+   * If all elements satisfy `f`, the current stream will be returned.
+   * If no elements satisfy `f`, an empty stream will be returned.
+   *
+   * For example:
+   *
+   *   Stream(1, 2, 3, 4, 5, 6, 7).takeWhile(n => n != 4)
```
fthomas: There are still some references to `Stream` instead of `Streaming` in the ScalaDocs (see also `dropWhile` and the `syntax` object). Some day we should start using [sbt-doctest](https://github.com/tkawachi/sbt-doctest). :-)

non:
Awesome, given the two sign-offs I'm going to merge this.

If we find problems (or think of improvements) we can fix those in a future PR.

Thanks to all the reviewers for their careful review, and especially to @xuwei-k for catching numerous mistakes on my part! :cat2:

PR #460, opened 2015-08-17
Monoid docs

philwills:
I can't help feeling there's more profound practical motivations for Monoid, beyond `combineAll`, but here's at least a first attempt at docs for Monoid.

### DIFF
--- a/docs/src/main/tut/monoid.md
+++ b/docs/src/main/tut/monoid.md
@@ -8,12 +8,12 @@ source: "https://github.com/non/algebra/blob/master/core/src/main/scala/algebra/
 # Monoid
 
 `Monoid` extends the [`Semigroup`](semigroup.html) typeclass, adding an `empty` method to semigroup's 
-`combine`. The `empty` method must return a value that combined with any other instance of that type
+`combine`. The `empty` method must return a value that when combined with any other instance of that type
 returns the other instance, i.e.
 
     (combine(x, empty) == combine(empty, x) == x)
     
-For example, if we have `Monoid[String]` with `combine` as string concatenation, then `empty = ""`.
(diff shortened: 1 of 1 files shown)


adelbertc:
One compelling use case I run into quite often is that "monoids beget monoids." If A and B form monoids, then their product (A, B) also forms a monoid. This becomes particularly useful when using something like `Foldable#foldMap`

```scala
scala> import cats._; import cats.implicits._
import cats._
import cats.implicits._

scala> val l = List(1, 2, 3, 4, 5)
l: List[Int] = List(1, 2, 3, 4, 5)

scala> l.foldMap(identity) // use Monoid[Int], adding numbers with 0 as identity
res0: Int = 15

scala> l.foldMap(i => i.toString) // use Monoid[String], concatenating strings with "" as identity
res1: String = 12345

scala> implicit def tupleMonoid[A : Monoid, B : Monoid]: Monoid[(A, B)] =
     |   new Monoid[(A, B)] {
     |   def combine(x: (A, B), y: (A, B)): (A, B) =
     |     (Monoid[A].combine(x._1, y._1), Monoid[B].combine(x._2, y._2))
     |
     |   def empty: (A, B) = (Monoid[A].empty, Monoid[B].empty)
     | }
tupleMonoid: [A, B](implicit evidence$1: cats.Monoid[A], implicit evidence$2: cats.Monoid[B])cats.Monoid[(A, B)]

scala> l.foldMap(i => (i, i.toString)) // do both of the above in one pass, hurrah!
res3: (Int, String) = (15,12345)
```

It looks like we currently don't have the `Monoid` instance for tuples? @non

ceedubs:
It sounds like @adelbertc might have some interesting stuff to add to this in the future, but I think this is a great start. :+1: from me.

non:
:+1:

PR #461, opened 2015-08-17
Remove section about adding cats-free to build.sbt

fthomas:
The reasons are:
- The version number would be mostly out of date because it currently
  needs to be updated manually.
- We can assume that the reader has the "cats" module in their
  libraryDependencies and all examples work with just this.
- Other docs don't have this.

see also #450


adelbertc:
:+1:

non:
:+1:

PR #464, opened 2015-08-19
NaturalTransformation combinators

adelbertc:


core/src/main/scala/cats/arrow/NaturalTransformation.scala
```diff
@@ -1,6 +1,23 @@
 package cats
 package arrow
 
-trait NaturalTransformation[F[_], G[_]] extends Serializable {
+trait NaturalTransformation[F[_], G[_]] extends Serializable { outer =>
```
fthomas: Consistency robot here: I think in other cases the self type is called `self`. Do we want to have this consistent across the code base?

fthomas:
LGTM.

With `id` and `compose`, I'm wondering if we can have a `Category` instance for `NaturalTransformation`.

core/src/main/scala/cats/arrow/NaturalTransformation.scala
```diff
@@ -1,6 +1,23 @@
...
+
+  def compose[E[_]](f: NaturalTransformation[E, F]): NaturalTransformation[E, G] =
+    new NaturalTransformation[E, G] {
+      def apply[A](fa: E[A]): G[A] = outer.apply(f(fa))
+    }
+
+  def andThen[H[_]](f: NaturalTransformation[G, H]): NaturalTransformation[F, H] =
+    f.compose(outer)
+
+  def function[A]: F[A] => G[A] = fa => apply(fa)
```
julien-truffaut: why do we need both `function` and `apply` ?

fthomas: Mayber nicer syntax? `nt.function[A] == nt.apply[A] _`

julien-truffaut: true but it would be good to be consistent, we could have the same discussion regarding parameter lists. For example:
```scala
def foo(a: A)(b: B): C
def foo2(a: A): B => C
def foo3: A => B => C
```

milessabin: It would be nice to be able to abstract over all three of those.

fthomas:
Still looking good after @adelbertc's latest change. :+1:

non:
:+1:

PR #465, opened 2015-08-19
Applicative product

adelbertc:


adelbertc:
A thought: If users want a product of higher arity (`TupleN` for N > 2), things can get messy with nested pairs (e.g. `Applicative[((F[A], G[A]), H[A])]`). We could have it auto-generated by SBT (not sure how to do this..) or write a macro ([Twitter does something similar with Semigroup](https://github.com/twitter/algebird/blob/develop/algebird-core/src/main/scala/com/twitter/algebird/macros/SemigroupMacro.scala)) to do it, or both. Thoughts?

xuwei-k:
This is similar to https://github.com/non/cats/pull/388 `cats.data.Prod` /cc @eed3si9n 
just use `Tuple2` or new case class

adelbertc:
Thanks for the note @xuwei-k ! Closed in favor of https://github.com/non/cats/pull/388/files

PR #466, opened 2015-08-20
Hide data constructors for FreeApplicative, expose only smart constru…

adelbertc:
…ctors

* Also rename run to foldMap to be consistent with Free

### DIFF
--- a/free/src/main/scala/cats/free/FreeApplicative.scala
+++ b/free/src/main/scala/cats/free/FreeApplicative.scala
@@ -48,9 +48,9 @@ sealed abstract class FreeApplicative[F[_], A] { self =>
 object FreeApplicative {
   type FA[F[_], A] = FreeApplicative[F, A]
 
-  private[free] final case class Pure[F[_], A](a: A) extends FA[F, A]
+  final case class Pure[F[_], A](a: A) extends FA[F, A]
 
-  private[free] abstract class Ap[F[_], A] extends FA[F, A] {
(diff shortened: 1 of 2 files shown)


adelbertc:
@mandubian

mandubian:
Seems ok to me!

free/src/main/scala/cats/free/FreeApplicative.scala
```diff
@@ -1,49 +1,56 @@
 package cats
 package free
 
+import cats.arrow.NaturalTransformation
+
 /** Applicative Functor for Free */
 sealed abstract class FreeApplicative[F[_], A] { self =>
-
-  import FreeApplicative.{FA, Pure, Ap}
+  import FreeApplicative.{FA, Pure, Ap, ap => apply, lift}
```
non: I think this example would be a little bit more clear without aliasing `ap` to `apply`.

What do you think? If you disagree I am ok to merge this as-is.

non:
:+1: 

You might want to leave a comment explaining why aliasing `ap` to `apply` is useful. (Namely that the class' method shadows the smart constructor, right?)

adelbertc:
Comment added - I think the comment is enough, naming the smart constructor `ap` is consistent with the operation it's performing.

non:
Great. :+1:

non:
:+1:

ceedubs:
Nice! :+1:

PR #467, opened 2015-08-20
Rework Free to use Coyoneda behind the scenes

adelbertc:
Taken from
https://github.com/scalaz/scalaz/commit/783eb0ed7f6aec56d6245e9b2b9bac97aab74d22#diff-afd5b6310425d773ca3459ebb17ed594

### DIFF
--- a/free/src/main/scala/cats/free/Free.scala
+++ b/free/src/main/scala/cats/free/Free.scala
@@ -3,63 +3,41 @@ package free
 
 import scala.annotation.tailrec
 
-object Free {
+import cats.arrow.NaturalTransformation
 
+object Free {
   /**
    * Return from the computation with the given value.
    */
-  case class Pure[S[_], A](a: A) extends Free[S, A]
+  private[free] final case class Pure[S[_], A](a: A) extends Free[S, A]
 
   /** Suspend the computation with the given suspension. */
-  case class Suspend[S[_], A](a: S[Free[S, A]]) extends Free[S, A]
+  private[free] final case class Suspend[S[_], A](a: S[A]) extends Free[S, A]
 
   /** Call a subroutine and continue with the given function. */
-  sealed abstract class Gosub[S[_], B] extends Free[S, B] {
-    type C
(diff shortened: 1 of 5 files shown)


adelbertc:
Review by @mandubian

adelbertc:
Strange Travis errors, `free` tests seems to be hanging (same behavior when I try locally): https://travis-ci.org/non/cats/jobs/76387437

non:
So, I haven't studied it as closely as I need to, but does this change break the machinery that reassociates left-associated binds to the right?

adelbertc:
@non I believe that's covered in the `foldMap` (and `step`) case.

non:
Aha, I see. I was used to seeing it happen during `flatMap`, but yes, that looks correct.

free/src/main/scala/cats/free/Free.scala
```diff
@@ -3,63 +3,41 @@ package free
...
-   */
-  def runFC[S[_], M[_], A](fa: FreeC[S, A])(f: S ~> M)(implicit M: Monad[M]): M[A] =
-    fa.foldMap[M](new (Coyoneda[S, ?] ~> M) {
-      def apply[B](ca: Coyoneda[S, B]): M[B] = M.map(f(ca.fi))(ca.k)
-    })
+  /** Lift a pure value into Free */
+  def pure[S[_], A](a: A): Free[S, A] = Pure(a)
 
   /**
    * `Free[S, ?]` has a monad if `S` has a `Functor`.
```
fthomas: This is obviously not true anymore.

adelbertc:
......Travis jobs aren't hanging, I introduced an infinite loop somewhere. Looking into it.

mandubian:
I think I need to think a bit more about what Runar has done before answering to this!
It's really interesting actually!

non:
:+1:

mandubian:
as usual Runar's work is great (and yours too) and I say :+1: for this

stew:
:+1:

PR #471, opened 2015-08-20
How about these credentials?

non:


non:
As before, once this is green I'll merge it.

Issue #441, opened 2015-08-11
Spelling: type class vs. typeclass


fthomas:
I've noticed that we are inconsistent wrt to the spelling of "type class" and "typeclass". The README uses  mostly "type class" while docs/src/site/index.md uses "typeclass" exclusively. For the sake of robotic consistency, can we agree on one spelling?

travisbrown:
:+1: for consistency, and :+1: for two words.

There's a discussion of this question at the end of [the second section of the Typeclassopedia](https://wiki.haskell.org/Typeclassopedia), which makes an argument from history for "type class".

non:
I agree with "type class" and also with the push for consistency. Thanks for noticing this.

adelbertc:
:+1: type class

milessabin:
+1

fthomas:
I'm also for "type class". As an additional data point, Wikipedia also uses two words: https://en.wikipedia.org/wiki/Type_class

I'll prepare a PR then later this week.

fthomas:
Closed by #446

PR #473, opened 2015-08-21
Add some basic docs for Foldable

philwills:
@adelbertc's comment in #460 made me realise that Foldable doesn't currently have any docs.

Once again, these are largely based off the Scaladoc, with a few examples added. 

This was based on the branch for #460, but if for some reason, you don't want that and do want this, I'm happy to rebase.

### DIFF
--- a/docs/src/main/tut/foldable.md
+++ b/docs/src/main/tut/foldable.md
@@ -1,0 +1,42 @@
+---
+layout: default
+title:  "Foldable"
+section: "typeclasses"
+source: "https://github.com/non/cats/blob/master/core/src/main/scala/cats/Foldable.scala"
+scaladoc: "#cats.Foldable"
+---
+# Foldable
+
+Foldable typeclass instances can be defined for data structures that can be 
+folded to a summary value.
+
+In the case of a collection (such as `List` or `Set`), these methods will fold 
+together (combine) the values contained in the collection to produce a single 
+result. Most collection types have `foldLeft` methods, which will usually be 
+used by the associated `Foldable[_]` instance.
+
+`Foldable[F]` is implemented in terms of two basic methods:
+
+ - `foldLeft(fa, b)(f)` eagerly folds `fa` from left-to-right.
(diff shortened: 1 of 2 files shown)


adelbertc:
Could you add a section on a the monoid for product/tuple types? Not sure if you should put it on Foldable or Monoid or both (maybe mention it in Monoid and then give an example in Foldable) - its fairly common (esp. in data processing contexts) to be able to compute multiple "summary" values over a large set of data in a single pass (think streaming data).

non:
This is great, thanks!

I have an ask which is smaller than @adelbertc's: we should make sure to point out that the built-in signature of `List#foldRight` is different than Cats'. For now this could just be a sentence but I think it's worth mentioning.

That all said, I'm also happy to merge this now and then update the document in future PRs. :+1:

non:
:+1: 

I'd be happy to merge this now and expand on it later. Thanks @philwills!

fthomas:
Could you replace "typeclass" with "type class" since we have recently agreed on that spelling (#441).

Other than that, LGTMT and I'm also fine with merging it now.

ceedubs:
It looks like this PR includes the same monoid doc that is in #460. Is that an accident?

It would probably be good to add some clarification that "Right" vs "Left" is really about association as opposed to starting at the end vs beginning of the data structure. That is, we can implement `exists`, `forall`, etc in terms of `foldRight` on an infinite stream, and they can still terminate (depending on the values in the stream). However, with documentation I'm usually pretty inclined to get something out there and people can expand upon it over time, so I give a :+1: for this once the monoid thing is sorted out.

philwills:
@ceedubs I've rebased against master, so that this is clean of the changes from #460. 

I'm not sure I understand the distinction around right/left being about association well enough to write about it yet, but might do some more digging/thinking, then give it a go.

ceedubs:
The Travis error is unrelated.

I can't speak for the CSS change, but the rest looks good to me.

PR #474, opened 2015-08-21
Cache SBT assets on Travis

philwills:
Currently the Cats Travis build seems to be resolving and downloading SBT dependencies every run.  This uses the container based infrastructure to avoid having to cache assets between runs and so shorten builds.

Hat-tip to @rtyley for doing the grunt work of working out what these settings were on another project.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,6 @@ env:
   global:
   - secure: Kf44XQFpq2QGe3rn98Dsf5Uz3WXzPDralS54co7sqT5oQGs1mYLYZRYz+I75ZSo5ffZ86H7M+AI9YFofqGwAjBixBbqf1tGkUh3oZp2fN3QfqzazGV3HzC+o41zALG5FL+UBaURev9ChQ5fYeTtFB7YAzejHz4y5E97awk934Rg=
   - secure: QbNAu0jCaKrwjJi7KZtYEBA/pYbTJ91Y1x/eLAJpsamswVOvwnThA/TLYuux+oiZQCiDUpBzP3oxksIrEEUAhl0lMtqRFY3MrcUr+si9NIjX8hmoFwkvZ5o1b7pmLF6Vz3rQeP/EWMLcljLzEwsrRXeK0Ei2E4vFpsg8yz1YXJg=
-sudo: false
 cache:
   directories:
   - $HOME/.sbt/0.13
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ env:
   - secure: QbNAu0jCaKrwjJi7KZtYEBA/pYbTJ91Y1x/eLAJpsamswVOvwnThA/TLYuux+oiZQCiDUpBzP3oxksIrEEUAhl0lMtqRFY3MrcUr+si9NIjX8hmoFwkvZ5o1b7pmLF6Vz3rQeP/EWMLcljLzEwsrRXeK0Ei2E4vFpsg8yz1YXJg=
 cache:
   directories:
-  - $HOME/.sbt/0.13
+  - $HOME/.sbt/0.13/dependency
   - $HOME/.sbt/boot/scala*
   - $HOME/.sbt/launchers
   - $HOME/.ivy2/cache
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,9 +32,9 @@ cache:
   - $HOME/.sbt/0.13
   - $HOME/.sbt/boot/scala*
   - $HOME/.sbt/launchers
-  - $HOME/.ivy2
+  - $HOME/.ivy2/cache
 before_cache:
-  - du -h -d 1 $HOME/.ivy2/
+  - du -h -d 1 $HOME/.ivy2/cache
   - du -h -d 2 $HOME/.sbt/
   - find $HOME/.sbt -name "*.lock" -type f -delete
   - find $HOME/.ivy2/cache -name "ivydata-*.properties" -type f -delete
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,3 +26,16 @@ env:
   global:
   - secure: Kf44XQFpq2QGe3rn98Dsf5Uz3WXzPDralS54co7sqT5oQGs1mYLYZRYz+I75ZSo5ffZ86H7M+AI9YFofqGwAjBixBbqf1tGkUh3oZp2fN3QfqzazGV3HzC+o41zALG5FL+UBaURev9ChQ5fYeTtFB7YAzejHz4y5E97awk934Rg=
   - secure: QbNAu0jCaKrwjJi7KZtYEBA/pYbTJ91Y1x/eLAJpsamswVOvwnThA/TLYuux+oiZQCiDUpBzP3oxksIrEEUAhl0lMtqRFY3MrcUr+si9NIjX8hmoFwkvZ5o1b7pmLF6Vz3rQeP/EWMLcljLzEwsrRXeK0Ei2E4vFpsg8yz1YXJg=
+sudo: false
+cache:
+  directories:
+  - $HOME/.sbt/0.13
+  - $HOME/.sbt/boot/scala*
+  - $HOME/.sbt/cache
+  - $HOME/.sbt/launchers
+  - $HOME/.ivy2
+before_cache:
+  - du -h -d 1 $HOME/.ivy2/
+  - du -h -d 2 $HOME/.sbt/
+  - find $HOME/.sbt -name "*.lock" -type f -delete
+  - find $HOME/.ivy2/cache -name "ivydata-*.properties" -type f -delete
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,6 @@ cache:
   directories:
   - $HOME/.sbt/0.13
   - $HOME/.sbt/boot/scala*
-  - $HOME/.sbt/cache
   - $HOME/.sbt/launchers
   - $HOME/.ivy2
 before_cache:


philwills:
Of course, it does need to go through at least once to have any benefit and currently Travis is timing out before it has anything to cache :-(

.travis.yml
```diff
@@ -24,3 +24,16 @@ env:
   global:
   - secure: "X3x7EJUTc2E0VzAC1f4lzeJ4kcU8NI1quLFk/uigEx5qqranzipqE4wXLXhRcub4akMnBrDdW3kahFAU28cSgeWxpbYC3f26MNlahSpJIrWpwEMSmfSlwJSZCJYP6cbVY7GCNjk2jl+4BdT8JAdM5j1ozWLFV0QAhfvV5GIj8tg="
   - secure: "KMngIPhnOE0WEonCi6ar2Qxxc3D0pe+DRalrveF2mWHuz8Xncphz5f5VhkZmxY/v8nM0EjiaN9L2KpAwiPrO7vIt152ZHYH+Ih6ojcBfdxB0N/F5LsRFWseNdJDkGX04teWjDFWWG1a9+rnbkQZ3S2sAT0QmSZT621ByEm3m438="
+sudo: false
+cache:
+  directories:
+  - $HOME/.sbt/0.13
+  - $HOME/.sbt/boot/scala*
+  - $HOME/.sbt/cache
```
dwijnand: To my knowledge there's no such thing as `~/.sbt/cache`

philwills:
One other note, is that I've got no experience with scalajs builds. I don't think this should have any negative impact on them, but that's supposition at present.

.travis.yml
```diff
@@ -24,3 +24,16 @@ env:
   global:
   - secure: "X3x7EJUTc2E0VzAC1f4lzeJ4kcU8NI1quLFk/uigEx5qqranzipqE4wXLXhRcub4akMnBrDdW3kahFAU28cSgeWxpbYC3f26MNlahSpJIrWpwEMSmfSlwJSZCJYP6cbVY7GCNjk2jl+4BdT8JAdM5j1ozWLFV0QAhfvV5GIj8tg="
   - secure: "KMngIPhnOE0WEonCi6ar2Qxxc3D0pe+DRalrveF2mWHuz8Xncphz5f5VhkZmxY/v8nM0EjiaN9L2KpAwiPrO7vIt152ZHYH+Ih6ojcBfdxB0N/F5LsRFWseNdJDkGX04teWjDFWWG1a9+rnbkQZ3S2sAT0QmSZT621ByEm3m438="
+sudo: false
```
InTheNow: The cats build already uses the container infrastructure so this is redundant: eg see https://travis-ci.org/non/cats/jobs/76398959#L78

```
 78 This job is running on container-based infrastructure,
```

InTheNow:
As a general point, using cached dependencies for a master build is generally considered bad practice. One of the main reasons for a having a "build server" is to ensure clean builds that can't pick up side-effects of other local, historic processes. 

On a local build its OK, or for a temporary hack-fix it's OK too.

InTheNow:
As I realize caching is a tempting feature (they wouldn't have added it otherwise, right?), I'll add a reference to my previous comment, from [Travis-CI Clearing Caches](http://docs.travis-ci.com/user/caching/#Clearing-Caches):

```
Sometimes you spoil your cache by storing bad data in one of the cached directories.
```

So how do you know when/if the cache is spoilt? One doesn't, therefore you have to clear the cache before each run to be sure, and that sort of defeats the point of having a cache in the first place. Expect perhaps....

In the travis build, we call sbt many times. In this case, perhaps having caching, cleared first, would actually speed up the build? But looking at [things not too cache](http://docs.travis-ci.com/user/caching/#Things-not-to-cache) I'm not so sure.

milessabin:
@InTheNow I have caching enabled for all my Travis projects. I _only_ use it for caching upstream binary dependencies, and I really don't see a problem with that which wouldn't equally be a problem with relying on Central.

InTheNow:
@milessabin I see you use `$HOME/.ivy2/cache` rather than  `$HOME/.ivy2` proposed here, that also include  `$HOME/.ivy2/local`.

So yes, just caching upstream binary dependencies may not be an issue (providing you __never__ have SNAPSHOT dependencies), but, as the link says "You possibly won’t see a speedup putting them into the cache."

milessabin:
I cargo-culted my config from @fommil during one of the Travis-repo connectivity screwups last year ... it doesn't seem to have caused me any problems and builds were slightly faster.

rtyley:
See also https://github.com/travis-ci/travis-ci/issues/4556 - time is
wasted doing unnecessary dependency repacked.

InTheNow:
And also see [cache should not be saved on failed builds](https://github.com/travis-ci/travis-ci/issues/4472). 

Just to be clear: @philwills @rtyley @milessabin - I'm not a Travis-CI Guru and have little desire to become one, so I am __really__ appreciative when others, like you, show an interest! And I am for sure not saying that this PR is wrong - just trying to get the best, long-ish term solution.

Special welcome to @rtyley, btw :)

fommil:
I've long since moved to shippable for small projects and self hosted drone for ENSIME. Both awesome and use docker instead of the homebrew Travis provisioning.

InTheNow:
@fommil That's interesting: https://app.shippable.com/ Looks nice.

fommil:
@InTheNow please feel free to use or take inspiration from our docker images https://github.com/ensime/ensime-docker/branches we publish on docker hub.

fommil:
fyi, we host our own https://github.com/drone/drone/ which is by far the best CI tool out there... but their service offering is actually a completely different codebase. The ENSIME build is pretty intensive (heavy I/O for file indexing tests, plus Emacs integration tests) so we find that travis/shippable/other all let us down unless we have our own hardware doing the work.

fommil:
btw, I'm deleting our `shippable-1.0` branch because we don't use it anymore. This is the code we were using on top of the `v1.x` image to preload a bunch of stuff that we find they were installing:

```
################################################
# Shippable
#
# Shippable tries to install these scripts on startup, so we may as
# well add them here since it is our CI platform.
RUN\
  apt-get install -y locales git-core python-pip ssh &&\
  apt-get clean &&\
  pip install virtualenv &&\
  pip install --install-option="--prefix=/root/ve" pika boto glob2 &&\
  echo "en_US.UTF-8 UTF-8" >> /etc/locale.gen &&\
  locale-gen
```

our `master` branch starts from the `v1.x` image and then caches the hell out of ivy so that all PRs are already preloaded with maven central goodness.

fommil:
also, note that we use `openjdk` but you have access to all the SUN / Oracle JDKs in that docker image.

.travis.yml
philwills: Think you're right, will remove.

philwills:
@InTheNow I agree it's a trade off. On some other, smaller, projects I've found it to be the dominant feature of build time, at which point I think it's definitely a feature worth exploiting. 

For the Cats build, it's a relatively small proportion of the build, so I'm much less convinced of it's benefit here, but it does also help with transient problems like we had the other day due to the issues with bintray.

non:
I'm torn. I understand @InTheNow's comments, and agree that using caching will possibly mask problems with resolution which would affect Cats' users. On the other hand, it's really frustrating how flaky Travis' resolvers have been, and so I am tempted to try this out just to see how it works.

InTheNow:
@non I saw your build fail earlier whilst downloading an sbt-plugin and instantly gave up on the theoretical approavh :)

But I would suggest we only cache the ivy2/cache

non:
Alright, @philwills, would you be willing to limit the caching to `~/.ivy2/cache`? If so, I give this +1.

InTheNow:
> .... we only cache the ivy2/cache

Just to clarify that, I mean still cache the `.sbt` directories, but `$HOME/.ivy2` => `$HOME/.ivy2/cache`.

Not only is `.ivy2/local` not required, the builds do actually `publishLocal` so they should not be cached.

InTheNow:
@philwills I had to make a bunch of other changes to `.travis.yml` in #488, but your PR needed rebasing on master before in any case (I did look first).

At the same time, I took the liberty of adding _very_ simple cache support as the travis issues were persisting: I have not added the clever bits of this PR, so I don't think my PR has shadowed this PR

non:
What's the status of this PR? It looked good to me, but it seemed like it also needed minor changes to which directories will be cached (and to resolve conflicts).

philwills:
@non It's on my, I should get round to fixing it list. I'll try and take a look tonight.

philwills:
OK, I've rebased against master and more narrowly defined the caching.

non:
:+1:

.travis.yml
```diff
@@ -26,3 +26,14 @@ env:
   global:
   - secure: Kf44XQFpq2QGe3rn98Dsf5Uz3WXzPDralS54co7sqT5oQGs1mYLYZRYz+I75ZSo5ffZ86H7M+AI9YFofqGwAjBixBbqf1tGkUh3oZp2fN3QfqzazGV3HzC+o41zALG5FL+UBaURev9ChQ5fYeTtFB7YAzejHz4y5E97awk934Rg=
   - secure: QbNAu0jCaKrwjJi7KZtYEBA/pYbTJ91Y1x/eLAJpsamswVOvwnThA/TLYuux+oiZQCiDUpBzP3oxksIrEEUAhl0lMtqRFY3MrcUr+si9NIjX8hmoFwkvZ5o1b7pmLF6Vz3rQeP/EWMLcljLzEwsrRXeK0Ei2E4vFpsg8yz1YXJg=
+cache:
+  directories:
+  - $HOME/.sbt/0.13
```
dwijnand: Anything in particular you're looking to cache here?

My stance is that you should cache as little as possible to avoid false positives, and very carefully start opting into what you want to cache from one CI run to another.

For instance here there are (what I think are) dependency graph results under `~/.sbt/0.13/dependency/` which I'm not sure you want to uncache and recache every run.

non:
:+1:

milessabin:
:+1:

.travis.yml
```diff
@@ -26,3 +26,14 @@ env:
   global:
   - secure: Kf44XQFpq2QGe3rn98Dsf5Uz3WXzPDralS54co7sqT5oQGs1mYLYZRYz+I75ZSo5ffZ86H7M+AI9YFofqGwAjBixBbqf1tGkUh3oZp2fN3QfqzazGV3HzC+o41zALG5FL+UBaURev9ChQ5fYeTtFB7YAzejHz4y5E97awk934Rg=
   - secure: QbNAu0jCaKrwjJi7KZtYEBA/pYbTJ91Y1x/eLAJpsamswVOvwnThA/TLYuux+oiZQCiDUpBzP3oxksIrEEUAhl0lMtqRFY3MrcUr+si9NIjX8hmoFwkvZ5o1b7pmLF6Vz3rQeP/EWMLcljLzEwsrRXeK0Ei2E4vFpsg8yz1YXJg=
+cache:
+  directories:
+  - $HOME/.sbt/0.13/dependency
```
dwijnand: Uh.. I meant the opposite, you probably *don't* want to cache this directory.

My question was "what is it you want to cache in `~/.sbt/0.13`?".

philwills: My understanding was that that held the dependency resolution graphs, which are what I was intending to cache.

Am I confused about what's in there, or is there a reason I'm missing that you wouldn't want to do that?

dwijnand: I'm not sure about them (@eed3si9n might be able to help), but in absence of solid knowledge I'd rather not cache it, and avoid possible false positives in Travis. Which might be a bit paranoid :)

eed3si9n: It should be safe to cache that directory, except it won't be of much use without enabling cached resolution. Given Cats is light on dependency graph, not sure cached resolution would buy much.

philwills: Gah, I had stupidly assumed that cached resolution was being used without checking.

PR #476, opened 2015-08-22
Add Choice type class, closes #463 redux

### LINKED ISSUE #463: add cats.arrow.Choice

adelbertc:
Accidentally merged previous, reverted that one. See https://github.com/non/cats/pull/468 for previous discussion.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,5 +22,5 @@ notifications:
     on_start: false
 env:
   global:
-  - secure: Kf44XQFpq2QGe3rn98Dsf5Uz3WXzPDralS54co7sqT5oQGs1mYLYZRYz+I75ZSo5ffZ86H7M+AI9YFofqGwAjBixBbqf1tGkUh3oZp2fN3QfqzazGV3HzC+o41zALG5FL+UBaURev9ChQ5fYeTtFB7YAzejHz4y5E97awk934Rg=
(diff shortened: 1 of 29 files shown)


julien-truffaut:
`Choice` is actually missing

julien-truffaut:
@fthomas as far as I can see, the only reason why `scalaz.Choice` inherit `Category` and not `Compose` is `codiagonal` https://github.com/scalaz/scalaz/blob/series/7.2.x/core/src/main/scala/scalaz/Choice.scala#L15

fthomas:
Hmm, okay. I looked through our current type classes and noted that we have a similar situation with `MonadFilter`. The `MonadFilterLaws` only make use of `empty` and `flatMap` which suggests that `MonadFilter` should actually be called `FlatMapFilter` and inherit `FlatMap`, but `MonadFilter` has derived functions that are implemented in terms of `flatMap`, `pure`, and `empty` (like `diagonal` in this case).

Under the assumption that we tolerate such artificial dependencies, I'm fine with `Choice` as it is right now.

julien-truffaut:
@adelbertc there is no `Choice.scala` in the changelist

@fthomas good point, I am fine either way: remove `diagonal` and inherit from `Compose` or let it is as it is. We should probably create an issue to discuss this general situation and get feedback from other

fthomas:
@julien-truffaut Sounds good to me. I'm slightly in favor of keeping `diagonal` for now.

non:
Let's leave it as-is (once @adelbertc re-adds `Choice.scala` that is). :+1:

julien-truffaut:
:+1:

PR #480, opened 2015-08-25
Remove Functor[Function0] constraint from state monad

ceedubs:
This now lives within core, so we can use it directly instead of
requiring it as a parameter.


non:
:+1:

adelbertc:
:+1:

PR #481, opened 2015-08-25
add .left and .right Xor syntax

drostron:
Discussion:
https://gitter.im/non/cats?at=55db66fd6ebe0cd61263f916

Is there interest in having the following syntax available?

```scala
scala> import cats.data.Xor.syntax._
import cats.data.Xor.syntax._

scala> val xor1: Xor[String, Int] = 7.right[String]
xor1: cats.data.Xor[String,Int] = Right(7)

scala> val xor2: Xor[String, Int] = "hello 🐈s".left[Int]
xor2: cats.data.Xor[String,Int] = Left(hello 🐈s)
```


travisbrown:
:+1: from me, but I'd vote to move the `Ops` class to `cats.syntax` and to make the operations available via `import cats.syntax.all._`.

milessabin:
What @travisbrown said.

drostron:
Happy to move it. A gitter discussion with @non gave me the impression that a `syntax` object in `Xor` would be [less controversial](https://gitter.im/non/cats?at=55db684f36e894436a9ada1a). Perhaps it isn't controversial to have it in `cats.syntax`?

travisbrown:
@drostron Ah, missed that conversation. I'd personally rather see both in `cats.syntax`, since both `syntax.all._` not actually meaning _all_ and having to guess whether to write `cats.data.Xor.syntax._` or `cats.syntax.xor._` don't seem ideal.

adelbertc:
I like having the a la carte import under `cats.syntax.xor._` for consistency, as well as buffet import under `cats.syntax.all._`

non:
Great. I don't have a problem with this kind of syntax living in `cats.syntax` either, but I remember previously seeing some folks push back against it.

Let's target there for now. I'll move the syntax in my streaming PR there too.

drostron:
Excellent. Pushed a commit with the move to `cats.syntax`.

non:
Really small request: currently the files in `src/main/scala/cats/syntax` are all lowercased to reflect the object file they relate to. Would you mind lowercasing `Xor.scala` as well?

adelbertc:
While you're there do we want to make the class a value class?

drostron:
`Xor.scala` is now lowercase and `XorOps` is a value class.

adelbertc:
:+1: from me once Travis gives the green light

non:
:+1: same

PR #483, opened 2015-08-26
add a touch of OptionSyntax

drostron:
Is there interest in the following syntax?

```scala
scala> none[Int]
res0: Option[Int] = None

scala> 1.some
res1: Option[Int] = Some(1)

scala> none[Int].toRightXor("a")
res2: cats.data.Xor[String,Int] = Left(a)

scala> 2.some.toRightXor("b")
res3: cats.data.Xor[String,Int] = Right(2)

scala> 3.some.toLeftXor("c")
res4: cats.data.Xor[Int,String] = Left(3)

scala> none[List[Int]].orEmpty
res5: List[Int] = List()
```

### DIFF
--- a/core/src/main/scala/cats/syntax/option.scala
+++ b/core/src/main/scala/cats/syntax/option.scala
@@ -6,15 +6,15 @@ import cats.data.Xor
 trait OptionSyntax {
   def none[A] = Option.empty[A]
   implicit def optionIdSyntax[A](a: A): OptionIdOps[A] = new OptionIdOps(a)
-  implicit def optionSyntax[A](a: Option[A]): OptionOps[A] = new OptionOps(a)
+  implicit def optionSyntax[A](oa: Option[A]): OptionOps[A] = new OptionOps(oa)
 }
 
 class OptionIdOps[A](val a: A) extends AnyVal {
   def some: Option[A] = Option(a)
 }
 
-class OptionOps[A](val a: Option[A]) extends AnyVal {
-  def toLeftXor[B](b: => B): A Xor B = a.fold[A Xor B](Xor.Right(b))(Xor.Left(_))
-  def toRightXor[B](b: => B): B Xor A = a.fold[B Xor A](Xor.Left(b))(Xor.Right(_))
-  def orEmpty(implicit A: Monoid[A]): A = a.getOrElse(A.empty)
+class OptionOps[A](val oa: Option[A]) extends AnyVal {
+  def toLeftXor[B](b: => B): A Xor B = oa.fold[A Xor B](Xor.Right(b))(Xor.Left(_))
+  def toRightXor[B](b: => B): B Xor A = oa.fold[B Xor A](Xor.Left(b))(Xor.Right(_))
+  def orEmpty(implicit A: Monoid[A]): A = oa.getOrElse(A.empty)
 }
--- a/core/src/main/scala/cats/syntax/option.scala
+++ b/core/src/main/scala/cats/syntax/option.scala
@@ -1,0 +1,20 @@
+package cats
+package syntax
+
+import cats.data.Xor
+
+trait OptionSyntax {
+  def none[A] = Option.empty[A]
+  implicit def toOptionSyntax[A](a: A): ToOptionOps[A] = new ToOptionOps(a)
+  implicit def optionSyntax[A](a: Option[A]): OptionOps[A] = new OptionOps(a)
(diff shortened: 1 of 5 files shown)


core/src/main/scala/cats/syntax/option.scala
```diff
@@ -0,0 +1,20 @@
...
+
+import cats.data.Xor
+
+trait OptionSyntax {
+  def none[A] = Option.empty[A]
+  implicit def toOptionSyntax[A](a: A): ToOptionOps[A] = new ToOptionOps(a)
+  implicit def optionSyntax[A](a: Option[A]): OptionOps[A] = new OptionOps(a)
+}
+
+class ToOptionOps[A](val a: A) extends AnyVal {
```
drostron: This makes me think [`EitherOps`](https://github.com/non/cats/blob/d128c060b92751179276882aa375096cbad209b2/core/src/main/scala/cats/syntax/xor.scala#L10) should be `ToEitherOps` for consistency.

travisbrown: I think I'd probably prefer `OptionIdOps` and `XorIdOps` (following Scalaz's naming convention) for related operations added to any old type, since it's at least conceivable that these wouldn't be _conversions_. That's not a strong opinion, though, and as long as the value class names are consistent I'm all for adding `none` and `some`.

core/src/main/scala/cats/syntax/option.scala
```diff
@@ -0,0 +1,20 @@
...
+  def some: Option[A] = Option(a)
+}
+
+class OptionOps[A](val a: Option[A]) extends AnyVal {
+  def toLeftXor[B](b: => B): A Xor B = a.fold[A Xor B](Xor.Right(b))(Xor.Left(_))
+  def toRightXor[B](b: => B): B Xor A = a.fold[B Xor A](Xor.Left(b))(Xor.Right(_))
```
travisbrown: If we add `toLeftXor` and `toRightXor` it seems like we'd want a `toXor` for `Either` as well. I'm not as certain we want all of these as I am about `none`, `some`, and `orEmpty`, though.

core/src/main/scala/cats/syntax/option.scala
drostron: The `OptionIdOps` naming convention does makes more sense.

core/src/main/scala/cats/syntax/option.scala
drostron: Is there interest in keeping these sort of syntax additions to a minimum? I can understand that stance. Perhaps I was in the minority by leveraging them in scalaz.

tpolecat: I use scalaz `\/>` all the time so I'm all in favor of `toRightXor`, et al.

travisbrown: I guess my position is that if it helps significantly with type inference, I'm all for it, but if it just lets you write arguments in a different order, I'm not necessarily all for it (but don't care strongly).

`Xor.fromOption(oa, b)` seems just as convenient as `oa.toXorRight(b)`, and a lot less mysterious.

tpolecat: It can be nice in a context where you have several things in a comprehension that you want to lift. 

```scala
for {
  a <- foo(...) \/> "oops foo"
  b <- bar(a)   \/> "oops bar"
} yield whatevs
```

scalaz is full of little things like this and I like them, but as you say it's not a big deal. You could argue that this is the kind of thing that scares people away.

non: What if we put this kind of syntax in a separate package/module? I guess an extra import could be a bit annoying, but maybe it's worth it to keep things a bit more modular/cleaner?

non: For example `import cats.rich.syntax._`?

Alternately, we could create a new package and then  `import fatcats.syntax._`.

travisbrown: If we include these I'd vote to make them available as they are in the current PR—the cats import situation is already a little more complicated than I'd prefer.

non: Great. Let's just include them as-is. :+1:

travisbrown:
:+1: from me once Travis CI says it's okay.

non:
:+1:

PR #485, opened 2015-08-26
Free touch up

adelbertc:
* Have Free extend Product to avoid strange type inference
* FreeApplicative extends Product and Serializable to avoid strange type inference
* Make FreeApplicative.Ap a case class and remove type casing
* Remove FreeC type alias - Free uses Coyoneda behind the scenes so F[_] is unconstrained


non:
:+1:

fthomas:
:+1:

PR #488, opened 2015-08-26
Build tweaks

InTheNow:
Fixes https://github.com/non/cats/issues/155
Fixes https://github.com/non/cats/issues/429

Site push is untested.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ TAGS
 .idea/*
 .idea_modules
 .DS_Store
+.sbtrc
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,20 @@
 language: scala
+
 git:
   depth: 9999
+
 scala:
 - 2.10.5
 - 2.11.7
+
+cache:
+  directories:
+    - $HOME/.ivy2/cache
+    - $HOME/.sbt
+
 script:
-- if [[ "$TRAVIS_PULL_REQUEST" == "false" &&
-        "$TRAVIS_BRANCH" == "master" &&
-        $(cat version.sbt) =~ "-SNAPSHOT"
-  ]]; then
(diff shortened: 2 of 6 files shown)


non:
Looks great to me. :+1:

non:
It seems like this worked! Or at least, it looks like we published a 2.10 snapshot. The 2.11 Travis build failed repeatedly though. Anyway, we'll see if it starts working moving forward.

Thanks! :cat2:

InTheNow:
@non

The 2.11 build is publishing snapshots but is failing on the newly added (and untested) web site generation. To fix this, `git` has to get a password. This is a summary taken from @rossabaker site - he's probably the best to ask for advice on this matter. See [the travis docs section on encripting files](http://docs.travis-ci.com/user/encrypting-files/)

- add another secure variable [eg from http4s](https://github.com/http4s/http4s/blob/v0.7.0/.travis.yml#L7)
- add a travis encryption file [eg `.travis.secrets.enc` from http4s](https://github.com/http4s/http4s/blob/v0.7.0/.travis.secrets.enc)
- add some `before_script` commands  [eg from http4s](https://github.com/http4s/http4s/blob/v0.7.0/.travis.yml#L10-L16)

non:
Here's how I think we should move forward.

1. Temporarily disable the web site generation to get our Travis builds working again.
2. Confirm that everything else seems to work (modulo Travis resolution issues)
3. Start looking at adding authentication and re-enabling the web site generation.

How does that sound? @rossabaker?

InTheNow:
That would make sense to me. ( just knockout [this call](https://github.com/non/cats/blob/master/scripts/travis-publish.sh#L10)).

OTOH - we're so nearly there.... just a bit more pain to go....

@non Your call, I'm easy on this

non:
I opened #491 just because I want to get things into a stable state to proceed from. As soon as we get successful builds with it I will create a branch to get site pushing working -- I just wanted to be sure everything else was OK first.

InTheNow:
Perfect - I added https://github.com/non/cats/pull/490 mainly as a convenient test case, btw.

PR #491, opened 2015-08-27
Temporarily disable site generation to get Travis building.

non:
Fixing these issues will hopefully be easy, but I want to
get everything into a working state first.


ceedubs:
:+1: if/when Travis goes green.

PR #495, opened 2015-08-29
Fix #458: Implement FutureInsantances for Scala-js

### LINKED ISSUE #458: Implement FutureInsantances for Scala-js
Currently, the [`FutureTests`](https://github.com/non/cats/blob/master/tests/jvm/src/test/scala/cats/tests/FutureTests.scala) are for JVM only. This is due to the use of [`Await` in `FutureInsantances`](https://github.com/non/cats/blob/master/core/src/main/scala/cats/std/future.scala#L4) and so even though this compiles in scala-js, it does not actually work.

InTheNow:
Also introduces exported tests


ceedubs:
:+1:

https://gitter.im/non/cats?at=55e1fb611b9798c846ecbc18 has some conversation about this.

ceedubs:
This has merge conflicts.

InTheNow:
this in now waiting on #507, and will need refactoring/re-doing, as in https://github.com/non/alleycats/pull/10#issuecomment-136675398

non:
Now that #507 has landed, what parts of this need redoing? As far as I can tell the issue (here at least) is fixed.

InTheNow:
@non - What still needs doing is the js version (originally in https://github.com/non/alleycats/pull/10) needs adding.

So as you say, I'll close this and add a new PR (sometime...) for the js parts.

PR #497, opened 2015-08-29
add setMonoid

drostron:
[gitter discussion](https://gitter.im/non/cats?at=55e23450fbbdfcbd3b5de06f)

### DIFF
--- a/core/src/main/scala/cats/std/set.scala
+++ b/core/src/main/scala/cats/std/set.scala
@@ -24,5 +24,5 @@ trait SetInstances extends algebra.std.SetInstances {
       override def isEmpty[A](fa: Set[A]): Boolean = fa.isEmpty
     }
 
-    implicit def setMonoidInstance[A]: Monoid[Set[A]] = MonoidK[Set].algebra[A]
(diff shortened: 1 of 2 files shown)


non:
:+1: Thanks!

ceedubs:
It will probably require adding a questionable `Eq` instance for sets, but could we test this against the monoid laws?

drostron:
Sure. I need to familiarize with law checking a bit more first (been itching to do so for awhile so this is a welcomed opportunity). The next few days are filled, but should have some time after that.

drostron:
Pushed a commit adding law checking for `setMonoidInstance`. @ceedubs, I'm curious if this PR introduces a questionable `Eq` instance for sets. If so I'm interested to learn more.

ceedubs:
:+1: looks good to me.

I was surprised to see that `Eq` instance for sets was in scope. It looks like algebra has provided one for quite a while (not introduced with this PR). I suppose it makes sense to have it if you are working with `Set`, since it's based on universal equality anyway.

ceedubs:
The only thing added since @non gave a :+1: is law-checking which I can't imagine he would be opposed to :). I'm going to merge this. Thanks!

PR #498, opened 2015-08-30
start adding some tests to task, 

stew:
obviously this is just scratching the surface of what should be tested


stew:
more to come, obviously, I'm just tossing up what I've got because I'm about to go AFK for the rest of the weekend

PR #499, opened 2015-08-30
TaskApp

adelbertc:


non:
If people are sure we want to use `List[String]` here, then I'm fine with this.

refried:
Nice idea.

ceedubs:
@non is there a type you would prefer to `List[String]` for the args?

non:
I think I was just bike shedding. Let's use `List[String]` and see how it goes.

@adelbertc I am moving `Task.scala` from `core/shared` to `core/jvm` due to its dependence on `java.util.concurrent` classes. Once that is in you'll need to move this as well.

non:
(More generally, there is a big reckoning around `Task` versus JS, which we should figure out sooner than later.)

non:
Alright, can you retarget this code to `core/jvm/src/main/scala/cats`?

Eventually we will probably want to move `Task` and `TaskApp` into a new submodule, but for now this will work.

non:
Also -- just a warning, there is a discussion on Gitter about creating a `cats.jvm` project, where we'd put things that require `Await`, `j.u.c`, etc. So you may also want to wait a bit on this until the dust settles there.

dchenbecker:
Would anyone be opposed to making the types a little richer? E.g.

```
sealed trait AppRunResult
case object AppSuccess extends AppRunResult
case class AppFailure(message: String, returnCode: Option[Int]) extends AppRunResult
trait TaskApp {
  def run(args: List[String]): Task[AppRunResult]
  ...
}
```

I'm just wary of returning Unit and having the compiler silently convert anything and everything for me (I know there's a flag for that, but still...). I also like being able to set the RC for apps.

adelbertc:
I'll wait for things to fall into place a bit, maybe even wait for `Task` to be pushed before continuing with this PR :-)

stew:
heh, I almost requested List => Vector (as it may be common to access by index) but ultimately decided it would be bikeshedding and we don't expect this to be a long list anyway. :)

InTheNow:
> things that require Await, j.u.c, etc.

The Monifu project by @alexandru has a good concurrent implementation that is platform independent. Perhaps such an approach would be useful? [Shared](https://github.com/monifu/monifu/tree/master/core/shared/src/main/scala/monifu/concurrent), [jvm](https://github.com/monifu/monifu/tree/master/core/jvm/src/main/scala/monifu/concurrent), [js](https://github.com/monifu/monifu/tree/master/core/js/src/main/scala/monifu/concurrent)

alexandru:
Hi guys, not sure my code can help. What I do is to provide some utilities in [FutureUtils](https://github.com/monifu/monifu/blob/c05ef4bbb04a9e6749afdd2c9c4b436ffdfa300b/core/shared/src/main/scala/monifu/concurrent/FutureUtils.scala) made possible with the introduction of the introduction of the [Scheduler](https://github.com/monifu/monifu/blob/c05ef4bbb04a9e6749afdd2c9c4b436ffdfa300b/core/shared/src/main/scala/monifu/concurrent/Scheduler.scala) (an `ExecutionContext` that can also do delayed execution), plus [atomic references](https://github.com/monifu/monifu/tree/c05ef4bbb04a9e6749afdd2c9c4b436ffdfa300b/core/jvm/src/main/scala/monifu/concurrent/atomic) that on top of Javascript are just boxed variables (still useful though), plus some utilities built on top, like the [AsyncQueue](https://github.com/monifu/monifu/blob/c05ef4bbb04a9e6749afdd2c9c4b436ffdfa300b/core/shared/src/main/scala/monifu/concurrent/async/AsyncQueue.scala) and the [AsyncSemaphore](https://github.com/monifu/monifu/blob/c05ef4bbb04a9e6749afdd2c9c4b436ffdfa300b/core/shared/src/main/scala/monifu/concurrent/async/AsyncSemaphore.scala). I am not modeling a `Task`, but I am modeling an [Observable](https://github.com/monifu/monifu/blob/563fc7ffbee72e2fd3254a9b0489daa5f4fa05c5/monifu/shared/src/main/scala/monifu/reactive/Observable.scala#L362), similar to the one in RxJava / Rx.NET, being like super-set of `Future` (i.e. a `Future` is just an `Observable` emitting a single item :)).

As a personal opinion - as long as `Task.run` returns an `A`, this imposes the limitation that execution must always be synchronous. And there's no way around it on platforms that cannot block threads and without some sort of continuations support in the runtime. I'm not familiar with what `Task` should be (is it a synchronous equivalent for a `Future`?). If it is, that's fine, but if `Task` is an alternative to `Future`, then personally I consider things that need blocking to work to be evil and I avoid such constructs, as I believe asynchronous execution should be exposed starting from the API. Personally I would do `Task.run(T => Unit)(implicit ctx: Context): Unit`, which would be a `Future` variant I might like, because I sometimes don't like how `Future` submits tasks in the thread-pool as a side-effect on each call of its operators.

InTheNow:
@alexandru For `Task` specifically, maybe... maybe not. But for @non's general comment about `Await` and `j.u.c`, your work does look like that you have more than scratched the surface :wink:

alexandru:
Related to things in `j.u.c`, as long as they don't block, in order to support Scala.js I just wrap things in my own constructs. For example I've got a `ConcurrentQueue` that on top of the JVM is `j.u.c.ConcurrentLinkedQueue` we all know and love, whereas on top of JS it is wrapping just [a regular mutable.Queue](https://github.com/monifu/monifu/blob/c05ef4bbb04a9e6749afdd2c9c4b436ffdfa300b/core/js/src/main/scala/monifu/collection/mutable/ConcurrentQueue.scala). I also have my [own Lock](https://github.com/monifu/monifu/blob/c05ef4bbb04a9e6749afdd2c9c4b436ffdfa300b/core/jvm/src/main/scala/monifu/concurrent/locks/SpinLock.scala) implementation (useful sometimes because Java's intrinsic lock has awful reentry performance) and on top of Scala.js it's [just a macro](https://github.com/monifu/monifu/blob/c05ef4bbb04a9e6749afdd2c9c4b436ffdfa300b/core/js/src/main/scala/monifu/concurrent/locks/locks.scala#L38) that does absolutely nothing with zero overhead.

ceedubs:
I'm going to go ahead and close this out since it sounds like cats is going to defer to other libraries for `Task`.

Issue #500, opened 2015-08-31
Traverse - FoldMap consistency


julien-truffaut:
`foldMap` and `traverse` are not always consistent, i.e.

```scala
  def foldMap[F[_], A, B: Monoid](fa: F[A])(f: A => B): B = 
    traverse(fa)(a => Const(f(a))).getConst
```

for example:

```scala
List(1,2,3).traverseU(i => Const(List(i))).getConst == List(3,2,1)
List(1,2,3).foldMap(List(_)) == List(1,2,3)
```

Shall we add a law to check this?

see: https://hackage.haskell.org/package/base-4.8.1.0/docs/src/Data.Traversable.html#foldMapDefault

non:
:+1: 

In fact, we may want laws to check all our derived methods which can be overridden for performance reasons (i.e. we'd either want laws, or to make derived methods `final`).

adelbertc:
Fixed

PR #142, opened 2015-02-08
List's traverse method was applying effects in reverse.

non:
This commit fixes this issue (which had to do with passing
arguments to G.map2 in the wrong order) and adds a regression
test to be sure we don't accidentally mess this up in the
future.

Fixes #140 

Review by @refried et al.

PR #502, opened 2015-08-31
#500 add derived foldMap from traverse law

julien-truffaut:
I probably missed something in the definition of the law because they all pass but a simple test shows that it is not the case for `List`


julien-truffaut:
@non could it be related to the limitation of `Arbitrary` function you mentioned recently?

julien-truffaut:
`Traverse` laws still pass with non commutative `Monoid`. If I understand correctly it is because because the arbitrary instance of `Function` only generates constant function

non:
@julien-truffaut Correct. Once we have a new ScalaCheck release we will start getting real functions which actually depend on their input.

julien-truffaut:
@non so what do you recommend to do? 
*   commit new law (remove unit test in List) and wait for scalacheck upgrade to fix instances
*   commit new law and try to fix as many instances as possible
*   something else?

non:
@julien-truffaut I guess I would say we should do (1). To me that seems like the best path, and will avoid duplicate (or redundant) work. But I'm fine with any of these to be honest.

julien-truffaut:
I went for (2) but only `List`, `Vector` and `Stream` (I need them for monocle test to pass).

I could fix `List` and `Vector`, we can probably have more efficient impl for `Vector`. However, I am struggling with `Stream`, I wouldn't mind if someone else would like to give it a try.

non:
OK, I'll take a look.

ceedubs:
@julien-truffaut I was perplexed by the discrepancy between the traverse/const and foldMap for List, and I realized that our implementation of `ap` for `Const` is the reverse of what is in scalaz. Are you aware of that? At first glance, both implementations appear to satisfy the necessary laws to me, but I could definitely be mistaken.

Cats:
```scala
scala> val x = Const[List[Int], String](List(1)) ap Const[List[Int], String => String](List(2))
x: cats.data.Const[List[Int],String] = Const(List(1, 2))
```

Scalaz:
```scala
scala> val x = Const[List[Int], String](List(1)) <*> Const[List[Int], String => String](List(2))
x: scalaz.Const[List[Int],String] = Const(List(2, 1))
```

non:
@ceedubs I also noticed some interesting order issues in the *monoidal* branch I'm working on.

refried:
flashback to #142

julien-truffaut:
@ceedubs no I didn't notice. It seems that haskell use the same order than cats http://hackage.haskell.org/package/base-4.8.1.0/docs/src/Control.Applicative.html#line-84

ceedubs:
@julien-truffaut I think Haskell's is actually the opposite of Cats. It looks the same because the value from the first const is "prepended" to the value from the second const. However, the function argument comes first in Haskell while it comes second in Cats. I think that it makes sense to have the function come first in Haskell but last in Scala, due to differences in the languages. However, it does mean that the version of `Const.ap` that we would probably want for `traverse` would probably seem backwards if used by itself. I suspect this might become a non-issue if @non's monoidal work replaces `ap` (though I'm not sure - the plan might be to leave it in).

julien-truffaut:
@ceedubs maybe I missed something but it seems to me that haskell behave in the same way than cats

```
Prelude Control.Applicative Data.Monoid> Const(3 : []) <*> Const(5 : [])
Const [3,5]
```

julien-truffaut:
some additional background of scalaz impl: https://github.com/scalaz/scalaz/commit/084e9c9a5d07144892db8e8dea3033d2d32d31bd#commitcomment-13182666

julien-truffaut:
@ceedubs found out exactly what was the issue!

non:
:+1:

non:
Sorry for the new merge conflicts. If this gets updated I'm happy to see it merged. Thanks!

adelbertc:
:+1: from me once merge conflicts resolved (and Travis is green light)

ceedubs:
What @adelbertc said ^

PR #507, opened 2015-09-02
Introduce cats-jvm and cats-js.

non:
This commit does a bunch of things:

 * Adds a cats.macros.Platform object
 * Converts cats-macros to shared/jvm/js.
 * Converts other Cats modules to CrossType.Pure.
 * Moves JVM-specific code into cats-jvm.
 * Supports limited platform checking in laws/tests.

The new Platform object provides .isJvm and .isJs. These are
notable for being macros: they will be expanded at compile-time
to be either true or false. This allows scalac to eliminate
the "other" branch at compile-time, which will be useful to
us in laws/tests.

Moving forward, the goal is to only admit platform-specific code
(e.g. using java.io.*, or thread-blocking) into cats-jvm and
cats-js (which will not be required to provide the same APIs).
Other modules should be of CrossType.Pure (that is, they should
not have JVM- or JS-specific code in them).

The platform checking using .isJvm and .isJs should only be
used in cats-laws and cats-tests (although of course users of
the Cats library may use them however they like).


non:
As I hinted in the commit message, I think we should allow a limited amount of platform-specific code in laws/tests. The reason is that splitting the tests across projects feels really cumbersome, and IMO using things like `Platform.isJvm` to branch in tests is quite readable. However, the `jvm` module also has its own tests, so for code in `cats-jvm` we will put the tests there.

build.sbt
```diff
@@ -251,7 +267,7 @@ lazy val commonScalacOptions = Seq(
   "-language:implicitConversions",
   "-language:experimental.macros",
   "-unchecked",
-  "-Xfatal-warnings",
+  //"-Xfatal-warnings",
```
adelbertc: Is something necessitating this?

macros/js/src/main/scala/cats/macros/Platform.scala
```diff
@@ -0,0 +1,20 @@
+package cats.macros
+
+import scala.reflect.macros.Context
+
+object Platform {
```
adelbertc: clever :-)

adelbertc:
Just a comment about turning off `-X-fatal-warnings`, I'd like to keep it on if possible. Other than that, awesome stuff - to more reliable builds! :+1:

build.sbt
non: Yes, this is necessary for now for `cats-macros` due to differences between 2.10 and 2.11. I think I can fix this though so I'll re-enable it when I do.

non:
Just to be clear -- I think the builds have already stabilized as far as JVM/JS split is concerned. This is about trying to structure the project to encourage cross-platform support in a way that is easier to get right.

non:
To expand on it a bit more -- in practice it seems like we haven't done a great job of dealing with projects whose code is split between `shared/src`, `jvm/src`, and `js/src`. So the strategy here was to limit us to a single project that does that (`cats-macros`) and then add in two more projects where all the JVM-specific and JS-specific code can go.

This might not be flexible enough (for example, if `cats-state` needed JVM-specific code we'd either have to make `cats-jvm` depend on `cats-state`, create a new module, or split up `state`). However, I think it is flexible enough, and a lot easier to keep straight.

build.sbt
ceedubs: If you won't be fixing it in this PR, what about omitting this option from only the macros module?

non:
Just re-enabled fatal warnings in everything except macros.

The "right" thing to do would be to split macros on 2.10/2.11 as well as JVM/JS, to remove the warnings. But that's a bit too fiddly to supress 2 warnings IMO. Once @milessabin finishes `macro-compat` (or someone else publishes something similar) we could consider using that.

ceedubs:
I agree that it would be overkill to split macros on 2.10/2.11 to get rid of 2 warnings.

I do worry a bit about how complicated our project build/structure has gotten. It seems like it could be tough for potential contributors to jump in. This looks good to me though (with the disclaimer that I'm no expert on these matters). :+1:

milessabin:
macro-compat-1.0.0 is good to go ... have at it :-)

non:
Hey folks, due to extensive work since this PR was created, there were conflicts that had to be updated.

I also pushed our *scala-bricks* dependency forward to the stable `0.0.1`. This means that we no longer need to include the `cats.macros.Platform` object, since that exact object is included in *scala-bricks*. In the future I hope this kind of "general compatibility code" between JS and JVM will come from @InTheNow's *scala-bricks* (which includes @milessabin's *macro-compat*).

I'd like to get this merged (or hear objections) relatively soon, since this PR is likely to conflict with any other changes to laws/tests that we want to make. To sum up the changes:

1. All Cats projects are "pure" Cross projects in SBT (meaning they only contain code that is valid on both the JVM and for JS).
2. There are special `cats.jvm` and `cats.js` projects where we can put useful instances and code for end users (and which none of the "pure" modules depend on).
3. All the tests pass reliably on the JVM and in JS.
4. We depend on *scala-bricks*, a compat library by @InTheNow to make it easy to support both the JVM and JS (and which uses @milessabin's *macro-compat* library to make it easy to support 2.10 and 2.11 in macros).

What do you all think?

adelbertc:
Looks like there is some merge conflicts, but to quote Cody, "This looks good to me though (with the disclaimer that I'm no expert on these matters)" :+1: 

I do like the idea of the `scala-bricks` library to simplify the build structure.

non:
Thanks! This most recent commit should resolve the conflicts. If Travis goes green I'll merge it. :cat2:

Issue #509, opened 2015-09-02
Add Bifunctor laws


fthomas:
`Bifunctor` currently has no laws. It should be straightforward to add `BifunctorLaws` based on the laws for `Functor`.

fthomas:
Duplicate of #557. I'm closing this because #557 has a more detailed description.

PR #511, opened 2015-09-02
Add a Bimonad law.

non:
The law requires that extract(pure(a)) = a.

This commit also updates the Monad* tests to use the new EqK
typeclass, and also relocates some laws from ComonadLaws to
CoflatMapLaws.

There is a fair amount of plubming that had to change to get
all of this working. A really important but thankless task
will be to go through all of our tests and use EqK and
ArbitraryK where possible to create a more consistent
experience. This will only get harder once we have a new
ScalaCheck release and we have to worry about Cogen as well.

### DIFF
--- a/laws/shared/src/main/scala/cats/laws/discipline/BimonadTests.scala
+++ b/laws/shared/src/main/scala/cats/laws/discipline/BimonadTests.scala
@@ -9,15 +9,19 @@ import Prop._
 trait BimonadTests[F[_]] extends MonadTests[F] with ComonadTests[F] {
   def laws: BimonadLaws[F]
 
-  def bimonad[A: Arbitrary: Eq, B: Arbitrary: Eq, C: Arbitrary: Eq]: RuleSet =
+  def bimonad[A: Arbitrary: Eq, B: Arbitrary: Eq, C: Arbitrary: Eq]: RuleSet = {
+    implicit val arbfa: Arbitrary[F[A]] = ArbitraryK[F].synthesize[A]
+    implicit val eqfa: Eq[F[A]] = EqK[F].synthesize[A]
     new RuleSet {
       def name: String = "bimonad"
       def bases: Seq[(String, RuleSet)] = Nil
       def parents: Seq[RuleSet] = Seq(monad[A, B, C], comonad[A, B, C])
       def props: Seq[(String, Prop)] = Seq(
-        "pure and extract compose" -> forAll(laws.pureExtractComposition[A] _)
+        "pure andThen extract = id" -> forAll(laws.pureExtractIsId[A] _),
+        "extract andThen pure = id" -> forAll(laws.extractPureIsId[A] _)
       )
     }
+  }
 }
 
 object BimonadTests {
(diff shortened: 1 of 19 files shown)


InTheNow:
http://reh.math.uni-duesseldorf.de/~wisbauer/Hopfmonad.pdf . I concede that I absolutely no idea as to whether this is useful here or not

non:
Fixes #509

non:
@InTheNow I've seen that paper too. I do think the concept there is related, but most of the issues in the paper are quite far from us I think. Given how useful `Bifunctor`, `Bitraversable`, etc. are I think changing the name of this useful-but-unusal type class makes the most sense.

fthomas:
@non #509 is about laws for Bifunctor not Bimonad. These are not (yet?😊) in this PR.

non:
@fthomas Oops, you are right. I updated the comment, hopefully this won't accidentally close that bug.

laws/shared/src/main/scala/cats/laws/discipline/BimonadTests.scala
```diff
@@ -0,0 +1,30 @@
...
+trait BimonadTests[F[_]] extends MonadTests[F] with ComonadTests[F] {
+  def laws: BimonadLaws[F]
+
+  def bimonad[A: Arbitrary: Eq, B: Arbitrary: Eq, C: Arbitrary: Eq]: RuleSet =
+    new RuleSet {
+      def name: String = "bimonad"
+      def bases: Seq[(String, RuleSet)] = Nil
+      def parents: Seq[RuleSet] = Seq(monad[A, B, C], comonad[A, B, C])
+      def props: Seq[(String, Prop)] = Seq(
+        "pure and extract compose" -> forAll(laws.pureExtractComposition[A] _)
```
adelbertc: "pure and extract are inverses" ?

tests/jvm/src/test/scala/cats/tests/FutureTests.scala
```diff
@@ -14,27 +15,29 @@ import org.scalacheck.Arbitrary.arbitrary
...
 
+  implicit val throwableEq: Eq[Throwable] =
+    Eq.fromUniversalEquals
+
   implicit val comonad: Comonad[Future] = futureComonad(timeout)
 
   // Need non-fatal Throwables for Future recoverWith/handleError
   implicit val nonFatalArbitrary: Arbitrary[Throwable] =
-    Arbitrary(arbitrary[Exception].map(e => e.asInstanceOf[Throwable]))
+    //Arbitrary(arbitrary[Exception].map(_.asInstanceOf[Throwable]))
```
adelbertc: Leftover comment? And reason for using a const generator instead of this one? (though in practice it may not matter)

non: Oh yeah I should fix this. It helped when I was tracking down a test failure.

laws/shared/src/main/scala/cats/laws/discipline/BimonadTests.scala
non: Sure, that's a better name.

ceedubs:
:+1:

adelbertc:
:+1:

PR #517, opened 2015-09-03
Add an example using foldMap on to a tuple

philwills:
This is based on feedback from @adelbertc to a previous PR.

I've also changed 'typeclass' to 'type class'.


adelbertc:
Hurrah! We currently don't have `Monoid` instances for `TupleN` (we need to fix that..) but until then this LGTM :+1:

PR #518, opened 2015-09-08
Return Xor instead of Either in Free.resume

ceedubs:
This makes it consistent with conventions used elsewhere in Cats.


non:
Makes sense to me.

:+1: when Travis passes.

mpilquist:
:+1:

PR #519, opened 2015-09-09
Set up a benchmark for our trampolines.

non:
This commit adds a new benchmark to the bench project.
It updates the version of JMH that we are using, and
also adds a new, hidden Eval subtype which improves
the performance of Eval.defer.


non:
Here's the benchmark result running on my machine:

```
Benchmark                    Mode  Cnt      Score      Error  Units
TrampolineBench.eval        thrpt    3  23049.351 ± 3891.576  ops/s
TrampolineBench.stdlib      thrpt    3  11827.608 ± 1919.654  ops/s
TrampolineBench.trampoline  thrpt    3   5743.017 ±  197.309  ops/s
```

(Run with `bench/jmh:run -i 3 -wi 3 -f1 -t1 cats.bench.TrampolineBench`)

ceedubs:
Are the build failures random or related to this?

non:
I'm looking into it. It doesn't seem like Travis flakiness.

non:
So `validate` seems to reliably pass locally. It could be an issue of memory/resources available.

ceedubs:
I restarted the builds and this time they worked in 2.10 but not 2.11. It does seem like a quirky sporadic/resource issue, but if we aren't seeing this outside of this branch, I'm also a little hesitant to merge this PR simply because I don't want us to start completely ignoring build statuses. @InTheNow do you have any ideas?

non:
Maybe we should set up Contained-Based builds? I wonder if that would help: http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/

non:
Awesome. We seem to have overcome the previous build flakiness!

ceedubs:
:+1:

adelbertc:
:+1:

PR #521, opened 2015-09-10
add xorMonoid

drostron:


tests/shared/src/test/scala/cats/tests/XorTests.scala
```diff
@@ -12,6 +13,8 @@ import org.scalacheck.Arbitrary._
 import scala.util.Try
 
 class XorTests extends CatsSuite {
+  checkAll("Xor[String, Int]", algebra.laws.GroupLaws[Xor[String, Int]].monoid)
```
adelbertc: Check Serializability

drostron: I see `- Xor[String, Int].monoid.serializable` in the test output. It appears serializability is checked via:

https://github.com/non/algebra/blob/v0.3.1/laws/shared/src/main/scala/algebra/laws/GroupLaws.scala#L28

Is that sufficient or is there a reason to have the following in addition?

```scala
checkAll("Xor[String, Int]", SerializableTests.serializable(Monoid[Xor[String, Int]]))
```

adelbertc: Oh you're right, `algebra` includes `Serializable` checks in their laws - @non confirm?

adelbertc:
:+1:

non:
:+1:

PR #522, opened 2015-09-10
Hook Eq up to scalactic/scalatest's equality

### LINKED ISSUE #219: hook up scalatest === to cats' eq
spoke @bvenners:

Not sure if you're aware, but you can teach ScalaTest's === to look for Cat's Eq instances, so the type checkiung comes from the presence of a Cats Eq instance but you still get decent error messages from ScalaTest when assertions fail.
I demoed how to do this in scalaz and spire a while back here: 

https://github.com/bvenners/equality-integration-demo

ceedubs:
Fixes #219

This allows us to use matchers like `should ===`. I changed a few
random places to do this, but I didn't want to go through the effort
of changing tons of places until this PR was reviewed/merged.

I think we could clean a lot of things up now by having CatsSuite
extend GeneratorDrivenPropertyChecks, which will allow using matchers
like `should ===` inside of `forAll` properties.


ceedubs:
Thanks @bvenners for setting up a great [example project](https://github.com/bvenners/equality-integration-demo) for this. Bill, if you happen to take a look at this and see that I'm doing anything silly, let me know.

adelbertc:
I'm guessing `should ===` gives better failure messages than just `Object#==` , similar to Specs2's `mustEqual`s? We got a lot of usages of `==` scattered through our tests, though we can probably put that in a separate PR?

:+1:

ceedubs:
@adelbertc I think `assert(a == b)` actually returns good error messages because `assert` is a fancy scalatest macro. However, as you mentioned, it uses universal equality instead of `Eq`. This will allow us to have `===` provide a good error message _and_ use our `Eq` instances (though I should test check that it's actually successfully requiring the `Eq` instance).

I agree that we should change a lot of `==` usage, and I can lead this effort, but I wanted this PR to go through first so the tedious effort wasn't in vain :)

adelbertc:
Yep sounds good, :+1: for me

tests/shared/src/test/scala/cats/tests/CatsEquality.scala
```diff
@@ -0,0 +1,29 @@
+package cats
```
InTheNow: I think this whole file would be useful in tests that use the cats library, and not just cats itself. So, if that is the case, consider moving the file to `main`, so it could be exported - i.e. `tests/shared/src/main/scala/cats/tests/CatsEquality.scala`

tests/shared/src/test/scala/cats/tests/CatsSuite.scala
```diff
@@ -16,13 +16,15 @@ import scala.util.{Failure, Success, Try}
...
   implicit override val generatorDrivenConfig: PropertyCheckConfiguration =
     PropertyCheckConfiguration(
       minSuccessful = Platform.minSuccessful,
       maxDiscardedFactor = Platform.maxDiscardedFactor)
-  // disable scalatest's ===
-  override def convertToEqualizer[T](left: T): Equalizer[T] = ???
+
+  // disable Eq syntax (by making `eqSyntax` not implicit), since it collides
+  // with scalactic's equality
+  override def eqSyntax[A: Eq](a: A): EqOps[A] = new EqOps[A](a)
```
InTheNow: As per previous comment, consider moving this into its own trait in `main`, perhaps in conjunction with the other traits. 
This would also mean that users need not import `cats.syntax.EqOps`

ceedubs:
@InTheNow I like the ideas in your comments. However, we currently aren't publishing the `tests` module at all. I've created #529 for this as a future idea. I think I'd prefer to keep this PR lightweight and hopefully uncontroversial and defer more complicated decisions/work. Does that sound okay?

Once this gets through I plan to update a lot of tests to take advantage of this syntax. I know that I've made type errors with `==` several times when writing Cats tests, and this seems like a mistake we shouldn't have to make when we have `Eq` instances available in all of our tests :)

InTheNow:
@ceedubs Absolutely OK! Note I used "consider" to introduce the concept, but more for a future issue/PR, where I shall now continue the conversation :+1:

non:
Thanks for this. Looks good! :+1:

ceedubs:
I merged in master and resolved a couple merge conflicts due to #507 being merged. Since there are several :+1:s, I'll merge once Travis goes green if there are no objections.

ceedubs:
The 2.11 build is failing with the following error:

```
[error] possible missing interpolator: detected interpolated identifier `$conforms`
[error] one error found
[error] (freeJVM/test:compileIncremental) Compilation failed
```

I don't see any reference to `conforms` in Cats, so I'm guessing this is coming from one of these changes not playing nicely with the recent macro/compiler plugins changes from #507. Any idea where this is coming from and what I've done to trigger it @non or @InTheNow or @milessabin?

InTheNow:
@ceedubs [in build.sbt](https://github.com/non/cats/blob/master/build.sbt#L277), try using  `   "-Xlint:-missing-interpolator,_",```

ceedubs:
@InTheNow thanks for the suggestion. That makes it work for 2.11, but that isn't supported in 2.10, so now that build fails :(

We could hack around the compiler flags based on the cross-version, but it seems like it would be nice to figure out where this is coming from and fix it in a more elegant way. Any idea?

InTheNow:
@ceedubs I'm guessing it's the two calls to `===` in https://github.com/non/cats/blob/master/free/src/test/scala/cats/free/YonedaTests.scala and https://github.com/non/cats/blob/master/free/src/test/scala/cats/free/CoyonedaTests.scala that are noe using a different `===`

If so, you can remove the `"-Xlint:-missing-interpolator,_",`

ceedubs:
@InTheNow ah thanks good catch! I think that solved the issue. We'll see if Travis agrees.

ceedubs:
We've gone green. I'm going to go ahead and merge since there are three +1s.

PR #524, opened 2015-09-11
FreeApplicative#hoist is same as compile

adelbertc:


fthomas:
`compile` requires an `Applicative[G]` but `hoist` does not. If they are the same why don't we remove the one which is more restricted?

adelbertc:
Good catch - `Applicative` constraint was not necessary :-) The 2 are the same

free/src/main/scala/cats/free/FreeApplicative.scala
```diff
@@ -45,7 +38,8 @@ sealed abstract class FreeApplicative[F[_], A] extends Product with Serializable
   final def fold(implicit F: Applicative[F]): F[A] =
     foldMap(NaturalTransformation.id[F])
 
-  final def compile[G[_]](f: F ~> G)(implicit G: Applicative[G]): FA[G, A] =
+  /** Interpreter this FreeApplicative algebra into another using FreeApplicative. */
```
ceedubs: Minor typo here. I think you probably meant "Interpret this..."

ceedubs:
:+1:

non:
:+1:

PR #525, opened 2015-09-11
Yoneda tests

adelbertc:


free/src/test/scala/cats/free/YonedaTests.scala
```diff
@@ -0,0 +1,30 @@
...
+import org.scalacheck.Prop.forAll
+
+class YonedaTests extends CatsSuite {
+  implicit def yonedaArbitraryK[F[_] : Functor](implicit F: ArbitraryK[F]): ArbitraryK[Yoneda[F, ?]] =
+    new ArbitraryK[Yoneda[F, ?]]{
+      def synthesize[A: Arbitrary]: Arbitrary[Yoneda[F, A]] =
+        Arbitrary(F.synthesize[A].arbitrary.map(Yoneda(_)))
+    }
+
+  implicit def yonedaArbitrary[F[_] : Functor : ArbitraryK, A : Arbitrary]: Arbitrary[Yoneda[F, A]] = yonedaArbitraryK[F].synthesize[A]
```
ceedubs: So far I think we have been putting Arbitrary and ArbitraryK instances in the `Arbitrary.scala` and `ArbitraryK.scala` files. Do we want these there as well?

adelbertc: I put it in the test to be the same as that of `FreeTests` - I think the reason we put them here is because `tests` does not depend on `free` at the moment, which leads us to doing strange things like placing `Arbitrary` instances inside the test. Perhaps we want `tests` to `dependOn(free)` ?

ceedubs: Ah I see. I think the problem with `tests` depending on `free` is that `free` depends on `tests` for `CatsSuite` :\

I forgot that `Free` had its own module. I think it makes sense for modules to contain their own arbitrary instances. At some point we may want these other modules to publish their own generators for others to consume instead of just having them within the test classes. I think that definitely should be a separate PR if we choose to do that though.

ceedubs:
:+1:

non:
:+1:

PR #527, opened 2015-09-11
Between releases we should probably be using -SNAPSHOT.

non:


stew:
:+1:

rossabaker:
:+1: pending Travis.

milessabin:
Don't we have the release plugin set up to do this automatically?

PR #530, opened 2015-09-12
Streaming.take(n).toList shouldn't blow up when n+1 is an error

ceedubs:
The `Cons(a, lt.map(_.take(n - 1)))` seems reasonable enough, but since `toList` calls `lt.value`, this blows up when `lt` is a deferred error.


core/src/main/scala/cats/data/Streaming.scala
```diff
@@ -473,7 +473,9 @@ sealed abstract class Streaming[A] { lhs =>
     if (n <= 0) Empty() else this match {
       case Empty() => Empty()
       case Wait(lt) => Wait(lt.map(_.take(n)))
-      case Cons(a, lt) => Cons(a, lt.map(_.take(n - 1)))
+      case Cons(a, lt) =>
+        if (n == 1) Cons(a, Now(Empty()))
+        else Cons(a, lt.map(_.take(n - 1)))
```
non: Maybe I'm too fussy, but could we rewrite that line?

```scala
Cons(a, if (n == 1) Now(Empty()) else lt.map(_.take(n - 1))
```

non:
:+1: (with-or-without my fussy suggestion honestly)

non:
:+1:

adelbertc:
Merge conflicts. Also one nitpicky comment :-)

core/src/main/scala/cats/data/Streaming.scala
```diff
@@ -473,7 +473,8 @@ sealed abstract class Streaming[A] { lhs =>
     if (n <= 0) Empty() else this match {
       case Empty() => Empty()
       case Wait(lt) => Wait(lt.map(_.take(n)))
-      case Cons(a, lt) => Cons(a, lt.map(_.take(n - 1)))
+      case Cons(a, lt) =>
+        Cons(a, if (n ==1) Now(Empty()) else lt.map(_.take(n - 1)))
```
adelbertc: Eeeek, space between `==` and `1` :-)

ceedubs:
@adelbertc thanks. Hopefully this should pass inspection now :)

non:
:+1:

adelbertc:
:+1:

PR #531, opened 2015-09-13
Add Coyoneda tests and remove `by`

adelbertc:
Not sure what `by` does and how the implicit `F[A]` param would be used.


non:
Seems good to me. If someone wants to chime in on `by` usage we can re-add it later.

:+1:

ceedubs:
:+1:

PR #532, opened 2015-09-15
Use type-safe equals checks in tests

ceedubs:
I grabbed the ones that I could easily grep for with `assert`. There are more that I want to change inside Scalacheck `forAll` blocks, but I think I'm going to make that a separate PR that introduces `GeneratorDrivenPropertyChecks` into `CatsSuite`.

### DIFF
--- a/free/src/test/scala/cats/free/FreeApplicativeTests.scala
+++ b/free/src/test/scala/cats/free/FreeApplicativeTests.scala
@@ -40,7 +40,7 @@ class FreeApplicativeTests extends CatsSuite {
     val y = FreeApplicative.pure[Option, Int](n)
     val f = x.map(i => (j: Int) => i + j)
     val r = y.ap(f)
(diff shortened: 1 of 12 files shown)


non:
Weird question -- are those parens always necessary?

ceedubs:
@non I think they are when `should` is used (from past experience - I haven't checked all of these). There may be other Scalatest matchers that are better for this? I don't know Scalatest that well.

non:
:+1: 

In the long run it might be nice to replace `x should === (y)` with something like `x shouldEq y` which be a bit easier to parse.

ceedubs:
@non yeah, I would like something like that. As far as I can tell, the only way to get a type-safe equals check with Scalatest is to use `should ===` with parens. We would probably need to add support for this ourselves (either as a PR to scalatest or within Cats tests).

non:
Right. To be clear, I think we should merge what you have and then explore other strategies independently.

adelbertc:
:+1:

PR #535, opened 2015-09-17
Add diagrams to scaladoc

InTheNow:
Partial implementation for https://github.com/non/cats/issues/95. Example output [in gitter](https://gitter.im/non/cats?at=55fa781618e0111d7e4f4a8b)

If accepted, gh-pages will have to be pushed too.


stew:
:+1:

PR #536, opened 2015-09-17
Add future{Comonad, Eq, Order, PartialOrder} to js

InTheNow:
Fix https://github.com/non/cats/issues/458

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -198,6 +198,7 @@ lazy val js = project
   .settings(moduleName := "cats-js")
   .settings(catsSettings:_*)
   .settings(commonJsSettings:_*)
+  .enablePlugins(ScalaJSPlugin)
 
 lazy val publishSettings = Seq(
   homepage := Some(url("https://github.com/non/cats")),
@@ -217,11 +218,11 @@ lazy val publishSettings = Seq(
 ) ++ credentialSettings ++ sharedPublishSettings ++ sharedReleaseProcess 
 
 // These aliases serialise the build for the benefit of Travis-CI.
-addCommandAlias("buildJVM", ";macrosJVM/compile;coreJVM/compile;freeJVM/compile;freeJVM/test;stateJVM/compile;stateJVM/test;lawsJVM/compile;testsJVM/test;bench/test")
+addCommandAlias("buildJVM", ";macrosJVM/compile;coreJVM/compile;freeJVM/compile;freeJVM/test;stateJVM/compile;stateJVM/test;lawsJVM/compile;testsJVM/test;jvm/test;bench/test")
 
 addCommandAlias("validateJVM", ";scalastyle;buildJVM;makeSite")
 
-addCommandAlias("validateJS", ";macrosJS/compile;coreJS/compile;lawsJS/compile;testsJS/test;freeJS/compile;freeJS/test;stateJS/compile;stateJS/test")
+addCommandAlias("validateJS", ";macrosJS/compile;coreJS/compile;lawsJS/compile;testsJS/test;js/test;freeJS/compile;freeJS/test;stateJS/compile;stateJS/test")
 
 addCommandAlias("validate", ";validateJS;validateJVM")
 
--- a/js/src/main/scala/cats/jvm/std/future.scala
+++ b/js/src/main/scala/cats/jvm/std/future.scala
@@ -1,0 +1,50 @@
+package cats
+package js
+package std
+
+import scala.concurrent.Future
+import scala.concurrent.{ExecutionContext => E}
+import scala.concurrent.duration.FiniteDuration
+
+import cats.std.FutureCoflatMap
+import cats.syntax.all._
+
+object future extends FutureInstances0
+
+object Await {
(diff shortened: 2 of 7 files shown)


InTheNow:
As I'm not sure where else to put these links, some references:

* [Event-based Concurrency (Advanced)](http://pages.cs.wisc.edu/~remzi/OSTEP/threads-events.pdf)
* [How JavaScript Timers Work](http://ejohn.org/blog/how-javascript-timers-work/)

non:
Looks good to me. :+1:

js/src/main/scala/cats/jvm/std/future.scala
```diff
@@ -0,0 +1,50 @@
...
+}
+
+trait FutureInstances0 extends FutureInstances1 {
+  def futureComonad(atMost: FiniteDuration)(implicit ec: E): Comonad[Future] =
+    new FutureCoflatMap with Comonad[Future] {
+      def extract[A](x: Future[A]): A =
+        Await.result(x, atMost)
+    }
+
+  def futureOrder[A: Order](atMost: FiniteDuration)(implicit ec: E): Eq[Future[A]] =
```
ceedubs: I think this return type should be `Order` instead of `Eq`.

build.sbt
```diff
@@ -216,11 +217,11 @@ lazy val publishSettings = Seq(
 ) ++ credentialSettings ++ sharedPublishSettings ++ sharedReleaseProcess 
 
 // These aliases serialise the build for the benefit of Travis-CI.
-addCommandAlias("buildJVM", ";macrosJVM/compile;coreJVM/compile;freeJVM/compile;freeJVM/test;stateJVM/compile;stateJVM/test;lawsJVM/compile;testsJVM/test;bench/test")
+addCommandAlias("buildJVM", ";macrosJVM/compile;coreJVM/compile;freeJVM/compile;freeJVM/test;stateJVM/compile;stateJVM/test;lawsJVM/compile;testsJVM/test;jvm/test;bench/test")
```
ceedubs: What's the difference between `jvm/test` and `testsJVM/test`? Similarly below for `js`.

js/src/test/scala/cats/tests/FutureTests.scala
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+  implicit val throwableEq: Eq[Throwable] =
+    Eq.fromUniversalEquals
+
+  implicit val comonad: Comonad[Future] = futureComonad(timeout)
+
+  // Need non-fatal Throwables for Future recoverWith/handleError
+  implicit val nonFatalArbitrary: Arbitrary[Throwable] =
+    Arbitrary(arbitrary[Exception].map(_.asInstanceOf[Throwable]))
```
ceedubs: `asInstanceOf` shouldn't be necessary here. I think it should compile if you just use `.map(identity)`.

ceedubs:
I left a few minor comments/questions, but in general this looks good to me.

build.sbt
InTheNow: `testsJVM` and `testsJS` are the projects for the `tests` cross project.So neither contain actual code

`jvm` and `js` are top level projects that are _not_ cross projects, and do contain code

js/src/main/scala/cats/jvm/std/future.scala
InTheNow: Perhaps - but so should [the lvm version](https://github.com/non/cats/blob/master/jvm/src/main/scala/cats/jvm/std/future.scala#L2)

build.sbt
ceedubs: Ah okay I think I understand now. Thanks for the explanation.

js/src/test/scala/cats/tests/FutureTests.scala
InTheNow: ditto ^^  Perhaps - but so should [the lvm version](https://github.com/non/cats/blob/master/jvm/src/test/scala/cats/tests/FutureTests.scala)

js/src/main/scala/cats/jvm/std/future.scala
InTheNow: I've also noticed that the dir is wrong (has jvm in path). 

`js/src/main/scala/cats/jvm/std/future.scala` to `js/src/main/scala/cats/js/std/future.scala`

InTheNow:
Shall I fix `jvm` version in this PR (easy) or a new one (strictly correct)?

ceedubs:
I say throw the change in here. It doesn't bother me, and I have a suspicion that you'll still get a :+1: from @non :)

non:
@ceedubs is correct :+1:

InTheNow:
Cool, done already, ready for review

build.sbt
non: Yeah, I am actually the origin of this confusing difference, but I think in the long-run it will make sense ;)

non:
Looks good to me. Thanks @InTheNow!

ceedubs:
:+1: thanks!

PR #538, opened 2015-09-22
Adjust implicit priorities in OneAnd.

paulp:
```
scala> implicitly[Functor[NonEmptyList]]
<console>:27: error: ambiguous implicit values:
 both method oneAndFunctor in trait OneAndInstances ...
 and method oneAndMonad in trait OneAndInstances ...
 match expected type cats.Functor[cats.data.NonEmptyList]
```

There were two bugs: one is that the Monad implicit
conflicts with the Functor implicit. The other is that
this construction doesn't work:

`object OneAnd extends OneAndInstances with OneAndLowPriority`

If it DID work, this ordering would give priority to the
implicits found in OneAndLowPriority, so it's reversed - but
more to the point, implicit priority isn't based on linearization.
One of the traits has to be a subtrait of the other to have
higher priority, and as written they had no relationship.

### DIFF
--- a/core/src/main/scala/cats/data/OneAnd.scala
+++ b/core/src/main/scala/cats/data/OneAnd.scala
@@ -87,7 +87,7 @@ final case class OneAnd[A, F[_]](head: A, tail: F[A]) {
     s"OneAnd(${A.show(head)}, ${FA.show(tail)})"
 }
 
-trait OneAndInstances {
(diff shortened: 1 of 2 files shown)


adelbertc:
:+1: once Travis is green

non:
:+1:

Great catch, thanks Paul!

PR #542, opened 2015-09-25
Added minimal WriterT implementation.

sungiant:


core/src/main/scala/cats/data/WriterT.scala
```diff
@@ -0,0 +1,73 @@
+package cats
+package data
+
+final case class WriterT[T[_], L, V] (run: T[(L, V)]) {
```
ceedubs: Should we change `T[_]` to `F[_]`? If memory serves, everywhere else we are pretty much using `F[_]` (and `G[_]` if there are two type constructor params)

core/src/main/scala/cats/data/WriterT.scala
```diff
@@ -0,0 +1,73 @@
...
+    functorT.map (run)(_._2)
+
+  def map[Z](fn: V => Z)(implicit functorT: Functor[T]): WriterT[T, L, Z] =
+    WriterT {
+      functorT.map (run) { z =>
+        (z._1, fn (z._2))
+      }
+    }
+
+  def flatMap[U](f: V => WriterT[T, L, U])(implicit monadT: Monad[T], semigroupL: Semigroup[L]): WriterT[T, L, U] =
```
ceedubs: Can this `Monad` constraint be relaxed to `FlatMap`?

core/src/main/scala/cats/data/WriterT.scala
```diff
@@ -0,0 +1,73 @@
...
+    WriterT {
+      monadT.flatMap (run) { lv =>
+        monadT.map (f (lv._2).run) { lv2 =>
+          (semigroupL.combine (lv._1, lv2._1), lv2._2)
+        }
+      }
+    }
+
+  def mapValue[M, U] (f: ((L, V)) => (M, U))(implicit functorT: Functor[T]): WriterT[T, M, U] =
```
ceedubs: It seems like this is more of a `mapBoth` than a `mapValue`. I think `map` already covers you if you just want to map the value component.

core/src/main/scala/cats/data/WriterT.scala
```diff
@@ -0,0 +1,73 @@
...
+    mapValue (wa => (wa._2, wa._1))
+
+  def reset (implicit monoidL: Monoid[L], functorT: Functor[T]): WriterT[T, L, V] =
+    mapWritten (_ => monoidL.empty)
+}
+object WriterT extends WriterTInstances with WriterTFunctions
+
+sealed abstract class WriterTInstances {
+  implicit def writerTMonad[T[_], L] (implicit monadT: Monad[T], monoidL: Monoid[L]) = {
+    new Monad[({type WT[$] = WriterT[T, L, $]})#WT] {
```
ceedubs: Would you mind converting these to use [kind-projector](https://github.com/non/kind-projector) to be consistent with other Cats code?

core/src/main/scala/cats/data/WriterT.scala
```diff
@@ -0,0 +1,73 @@
...
+        fa.flatMap (a => ff.map (f => f (a)))
+    }
+  }
+}
+
+trait WriterTFunctions {
+  def putT[T[_], L, V] (vt: T[V])(l: L)(implicit functorT: Functor[T]): WriterT[T, L, V] =
+    WriterT (functorT.map (vt)(v => (l, v)))
+
+  def put[T[_], L, V] (v: V)(l: L)(implicit functorT: Functor[T], applicativeT: Applicative[T]): WriterT[T, L, V] =
```
ceedubs: You should be able to drop this `functorT`. It's redundant with `applicativeT`.

ceedubs: This is true of a couple places below too.

ceedubs:
Thanks for submitting this! `WriterT` is a great addition to Cats.

I left some comments, but in general this looks great to me.

We should hook this up to some law-checking and tests. If you would be willing to do that on this PR, that would be great. Otherwise, I'm happy to work on that.

One other small stylistic note: in other parts of Cats we aren't putting a space in front of `(` for `def`s and function calls. We probably should have a Scalastyle rule enforcing consistency for that, but in the meantime, would you be willing to match the convention?

Thanks again, @sungiant!

sungiant:
Hi @ceedubs, thanks for the feedback, I've made a few stylistic changes as per your suggestions.

I've not made the Kind-Projector change as I'm not very familiar with how to make the change, how would `({type WT[$] = WriterT[F, L, $]})#WT[A]` be rewritten?

As for tests, I took a look, but I don't fully understand how the law-checking part works, it would be great if someone in the know could get the ball rolling.

ceedubs:
@sungiant thank you. You should be able to rewrite `new Monad[({type WT[$] = WriterT[F, L, $]})#WT]` as `new Monad[WriterT[F, L, ?]]` using kind-projector. I can add the law-checking (and maybe some other tests), but I probably won't be able to get around to it until mid-to-late next week.

ceedubs:
@sungiant I've submitted a PR to your repo that uses kind projector and adds some law-checking and a couple tests.

Some things that we should probably do (but I think could wait until subsequent PRs):
* Add a Writer alias for when `F` is `Id`
* Add some more instances (Functor, Applicative, Bifunctor, Semigroup, Monoid, etc) along with law-checking for those instances
* Add more ad-hoc tests

ceedubs:
:+1:

Thanks @sungiant!

non:
Looks great, thanks!

:+1:

PR #545, opened 2015-09-27
Add Semigroup instance for OneAnd

liff:
Add `Semigroup` instance for OneAnd based on its `SemigroupK`.

I couldn't find any examples on how to write tests for this, so any help is appreciated.

### DIFF
--- a/core/src/main/scala/cats/data/OneAnd.scala
+++ b/core/src/main/scala/cats/data/OneAnd.scala
@@ -103,6 +103,9 @@ trait OneAndInstances extends OneAndLowPriority1 {
         a combine b
     }
 
+  implicit def oneAndSemigroup[F[_]: MonadCombine, A]: Semigroup[OneAnd[A, F]] =
+    oneAndSemigroupK.algebra
+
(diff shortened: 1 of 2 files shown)


ceedubs:
:+1: (assuming Travis turns green)

There's a [gitter conversation](https://gitter.im/milessabin/export-hook?at=56081cd4efd97f3e4669b170) going on as to whether we might support an automatic `SemigroupK` => `Semigroup` conversion with some help from export-hook. The reason export-hook would probably need to be involved is that such a conversion currently needs to live in the `Semigroup` companion object, which lives in Algebra while `SemigroupK` lives in Cats. I suppose another option is to move `SemigroupK` into Algebra. I'm curious what @non thinks about these possibilities. But I think we should move forward with this PR for now and talk about the other possibilities as long-term possibilities.

non:
I think supporting export-hook in algebra is a good idea, but I think we should move forward in Cats until that is all set up. :+1:

non:
@liff I think the `checkAll` calls you made are fine. They will test that the semigroups work.

PR #547, opened 2015-09-28
Fix monoid zeros for Float/Double

### LINKED ISSUE #301: Monoid[Double] and Monoid[Float] are incorrect
https://github.com/xuwei-k/cats/commit/6dbad2ddff10

ceedubs:
Fixes #301.

This was ignored for far too long (even though there was #301 for it).
We should really have test coverage on this. I say we merge this for now
but discuss whether this should just be coming from algebra and add
tests to whichever project it comes from.


fthomas:
:+1:

non:
:+1:

PR #548, opened 2015-09-30
Add an examples subproject.

non:
The goal here is for people to be able to add
interesting code snippets here.

It's possible that many pieces of example code
which might live here could (or should) be expanded
into tutorials with narrative text. I still think
it might be smart to have a file like this, but
it wouldn't be hard to work the single example
I provided (dealing with Streaming) into a Tut
document.


non:
Oh wait, looks like an errant change ended up in here.

ceedubs:
@non I think you mentioned that you became convinced that an examples subproject shouldn't exist. Should we close this out? Or did you want to turn this PR into a tut doc?

PR #549, opened 2015-09-30
Clean up Either tests

ceedubs:
I think some of this code must have been written before we had
law-checking infrastructure in place for Eq/Order.

### DIFF
--- a/tests/src/test/scala/cats/tests/EitherTests.scala
+++ b/tests/src/test/scala/cats/tests/EitherTests.scala
@@ -2,10 +2,11 @@ package cats
 package tests
 
 import cats.laws.discipline.{TraverseTests, MonadTests, SerializableTests}
+import algebra.laws.OrderLaws
 import org.scalacheck.Prop._
 
 class EitherTests extends CatsSuite {
-  checkAll("Either[Int, Int]", MonadTests[Either[Int, ?]].flatMap[Int, Int, Int])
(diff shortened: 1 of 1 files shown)


tests/src/test/scala/cats/tests/EitherTests.scala
```diff
@@ -17,6 +18,11 @@ class EitherTests extends CatsSuite {
   val monad = implicitly[Monad[Either[Int, ?]]]
   val show = implicitly[Show[Either[Int, String]]]
 
+  val orderLaws = OrderLaws[Either[Int, String]]
+  checkAll("Either[Int, String]", orderLaws.eqv)
+  checkAll("Either[Int, String]", orderLaws.partialOrder(partialOrder))
+  checkAll("Either[Int, String]", orderLaws.order(order))
```
adelbertc: Doesn't `order` include both `partialOrder` and `eqv` ?

ceedubs: Please see comment by one @adelbertc: https://github.com/non/cats/blob/master/tests/src/test/scala/cats/tests/ConstTests.scala#L27-L30

;)

This probably isn't quite as good as the `Const` approach of using types that _only_ have the instance for the laws being checked.

adelbertc: Ahh I see :-) I think I actually prefer your way over figuring our which types happen to have the instances I want

adelbertc:
:+1:

non:
:+1:

PR #552, opened 2015-10-01
Use GeneratorDrivePropertyChecks in tests

ceedubs:
This makes CatsSuite extend Scalatest's GeneratorDrivenPropertyChecks, which allows us to use Scalatest matches inside of our `forAll`-style generator-driven property checks.

Example:

### old
```scala
test("OptionT[Id, A].isEmpty consistent with Option.isEmpty")(check {
  forAll { o: Option[Int] =>
    o.isEmpty == OptionT[Id, Int](o).isEmpty
  }
})
```

### new
```scala
test("OptionT[Id, A].isEmpty consistent with Option.isEmpty") {
  forAll { o: Option[Int] =>
    o.isEmpty should === (OptionT[Id, Int](o).isEmpty)
  }
}
```

This allows us to drop the awkward `check` call and use a type-safe equals comparison.

For those of you looking at this and thinking "wow, I can't believe this person is meticulous enough to go through and change all of these": hi, I'm Cody; I don't believe we've met. In my defense, a vim macro or two did a lot of the work.

### DIFF
--- a/free/src/test/scala/cats/free/CoyonedaTests.scala
+++ b/free/src/test/scala/cats/free/CoyonedaTests.scala
@@ -7,9 +7,8 @@ import cats.laws.discipline.{ArbitraryK, FunctorTests, SerializableTests}
 
 import org.scalacheck.{Arbitrary, Gen}
 import org.scalacheck.Arbitrary.arbitrary
-import org.scalatest.prop.GeneratorDrivenPropertyChecks
 
-class CoyonedaTests extends CatsSuite with GeneratorDrivenPropertyChecks {
+class CoyonedaTests extends CatsSuite {
   implicit def coyonedaArbitraryK[F[_] : Functor](implicit F: ArbitraryK[F]): ArbitraryK[Coyoneda[F, ?]] =
     new ArbitraryK[Coyoneda[F, ?]]{
       def synthesize[A: Arbitrary]: Arbitrary[Coyoneda[F, A]] = coyonedaArbitrary[F, A]
--- a/free/src/test/scala/cats/free/CoyonedaTests.scala
+++ b/free/src/test/scala/cats/free/CoyonedaTests.scala
@@ -7,9 +7,9 @@ import cats.laws.discipline.{ArbitraryK, FunctorTests, SerializableTests}
 
 import org.scalacheck.{Arbitrary, Gen}
 import org.scalacheck.Arbitrary.arbitrary
-import org.scalacheck.Prop.forAll
+import org.scalatest.prop.GeneratorDrivenPropertyChecks
 
-class CoyonedaTests extends CatsSuite {
+class CoyonedaTests extends CatsSuite with GeneratorDrivenPropertyChecks {
   implicit def coyonedaArbitraryK[F[_] : Functor](implicit F: ArbitraryK[F]): ArbitraryK[Coyoneda[F, ?]] =
     new ArbitraryK[Coyoneda[F, ?]]{
       def synthesize[A: Arbitrary]: Arbitrary[Coyoneda[F, A]] = coyonedaArbitrary[F, A]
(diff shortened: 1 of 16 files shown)


non:
Thanks! I'm really glad we are converging on a good convention here. :+1:

ceedubs:
I think one other benefit of this approach that I didn't mention is that we should get some more helpful failure messages in places that we were previously returning a `Boolean` Scalacheck `Prop` result like `x == y` where we are now using `x should === (y)`.

InTheNow:
@ceedubs At risk of being a killjoy....

As well as CatsSuite there is also [CatsProps](https://github.com/non/cats/blob/master/tests/src/test/scala/cats/tests/CatsSuite.scala#L44) that extends [PropertyChecks](http://doc.scalatest.org/2.2.4/#org.scalatest.prop.PropertyChecks) that extends `GeneratorDrivenPropertyChecks`

So I'm guessing the idea was (and that may have changed) to have two classes, a _Tests_ one and a `Props` one, eg for`Ior`:

 ```scala
class IorTests extends CatsSuite {
  checkAll("Ior[String, Int]", MonadTests[String Ior ?].monad[Int, Int, Int])
  checkAll("Monad[String Ior ?]]", SerializableTests.serializable(Monad[String Ior ?]))

  checkAll("Ior[String, Int] with Option", TraverseTests[String Ior ?].traverse[Int, Int, Int, Int, Option, Option])
  checkAll("Traverse[String Ior ?]", SerializableTests.serializable(Traverse[String Ior ?]))
}

class IorProps  extends CatsProps {

  property("left Option is defined left and both") {
    forAll { (i: Int Ior String) =>
      (i.isLeft || i.isBoth) should === (i.left.isDefined)
    }
  }

  property("right Option is defined for right and both") {
    forAll { (i: Int Ior String) =>
      (i.isRight || i.isBoth) should === (i.right.isDefined)
    }
  }
....
}
```

InTheNow:
Additionally,  https://github.com/non/cats/issues/529#issuecomment-139799747 is then made almost impossible see the two closed linked issues at the end

ceedubs:
@InTheNow thanks for the feedback. It looks like @non introduced `CatsProps` with https://github.com/non/cats/commit/95a38524d1b71d8957dfd0c7db1190dc856256b1 so he may be able to weigh in on why it was introduced. I assumed it was for the same reason that I introduced `GeneratorDrivenPropertyChecks`: using Scalatest matchers inside `forAll` is clunky. He may have been going with something more like what you were saying.

I'm not sure I understand your comment in https://github.com/non/cats/pull/552#issuecomment-145365384. `CatsSuite` already has a couple Scalatest traits mixed into it, so does this really make it that much worse? My thought is that we should have an opinionated `CatsSuite` that makes it easy to write consistently-styled tests for Cats, but that we should expose useful pieces as standalone components. That is, we could have a `cats-scalatest` module to expose some of the useful Scalatest pieces (like `CatsEquality`.

What do you think?

InTheNow:
@ceedubs The short answer here is not so short. If you look at https://github.com/InTheNow/catalysts/tree/master/testkit/src/main/scala/catalysts/testkit I've abstracted most of the `CatsSuite` traits so they can be used in cats, spire, algebra, etc. 

Further more, as per links in some of the other issues, they are agnostic to the underlying test framework. And almost there (waiting for some reviews in catalysts) I've also abstracted the common test framework syntax - including equality.

So really I'm saying.... why don't you look at using something that is not quite finished :wink: And I realise that's not an easy thing to ask for, maybe the test framework, like the build, should be spec'ed out a bit first/ discussed, I flesh out my proposols,  whatever. In terms of timescales on my side, < 1 week.

ceedubs:
@InTheNow sorry; I think I don't quite understand what catalysts is. It's a bit light on documentation at the moment :P.

I can understand not wanting to add Scalatest-specific code to laws. However, I don't see why it's a problem to add it to tests for Cats classes. If catalyst's goal is to abstract over test frameworks, then shouldn't it allow for people to use Scalatest? I'm sure I'm missing the point, so forgive me.

InTheNow:
I've just this minute pushed the latest catalysts code, so tomorrow I can look at adding the latest test related code. 

Might be easier/quicker to wait until then :wink: I guess my main reason to bring this up now was just a heads-up

InTheNow:
@ceedubs Heads up for catalysts is just to make sure this PR does not duplicate/override much. If/when you look at catalysts, we can just add to this.

ceedubs:
@InTheNow and I chatted about this. He has me convinced that we should consider using https://github.com/InTheNow/catalysts so that things like `FoldableTests` are easy for people to hook into from multiple test frameworks (scalatest, specs2, etc). Right now there is quite a bit of boilerplate involved in going from laws to tests.

We also decided that this PR probably doesn't make us much worse-off in terms of being able to change to using catalysts (or something like it). And it has some general cleanup that might be useful. So I've merged in the latest master, and I think we should give this another go. Preferably we don't let it hang around too long, because it's pretty prone to merge conflicts since it touches most tests.

@non are you still good with this?

Does anyone else want to give a +1?

InTheNow:
not an official +1, but just confirming that I'm 100% with @ceedubs on this. Also confirming that Cats is no way obliged to use catalysts, just that it might be worth a look.

ceedubs:
What do you think? @non @adelbertc?

adelbertc:
LGTM :+1:

non:
:+1:

PR #555, opened 2015-10-02
Generalize ApplyBuilder to Monoidal structures

julienrf:
I’d like to get your feedback on this work in progress. Here is my plan:

- [x] Introduce a `Monoidal` type class ;
- [x] Generalize `ApplySyntax` to monoidal structures ;
- [x] Add `imap` and `contramap` methods to `MonoidalBuilder` (that’s why this PR is actually useful) ;
- [x] Monoidal laws ;
- [x] Tests ;
- [ ] Documentation.

The motivation is to be able to express function application (for any arity) within a context, like we can currently do with `Apply`, but with contravariant (e.g. printers) and invariant (e.g. codecs) structures.


julienrf:
As you can see, there is a type inference issue: scalac is unable to find a `Monoidal` instance for `State[A, ?]` for a fixed `A`. I tried to play with `Unapply` but was unable to solve this problem. If we introduce a type alias (e.g. `type StateString[A] = State[String, A]`) the existing tests happily pass.

julienrf:
Ok, it seems that I eventually managed to fix the typing problem.

ceedubs:
@julienrf thanks - exciting stuff! It will take me a little while to digest this, but it looks promising. I'm curious what @mpilquist and @non think, as I know they've talked about this recently.

non:
First of all, thanks @julienrf for getting the ball rolling on this!

I do have some design concerns. I think it would be a mistake to introduce a `Monoidal` type class that is independent of `Applicative`, since they have equivalent definitions (i.e. given one of them you can build the other).

I had been imagining (and actually started sketching) a version where `product` was added to `Applicative`, and many definitions prefer to use `product` + `map` instead of `ap`. The nice thing here is that folks who have code with `Applicative` don't have to change much, but that we can rewrite more of our default implementations in terms of `product` and `map` (which I think will be more efficient).

For example:

```scala
def ap[A, B](fa: F[A], ff: F[A => B]): F[B] =
  product(ff, fa).map { case (f, a) => f(a) }
```

Does this make sense? Additionally, I could imagine adding a `Monoidal` type class which extends `Functor` but only provides `product` (i.e. there is no `pure` or equivalent method). This would sit alongside `Apply` as a precursor to `Applicative`:

```scala
trait Monoidal[F[_]] extends Functor[F] {
  def product(fa: F[A], fb: F[B]): F[(A, B)]
}
```

(Disclaimer: I've spent the weekend getting sick, so I may not be thinking clearly. Please feel free to push back against this design or complicate it with issues I haven't though of.)

julienrf:
> I do have some design concerns. I think it would be a mistake to introduce a `Monoidal` type class that is independent of `Applicative`, since they have equivalent definitions (i.e. given one of them you can build the other).

My point is that `Monoidal` + `Functor` is indeed equivalent to `Apply` (ie `ap` can be defined in terms of `product` and _vice versa_), but `Monoidal` is also useful with contravariant and invariant functors, that’s why I think `Monoidal` should **not** be coupled with `Functor`.

Again, the purpose of this PR is to make the current “`ApplyBuilder` syntax” available also for contravariant and invariant structures. For instance, suppose that you have the following contravariant typeclass:

~~~ scala
trait Showable[A] {
  def show(a: A): String
}
~~~

You can implement `Monoidal[Showable]` and `Contravariant[Showable]`:

~~~ scala
implicit val monoidalShowable: Monoidal[Showable] =
  new Monoidal[Showable] {
    def product[A, B](fa: Showable[A], fb: Showable[B]): Showable[(A, B)] =
      new Showable[(A, B)] {
        def show(ab: (A, B)) = {
          val (a, b) = ab
          fa.show(a) ++ " " ++ fb.show(b)
        }
      }
  }

implicit val contravariantShowable: Contravariant[Showable] =
  new Contravariant[Showable] {
    def contramap[A, B](fa: Showable[A], f: B => A): Showable[B] =
      new Showable[B] {
        def show(b: B) = fa.show(f(b))
      }
  }
~~~

Then you can easily build a `Showable[Foo]`, given the following `Foo` definition:

~~~ scala
case class Foo(s: String, i: Int, b: Boolean)
~~~

~~~ scala
implicit val showableFoo: Showable[Foo] =
  (
    Showable[String] |@|
    Showable[Int] |@|
    Showable[Boolean] |@|
  ).contramap(foo => (foo.s, foo.i, foo.b))
~~~

A really useful use case for this would be, in circe, to build JSON codecs for a case class from the JSON codecs of each field of the case class (codecs are invariant structures).

julienrf:
On the other hand, note that my PR changed `Apply` so that now `Apply` extends `Monoidal` (and `product` is implemented in terms of `ap` and `map`). That’s what makes existing code works without any change (modulo some imports renaming).

core/src/main/scala/cats/Apply.scala
```diff
@@ -45,6 +37,9 @@ trait Apply[F[_]] extends Functor[F] with ApplyArityFunctions[F] { self =>
       def F: Apply[F] = self
       def G: Apply[G] = GG
     }
+
+  override def product[A, B](fa: F[A], fb: F[B]): F[(A, B)] = ap(fb)(map(fa)(a => b => (a, b)))
```
julienrf: Note that I’m not actually _overriding_ this method, just _implementing_ it, but simulacrum does not accept my code if I omit the `override` qualifier.

non:
Aha, I see. OK, that totally makes sense. As long as `Applicative` ends up being `Monoidal` I am happy. I'll have to go back and read this more closely.

core/src/main/scala/cats/Apply.scala
non: Would you be willing to reverse this? I.e. leave `product` abstract, and implement `ap` in terms of `product` and `map`? I think the implementation will end up being a bit more efficient.

core/src/main/scala/cats/data/Streaming.scala
```diff
@@ -887,7 +887,7 @@ trait StreamingInstances extends StreamingInstances1 {
       def combine[A](xs: Streaming[A], ys: Streaming[A]): Streaming[A] =
         xs concat ys
 
-      override def map2[A, B, Z](fa: Streaming[A], fb: Streaming[B])(f: (A, B) => Z): Streaming[Z] =
+      override def map2[A, B, Z](fa: Streaming[A], fb: Streaming[B])(f: (A, B) => Z)(implicit ev: Monoidal[Streaming]): Streaming[Z] =
```
non: Why was this change necessary? If you have a `Streaming` in hand you can easily do `map` or `flatMap` without needing a `Monoidal` instance.

core/src/main/scala/cats/package.scala
```diff
@@ -35,7 +35,7 @@ package object cats {
       override def map[A, B](fa: A)(f: A => B): B = f(fa)
       override def ap[A, B](fa: A)(ff: A => B): B = ff(fa)
       override def flatten[A](ffa: A): A = ffa
-      override def map2[A, B, Z](fa: A, fb: B)(f: (A, B) => Z): Z = f(fa, fb)
+      override def map2[A, B, Z](fa: A, fb: B)(f: (A, B) => Z)(implicit ev: Monoidal[Id]): Z = f(fa, fb)
```
non: Relatedly to `Streaming`, why do we need to add this here?

non:
So yeah, everything seems fine except a lot of the added `(implicit ev: Monoidal[...])` code seems unnecessary (and `ev` seems unused). Would it be possible to remove those? Given that the instances in question should be `Monoidal` themselves, passing another one seems unnecessary.

xuwei-k:
What is the Monoidal laws? Maybe can't define laws only Monoidal.

I think
- `Monoidal` + `Functor` = `Apply`
- `Monoidal` + `Contravariant` = [`Divide`](https://github.com/scalaz/scalaz/blob/v7.1.4/core/src/main/scala/scalaz/Divide.scala#L270-L273)

and `Divide` has laws.

non:
@xuwei-k Right. If `Monoidal` abstracts across `Functor`, `Invariant`, and `Contravariant` then the laws need to be specified for each of those combinations.

I wonder if we should add `Divide` to Cats? I like the way it mirrors `Apply`.

julienrf:
> So yeah, everything seems fine except a lot of the added (implicit ev: Monoidal[...]) code seems unnecessary (and ev seems unused). Would it be possible to remove those?

This parameter is needed just because in all these places we _override_ `map2`, which is originally defined in [`Functor`](https://github.com/non/cats/pull/555/files#diff-cb5da707ed3ff9498d040873dd1b511aR70) and takes this implicit `Monoidal[F]` parameter.

My reasoning for pulling up `map2` into `Functor` was that `Monoidal` is sufficient to support the `|@|` syntax, and then all we need to support `mapN` is to know if we have a covariant (`Functor`), contravariant (`Contravariant`) or invariant (`Invariant`) structure. So, `Monoidal` gives you `|@|` and then `Functor` gives you `map`.

Maybe we could find a better design. I need to think a bit about that.

core/src/main/scala/cats/Apply.scala
julienrf: Ok but then I have to update every place in the code where we create instances of `Apply` to implement `product` instead of `ap`. Are you ok with that?

julienrf: Oh, and we have another problem: `Applicative` implements `map` in terms of `ap`, so if I implement `ap` in terms of `map` we hit a infinite recursion.

non: I'm leaving a comment on the overall issue which will (hopefully) clear up where I see this going.

non:
So first of all thanks for bearing with me on this. It's not always obvious how much context one is carrying around in there head until one has to explain it.

One thing we talked about (and agreed on) was that Cats was moving away from putting implementations in the default type class traits. Particularly, we wanted to move await from designs which allow one to implement `Applicative` (or `Monad`) in a terse but inefficient way (i.e. `map` via `pure` and `flatMap`, or `pure` and `ap`). We haven't gone through and made this change everywhere yet, but I was taking it as implicit that this should be part of the design.

Concretely, for your PR, here is how I see it fitting together:

1. Implement all derived methods (like `map2`) in terms of `product` and `map`.
2. Leave `map` entirely abstract on `Applicative`
3. Add `Divide` as per @xuwei-k's suggestion
4. Fix instances to keep them compiling
5. Add custom syntax for `Functor` + `Monoidal`, but leave `map2` off of `Functor` itself

Does this make sense? If it seems too big, I would be happy to try to get the redesign (away from derived methods) pushed through first, which might make it easier for you to make your changes.

Thanks again for your work on this.

julienrf:
Hi, I reworked a bit the PR. Notably, I achieved points 1, 2, 4 and 5.

I kept implementing `product` as a derived method in `FlatMap`, though.

`map2`, `map3`, etc., `contramap2`, `contramap3`, etc., `imap2`, `imap3`, etc. and `tuple2`, `tuple3`, etc. are implemented in `Monoidal` companion object. Then, they are referenced by monoidal syntax (ie. `(fa |@| fb |@| fc).contramap` calls `Monoidal.contramap3`). I also kept the aliases to `map3`, `map4`, etc. and `tuple3`, `tuple4`, etc. in `Apply` (we could remove them, though).

Sorry for all the changes…

Before I fix the tuts, can you confirm me that the direction this work is going looks good to you?

non:
Actually let's hold off on (3) for now. I think that should happen in another issue/PR.

This looks great!

In the long run I'd like to try to deprecate the apply builder syntax (i.e. `|@|`) but let's hold off doing that for now to limit the scope of this PR.

I'm interested to see what everyone else (particularly @mpilquist) thinks of this.

laws/src/main/scala/cats/laws/MonoidalLaws.scala
```diff
@@ -0,0 +1,16 @@
+package cats.laws
+
+import cats.Monoidal
+
+trait MonoidalLaws[F[_]] {
```
mpilquist: Can we define the left/right identity and associativity laws here, but instead of returning `IsEq[X]`, return a tuple, because the first and second elements will have different but isomorphic types?  For example, the left identity law would return a `(F[(Unit, A)], F[A])`. This stresses that this type class is not lawless, even if we need some type of functor in order to do the equality check for most `F`s.

core/src/main/scala/cats/Monoidal.scala
```diff
@@ -0,0 +1,15 @@
...
+
+import simulacrum.typeclass
+
+/**
+ * Monoidal allows us to express uncurried function application within a context,
+ * whatever the context variance is.
+ *
+ * It is worth noting that the couple Monoidal and [[Functor]] is interdefinable with [[Apply]].
+ */
+@typeclass trait Monoidal[F[_]] {
```
mpilquist: Any opinions on renaming this to `Product`?

laws/src/main/scala/cats/laws/MonoidalLaws.scala
mpilquist: We'd then bind these laws in `InvariantTests`, using the `Invariant[F]` instance.

julienrf:
Hi,

I think I’ll have some time to work on this in the next few days this week, so I will hopefully update the PR with checked laws. I’ve started working on that (though I didn’t push that yet) but I must admit that I’m a bit lost.

I tried to define some laws but I’m not sure they are good, and I was unable to formulate useful laws. For instance I didn’t succeed in expressing `Apply` laws with just `Functor` and `Monoidal`. So far, I only have one law that checks that `product` is associative:

~~~ scala
map(product(fa, product(fb, fc))) { case (a, (b, c)) => (a, b, c) } <->
map(product(product(fa, fb), fc)) { case ((a, b), c) => (a, b, c) }
~~~

I also have some design considerations. I’d like to define the laws in terms of `Functor` and `Monoidal` (taken independently) rather than `Apply` (which _extends_ `Functor` and `Monoidal`), but it does not fit with the pattern currently used in the code base where laws are expressed against one `F[_]` algebra that has *one* typeclass instance (e.g. `Apply`) that brings everything. So, instead of that pattern I’d like to define my laws as follow:

~~~ scala
trait CovariantMonoidalLaws[F[_]] {
  implicit def FMonoidal: Monoidal[F]
  implicit def FFunctor: Functor[F]
  // … laws expressed here
}
~~~

My motivation is that we can implement `Monoidal` *independently* of `Functor`, `Contravariant` or `Invariant`, and then benefit from the `|@|` operator and everything else, so our laws should be written without relying on a type that ties together `Monoidal` and `Functor` (or `Contravariant` or `Invariant`).

I’ve seen @mpilquist’s comment suggesting a way to express laws without even needing `Functor`, `Contravariant` or `Invariant`. His solution is interesting, but works only if we are able to define, with just `Monoidal`, all the laws we need. It will work for left and right identity and associativity, at least. Do we want other laws?

mpilquist:
> It will work for left and right identity and associativity, at least. Do we want other laws?
What other laws were you thinking? As far as I know, these 3 are complete.

julienrf:
As suggested [above](#issuecomment-145408607) by @xuwei-k, I thought that we could also “mirror” `Apply` laws for `Functor` + `Monoidal`, and so on for `Contravariant` and `Invariant`.

But I’m happy to enrich this PR with just the 3 laws you mentioned :)

By the way, what makes you think they are “complete”? Do you have some references about that? None of the papers I read about `Monoidal` did mention any laws.

mpilquist:
@julienrf The applicative laws (and apply laws) are free theorems given a lawful `Functor[F] with Monoidal[F]`. The most accessible reference I have for that is Typeclassopedia, though there's no proof: https://wiki.haskell.org/Typeclassopedia#Alternative_formulation.

laws/src/main/scala/cats/laws/MonoidalLaws.scala
julienrf: I need to add `pure` to `Monoidal` to express identity laws.

laws/src/main/scala/cats/laws/discipline/MonoidalTests.scala
```diff
@@ -0,0 +1,55 @@
...
+  object Isomorphisms {
+    import algebra.laws._
+    implicit def covariant[F[_]](implicit F: Functor[F]): Isomorphisms[F] =
+      new Isomorphisms[F] {
+        def `((a, b), c) ≅ (a, (b, c))`[A, B, C](fs: (F[(A, (B, C))], F[((A, B), C)]))(implicit EqFABC: Eq[F[(A, B, C)]]) =
+          F.map(fs._1) { case (a, (b, c)) => (a, b, c) } ?== F.map(fs._2) { case ((a, b), c) => (a, b, c) }
+        def `(unit, a) ≅ a`[A](fs: (F[(Unit, A)], F[A]))(implicit EqFA: Eq[F[A]]) =
+          F.map(fs._1)(_._2) ?== fs._2
+        def `(a, unit) ≅ a`[A](fs: (F[(A, Unit)], F[A]))(implicit EqFA: Eq[F[A]]) =
+          F.map(fs._1)(_._1) ?== fs._2
```
julienrf: I’d like to find better names for the above methods. Any idea?

tests/src/test/scala/cats/tests/CokleisliTests.scala
```diff
@@ -19,6 +19,10 @@ class CokleisliTests extends SlowCatsSuite {
   def cokleisliEqE[F[_], A](implicit A: Arbitrary[F[A]], FA: Eq[A]): Eq[Cokleisli[F, A, A]] =
     Eq.by[Cokleisli[F, A, A], F[A] => A](_.run)
 
+  implicit val iso = MonoidalTests.Isomorphisms.covariant[Cokleisli[Option, Int, ?]]
```
julienrf: For all kinds that are different from `* −> *`, I had to add this line. I’d be happy to find a workaround.

julienrf:
Hi all,

I have a problem: I think I need `unit` to be able to formulate laws.I currently get [arbitrary `F[Unit]` values](https://github.com/julienrf/cats/blob/monoidal-functors/laws/src/main/scala/cats/laws/discipline/MonoidalTests.scala#L26-L27) but obviously that doesn’t work (e.g. an arbitrary `Xor[A, Unit]` can be a `Left[A]`).

Alternatively, I’m considering to manually passing the `F[Unit]` I need, but it seems kind of ugly to me. What do you think?

Maybe we should enrich `Monoidal` to have both `product` and `unit` as defined [here](https://wiki.haskell.org/Typeclassopedia#Alternative_formulation).

non:
I think it makes sense for `Monoidal` to have either `unit` or `pure`.

mpilquist:
@non Even if we are keeping `Apply` as its own type class? I think whatever reasoning supports `Apply` existing applies equally to having `product` without `pure`.

julienrf:
Well, we can leave `Monoidal` without `unit`, but then we are only able to express the associativity law.

mpilquist:
@julienrf Right, which is consistent with the laws for `Apply` and `Bind`. I'm just arguing that if `Apply`/`Bind` exist, then a type class for the `product` operation should probably exist too.

julienrf:
Hi,

I decided not to include `pure` neither `unit` into `Monoidal`. I first tried to include `unit`, but it had cumbersome consequences: in several places we had to require a `Monoid` where we previously just required a `SemiGroup`. When I realized that I had to provide such a `Monoid` for `NonEmptyList`, I gave up.

So, here is a `Monoidal` typeclass with just an associativity law.

core/src/main/scala/cats/Monoidal.scala
mandubian: Why not "Cartesian" because Monoidal sounds different ?

adelbertc:
@julienrf sorry for the lack of chatter - I'm taking a closer look now

I may be missing something but I'm not sure where the `Semigroup` or `Monoid` comes in. It seems we could split `Monoidal` out similar to how we split `Applicative` - one with `pure` and one without. We'll of course figure out what we want to name the one without `pure`. It would look something like

```scala
trait Semigroupal[F[_]] { // yeah i know
  def zip[A, B](fa: F[A], fb: F[B]): F[(A, B)]
}

trait Monoidal[F[_]] extends Semigroupal[F] {
  def pure[A](a: A): F[A]
}
```

The `Semigroupal` laws would just be associativity, and the `Monoidal` laws would have left and right identity, much like `Semigroup` and `Monoid` laws.

julienrf:
Hi, your idea totally makes sense.

The issue I mentionned above with `SemiGroup` and `Monoid` is that if `Monoidal` has this `pure` method and if I want to implement e.g. `Monoidal[Validated]` then I need to implement a `Monoid[NonEmptyList]`. If only implement `SemiGroupal[Validated]` nothing has to be changed, though (i.e. I only need a `SemiGroup[NonEmptyList]`, which is ok).

So, if I find some more time, it could be interesting to follow your suggestion and rename the `Monoidal` I introduced into `SemiGroupal` and then define `Monoidal` as you did.

adelbertc:
Right, so some data types would only be able to define `Semigroupal`, others `Monoidal`

If you are busy these next couple weeks, perhaps we can resolve the merge conflicts and see if we can push what you have here, and then someone else can take over the other work. What are people's thoughts on that?

julienrf:
OK, I just updated my PR to master and resolved the conflicts.

adelbertc:
Pinging @mpilquist and @non - do we want to merge this in the interim and add the version with `unit` later on, or would we prefer a more complete solution?

julienrf:
Hi, keeping my branch free of conflicts is a tedious job, if you are ok with that let’s merge this work as it is and we will improve it later by:

- renaming `Monoidal` to `Semigroupal` ;
- introduce `Monoidal`, which extends `Semigroupal` and adds `pure[A](a: A): F[A]`, and has associativity laws ;
- implement some `Monoidal` instances (e.g. `Const`, etc.).

I can create the ticket to track progress on this path.

milessabin:
I think it's a shame we have to duplicate the `pure` signature. A `Pure` type class, without laws of it's own, but piggybacking on the laws of the type classes that extend it, would be a win here.

mpilquist:
Sorry, I haven't had time to look in detail but I  agree that we should merge this and address any issues subsequently.

adelbertc:
Alright, I'm going to merge this in the interest of getting something up there, and let's address improvements moving forward.

Thanks so much @julienrf , sorry it took a while!

Also, will leave this here for posterity https://gitter.im/non/cats?at=5669e953de5536717680f8b0

stew:
yay

PR #558, opened 2015-10-04
Add laws for Bifunctor

### LINKED ISSUE #557: Proposed laws for Bifunctor
Currently there are no laws associated with Bifunctor, I'd like to propose the following:

Bifunctor Identity:

`fa.bimap(identity, identity) <-> fa`

Bifunctor Composition

`fa.bimap(f1, f2).bimap(g1, g2) <-> fa.bimap(f1 andThen g1, f2 andThen g2)`

agenovese:
Added 2 laws for Bifunctor

BifunctorIdentity enforces that applying the identity function via bimap returns the original value

BifunctorComposition enforces that applying functions via consecutive applications of bimap returns the same result as applying the composition of the functions via a single application of bimap

Fixes #557

### DIFF
--- a/laws/src/main/scala/cats/laws/BifunctorLaws.scala
+++ b/laws/src/main/scala/cats/laws/BifunctorLaws.scala
@@ -1,0 +1,25 @@
+package cats.laws
+
+import cats.functor.Bifunctor
+import cats.syntax.bifunctor._
+
+/**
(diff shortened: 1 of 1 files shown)


agenovese:
Sorry, shouldn't be putting in PRs while half asleep. Will resubmit tomorrow with the tests for Validated, Ior, Xor, and XorT updated.

PR #559, opened 2015-10-04
Introduce Bifunctor Laws

agenovese:
- Create BifuncorLaws
- Create BifunctorTests
- Add Bifunctor instances for Validated, Ior, Xor, and XorT
- Add Bifunctor tests for Validated, Ior, Xor, and XorT

Fixes: 557

### DIFF
--- a/core/src/main/scala/cats/data/Ior.scala
+++ b/core/src/main/scala/cats/data/Ior.scala
@@ -1,6 +1,8 @@
 package cats
 package data
 
+import cats.functor.Bifunctor
+
 /** Represents a right-biased disjunction that is either an `A`, or a `B`, or both an `A` and a `B`.
  *
  * An instance of `A [[Ior]] B` is one of:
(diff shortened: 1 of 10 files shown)


non:
Thanks! This looks good to me.

:+1:

core/src/main/scala/cats/data/Ior.scala
```diff
@@ -142,6 +144,11 @@ sealed abstract class IorInstances extends IorInstances0 {
     def pure[B](b: B): A Ior B = Ior.right(b)
     def flatMap[B, C](fa: A Ior B)(f: B => A Ior C): A Ior C = fa.flatMap(f)
   }
+
+  implicit def bifunctor: Bifunctor[Ior] =
```
ceedubs: Sorry, this is super nitpicky, but could you name this `iorBifunctor` just to match the convention of the instances above it?

ceedubs: Similarly for the instances in the other files.

ceedubs:
Other than my super nitpicky comment, this looks good. :+1:

core/src/main/scala/cats/data/Ior.scala
```diff
@@ -142,6 +144,11 @@ sealed abstract class IorInstances extends IorInstances0 {
...
+    new Bifunctor[Ior] {
+      override def bimap[A, B, C, D](fab: A Ior B)(f: (A) => C, g: (B) => D): C Ior D = fab.bimap(f, g)
```
fthomas: Could you also remove the parens from `f` and `g`? If functions have only one parameter we usually omit them.

fthomas:
Also LGTM.

agenovese:
Fixed, and sorry about those. I should have noticed the naming and I never use parens like that (going to blame the IDE and the fact that it was after 2am).

fthomas:
Merging, thanks!

ceedubs:
@agenovese no worries. What you had was completely fine! I'm just nitpicky :P

PR #562, opened 2015-10-05
Add some helper methods to MonadError

ceedubs:
You should make me write unit tests and ScalaDocs for these before you
merge. I wanted to open a PR before I forgot about this though.

### DIFF
--- a/core/src/main/scala/cats/MonadError.scala
+++ b/core/src/main/scala/cats/MonadError.scala
@@ -2,27 +2,77 @@ package cats
 
 import cats.data.{Xor, XorT}
 
-/** A monad that also allows you to raise and or handle an error value.
-  *
-  * This type class allows one to abstract over error-handling monads.
-  */
+/**
+ * A monad that also allows you to raise and or handle an error value.
+ *
+ * This type class allows one to abstract over error-handling monads.
+ */
 trait MonadError[F[_], E] extends Monad[F] {
+  /**
+   * Lift an error into the `F` context.
+   */
   def raiseError[A](e: E): F[A]
 
-  def handleError[A](fa: F[A])(f: E => F[A]): F[A]
+  /**
+   * Handle any error, potentially recovering from it, by mapping it to an
+   * `F[A]` value.
+   *
+   * @see [[handle]] to handle any error by simply mapping it to an `A`
+   * value instead of an `F[A]`.
+   *
+   * @see [[recoverWith]] to recover from only certain errors.
+   */
+  def handleErrorWith[A](fa: F[A])(f: E => F[A]): F[A]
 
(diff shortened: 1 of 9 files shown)


ceedubs:
Also I should override some of these with optimized versions in some of the instances.

adelbertc:
Awesome possum

ceedubs:
I started to add some laws, but I went down a rabbit hole when realizing that Scalacheck doesn't provide generators for partial functions.

non:
You can always create a `PF[A, B]` from `A => Option[B]` right?

```scala
Arbitrary(arbitrary[A => Option[B]].map { f =>
  new PartialFunction[A, B] {
    def isDefinedAt(a: A): Boolean = f(a).isDefined
    def apply(a: A): B = f(a).get
  })
```

ceedubs:
@non right. There's even a `Function.unlift` method for that. Hence the rabbit hole :P

ceedubs:
I added some `MonadError` laws that hit the methods I've added. Their names probably aren't the greatest. I'm open to naming suggestions.

I'm going to add some optimized overrides, and then this PR will be ready for review (but feel free to comment now).

ceedubs:
As I was documenting some of the `MonadError` methods, I convinced myself that `handleError` should become `handleErrorWith` and there should be a new `handleError` method that takes `f: E => A` instead of `f: E => F[A]`. This makes it line up with `recover` and `recoverWith` (from this PR on MonadError, but currently existing on `Xor`, `XorT`, and `scala.concurrent.Future`. So now I'm down a bit of another rabbit hole. Do people like the idea of that? It doesn't need to be an abstract method, because it can be implemented in terms of `handleErrorWith` and `pure`.

non:
That sounds good to me. I say go for it! We have time to change it if it doesn't end up being good in practice.

ceedubs:
Okay I pushed some more changes.

Many of these "laws" are basically tests for default implementations. I'm not sure whether or not they should exist in their current form. This is something that has been talked about in the past, but I don't think we really reached a verdict on how to go about this. I figured it probably doesn't hurt much to add them for now, and we can always rip them out later.

ceedubs:
> Okay I pushed some more changes.

That didn't make it very clear, but I think this is ready to review, and I don't have any pending changes.

adelbertc:
Woaaa awesome :+1: from me

non:
I 100% support laws for testing default implementations (I think they are great examples of coherence that a user can expect), so +1 from me.

non:
Looks like there is a Scaladoc error here, but if that is cleaned up :+1: from me.

ceedubs:
Build is green again. I'm going to go ahead and merge.

PR #566, opened 2015-10-11
Edit Kleisli intro for brevity

philwills:
I found the Kleisli docs to generally be really helpful, but I thought that they could do with an intro that described their utility a bit more quickly, which is what I've attempted here.


ceedubs:
:+1:

non:
:+1:

PR #568, opened 2015-10-15
Add new shapes to Unapply

stew:
This commit adds Unapply shapes which will match a datum with a type in the
shape `F[X[_], Y]` with a typeclass expecting a `F[_]`

Thanks to @smungee who noticed that apply syntax was not working with
FreeApplicative because of this.


adelbertc:
:+1:

non:
:+1:

PR #570, opened 2015-10-15
Add syntax test for |@| on FreeApplicative

ceedubs:
This is a test that is fixed by #568. I just thought it might be nice to have a regression test for the specific case that was discovered to not work before that PR.

Big thanks to @stew for the fix and @smungee for reporting the issue.


non:
:+1:

fthomas:
:+1:

PR #571, opened 2015-10-15
Add transform and subflatMap to OptionT and XorT

liff:
As per [gitter discussion](https://gitter.im/non/cats?at=561f87d9c1c9c8640b034d75); proposing to add `transform` and `subflatMap` methods to `OptionT` and `XorT`. I suppose they should be in `StreamingT` as well but it's not currently obvious to me how to go about the implementation.

TODO:
- [x] agree on the name :)
- [x] add tests
- [ ] implement for `StreamingT`?

### DIFF
--- a/core/src/main/scala/cats/data/OptionT.scala
+++ b/core/src/main/scala/cats/data/OptionT.scala
@@ -39,6 +39,12 @@ final case class OptionT[F[_], A](value: F[Option[A]]) {
         case None => F.pure(None)
       })
 
+  def transform[B](f: Option[A] => Option[B])(implicit F: Functor[F]): OptionT[F, B] =
+    OptionT(F.map(value)(f))
+
+  def subflatMap[B](f: A => Option[B])(implicit F: Functor[F]): OptionT[F, B] =
+    transform(_.flatMap(f))
+
   def getOrElse(default: => A)(implicit F: Functor[F]): F[A] =
     F.map(value)(_.getOrElse(default))
 
--- a/core/src/main/scala/cats/data/XorT.scala
+++ b/core/src/main/scala/cats/data/XorT.scala
@@ -61,6 +61,12 @@ case class XorT[F[_], A, B](value: F[A Xor B]) {
   def flatMapF[AA >: A, D](f: B => F[AA Xor D])(implicit F: Monad[F]): XorT[F, AA, D] =
     flatMap(f andThen XorT.apply)
 
+  def transform[C, D](f: Xor[A, B] => Xor[C, D])(implicit F: Functor[F]): XorT[F, C, D] =
+    XorT(F.map(value)(f))
+
+  def subflatMap[AA >: A, D](f: B => AA Xor D)(implicit F: Functor[F]): XorT[F, AA, D] =
+    transform(_.flatMap(f))
+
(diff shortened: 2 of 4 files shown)


ceedubs:
Thanks, @liff!

I think you can relax all of the `Monad` constraints to `Functor`.

non:
I'm fine with the names. I agree with @ceedubs that you can relax the constraints. I also think it might be nice to write `value.map(f)` instead of `F.map(value)(f)` but that is somewhat subjective.

I'd say you should proceed with the tests and then we can think about how (or if) to make this work for `StreamingT` :wink:

refried:
`transform` vs `subMap`? Or is there history behind the names?

liff:
Relaxed constraints.

@non, wrt. `value.map(f)`, that would require importing `cats.syntax.functor` and it doesn't seem to be consistent with the other methods around?

@refried, the only history is in the linked gitter conversation :)

ceedubs:
This looks good to me. I think it's good to merge once it gets some tests.

I think the lack of consistency of where we are using syntax and where we aren't is a bigger issue that we can tackle separately. It's made a bit trickier by the fact that Machinist and Simulacrum currently don't play together very nicely. Hopefully [typeclassic](https://github.com/typelevel/typeclassic) saves the day! :)

non:
Sure, let's leave my syntax suggestion out of this.

liff:
Sorry guys, somewhat busy at the moment; going to get back to this later this week.

ceedubs:
@liff no worries; no rush.

liff:
Some tests added. Let me know if that looks sufficient.

Implementation for `StreamingT` still eludes me, unfortunately. It's quite different from these two specimen.

I also feel obliged to add @TomasMikula's [comments about `Hoist`](https://gitter.im/non/cats?at=561ff1d077333bcf6d92c5f7) here, for the record.

TomasMikula:
On that note, both `OptionT` and `XorT` can be made `Hoist`s (reference from scalaz: [MaybeTHoist](https://github.com/scalaz/scalaz/blob/a89b6d63e4ab30cced7178fd813d7c004b5ccb32/core/src/main/scala/scalaz/MaybeT.scala#L177-L188), [EitherTHoist](https://github.com/scalaz/scalaz/blob/a89b6d63e4ab30cced7178fd813d7c004b5ccb32/core/src/main/scala/scalaz/EitherT.scala#L417-L425)) and then `subflatMap` can be implemented in terms of `hoist`.

TomasMikula:
And then so can be [`StreamingT`](https://github.com/scalaz/scalaz/blob/e9b61483036ee3745a984f01048a71d12da9f1d8/core/src/main/scala/scalaz/StreamT.scala#L282-L297).

TomasMikula:
It's still open (at least to me) whether there are interesting monad transformers that are not `Hoist`s but have a meaningful `subflatMap`, but these three are not such.

ceedubs:
:+1:

I could see `Hoist` being handy for this sort of thing in the future, but I think these methods are also useful as they are in this PR.

ceedubs:
Sorry for the hassle @liff, but could you please resolve the merge conflict?

liff:
Done,

ceedubs:
:+1: thanks @liff!

non:
Looks good to me. :+1:

PR #572, opened 2015-10-15
Add FreeApplicative.analyze

smungee:
In trying to port the FreeApplicative example from @jdegoes's ScalaWorld
2015 talk[1] to cats, I found that cats does not have a
FreeApplicative.analyze method, which allows us to compile a
FreeApplicative into a monoid.

This commit adds FreeApplicative.analyze, as well as a new test for
this, as per a discussion on gitter [2].

[1] https://github.com/jdegoes/scalaworld-2015/blob/master/src/main/scala/asmfree.scala#L32
[2] https://gitter.im/non/cats?at=561ed5373a0d354869528194

### DIFF
--- a/free/src/main/scala/cats/free/FreeApplicative.scala
+++ b/free/src/main/scala/cats/free/FreeApplicative.scala
@@ -2,6 +2,7 @@ package cats
 package free
 
 import cats.arrow.NaturalTransformation
+import cats.data.Const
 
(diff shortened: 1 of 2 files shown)


adelbertc:
:+1:

non:
:+1:

non:
Thanks!

PR #575, opened 2015-10-16
Improve the type signature of orElse

jedesah:
The whole point of orElse is to drop the left hand value of a Disjunction, no sense having the type that is dropped influence the resulting type.

Let me know if I am missing something! :-)

### DIFF
--- a/core/src/main/scala/cats/data/Xor.scala
+++ b/core/src/main/scala/cats/data/Xor.scala
@@ -38,8 +38,10 @@ sealed abstract class Xor[+A, +B] extends Product with Serializable {
 
   def getOrElse[BB >: B](default: => BB): BB = fold(_ => default, identity)
 
-  def orElse[AA >: A, BB >: B](fallback: => AA Xor BB): AA Xor BB =
-    fold(_ => fallback, _ => this)
+  def orElse[C, BB >: B](fallback: => C Xor BB): C Xor BB = this match {
+    case Xor.Left(_)      => fallback
+    case r @ Xor.Right(_) => r
+  }
 
   def recover[BB >: B](pf: PartialFunction[A, BB]): A Xor BB = this match {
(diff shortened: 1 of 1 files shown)


ceedubs:
:+1: for the idea and type signature change.

If the original `Xor` is a `Right`, it would be nice to save an allocation by returning that same instance. I don't think `fold` will let you do that (at least not without an `asInstanceOf`), but I think if we replaced the implementation with a `match` we could. Should we give that a try?

jedesah:
@ceedubs Sure!

ceedubs:
:+1: thanks!

core/src/main/scala/cats/data/Xor.scala
```diff
@@ -38,8 +38,10 @@ sealed abstract class Xor[+A, +B] extends Product with Serializable {
 
   def getOrElse[BB >: B](default: => BB): BB = fold(_ => default, identity)
 
-  def orElse[AA >: A, BB >: B](fallback: => AA Xor BB): AA Xor BB =
-    fold(_ => fallback, _ => this)
+  def orElse[C, BB >: B](fallback: => C Xor BB): C Xor BB = this match {
+    case Xor.Left(_)      => fallback
+    case r @ Xor.Right(_) => r
```
stew: is there a difference in 

    case r @ Xor.Right(_) => r

and

    case r => r

?

stew:
:+1:

core/src/main/scala/cats/data/Xor.scala
jedesah: I would be surprised if it did not come at a cost, but who knows, maybe the scala pattern matcher is smarter than I give it credit for. Happy to change it to `r => r`.

non: I think it will essentially correspond to an *else* case, so it seems fine to me.

PR #577, opened 2015-10-17
Rename *Aux helper classes to *Curried

fthomas:
We currently have a few helper classes that have an `Aux` suffix
for emulating currying type parameters. These helper classes have
an `apply` method which allows some type parameters of a function
to be inferred while others need to be specified. @tpolecat has
nice explanation of this trick here:

http://tpolecat.github.io/2015/07/30/infer.html

Unfortunately the suffix `Aux` is an overloaded term that is most
prominently used to provide terser syntax for type refinements, e.g.:

```scala
type Aux[F[_], A, B] = Coyoneda[F, A] { type Pivot = B }
```

I'm wondering if we should use `Curried` instead of `Aux` for classes
that enable type parameter currying. This could make the purpose of
those classes clearer to those who are unfamiliar with this trick.

### DIFF
--- a/core/src/main/scala/cats/data/OptionT.scala
+++ b/core/src/main/scala/cats/data/OptionT.scala
@@ -97,8 +97,8 @@ object OptionT extends OptionTInstances {
   /**
    * Transforms an `Option` into an `OptionT`, lifted into the specified `Applicative`.
    *
-   * Note: The return type is a FromOptionCurried[F], which has an apply method on it, allowing
-   * you to call fromOption like this:
+   * Note: The return type is a FromOptionPartiallyApplied[F], which has an apply method
+   * on it, allowing you to call fromOption like this:
    * {{{
    * val t: Option[Int] = ...
    * val x: OptionT[List, Int] = fromOption[List](t)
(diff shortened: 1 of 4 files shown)


ceedubs:
I don't have any very strong opinions on this, but it seems reasonable to me :+1:

milessabin:
I've used `Aux` as a suffix for this purpose, but I agree that it is very overloaded, and the primary use is, as @fthomas says, as terser syntax for refinements.

I think that this is about partial application rather than currying however, so maybe `Partial` or `PartiallyApplied` as a suffix? It's not expected that a user would have to name these types explicitly, so a little verbosity shouldn't be an issue.

travisbrown:
I've used `-Helper` or just `-er` for these things, in an attempt to make it clear that they're just a type inference hack and have nothing interesting going on, but I like `-PartiallyApplied` a lot better. I definitely agree that `-Aux` should be changed.

fthomas:
:+1: for `PartiallyApplied`

ceedubs:
:+1:

stew:
:+1:

julienrf:
Conversely, do you think we should rename `Aux` to `Uncurried`?

PR #578, opened 2015-10-18
Test the Contravariant instance of Function1

fthomas:


ceedubs:
:+1: wonderful thank you!

This makes me realize that we are missing a `Contravariant` instance for `Kleisli`, aren't we? Are there any other types that come to mind that should have a `Contravariant` instance?

ceedubs:
I think I've found a few. I'll open a separate issue.

non:
:+1:

PR #583, opened 2015-10-26
Add test of Bimonad laws for Eval

mikejcurry:
* Adds basic law checking for the Eval Bimonad instance.
* Adds a check that the instance is also serializable

Additionally, in Function Tests
* Changes serializable test for Bimonad[Function0] to use Bimonad instance instead of Comonad instance.

Let me know if these make sense or if some additional checks would be better.

Also, just kind of learning the discipline stuff, so if I’ve misunderstood anything about how they work let me know.

Also, my commits got a bit messy and reverted. I can tidy them if needed. It is really just the last two commits - as the middle two commits were reverts.

### DIFF
--- a/tests/src/test/scala/cats/tests/EvalTests.scala
+++ b/tests/src/test/scala/cats/tests/EvalTests.scala
@@ -2,7 +2,8 @@ package cats
 package tests
 
 import scala.math.min
-import cats.laws.discipline.{BimonadTests, SerializableTests}
+
+// TODO: monad laws
 
 class EvalTests extends CatsSuite {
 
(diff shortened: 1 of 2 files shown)


ceedubs:
Thanks a bunch @mikejcurry!

These changes look great to me. Like you said, the git history is a bit messy for a relatively small change set. Would you be willing to open up a PR from a different branch that has a cleaner git history?

mikejcurry:
Sure, no problem. I'll create a clean branch and reapply the changes and send a fresh PR.

mikejcurry:
#584 created to replace this with cleaner commit history. Thanks!

PR #584, opened 2015-10-26
Add test of Bimonad laws for Eval

mikejcurry:
Recreating PR: #583 from a fresh branch to clean commit history.

* Adds some law checking for the Eval Bimonad instance.
* Adds a check that the instance is also serializable

Additionally, in Function Tests
* Changes serializable test for Bimonad[Function0] to use Bimonad instance instead of Comonad instance.

Let me know if these make sense or if some additional checks would be better.


non:
:+1: Thanks!

ceedubs:
:+1: good to merge if/when the build goes green

PR #586, opened 2015-10-26
Add sublime files to gitignore.

coltfred:
Just trying to make it less likely people will commit sublime "stuff".


adelbertc:
:+1:

Issue #585, opened 2015-10-26
Add syntax for Validated


coltfred:
We should support 
```
.valid
.invalid
.invalidNel
```
as syntax to lift values into Validated objects.

This should be added by importing `cats.syntax.validated._`.

coltfred:
I'll start on this and should have a PR ready shortly.

ceedubs:
:+1:

It would probably be nice to have `toValid` and `toValidNel` (and maybe `toInvalid` and `toInvalidNel`?) added to `OptionOps` if you are feeling ambitious :)

coltfred:
@ceedubs Correct. I'll do that.

coltfred:
This has been addressed and merged. Closing.

PR #587, opened 2015-10-27
Add syntax for validated.

coltfred:
Add syntax to values for lifting them into `Valid`, `Invalid` and the Nel versions of those. Currently no tests as I was unable to figure out which mechanism was preferred for testing syntax.

Let me know how that should be done.

Addresses #585.

### DIFF
--- a/core/src/main/scala/cats/syntax/all.scala
+++ b/core/src/main/scala/cats/syntax/all.scala
@@ -30,3 +30,4 @@ trait AllSyntax
     with StrongSyntax
     with TraverseSyntax
     with XorSyntax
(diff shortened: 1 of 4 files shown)


ceedubs:
:+1:

I would vote for merging this and creating a separate issue for testing syntax. And possibly another one for whether we should be using `Eval` instead of by-name params for these?

ceedubs:
By the way, thanks @coltfred!

julien-truffaut:
:+1: I agree with @ceedubs we should discuss `Eval` and test in a separate issue

coltfred:
No problem!

PR #588, opened 2015-10-27
Add IdT, the identity monad transformer

vikraman:


vikraman:
I don't see a `MonadTrans` yet, should that be added?

ceedubs:
Thanks, @vikraman!

I've never used `IdT` before. Do you have an example of how/why one might want to use it?

xuwei-k:
I think `IdT` should be an own data type, not type alias.

vikraman:
Yes, I just ran into diverging implicits with this alias, I'll change it to a proper data type.

vikraman:
Updated.

refried:
Ditto to @ceedubs's question

core/src/main/scala/cats/data/IdT.scala
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+ * `IdT[F[_], A]` is the identity monad transformer.
+ */
+final case class IdT[F[_], A](value: F[A]) {
+
+  def map[B](f: A => B)(implicit F: Functor[F]): IdT[F, B] =
+    IdT(F.map(value)(f))
+
+  def flatMap[B](f: A => IdT[F, B])(implicit F: Monad[F]): IdT[F, B] =
```
julien-truffaut: should be replace the constraint from `Monad` to `FlatMap`?

vikraman:
I think it is useful to have `IdT` since it is the most trivial transformer. It would be useful to have the definitions of `liftM` once there is a `MonadTrans` class.

xuwei-k:
https://github.com/xuwei-k/cats/commit/5b426cc249bd1f46d92b8d2

```
[error] /home/travis/build/xuwei-k/cats/tests/src/test/scala/cats/tests/IdTTests.scala:10: ambiguous implicit values:
[error]  both method idTFunctor in class IdTInstances of type [F[_]](implicit F: cats.Functor[F])cats.Functor[[β]cats.data.IdT[F,β]]
[error]  and method idTMonad in class IdTInstances of type [F[_]](implicit F: cats.Monad[F])cats.Monad[[β]cats.data.IdT[F,β]]
[error]  match expected type cats.Functor[[β]cats.data.IdT[F,β]]
[error]   def idTFunctor[F[_]: Monad] = implicitly[Functor[IdT[F, ?]]]
[error]                                           ^
```

vikraman:
Split the instances to avoid ambiguous implicits.

non:
Sorry for the long silence.

I'm fine with merging this -- I do think we probably want to write a few more words about how it works (e.g. "If M[_] is a monad, then IdT[M, ?] provides an equivalent monad" or similar).

:+1:

julien-truffaut:
LGTM 👍

PR #590, opened 2015-10-28
Add some Contravariant instances

### LINKED ISSUE #579: Missing Contravariant instances
There are a number of types that we are either missing `Contravariant` instances or are not being tested by the `ContravariantTests` introduced in #578.

I think the following fit into this category: `Kleisli`, `Const`, `Show`, `Equal`, `Order`, `StateT`. There could very possibly be more.

We should add these instances and tests.

https://github.com/non/cats/pull/578/files#diff-5a256f6c87581d09dae79f590a0a91e4R23 should serve as a good example of testing a type's `Contravariant` instance.

vikraman:
This is an attempt to fix #579

### DIFF
--- a/core/src/main/scala/cats/data/Kleisli.scala
+++ b/core/src/main/scala/cats/data/Kleisli.scala
@@ -2,7 +2,7 @@ package cats
 package data
 
 import cats.arrow.{Arrow, Choice, Split}
-import cats.functor.Strong
+import cats.functor.{Contravariant, Strong}
 
 /**
  * Represents a function `A => F[B]`.
@@ -20,6 +20,9 @@ final case class Kleisli[F[_], A, B](run: A => F[B]) { self =>
 
   def map[C](f: B => C)(implicit F: Functor[F]): Kleisli[F, A, C] =
     Kleisli(a => F.map(run(a))(f))
+
+  def contramap[C](f: C => B)(implicit F: Contravariant[F]): Kleisli[F, A, C] =
+    Kleisli(a => F.contramap(run(a))(f))
 
   def mapF[N[_], C](f: F[B] => N[C]): Kleisli[N, A, C] =
     Kleisli(run andThen f)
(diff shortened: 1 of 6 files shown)


vikraman:
The CovariantTests don't work because of the Eq requirement.

core/src/main/scala/cats/data/Kleisli.scala
```diff
@@ -21,6 +21,9 @@ final case class Kleisli[F[_], A, B](run: A => F[B]) { self =>
   def map[C](f: B => C)(implicit F: Functor[F]): Kleisli[F, A, C] =
     Kleisli(a => F.map(run(a))(f))
 
+  def contramap[C](f: C => B)(implicit F: Contravariant[F]): Kleisli[F, A, C] =
+    Kleisli(a => F.contramap(run(a))(f))
+
```
ceedubs: I think the `contramap` that I would expect for `Kleisli` would be the `local` method that already exists on it. I think in practice the `F` in a `Kleisli` is often going to be covariant (not contravariant) in nature (`Future`, `Task`, `Xor`, etc), so I don't know how often this would be helpful. If you have some good examples of how you might use this, I'd definitely be interested in them.

Otherwise, should we get rid of this one and use `local` for the `Contravariant` instance?

tests/src/test/scala/cats/tests/ConstTests.scala
```diff
@@ -31,4 +32,10 @@ class ConstTests extends CatsSuite {
   checkAll("Const[Map[Int, Int], String]", OrderLaws[Const[Map[Int, Int], String]].eqv)
   checkAll("PartialOrder[Const[Set[Int], String]]", OrderLaws[Const[Set[Int], String]].partialOrder)
   checkAll("Order[Const[Int, String]]", OrderLaws[Const[Int, String]].order)
+
+  {
+    implicit val constContravariant = Const.constContravariant[Int]
```
ceedubs: Was this necessary to get the law-checking to compile? It seems like we shouldn't have to do this.

ceedubs:
Thanks a bunch for working on this, @vikraman!

ceedubs:
For the missing `Eq` instances, I think you could add something like this to cats/laws/src/main/scala/cats/laws/discipline/Eq.scala:

```scala

  implicit def kleisliEq[F[_], A:Arbitrary, B](implicit FB: Eq[F[B]]): Eq[Kleisli[F, A, B]] =
    function1Eq[A, F[B]].on(_.run)
```

core/src/main/scala/cats/data/Kleisli.scala
vikraman: Yes, that does seem more useful. Is there a way to keep both the instances?

ceedubs: For simplicity, I'd be inclined to just keep the one until someone runs into a compelling use-case for the other.

ceedubs:
I believe this has been made obsolete by #626. I'm going to go ahead and close this. @vikraman please let me know if I'm missing something.

Issue #591, opened 2015-10-30
fromTryCatch shouldn't infer Null


ceedubs:
If you don't explicitly provide a type parameter to `fromTryCatch` (on either `Xor` or `Validated`), the type parameter defaults to `Null` and it doesn't actually catch exceptions.

```scala
scala> Xor.fromTryCatch(throw new RuntimeException("blargh"))
java.lang.RuntimeException: blargh
  at $anonfun$1.apply(<console>:17)
  at $anonfun$1.apply(<console>:17)
  at cats.data.XorFunctions$FromTryCatchAux.apply(Xor.scala:229)
  ... 43 elided

scala> Xor.fromTryCatch
res1: cats.data.Xor.FromTryCatchAux[Null] = cats.data.XorFunctions$FromTryCatchAux@1039a2fc

scala> Xor.fromTryCatch[Throwable](throw new RuntimeException("blargh"))
res2: cats.data.Xor[Throwable,Nothing] = Left(java.lang.RuntimeException: blargh)
```

I think we should do two things here:

1) Add a helper to catch `NonFatal` exceptions, as that is probably generally what people want. This could be called `fromTryCatchNonFatal` or just `catchNonFatal` if the former is too verbose.
2) Make `fromTryCatch` take implicit evidence that the type parameter is not an unhelpful type such as `Nothing` (do we need to cover `Null` too?). Scalaz does this: https://github.com/scalaz/scalaz/blob/series/7.2.x/core/src/main/scala/scalaz/NotNothing.scala

ceedubs:
Fixed in #592

PR #592, opened 2015-10-31
Don't infer Null in fromTryCatch

ceedubs:
This is the start of a fix for #591. So far I've only change `Xor`, because I wanted to run this approach by people. If people like this approach, then I can also update `Validated` and try to add some more tests to this PR.

I also renamed `fromTryCatch` to `catching`, because it's a bit less verbose and I like the sound of it a little more. I can change it back if people would prefer though.

### DIFF
--- a/core/src/main/scala/cats/NotNull.scala
+++ b/core/src/main/scala/cats/NotNull.scala
@@ -1,0 +1,24 @@
+package cats
+
+/**
+ * An instance of `NotNull[A]` indicates that `A` does not have a static type
+ * of `Null`.
+ *
+ * This can be useful in preventing `Null` from being inferred when a type
+ * parameter is omitted.
+ */
+sealed trait NotNull[A]
+
+object NotNull {
+  /**
+   * Since NotNull is just a marker trait with no functionality, it's safe to
+   * reuse a single instance of it. This helps prevent unnecessary allocations.
+   */
+  private[this] val singleton: NotNull[Any] = new NotNull[Any] {}
+
+  implicit def ambiguousNull1: NotNull[Null] = throw new Exception("An instance of NotNull[Null] was used. This should never happen. Both ambiguousNull1 and ambiguousNull2 should always be in scope if one of them is.")
+
+  implicit def ambiguousNull2: NotNull[Null] = ambiguousNull1
+
+  implicit def notNull[A]: NotNull[A] = singleton.asInstanceOf[NotNull[A]]
+}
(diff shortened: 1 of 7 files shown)


ceedubs:
By the way, this is what happens now if you omit the type parameter on `catching`:

```
scala> Xor.catching(throw new RuntimeException("blargh"))
<console>:14: error: ambiguous implicit values:
 both method ambiguousNull1 in object NotNull of type => cats.NotNull[Null]
 and method ambiguousNull2 in object NotNull of type => cats.NotNull[Null]
 match expected type cats.NotNull[Null]
       Xor.catching(throw new RuntimeException("blargh"))
```

ceedubs:
Looking at that error message, I'm a bit tempted to hack around the fact that you can't do something like `implicitNotFound` for ambiguous implicit values and encode a helpful message in the function names instead of just using `ambigrousNull1` and `ambiguousNull2` :)

non:
Using descriptive method names there sounds smart.

non:
Also, I feel like we talked about this before, but how would you feel about a simpler variation that was something like this:

```scala
def catching(body: => A): Xor[Throwable, A] =
  try { Right(body) } catch { case e: NonFatal => Left(e) }
```

Often when working with exceptions I don't care about the type, and the syntax is a lot nicer if you can leave off the annotations.

refried:
@ceedubs: you’ve probably seen this, but:

```
implicitAmbiguous annotation

The @implicitAmbiguous annotation allows customizing the error message when an implicit search finds multiple ambiguous values. Refer to the Scaladoc of the annotation class for an example.

Thanks to @puffnfresh for this contribution!
```
in scala 2.12.0-M3

ceedubs:
@refried @puffnfresh I didn't know about `@implicitAmbiguous`. That's wonderful!

@non is you preference for `catchAll` because just because it's more concise than `catchingNonFatal`? I would lean toward being consistent with `catching` vs `catch`, and I have a slight leaning toward `NonFatal` as more descriptive than `All`, but I don't have particularly strong feelings about this, so I'm fine with changing it.

ceedubs:
I've updated my PR. What do people think?

Specifically @non and @mpilquist, since both of you discussed this with me.

Also big thanks to @tpolecat because tut caught several places where I had neglected to update the documentation after these changes :)

non:
This seems fine to me. :+1: 

(I think I'd prefer `catchAll` but I can live with `catchNonFatal` -- it reads pretty well in an imperative way: "catch non-fatal exceptions")

ceedubs:
@non for what it's worth, [this](https://gitter.im/non/cats?at=5638b4a4a530033014e3ce6a) is the conversation that led me to stick with `catchNonFatal` instead of `catchAll`. If you feel strongly about this, I'm fine with changing it though.

PR #593, opened 2015-11-02
Added streaming.tut, a tutorial on cats.data.Streaming.

non:
This tutorial mostly just shows some uses of Streaming[A].
It doesn't go as deeply as it could into the relationship
with Eval[A], or with things like Monad[A] and Traverse[A].

It also adds the Streaming syntax to implicits, and
renames concat to ++ for consistency.

This pull request replaces #548. Review by @ceedubs et al.

### DIFF
--- a/docs/src/main/tut/streaming.md
+++ b/docs/src/main/tut/streaming.md
@@ -25,10 +25,10 @@ library) or `Streaming` (the specific type from Cats).
 ## Introduction
 
 A non-empty `Streaming` instance is structured like a `List`: it has a
-*cons* cell containing a single value, as well as a reference to a tail
-which will the subsequent values (if any). This means that adding
-values to the beginning is very efficient, whereas adding values to the
-end is potentially expensive.
+*cons* cell containing a single value, as well as a reference to a
+tail which will calculate the subsequent values (if any). This means
+that adding values to the beginning is very efficient, whereas adding
+values to the end is potentially expensive.
 
 The major difference between `List` and `Streaming` is evaluation.
 `List` is strict: this means that if you have an instance of `List`
--- a/docs/src/main/tut/streaming.md
+++ b/docs/src/main/tut/streaming.md
@@ -24,7 +24,7 @@ library) or `Streaming` (the specific type from Cats).
 
 ## Introduction
 
-A non-empty `Streaming` instances is structured like a `List`: it has a
+A non-empty `Streaming` instance is structured like a `List`: it has a
 *cons* cell containing a single value, as well as a reference to a tail
 which will the subsequent values (if any). This means that adding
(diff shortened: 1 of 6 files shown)


refried:
This is great!  Are all of the tutorials as good?
(disclaimer: I should admit I tuned out at "gcd")

docs/src/main/tut/streaming.md
```diff
@@ -0,0 +1,432 @@
...
+val t = pi.drop(1).take(40)
+str(h) + "." + str(t) + "..."
+```
+
+## Conclusion
+
+Lazy, immutable streams are a powerful way to model an in-progress
+calculation, especially when those sequences are potentially
+unbounded. While these examples were based on mathematical problems,
+streams are a great way to model ny case where waiting to collect all
```
fthomas: s/ny/any/

fthomas:
That was a great read! :+1:

While reading, I noted that `Streaming.infinite` has the same signature as `Stream.iterate`. Would it makes sense to add an alias with this name for consistency with the std library?

non:
I don't really like aliases (except for textual/symbolic pairs). I am not a huge fan of `.iterate` but I would be willing to rename `.infinite` to `.iterate` if you think it would be better.

fthomas:
My opinion on this is not that strong. Let's leave `.infinite` as it is.

docs/src/main/tut/streaming.md
```diff
@@ -0,0 +1,432 @@
...
+Sometimes the `Streaming` documentation will refer to a *stream*. In
+these cases, we will use lowercase *stream* as a general term,
+distinguished from `Stream` (the specific type from the standard
+library) or `Streaming` (the specific type from Cats).
+
+## Introduction
+
+A non-empty `Streaming` instance is structured like a `List`: it has a
+*cons* cell containing a single value, as well as a reference to a tail
+which will the subsequent values (if any). This means that adding
```
mikejcurry: There is something missing here. Maybe the word "contain" or the word "hold" or something. I'm not quite sure what the best word is to describe for Streaming, but the sentence as stands is missing something.

non: Good catch, thanks! I think the meaning I was going for was something like

> as well as a reference to a tail which will calculate the subsequent values (if any).

mikejcurry: Makes sense to me - rest of this all looks fantastic!

adelbertc:
Awesome! :+1: from me

non:
Seems like folks are OK with this so I'll merge it now

FWIW, I think there is still a lot of room to make this better, so others shouldn't feel bad about updating/rewriting it if they are inspired.

PR #596, opened 2015-11-03
Add two new laws for Bimonad.

### LINKED ISSUE #595: review bimonad laws
Kind of random ... @jedws just referenced [this paper](http://arxiv.org/pdf/0710.1163v3.pdf) (pdf link) which describes an entwining property of bimonads that may or may not be captured in the current laws. It looks like it goes beyond the point/extract symmetry but I didn't try to make sense of it. Mostly I just like the word "entwining".

non:
These laws are taken from this paper:
http://arxiv.org/pdf/0710.1163v3.pdf

Thanks to @jedws and @tpolecat for bringing this up.

We may also want to remove one of our existing
laws (F.pure(F.extract(fa)) <-> fa), but for now
I have left it alone.

This fixes #595

### DIFF
--- a/laws/src/main/scala/cats/laws/BimonadLaws.scala
+++ b/laws/src/main/scala/cats/laws/BimonadLaws.scala
@@ -3,6 +3,9 @@ package laws
 
 /**
  * Laws that must be obeyed by any `Bimonad`.
+ *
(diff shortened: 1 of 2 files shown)


adelbertc:
:+1:

fthomas:
:+1:

PR #597, opened 2015-11-03
Remove Bimonad law: F.pure(F.extract(fa)) <-> fa

fthomas:
This picks up the question @non asked in https://github.com/non/cats/issues/595#issuecomment-153259535 and removes one of the `Bimonad` laws: `F.pure(F.extract(fa)) <-> fa`. @milessabin noted on Gitter that this law makes any `Bimonad` isomorphic to `Id` (modulo evaluation order) and I think that makes `Bimonad` less useful. For example, `NonEmptyList` which is both monadic and comonadic isn't a lawful `Bimonad` because of the above law and I think it could be a lawful `Bimonad` if we remove that law (but I haven't proved that).

If people find this law useful I'd suggest introducing another type class which extends `Bimonad` and has this additional law.

### DIFF
--- a/laws/src/main/scala/cats/laws/BimonadLaws.scala
+++ b/laws/src/main/scala/cats/laws/BimonadLaws.scala
@@ -13,9 +13,6 @@ trait BimonadLaws[F[_]] extends MonadLaws[F] with ComonadLaws[F] {
   def pureExtractIsId[A](a: A): IsEq[A] =
     F.extract(F.pure(a)) <-> a
 
-  def extractPureIsId[A](fa: F[A]): IsEq[F[A]] =
-    F.pure(F.extract(fa)) <-> fa
-
   def extractFlatMapEntwining[A](ffa: F[F[A]]): IsEq[A] =
     F.extract(F.flatten(ffa)) <-> F.extract(F.map(ffa)(F.extract))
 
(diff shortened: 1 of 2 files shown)


non:
:+1:

If we do end up needing that type, we could call it `IdMonad` or something.

adelbertc:
:+1: from me. I remember @milessabin had objections to it but not sure what happened after, so I'll let someone else with more context merge

ceedubs:
@adelbertc (or @milessabin), are you saying that Miles had objections to this law, or objections to the removal of this law?

adelbertc:
The latter I think

non:
I think Miles wanted to keep this law. I don't want to put words in his mouth, but I think part of his concern was that we only had two laws and it seems arbitrary/asymmetrical to remove one of them.

refried:
I haven’t seen the discussion myself or closely followed the tickets, but I thought we were heading towards removing it because we didn’t have a source defining it was a real law and felt it was arbitrary to have arbitrary laws, even ones like this which are reasonable at a glance. Hope this helps. :)

non:
@refried Thats right, although now we have a good source defining bimonads. See https://github.com/non/cats/pull/596 for details.

non:
See also http://ncatlab.org/nlab/show/Hopf+monad

ceedubs:
Sounds good to me :+1:. It might be nice to give @milessabin a chance to weigh in though.

fthomas:
For posterity: https://gitter.im/non/cats/archives/2015/10/25 is the Gitter discussion I mentioned earlier.

ceedubs:
I think Miles is distracted with Shapeless 2.3.0 and Kittens at the moment :)

Since this has a bunch of :+1:s and through the glory of source control we can always revert it, I'm going to go ahead and merge.

PR #292, opened 2015-04-25
Swap order of G[_] and A in Applicative.sequence

ceedubs:
Fixes #285.

Scalaz has a style convention that kinds always go before types in
method signatures. Should we adopt the same convention?

PR #598, opened 2015-11-04
Swap type parameter order in OneAnd

ceedubs:
This is a delayed follow up to #292 which sets a precedent of a
preference for higher-kinded types coming first in type parameters.


non:
Makes sense to me. :+1:

adelbertc:
:+1:

PR #600, opened 2015-11-06
Add Show for Option and OptionT

ceedubs:
The OptionT Show instance just uses the Show instance for the wrapped
`F[Option[A]]` as opposed to wrapping it in `OptionT(...)`. This is to
match the current behavior of `XorT`, but I'm wondering if that's
something we would want to change on both. It seems to me like the Show
output should reflect the fact that the value is wrapped in a monad
transformer, but I don't have particularly strong feelings about this.


mpilquist:
:+1:

non:
:+1:

PR #601, opened 2015-11-06
Add Show[List]

ceedubs:


core/src/main/scala/cats/std/list.scala
```diff
@@ -70,6 +71,11 @@ trait ListInstances extends ListInstances1 {
 
   implicit def listAlgebra[A]: Monoid[List[A]] = new ListMonoid[A]
   implicit def listOrder[A: Order]: Order[List[A]] = new ListOrder[A]
+
+  implicit def listShow[A:Show]: Show[List[A]] =
+    new Show[List[A]] {
+      def show(fa: List[A]): String = s"List(${fa.map(_.show).mkString(", ")})"
```
fthomas: If you're already using `mkString`: `fa.map(_.show).mkString("List(", ", ", ")")`

ceedubs: Oh I forgot you could do that. Thanks will do.

ceedubs: To save you a comment: I'll do that in the Set PR too :)

mpilquist:
:+1:

adelbertc:
:+1:

PR #276, opened 2015-04-09
Miscellaneous cleanup to Xor

mpilquist:
- Removes `Xor#filter`, as the dependency on a `Monoid[A]` is ad-hoc
 - Adds `orElse`, `valueOr`, and `ensure` methods to `Xor`
 - Adds `Xor.fromOption`
 - Miscellaneous ScalaDoc
 - Miscellaneous formatting cleanup

PR #605, opened 2015-11-07
Remove Validated.filter

ceedubs:
I think this was leftover from copy/paste from `Xor`. It mentions that
it's useful when using `Validated` in for-comprehensions, but you can't
actually use `Validated` in for-comprehensions since it doesn't have a
`flatMap` method :).

Also, `Xor` no longer has a `filter` method. It was removed in #276
since using the monoid zero when the predicate doesn't match is rather
arbitrary and can lead to surprises. I think we should follow that
precedent for `Validated`.


non:
:+1:

adelbertc:
:+1:

PR #606, opened 2015-11-07
Remove EE >: E constraint from Validated.orElse

ceedubs:
This is the same as #575 but for Validated instead of Xor.


adelbertc:
:+1:

non:
:+1:

PR #610, opened 2015-11-07
XorT.toOption returns an OptionT

ceedubs:
This makes `toOption` on `XorT[F, A, B]` return `OptionT[F, B]` instead
of `F[Option[B]]`. This makes it symmetrical with `toRight` and `toLeft`
on `OptionT`. My assumption here is that if you are working with `XorT`
instead of `F[Xor...]` that you would also prefer to work with `OptionT`
rather than `F[Option[...]]`. And if not, it's easy enough to call
`.value`.

### DIFF
--- a/core/src/main/scala/cats/data/XorT.scala
+++ b/core/src/main/scala/cats/data/XorT.scala
@@ -39,7 +39,7 @@ case class XorT[F[_], A, B](value: F[A Xor B]) {
 
   def toEither(implicit F: Functor[F]): F[Either[A, B]] = F.map(value)(_.toEither)
 
-  def toOption(implicit F: Functor[F]): F[Option[B]] = F.map(value)(_.toOption)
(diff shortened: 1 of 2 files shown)


frosforever:
As a user, I am super on board with this. There's yet to be a time where I didn't want `OptionT` from `XorT`.
Would it make sense to name it explicitly `toOptionT` instead?

ceedubs:
@frosforever I considered `toOptionT`. The reason I didn't go forward with it was that then it seemed like `toRight` and `toLeft` on `OptionT` should be named something else. And `toRightT` didn't quite seem right. I'm definitely open to suggestions here.

adelbertc:
:+1: from me

frosforever:
@ceedubs fair enough, that would be a mess. The types should make it clear what it's returning anyway.

mpilquist:
:+1:

Issue #496, opened 2015-08-29
instance/syntax traits and binary compatibility


ceedubs:
Currently we have quite a few places where we have traits that are neither sealed nor private that house instances and syntax. For example the [`Option` instances](https://github.com/non/cats/blob/master/core/src/main/scala/cats/std/option.scala) and the [`Option` syntax](https://github.com/non/cats/blob/master/core/src/main/scala/cats/syntax/option.scala).

Adding a new instance or method to these traits is a change that is not binary compatible. Ideally, we could freely add new instances or syntax and not have to worry about waiting until the next minor version to merge them.

Personally I know that I always get my syntax/instances via imports as opposed to mixing in traits. How much do other people value being able to mix these in instead of importing them?

travisbrown:
How would you support à la carte vs. `cats.syntax.all._` if these were objects?

I may have missed some discussion about this, but the project is young enough that it seems like it might be better to bump the minor version after every release (and save patch versions for really bad mistakes) instead of distorting code to ensure 100% binary compatibility (which we explicit _don't_ guarantee yet, anyway).

ceedubs:
@travisbrown I was envisioning the traits being `private[cats]`, which as far as I know would alleviate the binary compatibility issue (please correct me if I'm wrong).

I agree that bumping the minor version is fine for now. This is something I was aiming to address by a first 1.0 RC.

travisbrown:
@ceedubs Oh, got it. I wouldn't be strongly opposed to making them `private[cats]`. (Still not sure being able to add syntax in a patch version is an especially good thing anyway, though, but I think I'm more of a syntax skeptic than most people.)

ceedubs:
I think I'm more interested in being able to add type class instances without waiting until the next minor version. For example "oops we defined the monad for `FooT[F, ?]` if `F` has a monad, but we neglected to define the functor for `FooT[F, ?]` if `F` has a functor but not a monad. In the past it's been annoying for me to have to wait 6 months for the next version of scalaz for something like that.

ceedubs:
@travisbrown at one point on Gitter you mentioned that you've grown fond of having the instance traits exposed. However you said you'd wait until you weren't on your phone to expand upon why. Do you want to weigh in on this issue with your use-case?

travisbrown:
@ceedubs It's partly just a vague dislike for introducing type class instances with imports, and I won't try to defend that.

[...]

PR #612, opened 2015-11-07
Make a bunch of instances traits/classes private/sealed

ceedubs:
This is related to #496, but it's a broader issue than just this change.

These instances traits/classes are used internally for code organization
and implicit prioritization, but they aren't really meant to be part of
the public API. Making them private/sealed may be helpful for binary
compatibility in the future. It also may be a helpful indicator to
newcomers which traits/classes are actually meant for consumption, so
they can ignore some of the noise.

So far, I have left the traits for orphans from `std` (such as
`StringInstances`), and I have left the omni-instance traits exposed.

### DIFF
--- a/core/src/main/scala/cats/Eval.scala
+++ b/core/src/main/scala/cats/Eval.scala
@@ -271,7 +271,7 @@ object Eval extends EvalInstances {
   }
 }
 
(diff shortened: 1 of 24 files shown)


adelbertc:
:+1:

mpilquist:
:+1:

PR #618, opened 2015-11-10
Release notes for 0.3.0

travisbrown:
I just wrote up a draft of some release notes (rendered version [here](https://github.com/travisbrown/cats/blob/topic/0.3.0-release-notes/CHANGES.md)). I didn't do much synthesizing, but it seems like the nature of this release is that there are a lot of little improvements and additions that don't really fit into a bigger story.

Let me know if I'm missing anything important, or if we want to aim for less detail and more of a narrative.

Also @non, I don't see a `v0.3.0` tag, so I'm not sure exactly when you cut the release yesterday—we might need to mention a few other changes here if they made it in.

It'd be nice to include the release notes on the GitHub page for the tag as well (see e.g. [the last circe release](https://github.com/travisbrown/circe/releases/tag/v0.2.0)), since that includes other useful links and is likely to be the first place people look. It introduces some redundancy, but not in the repo itself, and I think it's worth it.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,10 @@ This version includes new type class instances:
   `Validated`, `Ior`, `Xor`, and `XorT`
 * [#569](https://github.com/non/cats/pull/569): `Functor` instances for
   `OptionT` when `F` has a `Functor` instance but not a `Monad`
+* [#600](https://github.com/non/cats/pull/600): `Show` instances for `Option`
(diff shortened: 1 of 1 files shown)


non:
This is perfect, thanks! :+1:

adelbertc:
:+1:

refried:
Nice!

PR #621, opened 2015-11-10
Add explicit Id instances for Kleisli

### LINKED ISSUE #580: Inference issues with Id
I can fetch the `Applicative` instance for a `Kleisli` of a `List` no problem:

```scala
scala> import cats._, data._, implicits._
import cats._
import data._
import implicits._

scala> Applicative[Kleisli[List, Int, ?]]
res0: cats.Applicative[[γ]cats.data.Kleisli[[+A]List[A],Int,γ]] = cats.data.KleisliInstances$$anon$2@10c03632
```

However, if I use `Id` instead of `List`, I run into issues:

```scala
scala> Applicative[Kleisli[Id, Int, ?]]
<console>:20: error: could not find implicit value for parameter instance: cats.Applicative[[γ]cats.data.Kleisli[[A]A,Int,γ]]
       Applicative[Kleisli[Id, Int, ?]]
                  ^
```

This smells like a scala compiler bug, because it works if I call `kleisliApplicative` explicitly:

```scala
scala> val F: Applicative[Kleisli[Id, Int, ?]] = Kleisli.kleisliApplicative[Id, Int]
F: cats.Applicative[[γ]cats.data.Kleisli[[A]A,Int,γ]] = cats.data.KleisliInstances1$$anon$4@12dd5884
```

Questions:
* Why exactly does this happen?
* Does this have to do with the way we have Id defined? If so, can we change the way we define it to get around this?
* If we can't work around this with the way we have Id defined, should we work around it by providing explicit instances related to Id in a bunch of places?

ceedubs:
Fixes #580.

### DIFF
--- a/core/src/main/scala/cats/data/Kleisli.scala
+++ b/core/src/main/scala/cats/data/Kleisli.scala
@@ -89,8 +89,14 @@ private[data] sealed abstract class KleisliInstances extends KleisliInstances0 {
   implicit def kleisliMonoidK[F[_]](implicit M: Monad[F]): MonoidK[Lambda[A => Kleisli[F, A, A]]] =
     new KleisliMonoidK[F] { def F: Monad[F] = M }
 
+  implicit val kleisliIdMonoidK: MonoidK[Lambda[A => Kleisli[Id, A, A]]] =
+    kleisliMonoidK[Id]
+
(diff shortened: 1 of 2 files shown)


adelbertc:
:+1:

non:
:+1: but needs a re-merge

ceedubs:
Thanks @non. Updated.

PR #622, opened 2015-11-11
Make a bunch of classes final and a few private

ceedubs:
Probably best to default to `final` and `private` until we run into a
reason to expose these for extension.

### DIFF
--- a/core/src/main/scala/cats/data/Prod.scala
+++ b/core/src/main/scala/cats/data/Prod.scala
@@ -33,35 +33,34 @@ private[data] sealed abstract class ProdInstances extends ProdInstance0 {
   }
 }
 
-sealed abstract class ProdInstance0 extends ProdInstance1 {
+private[data] sealed abstract class ProdInstance0 extends ProdInstance1 {
   implicit def prodMonoidK[F[_], G[_]](implicit FF: MonoidK[F], GG: MonoidK[G]): MonoidK[Lambda[X => Prod[F, G, X]]] = new ProdMonoidK[F, G] {
     def F: MonoidK[F] = FF
     def G: MonoidK[G] = GG
   }
 }
 
-sealed abstract class ProdInstance1 extends ProdInstance2 {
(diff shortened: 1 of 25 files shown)


non:
Looks good to me. :+1:

core/src/main/scala/cats/data/Prod.scala
```diff
@@ -33,35 +33,34 @@ private[data] sealed abstract class ProdInstances extends ProdInstance0 {
...
+private[data] sealed abstract class ProdInstance2 extends ProdInstance3 {
   implicit def prodApplicative[F[_], G[_]](implicit FF: Applicative[F], GG: Applicative[G]): Applicative[Lambda[X => Prod[F, G, X]]] = new ProdApplicative[F, G] {
     def F: Applicative[F] = FF
     def G: Applicative[G] = GG
   }
 }
 
-sealed abstract class ProdInstance3 extends ProdInstance4 {
-  implicit def prodApply[F[_], G[_]](implicit FF: Apply[F], GG: Apply[G]): Apply[Lambda[X => Prod[F, G, X]]] = new ProdApply[F, G] {
+private[data] sealed abstract class ProdInstance3 extends ProdInstance4 { implicit def prodApply[F[_], G[_]](implicit FF: Apply[F], GG: Apply[G]): Apply[Lambda[X => Prod[F, G, X]]] = new ProdApply[F, G] {
```
travisbrown: Super nitpicky but do you mind re-inserting the line-break here for the sake of consistency and the diff?

travisbrown:
:+1: from me.

core/src/main/scala/cats/data/Prod.scala
ceedubs: Good catch. I didn't mean to do this.

ceedubs: Hmm weird. This is on two lines for me locally. Am I hitting some bizarre LF/CR/CRLF inconsistency?

PR #623, opened 2015-11-11
Fix Foldabe#dropWhile_ bug.

### LINKED ISSUE #563: Foldable.dropWhile_() is not working as advertised
this is what i get when using Foldable.dropWhile_() in the repl.  the results do not conform with the spec.
```
scala> Foldable[List].dropWhile_(List(2, 4, 5))(_ % 2 == 0)
res46: List[Int] = List(2, 4, 5)

scala> // should return List(5)

scala> Foldable[List].dropWhile_(List(1, 2, 4, 5))(_ % 2 == 0)
res47: List[Int] = List(2, 4, 5)

scala> // should return List(1, 2, 4, 5)
```

non:
In addition to fixing this bug, the commit defines several
FoldableCheck tests for various types that have foldable
instances.

Finally, the commit adds takeWhile_.

Fixes #563.


core/src/main/scala/cats/std/iterable.scala
```diff
@@ -0,0 +1,18 @@
+package cats
+package std
+
+trait IterableInstances {
+  implicit val iterableInstance: Foldable[Iterable] =
```
ceedubs: Does this belong in cats core? In a recent [gitter conversation](https://gitter.im/non/cats?at=5633b2d0e4bb7eee537ffb51), the consensus was to put `IndexedSeq` instances in alleycats because `IndexedSeq` can potentially be mutable. Wouldn't the same arguments apply to `Iterable`?

ceedubs:
You didn't mention that this PR also adds a `Foldable[Iterable]` instance :). I've left a question about that. Other than that, these changes look great to me.

non:
My claim is that foldable instances are different (i.e. we have one for `Set`), and none of the foldable methods return `F[A]`. But I'm fine removing the instance too, it just seemed to me to be an oversight.

ceedubs:
@tpolecat what are your thoughts?

Are we checking `FoldableLaws` for the `Foldable[Iterable]`? It doesn't look like it to me, but I might be overlooking some inheritance or something.

tpolecat:
Bad Cop to the rescue! So, yeah I think `Foldable[Iterable]` is bogus because the internal call to `.iterator` is observably side-effecting. That is,

```scala
val a = F.fold(xs)
val b = F.fold(xs)
```

is not necessarily the same program as

```scala
val a = F.fold(xs)
val b = a
```

even though `fold` is called with a pure value.

non:
Ugh. OK, sure.

Personally I'd prefer to use *correctness-preserving* when dealing with something like Scala-collections but I see your point. I'll remove it. :hammer: 🔥

ceedubs:
:+1:

adelbertc:
:+1:

PR #627, opened 2015-11-12
Fix typo

antoinecomte:


ceedubs:
Wonderful. Thanks @antoinecomte!

docs/src/main/tut/const.md
```diff
@@ -247,7 +247,7 @@ implicit def constApplicative[Z : Monoid]: Applicative[Const[Z, ?]] =
 
 We have our `Applicative`!
 
-Going back to `Traverse`, we fill in the first paramter of `traverse` with `fa` since that's
+Going back to `Traverse`, we fill in the first paramater of `traverse` with `fa` since that's
```
ceedubs: Oh actually this one is still off. Should be `parameter` not `paramater` :)

antoinecomte: Oops …

It’s fixed now

-- 
Antoine Comte


On 12 novembre 2015 at 14:23:50, Cody Allen (notifications@github.com) wrote:

In docs/src/main/tut/const.md:

> @@ -247,7 +247,7 @@ implicit def constApplicative[Z : Monoid]: Applicative[Const[Z, ?]] =
>   
>  We have our `Applicative`!
>   
> -Going back to `Traverse`, we fill in the first paramter of `traverse` with `fa` since that's
> +Going back to `Traverse`, we fill in the first paramater of `traverse` with `fa` since that's
Oh actually this one is still off. Should be parameter not paramater :)

ceedubs:
Thanks! :+1:

Issue #620, opened 2015-11-10
Add more WriterT instances


ceedubs:
`WriterT` currently has `Monad` and `Eq` instances, but there are many more that it could have. Some examples that I think can be added are:
- [x] Functor (#628)
- [x] Apply (#628)
- [x] Applicative (#628)
- [x] FlatMap (#628)
- [x] MonadFilter (#696)
- [x] Alternative (#696)
- [x] MonadCombine (#696)
- [x] ApplicativeError (#689)
- [x] MonadError (#689)
- [x] Show (#928)
- [ ] Foldable
- [ ] Traverse
- [x] Bifunctor (#673)
- [x] CoflatMap (#1049)
- [ ] Comonad
- [x] Semigroup (#1014)
- [x] Monoid (#1014)
- [x] SemigroupK (#696)
- [x] MonoidK (#696)

ceedubs:
I'll work on this.

akilegaspi:
Is this still in progress? I'm thinking of contributing for this enhancement.

kailuowang:
that'll be awesome. I updated the checklist, three instances still missing.

akilegaspi:
Just a quick question.

Is the resolving type for the `WriterT[F, L,  A]` after passing through `extract` is `F[(L, A)]`

barambani:
I would look into this one if that's ok.

akilegaspi:
I think this should be closed since @barambani already finished it and it has been merged to master

PR #628, opened 2015-11-12
Add WriterT instances for supertypes of Monad

ceedubs:
This is a partial fix for #620. There are still more `WriterT` instances that could be added, but this is already a sizable PR, so I thought getting in the supertypes of `Monad` would be a good chunk of work to split off.

I've gone to fairly great lengths to ensure that implicit resolution works in various cases in the tests. Perhaps I've even gone a bit overboard, but I think that these tests are kind of necessary, because there are some inference issues with `Id` that are compounded by issues with type aliases. In fact I had to rework the prioritization hierarchy several times as these tests uncovered issues.


non:
:+1:

stew:
:+1:

PR #634, opened 2015-11-14
Avoid `whenever` in tests

### LINKED ISSUE #632: Remove usage of `whenever` in tests
Currently some tests, such as some in `IorTests` use scalatest's `whenever` to filter generated values in tests. Occasionally, this leads to too many values being discarded, which causes the test to be abandoned and the build to fail.

We should take alternative approaches. Either we should use a generator that only generates values that we want, or we should change the test to handle the other values and perform a relevant check on them.

ceedubs:
Fixes #632.


non:
Love it! :heart: :+1:

adelbertc:
:+1:

PR #635, opened 2015-11-14
Remove Kleisli.lmap

### LINKED ISSUE #624: Kleisli: local and lmap are the same
@philwills has pointed out [on gitter](https://gitter.im/non/cats?at=564314316420c33467a16021) that `local` and `lmap` on `Kleisli` are essentially the same thing.

```scala
def lmap[C](f: C => A): Kleisli[F, C, B] =
  Kleisli(run compose f)

def local[AA](f: AA => A): Kleisli[F, AA, B] =
  Kleisli(f.andThen(run))
```

We should probably either remove one of them or make one just an alias for the other (and document that it is so).

My personal preference would be to drop one of them. I think that aliases tend to just make it harder on the user, as it just gives them more to learn.

ceedubs:
It is the same as Kleisli.local.

Fixes #624.


non:
:+1:

adelbertc:
:+1:

PR #636, opened 2015-11-14
Check more Eval instance laws

### LINKED ISSUE #630: Law-checking on Eval instances
The [code coverage report](https://codecov.io/github/non/cats/core/src/main/scala/cats/Eval.scala) for `Eval` shows a number of type class instances that currently aren't hit in tests. We should add law-checking for these instances (`Group`, `Monoid`, etc).

It looks like we currently aren't hitting the `memoized` method in tests either. We should probably add some tests for that.

ceedubs:
This makes `Eval` serializable and adds law-checking for its `Group`, `Monoid`,
`Semigroup`, `Order`, `PartialOrder`, and `Eq` instances.

Fixes #630.

### DIFF
--- a/core/src/main/scala/cats/Eval.scala
+++ b/core/src/main/scala/cats/Eval.scala
@@ -33,7 +33,7 @@ import cats.syntax.all._
  * Eval instance -- this can defeat the trampolining and lead to stack
  * overflows.
  */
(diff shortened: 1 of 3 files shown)


non:
:+1: rockstar :guitar: :speaker: :notes:

adelbertc:
:+1:

Issue #629, opened 2015-11-12
Add Reducible laws


ceedubs:

adelbertc:
Working on it: https://github.com/typelevel/cats/pull/893

ceedubs:
Resolved by #893

PR #639, opened 2015-11-14
Add Traverse[Id] and Id law-checking

ceedubs:
We should also be able to add a `Reducible` instance for `Id`, but I think that should wait for #629.


non:
:+1:

adelbertc:
:+1:

PR #642, opened 2015-11-15
Add a test for Show[Map]

mikejcurry:
Adds a test for Show[Map]


ceedubs:
Thanks, @mikejcurry!

Since Simulacrum adds syntax for `Show` instances, it should be possible to just use `map.show` instead of bothering with `mapShow`. Would you want to give that a try?

mikejcurry:
Sure, no problem - straightforward enough; I'll probably add a check that both approaches produce the same results also as it makes sense they produce consistent values. In most instances unlikely for anything to go wrong but a test shouldn't really hurt.

ceedubs:
:+1:

rossabaker:
:+1:

PR #645, opened 2015-11-15
Tests: XorT with Id as F should be consistent with Xor

ceedubs:
Add a bunch of tests that XorT[Id, A, B] methods are consistent with
Xor[A, B].


stew:
:+1:

PR #647, opened 2015-11-15
Change ProdInstance to ProdInstances for consistency

ceedubs:
This makes the `Prod` instances hierarchy names consistent with instances for
other data types. Mostly this is to make them easier to find with
grepping.


stew:
:+1:

rossabaker:
:+1:

PR #648, opened 2015-11-15
Streaming and StreamingT extend Product with Serializable

ceedubs:
This helps type inference by fixing the least upper bound of subtypes of
Streaming(T).

Before:
```scala
scala> val x = if (true) Cons(1, Now(Streaming.empty[Int])) else Streaming.Empty[Int]()
x: Product with Serializable with cats.data.Streaming[Int] = Streaming(1, ...)
```

After:
```scala
scala> val x = if (true) Cons(1, Now(Streaming.empty[Int])) else Streaming.Empty[Int]()
x: cats.data.Streaming[Int] = Streaming(1, ...)
```

### DIFF
--- a/core/src/main/scala/cats/data/Streaming.scala
+++ b/core/src/main/scala/cats/data/Streaming.scala
@@ -58,7 +58,7 @@ import scala.collection.mutable
  *     constructed with `Foldable#foldRight`, and that `.map` and
  *     `.flatMap` operations over the tail will be safely trampolined.
  */
-sealed abstract class Streaming[A] { lhs =>
+sealed abstract class Streaming[A] extends Product with Serializable { lhs =>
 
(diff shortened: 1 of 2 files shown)


hakuch:
Off-hand comment: There was a brief conversation on the Gitter channel on the need to explicitly inherit from `Product` and `Serializable` in the base trait or class when the sub-classes are case classes in order to produce a more helpful "least upper-bound". I wish there was first-class support in Scala for algabraic data types like in Haskell or OCaml. Do intersection types in Dotty address this?

ceedubs:
@hakuch That's an interesting question. I haven't really read up on Dotty yet. Someone else might have an answer for that.

fthomas:
:+1:

stew:
:+1:

PR #649, opened 2015-11-15
Make Arbitrary StreamingT more arbitrary

ceedubs:
This still isn't a perfect Arbitrary instance for StreamingT, but I
think it will help us hit more test cases than the current
implementation. For example, I don't think the current instance ever
results in `Wait` instances.

@non could you please be one of the reviewers of this?


non:
:+1:

Another improvement I could imagine is using weights to try to help ensure that we have longer streams (since right now there is a 1/3 chance of being empty at each step). But let's merge this first.

stew:
:+1:

Issue #487, opened 2015-08-26
sbt-doctest for Scaladoc


adelbertc:
https://github.com/tkawachi/sbt-doctest

fthomas:
:+1:

I've added it to scalaz-stream (https://github.com/scalaz/scalaz-stream/pull/248) a year ago and it hadn't caused any issues there. But I'm biased since I'm also a contributor of sbt-doctest. :-)

non:
In the interest of type-checking our documentation I definitely support this. I guess I'd want to see it in action here to make sure it is reasonable. I worry that there would be a sea of imports required, but maybe not?

travisbrown:
@non If demonstrating basic functionality of a method requires a sea of imports I think that's a problem in itself, and sbt-doctest is a good way to make _us_ feel that pain first. :smile:

stew:
I'd love to see doctest

ceedubs:
I just gave this a try and ran into https://github.com/tkawachi/sbt-doctest/issues/56. I tried to selectively disable the unused import flag in tests, but my SBT skills failed me. Does anyone with a better grasp on SBT want to give it a try?

ceedubs:
Big thanks to @fthomas for the quick turnaround on a fix for https://github.com/tkawachi/sbt-doctest/issues/56. I'll start on this.

fthomas:
fixed by #724

PR #651, opened 2015-11-15
Fix outdated example in Scaladoc for catchOnly

travisbrown:
Just noticed that the example didn't get the rename.


ceedubs:
Oops, thanks! :+1:

This is a good example of where #487 would be useful :)

rossabaker:
:+1:

PR #370, opened 2015-06-23
Sketch of approach to fix #168

woparry:
This is the first set of tests that verify your typeclass implementations are interchangeable with the defaults.

The pieces:

FunctorFromApplicative and ApplicativeFromMonad are mixins that define (map, ap) in terms of (ap, flatMap). They make it easy to build instances of Monad and Applicative by only defining the core functions (pure and (ap or flatMap)).

Applicative, FlatMap, Traverse, etc no longer provide default implementations of ap and map.
You must override them yourself with an efficient version (recommended) or mix in the appropriate trait from canon to get the defaults.

The canon package object get canonicalMonad and canonicalApplicative, which use the mixins to build an implementation using the default versions of derived functions.

The Laws use these to verify that the instance implementations of (map, ap, map2) match the canonical implementations.

PR #652, opened 2015-11-15
Add tests to cover derived methods leftMap and rightMap on BiFunctor

mikejcurry:
This PR would add tests that should cover identity for the derived leftMap and rightMap on BiFunctor.

I guess this PR is exploratory:
* Is this a suitable use of the laws infrastructure? Seems to be a good use, but not sure if you would actually consider them laws.

Also:
* There is no syntax for leftMap and rightMap at this stage, just bimap; would it make sense to add these as syntax also. I’ve not done that in this PR, but if it is generally in line with what we want to do I can add.

### DIFF
--- a/core/src/main/scala/cats/syntax/bifunctor.scala
+++ b/core/src/main/scala/cats/syntax/bifunctor.scala
@@ -10,5 +10,10 @@ trait BifunctorSyntax {
 }
 
 class BifunctorOps[F[_, _], A, B](fab: F[A, B])(implicit F: Bifunctor[F]) {
+  
   def bimap[C, D](f: A => C, g: B => D): F[C,D] = F.bimap(fab)(f,g)
+
+  def leftMap[C](f: A => C): F[C, B] = F.leftMap(fab)(f)
+
+  def rightMap[D](f: B => D): F[A, D] = F.rightMap(fab)(f)
 }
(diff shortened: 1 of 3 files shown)


ceedubs:
:+1: thanks!

> Is this a suitable use of the laws infrastructure?

At some point #370 may provide some answers to this. However, until we have something better in place, I think these sorts of checks are good. There is also a precedent [here](https://github.com/non/cats/pull/562#issuecomment-146380855).

> There is no syntax for leftMap and rightMap at this stage, just bimap; would it make sense to add these as syntax also.

Yes, I think so. We would inherit them for free if Simulacrum supported type classes for type constructors of kind `F[_, _]`. Since it doesn't, it would probably be handy to add them manually.

mikejcurry:
@ceedubs - Cool, Thanks. I've gone ahead and locally made changes to add leftMap and rightMap to syntax, added associativity checks for leftMap and rightMap, and updated the tests to use the syntax additions. Would you folks prefer pushing to this PR or a separate PR after this one?

ceedubs:
@mikejcurry I have no real preference. I have no objections to you adding those changes to this PR. The more changes you add to a PR the more likely one of them will raise objections, but these seem pretty uncontroversial to me :)

mikejcurry:
@ceedubs Sounds good - I'll push the changes to this PR and keep the set of changes in the same area together.

ceedubs:
:+1:

non:
:+1:

PR #660, opened 2015-11-16
Make Arbitrary Streaming instances more arbitrary

ceedubs:
This commit does two things.

1. Use a frequency distribution in the `StreamingT` generator as
suggested by @non [here](https://github.com/non/cats/pull/649#issuecomment-156856211).
It is now twice as likely to generate a Wait as an Empty and 6 times as
likely to generate a Cons as an Empty.
2. Update the `Streaming` generator to be similar to the `StreamingT`
generator (use a frequency distribution generator of Empty, Wait, and
Cons).


non:
:+1:

adelbertc:
:+1:

PR #662, opened 2015-11-16
DO NOT MERGE: repeat evaluation with Streaming.memoize

ceedubs:
This is a potential bug report and is not meant to be merged in its
current form.

This test fails. It attempts to check that if a `Streaming` instance is
created via `.memoize`, then its elements won't be evaluated multiple
times (such as when `.toList` is called).

One could argue that this is a bogus test case, since side effects
should be tracked in a type other than `Unit`. However, I believe that
this means that we might be evaluating arguments overly-eagerly, which
is a problem in itself.

I made a naive attempt at fixing this by making `Streaming.map` a little
lazier in the `Cons` case, but that didn't seem to do the trick.

@non do you have any thoughts?

### DIFF
--- a/tests/src/test/scala/cats/tests/StreamingTests.scala
+++ b/tests/src/test/scala/cats/tests/StreamingTests.scala
@@ -43,6 +43,18 @@ class AdHocStreamingTests extends CatsSuite {
     }
   }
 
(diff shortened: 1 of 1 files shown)


ceedubs:
Fixed by #665.

PR #665, opened 2015-11-16
Fix Streaming#memoization.

non:
This commit fixes a bug noticed by @ceedubs. The basic issue
is that we have to be sure to memoize the Eval instance holding
the tail while *also* memoizing the tail itself when it is
calculated. Previously we were only memoizing the head node.

### DIFF
--- a/core/src/main/scala/cats/data/Streaming.scala
+++ b/core/src/main/scala/cats/data/Streaming.scala
@@ -642,14 +642,21 @@ sealed abstract class Streaming[A] extends Product with Serializable { lhs =>
    * Ensure that repeated traversals of the stream will not cause
    * repeated tail computations.
    *
-   * By default stream does not memoize to avoid memory leaks when the
-   * head of the stream is retained.
+   * By default this structure does not memoize to avoid memory leaks
+   * when the head of the stream is retained. However, the user
+   * ultimately has control of the memoization approach based on what
+   * kinds of Eval instances they use.
+   *
+   * There are two calls to memoized here -- one is a recursive call
+   * to this method (on the tail) and the other is a call to memoize
+   * the Eval instance holding the tail. For more information on how
+   * .memoize works see Eval#memoize.
    */
   def memoize: Streaming[A] =
     this match {
       case Empty() => Empty()
-      case Wait(lt) => Wait(lt.memoize)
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/data/Streaming.scala
```diff
@@ -642,14 +642,21 @@ sealed abstract class Streaming[A] extends Product with Serializable { lhs =>
...
-   * head of the stream is retained.
+   * By default this structure does not memoize to avoid memory leaks
+   * when the head of the stream is retained. However, the user
+   * ultimately has control of the memoization approach based on what
+   * kinds of Eval instances they use.
+   *
+   * There are two calls to memoized here -- one is a recursive call
+   * to this method (on the tail) and the other is a call to memoize
+   * the Eval instance holding the tail. For more information on how
+   * .memoize works see Eval#memoize.
```
ceedubs: Minor: could we turn this into a proper ScalaDoc link? `[[cats.Eval.memoize]]`

ceedubs:
This fixes (and replaces) #662.

I left a minor request about the ScalaDoc, but :+1: from me.

core/src/main/scala/cats/data/Streaming.scala
non: Oh, sure.

tests/src/test/scala/cats/tests/StreamingTests.scala
```diff
@@ -37,6 +37,28 @@ class StreamingTests extends CatsSuite {
...
+      xs should === (ns)
+
+      val ys = memoized.toList
+      i should === (size)
+      ys should === (ns)
+    }
+  }
+
+  // convert List[A] to Streaming[A]
+  def convert[A](as: List[A]): Streaming[A] =
```
adelbertc: Is this used anywhere, or just added for possible future use?

adelbertc:
:+1: once question about function is answered :-)

tests/src/test/scala/cats/tests/StreamingTests.scala
non: Oh, looks like I left that in. I'll remove it.

adelbertc:
Strange error, rebuilding, once it's green :+1:

Issue #659, opened 2015-11-16
Add OptionT.fromF


lambdista:
I noticed that the 0.3.0 version includes a useful `OptionT.fromOption` function. It lets you build an `OptionT` from an `Option`. 

Wouldn't it be the case to add an `OptionT.fromF` that lets you build an `OptionT` from an `Applicative F`? As in:

```scala
scala> OptionT.fromF(List(8))
res1: cats.data.OptionT[List,Int] = OptionT(List(Some(8)))
```

Of course I'd accept suggestions for a better name than `fromF`.

ceedubs:
Thanks for the suggestion @lambdista. Just to make sure I understand clearly, you're talking about a function that will take an `F[A]` and lift it into an `OptionT[F, A]` where we essentially `map` on the `F[A]` and wrap every `A` in `Some`, right?

lambdista:
Thanks for your answer @ceedubs. Yes, that's exactly what I mean. About your thoughts:

* You're totally right, I think we could get by with just a `Functor` too.
* Yes, it's essentially `liftM` from `MonadTrans` but, as you correctly wrote, having an `OptionT`-specific method would do no harm and require a less constrained `Functor` requirement compared to the `Monad` of `liftM`.

Let me know if you want me to add it and send a PR. Thank you.

ceedubs:
I'd like this :+1:. Anyone else?

I'd propose `liftF` as a name to parallel `liftM` but with `Functor` instead of `Monad`.

non:
:+1:

lambdista:
That's great! `liftF` is perfect. :+1:

lambdista:
Implemented, added test and updated related docs. Let me know if you want to me to change/add anything. Thank you.

ceedubs:
Completed by #667.

PR #667, opened 2015-11-17
add OptionT.liftF and update related docs and tests. Close issue #659

lambdista:


core/src/main/scala/cats/data/OptionT.scala
```diff
@@ -120,6 +120,12 @@ object OptionT extends OptionTInstances {
     def apply[A](value: Option[A])(implicit F: Applicative[F]): OptionT[F, A] =
       OptionT(F.pure(value))
   }
+
+  /**
+    * Lifts the `F[A]` Functor into an `OptionT[F, A]`.
+    *
+    */
+  def liftF[F[_], A](fa: F[A])(implicit F: Functor[F]): OptionT[F, A] = OptionT(F.map(fa)(Option(_)))
```
ceedubs: This may not really be something to worry about, but I'm thinking that maybe this should use `Some(_)` instead of `Option(_)`. The `Option.apply` method has logic to turn `null` into `None`. In Cats we are generally assuming that people aren't passing around `null`. But if for some reason they really want to lift a `null` value into the `OptionT` structure, I don't think we should prevent them from doing that here.

docs/src/main/tut/optiont.md
```diff
@@ -51,6 +51,33 @@ val noWelcome: OptionT[Future, String] = customGreetingT.filterNot(_.contains("w
...
+## From `Option[A]` and/or `F[A]` to `OptionT[F, A]`
+
+Sometimes you may have an `Option[A]` and/or `F[A]` and want to *lift* them into an `OptionT[F, A]`. For this purpose `OptionT` exposes two useful methods, namely `fromOption` and `liftF`, respectively. E.g.:
+
+```tut:silent
+import scala.concurrent.Future
+import scala.concurrent.ExecutionContext.Implicits.global
+
+import cats.data.OptionT
+import cats.std.future._
```
ceedubs: These imports already happen earlier in the file. Do we want them again here?

ceedubs:
Thanks, @lambdista! In general, this looks great. I left a couple very minor comments.

docs/src/main/tut/optiont.md
lambdista: Yes, I noticed but I wanted to example to be *self-contained*. 

By the way, if you don't import `scala.concurrent.ExecutionContext.Implicits.global` you get a misleading: 
```scala
scala> OptionT.liftF(f)
<console>:17: error: could not find implicit value for parameter F: cats.Functor[scala.concurrent.Future]
       OptionT.liftF(f)
```
Do you think I should file this in a different issue or it's not a problem? I mean, even if you import `cats.std.future._` you get that error if no `ExecutionContext` is found in scope and people could get confused by this because they would think the problem is it cannot find a `Functor[Future]` instance while it actually can't just find the `ExecutionContext`. 

Anyway if you think it's not a problem I can remove those imports.

core/src/main/scala/cats/data/OptionT.scala
lambdista: I used `Option(_)` to prevent `null` and getting `None` instead of `Some(null)` in that case, but if you confirm we should not prevent them from doing that I'll change it to `Some(_)`.

docs/src/main/tut/optiont.md
ceedubs: I can see the benefit in self-contained examples. My only concern is that currently with tut I don't think we have a way to clear previous imports in the same file. So we run the risk of duplicating imports without actually guaranteeing that the example _is_ self-contained. I don't have any really strong feelings on this though, so whatever you think is best.

Regarding the misleading missing implicit value message:
This came up in the gitter channel [yesterday](https://gitter.im/non/cats?at=564a3a658b242470793e3053). I don't particularly like any of the solutions that we have in mind, but if there's something we can do to help, then we should. If you want to file an issue for that, that would be great!

lambdista:
Thank you @ceedubs for your support!

core/src/main/scala/cats/data/OptionT.scala
ceedubs: I think that it should be consistent with our `Applicative[Option].pure` implementation, which uses `Some`: https://github.com/non/cats/blob/master/core/src/main/scala/cats/std/option.scala#L14. So yeah, I think we should change this here. I don't think we should conflate the issues of lifting and de-nulling.

lambdista: I totally agree! I'll change it to `Some(_)`. Thank you.

docs/src/main/tut/optiont.md
lambdista: I removed those imports and filed the #670 for that issue.

ceedubs:
:+1: (assuming the build passes). Thanks again!

lambdista:
Thank you! :+1:

non:
:+1: Good work! Thanks! :cat2:

PR #668, opened 2015-11-17
Add PartialOrder and Eq instances for XorT

### LINKED ISSUE #664: Add Order and PartialOrder instances for XorT
The `compare` and `partialCompare` methods are already defined on `XorT`, but we need to add the type class instances.

ceedubs:
Fixes #664.


non:
:+1:

adelbertc:
:+1:

PR #671, opened 2015-11-17
Request to Include `cats.data.Coproduct` and `cats.free.Inject`

raulraja:
The following PR adds `cats.data.Coproduct` and `cats.free.Inject` largely based on those found in Scalaz as in ["Data types a la carte"](http://www.staff.science.uu.nl/~swier004/Publications/DataTypesALaCarte.pdf) (Swierstra 2008).

`Coproduct` & `Inject` can be used to compose Free based Algebras where an Application is nothing but the Coproduct resulting from the interaction of it's algebras.

In other words:

```scala
import cats.arrow.NaturalTransformation
import cats.data.{Xor, Coproduct}
import cats.free.{Inject, Free}
import cats.{Id, ~>}
import scala.collection.mutable.ListBuffer

/* Handles user interaction */
sealed trait Interact[A]
case class Ask(prompt: String) extends Interact[String]
case class Tell(msg: String) extends Interact[Unit]

/* Represents persistence operations */
sealed trait DataOp[A]
case class AddCat(a: String) extends DataOp[Unit]
case class GetAllCats() extends DataOp[List[String]]

type CatsApp[A] = Coproduct[DataOp, Interact, A]

def lift[F[_], G[_], A](fa: F[A])(implicit I: Inject[F, G]): Free[G, A] =
  Free.liftF(I.inj(fa))

class Interacts[F[_]](implicit I: Inject[Interact, F]) {
  def tell(msg: String): Free[F, Unit] = lift(Tell(msg))
  def ask(prompt: String): Free[F, String] = lift(Ask(prompt))
}

object Interacts {
  implicit def interacts[F[_]](implicit I: Inject[Interact, F]): Interacts[F] = new Interacts[F]
}

class DataSource[F[_]](implicit I: Inject[DataOp, F]) {
  def addCat(a: String): Free[F, Unit] = lift[DataOp, F, Unit](AddCat(a))
  def getAllCats: Free[F, List[String]] = lift[DataOp, F, List[String]](GetAllCats())
}

object DataSource {
  implicit def dataSource[F[_]](implicit I: Inject[DataOp, F]): DataSource[F] = new DataSource[F]
}

def program(implicit I : Interacts[CatsApp], D : DataSource[CatsApp]) = {

  import I._, D._

  for {
    cat <- ask("What's the kitty's name")
    _ <- addCat(cat)
    cats <- getAllCats
    _ <- tell(cats.toString)
  } yield ()
}
```
Interpreters omitted for simplicity but available in the free monads tutorial and Inject tests.

@non @ceedubs thanks for your consideration.


non:
This looks like a great PR! I'm on a train so I can't do a detailed review yet, but on the surface it looks great. Thanks for updating the tutorial and including tests as well! :cat2:

adelbertc:
This is awesome, thanks! Looks like there is a legitimate build failure - I'm guessing `Option#contains` was added in 2.11 and missing from 2.10 :/ 

Other than that, lots of tests, lots of doc, I like it :+1:

core/src/main/scala/cats/syntax/coproduct.scala
```diff
@@ -0,0 +1,13 @@
+package cats
+package syntax
+
+import cats.data.Coproduct
+
+trait CoproductSyntax {
+  implicit def coproductSyntax[F[_], A](a: F[A]): CoproductOps[F, A] = new CoproductOps(a)
+}
+
+class CoproductOps[F[_], A](val a: F[A]) extends AnyVal {
```
ceedubs: Minor: let's go ahead and make this class `final` to be consistent with other `Ops` classes.

docs/src/main/tut/freemonad.md
```diff
@@ -348,6 +348,120 @@ it's not too hard to get around.)
 val result: Map[String, Int] = compilePure(program, Map.empty)
 ```
 
+## Composing Free monads ADTs.
+
+Real world applications often time combine different algebras. 
+The `Inject` typeclass described by Swierstra in [Data types à la carte](http://www.staff.science.uu.nl/~swier004/Publications/DataTypesALaCarte.pdf)
+let us compose different algebras in the context of `Free`.
```
ceedubs: Minor typo: "lets us"

docs/src/main/tut/freemonad.md
```diff
@@ -348,6 +348,120 @@ it's not too hard to get around.)
...
+
+Let's see a trivial example of unrelated ADT's getting composed as a `Coproduct` that conform a more complex program.
```
ceedubs: Should "conform" here be "can form" or just "form"?

docs/src/main/tut/freemonad.md
```diff
@@ -348,6 +348,120 @@ it's not too hard to get around.)
...
+
+```tut
+type CatsApp[A] = Coproduct[DataOp, Interact, A]
+```
+
+In order to take advantage of monadic composition we use smart constructors to lift our Algebra to the `Free` context.
+
+```tut
+def lift[F[_], G[_], A](fa: F[A])(implicit I: Inject[F, G]): Free[G, A] =
+  Free.liftF(I.inj(fa))
```
ceedubs: Should this method be added to Cats?

docs/src/main/tut/freemonad.md
```diff
@@ -348,6 +348,120 @@ it's not too hard to get around.)
...
+  for {
+    cat <- ask("What's the kitty's name")
+    _ <- addCat(cat)
+    cats <- getAllCats
+    _ <- tell(cats.toString)
+  } yield ()
+}
+```
+
+Finally we write one interpreter per ADT and combine them with a `NaturalTransformation` to `Coproduct` so when they can be
```
ceedubs: I think we can drop "when" from this sentence.

docs/src/main/tut/freemonad.md
```diff
@@ -348,6 +348,120 @@ it's not too hard to get around.)
...
+  }
+}
+
+def or[F[_], G[_], H[_]](f: F ~> H, g: G ~> H): Coproduct[F, G, ?] ~> H =
+  new NaturalTransformation[Coproduct[F, G, ?], H] {
+    def apply[A](fa: Coproduct[F, G, A]): H[A] = fa.run match {
+      case Xor.Left(ff) => f(ff)
+      case Xor.Right(gg) => g(gg)
+    }
+  }
```
ceedubs: Should this method live within Cats?

free/src/main/scala/cats/free/Inject.scala
```diff
@@ -0,0 +1,49 @@
...
+  *
+  * @see [[http://www.staff.science.uu.nl/~swier004/Publications/DataTypesALaCarte.pdf]]
+  */
+sealed abstract class Inject[F[_], G[_]] {
+  def inj[A](fa: F[A]): G[A]
+
+  def prj[A](ga: G[A]): Option[F[A]]
+}
+
+sealed abstract class InjectInstances {
```
ceedubs: Could you please add explicit return types to the methods in this object?

tests/src/test/scala/cats/tests/CoproductTests.scala
```diff
@@ -0,0 +1,38 @@
...
+
+  checkAll("Coproduct[Show, Show, ?]", ContravariantTests[Coproduct[Show, Show, ?]].contravariant[Int, Int, Int])
+  checkAll("Contravariant[Coproduct[Show, Show, ?]]", SerializableTests.serializable(Contravariant[Coproduct[Show, Show, ?]]))
+
+  test("double swap is identity") {
+    forAll { (x: Coproduct[Option, Option, Int]) =>
+      x.swap.swap should ===(x)
+    }
+  }
+
```
ceedubs: Looks like we might be missing law-checking for the `Eq` instance. Are there any others we aren't hitting?

ceedubs:
Thanks a bunch @raulraja. I left some minor comments, but in general this is very clean, has good test coverage, and even comes with a great tutorial! :100:

docs/src/main/tut/freemonad.md
raulraja: As a user, this is something I use all the time and it's a the core of the Smart Constructors pattern to lift ADTs to `Free`. So I'd be inclined to say YES but I'm not sure where it would belong. Maybe someone in the more theoretical side of CT among us may know?

docs/src/main/tut/freemonad.md
raulraja: Same as the discussed `lift` above. I'd love to see it included but not sure where it belongs. This is needed to create interpreters and it's used in the same way as the `Coproduct` type alias to stack up multiple algebras. `CoProduct` builds the Tree of the coproduct definition and `or` builds the same tree but with the interpreters. They mirror each other

tests/src/test/scala/cats/tests/CoproductTests.scala
raulraja: @ceedubs Just added a test for that. I looked over the rest and I could not find one that wasn't in the hierarchy already but I'll be happy to add any extra ones if there are any missing.

raulraja:
@ceedubs @adelbertc @non addressed all your comments, let me know if anything else can be improved. thx!

free/src/main/scala/cats/free/Inject.scala
raulraja: done!

core/src/main/scala/cats/data/Coproduct.scala
```diff
@@ -0,0 +1,217 @@
...
+    def apply[G[_], A](ga: G[A]): Coproduct[F, G, A] = Coproduct(Xor.right(ga))
+  }
+
+  def left[G[_]]: CoproductLeft[G] = new CoproductLeft[G]
+
+  def right[F[_]]: CoproductRight[F] = new CoproductRight[F]
+
+}
+
+sealed abstract class CoproductInstances3 {
```
ceedubs: Sorry for the minor details, but could you make the `CodproductInstances*` traits `private[data]`? That will make them consistent with the work done in #612.

docs/src/main/tut/freemonad.md
ceedubs: I'm using absolutely zero CT here, but what if we added it to the `Free` companion object as `Free.inject`? It might also be helpful to follow the `PartiallyApplied` pattern we have in a couple other places to allow you to do something like `Free.inject[G](fa)` since the `F` and `A` type parameters could be inferred in that case. Just a thought.

core/src/main/scala/cats/data/Coproduct.scala
raulraja: No worries, I want to get it right so I understand what the expectations are for future contributions.:) Added those access modifiers also on `InjectInstances`

docs/src/main/tut/freemonad.md
ceedubs: I could see this living in either the `Coproduct` companion object or as a concrete method in the `NaturalTransformation` trait. I think that it might result in better inference and slightly nicer syntax to put it into `NaturalTransformation`, so I'm inclined to suggest that if nobody objects.

docs/src/main/tut/freemonad.md
raulraja: I'll be happy to add it there then if everyone else thinks it makes sense. @non @adelbertc thoughts?

docs/src/main/tut/freemonad.md
raulraja: +1 on adding it to `NaturalTransformation` @non @adelbertc thoughts?

free/src/main/scala/cats/free/Inject.scala
ceedubs: Oops sorry I wasn't very clear. The return types you added are nice, but I was thinking of the `implicit def`s. I think generally it's a good practice to put an explicit return type on all `implicit`s since it's a bit easy to accidentally have an overly specific type inferred.

tests/src/test/scala/cats/tests/CoproductTests.scala
ceedubs: If you look at the [code coverage diff](https://codecov.io/github/non/cats/compare/246512e0d813a7137f6c463779bef373bd126fec...4d00293c5ffc43fc2cb66cb3fb9fda94b4e6306c) it looks like the `Foldable` and `Coflatmap` instances aren't checked. The `Foldable` instance should be easy to check with something like [this](https://github.com/non/cats/blob/master/tests/src/test/scala/cats/tests/OneAndTests.scala#L31-L35) where you put only a `Foldable` but not a `Traverse` instance into the implicit scope. I don't think we have a handy helper for `Coflatmap`, so that might just require an explicit call to fetch the `Coflatmap` instance.

You've already done a lot with this PR. If you'd prefer, I'm happy to add these afterwards.

docs/src/main/tut/freemonad.md
non: Sounds good to me.

docs/src/main/tut/freemonad.md
non: Let's do it! I'm happy to do it in this PR, or another, whichever you feel like.

tests/src/test/scala/cats/tests/CoproductTests.scala
raulraja: I thought `ComonadTests` covered `CoflatMap` and `TraverseTests` covers `Foldable`. Is that not the case?. I see those tests have hierarchical laws for those types and I see my tests output showing those instances getting tested (Unless is testing some other instances not the ones I provided).

ceedubs: @raulraja you are correct. However, you have defined separate `Traverse` and `Foldable` instances. So when you check `TraverseTests`, you are validating that your `Traverse` instance is law-abiding, but the separate `Foldable` instance that you have defined isn't getting checked at all. It's a subtle issue that as of now requires a fair amount of boilerplate in our tests :(

raulraja: Makes sense, I'll add tests for those two instances as well

docs/src/main/tut/freemonad.md
raulraja: Done, promoted to `NaturalTransformation.or` and included test.

docs/src/main/tut/freemonad.md
raulraja: I just realized this same method is already part of `Inject`:
```scala
def inject[F[_], G[_], A](ga: G[Free[F, A]])(implicit I: Inject[G, F]): Free[F, A] =
    Free.liftF(I.inj(ga)) flatMap identity
```
Inference improvements apart, does it make sense to have it also on `Free`?
Currently is being used like so...
```scala
Inject.inject[T, Test1Algebra, Int](Test1(1))
```
Also @ceedubs can you point me to an example elsewhere where the `PartiallyApplied pattern` is in use? Where you referring to [something like this](https://github.com/propensive/rapture/blob/dev/core/shared/src/main/scala/result.scala#L30)

fthomas: @raulraja WRT `PartiallyApplied` see https://github.com/non/cats/pull/577

raulraja: Thanks @fthomas!

raulraja: @ceedubs @fthomas Do you think we can use the `PartiallyApplied` trick given `Free.inject[App](Test1(x))` would not be able to infer `TestAlgebra1`? `Test1` is an `F[_]` but our `Coproduct` composition depends on `TestAlgebra1`. I can take it as far as inferring `A` in `Test1[A]` with that trick but not sure how would I go about auto inferring the parent sealed trait in the ADT. 
This would fail not finding a `Inject[X, Test1]` implicit when really the one we are supposed to get is `Inject[X, Test1Algebra]`.
```scala
final class FreeInjectPartiallyApplied[G[_]] private[free] {
    def apply[F[_], A](fa: F[A])(implicit I : Inject[F, G]): Free[G, A] =
      Free.liftF(I.inj(fa))
  }

  def inject[G[_]]: FreeInjectPartiallyApplied[G] = new FreeInjectPartiallyApplied
```

raulraja: `Free.inject[F, G]` added. Unfortunately we need two params because the `Coproduct` you lift to is a parent of the type you pass to Free.inject which can't be inferred automatically as far as I know. If interested there is some more in depth discussion around it here https://gitter.im/non/cats?at=564cd3fa7ae518a71d2f9f98

raulraja:
@ceedubs @adelbertc @non Added `Free.inject[F[_], G[_]]`, `NaturalTransformation.or` with tests and `CoflatMap` and `Foldable` tests for `Coproduct`

tests/src/test/scala/cats/tests/CoproductTests.scala
```diff
@@ -0,0 +1,48 @@
...
+import cats.laws.discipline._
+import cats.laws.discipline.arbitrary._
+import org.scalacheck.Arbitrary
+
+class CoproductTests extends CatsSuite {
+
+  checkAll("Coproduct[Option, Option, ?]", TraverseTests[Coproduct[Option, Option, ?]].traverse[Int, Int, Int, Int, Option, Option])
+  checkAll("Traverse[Coproduct[Option, Option, ?]]", SerializableTests.serializable(Traverse[Coproduct[Option, Option, ?]]))
+
+  checkAll("Coproduct[Option, Option, ?]", FoldableTests[Coproduct[Option, Option, ?]].foldable[Int, Int])
```
ceedubs: Unfortunately, I think this will pick up the same instances as the `.traverse` check above. That's why I recommended the somewhat bizarre `ListWrapper.foldable` approach in another comment.

ceedubs:
:+1: :100:

non:
:+1: Looks good to me!

adelbertc:
:+1: Thanks @raulraja , awesome PR

PR #673, opened 2015-11-18
Add WriterT Bifunctor

ceedubs:
Knocks one item off of the checklist on #620.

### DIFF
--- a/core/src/main/scala/cats/data/WriterT.scala
+++ b/core/src/main/scala/cats/data/WriterT.scala
@@ -1,7 +1,7 @@
 package cats
 package data
 
-import functor.Bifunctor
+import cats.functor.Bifunctor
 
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/data/WriterT.scala
```diff
@@ -1,6 +1,8 @@
 package cats
 package data
 
+import functor.Bifunctor
```
adelbertc: Nitpicking, but given we seem to use the fully qualified name in other imports, maybe `import cats.functor.Bifunctor`

adelbertc:
:+1:

core/src/main/scala/cats/data/WriterT.scala
ceedubs: I like your style. Will do.

fthomas:
:+1:

PR #674, opened 2015-11-19
Minor cleanup in free monad doc

ceedubs:
This makes a couple small changes to the free monad docs.

1. Change most of the `tut` sheds to `tut:silent`. In most of these blocks, we are just defining objects/methods. There are only a few blocks where we are actually running things and care about the output. This makes things look a little cleaner.
2. One of the code blocks didn't use tut because it had a `readLine()` call that caused tut to hang. Unfortunately, this means that the code block wasn't compiler-verified, and it couldn't display its output. I've hacked around this by creating an invisible tut shed that defines `readLine` as a method that returns a canned String.

### DIFF
--- a/docs/src/main/tut/freemonad.md
+++ b/docs/src/main/tut/freemonad.md
@@ -76,7 +76,7 @@ recursive values.
 
 We need to create an ADT to represent our key-value operations:
 
-```tut
+```tut:silent
 sealed trait KVStoreA[+Next]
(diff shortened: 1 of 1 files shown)


ceedubs:
@raulraja what do you think about the canned string approach?

non:
:+1:

raulraja:
Looks great! Thanks for improving it! :+1:

fthomas:
:+1:

PR #676, opened 2015-11-19
Improve Streaming test coverage

ceedubs:
Previously many Streaming tests were always generating streams via
`fromList`. This meant that the Wait cases of streams were not getting
hit in tests.

Many of these tests now use an `Arbitrary[Streaming[A]]` instance to
generate streams, which should help hit the `Wait` case in tests.


fthomas:
:+1:

non:
:+1:

PR #680, opened 2015-11-19
Check Traverse laws for XorT

ceedubs:
There was a typo before that caused the XorT tests to only check the
Foldable laws for the Traverse instance.


non:
:+1:

adelbertc:
:+1:

PR #681, opened 2015-11-19
Add law-checking for Either Eq instance

ceedubs:
I think we were trying to do this before, but the approach we were
taking was testing the Order instance instead of the Eq instance.


non:
:+1:

adelbertc:
:+1:

PR #682, opened 2015-11-19
Check laws on Function1 Monoid

ceedubs:


non:
:+1:

tests/src/test/scala/cats/tests/FunctionTests.scala
```diff
@@ -25,4 +26,6 @@ class FunctionTests extends CatsSuite {
 
   checkAll("Function1[Int, Int]", ContravariantTests[? => Int].contravariant[Int, Int, Int])
   checkAll("Contravariant[? => Int]", SerializableTests.serializable(Contravariant[? => Int]))
+
+  checkAll("Function1[String, Int]", GroupLaws[Function1[String, Int]].monoid)
```
adelbertc: You've probably checked for this, but just to make sure :-) My first inclination is there's no `SerializableTests` because it can't be serialized, though it seems we're testing for it in the `Contravariant` instance?

ceedubs: The laws that come algebra [include serializability checks](https://github.com/non/algebra/blob/master/laws/shared/src/main/scala/algebra/laws/GroupLaws.scala#L28), so I don't think we need separate checks for those.

adelbertc: Ah right :-)

adelbertc:
:+1:

PR #683, opened 2015-11-19
Fix bug with Streaming#thunk.

### LINKED ISSUE #677: Streaming.thunk doesn't seem to reevaluate thunk
The ScalaDoc for `Streaming.thunk` says that the thunk may not be pure. However, it doesn't seem to support this use-case. In the following example, I would expect the output to be `List(1, 2, 3, 4)`, but it is `List(1, 1, 1, 1)`.

```scala
scala> var i = 0
i: Int = 0

scala> Streaming.thunk{() => i += 1; i}.take(4).toList
res6: List[Int] = List(1, 1, 1, 1)
```

Here is a test that helped me catch this:

```scala

  test("thunk is evaluated for each item") {
    // don't want the stream to be too big
    implicit val arbInt = Arbitrary(Gen.choose(-10, 20))
    forAll { (start: Int, end: Int) =>
      var i = start - 1
      val stream = Streaming.thunk{ () => i += 1; i}.takeWhile(_ <= end)
      stream.toList should === ((start to end).toList)
    }
  }
```

non:
The .thunk method is not supposed to memoize the results,
allowing the caller to provide an impure function which
returns different values on different calls.

However, .thunk's definition used .knot with memoization,
which completely defeated the point. This commit removes
the memoization allowing this method to function correctly.

Fixes #677.

### DIFF
--- a/core/src/main/scala/cats/data/Streaming.scala
+++ b/core/src/main/scala/cats/data/Streaming.scala
@@ -808,8 +808,8 @@ object Streaming extends StreamingInstances {
    * Continually return the result of a thunk.
    *
    * This method only differs from `continually` in that the thunk may
-   * not be pure. The stream is memoized to ensure that repeated
-   * traversals produce the same results.
+   * not be pure. Thus, repeated traversals may produce different
+   * results.
    */
   def thunk[A](f: () => A): Streaming[A] =
     knot(s => Cons(f(), s), memo = false)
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/data/Streaming.scala
```diff
@@ -812,7 +812,7 @@ object Streaming extends StreamingInstances {
    * traversals produce the same results.
```
ceedubs: I think this doc needs to be updated to no longer say that the result is memoized.

ceedubs:
:+1:

adelbertc:
:+1:

PR #685, opened 2015-11-19
Fix copy-paste naming errors (xor -> validated)

ms-tg:
Am I correct in assuming that these were:
  1.  supposed to say "validated"
  2.  probably copy-paste errors
  3.  in any case, not semantically meaningful, as they are usually imported as a group of implicits rather than individually?


ceedubs:
I'm pretty sure the answer the answer to all 3 questions is yes.

Thanks! :+1:

adelbertc:
:+1:

PR #686, opened 2015-11-19
Add test for Show[OneAnd]

mikejcurry:
Adds a test for the Show instance for OneAnd

### DIFF
--- a/tests/src/test/scala/cats/tests/OneAndTests.scala
+++ b/tests/src/test/scala/cats/tests/OneAndTests.scala
@@ -47,6 +47,14 @@ class OneAndTests extends CatsSuite {
   checkAll("NonEmptyList[Int]", ComonadTests[NonEmptyList].comonad[Int, Int, Int])
   checkAll("Comonad[NonEmptyList[A]]", SerializableTests.serializable(Comonad[NonEmptyList]))
 
+  test("Show is not empty and is formatted as expected") {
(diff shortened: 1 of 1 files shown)


ceedubs:
:+1: thanks!

It might be nice to have an exact value check for a particular fixed input too.

tests/src/test/scala/cats/tests/OneAndTests.scala
```diff
@@ -47,6 +47,14 @@ class OneAndTests extends CatsSuite {
   checkAll("NonEmptyList[Int]", ComonadTests[NonEmptyList].comonad[Int, Int, Int])
   checkAll("Comonad[NonEmptyList[A]]", SerializableTests.serializable(Comonad[NonEmptyList]))
 
+  test("Show is not empty and is formatted as expected") {
+    forAll { (nel: NonEmptyList[Int]) =>
+      nel.show.nonEmpty should === (true)
+      nel.show.startsWith("OneAnd") should === (true)
```
adelbertc: Unsure if we want to test specifically for "OneAnd" but I can see there being value in testing if the `.show` of the head is in the string or something

mikejcurry:
I think both of those make sense here. I'll look at them later today.

ceedubs:
:+1:

fthomas:
:+1:

PR #687, opened 2015-11-20
Comment out unused XorTMonadCombine

ceedubs:
This trait is currently unused, because the implicit def that uses it is
commented out until we have separated weak/strong laws for `MonadCombine`. We probably shouldn't expose it if it doesn't satisfy our current `MonadCombine` laws.


non:
:+1:

adelbertc:
:+1:

PR #689, opened 2015-11-21
Add WriterT MonadError instance

ceedubs:
This adds a `MonadError` instance for `WriterT`. However, it's not very useful in its current form. The following fails to compile, as the instance can't be found: `MonadError[WriterT[String Xor ?, Long, ?], String]`. Calling `writerTMonadError[String Xor ?, Long, String]` directly works though.

Does anyone know how to convince Scala to find this instance?

Even once it's found, there are some issues with the proper `Arbitrary` instances being found, but I guess we'll take it one step at a time.


core/src/main/scala/cats/data/WriterT.scala
```diff
@@ -149,6 +157,15 @@ private[data] sealed trait WriterTMonad[F[_], L] extends WriterTApplicative[F, L
     fa.flatMap(f)
 }
 
+private[data] sealed trait WriterTMonadError[F[_], L, E] extends MonadError[WriterT[F, L, ?], E] with WriterTMonad[F, L] {
+  override implicit def F0: MonadError[F, E]
+
+  def raiseError[A](e: E): WriterT[F, L, A] = WriterT(F0.raiseError[(L, A)](e))
+
+  def handleErrorWith[A](fa: WriterT[F, L, A])(f: E => WriterT[F, L, A]): WriterT[F, L, A] =
+    WriterT(F0.handleErrorWith(fa.run)(e => f(e).run))
```
lukiano: I wonder if you should append the L from fa.run and the L that results from f(e).run ?
In other words, if the Log that came from fa will be lost (maybe it's ok for it to be forgotten)

non:
@ceedubs Is this something you want to keep open? Still working on it?

core/src/main/scala/cats/data/WriterT.scala
ceedubs: @lukiano sorry I'm just seeing this. I don't think we can do what you have recommended. Let's consider `Future` as `F` (and thus `Throwable` as `E`). If the `Future` fails, then there is no log. We _do_ have a log if the future succeeds, but in that case the function that we pass to `handleErrorWith` will never be called, because there is no error to handle.

Does that sound right?

ceedubs:
@non I'll check if this is found for an `F[_]` type like `Future`. If so it may be at least a little useful. I also am interested in whether the `Xor` example works when the SI-2712 fix plugin is used. I'll get back on this shortly.

core/src/main/scala/cats/data/WriterT.scala
lukiano: I think you are right. I was thinking of something like XorT -> WriterT -> Future, but here F is just Future so there's no monadError instance for it. XorT will provide the MonadError and will already do the right thing.

ceedubs:
@non @lukiano @kailuowang I just updated this PR. Thanks to @kailuowang's `MonadError` instance for `Option`, I can see that this can potentially be useful.

Sorry -- I force pushed a rebase because I had made a mess of things.

ceedubs:
@non @lukiano @kailuowang updated again

kailuowang:
LGTM. Did you test `Xor` with SI-2712 fix, I will be surprised if it doesn't work.

ceedubs:
@non @kailuowang sorry about the delay on this one. I've resolved the merge conflict. I just did a check with a locally-published version, and `MonadError[WriterT[String Xor ?, Long, ?], String]` does indeed resolve just fine with the SI-2712 fix plugin. I think this should be good to merge if the Travis build succeeds.

ceedubs:
I've added some (gnarly) tests for the `ApplicativeError` instance. I look forward to SI-2712 making this a lot cleaner.

adelbertc:
LGTM 👍

ceedubs:
@kailuowang does this still look good to you?

kailuowang:
👍 LGTM

Issue #117, opened 2015-02-07
Add instances for std Either


ceedubs:
`Either` from the standard library is isomorphic to `Or` and should have instances for all of the same type classes that `Or` does.

wedens:
I'll take it. Mark as in progress, please.

PR #690, opened 2015-11-21
Pass along error when serializable tests fail

ceedubs:
This is the same work that was done in algebra in
[#117](https://github.com/non/algebra/pull/117).


adelbertc:
:+1:

non:
:+1:

Issue #684, opened 2015-11-19
SemigroupK and MonoidK instances for Function1


ceedubs:
#478 added `SemigroupK` and `MonoidK` instances for `Kleisli`. We should be able to provide similar instances for `Function1`.

I think we can also add a `Semigroup` instance for `Function1`, as it currently is missing this but has a `Monoid` instance (with a stricter constraint on `B`).

lvicentesanchez:
I could do this; I have some free time this weekend.

ceedubs:
Thanks @lvicentesanchez! I'v marked it as in progress.

mikejcurry:
Should this close now? I don't have rights to do so.

adelbertc:
Yep, thanks @mikejcurry

PR #691, opened 2015-11-21
Add MonoidK and SemigroupK instances for Function1 and also a Semigroup instance for Function1

lvicentesanchez:
This PR address #684 by adding MonoidK and SemigroupK instances for Function1; it also adds a Semigroup instance for Funcion1.

### DIFF
--- a/core/src/main/scala/cats/std/function.scala
+++ b/core/src/main/scala/cats/std/function.scala
@@ -6,7 +6,7 @@ import cats.arrow.{Arrow, Choice}
 import cats.data.Xor
 import cats.functor.Contravariant
 
-trait Function0Instances {
+private[std] sealed trait Function0Instances {
   implicit val function0Instance: Bimonad[Function0] =
     new Bimonad[Function0] {
       def extract[A](x: () => A): A = x()
@@ -26,7 +26,7 @@ trait Function0Instances {
     }
 }
 
-trait Function1Instances extends Function1Instances0 {
+private[std] sealed trait Function1Instances extends Function1Instances0 {
   implicit def function1Contravariant[R]: Contravariant[? => R] =
     new Contravariant[? => R] {
       def contramap[T1, T0](fa: T1 => R)(f: T0 => T1): T0 => R =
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/std/function.scala
```diff
@@ -71,13 +71,41 @@ trait Function1Instances {
...
-      }
-    }
+  implicit def function1Monoid[A,B](implicit M: Monoid[B]): Monoid[A => B] =
+    new Function1Monoid[A, B] { def B: Monoid[B] = M }
+
+  implicit val function1MonoidK: MonoidK[Lambda[A => A => A]] =
+    new Function1MonoidK {}
+}
+
+trait Function1Instances0 {
```
ceedubs: Could you make this `private[std] sealed trait Function1Instances0` please? This will make it consistent with the work done in #612.

core/src/main/scala/cats/std/function.scala
```diff
@@ -71,13 +71,41 @@ trait Function1Instances {
...
+  implicit def function1Semigroup[A,B](implicit S: Semigroup[B]): Semigroup[A => B] =
+    new Function1Semigroup[A, B] { def B: Semigroup[B] = S }
+
+  implicit val function1SemigroupK: SemigroupK[Lambda[A => A => A]] =
+    new Function1SemigroupK {}
+}
+
+private[std] trait Function1Semigroup[A, B] extends Semigroup[A => B] {
```
ceedubs: It would probably be good to go ahead make this and the other similar traits `sealed`.

ceedubs:
I left a couple very minor comments, but other than that, this looks great! Thank you.

core/src/main/scala/cats/std/function.scala
lvicentesanchez: Sure, I would do both things; and extend it to all traits in the file as per #612 :)

lvicentesanchez:
Ok, that should do it :)

ceedubs:
:+1: looks great!

non:
Thanks! :+1:

PR #697, opened 2015-11-23
Move NaturalTransformation.or from companion to trait

ceedubs:
I think this provides slightly nicer syntax for combining natural
transformations, and it makes it more consistent with `andThen` and
`compose`. If people like it, they can merge, but if not it's no
problem.


ceedubs:
@raulraja as the original contributor of `or`, what do you think?

raulraja:
LGTM! Also `or` as infix in the examples looks so much cleaner

lvicentesanchez:
I like it. That would help me to get rid of my own implicit/explicit `or` that I use to combine more than 2 NT.

non:
:+1:

adelbertc:
:+1:

Issue #694, opened 2015-11-22
Separate Strong/Weak Laws for MonadFilter/MonadCombine


mikejcurry:
#693 and #687 refer to Strong and Weak Laws for `MonadFilter`/`MonadCombine`. In #687 the `MonadCombine` instance for `XorT` was commented out as it was unused. It was unused due to the fact that the `MonadCombine` instance for `XorT` could not meet some of the laws for `MonadFilter`, namely: right absorbtion, right distributivity, and left distributivity. #693 will make a similar change for `OptionT` and reduce the available instance from `MonadCombine` to `Monad`.

It is suggested that these need to be separated out into Weak and Strong Laws, at which point an instance, supporting the appropriate weaker laws, can be reinstated for both `XorT` and `OptionT`.

kailuowang:
Some of the problematic laws are in `AlternativeLaws`, which is part of the #1210 refactor. Hence, blocked by #1210

kailuowang:
No longer needed due to #1616, moved out of the 1.0.0-MF, will close once mtl removal plan is completed.

Issue #693, opened 2015-11-21
OptionT implements MonadCombine, but doesn't meet the laws


mikejcurry:
Whilst looking at `OptionT` today and adding some tests in #692, I noticed that the `OptionT` has a `MonadCombine` instance, plus implementations for empty and combine. 

The tests exercise `MonadCombineTests.monad` instead of `MonadCombineTests.monadCombine`, which means that the implementations of `empty` and `combine` are not tested against the laws (or anything).

I briefly looked at increasing the test to `monadCombine`, but the current implementation of `OptionT` does not pass the laws, at least not for `OptionT[List, Int]`.

I noticed that #687 proposes to remove, at least temporarily, the `MonadCombine` for `XorT` - if a similar approach is warranted here I can take a look. By similiar in this instance, I mean that it might make sense to reduce the implementation to `Monad` until the laws pass.

ceedubs:
Good catch! Thank you for bringing this to attention.

I think that the right approach would probably be to:
1. Remove this instance
2. Make sure we have an issue for separating the weak and strong laws suggested by those comments in the `XorT` code
3. Create new issues for adding back in the `XorT` and `OptionT` instances once the laws work has been finished.

How does that sound?

mikejcurry:
That sounds sensible. I can put together a PR(after #692 - as both PRs will touch OptionTTests) and create issues for tracking.

ceedubs:
I'm going to go ahead and close this out since it was resolved with #698 and #695 tracks the follow-up.

PR #698, opened 2015-11-24
Reduce OptionT from MonadCombine to Monad

mikejcurry:
Reduces `OptionT` `MonadCombine` instance to `Monad`. 

The `MonadCombine` instance for `OptionT` was not tested and does not pass laws.

This addresses #693. 

#694 exists for tracking potential changes to laws which may allow this instance, or an alternative type class instance, to be reinstated.


ceedubs:
:+1: thanks!

adelbertc:
:+1:

PR #702, opened 2015-11-27
fix foldMap stack safety

svalaskevicius:


free/src/main/scala/cats/free/Free.scala
```diff
@@ -130,7 +130,20 @@ sealed abstract class Free[S[_], A] extends Product with Serializable {
...
-  final def foldMap[M[_]](f: S ~> M)(implicit M: Monad[M]): M[A] =
+  @tailrec
+  final def foldMap[M[_]](f: S ~> M)(implicit M: Monad[M]): M[A] = {
+    step match {
+      case Free.Pure(a) => M.pure(a)
+      case Free.Suspend(s) => f(s)
+      case Free.Gosub(c, g) => c match {
+        case Free.Pure(a) => g(M.pure(a)).foldMap(f)
+        case Free.Suspend(s) => g(f(s)).foldMap(f)
+        case Free.Gosub(c1, g1) =>  g(M.flatMap(c1.foldMapRecursiveStep(f))(cc => g1(cc).foldMapRecursiveStep(f))).foldMap(f)
```
svalaskevicius: tbh this line should never happen as its resolved by `step` - should we simple throw an exception and remove `foldMapRecursiveStep` function?

svalaskevicius: will remove `Free.` prefixes

also, looks like the only thing thats not handled by step is `Suspend`..

svalaskevicius:
how to kick @codecov-io again?

free/src/main/scala/cats/free/Free.scala
```diff
@@ -130,12 +130,17 @@ sealed abstract class Free[S[_], A] extends Product with Serializable {
...
-  final def foldMap[M[_]](f: S ~> M)(implicit M: Monad[M]): M[A] =
+  @tailrec
+  final def foldMap[M[_]](f: S ~> M)(implicit M: Monad[M]): M[A] = {
     step match {
       case Pure(a) => M.pure(a)
       case Suspend(s) => f(s)
-      case Gosub(c, g) => M.flatMap(c.foldMap(f))(cc => g(cc).foldMap(f))
+      case Gosub(c, g) => c match {
+        case Suspend(s) => g(f(s)).foldMap(f)
+        case _ => throw new Error("Unexpected operation. The case should have been eliminated by `step`.")
```
svalaskevicius: Need to check where else is step used, could it be moved to this function to avoid the exception throwing or does it add value on its own?

adelbertc:
:+1:

ceedubs:
Thanks, @svalaskevicius!

This looks good to me. Does anyone else have any concerns? @mandubian?

mandubian:
:+1: good catch... I wonder why we have missed this before ;)

PR #704, opened 2015-11-28
Adds test for Show[Const]

mikejcurry:
Adds a test for `Show[Const]`, includes:
* Simple test of a specific value
* Test that all output starts with “Const(“
* Test that the `Show` for the result of `const.getConst` is contained in the result
* Test that implicitly obtained `Show` instance produces a result consistent with direct method invocation.
* Tests that retagging the `Const` does not affect the `Show` result

The last one raises an interesting question. The phantom type of the `Const` might be interesting enough information in some cases. The question is whether it is interesting enough to show.

Basically, `Const[Int, String]` and `Const[Int, Boolean]` produce the same show result. I implemented a test for the behaviour primarily to demonstrate that it is the case, and to raise a discussion on whether it warrants inclusion.


stew:
:+1:

non:
:+1:

Issue #537, opened 2015-09-21
Should NonEmptyList have its own data type?


coltfred:
NonEmptyList is currently `type NonEmptyList[A] = OneAnd[A, List]`. There has been discussion about writing a version of it that is more so that we could support a more list like interface (++, tails, sortBy, etc).

How does everyone feel about it? Is it worth doing?

julien-truffaut:
it will be convenient to define custom methods but what about `NonEmptySet`, `NonEmptyMap`, `NonEmptyVector` ...

I recall that @stew wanted to use a separate project for "fancy" data structure

refried:
I'm not necessarily suggesting it, but another possibility would be to define List-like syntax for `OneAnd[A, List]` etc.

coltfred:
@refried I had thought about that... Doesn't seem like an awful idea. Does anyone have reasons why that's an awful idea?

refried:
@coltfred The main down-side I can think of is that it requires a non-obvious import to get the `NonEmptyList` syntax and another import to get the `List` TC instances.  (Maybe they could be combined).  Up-side is not having to define `NonEmptyList` TC instances.

stew:
@julien-truffaut I did recently start the https://github.com/stew/dogs project, but I think NonEmptyList is going to end up staying here, because we want to use it in cats itself.

refried:
@stew lol at bird beaks

julien-truffaut:
@stew could we use `OneAnd[List]` within cats and define all NonEmpty classes in dogs? I don't know if it makes sense

adelbertc:
On a related note I wonder if `NonEmpty` would be a better name than `OneAnd`

refried:
@adelbertc I like this.

adelbertc:
Going back, I think `NonEmptyList` has enough usages to warrant it being in Cats. If people agree with that, I don't think it's a stretch to want to include `Vector`, if only for it's faster `++` operation in the case of say, using it as the `Semigroup` for `Validated` (in which case maybe we change (or add) `ValidatedNel` with `ValidatedNev`) and it's fast random access properties.

Personally I've only found myself reaching for a non-empty `Set` once, and never for a `NonEmptyMap`. On that note, both of those have unique characteristics that may warrant their own special implementation instead of using something like `OneAnd`.

[...]

PR #706, opened 2015-12-01
Add documentation for `OneAnd`

markus1189:
Hi all, I wrote a quick introduction to the `OneAnd` data type.  It is not especially fancy but it should at least give a brief overview and is better than an empty page ;)

I'd be happy to apply any additions / modifications, so let me know.

Best,
Markus

-- 24pullrequests

### DIFF
--- a/docs/src/main/tut/oneand.md
+++ b/docs/src/main/tut/oneand.md
@@ -7,3 +7,24 @@ scaladoc: "#cats.data.OneAnd"
 ---
 # OneAnd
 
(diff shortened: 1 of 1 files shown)


ceedubs:
:+1: thanks!

If #537 results in `NonEmptyList` having its own data type, we should try to remember to update these docs to omit "which is the actual implementation of non-empty lists in cats" :)

adelbertc:
:+1:

PR #709, opened 2015-12-02
make tut doc ever so slightly less hideous

tpolecat:
This PR adds the `:silent` modifier to code blocks that contain only definitions (and in a few cases splits mixed blocks). The resulting **tut** doc is much cleaner.

### DIFF
--- a/docs/src/main/tut/validated.md
+++ b/docs/src/main/tut/validated.md
@@ -41,7 +41,7 @@ As our running example, we will look at config parsing. Our config will be repre
 `Map[String, String]`. Parsing will be handled by a `Read` type class - we provide instances
 just for `String` and `Int` for brevity.
 
-```tut
+```tut:silent
 trait Read[A] {
   def read(s: String): Option[A]
 }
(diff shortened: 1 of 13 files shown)


docs/src/main/tut/validated.md
```diff
@@ -296,16 +302,21 @@ val houseNumber = config.parse[Int]("house_number").andThen{ n =>
...
 
+Thus.
+
+```tut
 val houseNumber = config.parse[Int]("house_number").withXor{ xor: ConfigError Xor Int =>
   xor.flatMap{ i =>
-    positive(i)
+    positive("house_number", i)
   }
+}
```
tpolecat: This compile error wasn't detected because the file ended with a code block in the didn't have its closing `}` so the REPL was left hanging. Nice work isolating a corner case!

ceedubs:
Wow, thanks for going through and doing this, @tpolecat! :+1:

docs/src/main/tut/validated.md
ceedubs: For the record, this is because of https://github.com/tpolecat/tut/issues/65.

adelbertc:
:+1:

PR #710, opened 2015-12-02
Documentation for the `Invariant` typeclass

### LINKED ISSUE #164: Add better documentation for Invariant
We just had a conversation about what `Invariant` / `imap` was for on Gitter.

@mpilquist provided a link to a blog post he wrote about it: http://mpilquist.github.io/blog/2013/06/01/scodec-part-2/

@coltfred suggests that we could either link to that, or summarize it in a block comment. Obviously we don't have to get this perfect immediately, we can just make incremental progress.

markus1189:
Adds a `tut` section about `Invariant` as discussed in #164.  I oriented the explanation according to @ceedubs example, I hope this is okay.

I wasn't able to test the `jekyll serve` because it fails on my pc due to a missing ruby dependency, so I am not sure if adding the file is enough to make it show up on the [typeclasses](http://non.github.io/cats//typeclasses.html) site?  If someone could check that that'd be great :).

If merged this fixes #164.

-- 24pullrequests

### DIFF
--- a/docs/src/main/tut/invariant.md
+++ b/docs/src/main/tut/invariant.md
@@ -1,0 +1,107 @@
+---
+layout: default
+title:  "Invariant"
+section: "typeclasses"
+source: "https://github.com/non/cats/blob/master/core/src/main/scala/cats/functor/Invariant.scala"
+scaladoc: "#cats.functor.Invariant"
+---
+# Invariant
+
+The `Invariant` typeclass is for functors that define an `imap`
+function with the following type:
+
+```scala
+def imap[A, B](fa: F[A])(f: A => B)(g: B => A): F[B]
+```
+
+Every covariant (as well as contravariant) functor gives rise to an invariant
+functor, by ignoring the `f` (or in case of contravariance, `g`) function.
+
+Examples for instances of `Invariant` are `Semigroup` and `Monoid`, in
+the following we will explain why this is the case using `Semigroup`, the
+reasoning for `Monoid` is analogous.
+
+## Invariant instance for Semigroup
+
+Pretend that we have a `Semigroup[Long]` representing a standard UNIX
+timestamp.  Let's say that we want to create a `Semigroup[Date]`, by
+*reusing* `Semigroup[Long]`.
+
+### Semigroup does not form a covariant functor
+
+If `Semigroup` had an instance for the standard covariant `Functor`
+typeclass, we could use `map` to apply a function `longToDate`:
+
+```tut:silent
+import java.util.Date
+def longToDate: Long => Date = new Date(_)
+```
+
+But is this enough to give us a `Semigroup[Date]`?  The answer is no,
+unfortunately.  A `Semigroup[Date]` should be able to combine two
+values of type `Date`, given a `Semigroup` that only knows how to
+combine `Long`s!  The `longToDate` function does not help at all,
+because it only allows us to convert a `Long` into a `Date`.  Seems
+like we can't have an `Functor` instance for `Semigroup`.
+
+### Semigroup does not form a contravariant functor
+
+On the other side, if `Semigroup` would form a *contravariant* functor
+by having an instance for `Contravariant`, we could make use of
+`contramap` to apply a function `dateToLong`:
+
+```tut:silent
+import java.util.Date
+def dateToLong: Date => Long = _.getTime
+```
+
+Again we are faced with a problem when trying to get a
+`Semigroup[Date]` based on a `Semigroup[Long]`.  As before consider
+the case where we have two values of `Date` at hand.  Using
+`dateToLong` we can turn them into `Long`s and use `Semigroup[Long]`
+to combine the two values.  We are left with a value of type `Long`,
+but we can't turn it back into a `Date` using only `contramap`!
+
+### Semigroup does form an invariant functor
+
+From the previous discussion we conclude that we need both the `map`
+from (covariant) `Functor` and `contramap` from `Contravariant`.
+There already is a typeclass for this and it is called `Invariant`.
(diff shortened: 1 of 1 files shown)


ceedubs:
Thanks! Looks good at a first glance. I'll take a more detailed look.

docs/src/main/tut/invariant.md
```diff
@@ -0,0 +1,78 @@
+---
+layout: default
+title:  "Invariant"
+section: "typeclasses"
+source: "https://github.com/non/cats/blob/master/core/src/main/scala/cats/functor/Invariant.scala"
+scaladoc: "#cats.functor.Invariant"
+---
+# Invariant
+
+The `Invariant` typeclass is for `Functor`s that define an `imap`
```
ceedubs: We should probably be careful with wording here. In Cats, `Functor` means covariant functor. To me, this sentence makes it sound like `Invariant` is a subtype of `Functor`, while it's actually the other way around. This may be solved by just changing `Functor` to a generic "functor". That could potentially also be confusing to people who think that `Functor` and "functor" would be synonymous. It may be fine though.

docs/src/main/tut/invariant.md
```diff
@@ -0,0 +1,78 @@
...
+## Invariant instance for Semigroup
+
+Pretend that we have a `Semigroup[Long]` representing a standard UNIX
+timestamp.  Let's say that we want to create a `Semigroup[Date]`, by
+*reusing* `Semigroup[Long]`.
+
+### Semigroup does not form a covariant functor
+
+If `Semigroup` had an instance for the standard covariant `Functor`
+typeclass, we could use `map` to apply a function `longtoDate`:
```
ceedubs: Minor: `t` in `longtoDate` should be capitalized.

docs/src/main/tut/invariant.md
```diff
@@ -0,0 +1,78 @@
...
+
+```scala
+def longToDate: Long => Date = new Date(_)
```
ceedubs: Can we turn this into a `tut:silent` shed to enforce that it compiles?

ceedubs: This comment applies to some other places below. You can use `sbt docs/tut` to make sure the docs compile after these changes.

docs/src/main/tut/invariant.md
```diff
@@ -0,0 +1,78 @@
...
+
+```scala
+def imap[A, B](fa: F[A])(f: A => B)(g: B => A): F[B]
+```
+
+Reusing the example of turning `Semigroup[Long]` into
+`Semigroup[Date]`, we can use the `g` parameter to turn `Date` into a
+`Long`, combine our two values using `Semigroup[Long]` and then
+convert the result back into a `Date` using the `f` parameter of
+`imap`.
```
ceedubs: As a grand finale, can we show the code that actually converts the `Semigroup[Long]` to a `Semigroup[Date]`?

markus1189: Do you have an specific way in mind to show this?  By using it in an example, or did you mean something else?

ceedubs: I simply meant adding something like:

```scala
val dateSemigroup: Semigroup[Date] = Semigroup[Long].imap(longToDate)(dateToLong)
```

markus1189:
Added a full example at the end ;)

docs/src/main/tut/invariant.md
```diff
@@ -0,0 +1,99 @@
...
+import cats.syntax.all._
+
+def longToDate: Long => Date = new Date(_)
+def dateToLong: Date => Long = _.getTime
+
+implicit val semigroupDate: Semigroup[Date] =
+  Semigroup[Long].imap(longToDate)(dateToLong)
+
+val today: Date = longToDate(1449088684104l)
+val timeLeft: Date = longToDate(1900918893l)
```
ceedubs: In the spirit of the work that has just been done for #709, maybe we should put all but the last line of this into a `tut:silent` shed, and just use `tut` (with output) for the `today |+| timeLeft`. The output for imports, etc is pretty noisy.

docs/src/main/tut/invariant.md
```diff
@@ -0,0 +1,99 @@
...
+`Long`, combine our two values using `Semigroup[Long]` and then
+convert the result back into a `Date` using the `f` parameter of
+`imap`:
+
+```tut
+import java.util.Date
+
+import cats._
+import cats.std.all._
+import cats.syntax.all._
```
ceedubs: Warning: this is definitely me being nitpicky, so take it or leave it.

It seems to me like we should either combine the two `.all._` imports into the single `import cats.implicits._` or we should selectively import just what's needed for this example (which I think would be `cats.std.long._` and `cats.syntax.semigroup._`). The former might be a little more friendly for a tutorial.

docs/src/main/tut/invariant.md
```diff
@@ -0,0 +1,99 @@
...
+
+The `Invariant` typeclass is for functors that define an `imap`
+function with the following type:
+
+```scala
+def imap[A, B](fa: F[A])(f: A => B)(g: B => A): F[B]
+```
+
+Every covariant (as well as contravariant) functor gives rise to an invariant
+functor, by ignorig the `f` (or in case of contravariance, `g`) function.
```
mikejcurry: ignorig => ignoring

markus1189:
I added two versions of the import, the version that imports all the stuff and a commented out version that is more explicit.   @mikejcurry good catch for the typo :)

ceedubs:
I ran `jekyll serve` locally and the `Invariant` type class auto-populated in the links and looks good. :+1: thanks!

adelbertc:
:+1:

Issue #712, opened 2015-12-03
Free.foldMap throws a java.lang.ClassCastException


franciscodr:
As commented in a [Gitter conversation](https://gitter.im/non/cats?at=5660b8dd1da31ad605a8c321), when I try to transform a free monad to another monad like `Option`, `List` or `Future` using `foldMap`, I get a `java.lang.ClassCastException`.

I'm using cats 0.4.0-SNAPSHOT

ceedubs:
Thanks for reporting, @franciscodr! I'll start to take a look at this.

franciscodr:
@ceedubs After merging https://github.com/non/cats/pull/713, the example code works fine again. Thanks!

PR #713, opened 2015-12-04
Revert regression in Free.foldMap

ceedubs:
This PR does a couple of things.

Firstly, it includes `Gosub` in the instances generated by the `Free` `Arbitrary` instance. This reveals a bug that caused the "mapSuspension id" test to fail.

It also reverts most of the changes made in #702. This fixes the failing "mapSuspension id" test and #712. Unfortunately, it comes at the cost of reintroducing the stack safety issue with `Free.foldMap`, and therefore I have its stack-safety unit test ignored for now. I'm hoping that we can address this issue with a separate PR in the future.

### DIFF
--- a/free/src/main/scala/cats/free/Free.scala
+++ b/free/src/main/scala/cats/free/Free.scala
@@ -74,6 +74,14 @@ sealed abstract class Free[S[_], A] extends Product with Serializable {
   final def fold[B](r: A => B, s: S[Free[S, A]] => B)(implicit S: Functor[S]): B =
     resume.fold(s, r)
 
+  /** Takes one evaluation step in the Free monad, re-associating left-nested binds in the process. */
+  @tailrec
+  final def step: Free[S, A] = this match {
+    case Gosub(Gosub(c, f), g) => c.flatMap(cc => f(cc).flatMap(g)).step
+    case Gosub(Pure(a), f) => f(a).step
+    case x => x
+  }
+
   /**
    * Evaluate a single layer of the free monad.
    */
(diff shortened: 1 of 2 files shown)


ceedubs:
@svalaskevicius to be honest I don't understand why your change broke this. If you can find an alternative that fixes the "mapSuspension id" test and #712 but is still stack safe, that would be ideal.

svalaskevicius:
Hmm it looks I've made an error earlier in gosub suspend case and missed flatmap.. Will check later to confirm

non:
I think we should revert now (to remove bug) and then submit a new PR to fix the actual issue.

:+1:

svalaskevicius:
had a quick look.. looks like I made it tail recursive but removing the flatMap... which is required if we want to keep the structure.

so when I add it back:
```scala
case Gosub(c, g) => c match {
    case Suspend(s) => M.flatMap(f(s))(cc => g(cc).foldMap(f))
```

the recursion is not in the tail position any more and I'm not sure how to make it so.. any ideas?

svalaskevicius:
@non, @ceedubs:

actually, does foldMap have to preseve the structure? 

e.g. on hackage its `Map each element of the structure to a monoid, and combine the results.`

maybe it would make sense to leave structure preservation to mapSuspension only?

the reason being is that I don't see a way to make it both stack safe and structure preserving (using flatmap there) - that would be the ultimate solution :)

ceedubs:
I'm going to go ahead and merge this since `Free` isn't really usable at the moment. I'll create a new issue for its stack safety, as @non has suggested.

PR #714, opened 2015-12-05
Add XorT.orElse

rossabaker:


rossabaker:
The second test is so we don't repeat scalaz/scalaz#1037.

stew:
:+1:

adelbertc:
:+1:

ceedubs:
I don't think AA needs to be a subtype of A.

ceedubs:
I don't think As needs to be a supertype of A.

PR #715, opened 2015-12-05
Adds Monoid instance for Validated

mikejcurry:
Adds a `Monoid` instance for `Validated`


adelbertc:
:+1:

ceedubs:
:+1: thanks!

It would also be nice to have a `Semigroup` instance if `B` has a `Semigroup` but not a `Monoid`.

PR #716, opened 2015-12-06
Adds Semigroup for Validated

mikejcurry:
Added a `Semigroup` for `Validated` based on comment:
https://github.com/non/cats/pull/715#issuecomment-162326224

On a side note, I think there is at least one other `Monoid` instance (`Xor`) that has the same characteristics, i.e. no separate `Semigroup` instance . Seems based on the referenced comment it would be good to add these too?

### DIFF
--- a/core/src/main/scala/cats/data/Validated.scala
+++ b/core/src/main/scala/cats/data/Validated.scala
@@ -236,6 +236,12 @@ private[data] sealed abstract class ValidatedInstances extends ValidatedInstance
 }
 
 private[data] sealed abstract class ValidatedInstances1 extends ValidatedInstances2 {
+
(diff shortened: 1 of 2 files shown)


ceedubs:
Thanks! Looks good to me. :+1: if/when the build passes.

:+1: for your suggestions of an `Xor` `Semigroup` as well. Good catch.

mikejcurry:
Pushing a change as per #717 to keep things in line with naming conventions for tests. More specifically, changing the "Monoid[Validated[String, Int]]" to "Validated[String, Int]"

non:
:+1:

non:
Thanks!

PR #717, opened 2015-12-06
Adds Semigroup for Xor

mikejcurry:
As per #716, adding a `Semigroup` for the `Xor` data type which currently
has a `Monoid` instance but no `Semigroup` instance.

### DIFF
--- a/tests/src/test/scala/cats/tests/XorTests.scala
+++ b/tests/src/test/scala/cats/tests/XorTests.scala
@@ -12,8 +12,9 @@ import org.scalacheck.Arbitrary._
 import scala.util.Try
 
 class XorTests extends CatsSuite {
-  checkAll("Monoid[Xor[String, Int]]", GroupLaws[Xor[String, Int]].monoid)
-  checkAll("Semigroup[Xor[String, NonEmptyList[Int]]]", GroupLaws[Xor[String, NonEmptyList[Int]]].semigroup)
+  checkAll("Xor[String, Int]", GroupLaws[Xor[String, Int]].monoid)
+
+  checkAll("Xor[String, NonEmptyList[Int]]", GroupLaws[Xor[String, NonEmptyList[Int]]].semigroup)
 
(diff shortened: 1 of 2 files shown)


tests/src/test/scala/cats/tests/XorTests.scala
```diff
@@ -12,7 +12,8 @@ import org.scalacheck.Arbitrary._
 import scala.util.Try
 
 class XorTests extends CatsSuite {
-  checkAll("Xor[String, Int]", GroupLaws[Xor[String, Int]].monoid)
+  checkAll("Monoid[Xor[String, Int]]", GroupLaws[Xor[String, Int]].monoid)
+  checkAll("Semigroup[Xor[String, NonEmptyList[Int]]]", GroupLaws[Xor[String, NonEmptyList[Int]]].semigroup)
```
ceedubs: Minor: `monoid`/`semigroup` are already included in the test names, so I think the convention has been to not include `Monoid`/`Semigroup` in the string argument.

ceedubs:
:+1: if/when the build passes.

@mikejcurry you are on a roll again! :)

tests/src/test/scala/cats/tests/XorTests.scala
mikejcurry: I see - I can certainly change this.

mikejcurry:
Pushed a change to keep things in line with naming conventions for tests. Maybe I should add this convention to the contributors guide?

ceedubs:
@mikejcurry that would be great! Thank you. I think the sole exception is the Cats serializability test, since the same one is used for all types. Actually, if you are already going to make a change, could you add that all type class instances should also be checked with the serializability tests unless they are being checked by Algebra's laws (`GroupLaws`, `OrderLaws`, etc), which already test serializability? This helps keep us Spark-friendly (and other tools that require java Serializability).

mikejcurry:
Sure, have to head away for a while now, so it may not be today, but I'll get to it some time soon.

adelbertc:
:+1:

PR #718, opened 2015-12-06
Adds Semigroup for Const

mikejcurry:
Adding a `Semigroup` for `Const` to allow `Semigroup` behaviour for data
types with no valid `Monoid` instance.


non:
Looks good to me. :+1:

ceedubs:
:+1:

PR #720, opened 2015-12-07
More tut for FreeApplicative

adelbertc:


docs/src/main/tut/freeapplicative.md
```diff
@@ -5,10 +5,167 @@ section: "data"
...
+    def apply[A](fa: ValidationOp[A]): String => A =
+      str =>
+        fa match {
+          case Size(size) => str.size >= size
+          case HasNumber  => str.exists(c => "0123456789".contains(c))
+        }
+  }
+
+val validator = prog.foldMap[String => ?](compiler)
+validator("1234")
```
ceedubs: Minor formatting suggestion: maybe put all but the last two lines of this into a `tut:silent` block to avoid noise?

docs/src/main/tut/freeapplicative.md
```diff
@@ -5,10 +5,167 @@ section: "data"
...
+val logCompiler =
+  new NaturalTransformation[ValidationOp, Log] {
+    def apply[A](fa: ValidationOp[A]): Log[A] =
+      fa match {
+        case Size(size) => Const(List(s"size >= $size"))
+        case HasNumber  => Const(List("has number"))
+      }
+  }
+
+val logValidation = prog.foldMap[Log](logCompiler)
```
ceedubs: It might be nice to put a non-silent tut block under this one that feeds some input in and shows the output.

ceedubs:
I left a couple minor comments, but this is great! Thanks @adelbertc.

docs/src/main/tut/freeapplicative.md
```diff
@@ -5,10 +5,176 @@ section: "data"
 source: "https://github.com/non/cats/blob/master/core/src/main/scala/cats/free/FreeApplicative.scala"
 scaladoc: "#cats.free.FreeApplicative"
 ---
-# Free Applicative Functor
+# Free Applicative
 
-Applicative functors are a generalization of monads allowing expressing effectful computations in a pure functional way.
+`FreeApplicative`s are similar to `Free` (monads) in that they provide a nice way to represent
+computations as data and are useful for building embedded DSLs (EDSLs). However, they differ in
```
fthomas: s/differ in from/differ from/ ?

docs/src/main/tut/freeapplicative.md
```diff
@@ -5,10 +5,176 @@ section: "data"
...
+```tut
+logValidation(prog)
+logValidation(size(5) *> hasNumber *> size(10))
+logValidation((hasNumber |@| size(3)).map(_ || _))
+```
+
+### Why not both?
+It is perhaps more plausible and useful to have both the actual validation function and the logging
+strings. While we could easily compile our program twice, once for each interpreter as we have above,
+we could also do it in one go - this would avoid multiple traversals or the same structure.
```
fthomas: s/or/of/

fthomas:
Great work, @adelbertc!

mikejcurry:
I think this doc is great and really walks you through this - and the example is great because it is practical!

ceedubs:
:+1:

PR #722, opened 2015-12-07
Make StreamingT toString more like Streaming toString

ceedubs:
Before this, StreamingT was always returning "StreamingT(...)" for its
toString. It's now similar to Streaming in that it will return
"StreamingT()" in the case of Empty and will show the head element in
the case of Cons.

I also added a pretty weak test for its toString (which actually doesn't
capture my changes at all). It may be good to add further tests in the
future, but I've got to get to work, and I think this is at least an
improvement to the current status.


adelbertc:
:+1:

PR #725, opened 2015-12-08
Remove AA >: A constraint from XorT.orElse

ceedubs:


adelbertc:
:+1:

core/src/main/scala/cats/data/XorT.scala
```diff
@@ -29,11 +29,11 @@ final case class XorT[F[_], A, B](value: F[A Xor B]) {
     }
   }
 
-  def orElse[AA >: A, BB >: B](default: => XorT[F, AA, BB])(implicit F: Monad[F]): XorT[F, AA, BB] = {
+  def orElse[AA, BB >: B](default: => XorT[F, AA, BB])(implicit F: Monad[F]): XorT[F, AA, BB] = {
     XorT(F.flatMap(value) { xor =>
       xor match {
         case Xor.Left(_) => default.value
```
fthomas: Since we are not using `xor` after `match` I think we can just remove `xor => xor match {` ... `}`.

ceedubs: Yet another good catch. I pushed an update.

fthomas:
:+1:

PR #727, opened 2015-12-08
Update to scalastyle 0.8.0

fthomas:


ceedubs:
I suspect that the build failures are due to my recent changes to the `Arbitrary` instance for `Free`. I'm hoping they don't happen (at least as much as the currently do) after https://github.com/ceedubs/cats/commit/fbf562f3be82b0c21fbf5e48a3d1663221d548c4 which is in #724.

ceedubs:
@fthomas sorry I think #724 caused a merge conflict here.

ceedubs:
:+1:

PR #728, opened 2015-12-09
Add MonadFilter consistency to MonadFilter tests

ceedubs:
A `monadFilterConsistency` law was added in
8f7a110d536c087c8b70032103fb5a51fab5f34f but wasn't hooked up in the
tests. This might be better accomplished with something like #370, but
in the interim I think we should at least hook it up.


fthomas:
:+1:

PR #733, opened 2015-12-09
Test associativity of (Co)Kleisli composition

### LINKED ISSUE #732: Untested CoflatMap laws
There are several laws defined in `CoflatMapLaws` that aren't checked in `CoflatMapTests`. It looks like they require some additional `Eq` instances, but other than that I'm not sure whether or not there is a good reason they aren't currently included. Does anyone know?

Paging @fthomas and @non due to git history :)

fthomas:
closes #732


ceedubs:
Nice :+1:

Issue #719, opened 2015-12-07
Add some notes on testing conventions to contributors guide


mikejcurry:
As per comment https://github.com/non/cats/pull/717#issuecomment-162338024 on #717, creating an issue to add some notes on testing conventions applied to cats tests in the contributors guide. 

Hopefully I'll get time to look at this over the next few days.

mikejcurry:
Closing based on #734

PR #734, opened 2015-12-09
Add some notes on conventions in test writing

mikejcurry:
This PR looks to address #719 which was raised to add some guidance on test writing to the contributor guide related to conventions which have evolved over the course of cats development.

* Note, that there seems to be a mix of “type class” and “typeclass” used throughout the codebase/docs so I chose “type class”; happy to change if a decision has been made on that.

* This is far from perfect, but it’s a start and hopefully accurate.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -114,10 +114,27 @@ Write about https://github.com/non/cats/pull/36#issuecomment-72892359
 
 ### Write tests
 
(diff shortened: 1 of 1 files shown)


ceedubs:
Fantastic. Thanks! :+1:

stew:
this is great, thanks @mikejcurry !

PR #736, opened 2015-12-10
Test Monad map/flatMap coherence law

ceedubs:
This is basically the same story as #728, but it was introduced in
fa6457a61b527736d88a5192e9fae060db2b3ffe.


stew:
:+1:

PR #737, opened 2015-12-10
Add ScalaDoc for Arrow.split

ceedubs:


ceedubs:
I was thinking this would count as code coverage. Might have to check the build script when I'm at my computer.

fthomas:
`Arrow[Function1]` overrides the default `split` implementation and this override is already covered by the `ArrowLaws` (which use `Arrrow[Function1].split` explicitly).

ceedubs:
Ah yet another good catch, @fthomas. `Function1` seems like it's probably the most straightforward `Arrow` to show a `split` example, so I think I'll leave this as-is. We can address the lack of test coverage on the default implementation of split as a separate PR. Does that sound good?

fthomas:
Sounds good to me. :+1:

adelbertc:
:+1:

PR #738, opened 2015-12-10
Add ScalaDocs for Choice methods

ceedubs:
This is an attempt to be helpful while learning more about some methods that I've never directly used. So please feel free to chime in if the descriptions could be improved or if there are examples that might provide more insight.


ceedubs:
Ah it looks like ScalaDoc is trying to interpolate my usage of `$x`. I'll work around this.

adelbertc:
:+1:

adelbertc:
Merging because it's doc

PR #744, opened 2015-12-13
Make table in Kleisli readable

markus1189:
The table in the docs for [Kleisli](http://non.github.io/cats//tut/kleisli.html) is not readable, because it is missing markup.  I added ``` such that the table should now display properly

-- 24pullrequests


fthomas:
:+1:

PR #745, opened 2015-12-13
typeclass -> type class

markus1189:
In the documentation, most occurrences use "type class", but there are some occurrences of "typeclass" as well, this commit replaces all occurrences of "typeclass" with "type class" for a uniform naming.

-- 24pullrequests


fthomas:
We agreed on "type class" in https://github.com/non/cats/issues/441. So thanks for improving consistency! :+1:

PR #746, opened 2015-12-13
Check alternate Comonad laws

ceedubs:
cc @fthomas


fthomas:
:+1: (if Travis goes green)

I feel somehow bad about adding all those alternate formulations of laws and then never ensuring that they're correct. I hope I'm not the only one who finds those instructive...

adelbertc:
:+1:

PR #747, opened 2015-12-14
Fix monadErrorHandleWith/monadErrorHandle mix up in laws

ceedubs:
`monadErrorHandleWith` was being checked twice, but the second one was
meant to be `monadErrorHandle`.


fthomas:
:+1:

PR #748, opened 2015-12-14
Add tests of kleisli left and right identity laws

ceedubs:
cc @fthomas :)


fthomas:
:+1:

That should be the last one, right? :-)

ceedubs:
Yes, I think so! :)

PR #751, opened 2015-12-15
SemigroupK consistency and fix; added compose tests.

milessabin:
+ Renamed SemigroupK composedWith to compose for consistency with
  Applicative, Foldable, Reducible.
+ Fixed non-termination bug in SemigroupK compose(dWith).
+ Added tests for Applicative, Foldable, Reducible, SemigroupK
  composition.

### DIFF
--- a/core/src/main/scala/cats/SemigroupK.scala
+++ b/core/src/main/scala/cats/SemigroupK.scala
@@ -31,9 +31,9 @@ import simulacrum.{op, typeclass}
   /**
    * Compose two SemigroupK intsances.
    */
-  def composedWith[G[_]: SemigroupK]: SemigroupK[λ[α => F[G[α]]]] =
+  def compose[G[_]: SemigroupK]: SemigroupK[λ[α => F[G[α]]]] =
     new SemigroupK[λ[α => F[G[α]]]] {
-      def combine[A](x: F[G[A]], y: F[G[A]]): F[G[A]] = combine(x, y)
+      def combine[A](x: F[G[A]], y: F[G[A]]): F[G[A]] = self.combine(x, y)
     }
 
   /**
--- a/tests/src/test/scala/cats/tests/ComposeTests.scala
+++ b/tests/src/test/scala/cats/tests/ComposeTests.scala
@@ -1,0 +1,58 @@
+package cats
+package tests
+
+import cats.data.{ NonEmptyList, NonEmptyVector, OneAnd }
+import cats.laws.discipline.{ ApplicativeTests, FoldableTests, MonoidalTests, SemigroupKTests, arbitrary, eq }, arbitrary._, eq._
+import org.scalacheck.Arbitrary
+
+class ComposeTests extends CatsSuite {
+  // we have a lot of generated lists of lists in these tests. We have to tell
+  // Scalacheck to calm down a bit so we don't hit memory and test duration
+  // issues.
+  implicit override val generatorDrivenConfig: PropertyCheckConfiguration =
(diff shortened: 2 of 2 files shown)


core/src/main/scala/cats/SemigroupK.scala
```diff
@@ -31,9 +31,9 @@ import simulacrum.{op, typeclass}
   /**
    * Compose two SemigroupK intsances.
    */
-  def composedWith[G[_]: SemigroupK]: SemigroupK[λ[α => F[G[α]]]] =
+  def compose[G[_]: SemigroupK]: SemigroupK[λ[α => F[G[α]]]] =
```
ceedubs: We have [Alternative](https://github.com/non/cats/blob/master/core/src/main/scala/cats/Alternative.scala) which extends both `Applicative` and `MonoidK`. Not that calling it `composedWith` is a particularly elegant solution, but do we run the risk of overloading ambiguities/annoyances with both `compose`s being on the same instance?

tests/src/test/scala/cats/tests/ComposeTests.scala
```diff
@@ -0,0 +1,58 @@
...
+
+    val nelReducible =
+      new NonEmptyReducible[NonEmptyList, List] {
+        def split[A](fa: NonEmptyList[A]): (A, List[A]) = (fa.head, fa.tail)
+      }
+
+    val nevReducible =
+      new NonEmptyReducible[NonEmptyVector, Vector] {
+        def split[A](fa: NonEmptyVector[A]): (A, Vector[A]) = (fa.head, fa.tail)
+      }
```
ceedubs: I was surprised to see that these aren't already provided. I consider it outside of the scope of this PR, but I think we should follow up on this and add these instances in the `OneAnd` companion object.

milessabin: Agreed on both counts.

core/src/main/scala/cats/SemigroupK.scala
milessabin: I think the right thing to do here is to add a `compose` to `Alternative` which composes a pair of `Alternatives` to an `Alternative`. I can add that to this PR or follow up afterwards.

ceedubs: That sounds like a good solution to me. It might also mean we want to add something similar to `MonoidK`. I'm okay with that either going into this PR or a follow up one.

core/src/main/scala/cats/SemigroupK.scala
```diff
@@ -31,9 +31,9 @@ import simulacrum.{op, typeclass}
...
     new SemigroupK[λ[α => F[G[α]]]] {
-      def combine[A](x: F[G[A]], y: F[G[A]]): F[G[A]] = combine(x, y)
+      def combine[A](x: F[G[A]], y: F[G[A]]): F[G[A]] = self.combine(x, y)
```
ceedubs: Good catch!

ceedubs:
:+1:

As mentioned [here](https://github.com/non/cats/pull/751#discussion_r47636556), there are probably some additional `compose` methods that should be added, but I'm okay with those either being added here or in a separate PR.

milessabin:
I'll do a couple of follow up PRs for the issues that @ceedubs raised.

non:
I agree with @ceedubs' comments, and also his :+1: 

Thanks!

ceedubs:
Thanks, @milessabin!

PR #752, opened 2015-12-15
Remove unused Cokleisli.cokleisli method

ceedubs:
This method currently isn't used. Since `Cokleisli` is already a case
class with an equivalent auto-generated `apply` method, I don't think I
see the usefulness of it. Feel free to let me know if you would prefer
to keep this around.

### DIFF
--- a/core/src/main/scala/cats/data/Cokleisli.scala
+++ b/core/src/main/scala/cats/data/Cokleisli.scala
@@ -38,15 +38,9 @@ final case class Cokleisli[F[_], A, B](run: F[A] => B) { self =>
     Cokleisli(fca => F.extract(F.map(fca)(_._1)) -> run(F.map(fca)(_._2)))
 }
 
-object Cokleisli extends CokleisliInstances with CokleisliFunctions {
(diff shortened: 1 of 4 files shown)


stew:
:+1:

fthomas:
@ceedubs There is a similar function for `Kleisli` called `function`. Do you also want to remove that?

adelbertc:
:+1: 

@fthomas I believe the existence of `Kleisli.function` is in part due to me - due to our `ReaderT` alias, it may make readability strange to see `ReaderT.kleisli`. Though as I type this I wonder if the auto-generated `Kleisli.apply` will also be useable from the [value alias](https://github.com/non/cats/blob/master/core/src/main/scala/cats/data/package.scala#L38) we have..

fthomas:
I would be surprised if `ReaderT(f)` would not be the same as `Kleisli(f)` or `Kleisli.function(f)`.

ceedubs:
@adelbertc @fthomas I removed `Kleisli.function` as well.

adelbertc:
:+1:

PR #754, opened 2015-12-15
Remove now unnecessary tut resolver

fthomas:
see https://github.com/tpolecat/tut/issues/71 for more info

PR #755, opened 2015-12-15
remove tut resolver

tpolecat:
In theory **tut** is now part of the great wheel in the sky so it should not require a special resolver. This PR is in large part a test of that theory.


tpolecat:
Alright alright, @fthomas beat me to it. Duplicate of #754

PR #757, opened 2015-12-16
Follow up to #751

milessabin:
+ Added Reducible instance for OneAnd.
+ Added Alternative#compose and MonoidK#compose.

### DIFF
--- a/core/src/main/scala/cats/Alternative.scala
+++ b/core/src/main/scala/cats/Alternative.scala
@@ -6,10 +6,10 @@ import simulacrum.typeclass
   /**
    * Compose two Alternative intsances.
    */
-  def compose[G[_]](implicit GG: Alternative[G]): Alternative[λ[α => F[G[α]]]] =
+  override def compose[G[_]](implicit GG: Applicative[G]): Alternative[λ[α => F[G[α]]]] =
     new CompositeAlternative[F, G] {
       implicit def F: Alternative[F] = self
-      implicit def G: Alternative[G] = GG
+      implicit def G: Applicative[G] = GG
     }
 }
 
@@ -17,5 +17,5 @@ trait CompositeAlternative[F[_], G[_]]
   extends Alternative[λ[α => F[G[α]]]] with CompositeApplicative[F, G] with CompositeMonoidK[F, G] {
 
   implicit def F: Alternative[F]
-  implicit def G: Alternative[G]
+  implicit def G: Applicative[G]
 }
--- a/core/src/main/scala/cats/Alternative.scala
+++ b/core/src/main/scala/cats/Alternative.scala
@@ -2,5 +2,20 @@ package cats
 
 import simulacrum.typeclass
 
-@typeclass trait Alternative[F[_]] extends Applicative[F] with MonoidK[F]
+@typeclass trait Alternative[F[_]] extends Applicative[F] with MonoidK[F] { self =>
+  /**
+   * Compose two Alternative intsances.
+   */
+  def compose[G[_]](implicit GG: Alternative[G]): Alternative[λ[α => F[G[α]]]] =
+    new CompositeAlternative[F, G] {
+      implicit def F: Alternative[F] = self
+      implicit def G: Alternative[G] = GG
+    }
+}
 
+trait CompositeAlternative[F[_], G[_]]
+  extends Alternative[λ[α => F[G[α]]]] with CompositeApplicative[F, G] with CompositeMonoidK[F, G] {
+
+  implicit def F: Alternative[F]
+  implicit def G: Alternative[G]
+}
--- a/core/src/main/scala/cats/MonoidK.scala
+++ b/core/src/main/scala/cats/MonoidK.scala
@@ -30,6 +30,15 @@ import simulacrum.typeclass
   def empty[A]: F[A]
 
   /**
+   * Compose two MonoidK intsances.
+   */
+  def compose[G[_]](implicit GG: MonoidK[G]): MonoidK[λ[α => F[G[α]]]] =
+    new CompositeMonoidK[F, G] {
+      implicit def F: MonoidK[F] = self
+      implicit def G: MonoidK[G] = GG
+    }
+
+  /**
    * Given a type A, create a concrete Monoid[F[A]].
    */
   override def algebra[A]: Monoid[F[A]] =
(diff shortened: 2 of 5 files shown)


milessabin:
I don't know how to interpret that additional `Missed` line ... can someone help?

core/src/main/scala/cats/MonoidK.scala
```diff
@@ -38,3 +47,13 @@ import simulacrum.typeclass
       def combine(x: F[A], y: F[A]): F[A] = self.combine(x, y)
     }
 }
+
+trait CompositeMonoidK[F[_],G[_]]
+  extends MonoidK[λ[α => F[G[α]]]] {
+
+  implicit def F: MonoidK[F]
+  implicit def G: MonoidK[G]
```
ceedubs: We don't actually use this `G` at all. The same is true for the current `SemigroupK` - if you drop the `SemigroupK[G]` implicit evidence, everything still compiles.

We can freely take a `SemigroupK[F]` and turn it into a `SemigroupK[λ[α => F[G[α]]]` with no requirements on `G` (it can even be a polykinded type like `Nothing` or `Any`!). I think we should probably drop the requirements on `G`. I'm also curious if another name would make more sense, since we aren't actually composing two `MonoidK` instances together. However, we are creating a `MonoidK` instance for the composite type, so it still may be a reasonable name.

milessabin: Good point ... I'll remove the constraint.

milessabin: There is usage of the from `MonoidK[F] compose MonoidK[G]` (resp. `SemigroupK`) which depends on the constraint generating an implicit argument which which be supplied explicitly. I'll try providing two variants, one with the constraint and one without.

ceedubs: I'm not sure I understand this comment. I'm saying that I think we should be able to change `MonoidK[F] compose MonoidK[G]` to simply `MonoidK[F].compose[G]` since the `MonoidK[G]` is ignored as far as I can tell. I might be missing something.

milessabin:
I've added the unconstrained variants as `composedWith` taking only a type argument and no value argument. I'm open to alternative suggestions for the name. I think it makes sense to stick with `compose` for the constrained versions.

core/src/main/scala/cats/MonoidK.scala
milessabin: What I meant is that there is usage of the form `f compose g` for this entire family. I think we should preserve that, and also add an operator of the form `f.composedWith[G]` for the cases where the constraint isn't necessary.

core/src/main/scala/cats/Alternative.scala
```diff
@@ -2,5 +2,20 @@ package cats
 
 import simulacrum.typeclass
 
-@typeclass trait Alternative[F[_]] extends Applicative[F] with MonoidK[F]
+@typeclass trait Alternative[F[_]] extends Applicative[F] with MonoidK[F] { self =>
+  /**
+   * Compose two Alternative intsances.
+   */
+  def compose[G[_]](implicit GG: Alternative[G]): Alternative[λ[α => F[G[α]]]] =
```
ceedubs: I think the `GG` constraint can be relaxed to `Applicative` while still returning an `Alternative` (also on `CompositeAlternative` below).

core/src/main/scala/cats/MonoidK.scala
ceedubs: Hmm I guess that seems to be a bit misleading to me. You aren't really composing the instances at all. You are just grabbing a type parameter from the second one. My inclination would be to have `def compose[G[_]]: Semigroup[...]` on `SemigroupK`, `def compose[G[_]]: MonoidK[...]` on `MonoidK`, `def compose[G[_]:Applicative: Applicative[...]` on `Applicative`, and `def compose[G[_]:Applicative]: Alternative[...]` on `Alternative`. To me this seems to be the clearest about what's going on and which constraints are needed.

@non I think you originally added the `SemigroupK[G]` requirement to `compose`. Was that intentional or an oversight? What are your thoughts?

core/src/main/scala/cats/Alternative.scala
milessabin: Agreed ... updated accordingly ...

milessabin:
@ceedubs @non I'd like to move this forward if possible. I think the binary `compose` methods on `SemigroupK` and `MonoidK` are nice to have from a consistency PoV, but I'll take them out if they're blocking this.

We do need the renamed version with just an explicit type argument and no value argument though, and it should have a name other than `compose`. Do you have any suggestions? We're going from a `MonoidK[F]` to a `MonoidK[λ[t => F[G[t]]]]`, so if it isn't composition, is it some sort of lifting or wrapping operation?

milessabin:
Bump ...

non:
:+1: from me, generally speaking.

I would be fine using a name like `curryType` (or even `curry`) or any of your suggestions. It definitely seems like a useful operation.

milessabin:
@non `curryType`? Is that comment for this PR or a different one?

ceedubs:
Sorry for the delayed response. I've been on vacation.

I guess I still find am not convinced of the usefulness of the methods that take an instance but don't actually use it, and I'm concerned that they could result in unexpected behavior if someone is expecting a composition of effects from the two type constructors (and the ScalaDoc for `MonoidK.compose` currently says "Compose two MonoidK instances." which I think is misleading). If you have an example of why these methods are handy, please do share.

For the versions that take a type argument but no value, would `composeK` be a good name? It seems like these could be viewed as universal compositions in some sense.

mpilquist:
@milessabin Now that #772 is merged, can you fix conflicts?

stew:
I added PR #1 to patch up the merge

stew:
:+1:

stew:
actually cody has done a pretty good job of convincing me that although composeWith is useful, compose is not really.

ceedubs:
Just to clarify a couple points:

- I am not trying to convince anyone of anything. Just raising my concerns :)
- I completely think `compose` should exist on the subtypes of `Functor` (such as `Alternative`), where the instance is actually used. It's the other places (such as `MonoidK`), where I don't see the value.

travisbrown:
I agree that it makes sense not to include `compose` here (at least until someone can make a case for it), and I'm happy with either `composeWith` or `composeK` for the other name.

milessabin:
I have some general concerns about the way we're using `K` as a suffix (TL;DR: the `K` implies something higher kind related which isn't quite right .. I think we're after a related, but distinct, notion) but at least `composeK` would be consistent with things that we do elsewhere, so I won't object.

ceedubs:
It sounds like there is some consensus on this. Miles is currently on holiday, so I'll try to put together a PR that includes these changes along with the discussed updates.

PR #759, opened 2015-12-16
Add ScalaDocs for Option syntax

ceedubs:


milessabin:
:+1: when Travis goes green.

Issue #760, opened 2015-12-17
Move cats-free and cats-state in to cats-core?


mpilquist:
I don't recall the rationale for making the `cats.free` package its own module, but the way in which it forces `State` to be in its own module is a big enough of a disadvantage that I think we should reconsider.

This would also allow us to close https://github.com/non/cats/issues/573.

travisbrown:
:+1:. I think modularity is a good goal, but I don't think this is a useful way to cut things. Together these two jars are less than 5% of the size of cats core, and I don't think separating them makes the API more intelligible in any way.

stew:
:+1:

refried:
Hooray!

milessabin:
:+1:

dwijnand:
Can be closed now that #765 is merged.

PR #765, opened 2015-12-19
Move cats-free and cats-state in to cats-core

DavidGregory084:
To address #760


mpilquist:
:+1:

milessabin:
:+1:

mikejcurry:
Minor, but I had added the following text to the contributors guide a short time ago: *"Tests for additional modules, such as free, go into the tests directory within that module."*

If this goes through, and free moves into core, then it might make sense to change the above text to refer to the jvm module instead of referring to free, as free will no longer be a separate module and the tests have been factored to the main tests directory. I can push through a separate change at some stage if it doesn't make this one, but for consistency it might be better to change the docs and the code together.

DavidGregory084:
Thanks for pointing that out; not near a PC at the minute but I'll try and address the documentation changes this evening.

non:
Rerunning tests now. :+1: once they pass.

non:
Thanks!

PR #769, opened 2015-12-22
Improve stack consumption of `Eval.Call`

DavidGregory084:
I think that the awesome speed boost for `Eval` introduced by `Eval.Call` has broken the stack-safety of `Eval.defer` and that this is the reason that the `FoldableTestsAdditional` suite is failing now and then.
I have tried to reduce the stack consumption of `Eval.Call` by forcing thunks in such a way that the compiler can eliminate most of the stack.

### DIFF
--- a/core/src/main/scala/cats/Eval.scala
+++ b/core/src/main/scala/cats/Eval.scala
@@ -222,8 +222,22 @@ object Eval extends EvalInstances {
    * they will be automatically created when needed.
    */
   sealed abstract class Call[A](val thunk: () => Eval[A]) extends Eval[A] {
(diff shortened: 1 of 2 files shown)


non:
Thanks! (And sorry for the late comments.)

The approach seems good, and I'm glad you added a test. (I assume it was failing before?)

:+1:

DavidGregory084:
It was failing, but would only do so reliably with a small thread stack size, e.g. `-Xss256k`. Do you have any thoughts as to how we can test this without setting such an artificially small stack size?

I forgot to say but the benchmark results were not significantly worse that I could tell. :)

ceedubs:
:+1: thanks!

PR #770, opened 2015-12-22
Rewrite OptionT.flatMap in terms of flatMapF

dwijnand:
.. and fix a typo I noticed.

### DIFF
--- a/core/src/main/scala/cats/NotNull.scala
+++ b/core/src/main/scala/cats/NotNull.scala
@@ -18,7 +18,7 @@ object NotNull {
 
   private[this] def ambiguousException: Exception = new Exception("An instance of NotNull[Null] was used. This should never happen. Both ambiguous NotNull[Null] instances should always be in scope if one of them is.")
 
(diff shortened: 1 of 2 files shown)


fthomas:
:+1: 

It seems the same could be done for `orElse` and `orElseF`.

dwijnand:
@fthomas 

> It seems the same could be done for `orElse` and `orElseF`.

True, I had to check if by-name chain properly and they do, so added and updated the pull request.

fthomas:
I haven't thought of a potential issue with the by-name params, so thanks for checking that. :+1:

dwijnand:
For reference (and cross-checking), here's my proof:

```scala
Welcome to Scala version 2.11.7 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_60).
Type in expressions to have them evaluated.
Type :help for more information.

scala> def foo(b: Boolean, default: => String) = if (b) println(default)
foo: (b: Boolean, default: => String)Unit

scala> def bar(default: => String) = foo(false, default)
bar: (default: => String)Unit

scala> def mkString: String = { println("making!"); "abc" }
mkString: String

scala> bar(mkString)

scala>
```

ceedubs:
:+1: thanks!

PR #772, opened 2015-12-29
add Reducible instance for OneAnd, deprecate Foldable instance

refried:
I didn't add a test for `oneAndReducible` because I couldn't find tests for `Reducible`.  If they're here already, could you point me to them?

Because the new type suggested a name change, I left a `Foldable` instance in as `@deprecated`, non-`implicit`.  Let me know if I should do something else.


refried:
Duplicates #757 I guess?

non:
We don't have to maintain binary compatibility (yet) so let's just remove the foldable instance (since it is subsumed by the reducible one). Thanks!

ceedubs:
:+1:

As mentioned above, this overlaps with some work done in #757, but I think we want this approach (removing the `Foldable` instance). We may need to end up breaking #757 down into a couple separate PRs.

PR #774, opened 2016-01-03
Adds Show[Vector]

mikejcurry:
Adds a `Show` instance for `Vector`.

### DIFF
--- a/core/src/main/scala/cats/std/vector.scala
+++ b/core/src/main/scala/cats/std/vector.scala
@@ -2,6 +2,7 @@ package cats
 package std
 
 import cats.data.Streaming
+import cats.syntax.show._
 
 trait VectorInstances {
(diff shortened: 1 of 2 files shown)


non:
:+1:  Thanks!

core/src/main/scala/cats/std/vector.scala
```diff
@@ -43,6 +44,11 @@ trait VectorInstances {
         Streaming.fromVector(fa)
     }
 
+  implicit def vectorShow[A:Show]: Show[Vector[A]] =
+    new Show[Vector[A]] {
+      def show(fa: Vector[A]): String = fa.map(_.show).mkString("Vector(", ", ", ")")
```
ceedubs: Same question as [here](https://github.com/non/cats/pull/775#discussion_r48845132)

ceedubs:
I'm going to go ahead and merge. This is a useful instance and we can always revisit the syntax usage later.

PR #775, opened 2016-01-03
Adds Show[Stream]

mikejcurry:
Adds a `Show` instance for `Stream`.

Of note, unlike `Stream.toString`, this proposed implementation is referentially transparent.

A test exists for this property. 

I thought about adding some checks in that test to demonstrate that `toString` exhibits behaviour that violates referential transparency, but I thought that would be too brittle to included in this test, as if a future Scala version rectifies this behaviour in`Stream`, then the test might break, which might not be nice. It is straightforward to add something that shows this however.

### DIFF
--- a/core/src/main/scala/cats/std/stream.scala
+++ b/core/src/main/scala/cats/std/stream.scala
@@ -2,6 +2,7 @@ package cats
 package std
 
 import scala.collection.immutable.Stream.Empty
+import cats.syntax.show._
 
 trait StreamInstances {
   implicit val streamInstance: Traverse[Stream] with MonadCombine[Stream] with CoflatMap[Stream] =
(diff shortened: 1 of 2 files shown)


non:
Looks great to me! Thanks for ensuring referential transparency.

:+1:

core/src/main/scala/cats/std/stream.scala
```diff
@@ -58,6 +59,11 @@ trait StreamInstances {
       override def isEmpty[A](fa: Stream[A]): Boolean = fa.isEmpty
     }
 
+  implicit def streamShow[A: Show]: Show[Stream[A]] =
+    new Show[Stream[A]] {
+      def show(fa: Stream[A]): String = if(fa.isEmpty) "Stream()" else s"Stream(${fa.head.show}, ?)"
```
ceedubs: Using the `.show` syntax here will result in an extra allocation for the `ShowOps` object, right? In this particular case that probably doesn't matter, but it may be worth having a convention of not using syntax when it comes at an allocation cost in Cats. I don't know. What do people think?

ceedubs:
I'm going to go ahead and merge. This is a useful instance and we can always revisit the syntax usage later.

mikejcurry:
@ceedubs - sorry for the delay in responding. Thanks for merging!

With regard to the use of the `.show` syntax, I *think* you're right and with the current syntax it would be an extra allocation here.

I'm just wondering aloud whether it might be better to investigate what might be possible with value classes/simulacrum/machinist to reduce/remove the cost of using syntax here, instead of imposing the convention and fall back to the convention if we can't get the desired results.

It would be nice for users of cats not to have to consider this also so it would be valuable in that sense too.

PR #779, opened 2016-01-05
factor out a static inner method from #769 for better GC

refried:
I think the existing implementation will retain a reference to the outer-most `Call` by referencing its ` cats$Eval$Call$$loop$1` member.  This patch makes `loop` private/static instead.


refried:
P.S. I don't know how to test this without assuming that `System.gc()` will do a full GC.

ceedubs:
I don't claim to be an expert in these matters, but I certainly don't see any harm in factoring this out to be static.

:+1:

refried:
Bump :)

adelbertc:
:+1:

PR #780, opened 2016-01-05
Add StateT transformS

adelbertc:


milessabin:
:+1:

core/src/main/scala/cats/state/StateT.scala
```diff
@@ -73,6 +73,20 @@ final class StateT[F[_], S, A](val runF: F[S => F[(S, A)]]) extends Serializable
   def transformF[G[_], B](f: F[(S, A)] => G[(S, B)])(implicit F: FlatMap[F], G: Applicative[G]): StateT[G, S, B] =
     StateT(s => f(run(s)))
 
+  /** Transform the state used
```
ceedubs: Unjustifiably nitpicky: could you bump "Transform..." down to the next line to be consistent with the other ScalaDoc comments?

Also, what do you think about adding an sbt-doctest example here?

core/src/main/scala/cats/state/StateT.scala
```diff
@@ -74,6 +74,33 @@ final class StateT[F[_], S, A](val runF: F[S => F[(S, A)]]) extends Serializable
...
+   * scala> type GlobalEnv = (Int, String)
+   * scala> val x: StateT[Option, Int, Double] = StateT((x: Int) => Option((x + 1, x.toDouble)))
+   * scala> val xt: StateT[Option, GlobalEnv, Double] = x.transformS[GlobalEnv](_._1, (t, i) => (i, t._2))
+   * scala> val input = 5
+   * scala> x.run(input)
+   * res0: Option[(Int, Double)] = Some((6,5.0))
+   * scala> xt.run((input, "hello"))
+   * res1: Option[(GlobalEnv, Double)] = Some(((6,hello),5.0))
+   * }}}
+   */
```
ceedubs: Love it.

ceedubs:
:+1:

PR #782, opened 2016-01-07
Use Eval instead of Trampoline for State

ceedubs:
This may provide the following benefits:

1) It should be a bit more efficient.
2) It removes the need to `import cats.std.function._` for some
operations, which may make it a bit more newcomer friendly.
3) We are already using Eval for things like foldRight, so it seems more
consistent to use it for `State`.
4) I think it's a bit easier to explain `Eval` than `Trampoline`, since
it's hard to explain the latter without first explaining `Free`. So
again, this may be a bit more newcomer friendly.

However, there may be downsides I haven't thought of. I thought I would throw this PR together for discussion. Thoughts?

### DIFF
--- a/core/src/main/scala/cats/state/StateT.scala
+++ b/core/src/main/scala/cats/state/StateT.scala
@@ -1,9 +1,7 @@
 package cats
 package state
 
-import cats.free.Trampoline
 import cats.data.Kleisli
-import cats.std.function.function0Instance
 
 /**
  * `StateT[F, S, A]` is similar to `Kleisli[F, S, A]` in that it takes an `S`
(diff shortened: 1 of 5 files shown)


ceedubs:
Paging @refried @non @jdegoes @adelbertc @stew @mpilquist as people who have shown interest in State and/or Eval previously.

milessabin:
:+1:

I found working with `Eval` in Kittens a lot more straightforward than using `Trampoline`.

ceedubs:
I noticed there was an unused instance specialized to `State`. It may have been necessary before this change - I'm not sure - but scalac happily picks up the `StateT` version, so I've removed it which should fix the reported decrease in code coverage from this PR.

adelbertc:
:+1:

Issue #777, opened 2016-01-04
Move basic typeclasses into cats-kernel


rklaehn:
Cats will lose the dependency on algebra. In order for algebra and spire to depend on cats, there needs to be a small and stable core of essential typeclasses (algebra and spire will not want to depend on all of cats).

This core should contain at least

`Eq`
`Order`/`PartialOrder`
`Semigroup`/`Monoid`/`Group`
`Show`

I think it would be OK to have the core not that small, since the main concern from the algebra side is not .jar size but binary compatibility concerns. So if it is an advantage to have e.g. `SemigroupK` and `MonoidK` in there as well, that would be fine IMHO.

@non maybe just having all of algebra depend on cats-core is also acceptable. The only thing I really want to avoid is having duplicated typeclasses in cats and algebra.

rklaehn:
here is a proof of concept for a library that uses simulacrum to generate typeclass boilerplate, but does not have it in the .pom
https://github.com/rklaehn/hash

Here is how to hide simulacrum from the .pom (in case "provided" is not good enough):
https://github.com/rklaehn/hash/commit/2f11a6c3d0ce8974907cf309d4fe47863a1decc6

ceedubs:
This was resolved with #1001.

PR #786, opened 2016-01-07
Proof of concept creating a cats-kernel module for Cats/Algebra.

milessabin:
This is a follow on from https://github.com/non/cats/pull/762 taking into consideration the discussion there and on https://github.com/non/algebra/issues/142.

This new PR has essentially the same content as #762 but instead of merging the Algebra type classes in with the core module it adds them to a new kernel module on which we can impose binary compatibility constraints that are sufficient to meet the Algebra project's needs.


core/src/main/scala/cats/std/bigInt.scala
```diff
@@ -1,7 +1,14 @@
 package cats
 package std
 
-trait BigIntInstances extends algebra.std.BigIntInstances {
+trait BigIntInstances {
+  implicit val bigIntAlgebra: BigIntAlgebra =
```
ceedubs: I think this is out of the scope of this PR, but just to jot it down while I'm thinking of it: is there any benefit to having this statically typed to `BitIntAlgebra` instead of just `Order[BigInt]`?

milessabin: What are you proposing as an alternative?

core/src/main/scala/cats/std/map.scala
```diff
@@ -64,4 +66,72 @@ trait MapInstances extends algebra.std.MapInstances {
...
+}
+
+object MapMethods {
+  def initMutableMap[K, V](m: Map[K, V]): mutable.Map[K, V] = {
+    val result = mutable.Map.empty[K, V]
+    m.foreach { case (k, v) => result(k) = v }
+    result
+  }
+
+  def wrapMutableMap[K, V](m: mutable.Map[K, V]): Map[K, V] =
```
ceedubs: Are `initMutableMap` and `wrapMutableMap` methods that Cats wants to expose in its kernel API? As they return mutable values, they stick out when compared to the rest of the Cats API. Since binary compatibility is a big concern in `kernel`, I'm wondering if it would be better to make this `private` for now, as it's always easier to open them up later than it is to remove them once they are exposed in a release.

core/src/main/scala/cats/std/string.scala
```diff
@@ -1,7 +1,24 @@
...
 
-trait StringInstances extends algebra.std.StringInstances {
+trait StringInstances {
+  implicit val stringOrder: Order[String] = new StringOrder
+  implicit val stringMonoid = new StringMonoid
   implicit val stringShow: Show[String] =
     Show.fromToString[String]
 }
+
+class StringOrder extends Order[String] {
```
ceedubs: I'm wondering if we should make many of the classes in this PR, such as `StringOrder` `private[cats]`. This would follow the precedent set in #612, and I think it's especially important considering the binary compatibility constraints on the `kernel` module. If people find themselves really wanting access to these classes, then we can always open them for extension, but it's much easier to do that then it is to make changes once they have been exposed as part of the API.

kernel/src/main/scala/cats/Eq.scala
```diff
@@ -0,0 +1,106 @@
...
+import scala.{specialized => sp}
+
+import scala.math.Equiv
+
+/**
+ * A type class used to determine equality between 2 instances of the same
+ * type. Any 2 instances `x` and `y` are equal if `eqv(x, y)` is `true`.
+ * Moreover, `eqv` should form an equivalence relation.
+ */
+trait Eq[@sp A] extends Any with Serializable { self =>
```
ceedubs: If I recall, one of the goals of moving these traits from algebra to cats was to unify usage of Simulacrum, etc. Does Simulacrum work with specialization, and could we use it here? If you just wanted to get a POC out and we will follow up with this sort of thing, that's fine - I was just wondering what the plan is.

ceedubs:
Is there a near-term plan for making laws/tests consistent between kernel and the rest of cats?

core/src/main/scala/cats/std/bigInt.scala
ceedubs: I was just proposing `: Order[BigInt]` instead of `: BigIntAlgebra` for the return type. As far as I know that's the approach we are taking pretty much everywhere else.

kernel/src/main/scala/cats/Eq.scala
mpilquist: > Does Simulacrum work with specialization, and could we use it here?

Yes it does. Also, simulacrum 0.6.0-SNAPSHOT makes type classes extend `Any with Serializable`.

core/src/main/scala/cats/std/string.scala
milessabin: Just one observation on your comment: the class isn't part of the kernel, so won't be subject to quite the same strictures as the kernel types.

kernel/src/main/scala/cats/Eq.scala
milessabin: Totally. This PR is a proof of concept and the starting point for discussion about simulacrum, Typeclassic etc. Paging @mpilquist and @non ...

milessabin:
@ceedubs this is a proof of concept for discussion. I thought it would be better to keep the edit distance from Algebra as small as possible for the time being. Once it's completely clear that this is the direction we and Algebra want to go then we should deal with laws/tests/docs/simulacrum as a matter of urgency.

core/src/main/scala/cats/std/string.scala
ceedubs: Oh good point. I had in my head that all of the changes in this PR were part of the kernel module, but now I see the flaw in that reasoning :)

I still would advocate the suggested changes to be consistent with the rest of cats, and I think binary compatibility is still a worthwhile goal outside of kernel.

ceedubs:
@milessabin that makes sense, and I agree with your approach. We don't want this to turn into a never-ending PR. This looks good to me. I'd like to see what some algebra committers such as @non, @johnynek, @TomasMikula and @rklaehn think.

InTheNow:
>we should deal with laws/tests/docs/simulacrum as a matter of urgency

I've been following all of this in Algebra/Cats and on many occasions I have desperately wanted to shout out about how catalysts can do this and that... but (thankfully) kept quiet so as not to interrupt the main discussion.

As you have all done a great job on consensus, now might be a good time to raise the point that `sbt-catalysts/catalysts` needs sorting in Cats/Algebra/Spire as well... on the _build/testing side only_.

kernel/src/main/scala/cats/Eq.scala
striation: So given the binary compatibility constraints, and the fact that I think we're hoping to release cats 1.0 pretty soon now, The basic question for me is "will we get typeclassic working before we want to release cats 1.0?"

If yes, then things are easy: we can plug in simulacrum now (or not) and either way we can easily change to typeclassic without worrying too much about what exactly is produced. Ideally it wouldn't change but I wouldn't be scared of it.

If not, then I think we'll need to proceed with a bit more caution, in order to avoid reifying simulacrum's current behavior as the output that typeclassic *must* conform to.

core/src/main/scala/cats/std/anyval.scala
```diff
@@ -53,90 +52,80 @@ trait CharInstances /* missing algebra type classes */ {
...
 
   implicit val unitShow: Show[Unit] =
     Show.fromToString[Unit]
 
-  implicit val unitAlgebra: CommutativeGroup[Unit] with Order[Unit] =
-    new CommutativeGroup[Unit] with Order[Unit] {
-      def combine(x: Unit, y: Unit): Unit = ()
-      def empty: Unit = ()
-      def inverse(x: Unit): Unit = ()
+  implicit val unitOrder: Order[Unit] =
```
johnynek: the group here can be useful no? (Allows you do use some Monads that want Monoids on some values, e.g. Writer?)

core/src/main/scala/cats/std/bigInt.scala
johnynek: Also, why not the `Group[BigInt]` here? We have it for Int, Long?

core/src/main/scala/cats/std/map.scala
```diff
@@ -64,4 +66,72 @@ trait MapInstances extends algebra.std.MapInstances {
...
+}
+
+class MapMonoid[K, V](implicit V: Semigroup[V]) extends Monoid[Map[K, V]]  {
+  def empty: Map[K, V] = Map.empty
+
+  def combine(x: Map[K, V], y: Map[K, V]): Map[K, V] =
+    MapMethods.addMap(x, y)(V.combine)
+}
+
+class MapGroup[K, V](implicit V: Group[V]) extends MapMonoid[K, V] with Group[Map[K, V]] {
```
johnynek: we had a discussion in algebra about the corresponding `Eq` for this instance. Without one that considers `None == Some(V.empty)`, you might not follow the `a - a = empty` law.

core/src/main/scala/cats/std/map.scala
```diff
@@ -64,4 +66,72 @@ trait MapInstances extends algebra.std.MapInstances {
...
+  def initMutableMap[K, V](m: Map[K, V]): mutable.Map[K, V] = {
+    val result = mutable.Map.empty[K, V]
+    m.foreach { case (k, v) => result(k) = v }
+    result
+  }
+
+  def wrapMutableMap[K, V](m: mutable.Map[K, V]): Map[K, V] =
+    new WrappedMutableMap(m)
+
+  private[cats] class WrappedMutableMap[K, V](m: mutable.Map[K, V]) extends Map[K, V] {
```
johnynek: maybe add a comment that this is only safe if m is "owned" by the instance (so we know that no future modifications will happen).

core/src/main/scala/cats/std/map.scala
```diff
@@ -64,4 +66,72 @@ trait MapInstances extends algebra.std.MapInstances {
...
+    wrapMutableMap(m)
+  }
+
+  def subtractMap[K, V](x: Map[K, V], y: Map[K, V])(subtract: (V, V) => V)(negate: V => V): Map[K, V] = {
+    // even if x is smaller, we'd need to call map/foreach on y to
+    // negate all its values, so this is just as fast or faster.
+    val m = initMutableMap(x)
+    y.foreach { case (k, v2) =>
+      m(k) = m.get(k) match {
+        case Some(v1) => subtract(v1, v2)
```
johnynek: this will never remove a key, we would need `subtract: (V, V) => Option[V]` or something, so this can't return an empty map, right? So the Eq in play here needs to be correctly chosen to follow laws.

This has burned us *many* times in tests in Algebird and in production in scalding.

core/src/main/scala/cats/std/string.scala
```diff
@@ -1,7 +1,24 @@
...
+
+class StringMonoid extends Monoid[String] {
+  def empty: String = ""
+  def combine(x: String, y: String): String = x + y
+
+  override def combineAll(xs: TraversableOnce[String]): String = {
+    val sb = new StringBuilder
+    xs.foreach(sb.append)
+    sb.toString
+  }
```
johnynek: might be good to also override `combineAllOption` else it will be quadratic.

kernel/src/main/scala/cats/Eq.scala
johnynek: is there any value in making Eq contravariant? It could be (with proper modifications).

It might help get `Eq[Some[Int]]` inferred because we should use `Eq[Option[Int]]` in that case. But I can't think of others.

kernel/src/main/scala/cats/Monoid.scala
```diff
@@ -0,0 +1,52 @@
...
+  override def combineN(a: A, n: Int): A =
+    if (n < 0) throw new IllegalArgumentException("Repeated combining for monoids must have n >= 0")
+    else if (n == 0) empty
+    else repeatedCombineN(a, n)
+
+  /**
+   * Given a sequence of `as`, sum them using the monoid and return the total.
+   */
+  def combineAll(as: TraversableOnce[A]): A =
+    as.foldLeft(empty)(combine)
```
johnynek: I guess we should override `combineAllOption` to call `combineAll` so Monoid implementers don

kernel/src/main/scala/cats/Order.scala
```diff
@@ -0,0 +1,209 @@
...
+ * The truth table for compare is defined as follows:
+ *
+ * x <= y    x >= y      Int
+ * true      true        = 0     (corresponds to x == y)
+ * true      false       < 0     (corresponds to x < y)
+ * false     true        > 0     (corresponds to x > y)
+ *
+ * By the totality law, x <= y and y <= x cannot be both false.
+ */
+trait Order[@sp A] extends Any with PartialOrder[A] { self =>
```
johnynek: could be contravariant if we put min (and similar) in an implicit class. Thoughts on that?

johnynek: It is simpler to not be contravariant and use `Order[Option[Int]].on { s: Some[Int] => s }` but maybe worth a discussion about this choice.

kernel/src/main/scala/cats/PartialOrder.scala
```diff
@@ -0,0 +1,169 @@
...
+    new PartialOrder[B] {
+      def partialCompare(x: B, y: B): Double = self.partialCompare(f(x), f(y))
+    }
+
+  /**
+   * Defines a partial order on `A` where all arrows switch direction.
+   */
+  def reverse: PartialOrder[A] =
+    new PartialOrder[A] {
+      def partialCompare(x: A, y: A): Double = self.partialCompare(y, x)
```
johnynek: might be worth `override def reverse: PartialOrder[A] = self`

kernel/src/main/scala/cats/Semigroup.scala
```diff
@@ -0,0 +1,73 @@
...
+  /**
+   * Associative operation taking which combines two values.
+   */
+  def combine(x: A, y: A): A
+
+  /**
+   * Return `a` combined with itself `n` times.
+   */
+  def combineN(a: A, n: Int): A =
+    if (n <= 0) throw new IllegalArgumentException("Repeated combining for semigroups must have n > 0")
```
johnynek: should we include `: found n = $n` here? I hate to get an exception like this that doesn't tell me what the actual value was.

johnynek:
Re: typeclassic/simulacrum/machinist: can anyone confirm these have no runtime portion and are just generating code at compile time? Having a runtime portion makes the binary compat problem harder (obviously).

kernel/src/main/scala/cats/Order.scala
ceedubs: Contravariance and type classes don't play all that nicely together. [Here](http://www.scala-lang.org/old/node/4626.html) is a description of how this can be problematic.

striation:
This PR has the potential to get gigantic. Would folks be OK with creating a long-lived branch for this and then we can all issues PRs against it? That way we aren't relying on @milessabin to do all the work.

johnynek:
+1

On Thu, Jan 7, 2016 at 10:20 AM, Erik Osheim <notifications@github.com>
wrote:

> This PR has the potential to get gigantic. Would folks be OK with creating
> a long-lived branch for this and then we can all issues PRs against it?
> That way we aren't relying on @milessabin <https://github.com/milessabin>
> to do all the work.
>
> —
> <https://github.com/non/cats/pull/786#issuecomment-169762011>.
>



-- 

ceedubs:
I agree that we shouldn't shove all of the work into this PR. However, if people are confident that this is the way forward, I might propose merging and making sure we have issues for the items we want to fix before a 1.0 release. That way we don't have a branch that potentially diverges significantly from master and becomes a constant struggle to keep up to date.

striation:
The risk with merging now is that *master* will be in an unreleasable state until we straighten out the most pressing issues.

(For example, if *cats-kernel* does not export type class instances, then projects like *algebra* will have to either depend on *cats-core* or will have to duplicate those instances. IMO this is a flaw with the current PR.)

I think I'd prefer to do a branch which we will merge to master as soon as we feel like we hit a basic level of releasability. I'm also fine with merging now if other Cats devs aren't worried about this.

striation:
The reason the *cats-kernel* issue is subtle is that *cats.std* is currently implemented using a package object, which can't exist in two sub-projects. Therefore, the easy ways to resolve this would be:

1. Put types from *cats-kernel* into `cats.kernel` (and `cats.kernel.std`) and then export/alias them into `cats` in the *cats-core* sub-project.

2. Put the type class instances in a non-standard place (e.g. `cats.kernel.std`) and only realias those, leaving other types in `cats`.

3. Change how "std" is implemented so we don't need to use package objects, and can split these instances across *cats-kernel* and *cats-core*.

4. Something else I haven't thought of.

rklaehn:
On Thu, Jan 7, 2016 at 9:18 PM, Erik Osheim <notifications@github.com>
wrote:

> The risk with merging now is that *master* will be in an unreleasable
> state until we straighten out the most pressing issues.
>
> (For example, if *cats-kernel* does not export type class instances, then
> projects like *algebra* will have to either depend on *cats-core* or will
> have to duplicate those instances. IMO this is a flaw with the current PR.)
>
> I would consider duplicate typeclass *instances* much less of a problem
than duplicate typeclasses. E.g. if you have Eq[Int] in both algebra.std
and cats.std, it might not be very nice from a DRY POV, but as long as both
are law-abiding, it is not the end of the world.

The only thing that would be a problem would be

import cats.implicits._
import algebra.implicits._

// Eq[Int] does not work because of duplicate implicits

But if there is a way to prevent this, I would be in favour of leaving
instances out of cats-kernel to keep cats-kernel small. I think there are
quite a large number of instances that you can provide for e.g. Eq and
Order. Eq for all primitive types, Strings, Arrays, scala collections,
Order for all kinds of sequences, PartialOrder for sets, ...

And the number of Monoid or Group instances for primitives, arrays, scala
collections, ... is also very large.

Do we really want them all in cats-kernel? If not, where do we draw the
line?

Another thing we should not forget is to rename empty to something else
(neutral?) before doing the first release that uses cats-kernel (but not
necessarily before this PR is merged to master)

striation:
Yeah, I guess I am fine with not sharing instances, but we should be explicit that this is what is intended.

rklaehn:
Not sharing instances would definitely make things easier for now. If we
decide to share instances there should at least be some clear criterion for
which instance we share and which we don't. E.g. you could make a case that
Eq[Int] belongs into cats-kernel, but I think the various ways to define
partial orders for sorted sets definitely do not belong in cats-kernel.

How would you deal with the problem of collisions when importing all
implicits of e.g. algebra and cats. Not at all / just tell people not to do
that?

core/src/main/scala/cats/std/anyval.scala
milessabin: You mean plain `Group` rather than `CommutativeGroup`? Yes, agreed.

core/src/main/scala/cats/std/map.scala
milessabin: Can you point to that discussion?

core/src/main/scala/cats/std/map.scala
milessabin: I took this straight across from Algebra. Maybe it should stay in Algebra?

core/src/main/scala/cats/std/map.scala
milessabin: Again, I took this straight from Algebra, so if it's an issue here it's an issue there as well. @non?

milessabin:
I'm fine with this being a branch we can all contribute to ... shall I just do that?

milessabin:
Unless I'm misunderstanding, I think that quite a few of @johnynek's comments would be equally applicable to Algebra as it currently stands and don't relate to issues that I've introduced in the process of moving things over?

If that's the case then I think that would be a very strong argument in favour of merging or making this into a branch that multiple parties can iterate on ... from my PoV this PR was _just_ about the creation of cats-kernel, not also about refining its content.

non:
@milessabin Yes, I think some of @johnynek's comments do apply to algebra currently. Let's try the branch idea (with an eye to merging to master relatively soon to avoid the situation @ceedubs mentions).

ceedubs:
@striation thanks, you cleared up to me why it isn't too great of an idea to merge to master right away. :+1: for an official repo branch for this.

milessabin:
OK, I've rebased against master and pushed to https://github.com/non/cats/tree/topic/kernel.

core/src/main/scala/cats/std/anyval.scala
johnynek: Yes, sorry. I overlooked that you just removed the CommutativeGroup, but that's another issue to keep in mind with standard instances: many standard ones are commutative, but kernel won't have it, so something for algebra to keep in mind.

rklaehn:
Since there is now a separate branch to track this, I guess this PR should be closed and discussion continued somewhere else ( #777 )?

milessabin:
Agreed ... we should capture the comments made here as separate issues.

I'm completely overloaded for the next couple of days, so any help with this would be appreciated.

core/src/main/scala/cats/std/map.scala
johnynek: @milessabin https://github.com/non/algebra/pull/126#discussion_r45569352 

But yes, we can fork that off to a separate thread.

julienrf:
I think this issue can be closed now?

adelbertc:
Fixed https://github.com/typelevel/cats/pull/1001

PR #787, opened 2016-01-07
Make Scaladoc (but not Unidoc) warnings non-fatal

travisbrown:
This change addresses the errors mentioned [here](https://gitter.im/non/cats?at=568e7fc7d739f50a360243c2).

The issue is that Unidoc supports cross-module links, but Scaladoc does not. Since we're using Unidoc, it's reasonable to want to be able to have links between modules, but the publish task will still use Scaladoc to create the docs for the javadoc.jar artifacts. Any cross-module links will result in warnings, which will currently fail the build.

This change makes all warnings non-fatal for Scaladoc, but then turns fatal warnings back on for Unidoc. This allows publishing to succeed even with cross-module link warnings from Scaladoc, but the build will still fail if Unidoc runs into warnings.

An alternative solution would be to use `publishArtifact in (Compile, packageDoc) := false` to skip the javadoc.jar artifacts altogether, but that could cause other problems for repositories that expect these.

Note that there currently aren't any cross-module links in the API docs, but it's a reasonable thing to want to support even outside of the context of #786 (where the errors are first turning up).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -37,7 +37,8 @@ lazy val commonSettings = Seq(
     compilerPlugin("org.scalamacros" %% "paradise" % "2.1.0-M5" cross CrossVersion.full),
     compilerPlugin("org.spire-math" %% "kind-projector" % "0.6.3")
   ),
-  parallelExecution in Test := false
+  parallelExecution in Test := false,
+  scalacOptions in (Compile, doc) := (scalacOptions in (Compile, doc)).value.filter(_ != "-Xfatal-warnings")
 ) ++ warnUnusedImport
 
 lazy val commonJsSettings = Seq(
@@ -68,6 +69,7 @@ lazy val docSettings = Seq(
   ghpagesNoJekyll := false,
   siteMappings += file("CONTRIBUTING.md") -> "contributing.md",
   scalacOptions in (ScalaUnidoc, unidoc) ++= Seq(
+    "-Xfatal-warnings",
     "-doc-source-url", scmInfo.value.get.browseUrl + "/tree/master€{FILE_PATH}.scala",
     "-sourcepath", baseDirectory.in(LocalRootProject).value.getAbsolutePath,
     "-diagrams"


InTheNow:
@travisbrown I really cannot see a better way :+1: 

Any chance, please, that you could you also add to [sbt-catalysts](https://github.com/InTheNow/sbt-catalysts/blob/master/src/main/scala/sbtcatalysts/CatalystsPlugin.scala)?

build.sbt
```diff
@@ -37,7 +37,8 @@ lazy val commonSettings = Seq(
     compilerPlugin("org.scalamacros" %% "paradise" % "2.1.0-M5" cross CrossVersion.full),
     compilerPlugin("org.spire-math" %% "kind-projector" % "0.6.3")
   ),
-  parallelExecution in Test := false
+  parallelExecution in Test := false,
+  scalacOptions in (Compile, doc) := (scalacOptions in (Compile, doc)).value.filter(_ != "-Xfatal-warnings")
```
dwijnand:
```scala
scalacOptions in (Compile, doc) -= "-Xfatal-warnings"
```

TFW _you_ teach Travis Brown something for once ;-)

dwijnand:
1 code cleanup change, but other than that :+1: from me, I'd've gone this way as well.

Looks like TravisCI needs a kick.

build.sbt
travisbrown: @dwijnand Nice!—I had no idea that would work.

My approach to dealing with SBT is 95% cargo-cultery, and there's a similar `filter` a little further down in the file—I'll update both in this PR when I'm back at my desktop.

InTheNow: But not if this is added to sbt-catalsysts due to [typesafe activator](https://github.com/InTheNow/sbt-catalysts/blob/master/src/main/scala/sbtcatalysts/CatalystsPlugin.scala#L321-L323)

:wink:

dwijnand: @InTheNow Does activator seriously not respect the sbt version in built.properties?

ceedubs:
Fantastic. Thanks, @travisbrown!

striation:
:+1:

striation:
Whooops, was looking at an out-of-date page somehow. Great job everyone, thanks Travis!

Issue #740, opened 2015-12-11
`Monoidal` improvements


adelbertc:
The major pieces of adding `Monoidal` are in place, per https://github.com/non/cats/pull/555

[ ] As it stands, `Monoidal` is defined with just `product` (no `pure`) and as such only has the associativity law. We should perhaps have something similar to `Apply`/`Applicative` where we have a `Semigroupal` (?) with `product` and `Monoidal extends Semigroupal` with `pure`. If we go that route.. we'll have two different `pure`s that should really be the same thing - one for `Monoidal` the other for `Applicative`.
[ ] Documentation

OlivierBlanvillain:
@julienrf I've just read through #555 changes and I have the following question, what's the reason to not implement the new `product` method introduced by `Monoidal` directly on `Apply`? I feel like all implementations of this method on types `<: Apply` could be factorized by the following, to be added directly on `Apply`:

    def product[A, B](fa: F[A], fb: F[B]): F[(A, B)] =
      ap(fa)(map(fb)(b => (a => (a, b))))

julienrf:
Hi Olivier,

I just followed [this comment](https://github.com/non/cats/pull/555#discussion_r41104985) (though I didn’t run any sort of benchmark).

OlivierBlanvillain:
@julienrf I'm really sorry for the stupid question, I didn't read through the discussion and only looked at the commit... I didn't run any benchmarks but was simply surprised by the addition of new abstract methods introduced on `Apply`, I now understand it's this way for performance reasons.

Did you consider the option of splitting `Monoidal` into 3? (I did something similar [here](https://github.com/OlivierBlanvillain/validation/blob/scala-js-squashed-squashed/project/Boilerplate.scala#L90), but the motivation was to improve error reporting because `map`/`imap`/`contramap` where all overloaded `apply` to be user facing). Supposing the addition of equivalent to [`Divide`](https://github.com/scalaz/scalaz/blob/v7.1.4/core/src/main/scala/scalaz/Divide.scala) and [`Divisible`](https://github.com/scalaz/scalaz/blob/v7.1.4/core/src/main/scala/scalaz/Divisible.scala), it could end up as something like that:


| Internal                  | Functor          | Stuff with @ syntax |
| ------------------------- | ---------------- | ------------------- |
| `ApplicativeSyntax`       |  `Covariant`     | `Applicative`       |
| `DivideSyntax`            |  `Contravariant` | `Divide`            |
[...]

PR #793, opened 2016-01-09
Define Apply.product in term of ap and map

OlivierBlanvillain:
Following my comment on #740, I tried to define `Monoidal`'s `product` directly on `Apply`.


julienrf:
Hi Olivier,

Did you run some benchmarks on this?

PR #795, opened 2016-01-10
Add Semigroupal for Monoidal hierarchy, fixes #740

### LINKED ISSUE #740: `Monoidal` improvements

adelbertc:
- Make (new) Semigroupal extend Functor
- Add Monoidal and associated tests
- Applicative extends Monoidal

-----

So while I was doing this it has become even more apparent that we run into some strange-ness having both `Apply/Applicative` and `Semigroupal/Monoidal`. Currently we have `Apply extends Monoidal` (now `Semigroupal`) so we sort of give special treatment to `Apply` which doesn't particular make sense to me. Should we delete one or the other?

/cc @julienrf who did the initial work on `Monoidal` and @mpilquist who is our resident expert :-)


core/src/main/scala/cats/Semigroupal.scala
```diff
@@ -0,0 +1,16 @@
...
+import simulacrum.typeclass
+
+/**
+ * [[Semigroupal]] generalizes [[Functor]] by allowing functions of any
+ * arity to be applied to an effectful value. It encodes this by pairing
+ * together two effectful values (which can then be extended to be three
+ * or more) within a context, which can then be mapped over via the [[Functor]]
+ * instance.
+ */
+@typeclass trait Semigroupal[F[_]] extends Functor[F] {
```
julienrf: I think that’s not a good idea to restrict `Semigroupal` to be a specialization of `Functor`. For instance, you can not anymore implement [Monoidal[Show]](https://github.com/non/cats/pull/555#issuecomment-145351867).

mpilquist: Agreed, this would also prevent a `Semigroupal[Codec]`.

mpilquist:
I'm not really comfortable with the term `Semigroupal`, though I don't know the category theory well enough to know if removing the `unit` (here, `pure`) from a lax monoidal functor would yield a lax "semigroupal" functor. This is too loose of reasoning for me. I'd prefer names that were more descriptive of the operation.

I also don't think we should have type classes that are equivalent -- e.g., I don't think there's a need for both applicative and monoidal as they are represented in this PR.

My preference is that we leave the canonical type class hierarchy:

invariant ← functor ← apply ← applicative

We then create a new type class:

```scala
@typeclass trait Product[F[_]] {
}
```

We then change `Apply` to extend `Product`.

We then change the `Apply` and `Applicative` laws to be implemented in terms of `product`, `map`, and `pure`, along with adding an `ap` consistency test.

When creating an instance of `Apply` or `Applicative`, the implementor should have a choice of defining the instance in terms of `map`, `product`, and `pure`, or in terms of `ap` and `pure`. Allowing implementors this freedom is likely out of scope for this issue though, because it is better addressed when derived implementations are separated from the type class interfaces.

Fristi:
`Product` clashes with the trait `scala.Product`

adelbertc:
@mpilquist Ah that sounds good, new PR coming in shortly..

in terms of `Product` do we want to conflict with `scala.Product`? Perhaps we can call it `Zip` ? `Cartesian` ?

mpilquist:
Cartesian works for me.

ceedubs:
Sorry @adelbertc this has some merge conflicts.

adelbertc:
@mpilquist @julienrf review pls :D

core/src/main/scala/cats/syntax/monoidal.scala
```diff
@@ -1,25 +1,25 @@
...
 }
 
-abstract class MonoidalOps[F[_], A] extends Monoidal.Ops[F, A] {
-  def |@|[B](fb: F[B]): MonoidalBuilder[F]#MonoidalBuilder2[A, B] =
-    new MonoidalBuilder[F] |@| self |@| fb
+abstract class CartesianOps[F[_], A] extends Cartesian.Ops[F, A] {
+  def |@|[B](fb: F[B]): CartesianBuilder[F]#CartesianBuilder2[A, B] =
+    new CartesianBuilder[F] |@| self |@| fb
 
   def *>[B](fb: F[B])(implicit F: Functor[F]): F[B] = F.map(typeClassInstance.product(self, fb)) { case (a, b) => b }
```
julienrf: Maybe you should also rename the file to `cartesian.scala`

laws/src/main/scala/cats/laws/ApplicativeLaws.scala
```diff
@@ -31,6 +31,18 @@ trait ApplicativeLaws[F[_]] extends ApplyLaws[F] {
...
+    F.ap(fa)(f) <-> F.map(F.product(f, fa)) { case (f, a) => f(a) }
+
+  // The following are the lax monoidal functor identity laws - the associativity law is covered by
+  // Cartesian's associativity law.
+
+  def monoidalLeftIdentity[A](fa: F[A]): IsEq[F[A]] =
+    F.map(F.product(F.pure(()), fa))(_._2) <-> fa
+
+  def monoidalRightIdentity[A](fa: F[A]): IsEq[F[A]] =
+    F.map(F.product(fa, F.pure(())))(_._1) <-> fa
```
julienrf: These laws should not be expressed in terms of `map`. They should be expressed in a generic way and specialized for `Functor`, `Contravariant` and `Invariant`.

For instance, for the left identity law, instead of returning an `IsEq[F[A]]` you could return a `((F[Unit], F[A]), F[A])`, and, give a way to go from this type to `IsEq[F[A]]` in a specialized way for `Functor`, `Contravariant` and `Invariant`. I think there is something called `Isomorphisms` that can help you.

adelbertc: Hm what is the motivation for that? I mostly copied the identity laws from https://youtu.be/CtyzSZ3Z5_w?t=2464

julienrf: Having lawful `Contravariant` and `Invariant` monoidals.

julienrf:
Hi, thanks for working on this topic!

I appreciate the value of this PR, renaming `Monoidal` into `Cartesian` and adding two laws for `Apply`. I think we could push the work a bit further and do the following:

1. introduce a `Monoidal` typeclass

~~~ scala
trait Monoidal[F[_]] extends Cartesian[F] {
  def pure[A](a: A): F[A]
}
~~~

2. Define identity laws only in terms of `Monoidal`:

~~~ scala
def monoidalLeftIdentity[A](fa: F[A]): ((F[Unit], F[A]), F[A]) =
    (F.product(F.pure(()), fa)), fa)
def monoidalRightIdentity …
~~~

3. Provide ways to actually check them for `Functor`, `Contravariant` and `Invariant`, by implementing  isomorphisms between `(F[Unit], F[A])` and `F[A]` (and `(F[A], F[Unit])` and `F[A]`), as it was done for the [associativity law](https://github.com/non/cats/blob/master/laws/src/main/scala/cats/laws/discipline/MonoidalTests.scala#L33).

4. Implementing `Monoidal` for as many data types as possible (e.g. `Applicative` but also `Show`, `Const`, etc.)

If needed, we could capture `pure` in its own typeclass:

~~~ scala
trait Pure[F[_]] {
  def pure[A](a: A): F[A]
}
~~~

And then define `Monoidal` in terms of `Pure`:

~~~ scala
trait Monoidal[F[_]] extends Cartesian[F] with Pure[F]
~~~

I don’t remember if this is needed or not.

julienrf:
I’m OK for this work to be merged as it is, and to continue on working on it as per #817

adelbertc:
@mpilquist @ceedubs review? :D

laws/src/main/scala/cats/laws/discipline/CartesianTests.scala
```diff
@@ -0,0 +1,54 @@
...
+}
+
+object CartesianTests {
+  def apply[F[_] : Cartesian](implicit ev: Isomorphisms[F]): CartesianTests[F] =
+    new CartesianTests[F] { val laws: CartesianLaws[F] = CartesianLaws[F] }
+
+  trait Isomorphisms[F[_]] {
+    def associativity[A, B, C](fs: (F[(A, (B, C))], F[((A, B), C)]))(implicit EqFABC: Eq[F[(A, B, C)]]): Prop
+    def leftIdentity[A](fs: (F[(Unit, A)], F[A]))(implicit EqFA: Eq[F[A]]): Prop
+    def rightIdentity[A](fs: (F[(A, Unit)], F[A]))(implicit EqFA: Eq[F[A]]): Prop
```
julienrf: This is not very important, but I’m currently reading [this paper](http://math.ucr.edu/home/baez/rosetta.pdf) and it seems that the above isomorphisms should be named `associator`, `leftUnitor` and `rightUnitor`, to be consistent with the literature.

ceedubs:
@adelbertc sorry, I don't think I'm qualified to review this one. I'm hoping @mpilquist will weigh in. When you, @julienrf, and @mpilquist are all happy with it, then I think it's good to merge.

mpilquist:
:+1:

mpilquist:
@adelbertc Let's merge after the conflicts are fixed.

adelbertc:
Merge conflicts resolved, will merge when Travis gives the green light.

PR #798, opened 2016-01-12
Fix StreamingT.filter bug

ceedubs:
As reported by @liff [here](https://gitter.im/non/cats?at=5694ce218fdd3c0c382e20f1).

### DIFF
--- a/core/src/main/scala/cats/data/StreamingT.scala
+++ b/core/src/main/scala/cats/data/StreamingT.scala
@@ -70,7 +70,9 @@ sealed abstract class StreamingT[F[_], A] extends Product with Serializable { lh
    */
   def filter(f: A => Boolean)(implicit ev: Functor[F]): StreamingT[F, A] =
     this match {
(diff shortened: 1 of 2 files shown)


striation:
Wow! Jeez. Thanks @liff for catching this (and @ceedubs for the fix).

:+1:

fthomas:
Any idea why this wasn't caught by the `"filter with Id consistent with List.filter"` test?

ceedubs:
@fthomas I think because Scalacheck's auto-generated functions completely ignore their input and return a constant, so either the whole stream was filtered (`false`) or none of it was filtered (`true`). The effectiveness of these sorts of tests should improve once there is a new release of Scalacheck.

xuwei-k:
LGTM https://github.com/xuwei-k/cats/commit/06e542d8f60583024fe902

ceedubs:
Thanks @xuwei-k!

I'm going to go ahead and merge since this is a pretty obvious/straightforward bug.

PR #801, opened 2016-01-14
Fix order of effects in FreeApplicative.foldMap

### LINKED ISSUE #799: FreeApplicative applies effects in wrong order
Here is a fairly minimal way to reproduce the error:

```scala
import cats._
import cats.free.FreeApplicative
import cats.arrow.NaturalTransformation
import cats.implicits._

type Dsl[A] = FreeApplicative[Id, A]
val x: Dsl[String] = FreeApplicative.lift[Id, String]("x")
val y: Dsl[String] = FreeApplicative.lift[Id, String]("y")

// either one produces the wrong output, so it's not a problem with the |@| syntax
//val z: Dsl[Unit] = (x |@| y).map((_, _) => ())
val z = Apply[Dsl].map2(x, y)((_, _) => ())

val asString: Id ~> λ[α => String] = new (Id ~> λ[α => String]) {
  def apply[A](a: A): String = a.toString
}

// produces "yx" instead of "xy"
z.analyze(asString)
```

I believe this is a bug in the `ap` implementation.

Scalaz's implementation of `FreeAp` does the same thing, so it's possible that I'm wrong about what the expected result should be, but I don't think so. cc @runarorama

@adelbertc, @mandubian, or @raulraja would any of you want to take a look at this by any chance?

ceedubs:
Fixes #799

### DIFF
--- a/core/src/main/scala/cats/free/FreeApplicative.scala
+++ b/core/src/main/scala/cats/free/FreeApplicative.scala
@@ -30,7 +30,7 @@ sealed abstract class FreeApplicative[F[_], A] extends Product with Serializable
   final def foldMap[G[_]](f: F ~> G)(implicit G: Applicative[G]): G[A] =
     this match {
       case Pure(a) => G.pure(a)
-      case Ap(pivot, fn) => G.ap(f(pivot))(fn.foldMap(f))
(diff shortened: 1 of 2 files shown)


stew:
:+1:

non:
LGTM :+1:

PR #812, opened 2016-01-18
Add ApplicativeError

travisbrown:
This is a follow-up to [a question](https://gitter.im/non/cats?at=56983e475de13b3f15e3263c) I asked on Gitter a few days ago.

The basic idea is that it seems useful to be able to abstract over `Validated[E, ?]` as a context in which errors can be raised and handled, but `Validated` isn't monadic, so we can't provide a `MonadError` instance. None of the `MonadError` operations require monadic binding, though, and at least in the case of `Validated[E, ?]` there are reasonable implementations. The only `MonadError` law we lose is the left zero law.

This PR adds an `ApplicativeError` type class above `MonadError`, together with laws for `ApplicativeError` and an instance and tests for `Validated`. Current `MonadError` usage should be unaffected.

### DIFF
--- a/core/src/main/scala/cats/ApplicativeError.scala
+++ b/core/src/main/scala/cats/ApplicativeError.scala
@@ -1,0 +1,81 @@
+package cats
+
+import cats.data.{Xor, XorT}
+
+/**
+ * An applicative that also allows you to raise and or handle an error value.
+ *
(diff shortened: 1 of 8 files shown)


non:
Makes sense to me. Looks great! :+1:

adelbertc:
Awesome possum :+1:

PR #815, opened 2016-01-18
Link to actual Github Contributors

raulraja:
The following PR updates the AUTHORS file to include a link to the Github contributors list and charts which is automatically kept up to date.

@ceedubs @adelbertc @non Thanks for considering it.


refried:
I kind of like the existing version better :-\  Maybe it's too much work to maintain though?

raulraja:
For reference here is the link to the comment that triggered this PR https://gitter.im/non/cats?at=569afe4759e3d04215bc2fc4 
I'm fine if folks would rather maintain a manual list though PR authors won't update it and the list will become obsolete in no time given the fast pace in which Cats is evolving.

ceedubs:
Thanks @raulraja. To me, there is some appeal of a manually curated list, because it seems more personal (I'm guessing this is where @refried is coming from). However, I do agree that it will constantly slip out of date, which is probably worse than not having the list at all. @non do you want to weigh in on this?

If we take this approach, what do you think about changing it to something like this?

```
Thank you to [the contributors](https://github.com/non/cats/graphs/contributors) whose work has made Cats possible
```

Also: this PR currently has a merge conflict.

raulraja:
@non @ceedubs I'm happy with a manual approach as well, just added this since it seem like people liked the way Ensime handled it but I understand that the link is a bit of indirection and not as personable. I'm gonna close sit since it looks like there are mixed feelings and people are not convinced this is what is best for Cats. Feel free to reopen or do whatever. Thanks for your consideration :)

PR #818, opened 2016-01-19
Request to Add CoflatMap Instance for Vector

juanpedromoreno:
This PR brings the `CoflatMap` typeclass instance for `Vector`.

Here an example session:

```
scala> :paste
// Entering paste mode (ctrl-D to finish)

import cats._, cats.std.all._, cats.syntax.coflatMap._

def foo(vector: Vector[Int]) = vector map (_ + 1)

def bar(vector: Vector[Int]) = vector map (_ - 1)

scala> Vector(1).coflatMap(foo(_))
res0: Vector[scala.collection.immutable.Vector[Int]] = Vector(Vector(2))

scala> Vector(1).coflatMap(bar(_))
res1: Vector[scala.collection.immutable.Vector[Int]] = Vector(Vector(0))
```

@non @ceedubs @adelbertc @milessabin Thanks for your consideration.

### DIFF
--- a/core/src/main/scala/cats/std/vector.scala
+++ b/core/src/main/scala/cats/std/vector.scala
@@ -6,7 +6,7 @@ import cats.syntax.show._
 
 import scala.annotation.tailrec
 import scala.collection.+:
-import scala.collection.mutable.ArrayBuffer
+import scala.collection.mutable.ListBuffer
 
 trait VectorInstances {
   implicit val vectorInstance: Traverse[Vector] with MonadCombine[Vector] with CoflatMap[Vector] =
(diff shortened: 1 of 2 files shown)


adelbertc:
:+1:

core/src/main/scala/cats/std/vector.scala
```diff
@@ -23,6 +27,15 @@ trait VectorInstances {
       override def map2[A, B, Z](fa: Vector[A], fb: Vector[B])(f: (A, B) => Z): Vector[Z] =
         fa.flatMap(a => fb.map(b => f(a, b)))
 
+      def coflatMap[A, B](fa: Vector[A])(f: Vector[A] => B): Vector[B] = {
+        @tailrec def loop(buf: ListBuffer[B], as: Vector[A]): Vector[B] =
```
fthomas: Would it make sense to use `VectorBuilder` instead of `ListBuffer` here?

juanpedromoreno: @fthomas Yes, I think so. Done!

fthomas:
:+1:

Thanks, @juanpedromoreno!

Issue #803, opened 2016-01-14
Reverse the order of arguments to `ap`?


ceedubs:
Currently the function signature for `Apply.ap` looks like this:
```scala
  def ap[A, B](fa: F[A])(ff: F[A => B]): F[B]
```

However, there's a requirement (an unwritten one as far as I know) that the `F` effect for the second argument (`ff`) is applied before the effect of the first argument (`fa`). This dependency is relied upon for things like an implementation of `foldMap` in terms of `Const`.

This reverse order of effects [has](https://github.com/non/cats/pull/502#issuecomment-138013017) [tripped](https://github.com/non/cats/issues/799#issuecomment-171284444) [us](https://github.com/non/cats/pull/242) up a few times.

While we generally prefer to put functions as the last argument to a method, I recommend that we make an exception here and put `ff` before `fa` in the `ap` definition to help avoid this confusion. Also, since the `ff` function is wrapped in an `F` context, the usual arguments for putting the function last probably don't apply quite as strongly. I believe @mandubian also is in favor of this change (correct me if I'm wrong, Pascal).

The proposed new signature would be:
```scala
  def ap[A, B](ff: F[A => B], fa: F[A]): F[B]
```

Or we could continue to use two parameter lists, but I don't know if there's any advantage in this case.

mandubian:
I'm in favor of it as correction has already be made to be consistent with Haskell version but until we reverse params order, we can't be consistent and it leads to not very logic behavior as you've experimented yourself...

julien-truffaut:
I agree, it seems that reordering `ap` arguments will make effect order more clear.

ceedubs:
Okay we have at least 3 votes for this. Let's consider it ready for a PR.

lukewyman:
And it looks like truly low-hanging fruit, so I'll take it :)

lukewyman:
Looking for feedback

I set about making the changes as specified in this ticket with the uncurried signature of: 
def ap[A, B](ff: F[A => B], fa: F[A]): F[B].  

I ran into the type inference problems that can happen with this kind of signature, so I switched to the curried version to remedy that, as follows: 
def ap[A, B](ff: F[A => B])(fa: F[A]): F[B]

I was able to make all the necessary changes project-wide with this solution, except for the generated code in the ApplyArityFunctions, which always resulted in type inference compile errors. I coded up a little experiment to get to the bottom of this, and also to isolate the problem from the task of codegen, and found the following:

[...]

PR #821, opened 2016-01-20
Topic/apply ap reverse

lukewyman:
This PR will not pass validation, because the laws won't pass. I still submitted it so that others can look at the changes and see what the issue is. I changed the signature of Apply.ap to reverse the order of the parameter lists from:
```
def ap[A, B](fa: F[A])(ff: F[A => B]): F[B]
```
to
```
```
which broke a whole bunch of laws. So I switched the laws around, for example, from:
```
def alternativeRightAbsorption[A, B](ff: F[A => B]): IsEq[F[B]] =
    (F.empty[A] ap ff) <-> F.empty[B]
```
to
```
def alternativeRightAbsorption[A, B](ff: F[A => B]): IsEq[F[B]] =
    (ff ap F.empty[A]) <-> F.empty[B]
```
and the the law still errors with a message of:
```
cats/laws/AlternativeLaws.scala:11: Cannot prove that A => B =:= A => B.
[error]     (ff ap F.empty[A]) <-> F.empty[B]
```
It seems that A => B =:= A => B should be a provable thing. Is there anywhere else that test data "feeds" the laws that I also need to switch around to compensate for my switching around the parameter lists of Apply.ap?

### DIFF
--- a/core/src/main/scala/cats/Apply.scala
+++ b/core/src/main/scala/cats/Apply.scala
@@ -14,12 +14,12 @@ trait Apply[F[_]] extends Functor[F] with Monoidal[F] with ApplyArityFunctions[F
    * Given a value and a function in the Apply context, applies the
    * function to the value.
    */
-  def ap[A, B](fa: F[A])(ff: F[A => B]): F[B]
+  def ap[A, B](ff: F[A => B])(fa: F[A]): F[B]
 
   /**
    * ap2 is a binary version of ap, defined in terms of ap.
    */
-  def ap2[A, B, Z](fa: F[A], fb: F[B])(ff: F[(A, B) => Z]): F[Z] =
+  def ap2[A, B, Z](ff: F[(A, B) => Z])(fa: F[A], fb: F[B]): F[Z] =
     map(product(fa, product(fb, ff))) { case (a, (b, f)) => f(a, b) }
 
   /**
@@ -53,8 +53,8 @@ trait CompositeApply[F[_], G[_]]
   def F: Apply[F]
   def G: Apply[G]
 
-  def ap[A, B](fa: F[G[A]])(f: F[G[A => B]]): F[G[B]] =
-    F.ap(fa)(F.map(f)(gab => G.ap(_)(gab)))
+  def ap[A, B](f: F[G[A => B]])(fa: F[G[A]]): F[G[B]] =
+    F.ap(F.map(f)(gab => G.ap(gab)(_)))(fa)
 
   def product[A, B](fa: F[G[A]], fb: F[G[B]]): F[G[(A, B)]] =
     F.map2(fa, fb)(G.product)
--- a/core/src/main/scala/cats/FlatMap.scala
+++ b/core/src/main/scala/cats/FlatMap.scala
@@ -26,7 +26,7 @@ import simulacrum.typeclass
   def flatten[A](ffa: F[F[A]]): F[A] =
     flatMap(ffa)(fa => fa)
 
-  override def ap[A, B](fa: F[A])(ff: F[A => B]): F[B] =
+  override def ap[A, B](ff: F[A => B])(fa: F[A]): F[B] =
     flatMap(ff)(f => map(fa)(f))
 
   override def product[A, B](fa: F[A], fb: F[B]): F[(A, B)] =
--- a/core/src/main/scala/cats/data/Const.scala
+++ b/core/src/main/scala/cats/data/Const.scala
@@ -86,7 +86,7 @@ private[data] sealed abstract class ConstInstances0 extends ConstInstances1 {
     def pure[A](x: A): Const[C, A] =
       Const.empty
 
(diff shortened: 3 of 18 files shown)


lukewyman:
Forgot to add the waffle ticket - it's #803

ceedubs:
@lukewyman thanks a lot for all the work you've done on this. Sorry that I've been unresponsive - I've been traveling.

I didn't think about how changing the order of the arguments would play with Simulacrum. This does seem a bit unfortunate. Separately, I'm not sure why the compiler wouldn't be able to prove `A => B =:= A => B`.

Perhaps having the arguments line up with the order of effects isn't worth the hassle. But it also seems unfortunate if we arrive at this decision because of some syntax. @mpilquist do you have any thoughts on this?

lukewyman:
Thanks, @ceedubs. I'll continue working on this IF/when the Simulacrum issue proves resolvable.  Interesting, too, that there are a different number of errors for travis between Scala 2.10 and 2.11. Also, there are a couple of broken laws that clearly aren't equal and thus shouldn't be provable. I double checked my code on those, and I'm pretty confident that I did the switch-around right, but perhaps an issue there too.

travisbrown:
I'm pretty sure the issue here is just that the type equality constraint doesn't play as well with type inference as evidence of a subtype relationship:

```scala
scala> class Wrapper[A](a: A) {
     |   def foo[B, C](b: B)(implicit ev: A =:= (B => C)): C = ev(a)(b)
     |   def bar[B, C](b: B)(implicit ev: A <:< (B => C)): C = ev(a)(b)
     | }
defined class Wrapper

scala> val wrapped = new Wrapper((i: Int) => i.toString)
wrapped: Wrapper[Int => String] = Wrapper@3d7e2448

scala> wrapped.foo(0)
<console>:10: error: Cannot prove that Int => String =:= Int => C.
              wrapped.foo(0)
                         ^

scala> wrapped.bar(0)
res8: String = 0
```

I'm looking now at Simulacrum to see whether the constraint could be generally relaxed without causing other problems, but maybe @mpilquist could answer that more quickly?

ceedubs:
I don't think I've ever seen `ap` used directly in real-world code (outside of using it to implement `map2`, etc). For this reason, I'd be okay with us just documenting that the effect order is reversed and not trying to fight this battle. And since `ap` is often used to implement higher-level methods, if we suddenly need to start allocating extra implicit evidence instances to prove `A => B =:= A => B`, for all of our `ap` calls, that could be a further reason that this might not be worth it.

Just some thoughts.

travisbrown:
@ceedubs In my opinion a much better solution would be to require type annotations on all uses of `ap`, `ap2`, etc., which would work right now with no changes to Simulacrum. I agree that the extra evidence instances aren't ideal, but off the top of my head I don't see any reason to use `ap` (or `ap2`, etc.) to implement `map2` or other operations, which (I believe) can always be written in terms of the more primitive `map` and `product`. 

I've also just confirmed that changing [this one line](https://github.com/mpilquist/simulacrum/blob/master/core/src/main/scala/simulacrum/typeclass.scala#L202) in Simulacrum to use `<:<` instead of `=:=` and publishing locally seems to fix everything. I've rebased this PR and added some minor fixes and an update to the Simulacrum dependency (which will only be available if you publish it locally) in [this branch](https://github.com/travisbrown/cats/tree/topic/apply-ap-reverse).

I think there are three ways forward:
1. Update Simulacrum with this one line fix and merge this PR more or less as-is once the new Simulacrum (either 0.6.2 or 0.7.0) is published.
2. Don't worry about changing Simulacrum for now and just merge this PR after adding type annotations to every `ap` call (or at least the syntax ones).
3. Just don't reverse the order of arguments to the syntax `ap`.

I'd prefer the first if possible, but I don't think the second is that bad.

ceedubs:
@travisbrown wow thanks for the fast and thorough assessment. The proposed Simulacrum change sounds good to me. And even if we _do_ want to use `ap` to derive other methods, I'm realizing that we only pay the evidence allocation if we use syntax, which we probably wouldn't do. So I vote for proposal number 1 as long as @mpilquist doesn't object.

mpilquist:
:+1: for simulacrum change. Happy to merge and publish a 0.6.2 today.

travisbrown:
@mpilquist Great, thanks! Do you want the PR against master and then a backport?

travisbrown:
One question: I was just updating the docs and noticed that the name of the `ap` method on `CartesianBuilderN` isn't really consistent now, since the function comes second, although the operation is still potentially useful. Maybe it should be `apWith`? `mapF`? Other suggestions?

lukewyman:
@ceedubs, will anything further be required of me on this ticket once the simulacrum change is done?

ceedubs:
@lukewyman it looks like @travisbrown has incorporated your changes into #833, so I think we are all set here. I'm going to go ahead and close this out, but let me know if I missed anything. Thank you!

PR #822, opened 2016-01-20
Add diagram of typeclass hierarchy

rklaehn:
Not sure where the best place would be to put this. But I think it would be useful.

### DIFF
--- a/docs/src/main/tut/typeclasses.md
+++ b/docs/src/main/tut/typeclasses.md
@@ -86,6 +86,10 @@ def log[A](a: A)(implicit s: Show[A]) = println(s.show(a))
 That is that declaring the type parameter as `A : Show`, it will add
 an implicit parameter to the method signature (with a name we do not know).
 
+# Typeclass hierarchy
+
+Typeclass hierarchy for types parameterized on a type `F[_]`
+
 ![Typeclass hierarchy](http://g.gravizo.com/g?
   digraph G {
     aize ="4,4";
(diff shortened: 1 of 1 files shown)


InTheNow:
Would it also be useful to have a link to the scaladoc hierarchy under "content hierarchy in  http://non.github.io/cats/api/#cats.package ?

InTheNow:
Also see https://github.com/non/cats/issues/95, and http://plastic-idolatry.com/erik/cats-graph.png from https://github.com/non/cats/issues/95#issuecomment-114023955

ceedubs:
Nice! Thanks, @rklaehn.

@non I know you had previously created a diagram, but I don't think we ever got it into the site itself. Do you have any thoughts on this approach vs yours?

ceedubs:
:+1:

We can always put in another diagram in the future if we want, but I think this is a great start.

docs/src/main/tut/typeclasses.md
```diff
@@ -85,3 +85,52 @@ def log[A](a: A)(implicit s: Show[A]) = println(s.show(a))
...
 That is that declaring the type parameter as `A : Show`, it will add
 an implicit parameter to the method signature (with a name we do not know).
+
+# Typeclass hierarchy
+
+Typeclass hierarchy for types parameterized on a type `F[_]`
+
+![Typeclass hierarchy](http://g.gravizo.com/g?
+  digraph G {
+    aize ="4,4";
```
DavidGregory084: `size`?

stew:
here is an svg of this same diagram that is free of licensing restrictions:

PR #824, opened 2016-01-23
add more examples to foldable.md

stew:


adelbertc:
:+1:

docs/src/main/tut/foldable.md
```diff
@@ -86,5 +90,35 @@ as opposed to
...
+} catch {
+  case e:StackOverflowError => println(e)
+}
+```
+
+With the lazy `foldRight` on `Foldable`, the calculation terminates
+after looking at only one value:
+
+```tut
+Foldable[Stream].foldRight(allFalse, Eval.now(true))((a,b) => Eval.now(a && b.value)).value
```
ceedubs: I think that `(a,b) => Eval.now(a && b.value)` would need to be something like `(a,b) => if (a) b else Eval.now(false)` in order to be stack safe if there were say 10k `true` values prepending the continually `false` values. So it might be a good idea to avoid calling `.value` until the very end, since that's what you usually want to do with `Eval` to achieve stack safety.

Having said that, I just gave it a try and it looks like the Cats `foldRight` implementation for `Stream` isn't actually stack safe. I'll create a separate issue for that.

stew: I really mean this as a toy example to demonstrate the laziness. I think that it is clearer to read as is. but if you say please I'll change it :)

ceedubs: Hmm yeah, I see your point. It just seems bad to me to write an implementation that isn't stack-safe when a few lines above calls out that the std lib `foldRight` would hit a stack overflow. And if this is where people come to learn about `foldRight`, I'd want them to see examples of how to do it "right". I don't want to be a stick in the mud here though, so maybe I should just let it go :)

ceedubs:
I guess I should have just said "please" :)

Thanks, @stew! Good stuff.

PR #825, opened 2016-01-25
Fix two small errors in OptionT docs

rgcase:
Fixed a small error I noticed where Option[Future[String]] was supposed to say Future[Option[String]]. Also added a "]" where one was missing.

I'm really interested in this project so I'll probably be reading a lot of the documentation. I'll try to keep an eye out for things like this as I go.


stew:
:+1:

ceedubs:
Wonderful. Thanks, @rgcase! :+1:

Issue #826, opened 2016-01-26
Support findMap on Foldable


benhutchison:
Id like to send a PR to add a `findMap` operations over Foldable, expressed in terms of `foldRight`, that support mapping over the container to an optional value until a non-empty result is found.

`def findMap[A, B](fa: F[A], f: A => Option[B]): Option[B] =
    foldRight(fa, Eval.now(Option.empty[B]))((x, ly) => Eval.now(f(x).orElse(ly.value))).value`

Operates similarly to `fa.collectFirst(Function.unlift(f))` from the std library.

Some background context: https://groups.google.com/forum/#!topic/scala-melb/TMwPVCqh6bs

Issue #512, opened 2015-09-02
Implements foldMap in terms of foldRight


julien-truffaut:
A lazy `foldMap` permit to have a relatively efficiently `headOption`, `all`, `exist` ... e.g.

```scala
case class First[A](value: Option[A])
implicit def firstMonoidK[First] = new MonoidK[First]{
  def empty[A] = First(None)
  def combine[A](f1: First[A], f2: First[A]): First[A] =
    First(f1.value orElse f2.value)
}

trait Foldable[F[_]]{
  def headOption(fa: F[A]): Option[A] =
     foldMap(fa)(a => First(Some(a)))
}
```

julien-truffaut:
remark `headOption` and so on can still be implemented using `foldRight` directly without passing by a new type.

PR #827, opened 2016-01-26
Support findMap on Foldable issue #826

benhutchison:


ceedubs:
Thanks, @benhutchison! I like this. :+1:

Can anyone think of a reason that we might not want to include this?

benhutchison:
Actually, *I'm* having a bit of doubt... would like to get your thoughts on it:

One thing that kind of bugged me about it was, why concrete `Option` specifically rather than some typeclass? Been asking myself, what is the *essence* of Option here?

The answer Im coming to is that in `findMap` the Option makes choice between two alternatives with its `orElse` operator, ie "do I accept the current item Or keep folding". So I asked myself and the internet, whats the generalisation of that choice?

Internet reminded me that Monoids can be as much about *picking* between alternatives rather than *combining* them. And that got me thinking its actually a kind of First `MonoidK[A]` where `A: Foldable` (see also eg at http://eed3si9n.com/herding-cats/Monoid.html). Foldable defines the idea of something being `empty` or `nonEmpty`, so this monoid combine op chooses the first item if non-empty, or else the second (which continues the fold op).

To cut a long story short, I suspect its possible to implement `findMap` as an invocation of `foldMap`, but I havent been able to write it out yet. Also, because `foldMap` is currently based on eager left fold, it will fold the entire collection even if the first element is matched, which is a showstopper.

It seems that @julien-truffaut has already raised this  in #512. If #512 gets addressed, `findMap` might be a piece of documentation about using `foldMap`...?

ceedubs:
@benhutchison ah, you raise some good points. I have done this sort of thing before (with scalaz) using `foldMap` and the `First` tag. However if I recall correctly, it's a bit of a hassle, because you have to first `subst` `First` into all of the items, and then after the `foldMap` you have to `unsubst` to get rid of the tag. I could see it being handy to have a specialized version like this one if this is a common enough use-case.

Regardless of what we decide, I do agree that #512 is an issue that should be addressed.

benhutchison:
+1 on "I could see it being handy to have a specialized version like this one if this is a common enough use-case". 

From playing around today, it seems that deriving a monoid that generalizes `findMap` beyond `Option` to any `Foldable` is pretty involved, might need a `Pure` typeclass, *and* requires #512 to work properly, so.. I concluded this PR *is* worthwhile.

benhutchison:
In the end I was able to craft a `Monoid` that abstracts the search output type in `findMap` and allows it be to expressed as a invoke of `foldMap`. It seems to require:

- Alleycats, for the `EmptyK` typeclass
- Some pretty heavy type annotations
- Resolution of #512, to be viably efficient

Practically it could be out of reach for ordinary users who just want to get their problems solved, who might prefer the less general but more comprehensible `findMap`. I'll try to write it up as an article, but here's a repl trace showing the derivation:
```
Welcome to the Ammonite Repl 0.5.2
(Scala 2.11.7 Java 1.8.0_51)
@ load.ivy("org.typelevel" %% "alleycats-core" % "0.1.3")

@ import alleycats._
import alleycats._
@ import cats._
import cats._
@ import cats.std.all._
import cats.std.all._
@ import cats.data._
import cats.data._
@ import alleycats.std.OptionInstances._
import alleycats.std.OptionInstances._
@ import cats.syntax.foldable._
import cats.syntax.foldable._
@ implicit def firstNonEmpty[F[_], A](implicit F: Foldable[F], E: EmptyK[F]) = new Monoid[F[A]]{
  def empty = E.empty[A]
  def  combine(x: F[A], y: F[A]): F[A] = if (F.nonEmpty(x)) x else y
  }
defined function firstNonEmpty
@ val l = List(1, 2, 3, 4, 5)
l: List[Int] = List(1, 2, 3, 4, 5)
@ //example 1: findMap a collection using Option
@ val fOpt: Int => Option[String] = (x) => if (x < 3) Some(x.toString) else None
fOpt: Int => Option[String] = <function1>
@ l.foldMap(fOpt)(firstNonEmpty)
res10: Option[String] = Some(1)
@ l.reverse.foldMap(fOpt)(firstNonEmpty)
res11: Option[String] = Some(2)
@ //example 2: more difficult, search a collection using Xor
@ //Because Xor has more degrees of freedom, we need to specify the value returned if search fails
@ def emptyKXor[L](l: L) = new EmptyK[({type X[A] = Xor[L, A]})#X] {
  def empty[A]: Xor[L, A] = Xor.left(l)
  }
defined function emptyKXor
@ implicit val emptyXorS = emptyKXor("Not found")
emptyXorS: AnyRef with EmptyK[Xor[String, A]] = cmd12$$anon$1@7c590893
@ val fXor: Int => Xor[String, Double] = (x) => if (x < 3) Xor.left(x.toString) else Xor.right(x.toDoubl  e)
fXor: Int => Xor[String, Double] = <function1>
@ //type inference fails here
@ l.foldMap(fXor)(firstNonEmpty[({type X[A] = Xor[String, A]})#X, Double])
res15: Xor[String, Double] = Right(3.0)
@ val fXor2: Int => Xor[String, Double] = x => Xor.left("Bzzzzt")
fXor2: Int => Xor[String, Double] = <function1>
@ l.foldMap(fXor2)(firstNonEmpty[({type X[A] = Xor[String, A]})#X, Double])
res17: Xor[String, Double] = Left(Bzzzzt)
```

core/src/main/scala/cats/Foldable.scala
```diff
@@ -266,6 +266,12 @@ import simulacrum.typeclass
     !isEmpty(fa)
 
   /**
+    * Find the first element that when mapped by `f` yields a non-empty Option, returning the mapped value. Otherwise return `None`.
+    */
+  def findMap[A, B](fa: F[A])(f: A => Option[B]): Option[B] =
+    foldRight(fa, Eval.now(Option.empty[B]))((x, ly) => Eval.now(f(x).orElse(ly.value))).value
```
non: I think a better way to do this would be:

```scala
def findMap[A, B](fa: F[A])(f: A => Option[B]): Option[B] =
  foldRight(fa, Eval.now(Option.empty[B])) { (x, ly) =>
    f(x) match {
      case None => ly
      case o => Eval.now(o)
    }
  }
```

This probably means I didn't do a good job of documenting `Eval`. The basic issue here is that you generally don't want to see a call to `.value` inside an `Eval` constructor (e.g. `.now`) if you can help it, especially a recursive call like the function call in `.foldRight`. The reason is that this ends up defeating the trampolining that `Eval` can do (and is a reason that some people objected to providing the `.value` method at all).

One way to see if your implementation has this issue is to construct a very large list and then search for the last element, e.g. search `(0 to 20000).toList` for `20000`.

non:
One alternate approach here would be to return the stream of matching values. This has the advantage of being lazy and allowing the caller to choose the value(s) to keep any way they want. It also makes it clear that choosing the first value is the most efficient, but allows choosing the last value (or all the values) as well.

non:
Here's a possible implementation:

```scala
Streaming.wait(foldRight(fa, Now(Streaming.empty[B])) { (a, lbs) =>
  f(a).fold(lbs)(a => Now(Streaming.cons(a, lbs)))
})
```

core/src/main/scala/cats/Foldable.scala
benhutchison: Just checked, yes it stack overflows!

benhutchison:
Hi Erik, Im liking the alternative proposal for a `collectNonEmpty` (naming?) method that:_ lazily maps to `Option[B]` and returns a `Streaming[B]` of non-empty values_. I agree its more flexible and hits my needs. 

I tried your suggested implementation out, it worked for me. Might close this PR and submit a new one for `collectNonEmpty`?

benhutchison:
replaced by #868

Issue #823, opened 2016-01-22
Addition derived combinators for Monad, Appliicative


lukewyman:
Do we want to have any/all of the following derived combinators for cats structures such as Monad and Applicative?

```
def compose[A,B,C](f: A => F[B], g: B => F[C]): A => F[C]
```
```
def replicateM[A](n: Int, fa: F[A]): F[List[A]]
```
```
def filterM[A](as: List[A])(f: A => F[Boolean]): F[List[A]]
```

adelbertc:
`compose` I think is handled by the `Kleisli` data type

`filterM` we have via `MonadCombine`

`replicateA` would be useful to have, i'm guessing on `Applicative` (maybe `Traverse`)

```scala
def replicateA[F[_] : Applicative, A](n: Int, fa: F[A]): F[List[A]] =
  List.fill(n)(fa).sequence
```

lukewyman:
I took a shot at adding `replicateA`, as noted above, and immediately ran into the following compiler error:

```
/cats/Applicative.scala:35: could not find implicit value for evidence parameter of type cats.Traverse[List]
```
Since `sequence` is dependent on what type of collection the `Traverse` instance is for, it won't work on a concrete type like `List`

Is there another way I should try and implement this, or am I being naive in wanting to add this combinator?

ceedubs:
@lukewyman I think `import cats.std.list._` should make that compiler error go away.

lukewyman:
...and indeed, it did. Thanks, @ceedubs.

mikejcurry:
Should this one be closed now that #830 has been merged - or am I wrong and is there something left on it?

ceedubs:
It sounds like all of these are taken care of except for `filterM`, which is being tracked by #1054, so I'm going to go ahead and close this out. Feel free to reopen if I've missed something.

PR #830, opened 2016-01-29
Added replicateA to Applicative  #823

lukewyman:
Added replicateA to Applicative. Other combinators suggested already existed by other names.


adelbertc:
:+1:

lukewyman:
I don't "have the skill" for that yet. Scalacheck/property-based testing are pretty much next on my list. Should I hold off on this PR until I can get that grokked and wrapped up?

lukewyman:
@adelbertc At this point the tests aren't making a ton of sense to me. Is there a similar test on a similar method I can look at to get the idea enough to just get that done?

adelbertc:
@lukewyman The process will go something like this:

1. Add ApplicativeTests.scala to https://github.com/non/cats/tree/master/tests/src/test/scala/cats/tests
2. Add some tests checking for things like the result list has size equal to the `Int`, or anything else you may think of

An example of a test can be found https://github.com/non/cats/blob/master/tests/src/test/scala/cats/tests/FoldableTests.scala#L56

so you would have something like

```scala
class ApplicativeTests extends CatsSuite {
  test("replicateA") {
    val applicative = Applicative[Option]
    /* test here */
  }
}
```

non:
Also one very small quibble with the comment on the method. `fa` is a `F[A]` value, but is itself not an applicative functor. We could say that `fa` *has* an applicative functor. This does make writing the docs a bit more awkward.

Also, I think it's worth explaining the effect of using `sequence` to implement this. For example, another method that has the same types would be:

```scala
def alsoReplicate[A](n: Int, fa: F[A]): F[List[A]] =
  fa.map { a => List.fill(n)(a) }
```

Do you all think something like this would work?

> Given `fa` and `n`, apply `fa` to `n` times to construct an `F[List[A]]` value.

mikejcurry:
In general, given the prevalence of `List` in the ecosystem, it makes sense to have `toList` methods, but not sure about these specialisations to `List` in general. There is precedent in the various `Foldable` methods, e.g. `dropWhile_`, but this operation, and the `Foldable` operations, could equally well apply to other data structures, e.g. `Stream`, `Streaming` and I'm struggling to see what makes `List` important enough that it stands out. 

I don't want to derail this PR, which seems fine, but just wondering is there some principle about when such specialisations to something like `List` should be included and when not? Or shall things just remain organic for the moment?

@non that proposed comment update seems fine to me. Difficult operation to capture the implications of succinctly, but a couple of examples would also go a long way to clarifying the effect and add a lot of value, particularly the implications if `F` is also a `List` like data structure, which may not be entirely intuitive.
```scala
scala> Applicative[List].replicateA(2, List(1, 2))
res8: List[List[Int]] = List(List(1, 1), List(1, 2), List(2, 1), List(2, 2))

scala> Applicative[Option].replicateA(2, Option(2))
res9: Option[List[Int]] = Some(List(2, 2))

**Comment by lukewyman** [comment]
@mikejcurry I'm still very much a newbie to Scala and FP in general, so this was a learning experience for me. If `List` is a too specific considering the more general nature of `Traverse` that the rest of the combinators in `Applicative`use, I'm happy to consider another approach or abandon this PR. I basically came up with the idea of identifying Monadic and Applicative combinators in "FP in Scala", and comparing that to what cats has/doesn't have, and what might be named differently.

**Comment by mikejcurry** [comment]
Don't take my comment the wrong way. I'm pretty new myself. It was just a question. :-). Same as you, just learning. :-)

**Comment by lukewyman** [comment]
@adelbertc I've added a unit test for #823. Good to go?

File: core/src/main/scala/cats/Applicative.scala [hunk 1]
```diff
@@ -29,6 +30,12 @@ import simulacrum.typeclass
   def pureEval[A](x: Eval[A]): F[A] = pure(x.value)
 
   /**
+    * Given fa and n, apply fa n times to construct an F[List[A]] value.
```
ceedubs: This might render a little cleaner in the ScalaDoc if you use backticks around `fa`, `n`, and `F[List[A]]`.

ceedubs:
:+1: thanks!

adelbertc:
Looks like Travis 2.10 failed, rebuilding. :+1: on green.

ceedubs:
The build had failed again with a mysterious error. I restarted it again, but I just realized that there are now merge conflicts :(. Sorry @lukewyman would you mind updating your branch?

lukewyman:
Hey @ceedubs, I've updated the PR so that it merges. All green :)

ceedubs:
:+1: thanks!

lukewyman:
Hi @mikejcurry, I believe you can go ahead and close this one.

PR #832, opened 2016-01-30
Update free monad docs

### LINKED ISSUE #804: Update free monad tut for newer Free structure
The Cats `Free` monad has `Coyoneda` "built-in". This means that when using a custom ADT with `Free` and an interpreter, you don't need to create a `Functor` for your data type. This means that the ADT in our [doc](http://non.github.io/cats//tut/freemonad.html) could be simplified, which would probably be better to do for an introductory doc.

ceedubs:
This resolves #804.

I also updated the "pure" compiler to use `State`. It previously had
said that a pure compiler couldn't be written and used with `foldMap`,
but that's not really true if you use something like `State`. I think
this is probably a cleaner and more useful solution.

@mandubian @raulraja would you be available to review?

### DIFF
--- a/docs/src/main/tut/freemonad.md
+++ b/docs/src/main/tut/freemonad.md
@@ -77,30 +77,21 @@ recursive values.
 We need to create an ADT to represent our key-value operations:
 
 ```tut:silent
-sealed trait KVStoreA[+Next]
-case class Put[T, Next](key: String, value: T, next: Next) extends KVStoreA[Next]
-case class Get[T, Next](key: String, onResult: T => Next) extends KVStoreA[Next]
-case class Delete[Next](key: String, next: Next) extends KVStoreA[Next]
+sealed trait KVStoreA[A]
+case class Put[T](key: String, value: T) extends KVStoreA[Unit]
+case class Get[T](key: String) extends KVStoreA[Option[T]]
+case class Delete(key: String) extends KVStoreA[Unit]
 ```
-
-The `next` field in each of the types provides a way to link an
-operation with successive values. The `Next` type parameter can be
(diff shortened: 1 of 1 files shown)


docs/src/main/tut/freemonad.md
```diff
@@ -77,30 +77,21 @@ recursive values.
...
 
 There are six basic steps to "freeing" the ADT:
 
 1. Create a type based on `Free[_]` and `KVStoreA[_]`.
-2. Prove `KVStoreA[_]` has a functor.
-3. Create smart constructors for `KVStore[_]` using `liftF`.
-4. Build a program out of key-value DSL operations.
-5. Build a compiler for programs of DSL operations.
-6. Execute our compiled program.
+2. Create smart constructors for `KVStore[_]` using `liftF`.
```
raulraja: An alternative to `liftF` is to use `Free.inject` when targeting `cats.data.Coproduct` as a composition mechanism maybe it should be mentioned here.

raulraja: I guess it's mentioned below. Perhaps we should consider adding some kind of TOC at the beginning of each tutorial to show at a glance the different sections and topics being described.

raulraja:
@ceedubs LGTM!. My comments are just suggestion for future consideration in case the tutorials become too long.

non:
:+1:

mandubian:
yes that was required! thanks for the effort ;)
:+1:

ceedubs:
Thanks for the quick feedback everyone. @raulraja I agree that tables of contents might be helpful in docs. I'll create a separate issue to track that.

PR #833, opened 2016-01-30
Reverse order of arguments for ap

travisbrown:
This is a follow-up to #821 that rebases @lukewyman's changes, makes some fixes supported by the new version of Simulacrum, and updates the docs.

Note that I've renamed `ap` on `CartesianBuilderN` to `apWith` to avoid confusion (since the function comes after the arguments in that case).


non:
Seems reasonable to me. :+1:

ceedubs:
Looks good. :+1: as long as the build passes.

Issue #785, opened 2016-01-07
Move Free/State to data package?


ceedubs:
Currently there are separate (but small) packages for `free` and `state`. This is because up until #760 they were in their own separate modules. Now that they are within `core`, should we drop the `free`/`state` from their package name and move them to the `data` package to be more consistent with other types like `WriterT`?

I'm not sure `data` is a particularly meaningful package name for some of these types, but that's probably a tangential issue (if it's an "issue" at all).

dwijnand:
:+1: from me, for either `cats.data` or `cats`.

travisbrown:
:+1:, and I'm fine with having them all directly under `cats.data`.

ceedubs:
Update: Since #990 was merged, this no longer applies to `free`.

ceedubs:
Oops I just realized that `State` has already been moved to the `data` package. Looks like it's time to close this out.

PR #835, opened 2016-01-30
Move State to data package

ceedubs:
This is a partial partial fix for #785. So far I've moved `state` but not `free`. There are several different types within the `free` package, so it might warrant its own package. However, `State`/`StateT` was just sitting in the `state` package by itself, so I've gone ahead and moved it (so it now parallels `Writer/WriterT`).


adelbertc:
:+1: once we have green light

fthomas:
:+1:

PR #837, opened 2016-01-30
Update URLs in readme after org change

ceedubs:
@non can you update the site URL at the top of the repo? It won't let me edit it, so I'm guessing you need to be owner.


fthomas:
:+1:

PR #844, opened 2016-01-31
Adds a MonadTrans typeclass

stew:
This is similar to MonadTrans in scalaz. It allows you to lift a Monad into a transformer wrapping that monad

I skipped Hoist, since I've never seen that actually used in the wild.

I was shamed into doing this by @djspiewak :)

### DIFF
--- a/core/src/main/scala/cats/MonadTrans.scala
+++ b/core/src/main/scala/cats/MonadTrans.scala
@@ -1,0 +1,26 @@
+package cats
+
+/**
+ * A typeclass which abstracts over monad transformers, providing the
+ * ability to lift a monad, into the the monad transformer for another
+ * monad.
+ */
+trait MonadTrans[MT[_[_], _]] {
+  /**
+   * Lift a value of type M[A] into a monad transformer MT[M, A]
+   */
+  def liftM[M[_]: Monad, A](ma: M[A]): MT[M, A]
+
+  /**
+   * Lift a value of type M[A] into a monad transformer MT[M, A] using
+   * Unapply, this will be useful when the M[A] monad is actually not
+   * in the * -> * shape. For example Xor[E,A].
+   */
+  def liftMU[MA](ma: MA)(implicit U: Unapply[Monad, MA]): MT[U.M, U.A] = {
+    liftM[U.M, U.A](U.subst(ma))(U.TC)
+  }
+}
+
+object MonadTrans {
+  def apply[MT[_[_], _]](implicit MT: MonadTrans[MT]) = MT
+}
--- a/core/src/main/scala/cats/data/OptionT.scala
+++ b/core/src/main/scala/cats/data/OptionT.scala
@@ -136,6 +136,11 @@ private[data] sealed trait OptionTInstances1 {
       override def map[A, B](fa: OptionT[F, A])(f: A => B): OptionT[F, B] =
         fa.map(f)
     }
+
+  implicit val optionTMonadTrans: MonadTrans[OptionT] = new MonadTrans[OptionT] {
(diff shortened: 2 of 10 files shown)


core/src/main/scala/cats/MonadTrans.scala
```diff
@@ -0,0 +1,26 @@
...
+   * Unapply, this will be useful when the M[A] monad is actually not
+   * in the * -> * shape. For example Xor[E,A].
+   */
+  def liftMU[MA](ma: MA)(implicit U: Unapply[Monad, MA]): MT[U.M, U.A] = {
+    liftM[U.M, U.A](U.subst(ma))(U.TC)
+  }
+}
+
+object MonadTrans {
+  def apply[MT[_[_], _]](implicit MT: MonadTrans[MT]) = MT
```
ceedubs: Missing return type

core/src/main/scala/cats/syntax/monad.scala
```diff
@@ -0,0 +1,18 @@
+package cats
+package syntax
+
+trait MonadSyntax1 {
```
ceedubs: Should this be private?

ceedubs:
Looks good.  I just left a couple very minor comments.

I think we could also have instances for Kleisli and StateT, though that could be a separate PR.

core/src/main/scala/cats/data/OptionT.scala
```diff
@@ -136,6 +136,11 @@ private[data] sealed trait OptionTInstances1 {
       override def map[A, B](fa: OptionT[F, A])(f: A => B): OptionT[F, B] =
         fa.map(f)
     }
+
+  implicit val optionTMonadTrans: MonadTrans[OptionT] = new MonadTrans[OptionT] {
+    def liftM[M[_], A](ma: M[A])(implicit M: Monad[M]): OptionT[M, A] =
+      OptionT(M.map(ma)(Some.apply))
```
ceedubs: I think this could just delegate to liftF

ceedubs:
I think that `MonadTrans` is really helpful. However, there is one thing that bothers me a little bit about it. It seems overly-specific to `Monad`. Most of these transformers actually have constraints that are weaker than `Monad` in order to lift into them. For example, we can lift any `F[A]` into a `Kleisli` without constraints on `F`. We can lift any `F[A]` where `F` has a `Functor` into an `OptionT`.

I'm pondering whether it might make sense to have something like `Transformer[F[_], T[_[_], _]]` that also abstracts over what the type class needed by the constructor is. I might be trying to get a little too creative, and it might run into all sorts of inference issues. But I'd kind of like to give it a try. Any thoughts?

ceedubs:
Hmm I'm confused by this. Your branch seems to not have the latest from master, so `Cartesian` is still called `Monoidal`. Yet you've added `CartesianSyntax` and somehow the build passed?

travisbrown:
@ceedubs Travis CI will use GitHub's merge commit, which in this case is ece069f588077426bfc41f380ec341b38eaf6874, which has 7c9c20975cd9a2bc0c18926fbd5e519ab8e7931e for a parent and includes the `Cartesian` change. (I thought there was some way to have the merge commit automatically update for new commits on the target branch, but I might have been making that up.)

ceedubs:
Thanks @travisbrown. It works fine if I merge in the latest from master.

I tried out my idea and I think it looks like it could work. I'll probably submit a PR (maybe to Stew's branch) tomorrow for discussion.

refried:
thanks @stew, this helped me finally understand MonadTrans.

stew:
OMG @ceedubs YES, I have it. I'm closing this PR in favor of #853 which generalizes this to not always need monads to lift

Issue #802, opened 2016-01-14
InvariantMonoidal + FreeInvariant?


OlivierBlanvillain:
Would you be interested in a new type class:

``` scala
@typeclass trait InvariantMonoidal[F[_]] extends Invariant[F] with Monoidal[F]
```

And a new Free structure for it? Something along the lines of the stuff we built with 
@Fristi, https://gist.github.com/OlivierBlanvillain/b6322caec0318734c12b

I think this is what would be needed in a PR, let me know if I forgot something:
- cats.InvariantMonoidal
- cats.laws.InvariantMonoidalLaws
- cats.free.FreeInvariant
- cats.tests.FreeInvariantTests
- tut.freeinvariant.md

With s/Monoidal/Cartesian/ after #795 :)

Fristi:
:+1: 

Very useful stuff, for me atleast. Though I wonder what other Free variants there are. You can combine several typeclasses and lift their operations in a Free variant. But what combinations would work and be useful? Is there any work done on that?

OlivierBlanvillain:
@mpilquist What's your opinion on that?

All credits goes to you for coming with the name in your [SBTB talk](http://www.youtube.com/watch?v=CtyzSZ3Z5_w&t=51m15s) :)

mpilquist:
+1 This looks good to me!

ceedubs:
Sorry for not responding for so long but I started pondering a project today that I think could really benefit from this. Now that free has its own module again, I'm less concerned about adding more structures to it. :+1: from me.

ceedubs:
Uh ignore my last comment. I had forgotten there is an open PR for this.

peterneyens:
Added in #845.

Issue #817, opened 2016-01-19
Introduce `Monoidal`


julienrf:
#795 renamed `Monoidal` into `Cartesian` and introduced new laws for `Applicative`. To push further this work and have a proper support of `Monoidal` invariant and contravariant structures, I think we should follow these steps:
- Introduce a `Monoidal` typeclass

``` scala
trait Monoidal[F[_]] extends Cartesian[F] {
  def unit: F[Unit]
}
```
- Move the identity laws from `ApplicativeLaws` to `MonoidalLaws`
- Define `Applicative` (resp. `ApplicativeLaws`) in terms of `Monoidal` (resp. `MonoidalLaws`)
- Implement `Monoidal` for as many data types as possible (e.g. `Applicative` but also `Show`, `Const`, etc.)

mpilquist:
:+1: for adding a type class that is `Cartesian` with a `pure`. Does anyone have any examples on where we would abstract over a type constructor using `Monoidal`? I.e., is the primary motivation of providing it free theorems or is there some other utility?

I'm not sold on the name `Monoidal`. I mentioned this in the other PR -- feels like too loose of reasoning, though I'm happy to be wrong here. :smile:

julienrf:
@mpilquist Yes, I’ve already written something like the following:

~~~ scala
trait Something[F[_], A] {
  …
}
object Something {
    implicit def monoidal[F[A]](implicit F: Monoidal[F]): Monoidal[Something[F, ?]] = …
}
~~~

Basically, my `Something` typeclass was essentially a wrapper around an `F[_]` type, and I derived a `Monoidal[Something[F, ?]]` from an existing `Monoidal[F]`.

julienrf:
I also use it here: https://github.com/julienrf/play-products/blob/master/src/main/scala/julienrf/products/Builder.scala#L11 (Play’s `FunctionalCanBuild` is our `Cartesian`).

julienrf:
@mpilquist I don’t remember if we already debate of this, but what would be your opinion on having a member `unit: F[Unit]` instead of `pure[A](a: A): F[A]`? `unit` is sufficient to check the laws and maybe less constraining than `pure` (ie. maybe for some type `F[?]` it is not possible to implement `pure` but it is possible to implement `unit`).

mpilquist:
@julienrf I'm neutral on `unit` vs `pure`. I can see benefits to either approach. I initially was leaning towards `pure` but the more I think about it, I think `unit` might be better.

[...]

PR #845, opened 2016-02-01
Add InvariantMonoidal and FreeInvariantMonoidal

OlivierBlanvillain:
This PR adds `InvariantMonoidal` and `FreeInvariantMonoidal`, the invariant analogous to `Applicative` and `FreeApplicative`. Following this, we could consider adding a `Monoidal` type class and explore the `Contravariant` side of things.

--

Edit: I'm done with this one, it's good for review (ping @Fristi & @mpilquist)!

I realized that for every instance of `InvariantCartesian` I had in mind its also possible to define a `pure`, so I reworked everything to define an `InvariantMonoidal` and `FreeInvariantMonoidal` instead of their `Cartesian` counterpart.

--

This is work in progress on #802.

Before I finish it (doc is surprisingly long to write :smile:), I have concerns regarding the legitimacy of this new `InvariantCartesian` type class. I could not find new laws for it that cannot be proved by combining laws of `Invariant` and `Cartesian`, which makes me think that `InvariantCartesian` does nothing more than pairing an `Invariant` with a `Cartesian`. As such, is it interesting at all? I'm pretty sure that the `FreeInvariant` stuff (the original motivation for this PR), could be formulated directly in term of `Invariant` and `Cartesian`.

Are there other example of type classes already in cats that add nothing more than "hey, I'm this one plus this one, that it!"? If we look at `Apply` as an example, is there a thing which has a `Functor` and a `Cartesian`, but not an `Apply`? If the answer is no, then why do we need `Apply` at all, given that all code using it could be rewritten in term of `Functor` and `Cartesian`?

I'm a bit confused, not really sure if my questions/concerns make sense here...

### DIFF
--- a/core/src/main/scala/cats/Cartesian.scala
+++ b/core/src/main/scala/cats/Cartesian.scala
@@ -16,4 +16,13 @@ import simulacrum.typeclass
   def product[A, B](fa: F[A], fb: F[B]): F[(A, B)]
 }
 
-object Cartesian extends CartesianArityFunctions
+object Cartesian extends CartesianArityFunctions with AlgebraCartesianInstances
+
+/**
+ * Cartesian instances for types that are housed in Algebra and therefore
+ * can't have instances for Cats type classes in their companion objects.
+ */
+private[cats] sealed trait AlgebraCartesianInstances {
+  implicit val invariantSemigroup: Cartesian[Semigroup] = InvariantMonoidal.invariantMonoidalSemigroup
+  implicit val invariantMonoid: Cartesian[Monoid] = InvariantMonoidal.invariantMonoidalMonoid
+}
--- a/core/src/main/scala/cats/Cartesian.scala
+++ b/core/src/main/scala/cats/Cartesian.scala
@@ -16,13 +16,13 @@ import simulacrum.typeclass
   def product[A, B](fa: F[A], fb: F[B]): F[(A, B)]
 }
 
-object Cartesian extends CartesianArityFunctions with AlgebraCartesianInstances
+object Cartesian extends CartesianArityFunctions with KernelCartesianInstances
 
 /**
- * Cartesian instances for types that are housed in Algebra and therefore
+ * Cartesian instances for types that are housed in Kernel and therefore
  * can't have instances for Cats type classes in their companion objects.
  */
-private[cats] sealed trait AlgebraCartesianInstances {
+private[cats] sealed trait KernelCartesianInstances {
   implicit val catsInvariantSemigroup: Cartesian[Semigroup] = InvariantMonoidal.catsInvariantMonoidalSemigroup
   implicit val catsInvariantMonoid: Cartesian[Monoid] = InvariantMonoidal.catsInvariantMonoidalMonoid
 }
--- a/core/src/main/scala/cats/Cartesian.scala
+++ b/core/src/main/scala/cats/Cartesian.scala
@@ -23,6 +23,6 @@ object Cartesian extends CartesianArityFunctions with AlgebraCartesianInstances
  * can't have instances for Cats type classes in their companion objects.
  */
 private[cats] sealed trait AlgebraCartesianInstances {
-  implicit val invariantSemigroup: Cartesian[Semigroup] = InvariantMonoidal.invariantMonoidalSemigroup
-  implicit val invariantMonoid: Cartesian[Monoid] = InvariantMonoidal.invariantMonoidalMonoid
+  implicit val catsInvariantSemigroup: Cartesian[Semigroup] = InvariantMonoidal.catsInvariantMonoidalSemigroup
+  implicit val catsInvariantMonoid: Cartesian[Monoid] = InvariantMonoidal.catsInvariantMonoidalMonoid
 }
--- a/core/src/main/scala/cats/InvariantMonoidal.scala
+++ b/core/src/main/scala/cats/InvariantMonoidal.scala
@@ -1,0 +1,52 @@
+package cats
+
+import cats.functor.Invariant
+import simulacrum.typeclass
+
+/**
+ * Invariant version of a Monoidal.
+ *
+ * Must obey the laws defined in cats.laws.InvariantMonoidalLaws.
+ */
+@typeclass trait InvariantMonoidal[F[_]] extends Invariant[F] with Cartesian[F] {
+  def pure[A](a: A): F[A]
+}
+
+object InvariantMonoidal extends AlgebraInvariantMonoidalInstances
+
+/**
+ * InvariantMonoidal instances for types that are housed in Algebra and therefore
+ * can't have instances for Cats type classes in their companion objects.
+ */
+private[cats] trait AlgebraInvariantMonoidalInstances {
+  implicit val invariantMonoidalSemigroup: InvariantMonoidal[Semigroup] = new InvariantMonoidal[Semigroup] {
+    def product[A, B](fa: Semigroup[A], fb: Semigroup[B]): Semigroup[(A, B)] = new Semigroup[(A, B)] {
+      def combine(x: (A, B), y: (A, B)): (A, B) = fa.combine(x._1, y._1) -> fb.combine(x._2, y._2)
+    }
+
+    def imap[A, B](fa: Semigroup[A])(f: A => B)(g: B => A): Semigroup[B] = new Semigroup[B] {
+      def combine(x: B, y: B): B = f(fa.combine(g(x), g(y)))
+    }
+
+    def pure[A](a: A): Semigroup[A] = new Semigroup[A] {
+      def combine(x: A, y: A): A = a
+    }
+  }
+
+  implicit val invariantMonoidalMonoid: InvariantMonoidal[Monoid] = new InvariantMonoidal[Monoid] {
+    def product[A, B](fa: Monoid[A], fb: Monoid[B]): Monoid[(A, B)] = new Monoid[(A, B)] {
+      val empty = fa.empty -> fb.empty
+      def combine(x: (A, B), y: (A, B)): (A, B) = fa.combine(x._1, y._1) -> fb.combine(x._2, y._2)
+    }
+
+    def imap[A, B](fa: Monoid[A])(f: A => B)(g: B => A): Monoid[B] = new Monoid[B] {
+      val empty = f(fa.empty)
+      def combine(x: B, y: B): B = f(fa.combine(g(x), g(y)))
+    }
+
+    def pure[A](a: A): Monoid[A] = new Monoid[A] {
+      val empty = a
+      def combine(x: A, y: A): A = a
+    }
+  }
+}
--- a/core/src/main/scala/cats/InvariantMonoidal.scala
+++ b/core/src/main/scala/cats/InvariantMonoidal.scala
@@ -12,14 +12,14 @@ import simulacrum.typeclass
   def pure[A](a: A): F[A]
 }
 
-object InvariantMonoidal extends AlgebraInvariantMonoidalInstances
+object InvariantMonoidal extends KernelInvariantMonoidalInstances
 
 /**
- * InvariantMonoidal instances for types that are housed in Algebra and therefore
- * can't have instances for Cats type classes in their companion objects.
+ * InvariantMonoidal instances for types that are housed in cats.kernel and therefore
+ * can't have instances for this type class in their companion objects.
  */
-private[cats] trait AlgebraInvariantMonoidalInstances {
-  implicit val invariantMonoidalSemigroup: InvariantMonoidal[Semigroup] = new InvariantMonoidal[Semigroup] {
+private[cats] trait KernelInvariantMonoidalInstances {
+  implicit val catsInvariantMonoidalSemigroup: InvariantMonoidal[Semigroup] = new InvariantMonoidal[Semigroup] {
     def product[A, B](fa: Semigroup[A], fb: Semigroup[B]): Semigroup[(A, B)] = new Semigroup[(A, B)] {
       def combine(x: (A, B), y: (A, B)): (A, B) = fa.combine(x._1, y._1) -> fb.combine(x._2, y._2)
     }
@@ -33,7 +33,7 @@ private[cats] trait AlgebraInvariantMonoidalInstances {
     }
   }
 
-  implicit val invariantMonoidalMonoid: InvariantMonoidal[Monoid] = new InvariantMonoidal[Monoid] {
+  implicit val catsInvariantMonoidalMonoid: InvariantMonoidal[Monoid] = new InvariantMonoidal[Monoid] {
     def product[A, B](fa: Monoid[A], fb: Monoid[B]): Monoid[(A, B)] = new Monoid[(A, B)] {
       val empty = fa.empty -> fb.empty
       def combine(x: (A, B), y: (A, B)): (A, B) = fa.combine(x._1, y._1) -> fb.combine(x._2, y._2)
--- a/core/src/main/scala/cats/data/Const.scala
+++ b/core/src/main/scala/cats/data/Const.scala
@@ -107,6 +107,17 @@ private[data] sealed abstract class ConstInstances0 extends ConstInstances1 {
 }
 
 private[data] sealed abstract class ConstInstances1 {
+  implicit def constInvariantMonoidal[C: Monoid]: InvariantMonoidal[Const[C, ?]] = new InvariantMonoidal[Const[C, ?]] {
+    def pure[A](a: A): Const[C, A] =
+      Const.empty
+
+    def imap[A, B](fa: Const[C, A])(f: A => B)(g: B => A): Const[C, B] =
+      fa.retag[B]
+
+    def product[A, B](fa: Const[C, A],fb: Const[C, B]): Const[C, (A, B)] =
+      fa.retag[(A, B)] combine fb.retag[(A, B)]
+  }
+
   implicit def catsDataEqForConst[A: Eq, B]: Eq[Const[A, B]] = new Eq[Const[A, B]] {
     def eqv(x: Const[A, B], y: Const[A, B]): Boolean =
       x === y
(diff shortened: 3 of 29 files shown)


julienrf:
Interesting. I’m not sure `FreeInvariant` is a good name, though. What you actually have is a “free invariant cartesian functor”.

Also, are we going to have a `FreeXxx` for each combination of `(Functor + Invariant + Contravariant) × (Monad + Cartesian + Monoidal)`?

OlivierBlanvillain:
> Also, are we going to have a FreeXxx for each combination of (Functor + Invariant + Contravariant) × (Monad + Cartesian + Monoidal)?

@julienrf I think we are almost there. Unfolding your ×, I think these are the practical cases:

- `FreeFunctorMonad` → That's `cats.free.Free`.
- `FreeFunctorMonoidal` → That's `cats.free.FreeApplicative`.
- `FreeContravariantMonoidal` → Probably interesting for pretty printers, to be investigated. (Would this be `FreeDivisible`?)
- `FreeInvariantMonoidal` → That's the `cats.free.InvariantMonoidal` I would like to contribute :)

I this there would be no use cases of the other ones:

- `FreeContravariantMonad` → I don't think a `ContravariantMonad` would make sense.
- `FreeInvariantMonad` → Same.
- `FreeFunctorCartesian` → This would be a `FreeApply`, but if for every `Apply` instance we also have an `Applicative`, it would be useless.
- `FreeInvariantCartesian` → Doable but also useless if for every `InvariantCartesian` instance we also have an `InvariantMonoidal`.
- `FreeContravariantCartesian` → ?

That being said ìt might be possible to factor out the implementation of `FreeApplicative` and `FreeInvariantMonoidal` (and `FreeDivisible`?) into a more abstract structure. The related work of [Free Applicative Functors by Paolo Capriotti](http://www.paolocapriotti.com/assets/applicative.pdf) mentions the [HFree](https://hackage.haskell.org/package/free-functors-0.6.5/docs/Data-Functor-HFree.html) package in Haskell as some sort of "higher order free".

core/src/main/scala/cats/Cartesian.scala
```diff
@@ -16,4 +16,13 @@ import simulacrum.typeclass
   def product[A, B](fa: F[A], fb: F[B]): F[(A, B)]
 }
 
-object Cartesian extends CartesianArityFunctions
+object Cartesian extends CartesianArityFunctions with AlgebrCartesianInstances
```
adelbertc: s/AlgebrCartesian/AlgebraCartesian?

core/src/main/scala/cats/data/Const.scala
```diff
@@ -98,6 +97,17 @@ private[data] sealed abstract class ConstInstances0 extends ConstInstances1 {
 }
 
 private[data] sealed abstract class ConstInstances1 {
+  implicit def contInvariantMonoidal[C: Monoid]: InvariantMonoidal[Const[C, ?]] = new InvariantMonoidal[Const[C, ?]] {
```
adelbertc: s/cont/const

laws/src/main/scala/cats/laws/InvariantMonoidalLaws.scala
```diff
@@ -0,0 +1,27 @@
+package cats
+package laws
+
+/**
+ * Laws that must be obeyed by any `cats.InvariantMonoidal`.
+ */
+trait InvariantMonoidalLaws[F[_]] extends InvariantLaws[F] with CartesianLaws[F]{
```
adelbertc: What do these give you over the usual Cartesian laws?

adelbertc:
Hey @OlivierBlanvillain - sorry this took so long to review.

I have similar concerns to Julien regarding having a bunch of `Free` variants - I'm struggling because I can justify `FreeApplicative` and `FreeMonad` since `Applicative` and `Monad` capture fundamental properties of evaluation (e.g. context free vs. context dependent) though I also don't know where Cats should draw the line on including such structures. Including something like `FreeInvariantX` would tempt me to want to have the others if only for consistency's sake.. but then again I have an obsession with consistency.

Code-wise everything looks good modulo a couple typos and questions, but I'd like to hear other people's thoughts as well. @ceedubs ?

OlivierBlanvillain:
I would argue that `Applicative`s are only one part of the story as far as capturing fundamental properties of context free evaluations goes. If `Divisible` & `FreeDivisible` where also included it would lead to a nice symmetry between the available "composition syntaxes" and the Free structures:

| Algebraic Beast | Syntax                   | Free version        | Free use case      |
| --------------- | ------------------------ | ------------------- | ----------         |
| `Monad`         | `for {} yield f(...)   ` | `Free Monad`        | Tech talks :smile: |
| `Functor`       | `(x |@| y).map(f)      ` | `Free Applicative`  | Parsers            |
| `Contravariant` | `(x |@| y).contramap(f)` | `Free Divisible`?   | Pretty printers    |
| `Invariant`     | `(x |@| y).imap(f)(g)  ` | Stuff I wrote       | Codecs             |

ceedubs:
Warning: rambling brain dump below.

@OlivierBlanvillain thank you for this PR and like @adelbertc, I have to apologize for the radio silence for so long. Your code is high quality, but as mentioned it's tough to make decisions on what to include in cats and what not to. Some people have already complained about the cats-core jar being pretty large. I think that covariant forms of free (free monad, free applicative) are getting most of the attention in the FP Scala community right now, but I agree with you that the contravariant and invariant forms can also be really useful.

This puts my brain into a spiral of "So do we include this? Do we break free structures into a separate module/library? Do we have different modules for different forms of free structures?". At the moment I think going back to having free in a separate module is most appealing to me, but we've already been there and back, so I'm afraid that might frustrate people. If I recall, one of the biggest reasons that we broke free into a separate module was for trampolining, which is now handled by `Eval`. Personally I'm not too bothered by a large jar size, but I know that there are people who for various reasons are.

That's a partial braindump of my thoughts. Clearly there are no answers in there. I think I'm pretty happy with any route forward from here. Would people hate me if I put together a proof of concept PR of separating out a module or two now that the cats landscape has shifted a bit thanks to `Eval` and some other things?

ceedubs:
Sorry for not responding for so long but I started pondering a project yesterday that I think could really benefit from this. Now that free has its own module again, I'm less concerned about adding more structures to it.

The next week is really busy for me, but I'm hoping to do a more thorough review of this in the near future. I'm also open to any naming ideas that people have.

OlivierBlanvillain:
Cool :). I played a bit more with it (still in the context of codecs) and I have a cool example where [the same "configuration"](https://github.com/OlivierBlanvillain/free-validation/blob/master/src/main/scala/CustomUsage.scala#L38-L52) is used to generate a json parser, a json pretty printer, and a json schema. This example has quite a bit of boilerplate / manual wiring to combine unrelated algebras, but I have the feeling that this approach could lead to a neat alternative to automatic type class derivation entirely with cats.

core/src/main/scala/cats/InvariantMonoidal.scala
```diff
@@ -0,0 +1,52 @@
...
+  implicit val invariantMonoidalSemigroup: InvariantMonoidal[Semigroup] = new InvariantMonoidal[Semigroup] {
+    def product[A, B](fa: Semigroup[A], fb: Semigroup[B]): Semigroup[(A, B)] = new Semigroup[(A, B)] {
+      def combine(x: (A, B), y: (A, B)): (A, B) = fa.combine(x._1, y._1) -> fb.combine(x._2, y._2)
+    }
+
+    def imap[A, B](fa: Semigroup[A])(f: A => B)(g: B => A): Semigroup[B] = new Semigroup[B] {
+      def combine(x: B, y: B): B = f(fa.combine(g(x), g(y)))
+    }
+
+    def pure[A](a: A): Semigroup[A] = new Semigroup[A] {
```
ceedubs: This seems a little strange. Does it make sense to have this?

core/src/main/scala/cats/Cartesian.scala
```diff
@@ -16,4 +16,13 @@ import simulacrum.typeclass
...
 }
 
-object Cartesian extends CartesianArityFunctions
+object Cartesian extends CartesianArityFunctions with AlgebraCartesianInstances
+
+/**
+ * Cartesian instances for types that are housed in Algebra and therefore
+ * can't have instances for Cats type classes in their companion objects.
+ */
+private[cats] sealed trait AlgebraCartesianInstances {
```
ceedubs: Probably should change the name of this now that cats-kernel is merged.

core/src/main/scala/cats/InvariantMonoidal.scala
```diff
@@ -0,0 +1,52 @@
...
+  def pure[A](a: A): F[A]
+}
+
+object InvariantMonoidal extends AlgebraInvariantMonoidalInstances
+
+/**
+ * InvariantMonoidal instances for types that are housed in Algebra and therefore
+ * can't have instances for Cats type classes in their companion objects.
+ */
+private[cats] trait AlgebraInvariantMonoidalInstances {
```
ceedubs: Probably should change the name of this now that cats-kernel is merged.

ceedubs:
Should `Applicative` now extend `InvariantMonoidal`?

julienrf:
> Should Applicative now extend InvariantMonoidal?

Yes, I think so.

julienrf:
It would be great to also have `Monoidal` on its own, which would extend `Cartesian` and add `pure[A])(a: A): F[A]`.

@OlivierBlanvillain do you really need `pure` or would `unit: F[Unit]` be sufficient? (see also the discussion in #817)

core/src/main/scala/cats/InvariantMonoidal.scala
OlivierBlanvillain: I don't think it's wrong, but I'm not it's useful...

OlivierBlanvillain:
> Should Applicative now extend InvariantMonoidal?

Won't this lead to a weird diamond inheritance with `Invariant` being on both sides? I tried it but could not make scalac/simulacrum happy :/ https://github.com/OlivierBlanvillain/cats/commit/6226b5c30e4f36c7ffb84c0013e4297b531c3634

> It would be great to also have `Monoidal`

Shall I do it in the same PR?

> @OlivierBlanvillain do you really need pure or would unit: F[Unit] be sufficient? (see also the discussion in #817)

`def unit: F[Unit]` is sufficient since `def pure[A](x: A): F[A] = imap(unit)(_ => x)(_ => ())`. But since they are equivalent, I feel that pure is more elegant in that case.

julienrf:
>> It would be great to also have Monoidal
> Shall I do it in the same PR?

I think we could it afterward, let’s progress iteratively!

> def unit: F[Unit] is sufficient since def pure[A](x: A): F[A] = imap(unit)(_ => x)(_ => ()). But since they are equivalent, I feel that pure is more elegant in that case.

They are indeed equivalent once add functor capabilities to your `F[?]` (which is eventually always the case, I think…). So maybe it is OK to keep `pure`… Do @mpilquist or @travisbrown have a comment on that?

Fristi:
So these codecs (which are encode/decode really) are also encoded in Haskell as partial isomorphisms. See a example on this:

Bidirectional Parsing - https://youtu.be/NHRIV7UNiPU?t=41m24s

He points us to 

- [web-routes-boomerang](https://hackage.haskell.org/package/web-routes-boomerang-0.28.4.2)
- [JsonGrammer](https://hackage.haskell.org/package/JsonGrammar)
- [Invertible Syntax Descriptions: Unifying Parsing and Pretty Printing](http://www.informatik.uni-marburg.de/~rendel/unparse/rendel10invertible.pdf)

What we are trying to achieve with `FreeInvariantMonoidal` is something similar to the partial isomorphisms. I am considering to write my routing lib using, partial isos, what do you think @OlivierBlanvillain. Would this be applicable for your lib aswell?

OlivierBlanvillain:
Hey @Fristi :) I gave a superficial read to *Invertible Syntax Descriptions* a while ago, and indeed, it's trying to achieve the same same thing WRT encoders/decoders. However, I think that the `FreeInvariantMonoidal` is more general that what is done in this paper, my very fuzzy memory tells me that their approach would not be appropriate to generate something like a documentation type class (along with an encoder/decoder), while `FreeInvariantMonoidal` let's you do that.

Fristi:
Hmm you are right. Introspection is not included. A partial iso is just two functions. To make a category of it you should use kleisli arrows to make them compose. Monoids are used to denote choice between partial isos.

I've skimmed through JsonGrammer2 sources and it appears he using a free category (see: https://github.com/MedeaMelana/JsonGrammar2/blob/master/src/Language/JsonGrammar/Grammar.hs#L40). As you can see he also lifts; category and monoid operations to the algebra. After that he just makes instances category and monoid instances for his algebra. This is exactly what we are doing at the moment. We also lift some operations of Monoidal and Invariant to a 'generic' and free algebra, but it has no real strong name (and that's something with itches for me ;p).

ceedubs:
@Fristi @OlivierBlanvillain do you have a consensus on what the right way to go forward is? It seems like there's a lot of great stuff in this PR, so I don't want to hold it up forever, but if you have better ideas brewing, then we can of course put it on the back burner.

OlivierBlanvillain:
@ceedubs I would say that the `InvariantMonoidal` and `FreeInvariantMonoidal` proposed in the PR are *simply* the invariant analogous to `Applicative` and `FreeApplicative`. That's definitely a niche, but it's also unlikely to change. I would not be concerned about future work undoing what's is done here.

This is how I see things going forward:

- Merge/close this (depending on whether or not this stuff could be useful for someone else)
- Work on `Monoidal`
- Explore the `Contravariant` side of things

ceedubs:
@OlivierBlanvillain gotcha. That sounds good to me, and @mpilquist expressed interest in #802, so I suspect that he's on board.

It looks like there are merge conflicts again. Could you please refrain from making unrelated formatting changes in future PRs -- it can lead to lots of merge conflicts and makes the diffs a little tougher to review. It'd be great if you could also adopt the new naming convention for implicit instances (#1061), but that can wait until a subsequent PR if you'd prefer.

Fristi:
@ceedubs @OlivierBlanvillain Sounds fine by me! Let's keep this train going 👍

OlivierBlanvillain:
I resolve the conflicts. All the formatting stuff is isolated in the first commit (which I rebased with a brutal `git checkout --ours .` :P), sorry if it interfered with the review...

docs/src/main/tut/invariantmonoidal.md
```diff
@@ -0,0 +1,235 @@
...
+
+def product[A, B](fa: Semigroup[A], fb: Semigroup[B]): Semigroup[(A, B)] =
+  new Semigroup[(A, B)] {
+    def combine(x: (A, B), y: (A, B)): (A, B) = (x, y) match {
+      case ((xa, xb), (ya, yb)) => fa.combine(xa, ya) -> fb.combine(xb, yb)
+    }
+  }
+```
+
+Given an instance of `InvariantMonoidal` for `Semigroup`, we are able to combine existing `Semigroup` instances to form a new `Semigroup` by using the `Catesian` syntax:
```
DavidGregory084: `Cartesian`

laws/src/main/scala/cats/laws/discipline/Eq.scala
```diff
@@ -83,4 +83,12 @@ object eq {
   implicit val unitEq: Eq[Unit] = new Eq[Unit] {
     def eqv(a: Unit, b: Unit): Boolean = true
   }
+
+  // To be removed once https://github.com/non/algebra/pull/125 is published
+  implicit class EqAnd[A](self: Eq[A]) {
```
ceedubs: This shouldn't be needed any more.

core/src/main/scala/cats/Cartesian.scala
```diff
@@ -16,4 +16,13 @@ import simulacrum.typeclass
...
-object Cartesian extends CartesianArityFunctions
+object Cartesian extends CartesianArityFunctions with KernelCartesianInstances
+
+/**
+ * Cartesian instances for types that are housed in cats.kernel and therefore
+ * can't have instances for this type class in their companion objects.
+ */
+private[cats] sealed trait KernelCartesianInstances {
+  implicit val catsInvariantSemigroup: Cartesian[Semigroup] = InvariantMonoidal.catsInvariantMonoidalSemigroup
+  implicit val catsInvariantMonoid: Cartesian[Monoid] = InvariantMonoidal.catsInvariantMonoidalMonoid
```
ceedubs: Since `InvariantMonoidal` extends `Cartesian`, can these just live in the `InvariantMonoidal` companion object without also living here?

OlivierBlanvillain: Good catch, indeed it looks like scalac is able to find the ones in `InvariantMonoidal` companion object!

OlivierBlanvillain: Actually I take that back, the `tut` compilation fails because it can't find `Cartesian[Semigroup]` with this change. I reverted it and made this commit for reference: https://github.com/OlivierBlanvillain/cats/commit/07d2c14471a0fa8f105961685f1612c28a3785af

core/src/main/scala/cats/Cartesian.scala
ceedubs: Looks like this went back to `Algebra` when you backed out the change.

ceedubs:
I have one minor comment about an outdated reference to `Algebra`, but other than that, this looks good to me. @mpilquist or @adelbertc wdyt?

core/src/main/scala/cats/Cartesian.scala
OlivierBlanvillain: oops :smile:

kailuowang:
👍  It looks by and large good to me. Given the size of the PR, it certainly has space for further improvements (e.g. documentation on the free side), but I think those non-urgent improvements can come later. Would love to see people start profiting from this sooner rather than later.

ceedubs:
🎉

Fristi:
Nice work @OlivierBlanvillain  👍 🎉

OlivierBlanvillain:
Thanks all for your help! :smile:

PR #847, opened 2016-02-01
Fix State import in free monad docs

ceedubs:


ceedubs:
merging this without :+1: because I'm a renegade

PR #848, opened 2016-02-01
Draft of 0.4.0 release notes

travisbrown:
I'll update if any last-minute stuff makes it in before the release. Rendered version available [here](https://github.com/travisbrown/cats/blob/topic/0.4.0-release-notes/CHANGES.md) for now.

PR #809, opened 2016-01-15
Enforce binary compatibility with MiMa SBT plugin

DavidGregory084:
~~This should not be merged until 0.4.0 is released with cats-kernel as a distinct artifact.~~

Enables the Migration Manager SBT plugin for the kernel module and ensures binary compatibility issues are reported by `validateJVM`.

PR #850, opened 2016-02-01
Remove premature item in release notes

travisbrown:
@DavidGregory084 just [pointed out](https://github.com/typelevel/cats/pull/848#issuecomment-178183604) that I jumped the gun on #809.


non:
Oops! Good catch @DavidGregory084

ceedubs:
:+1:

PR #851, opened 2016-02-01
Update README and other documents for 0.4.0

travisbrown:
…and also the organization change.


non:
:+1:

non:
Since this is documentation, I think we can apply a laxer 1-signoff rule here. Thanks!

PR #852, opened 2016-02-01
Update build with org change

adelbertc:


non:
:+1: 

Thanks for fixing this @adelbertc! And apologies to everyone who's contributed to Cats for (unintentionally) taking sole credit for the project in the POM metadata.

travisbrown:
:+1: with one nitpick: would you mind using `https` URLs in the `developer` entries?

PR #853, opened 2016-02-02
Adds a new LiftTrans typeclass

stew:
This typeclass adds a `liftT` function which is similar to the liftM
function on the scalaz MonadTrans typeclass, however this one takes into
account that you might not need the power of a monad in order to be able
to lift into a transformer, for example, we are able to Lift any M[A]
into a Kleisli[M, B, A] regardless of whether or not M is a Monad,
Functor, or any other such thing. Thish would be useful in cases where
you are constructing an applicative computation using values for which
you don't have a monad.

### DIFF
--- a/core/src/main/scala/cats/syntax/transLift.scala
+++ b/core/src/main/scala/cats/syntax/transLift.scala
@@ -1,0 +1,10 @@
+package cats
+package syntax
+
+trait TransLiftSyntax {
+  implicit def transLiftSyntax[M[_], A](ma: M[A]): TransLiftOps[M, A] = new TransLiftOps(ma)
+}
+
+final class TransLiftOps[M[_], A](val ma: M[A]) extends AnyVal {
+  def liftT[MT[_[_],_]](implicit TL: TransLift[MT, M]): MT[M,A] = TL.liftT(ma)
+}
--- a/tests/src/test/scala/cats/tests/TransLiftTests.scala
+++ b/tests/src/test/scala/cats/tests/TransLiftTests.scala
@@ -1,0 +1,42 @@
+package cats
+package tests
+
+import data.{OptionT,XorT,WriterT,StreamingT, Kleisli, StateT}
+
+class TransLiftTests extends CatsSuite {
+
+  case class NoTypeclass[A](a: A)
+
+  case class JustFunctor[A](a: A)
+
+  implicit val jfFunctor: Functor[JustFunctor] = new Functor[JustFunctor] {
+    override def map[A,B](fa: JustFunctor[A])(f: A => B): JustFunctor[B] = JustFunctor(f(fa.a))
+  }
+
+  case class JustAp[A](a: A)
+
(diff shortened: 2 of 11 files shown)


core/src/main/scala/cats/syntax/transLift.scala
```diff
@@ -0,0 +1,10 @@
+package cats
+package syntax
+
+trait TransLiftSyntax {
+  implicit def transLiftSyntax[M[_], A](ma: M[A]): TransLiftOps[M, A] = new TransLiftOps(ma)
```
ceedubs: Do we need an `Unapply` version here?

tests/src/test/scala/cats/tests/TransLiftTests.scala
```diff
@@ -0,0 +1,42 @@
...
+
+  case class JustFunctor[A](a: A)
+
+  implicit val jfFunctor: Functor[JustFunctor] = new Functor[JustFunctor] {
+    override def map[A,B](fa: JustFunctor[A])(f: A => B): JustFunctor[B] = JustFunctor(f(fa.a))
+  }
+
+  case class JustAp[A](a: A)
+
+  implicit val jfApp: Applicative[JustAp] = new Applicative[JustAp] {
```
ceedubs: FWIW, there are `ListWrapper` helpers for this sort of thing.

tests/src/test/scala/cats/tests/TransLiftTests.scala
```diff
@@ -0,0 +1,42 @@
...
+    import StreamingT._
+    
+    val b: StreamingT[JustAp, Int] = JustAp(1).liftT[StreamingT]
+    val f: StateT[JustAp, Int, Int]  = JustAp(1).liftT[({type λ[α[_], β] = StateT[α, Int, β]})#λ]
+  }
+
+  test("transLift for, Kleisli doesn't require anything of the wrapped value"){
+    val e: Kleisli[NoTypeclass, Int, Int] = NoTypeclass(1).liftT[({type λ[α[_], β] = Kleisli[α, Int, β]})#λ]
+  }
+}
```
ceedubs: I think it would be good to have a test to see if we can lift a `ValidatedNel` into some transformers. I suspect it would reveal that we currently can't use syntax to do that because we don't have `Unapply`-based syntax.

ceedubs:
Excellent work, @stew. This is better than what I came up with yesterday and was planning on submitting via PR today.

:+1: in general. I left some minor comments that don't necessarily need to be addressed in this PR.

stew:
I'm stuck on getting this to work for Validation. I'd love for other ideas, or to see what @ceedubs has on this, or just merge this and say "tough luck, more type lambdas" to people ;)

ceedubs:
What's Validation? ;)

My current inclination is to merge this and say "tough luck, more type lambdas". But I'm also interested in exploring a bit more in this area. I'd say :+1: for a merge but we shouldn't feel set in stone on this approach.

adelbertc:
:+1: , leaving `Unapply` for future work

PR #856, opened 2016-02-03
Fix Streaming and StreamingT dropWhile functions

mikejcurry:
Default `Int => Boolean` in scalacheck just returns a function which
always returns the same value (`true` or `false`). This means that
operations like `dropWhile` are not tested accurately, as whilst the bounds (all `true` and all `false`) are checked, intermediate results which return different values based on input are not tested.

This fix creates a temporary `Arbitrary` instance for `Int => Boolean` and fixes the `dropWhile` on `Streaming` and `StreamingT` which were broken, and didn’t show up due to the deficiency in the `Arbitrary[Int => Boolean]` instance provided by scalacheck.

Maybe there is a better way to do these instances? This kind of hardcodes that the `Int => Boolean` returned is always just a comparison between the `Int` values, but seems better than always `true`/`false`.

### DIFF
--- a/core/src/main/scala/cats/data/Streaming.scala
+++ b/core/src/main/scala/cats/data/Streaming.scala
@@ -534,7 +534,7 @@ sealed abstract class Streaming[A] extends Product with Serializable { lhs =>
     this match {
       case Empty() => Empty()
       case Wait(lt) => Wait(lt.map(_.dropWhile(f)))
-      case Cons(a, lt) => if (f(a)) Wait(lt.map(_.dropWhile(f))) else Cons(a, lt)
+      case s @ Cons(a, lt) => if (f(a)) Wait(lt.map(_.dropWhile(f))) else s
     }
 
   /**
--- a/core/src/main/scala/cats/data/Streaming.scala
+++ b/core/src/main/scala/cats/data/Streaming.scala
@@ -524,7 +524,7 @@ sealed abstract class Streaming[A] extends Product with Serializable { lhs =>
    *
    * For example:
    *
-   *   Streaming(1, 2, 3, 4, 5, 6, 7).takeWhile(n => n != 4)
+   *   Streaming(1, 2, 3, 4, 5, 6, 7).dropWhile(n => n != 4)
    *
    * Will result in: Streaming(4, 5, 6, 7)
    */
@@ -532,7 +532,7 @@ sealed abstract class Streaming[A] extends Product with Serializable { lhs =>
     this match {
       case Empty() => Empty()
       case Wait(lt) => Wait(lt.map(_.dropWhile(f)))
-      case Cons(a, lt) => if (f(a)) Empty() else Cons(a, lt.map(_.dropWhile(f)))
+      case Cons(a, lt) => if (f(a)) Wait(lt.map(_.dropWhile(f))) else Cons(a, lt)
     }
 
   /**
--- a/core/src/main/scala/cats/data/StreamingT.scala
+++ b/core/src/main/scala/cats/data/StreamingT.scala
@@ -251,7 +251,7 @@ sealed abstract class StreamingT[F[_], A] extends Product with Serializable { lh
    */
   def dropWhile(f: A => Boolean)(implicit ev: Functor[F]): StreamingT[F, A] =
     this match {
-      case Cons(a, ft) => if (f(a)) Empty() else Cons(a, ft.map(_.dropWhile(f)))
+      case Cons(a, ft) => if (f(a)) Wait(ft.map(_.dropWhile(f))) else Cons(a, ft)
       case Wait(ft) => Wait(ft.map(_.dropWhile(f)))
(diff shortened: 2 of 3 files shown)


core/src/main/scala/cats/data/Streaming.scala
```diff
@@ -524,7 +524,7 @@ sealed abstract class Streaming[A] extends Product with Serializable { lhs =>
    *
    * For example:
    *
-   *   Streaming(1, 2, 3, 4, 5, 6, 7).takeWhile(n => n != 4)
+   *   Streaming(1, 2, 3, 4, 5, 6, 7).dropWhile(n => n != 4)
```
ceedubs: Now that we have sbt-doctest in place, we should change these to be compiler-verified!

ceedubs:
Yikes good catch. Thanks @mikejcurry!

ceedubs:
This probably warrants a 4.0.1 release :| cc @non @travisbrown

core/src/main/scala/cats/data/Streaming.scala
ceedubs: I submitted https://github.com/mikejcurry/cats/pull/1 to your PR branch

adelbertc:
I hope you mean 0.4.1 :-)

ceedubs:
@adelbertc oops quite right :)

core/src/main/scala/cats/data/Streaming.scala
```diff
@@ -523,16 +524,17 @@ sealed abstract class Streaming[A] extends Product with Serializable { lhs =>
...
+   * scala> s.dropWhile(n => n != 4).toList
+   * res0: List[Int] = List(4, 5, 6, 7)
+   * }}}
    */
   def dropWhile(f: A => Boolean): Streaming[A] =
     this match {
       case Empty() => Empty()
       case Wait(lt) => Wait(lt.map(_.dropWhile(f)))
-      case Cons(a, lt) => if (f(a)) Empty() else Cons(a, lt.map(_.dropWhile(f)))
+      case Cons(a, lt) => if (f(a)) Wait(lt.map(_.dropWhile(f))) else Cons(a, lt)
```
ceedubs: I think we can save an allocation by doing something like this here:
`case s @ Cons(a, lt) => if (f(a)) Wait(lt.map(_.dropWhile(f))) else s`

core/src/main/scala/cats/data/StreamingT.scala
```diff
@@ -244,14 +246,16 @@ sealed abstract class StreamingT[F[_], A] extends Product with Serializable { lh
...
+   * scala> import cats.std.list._
+   * scala> val s = StreamingT[List, Int](1, 2, 3, 4, 5, 6, 7)
+   * scala> s.dropWhile(n => n != 4).toList.flatten
+   * res0: List[Int] = List(4, 5, 6, 7)
+   * }}}
    */
   def dropWhile(f: A => Boolean)(implicit ev: Functor[F]): StreamingT[F, A] =
     this match {
-      case Cons(a, ft) => if (f(a)) Empty() else Cons(a, ft.map(_.dropWhile(f)))
+      case Cons(a, ft) => if (f(a)) Wait(ft.map(_.dropWhile(f))) else Cons(a, ft)
```
ceedubs: Same comment as https://github.com/typelevel/cats/pull/856/files#r51714784

core/src/main/scala/cats/data/Streaming.scala
mikejcurry: Good spot - Thanks! - In general, is there any reason to do `s @` instead and evaluate the expression to `s` instead of just evaluating to `this`? Aesthetically I prefer using `this`, but that is just personal preference - was wondering if there is any real reason I not aware of to prefer one of the other. Eitherways, I'll push through with the `s @` ; because the only reason I have any preference for using `this` is aesthetic, and that likely comes from Java/C++

non: To answer your question: there are times when you get a finer-grained type from `s @ ...` (due to the pattern match `...`) and then you would need to use `s` instead of `this` because you have better type information. If you don't need the extra type information then using `this` would be just fine; at that point it's just a stylistic preference.

mikejcurry: Cool - that makes a lot of sense - and thanks a million for following up!!! 

I guess in this instance there would have been no value, as there is no additional type information required, but I can definitely imagine scenarios, where this could be valuable.

ceedubs: Yep, totally agree with what @non said. It didn't occur to me that `this` would work fine in this example. I have no strong preference.

mikejcurry: Thanks! - I actually don't have much preference - it's more natural defaults :-). I went with your suggestion anyway.

non:
:+1:

PR #859, opened 2016-02-03
Add Contravariant documentation page

vpavkin:
Preview http://vpavkin.github.io/cats/tut/contravariant.html

### DIFF
--- a/docs/src/main/tut/contravariant.md
+++ b/docs/src/main/tut/contravariant.md
@@ -15,10 +15,10 @@ def contramap[A, B](f: B => A): F[A] => F[B]
 ```
 
 It looks like regular (also called `Covariant`) [`Functor`](functor.html)'s `map`,
-but with `f` transformation reversed.
+but with the `f` transformation reversed.
 
 Generally speaking, if you have some context `F[A]` for type `A`,
-and you can get `A` value out of `B` value — `Contravariant` allows you to get the `F[B]` context for `B`.
+and you can get an `A` value out of a `B` value — `Contravariant` allows you to get the `F[B]` context for `B`.
 
 Examples of `Contravariant` instances are [`Show`](show.html) and `scala.math.Ordering` (along with `algebra.Order`).
 
@@ -33,7 +33,7 @@ import cats.implicits._
 case class Money(amount: Int)
 case class Salary(size: Money)
 
-implicit val showMoney: Show[Money] = Show.show(m => s"${m.amount}$$")
+implicit val showMoney: Show[Money] = Show.show(m => s"$$${m.amount}")
 ```
 
 If we want to show a `Salary` instance, we can just convert it to a `Money` instance and show it instead.
--- a/docs/src/main/tut/contravariant.md
+++ b/docs/src/main/tut/contravariant.md
@@ -1,0 +1,78 @@
+---
+layout: default
+title:  "Contravariant"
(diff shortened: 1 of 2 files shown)


vpavkin:
Appreciate any input/corrections :)

docs/src/main/tut/contravariant.md
```diff
@@ -0,0 +1,78 @@
+---
+layout: default
+title:  "Contravariant"
+section: "typeclasses"
+source: "https://github.com/non/cats/blob/master/core/src/main/scala/cats/functor/Contravariant.scala"
```
ceedubs: This should switch from `non` to `typelevel`. We probably have the same mistake in lots of docs. We could probably just give relative paths here and use site-wide config for the base source URL.

docs/src/main/tut/contravariant.md
```diff
@@ -0,0 +1,78 @@
...
+
+The `Contravariant` type class is for functors that define a `contramap`
+function with the following type:
+
+```scala
+def contramap[A, B](f: B => A): F[A] => F[B]
+```
+
+It looks like regular (also called `Covariant`) [`Functor`](functor.html)'s `map`,
+but with `f` transformation reversed.
```
ceedubs: Might read more smoothly with as "but with _the_ `f` transformation..."

docs/src/main/tut/contravariant.md
```diff
@@ -0,0 +1,78 @@
...
+
+Generally speaking, if you have some context `F[A]` for type `A`,
+and you can get `A` value out of `B` value — `Contravariant` allows you to get the `F[B]` context for `B`.
```
ceedubs: maybe "you can get _an_ `A` value out of a `B` value"?

docs/src/main/tut/contravariant.md
```diff
@@ -0,0 +1,78 @@
...
+Say we have class `Money` with a `Show` instance, and `Salary` class. 
+
+```tut:silent
+import cats._
+import cats.implicits._
+
+case class Money(amount: Int)
+case class Salary(size: Money)
+
+implicit val showMoney: Show[Money] = Show.show(m => s"${m.amount}$$")
```
ceedubs: I'm used to the `$` being before the amount. Is this another weird thing that only the US does? :)

docs/src/main/tut/contravariant.md
```diff
@@ -0,0 +1,78 @@
...
+```
+
+If we want to show a `Salary` instance, we can just convert it to a `Money` instance and show it instead.
```
ceedubs: I was fully expecting a Jerry Maguire quote to make it's way in here. Have I shown my age? :)

ceedubs:
I made a few minor comments, but this is great. Thanks, @vpavkin!

docs/src/main/tut/contravariant.md
vpavkin: Strange, but currently `useful links` section, with those `source` links is hidden... 
I think this issue and all links should be fixed in another pr. I'll take care of that.

vpavkin:
@ceedubs I'm not a native speaker, thank you for covering my back here! :)
Fixed the minors, will open a separate pr with sources links fixed

ceedubs:
:+1:

non:
Thanks so much! :+1:

PR #862, opened 2016-02-04
Update to ScalaCheck 1.13.0

fthomas:
I haven't tested this locally, so let's see how `Cogen` affects Cats.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -54,7 +54,7 @@ lazy val commonJvmSettings = Seq(
 
 lazy val catsSettings = buildSettings ++ commonSettings ++ publishSettings ++ scoverageSettings
 
-lazy val scalacheckVersion = "1.12.5"
+lazy val scalacheckVersion = "1.13.0"
 
 lazy val disciplineDependencies = Seq(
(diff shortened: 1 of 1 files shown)


travisbrown:
I ran into ScalaTest runtime errors when I tried this for circe, so I'd guess we'll need to wait for a new milestone here as well.

fthomas:
The CI failed because of errors like this:
    
    [error] /home/travis/build/typelevel/cats/laws/src/main/scala/cats/laws/discipline/AlternativeTests.scala:30:
    could not find implicit value for parameter a3: org.scalacheck.Arbitrary[A => B]

mikejcurry:
This could possibly be related to #856, which added a temporary Int => Boolean arbitrary instance. This should be superfluous with the new scalacheck so can probably be removed.

I can investigate later, but won't be home for a few hours to do so.

ceedubs:
See also https://github.com/scalatest/scalatest/issues/837

fthomas:
Ok, so updating ScalaCheck is more involved than just updating the version number. Since I currently don't have enough time to work on this, I'll close this PR so that someone else can take over.

PR #863, opened 2016-02-05
Version 0.4.1 release notes and prep

ceedubs:


travisbrown:
+1. Once the release tag is pushed you should be able to choose "Add release notes" from the tag page (which you get to from the releases page) and copy the new Markdown text from here.

ceedubs:
Thanks @travisbrown. I'll give it a try once the tag is pushed.

non:
:+1:

Issue #800, opened 2016-01-13
Bitraverse type class


adelbertc:

ceedubs:
This would extend `Bifoldable`, which it seems like people have agreed on adding in #94 but that nobody has gotten around to yet.

ceedubs:
oh and :+1: :)

adelbertc:
Does anyone have a use case for this?

travisbrown:
I've used both `bisequence` and `bitraverse` fairly often, for things like a more generic `zip`:

```scala
scala> import scalaz._, Scalaz._
import scalaz._
import Scalaz._

scala> (some('a'), some(1)).bisequenceU
res0: Option[(Char, Int)] = Some((a,1))
```

…or printing / logging both sides of an option, etc.

adelbertc:
working on this

PR #864, opened 2016-02-05
Add Bifoldable, fixes #94

### LINKED ISSUE #94: Bifoldable
`Bifoldable` is a type class (currently not in cats) for a type that has two separate `Foldable`s. For example, `Or` has a `Foldable` instance (via its `Traverse` instance) that folds over the `RightOr` case, but `Or` could equally fold over the `LeftOr` case.

One of my favorite methods from scalaz is [`MonadPlus.separate`](https://github.com/scalaz/scalaz/blob/aab1891bb193432bab21a5e6c5055346ec0d55d1/core/src/main/scala/scalaz/MonadPlus.scala#L24). This can do things like take a `List[A Or B]` and turn it into a `(List[A], List[B])`. Or similarly, take a `List[(A, B)]` and turn it into a `(List[A], List[B])`. In Cats, I believe we could implement the same sort of method on `MonadFilter` if we were to bring in a `Bifoldable` type class.

What do people think? Is this something that would belong in `core`?

adelbertc:
Also:
- Add MonadCombine separate
- Add serializability check for Bifunctor[Xor]

### DIFF
--- a/core/src/main/scala/cats/Bifoldable.scala
+++ b/core/src/main/scala/cats/Bifoldable.scala
@@ -3,7 +3,7 @@ package cats
 /**
  * A type class abstracting over types that give rise to two independent [[cats.Foldable]]s.
  */
-trait Bifoldable[F[_, _]] extends Serializable { self =>
+trait Bifoldable[F[_, _]] extends Any with Serializable { self =>
   /** Collapse the structure with a left-associative function */
   def bifoldLeft[A, B, C](fab: F[A, B], c: C)(f: (C, A) => C, g: (C, B) => C): C
 
@@ -11,7 +11,7 @@ trait Bifoldable[F[_, _]] extends Serializable { self =>
   def bifoldRight[A, B, C](fab: F[A, B], c: Eval[C])(f: (A, Eval[C]) => Eval[C], g: (B, Eval[C]) => Eval[C]): Eval[C]
 
   /** Collapse the structure by mapping each element to an element of a type that has a [[cats.Monoid]] */
-  def bifoldMap[A, B, C](fab: F[A, B])(f: A => C)(g: B => C)(implicit C: Monoid[C]): C =
+  def bifoldMap[A, B, C](fab: F[A, B])(f: A => C, g: B => C)(implicit C: Monoid[C]): C =
     bifoldLeft(fab, C.empty)(
       (c: C, a: A) => C.combine(c, f(a)),
       (c: C, b: B) => C.combine(c, g(b))
--- a/core/src/main/scala/cats/Bifoldable.scala
+++ b/core/src/main/scala/cats/Bifoldable.scala
@@ -3,7 +3,7 @@ package cats
 /**
  * A type class abstracting over types that give rise to two independent [[cats.Foldable]]s.
  */
-trait Bifoldable[F[_, _]] extends Serializable {
+trait Bifoldable[F[_, _]] extends Serializable { self =>
   /** Collapse the structure with a left-associative function */
   def bifoldLeft[A, B, C](fab: F[A, B], c: C)(f: (C, A) => C, g: (C, B) => C): C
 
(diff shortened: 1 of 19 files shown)


adelbertc:
Hmm Travis isn't running - closing and re-opening to see what happens.

adelbertc:
There we go

core/src/main/scala/cats/data/Xor.scala
```diff
@@ -166,9 +166,19 @@ private[data] sealed abstract class XorInstances extends XorInstances1 {
       def combine(x: A Xor B, y: A Xor B): A Xor B = x combine y
     }
 
-  implicit def xorBifunctor: Bifunctor[Xor] =
-    new Bifunctor[Xor] {
+  implicit def xorBifunctor: Bifunctor[Xor] with Bifoldable[Xor] =
```
ceedubs: It looks like we are combining the `Bifunctor` and `Bifoldable` instances in some places but not others. I guess if we are going to go forward with #800 then they will eventually be combined anyway. Was there any particular reason behind when they've been combined and when they haven't?

core/src/main/scala/cats/std/tuple.scala
```diff
@@ -0,0 +1,15 @@
+package cats
+package std
+
+trait TupleInstances extends Tuple2Instances
```
ceedubs: I think this will cause merge conflicts with #867, but I suppose we can just cross that bridge when we get to it.

core/src/main/scala/cats/data/Xor.scala
adelbertc: The `Bifunctor[Xor]` instance was there prior to this PR. It wasn't in other places so I figured I would do those separately, hence https://github.com/typelevel/cats/pull/867. However because it looks like we have a use case for https://github.com/typelevel/cats/issues/800 I may abandon #867  and add `Bitraverse` once this is OK'ed and merged :-)

core/src/main/scala/cats/std/tuple.scala
adelbertc: Since it looks like we're doing https://github.com/typelevel/cats/issues/800 I'll be closing that

adelbertc:
As a note there an also be `Bifunctor[Validated]` but I'll leave that alone since I'll just define `Bitraverse` for it in a future PR, unless someone beats me to it :-)

ceedubs:
:+1:

ceedubs:
Thanks, @adelbertc!

adelbertc:
:+1: @stew  ? :D

core/src/main/scala/cats/Bifoldable.scala
```diff
@@ -0,0 +1,46 @@
+package cats
+
+/**
+ * A type class abstracting over types that give rise to two independent [[cats.Foldable]]s.
+ */
+trait Bifoldable[F[_, _]] extends Serializable { self =>
```
mpilquist: `extends Any with Serializable` ?

mpilquist:
:+1: I didn't see syntax for `Bifoldable`. Is that intentional?

core/src/main/scala/cats/Bifoldable.scala
adelbertc: Hmm I see why we would want to do that, but it doesn't seem like we do it in other type classes.. do we?

mpilquist: Simulacrum does it automatically. We probably aren't consistent with the `F[_,_]` type classes though.

adelbertc: Alright I added it for `Bifoldable`, I've added [an issue](https://github.com/typelevel/cats/issues/887) to do it for the others and I'll probably end up doing it in a future PR unless someone beats me to it

adelbertc:
Added some more stuff per @mpilquist 's comments - @ceedubs re-review? :D

ceedubs:
:+1:

Out of curiosity, is the reason we want to extend `Any` so that it's a universal trait that can be extended by values classes? Are there other reasons?

I could potentially see some value in providing methods to extract a `Foldable` for the left and a `Foldable` for the right. These could be handy for checking the `Foldable` laws for each side and for right-biased types like `Xor`, ensuring that the right `Foldable` is consistent with the `Foldable` instance. However, I don't know if people would find themselves using these in real-world code, so even if we did this maybe they should just reside in tests until someone finds themself wanting them?

ceedubs:
I'll give @mpilquist another chance to review since there were some changes since his last :+1:

mpilquist:
:+1:

PR #873, opened 2016-02-10
Fix OptionIdOps.some to always return Some

### LINKED ISSUE #871: OptionIdOps.some inconsistent
While obviously `null` shouldn't be used, if it were to creep in then `x.some` violates the principle of least surprise. Given this code:

```scala
val s: String = null
val o = s.some
```

I might expect `o` to be `Some(...)`, but it ends up as `None`.

vpavkin:
Fix #871

### DIFF
--- a/tests/src/test/scala/cats/tests/OptionTests.scala
+++ b/tests/src/test/scala/cats/tests/OptionTests.scala
@@ -77,10 +77,11 @@ class OptionTests extends CatsSuite {
     }
   }
 
-  // a test for OptionIdOps.some to always return Some
+  // a test for OptionIdOps.some to return Some even if the argument is null
 
-  test("OptionIdOps.some"){
+  test(".some with null argument still results in Some #871") {
     val s: String = null
-    s.some.exists(_ == null) should === (true)
+    // can't use `s.some should === (Some(null))` here, because it leads to NullPointerException
+    s.some.exists(_ == null) should ===(true)
   }
 }
--- a/tests/src/test/scala/cats/tests/OptionTests.scala
+++ b/tests/src/test/scala/cats/tests/OptionTests.scala
@@ -81,6 +81,6 @@ class OptionTests extends CatsSuite {
 
   test("OptionIdOps.some"){
     val s: String = null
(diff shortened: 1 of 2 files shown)


tests/src/test/scala/cats/tests/OptionTests.scala
```diff
@@ -76,4 +76,11 @@ class OptionTests extends CatsSuite {
       isEq.lhs should === (isEq.rhs)
     }
   }
+
+  // a test for OptionIdOps.some to always return Some
+
+  test("OptionIdOps.some"){
+    val s: String = null
+    s.some.contains(s) should === (true)
```
vpavkin: Due to NPE I wasn't able to write it as `s.some should === (Some(null))`

ceedubs: Thanks, @vpavkin!

Would you mind leaving a comment in the code itself about why it's done this way?

Also, would you mind naming the test something that better identifies that it's checking for behavior when `null` is passed in? Maybe something like `test(".some with null argument still results in Some #871")`.

ceedubs: Looks like `Option.contains` isn't available on Scala 2.10

vpavkin:
@ceedubs my bad. Added more comments and changed code to use `exists`, which should work for 2.10.
Let's see what travis thinks about it.

tests/src/test/scala/cats/tests/OptionTests.scala
```diff
@@ -76,4 +76,12 @@ class OptionTests extends CatsSuite {
       isEq.lhs should === (isEq.rhs)
     }
   }
+
+  // a test for OptionIdOps.some to return Some even if the argument is null
```
ceedubs: This comment probably isn't really needed since the test name says basically the same thing.

ceedubs:
:+1: thanks a bunch, @vpavkin!

vpavkin:
updated comment :)

stew:
:+1:

Issue #791, opened 2016-01-09
Support for enrichment of standard library classes


benhutchison:
This issue began with I posted a question to cats gitter - [is there support for conversion of `Boolean` to `Option` values, like the `option[A](a: A)` method in Scalaz](https://gitter.im/non/cats?at=568f09bd7669220736028310). 

Two other example where people have/want-to enrich standard lib classes with extra behavior:
- Issue #749 
-  https://github.com/non/cats/blob/master/core/src/main/scala/cats/syntax/option.scala#L26

The [scalaz.syntax.std](https://github.com/scalaz/scalaz/tree/series/7.3.x/core/src/main/scala/scalaz/syntax/std) package has a bunch more that could be supported if/when useful or requested, egs:
- [BooleanOps](https://github.com/scalaz/scalaz/blob/series/7.1.x/core/src/main/scala/scalaz/syntax/std/BooleanOps.scala) 
- [ListOps](https://github.com/scalaz/scalaz/blob/series/7.3.x/core/src/main/scala/scalaz/syntax/std/ListOps.scala)

Opinion seemed to run against adding such "conveniences" into cats core in general, eg @travisbrown [response](https://gitter.im/non/cats?at=568f0f73b1f439094a06ee6a). But in following discussion there seemed to be agreement they have value and for putting them _somewhere_ in a new module or project (working title `extra-syntax`).

Our immediate goals are to answer:
1. What should the scope/charter of `extra-syntax` be? I'd say the primary goal is _Improvement/enrichment of the standard library classes_, with secondary goal _increase familiarity & portability for Scalaz programmers_. 
   
   The principle I stumble to articulate though is, _when does code go into `extra-syntax` vs [cats syntax](https://github.com/non/cats/blob/master/core/src/main/scala/cats/syntax/)_? 
   
   Proposal:
   - conversions of std types to cats datatypes (eg `Option` => `Xor`) go into cats core
   - methods that enable/support using cats typeclasses go into cats core
   
   Anything else goes to extra, such as conversions between std types, or extra methods like `option.cata`.
   
   Or is the distinction more about barrier-to-entry/level-of-consensus? If everyone agrees its central, it can go to cats core, if there's doubt its goes to extras?  Not too keen on this, I'd prefer a more objective criteria ideally.. 
[...]

PR #876, opened 2016-02-12
Adds syntax for Strings to parse numbers and booleans.

cwmyers:
Includes tests for `parseInt`, `parseLong`, `parseByte`, `parseShort`, `parseDouble`, `parseFloat` and `parseBoolean`.

This also tests the typeclass for `Show`.

### DIFF
--- a/core/src/main/scala/cats/syntax/all.scala
+++ b/core/src/main/scala/cats/syntax/all.scala
@@ -29,6 +29,7 @@ trait AllSyntax
     with Show.ToShowOps
     with SplitSyntax
     with StreamingSyntax
+    with StringSyntax
     with StrongSyntax
     with TraverseSyntax
     with XorSyntax
--- a/core/src/main/scala/cats/syntax/string.scala
+++ b/core/src/main/scala/cats/syntax/string.scala
@@ -1,0 +1,42 @@
+package cats
+package syntax
+
+import cats.data.Xor
+
+trait StringSyntax {
+  implicit def stringSyntax(s: String): StringOps = new StringOps(s)
+}
+
+final class StringOps(val s: String) extends AnyVal {
+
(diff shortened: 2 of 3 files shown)


ceedubs:
Thank you for your contribution, @cwmyers.

I think that these methods are useful, but I find myself wondering whether they fit into cats. They fall into the category of "helper methods" as opposed to providing abstractions or data types that are useful for functional programming.

What do other people think?

cwmyers:
Hi @ceedubs 

You have a fair point.  This PR can from my assessment of moving our code base from Scalaz to cats.  I'm conscious that cats' goals may not necessarily be a drop in replacement for scalaz but these parse functions were useful to us.  Also, I was wondering of the overhead of bringing an extra helper library that returned cats data types.

That said, if the community recommends this goes to a helper library I'm happy for that to happen.  Is there general interest in incepting a helper library project?

Many thanks,
Chris

travisbrown:
This question has come up a few times, and as far as I know nobody's actually started a project for functions like this, but I think one would be very useful.

I agree that these don't belong in cats-core, but I'd personally be happy to see something like a cats-misc module in the cats repo—that way when people come from Scalaz the path is clear. I guess this could happen outside the cats repo as well, but I'm afraid that that'd mean we just keep kicking the can down the road until someone comes up with a funny project name or whatever.

fsvehla:
Could this be part of Cats if it would conform to the idea of `Read`?

benhutchison:
Hi @cwmyers I think your string enrichments are similar to #791 ? Seems clear that a module to support these kinds of extensions is needed somewhere. I posed some questions in #791 I was hoping to get feedback on, but no response yet, will bump it in gitter since you've hit same issue...

DavidGregory084:
Can I suggest that you call this `fat-cats` :wink:

cwmyers:
I'm going to close this PR as @benhutchison has created the mouse project for non-core helper features.

This PR has moved here: https://github.com/benhutchison/mouse/pull/1

PR #877, opened 2016-02-13
Remove required laziness in Prod, fixes #615

### LINKED ISSUE #615: Reconsider lazy nature of Prod
## The issue

This is a follow-up to https://github.com/non/cats/issues/607#issuecomment-154830206.

Currently `Prod` has lazy/memoized semantics baked into it: https://github.com/non/cats/blob/c29318233031ec8ce0f462de4c89f3bf3429bb66/core/src/main/scala/cats/data/Prod.scala#L14-L19. The `apply` method is the only way to create a `Prod` instance, and it forces `Eval.Later` as an evaluation strategy.

This makes it unclear how to approach `Show` instances for `Prod`. Ideally, we could just generate a trivial `Show[F, G, A]` instance whenever we have `Show[F[A]]` and `Show[G[A]]` instances available. However, once laziness and memoization is in the picture, how to approach this becomes unclear. Should a `Show` instance force a lazy value to be computed?

## My opinion

The forced lazy nature of `Prod.apply` seems a bit out of place in Cats to me. It feels like forcing Haskell-semantics in a Scala library. My gut reaction is that if people want laziness they should make it explicit in their `F` and `G` types (likely via `Eval`) instead of it happening under the hood without an alternative.

I'd be okay with a "constructor" method that is similar to the existing one or one that takes explicit `Eval[F[A]]` and `Eval[G[A]]` arguments, but I don't think it should be the only option.

adelbertc:


ceedubs:
Thanks for bringing this out of the archives, @adelbertc! :)

:+1: for this change in general. Is there anything in particular about `Prod` that makes it useful to create the `Always` helpers? This isn't something that we've generally done for other data types.

adelbertc:
I mostly just added it since I figured if we would have convenience for `later` might as well have it for `always`. Say the word and it's gone :-)

ceedubs:
@adelbertc I guess I really meant having `Eval` helpers at all (both `always` and `later`). I have never used `Prod`, so I'm not familiar with its common usage.

adelbertc:
Ah I see. I've never found myself wanting them - @eed3si9n do you know of the motivation behind this?

eed3si9n:
> @eed3si9n do you know of the motivation behind this?

If I understand it correctly, the usefulness of `Prod`, described in EIP, is that it's an automatically derived return type of an applicative function composition [`product`](https://github.com/typelevel/cats/blob/v0.4.1/core/src/main/scala/cats/data/Func.scala#L81-L88).
This represents a split in the calculation graph. An example of this might be calculating rows and columns of sudoku "in parallel" - http://eed3si9n.com/sudoku-using-func.
I'm putting the scare quotes because the parallelism is only in semantics, but the calculation is actually not done concurrently. The motivation behind putting lazy evaluation here is to delay the calculation of a branch so we can avoid them, when not needed.

ceedubs:
Thanks for the explanation @eed3si9n. There are quite a few places where it _could_ be helpful to have a lazy argument (`Semigroup.combine`, etc) but that we have chosen to let people handle this via `Eval` if they choose to instead of baking the laziness into things. I'm inclined to say that people could do the same here and that we don't really need helpers for `Lazy` etc. I guess they don't really cause any harm, but it seems inconsistent with other places. For example we have `Applicative.pureEval` but we don't provide helpers for `Lazy`, `Always`, etc. I guess I'm inclined to take the same approach here unless we have a compelling reason to do otherwise.

ceedubs:
:+1:

non:
👍

PR #878, opened 2016-02-13
Fix bug in freemonad doc

ceedubs:
Thanks to @tek for pointing out that the "pure" `State`-based compiler
was using the mutable map from the "impure" example.


adelbertc:
:+1: , merging because doc

PR #879, opened 2016-02-14
fixedpoints, folds and traverse

LucDupAtGitHub:
first humble contribution, 
hoping (later) to be able to implement `FixTraverse` 's  `traverseAlgebra` for all polynomial functors

### DIFF
--- a/laws/src/main/scala/cats/laws/FixLaws.scala
+++ b/laws/src/main/scala/cats/laws/FixLaws.scala
@@ -1,0 +1,24 @@
+package cats.laws
+
+import cats.Functor
+import cats.syntax.functor._
+
+import cats.fix.Fix
+
+trait FixLaws[F[_]] {
+  implicit def F: Functor[F]
+
+
+  def foldReflectionLaw(fix: Fix[F]): IsEq[Fix[F]] =
+    fix.fold[Fix[F]](Fix(_)) <-> fix
+
+  def foldCancellationLaw[A](unFix: F[Fix[F]], algebra: F[A] => A) =
+    Fix(unFix).fold[A](algebra) <-> algebra(unFix.map(_.fold[A](algebra)))
+
+}
+
+object FixLaws {
+  def apply[F[_]](implicit ev: Functor[F]): FixLaws[F] =
+    new FixLaws[F] { def F: Functor[F] = ev }
+}
+
--- a/laws/src/main/scala/cats/laws/FixLaws.scala
+++ b/laws/src/main/scala/cats/laws/FixLaws.scala
@@ -12,13 +12,15 @@ trait FixLaws[F[_]] {
   def foldReflectionLaw(fix: Fix[F]): IsEq[Fix[F]] =
     fix.fold[Fix[F]](Fix(_)) <-> fix
 
-  def foldCancellationLaw[A](unFix: F[Fix[F]], algebra: F[A] => A) =
+  def foldCancellationLaw[A](unFix: F[Fix[F]], algebra: F[A] => A): IsEq[A] =
     Fix(unFix).fold[A](algebra) <-> algebra(unFix.map(_.fold[A](algebra)))
 
 }
 
 object FixLaws {
   def apply[F[_]](implicit ev: Functor[F]): FixLaws[F] =
-    new FixLaws[F] { def F: Functor[F] = ev }
+    new FixLaws[F] {
+      def F: Functor[F] = ev
+    }
 }
 
--- a/core/src/main/scala/cats/fix/Fix.scala
+++ b/core/src/main/scala/cats/fix/Fix.scala
@@ -1,0 +1,70 @@
+package cats.fix
+
+import cats.{Applicative, Traverse, Functor}
+import cats.syntax.functor._
+
+/**
+  * `Fix`
+  *
+  * `Fix[F]` is the least fixed point of the functor `F`
+  *
+  *
+  * `Algebra`
+  *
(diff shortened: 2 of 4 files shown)


julien-truffaut:
there seems to be an error in the doc, can you `sbt validate` locally

benhutchison:
I'd like to understand the use cases that this PR addresses? Would you be able to describe the purpose/use-cases in the PR comment or in some associated doc? I see that there's demo code included, but it wasn't clear to me what problems it was tackling. 

For context, @stew outline a design principle of cats library that I support here [https://youtu.be/AAMncMqpOds?t=925], namely that everything in the library is there because it has clear use cases.

edmundnoble:
I believe it would benefit this pull request to use `Eval` where `Fix` is recursed over, to provide a degree of stack safety.

laws/src/main/scala/cats/laws/FixLaws.scala
```diff
@@ -0,0 +1,24 @@
...
+import cats.fix.Fix
+
+trait FixLaws[F[_]] {
+  implicit def F: Functor[F]
+
+
+  def foldReflectionLaw(fix: Fix[F]): IsEq[Fix[F]] =
+    fix.fold[Fix[F]](Fix(_)) <-> fix
+
+  def foldCancellationLaw[A](unFix: F[Fix[F]], algebra: F[A] => A) =
```
adelbertc: Annotate return type

DavidGregory084:
It might be worth linking to Bartosz Milewski's [excellent article](https://www.schoolofhaskell.com/user/bartosz/understanding-algebras) in the doc for a bit of additional background; like @benhutchison I am interested to know if you have some more motivating examples as to how this may be useful.

LucDupAtGitHub:
Folks,

`fold` generalizes `foldr` on lists to any fixed point (i.e. initial algebra) of a functor
moreover

the idea behind 'fold` is to first do sums and producta and then do recursion
(instead of doing everything in one step which is not modular)

also `traverse` can easily be defined in terms of `fold`

I am convinced that for polynomial functors (sums of products of functors)
`traverse` can be written once and for all

One of the goals of cats is to adhere to the DRY principle, yes?

So if only for that reason I think the code is useful.

More importantly, like I said above, I still think it is a valuable alternative (to using Free Monads, Free Applicatives or Free Arrows) for modular syntax/semantics

By the way: this is *very* old stuff  

cfr 
https://www.researchgate.net/publication/2550340_Using_Catamorphisms_Subtypes_and_Monad_Transformers_for_Writing_Modular_Functional_Interpreters

LucDupAtGitHub:
By the way: thanks for all replies.
I will study the `Eval` stuff and try to incorporate it in the code if possible at all.

LucDupAtGitHub:
about Travis:

running `sbt validate` locally gives me

```
[info] - Traverse[Map[Int, ?]].serializable.can serialize and deserialize
[info] FoldableTestsAdditional:
[info] - Foldable[List] *** FAILED ***
[info]   scala.scalajs.js.JavaScriptException:

```

laws/src/main/scala/cats/laws/FixLaws.scala
LucDupAtGitHub: done (btw: return type is `IsEq[A]`)

mikejcurry:
I personally use -Xmx5G if I am going to be doing the Scala JS build - that might be overkill. 

`sbt validate` will perform the Scala JS build, at least by default. Another option, if you're confident JS specific issues are unlikely, is to use `validateJVM` instead of `validate`, which will only run the build for the JVM only - Travis will still build Scala JS, so if you missed anything it will still get caught.

With regard to the Scala JS stuff and the different modes, see: https://github.com/typelevel/cats#scala-and-scala-js

LucDupAtGitHub:
more info about `sbt validate` locally

```
[error] /opt/home/data/scala/git/cats/core/src/main/scala/cats/fix/FixApp.scala:33: discarding unmoored doc comment
[error]      /**
[error]      ^
[error] /opt/home/data/scala/git/cats/core/src/main/scala/cats/fix/FixApp.scala:41: discarding unmoored doc comment
[error]      /**
[error]      ^
[error] /opt/home/data/scala/git/cats/core/src/main/scala/cats/fix/FixApp.scala:92: discarding unmoored doc comment
[error]      /**
[error]      ^
[info] No documentation generated with unsuccessful compiler run

```

what's wrong with doc comments ???

DavidGregory084:
@LucDupAtGitHub I think this is because there are newlines between the doc comments and the definitions that they document? Scaladoc is very picky I guess.

LucDupAtGitHub:
man!

adelbertc:
Thanks for this! Are you still working on this? I see there's a FixApp there for demonstration purposes, perhaps that can be turned into either a tutorial ([example](https://github.com/typelevel/cats/blob/master/docs/src/main/tut/kleisli.md)) or factored out into tests. Also looks like FixTests is incomplete

DavidGregory084:
@LucDupAtGitHub just had a chance to properly read your reply above; please accept my apologies if it sounded as if I didn't think this was a valuable contribution to cats (I am certainly not one of the persons who would decide this by any means) - in fact I feel quite the opposite.

LucDupAtGitHub:
@adelbertc 

Yes, it would be better to turn the app into a tutorial exmaple

... gonna work on that ...

LucDupAtGitHub:
@DavidGregory084 

Hey, no worries, my "man!" remark was a remark about scaladoc.
Not at all about you (on the contrary)!

PR #880, opened 2016-02-15
Optimize Eq[Vector[A]] instance

TomasMikula:
* Check size equality first.
 * Avoid pattern matching on Vector via `x +: xs`, which creates a new Vector instance via `Vector.tail`.
 * Added `@tailrec` annotation to the `loop` method.

I used this mini-benchmark to assess the improvement:

```scala
import cats.Eq
import cats.std.vector._
import cats.std.int._

val v = (1 to 10000000).toVector
val EqV = Eq[Vector[Int]]

{
  val t0 = System.currentTimeMillis
  EqV.eqv(v, v)
  val t1 = System.currentTimeMillis
  println(s"${t1 - t0}ms")
}
```

The result is 

before refactoring | after refactoring
--------- | -------
6542ms | 161ms

### DIFF
--- a/core/src/main/scala/cats/std/vector.scala
+++ b/core/src/main/scala/cats/std/vector.scala
@@ -68,16 +68,11 @@ trait VectorInstances {
   implicit def eqVector[A](implicit ev: Eq[A]): Eq[Vector[A]] =
     new Eq[Vector[A]] {
       def eqv(x: Vector[A], y: Vector[A]): Boolean = {
-        def loop(xs: Vector[A], ys: Vector[A]): Boolean =
-          xs match {
-            case Seq() => ys.isEmpty
-            case a +: xs =>
-              ys match {
-                case Seq() => false
-                case b +: ys => if (ev.neqv(a, b)) false else loop(xs, ys)
-              }
-          }
-        loop(x, y)
+        @tailrec def loop(from: Int): Boolean =
+          if(from == x.size) true
+          else ev.eqv(x(from), y(from)) && loop(from + 1)
+
+        (x.size == y.size) && loop(0)
       }
     }
 }
--- a/core/src/main/scala/cats/std/vector.scala
+++ b/core/src/main/scala/cats/std/vector.scala
@@ -68,11 +68,11 @@ trait VectorInstances {
   implicit def eqVector[A](implicit ev: Eq[A]): Eq[Vector[A]] =
     new Eq[Vector[A]] {
       def eqv(x: Vector[A], y: Vector[A]): Boolean = {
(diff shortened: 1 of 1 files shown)


core/src/main/scala/cats/std/vector.scala
```diff
@@ -68,16 +68,11 @@ trait VectorInstances {
...
-            case Seq() => ys.isEmpty
-            case a +: xs =>
-              ys match {
-                case Seq() => false
-                case b +: ys => if (ev.neqv(a, b)) false else loop(xs, ys)
-              }
-          }
-        loop(x, y)
+        @tailrec def loop(from: Int): Boolean =
+          if(from == x.size) true
```
ceedubs: I admit to being completely ignorant when it comes to JVM optimizations. Is it worth storing `x.size` as a `val` instead of calling it on each iteration?

TomasMikula: Not that my benchmarking is any rigorous, but taking the average of 20 runs in a REPL of the above current state of this PR and this new code that doesn't call `x.size` repeatedly:

```scala
def eqv(x: Vector[A], y: Vector[A]): Boolean = {
  @tailrec def loop(to: Int): Boolean =
    if(to == -1) true
    else ev.eqv(x(to), y(to)) && loop(to - 1)

  (x.size == y.size) && loop(x.size - 1)
}
```

it appears this new version is on average faster by some 21ms (167.95ms vs 146.81ms).

Also, I haven't tried to tweak any optimization flags.

TomasMikula: A added that to this PR.

TomasMikula:
Travis failure doesn't seem to be related to the commit. On my machine, all tests pass.

adelbertc:
Restarting build

TomasMikula:
Thanks, now it has passed.

ceedubs:
:+1: thanks!

We'll want to make sure that this gets picked up in the #786 effort.

mpilquist:
:+1:

japgolly:
Is there a deliberate reason that a manual loop was used instead of `.corresponds` here? This should result in the same performance:
```scala
def eqv(x: Vector[A], y: Vector[A]): Boolean =
  (x.size == y.size) && x.corresponds(y)(ev.eqv)
```

TomasMikula:
On my side, no. I was just improving on the pre-existing `loop`.

japgolly:
Ah cool. `.corresponds` is a bit of a hidden gem. Anyway, I'd imagine this should be the slightest, slightest bit faster cos there's one index counter instead of two.

PR #883, opened 2016-02-16
Delegate to Traverse.sequence in Applicative.sequence

ceedubs:
The previous approach duplicated the implementation of `sequence`, which
is a slight code stink. More importantly though, this approach will
benefit from an optimized override implementation of `sequence` if the
`Traverse` instance has one.


julien-truffaut:
:+1:

adelbertc:
:+1:

Issue #885, opened 2016-02-16
OneAnd missing `map` method


pvillega:
`OneAnd` has utility methods like `exist` or `filter` but is missing `map` 

It should be added

pvillega:
Fixed by PR

PR #886, opened 2016-02-16
Add map method to OneAnd. Fix for #885

pvillega:
OneAnd was missing an instance of map. This commit tries to solve this.

### DIFF
--- a/core/src/main/scala/cats/data/OneAnd.scala
+++ b/core/src/main/scala/cats/data/OneAnd.scala
@@ -3,6 +3,7 @@ package data
 
 import scala.annotation.tailrec
 import scala.collection.mutable.ListBuffer
+import cats.std.list._
 
 /**
  * A data type which represents a single element (head) and some other
@@ -110,7 +111,7 @@ private[data] sealed trait OneAndInstances extends OneAndLowPriority2 {
     }
 
   implicit def oneAndSemigroup[F[_]: MonadCombine, A]: Semigroup[OneAnd[F, A]] =
-    oneAndSemigroupK.algebra
+    oneAndSemigroupK(MonadCombine[F]).algebra
 
   implicit def oneAndReducible[F[_]](implicit F: Foldable[F]): Reducible[OneAnd[F, ?]] =
     new NonEmptyReducible[OneAnd[F,?], F] {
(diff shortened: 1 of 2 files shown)


adelbertc:
:+1: - could you also replace the existing `map` definitions in the type class instances with the new `map`?

pvillega:
will do :)

pvillega:
Done! I'm not sure about the `val functorList: Functor[List]` I had to define for the `CoMonad` implicit. I couldn't see an existing alternative but there may be one, if so please let me know and I'll correct it tomorrow (UK time).

adelbertc:
You probably added it because the type checker couldn't find a `Functor[List]` in implicit scope - you can get it from `import cats.std.list._`.

core/src/main/scala/cats/data/OneAnd.scala
```diff
@@ -131,9 +137,10 @@ private[data] sealed trait OneAndInstances extends OneAndLowPriority2 {
 }
 
 trait OneAndLowPriority0 {
+  import cats.std.list._
```
adelbertc: Nitpicking: would prefer this to be moved to the top

adelbertc:
One minor comment, but :+1: on Travis green

adelbertc:
Hm seems moving it to the top does [funky things](https://gitter.im/typelevel/cats?at=56c3acae5fa70a130ea500bc) - anyone know why this is the case? Either way, :+1: from me

pvillega:
Fixed compilation issue when moving import to top, thanks to @julien-truffaut for his support (he was the one fixing it in fact, I'm just stealing the contribution :P)

core/src/main/scala/cats/data/OneAnd.scala
```diff
@@ -104,7 +111,7 @@ private[data] sealed trait OneAndInstances extends OneAndLowPriority2 {
     }
 
   implicit def oneAndSemigroup[F[_]: MonadCombine, A]: Semigroup[OneAnd[F, A]] =
-    oneAndSemigroupK.algebra
+    oneAndSemigroupK(MonadCombine[F]).algebra
```
ceedubs: Sorry, this is nitpicky, but could you please change this to `oneAndSemigroupK[F].algebra`? That will save a `MonadCombine.apply` call.

julien-truffaut: for some reason the compiler get confused with `oneAndSemigroupK[F]` after adding `import cats.std.list._` on the top of the file. It says that there is an ambiguity between `MonadCombine[List]` and `MonadCombine[F]` ...

julien-truffaut: sorry I misread your comment @ceedubs, you were right `oneAndSemigroupK[F]` is enough to disambiguate.

julien-truffaut:
:+1: when the build is green

PR #894, opened 2016-02-23
add Task

stew:
So I think this contains a working Task implementation. There are some tests, they pass on the JVM, but we are using java.util.concurrent, which won't work on JS, so althogh validateJVM passes, validateJS will not pass.

So clearly this isn't ready for merge.

I don't think we currently have a way to have JVM only stuff currently, right?


stew:
aha, ok, I hadn't seen the jvm sub-module. I'm moving Task there now...

stew:
still left to do:

- [ ] eq instance (which we will need for ...:)
- [ ] test monad laws
- [ ] semigroup instance
- [ ] have a debate about whether or not it is a comonad (I vote no)

jvm/src/main/scala/cats/jvm/Task.scala
```diff
@@ -0,0 +1,166 @@
...
+   * thrown by the running of the computation.
+   */
+  def catchOnly[T >: Null <: Throwable: ClassTag]: Task[T Xor A] =
+    new AttemptOnly[T, A](this)
+
+
+  /**
+   * Run the computation to produce an A
+   */
+  final def unsafePerformIO(): A = eval.value
```
alexandru: This doesn't work in Javascript, because `Task` is inherently asynchronous and you can't get an `A` from it without potentially blocking threads.

jvm/src/main/scala/cats/jvm/Task.scala
```diff
@@ -0,0 +1,166 @@
...
+
+  /**
+   * Run this computation asynchronously, and call the callback with the result
+   */
+  final def unsafeAsyncIO(cb: A => Unit)(implicit pool: ExecutorService): Unit = {
```
alexandru: In Scala.js we don't have the `ExecutorService` interface. What you have is Scala's own `ExecutionContext`. But it cannot execute tasks with a delay (like Java's `ScheduledExecutorService`). This is why in Monix we have the [Scheduler](https://github.com/monixio/monix/blob/master/monix-execution/shared/src/main/scala/monix/execution/Scheduler.scala).

jvm/src/main/scala/cats/jvm/Task.scala
alexandru: Just as a note, this also goes for some of the Cats type classes, like `Foldable`, which extends by inheritance to `Traversable`.

jvm/src/main/scala/cats/jvm/Task.scala
adelbertc: These are in the JVM specific subproject

jvm/src/main/scala/cats/jvm/Task.scala
```diff
@@ -0,0 +1,166 @@
...
+      t
+    }
+  }
+
+  /**
+   * The default executor service is a fixed thread pool with N daemon threads,
+   * where N is equal to the twice number of available processors.
+   */
+  def createFixedExecutor(numThreads: Int): ExecutorService = {
+    Executors.newFixedThreadPool(numThreads, DaemonThreadFactory)
```
alexandru: The looks like the default thread-pool for `Task`, also used in Scalaz. I think it's a poor choice. Scala's own default is a ForkJoinPool that is sensitive to the [BlockContext](http://www.scala-lang.org/api/current/#scala.concurrent.BlockContext), Scala's baked-in mechanism for dealing with blocking calls in combination with limited thread-pools.

adelbertc:
@alexandru these are in the JVM specific subproject

jvm/src/main/scala/cats/jvm/Task.scala
```diff
@@ -0,0 +1,166 @@
...
+  private[cats] final case class AttemptOnly[T >: Null <: Throwable : ClassTag, A](task: Task[A]) extends Task[T Xor A] {
+    override def eval: Eval[Xor[T, A]] =
+      Eval.always(Xor.catchOnly[T](task.eval.value))
+  }
+
+  private[cats] final case class Async[A](complete: (A => Unit) => Unit) extends Task[A] {
+    override def eval: Eval[A] = {
+      val cdl = new CountDownLatch(1)
+      var result: Option[A] = None
+      complete((a: A) => {result = Some(a); cdl.countDown})
```
alexandru: `CountDownLatch` won't work in Javascript and besides, it's usage is a red flag on the JVM as well. There's absolutely no reason to block a thread in order to get an asynchronous value out of a `Task`.

jvm/src/test/scala/cats/tests/TaskSpec.scala
```diff
@@ -0,0 +1,159 @@
...
+    run should be (1)
+  }
+
+  test("Task should only evaulate side-effects once for Task.later") {
+    var run: Int = 0
+    val t: Task[Int] = Task.later { run += 1; run }
+    t.unsafePerformIO()
+    run should be(1)
+
+    t.unsafePerformIO()
```
alexandru: These tests won't work on Javascript. For one because you're using the synchronous `unsafePerformIO`. The advantage of building your own `Scheduler` is that you can then have a [TestScheduler](https://github.com/monixio/monix/blob/master/monix-execution/shared/src/main/scala/monix/execution/schedulers/TestScheduler.scala) that can simulate time. So you can do something like ...

```scala
val s = TestScheduler()
// ...

s.tick() // execute immediate tasks
s.tick(1.minute) // simulate the passage of time
```

alexandru: As an alternative to simulating asynchronous execution and time (which is much cleaner), on top of Scala.js we can also have real asynchronous tests, which my own [Minitest](https://github.com/monixio/minitest) supports.

jvm/src/main/scala/cats/jvm/Task.scala
alexandru: Another note on `unsafeAsyncIO`. Its signature is broken because there's no way to get a potential error out. It should be ...

```scala
final def unsafeAsyncIO(cb: Throwable Xor A => Unit) = ???
```

jvm/src/main/scala/cats/jvm/Task.scala
stew: yeah, totally valid point. How about I just avoid providing these and we can let the users just "new ForkJoinPool"

jvm/src/main/scala/cats/jvm/Task.scala
stew: if you need to cats errors, you can use `task.catchNonFatal.unsafeAsyncIO` instead of `task.unsafeAsyncIO`

jvm/src/main/scala/cats/jvm/Task.scala
stew: How do you suggest we wait for the value to arrive?

stew: just so we are on the same page, I'm going to want to be able to write something like:

   def FutureToTask[A](fa: Future[A]): Task[Throwable Xor A] = 
       Task.async { cb =>
            fa.onSuccess { 
               case Success(a) => cb(Xor.right(a))
               case Failure(t) => cb(Xor.left(t))
        }

jvm/src/main/scala/cats/jvm/EvalAsync.scala
```diff
@@ -0,0 +1,41 @@
...
+
+import data.Reader
+import java.util.concurrent.{Callable, CountDownLatch, ExecutorService}
+
+object EvalAsync {
+  def apply[A](cb: (A => Unit) => Unit): Eval[A] = Eval.always {
+    val cdl = new CountDownLatch(1)
+    var result: Option[A] = None
+    cb((a: A) => {result = Some(a); cdl.countDown})
+    cdl.await
```
alexandru: This is not asynchronous and it's not equivalent to a `Task`, because this is cheating :-)

I think Cats should strive to expose the same API for both the JVM and Scala.js. For one because it would allow people to write portable code without worrying about which types and functions are available where. But also because it keeps you honest.

When you do a `countDownLatch.await`, in essence blocking a thread for a result, the JVM pretends that the processing you're doing is synchronous, with a result available immediately. But it's a trick that can backfire. Ever heard of EJB (Enterprise Java Beans)? Ever heard of Corba? THE reason for why these sucked so badly was because their original design was a game of pretend. As in, lets pretend that where the processing happens for these synchronous methods don't matter. But of course, making local and remote method calls look the same ended up being a problem, because in the real world you can have significant latency, bandwidth limitations, security issues, dropped packages, timing issues and all sorts of other networking problems.

But back to the point, blocking threads is dangerous and in order to block safely, you have to know the configuration of the underlying thread-pool. Lets say that our thread-pool is limited to a single thread. Lets say that our `countDownLatch.await` will block that single thread. At that point there will be no threads left to do the actual processing. A one-thread pool is simplifying the problem of course, but this can happen with *any fixed sized thread-pool*. 

This is why we have a [BlockContext](https://github.com/alexandru/scala-best-practices/blob/master/sections/4-concurrency-parallelism.md#44-must-use-scalas-blockcontext-on-blocking-io) in Scala, being able to *tell the thread-pool* that this and that operations are potentially blocking so you might want to add more threads or whatever. Such a strategy is acceptable for blocking I/O. But this is very inefficient, as it defeats the purpose of using a limited thread-pool, so when doing a lot of synchronous I/O (like for example JDBC stuff), we might as well use a `CachedThreadPool` with unlimited threads.

Which leads us to the reason for why we want to use limited thread pools (e.g. with a number of threads directly proportional to the number of CPU cores). It's because preemptive multi-threading of more threads than CPU-cores is done by a technique called [time slicing](https://en.wikipedia.org/wiki/Super-threading) and when a CPU core switches between threads/processes it has to do a [context switch](https://en.wikipedia.org/wiki/Context_switch) and context switches are freaking expensive. This is why it's considered a best practice to do async I/O and to do CPU-bound processing on a limited thread-pool and if you have to do synchronous I/O, then people instantiate a second unlimited thread-pool meant only for synchronous I/O.

In other words, on top of the JVM, we have 3 options:

1. thread-block a limited thread-pool and suffer from non-deterministic *deadlocks*
2. thread-block an unlimited thread-pool and suffer the performance penalties
3. do not block any threads, ever

So you see, from my point of view, this isn't even an issue related to Javascript :-)

alexandru: TL;DR - sorry for the long message - what I'm saying is that if the design involves any form of blocking threads (like usage of a `CountDownLatch`), for a foundational library like Cats I think the design is wrong.

jvm/src/main/scala/cats/jvm/EvalAsync.scala
```diff
@@ -0,0 +1,41 @@
...
+    result.get // YOLO
+  }
+
+  implicit class EvalFork[A](val eval: Eval[A]) extends AnyVal {
+    def callable(cb: A => Unit): Callable[Unit] = new Callable[Unit] {
```
travisbrown: At a glance I have no idea why you're using `Callable` instead of `Runnable` here. Do you actually need the ability to throw checked exceptions?

jvm/src/main/scala/cats/jvm/EvalAsync.scala
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * Returns an Eval that will produce the same value as the wrapped
+     * Eval, but extracting the value from the resulting Eval will
+     * submit the work to the given ExecutorService
+     */
+    def fork: Reader[ExecutorService, Eval[A]] =
+      Reader { pool =>
+        EvalAsync { cb =>
+          val _ = pool.submit(callable(cb))
```
travisbrown: What does the `val _ =` accomplish?

travisbrown:
:+1: to @alexandru's critique (although personallly I don't really care about Scala.js for its own sake). More generally I'm still not at all convinced that any version of a `Task`-like type should be in cats-core, though (or possibly in cats at all).

I want to be able to write code that's generic up to something like `MonadError` and be confident that it will work for whatever abstraction I choose to use to model asynchronous computation. To take just one example, the core module of [iteratee.io](https://github.com/travisbrown/iteratee) will never refer to any specific implementation of `Future`, `Task`, or whatever.

jvm/src/main/scala/cats/jvm/EvalAsync.scala
stew: to get rid of a fatal warning about discarding a non-unit value

travisbrown: Hmm, `val _ =` is both more line noise (at least to my eye) than `; ()` and more bytecode (there actually is a local variable being assigned to).

travisbrown:
What I'd really like to see is a set of requirements for a `Task` type that everyone more or less agrees on, and then some experimentation with implementations until we have a better sense of which requirements (if any) need to be relaxed.

@alexandru's suggestion that we should aim for parity for Scala.js (largely for non-Scala.js-related reasons) seems like a good starting point. I'd also like to see better thread affinity than the standard library's futures, no implicit scheduling stuff floating around, close attention to allocations, opt-in-only exception catching, cancellation (maybe), etc.

I've got at least three of my own implementations of something like this—one a much cleaned-up port of Scalaz's `Task` and a couple of attempts at a little free-like thing that wraps `CompletableFuture` and a bunch of NIO.2 stuff. None of these are as mature or well-thought-through as @alexandru's `Task`, but they make different trade-offs that reflect my personal preferences.

Ideally at some point in the near-ish future we'd round up all these proposals and implementations and put together some kind of matrix of the features we care about (and some benchmarks, etc.). Something built on `Eval` might belong on that list, but I don't think it (or any of the other attempts) should be considered candidates for inclusion in cats for a long, long time, if ever.

adelbertc:
@travisbrown I think somewhere we need at least an `IO` - one thing I really like to use in Scalaz is `SafeApp` or `TaskApp`

```scala
abstract class SafeApp {
  def run(args: Vector[String]): IO[Unit]

  final def main(args: Array[String]): Unit =
    run(args.toVector).unsafePerformIO()
}
```

If we agree that `IO` would be useful, the question becomes do we want to just have a plain `IO` or do we want `Task`? If the former, would a `Task` implementation interpret into `IO` ? Why not just have `Task` be our `IO` ? etc. etc.

mpilquist:
I've mentioned it before in chat, though I suppose I should record it here. I agree with @travisbrown's position. 

It isn't clear to me that cats needs `IO`/`Task` when there are implementations available that are easy to integrate. I use FS2's `Task` quite happily and it only requires a single instance to bridge to cats.

julien-truffaut:
I would also prefer to see `IO` or `Task` defined as a separate project.

PR #905, opened 2016-03-01
make Unapply serializable

kailuowang:
This will enable typeclasses that uses Unapply to be serializable themselves.


ceedubs:
Thanks, @kailuowang. Could you please add a test (probably something like https://github.com/typelevel/cats/blob/9780e8ff36c69c959ee4acc2b2bb74a55d802995/tests/src/test/scala/cats/tests/FunctionTests.scala#L19)?

kailuowang:
@ceedubs test added.

adelbertc:
:+1: on Travis green

ceedubs:
:+1: thanks!

PR #906, opened 2016-03-02
Add BiProd, fixes #874

### LINKED ISSUE #874: Rethinking cats.data.Prod
For product composition of type classes we are using `cats.data.Prod` which is defined like

```scala
sealed trait Prod[F[_], G[_], A]
```

which fixes `F` and `G` to be unary type constructors. This works fine when defining product composition for type classes abstracting over unary type constructors like `Functor` and `Apply`. But for type classes like `Bifunctor` and `Bifoldable` which abstract over binary type constructors, they also have product composition but they can't use `cats.data.Prod`. We can just have the `compose` on those type classes return the tuple directly `Lambda[(A, B) => (F[A, B], G[A, B])]`, but that would be inconsistent with our use of `Prod` in the other type classes.

I looked in the [Essence of the Iterator Pattern](https://www.cs.ox.ac.uk/jeremy.gibbons/publications/iterator.pdf) paper which I believe inspired `cats.data.Prod` and dont' see them using it with the `Bifoldable`-y stuff.

/cc @eed3si9n since he added `Prod`

adelbertc:


ceedubs:
@adelbertc could you provide some example use-cases? If we decide to go forward with this, I'd like to see some more documentation.

adelbertc:
@ceedubs I confess I haven't had a use for this myself. I've definitely had uses for `Prod`, especially in the context of evaluating algebras. The motivation behind this was primarily for consistency in the sense that there is product composition for `Functor`, but none for `Bifunctor`.

PR #907, opened 2016-03-02
Add cats.io and IO monad

adelbertc:
This PR serves two purposes

1. Add an IO monad in a separate package
2. Get a discussion going around how Cats wants to approach IO/Task. It seems there is a mismatch in the nature of async on JVM and JS and it's tricky to get a Task with the same API on both JVM and JS. Furthermore, many folks will be using Task for FS2, and perhaps others the Task from Scalaz. There is also a relatively simple formulation of Task in the form of a thin wrapper around Eval [in dogs](https://github.com/stew/dogs/blob/64451b9083f78d3148c738d806ef0bdc07b2b2d2/core/src/main/scala/Task.scala), but that works on the JVM only. There's [another](https://github.com/scalaz/scalaz/compare/series/7.3.x...puffnfresh:feature/concurrent-io) that builds on Task using IO and MVars... I'm sure there are other approaches I don't know about.

I am proposing the possibility of Cats just having a simple IO and letting users of Cats decide where they want to get their Task from and how much cross-platform support they want.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -95,22 +95,22 @@ lazy val cats = project.in(file("."))
   .settings(moduleName := "root")
   .settings(catsSettings)
   .settings(noPublishSettings)
-  .aggregate(catsJVM, catsJS)
-  .dependsOn(catsJVM, catsJS, testsJVM % "test-internal -> test", bench % "compile-internal;test-internal -> test")
+  .aggregate(catsJVM, catsJS, ioJVM, ioJS)
+  .dependsOn(catsJVM, catsJS, ioJVM, ioJS, testsJVM % "test-internal -> test", bench % "compile-internal;test-internal -> test")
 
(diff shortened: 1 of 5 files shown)


build.sbt
```diff
@@ -139,6 +139,17 @@ lazy val core = crossProject.crossType(CrossType.Pure)
...
+  .dependsOn(core, tests % "test-internal -> test")
+  .settings(moduleName := "cats-io")
+  .settings(catsSettings: _*)
+  .jsSettings(commonJsSettings: _*)
+  .jsSettings()
+  .jvmSettings(commonJvmSettings: _*)
+
+lazy val ioJVM = io.jvm
+lazy val ioJS = io.js
+
```
DavidGregory084: I think the new module needs adding to the command aliases (`buildJVM` / `validateJVM`) below & `scripts/travis_publish.sh`

PR #910, opened 2016-03-03
Remove Streaming and StreamingT

### LINKED ISSUE #858: Move Streaming(T) to a separate module
I'm opening this issue for discussion, but if there's strong disagreement we can close it right away.

I think it'd make sense to move `Streaming` and `StreamingT` to a separate cats-streaming module (or 
possibly even a separate project).

These two types currently make up about a sixth of the cats-core codebase (~1,450 lines out of just under 9,000). They're only referred to elsewhere in cats-core to implement `toStreaming` methods that are redundant anyway, since `Streaming` has `fromFoldable`, `fromList`, `fromVector`, etc. Most importantly, these two types are the target of a majority of recent bug fixes and are responsible for a large number of open issues and PRs (most recently #856).

I haven't used either yet (and rarely used `StreamT` in Scalaz). If we care about modularity this seems like an easy way to split a big chunk of less-stable, less-essential code out into a separate module.

ceedubs:
This resolves #858 and is based on a discussion at dinner after the first day of the typelevel summit including @travisbrown, @non, @stew, and several others.

Per the discussion, Streaming will be moved to the dogs project.
StreamingT will be removed for now, and we can consider adding it to
dogs in the future if people find themselves wanting it.


stew:
:+1:

travisbrown:
:+1:

adelbertc:
:+1:

PR #911, opened 2016-03-03
Monad writer, fixes #889

### LINKED ISSUE #889: MonadWriter
https://github.com/purescript/purescript-transformers/blob/master/docs/Control/Monad/Writer/Class.md

adelbertc:
Other instances are possible: https://hackage.haskell.org/package/mtl-2.2.1/docs/Control-Monad-Writer-Class.html

But I realized we also don't have similar instances for the other MTL type classes, so I'll add it all in one fell swoop

### DIFF
--- a/core/src/main/scala/cats/MonadWriter.scala
+++ b/core/src/main/scala/cats/MonadWriter.scala
@@ -11,8 +11,16 @@ trait MonadWriter[F[_], W] extends Monad[F] {
   /** Apply the effectful function to the accumulator */
   def pass[A](fa: F[(A, W => W)]): F[A]
 
-  /** An effect that when run, logs w */
+  /** Lift the log into the effect */
   def tell(w: W): F[Unit] = writer(((), w))
+
+  /** Pair the value with an inspection of the accumulator */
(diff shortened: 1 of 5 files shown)


stew:
:+1:

ceedubs:
In several places this has `(A, W)`, but `(W, A)` seems like it would line up more nicely with our `WriterT`. Is there any particular reason you went with the former?

Do people have experience using `listen` and `listens`? I will admit to never using `MonadWriter`, but it seems to me like working with the tuples could be a bit of a hassle. I find myself wondering if it would often be simpler to discard the `A` component and let people use `flatMap` and then a method like these if they want a reference to the `A` component. Just a brain dump: I'm curious about the experience of people who have actually used `MonadWriter`.

adelbertc:
So I copied it from https://github.com/purescript/purescript-transformers/blob/master/docs/Control/Monad/Writer/Class.md - I think it uses tuples instead of separate args because the underlying representation for `Writer` is a tuple.

non:
Looks good to me. 👍

PR #915, opened 2016-03-04
Added support for F[A, ?] to kleisli instances

kailuowang:
Also moved MonadReader definition to higher priority. Placing it in the lowest priority causes implicit ambiguity.


kailuowang:
This issue would be addressed by this [SI-2712 fix](https://github.com/scala/scala/pull/5102)

PR #916, opened 2016-03-04
add -P:scalajs:mapSourceURI option

xuwei-k:


milessabin:
Does this change the release tag name?

xuwei-k:
> Does this change the release tag name?

No, this is `sbt-release` plugin's default `TaskKey[String]` value.
 
https://github.com/sbt/sbt-release/blob/v1.0.3/src/main/scala/ReleasePlugin.scala#L194

milessabin:
Is there any way of using the release plugin's default value without duplicating the definition here?

xuwei-k:
`releaseTagName` is `Task` but `scalacOptions` is `Setting`, `Setting` can not depends on `Task`.

milessabin:
Gotcha.

milessabin:
:+1:

non:
Thanks! :+1:

Issue #917, opened 2016-03-05
TransLift expressiveness is limited


djspiewak:
As discussed with @stew, capturing the type constructor to be lifted in the enclosing typeclass is neat in that it does allow the constraint to vary, but it also causes some problems at the call site.  It also unfortunately makes it impossible for cases like shims, since I cannot abstract over *both* `TransLift` and scalaz's `Hoist`, and I'm fairly certain even `TransLift` alone would not be usable for emm.

As a sidebar, I suspect that the insertion of the type constructor into the typeclass also gives rise to problems related to SI-2712 for users writing functions polymorphic in that effect type; problems which wouldn't occur with the type attached to `liftM`.

We proposed adjusting the type of `TransLift` to something more like the following signature:

```scala
trait TransLift[MT[_[_], _]] {
  type TC[M[_]]
  def liftT[M[_], A](ma: M[A])(implicit M: TC[M]): MT[M,A]
}
```

This would retain the power of the variable constraint by binding that power to the transformer itself, not the pair of the transformer and the inner effect.  @ceedubs noted that he had tried this approach in the past and ran into some problems with `Unapply` cases, but I think it's worth exploring a bit further.

ceedubs:
I started working on cleaning up my stash of this approach that I had. I remembered that a strange thing about the proposed change is that for types that don't require any constraint to `liftT` (like `Kleisli`), I think you are forced to write a dummy typeclass for which there is a trivial implicit instance.

djspiewak:
> I started working on cleaning up my stash of this approach that I had. I remembered that a strange thing about the proposed change is that for types that don't require any constraint to liftT (like Kleisli), I think you are forced to write a dummy typeclass for which there is a trivial implicit instance.

You can do that, or you can cheat with something ugly like I did and set the typeclass to `Lambda[X[_] => Unit =:= Unit]`, which will obviously trivially resolve.

adelbertc:
> set the typeclass to `Lambda[X[_] => Unit =:= Unit]`

*inches away*

ceedubs:
@djspiewak clever. I like it.

adelbertc:
@djspiewak can we call this closed?

djspiewak:
Yep

PR #918, opened 2016-03-05
First stab at a more flexible TransLift

djspiewak:
This is a review PR *only*!  Please do not merge yet.  I have some things I'd like to clean up, but I wanted to get some thoughts.

First off, this is an attempted fix for #917, which is a blocker for [emm](https://github.com/djspiewak/emm) and [shims](https://github.com/djspiewak/shims).  It rearranges `TransLift` so that a single instance is no longer fixed to one inner effect type while keeping the flexibility of *not* over-constraining said effect.  For example, the `TransLift` instance for `Kleisli` does not constrain the effect in any way, while the `OptionT` instance requires a `Functor` and `StateT` requires `Applicative`.

Second off…  I think I hit a few compiler bugs (`def` vs `val` is apparently significant in implicit resolution?!?!) and also may have made the test compilation time even longer.  Sorry…

The magic is in **TransLift.scala** and **transLift.scala** (the latter is the syntax class and contains some new implicit machinery)  I'm very open to nicer ways of doing this, btw, I just don't want to add a second type parameter to the `liftT` syntax.

BTW, @stew, does this mean I win the race?  ;-P

### DIFF
--- a/core/src/main/scala/cats/data/OptionT.scala
+++ b/core/src/main/scala/cats/data/OptionT.scala
@@ -129,9 +129,12 @@ private[data] sealed trait OptionTInstances1 {
         fa.map(f)
     }
 
-  implicit def optionTTransLift[M[_]: Functor]: TransLift[OptionT, M] =
-    new TransLift[OptionT, M] {
-      def liftT[A](ma: M[A]): OptionT[M, A] = OptionT.liftF(ma)
+  // do NOT change this to val! I know it looks like it should work, and really I agree, but it doesn't (for... reasons)
+  implicit def optionTTransLift: TransLift.Aux[OptionT, Functor] =
+    new TransLift[OptionT] {
+      type TC[M[_]] = Functor[M]
+
+      def liftT[M[_]: Functor, A](ma: M[A]): OptionT[M, A] = OptionT.liftF(ma)
     }
 }
 
--- a/core/src/main/scala/cats/data/StateT.scala
+++ b/core/src/main/scala/cats/data/StateT.scala
@@ -143,9 +143,11 @@ private[data] sealed abstract class StateTInstances {
         fa.map(f)
     }
 
-  implicit def stateTLift[M[_], S](implicit M: Applicative[M]): TransLift[({type λ[α[_], β] = StateT[α, S, β]})#λ, M] =
-    new TransLift[({type λ[α[_], β] = StateT[α, S, β]})#λ, M] {
-      def liftT[A](ma: M[A]): StateT[M, S, A] = StateT(s => M.map(ma)(s -> _))
+  implicit def stateTLift[S]: TransLift.Aux[StateT[?[_], S, ?], Applicative] =
+    new TransLift[StateT[?[_], S, ?]] {
+      type TC[M[_]] = Applicative[M]
+
+      def liftT[M[_]: Applicative, A](ma: M[A]): StateT[M, S, A] = StateT(s => Applicative[M].map(ma)(s -> _))
     }
 
 }
--- a/core/src/main/scala/cats/syntax/transLift.scala
+++ b/core/src/main/scala/cats/syntax/transLift.scala
@@ -6,5 +6,24 @@ trait TransLiftSyntax {
 }
 
 final class TransLiftOps[M[_], A](val ma: M[A]) extends AnyVal {
-  def liftT[MT[_[_],_]](implicit TL: TransLift[MT, M]): MT[M,A] = TL.liftT(ma)
+  def liftT[MT[_[_],_]](implicit extract: TLExtract[MT, M]): MT[M, A] = extract.TL.liftT(ma)(extract.TC)
+}
+
+trait TLExtract[MT[_[_], _], M[_]] {
+  val TL: TransLift[MT]
+  val TC: TL.TC[M]
+}
+
+object TLExtract {
+
+  implicit def extract1[MT[_[_], _], M[_], TC[_[_]]](implicit TL0: TransLift.Aux[MT, TC], TC0: TC[M]): TLExtract[MT, M] = new TLExtract[MT, M] {
+    val TL = TL0
+    val TC = TC0
+  }
+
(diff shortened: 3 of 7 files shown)


adelbertc:
Must resist urge to merge just so it's in before stew even gets a chance to see it.......

core/src/main/scala/cats/data/OptionT.scala
```diff
@@ -129,9 +129,12 @@ private[data] sealed trait OptionTInstances1 {
         fa.map(f)
     }
 
-  implicit def optionTTransLift[M[_]: Functor]: TransLift[OptionT, M] =
-    new TransLift[OptionT, M] {
-      def liftT[A](ma: M[A]): OptionT[M, A] = OptionT.liftF(ma)
+  // do NOT change this to val! I know it looks like it should work, and really I agree, but it doesn't (for... reasons)
+  implicit def optionTTransLift: TransLift.Aux[OptionT, Functor] =
```
ceedubs: That's really bizarre. In what way does it fail if it's a `val`?

core/src/main/scala/cats/data/StateT.scala
```diff
@@ -143,9 +143,11 @@ private[data] sealed abstract class StateTInstances {
...
     }
 
-  implicit def stateTLift[M[_], S](implicit M: Applicative[M]): TransLift[({type λ[α[_], β] = StateT[α, S, β]})#λ, M] =
-    new TransLift[({type λ[α[_], β] = StateT[α, S, β]})#λ, M] {
-      def liftT[A](ma: M[A]): StateT[M, S, A] = StateT(s => M.map(ma)(s -> _))
+  implicit def stateTLift[S]: TransLift.Aux[StateT[?[_], S, ?], Applicative] =
+    new TransLift[StateT[?[_], S, ?]] {
+      type TC[M[_]] = Applicative[M]
+
+      def liftT[M[_]: Applicative, A](ma: M[A]): StateT[M, S, A] = StateT(s => Applicative[M].map(ma)(s -> _))
```
ceedubs: It's probably worth having a `StateT.liftA` (or whatever it will be called) that this delegates through to. I know this is a bit tangential to your PR, but I'm writing it down while it's on my mind.

core/src/main/scala/cats/syntax/transLift.scala
```diff
@@ -6,5 +6,24 @@ trait TransLiftSyntax {
...
+
+  implicit def extract1[MT[_[_], _], M[_], TC[_[_]]](implicit TL0: TransLift.Aux[MT, TC], TC0: TC[M]): TLExtract[MT, M] = new TLExtract[MT, M] {
+    val TL = TL0
+    val TC = TC0
+  }
+
+  implicit def extract1Id[MT[_[_], _], M[_]](implicit TL0: TransLift.Aux[MT, λ[X[_] => Unit =:= Unit]], TC0: Unit =:= Unit): TLExtract[MT, M] = extract1[MT, M, λ[X[_] => Unit =:= Unit]]
+
+  // sigh...
+  implicit def extract2[MT[_[_], _, _], Z, M[_], TC[_[_]]](implicit TL0: TransLift.Aux[MT[?[_], Z, ?], TC], TC0: TC[M]): TLExtract[MT[?[_], Z, ?], M] = extract1[MT[?[_], Z, ?], M, TC]
```
ceedubs: I don't suppose there's any way we could get `Unapply` to help us here?

core/src/main/scala/cats/syntax/transLift.scala
```diff
@@ -6,5 +6,24 @@ trait TransLiftSyntax {
 }
 
 final class TransLiftOps[M[_], A](val ma: M[A]) extends AnyVal {
-  def liftT[MT[_[_],_]](implicit TL: TransLift[MT, M]): MT[M,A] = TL.liftT(ma)
+  def liftT[MT[_[_],_]](implicit extract: TLExtract[MT, M]): MT[M, A] = extract.TL.liftT(ma)(extract.TC)
 }
+
+trait TLExtract[MT[_[_], _], M[_]] {
+  val TL: TransLift[MT]
+  val TC: TL.TC[M]
```
ceedubs: Should these be `def`s instead of `val`s? I get uneasy about abstract `val`s in `trait`s.

ceedubs:
I think that this is a more flexible solution than the current implementation. I too would be quite happy to see someone show a cleaner approach, but barring that, I support going forward with something like this.

I'm curious if we could ease the case of lifting a `F[_, _]` with some helper methods for common cases like `Applicative`, `Functor`, and `Monad` with specialized syntax for those (probably based on `Unapply`).

core/src/main/scala/cats/data/OptionT.scala
djspiewak: It just fails to resolve for some reason.  I didn't look at `-Xlog-implicits`, but it seemed like it was simply not considering the possibility.  Passing it explicitly worked fine, and obviously changing it to a def is also fine.

core/src/main/scala/cats/syntax/transLift.scala
djspiewak: I get uneasy too.  :-)  In this case, it's perfectly safe since we're using the trait as a pure interface (no definitions) and we aren't mixing it into anything, so there are no eval orders to get screwed up.

core/src/main/scala/cats/syntax/transLift.scala
djspiewak: Not as such, no.  I don't think `Unapply` deals with the higher order case (`MT[_[_], _]`), which is what we're extracting.  Additionally, we would need to bind the results of the unapply to a type variable.  That might be possible using @milessabin's "project on a variable which is the subtype of a structural type" trick, but it would take a bit more machinery.

ceedubs:
Anyone else have thoughts on this? @stew @tpolecat?

stew:
This LGTM. This is essentially the path I had started heading down, but yeah, djspeiwak wins the race :)

djspiewak:
Alrighty, if the consensus is that this isn't the worst idea in the world, I'll clean it up and push a full PR.

adelbertc:
:+1:

djspiewak:
/me slaps @adelbertc for not using the github reaction feature

^_^

djspiewak:
Closing this PR so I can open a new one with clear intent to merge…

PR #921, opened 2016-03-08
Polynomial functor fixed points with documentation

LucDupAtGitHub:


docs/src/main/tut/poly.md
```diff
@@ -0,0 +1,229 @@
...
+  def showFixList[Y]: FixList[Y] => String =
+    _.cata[String](showListAlgebra)
+
+  def incrIntFixList: FixList[Int] => FixList[Int] =
+    _.cata[ FixList[Int]](incrIntListAlgebra)
+
+```
+
+and we can use all we defined as follows:
+evaluating `showFixList[Int](incrIntFixList(cons[Int].apply(1, cons[Int].apply(2, nil[Int].apply(())))))`
```
DavidGregory084: It may be clearer here not to call `apply` explicitly, i.e. `showFixList[Int](incrIntFixList(cons[Int](1, cons[Int](2, nil[Int](())))))`

docs/src/main/tut/fix.md
```diff
@@ -0,0 +1,199 @@
+---
+layout: default
+title:  "Fixed Points"
+section: "fix"
+source: "https://github.com/non/cats/blob/master/core/src/main/scala/cats/fix/Fix.scala"
+scaladoc: "#cats.fix.Fix"
+-------------------------
+# Fixed Points
+
+`Fix` uses the [`Functor`](functor.html) type class to define a *structural recursive* function `cata` over the least fixed point `Fix[F]` of a functor `F`.
```
DavidGregory084: Nitpick, but should this be *structurally recursive*?

docs/src/main/tut/fix.md
```diff
@@ -0,0 +1,199 @@
...
+
+    val badTryingOneTwoThreeInts: Try[List[Int]] =
+      traverse[Try, String, Int](badOneTwoThreeStrings) { s => `try`(Integer.parseInt(s)) }
+
+```
+
+## What about evaluation?
+
+As you may know, structural recursion may benefit from different evaluation strategies.
+Since `cata` works for *all* algebras, it suffices to replace `A` with one of the `Eval[]` case classes (`Now[A]`, `Later[A]` or `Always[A]`)
```
DavidGregory084: Nitpick again :smile: should this be `Eval[_]` or `Eval[A]`?

DavidGregory084:
This is really cool, I wonder what you think of https://github.com/slamdata/matryoshka & whether there are ideas from that library which you may find useful?

docs/src/main/tut/poly.md
LucDupAtGitHub: I'll fix that one

docs/src/main/tut/fix.md
LucDupAtGitHub: I'll fix that one
￼Add a line note

LucDupAtGitHub:
@DavidGregory084 thanks for the appreciation .. I will look at https://github.com/slamdata/matryoshka this evening

docs/src/main/tut/poly.md
LucDupAtGitHub: Hello, the `apply` is necessary because of the implicit functor argument (a somewhat unfortunate thing in Scala).
Maybe there is a way around this, I'll have a look

ceedubs:
@LucDupAtGitHub you definitely have some neat stuff here. Thanks especially for putting together docs. After looking at matryoshka (which I believe slamdata intends to become a typelevel library at some point), do you have thoughts on which sorts of things belong in Cats and which belong in a separate library dedicated to this sort of thing? cc @sellout

sellout:
Yep, Matryoshka is a Typelevel incubator project, just hasn't been added to the site yet.

This looks cool (and clearly some overlap with Matryoshka). I'll take a deeper look when I'm at a better device than my phone.

core/src/main/scala/cats/data/Prod.scala
```diff
@@ -9,7 +9,11 @@ package data
 sealed trait Prod[F[_], G[_], A] {
   def first: F[A]
   def second: G[A]
+
+  def bimap[B](f: F[A] => F[B], g: G[A] => G[B]): Prod[F, G, B] =
```
sellout: This should probably be a `Bifunctor` or even a `Bitraverse` instance.

sellout:
Ok, a couple thoughts. First of all, Matryoshka doesn’t support Cats yet 😕. It’s still a Scalaz library. I’d like to support both, but I’m not sure how (https://github.com/djspiewak/shims?).

I’m also not a big fan of `Fix`. If some of these types were to be in Cats directly, I’d prefer if it were `Mu` and `Nu`. But without the type classes, I don’t know if they’re much use.

But then I don’t know where that leaves the polynomial functors. I personally like tiny libraries, but that’s not that easy in the Scala world, it seems.

core/src/main/scala/cats/data/Prod.scala
LucDupAtGitHub: I will have a look at that one

LucDupAtGitHub:
about : "I’m also not a big fan of Fix. If some of these types were to be in Cats directly, I’d prefer if it were Mu and Nu."

Do you mean you do not like the name Fix or the concept?

gclaramunt:
Pretty cool! I've been trying to do the same (in a slightly different approach) with https://github.com/gclaramunt/scala-reggen
At first look, is not clear for me (I think is because of the type aliases), but is possible to define a generic "show" function that works for all the data-types, right?
(just pattern match in all the elements of the algebra: K, I, +, * )

easel:
This PR is pretty stale, but it looks like it has some worthwhile contributions I'd hate to lose. @LucDupAtGitHub are you in a position to bring this up-to-date?

larsrh:
I'm personally of the opinion that this should go into [turtles](https://github.com/sellout/turtles).

easel:
@larsrh that seems reasonable given the conversation already had. Unless somebody objects I'd say we should close this with the note 'please submit against turtles'.

larsrh:
@LucDupAtGitHub Thanks for your contribution. Consensus appears to be that this would rather belong to the `turtles` library.

ceedubs:
Thanks to all involved. I agree in closing this out. For anyone coming to this I just wanted to put in a reference to https://github.com/andyscott/droste as a potential alternative to turtles, since turtles seems to have stalled as a project.

PR #922, opened 2016-03-08
Add .tell and .writer syntax for creating Writers. Fixes #920.

### LINKED ISSUE #920: New constructors for `Writer`?
I'd like to propose adding direct support for constructing a `Writer` from a log, and ensuring that there are syntactically similar equivalents for creating from a log, a result, and a log and a result together.

There's currently support for creating `Writers` using the `pure` syntax and the `Writer.apply` method. However, any other constructors require you to know about `WriterT`.

One possibility would be to add a `tell` syntax like the one in Scalaz:

~~~ scala
scala> import scalaz._, Scalaz._

scala> Vector("Log message 1", "Log message 2").tell
// res1: Writer[Vector[String],Unit] = 
//   WriterT((Vector(Log message 1, Log message 2),()))
~~~

This would give us `pure` and `tell` to create from a result or a log, and `Writer.apply` to create from both values together.

One could also add an equivalent of Scalaz' `set` syntax, although IMO the name would have to change:

~~~ scala

scala> "Result".set(Vector("Log message 1", "Log message 2"))
// res2: Writer[Vector[String],String] =
//   WriterT((Vector(Log message 1, Log message 2),Result))
~~~

Another possibility would be adding `Writer.tell` and `Writer.put` / `Writer.value` to mirror the constructors on `WriterT`.

I realise this is duplicating functionality that's already there with `WriterT`, but I think there's a cleaner learning curve if people can learn about `Reader`, `Writer`, and `State` before they jump into monad transformers.

Keen to know peoples' thoughts!

davegurnell:
This adds the following syntax for the `Writer` monad, as outlined in #920:

~~~ scala
log.tell

result.writer(log)
~~~

This is my first PR to Cats so there's probably stuff that's wrong in here. I'm happy to fix stuff if required, or close the PR if people think it's not worthwhile. I figured it was easier to have a discussion over a concrete PR than an issue.

### DIFF
--- a/core/src/main/scala/cats/syntax/all.scala
+++ b/core/src/main/scala/cats/syntax/all.scala
@@ -37,3 +37,4 @@ trait AllSyntax
     with XorSyntax
     with ValidatedSyntax
     with CoproductSyntax
+    with WriterSyntax
--- a/core/src/main/scala/cats/syntax/package.scala
+++ b/core/src/main/scala/cats/syntax/package.scala
@@ -35,4 +35,5 @@ package object syntax {
   object traverse extends TraverseSyntax
   object xor extends XorSyntax
   object validated extends ValidatedSyntax
+  object writer extends WriterSyntax
 }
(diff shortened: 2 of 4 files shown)


non:
The syntax seems reasonable to me, as do the names. :+1:

stew:
:+1: Thanks Dave!

PR #923, opened 2016-03-10
Make Call.loop @tailrec optimized.

TomasMikula:
Ensure `Call.loop` is `@tailrec` optimized. Elegance of the code suffered a bit, but I guess it's worth it?

The fact that `@tailrec` could not be put on `loop` as it was actually seems like a Scala bug, because the second call to `loop` is not a recursive call.


ceedubs:
:+1:. Ensuring `@tailrec` optimization seems like a good enough reason to make this minor sacrifice in code cleanliness.

stew:
:+1:

Issue #926, opened 2016-03-10
XorT.ensure method


ImLiar:
Underlying `Xor` has `ensure` method which allows to check `B` value with predicate. It's useful to have this method in `XorT` too.

PR #927, opened 2016-03-10
XorT.ensure method

ImLiar:
#926

### DIFF
--- a/core/src/main/scala/cats/data/XorT.scala
+++ b/core/src/main/scala/cats/data/XorT.scala
@@ -51,6 +51,8 @@ final case class XorT[F[_], A, B](value: F[A Xor B]) {
 
   def exists(f: B => Boolean)(implicit F: Functor[F]): F[Boolean] = F.map(value)(_.exists(f))
 
+  def ensure[AA >: A](onFailure: => AA)(f: B => Boolean)(implicit F: Functor[F]): XorT[F, AA, B] = XorT(F.map(value)(_.ensure(onFailure)(f)))
+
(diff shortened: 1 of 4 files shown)


core/src/main/scala/cats/data/XorT.scala
```diff
@@ -51,6 +51,8 @@ final case class XorT[F[_], A, B](value: F[A Xor B]) {
 
   def exists(f: B => Boolean)(implicit F: Functor[F]): F[Boolean] = F.map(value)(_.exists(f))
 
+  def ensure[AA >: A](ifLeft: => AA)(f: B => Boolean)(implicit F: Functor[F]): XorT[F, AA, B] = XorT(F.map(value)(_.ensure(ifLeft)(f)))
```
ceedubs: The name `ifLeft` surprised me. This is the value that is used if the input is a `Right` but the predicate fails, right? I see that you are just following the name used in `Xor`, but I'm thinking we should probably change both of these to `ifFalse`, `whenFalse`, `predicateFailed` or something like that. What do you think?

ceedubs:
Thanks, @ImLiar! I like this. I left one minor comment and am interested in your thoughts on it.

core/src/main/scala/cats/data/XorT.scala
ImLiar: Well, in case of predicate falling the whole `XorT` "falls" to the left branch, it might be an explanation :)

But no problem, I could rename both.

ImLiar:
@ceedubs Done

ceedubs:
@ImLiar great, thanks! :+1:

Are people happy with `onFailure` for the parameter name?

stew:
:+1:

Issue #607, opened 2015-11-07
Add missing Show instances


ceedubs:
Many data types already have `Show` instances (and more will after #600, #601, and #602 are merged). However, there are still a few that are missing. Two that I see at the moment are ~~`Prod`~~ and `WriterT`. We should add those and any others we can find that are missing.

Edit: removed `Prod` from the scope of this issue per https://github.com/non/cats/issues/607#issuecomment-155074440.

mikejcurry:
I was going to add an instance for `Prod`; and learn the functionality at the same time, but it turned out to be a bit like peeling an onion. Less the tears part and more the layers part, but... :-)

Anyway, the thing with `Prod` specifically, and maybe this deserves its own issue, is that it is constructed to be lazy and it feels wrong for the `Show` instance to bypass this and compute the value. So, I figured an `Eval` instance for `Show` which respected the laziness of `Eval` might be useful.

Before I go too deep down this road, the question I have is whether a `Show` for `Eval` makes sense. E.g. for something like `Now` it would show _Now(value)_ and something like Later it could show the value if already computed and could show something like _Later(() => ResultType)_ if the value has not been computed.

The next layer of the onion is that there are `Show` instances for `BigInt` and `BigDecimal` but not for primitives such as `Int`, `Char` etc (as far as I could see) - does it make sense to add these also? 

My feeling is that all of these make sense at some level, and that `Show` is pretty much universally applicable as you may want to `Show` pretty much anything.

Let me know what you think.

ceedubs:
@mikejcurry I'm about to go somewhere, but there's one simple part of this that I can address first :). The `Show` instances for primitives are [here](https://github.com/non/cats/blob/master/core/src/main/scala/cats/std/anyval.scala).

mikejcurry:
Aah ok, I missed those for some reason. Sorry about that.

ceedubs:
@mikejcurry Thank you for brining up the rabbit hole you went down with `Prod`. I agree that it's awkward, and I've created #615 to address that. Let's drop `Prod` from the scope of this issue.

matt-martin:
I've tried to add a Show instance for WriterT with the following pull request: https://github.com/typelevel/cats/pull/928.  Just trying to get my feet wet here with some "low-hanging fruit" so I'd welcome any and all feedback.

rabbitonweb:
Since #928 is merged, shouldn't this issue be closed?

ceedubs:
[...]

Issue #631, opened 2015-11-12
Instances for tuples?


Centaur:
Did not find TupleInstances in `cats.std`. is this intentional or is it a feature to be implemented?

ceedubs:
The tuple instances for Algebra type classes (Monoid, Eq, etc) should be provided by https://github.com/non/algebra/pull/82 which has been merged but not yet published in a stable release.

For Cats-specific instances, this is probably just to-be-implemented. There's a question of whether we want to create instances for arbitrarily-sized tuples or just create instances for Tuple2 and leave the rest to a shapeless companion project (such as [kittens](https://github.com/milessabin/kittens)).

adelbertc:
I had concerns about the performance of the automatic derivation, but it seems the performance hit isn't [terrible, if any at all](https://github.com/travisbrown/circe/pull/96). 

That being said, I had added the Tuple codegen in non/algebra#82 in the name of performance, but I am now swaying towards deriving it. I feel it would be a bit strange/inconsistent to have codegen'ed Tuple instances in `algebra`, some hand-rolled in `cats`, and some delegated to `kittens`. Personally I would like to just have everything generically derived.. unsure how that fits with `algebra` goal of a small library though

/cc @non @johnynek

travisbrown:
For what it's worth, @adelbertc, I think I still prefer codegen for tuples and other non-user-defined types, not necessarily for the sake of performance, but for clarity of the API (and API docs), general simplicity, etc.

It's also worth noting that tuples need special-casing in generic derivation for some type classes—e.g. in circe the decoders we'd get from generic derivation for tuples-as-case-classes aren't generally the ones we want.

johnynek:
I'm personally pretty dependency phobic. An example for us is Kryo. This library is a dependency of Hive, Spark, Scalding, Storm and probably many others. We would like to upgrade, but there are binary incompatibility problems between releases, so this means if we upgrade in scalding, we can't mix that code with spark. Since the pain points aren't very great, we just let it sit there.

Such is the current pain of libraries on the jvm. Unfortunately, for core libraries that you expect (and hope) will have lots of adoption, it is much better that they have very few, if any, dependencies.

For my money, I'd rather use code-gen and macros, which don't introduce new dependencies than fix a new, even lower level dependency beneath algebra (which I guess would be shapeless in this example).

ceedubs:
[...]

PR #928, opened 2016-03-11
#607: Add missing Show instances

matt-martin:
Randomly picked up this issue as I saw it was marked as "low hanging fruit" and I wanted to get more familiar with cats (and associated FP concepts).

One thing of note: I had to add a show instance for Tuple2 to make this work.  I can't claim to be intimately familiar with Miles' Kittens project (see: https://github.com/milessabin/kittens), but it seemed like defining Show instances for Tuples and case classes would be particularly tedious and _possibly_ something that Kittens/Shapeless could make a lot easier.

Without further ado, here is my initial attempt to address issue #607 and add a Show instance for WriterT.

### DIFF
--- a/core/src/main/scala/cats/data/WriterT.scala
+++ b/core/src/main/scala/cats/data/WriterT.scala
@@ -44,6 +44,8 @@ final case class WriterT[F[_], L, V](run: F[(L, V)]) {
 
   def reset(implicit monoidL: Monoid[L], functorF: Functor[F]): WriterT[F, L, V] =
     mapWritten(_ => monoidL.empty)
+
+  def show(implicit F: Show[F[(L, V)]]): String = F.show(run)
 }
 object WriterT extends WriterTInstances with WriterTFunctions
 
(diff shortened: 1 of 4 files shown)


ceedubs:
Wonderful :+1:. Thanks @matt-martin!

See #631 for some discussion regarding the tuple instance generation that you brought up.

stew:
:+1:

matt-martin:
Thank you @stew and @ceedubs! Appreciate the reference to #631. Definitely interested in that discussion.

core/src/main/scala/cats/data/WriterT.scala
```diff
@@ -44,6 +44,8 @@ final case class WriterT[F[_], L, V](run: F[(L, V)]) {
 
   def reset(implicit monoidL: Monoid[L], functorF: Functor[F]): WriterT[F, L, V] =
     mapWritten(_ => monoidL.empty)
+
+  def show(implicit F: Show[F[(L, V)]]): String = F.show(run)
```
adelbertc: Hm we should be able to delete this line right?

ceedubs: @adelbertc isn't the `Show` instance delegating to this method?

matt-martin: Yeah, the `Show` instance is delegating to this method.  I modeled this off the change for OptionT (see: https://github.com/typelevel/cats/pull/600/files; and in particular: https://github.com/typelevel/cats/blob/6dbd4b04aff336867f563c9f6ec04fde828aa816/core/src/main/scala/cats/data/OptionT.scala#L90).

adelbertc: Oh I see now, I read it as `Show[(L, V)]` - ignore me :-)

PR #929, opened 2016-03-11
Adding more contravariant instances

DavidGregory084:
Some more work on #579.

### DIFF
--- a/core/src/main/scala/cats/data/WriterT.scala
+++ b/core/src/main/scala/cats/data/WriterT.scala
@@ -2,7 +2,7 @@ package cats
 package data
 
 import cats.kernel.std.tuple._
-import cats.functor.Bifunctor
+import cats.functor.{Bifunctor, Contravariant}
 
 final case class WriterT[F[_], L, V](run: F[(L, V)]) {
   def written(implicit functorF: Functor[F]): F[L] =
@@ -20,6 +20,11 @@ final case class WriterT[F[_], L, V](run: F[(L, V)]) {
   def map[Z](fn: V => Z)(implicit functorF: Functor[F]): WriterT[F, L, Z] =
     WriterT {
       functorF.map(run) { z => (z._1, fn(z._2)) }
+    }
+
+  def contramap[Z](fn: Z => V)(implicit F: Contravariant[F]): WriterT[F, L, Z] =
+    WriterT {
+      F.contramap(run) { z => (z._1, fn(z._2)) }
     }
 
   def flatMap[U](f: V => WriterT[F, L, U])(implicit flatMapF: FlatMap[F], semigroupL: Semigroup[L]): WriterT[F, L, U] =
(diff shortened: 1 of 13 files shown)


core/src/main/scala/cats/data/WriterT.scala
```diff
@@ -21,6 +21,11 @@ final case class WriterT[F[_], L, V](run: F[(L, V)]) {
       functorF.map(run) { z => (z._1, fn(z._2)) }
     }
 
+  def contramap[Z](fn: Z => V)(implicit F: Contravariant[F]): WriterT[F, L, Z] =
+    WriterT {
+      F.contramap(run) { z => (z._1, fn(z._2)) }
+    }
```
DavidGregory084: I wouldn't have thought this would be terribly useful (but I could be totally wrong), let me know what you think.

core/src/main/scala/cats/package.scala
```diff
@@ -58,4 +58,14 @@ package object cats {
...
+
+  implicit val partialOrderInstances: cats.functor.Contravariant[PartialOrder] =
+    new cats.functor.Contravariant[PartialOrder] {
+      def contramap[A, B](fa: PartialOrder[A])(f: B => A): PartialOrder[B] = fa.on(f)
+    }
+
+  implicit val orderInstances: cats.functor.Contravariant[Order] =
+    new cats.functor.Contravariant[Order] {
+      def contramap[A, B](fa: Order[A])(f: B => A): Order[B] = fa.on(f)
+    }
```
DavidGregory084: I'm not 100% sure where instances for types from Algebra belong - happy to move them somewhere a bit less ad-hoc

DavidGregory084:
I was able (I think) to write an implementation for StateT as:
```scala
def contramap[B](f: B => A)(implicit F: Contravariant[F], M: Monad[F]) =
  StateT(s1 =>
    F.contramap(run(s1)) {
      case (s2,  a) => (s2,  f(a)) 
    }
  ) 
```
However, this would typically mean inferring implicits for e.g `Contravariant[Cokleisli[Option, ?,  Int]]` and `Monad[Cokleisli[Option, Int, ?]]` at the same time -  I'm not really sure how to test that.

non:
Hi @DavidGregory084, sorry for the long wait!

These instances look great! Would you be willing to update this PR?

As far as instances for the kernel (e.g. contravariant functors for `Eq`, `PartialOrder`, and `Order`) I would actually recommend putting those in the `Contravariant` companion object (in `core`). I would also recommend adding a note that the resulting instances are only law-abiding if the functions provided are one-to-one (injections).

DavidGregory084:
No problem, this should be up to date now :smile_cat:

ceedubs:
Sorry - we've been merging some sweeping PRs recently that have caused a lot of merge conflicts :(. Would you mind updating again?

DavidGregory084:
@ceedubs Yeah, that's fine, should be up to date again :)

ceedubs:
👍 thanks a bunch, @DavidGregory084! @non would you be willing to do a review before this goes stale again?

non:
Looks great, thanks! 👍

Issue #934, opened 2016-03-16
Make Eval covariant in its type parameter


alexandru:
So I enthusiastically started using `Eval`, but then noticed that `Eval[A]` is not covariant in `A` and this creates problems for me.

Now I understand that variance is a hard and I understand why type-classes are invariant and so on, but `Eval` is supposed to be a replacement for by-name and for `Function0[+R]` parameters and the problem is that functions in Scala are covariant in their return, along with most immutable data-structures from the standard library. And so when you want covariance, then `Eval` ends up not being a replacement for by-name parameters.

My use-case is the ability to use `F[Nothing]` for empty containers when it comes to things like this:

```scala
trait Sequence[+A] {
  def continueWith[B >: A](other: => Sequence[B]): Sequence[B]
}
```

In the above, the ability to pass in a `Sequence[Nothing]` sure comes in handy. And replacing that parameter with an `Eval` doesn't work as well as it does with a by-name parameter.

Link to the Gitter conversation: https://gitter.im/typelevel/cats?at=56e95650c7364f7926bf7d38

non:
So, I'm cautiously :+1: on this proposal. I agree that the purpose of `Eval[A]` is to abstract over `=> A`, and so it should work for the same purposes. That said, there are definitely some contexts where variance causes real problems, so I think we should proceed a little bit carefully.

I tested a one-character change today (`Eval[A]` => `Eval[+A]`) and everything compiled, tests pass, etc. So at least as far as Cats goes this is not a big change.

non:
Also -- nothing is set in stone until 1.0, so we could always turn this on and keep an eye out for problems in the next few months.

milessabin:
:+1:

ceedubs:
Just to summarize what I've said on Gitter:

I'm okay either way on this one. Variance can be very problematic in some contexts (two that come to mind are on type classes and on abstract type constructor parameters). However, in this case, `sealed abstract class Eval[+A]` would look a lot like `sealed abstract class Option[+A]` or `sealed abstract class List[+A]` from a compiler perspective. While there are some minor annoyances with `List` and `Option` being covariant, I don't think it's ever caused serious issue, and there are some definite benefits.

Since there's been a lot of discussion around this, I propose that we keep this issue open for 2 weeks and reassess at that point. If people still seem to generally be in favor of this and nobody has provided a show-stopping example of how it could be problematic, let's go for it.

edmundnoble:
[...]

PR #937, opened 2016-03-18
Issue #934 - Make Eval covariant

alexandru:
The reason is that `Eval` should be a drop-in replacement for by-name parameters. Noticed that changing it wouldn't break anything within Cats.

### DIFF
--- a/core/src/main/scala/cats/Eval.scala
+++ b/core/src/main/scala/cats/Eval.scala
@@ -33,7 +33,7 @@ import cats.syntax.all._
  * Eval instance -- this can defeat the trampolining and lead to stack
  * overflows.
  */
-sealed abstract class Eval[A] extends Serializable { self =>
+sealed abstract class Eval[+A] extends Serializable { self =>
 
   /**
    * Evaluate the computation and return an A value.
@@ -279,7 +279,7 @@ object Eval extends EvalInstances {
                   cc.start().asInstanceOf[L],
                   cc.run.asInstanceOf[C] :: c.run.asInstanceOf[C] :: fs)
               case xx =>
-                loop(c.run(xx.value).asInstanceOf[L], fs)
+                loop(c.run(xx.value), fs)
             }
           case x =>
             fs match {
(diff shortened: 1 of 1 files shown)


core/src/main/scala/cats/Eval.scala
```diff
@@ -350,7 +350,7 @@ trait EvalMonoid[A] extends Monoid[Eval[A]] with EvalSemigroup[A] {
 trait EvalGroup[A] extends Group[Eval[A]] with EvalMonoid[A] {
   implicit def algebra: Group[A]
   def inverse(lx: Eval[A]): Eval[A] =
-    lx.map(_.inverse)
+    lx.map(_.inverse())
```
adelbertc: Why is this needed?

alexandru: It isn't needed, but the definition of that function in `cats.syntax.GroupOps` is:

```scala
def inverse(): A = macro Ops.unop[A]
```

Not sure why that's the definition, but when you've got a parens pair like that in your `def`, then you have tools like IntelliJ IDEA complaining about you not using the parens at the call site. Which I personally find cool because those parens usually indicate side-effects. Maybe we should change the `def inverse()` definition?

ceedubs: I tried removing the parens from `def inverse()` and got this:

```
[error] /Users/cody/code/cats/core/src/main/scala/cats/syntax/group.scala:15: macro implementation has incompatible shape:
[error]  required: (c: scala.reflect.macros.whitebox.Context): c.Expr[A]
[error]  or      : (c: scala.reflect.macros.whitebox.Context): c.Tree
[error]  found   : (c: reflect.macros.Context)(): c.Expr[R]
[error] number of parameter sections differ
[error]   def inverse: A = macro Ops.unop[A]
```

Paging @non, our macro optimization wizard.

milessabin: You would also need to remove the parens from the corresponding macro defn.

ceedubs:
It's been over 2 weeks since https://github.com/typelevel/cats/issues/934#issuecomment-197871798 and nobody has spoken up against this. :+1: from me.

non:
I agree with the rationale. Let's do it! 👍

Issue #810, opened 2016-01-16
Find a way to avert the CI failure `No output has been received in the last 10 minutes`


DavidGregory084:
CI builds occasionally fail due to the message:
```
No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
```
This is off-putting for newcomers as it's not immediately obvious that there might be nothing wrong with their PR.

Could we avert this by using the [travis_wait](https://docs.travis-ci.com/user/build-timeouts) command or is there some other way that we could speed up builds so they don't encounter this timeout?

InTheNow:
This is caused by a know bug that has been fixed in scalacheck, but not released. Might be best to wait on this for now.

DavidGregory084:
Ah, that makes sense. Shall we just close this, or wait for rickynils/scalacheck#195?

paulp:
Looking at  rickynils/scalacheck#195 I see this has been going on for six months. That's about 5.95 months too long when it comes to nondeterministic CI failures. They are huge time wasters.

DavidGregory084:
@paulp actually 1.13.0 is released now and it looks like cats is building against 1.12.5, maybe just needs a version bump?

ceedubs:
@DavidGregory084 I suspect we are going to run into https://github.com/rickynils/scalacheck/issues/218 if we do that.

DavidGregory084:
:cry:

paulp:
It doesn't even compile against 1.13.0. A little excerpt.
```
[error] /s/scala/cats/laws/src/main/scala/cats/laws/discipline/TraverseTests.scala:37: could not find implicit value for parameter a2: org.scalacheck.Arbitrary[A => M]
[error]         "traverse derive foldMap" -> forAll(laws.foldMapDerived[A, M] _)
[error]                                            ^
[error] /s/scala/cats/laws/src/main/scala/cats/laws/discipline/TraverseTests.scala:36: could not find implicit value for parameter a2: org.scalacheck.Arbitrary[A => X[B]]
[error]         "traverse parallel composition" -> forAll(laws.traverseParallelComposition[A, B, X, Y] _),
[error]                                                  ^
[error] 50 errors found
```

ceedubs:
[...]

PR #938, opened 2016-03-18
Onward to scala 2.11.8.

paulp:
Unused import detection has improved, leading to a number
of deletions.


paulp:
Oops, I accidentally snuck a couple new plugins in as well. That's actually what got me started on the whole thing, but I completely forgot until I skimmed the diff.

mpilquist:
:+1:

paulp:
So, let me know and I can carve off one or both of them. Coursier should be a slam dunk, the other one is a little more obscure - I just ran across it and found it handy.

screen shot 2016-03-18 at 7 04 28 am

paulp:
Great! In my initial push the build failed because `op` was not found. After restoring it, the build fails because `op` is unused. It is perhaps unused during normal compilation but used during documentation generation. I don't know how to deal with this.
```
[info] Compiling 119 Scala sources to /home/travis/build/typelevel/cats/core/.jvm/target/scala-2.11/classes...
[error] /home/travis/build/typelevel/cats/core/src/main/scala/cats/SemigroupK.scala:3: Unused import
[error] import simulacrum.{op, typeclass}
[error]                    ^
[error] one error found
[error] (coreJVM/compile:compileIncremental) Compilation failed
```

edmundnoble:
`op` is a macro annotation. Perhaps it's being expanded and removed, which is causing it to appear unused. I suggest fully qualifying the use of `op` for now and removing the import, seeing as it's the only use of `op` in the cats codebase. It looks like it will validate with that change.

paulp:
Hah, I can't believe I didn't think of doing that.

adelbertc:
:+1: on Travis green

paulp:
It all seems fine, but the builds hang up:
```
No output has been received in the last 10 minutes, this potentially
indicates a stalled build or something wrong with the build itself.
The build has been terminated
```
On my last push I removed those two plugins in case one was causing this, but it has happened again since. Has this happened before? My first guess would be memory exhaustion though I don't know why this commit would cross any threshold.

DavidGregory084:
@paulp this seems to happen fairly frequently at the minute, see #810 and associated issues

fthomas:
I've restarted the build.

adelbertc:
And restart again......

ceedubs:
Fantastic. Thanks @paulp!

PR #939, opened 2016-03-19
Add coursier plugin.

paulp:
If you haven't witnessed coursier download dependencies
you haven't yet lived.


ceedubs:
I had never seen coursier. It looks wonderful.

[This disclaimer](https://github.com/alexarchambault/coursier#important-sbt-plugin-might-mess-with-published-artifacts) looks a little scary though. Any cause for concern here?

paulp:
Good question. Obviously I'm not really able to test publishing. But as you can see, adding it is a one-line change as is removing it, so the worst-case scenario is pretty benign.

ceedubs:
@paulp that's a good point. :+1: I say why not.

stew:
:+1: 

shoot, we missed the boat on this one, sorry @paulp do you mind updating the pr so we can just merge this?

paulp:
I don't wish to politicize arenas which need not be politicized, but if that's how we'll be doing things from now on, I'm not sure why pull requests on github should be exempt.

alexarchambault:
@paulp Missed that too. Why did you close the PR? What's the problem?

paulp:
Typelevel has [set an example](http://typelevel.org/blog/2016/03/24/typelevel-boulder.html) that everything must be politicized and I am following their lead.

PR #940, opened 2016-03-20
More flexible TransLift

djspiewak:
Reimplements `TransLift` to no longer fix the type constructor.  This is substantially more flexible and makes it possible to achieve the same generality as scalaz's `Hoist`.  This is done while retaining the flexibility of cats's previous implementation, which is that `liftT` requires the minimal constraint necessary for the target `MT`.  In most cases, this is `Functor`, but in some cases (like `Kleisli`) it is nothing at all (i.e. no constraint).

To handle the "nothing at all" case, I have added a `Trivial` type, which is basically the unit typeclass.  It takes no parameters and has a single instance which is guaranteed to be in the implicit scope.  Several type aliases are provided to allow `Trivial` to be used in slots of varying shapes.  A `F[_]` typeclass (e.g. `Monoid`) can be satisfied by `Trivial.P1`, and a `F[_[_]]` typeclass (e.g. `Monad`) can be satisfied by `Trivial.PH1`.  I'm very open to nicer naming.

Oh, and for students of SI-2712, the way that `TLExtract` is encoded (specifically with `SingletonMT` and `SingletonM`) demonstrates an interesting workaround to some of its foibles.  One concern here is that the error messages will be horrendous.  But without control over the type `toString` in `@implicitNotFound`, we can't really do better.

This PR addresses #917.


djspiewak:
Er… what?  The build is hanging?  This change definitely increases the compilation time of `TransLiftTests`, but it's not increased by *that* much.  Furthermore, the hang appears to be happening at runtime.  I did confirm that `TransLiftTests` runs to completion locally, so that can't be it.  Anyone have any idea what's going on?

DavidGregory084:
@djspiewak yeah, this seems to be a big problem at the minute, see the discussion on #939, #810, #942 about trying to fix this

djspiewak:
@DavidGregory084 *Whew!*  Off the hook.  :-)

core/src/main/scala/cats/TransLift.scala
```diff
@@ -5,9 +5,24 @@ package cats
...
-trait TransLift[MT[_[_], _], M[_]] {
+trait TransLift[MT[_[_], _]] {
+
+  /**
+   * The typeclass which constrains liftT as a function of the type
+   * constructor it is given. A safe "identity" value for this type
+   * if your transformer does not constrain its lifted effects would
+   * be `type TC[M[_]] = Unit =:= Unit`.  A more common constraint
+   * might be `type TC[M[_]] = Monad[M]`.
+   */
```
DavidGregory084: Does this doc comment need to be updated to refer to `Trivial`?

djspiewak: Yes.  Will fix.

adelbertc:

ceedubs:
This looks good to me so far. @djspiewak would you be willing to make that comment update to refer to `Trivial` and to add a test or two so that that the codecov isn't reporting a coverage drop?

djspiewak:
@ceedubs Maybe I'm misreading the coverage report, but as far as I can tell, there's literally nothing to add.  All of the testable lines are already tested.  The reason the coverage is dropping is I added untestable lines (literally: whitespace and type-level declarations), which for some reason, Codecov includes in the computation which generates the coverage percentage.

Again, I may simply be misreading it.  If someone can point out to me what I missed, I will happily write tests.  :-)

adelbertc:
It looks that way to me as well. :+1: from me!

adelbertc:
Hm given the fancy nature of these types, could you also add some tests in `tests/src/test/scala/cats/tests/SyntaxTests.scala` ?

djspiewak:
Well so much for your :+1:! :-)

More seriously, the syntax is already tested by what previously existed. I could probably add one or two things that are now supported but which weren't before, but it wouldn't be too different from the tests we already have.

adelbertc:
I was halfway through amending my comment when I saw your new one :-)

Yeah I just realized there are pre-existing tests, I wrongly assumed this was a completely new addition. My :+1: stands!

ceedubs:
:+1: thanks, @djspiewak!

PR #941, opened 2016-03-20
Journey to >= 90% test coverage

adelbertc:
Type class composition and MonadState tests


ngbinh:
not sure why this one is still failing. The build was stopped only after ~20 minutes though. Let me try to upgrade scala.js to `0.6.8` (which contains some performance fixes) to see if the problems go away.

adelbertc:
Seems everyone else's builds are succeeding again except this one :( Even after `0.6.8`

ceedubs:
> Current coverage is 89.78%

So close to 90!

:+1:

PR #942, opened 2016-03-21
ScalaCheck 1.13.0

adelbertc:
Proposed fix for https://github.com/typelevel/cats/issues/810

This currently **compiles**, but **does not run**! When I try running it I get:

```
[error] Could not run test cats.tests.RegressionTests: java.lang.IncompatibleClassChangeError: Implementing class
[info] DeferredAbortedSuite:
[info] Exception encountered when attempting to run a suite with class name: org.scalatest.DeferredAbortedSuite *** ABORTED *** (0 milliseconds)
[info]   java.lang.IncompatibleClassChangeError: Implementing class
```

many many times. I thought it might be Discipline so I cloned Discipline, bumped ScalaCheck to 1.13.0, did a `publishLocal`, and bumped the Cats Discipline version. Still same error. Leaving here for comments/if anyone else has suggestions/wants to give it a shot.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -66,7 +66,7 @@ lazy val commonJvmSettings = Seq(
 
 lazy val catsSettings = buildSettings ++ commonSettings ++ publishSettings ++ scoverageSettings
 
-lazy val scalacheckVersion = "1.12.5"
+lazy val scalacheckVersion = "1.13.0"
 
 lazy val disciplineDependencies = Seq(
   libraryDependencies += "org.scalacheck" %%% "scalacheck" % scalacheckVersion,
--- a/laws/src/main/scala/cats/laws/discipline/AlternativeTests.scala
+++ b/laws/src/main/scala/cats/laws/discipline/AlternativeTests.scala
@@ -3,14 +3,13 @@ package laws
 package discipline
 
 import cats.laws.discipline.CartesianTests.Isomorphisms
-import org.scalacheck.Arbitrary
-import org.scalacheck.Prop
+import org.scalacheck.{Arbitrary, Cogen, Prop}
 import Prop._
 
 trait AlternativeTests[F[_]] extends ApplicativeTests[F] with MonoidKTests[F]  {
   def laws: AlternativeLaws[F]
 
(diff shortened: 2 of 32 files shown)


ceedubs:
I think you will also need a version of scalatest built against scalacheck 1.13

ceedubs:
See also: https://github.com/scalatest/scalatest/issues/837

adelbertc:
So steps needed for this to move forward:

1. ScalaTest release w/ ScalaCheck 1.13.0
2. Discipline with new ScalaTest and new ScalaCheck

ceedubs:
Update: @adelbertc has this working with locally published versions of Discipline and Algebra. It even seems to have caught a law-breaking instance! We should be able to move forward with this once there's a stable scalatest release.

Or perhaps we should update to the `SNAP` (stable snapshot) now and just make sure we wait until there's a stable version before the next cats release? That's actually sounding better to me considering builds are almost all failing.

ceedubs:
To my understanding, https://github.com/rickynils/scalacheck/issues/218 is a deal-breaker for upgrading to scalacheck 1.13.

Here is my understanding of the situation:

The `Cogen` support in scalacheck 1.13 is forced to call `Option.get` because of the fact that scalacheck generators return `Option[A]` instead of `A`. This can result in `None.get` exceptions when using `Cogen` along with generators that use `suchThat`/`filter`. This becomes quite problematic, because many common generators (such as `choose`, `frequency` and `oneOf`) in scalacheck use `suchThat`.

My understanding is based off of things that @xuwei-k and @non have said. It might not be entirely accurate.

non:
@ceedubs that's correct although it's a bit more complicated. Even if you remove almost all use of `suchThat` / `filter` you still get the failure. I think it has to do with the fact that shrinking can produce invalid `Gen` instances which will always fail. I'm hoping to get some time this week/weekend to test a proposed fix.

rossabaker:
Specs2 users: a downgrade to specs2-3.7 gets past the linker errors.  That is the latest version based on scalacheck-1.12.x.

ceedubs:
@adelbertc @non should this be closed in favor of https://github.com/typelevel/cats/pull/1345 ?

non:
Yes I think so.

Issue #935, opened 2016-03-16
Add an empty for OptionT


DomBlack:
Currently OptionT has a `pure` method, but it can be useful to have a `empty` method on the object. At the moment we're simply using the apply to work around it;

```scala
def something(empty: Boolean): OptionT[Future, String] = {
  if (empty) OptionT(Future.success(None)) else OptionT.pure("Hi")
}
```

However `empty` would look cleaner and be easier to understand

```scala
def something(empty: Boolean): OptionT[Future, String] = {
  if (empty) OptionT.empty else OptionT.pure("Hi")
}
```

ceedubs:
:+1: sounds good to me.

It'd look something like this, right?

```scala
def empty[F[_]:Applicative, A]: OptionT[F, A]
```

ceedubs:
I'm wondering if something like `OptionT.none` might be a better name. It might be a little counterintuitive to call `empty[List, String]` and get back a `List` that isn't actually empty (but instead contains a single `None` element). I'm also thinking of `empty[NonemptyList, String]`.

adelbertc:
+1 for `OptionT.none`

mikejcurry:
I think this should be closed after: #946

PR #946, opened 2016-03-24
Added OptionT.none, resolving #935

MaxWorgan:
I also added an alias from pure to some - that was never mentioned in the issue but seemed reasonable to me.

### DIFF
--- a/docs/src/main/tut/optiont.md
+++ b/docs/src/main/tut/optiont.md
@@ -72,6 +72,23 @@ val result: Future[Option[String]] = ot.value // Future(Some("Hello Jane Doe"))
 
 ```
 
+## From `A` to `OptionT[F,A]`
+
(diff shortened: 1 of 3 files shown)


docs/src/main/tut/optiont.md
```diff
@@ -72,6 +72,23 @@ val result: Future[Option[String]] = ot.value // Future(Some("Hello Jane Doe"))
 
 ```
 
+## From `A` to `OptionT[F,A]`
+
+If you have only an `A` and you wish to *lift* it into an `OptionT[F,A]` assuming you have an applicative instance for F you can use `some` which is an alias for `pure`. There also exists a `none` method which can be used to create an OptionT[F,A], where the Option wrapped A type is actually a `None`:
```
ceedubs: It might be nice to make `applicative` a link to the `Applicative` doc.

ceedubs:
:+1: thanks!

stew:
:+1:

Issue #561, opened 2015-10-05
Add syntax for MonadError, MonadReader, and MonadState


ceedubs:
Currently these type classes don't have syntax support.

stew:
what would syntax support allow you to do that you can't do now?

adelbertc:
i've found myself wanting `MonadError#handleError`

stew:
ah yes, that one makes sense. I'm not sure what makes sense for Reader. i guess with state we can expose put and and modify but read doesn't make sense, right?

adelbertc:
Agreed

davegurnell:
I'm in need of `raiseError`. I'll knock up a PR for `ApplicativeError` covering:

- `raiseError`
- `handleErrorWith`
- `handleError`
- `attempt`
- `attemptT`
- `recover`
- `recoverWith`

edmundnoble:
We have syntax for `MonadReader` and `MonadState` in cats-mtl; every transformer class has *all* of the syntax that makes sense for it.

PR #947, opened 2016-03-24
Syntax for ApplicativeError

davegurnell:
I've added syntax for the methods I mentioned in #561.

I'm happy with `raiseError`, `handleError`, `handleErrorWith`, `recover`, and `recoverWith`. However, I'm not sure about `attemptT` because the tests in `ApplicativeErrorTests.scala` required a bunch of heavyweight type hinting.

My inclination is to remove `attempt` and `attemptT` and leave the other methods intact. What do you think?

### DIFF
--- a/tests/src/test/scala/cats/tests/ApplicativeErrorTests.scala
+++ b/tests/src/test/scala/cats/tests/ApplicativeErrorTests.scala
@@ -1,0 +1,42 @@
+package cats
+package tests
+
+import cats.data.{Xor, XorT}
+
+class ApplicativeErrorCheck extends CatsSuite {
+
+  type ErrorOr[A] = String Xor A
+
+  val failed: String Xor Int =
+    "Badness".raiseError[ErrorOr, Int]
+
+  test("raiseError syntax creates an Xor with the correct type parameters") {
+    failed should === ("Badness".left[Int])
+  }
+
+  test("handleError syntax transforms an error to a success") {
+    failed.handleError(error => error.length) should === (7.right)
+  }
+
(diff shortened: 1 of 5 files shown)


tests/src/test/scala/cats/tests/ApplicativeErrorTests.scala
```diff
@@ -0,0 +1,42 @@
...
+  test("handleErrorWith transforms an error to a success") {
+    failed.handleErrorWith(error => error.length.right) should === (7.right)
+  }
+
+  test("attempt syntax creates a wrapped Xor") {
+    failed.attempt should === ("Badness".left.right)
+  }
+
+  test("attemptT syntax creates an XorT") {
+    type ErrorOrT[A] = XorT[ErrorOr, String, A]
```
ceedubs: This doesn't appear to be used. Please tell me that it's mere presence doesn't affect whether the next line compiles? :|

ceedubs:
@davegurnell thank you!

I'm probably missing something but the tests don't really appear to have more type gymnastics for `attempt` and `attemptT` than for other methods to me. Could you elaborate on the issue with those?

:+1: either with or without `attempt` and `attemptT`

davegurnell:
I guess it was late when I wrote the PR. The type annotations were on the [expected value](https://github.com/typelevel/cats/pull/947/files#diff-7b0332a0a4b928f389b743d893b4c960R30) in the test for `attemptT`, not on the use of `attemptT` itself.

adelbertc:

adelbertc:
:+1:

PR #950, opened 2016-03-26
Add curried version of .fold to Xor and Validated

tbrown1979:
This is just a personal preference, but I like the curried syntax for `.fold`ing things. This pr doesn't remove the current form of `.fold`ing, it just adds the curried form (to both `Xor` and `Validated`). You can use either! This is more inline with `Option.fold` which is maybe why I prefer it.

### DIFF
--- a/core/src/main/scala/cats/data/Validated.scala
+++ b/core/src/main/scala/cats/data/Validated.scala
@@ -15,7 +15,7 @@ sealed abstract class Validated[+E, +A] extends Product with Serializable {
       case Valid(a) => fa(a)
     }
 
-  def fold[B](fe: E => B): (A => B) => B = fold(fe, _)
+  def foldC[B](fe: E => B): (A => B) => B = fold(fe, _)
 
   def isValid: Boolean = fold(_ => false, _ => true)
   def isInvalid: Boolean = fold(_ => true, _ => false)
--- a/core/src/main/scala/cats/data/Validated.scala
+++ b/core/src/main/scala/cats/data/Validated.scala
@@ -15,6 +15,8 @@ sealed abstract class Validated[+E, +A] extends Product with Serializable {
       case Valid(a) => fa(a)
     }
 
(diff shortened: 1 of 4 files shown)


adelbertc:

ceedubs:
Thanks for your contribution, @tbrown1979!

I think we might want to avoid overloading the name `fold` to avoid confusion. In the case of `Xor` I think that the current version tends to have better inference, and sadly in scala I think it ends up being more performant. But I'm open to the idea of also having a curried version (maybe `foldC` or something?). How do others feel about this? Does anyone else find themselves wanting a curried version?

tbrown1979:
@ceedubs Great idea. I've renamed it to `.foldC`. We'll have to see if anyone else would find this a worthwhile addition.

tbrown1979:
Does anyone have a preference either way with this? I would still find this useful to have, but if no one else finds it of interest I can close it.

non:
Hi @tbrown1979. We've been reluctant to add curried `fold` methods in cases where it seems to make type inference worse.

For example, `.fold(_ => None, x => Some(x))` will correctly infer `Option`, whereas `.fold(_ => None)(x => Some(x))` will fail (because it will infer `None.type` from the first parameter list). So far we haven't been able to find cases where the currying improves type inference. (This includes the `.fold` method on `scala.Option`.)

I think if there are cases where we can show that currying the `.fold` parameters improves inference I'm happy to add it (or really, convert `.fold` to that style). For example, we do use currying with `foldLeft` / `foldRight` because it helps the inference of anonymous function parameters. If not, I think we'd rather be opinionated and leave this out.

What do you think? Does this make sense? Do you have examples where things are improved by currying in these cases?

tbrown1979:
Thanks for the detailed response @non. I appreciate it. That makes sense to me. 

I don't have any example that this would improve type inference in these cases. My only objective was a nicer looking syntax for folding on Validated/Xor. I'm perfectly fine with Cats remaining opinionated on this :) I appreciate you guys taking the time to respond.

PR #951, opened 2016-03-26
Are builds hanging because of 2.11.8?

adelbertc:
I have no reason to believe this would solve anything, the only basis I have is it seems builds have stalled more since bumping to 2.11.8. Wondering if this will "fix" the build... for science. And stuff.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ git:
 
 scala:
 - 2.10.6
-- 2.11.8
+- 2.11.7
 
 script:
   - scripts/travis-publish.sh
--- a/build.sbt
+++ b/build.sbt
@@ -14,8 +14,8 @@ lazy val scoverageSettings = Seq(
 
 lazy val buildSettings = Seq(
   organization := "org.typelevel",
-  scalaVersion := "2.11.8",
-  crossScalaVersions := Seq("2.10.6", "2.11.8")
+  scalaVersion := "2.11.7",
+  crossScalaVersions := Seq("2.10.6", "2.11.7")
 )
 
 lazy val catsDoctestSettings = Seq(


adelbertc:
guess not

mikejcurry:
Looks like this one is going to fail also. I'm not sure if anyone else has noticed, but even though these fail somewhat arbitrarily, there is also a consistent pattern.

Specifically, a large number of the failed builds I click on to see why they fail (I'm a bit weird, go figure). This is a completely random sample/observation so could be miles off, and I've definitely seen other failures also.

_[info] - transform
[info] - local_

This always seems to be during the scalajs compilation phase also, after running validateJVM for code coverage etc.

Just thinking out loud about a couple of potential things worth exploring:

- Would splitting scalajs into a separate build run completely and running in parallel in another run be an option and might that help?
- Is there an option of getting more memory (don't know much about the free Travis, and I think maybe it is capped at the 3GB currently allocated, but I don't really know)
- Maybe, and this would be 100% last gasp, temporarily disabling scalajs tests in the build process for CI might be useful - clearly not ideal but better than builds failing arbitrarily if it made the difference.

Just possibilities to throw out for discussion - someone else might have better ones.

ngbinh:
We are working on a big scala.js project and based on my experience, these two steps should fix the issue:
1. https://github.com/scala-js/scala-js/issues/1788#issuecomment-178346142 means we need to turn on `batchMode` for Travis.
2. By default, scala.js uses `Rhino` to compile scala.js to js. `Rhino` is atrocious in term of memory requirement so it is a must for large scala.js project to use other `Js` runtime such as `phantom.js` or `node.js`.

I can probably send these two PRs soon.

ceedubs:
@ngbinh thanks that sounds promising!

PR #952, opened 2016-03-27
Reduce maxSize and minSuccessful in Scalacheck params

ceedubs:
This is an experiment to see if builds time out less if we don't run as
many samples with Scalacheck.

### DIFF
--- a/laws/src/main/scala/cats/laws/discipline/Eq.scala
+++ b/laws/src/main/scala/cats/laws/discipline/Eq.scala
@@ -13,7 +13,7 @@ object eq {
    */
   implicit def function1Eq[A, B](implicit A: Arbitrary[A], B: Eq[B]): Eq[A => B] = new Eq[A => B] {
     def eqv(f: A => B, g: A => B): Boolean = {
-      val samples = List.fill(100)(A.arbitrary.sample).collect{
+      val samples = List.fill(10)(A.arbitrary.sample).collect{
         case Some(a) => a
         case None => sys.error("Could not generate arbitrary values to compare two functions")
       }
--- a/laws/src/main/scala/cats/laws/discipline/Eq.scala
+++ b/laws/src/main/scala/cats/laws/discipline/Eq.scala
@@ -2,6 +2,7 @@ package cats
 package laws
 package discipline
 
+import catalysts.Platform
 import algebra.Eq
 import org.scalacheck.Arbitrary
 
@@ -12,8 +13,10 @@ object eq {
    * and comparing the application of the two functions.
    */
   implicit def function1Eq[A, B](implicit A: Arbitrary[A], B: Eq[B]): Eq[A => B] = new Eq[A => B] {
+    val sampleCnt: Int = if (Platform.isJvm) 50 else 5
+
     def eqv(f: A => B, g: A => B): Boolean = {
-      val samples = List.fill(10)(A.arbitrary.sample).collect{
+      val samples = List.fill(sampleCnt)(A.arbitrary.sample).collect{
         case Some(a) => a
         case None => sys.error("Could not generate arbitrary values to compare two functions")
       }
--- a/tests/src/test/scala/cats/tests/CatsSuite.scala
+++ b/tests/src/test/scala/cats/tests/CatsSuite.scala
@@ -20,10 +20,10 @@ trait TestSettings extends Configuration with Matchers {
 
   lazy val checkConfiguration: PropertyCheckConfiguration =
     PropertyCheckConfiguration(
-      minSuccessful = PosInt(5),
+      minSuccessful = if (Platform.isJvm) PosInt(50) else PosInt(5),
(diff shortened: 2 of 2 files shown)


ceedubs:
This build passed the first time (after I fixed my compile error :|). I'm going to restart it and see if it succeeds again.

ceedubs:
This build succeeded twice in a row. We might want to do something like this. We can tweak the parameters as others see fit. I'm not _too_ concerned about how much this will reduce test cases because:

1) The current status of most builds failing is the worse of two evils I think.
2) These tests are run quite often (every PR, every merge to master), so over time we should still get quite a few test cases in.
3) As far as setting `maxSize` to `5` (from the previous `100`): Due to the abstract nature of Cats, I think that generally if things work with 5 elements they will work with 100. Stack overflows are the most likely exception, but `100` still probably isn't enough to trigger them unless we are doing something quadratic in nature.

adelbertc:
LGTM :+1:

tests/src/test/scala/cats/tests/CatsSuite.scala
```diff
@@ -20,8 +20,11 @@ trait TestSettings extends Configuration with Matchers {
 
   lazy val checkConfiguration: PropertyCheckConfiguration =
     PropertyCheckConfiguration(
-      minSuccessful = if (Platform.isJvm) PosInt(100) else PosInt(10),
-      maxDiscardedFactor = if (Platform.isJvm) PosZDouble(5.0) else PosZDouble(50.0))
+      minSuccessful = if (Platform.isJvm) PosInt(5) else PosInt(5),
```
mikejcurry: Minor, but the if expression is doing nothing here.

ceedubs: @mikejcurry thanks for catching my hastiness :). Fixed.

mikejcurry:
That's 3 passed builds in a row on this ticket I think. :-) - FWIW, I say :+1: - lot's of PRs with failed builds at the moment and this could resolve a lot of them.

It might also make sense to have an issue recorded to revisit and see if there is somewhere in between 5 and 100 that works - I guess the more tests run in a single run the better, so somewhere in between, there might be a sweet spot :-)

ceedubs:
Some changes on the scala.js side have improved the failing build situation a lot. I think there is still benefit in something like this but we might not want quite as drastic of a reduction as I had previously made.

travisbrown:
To be honest my vote would be to drop Scala.js from the Travis CI builds and check it locally before releases. That's annoying, but it would cut two-thirds (or more) of the time off of PR builds and eliminate the flakiness.

I'd be a little worried about ScalaCheck hitting the same cases most of the time for some types with `minSuccessful` this small.

ceedubs:
@travisbrown I can see your argument for dropping scala.js from the Travis builds. It makes me a bit nervous, though. I don't want us to get into a situation where a PR is merged and then months later when we go to do a release we realize that approach isn't going to work at all in the scala.js world. Would there be a way to avoid that?

For now I've used `Platform.isJVM` to reduce the sample size/count in the JS world while keeping it fairly high for the JVM world. If this PR build passes several times in a row then it might be a good-enough-for-now solution. I'd really like to see an end to the sporadic build failures, and since they have improved a lot after some recent scala.js changes, I'm hopeful that this could be the extra inch that's needed.

laws/src/main/scala/cats/laws/discipline/Eq.scala
```diff
@@ -12,8 +13,10 @@ object eq {
    * and comparing the application of the two functions.
    */
   implicit def function1Eq[A, B](implicit A: Arbitrary[A], B: Eq[B]): Eq[A => B] = new Eq[A => B] {
+    val sampleCnt: Int = if (Platform.isJvm) 50 else 5
```
mikejcurry: Would it be better to keep the JVM side at 100? Is there example builds where this has failed for lack of output when building the JVM side?

non:
I think we should do this. Making sure the JS tests run is important (and we should be running the tests manually e.g. before a release) but for Travis I think this is a better default.

non:
👍

laws/src/main/scala/cats/laws/discipline/Eq.scala
ceedubs: This is a fair question. When I wrote this my thought process was that these tests run (presumably with a different RNG seed) on every PR and every merge to master, so 50 samples should be plenty, especially given the general nature of cats. I thought this might just speed things up a bit. I don't have any strong opinions on this and am happy to change it if people would prefer.

ceedubs: @mikejcurry I'm going to go ahead and merge this PR to help make the build reliable. Feel free to open up a followup issue if you have any concerns about this lower number.

mikejcurry: @ceedubs - I've no strong feelings. I just wanted to raise the question, and the build speed is a good argument as the cats end to end build has certainly gotten longer. I just wanted to raise the question. :-)

PR #953, opened 2016-03-27
Enable `batchMode` on scala.js Optimizer

ngbinh:
The `batchMode` should help reduce the memory requirement to compile
scala.js code as per
https://github.com/scala-js/scala-js/issues/1788#issuecomment-178346142

Implementation details: `botBuild` is a sbt setting that will be turned
on on travisCi so that it will not interfere with local dev experience.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ env:
   global:
   - secure: Kf44XQFpq2QGe3rn98Dsf5Uz3WXzPDralS54co7sqT5oQGs1mYLYZRYz+I75ZSo5ffZ86H7M+AI9YFofqGwAjBixBbqf1tGkUh3oZp2fN3QfqzazGV3HzC+o41zALG5FL+UBaURev9ChQ5fYeTtFB7YAzejHz4y5E97awk934Rg=
   - secure: QbNAu0jCaKrwjJi7KZtYEBA/pYbTJ91Y1x/eLAJpsamswVOvwnThA/TLYuux+oiZQCiDUpBzP3oxksIrEEUAhl0lMtqRFY3MrcUr+si9NIjX8hmoFwkvZ5o1b7pmLF6Vz3rQeP/EWMLcljLzEwsrRXeK0Ei2E4vFpsg8yz1YXJg=
+  - CATS_BOT_BUILD=true
 cache:
   directories:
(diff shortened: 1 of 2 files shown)


ngbinh:
related to #951

.travis.yml
```diff
@@ -21,6 +21,7 @@ env:
   global:
   - secure: Kf44XQFpq2QGe3rn98Dsf5Uz3WXzPDralS54co7sqT5oQGs1mYLYZRYz+I75ZSo5ffZ86H7M+AI9YFofqGwAjBixBbqf1tGkUh3oZp2fN3QfqzazGV3HzC+o41zALG5FL+UBaURev9ChQ5fYeTtFB7YAzejHz4y5E97awk934Rg=
   - secure: QbNAu0jCaKrwjJi7KZtYEBA/pYbTJ91Y1x/eLAJpsamswVOvwnThA/TLYuux+oiZQCiDUpBzP3oxksIrEEUAhl0lMtqRFY3MrcUr+si9NIjX8hmoFwkvZ5o1b7pmLF6Vz3rQeP/EWMLcljLzEwsrRXeK0Ei2E4vFpsg8yz1YXJg=
+  - botBuild=true
```
ceedubs: Does it matter that this is using `=` instead of `:`?

build.sbt
```diff
@@ -55,7 +57,11 @@ lazy val commonJsSettings = Seq(
     s"-P:scalajs:mapSourceURI:$a->$g/"
   },
   scalaJSStage in Global := FastOptStage,
-  parallelExecution := false
+  parallelExecution := false,
+  // Only use for scala.js for now
+  botBuild := sys.props.getOrElse("botBuild", default="false") == "true",
```
ceedubs: This is probably me being paranoid, but since this is a global environment var should we namespace it a bit? Maybe `CATS_BOT_BUILD` or something?

ngbinh: :+1: will correct it

ngbinh:
updated and squashed

adelbertc:
Should/does this change need to be in conjunction with the switch away from Rhino? Seems the builds still take a long time, upwards of an hour.

ngbinh:
ok, I will merge this PR into #954

PR #954, opened 2016-03-27
Switch to use nodeJsEnv as default jsEnv to build scala.js

ngbinh:
By default, scala.js compiler uses `Rhino` as the js environment. But it
usually requires a lot of memory and often lead to OOM. This PR replace
the default to a node.js one that should be much faster and slimer.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,10 @@ scala:
 script:
   - scripts/travis-publish.sh
 
+# http://austinpray.com/ops/2015/09/20/change-travis-node-version.html
+install:
+  - rm -rf ~/.nvm && git clone https://github.com/creationix/nvm.git ~/.nvm && (cd ~/.nvm && git checkout `git describe --abbrev=0 --tags`) && source ~/.nvm/nvm.sh && nvm install $TRAVIS_NODE_VERSION
+
 notifications:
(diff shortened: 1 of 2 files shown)


adelbertc:
Awesome :+1: looks like this helped a lot. Build times are now in the 30 minute range as opposed to 1hr+. Do we want to do this together with https://github.com/typelevel/cats/pull/953 in one PR instead (e.g. closing #953 and putting it on this one) since it seems #953 is still suffering from long build times.

ceedubs:
Thanks @ngbinh!

Forgive my ignorance about scala.js: will this affect users Cats scala.js users, or does this only change the environment that our JS tests run in?

ngbinh:
@ceedubs it shouldn't affect scala.js Cats users. It only used a different jsEnv. Scala.js libraries are actually compiled in `nodejs` environment.

adelbertc:
Hurrah! :+1:

ceedubs:
:100:

adelbertc:
Hmm any ideas why this build is still failing? https://github.com/typelevel/cats/pull/941

ngbinh:
commented in #941. There is one more thing I'd like to do is upgrading scala.js plugin.

PR #958, opened 2016-03-28
Switch off scaladoc generation for Scala 2.10 due to macro problems

mikejcurry:
This PR would switch off scaladoc generation for Scala 2.10 during the makeSite process. The site generated in 2.10 mode would not have scaladoc generated, but would be otherwise identical to the site generated in 2.11 mode.

This allows documentation to be put together for all classes, and should allow #898 to pass the build.


ceedubs:
Sounds good to me if it should fix #898. Thanks!

rossabaker:
I did the same for http4s for the same reason, and haven't had any regrets.

PR #960, opened 2016-03-28
More Reducible tests

adelbertc:

Issue #516, opened 2015-09-02
Validate should validate Scaladocs too


non:
Right now it's possible to have Scaladoc errors which aren't caught by `validate`.

We should have it run `compile:doc` (or similar) for each project in addition to the other work it is doing.

ceedubs:
@non I'm unable to reproduce this. Do you have an example of a change that should cause it to fail but doesn't? It includes a `makeSite` call, which should compile the ScalaDocs.

non:
In the `BimonadLaws` scaladoc, change the "`Bimonad`" to a "[[Bimonad]]" and run `laws/compile:doc`.

ceedubs:
Ah, I suspect this and #510 have the same root cause.

ceedubs:
I think I see the issue. PR incoming.

ceedubs:
Hmm. I thought it was as easy as adding the laws project to https://github.com/non/cats/blob/672e0105f2d0136c3185ff79300a002384a2ec8b/build.sbt#L58 but I get some bizarre errors if I try to do that. I'm not sure what is going on.

non:
Ohhhh, now I'm almost remembering something. I think maybe we hit these errors a long time ago and disabled the laws project because we didn't understand them.

I also wonder if any of this is related to #510.

mikejcurry:
Investigated this for a little. After making the change outlined @ceedubs the doc build starts to fail. Some observations:
* For the most part the issues are confined to a couple of areas 
 * References to `Foldable` typeclass syntax 
 * References to `Apply` typeclass syntax 
* Example:
 * `TraverseLaws` references an implicit `Traversable[F]`
 * `TraverseLaws` also imports the `foldable` syntax 
 * At line 59, `TraverseLaws` tries to use this syntax `val rhs: B = fa.foldMap(f)`, but this fails compilation
 * Changing the code here to not use the syntax results in correct compilation, i.e. `val rhs: B = F.foldMap(fa)(f)` compiles successfully during doc generation 

Interestingly,
* The `Traverse` syntax used in the `TraverseLaws` is fine and compiles, so not all syntax is affected, and the `Traverse` syntax does not seem to use simulacrum for syntax, even though the typeclass uses simulacrum.
* References to the `Foldable` syntax in the `FoldableLaws` class is also failing, indicating that it is not specific to the usage in `Traverse`.
* A similar problem occurs with the `Apply` syntax in all areas that `Apply` syntax is used in the laws.

[...]

PR #961, opened 2016-03-28
Fix build for scaladoc generation of Laws

### LINKED ISSUE #510: Laws are missing in the published API docs
All `*Laws` are missing in http://non.github.io/cats/api/#package. They should be included in the API docs because the laws are an important part of our type classes.

mikejcurry:
This PR changes a number of laws classes to *not* use syntax, and updates build.sbt to enable scaladoc generation for the law project.

The problem outlined in #516 only occur if syntax is used to reference the operations on Apply and Foldable. This PR changes these references to use the type classes directly instead of using the syntax.

Once this is in place, the scaladoc generates correctly for the laws project during the makeSite project.

This PR would fix #510 as it publishes the scaladoc for laws as part of the site.

The fix is a bit ugly, because specific syntax has to be avoided, and a deeper investigation into how to make that syntax work with scaladoc is probably required, but given where things are, this is probably a useful step forward, despite the kludginess.

Note that this has conflicts with both #959 and #960 so will need to merge to latest once those have been merged.

### DIFF
--- a/laws/src/main/scala/cats/laws/BimonadLaws.scala
+++ b/laws/src/main/scala/cats/laws/BimonadLaws.scala
@@ -2,7 +2,7 @@ package cats
 package laws
 
 /**
- * Laws that must be obeyed by any `Bimonad`.
+ * Laws that must be obeyed by any [[Bimonad]].
  *
  * For more information, see definition 4.1 from this paper:
  * http://arxiv.org/pdf/0710.1163v3.pdf
--- a/build.sbt
+++ b/build.sbt
@@ -90,14 +90,14 @@ lazy val disciplineDependencies = Seq(
  * it does mean that the scaladocs cannot be generated when the build is in 2.10 mode.
  */
 def noDocProjects(sv: String): Seq[ProjectReference] = CrossVersion.partialVersion(sv) match {
-    case Some((2, 10)) => Seq[ProjectReference](coreJVM)
+    case Some((2, 10)) => Seq[ProjectReference](coreJVM, lawsJVM)
     case _ => Nil
   }
 
 lazy val docSettings = Seq(
   autoAPIMappings := true,
   unidocProjectFilter in (ScalaUnidoc, unidoc) :=
-    inProjects(coreJVM) -- inProjects(noDocProjects(scalaVersion.value): _*),
+    inProjects(coreJVM, lawsJVM) -- inProjects(noDocProjects(scalaVersion.value): _*),
   site.addMappingsToSiteDir(mappings in (ScalaUnidoc, packageDoc), "api"),
   site.addMappingsToSiteDir(tut, "_tut"),
   ghpagesNoJekyll := false,
--- a/laws/src/main/scala/cats/laws/AlternativeLaws.scala
+++ b/laws/src/main/scala/cats/laws/AlternativeLaws.scala
@@ -8,13 +8,13 @@ trait AlternativeLaws[F[_]] extends ApplicativeLaws[F] with MonoidKLaws[F] {
   implicit def algebra[A]: Monoid[F[A]] = F.algebra[A]
 
   def alternativeRightAbsorption[A, B](ff: F[A => B]): IsEq[F[B]] =
-    (ff ap F.empty[A]) <-> F.empty[B]
+    F.ap(ff)(F.empty[A]) <-> F.empty[B]
 
   def alternativeLeftDistributivity[A, B](fa: F[A], fa2: F[A], f: A => B): IsEq[F[B]] =
     ((fa |+| fa2) map f) <-> ((fa map f) |+| (fa2 map f))
 
   def alternativeRightDistributivity[A, B](fa: F[A], ff: F[A => B], fg: F[A => B]): IsEq[F[B]] =
-    ((ff |+| fg) ap fa) <-> ((ff ap fa) |+| (fg ap fa))
+    F.ap(ff |+| fg)(fa) <-> (F.ap(ff)(fa) |+| F.ap(fg)(fa))
 
 }
 
--- a/laws/src/main/scala/cats/laws/ApplicativeLaws.scala
+++ b/laws/src/main/scala/cats/laws/ApplicativeLaws.scala
@@ -1,7 +1,6 @@
 package cats
 package laws
 
-import cats.syntax.apply._
 import cats.syntax.functor._
 
 /**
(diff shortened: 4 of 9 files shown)


laws/src/main/scala/cats/laws/BimonadLaws.scala
```diff
@@ -2,7 +2,7 @@ package cats
 package laws
 
 /**
- * Laws that must be obeyed by any `Bimonad`.
+ * Laws that must be obeyed by any [[Bimonad]].
```
mikejcurry: Just added this to show that this compiles correctly during site generation and generates a link to Bimonad from BimonadLaws during site generation. This example was specifically referenced in #516

InTheNow:
Out of interest, what exactly is the cause of the scaladoc breaking the builds? By that, I mean is it scaladoc/unidoc/doctest/makesite...?

I ask just to be sure that the correct thing is being disabled in the build. For example, if it is doctest  that is failing, we could just disable that in 2.10 and still generate the docs

InTheNow:
You also might want to look at https://github.com/InTheNow/sbt-catalysts/blob/master/src/main/scala/sbtcatalysts/CatalystsPlugin.scala#L306-L308

Taken from the note at https://github.com/sbt/sbt-unidoc#how-to-unify-scaladoc

mikejcurry:
Hey, thanks for getting back. Yea, for sure, I guess I raised this one as much for discussion as anything else. If there are better options we should definitely go with those.

The other 2 PRs I sent were to disable scaladoc for 2.10 only. If theres a better way to do that, then all for it - they stop a problem which happens for macro expansion in 2.10 when generating scaladoc for `Comonad` - i.e. #898, to try to help that one add some docs for `Comonad`. Other projects have mentioned they have removed 2.10 from macro generation also, hence those PRs

But that's a different issue to this one.

This one references an issue that's been around for a while. The issue #510 is basically stating that the laws scaladoc should be published, because it is an important part of cats. The problem there affects scala 2.11. Being honest, I'm not sure what the underlying cause is. It may not be related to macro expansion at all. It only seems to affect two syntax types, ApplySyntax and FoldableSyntax. Other syntax types used in the laws code seem fine when generating docs. It could be related to the Unapply classes that are used when constructing the syntax for Foldable and Apply. 

So basically, when I noticed that it only relates to the syntax for those two type classes, I put this PR together to show that it could be resolved by avoiding the syntax for those two type classes. 

So, this definitely solves #510. But it is kludgy and if there is a different way to solve it without avoiding the syntax issues, I definitely think we should do that.

InTheNow:
@mikejcurry Might be worth a look to see if the setting I mention helps somehow, even if other issues still exist

mikejcurry:
Yea, I'd seen that alright, but couldn't find a way to get it to work with cats.

Unfortunately, `-Ymacro-no-expand` (or as that is deprecated, the current approach of `-Ymacro-expand:none`), results in a pile of errors like the below. I don't think that disabling macro expansion will be an option given how much macros are used in the core of the definition of the library.

```
[error] /Users/user/repos/open_source/cats/core/src/main/scala/cats/syntax/coflatMap.scala:11: not found: value CoflatMap
[error] trait CoflatMapSyntax extends CoflatMap.ToCoflatMapOps with CoflatMapSyntax1
[error]                               ^
[error] /Users/user/repos/open_source/cats/core/src/main/scala/cats/syntax/comonad.scala:12: not found: value Comonad
[error] trait ComonadSyntax extends Comonad.ToComonadOps with ComonadSyntax1
[error] /Users/user/repos/open_source/cats/core/src/main/scala/cats/syntax/contravariant.scala:15: type ToContravariantOps is not a member of object cats.functor.Contravariant
[error] trait ContravariantSyntax extends Contravariant.ToContravariantOps with ContravariantSyntax1
[error]                                                 ^
[error] /Users/user/repos/open_source/cats/core/src/main/scala/cats/syntax/foldable.scala:12: type ToFoldableOps is not a member of object cats.Foldable
[error] trait FoldableSyntax extends Foldable.ToFoldableOps with FoldableSyntax1 {
[error]                                       ^
```

mikejcurry:
Hey,
So, I merged this up with the latest to resolve conflicts.

I realize the changes might be a bit contentious. As far as I can ascertain, there is something that goes wrong with the application of `Unapply` when the compiler runs in scaladoc mode, which means that the syntax for the `Foldable` and `Apply` type classes that requires `Unapply` in order to be found, are not resolved in that compiler mode which in turn means that any references to this syntax fails to compile in scaladoc/unidoc generation mode.

The end result of all that is that the scaladoc for the laws project can't be published.

The way I see it, this workaround at least allows the scaladoc for the laws project to be published, which is a good step forward. The limitations related to the syntax and the scaladoc references could be followed up separately.

mikejcurry:
Kind of hoping that the amazing higher order unification work that has been going on makes this proposed workaround obsolete. I can close this if folks prefer, but might be worth keeping around for a while as a PR until an alternative is found. If the need for `Unapply` goes away, then I'd hope that the scaladoc generation which is failing in this area **might** just start working. :-).

InTheNow:
I cannot be certain, but https://github.com/akka/akka/issues/20533 may be a similar issue, 
as the errors are similar : 

```sbt
[error] value ap is not a member of type parameter F[B => C]
[error] value ap is not a member of type parameter F[A => B]
[error] Unused import
[error] import cats.syntax.apply._
```

ie the errors can be reproduced in the library `compile` build by removing the syntax import

@eed3si9n Any thoughts on this, please?

eed3si9n:
Unifying the source seem to interfere with implicits and macros.
What I've done in the past is workaround this by creating [a branch](https://github.com/eed3si9n/sbt/commit/91af1c61c46749bb8e92bd94c2028da6e0b9c374) and sprinkling something like:

```scala
implicit def fasttypetag[A]: FastTypeTag[A] = ???
```

liberally in the source, since you just need to get the code to compile to generate the docs.
Maybe you can create `object DocHacks`, which remains empty for normal compilation, but during the doc does all the hacks, and call `import DocHacks._`.

InTheNow:
@eed3si9n Firstly, thanks for the feedback, very much appreciated.

I wonder, when you say "Unifying the source seem...." if this would work better if we did not use unidoc for now (we have no Java code)? Or would it not make any difference?

mikejcurry:
Closing in anticipation of the great work in #1082 !!! Great stuff @InTheNow

Issue #969, opened 2016-04-04
Kleisli.apply isn't what it seems like it should be


ceedubs:
In [this gitter discussion](https://gitter.im/typelevel/cats?at=5702da0fe4a8384a1bbec87f) we realized that `Kleisli.apply` should probably be renamed `Kleisli.ap`.

We probably also want to create a new `apply` method that looks like `def apply(a: A): F[B] = run(a)`.

adelbertc:
fixed by #970

PR #970, opened 2016-04-04
Renamed `Kleisli#apply`

ZizhengTai:
Addressed #969. Renamed old `Kleisli#apply` to `Kleisli#ap`, and added `Kleisli#apply` which is an alias for `Kleisli#run`.


adelbertc:
👍  on green

ceedubs:
:+1:

It would probably be nice to add a test that apply and run are consistent.

PR #971, opened 2016-04-05
Add toValidatedNel to Xor

liff:
As a convenient shorthand for `xor.toValidated.toValidatedNel`.

I suppose `.toValidatedNel` should be the version more often used since the point of `Validated` is to accumulate errors?


stew:
:+1:

adelbertc:
Is the `[AA >: A]` needed here?

liff:
It doesn't compile without:

```
cats/core/src/main/scala/cats/data/Xor.scala:75: covariant type A occurs in invariant position in type => cats.data.ValidatedNel[A,B] of method toValidatedNel
  def toValidatedNel: ValidatedNel[A,B] = fold(Validated.invalidNel, Validated.valid)
      ^
```

Is it due to `OneAnd`'s invariance?

adelbertc:
Ah yes 👍

PR #973, opened 2016-04-09
Add flatMapF for StateT

travisbrown:
I wanted this today and it seems like a reasonable thing to have.


ceedubs:
:+1: thanks, @travisbrown.

While the situation has improved quite a lot after some scala.js changes, we are still seeing some builds (including one for this PR) stall. I'll restart the build, but would you want to comment on #952 as a potential solution?

adelbertc:
👍

PR #975, opened 2016-04-11
Add syntax for separate from MonadCombine.

peterneyens:
I based myself on the structure in syntax.bitraverse for the two implicit conversions (the
existing for unite and the new for separate).

MonadCombine separate was added in #864 by @adelbertc.

### DIFF
--- a/core/src/main/scala/cats/syntax/monadCombine.scala
+++ b/core/src/main/scala/cats/syntax/monadCombine.scala
@@ -1,15 +1,13 @@
 package cats
 package syntax
 
-trait MonadCombineSyntax extends MonadCombineSyntax1 {
+trait MonadCombineSyntax {
   // TODO: use simulacrum instances eventually
   implicit def monadCombineSyntax[F[_]: MonadCombine, G[_], A](fga: F[G[A]]): MonadCombineOps[F, G, A] =
     new MonadCombineOps[F, G, A](fga)
-}
 
-private[syntax] trait MonadCombineSyntax1 {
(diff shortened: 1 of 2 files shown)


ceedubs:
:+1: looks good. Thanks!

adelbertc:
Hm what is the motivation of calling it "nested" and having it in a `MonadCombineSyntax1` as opposed to just the `MonadCombineSyntax`?

peterneyens:
I followed the same structure as in the [bitraverse syntax](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/syntax/bitraverse.scala).

`MonadCombineSyntax1` doesn't seem necessary yes (and neither does `BitraverseSyntax1` ?)

Should I rename `NestedMonadCombineOps` to `SeparateOps` (like `FlattenOps` and `IfMOps` in the [flatMap syntax](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/syntax/flatMap.scala)?

peterneyens:
I have renamed the `NestedMonadCombineOps` to `SeparateOps` and moved the implicit conversion into `MonadCombineSyntax`.

I have also added a test for `MonadCombine` in [SyntaxTests](https://github.com/typelevel/cats/blob/master/tests/src/test/scala/cats/tests/SyntaxTests.scala). Am I right that this file just needs to compile, for the tests to be valid ?

ceedubs:
👍 looks good to me. Thanks @peterneyens!

non:
Looks good to me. 👍

Issue #974, opened 2016-04-11
Create a copair typeclass


Jacoby6000:
I was speaking with @tpolecat in the gitter channel, and I was looking to generalize some things, where I'd like to accept `Either`, `Xor` and `Validated`. Rob mentioned that this would be called a `Copair`.  I've done a very small (tiny) amount of work on this so far, but I know there's more.  I'm working to fully implement this now.  To provide an idea, my first pass implementation looked like this

``` scala
  trait Copair[F[_,_]] {
    def left[A,B](a: A): F[A,B]
    def right[A,B](b: B): F[A,B]

    def fold[A,B,C](f: F[A,B])(fa: A => C, fb: B => C): C

    final def swap[A,B](f: F[A,B]): F[B,A] = fold(f)(right[B,A](_), left[B,A](_))
  }

  implicit class CopairOps[F[_,_]: Copair, A, B](pair: F[A,B]) {
    val ev = implicitly[Copair[F]]

    def fold[C](fa: A => C, fb: B => C): C = ev.fold(pair)(fa, fb)
    def swap: F[B,A] = ev.swap(pair)
  }

  implicit class CopairBuilderOps[A](a: A) {
    def left[F[_,_]: Copair, A, B](a: A): F[A,B] = implicitly[Copair[F]].left(a)
    def right[F[_,_]: Copair, A, B](b: B): F[A,B] = implicitly[Copair[F]].right(b)
  }
```

Based on the conversations I had with @tpolecat it sounds like you should be able to derive a `Bifunctor` and `Bifoldable` given a `Copair`.  He also mentioned that there's probably a reason this thing doesn't exist, so this might be a dead end.

etorreborre:
I am wondering if `Copair` is not actually a `Coproduct` as described in Alissa Pajer's presentation: https://vimeo.com/channels/flatmap2016/165852490.

Jacoby6000:
Closing this issue in light of discussion on #976. Probably should've closed this ages ago 😅

PR #976, opened 2016-04-12
Created Copair typeclass

Jacoby6000:
This is a PR attempting to take care of #974. I'm mostly looking for feedback right now, as I'm new to FP, and contributing to Cats. 

A `Copair` is a generalization of `Xor`, `Either`, and `Validated`.

~~What's missing currently are tests, mainly because I'm not sure what laws `Copair` should have. I've thought of `f.swap.swap <-> f` but that's about it.~~  However I know there are more.  These things just aren't intuitive to me yet.  

I'm also looking for feedback on how to organize things. ~~As it stands now, there will be problems with ambiguous implicits, because of the `Bitraverse` derivation.~~ I can move it somewhere to make this better, but I'm not sure where it should go.  I'm also not sure where/how I should handle the `CopairIdOps` class.


adelbertc:
Ideas for some laws:

```scala
left(a).fold(f, g) == f(g)
right(b).fold(f, g) == g(b)

// left is associative
left(left(x, y), z) ~= left(x, left(y, z))

// similarly for right
```

It can also extend/implement `Bitraverse` (and by extension `Bifoldable` and `Bifunctor`).

I have had a use case for this once I think, but I don't see this around much so I'm wondering if there's a reason for that..

/cc @tpolecat since I saw him briefly discussing this with you on Gitter

Jacoby6000:
I had initially created this with an implicit function that would derive a `Bitraverse` from a `Copair`, but I've updated `Copair` to extend `Bitraverse`.  I also went ahead and replaced `Either`/`Xor`/`Validated`'s `Bitraverse` instances with `Copair` instances.

Jacoby6000:
@adelbertc I'm not sure what you mean by

```scala
left(left(x, y), z) ~= left(x, left(y, z))
```

left and right only take a single parameter.

adelbertc:
Ah yes good point

adelbertc:
I wonder if you can also define an instance for `Coproduct`.

This LGTM, I've had a use case for this before I think, though I have (purely aesthetic) reservations about the name Copair.

Jacoby6000:
I'm not married to the name Copair, it's just what @tpolecat recommended. There's still stuff missing here though. If you squint at it, you can derive a Monad from Copair. I've also not yet defined the typeclasses for checking equality. 

Another question, how to I get code coverage on the *Ops classes?  I followed the standard the other laws were following, but they seem to test the typeclass directly, and miss the Ops classes.

tpolecat:
I just made up `Copair`, I have no strong opinion.

Jacoby6000:
@adelbertc [Looks like Haskell has a Copair](https://hackage.haskell.org/package/TypeCompose-0.9.11/docs/Data-Pair.html#t:Copair) and it (from what I can tell) is different. So a different name should probably be here

Jacoby6000:
@adelbertc after my (Short) investigation, it looks like you cannot define an instance for `Coproduct`.

Whenever I tried to define fold (`F[A,B] => (A => C, B => C) => C`), I began to have trouble.

I've added a `to[G[_,_]: Copair]` function for converting between `Copair`s. I'm not sure what else can be added, but there very well may be more.

I'm confident with `Copair` where it sits now, but please feel free to suggest a new name, also please reference my link to hackage above, because I can't read Haskell.  If Haskell's `Copair` is different from this, then we should not name it `Copair`.

Jacoby6000:
Also something I just noticed, `Validated` appears to only have an instance for `Bifunctor`. Adapting `Validated` to use `Copair` would mean that `Validated` will now have a `Bitraverse` instance.  Is this a problem?

Jacoby6000:
Updated branch and resolved conflicts

Jacoby6000:
I don't quite understand the codecov/changes error.  Can I have some guidance on how to fix that?

ceedubs:
Thanks for putting this together @Jacoby6000.

I'm not sure what to think about this. Since `Xor` and `Either` are effectively the same thing, I don't know that I'd put a lot of value in an abstraction over the two of them - you'd just be encoding a specific data structure as a type class. The fact that this abstraction also covers `Validated` makes it much more interesting to me. However, it's troublesome to me that (as @Jacoby6000 has mentioned) `Copair` _could_ extend `Monad` without adding any abstract methods, and that would result in a lawful `Monad` for `Xor` but not for `Validated`. To me that makes it seem like something is off here.

Jacoby6000:
It may be that this construct doesnt belong in cats, but maybe something that extends cats in some way. Sometimes when developing libs, it's useful to accept/manipulate whatever Either construct somebody wants to use. Perhaps I can look at throwing this in some utility library. Perhaps it'd be a nice addition to [Daniel's "Shims"](https://github.com/djspiewak/shims)

kcsongor:
It looks to me like this type-class captures the notion of the sum of two types `A` and `B` (i.e. anything that is isomorphic to `Either[A, B]`, with `A` and `B` not necessarily parametric, see `Option` as an example below).

As @ceedubs said, `Xor` and `Either` are effectively the same thing, but I would argue that so is `Validated` - at least at the algebraic level. The data structures are all the same, the difference lies in the specific type-class instances of these structures, and in fact, this is the reason why multiple versions of the same data structure exist. (by 'data-structure', I mean the sum referred to above)

So your `Copair` (which I would call `Sum`, by the way) captures this specific algebraic structure. Now, `Option[A]`'s algebraic structure is isomorphic to `Either[(), A]`, and so it could also be made a `Copair`, with a type lambda that discards one of the arguments. (<- I'm not sure how this would play, I'm very new to scala)

Usually, the notion of abstracting over algebraic properties of ADTs (algebraic data types) is called 'generic' programming, and that is what the shapeless library is concerned with (https://github.com/milessabin/shapeless/blob/master/core/src/main/scala/shapeless/ops/coproduct.scala for your specific use-case)
I think this is a good example of using generic programming to derive type-class instances, which is in turn what the kittens library does (using shapeless).

Jacoby6000:
@kcsongor I opened a PR in the cats repo.  I feel like `Sum` is too general for this type.  A Sum can usually contain an arbitrary number of members, `shapeless.Coproduct` might be more of a `Sum` than this.  In haskell there appears to be a `Copair`, which from what I can tell is identical to this.

kcsongor:
Hm, what I meant was that kittens already has this abstraction, as this describes a binary sum of types. Of course, if you know how to represent the sum of two types, then you know how to represent the sum of n types, neither is more general. `Either[A, Either[B, Either[C, D]]]` is a long-winded way of writing the sum of 4 types. You might have to squint a bit to see how shapeless's coproduct is talking about the same thing, as it involves some extra machinery to support a more convenient type-level representation of the sums.

This is irrelevant, but: the link you posted earlier for Haskell is not quite the same thing. The docs say that Copair is dual to Unpair, where if Unpair eliminates a pair in a context, then Copair introduces one. The example states that this is useful for contravariant functors. That's because flipping the arrows of the product projections via some contravariant functor f, from `(a, b) -> a` and `(a, b) -> b`, you get exactly `f a -> f (a, b)` and `f b -> f (a, b)` respectively. Of course there are some similarities, because at the end of the day, all sums are created equal!

Jacoby6000:
@kcsongor I'm aware of what coproducts are and do.  In short, my vision of this `Copair` is a generalization to unify all coproducts with an arity of 2, because there are so many different ones in existence.. `Validated` and `Xor` are both coproducts, but they have different effects.  The typeclass I've created allows you to accept either one of those, provided you don't want behaviors specific to those two specific coproducts for example.  

The fact that this is a generalization of all coproducts with arity 2 might be seen more clearly here: 

```scala
  type ShapelessCoproduct2[A, B] = A :+: B :+: CNil
  implicit val shapelessCoproductCopair: Copair[ShapelessCoproduct2] =
    new Copair[ShapelessCoproduct2] {
      def fold[A, B, C](f: CopairInstances.this.ShapelessCoproduct2[A,B])(fa: A => C,fb: B => C): C =
        f match {
          case Inl(a) => fa(a)
          case Inr(Inl(b)) => fb(b)
          case Inr(Inr(_)) => sys.error("impossible")
        }

      def left[A, B](a: A): CopairInstances.this.ShapelessCoproduct2[A,B] = Coproduct[ShapelessCoproduct2[A,B]](a)
      def right[A, B](b: B): CopairInstances.this.ShapelessCoproduct2[A,B] = Coproduct[ShapelessCoproduct2[A,B]](b)
     }
```

kcsongor:
"The typeclass I've created allows you to accept either one of those, provided you don't want behaviors specific to those two specific coproducts for example." - this is what data-type generics programming is for. Shapeless is able to derive this binary coproduct for both Either and Xor, I suggest you to look up how the library works, quite interesting!
Your instance is not a proof that copair is more general, that would only be the case if you couldn't achieve with coproduct what you can with copair, but I think you can in all cases.

Jacoby6000:
I'm very familiar with shapeless, and I'm not at all arguing that my `Copair` is more general.  My `Copair` is more specific.  And because it's more specific, you can leverage things that you get because you have an Arity of 2, like `bitraverse` and `bifunctor`.  But then also `left[A](a: A)` and  `right[B](b: B)` which allow you to create "lefts" and "rights", which `bifunctor` and `bitraverse` do not give you.

kcsongor:
I'm confused about what the added value here is. Would this be for people who essentially need the coproduct from shapeless, but prefer to write the instance by hand and/or don't want to add it to their classpath?

kcsongor:
Could you elaborate on the arity argument? For the record, coproducts are binary in shapeless.

Jacoby6000:
@kcsongor [here's an example use of a `Copair`](https://gist.github.com/Jacoby6000/f47b664566c7cd6794d90e2016b5f99b). 

It allows the caller of `JsonParser` to pick what kind of `Copair` they want back.  If they primarily use eithers, they can ask for eithers, if they want a `Validated` because they're trying to do error accumulation they can do that.

Shapeless `Coproduct` is N-ary and is modeled as a singly linked list, just like `HList`. The difference being, HList contains an element for each type it contains, `Coproducts` contain only one element which may represent any of the N types.  In the [shapeless example](https://github.com/milessabin/shapeless/wiki/Feature-overview:-shapeless-2.0.0#coproducts-and-discriminated-unions), there is a `Coproduct` with arity 3.  

When you say "shapeless `Coproduct` is binary", are you sure you're not mixing it up with cats' Coproduct?  They're very different things.

kcsongor:
@Jacoby6000 
"In the shapeless example, there is a coproduct with arity 3". That is not exactly true.
I can say `type :+:[A, B] = Copair[A, B]`
Do you agree that that type alias didn't turn your Copair into more of an n-ary sum than it was before?

`type ISB = Int :+: String :+: Boolean :+: CNil` <- this can be written, however, in terms of Copair, but is still the ordinary binary coproduct, just a little bit in disguise. It seems to me that what confused you was the infix type alias.

As for you JSON example, I can only repeat what I've already said, that this is what data-type generic programming is, and it is precisely what shapeless is for. For this specific example, shapeless's coproduct.
Actually, you have put the coproduct in your example. But notice how you could write the 'parse' function in terms of coproduct as well. Then what you need, is a way of going back and forth between, say, `Either[A, B]` and `A :+: B :+: CNil`. This conversion is automatically provided by shapeless. It essentially fills in your `Copair` type class without asking the user to do so manually.

`Coproduct` is modeled as a list *at the type level*. At the term level, however, it is just a simple value. This is like saying that `Either[A, B]` is modeled as a tuple. Which is true, *at the type level*.

Jacoby6000:
I understand and agree with you on all fronts. I don't understand what conclusion you are making though.

kcsongor:
My point is that if you're the author of the library, you could write your functions with shapeless, and then the user wouldn't have to instantiate any type-class, just use their Xor or Validated, as they like, the transitions would be handled generically. That being said, there is indeed no facility in cats that generalises it.
My conclusion is that data-type generic programming is not a concern of cats, so I doubt this would be justifiable as a one-off.

PR #985, opened 2016-04-17
Add object `reducible` for reducible syntax

markus1189:
Seems like there is no object for `cats.syntax.reducible` ;)


adelbertc:
Hm did we just have the trait there and have no corresponding object for it? That's strange.. 👍

fthomas:
:+1:

Issue #965, opened 2016-04-03
Follow up on def vs val in #940


ceedubs:
In #940 some implicit instances that seemed like they should be able to be `val`s were forced to be `def`s. It would be nice to follow up on that - to try to figure out why that was and whether we can do something about it.

cc @djspiewak

djspiewak:
Here's a quick instance where this happens: https://github.com/typelevel/cats/blob/7cdd66136a0ddd13522a3f6492627ec38d657dc0/core/src/main/scala/cats/data/OptionT.scala#L140  Haven't tried in a while, but when I first wrote this line, swapping it from a `def` to a `val` made the implicit it provides unresolvable.

ceedubs:
@djspiewak yes, I saw the same behavior when I didn't believe you and tried to change it to a `val` :P

djspiewak:
@ceedubs Doubter…  ;-)

ceedubs:
This is no longer relevant after #1621.

PR #986, opened 2016-04-18
Add `Bitraverse` instances for `Validated` and `XorT`.

peterneyens:
While experimenting with `MonadCombine.separate` (for #975), I noticed there were no `Bitraverse` instances for `Validated` and `XorT`.

I would welcome suggestions, possible improvements, ... (maybe for `XorTBifoldable` ?)

### DIFF
--- a/core/src/main/scala/cats/data/Validated.scala
+++ b/core/src/main/scala/cats/data/Validated.scala
@@ -2,7 +2,6 @@ package cats
 package data
 
 import cats.data.Validated.{Invalid, Valid}
-import cats.functor.Bifunctor
 
 import scala.reflect.ClassTag
 import scala.util.{Failure, Success, Try}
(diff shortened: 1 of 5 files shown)


core/src/main/scala/cats/data/Validated.scala
```diff
@@ -228,10 +227,25 @@ private[data] sealed abstract class ValidatedInstances extends ValidatedInstance
     def show(f: Validated[A,B]): String = f.show
   }
 
-  implicit def validatedBifunctor: Bifunctor[Validated] =
-    new Bifunctor[Validated] {
-      override def bimap[A, B, C, D](fab: Validated[A, B])(f: A => C, g: B => D): Validated[C, D] = fab.bimap(f, g)
-      override def leftMap[A, B, C](fab: Validated[A, B])(f: A => C): Validated[C, B] = fab.leftMap(f)
+  implicit def validatedBifunctor: Bitraverse[Validated] =
```
ceedubs: It might be good to leave the `override def`s for `bimap` and `leftMap` in since they should be slightly more efficient than the versions derived from `Bitraverse`.

core/src/main/scala/cats/data/XorT.scala
```diff
@@ -185,11 +186,10 @@ private[data] abstract class XorTInstances extends XorTInstances1 {
   implicit def xorTShow[F[_], L, R](implicit sh: Show[F[L Xor R]]): Show[XorT[F, L, R]] =
     functor.Contravariant[Show].contramap(sh)(_.value)
 
-  implicit def xorTBifunctor[F[_]](implicit F: Functor[F]): Bifunctor[XorT[F, ?, ?]] = {
-    new Bifunctor[XorT[F, ?, ?]] {
-      override def bimap[A, B, C, D](fab: XorT[F, A, B])(f: A => C, g: B => D): XorT[F, C, D] = fab.bimap(f, g)
+  implicit def xorTBifunctor[F[_]](implicit F: Traverse[F]): Bitraverse[XorT[F, ?, ?]] =
```
ceedubs: We may want a `Bifunctor` instance for an `XorT` where we have a `Functor[F]` but not a `Traverse[F]` (in fact this comes up quite often with things like `Future`/`Task`). Therefore, I think we should leave this `xorTBifunctor` in place and add a separate `xorTBitraverse` at a separate layer of the `XorTInstances` priority hierarchy (to avoid ambiguous instances).

ceedubs:
@peterneyens thanks!

In general this looks good. I'd want to see https://github.com/typelevel/cats/pull/986/files#r60056173 addressed before this is merged.

> maybe for XorTBifoldable

Sounds good to me!

core/src/main/scala/cats/data/Validated.scala
ceedubs: A couple other notes:

I know this was already a `def`, but since I noticed it in the diff: could we switch it to a `val` to avoid repeated allocations? Or does it hit the #965 strangeness?

Are we keeping the name `validatedBifunctor` for compatibility reasons? If so, we should probably leave a comment that that's what we are doing. However, I think we can feel free to change this to `validatedBitraverse` since we won't really start to focus on compatibility until a 1.0 release (hopefully coming soon!).

core/src/main/scala/cats/data/XorT.scala
```diff
@@ -65,6 +65,9 @@ final case class XorT[F[_], A, B](value: F[A Xor B]) {
 
   def bimap[C, D](fa: A => C, fb: B => D)(implicit F: Functor[F]): XorT[F, C, D] = XorT(F.map(value)(_.bimap(fa, fb)))
 
+  def bitraverse[G[_], C, D](f: A => G[C], g: B => G[D])(implicit traverseF: Traverse[F], bitraverseXor: Bitraverse[Xor], applicativeG: Applicative[G]): G[XorT[F, C, D]] =
```
ceedubs: We shouldn't need to take `Bitraverse[Xor]` as a parameter, because that's already defined within cats, right?

peterneyens: Yeah, that felt strange. I wrote it this way because of the `traverseXorA` parameter in [`XorT.traverse`](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/XorT.scala#L99).
Can I just write `Bitraverse[Xor].bitraverse(axb)(f, g)` directly ?
Can I do the same with [`Bifoldable[Xor]`](https://github.com/typelevel/cats/pull/986/files#diff-443b0c0bbff3ee4ef3a4b8805699f07cR343) in `XorTBifoldable` ?

ceedubs: Seems reasonable to me. I don't know why we have `traverseXorA` in the `XorT.traverse` signature. I would think we could remove that too.

ceedubs:
Thanks again @peterneyens, especially for the cleanup you did outside of your initial PR. 👍 as long as the build goes green.

Jacoby6000:
If #976 gets merged, it gives you a bitraverse for these types for free (but I didn't clean things up)

non:
This looks good to me. 👍 

Unfortunately there are merge conflicts now. Sorry!

@peterneyens would you be able to merge and update? If not, I can give it a shot (I imagine the merge will be relatively easy).

peterneyens:
It was a simple merge. The conflict was just me renaming `xorBifunctor` to `xorBitraverse`.

non:
Works for me. 👍

PR #988, opened 2016-04-18
Use by-name parameter in Applicatve.replicateA

markus1189:
Changes `Applicative.replicateA` such that the `fa` parameter is by-name instead of by-value.

```
def replicateA[A](n: Int, fa: F[A]): F[List[A]] = sequence(List.fill(n)(fa))    // before
def replicateA[A](n: Int, fa: => F[A]): F[List[A]] = sequence(List.fill(n)(fa)) // after
```

I think this change deserves some discussion, because in an ideal world, there should not be implicit side effects and therefore it should not matter whether `fa` is by-name or by-value.

However I recently encountered a problem with scala's `Future` and the available `Applicative` instance in `cats`:

```
import cats.Applicative
import cats.std.future._
import scala.concurrent._
import scala.concurrent.duration._

Await.result(Applicative[Future].replicateA(3,Future(println("Hello World"))), Duration.Inf)
```

Which results in the following output after `:paste` in the REPL:

```
Hello World
import cats.Applicative
import cats.std.future._
import scala.concurrent._
import scala.concurrent.duration._
f: scala.concurrent.Future[Unit]
res1: List[Unit] = List((), (), ())
```

As you can see, the `Hello World` is printed only once, because the argument to `replicateA`, `fa`, is evaluated **before** passing it to `List.fill` (which itself does use a by-name param), so `Hello World` is printed only once, during the evaluation of the arguments before calling `Applicative.replicateA`.

To fix this, I think we should change `replicateA` to take the second parameter by-name.  Any other suggestions?

### DIFF
--- a/core/src/main/scala/cats/Applicative.scala
+++ b/core/src/main/scala/cats/Applicative.scala
@@ -35,7 +35,7 @@ import cats.std.list._
   /**
     * Given `fa` and `n`, apply `fa` `n` times to construct an `F[List[A]]` value.
     */
-  def replicateA[A](n: Int, fa: F[A]): F[List[A]] =
+  def replicateA[A](n: Int, fa: => F[A]): F[List[A]] =
     sequence(List.fill(n)(fa))
 
   /**
--- a/tests/src/test/scala/cats/tests/ApplicativeTests.scala
+++ b/tests/src/test/scala/cats/tests/ApplicativeTests.scala
@@ -2,7 +2,9 @@ package cats
 package tests
 
 import cats.Applicative
-
+import scala.concurrent.Future
+import scala.concurrent.duration._
+import scala.concurrent.Await
 
 class ApplicativeCheck extends CatsSuite {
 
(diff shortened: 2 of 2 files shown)


adelbertc:
The behavior you're seeing is because the constructed `Future` side-effects - if this was in an `IO` or `Task`-y type the behavior would be what is expected.

markus1189:
Okay, so don't change this then?  I am aware that `Future` in Scala is weird, but given that `cats` has the right type class instances, the above behavior is still surprising.

adelbertc:
Yeah a lot of the times `s.c.Future` is used with side-effects, but the semantics have nothing to say about it - e.g.

```scala
for {
  x <- future()
  y <- future()
} yield (x, y)
```

is different from

```scala
val f = future()
for {
  x <- f
  y <- f
} yield (x, y)
```

when `future()` side-effects. In a pure computation though everything is OK. The type class instances for it assume that the computation is pure - once it starts side effecting, as with any other "function," all bets are off.

cc @tpolecat to confirm my statements

markus1189:
Something to keep in mind, but I'm okay with not fixing this too.  Closing.

PR #989, opened 2016-04-20
Remove internal lazy val from Later.

non:
This commit replaces the previous Later implementation
with one that is not based on Scala's lazy val encoding.

Concretely, this new implementation uses a private,
internal lock to avoid potential deadlocks that could
arise if other objects synchronized on a Later instance.
The possibility of this happening is relatively remote
but not impossible.

Performance seems to be equivalent to lazy vals in the
cases that were tested. An additional Object is
allocated, so Later instance may be slightly larger.

Is this a good idea? Should we stick with `lazy val`?
Should we synchronize on `this` to have smaller
instances? Should we be doing something else?

### DIFF
--- a/core/src/main/scala/cats/Eval.scala
+++ b/core/src/main/scala/cats/Eval.scala
@@ -3,6 +3,8 @@ package cats
 import scala.annotation.tailrec
 import cats.syntax.all._
 
+import java.io.{ Externalizable, ObjectOutput, ObjectInput }
+
 /**
  * Eval is a monad which controls evaluation.
  *
@@ -134,23 +136,57 @@ final case class Now[A](value: A) extends Eval[A] {
  * by the closure) will not be retained, and will be available for
  * garbage collection.
  */
-final class Later[A](f: () => A) extends Eval[A] {
-  private[this] var thunk: () => A = f
+final class Later[A](f: () => A) extends Eval[A] with Externalizable {
 
-  // The idea here is that `f` may have captured very large
-  // structures, but produce a very small result. In this case, once
-  // we've calculated a value, we would prefer to be able to free
-  // everything else.
-  //
-  // (For situations where `f` is small, but the output will be very
-  // expensive to store, consider using `Always`.)
-  lazy val value: A = {
-    val result = thunk()
-    thunk = null // scalastyle:off
-    result
-  }
+  // we use lock to prevent possible problems with someone else
+  // synchronizing on 'this'.
+  private[this] val lock = new Object()
+  private[this] var thunk: Function0[A] = f
+  private[this] var result: A = _
+
+  final def value: A =
+    if (thunk == null) {
+      result
+    } else lock.synchronized {
+      if (thunk == null) result else {
+        result = thunk()
+        thunk = null
+        result
+      }
+    }
 
   def memoize: Eval[A] = this
+
+  // everything below here is necessary to convince java to serialize
+  // these things.
+
+  def this() =
+    this(() => null.asInstanceOf[A])
+  
+  def writeExternal(out: ObjectOutput): Unit =
+    thunk match {
+      case null =>
+        out.writeObject(new Integer(1))
+        out.writeObject(result)
+      case t =>
+        out.writeObject(new Integer(0))
+        out.writeObject(t)
+    }
+
+  def readExternal(in: ObjectInput): Unit =
+    in.readObject() match {
+      case n: Integer =>
+        n.intValue match {
+          case 0 =>
+            thunk = in.readObject().asInstanceOf[Function0[A]]
+            result = null.asInstanceOf[A]
+          case 1 =>
+            thunk = null
+            result = in.readObject().asInstanceOf[A]
+        }
+      case x =>
+        throw new RuntimeException(s"expected integer, got $x")
+    }
 }
 
 object Later {
--- a/core/src/main/scala/cats/Eval.scala
+++ b/core/src/main/scala/cats/Eval.scala
@@ -141,7 +141,7 @@ final class Later[A](f: () => A) extends Eval[A] with Externalizable {
   // we use lock to prevent possible problems with someone else
   // synchronizing on 'this'.
   private[this] val lock = new Object()
-  private[this] var thunk: Function0[A] = f
+  @volatile private[this] var thunk: Function0[A] = f
   private[this] var result: A = _
 
   final def value: A =
@@ -162,7 +162,7 @@ final class Later[A](f: () => A) extends Eval[A] with Externalizable {
 
   def this() =
     this(() => null.asInstanceOf[A])
-  
+
   def writeExternal(out: ObjectOutput): Unit =
     thunk match {
       case null =>
--- a/bench/src/main/scala/cats/bench/TrampolineBench.scala
+++ b/bench/src/main/scala/cats/bench/TrampolineBench.scala
@@ -30,11 +30,12 @@ class TrampolineBench {
       y <- Trampoline.suspend(trampolineFib(n - 2))
     } yield x + y
 
-  // TailRec[A] only has .flatMap in 2.11.
-
+  // // TailRec[A] only has .flatMap in 2.11.
+  // import scala.util.control.TailCalls
+  // 
   // @Benchmark
   // def stdlib(): Int = stdlibFib(N).result
-  //
+  // 
   // def stdlibFib(n: Int): TailCalls.TailRec[Int] =
   //   if (n < 2) TailCalls.done(n) else for {
   //     x <- TailCalls.tailcall(stdlibFib(n - 1))
--- a/bench/src/main/scala/cats/bench/TrampolineBench.scala
+++ b/bench/src/main/scala/cats/bench/TrampolineBench.scala
@@ -32,7 +32,7 @@ class TrampolineBench {
 
   // // TailRec[A] only has .flatMap in 2.11.
   // import scala.util.control.TailCalls
-  // 
+  //
   // @Benchmark
   // def stdlib(): Int = stdlibFib(N).result
   // 
--- a/bench/src/main/scala/cats/bench/TrampolineBench.scala
+++ b/bench/src/main/scala/cats/bench/TrampolineBench.scala
@@ -35,7 +35,7 @@ class TrampolineBench {
   //
   // @Benchmark
   // def stdlib(): Int = stdlibFib(N).result
-  // 
+  //
   // def stdlibFib(n: Int): TailCalls.TailRec[Int] =
   //   if (n < 2) TailCalls.done(n) else for {
   //     x <- TailCalls.tailcall(stdlibFib(n - 1))
--- a/tests/src/test/scala/cats/tests/EvalTests.scala
+++ b/tests/src/test/scala/cats/tests/EvalTests.scala
@@ -1,6 +1,7 @@
 package cats
 package tests
 
+import catalysts.Platform
 import scala.math.min
 import cats.laws.ComonadLaws
 import cats.laws.discipline.{CartesianTests, BimonadTests, SerializableTests}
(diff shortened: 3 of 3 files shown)


core/src/main/scala/cats/Eval.scala
```diff
@@ -134,23 +136,57 @@ final case class Now[A](value: A) extends Eval[A] {
...
+  // synchronizing on 'this'.
+  private[this] val lock = new Object()
+  private[this] var thunk: Function0[A] = f
+  private[this] var result: A = _
+
+  final def value: A =
+    if (thunk == null) {
+      result
+    } else lock.synchronized {
+      if (thunk == null) result else {
```
ceedubs: I think that double-checked locking on the JVM requires a `volatile` var because of...reasons. http://stackoverflow.com/questions/7855700/why-is-volatile-used-in-this-example-of-double-checked-locking hits on this. You are definitely more of a JVM expert than I am though, so I feel like you could probably convince me that this isn't a concern.

ceedubs:
I imagine that this might have been inspired in part by https://github.com/scalaz/scalaz/pull/1145 ? If so, I think that the `lazy val` in `Later` is significantly less of a concern than the places addressed in that PR (which was inspired by some profiling we had done at work). In that case, lazy vals were being used for evaluate-once semantics on method-level values, causing the entire object to be locked for something that really only needed to apply to a particular call of a particular method. For example, a call to `Apply[Option].apply7` would break down into a bunch of `ap` calls on the (singleton) `Apply[Option]` instance, so the lazy vals in those `ap` methods would end up hitting a lot of contention on the locking of the `Apply[Option]` instance.

In the case of `Later`, I think object-level locking is roughly the desired execution.

It's quite possible that this is still a good idea, and I definitely think you would be a better judge of that than I. I just wanted to bring to attention the difference between `Later` and the changes made in scalaz.

core/src/main/scala/cats/Eval.scala
non: So I think the reason this isn't a problem is because we are only setting `thunk = null` after `result = thunk()` has fully initialized the `result: A` value. This means that we don't have to worry about someone else returning a partially-evaluated `result` value. I could be wrong though. I looked at this code with @johnynek -- maybe he can weigh in here too?

non:
@ceedubs I would be comfortable moving the locking to `this` if everyone agrees that locking `Later` is acceptable and are not worried about cases where a user haplessly does their own locking. I'd rather remove `lazy val` in any case since we can do the encoding more directly.

xuwei-k:
https://gist.github.com/xuwei-k/4be690a606bec83683563cfff525b3a5

ceedubs:
@xuwei-k wow, super helpful. Thanks!

non:
Thanks!

johnynek:
I think if you check result != null this works.

On Wednesday, April 20, 2016, Erik Osheim <notifications@github.com> wrote:

> Thanks!
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/typelevel/cats/pull/989#issuecomment-212507416>
>


-- 

johnynek:
Actually result needs to volatile. See this paper:

http://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html

The solution at the end hat works with Java 5 (I think we can skip Java 4
support).

On Wednesday, April 20, 2016, P. Oscar Boykin <oscar.boykin@gmail.com>
wrote:

> I think if you check result != null this works.
>
> On Wednesday, April 20, 2016, Erik Osheim <notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>> wrote:
>
>> Thanks!
>>
>> —
>> You are receiving this because you were mentioned.
>> Reply to this email directly or view it on GitHub
>> <https://github.com/typelevel/cats/pull/989#issuecomment-212507416>
>>
>
>
> --
> P. Oscar Boykin, Ph.D. | http://twitter.com/posco |
> http://pobox.com/~boykin
>
>

-- 

johnynek:
Just to be clear, locking `this` can cause a deadlock. I don't see how the correct `volatile` solution has that risk. The risk is small (some external caller has to lock the `Eval[T]` instance and another object while the `Lazy` locks that same object), but I'd prefer it to be zero.

non:
Thanks for the example @xuwei-k! I added this as a test case.

ceedubs:
@non I see that you have gone to some length to ensure that this is serializable. Is that being checked by a test, or should we be adding one?

It's been mentioned by both @non and @johnynek that the chances of deadlock from locking on `this` are pretty remote. Considering that it's quite remote and people are kind of saying that they know what they are doing if they start locking on the instance, I'd be inclined to lock on `this` and avoid the extra `Object` allocation (which lingers as long as the `Eval` instance exists, increasing its memory footprint as far as I can tell). However, @non and @johnynek are much more performance-conscious than I am and seem to like this approach, so 👍 by me.

johnynek:
the serialization was exercised by the semigroup serialization check (indeed that is why we added the code).

I hear you that it is pretty defensive to worry about locks on `this`. I don't mind locking `this` I guess. I would never lock an object I didn't create since I have no idea who else might lock it, so I guess it is clearly dangerous, and this is scala, so we assume programmers know about sharp edges and avoid them.

core/src/main/scala/cats/Eval.scala
```diff
@@ -134,23 +136,57 @@ final case class Now[A](value: A) extends Eval[A] {
...
+
+  // we use lock to prevent possible problems with someone else
+  // synchronizing on 'this'.
+  private[this] val lock = new Object()
+  @volatile private[this] var thunk: Function0[A] = f
+  private[this] var result: A = _
+
+  final def value: A =
+    if (thunk == null) {
+      result
```
mikejcurry: I think there might be a nasty corner case here where both `thunk` and `result` evaluate to `null` from a secondary threads perspective, but would definitely love to find out I'm wrong.
Consider:
- Thread A enters `value` method and inspects `thunk`, seeing `thunk` as non `null`
- Thread A enters synchronized block
- Thread A inspects `thunk` again, and sees it as non `null`
- Thread A enters else branch and evaluates `thunk`, assigning the calculated `A` to `result`
- Thread A assigns `null` to `thunk`
- Thread B now enters `value` method and inspects `thunk`
- Thread B sees `thunk` as `null` due to volatility. The value has been written in Thread A as `null`.
- Thread B returns the value of `result`

I _think_ that it is possible for the Thread B `result` variable to still point to a `null` value, briefly in this scenario because I don't think either the volatility of `thunk` or the `synchronized` block which this code block does not take place in, make any guarantees about which value Thread B might see as the value in `result`. If I'm wrong here it would be great to find out why as it would improve my understanding of the JVM.

johnynek: yeah, I think that's a real issue. I think switching the `volatile` to
result and checking that would be safe, but requires a cast since `A >:
Null` is not guaranteed (but should be safe on the jvm).

On Mon, May 2, 2016 at 8:50 PM, Mike Curry <notifications@github.com> wrote:

> In core/src/main/scala/cats/Eval.scala
> <https://github.com/typelevel/cats/pull/989#discussion_r61843443>:
>
> > -  lazy val value: A = {
> > -    val result = thunk()
> > -    thunk = null // scalastyle:off
> > -    result
> > -  }
> > +final class Later[A](f: () => A) extends Eval[A] with Externalizable {
> > +
> > +  // we use lock to prevent possible problems with someone else
> > +  // synchronizing on 'this'.
> > +  private[this] val lock = new Object()
> > +  @volatile private[this] var thunk: Function0[A] = f
> > +  private[this] var result: A = _
> > +
> > +  final def value: A =
> > +    if (thunk == null) {
> > +      result
>
> I think there might be a nasty corner case here where both thunk and
> result evaluate to null from a secondary threads perspective, but would
> definitely love to find out I'm wrong.
> Consider:
>
>    - Thread A enters value method and inspects thunk, seeing thunk as non
>    null
>    - Thread A enters synchronized block
>    - Thread A inspects thunk again, and sees it as non null
>    - Thread A enters else branch and evaluates thunk, assigning the
>    calculated A to result
>    - Thread A assigns null to thunk
>    - Thread B now enters value method and inspects thunk
>    - Thread B sees thunk as null due to volatility. The value has been
>    written in Thread A as null.
>    - Thread B returns the value of result
>
> I *think* that it is possible for the Thread B result variable to still
> point to a null value, briefly in this scenario because I don't think
> either the volatility of thunk or the synchronized block which this code
> block does not take place in, make any guarantees about which value Thread
> B might see as the value in result. If I'm wrong here it would be great
> to find out why as it would improve my understanding of the JVM.
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/typelevel/cats/pull/989/files/39828830bb958ff9bc775ed34897e72d3f59a7c6#r61843443>
>



-- 

mikejcurry: Yea, agree - switching the `volatile` to `result` should close any loops. I don't think you need a cast though - at least not an explicit one - this should be fine.

```scala
 final class Later[A](f: () => A) extends Eval[A] {

   // we use lock to prevent possible problems with someone else
   // synchronizing on 'this'.
   private[this] val lock = new Object()
   private[this] var thunk: Function0[A] = f

   @volatile private[this] var result: A = _

   final def value: A =
     if (result != null) {
       result
     } else lock.synchronized {
       if (result != null) result else {
         result = thunk()
         thunk = null
         result
       }
     }

   def memoize: Eval[A] = this
 }
```
BTW - here is an alternate encoding. Not sure what would be better - but it is an interesting problem so I was thinking of a few different approaches and this one seemed interesting.

```scala
 final class Later[A](f: () => A) extends Eval[A] {

   private[this] var thunk: () => A = f

   private[this] trait Result {
     def value: A
   }

   private[this] val evaluated: Result = new Result {
     lazy val value = {
       val result = thunk()
       thunk = null
       result
     }
   }

   final def value: A = evaluated.value

   def memoize: Eval[A] = this
 }
```
This encoding avoids explicit use of `volatile` or `synchronized`, and moves the lock to an inner type so should avoid the `synchronization` problem mentioned. I'm not sure how the cost of the additional reference to the `lazy val `in the inner type would affect performance, I guess it would need to be measured.

johnynek: I guess in the case where we check `result != null` we also need to throw an NPE or something if `thunk() == null`.

mikejcurry: I'm not sure I understand.` thunk() == null` should never be `true`, as it is the evaluation of a `Function0[A] `- which evaluates to an `A`. Well - at least no more than any other scala function with a return type of `A` can evaluate to `null`. So, I think `thunk() == null` should not be considered here.

The reference to `thunk` can equal `null`, i.e. `thunk == null`, but there is only one dereference of `thunk` to and this is in the `synchronized` block, which is only entered when `result == null`. 

So, I think that the `volatile` nature of `result` would make that safe.

johnynek: I agree one should not return null, but if they do, this code will first
return null, but on a subsequent call would NPE (you would try to call
thunk again, but it would be null).

To me, this behavior sounds no good. I'd rather NPE the first time around
hen returning null once then failing if you happen to get the value again.

On Wednesday, May 4, 2016, Mike Curry <notifications@github.com> wrote:

> In core/src/main/scala/cats/Eval.scala
> <https://github.com/typelevel/cats/pull/989#discussion_r62152850>:
>
> > -  lazy val value: A = {
> > -    val result = thunk()
> > -    thunk = null // scalastyle:off
> > -    result
> > -  }
> > +final class Later[A](f: () => A) extends Eval[A] with Externalizable {
> > +
> > +  // we use lock to prevent possible problems with someone else
> > +  // synchronizing on 'this'.
> > +  private[this] val lock = new Object()
> > +  @volatile private[this] var thunk: Function0[A] = f
> > +  private[this] var result: A = _
> > +
> > +  final def value: A =
> > +    if (thunk == null) {
> > +      result
>
> I'm not sure I understand.thunk() == null should never be true, as it is
> the evaluation of a Function0[A]- which evaluates to an A. Well - at
> least no more than any other scala function with a return type of A can
> evaluate to null. So, I think thunk() == null should not be considered
> here.
>
> The reference to thunk can equal null, i.e. thunk == null, but there is
> only one dereference of thunk to and this is in the synchronized block,
> which is only entered when result == null.
>
> So, I think that the volatile nature of result would make that safe.
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/typelevel/cats/pull/989/files/39828830bb958ff9bc775ed34897e72d3f59a7c6#r62152850>
>


-- 

mikejcurry: I agree. That would be nasty.

non:
I'm going to rescind this for now, until @johnynek and I have a chance to think about this more and do more performance testing. I think the issue it's solving (what if someone locks a `Later` instance) are unlikely to be a problem for most people (especially compared to the "what if someone locks their own class that has a lazy val" problem).

PR #990, opened 2016-04-23
Separate free package into its own module

ceedubs:
It's quite possible that people are going to throw tomatoes at me for
proposing this change. There has been much discussion about
modularization in the past, and in fact free used to have its own
module, which was removed in #760/#765.

One significant change that has happened since then is that `State` has
started to use `Eval` instead of `Trampoline` for trampolining. At this
point, nothing outside of the `free` package in `cats-core` depends on
`free` (which is why this PR was so easy to create). To me that makes it
a fairly convenient border for modularization.

There was a [brief discussion](https://gitter.im/typelevel/cats?at=5716322b548df1be102e3f3c) about this on Gitter in which @sellout supported this change on the basis that they prefer to use fixpoint types (such as `Mu`) for free structures. Putting `free` into a separate module may be better for people who want to use different variations of these structures.

Of course in addition to these more recent updates, there are the usual
arguments for modularization and decreased jar size (which I believe is
especially important for `catsJS` users).


ceedubs:
I'm particularly interested in feedback from @mpilquist, @travisbrown, @stew, @refried, and @milessabin, as they had previously approved of pulling `free` into `core` (but a bit has changed since then as mentioned in the description above).

milessabin:
I think that `Eval` taking over from `Trampoline` makes a world of difference. This gets a :+1: from me.

ceedubs:
Another item that I meant to mention is #845. Basically there are many more free structures that could potentially be added, and if we choose to go down that path we might not want to add too much "bloat" to core.

julien-truffaut:
What about moving `free` to it's own project? it seems there is quite a lot of development in this area and it might be easier to experiment without being locked with cats releases.

adelbertc:
:+1: from me

adelbertc:
Woah there -2.61%

mpilquist:
:+1:, either for a separate module or even a separate repo.

ceedubs:
@adelbertc `validateJVM` wasn't running the `free` tests. I've fixed it, and codecov now shows no difference in coverage (though it didn't post an updated comment).

I think I'm going to add some notes in the relevant documentation about this being in a separate module and then merge this. I'll post a follow-up issue about whether it should be in its own project, but I don't want that indecision to keep it in core when it sounds like people are agreed that it shouldn't be there.

PR #991, opened 2016-04-23
Replace  `~>` with `NaturalTransformation`

markus1189:
After having read `NaturalTransformation.scala` several times for reference, I found it much easier to read by replacing `NaturalTransformation[A,B]` with the much better `A ~> B`.

This pr replaces the occurrences where possible in `NaturalTransformation.scala`, `Free.scala` and `FreeApplicative.scala`. 

I am sure there are more places where you could do this, but I first wanted to check if this change is welcome at all ;)

### DIFF
--- a/docs/src/main/tut/freemonad.md
+++ b/docs/src/main/tut/freemonad.md
@@ -165,12 +165,15 @@ DSL. By itself, this DSL only represents a sequence of operations
 
 To do this, we will use a *natural transformation* between type
 containers.  Natural transformations go between types like `F[_]` and
-`G[_]` (this particular transformation would be written as `F ~> G`).
+`G[_]` (this particular transformation would be written as
+`NaturalTransformation[F,G]` or as done here using the symbolic
+alternative as `F ~> G`).
 
 In our case, we will use a simple mutable map to represent our key
 value store:
 
 ```tut:silent
+import cats.arrow.NaturalTransformation
 import cats.{Id, ~>}
 import scala.collection.mutable
 
@@ -241,7 +244,7 @@ recursive structure by:
 This operation is called `Free.foldMap`:
 
 ```scala
-final def foldMap[M[_]](f: S ~> M)(M: Monad[M]): M[A] = ...
+final def foldMap[M[_]](f: NaturalTransformation[S,M])(M: Monad[M]): M[A] = ...
 ```
 
 `M` must be a `Monad` to be flattenable (the famous monoid aspect
--- a/core/src/main/scala/cats/arrow/NaturalTransformation.scala
+++ b/core/src/main/scala/cats/arrow/NaturalTransformation.scala
@@ -14,8 +14,8 @@ trait NaturalTransformation[F[_], G[_]] extends Serializable { self =>
   def andThen[H[_]](f: NaturalTransformation[G, H]): NaturalTransformation[F, H] =
     f.compose(self)
 
-  def or[H[_]](h: H ~> G): Coproduct[F, H, ?] ~> G =
-    new (Coproduct[F, H, ?] ~> G) {
+  def or[H[_]](h:  NaturalTransformation[H,G]): NaturalTransformation[Coproduct[F, H, ?],G] =
+    new (NaturalTransformation[Coproduct[F, H, ?],G]) {
       def apply[A](fa: Coproduct[F, H, A]): G[A] = fa.run match {
         case Xor.Left(ff) => self(ff)
         case Xor.Right(gg) => h(gg)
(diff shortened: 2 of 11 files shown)


ceedubs:
@markus1189 thanks for bringing this up!

Personally, I kind of prefer `~>` to `NaturalTransformation` as well. However, I think we've kind of settled on using the non-symbolic form of operators and types even when a symbolic form is provided as an alias. This tends to make it less bewildering and easier to google or ask questions about for newcomers. But I can't seem to find anywhere that we have documented this convention in the contributor guide. In fact, it looks like our contributor guide needs a lot of love. I've just created #993 for this.

markus1189:
That's a good point, also currently there already is a mixture of `NaturalTransformation` and `~>`, both in `NaturalTransformation` and throughout the rest of the code base. 

I agree with you that it might be confusing to see `~>` without knowing what that is supposed to be.  So is the tendency to convert more `~>` to `NaturalTransformation` instead of the other way around?  I think at least inside `NaturalTransformation` having a consistent style would be beneficial (currently e.g. `andThen` uses `NaturalTransformation` while `or` uses `~>`.  I'd be happy to change the pr tomorrow to translate it the other way around then ;)

ceedubs:
@markus1189 I tend to agree with you that it's best to be consistent. I hate to ask you to fill in the hole that you have just taken the time to dig, but I'd be in favor of changing the `~>` references to `NaturalTransformation` within the code. Maybe it's best if at least one other person chimes in with their opinion before you take the time to make such a change?

I think the docs for `FreeMonad` use `~>` as well. If we are going to use it there, we should probably at least point out that it's an alias for `NaturalTransformation`, so `F ~> G` is equivalent to `NaturalTransformation[F, G]`.

markus1189:
Since no one else chimed in, I assume that we go for the replacement of symbolic `~>` with `NaturalTransformation`.  I think I got all of them and added a short note in the free monad tutorial ;)

docs/src/main/tut/freemonad.md
```diff
@@ -162,7 +162,9 @@ DSL. By itself, this DSL only represents a sequence of operations
 
 To do this, we will use a *natural transformation* between type
 containers.  Natural transformations go between types like `F[_]` and
-`G[_]` (this particular transformation would be written as `F ~> G`).
+`G[_]` (this particular transformation would be written as
+`NaturalTransformation[F,G]` or as done here using the symbolic
+alternative as `F ~> G`).
```
markus1189: Here is the note in the tutorial

docs/src/main/tut/freemonad.md
```diff
@@ -238,7 +240,7 @@ recursive structure by:
 This operation is called `Free.foldMap`:
 
 ```scala
-final def foldMap[M[_]](f: S ~> M)(M: Monad[M]): M[A] = ...
+final def foldMap[M[_]](f: NaturalTransformation[S,M])(M: Monad[M]): M[A] = ...
```
ceedubs: The build is failing because `~>` is imported, but `NaturalTransformation` is not. We should probably pick either one or the other for the tutorial and stick to it, but I do like your note about one being an alias for the other.

has609:
> @@ -238,7 +240,7 @@ recursive structure by:
>  This operation is called `Free.foldMap`:
>  
>  ```scala
> -final def foldMap[M[_]](f: S ~> M)(M: Monad[M]): M[A] = ...
> +final def foldMap[M[_]](f: NaturalTransformation[S,M])(M: Monad[M]): M[A] = ...

markus1189:
Whoops, missed that one ;)

ceedubs:
@markus1189 thanks for helping keep us consistent.

I noticed that `~>` is still used in a few places in the docs. Is there a reason behind switching some occurrences over to `NaturalTransformation` but not others?

markus1189:
@ceedubs do you mean the ones in the tutorial for free monads?  I think that it doesn't hurt to have `~>` there, because it is explicitly mentioned and I think most people outside of cats would prefer `~>` anyway, such that it could be a good idea to make readers of the tutorial familiar with the symbolic operator.

Other than that I only found occurrences of `~>` inside of tests which I now also replaced.  Anything else I am not seeing?

ceedubs:
@markus1189 I think I was thrown off by the fact that `object InMemoryDatasourceInterpreter extends NaturalTransformation[DataOp,Id]` shows up in the free tutorial. If we are settling on `~>` for the tutorial, then maybe we should change that to `object InMemoryDatasourceInterpreter extends (DataOp ~> Id)`?

ceedubs:
Sorry but this PR has merge conflicts.

markus1189:
Should be good to go now ;)

ceedubs:
👍 thanks, @markus1189!

non:
I'm 👍 on this change too. Unfortunately it has conflicts!

PR #994, opened 2016-04-24
updated to latest algebra

kailuowang:


non:
This PR looks fine to me. 👍 if we can get Travis to pass (not sure why it failed)

kailuowang:
@non looks like that https://codecov.io/bash is down, causing the second part of the following command to fail 
```
$sbt_cmd coverage validateJVM coverageReport && bash <(curl -s https://codecov.io/bash)"
```

non:
OK, great. We can just hang out and wait for that to finish. Thanks for investigating!

kailuowang:
Looks like codecov is back online. Can we restart the build?

non:
Yeah let's try it again.

kailuowang:
Looks like that this time it failed on (or after) uploading coverage report to S3. Maybe codecov is still having issues.

laws/src/main/scala/cats/laws/discipline/Eq.scala
```diff
@@ -31,17 +32,9 @@ object eq {
       (a: A) => showInstance.show(a)
     }
 
-  // Temporary, see https://github.com/non/algebra/pull/82
-  implicit def tuple2Eq[A, B](implicit A: Eq[A], B: Eq[B]): Eq[(A, B)] =
-    new Eq[(A, B)] {
-      def eqv(x: (A, B), y: (A, B)): Boolean =
-        A.eqv(x._1, y._1) && B.eqv(x._2, y._2)
-    }
+  implicit def tuple2Eq[A, B](implicit A: Eq[A], B: Eq[B]): Eq[(A, B)] = tuple.tuple2Eq
```
adelbertc: We should be able to just remove these right? And do the correct import at the places that use this?

core/src/main/scala/cats/data/WriterT.scala
```diff
@@ -53,8 +53,6 @@ private[data] sealed abstract class WriterTInstances extends WriterTInstances0 {
   implicit def writerTIdMonad[L:Monoid]: Monad[WriterT[Id, L, ?]] =
     writerTMonadWriter[Id, L]
 
-  // The Eq[(L, V)] can be derived from an Eq[L] and Eq[V], but we are waiting
-  // on an algebra release that includes https://github.com/non/algebra/pull/82
   implicit def writerTIdEq[L, V](implicit E: Eq[(L, V)]): Eq[WriterT[Id, L, V]] =
```
adelbertc: Should we change this to `writerTIdEq[L : Eq, V : Eq]` instead of requiring the instance on the tuple

adelbertc:
Can you also add extend the appropriate tuple instances trait like we do https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/std/anyval.scala#L18 so users only need 1 import?

core/src/main/scala/cats/data/WriterT.scala
kailuowang: duh. Will change that way.

laws/src/main/scala/cats/laws/discipline/Eq.scala
kailuowang: I admit I was too lazy. Just updated.

kailuowang:
@adelbertc IRT your last comment about adding the algebra tuple to `cats.std`, the bad news is that the generated `algebra.std.TupleInstances` is package private. Do you want to open it up and release algebra 0.4.1 or shall we have this wait for the next cats release?

non:
@kailuowang let's open that up -- we just released 0.4.0 and i think that is pretty important

kailuowang:
@non males sense. I will send in a PR to algebra soon.

adelbertc:
https://github.com/non/algebra/pull/151

ceedubs:
The build issues should be fixed. I'm going to try to close and reopen this to see if it makes Travis CI happy.

kailuowang:
Don't merge yet, waiting on a new algebra release so we can address @adelbertc's last comment

ceedubs:
Do we have an issue with scala.js versions of algebra? https://travis-ci.org/typelevel/cats/jobs/125562475#L3310-L3313

kailuowang:
@ceedubs, that's due to the fact that I forgot to update the algebra.law version down in the build.sbt.  just pushed in a fix.

non:
Looks good to me. 👍

tests/src/test/scala/cats/tests/StateTTests.scala
```diff
@@ -5,6 +5,7 @@ import cats.laws.discipline.{CartesianTests, MonadStateTests, SerializableTests}
 import cats.data.{State, StateT}
 import cats.laws.discipline.eq._
 import cats.laws.discipline.arbitrary._
+import cats.std.tuple._
```
adelbertc: Is this needed?

adelbertc:
👍 from me, modulo that 1 comment

core/src/main/scala/cats/data/WriterT.scala
```diff
@@ -50,13 +50,14 @@ final case class WriterT[F[_], L, V](run: F[(L, V)]) {
...
 private[data] sealed abstract class WriterTInstances extends WriterTInstances0 {
+
   implicit def writerTIdMonad[L:Monoid]: Monad[WriterT[Id, L, ?]] =
     writerTMonadWriter[Id, L]
 
-  // The Eq[(L, V)] can be derived from an Eq[L] and Eq[V], but we are waiting
-  // on an algebra release that includes https://github.com/non/algebra/pull/82
-  implicit def writerTIdEq[L, V](implicit E: Eq[(L, V)]): Eq[WriterT[Id, L, V]] =
+  implicit def writerTIdEq[L: Eq, V: Eq]: Eq[WriterT[Id, L, V]] = {
+    import algebra.std.tuple.tuple2Eq
```
adelbertc: Can we move this up top with the rest of the imports?

kailuowang:
@adelbertc's comments addressed.

docs/src/main/tut/monoid.md
```diff
@@ -60,23 +60,9 @@ l.foldMap(i => i.toString)
 ```
 
 To use this
-with a function that produces a tuple, we can define a `Monoid` for a tuple 
+with a function that produces a tuple, cats also provides a `Monoid` for a tuple
```
travisbrown: Is this potentially confusing since these actually come from Algebra and the following sentence goes into detail about the role Algebra plays here?

kailuowang: I used cats here because it is available through `cats.std`. and I didn't notice the sentence below. I will add the detail about the instance below as well.

kailuowang: update: N.B. added below. Please help review the wording.

adelbertc:
👍  - if Travis is OK with the comment changes let's merge

ceedubs:
👍 the clarification about algebra tuple instances looks good to me. I'm going to go ahead and merge. @travisbrown if you have suggestions for further improvements it should be easy for us to make a doc update after merging this.

travisbrown:
Sorry for the delay, but yes, +1 from me.

PR #995, opened 2016-04-24
Move coverage away from bash

InTheNow:
Use new command in Travis for coverage, not the old "curl-and-bash" method.

see https://gitter.im/typelevel/cats?at=571ccb8dd47413c07dccc27c

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,9 @@ language: scala
 git:
   depth: 9999
 
+install:
+  - pip install --user codecov
+
 scala:
 - 2.10.6
 - 2.11.8
--- a/scripts/travis-publish.sh
+++ b/scripts/travis-publish.sh
@@ -27,7 +27,7 @@ fi
 
 sbt_cmd="sbt ++$TRAVIS_SCALA_VERSION"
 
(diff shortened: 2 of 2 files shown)


InTheNow:
This is failing as `pip install` is not run. I'm not sure why - perhaps as this is just a PR travis does not want to install this? The code is a straight copy from https://github.com/InTheNow/catalysts/blob/master/.travis.yml#L3-L4.

Also see https://libraries.io/pypi/codecov

ceedubs:
@InTheNow thanks! This looks like a big improvement. All of the other PR builds are currently failing as well, so I suspect this build failure isn't actually related to your change. The build logs don't exactly leave me enlightened as to what's going wrong.

To eliminate variables as we try to figure out what's going wrong with the build, we should probably hold off on merging this, but 👍 from me once we get things figured out.

ceedubs:
@InTheNow do you think that the problem could be that we already have another `install` block [here](https://github.com/InTheNow/cats/blob/topic/codecov/.travis.yml#L17)? I don't know yaml/travis all that well, but I'm wondering if the latter is overwriting your install block.

PR #996, opened 2016-04-24
Add SemigroupK instance for Xor + tests

aaronlevin:
Xor's SemigroupK instance follows that of XorT.

### DIFF
--- a/tests/src/test/scala/cats/tests/XorTTests.scala
+++ b/tests/src/test/scala/cats/tests/XorTTests.scala
@@ -14,8 +14,6 @@ class XorTTests extends CatsSuite {
   implicit val iso = CartesianTests.Isomorphisms.invariant[XorT[List, String, ?]]
   checkAll("XorT[List, String, Int]", MonadErrorTests[XorT[List, String, ?], String].monadError[Int, Int, Int])
   checkAll("MonadError[XorT[List, ?, ?]]", SerializableTests.serializable(MonadError[XorT[List, String, ?], String]))
-  checkAll("XorT[List, String, Int]", MonoidKTests[XorT[List, String, ?]].monoidK[Int])
-  checkAll("MonoidK[XorT[List, String, ?]]", SerializableTests.serializable(MonoidK[XorT[List, String, ?]]))
   checkAll("XorT[List, ?, ?]", BifunctorTests[XorT[List, ?, ?]].bifunctor[Int, Int, Int, String, String, String])
   checkAll("Bifunctor[XorT[List, ?, ?]]", SerializableTests.serializable(Bifunctor[XorT[List, ?, ?]]))
   checkAll("XorT[List, Int, ?]", TraverseTests[XorT[List, Int, ?]].traverse[Int, Int, Int, Int, Option, Option])
--- a/tests/src/test/scala/cats/tests/XorTTests.scala
+++ b/tests/src/test/scala/cats/tests/XorTTests.scala
@@ -20,6 +20,8 @@ class XorTTests extends CatsSuite {
   checkAll("Traverse[XorT[List, Int, ?]]", SerializableTests.serializable(Traverse[XorT[List, Int, ?]]))
   checkAll("XorT[List, String, Int]", OrderLaws[XorT[List, String, Int]].order)
   checkAll("Order[XorT[List, String, Int]]", SerializableTests.serializable(Order[XorT[List, String, Int]]))
+  checkAll("XorT[Option, ListWrapper[String], ?]", SemigroupKTests[XorT[Option, ListWrapper[String], ?]].semigroupK[Int])
+  checkAll("SemigroupK[XorT[Option, ListWrapper[String], ?]]", SerializableTests.serializable(SemigroupK[XorT[Option, ListWrapper[String], ?]]))
 
   {
     implicit val F = ListWrapper.foldable
@@ -43,11 +45,6 @@ class XorTTests extends CatsSuite {
     implicit val F = ListWrapper.eqv[String Xor Int]
     checkAll("XorT[ListWrapper, String, Int]", OrderLaws[XorT[ListWrapper, String, Int]].eqv)
     checkAll("Eq[XorT[ListWrapper, String, Int]]", SerializableTests.serializable(Eq[XorT[ListWrapper, String, Int]]))
-  }
-
-  {
-    checkAll("XorT[Option, ListWrapper[String], ?]", SemigroupKTests[XorT[Option, ListWrapper[String], ?]].semigroupK[Int])
-    checkAll("SemigroupK[XorT[Option, ListWrapper[String], ?]]", SerializableTests.serializable(SemigroupK[XorT[Option, ListWrapper[String], ?]]))
   }
 
   // make sure that the Monad and Traverse instances don't result in ambiguous
--- a/tests/src/test/scala/cats/tests/XorTTests.scala
+++ b/tests/src/test/scala/cats/tests/XorTTests.scala
@@ -46,7 +46,6 @@ class XorTTests extends CatsSuite {
   }
 
   {
-    implicit val L = ListWrapper.semigroup[String]
     checkAll("XorT[Option, ListWrapper[String], ?]", SemigroupKTests[XorT[Option, ListWrapper[String], ?]].semigroupK[Int])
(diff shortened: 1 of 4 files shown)


aaronlevin:
Note: one difference from the `SemigroupK` instance for `XorT` is that `XorT` seemed to create a separate trait (`XorTSemigroupK`) and I wasn't sure why this was necessary so I didn't follow suit. I'm happy to update if there's a reason for this (am also generally curious)

aaronlevin:
Build failure ended with: `The command "scripts/travis-publish.sh" exited with 6.` Unsure how to proceed.

aaronlevin:
There is another possible implementation of `SemigroupK[Xor[E,?]]` where we take the first `Right` or the last `Left` and put no constraints on `E`. This would give a `SemigroupK` for any `Xor`. This is also identical to the `Semigroup` instance for `Either` in haskell.

Thoughts? I'd be happy to update and change if people feel either way.

I like this because it's more semantically closer to `Xor`, whereas if you want the `Semigroup` constraint on `E` then you can use `Validated`.

/cc @ceedubs

ceedubs:
@aaronlevin thanks! We are currently having an issue with builds failing that we haven't tracked down, so I don't think the build failure has anything to do with your changes. Hopefully we get that figured out shortly.

> Note: one difference from the SemigroupK instance for XorT is that XorT seemed to create a separate trait (XorTSemigroupK) and I wasn't sure why this was necessary so I didn't follow suit. I'm happy to update if there's a reason for this (am also generally curious)

I would have done the same as you. In some cases defining a separate trait can be handy for code reuse (for example if you have a `Semigroup` instance as long as there's a `Semigroup[E]` but you also have a `Monoid` instance if there's a `Monoid[E]`). Some people prefer to always create a separate trait/class as opposed to creating a new anonymous type. I guess I'm not exactly sure what the motivation is there - perhaps they just like being a bit more explicit? I think @non might fall into that camp, so he might weigh in. Either way, I think what you have done is consistent with many parts of the cats codebase and isn't a problem.

> I like this because it's more semantically closer to Xor, whereas if you want the Semigroup constraint on E then you can use Validated.

I have the same inclination. It's a little troubling that this would be inconsistent with the `SemigroupK` instance for `XorT` though. I'm guessing that the `XorT` instance followed a precedent set in scalaz without too much thought into whether we wanted to take the same approach in cats. I'm wondering if it would make sense to change it. @xuwei-k do you have any thoughts on this?

Sorry that a seemingly simple change turned into a complicated discussion!

aaronlevin:
@ceedubs Good call on alignment with `XorT`. I suggest we:

1. wait to hear back from @xuwei-k on anything about `XorT` or thoughts.
2. I can incorporate any changes, including updating both `Xor` and `XorT` to have the desired "first-right, last-left" semantics.

ceedubs:
It occurs to me that the reason for the approach taken in scalaz's `EitherT` (equivalent to cats `XorT`) is probably because it also provides a `MonadPlus` instance (equivalent to cats `MonadCombine`), which needs to be consistent with the `Plus` instance (equivalent to cats `SemigroupK`). The `MonadCombine` instance for `XorT` was removed in cats because it violates right absorbtion, right distributivity, and left distributivity. Cats does, however still provide a `MonoidK` instance for `XorT`, that its `SemigroupK` instance should be consistent with. Personally, I'm not a big fan of that `MonoidK` instance and wouldn't mind seeing it go away. It seems kind of arbitrary to me to claim that the `empty` `XorT[F, L, A]` is based on the `empty` of the monoid of the left type.

ceedubs:

aaronlevin:
@ceedubs I'm inclined to agree. Taking the semantic, right-biased interpretation of `Xor`/`XorT` as success-or-failure, then it really doesn't make sense that an `empty` value should correspond to failure.

I'll proceed with the changes in a bit and if we have to roll-back, that's fine. This will be a breaking change, unfortunately, as it may require putting constraints on the `monoidK` instance.

ceedubs:
@aaronlevin thanks for being so willing to make changes.

> This will be a breaking change, unfortunately, as it may require putting constraints on the monoidK instance.

I thought we were on the same page about simply removing the `MonoidK` instance for `XorT`, so this didn't quite make sense to me. Am I misunderstanding?

We aren't too concerned about breaking changes until the upcoming 1.0 release, so it's best to fix it now even if it is a breaking change.

aaronlevin:
@ceedubs ah, sorry, I can remove the `MonoidK` instance :)

aaronlevin:
@ceedubs updated! :)

tests/src/test/scala/cats/tests/XorTests.scala
```diff
@@ -46,6 +47,12 @@ class XorTests extends CatsSuite {
     checkAll("Eq[ListWrapper[String] Xor ListWrapper[Int]]", SerializableTests.serializable(Eq[ListWrapper[String] Xor ListWrapper[Int]]))
   }
 
+  {
+    implicit val L = ListWrapper.semigroup[String]
```
ceedubs: This line is creating a local (to the surrounding `{ ... }` block) `Semigroup[ListWrapper[String]]` instance. We are taking this local instance approach in some tests, because we want to make sure that things work given the minimum required instances. For example, does this work when we have a `Semigroup[ListWrapper[String]]` but no `Monoid[ListWrapper[String]]`?

In this case, the `SemigroupK` instance you created doesn't have any requirements (specifically, no `Semigroup[L]` instance is required). So we shouldn't have to take this approach here. We should be able to just add the 2 `checkAll` tests around line 34 without creating a separate block to house locally-scoped instances.

I'd imagine all of these locally-scoped instances are somewhat terrifying for someone who is used to Haskell :)

tests/src/test/scala/cats/tests/XorTTests.scala
```diff
@@ -14,8 +14,6 @@ class XorTTests extends CatsSuite {
   implicit val iso = CartesianTests.Isomorphisms.invariant[XorT[List, String, ?]]
   checkAll("XorT[List, String, Int]", MonadErrorTests[XorT[List, String, ?], String].monadError[Int, Int, Int])
   checkAll("MonadError[XorT[List, ?, ?]]", SerializableTests.serializable(MonadError[XorT[List, String, ?], String]))
-  checkAll("XorT[List, String, Int]", MonoidKTests[XorT[List, String, ?]].monoidK[Int])
-  checkAll("MonoidK[XorT[List, String, ?]]", SerializableTests.serializable(MonoidK[XorT[List, String, ?]]))
```
ceedubs: This is minor, but now that the `SemigroupK` instance no longer requires a `Semigroup[L]`, the `SemigroupK` tests down at line ~49 don't need the local `Semigroup[ListWrapper[String]]` instance and could be moved up here, similarly to [this comment](https://github.com/typelevel/cats/pull/996/files#r61065904).

ceedubs:
@aaronlevin I suggested a couple very minor (and admittedly nitpicky) changes to the tests, but this looks good to me.

Are other people okay with removing the `MonoidK` instance for `XorT`?

tests/src/test/scala/cats/tests/XorTests.scala
aaronlevin: ah, good call. this is what I get for submitting a PR at 8am before breakfast :)

aaronlevin: and, yeah, coming from Haskell it feels like these instances are just flying around and I have no idea what's happening :) For example, I didn't know you could introduce a local instance by wrapping something in curly braces.

aaronlevin:
@ceedubs feedback has been incorporated (thanks!)

ceedubs:
👍 fantastic. Thanks a bunch @aaronlevin!

aaronlevin:
@ceedubs \o/

tests/src/test/scala/cats/tests/XorTests.scala
ceedubs: Yeah, the implicit scope rules are incredibly complicated (which admittedly we use as a "feature" quite a bit in cats to aid in inference). In this case we are using the curly braces to have it scoped locally to that block instead of to the whole class/object.

😈

ceedubs:
closing and reopening to see if it makes the code coverage report make more sense

non:
Looks good to me. 👍 Thanks @aaronlevin!

PR #998, opened 2016-04-24
Add SemigroupK instance for Validated + tests

aaronlevin:


aaronlevin:
Similar to my other PR on adding a `SemigroupK` instance for `Xor`. The instance requires only that the failure type has a `Semigroup` constraint.

ceedubs:
Looks great. Thanks @aaronlevin!

👍, but we should get the build issues figured out before we merge so we don't throw more variables into the mix.

ceedubs:
It occurs to me that it might also be handy to create a `combineK` (or perhaps there's a better name when speaking specifically about `Validated`) method on `Validated` directly and have the `SemigroupK` instance simply delegate to it. If we do this we should probably have a scaladoc comment describing how its behavior is a little different than `orElse` and `combine` (which wouldn't be a bad comment to have anyway). I'm not saying that you need to feel obligated to do all of this in your PR - just jotting this down before I forget.

aaronlevin:
@ceedubs since we're waiting for the build stuff to get fixed, I'm happy to update this PR tonight (EU time) with your suggestion :)

ceedubs:

ceedubs:
👍

At some point I think it would be good for us to follow up on [this comment](https://github.com/typelevel/cats/pull/998#issuecomment-214283934). I'm happy for that to be a separate effort though.

aaronlevin:
@ceedubs i think it's a good idea, but i can submit another PR with that change. is cats `0.5` about to be released? it'd be nice to get this in there! :)

ceedubs:
@aaronlevin it is and it would be :). We have a convention of waiting for two 👍s before merging, so that's all we are waiting on for this PR as far as I'm concerned.

adelbertc:
👍

PR #999, opened 2016-04-25
Add semiflatMap to OptionT and XorT

frosforever:
Adds `mapF` to OptionT and XorT making it easier in the case where you have `XorT[F, A, B]` and `B => F[C]` to get `XorT[F, A, C]`. 

I'm not sure if the name `mapF` is appropriate or even if this belongs in cats but I've found something like this very useful recently.

### DIFF
--- a/core/src/main/scala/cats/data/OptionT.scala
+++ b/core/src/main/scala/cats/data/OptionT.scala
@@ -25,6 +25,9 @@ final case class OptionT[F[_], A](value: F[Option[A]]) {
   def map[B](f: A => B)(implicit F: Functor[F]): OptionT[F, B] =
     OptionT(F.map(value)(_.map(f)))
 
+  def mapF[B](f: A => F[B])(implicit F: Monad[F]): OptionT[F, B] =
+    flatMap(a => OptionT.liftF(f(a)))
+
   def flatMap[B](f: A => OptionT[F, B])(implicit F: Monad[F]): OptionT[F, B] =
     flatMapF(a => f(a).value)
 
--- a/core/src/main/scala/cats/data/OptionT.scala
+++ b/core/src/main/scala/cats/data/OptionT.scala
@@ -25,7 +25,7 @@ final case class OptionT[F[_], A](value: F[Option[A]]) {
   def map[B](f: A => B)(implicit F: Functor[F]): OptionT[F, B] =
     OptionT(F.map(value)(_.map(f)))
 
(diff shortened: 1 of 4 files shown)


ceedubs:
Thanks @frosforever! This seems like a handy addition. I'm a bit hesitant with the name. This seems more closely related to `flatMap` than to `map`. I think `flatMapF` is already taken though :P.

We have some build issues that we need to figure out before we merge any PRs. Let's let this sit for a bit while we figure those out, and maybe someone will have a good name suggestion in the meantime.

liff:
This got me wondering. If we have monad `G` with transformer `GT[F[_], A]` = `F[G[A]]`, then the possible combinators are:

```scala
  map:        A => B
  mapF:       A => F[B] // this ticket
  subflatMap: A => G[B]
  flatMapF:   A => F[G[B]]
  flatMapG:   A => G[F[B]] // doesn't exist? can this be defined for any monad transformer?
  flatMap:    A => GT[F, B]
```

If `flatMapG` is viable and useful, does this make some kind of consistent hierarchy for which a consistent naming can be devised?

frosforever:
@ceedubs let me rephrase that. I'm certain the name isn't appropriate but `flatMapF` and `subflatMap` were already taken. I'm not sure what to name it. `superflatMap`?

ceedubs:

non:
What about `.semiflatMap`? I'm just brainstorming at this point.

non:
@frosforever What do you think? Does `semiflatMap` seem OK with you?

frosforever:
@non I don't think I'm qualified to have an opinion. `semiflatMap` sounds good to me. It at least has `flatmap` in there somewhere so the requirement of a monad isn't strange.

non:
@frosforever OK. I think @liff's `subflatMap` suggestion, or `semiflatMap`, would. I'd be happy to +1 this if you changed the name to either of those.

frosforever:
@non `semiflatMap` it is then. `subflatMap` was already taken.

ceedubs:
👍 nice. Thanks @frosforever! Sorry for the long delays in review and indecision about names.

frosforever:
@ceedubs no need to apologize 😀. Thanks for reviewing!

Tvaroh:
👍 Would be very cool to merge this. Thanks!

kailuowang:
👍  merging since all previous comments were addressed.

Tvaroh:
@kailuowang thank you! Is it supposed to be in next minor release?

PR #1000, opened 2016-04-25
Experiment - seeing if this fixes builds

ceedubs:


ceedubs:
🎉 @InTheNow's change fixed the build once it was moved to the existing `install` block in the travis configuration file.

ceedubs:
I'm going to break procedure and go ahead and merge this, because I can't imagine anyone objecting to such a simple change that simultaneously fixes our builds and makes them no longer execute arbitrary code that comes back from a `curl` to a third party service.

milessabin:
:+1:

dwijnand:
Just as a side-note, food-for-thought: now it executes arbitrary code that comes back from `pip`...

milessabin:
I agree that the `curl|bash` hate doing the rounds at the moment is a double standard, but still, this does fix the build as well ;-)

ceedubs:
@dwijnand fair point :)

I think this is at least a slight improvement (beyond fixing the build). At one point when the build was failing it was because what we were feeding into `bash` was the HTML of a 500 Internal Server Error page. Hopefully `pip` at least handles server errors a bit more gracefully :)

PR #1001, opened 2016-04-25
Introduce cats-kernel and remove algebra dependency

non:
### Overview

This branch pulls the code from algebra-core and algebra-laws into cats-kernel and cats-kernel-laws (respectively). The intention here is to reverse the Algebra/Cats dependency using the intersection of code that they are both interested in.

We already have two PRs that address this same issue (#786 and #978). Thanks to @milessabin and @has609 for that work -- I'm sorry I wasn't able to better communicate what I thought the requirements for this PR were (and that I couldn't get this PR done faster). I am proposing we merge this PR instead of those.

In order to satisfy everyone in Cats and Algebra as much as possible, I tried not to editorialize and just do the minimum job necessary get this working. However, I think we have the following requirements:

### Dependencies

Algebra needs to be able to depend on cats-kernel and have everything that currently works continue to work. This means that we need to export the same instances that Algebra uses here (rather than migrating the instances to cats-core). This also means that we need to use the package `cats.kernel` rather than `cats`, since otherwise we have a collision around who provides the `cats.std` package object.

This also means that we need to export `cats-kernel-laws` separately from `cats-laws`, and stick to the the current layout/strategy Algebra uses (for now).

### No new dependencies

Cats-kernel should not introduce new dependencies that Algebra didn't already have. I was able to accommodate this except for an extra compile-time dependency from cats-kernel-laws onto catalysts-platform (to get serialization tests working correctly on JVM and disabled on JS), which I thought was worth it and didn't pose a big risk.

### Style issues

Initially we need to start from the Algebra code, even if it is in a different style than Cats currently. We should be open to PRs improving the kernel, but the expectation is that this code needs to be very stable, and should also be less opinionated (to support Algebra's goal that it be usable from Algebird, Breeze, Spire, etc.). This means that we aren't using Simulacrum or Machinist for now, not supporting operators/syntax in the kernel, and being very explicit.

In some cases (i.e. the cats-kernel / algebra law testing code) I think we might want to try to adopt Cats' style, but let's leave that conversation until after this PR is merged if possible.

### Administrative issues

Even though we are moving this code to the Cats repo, I'd like to imagine that we'll solicit the Algebra maintainers' opinions before making big changes to this code. Would we be willing to consider the Algebra maintainers as Cats maintainers for the purposes of working in the Kernel? I'd also like to freeze this code sooner than later, and begin doing MiMA compatibility testing to ensure robust binary compatibility ASAP.

### Conclusion

I'd like to get sign-offs from at least two other Algebra maintainers on this PR in addition to the usual sign-offs from Cats maintainers. I'm happy to make substantive edits to this PR if we can get consensus amongst those two groups that the changes are warranted.

Also, I'd like to merge #994 and get another Cats release out before we actually merge this PR -- let's discuss the timing of the first Algebra-free release here as well.


non:
I'm also noticing that some files and large blocks are entirely commented out. Feel free to point these out -- I'll be removing them as I find them.

core/src/main/scala/cats/std/anyval.scala
```diff
@@ -15,129 +14,103 @@ trait AnyValInstances
...
-trait IntInstances extends algebra.std.IntInstances {
+trait IntInstances extends cats.kernel.std.IntInstances {
 
   implicit val intShow: Show[Int] =
     Show.fromToString[Int]
 
   implicit val intGroup: CommutativeGroup[Int] =
-    AdditiveCommutativeGroup[Int].additive
-
+    new CommutativeGroup[Int] {
```
johnynek: should this not be in `cats.kernel.std.IntInstances`?

kernel-laws/src/main/scala/cats/kernel/laws/GroupLaws.scala
```diff
@@ -0,0 +1,142 @@
...
+    Rules.rightInverse(A.empty)(A.combine)(A.inverse),
+    Rules.consistentInverse("remove")(A.remove)(A.combine)(A.inverse)
+  )
+
+  def commutativeGroup(implicit A: CommutativeGroup[A]) = new GroupProperties(
+    name = "commutative group",
+    parents = List(group, commutativeMonoid)
+  )
+
+  // // additive groups
```
johnynek: these should just be deleted right (will be in algebra).

kernel-laws/src/main/scala/cats/kernel/laws/package.scala
```diff
@@ -0,0 +1,45 @@
...
+
+import org.typelevel.discipline.Predicate
+
+import org.scalacheck._
+import org.scalacheck.util.Pretty
+import Prop.{False, Proof, Result}
+
+package object laws {
+
+  implicit def PredicateFromMonoid[A](implicit ev: Eq[A], A: Monoid[A]): Predicate[A] =
```
johnynek: can we name this `nonZeroPredicate` or something?

kernel/src/main/scala/cats/kernel/Group.scala
```diff
@@ -0,0 +1,48 @@
...
+  def remove(a: A, b: A): A = combine(a, inverse(b))
+
+  /**
+   * Return `a` appended to itself `n` times. If `n` is negative, then
+   * this returns `inverse(a)` appended to itself `n` times.
+   */
+  override def combineN(a: A, n: Int): A =
+    if (n > 0) repeatedCombineN(a, n)
+    else if (n == 0) empty
+    else if (n == Int.MinValue) combineN(inverse(combine(a, a)), 1073741824)
```
johnynek: this is due to:
1. `- Int.MinValue == Int.MinValue`
2. `combineN(x, n) == combineN(inverse(x), -n)`
3. `combineN(x, n*n) = combineN(combine(x, x), n)`

So, `1073741824 == -(Int.MaxValue / 2)`

Can we add some verbosity to the comments?

kernel/src/main/scala/cats/kernel/Order.scala
```diff
@@ -0,0 +1,209 @@
...
+    compare(x, y) > 0
+
+  /**
+   * Returns a new `Order[A]` instance that first compares by the original
+   * `Order` instance and uses the provided `Order` instance to "break ties".
+   *
+   * That is, `x.whenEqual(y)` creates an `Order` that first orders by `x` and
+   * then (if two elements are equal) falls back to `y` for the comparison.
+   */
+  def whenEqual(o: Order[A]): Order[A] = new Order[A] {
```
johnynek: out of curiosity, does this make a bounded semilattice like the equals ones does? Seems like, with `empty` being `allEqual`? I guess it is a Band, not a Semilattice since it is not commutative.

kernel/src/main/scala/cats/kernel/Order.scala
```diff
@@ -0,0 +1,209 @@
...
+   *
+   * `empty` returns a trivial `Order[A]` which considers all `A` instances to
+   * be equal.
+   *
+   * `combine(x: Order[A], y: Order[A])` creates an `Order[A]` that first
+   * orders by `x` and then (if two elements are equal) falls back to `y`.
+   *
+   * @see [[Order.whenEqual]]
+   */
+  def whenEqualMonoid[A]: Monoid[Order[A]] =
```
johnynek: actually, this is at least also a `Band` (it is idempotent).

core/src/main/scala/cats/std/anyval.scala
non: I was going to ask you about that -- we didn't have generic semigroups for numbers in Algebra. If everyone agrees that we want those I can move them in this PR.

kernel/src/main/scala/cats/kernel/std/bigInt.scala
```diff
@@ -0,0 +1,24 @@
+package cats.kernel
+package std
+
+package object bigInt extends BigIntInstances
+
+trait BigIntInstances {
```
johnynek: why not `Ring` here? (or is there no ring?) or at least Group.

kernel-laws/src/main/scala/cats/kernel/laws/GroupLaws.scala
non: Yup, total oversight.

kernel/src/main/scala/cats/kernel/Group.scala
non: Sure.

kernel/src/main/scala/cats/kernel/std/double.scala
```diff
@@ -0,0 +1,26 @@
+package cats.kernel
+package std
+
+import java.lang.Math
+
+trait DoubleInstances {
+  implicit val doubleOrder: Order[Double] = new DoubleOrder
+}
+
+class DoubleOrder extends Order[Double] {
```
johnynek: strictly, is this a PartialOrder due to `NaN`? Should we maybe have that and some `unsafeToOrder` method for people to opt out?

kernel/src/main/scala/cats/kernel/std/option.scala
```diff
@@ -0,0 +1,68 @@
+package cats.kernel
+package std
+
+package object option extends OptionInstances
+
+trait OptionInstances extends OptionInstances1 {
+  implicit def optionOrder[A: Order] = new OptionOrder[A]
+  implicit def optionMonoid[A: Semigroup] = new OptionMonoid[A]
```
johnynek: there are at least two other Monoids: `left.orElse(right)` `right.orElse(left)`. Should we expose those (maybe not as the default?)

kernel/src/main/scala/cats/kernel/std/set.scala
```diff
@@ -0,0 +1,28 @@
...
+    if (x.size < y.size) if (x.subsetOf(y)) -1.0 else Double.NaN
+    else if (y.size < x.size) -partialCompare(y, x)
+    else if (x == y) 0.0
+    else Double.NaN
+
+  override def eqv(x: Set[A], y: Set[A]): Boolean =
+    x == y
+}
+
+class SetMonoid[A] extends Monoid[Set[A]] {
```
johnynek: this can be a bounded semilattice.

kernel/src/main/scala/cats/kernel/std/unit.scala
```diff
@@ -0,0 +1,29 @@
+package cats.kernel
+package std
+
+package object unit extends UnitInstances
+
+trait UnitInstances {
+  implicit val unitAlgebra: Order[Unit] with BoundedSemilattice[Unit] =
```
johnynek: why not `CommutativeGroup` along with `BoundedSemilattice`?

kernel/src/main/scala/cats/kernel/std/double.scala
non: This implementation is using `java.lang.Double.compare` which means it is total (but ugly) -- `NaN` is greater than all non-NaN values:

```
scala> java.lang.Double.compare(Double.NaN, Double.PositiveInfinity)
res4: Int = 1
scala> java.lang.Double.compare(Double.NaN, Double.NaN)
res7: Int = 0
```

kernel/src/main/scala/cats/kernel/std/option.scala
non: I would be open to putting those in the `Monoid` companion object -- I don't think we should put them in `cats.kernel.std` since we wouldn't expose them by default.

kernel/src/main/scala/cats/kernel/std/bigInt.scala
non: I added a `CommutativeGroup`.

kernel/src/main/scala/cats/kernel/Order.scala
non: Yeah I updated the method to return a `Band[Order[A]]`.

kernel/src/main/scala/cats/kernel/Order.scala
```diff
@@ -194,16 +195,16 @@ object Order extends OrderFunctions {
    *
    * @see [[Order.whenEqual]]
    */
-  def whenEqualMonoid[A]: Monoid[Order[A]] =
-    new Monoid[Order[A]] {
+  def whenEqualMonoid[A]: Band[Order[A]] =
```
johnynek: I think we want `Band[Order[A]] with Monoid[Order[A]]` Band does not have `empty` but `BoundedSemilattice` is commutative.

non:
It seems like codecov.io may have a partial outage. I'm not sure the commits will turn green, but Travis passed.

build.sbt
```diff
@@ -24,6 +24,16 @@ lazy val catsDoctestSettings = Seq(
   doctestWithDependencies := false
 ) ++ doctestSettings
 
+lazy val kernelSettings = Seq(
+  scalacOptions ++= commonScalacOptions.filter(_ != "-Ywarn-value-discard"),
+  resolvers ++= Seq(
+    "bintray/non" at "http://dl.bintray.com/non/maven",
```
tixxit: What is this used for?

non: That may not be necessary -- I'm not sure what we're using it for.

tixxit:
👍

ceedubs:
@non regarding lowering code coverage:

I noticed that https://github.com/typelevel/cats/pull/990 reported a drop in code coverage, which seems wrong to me since I just moved something into another module, but I haven't looked into it in depth. It's possible that something is a bit wonky at the moment.

kernel/src/main/scala/cats/kernel/std/Util.scala
```diff
@@ -90,7 +77,7 @@ object StaticMethods {
     while (true) {
       if (xs.hasNext) {
         if (ys.hasNext) {
-          if (ev.eqv(xs.next, ys.next)) return true
```
johnynek: yay for tests!

kernel/src/main/scala/cats/kernel/std/map.scala
```diff
@@ -1,16 +1,11 @@
...
 
 package object map extends MapInstances
 
-trait MapInstances extends MapInstances0 {
-  implicit def mapGroup[K, V: Group]: MapGroup[K, V] =
-    new MapGroup[K, V]
-}
-
-trait MapInstances0 {
+trait MapInstances {
```
johnynek: below in `MapEq` should we add the optimization of `(x eq y) || { // rest of the method?` no need to do `O(N)` work when they are referentially equal, right?

non: Oh yeah, that's a great point, I'll add it.

non: I'll add that optimization to the other instances that would benefit (e.g. `String`, `Vector`, `List`, `Stream`, etc).

ceedubs:
@non sorry this has merge conflicts, which are probably from me merging #994.

build.sbt
```diff
@@ -80,8 +88,12 @@ lazy val scalacheckVersion = "1.12.5"
 
 lazy val disciplineDependencies = Seq(
   libraryDependencies += "org.scalacheck" %%% "scalacheck" % scalacheckVersion,
-  libraryDependencies += "org.typelevel" %%% "discipline" % "0.4"
-)
+  libraryDependencies += "org.typelevel" %%% "discipline" % "0.4")
+
+lazy val testingDependencies = Seq(
+  libraryDependencies += "org.typelevel" %%% "catalysts-platform" % "0.0.2",
```
ceedubs: Should `catalysts-platform` have a `% "test"` modifier?

kernel/src/main/scala/cats/kernel/CommutativeGroup.scala
```diff
@@ -0,0 +1,17 @@
+package cats.kernel
+
+import scala.{ specialized => sp }
+
+/**
+ * An abelian group is a group whose operation is commutative.
```
ceedubs: Minor: we might want something like "A `CommutativeGroup` (AKA abelian group) is a group whose `combine` operation is commutative".

kernel/src/main/scala/cats/kernel/Eq.scala
```diff
@@ -0,0 +1,138 @@
...
+   * Return an Eq that gives the result of the or of this and that
+   * Note this is idempotent
+   */
+  def or(that: Eq[A]): Eq[A] =
+    new Eq[A] {
+      def eqv(x: A, y: A) = self.eqv(x, y) || that.eqv(x, y)
+    }
+}
+
+trait EqFunctions {
```
ceedubs: Is this meant to be open for people to mix in elsewhere? If it's just separated out to break up code a bit, I'd default to `sealed abstract class` instead of `trait` so we can add methods to it without breaking binary compatibility.

kernel/src/main/scala/cats/kernel/Group.scala
```diff
@@ -0,0 +1,61 @@
...
+    //   combineN(combine(x, x), -1073741824)
+    //   combineN(inverse(combine(x, x)), 1073741824)
+    if (n > 0) repeatedCombineN(a, n)
+    else if (n == 0) empty
+    else if (n == Int.MinValue) combineN(inverse(combine(a, a)), 1073741824)
+    else repeatedCombineN(inverse(a), -n)
+  }
+}
+
+trait GroupFunctions[G[T] <: Group[T]] extends MonoidFunctions[Group] {
```
ceedubs: Same comment as https://github.com/typelevel/cats/pull/1001/files#r61071186

kernel/src/main/scala/cats/kernel/Order.scala
```diff
@@ -0,0 +1,210 @@
...
+    new Order[A] {
+      def compare(x: A, y: A): Int = 0
+    }
+
+
+  /**
+   * A `Band[Order[A]]` can be generated for all `A` with the following
+   * properties:
+   *
+   * `empty` returns a trivial `Order[A]` which considers all `A` instances to
```
ceedubs: Warning nitpicky comment: `empty` comes from `Monoid` not `Band`, so we may want to adjust the previous line of the comment to mention both `Band` and `Monoid`.

kernel/src/main/scala/cats/kernel/PartialOrder.scala
```diff
@@ -0,0 +1,169 @@
...
+   */
+  def gteqv(x: A, y: A): Boolean = partialCompare(x, y) >= 0
+
+  /**
+   * Returns true if `x` > `y`, false otherwise.
+   */
+  def gt(x: A, y: A): Boolean = partialCompare(x, y) > 0
+}
+
+trait PartialOrderFunctions {
```
ceedubs: Same as https://github.com/typelevel/cats/pull/1001/files#r61071186

kernel/src/main/scala/cats/kernel/Priority.scala
```diff
@@ -0,0 +1,65 @@
+package cats.kernel
```
ceedubs: This doesn't appear to be used anywhere in cats. Is there a particular reason to include it in cats-kernel?

kernel/src/main/scala/cats/kernel/Semigroup.scala
```diff
@@ -0,0 +1,79 @@
...
+  /**
+   * Given a sequence of `as`, combine them and return the total.
+   *
+   * If the sequence is empty, returns None. Otherwise, returns Some(total).
+   */
+  def combineAllOption(as: TraversableOnce[A]): Option[A] =
+    as.reduceOption(combine)
+}
+
+trait SemigroupFunctions[S[T] <: Semigroup[T]] {
```
ceedubs: Same as https://github.com/typelevel/cats/pull/1001/files#r61071186

kernel/src/main/scala/cats/kernel/std/Util.scala
```diff
@@ -0,0 +1,90 @@
+package cats.kernel
+package std.util
+
+import scala.collection.mutable
+
+object StaticMethods {
```
ceedubs: Suggested name change: `object SpiralStaircaseDownFromTheFunctionalIvoryTower` ;)

kernel/src/main/scala/cats/kernel/std/array.scala
```diff
@@ -0,0 +1,73 @@
+package cats.kernel
+package std
+
+import scala.{ specialized => sp }
+
+package object array extends ArrayInstances
```
ceedubs: Since cats focuses on _pure_ functional programming, would it make sense for `Array` instances to not be part of cats-kernel and instead stay in algebra or something?

kernel/src/main/scala/cats/kernel/std/list.scala
```diff
@@ -0,0 +1,95 @@
...
+import scala.collection.mutable
+
+package object list extends ListInstances
+
+trait ListInstances extends ListInstances1 {
+  implicit def listOrder[A: Order] = new ListOrder[A]
+  implicit def listMonoid[A] = new ListMonoid[A]
+}
+
+trait ListInstances1 extends ListInstances2 {
```
ceedubs: I'm inclined to make traits like `ListInstances1` and `ListInstances2` `sealed` and `private[cats.kernel.std]` just to signify that they are an internal detail and not really meant to be used directly. I think we are taking this approach in most of cats.

ceedubs: It's always easier to remove `sealed` and `private` than it is to add them in later.

kernel/src/main/scala/cats/kernel/std/option.scala
ceedubs: FWIW `left.orElse(right)` is exposed through the `SemigroupK[Option]` instance (not in kernel).

If we _do_ provide these other instances (I don't necessarily see a strong need for them), I guess that `std` seems like a more natural place to me than the `Monoid` companion object. We probably have lots of types with alternative type class instances, and putting those with the other instances for those types seems like a more modular approach to me than shoving all of them into the corresponding type class companion objects (especially once you start talking about type class hierarchies). Just my 2 cents.

kernel/src/main/scala/cats/kernel/std/stream.scala
```diff
@@ -0,0 +1,58 @@
+package cats.kernel
+package std
+
+import cats.kernel.std.util.StaticMethods
+
+package object stream extends StreamInstances
+
+trait StreamInstances extends StreamInstances1 {
+  implicit def streamOrder[A: Order] = new StreamOrder[A]
```
ceedubs: I think we should add explicit return types (`Order[Stream[A]]` as opposed to `StreamOrder[A]`) here and in other places in this file. I was thinking that we had a compiler flag that was enforcing this. Is it turned off in cats-kernel?

kernel/src/main/scala/cats/kernel/std/string.scala
```diff
@@ -0,0 +1,27 @@
+package cats.kernel
+package std
+
+package object string extends StringInstances
+
+trait StringInstances {
+  implicit val stringOrder: Order[String] = new StringOrder
+  implicit val stringMonoid = new StringMonoid
```
ceedubs: Explicit return type

ceedubs:
Thanks a bunch for all of your work on this @non!

build.sbt
non: I think not, since we are checking the platform in the law testing code that we are publishing.

kernel/src/main/scala/cats/kernel/std/array.scala
non: I'm fine with leaving the array instances in Algebra if people would prefer that.

kernel/src/main/scala/cats/kernel/std/stream.scala
non: Agreed -- I was trying to clean that up but clearly missed a few.

kernel/src/main/scala/cats/kernel/Eq.scala
non: So, I'd like @johnynek to weigh-in on this. It's possible that other libraries like Algebra and Algebird will want to be able to mix these into their own companion objects, which would make non-sealed types necessary, and might make traits more attractive.

I don't see any places where we need to be using traits, so it's possible we could use abstract classes instead.

kernel/src/main/scala/cats/kernel/Priority.scala
non: No, I'm happy to remove it.

non:
@ceedubs I think I took every piece of advice you gave except about the mix-ins, which I want to wait to hear from Oscar about.

kernel/src/main/scala/cats/kernel/Eq.scala
tixxit: Personal preference is that we leave them unsealed, but abstract classes. Making them abstract classes makes them slightly less flexible, but since the whole purpose of cats-kernel is to be stable, I think it's a trade-off worth making. Libraries higher up the stack can make their `*Functions` traits to allow for multiple inheritance, but it isn't needed here.

build.sbt
ceedubs: Ah I didn't realize this was only being included in laws/tests. 👍

kernel/src/main/scala/cats/kernel/Eq.scala
ceedubs: I have no problem with dropping `sealed`. Changing `trait` to `abstract class` for binary compatibility was my real concern.

johnynek: that sounds good to me. I can't think of a case where we want to mix two of these, so abstract class should be fine.

johnynek:
Seems like the coverage is actually pretty low.  Check:

https://codecov.io/gh/typelevel/cats/tree/a6b7dfd390d16d562eada8502c4e1fc4e14c053c/kernel/src/main/scala/cats/kernel

https://codecov.io/gh/typelevel/cats/tree/a6b7dfd390d16d562eada8502c4e1fc4e14c053c/kernel/src/main/scala/cats/kernel/std

Looks like we just need to connect a lot of the instances to the laws, but also some of the core methods are not tested, for instance:
https://codecov.io/gh/typelevel/cats/src/a6b7dfd390d16d562eada8502c4e1fc4e14c053c/kernel/src/main/scala/cats/kernel/PartialOrder.scala

pmin, pmax, reverse look untested.

kernel/src/main/scala/cats/kernel/PartialOrder.scala
```diff
@@ -0,0 +1,169 @@
...
+   * Result of comparing `x` with `y`. Returns None if operands are
+   * not comparable. If operands are comparable, returns Some[Int]
+   * where the Int sign is:
+   * - negative iff `x < y`
+   * - zero     iff `x = y`
+   * - positive iff `x > y`
+   */  
+  def tryCompare(x: A, y: A): Option[Int] = {
+    val c = partialCompare(x, y)
+    if (c.isNaN) None else Some(c.signum)
```
johnynek: this boxes doesn't it? Don't we need `java.lang.Double.isNaN(c)`?

non: Good catch! Yes, I'll change it.

non:
@johnynek Regarding coverage:

Yes, we're missing a lot of things. One thing I want to do is write laws to ensure that `compare(x, y) < 0` is always the same as `x < y` and so on. I think we'll get a ton of coverage (and correctness assurance) once we do this.

ceedubs:
@non is it fair to assume that this PR has as much coverage as was previously present in algebra for the same code? If so, I would probably advocate for getting this PR merged (after the 0.5 cats release) and then following up with code coverage PRs for two reasons: 1) so you aren't forced to do all of the work and 2) to avoid the hassle of repeated merge conflicts.

johnynek:
+1 on merging and beefing up coverage after.

kernel/src/main/scala/cats/kernel/PartialOrder.scala
johnynek: also, we might keep `val some1 = Some(1); val someNeg1 = Some(-1); val someZero = Some(0)` around and not allocate on each compare. These could be private in the PartialOrder object.

non:
OK I think all the extant issues (except for coverage) have been addressed.

ceedubs:
👍 thanks again for all of your work on this, @non! Are we waiting until after the cats 0.5 release for this?

milessabin:
:+1: awesome work! Thanks for picking up where I left off :-)

non:
@ceedubs Yes, I think we should do one more release before merging this.

non:
Looks like *master* merged cleanly. I'm fine with merging this anytime.

PR #1002, opened 2016-04-25
correct the url to Data types à la carte

channingwalton:
The URL to Data types à la carte seems to have changed to this URL http://www.staff.science.uu.nl/~swier004/publications/2008-jfp.pdf


ceedubs:
thanks @channingwalton!

PR #1005, opened 2016-04-27
fix broken link in foldable docs

hamishdickson:
The link for "A tutorial on the universality and expressiveness of fold" in the docs for foldable incorrectly uses `https` instead of `http` - this is enough to break the link. This change resolves this.


adelbertc:
👍

PR #1006, opened 2016-04-27
Revert "Disable scaladoc during packaging for scala 2.10"

kailuowang:
Disabling docs for scala 2.10 can't be achieved by this change alone. Sonatype release requires javadoc jars, this is now blocking the 0.5.0 release. My proposal is that we revert this for now and find a more comprehensive solution for https://github.com/typelevel/cats/pull/898 later. I know it's kicking the can down the road, but this PR itself does not bring any value to 0.5.0, so I think it is sensible to not having it blocking the release. 

Update: #898 is not included in the 0.5.0 release.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -233,12 +233,6 @@ lazy val publishSettings = Seq(
   scmInfo := Some(ScmInfo(url("https://github.com/typelevel/cats"), "scm:git:git@github.com:typelevel/cats.git")),
   autoAPIMappings := true,
   apiURL := Some(url("http://typelevel.org/cats/api/")),
-  publishArtifact in (Compile, packageDoc) := {
-    CrossVersion.partialVersion(scalaVersion.value) match {
-      case Some((2, 10)) => false  // don't package scaladoc when publishing for 2.10
-      case _ => true
-    }
-  },
   pomExtra := (
     <developers>
       <developer>


ceedubs:
👍

travisbrown:
👍

non:
👍 Thanks!

mikejcurry:
Hey folks - sorry about that. Thanks for reverting.

ceedubs:
@kailuowang This has come up again in #1041. I noticed that circe is doing [this](https://github.com/travisbrown/circe/blob/846b20b3cefd9800880ee5054ae1b6c1749f06ad/build.sbt#L209-L214) to omit sources in the doc. Did you ever try anything like that without changing `publishArtifact in (Compile, packageDoc)`? It seems to make `coreJVM/compile:doc` no longer fail. And it seems like it must satisfy sonatype, because circe is publishing there.

kailuowang:
@ceedubs yeah, that might work, let me give it a try (i wonder if it still creates the java-doc jar and thus pass the sonatype test)

kailuowang:
seems working, PR #1051

kailuowang:
per gitter chat with @InTheNow, we might be able to find a better solution than this. We might be able to generate 2.10 javadoc with sbt 0.13.11

PR #1011, opened 2016-04-30
Add missing type parameters.

non:
We had left out some type parameters in `cats.kernel` which leads to errors when trying to use those methods.


milessabin:
:+1:

adelbertc:
👍

Issue #1012, opened 2016-05-03
Add WriterT Monoid


yilinwei:
I would like to add an implicit `Monoid` for `WriterT`. I've copied the implementation from scalaz [here](https://github.com/yilinwei/cats/commit/542b971f1c442eb8f04389f99ff695c943d463bf).

I would like to add some tests for the commit but am having a little trouble understanding how the tests work and should be structured. My initial attempt is [here](https://github.com/yilinwei/cats/commit/cfc2ccb045bfffda53a371efa5ef21c8b4c08e15) where I've just made sure that the `Monoid` and `Semigroup` can be resolved.

adelbertc:
You'll also have to test the `Monoid` laws for it

yilinwei:
I had some issues finding them at first (which was what I meant to ask) because I didn't realise they were under `GroupLaws` in `kernel`.

I've now added the test [here](https://github.com/yilinwei/cats/commit/d48c051a8c530eeeedcbeb7e387cfe285d0ec70e).

PR #1014, opened 2016-05-03
Adding WriterT Monoid

yilinwei:
Adding WriterT Monoid as referenced in #1012. 

It's also come to my attention that it's one of the subtasks in #620.

### DIFF
--- a/core/src/main/scala/cats/data/WriterT.scala
+++ b/core/src/main/scala/cats/data/WriterT.scala
@@ -75,6 +75,11 @@ private[data] sealed abstract class WriterTInstances extends WriterTInstances0 {
   implicit def writerTShow[F[_], L, V](implicit F: Show[F[(L, V)]]): Show[WriterT[F, L, V]] = new Show[WriterT[F, L, V]] {
     override def show(f: WriterT[F, L, V]): String = f.show
   }
+
+  implicit def writerTMonoid[F[_], L, V](implicit W: Monoid[F[(L, V)]]): Monoid[WriterT[F, L, V]] = new Monoid[WriterT[F, L, V]] {
+    def empty = WriterT[F, L, V](W.empty)
+    def combine(x: WriterT[F, L, V], y: WriterT[F, L, V]): WriterT[F, L, V] = WriterT(W.combine(x.run, y.run))
+  }
 }
 
 private[data] sealed abstract class WriterTInstances0 extends WriterTInstances1 {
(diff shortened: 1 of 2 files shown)


yilinwei:
After using in our application code further I'm seeing similar behaviour that @ceedubs is seeing for `MonadError` in #689.

In particular I seem to have some trouble resolving the `Monoid` which should exist for `Id` and `Xor` as the `F[_]`. The kinds with a 'normal' `F[_]` shape (apart from `Id`) such as `Option` and `List` seem to work well.

```
  //explicit call works, implicitly fails, resolve fails
  writerTMonoid[String Xor ?, String, Unit]
  implicitly[Monoid[WriterT[String Xor ?, String, Unit]]]
  resolveMonoid[WriterT[String Xor ?, String, ?]]

  //implicitly and resolve both work
  implicitly[Monoid[WriterT[List, String, Unit]]]
  resolveMonoid[WriterT[List, String, ?]]

  //implicitly works, and resolve fails
  implicitly[Monoid[WriterT[Id, String, Unit]]]
  resolveMonoid[WriterT[Id, String, ?]]

  //implicitly and resolve both work
  implicitly[Monoid[WriterT[Option, String, Unit]]]
  resolveMonoid[WriterT[Option, String, ?]]
```

yilinwei:
I've checked in `scalaz` and can see the same issue with their `WriterT` implementation with `\/` and `Id`. Also, the semigroup `|+|` works with `Id` as `F[_]` but not `Xor` which I guess is vaguely expected since `implicitly` works for `Id` but not `Xor`.

adelbertc:
Is this SI-2712?

yilinwei:
I compiled against @milessabin proposed fix for SI-2712 and it didn't work.

A simplified repro,

```
trait TC[A]
trait Foo[A, B]
trait Bar[F[_], A]

object Check {
  implicit val a: TC[Unit] = ???
  implicit def fooTC[A, B](implicit a: TC[A]): TC[Foo[A, B]] = ???
  implicit def barTC[F[_], A](implicit fa: TC[F[A]]): TC[Bar[F, A]] = ???
  
  //works
  implicitly[TC[Foo[Unit, Unit]]]
  //doesn't work
  implicitly[TC[Bar[({type L[A] = Foo[Unit, A]})#L, Unit]]]
  //works
  barTC[({type L[A] = Foo[Unit, A]})#L, Unit]
}


```

ceedubs:
@yilinwei thanks a lot for getting the ball rolling on this and identifying implicit resolution issues!

I've submitted a PR to your branch that helps with the case of `Id`. If you merge it, then the updates should appear on this PR.

I'm not sure what to do about the `Xor` case. We might just have to punt on it for now.

yilinwei:
I think we can leave the Xor use case for now - I'm not really familiar with codecov.io - what does the check mean and is there anything I should be doing?

ceedubs:
👍 lgtm. I don't entirely understand what's going on with the code coverage report. I think something weird is happening because of the separation of the `free` module that has been merged to master. I wouldn't worry about it too much.

adelbertc:
👍

PR #1015, opened 2016-05-04
Add Apply.map2Eval and allow traverse laziness

### LINKED ISSUE #513: Make traverse lazy
I believe we can make `traverse` lazy by using thunk or `Eval` in `Applicative.map2` (please correct me if I am wrong)

```scala
def validate(i: Int): Option[Int] = if(i < 5) Some(i) else None

List(1,2,6,8).traverse(validate) // should not evaluate validate(8)
```

ceedubs:
Fixes #513.

I think this should solve the majority of use-cases for lazy traversal.
One thing I noticed is that both `reduceRightToOption` and `traverse1_`
(which uses the former) seem to be impossible to implement lazily such
that they allow for short-circuiting. At least I couldn't figure out a
way. I left a note in the `traverse1_` scaladoc, but in practice I don't
anticipate this being a problem, because at the cost of just having an
`Applicative` instance as opposed to `Apply`, you can use `traverse_`
which is sufficiently lazy.

### DIFF
--- a/core/src/main/scala/cats/Foldable.scala
+++ b/core/src/main/scala/cats/Foldable.scala
@@ -107,9 +107,9 @@ import simulacrum.typeclass
    * needed.
    */
   def traverse_[G[_], A, B](fa: F[A])(f: A => G[B])(implicit G: Applicative[G]): G[Unit] =
-    foldLeft(fa, G.pure(())) { (acc, a) =>
-      G.map2(acc, f(a)) { (_, _) => () }
-    }
+    foldRight(fa, Later(G.pure(()))) { (a, acc) =>
+      G.map2Eval(f(a), acc) { (_, _) => () }
+    }.value
 
   /**
    * Behaves like traverse_, but uses [[Unapply]] to find the
--- a/core/src/main/scala/cats/Foldable.scala
+++ b/core/src/main/scala/cats/Foldable.scala
@@ -107,7 +107,7 @@ import simulacrum.typeclass
    * needed.
    */
   def traverse_[G[_], A, B](fa: F[A])(f: A => G[B])(implicit G: Applicative[G]): G[Unit] =
-    foldRight(fa, Later(G.pure(()))) { (a, acc) =>
+    foldRight(fa, Always(G.pure(()))) { (a, acc) =>
       G.map2Eval(f(a), acc) { (_, _) => () }
     }.value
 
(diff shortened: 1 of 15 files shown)


ceedubs:
Another note: it kind of seems like there should be things like `apEval`, `ap2Eval`, etc now, because those are also methods that could benefit from laziness. I opted to omit them for now since `map2` probably satisfies most use-cases. We may want to add them in a future PR though.

ceedubs:
@julien-truffaut would you want to review this since you originally opened up #513?

non:
👍

core/src/main/scala/cats/Foldable.scala
```diff
@@ -107,9 +107,9 @@ import simulacrum.typeclass
    * needed.
    */
   def traverse_[G[_], A, B](fa: F[A])(f: A => G[B])(implicit G: Applicative[G]): G[Unit] =
-    foldLeft(fa, G.pure(())) { (acc, a) =>
-      G.map2(acc, f(a)) { (_, _) => () }
-    }
+    foldRight(fa, Later(G.pure(()))) { (a, acc) =>
```
non: You might want to use `Always` here, since it will be evaluated once and there's no need for memoization.

core/src/main/scala/cats/data/OneAnd.scala
```diff
@@ -170,9 +170,7 @@ trait OneAndLowPriority2 extends OneAndLowPriority1 {
   implicit def oneAndTraverse[F[_]](implicit F: Traverse[F]): Traverse[OneAnd[F, ?]] =
     new Traverse[OneAnd[F, ?]] {
       def traverse[G[_], A, B](fa: OneAnd[F, A])(f: (A) => G[B])(implicit G: Applicative[G]): G[OneAnd[F, B]] = {
-        val tail = F.traverse(fa.tail)(f)
-        val head = f(fa.head)
-        G.ap2[B, F[B], OneAnd[F, B]](G.pure(OneAnd(_, _)))(head, tail)
+        G.map2Eval(f(fa.head), Later(F.traverse(fa.tail)(f)))(OneAnd(_, _)).value
```
non: Same thing here, since we're immediately calling `.value` we can just use `Always`.

non:
👍

julien-truffaut:
👍

PR #1018, opened 2016-05-06
Remove blocking (JVM-only) Future instances

### LINKED ISSUE #589: jvm Future instances do not obey laws

ceedubs:
Fixes #589.

If anyone is set on using these, they can be added to alleycats.
However, I suspect that few to no people are using these. They all throw
exceptions on failed futures, which seems like it would be a
show-stopper in most codebases.


adelbertc:
Build failure? But 👍

ceedubs:
@adelbertc yeah, I don't completely understand the failure. The _build_ passed. The code coverage check is complaining. It seems that it doesn't know how to handle the fact that I removed a couple classes and their tests? I'm not sure.

ceedubs:
I've resolved the merge conflict.

non:
If no one object I'm fine with this. 👍

adelbertc:
👍

mpilquist:
👍

PR #1019, opened 2016-05-06
Use Future#successful in pureEval when possible

dialelo:
Since `Now` already has performed the work required to calculate the result value, we can lift the result into a Future using `Future#successful` instead of unwrapping the `Now` inside the block run by the `Future` (which potentially has to use another thread).


ceedubs:
👍 looks good to me as long as the build passes

adelbertc:
👍

PR #1020, opened 2016-05-07
Add Traverse.traverseM 

peterneyens:
I have added `Traverse.traverseM` and added some examples to the `traverse` syntax.

### DIFF
--- a/core/src/main/scala/cats/Traverse.scala
+++ b/core/src/main/scala/cats/Traverse.scala
@@ -30,6 +30,12 @@ import simulacrum.typeclass
     U.TC.traverse(fa)(a => U.subst(f(a)))(this)
 
   /**
+   * A traverse followed by flattening the inner result.
+   */
+  def traverseM[G[_], A, B](fa: F[A])(f: A => G[F[B]])(implicit G: Applicative[G], F: FlatMap[F]): G[F[B]] =
+    G.map(traverse(fa)(f))(F.flatten)
+
+  /**
    * Thread all the G effects through the F structure to invert the
    * structure from F[G[A]] to G[F[A]].
    */
--- a/core/src/main/scala/cats/syntax/traverse.scala
+++ b/core/src/main/scala/cats/syntax/traverse.scala
@@ -16,17 +16,98 @@ trait TraverseSyntax extends TraverseSyntax1 {
 }
 
 final class TraverseOps[F[_], A](fa: F[A])(implicit F: Traverse[F]) {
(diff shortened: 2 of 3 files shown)


core/src/main/scala/cats/Traverse.scala
```diff
@@ -30,6 +30,12 @@ import simulacrum.typeclass
     U.TC.traverse(fa)(a => U.subst(f(a)))(this)
 
   /**
+   * A traverse followed by flattening the inner result.
+   */
+  def traverseM[G[_], A, B](fa: F[A])(f: A => G[F[B]])(implicit G: Applicative[G], F: FlatMap[F]): G[F[B]] =
+    G.map(traverse(fa)(f))(F.flatten)
```
peterneyens: We got `F` (`FlatMap[F]`) as an `Applicative[F]`, but we could also use the `Traverse` itself : `G.map(traverse(fa)(f)(this))(F.flatten)`.

I am not sure if there is a reason to choose one over the other ?

ceedubs: I think the version that you currently have is probably better, because `Applicative.traverse` ends up delegating through to `Traverse.traverse` anyway.

ceedubs:
👍 thanks!

core/src/main/scala/cats/Traverse.scala
adelbertc: Should this be called `traverseA` ? Consistent with `replicateA`. I think the `M` suffix is usually used when it has a `Monad` constraint.

ceedubs: @adelbertc The `Applicative[G]` is required on `traverse` also. Here I think of the `M` as "monadic", referring to the `FlatMap[F]`. I guess we could call it `traverseF` instead, though that runs the risk of being confused with `Functor`.

peterneyens: I was not really sure about the `M`. `traverseA` seems a bit strange because we need `FlatMap`. `traverseF` could also be confusing with the `flatMapF` and `mapF` on (some) monad transformers.

Maybe we can call it `traverseFlatten` ?

adelbertc: Oh I totally did not see the `FlatMap` requirement.. the size of my screen stopped right at `Applicative` so I assumed that was all there was. I am good with `traverseM` in that case

adelbertc:
👍

PR #1022, opened 2016-05-08
Add law-checking for asMeetPartialOrder and asJoinPartialOrder

ceedubs:


ceedubs:
hmm I don't understand why this didn't affect code coverage. I'll have to
investigate.

ceedubs:
I think that the `build`/`validate` aliases in the `build.sbt` aren't quite right. I'm testing some changes locally and should be adding a commit to this PR shortly.

non:
Sounds good. I'm 👍 on this PR either way.

ceedubs:
Okay, I'm not quite sure what codecov is complaining about now, but it looks like the build changes resulted in +6% coverage being reported.

adelbertc:
:+1: woot woot!

PR #1024, opened 2016-05-10
Override reverse on reversed PartialOrder to return original instance

ceedubs:
Also add some law-checking for reversed and double-reversed
Order/PartialOrder instances.


adelbertc:
I like it 👍

fthomas:
:+1:

PR #1025, opened 2016-05-10
Streamline Travis build

ceedubs:
This commit changes a couple things related to the Travis build.

- For PR builds, don't run `validateJVM` again after we have already run
  it with code-coverage enabled. It's true that the coverage tool
  instruments the byte code, which could conceivably change behavior,
  but in the off-chance that that happens, it should be caught by the
  master-branch build. We should be able to trim the PR build times by a
  fair amount by not running all of the JVM-related tasks twice.
- Use the `validateJS` alias within the Travis build. Previously the
  Travis build was basically duplicating the `validateJS` behavior (but
  notably missing certain things, such as running the the kernel-laws
  tests.

I don't know a good way to test these changes, so scrutiny is appreciated. In particular I'm hoping that @mikejcurry, @InTheNow, and @non might be able to take a look, since I believe they have all worked on this part of the build.

### DIFF
--- a/scripts/travis-publish.sh
+++ b/scripts/travis-publish.sh
@@ -11,25 +11,22 @@
 # 3. The validateJVM target is executed again, due to the fact that producing coverage with the
 #    code coverage tool causes the byte code to be instrumented/modified to record the coverage
 #    metrics when the tests are executing. This causes the full JVM build to be run a second time.
+#    This step is omitted in PR builds to speed them up. In the off-chance that the coverage tool
+#    changes the behavior of code by instrumenting it, the master build should be able to catch
+#    the issue before a release happens.
 
 # Example setting to use at command line for testing:
 # export TRAVIS_SCALA_VERSION=2.10.5;export TRAVIS_PULL_REQUEST="false";export TRAVIS_BRANCH="master"
 
-export publish_cmd="publishLocal"
+sbt_cmd="sbt ++$TRAVIS_SCALA_VERSION"
+coverage_cmd="$sbt_cmd coverage validateJVM coverageReport && codecov"
 
 if [[ $TRAVIS_PULL_REQUEST == "false" && $TRAVIS_BRANCH == "master" && $(cat version.sbt) =~ "-SNAPSHOT" ]]; then
-  export publish_cmd="publish gitSnapshots publish"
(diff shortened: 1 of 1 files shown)


ceedubs:
Update now that the Travis build has finished:

The build time improvement is not huge, but I think it's significant. It looks like there's ~10 min gain in the combined 2.10 + 2.11 build time compared to other PRs. However, I think it's a bit better than that, because I think that the kernel laws were not previously tested for JS.

mikejcurry:
LGTM - the artifacts with coverage instrumentation will only be published locally on Travis - which should be fine.

The only thing is it would be important for folks building for release locally not to use this script as if they did, the version would not be -SNAPSHOT, so the artifacts with coverage instrumentation would be built.

Just mentioning this because I'm not sure how people build when they are releasing the final artifacts for a particular release - but this script could be a problem if used.

Other than that 👍

InTheNow:
@ceedubs Just so that you know the history (in case it repeats itself)

The reason we had different build settings for the scalajs build was that `validateJS` run in one instance of sbt, whereas the script chunks the build into multiple sbt sessions to prevent OOM -  hence the `&&` in `$sbt_cmd macrosJS/compile coreJS/compile lawsJS/compile && $sbt_cmd testsJS/test && $sbt_cmd js/test"` and the final run command.

Now, if thing work fine using `validateJS`, then that is of course the preferred solution. But some of the speed improvements that you see could well be down to thîs (ie fewer sbt start/stops), rather than the de-duplication of `validateJVM`.

> but in the off-chance that that happens, it should be caught by the master-branch 

I totally take your point, but professional build managers go ballistic at things like that.....Is the time reduction in absolute terms so huge? Ideally, the only difference between a release build and a non-release build is just the release part.

InTheNow:
Also, regarding scala.js build times. 

I'm pretty sure https://github.com/typelevel/cats/blob/master/build.sbt#L74 won't work as https://github.com/typelevel/cats/blob/master/.travis.yml#L30 defines an env var, not a sys property

ceedubs:
@mikejcurry @InTheNow thank you for the detailed reviews. You both raise some good points, and at the moment I'm inclined to say that the minor improvement in build time isn't worth the risk this introduces. I'm going to close this out, but I encourage anyone who has better ideas to feel free to submit then in a PR.

PR #1027, opened 2016-05-11
Remove unnecessary nelSemigroup from traverse doc

ceedubs:
The necessary `Semigroup` instance has been added to cats, so this is no
longer necessary.


adelbertc:
:+1: + merge because doc

Issue #987, opened 2016-04-18
Coproduct#fold


adelbertc:
Like `Xor#fold` but for functors

ceedubs:
@adelbertc just to be clear for someone who might want to work on this - this would take two `NaturalTransformation`s as opposed to ordinary functions, right?

adelbertc:
Yep

PR #1032, opened 2016-05-14
Add Coproduct fold

peterneyens:
Add the `fold` operation for `Coproduct` mentioned in #987.
I am not sure how I could test this ?

### DIFF
--- a/core/src/main/scala/cats/data/Coproduct.scala
+++ b/core/src/main/scala/cats/data/Coproduct.scala
@@ -1,6 +1,7 @@
 package cats
 package data
 
+import cats.arrow.NaturalTransformation
 import cats.functor.Contravariant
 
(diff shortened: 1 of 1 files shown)


ceedubs:
Thanks @peterneyens. I think this method is pretty much guaranteed to be implemented correctly if the types line up, so I'm not too worried about testing it. But it would also be nice to not have it reported as uncovered code. Would you be interested in writing an `sbt-doctest` example in the scaladoc? I think that would kind of knock out two birds with one stone, because it may be helpful for people and it should count as test coverage. You can could use simple `NaturalTransformation`s such as the `listToOption` one that is used in `NaturalTransformationTests`.

adelbertc:
👍

ceedubs:
Awesome. Thanks!

PR #1034, opened 2016-05-14
Don't combine lefts on Xor and XorT combine

### LINKED ISSUE #888: Confusing behavior of Monoid[Xor[A, B]]#combine
Don’t you feel that it’s strange that `Monoid[Xor[A, B]]` requires `A: Semigroup`?
I expected `combine` to be implemented like this
```
def combine(x: A Xor B, y: A Xor B): A Xor B =
  x.flatMap { xb =>
    y.map { yb => Monoid[B].combine(xb, yb) }
  }
```
In current implementation it also tries to combine `Left`s

ceedubs:
Resolves #888

I don't think there's really a _right_ answer here. Both methods of
combining `Xor`s are straightforward and law-abiding. However, I think
that since in pretty much every other context (including the
`SemigroupK` instances), `Xor` does not combine failures and `Validated`
does, it is less surprising behavior to not combine left values in
`combine` methods for `Xor` and `XorT`.

Some notes about related implementations:
- Scalaz's respective methods (and semigroup instances) _do_ combine
  errors, so this is a deviation from that.
- The `Semigroup` for `Either` in Haskell has doesn't combine values at
  all, but returns the first `Right` (if any), making it equivalent to
  the behavior of `orElse` and the `SemigroupK` instance for `Xor`.
  Since we have already decided to not go with `orElse`-like behavior
  for our `Option` semigroup, I'm inclined to not take this approach for
  `Xor`.

See also
https://github.com/typelevel/cats/pull/996#issuecomment-214010455

cc @notxcain @aaronlevin @travisbrown @tpolecat @stew whom I think have all weighed in on this or similar items in the past

### DIFF
--- a/core/src/main/scala/cats/data/Xor.scala
+++ b/core/src/main/scala/cats/data/Xor.scala
@@ -167,9 +167,9 @@ sealed abstract class Xor[+A, +B] extends Product with Serializable {
    * }}}
    */
   final def combine[AA >: A, BB >: B](that: AA Xor BB)(implicit BB: Semigroup[BB]): AA Xor BB = this match {
-    case l @ Xor.Left(_) => l
+    case left @ Xor.Left(_) => left
     case Xor.Right(b1) => that match {
-      case l @ Xor.Left(_) => l
+      case left @ Xor.Left(_) => left
       case Xor.Right(b2) => Xor.Right(BB.combine(b1, b2))
     }
   }
--- a/core/src/main/scala/cats/data/Xor.scala
+++ b/core/src/main/scala/cats/data/Xor.scala
@@ -133,13 +133,43 @@ sealed abstract class Xor[+A, +B] extends Product with Serializable {
 
   def merge[AA >: A](implicit ev: B <:< AA): AA = fold(identity, ev.apply)
 
-  final def combine[AA >: A, BB >: B](that: AA Xor BB)(implicit AA: Semigroup[AA], BB: Semigroup[BB]): AA Xor BB = this match {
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/data/Xor.scala
```diff
@@ -133,13 +133,43 @@ sealed abstract class Xor[+A, +B] extends Product with Serializable {
...
+   *
+   * scala> r3 combine r4
+   * res3: Xor[String, Int] = Right(7)
+   * }}}
+   */
+  final def combine[AA >: A, BB >: B](that: AA Xor BB)(implicit BB: Semigroup[BB]): AA Xor BB = this match {
+    case l @ Xor.Left(_) => l
     case Xor.Right(b1) => that match {
-      case Xor.Left(a2) => Xor.Left(a2)
+      case l @ Xor.Left(_) => l
```
non: Tiny thing, but let's call this `left` instead of `l` -- I think it reads a lot better.

non:
Tiny style comment, but 👍 from me!

aaronlevin:
I'm in favour. So, this makes `combine` semantically equivalent to:

```scala
for {
   x <- this
   y <- that
} yield SemigroupInstance.combine(x,y)
```

which make sense as you mostly want to use `Xor` like a monad.

core/src/main/scala/cats/data/Xor.scala
ceedubs: 👍 updated

non:
👍

PR #1036, opened 2016-05-15
Add FPiS to the "Resources for Learners" section

fthomas:
This adds Functional Programming in Scala to the "Resources for Learners" section in our docs. FPiS does not specifically cover Cats but it is IMO one of the best introduction to functional programming in general and to some of the abstractions that are provided by Cats.


adelbertc:
:+1:

PR #1040, opened 2016-05-16
Add FlatMapRec, MonadRec.

TomasMikula:
See #616.

PR #1041, opened 2016-05-17
Some MonadRec instances.

TomasMikula:
Some `MonadRec` instances to get started. This is a follow-up on #1040.

### DIFF
--- a/tests/src/test/scala/cats/tests/MonadRecInstancesTests.scala
+++ b/tests/src/test/scala/cats/tests/MonadRecInstancesTests.scala
@@ -1,0 +1,41 @@
+package cats
+package tests
+
+import cats.data.{OptionT, Xor, XorT}
+
+class MonadRecInstancesTests extends CatsSuite {
+  def tailRecMStackSafety[M[_]](implicit M: MonadRec[M], Eq: Eq[M[Int]]): Unit = {
+    val n = 50000
+    val res = M.tailRecM(0)(i => M.pure(if(i < n) Xor.Left(i + 1) else Xor.Right(i)))
+    res should === (M.pure(n))
+  }
+
+  test("tailRecM stack-safety for Id") {
+    tailRecMStackSafety[Id]
+  }
+
+  test("tailRecM stack-safety for Option") {
+    tailRecMStackSafety[Option]
+  }
+
+  test("tailRecM stack-safety for OptionT") {
+    tailRecMStackSafety[OptionT[Option, ?]]
+  }
+
+  test("tailRecM stack-safety for Either") {
+    tailRecMStackSafety[Either[String, ?]]
+  }
+
+  test("tailRecM stack-safety for Xor") {
+    tailRecMStackSafety[String Xor ?]
+  }
+
+  test("tailRecM stack-safety for XorT") {
+    tailRecMStackSafety[XorT[Option, String, ?]]
+  }
+
+  test("tailRecM stack-safety for List") {
+    tailRecMStackSafety[List]
+  }
+
+}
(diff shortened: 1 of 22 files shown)


tests/src/test/scala/cats/tests/MonadRecInstancesTests.scala
```diff
@@ -0,0 +1,37 @@
+package cats
+package tests
+
+import cats.data.{OptionT, Xor, XorT}
+
+class MonadRecInstancesTests extends CatsSuite {
+  def tailRecMStackSafety[M[_]](implicit M: MonadRec[M], Eq: Eq[M[Int]]): Unit = {
+    val n = 50000
```
adelbertc: Can this test be added as a `MonadRecLaws` and `MonadRecTests` in the `cats-laws` project, with tests for instances being put in `OptionTests`, `XorTests`, etc. ?

ceedubs:
@mpilquist ScalaDoc generation fails on [this line](https://github.com/typelevel/cats/pull/1041/files#diff-c7e2be9a5f6ea4caec6c3cc3e2b5745cR5) in scala 2.10 but succeeds if I remove the `@typeclass` annotation. I don't see this listed as a known issue with simulacrum. Any idea what's going on there?

ceedubs:
Thanks @TomasMikula! This looks like a good start to me.

tests/src/test/scala/cats/tests/MonadRecInstancesTests.scala
ceedubs: A law that `tailRecM` is consistent with `flatMap` would be nice.

TomasMikula: @adelbertc Is it OK to call that unit test a law?

@ceedubs The best I can think of is testing that `tailRecM` is consistent with the naive (non-stack safe) implementation of `tailRecM` that uses recursive `flatMap`s (and is implemented uniformly for any monad)

```scala
def naiveTailRecM[M[_], A, B](a: A)(f: A => M[A Xor B])(implicit M: Monad[M]): M[B] =
  M.flatMap(f(a))(_ match {
    case Xor.Left(a1) => naiveTailRecM(a1)(f)
    case Xor.Right(b) => M.pure(b)
  })
```

(This is also basically the implementation of `tailRecM` for `Free`, where it is OK, because `Free` is lazy.)

Did you have something else in mind?

ceedubs: @TomasMikula I was thinking of something like what [scalaz has](https://github.com/scalaz/scalaz/blob/series/7.3.x/core/src/main/scala/scalaz/BindRec.scala#L32).

Regarding @adelbertc's comment, I have a little hesitation adding something like that as a law. Mostly because it could potentially take a really long time to run for some types. I'm not writing it off as a possibility, just raising a potential concern.

TomasMikula: @ceedubs laws added.

ceedubs: @TomasMikula nice! Thank you for this. I'll do a more detailed review shortly.

adelbertc:
Unsure why it's failing but :+1: once we figure that out

ceedubs:
The build failure seems to be the same as https://github.com/typelevel/cats/pull/898#issuecomment-189621115

kailuowang:
As @InTheNow pointed out, the newly updated sbt 0.13.11 with `-Ymacro-no-expand`, might fix scala 2.10 javadoc problem. We might be able to have this PR passing the build by just updating to the latest master.

TomasMikula:
@adelbertc Do you want to review #1051 which should unblock this one?

ceedubs:
@TomasMikula #1051 has been merged. Sorry, but would you be willing to rebase?

non:
Great! 👍 when Travis is green.

PR #1043, opened 2016-05-17
Add notes about kind-projector usage in docs

### LINKED ISSUE #1033: Avoid kind-projector in docs where possible
The kind-projector usage in the docs trips up a lot of people (as evident by it frequently coming up in Gitter questions). I think that where possible we should use simple type aliases instead of kind-projector syntax.

If we have any cases where a simple type alias isn't possible, I like using kind-projector instead of a type lambda, but I think that we should preface the first such usage on each page with an explanation that kind-projector is being used.

ceedubs:
Resolves #1033 although in all but one place, a simple type alias was
not possible.


non:
👍

adelbertc:
👍

PR #1044, opened 2016-05-17
Add updating site and release notes to release process

ceedubs:
At the Typelevel Summit in Philly, people brought up confusion around
the site not always being up to date with the latest release.

Also, we always seem to forget to update the release notes until
@travisbrown kindly volunteers to :)

cc @non @travisbrown


ceedubs:
I guess we are more likely to need "Edit release" than "Draft a new release". I probably need to change this.

non:
Yeah, maybe add that if it happens after the release, that we can just edit it instead.

But 👍 from me.

PR #1045, opened 2016-05-17
Add a link to the OptionT documentation from the monad docs.

peterneyens:
The monad docs uses `OptionT` as an example about how to compose (some) monads, but doesn't include a link to the `OptionT` documentation.


ceedubs:
👍 thanks!

PR #1046, opened 2016-05-17
summon `ApplicativeErrorSyntax` for `F[_]` instead of `F[_, _]`

kailuowang:


core/src/main/scala/cats/syntax/applicativeError.scala
```diff
@@ -3,12 +3,18 @@ package syntax
...
+  implicit def applicativeErrorSyntax[F[_], E, A](fa: F[A])(implicit F: ApplicativeError[F, E]): ApplicativeErrorOps[F, E, A] =
+    new ApplicativeErrorOps[F, E, A](fa)
+}
+
+trait ApplicativeErrorSyntax extends ApplicativeErrorSyntax1 {
   implicit def applicativeErrorIdSyntax[E](e: E): ApplicativeErrorIdOps[E] =
     new ApplicativeErrorIdOps(e)
 
-  implicit def applicativeErrorSyntax[F[_, _], E, A](fa: F[E, A])(implicit F: ApplicativeError[F[E, ?], E]): ApplicativeErrorOps[F[E, ?], E, A] =
+  implicit def applicativeErrorSyntax2[F[_, _], E, A](fa: F[E, A])(implicit F: ApplicativeError[F[E, ?], E]): ApplicativeErrorOps[F[E, ?], E, A] =
```
julienrf: I’d rather replace this one with a variant using `Unapply`.

kailuowang: To me `Unapply` is hard to read, I thought this variant does the same thing, so if you don't mind, could you educate what's the benefit of using `Unapply` over this variant?

julienrf: `Unapply` is a more general solution to the type inference problem that you are currently solving with this PR. But it will (hopefull soon!) become useless because [SI-2712](https://issues.scala-lang.org/browse/SI-2712) is going to [be fixed](https://github.com/scala/scala/pull/5102).

Once this will be the case, projects like cats will follow some migration path to get rid of the current workarounds. One possible migration path could be to move every `Unapply`-related method to an external project that could be kept maintained just for compatibility with old Scala versions.

So, using `Unapply` clearly marks that the implicit definition will be useless once SI-2712 is fixed.

kailuowang: @julienrf, being consistent is a good point although I would imagine you want to cross build this code, meaning, I am not sure you can to simply remove the `Unapply` usage in the code base.  
@ceedubs mentioned in the chatroom that we don't want to over invest in `Unapply` post the SI-2712 fix. That being said, I am totally willing to turn this into `Unapply`, if that's the consensus. 
Maintainers any thoughts?

ceedubs: I am hopeful that `Unapply`'s days are numbered. However, if we _are_ going to add this sort of thing to help with inference (which is probably its own question now that the [SI-2712 fix plugin](https://github.com/milessabin/si2712fix-plugin) exists), I think that we should reuse `Unapply` instead of less general workarounds like this one. Having said that, I briefly tried hooking up `Unapply` to this and was having trouble getting the syntax to be recognized for `Xor` - maybe someone else can do a better job than I did?

kailuowang: hmm.. actually I couldn't see a way for `Unapply` to work here as well. In `Unapply` the actual type `TC` only takes one type parameter `M` (which is a high-kinded type itself), while in our case the actual type is `ApplicativeError[F, E]`, which takes not only a highkinded type parameter, but also another type parameter `E`. So I think `Unapply` as is won't help here, and if I am right, we either keep this limited workaround (it's just a syntax helper, so I don't really see more complicated inference needed most of the cases), or we just do as you said, get rid of the support for `F[_, _]` and let people use Si-2712 fix for that.

ceedubs: I thought that I'd be able to use something like `Unapply[({type f[l[_]] = ApplicativeError[l, E]})#l, FA]`, and that compiles, but it doesn't seem to result in the syntax actually being picked up.

kailuowang: hmmm. I couldn't get that working either.

kailuowang: @ceedubs do we want to do the originally proposed change, which is, just change it to `F[_]`?

ceedubs: @kailuowang yeah, that sounds reasonable to me. Have you by any chance checked that works when using the SI-2712 fix?

@travisbrown do you have any strong opinions on this? I think you originally added the applicative error syntax.

kailuowang: Just checked with the SI-2712 fix, it works fine (can find syntax for `F[_, _]` with only one `F[_]` variant.  
Removing the existing `F[_, _]` variant, however, would mean we need to change existing syntax tests to use `Option` instead of `Xor`, right?

julienrf: Yes, or introduce a type alias.

kailuowang: per @travisbrown at gitter chat, I am going to replace the `F[_, _]` with `F[_]` and change tests to use `Option`

travisbrown:
I don't have a strong opinion here, and I've never used the `ApplicativeError` syntax (it looks like it was introduced by @davegurnell, so ideally I guess we'd want to get him to sign off on this change), but off-hand I prefer the `F[_]` version.

non:
Seems reasonable to me. 👍

ceedubs:
👍

philwills:
Thanks @kailuowang, I think this could be really useful.

PR #1050, opened 2016-05-20
Use pygments instead of rouge for syntax highlighting

ceedubs:
It appears that GitHub is already using rouge instead of pygments for
the cats site, because it doesn't support pygments. This change is so I
stop getting the following email every time I push the cats website:

> You are attempting to use the 'pygments' highlighter, which is currently
> unsupported on GitHub Pages. Your site will use 'rouge' for highlighting
> instead. To suppress this warning, change the 'highlighter' value to
> 'rouge' in your '_config.yml' and ensure the 'pygments' key is unset.

### DIFF
--- a/docs/src/site/_config.yml
+++ b/docs/src/site/_config.yml
@@ -1,6 +1,6 @@
 name: Cats Documentation
 markdown: redcarpet
-highlighter: pygments
+highlighter: rouge
 baseurl: /cats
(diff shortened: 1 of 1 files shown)


non:
Seems reasonable. 👍

non:
I claim this change is a documentation change that only requires 1 sign-off so I'm merging now.

PR #1051, opened 2016-05-21
skip javadoc compilation for 2.10

kailuowang:
Hopefully this would unblock #898 and #1041

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -82,7 +82,7 @@ lazy val commonJvmSettings = Seq(
 // JVM settings. https://github.com/tkawachi/sbt-doctest/issues/52
 ) ++ catsDoctestSettings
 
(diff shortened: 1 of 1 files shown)


ceedubs:
Thanks @kailuowang - I'm glad to hopefully see the end of the doc troubles. Would you mind adding a brief comment about why this is being done?

kailuowang:
@ceedubs just consolidated with the existing skip-doc-at-210 settings together with its comment.

ceedubs:
👍

kailuowang:
per conversation with @InTheNow at gitter, we might be able find a better solution than this. Javadoc generation is probably fine with sbt 0.13.11 (turned on `-Ymacro-no-expand` by default). It's the makeSite that fails for laws

non:
Yeah, this makes sense. We should open an issue to re-enable these once @InTheNow's fixes land in SBT. 👍

PR #1052, opened 2016-05-21
Add a Why section to README

adelbertc:
Inspired by a colleague who had a bit of trouble figuring out what problem Cats was solving

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,10 +16,10 @@ The name is a playful shortening of the word *category*.
 
 #### Why?
 
-Scala provides many tools for functional programming such as algebraic data types, higher kinded types
-and implicits. However, common abstractions such as `Monoid`, `Monad`, and `Traverse` are currently missing from
-the standard library, leaving a bit to be desired. Among the goals of Cats is to fill this gap, and to provide a
-foundation for an [ecosystem of pure, typeful libraries](http://typelevel.org/projects/).
+Scala supports both object-oriented and functional programming, and this is reflected in the hybrid approach of the
+standard library. Cats augments the standard library with tools that further enable functional programming such as
+`Validated`, `Monad`, and `Traverse`. A broader goal of Cats is to provide a foundation for an
+[ecosystem of pure, typeful libraries](http://typelevel.org/projects/).
 
 ### Getting Started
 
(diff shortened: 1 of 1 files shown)


non:
Looks reasonable to me. 👍

DavidGregory084:
I think maybe you could expand on the phrase `leaving a bit to be desired` as this does not seem to describe the problem with these abstractions being absent - maybe something like `limiting the level of abstraction and therefore the reusability of code written against the standard library alone`?

README.md
```diff
@@ -14,6 +14,13 @@ The name is a playful shortening of the word *category*.
 
 ![cats image](http://plastic-idolatry.com/erik/cats2.png)
 
+#### Why?
+
+Scala provides many tools for functional programming such as algebraic data types, higher kinded types
+and implicits. However, common abstractions such as `Monoid`, `Monad`, and `Traverse` are currently missing from
+the standard library, leaving a bit to be desired. Among the goals of Cats is to fill this gap, and to provide a
+foundation for an [ecosystem of pure, typeful libraries](http://typelevel.org/projects/).
```
ceedubs: Warning: bike shedding - I'm sorry.

> However, common abstractions such as `Monoid`, `Monad`, and `Traverse` are currently missing from the standard library, leaving a bit to be desired.

The word _missing_ here suggests that `Monoid` etc _should_ be in the standard library, which is an opinion that many may hold, but at the end of the day _is_ an opinion. I'd be inclined to focus on the positive rather than the negative. What do you think of something like the following?

The scala language supports both object-oriented and functional programming, and this is reflected in the hybrid OO/FP approach of its standard library. Cats is an exploration into what the standard library might look like if it focused mostly on functional programming. It takes advantage of scala features such as implicits and higher-kinded types to provide type classes such as [Monoid](http://typelevel.org/cats/tut/monoid.html), [Monad](http://typelevel.org/cats/tut/monad.html), and [Traverse](http://typelevel.org/cats/tut/traverse.html), which provide composable abstractions for common use-cases. While Cats encourages reuse of some standard library data structures such as `List` and `Option`, it also provides some of its own algebraic data types such as [Xor](http://typelevel.org/cats/tut/xor.html) and [Validated](http://typelevel.org/cats/tut/validated.html). `<Typelevel plug here>`

adelbertc:
@ceedubs Good point, reworded it a bit, lmk what you think

zmccoy:
Thanks for adding this.

ceedubs:
👍

PR #1056, opened 2016-05-23
Left-adjoint of forgetful functor

Madder:
Isn't it supposed to be the free functor?

### DIFF
--- a/docs/src/main/tut/freemonad.md
+++ b/docs/src/main/tut/freemonad.md
@@ -420,14 +420,14 @@ _very simple_ Monad from any _functor_**.
 The above forgetful functor takes a `Monad` and:
 
  - forgets its *monadic* part (e.g. the `flatMap` function)
- - forgets its *applicative* part (e.g. the `pure` function)
+ - forgets its *pointed* part (e.g. the `pure` function)
  - finally keeps the *functor* part (e.g. the `map` function)
 
 By reversing all arrows to build the left-adjoint, we deduce that the
-forgetful functor is basically a construction that:
+free monad is basically a construction that:
 
  - takes a *functor*
- - adds the *applicative* part (e.g. `pure`)
+ - adds the *pointed* part (e.g. `pure`)
  - adds the *monadic* behavior (e.g. `flatMap`)
 
 In terms of implementation, to build a *monad* from a *functor* we use


docs/src/main/tut/freemonad.md
```diff
@@ -424,7 +424,7 @@ The above forgetful functor takes a `Monad` and:
  - finally keeps the *functor* part (e.g. the `map` function)
 
 By reversing all arrows to build the left-adjoint, we deduce that the
-forgetful functor is basically a construction that:
+free functor is basically a construction that:
```
non: I think in this context the author really meant this to read *free monad*. Thanks for pointing out the problem. Does this make sense to you?

Madder:
I agree. He certainly means Free monad in this context. Free functor being the more general concept of a left-adjoint of a forgetful functor. Another issue I see is that pure is not the "applicative" part, as both pure and flatmap are part of the monad structure: applicative is simply a by product of being a monad.

ceedubs:
@Madder thanks for pointing this out.

It sounds like we all agree that it should be changed to _free monad_.

As for at the "applicative part" - I can see the author's point (no pun intended, but I'll go with it) that by taking a `Functor` and adding `pure` you have an `Applicative`. But if you have ideas on how to improve the wording, they would be welcomed!

Would you be interested in submitting a PR for any changes you'd like to see?

Madder:
@ceedubs Possibly a stupid question, but don't you need an apply : F[A=>B] => F[A]  => F[B] or a map2 : (F[A],F[B]) => ((A,B) => C) => F[C] to get an Applicative out of a Functor? Or is pure enough somehow?

ceedubs:
@Madder oops you are quite right :)

Madder:
@ceedubs I've updated the PR with _free monad_ instead _free functor_. I've also replaced _applicative_ with _pointed_ which is more accurate but I'm not sure whether it fits in that context anymore... Let me know if you prefer another formulation. I can change the PR accordingly or submit a new one if preferable.

ceedubs:
@Madder thank you. If someone wants to further refine this, they should feel free to, but 👍 from me.

non:
👍 looks good to me. thanks for working on this!

PR #1057, opened 2016-05-23
Make code samples in the documentation copy-paste friendly

dialelo:
Changes `tut` blocks to use `tut:book` instead, making them copy-paste friendly. This way people following along the documentation can copy and paste examples in a REPL and try things out.

Before:
tut

After:
tutbook


ceedubs:
👍 I find it to be a hassle when I want to copy/paste an example as a starting point and end up needing to make a bunch of minor edits to get it to compile.

raulraja:
:+1: same here!

julien-truffaut:
👍

PR #1060, opened 2016-05-24
fix typo

levinotik:


ceedubs:
"most certainly" sounds right to me - is this another one of those things that only sounds good to me because I'm from Indiana? :)

rossabaker:
"Most certainly" typechecks as an absolute superlative.  "Must certainly" also compiles, and could be optimized to "must".

Disclaimer: I spent the last 13 years immersed in Cody's dialect. :trollface:

levinotik:
Wow, I think I hadn't had enough coffee when I submitted this and my eyes completely skipped over the word `certainly` so I read it as `we most have`.

PR #1062, opened 2016-05-25
Fixed a bug in the Order and PartialOrder instances for Tuple2+ where only the first element was used in comparisons

mpilquist:
Before this fix:

```
import cats.implicits._

scala> (1 -> 2) === (1 -> 3)
res0: Boolean = true

scala> import scala.reflect.runtime.universe._
import scala.reflect.runtime.universe._

scala> showCode(reify { (1 -> 2) === (1 -> 3) }.tree)
res1: String = implicits.eqSyntax(Predef.ArrowAssoc(1).->(2))(implicits.tuple2Order(implicits.intOrder, implicits.intOrder)).===(Predef.ArrowAssoc(1).->(3))
```

This is because we pick up the `Order[Tuple2[..]]` instance, which is erroneously defined like this:

```
  implicit def tuple2Order[A0, A1](implicit A0: Order[A0], A1: Order[A1]): Order[(A0, A1)] =
    new Order[(A0, A1)] {
      def compare(x: (A0, A1), y: (A0, A1)): Int =
        A0.compare(x._1, y._1)
    }
```

I think we should probably publish a 0.6.1 with this fix too as it is really nasty -- code compiles but values end up being report equal which aren't.

/cc @non

### DIFF
--- a/project/KernelBoiler.scala
+++ b/project/KernelBoiler.scala
@@ -147,13 +147,13 @@ object KernelBoiler {
         -  implicit def tuple${arity}Order[${`A..N`}](implicit ${constraints("Order")}): Order[${`(A..N)`}] =
         -    new Order[${`(A..N)`}] {
         -      def compare(x: ${`(A..N)`}, y: ${`(A..N)`}): Int =
-        -        ${binMethod("compare").find(_ != 0).getOrElse(0)}
+        -        ${binMethod("compare").mkString("Array(", ", ", ")")}.find(_ != 0).getOrElse(0)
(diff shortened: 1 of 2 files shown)


non:
👍 ugh thanks for the catch!

non:
(we need more tests for the generated code!)

ceedubs:
Yikes :(. Thanks!

ceedubs:
It'd probably be nice to put this into a patch release. I won't have much availability to help with that for the next few days though :(

Issue #1061, opened 2016-05-24
Uniquify implicit names


non:
Paul Phillips mentions that in a widely-used library it's important to minimize the chance that the names of its implicits will be used by others. To that end I suggest that we review our implicits and rename them according to the following rules:

1. Make sure that the implicits start with *cats* -- it's less likely that a third-party will do this.
2. Make sure that both the type and the type class are mentioned in the name.

So for example, for `Monoid[List[A]]` the name should be something like `catsMonoidForList` or similar.

non:
Since a single across-the-board PR is likely to conflict with every other open PR, we should either accomplish this piecemeal, or coordinate ahead of time. If you want to work on this issue please comment here so we can avoid duplicating work and plan for a minimum amount of conflict.

ceedubs:
As an example of how name collisions can be problematic, in [this](https://gitter.im/typelevel/cats?at=5743073fd3f431720bb2e52c) gitter conversation, an upgrade from cats 0.5 to 0.6 broke someone's code because of a collision on the name `listMonoid` (they had created one for `java.util.List`).

dwijnand:
Note that not only implicits that provide typeclass instances, but also syntax/enrich-my-type implicits suffer from the same exact problem.

A banal example is having two different implicit classes from different libraries called `AnyOps`.

InTheNow:
Would this also be an issue for simulacrum generated implicits? 
For example, for `Invariant`, we get `implicit def toInvariantOps`. Should that be `implicit def toCatsFunctorInvariantOps`? 

N.B. Note the inclusion of the full package names, should this be standard as well?

milessabin:
> Would this also be an issue for simulacrum generated implicits?

Potentially yes.

Bear in mind that this can't be fixed by using `freshName` in a macro, because the names, even if they're never typed by a human, are public definitions and have to be stable between builds. There's some [code in export-hook](https://github.com/milessabin/export-hook/blob/master/core/src/main/scala/export/export.scala#L110-L114) to deal with this.

kailuowang:
I can work on this one. How about just tackle the std first?

ceedubs:
@kailuowang do you think that this is done after #1066 and #1068? Or should we cover the other modules too? Collisions outside of the `std` package seem unlikely to me since people don't have to import the instances explicitly, but I guess there is something to be said about consistency.

kailuowang:
[...]

PR #1066, opened 2016-05-27
I1061 part 1 - renames in std package

kailuowang:


core/src/main/scala/cats/std/anyval.scala
```diff
@@ -14,37 +14,37 @@ trait AnyValInstances
   with    TupleInstances
 
 trait IntInstances extends cats.kernel.std.IntInstances {
-  implicit val intShow: Show[Int] = Show.fromToString[Int]
+  implicit val catsShowForcatsForInt: Show[Int] = Show.fromToString[Int]
```
ceedubs: You might have gone a little overboard here ;)

InTheNow: `catsShowForcatsForInt` => `catsShowForInt`

core/src/main/scala/cats/std/either.scala
```diff
@@ -23,7 +23,7 @@ trait EitherInstances extends EitherInstances1 {
         }
     }
 
-  implicit def eitherInstances[A]: Monad[Either[A, ?]] with Traverse[Either[A, ?]] =
+  implicit def catsMonadForEither[A]: Monad[Either[A, ?]] with Traverse[Either[A, ?]] =
```
ceedubs: For this and several other instances you've somewhat arbitrarily picked the first first type class when an instanced provides several type classes. This may not be a big deal and I don't think I really have a better suggestion, but it does feel a little off.

kailuowang: I see, I used a regex replace, so it inherited the original name which didn't have the `Traverse` in it. I meant to have all typeclass in it though as I did with several manual replacement. updated

ceedubs: There's probably no single good solution here, but I think that including _all_ of the type class names in the `implicit` name has its issues too. Changing the name of these `implicits` breaks binary compatibility. Ideally we could make an `implicit` extend another type class without breaking binary compatibility but then we wouldn't have the new type class in the name.

ceedubs: @larsrh do you have any thoughts on this? You did a really good job of maintaining binary compatibility in scalaz, and I'd imagine this sort of thing came up.

kailuowang: Yeah, I  thought about using a name like `catsInstancesForXXX` but it might conflict with cats.kernel

InTheNow: As per https://github.com/typelevel/cats/issues/1061#issuecomment-221559248, perhaps the full package name should be incuded? So here, `cats` => `catsStd`. An then `catsKernel`, `catsKernelStd` and so on

larsrh: @ceedubs I don't recall any situation where this sort of thing came up. Having said that, I now wonder whether these instances are necessarily orphan, or whether we can come up with a non-orphan encoding which doesn't duplicate code all over the place. (If they were not orphan this problem wouldn't have existed in the first place.)

kailuowang:
I am fixing the build. Also if no objection I am going to implement @InTheNow 's suggestion ( full package name). for instances with multiple typeclasses I am just going to name them with `catsStdInstancesForXxxxx`

non:
This is great, thanks!

Can you also update the names in `project/Boilerplate.scala` and `project/KernelBoiler.scala` too?

kailuowang:
@non I updated the kernelBoilerplate and `cats.kernel.std` but didn’t find implicit defs in the templates in`project/Boilerplate` ? Do you mind point me to the specific location?

non:
@kailuowang I was wrong. It looks like you got them all! Thanks! 👍

ceedubs:
👍

Issue #1065, opened 2016-05-27
Rename NaturalTransformation to FunctionK?


ceedubs:
Recently @paulp [suggested](https://gitter.im/typelevel/cats?at=573ddd230cb634927f80758c) that `NaturalTransformation` should be renamed to `FunctionK`. This does seem to fit nicely with the convention used by other types such as `SemigroupK` and `MonoidK`. What do people think?

Note: I (and I imagine others) would still want to keep the `~>` alias around.

kailuowang:
👍

ceedubs:
Resolved by #1072.

PR #1072, opened 2016-05-30
Rename NaturalTransformation to FunctionK.

peterneyens:
This would rename `NaturalTransformation` to `FunctionK` #1065, but we should probably wait until some more people have expressed their opinion about this change ?

I'm also not sure how to change the documentation, should we go from "a natural transformation" to "a `FunctionK`".


docs/src/main/tut/freemonad.md
```diff
@@ -166,14 +166,14 @@ DSL. By itself, this DSL only represents a sequence of operations
 To do this, we will use a *natural transformation* between type
 containers.  Natural transformations go between types like `F[_]` and
 `G[_]` (this particular transformation would be written as
-`NaturalTransformation[F,G]` or as done here using the symbolic
+`FunctionK[F,G]` or as done here using the symbolic
 alternative as `F ~> G`).
```
peterneyens: This could then be something like : 

> To do this, we will use  a `FunctionK` (also known as a natural transformation). A `FunctionK` is like a function but between type containers instead of values. A `FunctionK` between types `F[_]` and `G[_]` could be written as `FunctionK[F, G]` (or as `F ~> G` using the symbolic alternative).

core/src/main/scala/cats/arrow/FunctionK.scala
```diff
@@ -0,0 +1,31 @@
...
+  def compose[E[_]](f: FunctionK[E, F]): FunctionK[E, G] =
+    new FunctionK[E, G] {
+      def apply[A](fa: E[A]): G[A] = self.apply(f(fa))
+    }
+
+  def andThen[H[_]](f: FunctionK[G, H]): FunctionK[F, H] =
+    f.compose(self)
+
+  def or[H[_]](h: FunctionK[H,G]): FunctionK[Coproduct[F, H, ?],G] =
+    new (FunctionK[Coproduct[F, H, ?],G]) {
```
non: We don't need parenthesis on this constructor right?

docs/src/main/tut/freemonad.md
non: I would say that we can describe the type `FunctionK` as a *natural transformation*. So, when we would put the term in backticks (i.e. code) we'd say `FunctionK` but otherwise we'd say transformation or natural transformation. Does this make sense?

non:
This is great, thanks!

I think for now we should remove `type NaturalTransformation`. We can add it back in if a bunch of people complain but having too many names for the same thing will probably confuse the issue.

Relatedly, I think we should remove the `<~` alias. I don't think there's ever a situation where code can't just use `~>`.

docs/src/main/tut/freemonad.md
peterneyens: That does make sense, yes. It felt weird to use `FunctionK` in the documentation, that's why I left the documentation itself untouched.

core/src/main/scala/cats/arrow/FunctionK.scala
peterneyens: The parentheses seem to be a remnant of the use of `~>`  (`new (Coproduct[F, G, ?] ~> H) {`). I will remove them.

ceedubs:
@peterneyens thanks, this looks great! Sorry, but it has a merge conflict. Would you mind resolving it?

non:
Perfect. 👍 Thanks!

ceedubs:
👍

PR #1074, opened 2016-05-30
Add a FreeAp structure that acts as a Free monad & materializes the ap behavior #983

### LINKED ISSUE #983: Hybrid free monad / free applicative
Hello,

I am new to the cats project & open source in general - cats is a great project & I learnt a lot from it :)

I've been exploring free monads recently & my understanding is that we can't express parallel operations with them. Free applicatives on the other hand, allow us to express parallelism but we can't express sequential operations.

It seems that it could be interesting to have a data structure that could support both use cases - sequential & parallel operations. For instance, we may want to implement a key value store DSL and be able to write a program like "do these 2 put operations in parallel and if they succeeded, do these 2 get operations in parallel".

I've made a naive implementation of such a structure to show what I mean (naive implementation because I lose a few interesting properties like the tail recursive interpreters for instance) & wrote a simple client code example.

Naive implementation: https://gist.github.com/bqm/bee72c18fa4baa708f7cf2b146bbc872
Simple client code: https://gist.github.com/bqm/e8e28c06311be8d76071729dcaf5b836

What do you think of having an hybrid structure like that in cats?

bqm:
Proof-of-concept #983

### DIFF
--- a/free/src/main/scala/cats/free/FreeAp.scala
+++ b/free/src/main/scala/cats/free/FreeAp.scala
@@ -1,0 +1,95 @@
+package cats
+package free
+
+import cats.arrow.NaturalTransformation
+
+object FreeAp {
+
+  /**
+   * Our case classes are the same as those provided by Free with the addition of Ap
+   * which materializes the applicative ap behavior
+   * This is useful when folding the FreeAp instance to a different Monad:
+   * if that targeted monad has an `ap` function that differs from the usual flatMap / map sequence
+   * we will retain that behavior
+   */
+  private final case class Pure[S[_], A](a: A) extends FreeAp[S, A]
+
+  private final case class Suspend[S[_], A](a: S[A]) extends FreeAp[S, A]
+
+  private final case class Gosub[S[_], B, C](c: FreeAp[S, C], f: C => FreeAp[S, B]) extends FreeAp[S, B]
+
+  private final case class Ap[S[_], B, C](a: FreeAp[S, B], b: FreeAp[S, B => C]) extends FreeAp[S, C]
+
+  def liftF[F[_], A](value: F[A]): FreeAp[F, A] = Suspend(value)
+
+  def suspend[F[_], A](value: => FreeAp[F, A])(implicit F: Applicative[F]): FreeAp[F, A] =
+    liftF(F.pure(())).flatMap(_ => value)
+
+  def pure[S[_], A](a: A): FreeAp[S, A] = Pure(a)
+
+  /**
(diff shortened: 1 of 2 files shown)


ceedubs:
Thanks for putting together this proof of concept, @bqm.

I'm not sure that `FreeAp` is the right name for this structure. It's more than a free `ap` - it's a free monad. It just has a reified `ap`, meaning that `ap`'s behavior can potentially differ from that of `flatMap`. I think if we _do_ add something like this to cats it should just be adding this `Ap` constructor to the existing `Free` (monad). When `foldMap`ing into the free monad into an `M[_]`, the behavior wouldn't change at all for an `M` where `flatMap` and `ap` are consistent. The only difference would be that if there's a more efficient `ap` implementation on `M`, it would be used instead of `flatMap`. @travisbrown I think that at the Typelevel summit in Philly you mentioned that you would like to see something like this, didn't you?

free/src/main/scala/cats/free/FreeAp.scala
```diff
@@ -0,0 +1,95 @@
...
+   * which materializes the applicative ap behavior
+   * This is useful when folding the FreeAp instance to a different Monad:
+   * if that targeted monad has an `ap` function that differs from the usual flatMap / map sequence
+   * we will retain that behavior
+   */
+  private final case class Pure[S[_], A](a: A) extends FreeAp[S, A]
+
+  private final case class Suspend[S[_], A](a: S[A]) extends FreeAp[S, A]
+
+  private final case class Gosub[S[_], B, C](c: FreeAp[S, C], f: C => FreeAp[S, B]) extends FreeAp[S, B]
```
ceedubs: Sorry for the nitpick, but could we switch the two arguments around to line up with the `Apply.ap` type signature?

bqm:
> I'm not sure that FreeAp is the right name for this structure.

Agreed - I wasn't sure how to call it & decided that "Free monad + Ap behavior = FreeAp" but that's a bit confusing.

>  I think if we do add something like this to cats it should just be adding this Ap constructor to the existing Free (monad).

I went for a new structure as having an `Ap`  constructor means that we can't have a tail recursive `resume` / `step` functions like we have in `Free` & thought that we might want to keep it separate as a first step so we don't impact existing customers.

> When foldMaping into the free monad into an M[_], the behavior wouldn't change at all for an M where flatMap and ap are consistent.

I believe that's currently the case in my PR, at least that was the intent :)

easel:
This PR is pretty stale, @bqm iIt would be good to get this up-to-date or closed out if it's no longer of value.

kailuowang:

Issue #1069, opened 2016-05-28
Mixed use of symbols in Kind projections


ghost:
Independent to https://github.com/typelevel/cats/issues/993, there is currently a mixed use of `Lambda` and `λ` in kind projections. Eg

`def compose[G[_, _]](implicit ev: Bifoldable[G]): Bifoldable[Lambda[(A, B) => F[G[A, B], G[A, B]]]]`

`def compose[G[_]](implicit GG : Applicative[G]): Applicative[λ[α => F[G[α]]]] =`

Whatever the chosen convention, it should be applied consistently.

ceedubs:
I think this was fixed by #1079 (and even with a scalastyle rule to make sure we are consistent in the future!), so I'm going to close this out. Please feel free to reopen if I'm missing something.

PR #1079, opened 2016-06-01
Make the usage of kind projections more consistent.

peterneyens:
Like @InTheNow said in #1069: it's probably better to be consistent in our kind projection syntax.

Since the `Lambda` syntax was used quite a lot more than the `λ` syntax, I changed the occurrences of the last to the first.

I have also replaced the last type lambdas in `TransLiftTests` by kind projections.

### DIFF
--- a/core/src/main/scala/cats/Alternative.scala
+++ b/core/src/main/scala/cats/Alternative.scala
@@ -3,7 +3,7 @@ package cats
 import simulacrum.typeclass
 
 @typeclass trait Alternative[F[_]] extends Applicative[F] with MonoidK[F] { self =>
-  override def compose[G[_]: Applicative]: Alternative[Lambda[A => F[G[A]]]] =
+  override def compose[G[_]: Applicative]: Alternative[λ[α => F[G[α]]]] =
     new ComposedAlternative[F, G] {
       val F = self
       val G = Applicative[G]
--- a/core/src/main/scala/cats/Applicative.scala
+++ b/core/src/main/scala/cats/Applicative.scala
@@ -44,7 +44,7 @@ import simulacrum.typeclass
   def sequence[G[_], A](as: G[F[A]])(implicit G: Traverse[G]): F[G[A]] =
     G.sequence(as)(this)
 
-  def compose[G[_]: Applicative]: Applicative[Lambda[A => F[G[A]]]] =
+  def compose[G[_]: Applicative]: Applicative[λ[α => F[G[α]]]] =
     new ComposedApplicative[F, G] {
       val F = self
       val G = Applicative[G]
--- a/core/src/main/scala/cats/Apply.scala
+++ b/core/src/main/scala/cats/Apply.scala
@@ -58,7 +58,7 @@ trait Apply[F[_]] extends Functor[F] with Cartesian[F] with ApplyArityFunctions[
   def map2Eval[A, B, Z](fa: F[A], fb: Eval[F[B]])(f: (A, B) => Z): Eval[F[Z]] =
     fb.map(fb => map2(fa, fb)(f))
 
-  def compose[G[_]: Apply]: Apply[Lambda[A => F[G[A]]]] =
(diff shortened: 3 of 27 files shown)


non:
Hi @peterneyens, this seems OK to me. But would you mind distinguishing the type lambda parameters from normal types somehow? It's fine if we move away from greek letters but maybe we could use lowercase letters to help distinguish them?

peterneyens:
The type lambda parameters were indeed more easily visible with the Greek letters than with the upper-case letters. 

I don't mind lower casing or using Greek characters for the parameters, but I hope you don't me waiting on other opinions (doesn't make much sense doing this work twice :smile:).

InTheNow:
from gitter:

- https://gitter.im/typelevel/cats?at=54d39e88ea831b7605fb20e7
- https://gitter.im/typelevel/cats?at=54d3a07bea831b7605fb213d
- https://gitter.im/typelevel/cats?at=54d3a1a044f4db8b5063b64c
- https://gitter.im/typelevel/cats?at=54ebc5f9ea450f2f20888ba1
- https://gitter.im/typelevel/cats?at=54d3a9ecea831b7605fb2390
- https://gitter.im/typelevel/cats?at=54d3aa6eadca701740eb497f
- https://gitter.im/typelevel/cats?at=54d3a69826bc541370bef668

Not sure if that helps or not :)

InTheNow:
A few comments ^^ was about the general usage of Greeks in general (My IDE can't, etc). Hence, if you use greeks for the parameters, you might as well use them for the lambda, too.

One of my comments was they are confusing to newbies. Whilst I still hold to that, alot of that thought was based on code _not_ using kind projector _and_ using ascii symbols such as `\/`. There are no greeks below (using lower case), but OMG!

```scala
implicit def DisjunctionInstances1[L]: Traverse[({type l[a] = L \/ a})#l] 
  with Monad[({type l[a] = L \/ a})#l] 
  with Cozip[({type l[a] = L \/ a})#l] 
  with Plus[({type l[a] = L \/ a})#l] 
  with Optional[({type l[a] = L \/ a})#l] 
  with MonadError[\/, L] = new Traverse[({type l[a] = L \/ a})#l] 
  with Monad[({type l[a] = L \/ a})#l] 
  with Cozip[({type l[a] = L \/ a})#l] 
  with Plus[({type l[a] = L \/ a})#l] 
  with Optional[({type l[a] = L \/ a})#l] with MonadError[\/, L] {
```

But as cats does not make heavy use of ascii symbols, and does use Kind projector, I must say that I prefer

`val asString: FunctionK[Id,λ[α => String]] = new FunctionK[Id,λ[α => String]] {`

to

`val asString: FunctionK[Id, Lambda[X => String]] = new FunctionK[Id, Lambda[X => String]] {`

Luckily, it's not my decision - tough one :smile:

ceedubs:
I'm pretty ambivalent as to whether greek letters are used or not, but I echo @non's request that type lambda parameters are clearly distinguishable from "normal" types.

peterneyens:
After the consensus on using Greek characters on gitter yesterday, I have changed this PR to use the `λ[α => G[F[α]]` syntax.

The most important point for this decision are (if I summarized the discussion correctly) : 

- The type lambda parameters need to be clearly distinguishable from normal type parameters.
- The kind projector plugin makes it (to ?) easy to write type lambdas, so using Greek characters slightly raises the bar again.
- The Greek characters show that something different is going on, making it easier to refer to the kind projector plugin from something like a FAQ.

non:
👍 Thanks for this @peterneyens, and also putting up with our flip-flopping!

non:
Unfortunately it looks like I just created a merge conflict by merging #1021. Sorry! I'll try to fix it ASAP.

peterneyens:
I saw that one coming indeed :smile: , I hope I solved the conflicts.

I was also thinking that I could at a scalastyle regex check for `Lambda[`, but scalastyle doesn't seem to be working in cats here, I don't know if I am the only one having this problem ?

ceedubs:
Great. Thanks, @peterneyens!

PR #1082, opened 2016-06-01
Fix #1070: Reenable scaladoc for 2.10

### LINKED ISSUE #510: Laws are missing in the published API docs

### LINKED ISSUE #1070: Reenable scaladoc for 2.10
Revert #1051 once we are able to, as it sounds like SBT 0.13.12 will have a fix for the issue that caused disabling scaladoc for 2.10.

InTheNow:
This also:

- eliminates #516 
- fixes #510
- eliminates #961

Once sbt-unidoc adopts the change, we can then remove `project/Unidoc.scala` - see the big comment there for general details

/CC @mikejcurry @eed3si9n @kailuowang

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -30,8 +30,7 @@ lazy val kernelSettings = Seq(
   resolvers ++= Seq(
     Resolver.sonatypeRepo("releases"),
     Resolver.sonatypeRepo("snapshots")),
-  parallelExecution in Test := false,
-  scalacOptions in (Compile, doc) := (scalacOptions in (Compile, doc)).value.filter(_ != "-Xfatal-warnings")
+  parallelExecution in Test := false
 ) ++ warnUnusedImport
 
 lazy val commonSettings = Seq(
@@ -47,8 +46,7 @@ lazy val commonSettings = Seq(
     compilerPlugin("org.scalamacros" %% "paradise" % "2.1.0" cross CrossVersion.full),
     compilerPlugin("org.spire-math" %% "kind-projector" % "0.6.3")
   ),
-  parallelExecution in Test := false,
-  scalacOptions in (Compile, doc) := (scalacOptions in (Compile, doc)).value.filter(_ != "-Xfatal-warnings")
+  parallelExecution in Test := false
 ) ++ warnUnusedImport
 
 lazy val tagName = Def.setting{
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -95,37 +93,20 @@ lazy val testingDependencies = Seq(
...
   site.addMappingsToSiteDir(tut, "_tut"),
   ghpagesNoJekyll := false,
   siteMappings += file("CONTRIBUTING.md") -> "contributing.md",
   scalacOptions in (ScalaUnidoc, unidoc) ++= Seq(
-    "-Xfatal-warnings",
     "-doc-source-url", scmInfo.value.get.browseUrl + "/tree/master€{FILE_PATH}.scala",
     "-sourcepath", baseDirectory.in(LocalRootProject).value.getAbsolutePath,
     "-diagrams"
   ),
+  scalacOptions in (ScalaUnidoc, unidoc) ~= {_.filterNot("-Xfatal-warnings" == _)},
```
dwijnand: Can also

```scala
scalacOptions in (ScalaUnidoc, unidoc) -= "-Xfatal-warnings"
```

non:
Awesome! 👍 once Travis is green. Thanks @InTheNow

eed3si9n:
I'd be happy to review pull requests for sbt-unidoc or sbt, if the solution here fixes things generally.

InTheNow:
@eed3si9n I have a PR waiting to go :smile: Just waiting for the build here

mikejcurry:
Pretty awesome stuff. Glad to see any hacky proposals I'd put together be put to shame :-)

👍

InTheNow:
@mikejcurry All the previous work pointed the way to the next work :smile:

non:
Travis claims that the 2.10 build failed, but I don't see an obvious reason why.

InTheNow:
@non From my chat on gitter: 

>OK - I'm getting a build error in cats in pulblishLocal as the fix I've done for docs (ie unidoc) is not working for the individual projects compile:doc. Same fix, though.
But obviously there is no point adding to sbt-unidoc only
I have to go out, but will fix that later/tomorrow in the same way, just to complete the POC

All the _real_ testing I did was with the "fix" in a locally modified sbt, I then pulled the bits out to provide a short term workaround - this just slipped through the net.

InTheNow:
Right: As it looks like the fix is OK in general, ultimately needs to go into SBT and sbt 0.13.2 is not far way, I will focus on the the real fix that is in SBT.

So lets leave this PR open, and I'll push an new PR as and when the fix is in sbt - even as a milestone release so we can use this PR as a test for that

PR #1083, opened 2016-06-01
Add Traverse[Try]

### LINKED ISSUE #1081: Add Traverse[Try] instance
Follow up to https://github.com/typelevel/cats/pull/1059#issuecomment-222554220

ceedubs:
Resolves #1081

### DIFF
--- a/core/src/main/scala/cats/std/try.scala
+++ b/core/src/main/scala/cats/std/try.scala
@@ -3,14 +3,15 @@ package std
 
 import cats.syntax.all._
 import cats.data.Xor
+import TryInstances.castFailure
 
 import scala.util.control.NonFatal
 import scala.util.{Failure, Success, Try}
 
 trait TryInstances extends TryInstances1 {
 
-  implicit def catsStdInstancesForTry: MonadError[Try, Throwable] with CoflatMap[Try] =
-    new TryCoflatMap with MonadError[Try, Throwable]{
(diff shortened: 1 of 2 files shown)


non:
👍

ceedubs:
@johnynek would you be interested in reviewing this one, since you recently submitted #1059?

core/src/main/scala/cats/std/try.scala
```diff
@@ -78,6 +97,14 @@ trait TryInstances extends TryInstances1 {
     }
 }
 
+private[std] object TryInstances {
+  /**
+   * A `Failure` can be statically typed as `Try[A]` for all `A`, because it
+   * does not actually contain an `A` value (as `Success[A]` does).
+   */
+  def castFailure[A](f: Failure[_]): Try[A] = f.asInstanceOf[Try[A]]
```
johnynek: worth it to put an `@inline` here? I don't actually know how effective that is, but it seems like a candidate.

johnynek:
👍 modulo the minor comment on `@inline`

core/src/main/scala/cats/std/try.scala
ceedubs: Good suggestion. That seems reasonable to me.

ceedubs:
I've updated with `@inline`

johnynek:
:+1:

adelbertc:
:+1:

PR #1084, opened 2016-06-02
Provide concise .toString for free structures.

non:
Previously calling .toString on a Free value would traverse
the entire ADT of case classes. In some cases this would
actually cause a stack overflow in the printing code. It
also represented a (small) law violation (you could use the
output to distinguish x and x.map(identity) in some cases).

This change replaces the output with opaque "Free(...)"
and "FreeApplicative(...)" strings.


mandubian:
👍

fthomas:
:+1:

ceedubs:
@non would you be willing to add a regression test that a deeply-nested `Free` doesn't result in a stack overflow when `toString` is called?

non:
@ceedubs added.

ceedubs:
Thanks @non! 👍 as soon as the build passes.

PR #1085, opened 2016-06-02
Some clean up for Free (and friends).

non:
This commit does a bunch of small things:

 * Introduces `foldLeftM` and `runTailRec`
 * Renames `mapSuspension` to `compile`
 * Removes `Applicative[F]` requirement from `Free.suspend`
 * Adds some new comments
 * Moves the Free companion to the end of the file
 * Minor formatting changes

I think all of these are good, but I'm happy to revert
any of them that seem like they are bad news. The main
reason I removed `mapSuspension` is that its description
was written using "compile" which seems like a more
evocative name (there are already several "map" variants).

### DIFF
--- a/free/src/main/scala/cats/free/Free.scala
+++ b/free/src/main/scala/cats/free/Free.scala
@@ -207,6 +207,10 @@ object Free {
   /**
    * Perform a stack-safe monadic fold from the source context `F`
    * into the target monad `G`.
+   *
+   * This method can express short-circuiting semantics, but like
+   * other left folds will traverse the entire `F[A]` structure. This
+   * means it is not suitable for potentially infinite structures.
    */
   def foldLeftM[F[_], G[_]: MonadRec, A, B](fa: F[A], z: B)(f: (B, A) => G[B])(implicit F: Foldable[F]): G[B] =
     F.foldM[Free[G, ?], A, B](fa, z) { (b, a) => Free.liftF(f(b, a)) }.runTailRec
--- a/free/src/main/scala/cats/free/Free.scala
+++ b/free/src/main/scala/cats/free/Free.scala
@@ -139,6 +139,9 @@ sealed abstract class Free[S[_], A] extends Product with Serializable {
         def apply[B](fa: S[B]): Free[T, B] = Suspend(f(fa))
       }
     }(Free.freeMonad)
+
+  override def toString(): String =
+    "Free(...)"
 }
 
 object Free {
@@ -214,6 +217,4 @@ object Free {
    */
   def foldLeftM[F[_], G[_]: MonadRec, A, B](fa: F[A], z: B)(f: (B, A) => G[B])(implicit F: Foldable[F]): G[B] =
     F.foldM[Free[G, ?], A, B](fa, z) { (b, a) => Free.liftF(f(b, a)) }.runTailRec
-
-  override def toString(): String = "Free(...)"
 }
--- a/free/src/main/scala/cats/free/Free.scala
+++ b/free/src/main/scala/cats/free/Free.scala
@@ -6,61 +6,14 @@ import scala.annotation.tailrec
 import cats.data.Xor, Xor.{Left, Right}
 import cats.arrow.FunctionK
 
-object Free {
-  /**
-   * Return from the computation with the given value.
-   */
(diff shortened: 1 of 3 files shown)


mandubian:
👍 really good move!

ceedubs:
Thanks @non! I've submitted https://github.com/non/cats/pull/1 because I think that we can get short-circuiting semantics out of `foldLeftM`.

non:
@ceedubs Thanks! Great job.

ceedubs:
I'm thinking that we'll probably want to move `Free` (the monad - not all of the `free` module) into core before 1.0 so that we can have `foldLeftM` on `Foldable` and take similar approaches for some other types. But I think that we should move forward with this in the mean time 👍.

ceedubs:
@mandubian would you mind taking another look now that my change has been included in this PR?

non:
OK, this should be good to merge once it gets sign-offs.

mandubian:
@ceedubs code is good to me... Just need to think a bit more about this foldLeftM that is based on a foldRight and that becomes safe due to Free + MonadRec...

adelbertc:
👍 from me

free/src/main/scala/cats/free/Free.scala
```diff
@@ -130,31 +88,141 @@ sealed abstract class Free[S[_], A] extends Product with Serializable {
...
+   *
+   * This method can express short-circuiting semantics. Even when
+   * `fa` is an infinite structure, this method can potentially
+   * terminate if the `foldRight` implementation for `F` and the
+   * `tailRecM` implementation for `G` are sufficiently lazy.
+   */
+  def foldLeftM[F[_]: Foldable, G[_]: MonadRec, A, B](fa: F[A], z: B)(f: (B, A) => G[B]): G[B] =
+    unsafeFoldLeftM[F, Free[G, ?], A, B](fa, z) { (b, a) =>
+      Free.liftF(f(b, a))
+    }.runTailRec
```
TomasMikula: Did I miss it, or does this method not use the `G.tailRecM` method at all?

Also, it seems like there is double trampolining (once via `Free` and once via `Eval` in `F.foldRight`)? I wonder if a more direct implementation (using either `Free` or `Eval`, but not both) is possible.

ceedubs: @TomasMikula as the author of this code, I'll be the first to say that it is a bit bizarre. `G.tailRecM` should be being called by the `runTailRec` call on the `Free` structure. `Eval` makes an appearance due to the `foldRight` type signature, but really we are just using `foldRight` because of its laziness, and `unsafeFoldLeftM` is calling `.value` on the `Eval`, so we don't actually get any `Eval` trampolining here. It's effectively like we are using a scalaz-style `foldRight` that takes a byname parameter, since we aren't taking advantage `Eval`'s trampolining.

If you can figure out a way to provide a more direct implementation, I'd definitely be interested. I think that this has to be done in terms of `foldRight` if we want laziness, though, which means `Eval` will definitely be in play. You may be able to use `MonadRec` and `foldRight` without `Free`, and if so that'd be great - as it stands I'm kind of convinced that we should move `Free` (the monad but not the other free structures) back into core for this sort of thing.

ceedubs: One other note: it really isn't obvious to me that this implementation should be stack safe, but the tests kind of prove that it is ¯\_(ツ)_/¯

ceedubs:
I'm going to go ahead and merge, because this has some really good changes in it. @TomasMikula please feel free to experiment with `foldLeftM` - it is similar to a few other methods that I think we will want to support under a common approach to stack-safety.

free/src/main/scala/cats/free/Free.scala
TomasMikula: somehow I completely missed that `.runTailRec` at the end 😊

TomasMikula:
I might try to experiment to avoid the use of `Free` in `foldLeftM`, but I'm afraid that if I succeed I will just reinvent half of it in the process.

TomasMikula:
@ceedubs so, I was not even trying to move the computation to the type level, but I just managed to cause a stack overflow in the compiler 😄

TomasMikula:
See #1117.

PR #1090, opened 2016-06-04
Adding a way to create a Kleisli with just F[B]

tbrown1979:
I'm new to Kleislis and found myself wanting a function like this. Scalaz has a `.liftKleisli` function, but I wasn't sure what would be preferred: `Kleisli.pureF(fa)` or `fa.liftKleisli`.

### DIFF
--- a/core/src/main/scala/cats/data/Kleisli.scala
+++ b/core/src/main/scala/cats/data/Kleisli.scala
@@ -67,7 +67,7 @@ object Kleisli extends KleisliInstances with KleisliFunctions
 
 private[data] sealed trait KleisliFunctions {
 
-  def pureF[F[_], A, B](x: F[B])(implicit F: Applicative[F]): Kleisli[F, A, B] =
+  def pureF[F[_], A, B](x: F[B]): Kleisli[F, A, B] =
     Kleisli(_ => x)
 
   def pure[F[_], A, B](x: B)(implicit F: Applicative[F]): Kleisli[F, A, B] =
--- a/core/src/main/scala/cats/data/Kleisli.scala
+++ b/core/src/main/scala/cats/data/Kleisli.scala
@@ -67,7 +67,7 @@ object Kleisli extends KleisliInstances with KleisliFunctions
 
 private[data] sealed trait KleisliFunctions {
 
-  def const[F[_], A, B](x: F[B]): Kleisli[F, A, B] =
+  def lift[F[_], A, B](x: F[B]): Kleisli[F, A, B] =
     Kleisli(_ => x)
 
   def pure[F[_], A, B](x: B)(implicit F: Applicative[F]): Kleisli[F, A, B] =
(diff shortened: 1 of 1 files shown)


edmundnoble:
Personally I would be in favour of naming this `Kleisli.const`, as I think that illustrates its meaning more effectively. `pureF` might not work so well, because `pure` implies a lack of effects (as in `Applicative.pure`) but `pureF`'s argument would be an `F[B]`, which may have effects. `liftKleisli` as a postfix operator sounds alright if it's that common a use-case, but you'd need to also add `KleisliSyntax` for a place to put it.

tbrown1979:
@edmundnoble Thanks for the response. I was looking for a more appropriate name and `Kleisli.const` seems much better. I'll make that change. I'll also look into adding the `.liftKleisli` syntax as some people might prefer that.

non:
I second the `Kleisli.const` suggestion.

ceedubs:
I think I'd be inclined to call this `.liftT`, and then the [liftT](https://github.com/typelevel/cats/blob/fb87999b5ec8d1eb428df991752a172b08f0becc/core/src/main/scala/cats/data/Kleisli.scala#L124) implementation on the `TransLift` instance could simply delegate through to it.

tbrown1979:
Okay, so I've changed the name to `.const` but can certainly change it to `.liftT` if that is preferred. Either way, I have changed `TransLift` instance to use the `Kleisli.const` now.

tbrown1979:
I think I will forego the `.liftKleisli` for this pull request and add that in another, if we're okay with that.

non:
👍 Thanks!

ceedubs:
I think that everywhere else that we have an analogous method on other monad transformers it's called `liftF` or `liftT`, so I'd prefer to be consistent (or even just use `lift`). It may be an unfounded concern, but I don't want people to think that this is related to `Const`. I'll let another person weigh in -- I'm happy to have this merged if people align on `const`.

peterneyens:
Looking at similar methods on the other monad transformers there is no real consensus at the moment, probably because the methods on `XorT` and `WriterT` have to deal with their specific context.

| Tranformer    | `F[A] => MT[F, ... , A]`     |
| ------------- |--------------------------|
| [`OptionT`](http://typelevel.org/cats/api/index.html#cats.data.OptionT$)     | `OptionT.liftF`          |
| [`XorT`](http://typelevel.org/cats/api/index.html#cats.data.XorT$)        | `XorT.right` / `Xort.left` |
| [`WriterT`](http://typelevel.org/cats/api/index.html#cats.data.WriterT$)    | `WriterT.putT` / `WriterT.valueT` |
| [`StateT`](http://typelevel.org/cats/api/index.html#cats.data.StateT$)      | *none*                     |
| [`Kleisli`](http://typelevel.org/cats/api/index.html#cats.data.Kleisli$)     | *none*                     |

Even though `Free` is not a monad transformer it also uses `Free.liftF` to go from `F[A]` to `Free[F, A]`.

I like `Kleisli.const` in this case, but concur with @ceedubs that it is important to be consistent. Since we have the type class `TransLift` with `liftT`, I think that it's a good idea to use `MT.lift` or `MT.liftF` if there is no better specific name (like for `XorT` and `WriterT`).

```scala
OptionT.liftF(List(1,2))
List(1,2).liftT[Option]

Kleisli.lift[List, Int, Int](List(1,2)) // ?
List(1,2).liftT[({type λ[α[_], β] = Kleisli[α, Int, β]})#λ] // -> liftKleisli[Int] ?
```

ceedubs:
@peterneyens thanks for the helpful writeup. It looks like I definitely didn't have the facts on my side when I claimed that this was `liftT` or `liftF` everywhere else.

tbrown1979:
I've renamed from `.const` to `.lift` to remain a little more consistent with naming. 😄

ceedubs:
👍 thanks!

ceedubs:
@non your 👍 was for `const` -- are you okay with `lift`?

non:
👍

Issue #1092, opened 2016-06-05
Enforce kind projection style


peterneyens:
Now we have settled on using Greek characters in our kind projections in #1079, we should try to keep it consistent.

I was thinking about a scalastyle rule to check that files don't contain `Lambda[`.

This could look like the following (but I have been unable to test it, since running `scalastyle` gives me `[info] Processed 0 file(s)`) : 
```xml
 <check level="error" class="org.scalastyle.file.RegexChecker" enabled="true">
  <parameters>
   <parameter name="regex">Lambda\[</parameter>
  </parameters>
  <customMessage>Use Greek characters in a kind projection : λ[α => ...].</customMessage>
 </check>
```

fthomas:
@peterneyens I think we need to update our scalastyle configuration, see: https://github.com/scalastyle/scalastyle-sbt-plugin/issues/47

PR #1093, opened 2016-06-05
Fix Scalastyle

peterneyens:
Like @fthomas mentioned in #1092, we needed to update our scalastyle configuration because of https://github.com/scalastyle/scalastyle-sbt-plugin/issues/47.

I used the same workaround as circe (https://github.com/travisbrown/circe/pull/156) and refined (https://github.com/fthomas/refined/pull/122).

I am not sure if `commonSettings` is the right place in *build.sbt* to add it ?


kernel/src/main/scala/cats/kernel/std/Util.scala
```diff
@@ -85,4 +86,5 @@ object StaticMethods {
     }
     true
   }
+  // scalastyle:on magic.number
```
ceedubs: I might be reading or understanding this incorrectly, but it looks like the `scalastyle:off` has `return`, but `scalastyle:on` has `magic.number`. Is this a mistake?

ceedubs:
Thanks! I left one minor question, but 👍 - scalastyle helps keep us consistent.

kernel/src/main/scala/cats/kernel/std/Util.scala
peterneyens: That's a mistake indeed, thanks for spotting it !

kernel-laws/src/main/scala/cats/kernel/laws/IsSerializable.scala
```diff
@@ -19,7 +19,9 @@ private[laws] object IsSerializable {
       import java.io._
       val baos = new ByteArrayOutputStream()
       val oos = new ObjectOutputStream(baos)
+      // scalastyle:off null
       var ois: ObjectInputStream = null
+      // scalastyle:off on
```
fthomas: This should be `scalastyle:on null`, right?

fthomas: Btw, it seems that you can also disable scalastyle on single lines with
```scala
var ois: ObjectInputStream = null // scalastyle:ignore null
```
see http://www.scalastyle.org/configuration.html.

non:
It seems like a few of these rules (e.g. number.of.types) aren't really doing much for us. Does anyone think we should leave these rules on? I would be just as happy to turn them off as to disable them.

kernel-laws/src/main/scala/cats/kernel/laws/IsSerializable.scala
peterneyens: Thanks for finding another of my errors. I didn't know you could also specify the checker in the single line comment.

peterneyens:
@non You are right, I figured it was best to get scalastyle working first and then worry about which rules we might have to change with some input from you guys.

The NumberOfTypes rule seems indeed a prime candidate to be removed from our scalastyle config. I would be happy to remove this and other rules in this PR if that's what we want.

fthomas:
:+1: (when the build is green). Thanks @peterneyens!

I'd prefer merging this sooner than later and changing the scalastyle configuration in subsequent PRs. Is that okay for you, @non?

ceedubs:
I agree that number of types is a prime candidate for rule removal in a separate PR, but I'm going to go ahead and merge this to get scalastyle rolling again. thanks again @peterneyens!

PR #936, opened 2016-03-16
Imports section

noelmarkham:
Placement suggested in https://github.com/typelevel/cats/issues/912

Issue #912, opened 2016-03-03
document imports


stew:
We need a document which states exactly what needs to be import when, so that newcomers know exactly what imports do what, and what the basic set is they can use to get going.

longcao:
For posterity: @non mentioned in the round table discussion at the Typelevel Summit unconference that it should be the standard to publish only:

```scala
import cats.implicits._
```

as the fully-capable gateway to using `cats`, and that more specific imports can be used but with the caveat of being a) not exhaustively documented and b) for more advanced usage. IMO, part of this would include rewriting current `tut` documentation to use only `import cats.implicits._`, and publishing that this is the recommended pathway.

(hope I remembered correctly!)

lvicentesanchez:
+1 for using `import cats.implicits._` in the documentation, but I would love to have an advanced usage section with an extended version of what @noelmarkham has started [here](https://gist.github.com/noelmarkham/5815e3b1be81cf74c536).

longcao:
@lvicentesanchez :+1: that's pretty nice! I think that cheat sheet might be good to have in the documentation, perhaps as a sub-bullet to `core` in "Project Structure" (http://typelevel.org/cats/)

noelmarkham:
I've started a PR at https://github.com/typelevel/cats/pull/936

ceedubs:
There has been an imports guide for a while. One thing that it doesn't include is a guide on piecemeal imports, so I've created #1668 to track that.

PR #1095, opened 2016-06-07
Add FAQ section to site

ceedubs:
This is just a start, and there are some more items that we will
probably want to add.

The import section probably is really a start to #912, and may end up
simply referencing a separate import document, as discussed in #936. But
I thought it would be nice to get something out there as an interim
resolution.

### DIFF
--- a/docs/src/main/tut/faq.md
+++ b/docs/src/main/tut/faq.md
@@ -1,0 +1,64 @@
+---
+layout: default
+title:  "FAQ"
+section: "faq"
+---
+# Frequently Asked Questions
+
+## What imports do I need?<a id="what-imports" href="#what-imports"></a>
+
+The easiest approach to cats imports is to import everything that's commonly needed:
+
+```tut:silent
+import cats._
+import cats.data._
+import cats.implicits._
+```
+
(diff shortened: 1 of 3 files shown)


docs/src/main/tut/faq.md
```diff
@@ -0,0 +1,64 @@
...
+// cats adds a toXor method to the standard library's Either
+val e: Either[String, Int] = Right(3)
+e.toXor
+
+// cats adds an orEmpty method to the standard library's Option
+val o: Option[String] = None
+o.orEmpty
+```
+
+**Note**: if you import `cats.implicits._` (the preferred method), you should _not_ also use imports like `cats.std.syntax._` or `cats.std.either._`. This can result in ambiguous implicit values that cause bewildering compile errors.
```
peterneyens: I suspect you meant something different than `cats.std.syntax._` ?

ceedubs: Oops quite right thanks. I wish I could have doobie verify inline code! I'll fix this but I might not have time today.

ceedubs: Fixed.

non:
This is a great start. Let's merge it and iterate! 👍

ceedubs:
Invoking the Documentation Rule and merging :)

PR #1096, opened 2016-06-07
Fix Inject instance and reduceRightToOption null behavior

edmundnoble:
`Inject.prj(null)` with the provided reflexive Inject instance (`Inject[F, F]`) returns None, but `Inject.inj(null)` returns null. I am fairly sure that inj and prj should be mutual inverses, this fixes that. It works the same way now as in Scalaz.


ceedubs:
👍

non:
👍 Great catch, thanks!

non:
Should we add a law that checks this property, and/or an explicit test of the null behavior with option?

non:
I'm going to just merge this -- we can add a test in another PR.

PR #1097, opened 2016-06-07
Explicit functor variance

edmundnoble:
This uses casts to implement `Functor.widen` and `Contravariant.narrow` to take advantage of the natural variance of Functors with respect to subtyping. `Functor.widen` in Scalaz works similarly (though it uses Liskov and `@uncheckedVariance` in its implementation, it has the same underlying optimization).

### DIFF
--- a/core/src/main/scala/cats/Functor.scala
+++ b/core/src/main/scala/cats/Functor.scala
@@ -19,6 +19,12 @@ import simulacrum.typeclass
   // derived methods
 
   /**
+   * Lifts natural subtyping covariance of covariant Functors.
+   * could be implemented as map(identity), but the Functor laws say this is equivalent
+   */
+  def widen[A, B >: A](fa: F[A]): F[B] = fa.asInstanceOf[F[B]]
+
+  /**
    * Lift a function f to operate on Functors
    */
   def lift[A, B](f: A => B): F[A] => F[B] = map(_)(f)
--- a/core/src/main/scala/cats/functor/Contravariant.scala
+++ b/core/src/main/scala/cats/functor/Contravariant.scala
@@ -16,6 +16,12 @@ import simulacrum.typeclass
       val G = Contravariant[G]
     }
 
+  /**
+   * Lifts natural subtyping contravariance of contravariant Functors.
+   * could be implemented as contramap(identity), but the Functor laws say this is equivalent
+   */
+  def narrow[A, B <: A](fa: F[A]): F[B] = fa.asInstanceOf[F[B]]
+
   override def composeFunctor[G[_]: Functor]: Contravariant[λ[α => F[G[α]]]] =
     new ComposedContravariantCovariant[F, G] {
       val F = self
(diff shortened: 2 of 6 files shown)


edmundnoble:
I think the only useful tests for widen and narrow are going to be fairly low-level. I'll add some for codecov.

peterneyens:
You could probably add an sbt-doctest to solve the codecov difference ?

julienrf:
Can you point to a discussion about the pros and cons of this approach versus using a variance annotation?

core/src/main/scala/cats/Functor.scala
```diff
@@ -18,6 +18,12 @@ import simulacrum.typeclass
 
   // derived methods
 
+  /*
```
ceedubs: Could you please add another `*` so that this becomes a ScalaDoc comment?

core/src/main/scala/cats/functor/Contravariant.scala
```diff
@@ -16,6 +16,12 @@ import simulacrum.typeclass
       val G = Contravariant[G]
     }
 
+  /*
```
ceedubs: Same here.

ceedubs:
@edmundnoble thanks! Looking good so far. I agree with @peterneyens that some doctest examples would be good and will help out the code coverage. Another thing that you could do is add some property-tests that `widen` is equivalent to `map identity`, but then I guess you have the question of whether you should add that for a few random functors, or add it to the laws.

non:
I agree with @ceedubs specific comment syntax remarks, but other than that +1 from me. Tests and/or examples would be nice too if you felt like doing them.

edmundnoble:
@julienrf That's proving surprisingly hard to find. I can find a good argument against using a contravariance annotation on Contravariant: https://issues.scala-lang.org/browse/SI-2509.

edmundnoble:
I've made the comments into scaladoc comments, and added some doctest stuff and tests.

ceedubs:
👍 thanks!



@julienrf I don't know of a discussion about this approach vs adding a variance annotation to point to. I can add some notes here though.

If you were to write an invariant `InvList[A]`, your prepend method would look something like `def ::(a: A): InvList[A] = ?`. With a covariant list like the standard library's `List[+A]`, this becomes a bit more complicated. A `List[A]` _is_ a `List` of any supertype of `A`, so your method has to become something like `def ::[B >: A](x: B): List[B]`. This means that something that is likely an error like `1 :: List("b", "c")` compiles (but returns `List[Any]`). There are reasons that you might want that behavior, but I think that it's also reasonable to not want that behavior.

The previous paragraph referred to putting the variance annotation on the data structure. If you were to put the variance annotation on the type class instances (like `trait Functor[F[+_]]`), then you would prohibit the creation of a `Functor` instance for any invariant structure (such as the `NonEmptyList` in cats). I consider that a deal-breaker for adding variance to type classes.

non:
Looks good to me. 👍

PR #1098, opened 2016-06-08
Add a different MonoidK and SemigroupK instance for Kleisli

peterneyens:
I was just looking at a [StackOverflow question](http://stackoverflow.com/questions/37700755/how-to-compose-functions-returning-option-with-orelse) and was thinking this could be solved using `SemigroupK.combineK`, but while there is a `SemigroupK[Option]` there is nothing to give us a `SemigroupK[λ[α => Int => Option[α]]]`. So I have added a `SemigroupK` and `MonoidK` for a function `A => F[B]` if there is a `MonoidK[F]`.

I am not exactly sure how useful this is in practice. This is probably not something used very often (so I am not sure if this warrants inclusion in cats ?). It also seems you will always need to write a type lambda : 

``` scala
import cats.implicits._
val foo: Int => Option[Int] = x => if (x % 2 == 0) Some(x) else None
val bar: Int => Option[Int] = x => if (x % 3 == 0) Some(x) else None
SemigroupK[λ[α => Int => Option[α]]].combineK(foo, bar)
```

It would of course be nicer if you could write `foo <+> bar`. I'm not sure if that would be possible using the existing `semigroupSyntaxU` or using the SI-2712 fix (I tried with a locally published cats version with this PR and Miles' compiler plugin, but couldn't get it working) ?


peterneyens:
I got the syntax working with a type alias (so it can use the `Unapply` and it probably works with the SI 2712 plugin too).

```scala
type T[f[_], a, b] = Function[a, f[b]]
val foo: T[Option, String, Int] = s => Option(s.toInt)
val bar: T[Option, String, Int] = s => None

foo <+> bar
```
This looks quite a lot like `Kleisli` :stuck_out_tongue: , so I should probably change this PR to add these instances to `Kleisli` instead of `Function1`.

non:
Looks good to me! 👍

ceedubs:
@peterneyens thanks!

It seems like it could potentially be problematic to have two different `SemigroupK` instances for `Kleisli`. Would there be ambiguity in the case of something like `Kleilsi[List, Int, Int]`?

peterneyens:
@ceedubs You are right, with this PR you will always get the newly added `Kleisli[F, A, ?]` instances. The existing `SemigroupK[F, α, α]` will probably be used more and is also consistent with the `SemigroupK` from `Function1`.

I'm not sure how often these new instances which combine `Kleisli`s into one with the combined result will be used. Do I leave the them in and only make them explicitly available (dropping the `implicit` from the functions) ?

edmundnoble:
Any updates on this? I don't see an issue with providing these instances explicitly, provided they're documented.

peterneyens:
Made this instances explicit (notreally  sure about the naming of the instances and the traits) and added doctests for both instances.

Anything else you were thinking of @edmundnoble?

kailuowang:
@peterneyens do you still have time to work on this PR?

peterneyens:
Could come back to this in the weekend, but if anybody else wants to go ahead in the mean time, I don't mind.

kailuowang:
Thanks so much @peterneyens

aeons:
Awesome!

PR #1099, opened 2016-06-08
Enforce kind projection style using Scalastyle

peterneyens:
Now Scalastyle is working again, I have added the Scalastyle rule I mentioned in #1092 to make sure the `Lambda[...]` style is not use are used.


fthomas:
:+1:

non:
👍 Thanks!

PR #1100, opened 2016-06-08
Simplify the constraint on Xor/XorT/Ior to[] by using Alternative

peterneyens:
Replace seperate constraints `MonoidK` and `Applicative` with `Alternative`.


ceedubs:
👍 nice.

adelbertc:
:+1:

PR #1102, opened 2016-06-09
Add {C,c}omparison to Order, fixes #1101

### LINKED ISSUE #1101: cats.Order: Support for pattern matching on comparison results
When doing comparisons with cats.Order, would be desirable to pattern match on the cases, like eg:
```
a comparison b match {
  case GreaterThan => ???
  case Equivalent => ???
  case LessThan => ???
}
```
Order.compare returns an Int for performance and Java 1.7 compat reasons, which is fair enough, when you need that. But that requires an if (c <0) ... else if (c == 0) .. else ... construct to handle a comparison; less readable than the above alternative IMO.

Proposal is a function, implemented on top of `compare: Int`, perhaps named `comparison: Comparison` that returned a case object (GreaterThan etc) for pleasant pattern matching.

Gitter discussion: https://gitter.im/typelevel/cats?at=5758beb8064b9e7266f0c086

adelbertc:
Review by @non and @johnynek

### DIFF
--- a/kernel/src/main/scala/cats/kernel/Comparison.scala
+++ b/kernel/src/main/scala/cats/kernel/Comparison.scala
@@ -1,0 +1,19 @@
+package cats.kernel
+
+/** ADT encoding the possible results of a comparison */
+sealed abstract class Comparison extends Product with Serializable {
+  /** The signum of this comparison */
+  def toInt: Int = this match {
+    case Comparison.GreaterThan => 1
+    case Comparison.EqualTo     => 0
+    case Comparison.LessThan    => -1
+  }
+}
+
+object Comparison {
+  final case object GreaterThan extends Comparison
+  final case object EqualTo     extends Comparison
+  final case object LessThan    extends Comparison
+
+  implicit val catsKernelEqForComparison: Eq[Comparison] = Eq.fromUniversalEquals
+}
--- a/kernel/src/main/scala/cats/kernel/Comparison.scala
+++ b/kernel/src/main/scala/cats/kernel/Comparison.scala
@@ -1,19 +1,30 @@
 package cats.kernel
 
 /** ADT encoding the possible results of a comparison */
-sealed abstract class Comparison extends Product with Serializable {
-  /** The signum of this comparison */
-  def toInt: Int = this match {
-    case Comparison.GreaterThan => 1
-    case Comparison.EqualTo     => 0
-    case Comparison.LessThan    => -1
-  }
-}
+sealed abstract class Comparison(val toInt: Int, val toDouble: Double) extends Product with Serializable
 
 object Comparison {
-  final case object GreaterThan extends Comparison
-  final case object EqualTo     extends Comparison
-  final case object LessThan    extends Comparison
+  final case object GreaterThan extends Comparison(1, 1.0)
+  final case object EqualTo     extends Comparison(0, 0.0)
+  final case object LessThan    extends Comparison(-1, -1.0)
+
+  // Used for fromDouble
+  private val SomeGt = Some(Comparison.GreaterThan)
+  private val SomeEq = Some(Comparison.EqualTo)
+  private val SomeLt = Some(Comparison.LessThan)
+
+  def fromInt(int: Int): Comparison = {
+    if (int > 0)       Comparison.GreaterThan
+    else if (int == 0) Comparison.EqualTo
+    else               Comparison.LessThan
+  }
+
+  def fromDouble(double: Double): Option[Comparison] = {
+    if (double.isNaN) None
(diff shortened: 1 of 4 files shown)


kernel/src/main/scala/cats/kernel/Order.scala
```diff
@@ -29,6 +29,17 @@ trait Order[@sp A] extends Any with PartialOrder[A] { self =>
...
   def compare(x: A, y: A): Int
 
+  /**
+   * Like `compare`, but returns a [[cats.kernel.Comparison]] instead of an Int.
+   * Has the benefit of being able to pattern match on, but not as performant.
+   */
+  def comparison(x: A, y: A): Comparison = {
+    val r = compare(x, y)
+    if (r > 0)       Comparison.GreaterThan
+    else if (r == 0) Comparison.EqualTo
```
adelbertc: Is it OK we're using `==` instead of `Eq[Int]` ? I figure it's more performant but maybe not?

kernel/src/main/scala/cats/kernel/Comparison.scala
```diff
@@ -0,0 +1,19 @@
...
+    case Comparison.GreaterThan => 1
+    case Comparison.EqualTo     => 0
+    case Comparison.LessThan    => -1
+  }
+}
+
+object Comparison {
+  final case object GreaterThan extends Comparison
+  final case object EqualTo     extends Comparison
+  final case object LessThan    extends Comparison
```
ceedubs: It would probably be nice to have a `fromInt` method.

kernel/src/main/scala/cats/kernel/Comparison.scala
```diff
@@ -0,0 +1,19 @@
...
+
+  implicit val catsKernelEqForComparison: Eq[Comparison] = Eq.fromUniversalEquals
```
ceedubs: We could probably make this an `Order` as opposed to just an `Eq`. I don't know if it would ever really be used that way in practice, though.

kernel/src/main/scala/cats/kernel/Order.scala
non: Yes, when working with a known primitive (e.g. `Int`) using `==` is correct.

non: I agree with @ceedubs that this should probably use `Comparison.fromInt(compare(x, y))`.

non:
If we're going to do this, we should also create a `PartialComparison` object which parallels this but for partial orders.

ceedubs:
@non would it make sense to have a `partialComparison` method that just returns `Option[Comparison]` instead of creating a separate `IncomparableTo` type?

non:
Yeah, returning `Option[Comparison]` is fine too.

johnynek:
I like `Option[Comparison]` but I would allocate the values (other than None) once.

kernel/src/main/scala/cats/kernel/Comparison.scala
```diff
@@ -0,0 +1,19 @@
+package cats.kernel
+
+/** ADT encoding the possible results of a comparison */
+sealed abstract class Comparison extends Product with Serializable {
+  /** The signum of this comparison */
+  def toInt: Int = this match {
+    case Comparison.GreaterThan => 1
+    case Comparison.EqualTo     => 0
+    case Comparison.LessThan    => -1
+  }
```
johnynek: out of curiosity, @non, do you know it the above is faster or:
```scala
sealed abstract class Comparison(val toInt)
case object GreaterThan extends Comparison(1)
case object EqualTo extends Comparison(0)
case object LessThan extends Comparison(-1)
```
It seems to me the above would be.

non: Yeah I think having `Comparison` instances "know" their integer value (either through fields or through methods on the subtype) would be better than using pattern matching in `.toInt`.

ceedubs: It may be different with a `val` like that, but fwiw, at one point I benchmarked having scalaz's `Maybe` implement `fold` via pattern-matching in the parent vs dynamic dispatch to the child implementation and the former seemed to have a slight edge. It may be worth throwing together a benchmark if this is something that we care about. (I'm okay with it not being something that we care about, since people can always use the `Int` version of compare if they really want performance.)

non:
Looks great! Thanks so much. 👍

ceedubs:
Nice! Thanks @adelbertc.

non:
@ceedubs is that a +1 ?

ceedubs:
@non yep 👍 when the build succeeds.

Issue #1026, opened 2016-05-10
Use uber import in code examples


ceedubs:
Currently we have both ScalaDoc examples (via sbt-doctest) and tutorials (via tut) that import specific typeclass instances (ex: `import cats.std.option._`). At the Typelevel summit in Philly, several people mentioned that they had seen people struggle a lot with imports. There seemed to be a consensus that we should use `import cats.implicits._` in code samples to encourage people to start with the easier approach.

If users really want to be selective about what they bring into scope, then they can use more piecemeal imports. We should also provide some documentation (maybe even a flowchart?) for this use-case. However, I think more importantly we should use the uber import in code examples so people don't go down an unnecessary rabbit hole.

rabbitonweb:
I will be willing to contribute if u point to resources which should be updated.

ceedubs:
@rabbitonweb thanks for volunteering! Let's wait for a couple maintainers to 👍 this to make sure we want to go forward with this before you put too much effort into it. @non @adelbertc @tpolecat?

To find the resources, the best approach I can think of is to grep for `import cats.std` and `import cats.syntax` and ignore the ones that are happening inside tests or at the top of source files. Really we just want things within `/** ... */` blocks and `tut` sheds in `.md` files I think.

peterneyens:
While this will undoubtably remove some confusion for those getting started with (a library like) cats, I find that examples with the specific imports help those a little farther down the road to understand what is going on. It is an easy way to make the connection between a function and the underlying type class. Trying to use the right imports at least helped/helps me to improve my understanding of the underlying concepts.

What I am trying to say is that while reducing the first step, it may increase the next.

I am not sure how to make it easy to find the specific imports.

* Creating a good and easy to use flowchart to get the specific imports is probably not an easy task (If I just want the import for `|@|`, I wouldn't want to look at something similar to these [scalaz](http://xuwei-k.github.io/scalaz7.2.0.svg
) [charts](http://tpolecat.github.io/assets/scalaz.svg)).

* Just commenting out the specific imports probably adds to much noise (especially for those getting started). Comments also don't seem to play nice with sbt-doctest.

```scala
import cats.implicits._
// import cats.std.list._
// import cats.std.option._
// import cats.syntax.option._
// import cats.syntax.traverse._

List(1.some, 2.some).sequence
```

I am not sure if this should stop us from going ahead and using the “uber” import everywhere, just felt it is something to consider.

ceedubs:
[...]

PR #1104, opened 2016-06-09
Use uber import in docs. (#1026)

cranst0n:


non:
Love it! Thanks so much. 👍

ceedubs:
Oh fantastic @cranst0n. Thanks a bunch! Based on repeated Gitter questions, I think this will help a lot of people.

cranst0n:
@ceedubs @non 

I forgot to mention an issue that came up when making these changes. Specifically in `optiont.md` in the tut docs.

When I replaced the respective instances of `import cats.std.future._` tut fails to compile it:

```scala
[tut] *** Error reported at optiont.md:69
<console>:25: error: ambiguous implicit values:
 both value catsStdBimonadForFunction0 in trait Function0Instances of type => cats.Bimonad[Function0]
 and value catsStdInstancesForList in trait ListInstances of type => cats.Traverse[List] with cats.MonadCombine[List] with cats.MonadRec[List] with cats.CoflatMap[List]
 match expected type cats.Applicative[F]
         l <- OptionT.fromOption(lastnameO)
```

I didn't make the change because of this reason but I suppose it should be addressed in some way?

ceedubs:
@cranst0n oh thanks for the heads up. It looks like after changing the imports you need to change that line to `OptionT.fromOption[Future](lastnameO)`. I'm not really sure why that is. I'd be inclined to go ahead and change to the uber imports and do that anyway.

PR #1105, opened 2016-06-09
Add an Unapply with the signature of Nested.

peterneyens:
This would make it possible to use the syntax classes with `Nested`.


non:
Seems reasonable. Thanks! 👍

adelbertc:
Awesome, thanks! 👍

PR #1103, opened 2016-06-09
Add some more OptionT instances.

peterneyens:
- Added `Eq`, `PartialOrder` and `Order`.
- Added `Semigroup` and `Monoid`.
- Added `SemigroupK` and `MonoidK`
- Added `MonadError`, `MonadCombine` and `MonadFilter` (the last two
instances are dissabled, because they violate the same laws as the
equivalent `Xor`T instances).

PR #1106, opened 2016-06-09
XorTInstances + cleanup tests

yilinwei:
Add `Semigroup` and `Monoid` for `XorT`.

Also, cleanup the tests so the implicit resolution is clear.

### DIFF
--- a/core/src/main/scala/cats/data/Xor.scala
+++ b/core/src/main/scala/cats/data/Xor.scala
@@ -179,6 +179,10 @@ sealed abstract class Xor[+A, +B] extends Product with Serializable {
     a => s"Xor.Left(${AA.show(a)})",
     b => s"Xor.Right(${BB.show(b)})"
   )
+
+  def ap[AA >: A, BB >: B, C](that: AA Xor (BB => C)): AA Xor C = flatMap(
+    a => that.map(f => f(a))
+  )
 }
 
 object Xor extends XorInstances with XorFunctions {
--- a/core/src/main/scala/cats/data/Xor.scala
+++ b/core/src/main/scala/cats/data/Xor.scala
@@ -180,8 +180,8 @@ sealed abstract class Xor[+A, +B] extends Product with Serializable {
     b => s"Xor.Right(${BB.show(b)})"
   )
 
-  def ap[AA >: A, BB >: B, C](that: AA Xor (BB => C)): AA Xor C = flatMap(
-    a => that.map(f => f(a))
+  def ap[AA >: A, BB >: B, C](that: AA Xor (BB => C)): AA Xor C = that.flatMap(
+    f => this.map(a => f(a))
   )
 }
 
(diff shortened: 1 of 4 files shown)


non:
Looks good to me! Thanks! 👍

yilinwei:
As @peterneyens pointed out, the `Applicative` in this case isn't the same as dervied from the `Monad` (which one doesn't extend the other).

xuwei-k:
> Thought about this further and I think that it's not possible to have a Applicative for XorT

Yes, it's not possible. see https://github.com/scalaz/scalaz/pull/255

yilinwei:
I've removed the `Applicative`.

 I think it may be good to add a helper method onto `XorT` to find the `Nested` applicative like in scalaz but that's a different issue.

core/src/main/scala/cats/data/Xor.scala
```diff
@@ -179,6 +179,10 @@ sealed abstract class Xor[+A, +B] extends Product with Serializable {
     a => s"Xor.Left(${AA.show(a)})",
     b => s"Xor.Right(${BB.show(b)})"
   )
+
+  def ap[AA >: A, BB >: B, C](that: AA Xor (BB => C)): AA Xor C = that.flatMap(
+    f => this.map(a => f(a))
+  )
```
ceedubs: minor: I think this could simply be `that.flatMap(this.map)`

core/src/main/scala/cats/data/XorT.scala
```diff
@@ -316,8 +332,9 @@ private[data] trait XorTFunctor[F[_], L] extends Functor[XorT[F, L, ?]] {
 
 private[data] trait XorTMonad[F[_], L] extends Monad[XorT[F, L, ?]] with XorTFunctor[F, L] {
   implicit val F: Monad[F]
-  def pure[A](a: A): XorT[F, L, A] = XorT.pure[F, L, A](a)
+  def pure[A](a: A): XorT[F, L, A] = XorT(F.pure(Xor.right(a)))
   def flatMap[A, B](fa: XorT[F, L, A])(f: A => XorT[F, L, B]): XorT[F, L, B] = fa flatMap f
+  override def ap[A, B](x: XorT[F, L, A => B])(y: XorT[F, L, A]): XorT[F, L, B] = super.ap(x)(y)
```
ceedubs: Is there a reason to override this if we are just calling through to `super`?

tests/src/test/scala/cats/tests/ListWrapper.scala
```diff
@@ -44,14 +44,21 @@ object ListWrapper {
 
   def eqv[A : Eq]: Eq[ListWrapper[A]] = Eq[List[A]].on[ListWrapper[A]](_.list)
 
-  val foldable: Foldable[ListWrapper] =
-    new Foldable[ListWrapper] {
-      def foldLeft[A, B](fa: ListWrapper[A], b: B)(f: (B, A) => B): B =
-        Foldable[List].foldLeft(fa.list, b)(f)
+  val traverse: Traverse[ListWrapper] = {
```
ceedubs: Unfortunately I think there are going to be merge conflicts between this and #1103. I guess we'll have to get that sorted out after one of them is merged.

core/src/main/scala/cats/data/XorT.scala
yilinwei: Good spot - this is a left over from the conflicting `Applicative`.

tests/src/test/scala/cats/tests/ListWrapper.scala
yilinwei: #1103 is merged - should I resync this or would it tell me if there were conflicts?

ceedubs: Hmm interesting. It looks like maybe there's no merge conflict because the code is exactly the same?

ceedubs:
👍 thanks!

@non does this still look good to you?

tests/src/test/scala/cats/tests/ListWrapper.scala
peterneyens: Like I mentioned somewhere, I didn't reinvent the wheel and reused the `ListWrapper` instances @yilinwei wrote here. It's still interesting there is no merge conflict at all though.

non:
LGTM 👍

PR #1108, opened 2016-06-09
Relax Monad constraint for SemigroupK for XorT

yilinwei:
Can form a `SemigroupK` with an `Applicative` rather than `Monad`


non:
Looks good. 👍

yilinwei:
#1106 explains why this `Applicative` makes no sense and should be using a `Nested`.

Issue #1091, opened 2016-06-05
Size or length operation on Foldable


benhutchison:
Please provide a `size` method, with the ordinary meaning eg:
```
@ Foldable[List].foldLeft(List(1, 2, 3), 0){case (count, elem) => count + 1}
res6: Int = 3
```

Both (a) to provide a convenient default impl for a common operation, and (b) to allow instances to define their own faster size computation.

benhutchison:
background context from haskell: https://www.reddit.com/r/haskell/comments/xo05l/why_isnt_length_part_of_foldable/

ceedubs:
👍. I've also wondered if we should have a similar method that would be safe on infinite streams - something like "size, but stop calculating if it's over N".

ceedubs:
This was introduced in #1114.

PR #1114, opened 2016-06-11
Add size to Foldable

afiore:
Partially addresses what discussed in #1091.

### DIFF
--- a/core/src/main/scala/cats/Foldable.scala
+++ b/core/src/main/scala/cats/Foldable.scala
@@ -1,6 +1,7 @@
 package cats
 
 import scala.collection.mutable
+import cats.std.long._
 import simulacrum.typeclass
 
 /**
@@ -55,6 +56,16 @@ import simulacrum.typeclass
         case None => Later(Some(f(a)))
       }
     }
+
+  /**
+   * The size of this Foldable.
+   *
+   * This is overriden in structures that have more efficient size implementations
+   * (e.g. Vector, Set, Map).
+   *
+   * Note: will not terminate for infinite-sized collections.
+   */
+  def size[A](fa: F[A]): Long = foldMap(fa)(_ => 1)
 
   /**
    * Fold implemented using the given Monoid[A] instance.
--- a/tests/src/test/scala/cats/tests/OneAndTests.scala
+++ b/tests/src/test/scala/cats/tests/OneAndTests.scala
@@ -64,6 +64,12 @@ class OneAndTests extends CatsSuite {
   checkAll("NonEmptyList[Int]", ComonadTests[NonEmptyList].comonad[Int, Int, Int])
   checkAll("Comonad[NonEmptyList[A]]", SerializableTests.serializable(Comonad[NonEmptyList]))
 
+  test("size is consistent with toList.size") {
+    forAll { (oa: OneAnd[Vector, Int]) =>
+      oa.size should === (oa.toList.size.toLong)
+    }
+  }
+
   test("Show is not empty and is formatted as expected") {
     forAll { (nel: NonEmptyList[Int]) =>
       nel.show.nonEmpty should === (true)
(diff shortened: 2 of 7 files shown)


afiore:
Obviously, this naive implementation would not terminate on infinite streams. @yilinwei was suggesting that a possible approach could be to have `size` return an `Option[Long]`, which would allow tweaking the implementation for specific instances (e.g. Streams could return `None`).

non:
@afiore I would be fine with that behavior. On the other hand, we already have potentially non-terminating methods (like `.toList`) so I am fine with this method being unsafe as well (as long as we note that in the documentation).

ceedubs:
I like this approach as opposed to the `Option` approach. I think that if we want to have support for potentially infinite structures, it should be via a separate method that takes a bound so you can say something like "give me the size, but cap it at 1,000 instead blocking forever on an infinite stream" or something.

I think that this is a great start. There are some additional things that I think should be done:

- Add a ScalaDoc comment explaining that the default implementation is a naive fold over the structure meaning that 1) it may not terminate and 2) instances for structures that have efficient size implementations should override this
- Override `size` in instances for types that have efficient size implementations (`Vector` comes to mind - are there others?).

Would you be willing to do those? If not, I'm sure that someone else can get them in a followup PR.

core/src/main/scala/cats/Foldable.scala
```diff
@@ -56,6 +57,8 @@ import simulacrum.typeclass
       }
     }
 
+  def size[A](fa: F[A]): Long = foldMap(fa)(_ => 1)
```
ceedubs: 👍 I think that `Long` was a good choice here. `Int` is probably the first thing that would have come to my mind, but people could conceivably be using `Foldable` for some very large structures. Though hopefully if their structure is large enough to pass the max `Int` value they have overridden `size` to a more efficient implementation :)

ceedubs:
A couple other structures came to mind that could benefit from an override of `size`: `Set`, `Map, and `OneAnd`. There may be marginal performance improvements for `List` and `Stream`, but it probably doesn't matter much for those.

afiore:
@ceedubs I have added overrides for `Map`, `Set` and `Vector` using the standard library implementation of `size`, which I believe is defined in `scala.collection.TraversableOnce`. Not sure I understand how providing an override for `OneAnd` would help. Wouldn't it just use the instance corresponding to the structure used for the tail (e.g. `Foldable[Vector]` for `OneAnd[Vector, A]`)?

ceedubs:
> Not sure I understand how providing an override for OneAnd would help. Wouldn't it just use the instance corresponding to the structure used for the tail (e.g. Foldable[Vector] for OneAnd[Vector, A])?

@afiore It will use a `foldMap` derived from the first element plus the `foldLeft` on the tail structure. If instead there were an override for 1 + size(tail), then it could take advantage of any `size` optimizations that may be in the `Foldable[F]` type class for the tail (such as with `Vector`).  Does that make sense?

afiore:
@ceedubs makes sense! I have amended my commit adding an override for `AndOne`.

tests/src/test/scala/cats/tests/OneAndTests.scala
```diff
@@ -64,6 +64,12 @@ class OneAndTests extends CatsSuite {
   checkAll("NonEmptyList[Int]", ComonadTests[NonEmptyList].comonad[Int, Int, Int])
   checkAll("Comonad[NonEmptyList[A]]", SerializableTests.serializable(Comonad[NonEmptyList]))
 
+  test("Foldable[OneAnd[Vector, ?]]#size consistent with Vector#size") {
+    forAll { (oa: OneAnd[Vector, Int]) =>
+      oa.size should === (1L + oa.tail.size.toLong)
```
ceedubs: This seems to be duplicating the implementation within the test. Instead, it might be better to do something like `oa.size should === (oa.unwrap.size)` or `oa.size should === (oa.toList.size)`.

afiore:
@ceedubs I have updated `AndOneTests` with an assertion that doesn't repeat the actual implementation.

ceedubs:
👍 thanks, @afiore!

non:
👍

PR #1115, opened 2016-06-11
Use uber import in OptionT doc

ceedubs:
This (strangely?) required adding an explicit type parameter.

Follow up to
https://github.com/typelevel/cats/pull/1104#issuecomment-224989444


non:
👍

PR #1117, opened 2016-06-11
`foldLeftM` without `Free`.

TomasMikula:
This is an implementation of `foldLeftM` without using `Free`. Instead, it introduces an auxiliary structure

``` scala
case class Source[A](uncons: () => Option[(A, Source[A])])
```

and a conversion from any foldable `F[A]` to `Source[A]`, which provides lazy access to elements of `F[A]`.

It still uses `foldRight`, but I think there is a better separation of concerns here, because `G` is not involved in the `foldRight` at all.

~~I didn't bother to move this out of `Free.scala`, but `Free` is not used here at all.~~ I moved the implementation to `Foldable`.

Funny enough, if I make `Source` extend `AnyVal`, it crashes the compiler with a `StackOverflowError` ([SI-9600](https://issues.scala-lang.org/browse/SI-9600)).

### DIFF
--- a/core/src/main/scala/cats/Foldable.scala
+++ b/core/src/main/scala/cats/Foldable.scala
@@ -186,16 +186,31 @@ import simulacrum.typeclass
     foldLeft(fa, B.empty)((b, a) => B.combine(b, f(a)))
 
   /**
-   * Left associative monadic folding on `F`.
+   * Perform a stack-safe monadic left fold from the source context `F`
+   * into the target monad `G`.
    *
-   * The default implementation of this is based on `foldLeft`, and thus will
-   * always fold across the entire structure. Certain structures are able to
-   * implement this in such a way that folds can be short-circuited (not
-   * traverse the entirety of the structure), depending on the `G` result
-   * produced at a given step.
+   * This method can express short-circuiting semantics. Even when
+   * `fa` is an infinite structure, this method can potentially
+   * terminate if the `foldRight` implementation for `F` and the
+   * `tailRecM` implementation for `G` are sufficiently lazy.
+   *
+   * Instances for concrete structures (e.g. `List`) will often
+   * have a more efficient implementation than the default one
+   * in terms of `foldRight`.
    */
-  def foldM[G[_], A, B](fa: F[A], z: B)(f: (B, A) => G[B])(implicit G: Monad[G]): G[B] =
-    foldLeft(fa, G.pure(z))((gb, a) => G.flatMap(gb)(f(_, a)))
+  def foldM[G[_], A, B](fa: F[A], z: B)(f: (B, A) => G[B])(implicit G: Monad[G]): G[B] = {
+    val src = Foldable.Source.fromFoldable(fa)(self)
+    G.tailRecM((z, src)) { case (b, src) => src.uncons match {
+      case Some((a, src)) => G.map(f(b, a))(b => Left((b, src)))
+      case None => G.pure(Right(b))
+    }}
+  }
+
+  /**
+   * Alias for [[foldM]].
+   */
+  final def foldLeftM[G[_], A, B](fa: F[A], z: B)(f: (B, A) => G[B])(implicit G: Monad[G]): G[B] =
+    foldM(fa, z)(f)
 
   /**
    * Monadic folding on `F` by mapping `A` values to `G[B]`, combining the `B`
@@ -432,5 +447,35 @@ object Foldable {
       else M.pure(Right(b))
     }
     M.tailRecM(z)(go)
+  }
+
+
+  /**
+   * Isomorphic to
+   *
+   *     type Source[+A] = () => Option[(A, Source[A])]
+   *
+   * (except that recursive type aliases are not allowed).
+   *
+   * It could be made a value class after
+   * https://github.com/scala/bug/issues/9600 is resolved.
+   */
+  private sealed abstract class Source[+A] {
+    def uncons: Option[(A, Source[A])]
+  }
+
+  private object Source {
+    val Empty: Source[Nothing] = new Source[Nothing] {
+      def uncons = None
+    }
+
+    def cons[A](a: A, src: Eval[Source[A]]): Source[A] = new Source[A] {
+      def uncons = Some((a, src.value))
+    }
+
+    def fromFoldable[F[_], A](fa: F[A])(implicit F: Foldable[F]): Source[A] =
+      F.foldRight[A, Source[A]](fa, Now(Empty))((a, evalSrc) =>
+        Later(cons(a, evalSrc))
+      ).value
   }
 }
--- a/tests/src/test/scala/cats/tests/FoldableTests.scala
+++ b/tests/src/test/scala/cats/tests/FoldableTests.scala
@@ -204,6 +204,28 @@ class FoldableTestsAdditional extends CatsSuite {
     // test laziness of foldM
     dangerous.foldM(0)((acc, a) => if (a < 2) Some(acc + a) else None) should === (None)
   }
+
+  test(".foldLeftM short-circuiting") {
+    val ns = Stream.continually(1)
+    val res = Foldable[Stream].foldLeftM[Either[Int, ?], Int, Int](ns, 0) { (sum, n) =>
+      if (sum >= 100000) Left(sum) else Right(sum + n)
+    }
+    assert(res == Left(100000))
+  }
+
+  test(".foldLeftM short-circuiting optimality") {
+    // test that no more elements are evaluated than absolutely necessary
+
+    def concatUntil(ss: Stream[String], stop: String): Either[String, String] =
+      Foldable[Stream].foldLeftM[Either[String, ?], String, String](ss, "") { (acc, s) =>
+        if (s == stop) Left(acc) else Right(acc + s)
+      }
+
+    def boom: Stream[String] = sys.error("boom")
+    assert(concatUntil("STOP" #:: boom, "STOP") == Left(""))
+    assert(concatUntil("Zero" #:: "STOP" #:: boom, "STOP") == Left("Zero"))
+    assert(concatUntil("Zero" #:: "One" #:: "STOP" #:: boom, "STOP") == Left("ZeroOne"))
+  }
 }
 
 class FoldableListCheck extends FoldableCheck[List]("list") {
--- a/free/src/main/scala/cats/free/Free.scala
+++ b/free/src/main/scala/cats/free/Free.scala
@@ -254,23 +254,4 @@ object Free {
           case Right(b) => pure(b)
         }
     }
-
-  /**
-   * Perform a stack-safe monadic fold from the source context `F`
-   * into the target monad `G`.
-   *
-   * This method can express short-circuiting semantics. Even when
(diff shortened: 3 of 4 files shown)


ceedubs:
@TomasMikula thanks, this does seem quite a bit simpler than the `Free` approach. This week I'll try to take some time to figure out which other methods we want better stack-safe/lazy support for and whether we can do something similar there as opposed to needing to bring `Free` into `core`.

TomasMikula:
This approach basically gives an Iterator-like/Stream-like interface to (any) `Foldable`, which could potentially be useful elsewhere. Also, I think pretty much any concrete foldable data type will have a much more efficient implementation of `foldLeftM` (or "toStream") than this default one.

ceedubs:
@TomasMikula I like this. I'd be interested in moving it into core somewhere where we could use it in the default implementation of `Foldable.foldLeftM` I think (though we could maybe leave it `private` for now so it doesn't become a part of the official cats API for now). I'd also be interested in some benchmarks to see whether there's any performance difference if you were to use an abstract class with an abstract `def` as opposed to a case class with a `Function0` val. I'm happy to put these together, but I also don't want to steal your thunder if this is something that you want to see through. WDYT?

ceedubs:
cc @non since he and I have talked about `foldLeftM` and the possibility of moving `Free` to `core` to get the functionality that this achieves.

TomasMikula:
@ceedubs I made the changes. I made `Source` to be an abstract class. I was originally going for a value class, but hit [SI-9600](https://issues.scala-lang.org/browse/SI-9600).

TomasMikula:
Perhaps we need a better naming for `foldM` vs. `foldLeftM`? They are both left folds. They differ in the ability to short-circuit.

tests/src/test/scala/cats/tests/FoldableTests.scala
```diff
@@ -113,6 +114,25 @@ class FoldableTestsAdditional extends CatsSuite {
     val large = Stream((1 to 10000): _*)
     assert(contains(large, 10000).value)
   }
+
+  test(".foldLeftM") {
+    // you can see .foldLeftM traversing the entire structure by
+    // changing the constant argument to .take and observing the time
+    // this test takes.
```
ceedubs: This comment should no longer be true, should it?

ceedubs:
> Perhaps we need better naming for foldM vs. foldLeftM? They are both left folds. They differ in the ability to short-circuit.

@TomasMikula I agree. I think we are going to likely end up having several methods that have a `Monad` form and a separate `MonadRec` form. Maybe to fit the names of those two type classes we could use the convention of `foldLeftM` and `foldLeftMRec`?

core/src/main/scala/cats/Foldable.scala
```diff
@@ -280,4 +304,23 @@ object Foldable {
...
+
+    def cons[A](a: A, src: Eval[Source[A]]): Source[A] = new Source[A] {
+      def uncons = Some((a, src.value))
+    }
+
+    def fromFoldable[F[_], A](fa: F[A])(implicit F: Foldable[F]): Source[A] =
+      F.foldRight[A, Source[A]](fa, Now(Source.empty[A]))((a, evalSrc) =>
+        Now(cons(a, evalSrc))
+      ).value
+  }
```
ceedubs: This is probably me prematurely optimizing, but I'd be tempted to do a couple things:

- Make `empty` reuse a singleton either by making `Source` covariant or by giving it a `def widen[AA >: A]: Source[AA] = this.asInstanceOf[Source[AA]]` method.
- Possibly store a value of `Now(Source.empty)` around for reuse by `fromFoldable`.

tests/src/test/scala/cats/tests/FoldableTests.scala
TomasMikula: Right, it became obsolete earlier, though. [See my comment.](https://github.com/TomasMikula/cats/commit/3eca17fd1f05d861599ce6a697d96ae8a6c5b762#commitcomment-17900340) The whole test case is redundant in favor of the one below.

TomasMikula:
@ceedubs made the changes, except renaming. I like the -`Rec` suffix convention. Do you mean we should rename `foldM` to `foldLeftM`? Or the current `foldLeftM` to `foldMRec`?

ceedubs:
@TomasMikula I was thinking about renaming the current `foldM` to `foldLeftM` and having this new one be `foldLeftMRec`. I guess if we don't have a right version then there is no harm in `foldM` and `foldMRec`, but I think it makes it a little more obvious that it will be left-associated and parallels the `foldLeft` name better. WDYT?

TomasMikula:
@ceedubs Makes sense. My concern was just whether `foldM` was a standard name for this operation. Probably not. `scalaz` has `foldLeftM`, Haskell has `foldlM`.

TomasMikula:
How about now?

ceedubs:
👍 thanks @TomasMikula!

It might be nice to have the `foldLeftM` and `foldLeftMRec` scaladocs reference each other to point out why people might want one or the other. Also we'll probably want to optimize the implementations in some of our instances. But I think all of that can wait for subsequent PRs. I'd rather get this merged and then iterate.

kailuowang:
@TomasMikula I am sorry that this fell through the cracks, any interest in resolving the conflicts? It looks like that we are close to merging.

core/src/main/scala/cats/Foldable.scala
```diff
@@ -78,11 +80,33 @@ import simulacrum.typeclass
 
   /**
    * Left associative monadic folding on `F`.
+   * Stack safe, not short-circuiting (always folds the whole structure).
    */
-  def foldM[G[_], A, B](fa: F[A], z: B)(f: (B, A) => G[B])(implicit G: Monad[G]): G[B] =
+  def foldLeftM[G[_], A, B](fa: F[A], z: B)(f: (B, A) => G[B])(implicit G: Monad[G]): G[B] =
```
johnynek: we don't need this now, right?

core/src/main/scala/cats/Foldable.scala
```diff
@@ -78,11 +80,33 @@ import simulacrum.typeclass
...
+   * This method can express short-circuiting semantics. Even when
+   * `fa` is an infinite structure, this method can potentially
+   * terminate if the `foldRight` implementation for `F` and the
+   * `tailRecM` implementation for `G` are sufficiently lazy.
+   *
+   * Instances for concrete structures (e.g. `List`) will often
+   * have a more efficient implementation than the default one
+   * in terms of `foldRight`.
+   */
+  def foldLeftMRec[G[_], A, B](fa: F[A], z: B)(f: (B, A) => G[B])(implicit G: MonadRec[G]): G[B] = {
```
johnynek: can we rename this `foldLeftM` or just stick with `foldM`. I'm not sure we should be changing this name this late in the game, although I agree `foldLeftM` is a better name.

Maybe add a deprecated final `foldM` that redirects to `foldLeftM`?

TomasMikula:
How about now?

core/src/main/scala/cats/Foldable.scala
```diff
@@ -433,4 +448,26 @@ object Foldable {
...
+
+  private object Source {
+    val Empty: Source[Nothing] = new Source[Nothing] {
+      def uncons = None
+    }
+
+    val NowEmpty: Eval[Source[Nothing]] = Now(Empty)
+
+    def cons[A](a: A, src: Eval[Source[A]]): Source[A] = new Source[A] {
+      def uncons = Some((a, src.value))
```
johnynek: I'm a little nervous about these calls to value. Can we add a comment about the stack safety here? I think the answer is that we can't recursively call .value in an Eval and we should never be in an Eval when we call uncons, is that right?

core/src/main/scala/cats/Foldable.scala
```diff
@@ -186,16 +186,31 @@ import simulacrum.typeclass
...
+   * `fa` is an infinite structure, this method can potentially
+   * terminate if the `foldRight` implementation for `F` and the
+   * `tailRecM` implementation for `G` are sufficiently lazy.
+   *
+   * Instances for concrete structures (e.g. `List`) will often
+   * have a more efficient implementation than the default one
+   * in terms of `foldRight`.
+   */
+  def foldLeftM[G[_], A, B](fa: F[A], z: B)(f: (B, A) => G[B])(implicit G: Monad[G]): G[B] = {
+    val src = Foldable.Source.fromFoldable(fa)(self)
```
johnynek: Kind of a bummer that is someone has override on foldM they will have two different folds. Maybe one should be final and a pointer to the other?

core/src/main/scala/cats/Foldable.scala
TomasMikula: The presence of `Eval` in this case actually just obstructs the understanding, IMO. The `src` parameter of `cons` could just be by-name (the evaluation of `.value` doesn't need any trampolining built into `Eval`). However, since `foldRight` already uses `Eval`, I didn't want to mix both by-name and `Eval`.

A clearer definition of `Source` would be

```scala
type Source[A] = () => Option[(A, Source[A])]
```

except that recursive type aliases are not allowed. I guess I could make it an `AnyVal`, though.

core/src/main/scala/cats/Foldable.scala
TomasMikula: True.

TomasMikula:
I realized that in case of short-circuited fold, it evaluates 1 more element than necessary. Maybe I should fix that...

johnynek:
👍

kailuowang:
merged with 3 sign-offs

tests/src/test/scala/cats/tests/FoldableTests.scala
```diff
@@ -204,6 +204,28 @@ class FoldableTestsAdditional extends CatsSuite {
...
+
+  test(".foldLeftM short-circuiting") {
+    val ns = Stream.continually(1)
+    val res = Foldable[Stream].foldLeftM[Either[Int, ?], Int, Int](ns, 0) { (sum, n) =>
+      if (sum >= 100000) Left(sum) else Right(sum + n)
+    }
+    assert(res == Left(100000))
+  }
+
+  test(".foldLeftM short-circuiting optimality") {
```
kailuowang: This test fails if we remove `Foldable.iteratorFoldM`, because it actually no longer tests the default `foldM` implementation from `Foldable`  after #1414 which overrides foldLeftM in the instance of `Stream` using `iteratorFoldM`. This test fails when using the default `foldM` because, if I am not mistaken, the new `foldM` still have to evaluate one element after the stop. Also these two tests are now redundant with the test("Foldable[Stream]") above. I will fix this in a PR removing `iteratorFoldM`

TomasMikula: Huh, I didn't notice that I was testing some other code 😃. I suppose changing the signature of

```scala
def uncons: Option[(A, Source[A])]
```

to

```scala
def uncons: Option[(A, Eval[Source[A]])]
```

and doing the necessary changes would make it pass.

kailuowang: right, I updated #1740

PR #1119, opened 2016-06-12
Fix contramap signature in Contravariant doc page

afiore:
I think I have spotted a typo in the docs. Find the actual function definition [here](https://github.com/typelevel/cats/blob/f592db84facd88d51d3f64e0e309f320878b8c52/core/src/main/scala/cats/functor/Contravariant.scala#L10).


yilinwei:
:+1:

ceedubs:
👍 thanks!

Issue #1087, opened 2016-06-03
Create cats.data.NonEmptyList


ceedubs:
We seem to have arrived on a decision in #537 that `NonEmptyList` should have its own data type. Let's go for it!

ceedubs:
@WarFox is working on this in #1120.

ceedubs:
Resolved by #1231

PR #1120, opened 2016-06-12
NonEmptyList work in progress [do-not-merge-yet]

WarFox:
This is a work in progress branch for #1087 created during Hack the tower meet up in London on Saturday, 11 June 2016.

### DIFF
--- a/core/src/main/scala/cats/data/NonEmptyList.scala
+++ b/core/src/main/scala/cats/data/NonEmptyList.scala
@@ -14,7 +14,7 @@ final case class NonEmptyList[A](head: A, tail: List[A] = List[A]()) {
   /**
    * Return the head and tail into a single list
    */
-  def unwrap: List[A] = head :: tail
+  def toList: List[A] = head :: tail
 
   /**
    * remove elements not matching the predicate
--- a/core/src/main/scala/cats/data/NonEmptyList.scala
+++ b/core/src/main/scala/cats/data/NonEmptyList.scala
@@ -6,63 +6,154 @@ import cats.std.list._
 //it needs semigroup - combine, filter
 
 /**
-  *  A data type which represents a single element (head) and some other
-  * structure (tail).
-  */
-final case class NonEmptyList[A](head: A, tail: List[A]) {
+ *  A data type which represents a single element (head) and some other
+ * structure (tail).
+ */
+final case class NonEmptyList[A](head: A, tail: List[A] = List[A]()) {
 
   /**
-    * Return the head and tail into a single list
-    */
+   * Return the head and tail into a single list
+   */
   def unwrap: List[A] = head :: tail
 
   /**
-    * remove elements not matching the predicate
-    */
+   * remove elements not matching the predicate
+   */
   def filter(p: A => Boolean): List[A] = {
     val rest = tail.filter(p)
-    if(p(head)) head::rest else rest
+    if (p(head)) head :: rest else rest
   }
 
   /**
-    * Append another NonEmptyList
-    */
-  def combine(other: NonEmptyList[A]):NonEmptyList[A] = 
-    NonEmptyList(head, MonadCombine[List].combineK(tail, other.head::other.tail))
+   * Append another NonEmptyList
+   */
+  def combine(other: NonEmptyList[A]): NonEmptyList[A] =
+    NonEmptyList(head, MonadCombine[List].combineK(tail, other.head :: other.tail))
 
   /**
-    * Find the first element matching the predicate, if one exists
-    */
-  def find(p:A=>Boolean): Option[A] =
-    if(p(head)) Some(head) else tail.find(p)
+   * Find the first element matching the predicate, if one exists
+   */
+  def find(p: A => Boolean): Option[A] =
+    if (p(head)) Some(head) else tail.find(p)
 
   /**
-    * Check whether at least one element satisfies the predicate
-    */
+   * Check whether at least one element satisfies the predicate
+   */
   def exists(p: A => Boolean): Boolean =
     p(head) || tail.exists(p)
 
   /**
-    * Check whether all elements satisfy the predicate
-    */
+   * Check whether all elements satisfy the predicate
+   */
   def forall(p: A => Boolean): Boolean =
     p(head) && tail.exists(p)
 
   /**
-    * Left-associative fold on the structure using f.
-    */
+   * Left-associative fold on the structure using f.
+   */
   def foldLeft[B](b: B)(f: (B, A) => B): B =
-    (head::tail).foldLeft(b)(f)
+    (head :: tail).foldLeft(b)(f)
 
   /**
-    * Right-associative fold on the structure using f.
-    */
-  def foldRight[B](lb: Eval[B])(f: (A, Eval[B]) => Eval[B]):Eval[B] =
-    (head::tail).foldRight(lb)(f)
+   * Right-associative fold on the structure using f.
+   */
+  def foldRight[B](lb: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B] =
+    (head :: tail).foldRight(lb)(f)
 
   /**
-    *  Applies f to all the elements of the structure
-    */
-  def map[B](f: A => B):NonEmptyList[B] =
+   *  Applies f to all the elements of the structure
+   */
+  def map[B](f: A => B): NonEmptyList[B] =
     NonEmptyList(f(head), tail.map(f))
 }
+
+private[data] sealed trait NonEmptyListInstances extends NonEmptyListLowPriority2 {
+
+  implicit def catsDataEqForNonEmptyList[A](implicit A: Eq[A]): Eq[NonEmptyList[A]] =
+    new Eq[NonEmptyList[A]]{
+      def eqv(x: NonEmptyList[A], y: NonEmptyList[A]): Boolean = x === y
+    }
+
+  implicit def catsDataShowForNonEmptyList[A](implicit A: Show[A]): Show[NonEmptyList[A]] =
+    Show.show[NonEmptyList[A]](_.show)
+
+  implicit def catsDataSemigroupKForNonEmptyList[A]: SemigroupK[NonEmptyList[?]] =
+    new SemigroupK[NonEmptyList[?]] {
+      def combineK[A](a: NonEmptyList[A], b: NonEmptyList[A]): NonEmptyList[A] =
+        a combine b
+    }
+
+  implicit def catsDataSemigroupForNonEmptyList[A]: Semigroup[NonEmptyList[A]] =
+    catsDataSemigroupKForNonEmptyList[F].algebra
+
+  implicit def catsDataReducibleForNonEmptyList[A]: Reducible[NonEmptyList[?]] =
+    new NonEmptyReducible[NonEmptyList[?]] {
+      override def split[A](fa: NonEmptyList[A]): (A, F[A]) = (fa.head, fa.tail)
+    }
+
+  implicit def catsDataMonadForNonEmptyList[A]: Monad[NonEmptyList[?]] =
+    new Monad[NonEmptyList[?]] {
+      override def map[A, B](fa: NonEmptyList[F, A])(f: A => B): NonEmptyList[F, B] =
+        fa map f
+
+      def pure[A](x: A): NonEmptyList[F, A] =
+        NonEmptyList(x, monad.empty)
+
+      def flatMap[A, B](fa: NonEmptyList[F, A])(f: A => NonEmptyList[F, B]): NonEmptyList[F, B] = {
+        val end = monad.flatMap(fa.tail) { a =>
+          val fa = f(a)
+          monad.combineK(monad.pure(fa.head), fa.tail)
+        }
+        val fst = f(fa.head)
+        NonEmptyList(fst.head, monad.combineK(fst.tail, end))
+      }
+    }
+}
+
+trait NonEmptyListLowPriority0 {
+  implicit val nelComonad: Comonad[NonEmptyList[List, ?]] =
+    new Comonad[NonEmptyList[List, ?]] {
+      def coflatMap[A, B](fa: NonEmptyList[List, A])(f: NonEmptyList[List, A] => B): NonEmptyList[List, B] = {
+        @tailrec def consume(as: List[A], buf: ListBuffer[B]): List[B] =
+          as match {
+            case Nil => buf.toList
+            case a :: as => consume(as, buf += f(NonEmptyList(a, as)))
+          }
+        NonEmptyList(f(fa), consume(fa.tail, ListBuffer.empty))
+      }
+
+      def extract[A](fa: NonEmptyList[List, A]): A =
+        fa.head
+
+      def map[A, B](fa: NonEmptyList[List, A])(f: A => B): NonEmptyList[List, B] =
+        fa map f
+    }
+}
+
+trait NonEmptyListLowPriority1 extends NonEmptyListLowPriority0 {
+  implicit def catsDataFunctorForNonEmptyList[F[_]](implicit F: Functor[F]): Functor[NonEmptyList[F, ?]] =
+    new Functor[NonEmptyList[F, ?]] {
+      def map[A, B](fa: NonEmptyList[F, A])(f: A => B): NonEmptyList[F, B] =
+        fa map f
+    }
+
+}
+
+trait NonEmptyListLowPriority2 extends NonEmptyListLowPriority1 {
+  implicit def catsDataTraverseForNonEmptyList[F[_]](implicit F: Traverse[F]): Traverse[NonEmptyList[F, ?]] =
+    new Traverse[NonEmptyList[F, ?]] {
+      def traverse[G[_], A, B](fa: NonEmptyList[F, A])(f: (A) => G[B])(implicit G: Applicative[G]): G[NonEmptyList[F, B]] = {
+        G.map2Eval(f(fa.head), Always(F.traverse(fa.tail)(f)))(NonEmptyList(_, _)).value
+      }
+
+      def foldLeft[A, B](fa: NonEmptyList[F, A], b: B)(f: (B, A) => B): B = {
+        fa.foldLeft(b)(f)
+      }
+
+      def foldRight[A, B](fa: NonEmptyList[F, A], lb: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B] = {
+        fa.foldRight(lb)(f)
+      }
+    }
+}
+
+object NonEmptyList extends NonEmptyListInstances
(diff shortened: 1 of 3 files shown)


ceedubs:
Thanks @WarFox -- this looks like a great start!

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -0,0 +1,68 @@
...
+/**
+  *  A data type which represents a single element (head) and some other
+  * structure (tail).
+  */
+final case class NonEmptyList[A](head: A, tail: List[A]) {
+
+  /**
+    * Return the head and tail into a single list
+    */
+  def unwrap: List[A] = head :: tail
```
ceedubs: what do you think of us simply calling this `toList`?

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -0,0 +1,68 @@
...
+  def filter(p: A => Boolean): List[A] = {
+    val rest = tail.filter(p)
+    if(p(head)) head::rest else rest
+  }
+
+  /**
+    * Append another NonEmptyList
+    */
+  def combine(other: NonEmptyList[A]):NonEmptyList[A] = 
+    NonEmptyList(head, MonadCombine[List].combineK(tail, other.head::other.tail))
```
ceedubs: Minor: this could use `other.unwrap` (or `other.toList` if we go with that name) instead of `other.head :: other.tail`, and there are a few other places in this file that could do the same.

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -0,0 +1,68 @@
...
+    */
+  def foldRight[B](lb: Eval[B])(f: (A, Eval[B]) => Eval[B]):Eval[B] =
+    (head::tail).foldRight(lb)(f)
+
+  /**
+    *  Applies f to all the elements of the structure
+    */
+  def map[B](f: A => B):NonEmptyList[B] =
+    NonEmptyList(f(head), tail.map(f))
+}
```
ceedubs: I think in the companion object (when it's created), it would be nice to have a `def apply[A](head: A, tail: A*) = NonEmptyList(head, tail.toList)`. That way people can do something like `NonEmptyList(1, 2, 3, 4)` as they can with most other collections.

core/src/main/scala/cats/data/NonEmptyList.scala
WarFox: Thanks for the comments @ceedubs.. I'm learning!
I think it makes sense to call it `toList`. I was just trying to follow `OneAnd.scala`, thought it was a convention

core/src/main/scala/cats/data/NonEmptyList.scala
WarFox: yes, definitely

core/src/main/scala/cats/data/NonEmptyList.scala
ceedubs: I think in `OneAnd` it's called `unwrap` since `F` is abstract so there's not a straightforward name to call it. Since we are fixed to `List` here, it's easier to come up with a good name :)

WarFox: makes sense, I've pushed that change

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -0,0 +1,159 @@
...
+  def filter(p: A => Boolean): List[A] = {
+    val rest = tail.filter(p)
+    if (p(head)) head :: rest else rest
+  }
+
+  /**
+   * Append another NonEmptyList
+   */
+  def combine(other: NonEmptyList[A]): NonEmptyList[A] =
+    NonEmptyList(head, MonadCombine[List].combineK(tail, other.head :: other.tail))
```
non: I think this would be simpler as:

```scala
NonEmptyList(head, tail ::: other.toList)
```

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -0,0 +1,159 @@
...
+
+import cats.std.list._
+//it needs a Functor
+//it needs semigroup - combine, filter
+
+/**
+ *  A data type which represents a single element (head) and some other
+ * structure (tail).
+ */
+final case class NonEmptyList[A](head: A, tail: List[A] = List[A]()) {
```
non: I would prefer not to have a default parameter here -- we can support calling with one argument via an additional `apply` method on the companion object.

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -0,0 +1,159 @@
...
+   * Check whether all elements satisfy the predicate
+   */
+  def forall(p: A => Boolean): Boolean =
+    p(head) && tail.exists(p)
+
+  /**
+   * Left-associative fold on the structure using f.
+   */
+  def foldLeft[B](b: B)(f: (B, A) => B): B =
+    (head :: tail).foldLeft(b)(f)
```
non: I think this might be better-written as `tail.foldLeft(f(b, head))(f)`.

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -0,0 +1,159 @@
...
+
+  implicit def catsDataMonadForNonEmptyList[A]: Monad[NonEmptyList[?]] =
+    new Monad[NonEmptyList[?]] {
+      override def map[A, B](fa: NonEmptyList[F, A])(f: A => B): NonEmptyList[F, B] =
+        fa map f
+
+      def pure[A](x: A): NonEmptyList[F, A] =
+        NonEmptyList(x, monad.empty)
+
+      def flatMap[A, B](fa: NonEmptyList[F, A])(f: A => NonEmptyList[F, B]): NonEmptyList[F, B] = {
```
non: Could we move this method (and other type class methods) to `NonEmptyList` and then reference them in the instances? I think that will make the scaladoc for `NonEmptyList` a bit nicer.

non:
Great job, thanks! This is looking really good. Feel free to remove the commented-out type aliases.

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -0,0 +1,159 @@
...
+      override def split[A](fa: NonEmptyList[A]): (A, F[A]) = (fa.head, fa.tail)
+    }
+
+  implicit def catsDataMonadForNonEmptyList[A]: Monad[NonEmptyList[?]] =
+    new Monad[NonEmptyList[?]] {
+      override def map[A, B](fa: NonEmptyList[F, A])(f: A => B): NonEmptyList[F, B] =
+        fa map f
+
+      def pure[A](x: A): NonEmptyList[F, A] =
+        NonEmptyList(x, monad.empty)
```
non: Let's just use `Nil` instead of `monad.empty` here, since we know we are talking about `List`.

core/src/main/scala/cats/data/NonEmptyList.scala
WarFox: very much cleaner

core/src/main/scala/cats/data/NonEmptyList.scala
WarFox: 👍 agreed.
@ceedubs had also pointed it out here  https://github.com/typelevel/cats/pull/1120#discussion_r66786120
to have apply method in instances

```
def apply[A](head: A, tail: A*) = NonEmptyList(head, tail.toList)
```

core/src/main/scala/cats/data/NonEmptyList.scala
WarFox: Is `tail.foldLeft(f(b, head))(f)`  same as `(head :: tail).foldLeft(b)(f)`?

I'm quite new to this, it's a bit confusing. Sorry!

core/src/main/scala/cats/data/NonEmptyList.scala
WarFox: thanks for the detailed review :)
this is motivational
I shall do this changes tonight

core/src/main/scala/cats/data/NonEmptyList.scala
WarFox: got it!
First iteration of `f` will be `f(b, head)`, in (head :: tail).foldLeft(b)(f)

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -0,0 +1,159 @@
...
+   * Return the head and tail into a single list
+   */
+  def toList: List[A] = head :: tail
+
+  /**
+   * remove elements not matching the predicate
+   */
+  def filter(p: A => Boolean): List[A] = {
+    val rest = tail.filter(p)
+    if (p(head)) head :: rest else rest
```
yilinwei: Would be tempted to go with `(head :: tail).filter(p)`

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -0,0 +1,159 @@
...
+   */
+  def combine(other: NonEmptyList[A]): NonEmptyList[A] =
+    NonEmptyList(head, MonadCombine[List].combineK(tail, other.head :: other.tail))
+
+  /**
+   * Find the first element matching the predicate, if one exists
+   */
+  def find(p: A => Boolean): Option[A] =
+    if (p(head)) Some(head) else tail.find(p)
+
```
yilinwei: Again, `unwrap.find(p)`

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -0,0 +1,159 @@
...
+   * Right-associative fold on the structure using f.
+   */
+  def foldRight[B](lb: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B] =
+    (head :: tail).foldRight(lb)(f)
+
+  /**
+   *  Applies f to all the elements of the structure
+   */
+  def map[B](f: A => B): NonEmptyList[B] =
+    NonEmptyList(f(head), tail.map(f))
```
yilinwei: It would be nice to have variance on this particular method `def map[AA <: A, B](f: AA => B): NonEmptyList[B]` so that you can pass a function for a supertype of `A` into map.

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -0,0 +1,159 @@
...
+      def pure[A](x: A): NonEmptyList[F, A] =
+        NonEmptyList(x, monad.empty)
+
+      def flatMap[A, B](fa: NonEmptyList[F, A])(f: A => NonEmptyList[F, B]): NonEmptyList[F, B] = {
+        val end = monad.flatMap(fa.tail) { a =>
+          val fa = f(a)
+          monad.combineK(monad.pure(fa.head), fa.tail)
+        }
+        val fst = f(fa.head)
+        NonEmptyList(fst.head, monad.combineK(fst.tail, end))
```
yilinwei: Since we know the concrete type we can write this a little nicer as:

```
val h :: tail = f(head) ++ tail.flatMap(f)
NonEmptyList(h, tail)
```

yilinwei:
Thanks for finishing the issue from hack the tower and putting it up for a pull request! :+1:

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -0,0 +1,159 @@
...
+  implicit def catsDataReducibleForNonEmptyList[A]: Reducible[NonEmptyList[?]] =
+    new NonEmptyReducible[NonEmptyList[?]] {
+      override def split[A](fa: NonEmptyList[A]): (A, F[A]) = (fa.head, fa.tail)
+    }
+
+  implicit def catsDataMonadForNonEmptyList[A]: Monad[NonEmptyList[?]] =
+    new Monad[NonEmptyList[?]] {
+      override def map[A, B](fa: NonEmptyList[F, A])(f: A => B): NonEmptyList[F, B] =
+        fa map f
+
```
zainab-ali: `NonEmptyList` should have only one type parameter, so this should be `NonEmptyList[B]` instead of `NonEmptyList[F, B]`.  This should cause compilation errors.

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -0,0 +1,159 @@
...
+        NonEmptyList(f(fa), consume(fa.tail, ListBuffer.empty))
+      }
+
+      def extract[A](fa: NonEmptyList[List, A]): A =
+        fa.head
+
+      def map[A, B](fa: NonEmptyList[List, A])(f: A => B): NonEmptyList[List, B] =
+        fa map f
+    }
+}
```
zainab-ali: Since a `Comonad` is a `Functor`, the `Comonad` here can extend a `Functor`, in which case you won't need to rewrite `map`.  The same is true for `CoflatMap`.

```
trait NonEmptyListComonad extends Comonad[NonEmptyList] 
   with NonEmptyListFunctor with NonEmptyListCoflatMap {
   //extract is the only method you need to write
   def extract[A](fa: NonEmptyList[A]): A = fa.head
}
```

It would be good to change the `catsDataFunctorForNonEmptyList` to a `NonEmptyListFunctor` trait for this purpose

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -0,0 +1,159 @@
...
+    new SemigroupK[NonEmptyList[?]] {
+      def combineK[A](a: NonEmptyList[A], b: NonEmptyList[A]): NonEmptyList[A] =
+        a combine b
+    }
+
+  implicit def catsDataSemigroupForNonEmptyList[A]: Semigroup[NonEmptyList[A]] =
+    catsDataSemigroupKForNonEmptyList[F].algebra
+
+  implicit def catsDataReducibleForNonEmptyList[A]: Reducible[NonEmptyList[?]] =
+    new NonEmptyReducible[NonEmptyList[?]] {
```
zainab-ali: Now that `NonEmptyList` only takes one parameter, we should be able to remove the `?` here.
`Reducible[NonEmptyList]` should work fine.

core/src/main/scala/cats/data/NonEmptyList.scala
WarFox: thanks, I shall try that

core/src/main/scala/cats/data/NonEmptyList.scala
non: I actually prefer being explicit and not creating a list when not necessary. I'll leave it to @WarFox to decide which way to go.

tests/src/test/scala/cats/tests/NonEmptyListTests.scala
```diff
@@ -0,0 +1,165 @@
...
+  checkAll("NonEmptyList[List, Int]", OrderLaws[NonEmptyList[List, Int]].eqv)
+
+  checkAll("NonEmptyList[List, Int] with Option", TraverseTests[NonEmptyList[List, ?]].traverse[Int, Int, Int, Int, Option, Option])
+  checkAll("Traverse[NonEmptyList[List, A]]", SerializableTests.serializable(Traverse[NonEmptyList[List, ?]]))
+
+  checkAll("NonEmptyList[List, Int]", ReducibleTests[NonEmptyList[List, ?]].reducible[Option, Int, Int])
+  checkAll("Reducible[NonEmptyList[List, ?]]", SerializableTests.serializable(Reducible[NonEmptyList[List, ?]]))
+
+  implicit val iso = CartesianTests.Isomorphisms.invariant[NonEmptyList[ListWrapper, ?]](NonEmptyList.catsDataFunctorForNonEmptyList(ListWrapper.functor))
+
```
yilinwei: You don't need 2 `iso`'s if the second one resolves.

core/src/main/scala/cats/data/NonEmptyList.scala
ceedubs: For `OneAnd`, where `F` is abstract, you have to break down a bunch of separate instances. Since we are now specializing to `List`, I think we can probably get by without creating intermediate traits like `NonEmptyListFunctor`. I think we can probably have a bunch of instances smashed into one like we do [here](https://github.com/typelevel/cats/blob/5a38209b2a13ded2652528a8767b1683e74d339d/core/src/main/scala/cats/std/option.scala#L9) for `List` (only in this case we will also have `Comonad`, `Reducible`, etc.

yilinwei: @ceedubs wdyt of delegating applicable methods to the `List` instance and calling `NonEmptyList(list.head, list.tail)`?

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -0,0 +1,161 @@
...
+
+  implicit def catsDataEqForNonEmptyList[A](implicit A: Eq[A]): Eq[NonEmptyList[A]] =
+    new Eq[NonEmptyList[A]] {
+      def eqv(x: NonEmptyList[A], y: NonEmptyList[A]): Boolean = x === y
+    }
+
+  implicit def catsDataShowForNonEmptyList[A](implicit A: Show[A]): Show[NonEmptyList[A]] =
+    Show.show[NonEmptyList[A]](_.show)
+
+  implicit def catsDataSemigroupKForNonEmptyList[A]: SemigroupK[NonEmptyList[?]] =
```
xuwei-k: unnecessary type parameter

```scala
implicit val catsDataSemigroupKForNonEmptyList: SemigroupK[NonEmptyList] =
```

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -0,0 +1,161 @@
...
+
+  implicit def catsDataSemigroupForNonEmptyList[A]: Semigroup[NonEmptyList[A]] =
+    catsDataSemigroupKForNonEmptyList[F].algebra
+
+  implicit def catsDataReducibleForNonEmptyList[A]: Reducible[NonEmptyList[?]] =
+    new NonEmptyReducible[NonEmptyList[?]] {
+      override def split[A](fa: NonEmptyList[A]): (A, F[A]) = (fa.head, fa.tail)
+    }
+
+  implicit def catsDataMonadForNonEmptyList[A]: Monad[NonEmptyList[?]] =
```
xuwei-k: unnecessary type parameter

```scala
implicit val catsDataMonadForNonEmptyList: Monad[NonEmptyList] =
```

WarFox:
Thanks guys. I've been busy this week with work and other things. Please allow me time to update this pull requests with these review comments, and please add more comments. I can't promise a time frame now.

ceedubs:
Thanks for all of your work on this, @WarFox. I understand that you are busy. I'm hoping to get the `NonEmptyList` rework in before too long. Would you mind if I branch off of the commits that you have so far and then finish up the pending work?

WarFox:
@ceedubs Hi Cody, I am sorry about this. I couldn't work on this due to various reasons. It is great of you to pickup the work. Appreciate it, thank you.

ceedubs:
Hi @WarFox, it's no problem at all. Thanks for getting it off to a great start!

I'm going to close out this PR in favor of #1231 (which has these commits plus some more).

PR #1122, opened 2016-06-14
Yet some more uniquifying of implicit names

ceedubs:
I think this is the last of the implicit names for #1061.

@kailuowang would you be willing to review this?

### DIFF
--- a/kernel-laws/src/main/scala/cats/kernel/laws/package.scala
+++ b/kernel-laws/src/main/scala/cats/kernel/laws/package.scala
@@ -1,17 +1,10 @@
 package cats.kernel
 
-import org.typelevel.discipline.Predicate
-
 import org.scalacheck._
 import org.scalacheck.util.Pretty
 import Prop.{False, Proof, Result}
 
 package object laws {
-
-  implicit def nonEmptyPredicate[A](implicit ev: Eq[A], A: Monoid[A]): Predicate[A] =
-    new Predicate[A] {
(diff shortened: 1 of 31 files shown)


kernel-laws/src/main/scala/cats/kernel/laws/package.scala
```diff
@@ -1,18 +1,11 @@
...
 import Prop.{False, Proof, Result}
 
 package object laws {
 
-  implicit def nonEmptyPredicate[A](implicit ev: Eq[A], A: Monoid[A]): Predicate[A] =
-    new Predicate[A] {
-      def apply(a: A) = ev.neqv(a, A.empty)
-    }
-
   lazy val proved = Prop(Result(status = Proof))
```
kailuowang: Just curious what makes this one redundant?

ceedubs: Oh good catch I meant to ask people about this.

@non @johnynek: this seemed a little suspicious to me. I can understand the parameters being implicit, but why was the `def` itself implicit? It wasn't actually used in cats and didn't appear to be in algebra, so I removed it. Is this something that should be kept around?

non: Let's remove it. I'm not sure what I was thinking.

non:
👍

kailuowang:
👍

PR #1123, opened 2016-06-14
Include kernel in ScalaDoc

ceedubs:
Currently the [ScalaDoc](http://typelevel.org/cats/api/#package)
doesn't include the kernel module, so types like Semigroup aren't
present.


non:
👍 good catch

PR #1127, opened 2016-06-14
Add Xor.toTry

johnynek:
Add a method to convert an `Xor` to a `Try`, since `Try` is very commonly used in existing libraries.


non:
👍 This seems like something that people who need to use `Try` will really appreciate.

ceedubs:
👍

PR #1128, opened 2016-06-14
Add ApplySemigroup and ApplicativeMonoid

johnynek:
these are always lawful. The generalization to `Group` may not be (see Try, Failure has no inverse) or `Semilattice` (`Apply.map2(a, a)(_._1)` might not be idempotent, but if it is, we can add Band and Semilattice.

### DIFF
--- a/core/src/main/scala/cats/Applicative.scala
+++ b/core/src/main/scala/cats/Applicative.scala
@@ -50,3 +50,7 @@ import simulacrum.typeclass
       val G = Applicative[G]
     }
 }
+
+abstract class ApplicativeMonoid[F[_], A](f: Applicative[F], monoid: Monoid[A]) extends ApplySemigroup(f, monoid) with Monoid[F[A]] {
+  def empty: F[A] = f.pure(monoid.empty)
+}
--- a/core/src/main/scala/cats/Applicative.scala
+++ b/core/src/main/scala/cats/Applicative.scala
@@ -51,6 +51,11 @@ import simulacrum.typeclass
     }
 }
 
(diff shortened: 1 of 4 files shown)


non:
This looks great!

We should run the law-checker to verify these. I can try to dig up a snippet for you but it should be pretty easy to get it working.

johnynek:
@non I *think* we are testing try at least, so I think they are tested that way.

ceedubs:
👍 looks good to me, and the project code coverage has gone up, so they do appear to be tested.

@johnynek these are public classes so they will become a part of the cats API. That seems reasonable to me as other people could potentially benefit from these, but I just wanted to confirm that we are okay with that and the compatibility constraints that it adds.

johnynek:
I think we want them public so they can be reused indeed.

johnynek:
actually, I'm not sure they need to be abstract at all in that case.

That said, since they are so small, maybe we should hide them. Most users can easily reimplement. Instead we could put:
```scala
object Apply {
  def semigroup[F[_]: Apply, A: Semigroup]: Semigroup[F[A]] = ...
}

object Applicative {
  def monoid[F[_]: Apply, A: Monoid]: Monoid[F[A]] = ...
}
```
We may or may not make those implicit for someone to import.

Then the only people that miss out are those that know they have some other special structure (like they know they can make a Group or a Semilattice), but again, it is so little code to share, maybe no big deal.

Preferences?

non:
I'd lean toward private. As you say, reimplementation is not hard.

non:
Oh, right, I missed the fact that the `Try` instances were using them. 👍 from me (modulo making things private).

PR #1130, opened 2016-06-15
Add monoid syntax

### LINKED ISSUE #1124: Add Monoid syntax
Currently we have [Semigroup syntax](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/syntax/semigroup.scala) but lack something similar for `Monoid`. This means that you can do something like `1 |+| 2`, you can't do something like `1.isEmpty` (even though `Monoid[Int].isEmpty(1)` works).

hamishdickson:
This resolves #1124 

There are a couple of questions

- in `Group` and `SemiGroup` the comment `// TODO: use simulacrum instances eventually` has been made. I have copied this so all 3 can be changed at the same time. Is that the right thing to do here? I'm happy to use simulacrum if not

- in implementing `isEmpty` I originally attempted to use `Ops.binOp` but couldn't get it to work with my implicit `Eq`. Should I be using that here?

Sorry they are such basic questions, this is my first attempt at contributing to Cats and I'm trying to find my feet!

### DIFF
--- a/core/src/main/scala/cats/syntax/all.scala
+++ b/core/src/main/scala/cats/syntax/all.scala
@@ -23,6 +23,7 @@ trait AllSyntax
     with ListSyntax
     with MonadCombineSyntax
     with MonadFilterSyntax
+    with MonoidSyntax
     with OptionSyntax
     with OrderSyntax
     with PartialOrderSyntax
--- a/core/src/main/scala/cats/syntax/monoid.scala
+++ b/core/src/main/scala/cats/syntax/monoid.scala
@@ -1,0 +1,12 @@
+package cats
+package syntax
+
+trait MonoidSyntax extends SemigroupSyntax {
+  // TODO: use simulacrum instances eventually
+  implicit def catsSyntaxMonoid[A: Monoid](a: A): MonoidOps[A] =
+    new MonoidOps[A](a)
+}
+
+final class MonoidOps[A: Monoid](lhs: A) {
+  def isEmpty(implicit eq: Eq[A]): Boolean = Monoid[A].isEmpty(lhs)(eq)
+}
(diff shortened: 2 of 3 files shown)


ceedubs:
@hamishdickson thanks for working on this!

> in Group and SemiGroup the comment // TODO: use simulacrum instances eventually has been made. I have copied this so all 3 can be changed at the same time. Is that the right thing to do here? I'm happy to use simulacrum if not

I think that you can remove the comments. We've decided not to use simulacrum within cats-kernel for the time being.

> in implementing isEmpty I originally attempted to use Ops.binOp but couldn't get it to work with my implicit Eq. Should I be using that here?

I don't understand machinist very well. I thought that you'd be able to do something like `def isEmpty(implicit ev: Eq[A]): Boolean = macro Ops.unopWithEv[Eq[A], A]`, and that compiles, but then when you try to use it elsewhere, you get the error `Cannot extract subject of operator (tree = cats.implicits.catsSyntaxMonoid[Int](1)(cats.implicits.intGroup))`. Paging @non!

ceedubs:
PS @hamishdickson I had forgotten about the usage of machinist, and this issue wasn't as much "low-hanging-fruit" as I was thinking it was. Sorry about that!

hamishdickson:
@ceedubs my pleasure and no need to say sorry! I found it really interesting trying to get machinist to work (I also tried `unopWithEv`) but didn't want to spend too long on it unless it was obviously the wrong thing to do

If @non doesn't know off the top of his head, I will have a deeper look into machinist 😄 

Another noob question (sorry) - should I be updating any docs with this change?

ceedubs:
> Another noob question (sorry) - should I be updating any docs with this change?

There's definitely no need to apologize. If you feel inclined, you could add an example of using the `isEmpty` syntax on the [monoid tutorial](https://github.com/typelevel/cats/blob/master/docs/src/main/tut/monoid.md).

non:
@ceedubs so there are two evidence parameters here (`Monoid[A]` and `Eq[A]`) so in this case I think you'd need to use `unopWithEv2`. The documentation for Machinist needs to be a lot better before I can expect someone else to figure that one out!

PR #1131, opened 2016-06-15
Removed nesting in project structure listing

mgttlinger:
Is the project structure listing necessary at all given that nearly identical info is provided at the beginning?


ceedubs:
👍

> Is the project structure listing necessary at all given that nearly identical info is provided at the beginning?

I think that this is something that @non has cleaned up on a doc branch that he's going to submit a PR for soon.

kailuowang:
:+1:

Issue #872, opened 2016-02-10
add Unfoldable typeclass


julien-truffaut:
see https://github.com/purescript/purescript-unfoldable/blob/master/docs/Data/Unfoldable.md

adelbertc:
What are the laws? Can't seem to find it in the repo

julien-truffaut:
I guess we can express the first "law" if we assume `F` to have a `Foldable` instance:

```
forAll{b: B =>
  unfoldr(_ => None)(b).isEmpty 
}
```

for the second one, we would need some notion of `Cons` for `F`

PR #1132, opened 2016-06-15
POC of unfoldable

julien-truffaut:
Implement Unfoldable as of https://github.com/purescript/purescript-unfoldable https://github.com/typelevel/cats/issues/872

I don't know if it is a good idea but if `Unfoldable` extends `Foldable` then we can lift all methods from `List[A] => List[A]` to `F[A] => F[A]` e.g. `prepend`, `append`, `filter`, `concat`

This PR is just a proof of concept, I mainly wish to get some feedbacks.


non:
Interesting!

I like the idea of something like this. I'm a bit worried that the laws are too strict. For example, `Foldable` doesn't require a specific ordering (so we can say that sets and similar are foldable). In principle it should be fine to construct a set through *unfold* but this would currently violate the laws (due to the order of the prepends not being maintained).

Is this worth it? `Foldable` is somewhat lawless, and in some ways I'd be happy with an `Unfoldable` that was similar.

ceedubs:
@julien-truffaut neat! Do you have a specific use-case that prompted you to submit this? I'm always interested in how people are using things.

@non one possibility would be to start with these strict laws and later relax them if people find themselves wanting instances for types like `Set`. It's always easier to relax constraints than it is to add them.

julien-truffaut:
@ceedubs not really, I wanted to implement `Unfoldable` as in purescript but as @adelbertc and @non mentioned it is lawless. Then I wonder what could we do if we had both `fold` and `unfold` together and it seems that it is enough to implement all `Seq` / `List` functions

julien-truffaut:
What do you prefer:
- a lawless `Unfoldable` with no dependency on `Foldable`
- a stricter `Unfoldable` that guarantee the order is preserved with `Foldable`
- something else?

adelbertc:
I recall `Foldable` doesn't really have laws and that the laws we have defined are a bit sketchy - is this a similar case? If so, I may be persuaded to include a standalone `Unfoldable`. Otherwise I prefer the second, to have it be consistent with `Foldable`.

julien-truffaut:
Sorry for the inactivity on this PR. I am not sure how to go from this, I made `Unfoldable` lawless following the argument that we couldn't implement a `Set` instance if we maintain the order constraint between `Foldable` and `Unfoldable`. However, it seems wrong to have these two typeclasses without mutual laws as their methods are related.

All ideas are welcome! We can also simply close this PR until we have a clearer view, I mainly opened it to start a discussion.

etorreborre:
I am not sure about the laws of `Unfoldable` but today at work, someone asked me how to get all the consecutive values out of a `State` object and I thought I could use `unfold` for this.

adelbertc:
`Unfoldable` looks interesting to me for similar reasons as @etorreborre pointed out.

That being said, if we go with `Unfoldable[F] extends Foldable[F]` we run into issues described here: https://github.com/typelevel/cats/pull/1379#issuecomment-247878936 with

```scala
def foo[F[_]: Unfoldable, Traverse] = ... // ambiguous Foldable[F]
```

laws/src/main/scala/cats/laws/discipline/UnfoldableTests.scala
```diff
@@ -0,0 +1,28 @@
...
+import org.typelevel.discipline.Laws
+
+trait UnfoldableTests[F[_]] extends Laws {
+  def laws: UnfoldableLaws[F]
+
+  def unfoldable[A: Arbitrary](implicit EqFa: Eq[F[A]]): RuleSet =
+    new DefaultRuleSet(
+      name = "unfoldable",
+      parent = None,
+//        "noneConsistentWithDefault"         -> forAll(() => laws.noneConsistentWithDefault[A]),
```
johnynek: what's up here?

core/src/main/scala/cats/Unfoldable.scala
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+    * Build an F[A] from a `seed` and a generating function `f``
+    * The generating function `f` behave as follows:
+    * - If `f(b)` is `None`, then `unfoldLeft(b)(f)` should be empty.
+    * - If `f(b)` is `Some((b1, a))`, then `unfoldLeft(b)(f)` should consist of `a` appended by `unfoldLeft(b1)(f)`
+    */
+  def unfoldLeft[A, B](seed: B)(f: B => Option[(B, A)]): F[A]
+
+  def none[A]: F[A] =
```
johnynek: `def empty` seems better to me. Do you foresee that conflicting with `MonoidK`?

core/src/main/scala/cats/Unfoldable.scala
```diff
@@ -0,0 +1,59 @@
...
+    Unfoldable.DefaultImpl.none(this)
+
+  def singleton[A](value: A): F[A] =
```
johnynek: I think this will wind up being `pure` for `Alternative` instances.

core/src/main/scala/cats/Unfoldable.scala
```diff
@@ -0,0 +1,59 @@
...
+
+@typeclass trait Unfoldable[F[_]] { self =>
+
+  /**
+    * Build an F[A] from a `seed` and a generating function `f``
+    * The generating function `f` behave as follows:
+    * - If `f(b)` is `None`, then `unfoldLeft(b)(f)` should be empty.
+    * - If `f(b)` is `Some((b1, a))`, then `unfoldLeft(b)(f)` should consist of `a` appended by `unfoldLeft(b1)(f)`
+    */
+  def unfoldLeft[A, B](seed: B)(f: B => Option[(B, A)]): F[A]
```
johnynek: seems like `F[A]` may be a functor. Certainly if I have `B, B => Option[(B, A)]` to get `F[A]` and if I have `A => C` then I can make `F[C]`.

So, if for all `F[A]` there exists `B, B => Option[(B, A)]` to unfold to make `F[A]`, then `F[A]` is a functor.

core/src/main/scala/cats/Unfoldable.scala
```diff
@@ -0,0 +1,59 @@
+package cats
+
+import simulacrum.typeclass
+
+@typeclass trait Unfoldable[F[_]] { self =>
+
+  /**
+    * Build an F[A] from a `seed` and a generating function `f``
```
johnynek: it would be nice to have some stronger laws, even that might relate `Unfoldable` to `Foldable` or `Functor`.

Such as, if `F[_]: Unfoldable` and `F[_]: Functor` then,
```scala
unfoldLeft(seed)(fn).map(g) == unfoldLeft(seed)(fn.andThen { case (b, a) => (b, g(a)) })`
```

julien-truffaut: 👍

core/src/main/scala/cats/Unfoldable.scala
julien-truffaut: yes, I would prefer to have unique function name for all cats typeclasses

core/src/main/scala/cats/Unfoldable.scala
julien-truffaut: if `F` is both a `Foldable` and `Unfoldable` then `F` is a `Functor`

core/src/main/scala/cats/Unfoldable.scala
```diff
@@ -0,0 +1,59 @@
...
+
+    def fromFoldable[F[_], G[_], A](ga: G[A])(implicit F: Unfoldable[F], G: Foldable[G]): F[A] =
+      fromList(G.toList(ga))
+
+    def fromList[F[_], A](ga: List[A])(implicit F: Unfoldable[F]): F[A] = {
+      def go(l: List[A]): Option[(List[A], A)] = l match {
+        case Nil     => None
+        case x :: xs => Some((xs, x))
+      }
+      F.unfoldLeft(ga.reverse)(go)
```
non: Is there a reason we need to do the `.reverse` here? I don't see any laws about maintaining order -- is this just so that `Unfoldable[List].fromList(xs) <-> xs`?

adamgfraser:
This looks really cool. One thought that struck me is that right now I think this can only cover linear structures like List or Stream as opposed to structures like Tree with another branching factor. Also it can't really support structures that are always nonempty because when unfold returns none to stop the unfolding there is no value to put in the last layer. Is it worth thinking about how to abstract over these?

Maybe something like S => (A, F[S]) for nonempty structures (like the way unfoldTree is implemented in Scalaz) and S => Option[(A, F[S])] for structures that could be empty? I think there is a connection to Free and Cofree here as well. This may be out of scope though.

core/src/main/scala/cats/Unfoldable.scala
julien-truffaut: you're right, I think it is an artifact from the impl where I made `Unfoldable extends Foldable`. I guess we could express an order law if `F` is also `Foldable`

julien-truffaut:
@adamgfraser really interesting idea, it would be great if `Unfoldable` could also generate `Tree`. This PR is really a proof of concept, I am not sure it is worth to merge as it is right now. Please feel free to contribute to this branch or create your own version!

core/src/main/scala/cats/Unfoldable.scala
julienrf: In some cases, it could be useful to keep the `B` value at the end of the process, what do you think?

LukaJCB:
I think we should revive this, especially given that we now require our `Foldable`s to be orderered. Not sure if this PR can be salvaged as it's pretty out of date, but the main type class looks pretty good to me :)

LukaJCB:
So I think making `Unfoldable` lawless isn't a great idea. 
If we have `Unfoldable` extend `Foldable` we can add some nice laws, as well as get a `Functor` instance for free, at that point we can also get a `Traverse` instance, so maybe we want `Unfoldable extends Traverse`? That would also solve the problem adelbert mentioned about ambiguous `Foldable[F]`.

kailuowang:

PR #1134, opened 2016-06-16
Add list syntax object

ceedubs:
Currently `import cats.syntax._` brings in the list syntax but there is
no object to import only the list syntax.


non:
👍

adelbertc:
👍

PR #1136, opened 2016-06-16
Symbols Chart

zainab-ali:
Adding a symbols chart to the docs.
This is a copy of PR #1094 with commits merged.

### DIFF
--- a/docs/src/main/tut/symbols.md
+++ b/docs/src/main/tut/symbols.md
@@ -1,0 +1,31 @@
+#Symbols
+
+Below is a list of symbols used in cats.  
+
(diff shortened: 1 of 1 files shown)


ceedubs:
👍 thanks! Looks good to me. Do we need to add a link to this somewhere on the site, or is a link automatically being added?

kailuowang:
👍 thanks! LGTM. Just want to note here that either https://github.com/typelevel/cats/pull/1112 or this one might need to update the cartesian builder symbol depending on which PR gets merged first.

non:
Looks great! Thanks!

We'll have to work to keep this up-to-date but I think it will be very useful. 👍

Issue #1088, opened 2016-06-03
create cats.data.NonEmptyVector


ceedubs:
We seem to have arrived on a decision in #537 that NonEmptyVector should have its own data type. Let's go for it!

Since representing a `Vector` as `case class NonEmptyVector[A](head: A, tail: Vector[A])` is not very efficient for quite a few operations, @non has suggested that we create a class that wraps a `Vector` that it knows to be non-empty. We can guarantee this by making the companion object method that creates  an instance of the class from a `Vector` return an `Option`.

zainab-ali:
I'm working on this

ceedubs:
Resolved by #1137

PR #1137, opened 2016-06-17
Adding NonEmptyVector

zainab-ali:
Adding `NonEmptyVector` as discussed in #1088


ceedubs:
@zainab-ali thanks for working on this!

It looks like there are already merge conflicts, so you'll want to pull in the latest changes from master.

Also please note the second paragraph of the description for #1088 -- I think we'll want to change this implementation to not use the "cons" approach for `Vector`.

core/src/main/scala/cats/data/package.scala
```diff
@@ -11,11 +10,6 @@ package object data {
   def NonEmptyList[A](head: A, tail: A*): NonEmptyList[A] =
     OneAnd[List, A](head, tail.toList)
 
-  def NonEmptyVector[A](head: A, tail: Vector[A] = Vector.empty): NonEmptyVector[A] =
-    OneAnd(head, tail)
-  def NonEmptyVector[A](head: A, tail: A*): NonEmptyVector[A] =
```
kailuowang: IMO, this API would be handy.

docs/src/main/tut/oneand.md
```diff
@@ -21,10 +21,4 @@ type NonEmptyList[A] = OneAnd[List, A]
 
 which is the actual implementation of non-empty lists in cats.  By
 having the higher kinded type parameter `F[_]`, `OneAnd` is also able
-to represent other "non-empty" data structures, e.g.
```
kailuowang: I don't feel strong about this but shall we replace this example with `NonEmptyStream`, instead of just removing it.

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -0,0 +1,179 @@
...
+  /**
+   * Typesafe stringification method.
+   *
+   * This method is similar to .toString except that it stringifies
+   * values according to Show[_] instances, rather than using the
+   * universal .toString method.
+   */
+  def show(implicit A: Show[A]): String =
+    s"NonEmptyVector(${A.show(head)}, ${Show[Vector[A]].show(tail)})"
+}
```
zainab-ali: Is this show correct?  
Or should I have `s"NonEmptyVector(${Show[Vector[A]].show(vector)})"` instead?

kailuowang: I think it's better than the `cons` like show.

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -0,0 +1,179 @@
...
+        fa.foldLeft(b)(f)
+
+      def foldRight[A, B](fa: NonEmptyVector[A], lb: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B] =
+        fa.foldRight(lb)(f)
+    }
+}
+
+object NonEmptyVector extends NonEmptyVectorInstances {
+
+  def apply[A](head: A, tail: Vector[A]): NonEmptyVector[A] = NonEmptyVector(head +: tail)
```
non: I would prefer to support this as:

```scala
def apply[A](body: Vector[A], last: A): NonEmptyVector[A] = NonEmptyVector(body :+ last)
```

In general I don't like treating non-empty vector as a head/tail pair because it's so inefficient -- prepending to a vector is O(n) and reallocates the entire vector.

Honestly I'd probably prefer an interface like:

```scala
def apply[A](xs: Vector[A]): Option[NonEmptyVector[A]] = ...
```

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -0,0 +1,179 @@
...
+
+  /**
+   * remove elements not matching the predicate
+   */
+  def filter(f: A => Boolean): Vector[A] = vector.filter(f)
+
+  /**
+   * Append another NonEmptyVector to this
+   */
+  def combine(other: NonEmptyVector[A]): NonEmptyVector[A] = NonEmptyVector(vector ++ other.vector)
```
non: We probably also want a variant of this that works with `Vector[A]`, since appending a vector to a non-empty vector (or vice-versa) always produces a non-empty vector.

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -0,0 +1,179 @@
...
+package data
+
+import scala.annotation.tailrec
+import scala.collection.immutable.VectorBuilder
+import cats.std.vector._
+
+/**
+ * A data type which represents a single element (head) and a Vector
+ * (tail).  This can be used to represent a Vector which is guaranteed
+ * to not be empty.
```
non: This description should be changed, since we're just wrapping a vector.

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -0,0 +1,179 @@
...
+ */
+final case class NonEmptyVector[A] private (vector: Vector[A]) {
+
+  def unwrap: Vector[A] = vector
+
+  def head: A = vector.head
+
+  def tail: Vector[A] = vector.tail
+
```
non: I feel like it's worth supporting `apply` and `updated` for this structure, since indexed access is one of the best reasons to want to be using a vector. I'd probably use the standard signatures but I could be persuaded to have them return `Option[A]` and `Option[Unit]` respectively.

adelbertc: Or both? :-) But agree on wanting indexed access.

core/src/main/scala/cats/data/NonEmptyVector.scala
kailuowang: I probably missed something. I have been holding the belief that according to 
http://docs.scala-lang.org/overviews/collections/performance-characteristics.html
Prepending to a vector is as efficient as appending? I thought it makes sense based on my (not very thorough) understanding of the implementation of vector. Is there some caveat in this claim?

core/src/main/scala/cats/data/NonEmptyVector.scala
ceedubs: @non you mean `Option[NonEmptyVector[A]]` for `updated` right? :)

I'd be in favor of `updated` that returned an `Option` and `updatedUnsafe` that had the behavior of the the std lib (throwing an exception if the index doesn't exists). Similarly for accessing an index, but the name may be a little awkward there if we use `apply`. Maybe `atIndex`/`atIndexUnsafe` or `at` or `elem` or something?

core/src/main/scala/cats/data/NonEmptyVector.scala
ceedubs: And even though this will be called `combine` in the `Semigroup` instance, I'd be in favor of calling it `concat` (maybe with a `++` alias) since it's a little more descriptive in this context.

core/src/main/scala/cats/data/NonEmptyVector.scala
ceedubs: I might propose the following instantiation methods:

```scala
// to be consistent with var-arg `apply` methods for other collections
def apply[A](head: A, tail: A*): NonEmptyVector[A]

def fromVector[A](vector: Vector[A]): Option[NonEmptyVector[A]]

// this will throw an exception if the vector is empty
def fromVectorUnsafe[A](vector: Vector[A]): NonEmptyVector[A]

def prepend[A](head: A, tail: Vector[A]): NonEmptyVector[A]

def append[A](vector: Vector[A], last: A): NonEmptyVector[A]
```

I think that would cover all of the use-cases. WDYT?

docs/src/main/tut/oneand.md
ceedubs: I think that after this PR (or possibly as a part of this PR) we can probably rip out `OneAnd` as discussed [here](https://github.com/typelevel/cats/issues/537#issuecomment-222667534). Someone can follow up with a `NonEmptyStream` if it's something they really want -- I suspect it's not used as commonly as `NonEmptyList` and `NonEmptyVector`.

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -0,0 +1,179 @@
...
+import cats.std.vector._
+
+/**
+ * A data type which represents a single element (head) and a Vector
+ * (tail).  This can be used to represent a Vector which is guaranteed
+ * to not be empty.
+ */
+final case class NonEmptyVector[A] private (vector: Vector[A]) {
+
+  def unwrap: Vector[A] = vector
```
ceedubs: Since people can just call `vector`, I don't know if this alias is very helpful. We can probably remove it.

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -0,0 +1,179 @@
...
+   * Left-associative fold using f.
+   */
+  def foldLeft[B](b: B)(f: (B, A) => B): B =
+    vector.foldLeft(b)(f)
+
+  /**
+   * Right-associative fold using f.
+   */
+  def foldRight[B](lb: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B] =
+    Eval.defer(Foldable[Vector].foldRight(vector, lb)(f))
```
ceedubs: Is there a reason this `Eval.defer` is needed here? If so it's probably worth leaving a comment describing why.

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -0,0 +1,179 @@
...
+  implicit def catsDataSemigroupKForNonEmptyVector: SemigroupK[NonEmptyVector] =
+    new SemigroupK[NonEmptyVector] {
+      def combineK[A](a: NonEmptyVector[A], b: NonEmptyVector[A]): NonEmptyVector[A] =
+        a combine b
+    }
+
+  implicit def catsDataSemigroupForNonEmptyVector[A]: Semigroup[NonEmptyVector[A]] =
+    catsDataSemigroupKForNonEmptyVector.algebra
+
+  implicit def catsDataReducibleForNonEmptyVector: Reducible[NonEmptyVector] =
```
ceedubs: I think we should be able to combine a bunch of these instances into 1 like we do with the instances for `List` [here](https://github.com/typelevel/cats/blob/5a38209b2a13ded2652528a8767b1683e74d339d/core/src/main/scala/cats/std/list.scala#L13). This avoids the need to mess with the priority hierarchy (or at least for the most part).

ceedubs:
@zainab-ali thanks, this is getting really close I think!

core/src/main/scala/cats/data/NonEmptyVector.scala
ceedubs: Also when you do this you should be able to change it from a `def` to a `val`.

core/src/main/scala/cats/data/NonEmptyVector.scala
non: You're right @kailuowang. I forgot that vectors maintain a start index to allow this kind of thing to be *effectively constant time*. I was wrong.

I think maybe I got confused because for awhile `Vector(x) ++ xs` was very slow. But this was just a bug, not related to `x +: xs`.

core/src/main/scala/cats/data/NonEmptyVector.scala
non: Yes! Sorry, in my defense it was late! ;)

core/src/main/scala/cats/data/NonEmptyVector.scala
non: We could call the member `.toVector` so that `nev.toVector` would do the right thing.

core/src/main/scala/cats/data/NonEmptyVector.scala
non: It looks to me like `Eval.defer` could be removed.

core/src/main/scala/cats/data/NonEmptyVector.scala
johnynek: +1 to `.toVector`

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -0,0 +1,179 @@
...
+    */
+  def map[B](f: A => B): NonEmptyVector[B] =
+    NonEmptyVector(vector.map(f))
+
+  /**
+    *  Applies f to all elements and combines the result
+    */
+  def flatMap[B](f: A => NonEmptyVector[B]): NonEmptyVector[B] =
+    NonEmptyVector(vector.flatMap(a => f(a).vector))
+
```
johnynek: what about `reduce`? Since, we know that this is non-empty the `reduce(fn: (A, A) => A): A` is a safe signature. That or we could have: `def combineAll(implicit: sg: Semigroup[T]): T` (or we could have both).

core/src/main/scala/cats/data/NonEmptyVector.scala
johnynek: why not have:
```scala
def get(i: Int): Option[A]
// This is unsafe, it may throw on out-of-bounds, see `get` for a safe variant.
def apply(i: Int): A
```

for updated I would like:
```scala
def updatedOption(i: Int, a: A): Option[NonEmptyVector[A]]
// this will throw is i is out of bounds.
def updated(i: Int, a: A): NonEmptyVector[A]
```
I feel like adding `Option` as a suffix signals (like `headOption`, `reduceOption` is pretty standard), and I think the unsafe methods should keep the same name (lest we train people that this method is safe in some contexts but not others. Reading the code you have to know exactly if it is NonEmptyVector or Vector to see if `apply` or `updated` is safe, I'd rather always make them unsafe so we are consistent).

core/src/main/scala/cats/data/NonEmptyVector.scala
johnynek: +1 for `++` we should match Vector when possible.

core/src/main/scala/cats/data/NonEmptyVector.scala
johnynek: alternatively, since `toVector` is free, we could have only `get` and require `ne.toVector(12)` for unsafe apply.

ceedubs: This is tough. I think that `updated` and `apply` have names that seem too benign for throwing exceptions in common cases. It's nice to be consistent with the standard library, but I also have seen lots of scala devs mess up with `head` and `reduce` when I don't think they would have if these had been the safe options and separate `headUnsafe` and `reduceUnsafe` options had existed. I'm hesitant to pull in gotchas just to have consistent naming with the standard library. Hmmm.

core/src/main/scala/cats/data/NonEmptyVector.scala
ceedubs: `reduce` and several other methods are available via the `Reducible` instance, but I agree that it would probably be nice to have some of these helpers on the class itself.

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -0,0 +1,179 @@
...
+      def foldRight[A, B](fa: NonEmptyVector[A], lb: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B] =
+        fa.foldRight(lb)(f)
+    }
+}
+
+object NonEmptyVector extends NonEmptyVectorInstances {
+
+  def apply[A](head: A, tail: Vector[A]): NonEmptyVector[A] = NonEmptyVector(head +: tail)
+
+  def apply[A](head: A, tail: A*): NonEmptyVector[A] = NonEmptyVector(head +: tail.toVector)
```
ceedubs: I thought the same :). I admit to having no idea how big of a difference it would make. I think that there are probably some optimizations that could be made to the `Reducible` instance as well, but I think we should maybe wait for a follow-up PR for optimizations.

core/src/main/scala/cats/data/NonEmptyVector.scala
johnynek: I can see your point on safety, but there is also a cost (especially to reviewers of future code) of using names that are unsafe on Vector to be safe here (if they change the type signature).

My bias is using `cats` in a large codebase with many novices that mostly work with the standard library. I'd rather avoid confusion in naming (not at the cost of safety though). So, I would say `updateOption` and maybe `updateUnsafe` would be the best solution.

Maybe I would say: when we can safely follow standard naming conventions, we should. Otherwise, if we want to add methods that can throw, use `Unsafe` as a suffix (or prefix).

`head` has the same type signature, so I am not worried about it. Changing `updated` to be optional seems like it will create reading difficulties, so I prefer a new name for a safe variant.

ceedubs: @johnynek everything that you said sounds good to me. So it sounds like it'll be `updateOption` and `updateUnsafe` for the update methods. And maybe `get` and `getUnsafe` for index access? Or it could be `getOption` to be symmetric if you'd prefer. Either one is fine with me.

core/src/main/scala/cats/data/NonEmptyVector.scala
zainab-ali: I think they should be called `reduceLeft(f: (A, A) => A): A` and `reduce(implicit S: Semigroup[A]): A` to be consistent with `cats.Reducible`.  AFAK, the standard library has similar names.

zainab-ali:
I've been mulling over the naming conventions for safe / unsafe methods.
 
I agree with the following conventions:
 1. If a function is unsafe, it should be suffixed by `unsafe`.  This is because `cats` is safe by default. 
 2. We should stick to the standard library conventions whenever they are safe.
 
So when a function is safe, but is analogous to an unsafe standard library function, what should we do? 
e.g `updated(i: Int): Option[A]`, `apply(i: Int): Option[A]`, `Vector.apply(v: Vector[A]): Option[NonEmptyVector[A]]`

Since cats is safe by default, we could keep the names as is,  but as @johnynek points out, is causes confusion for users of the standard library, when functions they expect to be unsafe start returning options.

In the above comments, we've decided rename the functions instead:
`updated` to `updatedOption`
`apply(i: Int)` to `get`, and the unsafe variant to `getUnsafe`
`apply(v: Vector[A])` to `fromVector` and the unsafe variant to `fromVectorUnsafe`

TBH, I'm uneasy about this, because I find variation in method names more confusing than variation in signatures.  However, since I can't think of a better solution, I'll go with it.

We should use the same convention for the `NonEmptyList` in #1120

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -3,69 +3,99 @@ package data
...
+    * Left-associative reduce using f.
+    */
+  def reduceLeft(f: (A, A) => A): A =
+    tail.foldLeft(head)(f)
+
+  /**
+    * Left-associative reduce using the Semigroup of A
+    */
+  def reduce(implicit S: Semigroup[A]): A =
+    reduceLeft(S.combine)
```
johnynek: semigroup has `combineOption` which can possibly be faster than using `combine`.

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -85,37 +115,29 @@ final case class NonEmptyVector[A] private (vector: Vector[A]) {
...
-  implicit def catsDataEqForNonEmptyVector[A](implicit A: Eq[A]): Eq[NonEmptyVector[A]] =
-    new Eq[NonEmptyVector[A]]{
-      def eqv(x: NonEmptyVector[A], y: NonEmptyVector[A]): Boolean = x === y
-    }
-
-  implicit def catsDataShowForNonEmptyVector[A](implicit A: Show[A]): Show[NonEmptyVector[A]] =
-    Show.show[NonEmptyVector[A]](_.show)
+  implicit val catsDataInstancesForNonEmptyVector: SemigroupK[NonEmptyVector] with Reducible[NonEmptyVector]
+      with Monad[NonEmptyVector] with Comonad[NonEmptyVector] with Traverse[NonEmptyVector] =
+    new NonEmptyReducible[NonEmptyVector, Vector] with SemigroupK[NonEmptyVector] with Monad[NonEmptyVector]
```
johnynek: I think this can be a `MonadRec` using the same style as:
https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/instances/list.scala#L31

and also I am pretty sure `MonadCombine`

ceedubs: I don't think we can do `MonadCombine` because it has an `empty` method that you can't write for non-empty structures.

👍 for the `MonadRec` instance though.

core/src/main/scala/cats/data/NonEmptyVector.scala
zainab-ali: @ceedubs what optimisations did you have in mind?

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -0,0 +1,230 @@
...
+import cats.instances.vector._
+
+/**
+ * A data type which represents a non empty Vector.
+ */
+final case class NonEmptyVector[A] private (toVector: Vector[A]) {
+
+  /** Gets the element at the index, if it exists */
+  def get(i: Int): Option[A] =
+    Try(getUnsafe(i)).toOption
```
ceedubs: I think that `toVector.lift(i)` would probably be a more efficient implementation (though this is performance speculation without a benchmark).

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -0,0 +1,230 @@
...
+
+  /** Gets the element at the index, or throws an exception if none exists */
+  def getUnsafe(i: Int): A = toVector(i)
+
+  /** Updates the element at the index, if it exists */
+  def updated(i: Int, a: A): Option[NonEmptyVector[A]] =
+    Try(updatedUnsafe(i, a)).toOption
```
ceedubs: More performance speculation: `if (toVector.isDefinedAt(i)) Some(toVector.updated(i, a)) else None`.

ceedubs: Oops I guess that would need to wrap it in a `NonEmptyVector` too.

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -0,0 +1,230 @@
...
+  def show(implicit A: Show[A]): String =
+    s"NonEmptyVector(${Show[Vector[A]].show(toVector)})"
+}
+
+private[data] sealed trait NonEmptyVectorInstances {
+
+  implicit val catsDataInstancesForNonEmptyVector: SemigroupK[NonEmptyVector] with Reducible[NonEmptyVector]
+      with Monad[NonEmptyVector] with Comonad[NonEmptyVector] with Traverse[NonEmptyVector] with MonadRec[NonEmptyVector] =
+    new NonEmptyReducible[NonEmptyVector, Vector] with SemigroupK[NonEmptyVector] with Monad[NonEmptyVector]
+        with Comonad[NonEmptyVector] with Traverse[NonEmptyVector] with MonadRec[NonEmptyVector] {
```
ceedubs: Minor: since `MonadRec` extends `Monad`, you can replace the `with Monad[NonEmptyVector]` in the two places it shows up here.

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -0,0 +1,230 @@
...
+
+  def fromVector[A](vector: Vector[A]): Option[NonEmptyVector[A]] =
+    vector.headOption.map(apply(_, vector.tail))
+
+  def fromVectorUnsafe[A](vector: Vector[A]): NonEmptyVector[A] =
+    fromVector(vector) match {
+      case Some(v) => v
+      case None =>
+        throw new IllegalArgumentException("Cannot create NonEmptyVector from empty vector")
+    }
```
ceedubs: I'd probably just do an `isEmpty` check here to avoid the extra `Option` allocation.

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -0,0 +1,230 @@
...
+
+  def apply[A](head: A, tail: A*): NonEmptyVector[A] = {
+    val buf = Vector.newBuilder[A]
+    buf += head
+    tail.foreach(buf += _)
+    NonEmptyVector(buf.result)
+  }
+
+  def fromVector[A](vector: Vector[A]): Option[NonEmptyVector[A]] =
+    vector.headOption.map(apply(_, vector.tail))
```
ceedubs: More performance speculation: `if (vector.isEmpty) None else Some(new NonEmptyVector(vector))`.

tests/src/test/scala/cats/tests/NonEmptyVectorTests.scala
```diff
@@ -0,0 +1,185 @@
...
+  checkAll("NonEmptyVector[Int] with Option", TraverseTests[NonEmptyVector].traverse[Int, Int, Int, Int, Option, Option])
+  checkAll("Traverse[NonEmptyVector[A]]", SerializableTests.serializable(Traverse[NonEmptyVector]))
+
+  checkAll("NonEmptyVector[Int]", ReducibleTests[NonEmptyVector].reducible[Option, Int, Int])
+  checkAll("Reducible[NonEmptyVector]", SerializableTests.serializable(Reducible[NonEmptyVector]))
+
+  implicit val iso = CartesianTests.Isomorphisms.invariant[NonEmptyVector]
+
+  // Test instances that have more general constraints
+  {
```
ceedubs: I don't think we need to add braces for separate scopes for the following tests. That's just done for tests where there's an abstract `F[_]` and we need to test cases where there are different instances available for `F[_]`.

tests/src/test/scala/cats/tests/NonEmptyVectorTests.scala
```diff
@@ -0,0 +1,185 @@
...
+    checkAll("Monad[NonEmptyVector]", SerializableTests.serializable(Monad[NonEmptyVector]))
+  }
+
+  {
+    checkAll("NonEmptyVector[Int]", ComonadTests[NonEmptyVector].comonad[Int, Int, Int])
+    checkAll("Comonad[NonEmptyVector]", SerializableTests.serializable(Comonad[NonEmptyVector]))
+  }
+
+  {
+    checkAll("NonEmptyVector[Int]", MonadRecTests[NonEmptyVector].monadRec[Int, Int, Int])
```
ceedubs: `MonadRecTest` already runs the tests for `Monad`, `Functor`, and `Cartesian`, so you should be able to remove the test calls for the latter 3.

ceedubs:
👍 thanks @zainab-ali!

I've left some minor comments about possibly more performant implementations of a few things, but I'm happy to let those happen in a followup PR.

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -0,0 +1,230 @@
...
+    * Left-associative reduce using f.
+    */
+  def reduceLeft(f: (A, A) => A): A =
+    tail.foldLeft(head)(f)
+
+  /**
+    * Left-associative reduce using the Semigroup of A
+    */
+  def reduce(implicit S: Semigroup[A]): A =
+    S.combineAllOption(tail).foldLeft(head)(S.combine)
```
johnynek: personally I would do: `S.combineAllOption(toVector).get` since we know the get can't throw since `toVector.nonEmpty`.

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -0,0 +1,230 @@
...
+
```
johnynek: we could add `def zip[B](that: NonEmptyVector[B]): NonEmpty[(A, B)]` which could be nice.

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -0,0 +1,230 @@
...
+    NonEmptyVector(toVector.flatMap(a => f(a).toVector))
+
+  /**
+    * Left-associative reduce using f.
+    */
+  def reduceLeft(f: (A, A) => A): A =
+    tail.foldLeft(head)(f)
+
+  /**
+    * Left-associative reduce using the Semigroup of A
```
johnynek: I think we should remove this comment: "Left-associative" since `Semigroup` is associative, it does not matter which order you do the combine, the result is the same.

core/src/main/scala/cats/data/NonEmptyVector.scala
johnynek: +1. I think we should do the horrible, but fast stuff internally when we can prove it is correct (and then test it to be sure).

core/src/main/scala/cats/data/NonEmptyVector.scala
johnynek: +1

kailuowang:
👍

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -0,0 +1,226 @@
...
+
+  /**
+   * Typesafe stringification method.
+   *
+   * This method is similar to .toString except that it stringifies
+   * values according to Show[_] instances, rather than using the
+   * universal .toString method.
+   */
+  def show(implicit A: Show[A]): String =
+    s"NonEmptyVector(${Show[Vector[A]].show(toVector)})"
```
peterneyens: I think this will give us something like `NonEmptyVector(Vector(1, 2, 3))` for `NonEmptyVector(1, 2, 3)`.
Maybe we could do `"NonEmpty" + Show[Vector[A]].show(toVector)` which should give us `NonEmptyVector(1,2,3)` ?

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -0,0 +1,226 @@
...
+            case a +: as => consume(as, buf += f(NonEmptyVector(a, as)))
+            case _ => buf.result()
+          }
+        NonEmptyVector(f(fa), consume(fa.tail, new VectorBuilder[B]))
+      }
+
+      def extract[A](fa: NonEmptyVector[A]): A = fa.head
+
+      def traverse[G[_], A, B](fa: NonEmptyVector[A])(f: (A) => G[B])(implicit G: Applicative[G]): G[NonEmptyVector[B]] =
+      G.map2Eval(f(fa.head), Always(Traverse[Vector].traverse(fa.tail)(f)))(NonEmptyVector(_, _)).value
```
peterneyens: Slightly nitpicking but the 2 spaces for indentation seem to be missing.

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -0,0 +1,226 @@
...
+
+  /**
+   * Append another Vector to this
+   */
+  def concat(other: Vector[A]): NonEmptyVector[A] = NonEmptyVector(toVector ++ other)
+
+  /**
+    * Alias for concat
+    */
+  def ++(other: Vector[A]): NonEmptyVector[A] = concat(other)
```
peterneyens: Again some minor nitpicking about formatting, but both the `++` methods ScalaDoc's can be changed to 

```scala
/**
 * Alias for concat
 */
```

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -0,0 +1,226 @@
...
+    toVector.lift(i)
+
+  /** Gets the element at the index, or throws an exception if none exists */
+  def getUnsafe(i: Int): A = toVector(i)
+
+  /** Updates the element at the index, if it exists */
+  def updated(i: Int, a: A): Option[NonEmptyVector[A]] =
+    if (toVector.isDefinedAt(i)) Some(NonEmptyVector(toVector.updated(i, a))) else None
+
+  /** Updates the element at the index, or throws an exeption if none exists */
```
peterneyens: Small typo: exeption -> exception.

peterneyens:
This looks great @zainab-ali, I left some minor comments.

Another thing which might be useful is an `unapply` method on the `NonEmptyVector` object so you can pattern match on `NonEmptyVector`, but this can probably be added in a later PR.

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -0,0 +1,226 @@
...
+   */
+  def filter(f: A => Boolean): Vector[A] = toVector.filter(f)
+
+  /**
+   * Append another NonEmptyVector to this
+   */
+  def concat(other: NonEmptyVector[A]): NonEmptyVector[A] = NonEmptyVector(toVector ++ other.toVector)
+
+  /**
+    * Alias for concat
```
ceedubs: Would be nice to make this `Alias for [[concat]]` so that the scaladoc has a link.

ceedubs: Oh I see that we are using overloading, so I don't know how that works with scaladoc. I'm generally skeptical of overloading, though it's probably fine in this instance. I think it would prevent us from being able to turn `NonEmptyVector` into a value class though, wouldn't it?

non:
It looks we have have multiple thumbs-up. I'm going to give mine too 👍 and merge this. There is probably some follow-up work to do, but let's open a new PR for that.

Issue #1135, opened 2016-06-16
std in instance/syntax imports


ceedubs:
Currently if you want to import the type class _instances_ for `Option`, you `import cats.std.option._`. If you want to import the _syntax_ for `Option`, you `import cats.syntax.option._` (note the lack of `std`). This seems inconsistent and/or backwards to me. Since instances for Cats types are in their companion objects, I don't think I'll ever need to import instances for anything _other_ than `std` and thus the namespacing doesn't really seem to make much sense. However, when it comes to syntax, I may be adding syntax to std lib types (like `Option`) or cats types (like `Functor`).

I see several logical ways that we could do this:

1. Drop the `std` from the instance imports
2. Keep the `std` in the instance imports and add it to the syntax imports
3. Drop the `std` from the instance imports (since they are all std lib anyway) and add it to the syntax imports (since not all of them will actually be related to std lib)

I think that I prefer `2`. The thing that makes me uneasy about `1` is that it continues the current trend of not namespacing syntax for std lib types, so if the std lib were to gain a `Functor` or `Semigroup` in the future it could mess up our naming convention. I'm fine with `3`, but the lack of symmetry between the instance and syntax imports seems like it could make things less intuitive.

kailuowang:
I vote for 2 as well. I like the `std` package convention.

philwills:
I'd vote for 1. The `std` thing has always seemed a bit odd and unintuitive to me.

non:
What if we rename `std` to `instances`? To me `std` is the place where implicit instances go and `syntax` is the place where implicit syntax goes.

ceedubs:
@non I like that. I had thought of that but was thinking that we had some non-instance things in there like the `none[A]` syntax. It looks like that's in `syntax`, so as long as we don't think we'll have non-instance things in there, 👍 from me.

kailuowang:
👍  on renaming `std` to `instances`.

ceedubs:
Fixed in #1140.

PR #1140, opened 2016-06-18
Rename std package to instances

ceedubs:
For #1135.

So far I haven't renamed the `std` within `kernel`. I'm happy to make
those changes as well, but I wanted to check with @non and @johnynek.

I also didn't rename all of the `catsStdX` instances. Those names seemed
fine to me, and some of them have names like `catsStdInstancesForList`,
which doesn't translate well if you switch `Std` to `Instances`.


kailuowang:
:+1:

johnynek:
👍 please go ahead and change kernel. We should stabilize now before it is used too much.

ceedubs:
@johnynek @kailuowang I've updated the PR to include the change for kernel as well.

kailuowang:
:+1:

johnynek:
👍

non:
👍 (once the conflicts are resolved)

ceedubs:
Interesting - I didn't have to resolve any merge conflicts when I merged in master.

PR #1142, opened 2016-06-18
Add more scalastyle rules

ceedubs:
The intention of this is so that potential contributors see a consistent
style and so that they can check the scalastyle output and hopefully
avoid reviewers leaving nitpicky formatting comments on PRs.

I'd like to wait until #845 is merged before we merge this one so we don't potentially cause yet another merge conflict in that PR. I wanted to go ahead and submit this so that I don't forget to though.


peterneyens:
Maybe we could also add a rule to check we write type class instead of typeclass, so we don't need to clean this up regularly like in  #1149.

ceedubs:
@peterneyens that is an excellent idea. I'm very inconsistent on this point.

I think that there are a couple difficulties though.
- I don't think scalastyle runs on our docs, so it's still up to us to try to be consistent there.
- We'll probably have to make the pattern not match if there's a leading `@` to avoid our simulacrum annotations tripping this.

ceedubs:
@peterneyens updated to include "type class" rule.

peterneyens:
Maybe we could also remove the `NumberOfTypes` rule in this PR, like we mentioned in #1093 ?

ceedubs:
#845 has been merged and I've updated this PR. I think it's ready for review/merge.

kailuowang:
👍  LGTM

core/src/main/scala/cats/data/OneAnd.scala
```diff
@@ -27,7 +27,7 @@ final case class OneAnd[F[_], A](head: A, tail: F[A]) {
    */
   def filter(f: A => Boolean)(implicit F: MonadCombine[F]): F[A] = {
     val rest = F.filter(tail)(f)
-    if (f(head)) F.combineK(F.pure(head), rest) else rest
+    if(f(head)) F.combineK(F.pure(head), rest) else rest
```
fthomas: Later in this diff, spaces are inserted between `if` and `(`, for example [here](https://github.com/typelevel/cats/pull/1142/files#diff-26729163874fb5967d1a0c27b6db2457R60). Can we add a scalastyle rule to make this consistent?

fthomas:
I made a minor comment about spaces after `if`. Apart from that, this LGTM. :+1:

core/src/main/scala/cats/data/OneAnd.scala
ceedubs: Wow good catch. The other places were caught by `EnsureSingleSpaceAfterTokenChecker`. It seems to me like a scalastyle bug that this particular place isn't caught by that rule.

ceedubs:
@fthomas I've added a rule for space after `if` since for some reason `EnsureSingleSpaceAfterTokenChecker` wasn't working in certain cases.

Issue #957, opened 2016-03-28
s/ListWrapper/OptionWrapper


adelbertc:
Lots of Lists flying around makes build times slow and sometimes time out.

ceedubs:
I've thought about this in the past. One downside is that we only have some type class instances (`Semigroup`, `Monoid`) for `Option[A]` if we have a `Semigroup[A]` instance, which isn't the case with `List[A]`. You can create an alternative version of `Semigroup[A]` that uses `orElse` and doesn't require `Semigroup[A]`. I think that I advocated for this in an issue way back but ultimately we decided not to go with it. I guess we could have `OptionWrapper` do that, but I could see the difference between the two getting confusing.

ceedubs:
@adelbertc since our builds haven't been timing out for a long time now, I'm going to go ahead and close this out. See also the experiment in #1143.

PR #1143, opened 2016-06-18
Reduce size of arbitrary ListWrappers

ceedubs:
This is an experiment in response to #957. If this doesn't have much of
an effect on build times, then I think we can go ahead and close out
that ticket, since we are no longer seeing build timeouts. I suspect it
won't make much of a difference, because we already have changed to a
more performant JS environment and have reduced the scalacheck size
configuration in CatSuite.


ceedubs:
This seemed to have no meaningful impact on build times. I'm going to close this out along with #957.

Issue #1054, opened 2016-05-21
filterM broken?


3rdLaw:
I ran the following in sbt console after a git pull:
```scala
scala> import cats.implicits._
import cats.implicits._
scala> import cats._
import cats._
scala> List(1, 2, 3).filterM { x => List(true, false) }
res0: List[Int] = List(1, 2, 3)
```

Compare this to scalaz's results:
```scala
import scalaz._
import scalaz.Scalaz._
List(1, 2, 3).filterM { x => List(true, false) }
res0: List[List[Int]] = List(List(1, 2, 3), List(1, 2), List(1, 3), List(1), List(2, 3), List(2), List(3), List())
```

ceedubs:
@3rdLaw thanks for bringing this up. The `filterM` in cats indeed isn't the `filterM` in Haskell or scalaz. It doesn't even have quite the same type signature: the one in cats is not specialized to `List` and returns an `F[A]` result as opposed to `F[F[A]]`.

We should probably do something to avoid confusion. @non it looks like you originally added `filterM` - what are your thoughts?

3rdLaw:
Thanks for the clarification @ceedubs; I clearly assumed this was the same as Haskell/Scalaz's, so some documentation would be helpful.

Does cats have an equivalent to `filterM` in Haskell or scalaz?

ceedubs:
@non to avoid confusion, what do you think about renaming our `fliterM` to `flatFilter` to parallel the relationship between `map` and `flatMap`?

ceedubs:
@non on second thought, can you think of any practical example of using the method as it currently exists in cats? `Option` is about the only type that I imagine myself using it with, and in that case it would be pretty easy to use `flatMap` with `filter` instead.

Issue #842, opened 2016-01-31
collect and mapMaybe for MonadFilter?


alexknvl:
Along the lines of:

``` scala
  implicit class MonadFilterOps[F[_], A](val fa: F[A]) extends AnyVal {
    def collect[B](pf: PartialFunction[A, B])(implicit F: MonadFilter[F]): F[B] =
      F.map(F.filter[A](fa)(pf.isDefinedAt))(pf.apply)
    def mapMaybe[B](f: A => Option[B])(implicit F: MonadFilter[F]): F[B] =
      F.flatMap(fa) { a => f(a) match {
          case Some(x) => F.pure(x)
          case None => F.empty
        }
      }
  }
```

I am not sure about the names (I took `collect` from the standard Scala collection library and `mapMaybe` from [Haskell's Data.Maybe](https://hackage.haskell.org/package/base-4.8.2.0/docs/Data-Maybe.html#v:mapMaybe)).

Some random thoughts:
- Having a method that combines `filter` and `map` could be beneficial performance-wise (an implementation for `List` could do both in a single fold).
- Concise implementation of `List[Xor[A, B]] => List[B]` (or some other `MonadFilter` instance, since `List` already has `collect`):

``` scala
list.collect { case Xor.Right(x) => x }
list.mapMaybe { case Xor.Right(x) => Some(x); case Xor.Left(_) => None }
```
- `mapMaybe` might not seem much better than using `flatMap` with `pure` and `empty`.
- But more familiar and compatible with pre-existing `A => Option[B]` functions.

ceedubs:
@alexknvl thank you for the detailed suggestion. I agree that some things along these lines could be useful.

Some half-baked thoughts that come to mind as I ponder this:
- Something similar to `collect` but with slightly different constraints could exist. You could define it on `Foldable` and require that you collect into a `G[_]` type that has a `MonadCombine` instance. This alternative approach might have a more performant default implementation but it might not lead to efficient overrides. I think both could likely exist.
- My first reaction to `mapMaybe` was that the `Option` seems overly specific. But I guess you either need to either do this, map to `F[B]` instead of `Option[B]` and use `flatMap` (which can already be done), or put this on `MonadCombine` where I think you could generalize to `F[B]` for any `Foldable` `F`. This use-case might be common enough that it's worth supporting in the way that you've written it.
- We may want to call it `mapOption` or something else instead of `mapMaybe` :)

alexknvl:
[...]

PR #1148, opened 2016-06-19
TraverseFilter (AKA Witherable) type class

ceedubs:
This is not meant to be merged in its current form.

This is a proof of concept for adding a `TraverseFilter` type class, which is a port of Haskell's [Data.Witherable](https://hackage.haskell.org/package/witherable-0.1.3.3/docs/Data-Witherable.html). I'm not set on any of the naming, so feel free to offer other recommendations.

Would people be interested in something like this? Some things that I like about it are:
- It provides `filterM` (actually a more general version of it) #1054.
- It provides `collect` and `mapFilter` (aka `mapMaybe`) methods #842. Since `collect` is often used on standard library collections, I think this is a nice feature to support.
- `traverseFilter` supports operations that essentially do a `filter` and a `traverse` at the same time, which is something that you can't get from either `Traverse` or `MonadFilter`.

Some things that I'm not so sure about:
- There's definitely some overlap with `MonadFilter` here. For example, `filter` and `collect` can be defined for both. I'm not sure if there should be a closer relationship between the two. Also syntax enrichment for methods like `filter` could potentially be ambiguous between the two types (though I don't know how much of a problem this will be in practice as most concrete types that support `filter` will have it defined on them).
- (**edit**: I think this is resolved) I'm not so sure about the composition law. The Haskell wiki defines it as `Compose . fmap (wither f) . wither g ≡ wither (Compose . fmap (wither f) . g)`, but I can't get the types to line up on that. I don't know Haskell very well and struggle with point free style, so I'm probably losing something in translation. I have a different version of the right hand side of the composition law that compiles, but it seems much less direct than the haskell version, and I don't know know if it's meaningfully distinct from the left hand side. I'm hoping someone who knows Haskell well (such as @aaronlevin or @sellout) and/or someone who knows laws well (such as @fthomas) can weigh in on this.

If people want to go forward with this I need to provide more instances, tests, and documentation.


core/src/main/scala/cats/Collect.scala
```diff
@@ -0,0 +1,96 @@
...
+   * scala> import cats.implicits._
+   * scala> val m: Map[Int, String] = Map(1 -> "one", 3 -> "three")
+   * scala> val l: List[Int] = List(1, 2, 3, 4)
+   * scala> def asString(i: Int): Eval[Option[String]] = Now(m.get(i))
+   * scala> val result: Eval[List[String]] = l.mapOptionA(asString)
+   * scala> result.value
+   * res0: List[String] = List(one, three)
+   * }}}
+   */
+  def mapOptionA[G[_]: Applicative, A, B](fa: F[A])(f: A => G[Option[B]]): G[F[B]]
```
ceedubs: I guess another potential name for this would be `traverseOption`.

core/src/main/scala/cats/MonadFilter.scala
```diff
@@ -15,7 +15,4 @@ import simulacrum.typeclass
 
   def filter[A](fa: F[A])(f: A => Boolean): F[A] =
     flatMap(fa)(a => if (f(a)) pure(a) else empty[A])
-
-  def filterM[A](fa: F[A])(f: A => F[Boolean]): F[A] =
-    flatMap(fa)(a => flatMap(f(a))(b => if (b) pure(a) else empty[A]))
```
ceedubs: I've removed this based on https://github.com/typelevel/cats/issues/1054#issuecomment-226961198 and to prevent confusion with `filterA`, but that should maybe be a topic for a separate PR.

laws/src/main/scala/cats/laws/CollectLaws.scala
```diff
@@ -0,0 +1,36 @@
...
+  ): IsEq[Nested[M, N, F[C]]] = {
+
+    val lhs: Nested[M, N, F[C]] = Nested(fa.mapOptionA(f).map(_.mapOptionA(g)))
+    // TODO is this right and/or meaningful? I can't seem to get the types to
+    // line up for the a straight port of `wither (Compose. fmap (wither f). g)`
+    // which is how the law is stated for Haskell's `Witherable`
+    val rhs: Nested[M, N, F[C]] = fa.mapOptionA[Nested[M, N, ?], C](a =>
+      Nested(f(a).map(_.traverseM(g))))
+    lhs <-> rhs
+  }
```
sellout: So, I haven’t _tried_ it, but the only difference I see between your law and the Haskell one is that the Haskell uses `mapOptionA` where you use `traverseM`. And I _think_ the types line up if you make that change.

More superficially, the Haskell version also swaps the names `f` and `g`.

ceedubs: @sellout It's an `Option[B]` that I'm calling `traverseM` on. I think there's an extra layer of `Option` if I try to use `mapOptionA`. It seems off since I'm not actually calling `mapOptionA` twice.

sellout: I think it’s fine that it’s an `Option[B]` – the type signatures of `mapOptionA` and `traverseM` are both `F[A] =...> G[F[B]]`
```scala
def mapOptionA[G[_]: Applicative, A, B](fa: F[A])(f: A => G[Option[B]]): G[F[B]]
def traverseM[G[_]: Applicative, A, B](fa: F[A])(f: A => G[F[B]]): G[F[B]]
```
so in this specific case, they both go `Option[B] => (B => N[Option[C]]) => N[Option[C]]`. The only difference is how the `Option` in `N[Option[C]]` is handled – IIUC, `mapOptionA` throws away the `None` cases, while `traverseM` flatMaps them, which doesn’t throw them away.

ceedubs: Ah so do I need to write an instance for Option and use it in the law? I
assumed I would only need the instance for the abstract F.
On Jun 19, 2016 10:17 PM, "Greg Pfeil" <notifications@github.com> wrote:

In laws/src/main/scala/cats/laws/CollectLaws.scala
<https://github.com/typelevel/cats/pull/1148#discussion_r67630044>:

> +    fa: F[A],
> +    f: A => M[Option[B]],
> +    g: B => N[Option[C]]
> +  )(implicit
> +      M: Applicative[M],
> +      N: Applicative[N]
> +  ): IsEq[Nested[M, N, F[C]]] = {
> +
> +    val lhs: Nested[M, N, F[C]] = Nested(fa.mapOptionA(f).map(_.mapOptionA(g)))
> +    // TODO is this right and/or meaningful? I can't seem to get the types to
> +    // line up for the a straight port of `wither (Compose. fmap (wither f). g)`
> +    // which is how the law is stated for Haskell's `Witherable`
> +    val rhs: Nested[M, N, F[C]] = fa.mapOptionA[Nested[M, N, ?], C](a =>
> +      Nested(f(a).map(_.traverseM(g))))
> +    lhs <-> rhs
> +  }

I think it’s fine that it’s an Option[B] – the type signatures of mapOptionA
and traverseM are both F[A] =...> G[F[B]]

def mapOptionA[G[_]: Applicative, A, B](fa: F[A])(f: A =>
G[Option[B]]): G[F[B]]def traverseM[G[_]: Applicative, A, B](fa:
F[A])(f: A => G[F[B]]): G[F[B]]

so in this specific case, they both go Option[B] => (B => N[Option[C]]) =>
N[Option[C]]. The only difference is how the Option in N[Option[C]] is
handled – IIUC, mapOptionA throws away the None cases, while traverseM
flatMaps them, which doesn’t throw them away.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/typelevel/cats/pull/1148/files/139659472e3b698dd0e392099f01155f33cbe1b5#r67630044>,
or mute the thread
<https://github.com/notifications/unsubscribe/AA7sCUSHu09nEV8TRexRAl7JhuoU5XJiks5qNfhMgaJpZM4I5Mbj>
.

sellout: I actually hadn’t thought about it, but yeah, it looks like that’s the case. It looks like Witherable has an instance for Maybe as well as some others.

ceedubs: Thanks for the help @sellout. I've updated this PR to include what I think is the right representation of the law.

ceedubs:
I renamed the type class from `Collect` to `TraverseFilter`, which seems like a slightly better name to me (and relates to `MonadFilter`). Still open to other name ideas.

kailuowang:
Looks great to me. Thanks very much! @ceedubs 👍

core/src/main/scala/cats/TraverseFilter.scala
```diff
@@ -0,0 +1,116 @@
+package cats
+
+import simulacrum.typeclass
+
+/**
+ * `TraverseFilter`, also known as `Witherable` represents list-like structures
+ * that can essentially have a [[traverse]] and a [[filter]] applied as a single
+ * combined operation ([[traverseFilter]]).
+ *
+ * TODO ceedumbs more info
```
johnynek: Can we add the laws:
```
A definition of wither must satisfy the following laws:

identity
wither (pure . Just) ≡ pure
composition
Compose . fmap (wither f) . wither g ≡ wither (Compose . fmap (wither f) . g)
```

johnynek: By which I mean, mention in the comments the laws, I see they are encoded below in the tests.

johnynek:
This looks really nice.

👍

core/src/main/scala/cats/Traverse.scala
```diff
@@ -55,6 +55,12 @@ import simulacrum.typeclass
       val G = Traverse[G]
     }
 
+  def composeFilter[G[_]: TraverseFilter]: TraverseFilter[λ[α => F[G[α]]]] =
```
ceedubs: Are people happy with this? Since we only need a `Traverse` for `F` but a `TraverseFilter` for `G`, and there is a separate `compose` when `G` only has `Traverse`, it doesn't quite fit the common pattern.

core/src/main/scala/cats/data/Nested.scala
```diff
@@ -27,10 +27,36 @@ final case class Nested[F[_], G[_], A](value: F[G[A]])
...
+      override def traverse[H[_]: Applicative, A, B](fga: Nested[F, G, A])(f: A => H[B]): H[Nested[F, G, B]] =
+        Applicative[H].map(instance.traverse(fga.value)(f))(Nested(_))
+
+      override def map[A, B](fga: Nested[F, G, A])(f: A => B): Nested[F, G, B] =
+        Nested(instance.map(fga.value)(f))
+
+      override def imap[A, B](fga: Nested[F, G, A])(f: A => B)(g: B => A): Nested[F, G, B] =
+        Nested(instance.imap(fga.value)(f)(g))
+    }
+}
```
ceedubs: @adelbertc this duplicates a lot of what `catsDataTraverseForNested` has below. It is starting to feel like we should be separating out `trait NestedTraverse` etc traits like has been done for `Composed`. What are your thoughts?

ceedubs:
@kailuowang @johnynek thanks for the reviews! Since you've shown support for this change, I've fleshed it out a bit. Would you mind taking a second look?

I left a couple minor comments seeking feedback, but I think they are for things that could probably be addressed in a later PR.

The one item that I'm still a little uneasy about is this (from the original PR description):

> There's definitely some overlap with MonadFilter here. For example, filter and collect can be defined for both. I'm not sure if there should be a closer relationship between the two. Also syntax enrichment for methods like filter could potentially be ambiguous between the two types (though I don't know how much of a problem this will be in practice as most concrete types that support filter will have it defined on them).

core/src/main/scala/cats/data/Nested.scala
adelbertc: Yeah a lot of the duplication I couldn't think of a way around because I also wanted to make sure default implementations were overriden where possible, so I had to do a lot of manual overriding/delegation like you do here. If you have ideas on making it cleaner I'm all for it :-)

ceedubs:
I'm closing this in favor of #1225, where I've cleaned up the git history a bit and added `FunctorFilter` to resolve the concerns I had raised about overlap between `TraverseFilter` and `MonadFilter`.

PR #1153, opened 2016-06-20
Use Simulacrum for Traverse syntax

### LINKED ISSUE #1146: Use Simulacrum Ops for Traverse syntax
I just noticed that our `Traverse` syntax seems to be manually-built instead of using the simulacrum ops. It doesn't seem like this should be necessary to me. Does anyone know of a reason that we can't use the simulacrum ops instead?

cc @stew because I think that you originally added it and @mpilquist because simulacrum

ceedubs:
Fixes #1146.

I followed the convention for `Functor` (okay I copied it and did a find/replace). Considering the recent fix for SI-2712, we'll probably want to handle the `Unapply` differently, but I'd be inclined to solve all of that at once under #1073.


ceedubs:
For some reason this didn't trigger a Travis build. I'm going to try reopening it to see if that does.

ceedubs:
@kailuowang I just moved the examples from the Ops class into the type class. One thing that I don't quite like about this is that the examples don't directly use the method that they are attached to - they actually use the Ops version of the method. I'm wondering whether it's better to have the example use the syntax since that's the easier way to use these or they should summon the type class instance and call methods on it, since that would should an example of actually using the method the scaladoc is for. Sometimes the the data type used in the example will already have the method directly on it (such as `List.filter`), so the example wouldn't actually use the type class at all!

kailuowang:
Ah, didn't realize that. I agree that using syntax in examples makes more sense. Maybe we can address that in a different PR?

peterneyens:
This looks good :+1: 

We could probably create a new issue on how to handle the ScalaDoc examples for the type class methods versus the enriched syntax methods we get from Simulacrum.

PR #1155, opened 2016-06-20
Add syntax to create a XorT from an F[A] and from a Xor.

peterneyens:
I have added syntax to create a `XorT` from a 

- `F[A]` :  `fa.leftXorT` and `fa.rightXorT` and which use `XorT.left` and `XorT.right`
- `Xor` : `xor.toXorT` (which was suggested by @ceedubs in #1078)

The last one is a little bit difficult:
- If I define it on `Xor` I need to work around the covariance of `A` and `B` in `Xor` versus the invariance in `XorT`, so you would need to write `xor.toXorT[F, A, B]`. 
- I have also defined `toXorT` (currently `toXorT2` in `syntax.xor`) as a syntax function on `Xor` so you can write `xor.toXorT[F]`. But it's probably not expected that there are syntax functions for a data type defined in cats itself (at least I couldn't think of / find an earlier precedent).

I prefer the syntax function because it is nicer/easier to use, but it would probably be more difficult to find, since it won't be in the `Xor` scaladoc. WDYT ?

### DIFF
--- a/core/src/main/scala/cats/data/Xor.scala
+++ b/core/src/main/scala/cats/data/Xor.scala
@@ -19,6 +19,9 @@ import scala.util.{Failure, Success, Try}
  * `flatMap` apply only in the context of the "right" case. This right bias makes [[Xor]] more convenient to use
  * than `scala.Either` in a monadic context. Methods such as `swap`, and `leftMap` provide functionality
  * that `scala.Either` exposes through left projections.
+ *
+ * Some additional [[Xor]] methods can be found in [[Xor.XorOps XorOps]]. These methods are not defined on [[Xor]] itself because
+ * [[Xor]] is covariant in its types `A` and `B`.
  */
 sealed abstract class Xor[+A, +B] extends Product with Serializable {
 
@@ -188,6 +191,20 @@ sealed abstract class Xor[+A, +B] extends Product with Serializable {
 object Xor extends XorInstances with XorFunctions {
   final case class Left[+A](a: A) extends (A Xor Nothing)
   final case class Right[+B](b: B) extends (Nothing Xor B)
+
+  final implicit class XorOps[A, B](val value: A Xor B) extends AnyVal {
+    /**
+     * Transform the `Xor` into a [[XorT]] while lifting it into the specified Applicative.
+     *
+     * {{{
+     * scala> import cats.implicits._
+     * scala> val x: Xor[String, Int] = Xor.right(3)
+     * scala> x.toXorT[Option]
+     * res0: cats.data.XorT[Option, String, Int] = XorT(Some(Right(3)))
+     * }}}
+     */
+    def toXorT[F[_]: Applicative]: XorT[F, A, B] = XorT.fromXor(value)
+  }
 }
 
 private[data] sealed abstract class XorInstances extends XorInstances1 {
--- a/core/src/main/scala/cats/syntax/xort.scala
+++ b/core/src/main/scala/cats/syntax/xort.scala
@@ -1,0 +1,39 @@
+package cats
+package syntax
+
+import cats.data.XorT
+
+trait XorTSyntax extends XorTSyntax1 {
+  implicit def catsSyntaxXorT[F[_]: Functor, A](fa: F[A]): XorTFunctorOps[F, A] = new XorTFunctorOps(fa)
+}
+
+private[syntax] trait XorTSyntax1 {
+  implicit def catsSyntaxUXorT[FA](fa: FA)(implicit U: Unapply[Functor, FA]): XorTFunctorOps[U.M, U.A] =
+    new XorTFunctorOps[U.M, U.A](U.subst(fa))(U.TC)
+}
+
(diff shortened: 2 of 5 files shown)


tests/src/test/scala/cats/tests/XorTests.scala
```diff
@@ -228,7 +229,7 @@ class XorTests extends CatsSuite {
     }
   }
 
-  test("combine is right iff both operands are right") {
+  test("combine is right if both operands are right") {
     forAll { (x: Int Xor String, y: Int Xor String) =>
```
kailuowang: hmmm, I wonder why you changed this. This test seems testing `iff` to me

peterneyens: Oops, thought it was a typo :wink:

core/src/main/scala/cats/data/Xor.scala
```diff
@@ -83,6 +83,19 @@ sealed abstract class Xor[+A, +B] extends Product with Serializable {
...
+   *
+   * {{{
+   * scala> import cats.implicits._
+   * scala> val x: Xor[String, Int] = Xor.right(3)
+   * scala> x.toXorT[Option, String, Int]
+   * res0: cats.data.XorT[Option, String, Int] = XorT(Some(Right(3)))
+   * }}}
+   */
+  def toXorT[F[_], L, R](implicit F: Applicative[F], evL: A <:< L, evR: B <:< R): XorT[F, L, R] =
+    XorT(F.pure(bimap(evL, evR)))
```
kailuowang: Hmmm if `toXorT` has to respecify all type parameters I am not sure I will favor it over `fromXor`

peterneyens: Indeed, that is what I was thinking. What do you think about adding it as a syntax method (see `syntax.xor.toXorT2`) ?

kailuowang:
I vote for the `toXorT2` in syntax. The one on data class itself is hard to use and doesn't feel right to me.  The users I know are used to find methods in syntax so discoverability doesn't seem an issue to me.

core/src/main/scala/cats/data/Xor.scala
ceedubs: Yeah, I wasn't thinking about the variance on `Xor` when I proposed this previously. The syntax approach doesn't look bad, but I want to ponder our options a bit.

ceedubs:
For anyone who has asked about what the downsides of variance are, this is an example :)

What if we were to put something like this in the `Xor` companion object?

```scala
  final implicit class XorOps[A, B](val value: A Xor B) extends AnyVal {
    def toXorT[F[_]:Applicative]: XorT[F, A, B] = XorT.fromXor(value)
  }
```

If we do that then people should automatically pick up the syntax on their `Xor` instances and therefore it will probably show up in their IDE's autocomplete. As you've mentioned, people probably won't expect to need an import to add this sort of method onto `Xor`, which is why I'm inclined to go down this route as opposed to the normal syntax route; I would view this not being directly on `Xor` as an implementation detail and not something that people should have to think about.

We could also link to `XorOps` within the `Xor` scaladoc with a note like "some additional `Xor` methods can be found here" and maybe a brief explanation of why they are separate.

What do you think?

core/src/main/scala/cats/syntax/xort.scala
```diff
@@ -0,0 +1,39 @@
...
+trait XorTSyntax extends XorTSyntax1 {
+  implicit def catsSyntaxXorT[F[_]: Functor, A](fa: F[A]): XorTOps[F, A] = new XorTOps(fa)
+}
+
+trait XorTSyntax1 {
+  implicit def catsSyntaxUXorT[FA](fa: FA)(implicit U: Unapply[Functor, FA]): XorTOps[U.M, U.A] =
+    new XorTOps[U.M, U.A](U.subst(fa))(U.TC)
+}
+
+final class XorTOps[F[_]: Functor, A](val fa: F[A]) {
```
ceedubs: Should we call this `XorTFunctorOps` or something? We could conceivably want to add `XorT`-related syntax onto types other than `F[A]`. I think it should still be mixed in via `XorTSyntax` though.

kailuowang:
Many thanks! 👍

core/src/main/scala/cats/syntax/xort.scala
```diff
@@ -0,0 +1,39 @@
+package cats
+package syntax
+
+import cats.data.XorT
+
+trait XorTSyntax extends XorTSyntax1 {
+  implicit def catsSyntaxXorT[F[_]: Functor, A](fa: F[A]): XorTFunctorOps[F, A] = new XorTFunctorOps(fa)
+}
+
+trait XorTSyntax1 {
```
ceedubs: We probably want to make this private (as per #1160).

peterneyens:
Thanks for the comments @ceedubs and @kailuowang. Even a simple PR like this looks a lot nicer after improving the little issues you pointed out.

ceedubs:
👍 excellent. Thank you @peterneyens!

PR #1157, opened 2016-06-23
Use tut:book in invariantmonoidal doc

ceedubs:
To make it consistent with other docs.


kailuowang:
usage of `std` package fails the build. My 👍  after build pass

ceedubs:
@kailuowang I fixed the `std` issue in #1142 when I was merging in the changes from master. If we think that's good to go, then that should resolve this issue. Otherwise I can fix it separately and then resolve the conflicts in that PR.

kailuowang:
Good deal, just gave my thumbs up on #1142

OlivierBlanvillain:
:+1:

Issue #814, opened 2016-01-18
Update SBT MiMa `previousArtifacts` once cats-kernel is released


DavidGregory084:
Further to #809, which adds the Migration Manager plugin to the build, we should make sure to update the `previousArtifacts` in `build.sbt` once #777 is completed and cats-kernel is released.

ceedubs:
Closing in favor of #608, as this referred to a branch that didn't end up getting merged.

PR #1159, opened 2016-06-25
Add mima for kernel

johnynek:
This adds mima for kernel, but it does not pass, so it is not added to travis yet.

As discussed when moving kernel from algebra, the thinking was we could be more conservative there. I would like to turn this on after the next release (I guess 0.7.0) and I hope we can be conservative with kernel updates so that algebra is not too tightly coupled to cats.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -203,9 +203,9 @@ lazy val kernel = crossProject.crossType(CrossType.Pure)
   .settings(scoverageSettings: _*)
   .settings(sourceGenerators in Compile <+= (sourceManaged in Compile).map(KernelBoiler.gen))
   .jsSettings(commonJsSettings:_*)
-  .jvmSettings(commonJvmSettings:_*)
+  .jvmSettings((commonJvmSettings ++ (mimaPreviousArtifacts := Set("org.typelevel" %% "cats-kernel" % "0.6.0"))):_*)
 
-lazy val kernelJVM = kernel.jvm
+lazy val kernelJVM = kernel.jvm 
 lazy val kernelJS = kernel.js
 
 lazy val kernelLaws = crossProject.crossType(CrossType.Pure)
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -1,6 +1,7 @@
 addSbtPlugin("com.eed3si9n"         % "sbt-unidoc"            % "0.3.2")
 addSbtPlugin("com.github.gseitz"    % "sbt-release"           % "1.0.0")
 addSbtPlugin("com.jsuereth"         % "sbt-pgp"               % "1.0.0")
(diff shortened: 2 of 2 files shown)


johnynek:
relates to #814 (does not close, since we have failing test case now).

kailuowang:
We could also make the previousArtifacts dynamically check previous bug fix version, like 
```Scala 
  lazy val mimaSettings = Seq(
    mimaPreviousArtifacts := (if (scalaBinaryVersion.value != "2.10") {
      Version(version.value).flatMap {
        // we are currently only checking binary compat with the previous bugfix version
        case Version(major, Some(minor), Some(bugfix), _) if bugfix > 0 =>
          Some(Set(organization.value %% moduleName.value % Seq(major, minor, bugfix - 1).mkString(".")))
        case  _ => None
      }
    } else None).getOrElse(Set.empty)
  )


````

benhutchison:
In my workplace, many Scalaz/Argonaut-based projects have become binary- and source- compatibility minefields. Pleased to see cats taking version compatibility seriously from the outset; I hope concern from compatibility becomes part of Cats' "cultural DNA".

johnynek:
big +1 to this. binary compatibility should be taken very seriously for a library like cats that hopes to be at the bottom of a lot of dependencies.

johnynek:
@kailuowang https://github.com/typelevel/cats/pull/1159#issuecomment-228499175 that is a great idea. In fact, I think we can do something like this to verify semver:

if `major == 0` then we check previous bugfix, if `major > 0` we check previous bugfix and previous patch at version 0 perhaps?

That said, my overall point is that I think cats-kernel should enforce binary compatibility sooner (maybe like now), so that algebra can move close to a 1.0 release.

kailuowang:
Sorry not sure I follow completely. Are you suggesting that for `kernel` we maintain binary comp between minor versions between `0.7.0` and `1.0.0` ?

johnynek:
Yes, I am. That was my understanding when we discussed moving that code
from algebra to cats (since it has changed very little in two years, this
does not seem onerous).

On Saturday, June 25, 2016, Kai(luo) Wang <notifications@github.com> wrote:

> Sorry not sure I follow completely. Are you suggesting that for kernel we
> maintain binary comp between minor versions between 0.7.0 and 1.0.0 ?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/typelevel/cats/pull/1159#issuecomment-228579665>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AAEJdmQq7UieVdS2tgCSTeUlCo89hNuvks5qPdpLgaJpZM4I-OQp>
> .
>


-- 

ceedubs:
👍 thanks @johnynek. I'll create a follow-up issue to enable this after the next release.

PR #1145, opened 2016-06-18
Topic/doc overhaul wip

non:
This is a doc overhaul @ceedubs and I were working on.

I'm not 100% sure it's ready to go but I'm pushing this for early review. Suggestions, editions, and additions are all very welcome. /cc @ceedubs

PR #1162, opened 2016-06-27
Include kernel in modules doc

ceedubs:
As suggested in
https://github.com/typelevel/cats/pull/1145#discussion_r67731278


kailuowang:
👍

PR #1163, opened 2016-06-27
Break import guide out from FAQ

ceedubs:
This seemed like a bit too much information to shove into a FAQ answer. This also leaves room for a more detailed guide on piecemeal imports if someone gets around to writing it.


kailuowang:
👍

PR #1167, opened 2016-06-28
Add max/min and reduceOption methods

### LINKED ISSUE #1164: Create max/min methods for Foldable and Reducible
We should have methods like `maximumOption`/`minimumOption` on `Foldable` and `maximum`/`minimum` on `Reducible`. The default implementation can simply `foldLeft`/`reduceLeft` through the structure and keep track of the max/min. These methods would require an implicit `Order` instance.

ceedubs:
This adds:
 - `reduceLeftOption` and `reduceRightOption` methods to `Foldable`
 - `maximumOption` and `minimumOption` methods to `Foldable`
 - `maximum` and `minimum` methods to `Reducible`

Resolves #1164.

### DIFF
--- a/core/src/main/scala/cats/Foldable.scala
+++ b/core/src/main/scala/cats/Foldable.scala
@@ -58,6 +58,90 @@ import simulacrum.typeclass
     }
 
   /**
+   * Reduce the elements of this structure down to a single value by applying
+   * the provided aggregation function in a left-associative manner.
+   *
+   * @return `None` if the structure is empty, otherwise the result of combining
+   * the cumulative left-associative result of the `f` operation over all of the
+   * elements.
+   *
+   * @see [[reduceRightOption]] for a right-associative alternative.
+   *
+   * @see [[Reducible#reduceLeft]] for a version that doesn't need to return an
+   * `Option` for structures that are guaranteed to be non-empty.
+   *
+   * Example:
+   * {{{
+   * scala> import cats.implicits._
+   * scala> val l = List(6, 3, 2)
+   * This is equivalent to (6 - 3) - 2
+   * scala> Foldable[List].reduceLeftOption(l)(_ - _)
+   * res0: Option[Int] = Some(1)
+   *
+   * scala> Foldable[List].reduceLeftOption(List.empty[Int])(_ - _)
+   * res1: Option[Int] = None
+   * }}}
+   */
+  def reduceLeftOption[A](fa: F[A])(f: (A, A) => A): Option[A] =
+    reduceLeftToOption(fa)(identity)(f)
+
+  /**
+   * Reduce the elements of this structure down to a single value by applying
+   * the provided aggregation function in a right-associative manner.
+   *
+   * @return `None` if the structure is empty, otherwise the result of combining
+   * the cumulative right-associative result of the `f` operation over the
+   * `A` elements.
+   *
+   * @see [[reduceLeftOption]] for a left-associative alternative
+   *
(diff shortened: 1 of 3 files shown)


kailuowang:
👍 thanks!

core/src/main/scala/cats/Reducible.scala
```diff
@@ -112,6 +112,12 @@ import simulacrum.typeclass
       val F = self
       val G = Reducible[G]
     }
+
+  def minimum[A](fa: F[A])(implicit A: Order[A]): A =
```
johnynek: why not `.min` to be closer to stdlib (and fewer chars).

core/src/main/scala/cats/Foldable.scala
```diff
@@ -58,6 +58,90 @@ import simulacrum.typeclass
...
+   *
+   * @return `None` if the structure is empty, otherwise the minimum element
+   * wrapped in a `Some`.
+   *
+   * @see [[Reducible#minimum]] for a version that doesn't need to return an
+   * `Option` for structures that are guaranteed to be non-empty.
+   *
+   * @see [[maximumOption]] for maximum instead of minimum.
+   */
+  def minimumOption[A](fa: F[A])(implicit A: Order[A]): Option[A] =
```
johnynek: what about `minOption`. Do we need such a long name?

core/src/main/scala/cats/Reducible.scala
ceedubs: I did this specifically to avoid collision with the std lib. If I refactor something from a `NonEmptyList` to a `List` I'd prefer to get a compile error instead of something with no type signature change but that will start throwing exceptions at runtime.

Having said that, there's a question of how much we should bend over backwards to avoid collisions with std lib names. I'm not set in stone on this if people would prefer `min`.

johnynek: In this case you would have `Reducible[NonEmptyList].min(myThing)` how could there be a collision with the stdlib in practice? The collision could come with an implicit class that adds `.min` to any `Reducible`, and then we have anxiety, perhaps, when reading the code if we are getting the safe `Reducible` one or the stdlib (which would be safe in exactly the cases that the Reducible exists).

Is that what you mean?

ceedubs: @johnynek yes that's what I mean, but note that we already have syntax enrichment for `Reducible`, and it's generated by Simulacrum so it will automatically inherit the name of the method on the type class.

ceedubs: @johnynek do you have any further thoughts on this after my note about the syntax enrichment that currently exists?

If there's a strong preference for `min` then I can change it, but I do have a hesitation about potential refactoring bugs that could come out of it.

johnynek: I see...

I'm not a huge implicit syntax fan (this example, but also confusion for novices as to where methods are coming from).

But I recognize many cats users do like this, so avoiding the collision seems safest.

👍

core/src/main/scala/cats/Foldable.scala
ceedubs: This is just to be consistent with `minimum`. Is this okay considering the discussion [here](https://github.com/typelevel/cats/pull/1167#discussion_r69006590)?

johnynek: totally. Sorry. I'm +1 on the whole PR. My 👍 above may not have been clear about that.

johnynek:
👍

peterneyens:
Exemplary documentation :+1:

PR #1169, opened 2016-06-29
Adding lift and inspect to StateT

zainab-ali:
Adding utility functions `lift` and `inspect` to the `StateT` object.

### DIFF
--- a/core/src/main/scala/cats/data/StateT.scala
+++ b/core/src/main/scala/cats/data/StateT.scala
@@ -137,8 +137,17 @@ object StateT extends StateTInstances {
   def lift[F[_], S, A](fa: F[A])(implicit F: Applicative[F]): StateT[F, S, A] =
     StateT(s => F.map(fa)(a => (s, a)))
 
-  def inspect[F[_], S, A](f: S => F[A])(implicit F: Applicative[F]): StateT[F, S, A] =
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/data/StateT.scala
```diff
@@ -133,6 +133,12 @@ object StateT extends StateTInstances {
 
   def pure[F[_], S, A](a: A)(implicit F: Applicative[F]): StateT[F, S, A] =
     StateT(s => F.pure((s, a)))
+
+  def lift[F[_], S, A](fa: F[A])(implicit F: Applicative[F]): StateT[F, S, A] =
+    StateT(s => F.map(fa)(a => (s, a)))
+
+  def inspect[F[_], S, A](f: S => F[A])(implicit F: Applicative[F]): StateT[F, S, A] =
+    StateT(s => F.map(f(s))(a => (s, a)))
```
ceedubs: Should we have `inspect` take `S => A` and create a separate `inspectF` for `S => F[A]`? It seems like the simple `S => A` case might be pretty common, so it might be nice to make it concise.

zainab-ali: Agreed.  I'm thinking of adding a `modify` taking `S => S` and a `modifyF` taking `S => F[S]` while 
I'm at it.

ceedubs:
👍 looks great to me. Thanks @zainab-ali!

kailuowang:
👍  many thanks!

Issue #1160, opened 2016-06-25
Make implicit priority and similarly purposed traits private


adelbertc:
See: https://www.javadoc.io/doc/org.typelevel/cats-core_2.11/0.6.0 - I believe stuff like `OneAndLowPriorityN` should be private.

peterneyens:
Fixed by #1171.

PR #1171, opened 2016-06-30
Make the implicit priority traits private.

peterneyens:
Hide the traits which are used for implicit resolution prioritization #1160.


kailuowang:
Thanks! 👍

ceedubs:
Wonderful. Thanks @peterneyens!

Issue #1107, opened 2016-06-09
Applicative/Functor instance missing for OptionT


tbrown1979:
Missing for when F doesn't have a Monad

yilinwei:
This is absolutely the case because a monad transfromer is not the same as a `Nested`. 

See explanation here #1106.

peterneyens:
It would still be possible to add something like `scalaz.OptionT[F, ?].app`, but it would give the same result as `Nested[F, Option, ?].ap` so that wouldn't be that useful probably.

I was wondering if we should add some sort of `toNested` on  `OptionT` and `XorT`, in that way you could do `OptionT(List(Some(1))).toNested.ap( ... )`.
We could probably also use it to implement methods like `OptionT.foldLeft` for `Foldable[OptionT[F, ?]]` (which I just added in #1103, which also includes the `Functor` you mentioned)

```scala
def foldLeft[B](b: B)(f: (B, A) => B)(implicit F: Foldable[F]): B =
  toNested.foldLeft(value, b)(f)
  // F.foldLeft(value, b)((b, o) => optionInstance.foldLeft(o, b)(f))
```

ceedubs:
@peterneyens I like your idea of `toNested`.

ceedubs:
`toNested` was added in #1172, so I'm going to close this out.

PR #1172, opened 2016-07-01
Add toNested to OptionT and XorT.

peterneyens:
I have added `toNested` to `OptionT` and `XorT` like I mentioned in #1107.

This among other things makes it possible to use the "applicative `ap`" (called `app` in Scalaz) which is inconsistent the `ap` it gets from the monad instances of `OptionT` or `XorT` (see #1106).

```scala
import cats.implicits._
import cats.data.OptionT

val ff: OptionT[List, Int => String] = 
  OptionT(List(Option(_.toString), none[Int => String]))
val fa: OptionT[List, Int] = 
  OptionT(List(1.some, 2.some))

ff.ap(fa)
// OptionT[List,String] = OptionT(List(Some(1), Some(2), None))

OptionT(ff.toNested.ap(fa.toNested).value)
// OptionT[List,String] = OptionT(List(Some(1), Some(2), None, None))
```

The `toNested` for `XorT` is currently not that useful because we are not able to get the implicits for `Nested[F, A Xor ?, B]`, but it works with the SI-2712 fix : 

```scala
import cats.data.{Xor, XorT}
import cats.implicits._

val ff2: XorT[List, String, Int => String] = 
  XorT(List(Xor.right(_.toString), "error".left))
val fa2: XorT[List, String, Int] = 
  XorT(List(1.right, 2.right))

ff2.ap(fa2)
// XorT[List,String,String] = XorT(List(Right(1), Right(2), Left(error)))

XorT(ff2.toNested.ap(fa2.toNested).value)
// XorT[List,String,String] = XorT(List(Right(1), Right(2), Left(error), Left(error)))
```
---

In #1107 I also suggested that we could simplify some functions like `foldLeft` by using the one from `Nested`, but I benchmarked `optiont.foldLeft(0)(_ + _)` (the current implementation) vs `optiont.toNested.foldLeft(0)(_ + _)` and the one using `toNested` is quite a bit slower : 

```
[info] Benchmark                    Mode  Cnt         Score         Error  Units
[info] ToNestedBench.compose       thrpt   10  58633849.651 ±  646010.783  ops/s
[info] ToNestedBench.nestedDetour  thrpt   10  39380592.834 ± 4962640.124  ops/s
```

Even though a `foldLeft` implementation with `toNested` would like slightly nicer, I am not sure it's worth the performance hit ? (This doesn't feel that important, but if someone wants to voice their opinion ...)

```scala
F.compose(optionInstance).foldLeft(value, b)(f)
toNested.foldLeft(value, b)(f)
```
---

I am not sure which tests to add (I could check for consistency for `map`, `traverse`, ... ?)

### DIFF
--- a/core/src/main/scala/cats/data/OptionT.scala
+++ b/core/src/main/scala/cats/data/OptionT.scala
@@ -105,6 +105,25 @@ final case class OptionT[F[_], A](value: F[Option[A]]) {
   def foldRight[B](lb: Eval[B])(f: (A, Eval[B]) => Eval[B])(implicit F: Foldable[F]): Eval[B] =
     F.compose(optionInstance).foldRight(value, lb)(f)
 
+  /**
+   * Transform this `OptionT[F, A]` into a `[[Nested]][F, Option, A]`.
+   *
+   * An example where `toNested` can be used, is to get the `Apply.ap` function with the
+   * behavior from the composed `Apply` instances from `F` and `Option`, which is
+   * inconsistent with the behavior of the `ap` from `Monad` of `OptionT`.
+   *
+   * {{{
+   * scala> import cats.implicits._
+   * scala> import cats.data.OptionT
+   * scala> val ff: OptionT[List, Int => String] =
+   *      |   OptionT(List(Option(_.toString), None))
+   * scala> val fa: OptionT[List, Int] = OptionT(List(Option(1), Option(2)))
+   * scala> ff.ap(fa)
+   * res0: OptionT[List,String] = OptionT(List(Some(1), Some(2), None))
+   * scala> OptionT(ff.toNested.ap(fa.toNested).value)
+   * res1: OptionT[List,String] = OptionT(List(Some(1), Some(2), None, None))
+   * }}}
+   */
   def toNested: Nested[F, Option, A] = Nested(value)
 }
 
@@ -122,8 +141,8 @@ object OptionT extends OptionTInstances {
   /**
    * Transforms an `Option` into an `OptionT`, lifted into the specified `Applicative`.
    *
-   * Note: The return type is a FromOptionPartiallyApplied[F], which has an apply method
-   * on it, allowing you to call fromOption like this:
+   * Note: The return type is a `FromOptionPartiallyApplied[F]`, which has an apply method
+   * on it, allowing you to call `fromOption` like this:
    * {{{
    * scala> import cats.implicits._
    * scala> val o: Option[Int] = Some(2)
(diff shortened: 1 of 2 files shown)


ceedubs:
Nice @peterneyens. And thank you for the detailed explanation of the inconsistent `ap` and setting up a benchmark!

> I am not sure which tests to add (I could check for consistency for map, traverse, ... ?)

Since this is such a trivial method, I'm not really concerned about testing it. Maybe you could add a simple doctest example so the code coverage doesn't drop? Actually the examples you gave in this PR along with comments about potential `ap` inconsistencies would probably be great for the scaladoc.

peterneyens:
I have added the `ap` examples as ScalaDoc, the `XorT` one doesn't looks as nice because I had to work around SI-2712.

I am not sure which part of the diff codecov is complaining about ?

ceedubs:
👍 looks great to me. Thanks for the detailed examples @peterneyens.

> I am not sure which part of the diff codecov is complaining about ?

The "patch" code coverage stats often seem kind of bogus to me, though I'm sure this is just me not understanding what they represent. The overall project code coverage has gone up with this PR, so that's enough to make me happy.

ceedubs:
@adelbertc as our resident `Nested` expert, would you be willing to review this PR?

adelbertc:
LGTM 👍 Some questions:

1. Would we want to do this for other transformers like `StateT`, `IdT`, `WriterT`?
2. Should we add a `withNested` method similar to the `withValidated` method: https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/Xor.scala#L83 ?

peterneyens:
1. I think `OptionT` and `XorT` are probably the most useful.
 
  We could define `toNested` on `WriterT[F, L, V]` to give us `Nested[F, (L, ?), V]`, but we currently don't have instances for `Tuple2[A, ?]`. (I don' think `Nested[F, Writer[L, ?], V]` would make much sense ?)

  I don't know what `StateT[F, S, A].toNested` should give us ?

  `IdT[F, A]` to `Nested[F, Id, A]` would be easy, but I don't know how this could be useful (Though I have to admit that I have also never used `IdT`).

  `Kleisli[F, A, B].toNested` would be `Nested[A => ?, F, B]`.
2. I think `withNested` could be useful. The `ap` example above could then be `ff.withNested(_.ap(fa.toNested))`.

I am not sure if these `toNested` methods would be used a lot ?

ceedubs:
What's here looks really good to me and has a couple 👍 already, so I'm going to go ahead and merge it.

It sounds like there are some potential followups that could be addressed in later PRs. I agree with @peterneyens that `toNested` for the other transformers listed seems to be less useful. I could see `withNested` also being handy if you hit its use-case, but I'm not sure how often people will hit that -- either way is fine with me.

PR #1174, opened 2016-07-01
Adds a script to try cats in ammonite

sisirkoppaka:
Based on try-shapeless

### DIFF
--- a/scripts/try-cats.sh
+++ b/scripts/try-cats.sh
@@ -1,0 +1,9 @@
+#!/bin/sh
+COURSIER_URL=https://raw.githubusercontent.com/alexarchambault/coursier/v1.0.0-M12/coursier
+test -e ~/.coursier/cr || (mkdir -p ~/.coursier && curl -s --output ~/.coursier/cr $COURSIER_URL && chmod +x ~/.coursier/cr)
+CLASSPATH="$(~/.coursier/cr fetch -q -p \
+  \
+  org.typelevel:cats_2.11:0.6.0 \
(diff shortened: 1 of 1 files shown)


non:
Seems good to me. 👍

johnynek:
is there any way we can test that this is working in CI? Something like piping some input to it and then grepping the response might work. Things like this often bitrot because the core developers are not using it.

Also, if we add this, should we link to it in the readme and mention how it could be used (maybe with some horribly insecure curl pipe to bash?)

sisirkoppaka:
Sure, I can do that. Good suggestion @johnynek

kailuowang:
The fact that cats version is hardcoded here made me nervous. Trying cats in ammonite is actually quite easy 
```bash
brew install ammonite-repl
amm
```
then in ammonite-repl
```scala 
import $ivy.`org.typelevel::cats:0.7.2`, cats.implicits._

```
This is probably the quickist way to try cats in a repl. 
How about we just close this issue and add the above to the documentation (say FAQ? )

milessabin:
What is this "brew" of which you speak? ;-)

kailuowang:
@milessabin alright, this what @lihaoyi suggests
```
$ sudo curl -L -o /usr/local/bin/amm https://git.io/viYun && chmod +x /usr/local/bin/amm && amm
```

milessabin:
You lost me a `sudo`.

easel:
Looks like the preferred solution here is to document some preferred one-or-two-liners to the docs. Let's close out this PR in favor of that. Might not hurt to submit a PR to the ammonite docs as well.

kailuowang:

Issue #1177, opened 2016-07-04
Add Leibniz equality


tel:
This might fall a bit outside the standard "only the things you need" toolkit, but having a notion of Leibniz equality makes handling type equalities much easier and is useful for some pure functional idioms. Scalaz's implementation is more powerful to account for subtyping, but a simpler version might be the following

I also propose calling it something less scary than "Leibniz equality", so `Teq` seems reasonable.

``` scala
import scala.language.higherKinds
import scala.language.implicitConversions

/**
  * `Teq[A, B]` is a witness that the types `A` and `B` are equal. It is more
  * powerful than the standard `A =:= B` since it offers the ability to
  * convert types appearing within other structures.
  */
sealed abstract class Teq[A, B] {
  def subst[F[_]](fa: F[A]): F[B]

  final def apply(a: A): B =
    Teq.witness(this)(a)

  final def andThen[C](next: B Teq C): A Teq C =
    next.subst[A Teq ?](this)

  final def compose[C](prev: C Teq A): C Teq B =
    prev andThen this

  final def from: B Teq A =
    this.subst[? Teq A](Teq.refl)

  final def lift[F[_]]: F[A] Teq F[B] =
    subst[Lambda[X => F[A] Teq F[X]]](Teq.refl)
}

object Teq {

  /**
    * The only real value of `Teq` is the statement `A Teq A`.
    */
  implicit def refl[A]: A Teq A = new Teq[A, A] {
    def subst[F[_]](fa: F[A]): F[A] = fa
  }

  /**
    * A `Teq` immediately furnishes a coercion function.
    */
  implicit def witness[A, B](t: A Teq B): A => B =
    t.subst[A => ?](identity)

  /**
    * We can convert `A Teq B` to a `A =:= B` witness via substitution.
    */
  implicit def scalaEq[A, B](t: A Teq B): A =:= B =
    t.subst[A =:= ?](implicitly[A =:= A])

}
```

tpolecat:
I think both `Leibniz

tel:
I'm a little curious about Lizkov and the subtyping parameters around Leibniz.

Also, any thoughts on `TEq` as the name? The other option I thought about was `Is`.

adelbertc:
I've found these to be useful, :+1: 

As for naming, I would prefer `Leibniz` and `Liskov` if only to provide access to the general Google-able principle

tel:
[...]

PR #1178, opened 2016-07-05
Add Is constructor for Leibniz equality

tel:
(See #1177 for initial discussion)

This PR adds Leibniz type equality under the name `A Is B`. Leibniz equality is an improvement over `Predef.=:=` as it can be substituted into type expressions and as it has more computational meaning. Both of these add power and convenience to `Is` which makes it appealing for certain kinds of pure functional work. I am, for instance, using a local version of this in a [Cats-based lens library I'm working on](https://github.com/tel/scala-telescope) to implement a generalized `Equality[S, T, A, B]` optic at the base of the optical hierarchy. It also shows up nicely for making GADTs with more computational integrity.

This is my first Cats contribution so there's probably a lot missing—please let me know how to continue to develop it.
## Questions
- I've simplified the Scalaz version back to one they used previously which appears to not play as nicely with subtyping. I'm not sure if this is a good idea... but it's undeniable that this version is clearer.
- Is `Is` a good name? Also possible are `TEq`, `Leib`, and `Leibniz`. I like `Is` since it's very direct.
- How can I better document this?
- What is the right way to add tests?
- What methods or companion object functions are missing? Are there any I have right now that are suspect?
## Before we can land
- [X] Rename `cats.eq` to `cats.evidence`
- [ ] Decide about design to maximize allocation and inlining
- [ ] Determine whether or not to include `def fromPredef[A, B](eq: A =:= B): A Is B`
  - **Pros** greater convenience
  - **Cons** not strictly valid in a constructive sense; may be possible to generate invalid inferences?
- [ ] Add equality tests at the type level by porting some notion of Shapeless's `illtyped` macro
- [ ] Add basic tests showing design of implicits working without imports
- [ ] Add basic tests to show that methods don't throw
- [ ] Tests that we aren't reallocating
- [x] Decide what to call `from` ({`from`, `reverse`, `flip`, `mirror`, `symm`, `inverse`})
- [ ] Determine the appropriate `@inline` design
## Implicits spec
- [ ] `refl` is available whenever there is an implicit need for `A Is A` without extra imports

### DIFF
--- a/core/src/main/scala/cats/eq/Is.scala
+++ b/core/src/main/scala/cats/eq/Is.scala
@@ -1,5 +1,7 @@
 package cats.eq
 
+import cats.Id
+
 /**
   * A value of `A Is B` witnesses that the types `A` and `B` are the same.
   * You can use this value to coerce `A` into `B` and therefore `A Is B` is
@@ -30,6 +32,13 @@ abstract class Is[A, B] {
 
   final def lift[F[_]]: F[A] Is F[B] =
     subst[λ[α => F[A] Is F[α]]](Is.refl)
+
+  /**
+    * Substitution on identity brings about a direct coercion function of the
+    * same form that `=:=` provides.
+    */
+  def coerce: A => B =
+    subst[Id]
 }
 
 object Is {
@@ -43,13 +52,6 @@ object Is {
   implicit def refl[A]: A Is A = new Is[A, A] {
     def subst[F[_]](fa: F[A]): F[A] = fa
   }
-
-  /**
-    * In order to furnish `A Is B` values as their coercions we provide an
-    * implicit conversion to the coercion function.
-    */
-  implicit def witness[A, B](t: A Is B): A => B =
-    t.subst[A => ?](identity)
 
   /**
     * A value `A Is B` is always sufficient to produce a similar `Predef.=:=`
--- a/core/src/main/scala/cats/eq/Is.scala
+++ b/core/src/main/scala/cats/eq/Is.scala
@@ -3,61 +3,64 @@ package cats.eq
 import cats.Id
 
 /**
- * A value of `A Is B` witnesses that the types `A` and `B` are the same.
- * You can use this value to coerce `A` into `B` and therefore `A Is B` is
- * similar in feel to `A =:= B`; however, `A Is B` has more computational
- * content and offers greater functionality.
+ * A value of `A Is B` is proof that the types `A` and `B` are the same. More
+ * powerfully, it asserts that they have the same meaning in all type
+ * contexts. This can be a more powerful assertion than `A =:= B` and is more
+ * easily used in manipulation of types while avoiding (potentially
+ * erroneous) coercions.
  *
- * `A Is B` is also known as Leibniz equality and represents computationally
- * the idea that two things are equal if they take equal values on all
- * predicates. In other words, `A` equals `B` when for all predicates `P[_]`
- * we have that `P[A]` implies `P[B]` and visa versa. `A Is B` represents
- * this idea computationally providing exploitable structure to our equalities.
- *
- * The heart of `A Is B` is the `subst` method which converts values `F[A]`
- * into values `F[B]` for any `F[_]`. Creative choices of `F` enable type
- * equalities to be substituted into many more complex type expressions.
+ * `A Is B` is also known as Leibniz equality.
  */
 abstract class Is[A, B] {
   def subst[F[_]](fa: F[A]): F[B]
 
-  final def andThen[C](next: B Is C): A Is C =
+  @inline final def andThen[C](next: B Is C): A Is C =
     next.subst[A Is ?](this)
 
-  final def compose[C](prev: C Is A): C Is B =
+  @inline final def compose[C](prev: C Is A): C Is B =
     prev andThen this
 
-  final def from: B Is A =
+  @inline final def from: B Is A =
     this.subst[? Is A](Is.refl)
 
-  final def lift[F[_]]: F[A] Is F[B] =
+  @inline final def lift[F[_]]: F[A] Is F[B] =
     subst[λ[α => F[A] Is F[α]]](Is.refl)
 
   /**
    * Substitution on identity brings about a direct coercion function of the
    * same form that `=:=` provides.
    */
-  def coerce: A => B =
-    subst[Id]
+  @inline final def coerce(a: A): B =
+    subst[Id](a)
 }
 
 object Is {
+
+  /**
+   * In truth, "all values of `A Is B` are `refl`". `reflAny` is that
+   * single value.
+   */
+  private[this] val reflAny = new Is[Any, Any] {
+    def subst[F[_]](fa: F[Any]) = fa
+  }
 
   /**
    * In normal circumstances the only `Is` value which is available is the
    * computational identity `A Is A` at all types `A`. These "self loops"
    * generate all of the behavior of equality and also ensure that at its
    * heart `A Is B` is always just an identity relation.
+   *
+   * Implementation note: all values of `refl` return the same (private)
+   * instance at whatever type is appropriate to save on allocations.
    */
-  implicit def refl[A]: A Is A = new Is[A, A] {
-    def subst[F[_]](fa: F[A]): F[A] = fa
-  }
+  implicit def refl[A]: A Is A =
+    reflAny.asInstanceOf
 
   /**
    * A value `A Is B` is always sufficient to produce a similar `Predef.=:=`
    * value.
    */
-  implicit def scalaEq[A, B](t: A Is B): A =:= B =
+  implicit def predefEq[A, B](t: A Is B): A =:= B =
     t.subst[A =:= ?](implicitly[A =:= A])
 
 }
--- a/core/src/main/scala/cats/eq/Is.scala
+++ b/core/src/main/scala/cats/eq/Is.scala
@@ -3,21 +3,21 @@ package cats.eq
 import cats.Id
 
 /**
-  * A value of `A Is B` witnesses that the types `A` and `B` are the same.
-  * You can use this value to coerce `A` into `B` and therefore `A Is B` is
-  * similar in feel to `A =:= B`; however, `A Is B` has more computational
-  * content and offers greater functionality.
-  *
-  * `A Is B` is also known as Leibniz equality and represents computationally
-  * the idea that two things are equal if they take equal values on all
-  * predicates. In other words, `A` equals `B` when for all predicates `P[_]`
-  * we have that `P[A]` implies `P[B]` and visa versa. `A Is B` represents
-  * this idea computationally providing exploitable structure to our equalities.
-  *
-  * The heart of `A Is B` is the `subst` method which converts values `F[A]`
-  * into values `F[B]` for any `F[_]`. Creative choices of `F` enable type
-  * equalities to be substituted into many more complex type expressions.
-  */
+ * A value of `A Is B` witnesses that the types `A` and `B` are the same.
+ * You can use this value to coerce `A` into `B` and therefore `A Is B` is
+ * similar in feel to `A =:= B`; however, `A Is B` has more computational
+ * content and offers greater functionality.
+ *
+ * `A Is B` is also known as Leibniz equality and represents computationally
+ * the idea that two things are equal if they take equal values on all
+ * predicates. In other words, `A` equals `B` when for all predicates `P[_]`
+ * we have that `P[A]` implies `P[B]` and visa versa. `A Is B` represents
+ * this idea computationally providing exploitable structure to our equalities.
+ *
+ * The heart of `A Is B` is the `subst` method which converts values `F[A]`
+ * into values `F[B]` for any `F[_]`. Creative choices of `F` enable type
+ * equalities to be substituted into many more complex type expressions.
+ */
 abstract class Is[A, B] {
   def subst[F[_]](fa: F[A]): F[B]
 
@@ -34,9 +34,9 @@ abstract class Is[A, B] {
     subst[λ[α => F[A] Is F[α]]](Is.refl)
 
   /**
-    * Substitution on identity brings about a direct coercion function of the
-    * same form that `=:=` provides.
-    */
+   * Substitution on identity brings about a direct coercion function of the
+   * same form that `=:=` provides.
+   */
   def coerce: A => B =
     subst[Id]
 }
@@ -44,19 +44,19 @@ abstract class Is[A, B] {
 object Is {
 
   /**
-    * In normal circumstances the only `Is` value which is available is the
-    * computational identity `A Is A` at all types `A`. These "self loops"
-    * generate all of the behavior of equality and also ensure that at its
-    * heart `A Is B` is always just an identity relation.
-    */
+   * In normal circumstances the only `Is` value which is available is the
+   * computational identity `A Is A` at all types `A`. These "self loops"
+   * generate all of the behavior of equality and also ensure that at its
+   * heart `A Is B` is always just an identity relation.
+   */
   implicit def refl[A]: A Is A = new Is[A, A] {
     def subst[F[_]](fa: F[A]): F[A] = fa
   }
 
   /**
-    * A value `A Is B` is always sufficient to produce a similar `Predef.=:=`
-    * value.
-    */
+   * A value `A Is B` is always sufficient to produce a similar `Predef.=:=`
+   * value.
+   */
   implicit def scalaEq[A, B](t: A Is B): A =:= B =
     t.subst[A =:= ?](implicitly[A =:= A])
 
--- a/core/src/main/scala/cats/eq/Is.scala
+++ b/core/src/main/scala/cats/eq/Is.scala
@@ -61,3 +61,4 @@ object Is {
     t.subst[A =:= ?](implicitly[A =:= A])
 
 }
+
--- a/core/src/main/scala/cats/eq/Is.scala
+++ b/core/src/main/scala/cats/eq/Is.scala
@@ -1,0 +1,61 @@
+package cats.eq
+
+/**
+  * A value of `A Is B` witnesses that the types `A` and `B` are the same.
+  * You can use this value to coerce `A` into `B` and therefore `A Is B` is
+  * similar in feel to `A =:= B`; however, `A Is B` has more computational
+  * content and offers greater functionality.
+  *
+  * `A Is B` is also known as Leibniz equality and represents computationally
+  * the idea that two things are equal if they take equal values on all
+  * predicates. In other words, `A` equals `B` when for all predicates `P[_]`
+  * we have that `P[A]` implies `P[B]` and visa versa. `A Is B` represents
+  * this idea computationally providing exploitable structure to our equalities.
+  *
+  * The heart of `A Is B` is the `subst` method which converts values `F[A]`
+  * into values `F[B]` for any `F[_]`. Creative choices of `F` enable type
+  * equalities to be substituted into many more complex type expressions.
+  */
+abstract class Is[A, B] {
+  def subst[F[_]](fa: F[A]): F[B]
+
+  final def andThen[C](next: B Is C): A Is C =
+    next.subst[A Is ?](this)
+
+  final def compose[C](prev: C Is A): C Is B =
+    prev andThen this
+
+  final def from: B Is A =
+    this.subst[? Is A](Is.refl)
+
+  final def lift[F[_]]: F[A] Is F[B] =
+    subst[λ[α => F[A] Is F[α]]](Is.refl)
+}
+
+object Is {
+
+  /**
+    * In normal circumstances the only `Is` value which is available is the
+    * computational identity `A Is A` at all types `A`. These "self loops"
+    * generate all of the behavior of equality and also ensure that at its
+    * heart `A Is B` is always just an identity relation.
+    */
+  implicit def refl[A]: A Is A = new Is[A, A] {
+    def subst[F[_]](fa: F[A]): F[A] = fa
+  }
+
+  /**
+    * In order to furnish `A Is B` values as their coercions we provide an
+    * implicit conversion to the coercion function.
+    */
+  implicit def witness[A, B](t: A Is B): A => B =
+    t.subst[A => ?](identity)
+
+  /**
+    * A value `A Is B` is always sufficient to produce a similar `Predef.=:=`
+    * value.
+    */
+  implicit def scalaEq[A, B](t: A Is B): A =:= B =
+    t.subst[A =:= ?](implicitly[A =:= A])
+
+}
--- a/core/src/main/scala/cats/evidence/Is.scala
+++ b/core/src/main/scala/cats/evidence/Is.scala
@@ -89,9 +89,9 @@ object Is {
 
   /**
    * It can be convenient to convert a `Predef.=:=` value into an `Is` value.
-   * This is not strictly valid as while it is almost certainly true that `A
-   * =:= B` implies `A Is B` it is not the case that you can create evidence
-   * of `A Is B` except via a coercion. Use responsibly.
+   * This is not strictly valid as while it is almost certainly true that
+   * `A =:= B` implies `A Is B` it is not the case that you can create
+   * evidence of `A Is B` except via a coercion. Use responsibly.
    */
   @inline def unsafeFromPredef[A, B](eq: A =:= B): A Is B =
     reflAny.asInstanceOf[A Is B]
--- a/core/src/main/scala/cats/evidence/Is.scala
+++ b/core/src/main/scala/cats/evidence/Is.scala
@@ -1,4 +1,4 @@
-package cats.eq
+package cats.evidence
 
 import cats.Id
 
--- a/core/src/main/scala/cats/evidence/Is.scala
+++ b/core/src/main/scala/cats/evidence/Is.scala
@@ -71,11 +71,11 @@ object Is {
     reflAny.asInstanceOf[A Is A]
 
   /**
-    * It can be convenient to convert a `Predef.=:=` value into an `Is` value.
-    * This is not strictly valid as while it is almost certainly true that `A
-    * =:= B` implies `A Is B` it is not the case that you can create evidence
-    * of `A Is B` except via a coercion. Use responsibly.
-    */
+   * It can be convenient to convert a `Predef.=:=` value into an `Is` value.
+   * This is not strictly valid as while it is almost certainly true that `A
+   * =:= B` implies `A Is B` it is not the case that you can create evidence
+   * of `A Is B` except via a coercion. Use responsibly.
+   */
   def unsafeFromPredef[A, B](eq: A =:= B): A Is B =
     reflAny.asInstanceOf[A Is B]
 
--- a/core/src/main/scala/cats/evidence/Is.scala
+++ b/core/src/main/scala/cats/evidence/Is.scala
@@ -63,5 +63,14 @@ object Is {
   implicit def refl[A]: A Is A =
     reflAny.asInstanceOf[A Is A]
 
+  /**
+    * It can be convenient to convert a `Predef.=:=` value into an `Is` value.
+    * This is not strictly valid as while it is almost certainly true that `A
+    * =:= B` implies `A Is B` it is not the case that you can create evidence
+    * of `A Is B` except via a coercion. Use responsibly.
+    */
+  def unsafeFromPredef[A, B](eq: A =:= B): A Is B =
+    reflAny.asInstanceOf[A Is B]
+
 }
 
--- a/core/src/main/scala/cats/evidence/Is.scala
+++ b/core/src/main/scala/cats/evidence/Is.scala
@@ -21,15 +21,32 @@ abstract class Is[A, B] extends Serializable {
    */
   def substitute[F[_]](fa: F[A]): F[B]
 
+  /**
+    * `Is` is transitive and therefore values of `Is` can be composed in a
+    * chain much like functions. See also `compose`.
+    */
   @inline final def andThen[C](next: B Is C): A Is C =
     next.substitute[A Is ?](this)
 
+  /**
+    * `Is` is transitive and therefore values of `Is` can be composed in a
+    * chain much like functions. See also `andThen`.
+    */
   @inline final def compose[C](prev: C Is A): C Is B =
     prev andThen this
 
-  @inline final def from: B Is A =
+  /**
+    * `Is` is symmetric and therefore can be flipped around. Flipping is its
+    * own inverse, so `x.flip.flip == x`.
+    */
+  @inline final def flip: B Is A =
     this.substitute[? Is A](Is.refl)
 
+  /**
+    * Sometimes for more complex substitutions it helps the typechecker to
+    * wrap one layer of `F[_]` context around the types you're equating
+    * before substitution.
+    */
   @inline final def lift[F[_]]: F[A] Is F[B] =
     substitute[λ[α => F[A] Is F[α]]](Is.refl)
 
--- a/core/src/main/scala/cats/evidence/Is.scala
+++ b/core/src/main/scala/cats/evidence/Is.scala
@@ -12,33 +12,40 @@ import cats.Id
  * `A Is B` is also known as Leibniz equality.
  */
 abstract class Is[A, B] extends Serializable {
-  def subst[F[_]](fa: F[A]): F[B]
+
+  /**
+   * To create an instance of `A Is B` you must show that for every choice
+   * of `F[_]` you can convert `F[A]` to `F[B]`. Loosely, this reads as
+   * saying that `B` must have the same effect as `A` in all contexts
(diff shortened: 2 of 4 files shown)


core/src/main/scala/cats/eq/Is.scala
```diff
@@ -0,0 +1,61 @@
...
+  implicit def refl[A]: A Is A = new Is[A, A] {
+    def subst[F[_]](fa: F[A]): F[A] = fa
+  }
+
+  /**
+    * In order to furnish `A Is B` values as their coercions we provide an
+    * implicit conversion to the coercion function.
+    */
+  implicit def witness[A, B](t: A Is B): A => B =
+    t.subst[A => ?](identity)
```
adelbertc: Hm I don't know about this one as it might bring in implicit conversions that may be undesired.. perhaps turn this into a method on the type class itself?

core/src/main/scala/cats/eq/Is.scala
```diff
@@ -0,0 +1,61 @@
...
+  * the idea that two things are equal if they take equal values on all
+  * predicates. In other words, `A` equals `B` when for all predicates `P[_]`
+  * we have that `P[A]` implies `P[B]` and visa versa. `A Is B` represents
+  * this idea computationally providing exploitable structure to our equalities.
+  *
+  * The heart of `A Is B` is the `subst` method which converts values `F[A]`
+  * into values `F[B]` for any `F[_]`. Creative choices of `F` enable type
+  * equalities to be substituted into many more complex type expressions.
+  */
+abstract class Is[A, B] {
```
adelbertc: Usually these are `trait`s but not sure if this is a special case since it's a proof witness rather than a type class.. though I guess the two are the same.

core/src/main/scala/cats/eq/Is.scala
```diff
@@ -0,0 +1,61 @@
...
+    * implicit conversion to the coercion function.
+    */
+  implicit def witness[A, B](t: A Is B): A => B =
+    t.subst[A => ?](identity)
+
+  /**
+    * A value `A Is B` is always sufficient to produce a similar `Predef.=:=`
+    * value.
+    */
+  implicit def scalaEq[A, B](t: A Is B): A =:= B =
```
adelbertc: Is the other direction valid?

tel: Might be a point of contention. I don't think you can write `(A =:= B) => A Is B` function without coercion, but it might not be illegal.

core/src/main/scala/cats/eq/Is.scala
tel: I felt this route was justified because it's unlikely you want to extend let alone mix in `Is` values. I left it unsealed because there are times you want to instantiate `Is` directly... though even that is a little troubling.

OTOH, I'd definitely consider a trait implementation. I'm still not sure what all of the tradeoffs between trait and abstract class are at the end of the day.

core/src/main/scala/cats/eq/Is.scala
tel: Sounds good. I'll update this presently.

core/src/main/scala/cats/eq/Is.scala
non: I agree that `abstract class` is appropriate in this situation.

non:
Minor formatting nit: can you reformat your comments like this?

```scala
/**
 * We prefer the Java-style comment syntax.
 */
```

core/src/main/scala/cats/eq/Is.scala
non: I am happy to leave `(A =:= B) => A Is B` to folks that need it in a particular circumstance (and can convince themselves it is valid).

non: Also, if we want to seamlessly bridge `X Is Y` and `X =:= Y` we probably need to another implicit someplace else in implicit scope. I feel like we might want something like this in `cats.implicits._` somewhere:

```scala
def catsIsProvidesScalaEq[A, B](implicit t: A Is B): A =:= B = ...
```

tel:
Formatting fixed! :+1:

core/src/main/scala/cats/eq/Is.scala
tel: I am not sure if bidirectional implicits make as much sense to me. I'm in favor of the idea that we allow for `(A =:= B) => (A Is B)` as a backdoor, but that seems like something that should be done only with trepidation.

On a similar note, is there an "`unsafe*`" naming convention for Cats? I'd propose `unsafeFromPredef(eq: A =:= B): A Is B` perhaps?

core/src/main/scala/cats/eq/Is.scala
tel: 👍

tel:
Some more questions:

- I placed this in an `eq` package with expectation that someone will later want to add Liskov—but this might be premature.

kailuowang:
Thanks very much @tel, I had a discussion with @milessabin at gitter IRT testing such code in cats
For your reference
https://gitter.im/typelevel/cats?at=577e6d647aeb0805276ec94e

non:
This looks good to me. Thanks!

One question: how do we feel about a `cats.eq` package? I have to admit I'm nervous that it's a bit too specialized. Would something like `cats.witness` or `cats.evidence` work instead? I can imagine more types living in those packages.

tel:
@kailuowang Great, thanks! I'll take a look and make some plans to investigate this again perhaps this weekend.

@non I immediately agree `witness`, `evidence`, `proof` feels nicer than `eq` specifically. Of those, I think `evidence` is perhaps the best: `proof` feels too heady/academic, `witness` makes me think of the actual to and fro functions in an isomorphism—it feels specialized. Then, `evidence` feels both general and already shows up sometimes as the name of implicit arguments. I think that's appropriate. So color my bikeshed red.

tel:
Just to copy some of Miles' comments into this thread:

> @kailuowang: looks like we didn’t test the negative cases. I was thinking how do we test those without using shapeless’ illtyped

> @milessabin Typically you test by allowing a type to be inferred and then checking it by ascribing.
> You could have a test dependency on shapeless for that. Or copy and paste the implementation into the Cats tests.
> ...  it's so small it's probably less effort to have a local copy.

non:
@tel great! thanks for tolerating my bike shed commentary ;)

core/src/main/scala/cats/eq/Is.scala
```diff
@@ -0,0 +1,64 @@
...
+ * predicates. In other words, `A` equals `B` when for all predicates `P[_]`
+ * we have that `P[A]` implies `P[B]` and visa versa. `A Is B` represents
+ * this idea computationally providing exploitable structure to our equalities.
+ *
+ * The heart of `A Is B` is the `subst` method which converts values `F[A]`
+ * into values `F[B]` for any `F[_]`. Creative choices of `F` enable type
+ * equalities to be substituted into many more complex type expressions.
+ */
+abstract class Is[A, B] {
+  def subst[F[_]](fa: F[A]): F[B]
```
johnynek: bikeshed: could we call this `substitute` or something? I don't think it is worth saving a few chars personally.

core/src/main/scala/cats/eq/Is.scala
```diff
@@ -0,0 +1,64 @@
...
+
+object Is {
+
+  /**
+   * In normal circumstances the only `Is` value which is available is the
+   * computational identity `A Is A` at all types `A`. These "self loops"
+   * generate all of the behavior of equality and also ensure that at its
+   * heart `A Is B` is always just an identity relation.
+   */
+  implicit def refl[A]: A Is A = new Is[A, A] {
```
johnynek: couldn't we keep something like:
```scala
private[this] val reflAny = new Is[Any, Any] { ... }

// Due to erasure, this cast is safe:
implicit def refl[A]: A Is A = reflAny.asInstanceOf[A Is A]
```
seems like saving the allocation is worth it.

core/src/main/scala/cats/eq/Is.scala
```diff
@@ -0,0 +1,64 @@
+package cats.eq
+
+import cats.Id
+
+/**
+ * A value of `A Is B` witnesses that the types `A` and `B` are the same.
+ * You can use this value to coerce `A` into `B` and therefore `A Is B` is
+ * similar in feel to `A =:= B`; however, `A Is B` has more computational
```
johnynek: can you expand on this? I don't see what, other than additional methods, this class has that is different from `=:=`. Semantically the meaning seems precisely the same to me.

core/src/main/scala/cats/eq/Is.scala
```diff
@@ -0,0 +1,64 @@
...
+    def subst[F[_]](fa: F[A]): F[A] = fa
+  }
+
+  /**
+   * A value `A Is B` is always sufficient to produce a similar `Predef.=:=`
+   * value.
+   */
+  implicit def scalaEq[A, B](t: A Is B): A =:= B =
+    t.subst[A =:= ?](implicitly[A =:= A])
+
```
johnynek: why can't we go the opposite way too? `A =:= B => A Is B`?

core/src/main/scala/cats/eq/Is.scala
johnynek: put another way, why is it not true that if `A =:= B` then `F[A] =:= F[B]`?

core/src/main/scala/cats/eq/Is.scala
```diff
@@ -0,0 +1,64 @@
...
+    this.subst[? Is A](Is.refl)
+
+  final def lift[F[_]]: F[A] Is F[B] =
+    subst[λ[α => F[A] Is F[α]]](Is.refl)
+
+  /**
+   * Substitution on identity brings about a direct coercion function of the
+   * same form that `=:=` provides.
+   */
+  def coerce: A => B =
```
johnynek: why return a function here rather than `def coerce(a: A): B = subst[Id](a)` seems returning a function always has an allocation here, whereas the method could hopefully be inlined away completely (blind faith in the JIT).

core/src/main/scala/cats/eq/Is.scala
```diff
@@ -0,0 +1,64 @@
...
+ * the idea that two things are equal if they take equal values on all
+ * predicates. In other words, `A` equals `B` when for all predicates `P[_]`
+ * we have that `P[A]` implies `P[B]` and visa versa. `A Is B` represents
+ * this idea computationally providing exploitable structure to our equalities.
+ *
+ * The heart of `A Is B` is the `subst` method which converts values `F[A]`
+ * into values `F[B]` for any `F[_]`. Creative choices of `F` enable type
+ * equalities to be substituted into many more complex type expressions.
+ */
+abstract class Is[A, B] {
```
johnynek: should this not be `sealed`? Is there any case where you should ever make a new subclass of this outside of this file?

core/src/main/scala/cats/eq/Is.scala
johnynek: since many of the methods use `refl` it seems like saving the allocation may make almost all methods allocation-free.

johnynek: could we call this `def reflexive[A]` or something, or maybe even `reflexiveIs`. Short names on implicit defs can cause issues (though maybe not in this case due to importing being unlikely)

johnynek:
Can we have a simple test just to exercise a few things here:

1. implicit resolution works as expected without imports
2. each method doesn't do anything terrible like throw.
3. if `A Is B` then `subst(x) eq x` (check that we are never reallocating).

core/src/main/scala/cats/eq/Is.scala
tel: I'm fine with that personally. I defer to Cats House Style whatever that is.

core/src/main/scala/cats/eq/Is.scala
tel: - `refl` is a bit traditional—I'm sort of in favor of keeping it and generally agree that it's unlikely you'll import this without knowing what `refl` means. I'm curious what issues short names on implicits cause though.

- I defer to those more familiar with the JVM than I as to whether this allocation savings is worth it. In a certain light, literally every value of `Is` actually *is* `refl` at different types so there's something a little appealing about this.

core/src/main/scala/cats/eq/Is.scala
tel: In principle, `A =:= B` is weaker since it's only powerful enough to solve `F[A] => F[B]` for `F = Id`. I'm not offhand aware of any way to abuse this strengthening to cause issues but I'm also not sure of how you might show it's impossible.

core/src/main/scala/cats/eq/Is.scala
tel: I'm not familiar enough with the JIT to necessarily make great decisions for this sort of thing. I essentially did it because I was using it in places where `x.coerce` and `x.from.coerce` were useful arguments and I preferred to leave the trailing underscore off.

If fewer allocations is preferred and the JIT can hit this though then I see no reason to not def `coerce` as a method.

core/src/main/scala/cats/eq/Is.scala
tel: I'm actually doing just that in a pure profunctors lens lib that drove the need/desire for this: https://github.com/tel/scala-telescope/blob/master/src/main/scala-2.11/jspha/telescope/Equality.scala#L12 (I called it `A Teq B` there).

I'm not sure that it's exactly perfect justification to just say "I needed it" but unlike `Predef`s `=:=` it's actually (sort of, I think) ok to let users define values of `Is` since furnishing the implementation of `[F](f: F[A]): F[B]` is sufficient to ensure no cheating so long as nobody really, really subverts the type system.

core/src/main/scala/cats/eq/Is.scala
tel: It's _true_ that `A =:= B` ought to be enough to say that `F[A] =:= F[B]` but there's no internal mechanism to furnish proof of that. We have to believe this assertion by using knowledge and process _about_ Scala instead of _within_ Scala. This makes it harder to do type shenanigans which require notions of equality unless you're willing to do a lot of  hand-validated `asInstanceOf` calls.

So, the other way of saying this is to say that `A =:= B` produces two functions `f: A => B` and `g: B => A` such that `x = f(g(x))` for all `x`. This is not enough to get `F[A] =:= F[B]` directly even just for some `F`s since there's no way to use `f` and `g` to produce the required witnesses for just any `F` we can dream up. Immediately, `F` might be abstract and then we're done.

On the other hand, `A Is B` can trivially prove `A =:= B` (as shown) because having `to[F](f: F[A]): F[B]` and `fro[F](f: F[B]): F[A]` easily recovers `f = to[Id]` and `g = fro[Id]`. Generally this power is magnified with type lambdas (and heavily simplified with kind projector) so let us do what we sort of empirically want: use equalities to substitute out type variables in arbitrary type expressions.

So it's reasonably crucial to use `Is` to work with type equalities unless one wants to fall to coercions (which may be erroneous).

tel: I'm not precisely clear how to add that to the documentation appropriately. Guidance very welcome!

tel:
@johnynek I think a good spec around what allocations are acceptable (if any) is good. I added some todo items before landing including working with those. That said, I don't understand what you're trying to say with your example test. Could you rephrase?

tel:
@johnynek I also changed the docs on the class—let me know if you feel that's any better and, as necessary, how I can keep improving it.

core/src/main/scala/cats/eq/Is.scala
```diff
@@ -3,61 +3,64 @@ package cats.eq
...
    * heart `A Is B` is always just an identity relation.
+   *
+   * Implementation note: all values of `refl` return the same (private)
+   * instance at whatever type is appropriate to save on allocations.
    */
-  implicit def refl[A]: A Is A = new Is[A, A] {
-    def subst[F[_]](fa: F[A]): F[A] = fa
-  }
+  implicit def refl[A]: A Is A =
+    reflAny.asInstanceOf
```
johnynek: I think scala may infer `Nothing` as the type parameter here and you may need to instead write: `reflAny.asInstanceOf[A Is A]`

tel: Thanks! Fixed!

core/src/main/scala/cats/eq/Is.scala
```diff
@@ -3,61 +3,64 @@ package cats.eq
...
 
   /**
    * A value `A Is B` is always sufficient to produce a similar `Predef.=:=`
    * value.
    */
-  implicit def scalaEq[A, B](t: A Is B): A =:= B =
+  implicit def predefEq[A, B](t: A Is B): A =:= B =
```
johnynek: I am pretty sure this will need to be imported to use it. Do we want this to be where we have that import?

The scala implicit name issue is that you can't have two imported implicits with the same name. So:
```scala
object A {
  implicit val foo: Int = 0
}
object B {
  implicit val foo: String = "foo"
}

import A._
import B._ // now the implicit[Int] cannot be found
```
So, I think we want something like `def catsTypeEqualsFromIs` or something.

Secondly, I think an implicit conversion is only used on a call that fails to find a method. So I think we want:
```scala
implicit def catsTypeEqualsFromIs[A, B](implicit t: A Is B): A =:= B =
  t.subst[A =:= ?](implicitly[A =:= A])
```

johnynek:
What about adding this method to object:
```scala
def fromScala[A, B](implicit ev: A =:= B): A Is B =
  reflAny.asInstanceOf[A Is B]
```
I can't see how this cast is unsafe, though it is a bit ugly. Can you think of reasons (aside from casting distaste) not to add this? It cannot be unsound, right?

The reason I want this, I do sometimes have `A =:= B` instances and would like to use this more powerful tool when I only have access to `A =:= B`.

core/src/main/scala/cats/evidence/Is.scala
```diff
@@ -0,0 +1,67 @@
...
+abstract class Is[A, B] {
+  def subst[F[_]](fa: F[A]): F[B]
+
+  @inline final def andThen[C](next: B Is C): A Is C =
+    next.subst[A Is ?](this)
+
+  @inline final def compose[C](prev: C Is A): C Is B =
+    prev andThen this
+
+  @inline final def from: B Is A =
```
johnynek: what about calling this `reverse`?

core/src/main/scala/cats/evidence/Is.scala
```diff
@@ -0,0 +1,67 @@
...
+  @inline final def lift[F[_]]: F[A] Is F[B] =
+    subst[λ[α => F[A] Is F[α]]](Is.refl)
+
+  /**
+   * Substitution on identity brings about a direct coercion function of the
+   * same form that `=:=` provides.
+   */
+  @inline final def coerce(a: A): B =
+    subst[Id](a)
+}
```
johnynek: what about adding
```scala
def toScalaTypeEq: A =:= B = subst[A =:= ?](implicitly[A =:= A])
``` and then use this method below?

File: core/src/main/scala/cats/evidence/Is.scala [hunk 14]
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * A value of `A Is B` is proof that the types `A` and `B` are the same. More
+ * powerfully, it asserts that they have the same meaning in all type
+ * contexts. This can be a more powerful assertion than `A =:= B` and is more
+ * easily used in manipulation of types while avoiding (potentially
+ * erroneous) coercions.
+ *
+ * `A Is B` is also known as Leibniz equality.
+ */
+abstract class Is[A, B] {
```
johnynek: we usually make these extend `java.io.Serializable` since they are sometimes swept up in serializations on spark or scalding.

core/src/main/scala/cats/eq/Is.scala
tel: The other option now that this is getting more complex is just to ask the user to explicitly make the conversion. We could just add `toPredef` as a method!

I think increasingly that solution is the one I feel most strongly toward.

core/src/main/scala/cats/evidence/Is.scala
tel: I'm not a huge fan of `from`, but I was taking it from some other sources. Calling it `reverse` feels alright since underlying `A Is B` is an actual process from `A` to `B`, but I sort of want to wipe that notion away. It's incidental instead of important. So, in order of my preference I like:

* `flip`
* `mirror`
* `reverse`
* `from`

But any of the above are acceptable to me.

core/src/main/scala/cats/evidence/Is.scala
tel: I like this! See also above note on the implicit.

core/src/main/scala/cats/eq/Is.scala
tel: More than that, it's probably not terrible to add the implicit later if we give it a name like the one your proposed, right? So this is a reversible choice.

core/src/main/scala/cats/evidence/Is.scala
tel: Will do!

tel:
Another question with respect to performance: I've been recently sort of liberally decorating things with `@inline` but I have little understanding as to when is a good time to ask for inlining or not. Since ideally most of this code ought to just boil away I think aggressive `@inline`ing is probably correct, but I'd like the opinion of someone more familiar with JVM/Scala inlining.

core/src/main/scala/cats/evidence/Is.scala
johnynek: I like `flip`.

tel: Just changed it actually!

tel: Just renamed it!

johnynek:
What do you think of these:
```scala
  // A <: B
  sealed abstract class Sub[A, +B] {
    def co[F[+_]](f: F[A]): F[B]
    def contra[F[-_]](f: F[B]): F[A]
    def andThen[B1 >: B, C](that: Sub[B1, C]): Sub[A, C] =
      that.co[({type G[+T]=Sub[A, T]})#G](this)
  }
  object Sub {
    implicit def reflS[B, A <: B]: Sub[A, B] = new Sub[A, B] {
      def co[F[+_]](f: F[A]): F[B] = f
      def contra[F[-_]](f: F[B]): F[A] = f
    }

    def toSuper[A, B](s: Sub[A, B]): Super[B, A] =
      s.contra[({type G[-T]=Super[B, T]})#G](Super.reflSuper[B, B])
  }
  // A >: B
  sealed abstract class Super[A, -B] {
    def contra[F[-_]](f: F[A]): F[B]
    def co[F[+_]](f: F[B]): F[A]
    def andThen[B1 <: B, C](that: Super[B1, C]): Super[A, C] =
      that.contra[({type G[-T]=Super[A, T]})#G](this)
  }
  object Super {
    implicit def reflSuper[B, A >: B]: Super[A, B] = new Super[A, B] {
      def contra[F[-_]](f: F[A]): F[B] = f
      def co[F[+_]](f: F[B]): F[A] = f
    }
    def toSub[A, B](s: Super[A, B]): Sub[B, A] =
      s.co[({type G[+T]=Sub[B, T]})#G](Sub.reflS[B, B])
  }
```
which seem like the analogous versions of this for subtyping/supertyping.

tel:
There's [`Liskov` from `scalaz`](https://github.com/scalaz/scalaz/blob/series/7.3.x/core/src/main/scala/scalaz/Liskov.scala) that handles both of those ideas at once. Amusingly, it looks like Scala is smart enough to naturally have `B >: A` lift up to `[F]: F[B] => F[A]`

Here's a simplified version:

```scala
abstract class IsA[-A, +B] {
  def substitute[F[-_]](fb: F[B]): F[A]
}

object Sub {

  // principal basis
  def isa[A, B >: A]: A IsA B = new Isa[A, B] {
    def substitute[F[-_]](fb: F[B]) = F[A]
  }
}
```

Generally I'm not familiar enough with subtyping in a language like Scala to talk about the value of Liskov, but I'm sure others would know more.

tel:
The build error appears to be

```
[error] /home/travis/build/typelevel/cats/core/src/main/scala/cats/evidence/Is.scala:90: no additional content on same line after heading
[error]   /**
[error]   ^
```

Anyone know what's up with that?

ceedubs:
Thanks for submitting this and quickly addressing feedback, @tel.

I think that the inability to show the relationship between `F[A]` and `F[B]` is a compelling improvement upon the standard library's `=:=`. Having said that, in its current form this is completely standalone. Are there places that having `Is` might improve other parts of Cats? If it is mostly a standalone utility, then I wonder if, for the purposes of modularity and stable versioning, it might be better off as its own micro-library. WDYT?

non:
@tel I think the error may be caused by having a *backticks expression* getting split across lines. I tried reformatting your comment and then the build worked for me.

tel:
@ceedubs That's a good question. Right now it is totally standalone, I agree, but it's a useful part of a toolkit for type-level work and it could be argued that having it around is meaningful so that any parts of `cats` which intend to accept or return equality proofs can use `Is`.

A compelling place that it is used in `scalaz` is that it's an element of the `Unapply` trait since at the end of the day `Unapply` is all about showing the equality between two types. It also shows up, e.g., in the Syntax of `sequence` in `Traverse` and a few others as a mechanism for constraining the type parameter as needed.

So, there is some chance that it's useful in `cats`, but I definitely see the argument that it's reasonably standalone and could be externalized. If `cats` needs or wants it then the lib could just pull it in.

The only other thing to consider is whether `Is` wants some instances from `cats`. There's no necessarily a large need for this, but `Is` definitely forms a `Category` so if `cats` eventually has a `Category` trait and it's desirable to have that instance then finally we're in a little bit of a pickle.

I think generally, the question comes down to whether `cats` is trying to be more batteries included or is instead aiming for maximal compartmentalization of the libraries. I'll point to the Haskell ecosystem (or really the Kmett-verse within it, excluding `lens` which is an outlier) as an example of the latter.

tel:
@non thanks, that seems to do it

tel:
fwiw, there seems to be a similar debate over in #1184

javierfs89:
I found very useful this functionality but missed a `lift2` method ~> tel#1

core/src/main/scala/cats/evidence/Is.scala
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * A value of `A Is B` is proof that the types `A` and `B` are the same. More
+ * powerfully, it asserts that they have the same meaning in all type
+ * contexts. This can be a more powerful assertion than `A =:= B` and is more
+ * easily used in manipulation of types while avoiding (potentially
+ * erroneous) coercions.
+ *
+ * `A Is B` is also known as Leibniz equality.
+ */
+abstract class Is[A, B] extends Serializable {
```
Atry: I think `Is` is a better name as a replacement of `<:<`.

Did you think about the other names, like `Same`, for this type class?

stew:
There was a ton of activity and a ton of things addressed. Is there anything left holding this one up?

as for @ceedubs' question of whether or not it fits here, I feel like I'd like to err on the side of having it here for now, since people keep looking for it here, at least until we have a better place to point them to.

adelbertc:
I would prefer this be called `Leibniz` for the same reasons we use `Monoid` and not `Summable`

johnynek:
Can you say more? I think a compelling reason not to call Monoid Summable
is because it overlooks the identity/empty element and also many Monoids
are not at all sums. I don't think that's what you mean. Can you spell it
out for me in a paragraph or two?

That said, a more intuitive name is generally a better one, and Leibniz is
just a black box pointer into the history books. I care about adoption, and
I think adoption is improved when we lower the investment in learning.

For instance if we have Liebniz and Liskov, people have to memorize which
one is for subtyping. But if we have SameType and SubType or something, we
don't have that issue.

tel:
I'm not really one way or another about the name, and I totally understand and empathize with @johnynek's argument. Let me try to toss a point in the direction of `Leibniz`, though.

Similar to `Monoid` versus `Summable`, there's a good argument to be made in favor of "a black box pointer into the history books". Be being obscure it enables abstractness and puts a bookmark on the idea that `A SameType B` is something a little more than the intuitive reading of "some kind of property where A and B are the same".

Type equality is an equality relationship on type expressions generated by a type equality on type variables in the local context. "Leibniz equality" gives a particular and powerful form of this equality and, in particular, it's one that's more powerful than `Predef.=:=`. This makes it a much more physical thing with more properties than the handwavey intuitive reading above. Having a black box name gives space for that more sophisticated view.

So that's the best I can do in favor of Leibniz. Now let me tell you why I didn't put that in the PR:

- I agree that adoption is pretty important
- `A Is B` is short and sweet unlike `Leibniz[A, B]`—more pleasing to my eye at least
- This whole equality thing is much less likely than `Monoid` to end up in code that isn't "obviously a hairy ball of type madness". It almost always lives beyond the police tape, so more expertise can be assumed and a cute shorthand pays its way by being cute without as much downside for being subtle.

Or, in short, I like it for the same reason that when I write mathematics I'm completely happy with the rapid notational punning and infidelity that everyone uses. When it's by experts for experts then fast intuition holds water.

But then, finally, I don't think I actually care too much if it ends up as Leibniz. The delta here feels really small. If I were the final decision maker, I'd do it with a coin flip.

adelbertc:
My argument in favor of `Leibniz` is that like `Monoid` (perhaps a better example would've been `Monad`) we're trying to achieve something very specific here. `Leibniz` equality is a specific thing and is perfectly Google-able, and I think calling it that robs people of the opportunity of reaching into a rich history of existing literature around it.

I also think at the point where you need to wonder what `Is`/`Leibniz` is doing you're probably in a position to understand what it actually is.

On a related note, I could see a `Liskov`-esque proof type making its way into Cats and as I understand it much of OOP revolves around the Liskov substitution principle, which is another body of rich work. Calling it say, `SubtypeOf` may imply it's just something Cats invented to get Scala to do what we want, which isn't to say that's not true, as opposed to something more principled.

johnynek:
I hear you.

For `Monad` I can't think of a good description, personally. It is very abstract. For `Liebniz` I think we call can. We can add in the comments links to where to read more, but I'm not crazy about every deep principle, of which there are many, using a possibly obtuse name to encourage users to learn more. From my experience at large-ish companies, that does not help me sell scala as an appealing method to develop software, and like I said, that is really important to me.

My vote in this case is to keep `Is` or `Same` or something like that and we should feel free to link to as many interesting bodies of work for those that need to learn more.

PS: I since opened this PR to scala:
https://github.com/scala/scala/pull/5623

which I hope to have merged in 2.13, so this distinction with `=:=` and `<:<` will go away then. Lastly, I can't see any example where `=:=` is actually weaker than `Is[A, B]` or `Liebniz[A, B]` just that it lacks a method to prove that in the type system. Are there legitimate instances of `A =:= B` where it would have been impossible to construct `Is[A, B]`? Since we can only make `=:=` with `refl`, and it could have had `def sub[F[_]](f: F[A]): F[A] = f`, it seems to me that implies that in principle it is just as strong, just that the compiler can't see it.

non:
My vote is that we use both, and alias one to the other. I don't think there's a reason to forbid or force people to use either name, and unlike some other types that we have chosen not to do this for, there is literally only one instance with one method so I don't think there is a particularly high documentation burden created by supporting both.

adelbertc:
Wow, re-reading my previous comment, talk about terrible grammar. Guess that's when happens when I attempt to reply to things while amped on coffee.

> Are there legitimate instances of A =:= B where it would have been impossible to construct Is[A, B]

That's an interesting question, I'll have to think about that some more. I know S11 has written extensively on related topics:

* [A function from type equality to Leibniz](http://typelevel.org/blog/2014/07/02/type_equality_to_leibniz.html)
* [Higher Leibniz](http://typelevel.org/blog/2014/09/20/higher_leibniz.html)

> My vote is that we use both, and alias one to the other.

Not surprisingly my vote would be to use `Leibniz` for the actual definition and alias `Is`.

non:
@tel Would you be willing to add a type alias in `cats/evidence/package.scala`? Something like:

```scala
package cats
package object evidence {
  type Leibniz[A, B] = cats.evidence.Is[A, B]
}
```

tel:
@non Can do. In terms of moving to merger, should I rebase this patch to the mainline as well?

non:
Yes, I think so, thanks. Also, you might want to create a single test that exercises this to make the codecov warnings pass. Something like:

```scala
test("Is / Leibniz") {
  def cast1[A, B](as: List[A])(implicit ev: A Is B): List[B] = ev.substitute(as)
  cast1[Int, Int](1 :: 2 :: 3 :: Nil)

  def cast2[A, B](as: List[A])(implicit ev: Leibniz[A, B]): List[B] = ev.substitute(as)
  cast2[Int, Int](1 :: 2 :: 3 :: Nil)
}
```

Thanks for sticking with this!

non:
LGTM 👍 

(We can add more coverage tests at some point but this has been hanging out for awhile and I think it's worth merging soon.)

tel:
🎸

PR #1183, opened 2016-07-10
Update FAQ with some references for machinist, kind projector and simulacrum

mikejcurry:
Just noticed that these sections were basically left blank so threw in some text for them.

Happy to tweak if required, but I think this at least provides a pointer in the right direction.


milessabin:
👍

docs/src/main/tut/faq.md
```diff
@@ -16,7 +16,7 @@ section: "faq"
  * [What does `macro Ops` do? What is `cats.macros.Ops`?](#machinist)
  * [How can I help?](#contributing)
 
-## What imports do I need?<a id="what-imports" href="#what-imports"></a>
+## <a id="what-imports" href="#what-imports"></a>What imports do I need?
```
ceedubs: I put the links at the end of the header because I added a bit of CSS that makes an anchor pop up with a link for each header and if it's on the left side, the header text shifts around which is a bit disorienting. This is something that isn't noticeable if you are just looking at the markdown view, but it will be noticeable once the FAQ is deployed on the site.

I'm definitely open to better options -- I was just looking for low-hanging fruit for easy FAQ links that I could implement with very limited CSS skills :)

docs/src/main/tut/faq.md
```diff
@@ -28,15 +28,33 @@ import cats.implicits._
...
 
-## How can I help?<a id="contributing" href="#contributing"></a>
+## <a id="simulacrum" href="#simulacrum"></a>What does `@typeclass` mean?
+
+Cats defines and implements numerous type classes. Unfortunately, encoding these type classes in Scala can incur a large amount of boilerplate. To address this, [Simulacrum](https://github.com/mpilquist/simulacrum) introduces `@typeclass`, a macro annotation which generates a lot of this boilerplate. This elevates type classes to a first class construct and increases the legibility and maintainability of the code. Use of Simulacrum also ensures consistency in how the type classes are encoded across a project. Cats uses Simulacrum wherever possible to encode type classes, and you can read more about it at the [project page](https://github.com/mpilquist/simulacrum).
+
```
ceedubs: > Cats uses Simulacrum wherever possible to encode type classes

We have made a conscious decision not to use it within cats-kernel to ensure backwards compatibility. I'm not sure whether or not that's worth bringing up here.

ceedubs:
I left a couple minor comments, but this is great. Thanks @mikejcurry! Let me know what you think about the comments that I left.

docs/src/main/tut/faq.md
mikejcurry: Hey @ceedubs - so the reason I switched the order, was that when I tested locally, when the anchor was defined after the text, I built locally and tested the page, and the problem was that clicking the link brought me to just after the heading. So, as the page is a bit larger now, that actually meant that the heading was not visible, and I had to scroll in reverse to see the heading.  I thought this would be confusing for anyone who clicked a link not to be able to see the heading of the link they clicked on so I switched the anchor to earlier. I agree about the anchor, but I think it might actually be worse not to be able to see the heading.

docs/src/main/tut/faq.md
mikejcurry: Good point, I'll add some text this evening for this. Were the backward compatibility concerns related to algebra origins of the code in kernel? I think context would be useful here related to this for reference so I'd like to be able to include something to that effect.

docs/src/main/tut/faq.md
ceedubs: @mikejcurry ah I see. That makes sense. Let's go with your approach, and hopefully someone will be annoyed enough by the jumping text that they'll fix it for us :P

docs/src/main/tut/faq.md
ceedubs: Yes, the compatibility concerns are related to the algebra origins of cats-kernel. I agree that including context would be nice, but we probably don't need to bring in the historical context. I think we could just say something about how the `cats-kernel` module is expected to be a shared dependency for at least a couple of projects and there's an interest in ensuring that it is lightweight and very stable from a binary compatibility perspective. At some point there may be a transition from simulacrum to typeclassic, and we aren't confident that this transition will be 100% binary compatible.

docs/src/main/tut/faq.md
mikejcurry: My own CSS skills aren't particularly advanced, but if I get some time I'll come back to these links and see if I can find a better way.

docs/src/main/tut/faq.md
mikejcurry: I've incorporated your comment into the text. Let me know what you think!

non:
This looks great to me. I think the minor comments might be right, but I'm also happy to merge sooner and then fix issues in a future PR. 👍

mikejcurry:
Cheers, I think the minor comments have been addressed now - unless I missed something along the way.

Thanks!

ceedubs:
Fantastic. Thanks @mikejcurry!

PR #1185, opened 2016-07-11
Add return type in Traverse example

richardmiller:
The following line says note the return type but it is not in the definition.


johnynek:
👍

ceedubs:
Thanks @richardmiller! I would also be inclined to mark this as `tut:silent`, because the output in the current form is a little wonky:

```scala
scala> def profilesFor(users: List[User]) = users.map(userInfo)
profilesFor: (users: List[User])List[scala.concurrent.Future[Profile]]
```

Now that the return type is specified explicitly the output line can probably just be omitted. What do you think?

richardmiller:
@ceedubs makes sense to me to omit the output, I've opened #1186 with the change.

PR #1186, opened 2016-07-12
Use tut:silent for Traverse example

richardmiller:
As per comment on #1185


fthomas:
:+1: Thanks @richardmiller!

kailuowang:
:+1: many thanks!

PR #1188, opened 2016-07-13
Add XorT.toValidatedNel

markus1189:
Currently there is only `.toValidated`, but I frequently need `toValidatedNel` as well, so here it is ;)

### DIFF
--- a/core/src/main/scala/cats/data/XorT.scala
+++ b/core/src/main/scala/cats/data/XorT.scala
@@ -157,6 +157,9 @@ final case class XorT[F[_], A, B](value: F[A Xor B]) {
   def toValidated(implicit F: Functor[F]): F[Validated[A, B]] =
     F.map(value)(_.toValidated)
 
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/data/XorT.scala
```diff
@@ -157,6 +157,9 @@ final case class XorT[F[_], A, B](value: F[A Xor B]) {
   def toValidated(implicit F: Functor[F]): F[Validated[A, B]] =
     F.map(value)(_.toValidated)
 
+  def toValidatedNel(implicit F: Functor[F]): F[ValidatedNel[A,B]] =
```
ceedubs: Scalastyle wants there to be a space after the comma: `ValidatedNel[A, B]`

ceedubs:
I like it. Thanks @markus1189.

👍 after the minor formatting issue is addressed and Scalastyle is happy.

peterneyens:
:+1:

PR #1189, opened 2016-07-13
Add return type in Traverse example

richardmiller:
As with #1185 the return type is referred to in following line but not in the code example


ceedubs:
👍 thanks again @richardmiller!

PR #1190, opened 2016-07-13
Fix imports in Semigroup doc

### LINKED ISSUE #1187: Documentation - Example for Semigroup won't compile SBT/IntelliJ
Either via sbt compile on command line or in IntelliJ, the examples that are with the Semigroup documentation just don't work.

I had to change things up a bit (looking at your syntax unit tests) to get something to work...

I seems if I import cats.std.AllInstances & cats.syntax.AllSyntax and then add them to an object, I can get that object to compile in both.

i.e.: 

```scala
object Demo extends App with AllInstances with AllSyntax {
   val a = Semigroup[Int].combine(1,2)
   val b = List(1,2,3) |+| List(4,5,6)
}
```

If you don't extend those traits, then this will not compile with the imports that your documentation provides.

v0.6.0

ceedubs:
Note: this is a PR to the gh-pages branch for a manual edit of the website. We want to keep the website in sync with the latest stable release, so we can't just publish the site from the current master branch.

Fixes #1187.

### DIFF
--- a/_tut/semigroup.md
+++ b/_tut/semigroup.md
@@ -31,7 +31,6 @@ First some imports.
 
 ```scala
 import cats._
-import cats.std.all._
 import cats.implicits._
 ```
 


ceedubs:
It looks like Travis CI doesn't like PRs to the gh-pages branch. Oh well, I think this fix is fine. YOLO!

kailuowang:
👍

peterneyens:
:+1: 

[This StackOverflow question](http://stackoverflow.com/questions/37964102/why-i-have-a-compilation-error-with-monoid-from-cats) showed the same problem, so it seems important that we fix this issue in the docs before the next release.

PR #1191, opened 2016-07-13
Add piecemeal Monoid import

yilinwei:
Add piecemeal `monoid` import.


fthomas:
:+1:

peterneyens:
:+1: Thanks @yilinwei, it seems this was missed in #1130 (probably because of the Machinist confusion).

PR #1193, opened 2016-07-14
Implement Reducible.reduceM

benhutchison:
`foldM` for nonempty collections. First mention: https://gitter.im/typelevel/cats?at=5781d3f33eaf66535e62da69

### DIFF
--- a/core/src/main/scala/cats/Reducible.scala
+++ b/core/src/main/scala/cats/Reducible.scala
@@ -60,6 +60,12 @@ import simulacrum.typeclass
   def reduceLeftTo[A, B](fa: F[A])(f: A => B)(g: (B, A) => B): B
 
   /**
+   *  Monadic variant of [[reduceLeftTo]]
+   */
+  def reduceLeftM[G[_], A, B](fa: F[A])(f: A => G[B])(g: (B, A) => G[B])(implicit G: FlatMap[G]): G[B] =
+    reduceLeftTo(fa)(f)((gb, a) => G.flatMap(gb)(g(_, a)))
+
+  /**
    * Overriden from Foldable[_] for efficiency.
    */
   override def reduceLeftToOption[A, B](fa: F[A])(f: A => B)(g: (B, A) => B): Option[B] =
--- a/tests/src/test/scala/cats/tests/ReducibleTests.scala
+++ b/tests/src/test/scala/cats/tests/ReducibleTests.scala
@@ -1,0 +1,18 @@
+package cats
+package tests
+
+class ReducibleTestsAdditional extends CatsSuite {
+
+  test("Reducible[NonEmptyList].reduceLeftM stack safety") {
+    def nonzero(acc: Long, x: Long): Option[Long] =
+      if (x == 0) None else Some(acc + x)
+
+    val n = 100000L
+    val expected = n*(n+1)/2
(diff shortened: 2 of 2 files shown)


core/src/main/scala/cats/Reducible.scala
```diff
@@ -60,6 +60,12 @@ import simulacrum.typeclass
   def reduceLeftTo[A, B](fa: F[A])(f: A => B)(g: (B, A) => B): B
 
   /**
+    * Monadic variant of `reduceLeftTo`
+    * */
```
ceedubs: Minor nitpicks:

Could we change the last line of the scaladoc entry to just `*/`?

Also can we use `[[reduceLeftTo]]` so that the scaladoc has a handy link?

tests/src/test/scala/cats/tests/ReducibleTests.scala
```diff
@@ -0,0 +1,20 @@
...
+    def nonzero(acc: Long, x: Long): Option[Long] =
+      if (x == 0) None else Some(acc + x)
+
+    val n = 100000L
+    val expected = n*(n+1)/2
+    (1L to n).toList.toNel.foreach { nel =>
+      val actual = nel.reduceM(Option.apply)(nonzero)
+      assert(actual.get == expected)
+    }
+  }
```
ceedubs: This should be fine since `(1L to n).toList.toNel` should never return `None`, but I think I'd rather either see it explicitly use `Option.get` or do something like:

```scala
val actual = (1L to n).toList.toNel.flatMap(_.reduceM(Option.apply)(nonzero))
actual should === (Some(expected))
```

ceedubs:
Thanks @benhutchison! It looks good -- I just left a few very minor formatting/stylistic thoughts.

I also wonder about using the name `reduceLeftM` and potentially adding a `reduceLeftMRec` once #1117 goes through, but those should probably wait until that PR is merged.

benhutchison:
@ceedubs Updated.

 I went ahead with name `reduceLeftM` because I think its clearer & already well considered in #1117. Avoids busywork later to change it.

ceedubs:
👍 looks good to me once the build passes.

core/src/main/scala/cats/Reducible.scala
```diff
@@ -60,6 +60,12 @@ import simulacrum.typeclass
   def reduceLeftTo[A, B](fa: F[A])(f: A => B)(g: (B, A) => B): B
 
   /**
+    * Monadic variant of [[reduceLeftTo]]
+    */
+  def reduceLeftM[G[_], A, B](fa: F[A])(f: A => G[B])(g: (B, A) => G[B])(implicit G: Monad[G]): G[B] =
```
peterneyens: I think we can change `Monad` to `FlatMap` here.

peterneyens:
I left one minor comment, for the rest  :+1:

core/src/main/scala/cats/Reducible.scala
ceedubs: And since another change has been suggested I feel less guilty about this one: can you shift the asterisks one space to the left to be javadoc style like other comments in the file? :)

ceedubs:
Nice work, @benhutchison!

PR #1173, opened 2016-07-01
Backport tuple order fix for 0.6.x

ceedubs:
This PR (which is against the `series/0.6.x` branch that I just created) does a couple of things:
- Backport the tuple order fix from #1062 
- Add a binary compatibility check to all published modules (since we will want to maintain binary compatibility within the 0.6.x series)

I will largely unavailable for a little over a week, but I thought that if I got this branch set up someone else (@non?) might be able to get a 0.6.1 published with the fix for this pretty nasty bug.

cc @mpilquist and @travisbrown who have both expressed interest in getting this backported for a 0.6.1 release.

PR #1196, opened 2016-07-14
Update docs for version 0.6.1

ceedubs:
This shouldn't be merged until version 0.6.1 is released.

Currently I don't have permission to publish to Sonatype for cats, so for now I'm hoping that @non (or someone else -- maybe @travisbrown or @kailuowang or @mpilquist ?) can publish the release (off of the `series/0.6.x` branch). Hopefully by the time we are ready for the next release I can get the necessary permissions. I've guessed Saturday as the release date in the changelog but we can definitely change that depending on when it actually gets released :)

Note: there's a little bit of strangeness in that we need to make these changes to the master branch since that's the README that people see, but it kind of seems like these changes should also live on the `series/0.6.x` branch. However, the `0.6.x` branch was made off of the [v0.6.0 tag](https://github.com/typelevel/cats/releases/tag/v0.6.0) which doesn't actually include the release notes for the 0.6.0 release. I'm inclined to not worry about updating the documentation on the 0.6.0 branch and just make a manual adjustment to the suggested version number in `index.md` on the `gh-pages` branch. Hopefully we can form a better system for the next release.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,6 @@
 ## Version 0.6.1
 
-> 2016 July 16
+> 2016 July 14
 
 Version 0.6.1 is a patch release compatible with 0.6.0.
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,17 @@
+## Version 0.6.1
+
+> 2016 July 16
+
+Version 0.6.1 is a patch release compatible with 0.6.0.
+
+It contains one bug fix:
+
+* [#1062](https://github.com/typelevel/cats/pull/1173/commits/8dd682771557274a61f1e773df0f999b44a9819d): Fixed a bug in the Order and PartialOrder instances for Tuple2+ where only the first element was used in comparisons
+
+It also contains a change to the build:
+
+* [#1173](https://github.com/typelevel/cats/pull/1173/commits/5531d1ac7a6807c1842cd4b5b599173b14b652a9): Add binary compatibility check to all published modules
+
 ## Version 0.6.0
 
 > 2016 May 19
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ To get started with SBT, simply add the following to your `build.sbt`
 file:
 
 ```scala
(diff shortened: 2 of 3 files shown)


kailuowang:
👍  to me. I think I have the permission to release cats on sonatype, but I've never released cats before. Is it just a `sbt release` from the `series/0.6.x` branch? Any other manual step?

travisbrown:
I also don't have publishing permissions, but 👍 to @kailuowang running it.

ceedubs:
@kailuowang I haven't released either. There are instructions [here](https://github.com/typelevel/cats/blob/master/PROCESS.md#releasing) (but as you mentioned it should be from the `series/0.6.x branch`). Hopefully they are right :P

kailuowang:
@ceedubs okay I am going to give it shot. Will ask in gitter as well.

CHANGES.md
```diff
@@ -1,3 +1,17 @@
+## Version 0.6.1
```
johnynek: doesn't this change the name of many implicits? Should we mention that and mention that it is not strictly binary compatible with `0.6.1` due to that change?

Also, why not `0.7.0` to signify that change rather than `0.6.1`?

ceedubs: @johnynek this isn't a release of the current master. It's just a backport to the `series/0.6.x` branch that should be equivalent to `0.6.0` but with the addition of #1173.

ceedubs: A couple more notes:

It should be completely binary compatible (as verified by mima).

It's a little strange that the CHANGES.md PR is to the master branch. See the note in the original PR message about the current branch/release system being a little cumbersome. I'm hoping we can have a good system for this in the future.

johnynek: got it. Thanks for the clarification!

kailuowang:
Do you mind update the release date?

ceedubs:
@kailuowang thanks for the reminder. Release date updated.

kailuowang:
👍

ceedubs:
I'm just going to merge since the only thing that changed in the last commit was the release date in the `CHANGES.md` file.

PR #1197, opened 2016-07-14
Copied scaladoc content as a better placeholder

metasim:
The scaladoc description of `MonoidK` is pretty good, and I think it's better to see a copy of it here than to find a blank page.


ceedubs:
👍 certainly seems better than not having anything to me.

kailuowang:
many thanks! 👍

PR #1198, opened 2016-07-14
Change dependency version to 0.6.1 on website

ceedubs:
This is a change directly to the `gh-pages` branch. Normally this would be taken care of by publishing the whole site for a new release, but we currently don't have a great setup for doing this outside of the `master` branch.

We should wait until 0.6.1 has synced to maven central before merging this.


kailuowang:
maven is synced. 👍

PR #1200, opened 2016-07-14
fix memory problem when running sbt with forking all tests and doc ge…

kailuowang:
Right now running aggregated task such as `validate` or `release` in a single sbt session usually causes out of memory crash.
Forking some heavy tasks such as `tests` and `scalaUnidoc` seems to help. 
Noticed that it is able to run `validate` on travis now.


johnynek:
👍 looks good to me. I often get OOM.

non:
👍

non:
Thanks for tracking this down!

ceedubs:
I just noticed https://github.com/rickynils/scalacheck/issues/185. Could this be an issue?

kailuowang:
I briefly compared the counts with and without this change, didn't notice any difference.

PR #1201, opened 2016-07-15
Prefix instead of suffix unsafe methods with 'unsafe'

adelbertc:
Prefixing seems to be the more common pattern, so here ya go

Examples:
FS2: https://github.com/functional-streams-for-scala/fs2/search?utf8=%E2%9C%93&q=unsafe
Our own `Free`: https://github.com/typelevel/cats/blob/a4b4091d04a8ad4736713037a0fb6d58c23468cf/free/src/main/scala/cats/free/Free.scala#L224
Haskell: https://hackage.haskell.org/package/base-4.9.0.0/docs/System-IO-Unsafe.html

### DIFF
--- a/core/src/main/scala/cats/data/NonEmptyVector.scala
+++ b/core/src/main/scala/cats/data/NonEmptyVector.scala
@@ -15,14 +15,14 @@ final case class NonEmptyVector[A] private (toVector: Vector[A]) {
     toVector.lift(i)
 
   /** Gets the element at the index, or throws an exception if none exists */
-  def getUnsafe(i: Int): A = toVector(i)
+  def unsafeGet(i: Int): A = toVector(i)
 
   /** Updates the element at the index, if it exists */
   def updated(i: Int, a: A): Option[NonEmptyVector[A]] =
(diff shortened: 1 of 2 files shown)


ceedubs:
The thing I like about using `Unsafe` as a suffix (as opposed to a prefix) is that methods seem like they are a bit more discoverable via autocomplete that way. For example, if I have a `NonEmptyVector` and I want to update an element of it, I'll likely start typing `.upda...` and then I'll see the two safe/unsafe options. If `unsafe` is used as a prefix, I might not discover a method as easily.

Having said that, I don't feel super strongly about this and am happy to go with whatever people prefer, so long as we are consistent.

non:
I agree with @ceedubs -- to me the most interesting these about these methods is that they do a particular thing, not that they are *unsafe*. If we want to standardize I think I'd prefer to standardize in the other direction. (I've used the unsafe suffix convention in Jawn for example.)

adelbertc:
To be honest I like the suffix as well for similar reasons as @ceedubs , but the inconsistency bothered me enough to file a PR to see how others felt :-)

travisbrown:
For what it's worth I prefer the prefix in part _because of_ the discovery issue.

PR #1203, opened 2016-07-15
Change format of NonEmptyVector.show

ceedubs:
This follows up on several small suggestions from comments in #1137. The
only meaningful modification is changing the format of `show` from
something like `NonEmptyVector(Vector(1, 2, 3))` to something like
`NonEmptyVector(1, 2, 3)`.

I've also added a `length` method to `NonEmptyVector`.

### DIFF
--- a/core/src/main/scala/cats/data/NonEmptyVector.scala
+++ b/core/src/main/scala/cats/data/NonEmptyVector.scala
@@ -21,7 +21,11 @@ final case class NonEmptyVector[A] private (toVector: Vector[A]) {
   def updated(i: Int, a: A): Option[NonEmptyVector[A]] =
     if (toVector.isDefinedAt(i)) Some(NonEmptyVector(toVector.updated(i, a))) else None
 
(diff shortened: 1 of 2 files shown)


johnynek:
👍

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -138,7 +144,7 @@ private[data] sealed trait NonEmptyVectorInstances {
 
       override def split[A](fa: NonEmptyVector[A]): (A, Vector[A]) = (fa.head, fa.tail)
 
-      override def size[A](fa: NonEmptyVector[A]): Long = 1 + fa.tail.size.toLong
+      override def size[A](fa: NonEmptyVector[A]): Long = fa.length.toLong
```
kailuowang: good catch there.

kailuowang:
👍  LGTM pending the `tut` fix.

non:
👍

peterneyens:
:+1:

PR #1205, opened 2016-07-15
Make some Writer improvements

ceedubs:
These are some changes based on a [Gitter conversation](https://gitter.im/typelevel/cats?at=5788d4739f79ee4f2bb479a0).
Well at least some of them are. Others just seemed right to me.

* Add a `write` method to `WriterT` to easily append to the log.
  * This is the main item inspired by the Gitter conversation.
* Change `tell` to `write`.
  * This seems more consistent to me, because we have used `MonadWriter` as opposed to `MonadTell`.
* Add helper `Writer.value` and `Writer.write` instantiation methods.

### DIFF
--- a/tests/src/test/scala/cats/tests/WriterTTests.scala
+++ b/tests/src/test/scala/cats/tests/WriterTTests.scala
@@ -12,7 +12,7 @@ import cats.kernel.laws.OrderLaws
 class WriterTTests extends CatsSuite {
   type Logged[A] = Writer[ListWrapper[Int], A]
 
-  // we have a lot of generated lists of lists in these tests. We have to write
+  // we have a lot of generated lists of lists in these tests. We have to tell
   // Scalacheck to calm down a bit so we don't hit memory and test duration
   // issues.
   implicit override val generatorDrivenConfig: PropertyCheckConfiguration =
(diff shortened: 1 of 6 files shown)


tests/src/test/scala/cats/tests/WriterTTests.scala
```diff
@@ -12,7 +12,7 @@ import cats.kernel.laws.OrderLaws
 class WriterTTests extends CatsSuite {
   type Logged[A] = Writer[ListWrapper[Int], A]
 
-  // we have a lot of generated lists of lists in these tests. We have to tell
+  // we have a lot of generated lists of lists in these tests. We have to write
   // Scalacheck to calm down a bit so we don't hit memory and test duration
```
fthomas: :smile_cat:

ceedubs: Oopsies :)

ceedubs: Does anyone know the mailing address that I can use to write Scalacheck?

johnynek:
`listen` seems awkward now to me. `tell` and `listen` are dual, `write` and `listen` don't seem to be. `consume` or `read` seem like better pairings with `write`.

ceedubs:
@johnynek hmm that's a good point about `listen`. I won't have time to come back to this today, so hopefully that will give people a bit of time to bounce around ideas.

ceedubs:
Now @johnynek has me questioning everything :P. I think that `read` does sound like a good companion to `write`, but then I wanted to check if that would collide with `MonadReader`. I then saw that we are using `ask` there and I'm asking myself if that should be `read` instead 🙀

ceedubs:
I'm going to just close this PR and open a new one in which I add the helper methods but don't changing any naming. Naming is too hard :P

PR #1207, opened 2016-07-16
Add tell method to WriterT

ceedubs:
And also add `tell` and `value` methods to the `Writer` object to
conveniently create `Writer` instances.

### DIFF
--- a/tests/src/test/scala/cats/tests/WriterTTests.scala
+++ b/tests/src/test/scala/cats/tests/WriterTTests.scala
@@ -77,7 +77,7 @@ class WriterTTests extends CatsSuite {
     w2.tell("bar") should === (Writer("foobar", 3))
   }
 
-  test("MonadWriter's tell is consistent with write") {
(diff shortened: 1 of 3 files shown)


ceedubs:
This is a replacement for #1205, where I tried to do too many things at once and got tripped up on naming.

tests/src/test/scala/cats/tests/WriterTTests.scala
```diff
@@ -70,6 +70,25 @@ class WriterTTests extends CatsSuite {
...
   }
 
+  test("tell appends to log") {
+    val w1: Writer[String, Int] = Writer.value(3)
+    val w2 = w1.tell("foo")
+    w2 should === (Writer("foo", 3))
+    w2.tell("bar") should === (Writer("foobar", 3))
+  }
+
+  test("MonadWriter's tell is consistent with write") {
```
peterneyens: This probably should be "*... is consistent with tell*" ?

peterneyens:
:+1:

non:
👍

Issue #1067, opened 2016-05-27
`Profunctor` instance for `StateT`


ceedubs:
@vmarquez pointed out to me that `StateT` forms a `Profunctor`

kailuowang:
I gave it a try and failed, since `S` appears in both input and output, it looks to me that for `S` it has to be invariant (you need both `S => NewS` and `NewS => S`. 
@vikraman or @ceedubs mind give some further instruction? 
Here is my attempt
```Scala
def diMap[C, D](f: C => S)(g: A => D)(implicit F: Functor[F]): StateT[F, C, D] =
    new StateT(F.map(runF)(run => (c:C) => F.map(run(f(c)))(p => ((???) , g(p._2))) ))
````

I need a `C` in the place of `???` but in `p` I got only a `S`

vmarquez:
Err, I think I pointed out *Indexed*StateT is one. ;-)

kailuowang:
ah, makes sense. since cats doesn't have `IndexedStateT` shall we close this one?

ceedubs:
Oops I hadn't thought that through very well :)

Sorry for the false alert. Thanks for giving it a try @kailuowang and thanks for your input, @vmarquez.

PR #1211, opened 2016-07-17
Add Contravariant instance for Eq and test Show instance.

peterneyens:
As far I can tell this seemed to be the only missing `Contravariant` instance.

 #579 also mentioned `StateT`, but I think only an `IndexedStateT` would have a `Contravariant` (that's also why there is no `Profunctor` instance, see #1067).

I don't think it is possible to test the instance for `Eq`, since `CovariantTests` would need an `Arbitrary[Eq[A]]` and an `Eq[Eq[A]]` ?

### DIFF
--- a/laws/src/main/scala/cats/laws/discipline/Eq.scala
+++ b/laws/src/main/scala/cats/laws/discipline/Eq.scala
@@ -32,6 +32,22 @@ object eq {
   }
 
   /**
+   * Create an approximation of Eq[Eq[A]] by generating 100 values for A
+   * and comparing the application of the two eqv functions
+   */
+  implicit def catsLawsEqForEq[A](implicit arbA: Arbitrary[(A, A)], booleanEq: Eq[Boolean]): Eq[Eq[A]] = new Eq[Eq[A]] {
+    def eqv(f: Eq[A], g: Eq[A]): Boolean = {
+      val samples = List.fill(100)(arbA.arbitrary.sample).collect {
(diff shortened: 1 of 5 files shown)


adelbertc:
Perhaps we could do a heuristic test by generating a bunch of `A`'s and say two `Eq`s are equal if they return the same result when equal-ing a bunch of random `A`s?

ceedubs:
👍 great!

laws/src/main/scala/cats/laws/discipline/Eq.scala
```diff
@@ -32,6 +32,22 @@ object eq {
   }
 
   /**
+   * Create an approximation of Eq[Eq[A]] by generating 100 values for A
+   * and comparing the application of the two eqv functions
+   */
+  implicit def catsLawsEqForEq[A](implicit arbA: Arbitrary[(A, A)], booleanEq: Eq[Boolean]): Eq[Eq[A]] = new Eq[Eq[A]] {
```
johnynek: I don't really think we need the `Eq[Boolean]` here. I think we assume that primitives are okay to compare directly.

peterneyens: That makes sense, I adapted `catsLawsEqForOrder` where an `Eq[Int]` is used.

johnynek:
👍

ceedubs:
👍

Issue #1204, opened 2016-07-15
Reconsider overloading in NonEmptyVector


ceedubs:
Currently [NonEmptyVector](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/NonEmptyVector.scala) has overloaded `concat` and `++` methods supporting both `NonEmptyVector` and `Vector` as the arguments.

In general I'm a bit skeptical about overloading. In this specific case the main downside I see is that it prevents us from turning `NonEmptyVector` into a value class (which seems like it might be an otherwise reasonable thing to do).

What do people think?

mikejcurry:
:+1: on removing the overloading as a general principle. Also, having a value class would be useful in specific use cases. 

The syntax is still useful for working with Vectors directly though. Maybe a rename to `concatVector` or `appendVector` and a different operator such as `+++`.

As an aside, with regard to value classes, it is worth bearing in mind that, as I understand at least, invoking any general code that involves summoning an instance would negate the benefit of the value class. i.e. summoning the `Semigroup` for `NonEmptyVector` would require an instance of `NonEmptyVector` to be instantiated - because `NonEmptyVector` would be being used as a type argument to `Semigroup` and one of the limitations on value classes is that if they are used as type argument an instance needs to be instantiated. I think having `NonEmptyVector` as a value classes is still useful despite this, because it opens up options to users of the class.

mikejcurry:
Believe this should be resolved by #1212

PR #1212, opened 2016-07-17
Make NonEmptyVector a value class. #1204

mikejcurry:
As suggested might be useful in #1204 -
Merging this would make `NonEmptyVector` a value class, which may be useful in some scenarios. It also:
- Renames variant of concat taking a `Vector` to `concatVector`. 
- Renames variant of `++` taking a `Vector` to `+++`
- Updates test in line with the change. 
- Adds additional consistency test.

Updates scaladoc in some minor places, including turning references to concat and concatVector from `++` and `+++` respectively into links to the relevant methods.

Not sure about naming, especially `+++`, but figured a PR would be a good place to discuss:
(a) The naming,
(b) If this is something we would want to do.

### DIFF
--- a/core/src/main/scala/cats/data/NonEmptyVector.scala
+++ b/core/src/main/scala/cats/data/NonEmptyVector.scala
@@ -6,9 +6,9 @@ import scala.collection.immutable.VectorBuilder
 import cats.instances.vector._
 
 /**
- * A data type which represents a non empty Vector.
+ * A data type which represents a `Vector` guaranteed to contain at least one element.
  */
-final case class NonEmptyVector[A] private (toVector: Vector[A]) {
+final case class NonEmptyVector[A](val toVector: Vector[A]) extends AnyVal {
 
   /** Gets the element at the index, if it exists */
   def get(i: Int): Option[A] =
@@ -38,27 +38,27 @@ final case class NonEmptyVector[A] private (toVector: Vector[A]) {
   def filter(f: A => Boolean): Vector[A] = toVector.filter(f)
 
   /**
-   * Append another NonEmptyVector to this
+   * Append another `NonEmptyVector` to this, producing a new `NonEmptyVector`.
    */
   def concat(other: NonEmptyVector[A]): NonEmptyVector[A] = NonEmptyVector(toVector ++ other.toVector)
 
   /**
-   * Alias for concat
+   * Alias for [[concat]]
    */
   def ++(other: NonEmptyVector[A]): NonEmptyVector[A] = concat(other)
 
   /**
-   * Append another Vector to this
+   * Append another `Vector` to this, producing a new `NonEmptyVector`.
    */
-  def concat(other: Vector[A]): NonEmptyVector[A] = NonEmptyVector(toVector ++ other)
+  def concatVector(other: Vector[A]): NonEmptyVector[A] = NonEmptyVector(toVector ++ other)
 
   /**
-   * Alias for concat
+   * Alias for [[concatVector]]
    */
-  def ++(other: Vector[A]): NonEmptyVector[A] = concat(other)
+  def +++(other: Vector[A]): NonEmptyVector[A] = concatVector(other)
 
   /**
-   * find the first element matching the predicate, if one exists
+   * Find the first element matching the predicate, if one exists
    */
   def find(f: A => Boolean): Option[A] = toVector.find(f)
 
@@ -190,7 +190,7 @@ private[data] sealed trait NonEmptyVectorInstances {
               case Some(t) => go(t)
               case None => ()
             }
-          case Xor.Left(a) => go(f(a).concat(v.tail))
+          case Xor.Left(a) => go(f(a).concatVector(v.tail))
           }
         go(f(a))
         NonEmptyVector.fromVectorUnsafe(buf.result())
--- a/core/src/main/scala/cats/data/NonEmptyVector.scala
+++ b/core/src/main/scala/cats/data/NonEmptyVector.scala
@@ -7,8 +7,12 @@ import cats.instances.vector._
 
 /**
  * A data type which represents a `Vector` guaranteed to contain at least one element.
+ * <br/>
+ * Note that the constructor is `private` to prevent accidental construction of an empty
+ * `NonEmptyVector`. However, due to https://issues.scala-lang.org/browse/SI-6601, on
+ * Scala 2.10, this may be bypassed due to a compiler bug.
  */
-final case class NonEmptyVector[A](val toVector: Vector[A]) extends AnyVal {
+final class NonEmptyVector[A] private (val toVector: Vector[A]) extends AnyVal {
 
   /** Gets the element at the index, if it exists */
   def get(i: Int): Option[A] =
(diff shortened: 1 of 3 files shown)


mikejcurry:
Hmm, just noticed #1214, which effects this a lot. I wonder is there a nice way to still be a value class and prevent the issue on #1214  at the same time. Maybe turning `NonEmptyVector` into a sealed universal trait with a private implementation in the same file. Not sure if this would work though - will have to play around with it. I guess this implementation as it stands is not something that would be good though.

ceedubs:
@mikejcurry as far as supporting both this and #1214, I think it shouldn't be a problem as long as we make it no longer a case class (as that issue suggests). I think it would look like `final class NonEmptyVector[A] private (val toVector: Vector[A]) extends AnyVal`. I've left a few comments [here](https://github.com/typelevel/cats/issues/1214#issuecomment-233301210) about changes we would need to make to support it no longer being a case class.

Unfortunately, there are still some compiler errors related to value classes. Having said that, this seems like a reasonable change to me that I don't really expect to cause trouble. I'm curious whether @non, @johnynek, or @mpilquist  have any thoughts.

I feel slightly inclined to use `concat` and `++` for the `Vector` varieties and `concatNEV`/`+++` for the `NonEmptyVector` cases, just because it seems like you'll be more likely to have a `Vector` than a `NonEmptyVector` (and the latter is easy to turn into the former), but this is quite arbitrary and I have no strong opinions here.

mikejcurry:
Hmmm, so I thought (on scala 2.10), that a value class needed exactly one public val parameter in primary constructor and scala 2.11 private was fine. Which is the bit that concerned me. Will do some research/testing as it could have changed/I could be wrong.

mikejcurry:
Ok, cool, looks like I misinterpreted the implications of the value class limitation with regard to vals, and seems like private constructors work fine - did a quick test and seems like it should fit together. 

I can fold the work for #1214 into this also as this would be a bit deficient without it I think. Might be a few days before I get to it though.

johnynek:
I don't like the `+++` syntax, myself. I'm also not 100% that we really need a value class here since changing `Vectors` involves a lot of allocations anyway. But if we can reduce, that's great.

I agree that if you can only have one arg type, I would take `Vector` over `NonEmptyVector`. In fact, I can imagine an implicit conversion from `NonEmptyVector` to `Vector` (though it may be too magical). In any case calling `a.concat(b.toVector)` seems fine.

ceedubs:
@johnynek I'm not super happy about the `+++` syntax either. Maybe we could maybe `concatNonEmpty` or `concatNEV` (hmm don't like these names that much either) and just let people use `|+|` if they want a symbolic operator?

mikejcurry:
@ceedubs @johnynek @peterneyens 

So, based on feedback have done the following:
- Folded in #1214 - `NonEmptyVector` is now a `value class`, but **not** a `case class`.
- Added test that `new NonEmptyVector` cannot be called externally, which unfortunately, due to a bug in scala 2.10, can be bypassed on that version. Added related comments in tests and documentation comments.
- No implementation of `hashCode` or `equals` required, value classes implicitly delegate to the underlying implementation, i.e. in this case, `Vector`.
- Removed the `+++` operator which felt strange. `concat` and `++` now take a `Vector` and `concatNEV` included as a convenience.

Let me know what you think.

mikejcurry:
Hmmm, didn't build scala js locally, and I didn't think accessing the version would require a callout like that. Not sure whether to remove the test or factor it into a scala 2.11 specific location. Open to ideas. The test is somewhat useful, but the scala 2.10 defect is a bit of a pain.

mikejcurry:
Forgot about the `catalysts` project, so used the `Platform.isJvm` construct provided with that to only run the test if on JVM, which prevents the test from running under scala js.

The build failure looks like it might be spurious. `137`, so the build script was sent a `SIGKILL`. Perhaps something I've done, but a rebuild might be the best starting point to see if it was just some spurious issue, especially as 2.10 build completed fine.

ceedubs:
Thanks @mikejcurry! I just restarted the build. Hopefully it works this time.

ceedubs:
👍 excellent. Thanks again!

tests/src/test/scala/cats/tests/NonEmptyVectorTests.scala
```diff
@@ -218,4 +222,40 @@ class NonEmptyVectorTests extends CatsSuite {
...
+      if (!Properties.versionNumberString.startsWith("2.10")) {
+        // A bug in scala 2.10 allows private constructors to be accessed.
+        // We should still ensure that on scala 2.11 and up we cannot construct the
+        // object directly. see: https://issues.scala-lang.org/browse/SI-6601
+        "val bad: NonEmptyVector[Int] = new NonEmptyVector(Vector(1))" shouldNot compile
+      }
+    }
+  }
+
+  test("Cannot create a new NonEmptyVector from apply with an empty vector") {
```
kailuowang: test description a bit misleading, can remove the word "empty" 
"Cannot create a new NonEmptyVector from apply with a vector".

mikejcurry: Fair point on the wording. Will rename when I get to a laptop I can work on it from. Hopefully this evening.

ceedubs: This is the perfect use-case for living on the wild side and using the in-browser GitHub editor :D

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -19,14 +23,14 @@ final case class NonEmptyVector[A] private (toVector: Vector[A]) {
 
   /** Updates the element at the index, if it exists */
   def updated(i: Int, a: A): Option[NonEmptyVector[A]] =
-    if (toVector.isDefinedAt(i)) Some(NonEmptyVector(toVector.updated(i, a))) else None
+    if (toVector.isDefinedAt(i)) Some(new NonEmptyVector(toVector.updated(i, a))) else None
```
johnynek: I'm a bit nervous about `isDefinedAt` I'm afraid this could be coming from `Function` and not be specialized well for vector. What about:
```scala
@inline
private def inBounds(i: Int): Boolean = 0 <= i && i <= toVector.size
```
And use it here and on `get`?

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -38,27 +42,22 @@ final case class NonEmptyVector[A] private (toVector: Vector[A]) {
...
    */
-  def concat(other: Vector[A]): NonEmptyVector[A] = NonEmptyVector(toVector ++ other)
+  def concat(other: Vector[A]): NonEmptyVector[A] = new NonEmptyVector(toVector ++ other)
 
   /**
-   * Alias for concat
+   * Append another `NonEmptyVector` to this, producing a new `NonEmptyVector`.
    */
-  def ++(other: Vector[A]): NonEmptyVector[A] = concat(other)
+  def concatNEV(other: NonEmptyVector[A]): NonEmptyVector[A] = new NonEmptyVector(toVector ++ other.toVector)
```
johnynek: any reason not to do `= concat(other.toVector)`

core/src/main/scala/cats/data/NonEmptyVector.scala
mikejcurry: Yea we could do that. For reference `Vector` inherits `isDefinedAt` from `GenSeqLike` - https://github.com/scala/scala/blob/v2.11.8/src/library/scala/collection/GenSeqLike.scala#L64-L72

Definition is pretty much the same, with the exception that a local implementation can be private and inlined.

core/src/main/scala/cats/data/NonEmptyVector.scala
mikejcurry: No reason I can think of. I wouldn't have a huge preference either way because they are functionally equivalent.

tests/src/test/scala/cats/tests/NonEmptyVectorTests.scala
mikejcurry: Yea, that's true - :-). I didn't get a chance to get back to this unfortunately. I'll get to it, and the other comments, by the end of the week all going well.

core/src/main/scala/cats/data/NonEmptyVector.scala
johnynek: Okay, this is fine then. (Lacking an IDE, finding such definitions is really painful for me in the insane collections hierarchy).

johnynek:
👍

adelbertc:
👍

mikejcurry:
Thanks for merging - I'll follow up on @kailuowang comment - https://github.com/typelevel/cats/pull/1212#discussion_r72367628 separately pretty soon. I just hadn't gotten to it prior to this merge.

PR #1215, opened 2016-07-18
Put liftT on the companion object of XorT

zmccoy:
I found it useful to have liftT available to me for wrapping values in XorT.  I couldn't figure out what test should be added in this case, but if anyone has any examples or suggestions I'd love to add one.

### DIFF
--- a/core/src/main/scala/cats/data/XorT.scala
+++ b/core/src/main/scala/cats/data/XorT.scala
@@ -223,7 +223,7 @@ trait XorTFunctions {
    * Alias for [[XorT.right]]
    * {{{
    * scala> import cats.data.XorT
-   * scala> import cats.instances.option._
+   * scala> import cats.implicits._
    * scala> val o: Option[Int] = Some(3)
    * scala> val n: Option[Int] = None
    * scala> XorT.liftT(o)
--- a/core/src/main/scala/cats/data/XorT.scala
+++ b/core/src/main/scala/cats/data/XorT.scala
@@ -210,7 +210,11 @@ final case class XorT[F[_], A, B](value: F[A Xor B]) {
   def toNested: Nested[F, A Xor ?, B] = Nested[F, A Xor ?, B](value)
 }
 
-object XorT extends XorTInstances with XorTFunctions
+object XorT extends XorTInstances with XorTFunctions {
+
+  def liftT[M[_]: Functor, E, A](ma: M[A]): XorT[M, E, A] = XorT(Functor[M].map(ma)(Xor.right))
+
+}
 
 trait XorTFunctions {
   final def left[F[_], A, B](fa: F[A])(implicit F: Functor[F]): XorT[F, A, B] = XorT(F.map(fa)(Xor.left))
@@ -280,7 +284,7 @@ private[data] abstract class XorTInstances extends XorTInstances1 {
       type TC[M[_]] = Functor[M]
 
       def liftT[M[_]: Functor, A](ma: M[A]): XorT[M, E, A] =
-        XorT(Functor[M].map(ma)(Xor.right))
+        XorT.liftT(ma)
     }
 
   implicit def catsMonoidForXorT[F[_], L, A](implicit F: Monoid[F[L Xor A]]): Monoid[XorT[F, L, A]] =
--- a/core/src/main/scala/cats/data/XorT.scala
+++ b/core/src/main/scala/cats/data/XorT.scala
@@ -220,9 +220,9 @@ trait XorTFunctions {
   final def pure[F[_], A, B](b: B)(implicit F: Applicative[F]): XorT[F, A, B] = right(F.pure(b))
 
   /**
-   * Alias for XorT.right
+   * Alias for [[XorT.right]]
    */
-  final def liftT[F[_], A, B](fb: F[B])(implicit F: Functor[F]): XorT[F, A, B] = right(fb)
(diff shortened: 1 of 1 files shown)


ceedubs:
@zmccoy thanks! This looks good.

Only one minor suggestion: to be consistent with other functions, could you add this to the `XorTFunctions` trait that gets mixed into the `XorT` companion object instead of defining it directly on the companion object? I wouldn't really be opposed to us having all of the methods directly in the companion object instead of using the mix-in trait, but since it's there I'd rather be consistent.

peterneyens:
I think `XorT.liftT` is analogous to the existing `XorT.right`, I am not sure if we need both ?

We also have the syntax `rightXorT` :

```scala
1.some.rightXorT[String]
// XorT[Option,String,Int] = XorT(Some(Right(1)))

XorT.right[Option, String, Int](1.some)
// XorT[Option,String,Int] = XorT(Some(Right(1)))
```

It is a good idea to use `XorT.liftT` / `XorT.right` in the `TransLift` instance of `XorT`.

ceedubs:
@peterneyens ah good point. I had forgotten about that.

If a `liftT` method would be more discoverable for people, then I don't think I have any objection to adding it as an alias for `right`, but it also seems fine to me to either let people use the existing methods.

zmccoy:
@peterneyens I somehow completely missed that when I skimmed the source!  
@ceedubs I'm fine with either, just let me know what you all would prefer and I can get it up to par :)

zmccoy:
I've put the function in the correct place, added it as an alias to `XorT.right`, and it's used in the `Translift` typeclass impl for `XorT`.

core/src/main/scala/cats/data/XorT.scala
```diff
@@ -219,6 +219,11 @@ trait XorTFunctions {
 
   final def pure[F[_], A, B](b: B)(implicit F: Applicative[F]): XorT[F, A, B] = right(F.pure(b))
 
+  /**
+   * Alias for XorT.right
+   */
+  final def liftT[F[_], A, B](fb: F[B])(implicit F: Functor[F]): XorT[F, A, B] = right(fb)
```
johnynek: why do we even need `Functor` here?

zmccoy: Without the implicit evidence for a `Functor` it won't compile since `right` needs one.

johnynek: Indeed. I made the mistake of believing that was just `Xor.right` Sorry for the noise.

zmccoy: All good, I'm just happy I could answer :)

core/src/main/scala/cats/data/XorT.scala
```diff
@@ -219,6 +219,11 @@ trait XorTFunctions {
 
   final def pure[F[_], A, B](b: B)(implicit F: Applicative[F]): XorT[F, A, B] = right(F.pure(b))
 
+  /**
+   * Alias for XorT.right
```
ceedubs: Could you please make this a ScalaDoc link? `[[XorT.right]]`

ceedubs:
Thanks @zmccoy!

Would you want to add an sbt-doctest example (for example like [this](https://github.com/typelevel/cats/blob/258fac68fa4cab2bf0f282cf083b7ea43592e0ed/core/src/main/scala/cats/data/XorT.scala#L192-L197))? I think these are useful for people looking at the scaladoc but also help us to see whether there are any inference issues that make methods like this cumbersome to use in practice.

core/src/main/scala/cats/data/XorT.scala
```diff
@@ -219,6 +219,11 @@ trait XorTFunctions {
 
   final def pure[F[_], A, B](b: B)(implicit F: Applicative[F]): XorT[F, A, B] = right(F.pure(b))
 
+  /**
+   * Alias for [[XorT.right]]
+   */
+  final def liftT[F[_], A, B](fb: F[B]): XorT[F, A, B] = right(fb)
```
ceedubs: This currently isn't compiling, because it needs an `F[_]: Functor` constraint here for the call to `right`.

zmccoy: That was my bad, I pushed up a bad commit with the scaladoc link in it.  I ended up in a hurry and my laptop was dieing.  I'll get that all fixed along with the sbt-doctest later this evening / tomorrow

zmccoy:
@ceedubs I'm not very familiar with sbt-doctest but I think that's what you were looking for, but if it isn't I'd love to fix it up.  Let me know if you need anything else.

core/src/main/scala/cats/data/XorT.scala
```diff
@@ -219,6 +219,21 @@ trait XorTFunctions {
 
   final def pure[F[_], A, B](b: B)(implicit F: Applicative[F]): XorT[F, A, B] = right(F.pure(b))
 
+  /**
+   * Alias for [[XorT.right]]
+   * {{{
+   * scala> import cats.data.XorT
+   * scala> import cats.instances.option._
```
ceedubs: Could you please use `import cats.implicits._` here? See #1026 for justification.

zmccoy: Totally, thanks for the link.

ceedubs:
👍 thanks!

zmccoy:
Hi @ceedubs is there anything else I need to do on this for review? :)

kailuowang:
:+1:

Issue #1209, opened 2016-07-16
Check syntax for typeclass piecemeal imports


yilinwei:
As demonstrated by @ShaneDelmore in the gitter channel, `syntax.flatMap._` doesn't have the functor syntax. I suspect that was because `FlatMap` had some new syntax added on and it doesn't extend the ones that simulacrum generates in the `ToOps`.

We should probably catch these in some tests.

ceedubs:
I think that this is probably resolved as we moved more syntax to be simulacrum-generated. Feel free to open new issues for any lingering problems.

PR #1218, opened 2016-07-19
Use Simulacrum Ops for FlatMap syntax

ceedubs:
This partially addresses #1209, though it doesn't add much in the way of
the syntax tests suggested by that ticket.


kailuowang:
👍 
Travis error
> Could not create file /home/travis/build/typelevel/cats/free/.js/target/streams/$global/projectDescriptors/$global/streams/outjava.io.IOException: No such file or directory

 that's a weird one, shall we try re-run?

peterneyens:
:+1:

PR #1221, opened 2016-07-20
Use traits to avoid code duplication in Nested instances

ceedubs:
This came up [here](https://github.com/typelevel/cats/pull/1148#discussion_r71067154), but I didn't want to tie this change into #1148.

Among others, I think a review from @adelbertc would be helpful.


adelbertc:
👍

kailuowang:
:+1:

PR #1225, opened 2016-07-23
Add FunctorFilter and TraverseFilter

### LINKED ISSUE #1054: filterM broken?

ceedubs:
This is a replacement for #1148.

`TraverseFilter` is a port of Haskell's Data.Witherable, and represents
structures that support doing a combined `traverse` and `filter` as a
single operation.

`FunctorFilter` is similar but is limited to a combined `map` and
`filter` as a single operation. The main reason that I've added
`FunctorFilter` is to extract the commonality between `TraverseFilter`
and `MonadFilter` so that they don't have collisions in methods such as
`filter`.

Some benefits of these type classes:

- They provide `filterM` (actually a more general version of it -
  `filterA`), which resolves #1054.
- They provide `collect` which is a handy method that is often used on
  standard library collections but had been absent in Cats.


ceedubs:
Oops I have a couple things to add. Closing for the moment.

core/src/main/scala/cats/data/Const.scala
```diff
@@ -53,13 +56,16 @@ private[data] sealed abstract class ConstInstances extends ConstInstances0 {
       fa.retag[B]
   }
 
-  implicit def catsDataTraverseForConst[C]: Traverse[Const[C, ?]] = new Traverse[Const[C, ?]] {
-    def traverse[G[_]: Applicative, A, B](fa: Const[C, A])(f: A => G[B]): G[Const[C, B]] =
-      fa.traverse(f)
+  implicit def catsDataTraverseForConst[C]: TraverseFilter[Const[C, ?]] = new TraverseFilter[Const[C, ?]] {
```
kailuowang: nit: does the name need to be consistent with the pattern, i.e. `catsDataTraverseFilterForConst`?

ceedubs: Good catch. I think I might have this in a couple places in this PR. These changes area already not binary compatible, so there's no reason not to change this.

kailuowang:
Again, this looks awesome. Thank you @ceedubs 👍

core/src/main/scala/cats/MonadFilter.scala
```diff
@@ -4,15 +4,15 @@ import simulacrum.typeclass
...
- * create an "Empty" value for the Monad (for whatever "empty" makes
+ * create an "empty" value for the Monad (for whatever "empty" makes
  * sense for that particular monad). This is of particular interest to
  * us since it allows us to add a `filter` method to a Monad, which is
  * used when pattern matching or using guards in for comprehensions.
  */
-@typeclass trait MonadFilter[F[_]] extends Monad[F] {
+@typeclass trait MonadFilter[F[_]] extends Monad[F] with FunctorFilter[F] {
 
   def empty[A]: F[A]
```
johnynek: It seems like a FunctorFilter had to have an empty value too. We could just filter everything from any value and get there, no?

Seems like empty might almost be there.

core/src/main/scala/cats/FunctorFilter.scala
```diff
@@ -0,0 +1,62 @@
...
+   * {{{
+   * scala> import cats.implicits._
+   * scala> val m: Map[Int, String] = Map(1 -> "one", 3 -> "three")
+   * scala> val l: List[Int] = List(1, 2, 3, 4)
+   * scala> def asString(i: Int): Option[String] = m.get(i)
+   * scala> l.mapFilter(i => m.get(i))
+   * res0: List[String] = List(one, three)
+   * }}}
+   */
+  def mapFilter[A, B](fa: F[A])(f: A => Option[B]): F[B]
```
johnynek: I wonder if it is worth thinking about a typeclass for mapConcat? The law would be similar to the one here. Collections support this but also distributed compute libraries like spark and scalding.

laws/src/main/scala/cats/laws/discipline/TraverseFilterTests.scala
```diff
@@ -0,0 +1,42 @@
...
+    EqFA: Eq[F[A]],
+    EqFC: Eq[F[C]],
+    EqM: Eq[M],
+    EqXYFC: Eq[X[Y[F[C]]]],
+    EqXFA: Eq[X[F[A]]],
+    EqXFB: Eq[X[F[B]]],
+    EqYFB: Eq[Y[F[B]]]
+  ): RuleSet = {
+    new RuleSet {
+      def name: String = "collect"
```
mikejcurry: RuleSet name should probably be `traverseFilter` for consistency.

core/src/main/scala/cats/FunctorFilter.scala
```diff
@@ -0,0 +1,62 @@
...
+
+  /**
+   * Similar to [[mapFilter]] but uses a partial function instead of a function
+   * that returns an `Option`.
+   *
+   * Example:
+   * {{{
+   * scala> import cats.implicits._
+   * scala> val l: List[Int] = List(1, 2, 3, 4)
+   * scala> TraverseFilter[List].collect(l){
```
mikejcurry: Should this refer to `FunctorFilter` here instead of `TraverseFilter` as this is the docs for `FunctorFilter`.

core/src/main/scala/cats/FunctorFilter.scala
ceedubs: @johnynek what's `mapConcat`?

core/src/main/scala/cats/FunctorFilter.scala
ceedubs: Thanks for catching my cut/paste error :)

core/src/main/scala/cats/MonadFilter.scala
ceedubs: > We could just filter everything from any value and get there, no?

That's true, but you would have to have a value to start with. With `Applicative` you have `pure` for that, but with `Functor` this isn't available.

I was thinking about whether the `empty` from `MonadFilter` should be moved to `FunctorFilter` along with the `filter` method at first, but I think this constrains `FunctorFilter` more than we might want. For example, that would mean that the `FunctorFilter` instance for `OptionT[F, A]` would then require an `Applicative[F]` as opposed to the `Functor[F]` that it currently requires. And more generally, I think that what I have in this PR as a `composeFilter` method on `Functor` would need to instead be moved to `Applicative`.

I think that if we were to add an `ApplicativeFilter`, _that_ would be the place to add the `def empty[A]: F[A]`. With `ApplicativeFilter`, you could even implement `empty` and `pure` in terms of each other (with help from `map`).

At this point I haven't added `ApplicativeFilter` because I didn't have a particularly compelling use-case for it and there's already a lot going on in this type class hierarchy. However, I should be able to add it if people have good use-cases for it.

laws/src/main/scala/cats/laws/discipline/TraverseFilterTests.scala
ceedubs: Good catch, @mikejcurry! This was left over from when I had used `Collect` as this type class name at first.

ceedubs:
I've updated with a couple typo fixes based on feedback. I've also added a `TraverseFilter` instance for `Map`.

core/src/main/scala/cats/FunctorFilter.scala
johnynek: `def mapConcat[A, B](fa: F[A])(f: A => Iterable[B]): F[B]` something like that.

For instance there is no `FunctorMapConcat[Option]` but there is `FunctorMapConcat[Vector]` and in the case of scalding `FunctorMapConcat[TypedPipe]`.

johnynek: Actually, when you think about it, it might be something like this:
```scala
trait FunctorArrow[F[_], A[_, _]] extends Functor[F] {
  def arrow: Arrow[A]
  def mapArrow[T, U](f: F[T], arrow: A[T, U]): F[U]
}
```
And we are just talking about various `KleisliArrows` here.

Or maybe we mean something like:

```scala
trait FunctorTraverse[F[_], M[_]] extends Functor[F] {
  // this is similar to traverse, except we end up with `F[B]` not `M[F[B]]`
  def mapTraverse[A, B](fa: F[A])(f: A => M[B]): F[B]
}
```
It is not clear what the right level of generalization is. It seems to me `A => M[B]` where you have `Foldable[M]` is pretty generally useful. Maybe that is the thing you want instead of mapConcat:
```scala
trait FunctorFoldableFn[F[_]] extend Functor[F] {
  def mapFoldable[A, B, M[_]](f: F[A])(f: A => M[B])(implicit M: Foldable[M]): F[A]
}
```

ceedubs: There was some similar discussion a long time ago [here](https://github.com/typelevel/cats/issues/842#issuecomment-178583594). I wonder what the valid laws for this construct would be and whether you would run into issues where just working with `Foldable` would mean implementations would tend to be inefficient.

@johnynek were you suggesting something like this instead of `FunctorFilter` or `TraverseFilter`? I would tend to think that something like this should be in addition to these type classes. As you've mentioned you couldn't have an instance of this type class for `Option` or `OptionT`, which I think is a pretty big limitation, they are common types to call `filter` or `collect` on. I also feel like there is some benefit to knowing these methods can never make your structure grow (they can only reduce it or leave it the same); however this is just a fuzzy feeling and not something that I can confidently say is a useful property.

adelbertc:
Wow, that's quite a lot of work. LGTM 👍

kailuowang:
👍

core/src/main/scala/cats/FunctorFilter.scala
johnynek: I was mostly thinking out loud to figure out what this should be.

johnynek:
👍  Thanks for the hard work here indeed.

ceedubs:
Thanks for the great review comments, @johnynek, @kailuowang, and @mikejcurry!

PR #1226, opened 2016-07-23
Add ListT section to FAQ

ceedubs:


adelbertc:
👍

docs/src/main/tut/faq.md
```diff
@@ -36,6 +37,47 @@ If you have already followed the [imports advice](#what-imports) but are still g
...
+```
+
+```tut:book
+l.map(_.map(_ + 1))
+l.exists(_.exists(isEven))
+```
+
+A naive implementation of `ListT` suffers from associativity issues; see [this gist](https://gist.github.com/tpolecat/1227e22e3161b5816e014c00650f3b57) for an example. It's possible to create a `ListT` that doesn't have these issues, but it tends to be pretty inefficient. For many use-cases, [Nested](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/Nested.scala) can be used to achieve the desired results.
+
+Here is how we could achieve the effect ofthe previous example using `Nested`:
```
kailuowang: s/offthe/of the/

kailuowang:
:+1:  apart from the missing space.

ceedubs:
By the way @adelbertc I think that these examples show a bit of how useful your `Nested` contribution can be. Thanks for that!

Issue #556, opened 2015-10-02
Modify XorT and OptionT variance


frosforever:
`Xor` is currently covariant while `XorT` is invariant.

Assuming one has the following:

``` scala
sealed trait Error
case class ErrorA(message: String) extends Error
case object ErrorB extends Error
```

with `Xor` one can have something that looks like:

``` scala
val t: Xor[Error, Int] = for {
  i <- Xor.left[ErrorB.type, Int](ErrorB)
  r <- Xor.left[ErrorA, Int](ErrorA("foo"))
} yield r
```

However the same using `XorT` will not compile:

``` scala
val noCompile: XorT[List, Error, Int] = for {
  i <- XorT(List(Xor.left[ErrorB.type, Int](ErrorB)))
  r <- XorT(List(Xor.left[ErrorA, Int](ErrorA("foo"))))
} yield r
```

and one is instead forced to use

``` scala
val tf: XorT[List, Error, Int] = for {
  i <- XorT(List(Xor.left[Error, Int](ErrorB)))
  r <- XorT(List(Xor.left[Error, Int](ErrorA("foo"))))
} yield r
```

Because `Xor` is covariant, the non-monad-transformer version of `noCompile` compiles just fine:

``` scala
val compiles: List[Xor[Error, Int]] = List(Xor.left[ErrorB.type , Int](ErrorB)).flatMap { e1 =>
  List(Xor.left[ErrorA, Int](ErrorA("foo"))).map { e2 =>
    e1.flatMap { i =>
      e2.map { r =>
        r
      }
    }
  }
}
val compilesXorT = XorT[List, Error, Int](compiles)
```

I do not know which one is more correct but it is confusing for `Xor` and `XorT` not to behave in the same way. If this is indeed the expected and correct behavior perhaps we can add some documentation to prevent confusion.

Note: the same can be said for `OptionT` though I came across this using `Xor` and so these are the examples I've got.

ceedubs:
Thanks for raising this issue and providing a detailed explanation @frosforever! I agree that we should do something about this, even if that "something" is just better documentation. Here is a bit of a brain dump on my thoughts on the matter.

In Scala, variance has both positives and negatives (perhaps that's how they decided on variance notation! :P). These pros/cons have been discussed numerous times in various venues, so I won't go into them now, but in my opinion at the end of the day you kind of have to settle on "to each their own".

[...]

PR #1230, opened 2016-07-25
Adding leftWiden to XorT

zainab-ali:
Adding a `leftWiden` method to `XorT`, as `XorT` is invariant on errors, as referenced in #556

### DIFF
--- a/core/src/main/scala/cats/data/XorT.scala
+++ b/core/src/main/scala/cats/data/XorT.scala
@@ -93,6 +93,8 @@ final case class XorT[F[_], A, B](value: F[A Xor B]) {
 
   def leftMap[C](f: A => C)(implicit F: Functor[F]): XorT[F, C, B] = bimap(f, identity)
 
+  def leftWiden[AA >: A](implicit F: Functor[F]): XorT[F, AA, B] = leftMap(a => a)
+
   def compare(that: XorT[F, A, B])(implicit o: Order[F[A Xor B]]): Int =
     o.compare(value, that.value)
 
--- a/core/src/main/scala/cats/data/XorT.scala
+++ b/core/src/main/scala/cats/data/XorT.scala
@@ -93,8 +93,6 @@ final case class XorT[F[_], A, B](value: F[A Xor B]) {
 
   def leftMap[C](f: A => C)(implicit F: Functor[F]): XorT[F, C, B] = bimap(f, identity)
 
-  def leftWiden[AA >: A](implicit F: Functor[F]): XorT[F, AA, B] = leftMap(a => a)
-
   def compare(that: XorT[F, A, B])(implicit o: Order[F[A Xor B]]): Int =
     o.compare(value, that.value)
 
--- a/core/src/main/scala/cats/functor/Bifunctor.scala
+++ b/core/src/main/scala/cats/functor/Bifunctor.scala
@@ -30,6 +30,20 @@ trait Bifunctor[F[_, _]] extends Any with Serializable { self =>
       val F = self
       val G = G0
     }
+
+  /**
+   * Widens A into a supertype AA.
+   * Example:
+   * {{{
+   * scala> import cats.data.Xor
+   * scala> import cats.implicits._
+   * scala> sealed trait Foo
+   * scala> case object Bar extends Foo
+   * scala> val x1: Xor[Bar.type, Int] = Xor.left(Bar)
+   * scala> val x2: Xor[Foo, Int] = x1.leftWiden
+   * }}}
+   */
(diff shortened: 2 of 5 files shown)


adelbertc:
Should this perhaps be a cast instead? https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Functor.scala#L32

Also maybe a quick doctest?

ceedubs:
Or should we even perhaps add this as a method on `Bifunctor` like `widen` is on `Functor`?

adelbertc:
@ceedubs ah yes, that too :-)

core/src/main/scala/cats/data/XorT.scala
```diff
@@ -93,6 +93,8 @@ final case class XorT[F[_], A, B](value: F[A Xor B]) {
 
   def leftMap[C](f: A => C)(implicit F: Functor[F]): XorT[F, C, B] = bimap(f, identity)
 
+  def leftWiden[AA >: A](implicit F: Functor[F]): XorT[F, AA, B] = leftMap(a => a)
```
johnynek: will `leftMap(identity)` not work here?

johnynek:
Shouldn't we have `rightWiden` on `Bifunctor` too? Seems strange to only have `leftWiden`.

zainab-ali:
I'm not too sure about `rightWiden`.  All of the cats data types which have a `Bifunctor` also have a `Functor` on the "right" (I think).  Hence, they can use the `widen` on `Functor`.  
It would be useful to widen something which had a `Bifunctor[A, B]`, and no `Functor[B]`, but I can't think of a data type that needs it.

zainab-ali:
@adelbertc I've addded a doctest, but I'm unsure if it's getting run as it doesn't generate logs or coverage during testing.  It's definitely being generated locally, and is compiling.

johnynek:
@zainab-ali is that's the case, why does `Bifunctor` has `rightMap`? Why not just `map`? Seems inconsistent to me to have `leftMap` and `leftWiden` but only `rightMap`.

ceedubs:
@johnynek I see what you are saying, but I think at this point `rightMap` is the only method from `Functor` that we are putting on `Bifunctor`. If we were to go down the path that you suggested wouldn't we also be adding `rightLift`, `rightVoid`, `rightFproduct` (capitalization?), and `rightAs`?

johnynek:
I'm confused, I only see a few methods here:
https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/functor/Bifunctor.scala#L8

`bimap, leftMap, rightMap, compose`. I think what I am saying is if we add to that set of 4 methods, `leftWiden` I would suggest `rightWiden`.

Am I missing that there are more methods?

Another approach if we don't like the duplication: what about adding:
```scala
  def leftFunctor[B]: Functor[F[?, B]]
  def rightFunctor[A]: Functor[F[A, ?]]
``` 

Then you can just do `leftFunctor[B].widen[AA](f)` or `rightFunctor[A].widen[BB](f)`

johnynek:
But I still feel if we don't want to add `rightWiden` we should remove `rightMap` and somehow have the principle that aside from `bimap` we are just going to have the `left`-biased methods since otherwise a right-biased Functor should work.

ceedubs:
@johnynek I see what you are saying. I wouldn't be opposed to removing `rightMap`. I think exposing each individual `Functor` is reasonable (I thought about submitting a PR for this at some point in the past but never did). Even if we decide we are going to only have left-biased methods in `Bifunctor` (other than `bimap`), I'd still be inclined to prefix them with `left` to avoid name collisions with `Functor` syntax.

johnynek:
+1 to leaving `left` as a prefix.

+1 to dropping rightMap

(especially before `0.7.0`).

johnynek:
could we possibly merge this dropping `rightMap` before 0.7.0?

I'd like to get incompatible changes behind us ASAP (or I will start advocating for keeping rightMap forever)

johnynek:
👍

johnynek:
@ceedubs can we merge this now? Do you want to see anything else?

non:
FWIW I'm 👍 on this as well.

johnynek:
two +1 so, I think I'm good to merge. Thank you @zainab-ali !

ceedubs:
Oops sorry for being late to the show. 👍

Do we want to open an issue to expose the left and right functors on `Bifunctor`? It's something that I've never used.

johnynek:
I don't know. I might punt until someone felt like it was missing. I kind of think the person that wrote the `Bifunctor` should have given each of the `Functors`, but maybe there is an application where you have a `Bifunctor` and you want a `Functor` instance with minimal work?

I wouldn't oppose it, but the more I thought about it the more I was on the "punt" side.

Issue #1089, opened 2016-06-03
Create cats.data.NonEmptyStream?


ceedubs:
Once #1087 and #1088 have been completed, I think that we should probably drop `OneAnd` from cats. At that point, I think that `NonEmptyStream` would be the only data structure referencing it. We can probably drop `NonEmptyStream` at that point, but I created this issue to let people weigh on if `NonEmptyStream` is something that they would really want in cats. I suspect that it's not really used.

wjlow:
I'm at the point where I would like to have NonEmptyStream currently.

I have a `Stream[A]` at the moment that I know is never empty and `A` is a `Semigroup` without a `Monoid`. Would be good to be able to `reduce` it safely without having to do `stream.reduce(_ combine _)`, which is unsafe.

johnynek:
@wjlow or you can use OneAnd right? https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/OneAnd.scala

wjlow:
Yes, thanks!

PR #1231, opened 2016-07-25
Finish up work on NonEmptyList

ceedubs:
This continues work started by @WarFox in #1120 (see [this comment](https://github.com/typelevel/cats/pull/1120#issuecomment-234737694)).

At this point, I have left `OneAnd` in place. However, I think that
after merging this we may want to delete it. In practice it's pretty
awkward to use and sometimes prevents performant operations. See also #1089.


ceedubs:
It looks like I have some unit tests that I should add before this gets merged. I can probably add them tonight or tomorrow (Eastern time). I welcome code reviews before then.

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -0,0 +1,193 @@
...
+import cats.syntax.eq._
+
+import scala.annotation.tailrec
+import scala.collection.mutable.ListBuffer
+
+/**
+ * A data type which represents a non empty list of A, with
+ * single element (head) and optional structure (tail).
+ */
+final case class NonEmptyList[A](head: A, tail: List[A]) {
```
johnynek: what if we did something similar to `NonEmptyVector` and made this a value class around `::[A]`? Then we would not need to allocate in some cases.

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -0,0 +1,193 @@
...
+
+  /**
+   * Return the head and tail into a single list
+   */
+  def toList: List[A] = head :: tail
+
+  /**
+   *  Applies f to all the elements of the structure
+   */
+  def map[B](f: A => B): NonEmptyList[B] =
```
johnynek: if we did the `::[A]` trick, this would require a cast. Maybe the real class is the way to go.

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -0,0 +1,193 @@
...
+    toList.iterator.map(A.show).mkString("NonEmptyList(", ", ", ")")
+
+  override def toString: String = s"NonEmpty$toList"
+}
+
+object NonEmptyList extends NonEmptyListInstances {
+  def apply[A](head: A, tail: A*): NonEmptyList[A] = NonEmptyList(head, tail.toList)
+
+  def fromList[A](l: List[A]): Option[NonEmptyList[A]] =
+    if (l.isEmpty) None else Some(NonEmptyList(l.head, l.tail))
```
johnynek: what about:
```scala
l match {
  case Nil => None
  case h :: t => Some(NonEmptyList(h, t))
}
```
to make more clear that `.head` and `.tail` are safe?

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -0,0 +1,193 @@
...
+
+  def fromListUnsafe[A](l: List[A]): NonEmptyList[A] =
```
johnynek: what about
```scala
l match {
  case Nil => throw new IllegalArgumentException("Cannot create NonEmptyList from empty list")
  case h :: t => NonEmptyList(h, t)
}
```

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -0,0 +1,193 @@
...
+    Foldable[List].foldRight(toList, lb)(f)
+
+  /**
+   * Left-associative reduce using f.
+   */
+  def reduceLeft(f: (A, A) => A): A =
+    tail.foldLeft(head)(f)
+
+  def coflatMap[B](f: NonEmptyList[A] => B): NonEmptyList[B] = {
+    @tailrec def consume(as: List[A], buf: ListBuffer[B]): List[B] =
```
johnynek: why pass the mutable `buf`? Can we capture to make more clear there is only one:
```scala
val buf = ListBuffer.empty
@tailrec def consume(as: List[A]): List[B] = {
...
}
```

core/src/main/scala/cats/data/package.scala
```diff
@@ -1,33 +1,14 @@
...
-  def NonEmptyList[A](head: A, tail: A*): NonEmptyList[A] =
-    OneAnd[List, A](head, tail.toList)
-
   def NonEmptyStream[A](head: A, tail: Stream[A] = Stream.empty): NonEmptyStream[A] =
     OneAnd(head, tail)
   def NonEmptyStream[A](head: A, tail: A*): NonEmptyStream[A] =
     OneAnd(head, tail.toStream)
 
-  object NonEmptyList {
-    def fromReducible[F[_], A](fa: F[A])(implicit F: Reducible[F]): Eval[NonEmptyList[A]] =
```
johnynek: can we not move this to the `NonEmptyList` companion?

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -0,0 +1,193 @@
...
+      def flatMap[A, B](fa: NonEmptyList[A])(f: A => NonEmptyList[B]): NonEmptyList[B] =
+        fa flatMap f
+
+      def coflatMap[A, B](fa: NonEmptyList[A])(f: NonEmptyList[A] => B): NonEmptyList[B] =
+        fa coflatMap f
+
+      def extract[A](fa: NonEmptyList[A]): A = fa.head
+
+      def traverse[G[_], A, B](fa: NonEmptyList[A])(f: (A) => G[B])(implicit G: Applicative[G]): G[NonEmptyList[B]] =
+        G.map2Eval(f(fa.head), Always(Traverse[List].traverse(fa.tail)(f)))(NonEmptyList(_, _)).value
```
johnynek: why not have a `.traverse` method on `NonEmptyList`? Why is this different from the others?

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -0,0 +1,193 @@
...
+    NonEmptyList(head, tail ++ l)
+
+  def flatMap[B](f: A => NonEmptyList[B]): NonEmptyList[B] =
+    f(head) ++ tail.flatMap(f andThen (_.toList))
+
+  /**
+   * remove elements not matching the predicate
+   */
+  def filter(p: A => Boolean): List[A] =
+    toList.filter(p)
```
johnynek: this is maybe a little slower than:
```scala
def filter(p: A => Boolean): List[A] = {
  val ftail = tail.filter(p)
  if (p(head)) head :: ftail
  else ftail
}
```

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -0,0 +1,193 @@
...
+  /**
+   * Append another NonEmptyList
+   */
+  def concat(other: NonEmptyList[A]): NonEmptyList[A] =
+    NonEmptyList(head, tail ::: other.toList)
+
+  /**
+   * Find the first element matching the predicate, if one exists
+   */
+  def find(p: A => Boolean): Option[A] =
```
johnynek: what about:
```scala
if (p(head)) Some(head)
else tail.find(p)
```

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -0,0 +1,193 @@
...
+  def reduceLeft(f: (A, A) => A): A =
+    tail.foldLeft(head)(f)
+
+  def coflatMap[B](f: NonEmptyList[A] => B): NonEmptyList[B] = {
+    @tailrec def consume(as: List[A], buf: ListBuffer[B]): List[B] =
+      as match {
+        case Nil => buf.toList
+        case a :: as => consume(as, buf += f(NonEmptyList(a, as)))
+      }
+    NonEmptyList(f(this), consume(this.tail, ListBuffer.empty))
```
johnynek: why `this.tail`?

core/src/main/scala/cats/data/NonEmptyList.scala
ceedubs: One thing that we could do would be to define it as a `sealed abstract class` so we have some flexibility in changing it in the future. However, we still wouldn't be able to change it into a value class without breaking binary compatibility.

core/src/main/scala/cats/data/package.scala
ceedubs: Good catch.

ceedubs: I'm actually inclined to add this as a method on the `Reducible` type class so that `NonEmptyList` can override it with an efficient implementation. Any objections?

ceedubs:
Thanks for the great review, @johnynek! I've addressed many of your points. So far I haven't changed the representation from a case class with a `head` and a `tail`, because there didn't seem to be a strong inclination there. I'm open to ideas.

core/src/main/scala/cats/data/package.scala
kailuowang: I am thinking if we add a `toNonEmptyList`, why not a `toNonEmptyVector`, to that spirit, I feel there might be a need for a new typeclass for such conversions?  or does such a typeclass have to be less convenient than having the method in either `Reducible` or `NonEmptyList`?

core/src/main/scala/cats/Reducible.scala
```diff
@@ -113,6 +115,11 @@ import simulacrum.typeclass
   def sequence1_[G[_], A](fga: F[G[A]])(implicit G: Apply[G]): G[Unit] =
     G.map(reduceLeft(fga)((x, y) => G.map2(x, y)((_, b) => b)))(_ => ())
 
+  def toNonEmptyList[A](fa: F[A]): NonEmptyList[A] =
+    reduceRightTo(fa)(a => NonEmptyList(a, Nil)) { (a, lnel) =>
+      lnel.map { case NonEmptyList(h, t) => NonEmptyList(a, h :: t) }
```
johnynek: should we add `a :: nel` to make a new NonEmptyList?

```scala
def ::(a: A): NonEmptyList[A] = NonEmptyList(a, head :: tail)
```

core/src/main/scala/cats/data/package.scala
ceedubs: @kailuowang I agree that there should be a `toNonEmptyVector`. I think I'd like to opt to put that into a follow-up PR though.

As far as a type class for these conversions, do you have any ideas for improvements above `Reducible`? I like `Reducible` because it is a pretty general abstraction.

core/src/main/scala/cats/Reducible.scala
ceedubs: Added.

core/src/main/scala/cats/data/package.scala
kailuowang: @ceedubs  I felt a bit uneasy with `Reducible` having a dependency on the `NonEmptyList` data structure.  Then I just realized that we have `toList` in `Foldable`, now it seems all natural that we have `toNonEmptyList` in `Reducible`.  
Not sure if we need `toNonEmpty(Vector|Stream)` in the `Reducible`.

kailuowang:
👍

johnynek:
This looks great! Thanks! 👍

travisbrown:
👍 from me! Looking forward to this.

core/src/main/scala/cats/data/package.scala
johnynek: I tend to somewhat agree on this. I'm not sure it is worth putting `to(|NonEmpty)(Vector|List|Stream)` in `Foldable|Reducible`.

I would be somewhat more inclined to put them on the companion objects of the data structure.

The general idea is something like:
```scala
trait CollectionyThingy[C[_]] {
  def singleton[T](t: T): C[T]
  // maybe want some Eval below
  def cons[T](head: T, tail: C[T]): C[T]
}

trait MaybeEmptyCollection[C[_]] extends CollectionyThingy[C] {
  def empty[T]: C[T]
}
```
but so far that seems lawless except `cons(a, empty) == singleton(a)`

ceedubs: My main motivation for putting them in the type class as opposed to companion objects is to allow for efficient overrides. Having said that, I think it's actually probably pretty rare that you would want to ask for a `Reducible[F]` and then immediately call `.toNonEmptyList` on it. You can probably do whatever you want with the `Reducible` constraint (or another type class constraint that you add).

I don't feel too strongly about whether this should exist, so I don't mind taking it out if people would prefer. Though due to unfortunate binary compatibility with traits in Scala, it isn't one of those cases where "it's always easy to add it later" works :(

kailuowang: I vote for let's merge this now and have a separate discussion on whether it's worth having `to(|NonEmpty)(Vector|List|Stream)`  in `Foldable|Reducible`.

adelbertc:
Nice! LGTM 👍

PR #1234, opened 2016-07-27
Remove Applicative#pureEval

### LINKED ISSUE #1150: pureEval vs a typeclass
Apologies if this was discussed, but I wonder why `pureEval` was added to `Applicative` vs something like:

```scala
trait ApplicativeEval[F[_]] extends Applicative[F] {
  def pureEval[A](p: Eval[A]): F[A]
}
```
with the law that `Eq[F[A]].equiv(pureEval(e), pure(e.value))`

Seems like making that always present but by default forcing `.value` it removes the ability of the caller of `pureEval` to know if `value` is called immediately or not. With a new type, it would not be optional laziness, it would advertise true laziness support.

Maybe this is not a super meaningful distinction (unlike `tailRecM` which can be implemented in a non-tailrecursive way with flatMap), but still it seemed easy enough to signal at the type level.

ceedubs:
This probably resolves #1150, though it doesn't provide an alternative
type class as was brought up there.

@adelbertc has [gotten tripped up](https://github.com/typelevel/cats/issues/1150#issuecomment-235322721) by `pureEval` defaulting to a call to `pure` with `.value` called on the `Eval`.

@johnynek has also [questioned](https://github.com/typelevel/cats/issues/1150#issue-161092646) `pureEval` being directly on `Applicative`.

@tpolecat has [raised concerns](https://gitter.im/typelevel/cats?at=5794dc4c1b9de56c0edd887a) that `pureEval` looks like it might exist as a way to inject
side effects into a type, but that it doesn't actually make any claims
about execution semantics or value retention, so it can't be relied upon
to do this.

I think that there could be an argument for having an instantiation method for
a side-effect-y type (like a `Task`) that takes an `Eval[A]` and produces an instance
that will repeat the computation given `Always`, memoize the value given
`Later`, etc. However, I don't think that @tpolecat necessarily agrees
with me, and I think that even _if_ this is a reasonable thing to do,
`Applicative` probably isn't the place for it.

Since several people have questioned `pureEval`, it was entirely
untested, and I haven't heard any one argue for it staying, I'm inclined
to remove it. If someone has a compelling example of when a non-strict
argument can be useful for `pure`, then I may be convinced. But even
then we should clearly document what behavior can be expected of
`pureEval`.

@non I believe that you originally added `pureEval` as part of your `StreamingT` work. Do you have any thoughts on this?

### DIFF
--- a/core/src/main/scala/cats/Applicative.scala
+++ b/core/src/main/scala/cats/Applicative.scala
@@ -22,13 +22,6 @@ import simulacrum.typeclass
    */
   def pure[A](x: A): F[A]
 
-  /**
-   * `pureEval` lifts any value into the Applicative Functor.
-   *
-   * This variant supports optional laziness.
-   */
-  def pureEval[A](x: Eval[A]): F[A] = pure(x.value)
-
   override def map[A, B](fa: F[A])(f: A => B): F[B] =
     ap(pure(f))(fa)
 
--- a/core/src/main/scala/cats/Eval.scala
+++ b/core/src/main/scala/cats/Eval.scala
@@ -299,7 +299,6 @@ private[cats] trait EvalInstances extends EvalInstances0 {
     new Bimonad[Eval] with MonadRec[Eval] {
       override def map[A, B](fa: Eval[A])(f: A => B): Eval[B] = fa.map(f)
       def pure[A](a: A): Eval[A] = Now(a)
-      override def pureEval[A](la: Eval[A]): Eval[A] = la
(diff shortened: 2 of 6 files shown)


johnynek:
👍

tpolecat:
👍

adelbertc:
:+1:

kailuowang:
👍

kailuowang:
I apologize @ceedubs and @non, I forgot that @non is mentioned here as well, my bad. I will do the fix if needed.

PR #1236, opened 2016-07-27
Update to Scala.js 0.6.11

fthomas:


ceedubs:
👍 thanks! I saw the announcement and was thinking that we should do this.

ceedubs:
oh @fthomas could you also update the scala.js badge in the readme?

fthomas:
I tried updating the badge, but http://www.scala-js.org/assets/badges/scalajs-0.6.11.svg returns 404. There is no badge for any version above 0.6.8.

fthomas:
http://www.scala-js.org/libraries/ explains why 0.6.8+ is still fine:

> Show your true colors by adding a Scala.js badge on your site! Select the lowest Scala.js version your library is compatible with.

> If you don’t find your version of Scala.js, that’s probably because it is forward-compatible with its predecessor, in which case you should take the latest available version older than your version of Scala.js. For example, for Scala.js 0.6.7, use the 0.6.6+ badge.

ceedubs:
@fthomas ah, thanks for the explanation! 👍

kailuowang:
👍

PR #1238, opened 2016-07-28
Add bisequence and bimap examples

ceedubs:
This was originally part of #1194, but since it's looking uncertain
whether or not that will go forward, I'm submitting it separately.


adelbertc:
:+1: on Travis green

fthomas:
@ceedubs All result variables have the name `res0` which could be confusing. Could you rename them to `res0`, `res1`, `res2`, etc. so that the example looks like a real REPL session?

ceedubs:
Good catch @fthomas. Fixed.

fthomas:
:+1:

PR #1240, opened 2016-07-28
Add distinct method to NonEmptyList and NonEmptyVector.

Tvaroh:
Hello.

Does it make sense to have `distinct` method on `NonEmptyList`? If this is something that could be accepted I might work on unit tests for this.

The implementation is based on universal equality to compare values. Using `Eq` would require a typeclass for getting hash codes and some `Set` implementation that would use those.

### DIFF
--- a/core/src/main/scala/cats/data/NonEmptyList.scala
+++ b/core/src/main/scala/cats/data/NonEmptyList.scala
@@ -5,6 +5,7 @@ import cats.instances.list._
 import cats.syntax.order._
 
 import scala.annotation.tailrec
+import scala.collection.immutable.TreeSet
 import scala.collection.mutable.ListBuffer
 
 /**
@@ -104,6 +105,20 @@ final case class NonEmptyList[A](head: A, tail: List[A]) {
 
   def show(implicit A: Show[A]): String =
     toList.iterator.map(A.show).mkString("NonEmptyList(", ", ", ")")
+
+  /**
+   * Remove duplicates. Duplicates are checked using `Order[_]` instance.
+   */
+  def distinct(implicit O: Order[A]): NonEmptyList[A] = {
+    implicit val ord = O.toOrdering
+
+    val buf = ListBuffer.empty[A]
+    tail.foldLeft(TreeSet(head)) { (elementsSoFar, a) =>
+      if (elementsSoFar(a)) elementsSoFar else { buf += a; elementsSoFar + a }
+    }
+
+    NonEmptyList(head, buf.toList)
+  }
 
   override def toString: String = s"NonEmpty$toList"
 }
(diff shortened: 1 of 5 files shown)


ceedubs:
Thanks for submitting this @Tvaroh! I was thinking about wanting a `distinct` on `NonEmptyList`/`NonEmptyVector` just the other day.

There are a few complications:
- We've recently added first-class [NonEmptyList](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/NonEmptyList.scala) and [NonEmptyVector](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/NonEmptyVector.scala) implementations. Now that those are in place, I think we are likely going to drop `OneAnd` (see #1089).
- As you have brought up, this relies on universal equality. This seems a bit unnatural in Cats, where most things are type class based. However, as you also mentioned, we don't have a hash type class. One alternative would be to take an `Order` instance and use a tree-based set, but we don't even have one of those.

I'm not quite sure how to go forward. I wonder if anyone else has any thoughts on this?

Tvaroh:
@ceedubs I understand. Writing a balanced ordered tree-based set might be complicated, not sure if it's even in cats' scope.

adelbertc:
Short of writing our own tree, perhaps we can use `scala.collection.immutable.TreeSet` under the covers which requires a `scala.math.Ordering`, which we can get from our own `Order` ? Just need to be careful that none of the existing instances of `Ordering` are inconsistent with `Order`. I'd imagine they're not.

Tvaroh:
Gents, I've updated the PR with `distinct` on `NonEmptyList` and `NonEmptyVector` using `TreeSet` to keep track of duplicate elements. I tried to do it efficiently, so the implementation uses mutable builders (`ListBuffer` and `Vector.newBuilder`) underneath.

Let me know if this is acceptable and I can work on unit tests then.

Tvaroh:
Have added tests and resolved conflicts with master. Unfortunately `sbt test` hangs forever for me, do you run it using some specific options?

ceedubs:
@Tvaroh I usually use `testsJVM/test`.

Tvaroh:
@ceedubs thanks, it works and tests pass.

ceedubs:
This approach seems reasonable to me. I'd like for a couple others to weigh in on this.

@Tvaroh could you please remove the unrelated formatting changes from your PR? They don't follow the conventions used elsewhere in the code base.

I could see it being useful to have a `def distinctBy[B](f: A => B)(implicit O: Order[B])` generalization with `distinct` being a special case for `f` being `identity`. That could also wait for a separate PR though.

Tvaroh:
@ceedubs omg, didn't notice that, blame Intellij. Let me better create a new PR from master to not pollute the git log with unrelated changes.

Tvaroh:
See #1243.

Issue #1214, opened 2016-07-17
NonEmptyVector unsafe apply method


peterneyens:
With the `apply` method we get because `NonEmptyVector` is a case class, we can create an empty `NonEmptyVector`.

```scala
scala> NonEmptyVector(Vector.empty[Int])
res1: cats.data.NonEmptyVector[Int] = NonEmptyVector(Vector())

scala> res1.head
java.lang.UnsupportedOperationException: empty.head
  at scala.collection.immutable.Vector.head(Vector.scala:193)
  at cats.data.NonEmptyVector.head(NonEmptyVector.scala:31)
  ... 42 elided
```

Should we turn it into a regular class with the following private `apply` method ?

```scala
private def apply[A](vector: Vector[A]): NonEmptyVector[A] =
  new NonEmptyVector(vector)
```

ceedubs:
Ouch, good catch. For some reason I was thinking that the generated `apply` was private when a case class constructor is marked private.

> Should we turn it into a regular class

This is beginning to seem like the right solution to me. If it stays a case class we'll need to override both the `apply` and the `copy` method to make them `private`. And even then we don't _really_ want either of them to exist, because we'd prefer to just have `fromVectorUnsafe`.

We'll need to make sure that we do the right thing for `equals`, `toString`, and `hashCode` (where the easiest thing to do is probably to delegate through to the underlying `Vector` for each, and probably to prepend `NonEmpty` to the `toString`).

> with the following private apply method?

As mentioned above, I think we might just want `fromVectorUnsafe` instead.

mikejcurry:
I believe this is resolved by #1212

PR #1245, opened 2016-07-30
Additional test to ensure that semantics of apply on NonEmptyVector remains as expected

mikejcurry:
@kailuowang @ceedubs - Apologies I didn't get back to the [comment](https://github.com/typelevel/cats/pull/1212#discussion_r72367628) by @kailuowang on the test name prior to the PR being merged.

Here is a follow up to address the text of the test. I have renamed it a different way than was suggested, and I have also added an additional test.

The original intent of the test was that it was a test to ensure that supplying a` Vector[Int]` to the `apply` method could not result in a `NonEmptyVector[Int] `(rather it should result in a `NonEmptyVector[Vector[Int]]`. 

As noted in #1214, supplying an empty `Vector[Int] `to apply was causing an empty `NonEmptyVector[Int]` to be created. The new test more explicitly verifies that this cannot happen.

These might be overkill as far as test goes, but given #1214, I figured the tests would be useful. Let me know what you think.

### DIFF
--- a/tests/src/test/scala/cats/tests/NonEmptyVectorTests.scala
+++ b/tests/src/test/scala/cats/tests/NonEmptyVectorTests.scala
@@ -254,10 +254,13 @@ class NonEmptyVectorTests extends CatsSuite {
     }
   }
 
-  test("Cannot create a new NonEmptyVector from apply with an empty vector") {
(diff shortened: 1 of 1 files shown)


kailuowang:
👍  Thanks!

ceedubs:
👍

Issue #1064, opened 2016-05-27
Remove boilerplate for n-ary typeclasses


yilinwei:
Since simulacrum now supports [n-ary typeclass constructors](https://github.com/mpilquist/simulacrum/pull/59) though not typeclasses of the form `TC[F[_], A]` (yet!) I'd like to start removing the boilerplate and adding some symbolic ops to some of the less used packages.

In particular I'm thinking of the `arrow` package.

yilinwei:
So I've made a start on it [here](https://github.com/yilinwei/cats/tree/arrow).

I've got a few questions about some of where the operations have come from - in particular I'm wondering if the `choice` operation on the `Choice` is named correctly since it seems to be identical to fan in according to [hackage](https://hackage.haskell.org/package/base-4.9.0.0/docs/Control-Arrow.html).

I'm also a little confused by the hierarchy - I can understand inheriting some stuff from `Profunctor` and splitting out `Compose` and `Category`, but I'm having a hard time understanding why `Choice` doesn't inherit from `Arrow` given they are both in the `arrow` package.

As I'm learning more about arrows I'm thinking that if `free` is it's own module, then arrows should be as well.

@non I think you originally wrote the arrow implementations, so was wondering whether you could provide some insight.

ceedubs:
@yilinwei this is great!

Other than being in the same package, is there anything else that makes you think that `Choice` should inherit from `Arrow`?

The idea of `arrow` being its own package is interesting. To me `FunctionK` seems like something that should be in `core`, but I'm not so sure about the other stuff - it's not stuff that I see people use often. Anyone else have thoughts on this?

@mpilquist do you think that this effort is something that we should move forward with for cats 1.0? It seems like we are probably closing in on a 1.0 milestone within the next few months.

yilinwei:
@ceedubs 

It's not entirely clear to me where the `Choice` comes from - I had suspected it was the `ArrowChoice` in which case, the problem with not inheriting from `Arrow` is that certain operations on the `Choice` can't be derived (if choice is indeed written with the wrong signature currently, then `|||` can't be derived without the `lift` on the arrow) which means that it's of fairly limited use and we'd have an `ArrowChoice` which inherits from both anyway. 

If `Choice` is not from the same abstraction as `Arrow` then I think it ought to be moved to reflect that - mainly to help people understand where the terms come from; in a similar way that `free` and `functor` have their own separate package.

mpilquist:
[...]

PR #1247, opened 2016-07-30
Use Simulacrum for F[_, _] type classes

ceedubs:
This addresses #1064. I updated all of the type classes that I could find that I think Simulacrum 0.8 supports, but I'm not 100% I got them all. There was some discussion on that issue about moving arrow into its own package, but I think we should treat that as a separate concern.

~~I currently have this marked as `Do not merge`, because it depends on a snapshot version of Simulacrum. @mpilquist as far as I can tell this feature is working splendidly. Do you think Simulacrum is ready for an official 0.8 release?~~

@yilinwei I know you had started some work on this. I hope I'm not stepping on your toes. I wanted to try to wrap this up in preparation for cats 1.0.


ceedubs:
@mpilquist I was under the assumption that Simulacrum still doesn't support shapes like `ApplicativeError`. Is that right?

yilinwei:
@ceedubs Nope, not yet.

ceedubs:
Version 0.8.0 of Simulacrum has been released now. When I'm at the right computer I just need to update this PR to depend on it.

ceedubs:
I've updated this to Simulacrum 0.8.0. It should be ready for review/merge.

travisbrown:
I haven't really been following Simulacrum development closely enough to have anything useful to say here, but this looks reasonable and it works, so we might as well update before 0.7.0.

ceedubs:
I've resolved the merge conflict. @mpilquist does this look good to you?

core/src/main/scala/cats/Bifoldable.scala
```diff
@@ -1,9 +1,11 @@
 package cats
 
+import simulacrum.typeclass
+
 /**
  * A type class abstracting over types that give rise to two independent [[cats.Foldable]]s.
  */
-trait Bifoldable[F[_, _]] extends Any with Serializable { self =>
+@typeclass trait Bifoldable[F[_, _]] extends Any with Serializable { self =>
```
mpilquist: No need for `Any with Serializable` here -- Simulacrum adds that.

core/src/main/scala/cats/functor/Bifunctor.scala
```diff
@@ -1,11 +1,13 @@
...
 package functor
 
+import simulacrum.typeclass
+
 /**
  * A type class of types which give rise to two independent, covariant
  * functors.
  */
-trait Bifunctor[F[_, _]] extends Any with Serializable { self =>
+@typeclass trait Bifunctor[F[_, _]] extends Any with Serializable { self =>
```
mpilquist: Same as above

core/src/main/scala/cats/functor/Profunctor.scala
```diff
@@ -1,15 +1,30 @@
...
+import simulacrum.typeclass
+
 /**
  * A [[Profunctor]] is a [[Contravariant]] functor on its first type parameter
  * and a [[Functor]] on its second type parameter.
  *
  * Must obey the laws defined in cats.laws.ProfunctorLaws.
  */
-trait Profunctor[F[_, _]] extends Serializable { self =>
+@typeclass trait Profunctor[F[_, _]] extends Serializable { self =>
```
mpilquist: No need for Serializable

mpilquist:
I added a couple of minor comments but 👍 otherwise.

ceedubs:
Thanks @mpilquist. I updated the ones that you pointed out (and also `Compose`).

Issue #1242, opened 2016-07-29
Sproradic (but frequent) JS build failures


ceedubs:
Recently we started seeing the following sporadic build failures (during JS phase). Do any folks with scala.js experience know what might be going on here?

I _think_ that the first time that I noticed it was after #1231, but I'm not sure. I'm pretty sure that I _did_ see this happen at least once before the recent switch from scala.js 0.6.8 to 0.6.11.

Below is an excerpt from the logs. You can also look at [this Travis build](https://travis-ci.org/typelevel/cats/jobs/148468118) starting on line 9132.

```

java.lang.AssertionError: JS test interface sent bad reply: msg:org.scalatest.events.SuiteStarting
	at org.scalajs.testadapter.ComUtils$.badResponse$1(ComUtils.scala:54)
	at org.scalajs.testadapter.ComUtils$.receiveResponse(ComUtils.scala:84)
	at org.scalajs.testadapter.ComUtils$.receiveLoop(ComUtils.scala:21)
	at org.scalajs.testadapter.ComUtils$.receiveLoop(ComUtils.scala:16)
	at org.scalajs.testadapter.ScalaJSTask.execute(ScalaJSTask.scala:61)
	at sbt.TestRunner.runTest$1(TestFramework.scala:76)
	at sbt.TestRunner.run(TestFramework.scala:85)
	at sbt.TestFramework$$anon$2$$anonfun$$init$$1$$anonfun$apply$8.apply(TestFramework.scala:202)
	at sbt.TestFramework$$anon$2$$anonfun$$init$$1$$anonfun$apply$8.apply(TestFramework.scala:202)
	at sbt.TestFramework$.sbt$TestFramework$$withContextLoader(TestFramework.scala:185)
	at sbt.TestFramework$$anon$2$$anonfun$$init$$1.apply(TestFramework.scala:202)
	at sbt.TestFramework$$anon$2$$anonfun$$init$$1.apply(TestFramework.scala:202)
	at sbt.TestFunction.apply(TestFramework.scala:207)
	at sbt.Tests$.sbt$Tests$$processRunnable$1(Tests.scala:239)
	at sbt.Tests$$anonfun$makeSerial$1.apply(Tests.scala:245)
	at sbt.Tests$$anonfun$makeSerial$1.apply(Tests.scala:245)
	at sbt.std.Transform$$anon$3$$anonfun$apply$2.apply(System.scala:44)
[...]

PR #1249, opened 2016-07-30
Check Order laws for BigInt

ceedubs:
It appears that this test was overlooked.


kailuowang:
👍

johnynek:
👍 

2.10 build failed on the #1242 issue

PR #1250, opened 2016-07-31
[Do not merge] Experimenting with stabilizing builds

ceedubs:
I'm going to use this PR to take shots in the dark at stabilizing builds. Don't merge this branch and be aware that I'm likely to force push to it.

It has been `0` builds since the last build failure.


kailuowang:
It might be memory related. I am going to experiment with separating the tasks in travis, here is the experiment PR https://github.com/typelevel/cats/pull/1254

kailuowang:
I think #1256 fixes the build. The problem is still most likely memory.

ceedubs:
Thanks again @kailuowang.

PR #1251, opened 2016-07-31
Correct spelling in instance names

travisbrown:
This is trivial but we might as well do it sooner rather than later.


johnynek:
👍 

looks like a failure with the scalajs build. Related to #1242

adelbertc:
👍

ceedubs:
I'm going to assume that this compiles and merge it :)

PR #1257, opened 2016-08-02
[Do not merge] fix build with more memory / update to jdk8

kailuowang:
Another approach to fix build by giving jvm more memory and upgrade to jdk 8


kailuowang:
this one seems not working, closing

Issue #1253, opened 2016-08-01
Draft 0.7.0 release notes


travisbrown:
We'd like to publish an 0.7.0 release in the next few days, and we need release notes. Since this release includes a number of large breaking changes (I just updated circe and my other Cats-dependent projects to the 0.7.0 snapshot yesterday and it took a couple of hours longer than I expected), I think it'd be a good idea to include something like a migration guide in addition to the summary of pull requests.

ceedubs:
Resolved in #1259.

PR #1258, opened 2016-08-02
Update SBT and SBT plugin versions

travisbrown:
I just noticed while working on #1253 that some of these versions are out-of-date. I ran the benchmarks locally to confirm that the sbt-jmh update doesn't break anything.


kailuowang:
Looks like build is going to fail.  but #1256 should fix build

travisbrown:
@kailuowang Great, thanks—I'll rebase after that's merged. (For the record `sbt validate` passes locally if I bump the memory in `.jvmopts`.)

kailuowang:
@travisbrown thanks for the hint. I tried the same approach in this PR https://github.com/typelevel/cats/pull/1257 but it failed pretty miserably (maybe travis just doesn't have that much memory? )

mikejcurry:
@kailuowang -I understood Travis limitation in terms of memory to be 3G. From Travis FAQ, though don't know how up to date that is.

> What hardware are my tests running on?
> We use dedicated hardware to ensure full speed for your tests. You have 3GB of memory and up to 2 cores available. Your tests are running directly off SSDs with the most common databases optimized to run off RAM disks.

travisbrown:
Just rebased this to try out #1256 with the updates.

johnynek:
👍

fthomas:
:+1:

PR #1261, opened 2016-08-03
Add MonadRec for Future and reinstate Future tests for JVM

travisbrown:
This PR adds a new `MonadRec` instance for `Future` (as discussed on [Gitter](https://gitter.im/typelevel/cats?at=57a2073cc915a0e426b91de1)). The implementation of `tailRecM` is not literally tail recursive, but it is stack-safe.

I've also reinstated the `jvm` module that was removed in #1018, since it looks like the removal of the `MonadError` laws checking there wasn't intentional, and since I need a place to put the `MonadRec` laws checking for `Future`.

### DIFF
--- a/core/src/main/scala/cats/instances/future.scala
+++ b/core/src/main/scala/cats/instances/future.scala
@@ -8,11 +8,17 @@ import scala.concurrent.{ExecutionContext, Future}
 
 trait FutureInstances extends FutureInstances1 {
 
-  implicit def catsStdInstancesForFuture(implicit ec: ExecutionContext): MonadError[Future, Throwable] with CoflatMap[Future] =
-    new FutureCoflatMap with MonadError[Future, Throwable]{
+  implicit def catsStdInstancesForFuture(implicit ec: ExecutionContext): MonadError[Future, Throwable] with CoflatMap[Future] with MonadRec[Future] =
+    new FutureCoflatMap with MonadError[Future, Throwable] with MonadRec[Future] {
       def pure[A](x: A): Future[A] = Future.successful(x)
 
       def flatMap[A, B](fa: Future[A])(f: A => Future[B]): Future[B] = fa.flatMap(f)
+
+      final def tailRecM[B, C](b: B)(f: B => Future[(B Xor C)]): Future[C] =
+        f(b).flatMap {
+          case Xor.Left(b1) => tailRecM(b1)(f)
+          case Xor.Right(c) => Future.successful(c)
+        }
 
       def handleErrorWith[A](fea: Future[A])(f: Throwable => Future[A]): Future[A] = fea.recoverWith { case t => f(t) }
 
--- a/core/src/main/scala/cats/instances/future.scala
+++ b/core/src/main/scala/cats/instances/future.scala
@@ -14,6 +14,10 @@ trait FutureInstances extends FutureInstances1 {
 
       def flatMap[A, B](fa: Future[A])(f: A => Future[B]): Future[B] = fa.flatMap(f)
 
+      /**
+       * Note that while this implementation will not compile with `@tailrec`,
+       * it is in fact stack-safe.
+       */
       final def tailRecM[B, C](b: B)(f: B => Future[(B Xor C)]): Future[C] =
         f(b).flatMap {
           case Xor.Left(b1) => tailRecM(b1)(f)
--- a/jvm/src/test/scala/cats/tests/FutureTests.scala
+++ b/jvm/src/test/scala/cats/tests/FutureTests.scala
@@ -1,26 +1,17 @@
 package cats
-package js
+package jvm
 package tests
 
 import cats.data.Xor
 import cats.laws.discipline._
-import cats.js.instances.Await
-import cats.js.instances.future.futureComonad
 import cats.tests.CatsSuite
 
-import scala.concurrent.Future
+import scala.concurrent.{Await, Future}
 import scala.concurrent.duration._
+import scala.concurrent.ExecutionContext.Implicits.global
 
 import org.scalacheck.Arbitrary
 import org.scalacheck.Arbitrary.arbitrary
-
-// https://issues.scala-lang.org/browse/SI-7934
-@deprecated("", "")
-class DeprecatedForwarder {
-  implicit def runNow = scala.scalajs.concurrent.JSExecutionContext.Implicits.runNow
-}
-object DeprecatedForwarder extends DeprecatedForwarder
-import DeprecatedForwarder.runNow
 
 class FutureTests extends CatsSuite {
   val timeout = 3.seconds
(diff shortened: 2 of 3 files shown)


core/src/main/scala/cats/instances/future.scala
```diff
@@ -8,12 +8,18 @@ import scala.concurrent.{ExecutionContext, Future}
...
 
-  implicit def catsStdInstancesForFuture(implicit ec: ExecutionContext): MonadError[Future, Throwable] with CoflatMap[Future] =
-    new FutureCoflatMap with MonadError[Future, Throwable]{
+  implicit def catsStdInstancesForFuture(implicit ec: ExecutionContext): MonadError[Future, Throwable] with CoflatMap[Future] with MonadRec[Future] =
+    new FutureCoflatMap with MonadError[Future, Throwable] with MonadRec[Future] {
       def pure[A](x: A): Future[A] = Future.successful(x)
 
       def flatMap[A, B](fa: Future[A])(f: A => Future[B]): Future[B] = fa.flatMap(f)
 
+      final def tailRecM[B, C](b: B)(f: B => Future[(B Xor C)]): Future[C] =
```
kailuowang: shall we add `@tailrec` ?

ceedubs:
Oops sorry about that, it looks like I did remove some tests that I didn't mean to remove. Thanks for catching this @travisbrown.

I'm a bit confused though. Why is it that these tests need to be jvm-only? One explanation that I could imagine is that the `Eq[Future[A]]` instance uses `Await.result`. However, it looks like we are calling `Await.result` in the [JS Future tests](https://github.com/typelevel/cats/blob/master/js/src/test/scala/cats/tests/FutureTests.scala), so is that not an issue?

core/src/main/scala/cats/instances/future.scala
travisbrown: @kailuowang This implementation won't actually compile with `@tailrec`, since the recursive call is inside the `flatMap`. It is stack-safe, though, thanks to the way `Future#flatMap` works.

travisbrown:
@ceedubs The `js` version of the tests uses a [different `Await`](https://github.com/typelevel/cats/blob/master/js/src/test/scala/cats/tests/FutureTests.scala#L7), so having the tests be shared would be somewhat inconvenient (you'd still need _some_ platform-specific code for at least the `Eq` instance).

core/src/main/scala/cats/instances/future.scala
johnynek: @travisbrown is this stack safe on all versions of scala? I thought scala futures could have issues with this (at least in 2.10 or something. This is just a vague recollection).

travisbrown: @johnynek I'm pretty sure that was only an issue with the original Akka `Future` implementation, and that by the time SIP-14 landed `flatMap` was stack-safe (thanks to pressure from Twitter and inspired by `c.t.u.Future`). I'll double-check that, though.

js/src/test/scala/cats/tests/FutureTests.scala
```diff
@@ -47,4 +47,5 @@ class FutureTests extends CatsSuite {
 
   checkAll("Future[Int]", MonadErrorTests[Future, Throwable].monadError[Int, Int, Int])
   checkAll("Future[Int]", ComonadTests[Future].comonad[Int, Int, Int])
+  checkAll("Future", MonadRecTests[Future].monadRec[Int, Int, Int])
```
johnynek: is this test a reliable test of stack safety?

jvm/src/test/scala/cats/tests/FutureTests.scala
```diff
@@ -0,0 +1,39 @@
...
+import org.scalacheck.Arbitrary
+import org.scalacheck.Arbitrary.arbitrary
+
+class FutureTests extends CatsSuite {
+  val timeout = 3.seconds
+
+  def futureXor[A](f: Future[A]): Future[Xor[Throwable, A]] =
+    f.map(Xor.right[Throwable, A]).recover { case t => Xor.left(t) }
+
+  implicit def eqfa[A: Eq]: Eq[Future[A]] =
```
johnynek: should we add this but have it take a timeout to the object where we define the `Monad`?

`def futureEq[A: Eq](timeout: Duration): Eq[Future[A]] = ...`

js/src/test/scala/cats/tests/FutureTests.scala
travisbrown: @johnynek I don't believe so—it [looks like](https://github.com/typelevel/cats/blob/master/laws/src/main/scala/cats/laws/FlatMapRecLaws.scala#L14) it just checks for consistency with `flatMap`.

core/src/main/scala/cats/instances/future.scala
travisbrown: @johnynek I'm not finding the discussion I was thinking of at the moment, but I did try this on 2.9.3:
```scala
import scala.concurrent.{ Await, Future }, scala.concurrent.duration._

def test(i: Int)(f: Int => Future[Int]): Future[Int] = f(i).flatMap {
  case x if x > 100000 => Future.successful(x)
  case x => test(x)(f)
}

val f1 = test(0)(x => Future(x + 1))
val f2 = test(0)(x => Future(x))
```

Neither blows the stack, and `f1` succeeds as expected.

jvm/src/test/scala/cats/tests/FutureTests.scala
```diff
@@ -0,0 +1,39 @@
...
+
+  implicit val throwableEq: Eq[Throwable] =
+    Eq.fromUniversalEquals
+
+  // Need non-fatal Throwables for Future recoverWith/handleError
+  implicit val nonFatalArbitrary: Arbitrary[Throwable] =
+    Arbitrary(arbitrary[Exception].map(identity))
+
+  checkAll("Future with Throwable", MonadErrorTests[Future, Throwable].monadError[Int, Int, Int])
+  checkAll("Future", MonadRecTests[Future].monadRec[Int, Int, Int])
```
johnynek: isn't this repeated above?

jvm/src/test/scala/cats/tests/FutureTests.scala
travisbrown: @johnynek That's what I do for Twitter futures in catbird, but there's no chance for Scala.js there, anyway, and I think Scala.js support is the motivation for keeping all `Await` out of cats-core.

johnynek: I see. I didn't know that restriction.

jvm/src/test/scala/cats/tests/FutureTests.scala
travisbrown: @johnynek There's a bit of duplication here across `js` and `jvm`, since the `Eq` instance for Scala.js uses a custom `Await`.

core/src/main/scala/cats/instances/future.scala
johnynek: Yeah, looks legit. Thanks (I know Twitter's is stack safe, I didn't know scala was).

fthomas: Maybe a code comment mentioning that this construct is actually stack-safe would be appropriate here?

johnynek:
👍 

Love some MonadRec!

kailuowang:
👍 thanks a lot !

core/src/main/scala/cats/instances/future.scala
travisbrown: @fthomas Makes sense—done.

ceedubs:
@travisbrown thanks for the explanation regarding `Await`. That makes sense now.

👍 once the build passes. Thanks!

PR #1262, opened 2016-08-04
Configure external links in ScalaDoc

Atry:
Like what we did in https://github.com/milessabin/shapeless/commit/e1a620ddb172e6ae62bcad6395debd1f28c5d2fa


travisbrown:
Can you expand on why this is necessary? As far as I know, Cats doesn't have any external dependencies that need linking apart from the standard library, and there's no need for an extra plugin to get links to the standard library—we're already configured for that, and it seems to [work just fine](http://typelevel.org/cats/api/index.html#cats.data.OneAnd@find(f:A=>Boolean)(implicitF:cats.Foldable[F]):Option[A]).

Atry:
Sorry, I was just seeing [alleycats's ScalaDoc](https://oss.sonatype.org/service/local/repositories/releases/archive/org/typelevel/alleycats-core_2.11/0.1.5/alleycats-core_2.11-0.1.5-javadoc.jar/!/index.html), and missing in external links. Then I submitted these PRs for related projects.

I should have checked if it is necessary for each project.

PR #1265, opened 2016-08-04
Support pattern matching on NonEmptyVector

travisbrown:


kailuowang:
👍

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -223,6 +223,8 @@ object NonEmptyVector extends NonEmptyVectorInstances {
     new NonEmptyVector(buf.result)
   }
 
+  def unapply[A](nev: NonEmptyVector[A]): Some[(A, Vector[A])] = Some((nev.head, nev.tail))
```
ceedubs: Is there a foreseeable use-case where having the static return type as `Some` instead of `Option` will cause inference (or similar) issues?

travisbrown: @ceedubs There's some reason to prefer `Some` as the return type for extractors that can't fail—the compiler actually uses that information in some way, but I can't remember the details. I'll try to look up a reference…

larsrh: The compiler can check for exhaustiveness if it returns `Some`. It can't when it returns `Option`.

dwijnand: @larsrh Got a code snippet showing that?

AFAIK Some can't check exhaustiveness any more than Option as it's not sealed (sealedness isn't transitive nor does final doesn't imply sealed, double sadly).

larsrh:
```scala
scala> object MyCons1 { def unapply[A](xs: scala.collection.immutable.::[A]): Some[(A, List[A])] = Some(xs.head, xs.tail) }
defined object MyCons1

scala> object MyCons2 { def unapply[A](xs: scala.collection.immutable.::[A]): Option[(A, List[A])] = Some(xs.head, xs.tail) }
defined object MyCons2

scala> def x = List(1, 2) match { case MyCons1(_, _) => }
<console>:12: warning: match may not be exhaustive.
It would fail on the following input: Nil
       def x = List(1, 2) match { case MyCons1(_, _) => }
                   ^
x: Unit

scala> def x = List(1, 2) match { case MyCons2(_, _) => }
x: Unit
```

larsrh:
AFAICT it's unspecced behaviour, but I think it makes a lot of sense for the pattern matcher to behave that way.

core/src/main/scala/cats/data/NonEmptyVector.scala
larsrh: AFAICT it's unspecced behaviour, but I think it makes a lot of sense for the pattern matcher to behave that way.

ceedubs: Oh neat. Thanks, @larsrh!

ceedubs:
👍

johnynek:
👍

PR #1111, opened 2016-06-10
Rename Gosub to FlatMapped

adelbertc:

PR #1266, opened 2016-08-04
FreeT stack-safe transformer

### LINKED ISSUE #1158: FreeT

raulraja:
`FreeT` is a stack safe monad transfer that allows interleaving effects and other monads along with the process of building a Free AST. Fixes #1158 

@ceedubs @adelbertc @non Can one of you review? Thanks for your consideration.

Includes instances for
- [x] MonadState
- [x] MonadError
- [x] FlatMap
- [x] TransLift
- [x] Monad
- [x] SemigroupK
- [x] MonadCombine
- [x] FlatMapRec

Instances for `Foldable` and `Traverse` are not included but possible. We need a way to derive `foldRight` and `foldLeft` from `foldMap` which I will contribute after this one so those instances are stack safe as well. I believe the same applies to `Free` which is currently lacking several instances.

An example that illustrating interleaving `State` is shown below and also available in the `freemonads.tut` FreeT section included as part of this PR.

```scala
import cats.free._
import cats._
import cats.data._

/* A base ADT for the user interaction without state semantics */
sealed abstract class Teletype[A] extends Product with Serializable
final case class WriteLine(line : String) extends Teletype[Unit]
final case class ReadLine(prompt : String) extends Teletype[String]

type TeletypeT[M[_], A] = FreeT[Teletype, M, A]
type Log = List[String]

/** Smart constructors, notice we are abstracting over any MonadState instance
 *  to potentially support other types beside State 
 */
class TeletypeOps[M[_]](implicit MS : MonadState[M, Log]) {
  def writeLine(line : String) : TeletypeT[M, Unit] =
	FreeT.liftF[Teletype, M, Unit](WriteLine(line))
  def readLine(prompt : String) : TeletypeT[M, String] =
	FreeT.liftF[Teletype, M, String](ReadLine(prompt))
  def log(s : String) : TeletypeT[M, Unit] =
	FreeT.liftT[Teletype, M, Unit](MS.modify(s :: _))
}

object TeletypeOps {
  implicit def teleTypeOpsInstance[M[_]](implicit MS : MonadState[M, Log]) : TeletypeOps[M] = new TeletypeOps
}

type TeletypeState[A] = State[List[String], A]

def program(implicit TO : TeletypeOps[TeletypeState]) : TeletypeT[TeletypeState, Unit] = {
  for {
	userSaid <- TO.readLine("what's up?!")
	_ <- TO.log(s"user said : $userSaid")
	_ <- TO.writeLine("thanks, see you soon!")
  } yield () 
}

def interpreter = new (Teletype ~> TeletypeState) {
  def apply[A](fa : Teletype[A]) : TeletypeState[A] = {  
	fa match {
	  case ReadLine(prompt) =>
		println(prompt)
		val userInput = "hanging in here" //scala.io.StdIn.readLine()
		StateT.pure[Eval, List[String], A](userInput)
	  case WriteLine(line) =>
		StateT.pure[Eval, List[String], A](println(line))
	}
  }
}

import TeletypeOps._

val state = program.foldMap(interpreter)
val initialState = Nil
val (stored, _) = state.run(initialState).value
```
Largely based on scalaz's FreeT and ideas from http://www.haskellforall.com/2012/07/free-monad-transformers.html


free/src/main/scala/cats/free/FreeT.scala
```diff
@@ -0,0 +1,254 @@
...
+        Gosub(e.a.interpret(st), e.f.andThen(_.interpret(st)))
+      case Suspend(m) =>
+        Suspend(M.map(m)(_.map(s => st(s))))
+    }
+
+  /**
+    * Runs to completion, mapping the suspension with the given transformation
+    * at each step and accumulating into the monad `M`.
+    */
+  def foldMap(f: S ~> M)(implicit M0: FlatMapRec[M], M1: Applicative[M]): M[A] = {
```
travisbrown: Why not just a `MonadRec` constraint?

raulraja: Makes sense. Will change that there and in a couple of methods more.

free/src/main/scala/cats/free/FreeT.scala
```diff
@@ -0,0 +1,254 @@
...
+    * at each step and accumulating into the monad `M`.
+    */
+  def foldMap(f: S ~> M)(implicit MR: MonadRec[M]): M[A] = {
+    def go(ft: FreeT[S, M, A]): M[FreeT[S, M, A] Xor A] =
+      ft match {
+        case Suspend(ma) => MR.flatMap(ma) {
+          case Xor.Left(a) => MR.pure(Xor.Right(a))
+          case Xor.Right(sa) => MR.map(f(sa))(Xor.right)
+        }
+        case g @ Gosub(_, _) => g.a match {
```
johnynek: why are you not doing `case GoSub(a, f) => ` here and preferring `g @ GoSub(_, _)`?

raulraja: The abstract type `A` in `GoSub` in which both `a` and `f` depend seem to cause inference to not work properly unless the whole thing is named :P. 
This technique is also used in scalaz. Open to changes if anyone thinks there is a better way. 
https://github.com/scalaz/scalaz/pull/730

free/src/main/scala/cats/free/FreeT.scala
```diff
@@ -0,0 +1,254 @@
...
+    }
+}
+
+
+object FreeT extends FreeTInstances {
+  /** Suspend the computation with the given suspension. */
+  private[free] case class Suspend[S[_], M[_], A](a: M[A Xor S[A]]) extends FreeT[S, M, A]
+
+  /** Call a subroutine and continue with the given function. */
+  private[free] case class Gosub[S[_], M[_], A0, B](a0: FreeT[S, M, A0], f0: A0 => FreeT[S, M, B]) extends FreeT[S, M, B] {
```
fthomas: `Gosub` in `Free` was renamed to `FlatMapped` in https://github.com/typelevel/cats/pull/1111. Does it make sense to call it `FlatMapped` here too?

adelbertc: +1 for `FlatMapped`

raulraja: Done!

travisbrown:
This looks good to me now, but I'm not really up-to-date on the details of the implementation here or how it lines up with the current `Free`, so I think we should wait for two other maintainer sign-offs.

ceedubs:
Would it make sense to do some more reification as constructors, similarly to what has been done with baking coyoneda into `Free`? For example, currently the `pure` method requires an `Applicative[M]` instance and ends up using the `Suspend` constructor. Should there be a `Pure` constructor and potentially `SuspendS`, `SuspendM`, etc? That seems like it would reduce some of the constraints of the initialization methods. It could also possibly lead to minor performance improvements, but probably not enough so to be a good reason to do it.

WDYT?

raulraja:
@ceedubs I would agree with that but given you need a `MonadRec[M]` to do anything useful with `FreeT` and that is a harder constrain than `Applicative` not sure what the real benefit is since in most if not all cases you would already have those instances in scope. Fewer ADTs keeps the design easier to reason about IMHO. Having said that I'll be happy to change it and add those if it is what we usually follow in cats for other types. Let me know your thoughts and I'll be happy to add those. Thanks!

johnynek:
in light of #1275 and the fact that this is based on scalaz's implementation, if there are significant portions of copy-paste here (I didn't check) can we git-blame the original file and add a comment to that effect? Also, it might be nice to link to a particular git-sha from which this was based.

raulraja:
@johnynek out of a computer until tomorrow. I hope the tweet was not originated from this PR I clearly stated that it was based on the Scalaz impl in the PR description and at no time I wanted to discredit anyone for their work. Is the proper way to attribute them in the file header? I didn't feel that mentioning people there would be appropriate without their explicit permission. I'll run the git blame in the AM when I'm close to a computer. Where do those name belong since they are neither cats authors nor contributors? I thought that most of cats was based on the Scalaz design and mentioning the library in the PR was enough. Apologies if I skipped any protocol I was not aware of.

johnynek:
I don't think you need to apologize. I am not 100% sure what has motivated all of it, and I guess history had some role to play, far beyond any one 1 PR.

That said, I think no one can object to adding a link to the code it was based on an mentioning the authors in a comment. Others can voice their opinion.

raulraja:
@johnynek Given the current status of affairs, what should I do to get this PR to comply with how you guys want to proceed with attributions?. It seems Brian has already been added to Authors and not sure if anything extra should be sone as part of this PR. Looking for directions to push this forward.

free/src/main/scala/cats/free/FreeT.scala
```diff
@@ -0,0 +1,258 @@
...
+          case Xor.Right(s) => liftF[S, M, g1.A](s).flatMap(g1.f)
+        }
+        case g0 @ FlatMapped(_, _) => g0.a.flatMap(g0.f(_).flatMap(g1.f)).toM
+      }
+    }
+
+  @tailrec
+  private def step: FreeT[S, M, A] =
+    this match {
+      case g @ FlatMapped(_, _) => g.a match {
```
johnynek: what's the rationale for `g @ FlatMapped(_, _) => g.a` rather than `FlatMapped(a, _) =>`

same below.

Also, why not:
```scala
this match {
  case FlatMapped(FlatMapped(a, fn0), fn1) =>
    a.flatMap(a => fn0(a).flatMap(fn1)).step
  case nonNestedFlatMap => nonNestedFlatMap
}
```

raulraja:
@johnynek @puffnfresh Added attribution based on the ones @ceedubs added in #1276

johnynek:
just a minor comment on the style of pattern matching (which I'm not personally crazy about), but otherwise this looks good to me.

Thanks for porting this. I could have personally used it earlier in the week.

free/src/main/scala/cats/free/FreeT.scala
raulraja: These are for the same reasons as https://github.com/typelevel/cats/pull/1266#discussion_r73614872 Inference does not work when referring to an abstract type member.

johnynek:
👍

non:
👍

johnynek:
I think you need to merge master and make a couple of changes (MonadRec no longer exists). See `Free.scala` for the change.

free/src/main/scala/cats/free/FreeT.scala
```diff
@@ -0,0 +1,258 @@
...
+import scala.annotation.tailrec
+
+import cats.data.Xor
+
+/** FreeT is a monad transformer for Free monads over a Functor S
+  *
+  * This Scala implementation of `FreeT` and its usages are derived from
+  * [[https://github.com/scalaz/scalaz/blob/series/7.3.x/core/src/main/scala/scalaz/FreeT.scala Scalaz's FreeT]],
+  * originally written by Brian McKenna.
+  */
```
ceedubs: It might also be good to link to [Stack Safety for Free](http://functorial.com/stack-safety-for-free/index.pdf) here. Both as attribution to Phil Freeman and because this is the paper that inspired `FreeT` arriving in Scala.

ceedubs:
Thanks, @raulraja! This is looking great.

I think that this might suffer from the [toString issue that was fixed on Free](https://github.com/typelevel/cats/pull/1084). If so, shall we take the same approach we did there?

free/src/main/scala/cats/free/FreeT.scala
```diff
@@ -0,0 +1,263 @@
...
+        }
+      }
+
+    MR.tailRecM(this)(go)
+  }
+
+  /**
+    * Runs to completion, using a function that maps the resumption from `S` to a monad `M`.
+    */
+  def runM(interp: S[FreeT[S, M, A]] => M[FreeT[S, M, A]])(implicit S: Functor[S], MR: MonadRec[M]): M[A] = {
```
johnynek: this is no longer in master. You need to merge master or the merge is going to fail CI.

free/src/test/scala/cats/free/FreeTTests.scala
```diff
@@ -0,0 +1,218 @@
...
+  type FreeTListState[A] = FreeT[IntState, IntState, A]
+  type MonadRec[F[_]] = Monad[F] with RecursiveTailRecM[F]
+
+  case class JustFunctor[A](a: A)
+
+  implicit val listSemigroupK: SemigroupK[ListWrapper] = ListWrapper.semigroupK
+  
+  implicit val listWrapperMonad: MonadRec[ListWrapper] with Alternative[ListWrapper] = ListWrapper.monadCombine
+
+  implicit val intStateMonad: MonadRec[IntState] = throw("can't summon this instance!")
```
raulraja: @adelbertc @johnynek This is what I was mentioning in gitter. I can't find a way to summon the `StateT` `Monad with RecursiveTailRecM` instance

raulraja: Nevermind, I think i may have found the issue

free/src/main/scala/cats/free/FreeT.scala
```diff
@@ -0,0 +1,263 @@
...
+    }
+}
+
+private[free] sealed trait FreeTInstances0 extends FreeTInstances1 {
+  implicit def catsFreeMonadForFreeT[S[_], M[_]](implicit M0: Applicative[M]): Monad[FreeT[S, M, ?]] with RecursiveTailRecM[FreeT[S, M, ?]] =
+    new FreeTMonad[S, M] {
+      def M = M0
+    }
+
+  implicit def catsFreeCombineForFreeT[S[_], M[_]: Applicative: RecursiveTailRecM: SemigroupK]: SemigroupK[FreeT[S, M, ?]] =
```
johnynek: I don't think you need `RecursiveTailRecM` unless you also have `FlatMap` or `Monad` and I don't see that here.

free/src/main/scala/cats/free/FreeT.scala
```diff
@@ -0,0 +1,263 @@
...
+    new FreeTCombine[S, M] {
+      override def M = implicitly
+      override def M1 = implicitly
+      override def M2 = implicitly
+    }
+}
+
+private[free] sealed trait FreeTInstances extends FreeTInstances0 {
+  implicit def catsFreeMonadCombineForFreeT[S[_], M[_]: Alternative: RecursiveTailRecM]: MonadCombine[FreeT[S, M, ?]] =
```
johnynek: I don't think we should take `RecursiveTailRecM` here

free/src/main/scala/cats/free/FreeT.scala
```diff
@@ -0,0 +1,263 @@
...
+  /**
+   * Runs to completion, using a function that maps the resumption from `S` to a monad `M`.
+   */
+  def runM(interp: S[FreeT[S, M, A]] => M[FreeT[S, M, A]])(implicit S: Functor[S], MR: Monad[M], RT: RecursiveTailRecM[M]): M[A] = {
+    def runM2(ft: FreeT[S, M, A]): M[FreeT[S, M, A] Xor A] =
+      MR.flatMap(ft.resume) {
+        case Xor.Left(a) => MR.pure(Xor.right(a))
+        case Xor.Right(fc) => MR.map(interp(fc))(Xor.left)
+      }
+    MR.tailRecM(this)(runM2)
```
johnynek: you should do: `RT.sameType(MR).tailRecM(this)(runM2)` so we have the compiler check we have the right type on `RecursiveTailRecM`

free/src/main/scala/cats/free/FreeT.scala
```diff
@@ -0,0 +1,263 @@
...
+        case g1 @ FlatMapped(_, _) => g1.a match {
+          case Suspend(m1) => MR.map(m1) {
+            case Xor.Left(a) => Xor.left(g1.f(a))
+            case Xor.Right(fc) => Xor.right(Xor.right(S.map(fc)(g1.f(_))))
+          }
+          case g2 @ FlatMapped(_, _) => MR.pure(Xor.left(g2.a.flatMap(g2.f(_).flatMap(g1.f))))
+        }
+      }
+
+    MR.tailRecM(this)(go)
```
johnynek: same comment as below `RT.sameType(MR).tailRecM(this)(go)`

free/src/main/scala/cats/free/FreeT.scala
```diff
@@ -0,0 +1,263 @@
...
+        case g @ FlatMapped(_, _) => g.a match {
+          case Suspend(mx) => MR.flatMap(mx) {
+            case Xor.Left(x) => MR.pure(Xor.left(g.f(x)))
+            case Xor.Right(sx) => MR.map(f(sx))(g.f andThen Xor.left)
+          }
+          case g0 @ FlatMapped(_, _) => MR.pure(Xor.left(g0.a.flatMap(g0.f(_).flatMap(g.f))))
+        }
+      }
+
+    MR.tailRecM(this)(go)
```
johnynek: `RT.sameType(MR).tailRecM(this)(go)`

free/src/main/scala/cats/free/FreeT.scala
```diff
@@ -0,0 +1,263 @@
...
+      override def M = implicitly
+      override def get =
+        FreeT.liftT(M1.get)
+      override def set(s: E) =
+        FreeT.liftT(M1.set(s))
+    }
+}
+
+private[free] sealed trait FreeTInstances2 extends FreeTInstances3 {
+  implicit def catsFreeMonadErrorForFreeT[S[_], M[_]: RecursiveTailRecM, E](implicit E: MonadError[M, E]): MonadError[FreeT[S, M, ?], E] =
```
johnynek: why do we need `RecursiveTailRecM` here?

free/src/main/scala/cats/free/FreeT.scala
```diff
@@ -0,0 +1,263 @@
...
+private[free] sealed trait FreeTMonad[S[_], M[_]] extends Monad[FreeT[S, M, ?]] with RecursiveTailRecM[FreeT[S, M, ?]] with FreeTFlatMap[S, M] {
+  implicit def M: Applicative[M]
+
+  override final def pure[A](a: A): FreeT[S, M, A] =
+    FreeT.pure[S, M, A](a)
+}
+
+private[free] sealed trait FreeTCombine[S[_], M[_]] extends SemigroupK[FreeT[S, M, ?]] {
+  implicit def M: Applicative[M]
+  implicit def M1: RecursiveTailRecM[M]
```
johnynek: why is this here?

johnynek:
👍 looks good to me now. Thanks!

free/src/main/scala/cats/free/FreeT.scala
raulraja: Constrains the nested M to be also stack safe but if you think is not necessary we can remove this constrain.

raulraja:
@johnynek @adelbertc @travisbrown @non @ceedubs I think this is ready. Let me know if there is anything extra you'd like me to add or change. Thanks!

johnynek:
👍

adelbertc:
LGTM 👍 Given the size the of the change and the number of changes since @non 's previous 👍 we should let him (or another maintainer) OK it before merging.

adelbertc:
Hey @raulraja I just merged https://github.com/typelevel/cats/pull/1289 - doesn't cause any merge conflicts but moving forward Cats will be using `Either` so I made a PR to the branch which merges the latest stuff + changes usage to `Either`: https://github.com/47deg/cats/pull/1

johnynek:
:+1:

non:
Looks good to me. 👍

PR #1269, opened 2016-08-05
add catchNonFatal to MonadError

johnynek:
This method seems really convenient for many common `MonadError` cases.

It is somewhat limited due to requiring `Throwable`, we could put it on the `MonadError` object rather than instance, and we could introduce a new typeclass `Catchable[M]`?

I think this covers 90% of the cases without adding too much complexity.

### DIFF
--- a/core/src/main/scala/cats/ApplicativeError.scala
+++ b/core/src/main/scala/cats/ApplicativeError.scala
@@ -80,7 +80,7 @@ trait ApplicativeError[F[_], E] extends Applicative[F] {
    * Often E is Throwable. Here we try to call pure or catch
    * and raise.
    */
-  def catchNonFatal[A](a: => A)(implicit ev: Throwable =:= E): F[A] =
+  def catchNonFatal[A](a: => A)(implicit ev: Throwable <:< E): F[A] =
     try pure(a)
     catch {
       case NonFatal(e) => raiseError(e)
@@ -90,7 +90,7 @@ trait ApplicativeError[F[_], E] extends Applicative[F] {
    * Often E is Throwable. Here we try to call pure or catch
    * and raise
    */
-  def catchNonFatalEval[A](a: Eval[A])(implicit ev: Throwable =:= E): F[A] =
+  def catchNonFatalEval[A](a: Eval[A])(implicit ev: Throwable <:< E): F[A] =
     try pure(a.value)
     catch {
       case NonFatal(e) => raiseError(e)
@@ -99,7 +99,7 @@ trait ApplicativeError[F[_], E] extends Applicative[F] {
   /**
    * If the error type is Throwable, we can convert from a scala.util.Try
    */
-  def fromTry[A](t: Try[A])(implicit ev: Throwable =:= E): F[A] =
+  def fromTry[A](t: Try[A])(implicit ev: Throwable <:< E): F[A] =
     t match {
       case Success(a) => pure(a)
       case Failure(e) => raiseError(e)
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/ApplicativeError.scala
```diff
@@ -74,6 +75,25 @@ trait ApplicativeError[F[_], E] extends Applicative[F] {
   def recoverWith[A](fa: F[A])(pf: PartialFunction[E, F[A]]): F[A] =
     handleErrorWith(fa)(e =>
       pf applyOrElse(e, raiseError))
+  /**
+   * Often E is Throwable. Here we try to call pure or catch
+   * and raise.
+   */
+  def tryCatch[A](a: => A)(implicit ev: Throwable =:= E): F[A] =
```
mpilquist: What do you think about naming this `catchNonFatal` to match the method on `Xor`?

johnynek: good call. Updated.

tests/src/test/scala/cats/tests/TryTests.scala
```diff
@@ -30,6 +30,31 @@ class TryTests extends CatsSuite {
     }
   }
 
+  test("catchNonFatal works") {
+    forAll { e: Either[String, Int] =>
+      val str = e.fold(identity, _.toString)
+      val res = MonadError[Try, Throwable].catchNonFatal(str.toInt)
+      // the above shuold just never cause an uncaught exception
```
mikejcurry: s/shuold/should

mpilquist:
👍

core/src/main/scala/cats/ApplicativeError.scala
```diff
@@ -74,6 +76,34 @@ trait ApplicativeError[F[_], E] extends Applicative[F] {
   def recoverWith[A](fa: F[A])(pf: PartialFunction[E, F[A]]): F[A] =
     handleErrorWith(fa)(e =>
       pf applyOrElse(e, raiseError))
+  /**
+   * Often E is Throwable. Here we try to call pure or catch
+   * and raise.
+   */
+  def catchNonFatal[A](a: => A)(implicit ev: Throwable =:= E): F[A] =
```
travisbrown: My first inclination would be to make this `eq: Throwable <:< E`—that will work in all reasonable places where `E` is statically known, and it eliminates the need for guesswork about whether to write `E =:= Throwable` (which won't work) or `Throwable =:-=` (which will) for people who want to support these methods in generic contexts.

travisbrown:
:+1: but see my question above.

johnynek:
this has two 👍 is it okay to merge one's own PR?

core/src/main/scala/cats/ApplicativeError.scala
```diff
@@ -74,6 +76,34 @@ trait ApplicativeError[F[_], E] extends Applicative[F] {
...
+    try pure(a)
+    catch {
+      case NonFatal(e) => raiseError(e)
+    }
+
+  /**
+   * Often E is Throwable. Here we try to call pure or catch
+   * and raise
+   */
+  def catchNonFatalEval[A](a: Eval[A])(implicit ev: Throwable <:< E): F[A] =
```
adelbertc: Could this just delegate to `catchNonFatal` with `a.value` ?

johnynek: I think that would be slightly slower since you would allocate another closure to do the call by name.

adelbertc:
I've done it before, go ahead and merge :)

PR #1270, opened 2016-08-06
Unapply.apply: return implicit parameter

fthomas:
... instead of calling `implicitly`. This removes an unnecessary
function call which also just returns the `ev` parameter.


adelbertc:
:+1:

ceedubs:
👍

PR #1273, opened 2016-08-07
Add syntax import objects for FunctorFilter and TraverseFilter

ceedubs:
This is something that I missed on #1225.

@travisbrown this might be something that we want to make it into 0.7.0.


travisbrown:
:+1:

adelbertc:
:+1:

nima1363azad:
989212672037

Issue #1275, opened 2016-08-09
Attribution or original author clause as part of contributor agreement


jdegoes:
See [here](https://twitter.com/puffnfresh/status/762790754066440192) for context.

Quick thoughts:

1. I'd suggest the CLA (which the project does not appear to have yet) mandate original authorship or proper attribution. Once such a policy is instantiated, every contributor could be contacted to certify original authorship or provide proper attribution, thus catching contributions up to the policy change.
2. I'd also suggest amending the COC to explicitly forbid copy/paste without compatible license and proper attribution — mainly so people know that intentional or uncorrected failure to comply is grounds for ejection from the community.

travisbrown:
👎 to a CLA, 👍 to more explicitly forbidding copy-paste without proper attribution.

johnynek:
+1 to this. It is fine to share code, but copyright and license annotations should be preserved.

I don't know the exact solution here because I am not familiar with what fraction of the code is copy/paste or trivial modifications of existing code.

One solution Brian mentioned was: https://twitter.com/puffnfresh/status/762811494719291393 but I don't think that's such a great solution as:
1. many on that list may not want to be added to contributors.
2. many may not in fact be contributors.

It seems Brian wants to be added or acknowledged in some way, so the first thing we could do is try to make a quick pass for the obvious candidates of places where the ball was dropped in the past.

jdegoes:
For some, "contributors" may imply endorsement and / or intentional activity, so "Attribution" or some other section distinct from contributors and prefaced appropriately may be a more broadly appreciated way of providing attribution.

johnynek:
So, this file is here (as pointed out in gitter):
https://github.com/typelevel/cats/blob/master/COPYING

which contains the scalaz notice.

That may not be enough. I still think if you contribute something you should state where it came from if not original work (whatever that means in this area that is so highly derivative of other work generally).

jdegoes:
> That may not be enough. 

Indeed, the wording makes it seem like this was a one time thing that stopped in 2014 (based on the copyright). Not that it is being "continuously derived" from Scalaz, including recent 2016 code, and including many contributors not mentioned in the copyright notice, such as Brian McKenna.

> I still think if you contribute something you should state where it came from if not original work 

👍 In many projects, I often see these notices in the code itself, e.g. `This function was based on so and so's work ...` with proper attribution. That works.

[...]

PR #1276, opened 2016-08-09
Some attribution work

ceedubs:
This is inspired by #1275. It doesn't include a CLA, and there is
most likely more work that should be done with respect to attribution,
but I'm hoping that this is a step in the right direction.

I've added a link to the scalaz contributors page within AUTHORS.md.
I'm not sure who all out of that should and would want to be directly listed
in the authors page for Cats. I've explicitly added Brian McKenna to the
authors page, as my understanding of [this
tweet](https://twitter.com/puffnfresh/status/762901748772057088) is that
he would like to be on there.

@puffnfresh, if you care to, please let me know if there is any way that I could improve this.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -99,8 +99,9 @@ possible:
  * yilinwei
  * Zach Abbott
 
-Cats has been heavily inspired by [Scalaz](https://github.com/scalaz/scalaz),
-and some Cats code is only a slightly modified version of code originating in
+Cats has been heavily inspired by many libraries, including [Scalaz](https://github.com/scalaz/scalaz),
+Haskell's [Prelude](https://hackage.haskell.org/package/base-4.9.0.0/docs/Prelude.html), and others.
+In particular, some Cats code is only a slightly modified version of code originating in
 Scalaz. Therefore, we'd also like to credit and thank all of the
 [Scalaz contributors](https://github.com/scalaz/scalaz/graphs/contributors) for
 their work.
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -31,6 +31,7 @@ possible:
  * Binh Nguyen
  * Bobby Rauchenberg
  * Brendan McAdams
+ * Brian McKenna
  * Cody Allen
  * Colt Frederickson
  * Dale Wijnand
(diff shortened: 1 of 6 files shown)


mpilquist:
👍 Thanks

AUTHORS.md
```diff
@@ -98,6 +99,12 @@ possible:
  * yilinwei
  * Zach Abbott
 
+Cats has been heavily inspired by [scalaz](https://github.com/scalaz/scalaz),
+and some Cats code is only a slightly modified version of code originating in
+scalaz. Therefore, we'd also like to credit and thank all of the
```
travisbrown: This is nitpicky, but maybe capitalize "scalaz" throughout for consistency with "Cats" and other project names here, and with the Scalaz README?

milessabin:
I'm pretty sure that `NotNothing` was derived from a a gist of mine way back when.

AUTHORS.md
```diff
@@ -98,6 +99,12 @@ possible:
  * yilinwei
  * Zach Abbott
 
+Cats has been heavily inspired by [scalaz](https://github.com/scalaz/scalaz),
```
travisbrown: Suggestion:

> Cats has been inspired by many libraries, including Scalaz, Haskell's Prelude, and others. In particular, some Cats code is only…

To acknowledge stuff like witherable, etc.?

AUTHORS.md
ceedubs: 👍 thanks good catch. I tend to capitalize Cats to make it clear that I'm talking about the project and not the group of animals, but I definitely didn't mean to be inconsistent here.

milessabin:
Yes, [here](https://gist.github.com/milessabin/de58f3ba7024d51dcc1a) ... itself derived from an earlier uses of ambiguity to encode negation.

AUTHORS.md
ceedubs: 👍 sorry I missed this when you first posted it.

travisbrown:
:+1:

kailuowang:
👍

aloiscochard:
Thanks for the effort @ceedubs and @adelbertc, appreciated!

Just giving my opinion here as Adelbert did ask for it, just keep in mind that with only 4 commits I do not consider myself as a scalaz7 contributor at all.

That being said, in the process of writing a functional library that is fast, I steel ideas from @non whenever I can, and I would expect anybody to do the same with all the great ideas appearing in the diversity of open source projects around.

I like what you are doing here in term of attribution, and in essence it's how I did it (https://github.com/scalaz/scalaz/blob/series/8.0.x/meta/src/main/scala/Ops.scala#L6), once we'll move forward formalizing more the approach, I'll definitely push into following a similar process as the one put in place here.

:+1:

ceedubs:
@aloiscochard thank you, I really appreciate the feedback.

I'm going to go ahead and merge this, and if @puffnfresh has any further feedback, I can address it in a separate PR.

PR #1286, opened 2016-08-12
Add val alias for FunctionK

andyscott:
I find myself wanting to use the identity instance of FunctionK via `~>.id`. This can be achieved with an import `import cats.arrow.{ FunctionK ⇒ ~> }`, but an alias to the object would be a bit more convenient.

### DIFF
--- a/core/src/main/scala/cats/package.scala
+++ b/core/src/main/scala/cats/package.scala
@@ -7,6 +7,7 @@ import cats.data.Xor
 package object cats {
 
   type ~>[F[_], G[_]] = arrow.FunctionK[F, G]
+  val ~> = arrow.FunctionK
 
   type ⊥ = Nothing
   type ⊤ = Any


ceedubs:
👍 thanks, @andyscott!

To other reviewers: we have [a precedent](https://github.com/typelevel/cats/blob/e0f08498e4bb118626fe42fb50727cc659db430e/core/src/main/scala/cats/package.scala#L54-L66) of doing this for types coming from cats kernel.

travisbrown:
I'm a weak :-1: on this. I think aliases have to do a lot of work to justify themselves, and while I can see the arguments for `~>` as a type and the cats-kernel aliases in `cats`, I'm not as convinced here.

johnynek:
Similar to Travis here. I think operators are the main barrier to learning a library due to practical difficulties around discovery in many cases. I'd go so far as just preferring to never use ~> in favor of FunctionK personally (but I work with a number of people who are a bit skeptical of this stuff to begin with)

andyscott:
Here's my two cents:

I have mixed feelings on this as well.

I suspect it is slightly confusing, as an end user, to see the `~>` alias for `FunctionK` being used throughout cats yet be unable to access the companion with the same alias when writing your own code. If the alias exists, it seems simpler if it works for both cases. No aliases are probably even more straightforward.

non:
I think the `~>` alias is important and don't want to remove it.

I agree that `~>.id` seems a bit weird, but I don't see a ton of harm in aliasing the companion and letting end users use it if they want.

PR #1292, opened 2016-08-16
Eliminate macro build noise.

milessabin:
SBT 0.13.12 added a warning about recompilation due to incremental build cache invalidation in the presence of macros. This is enabled by default.

Whilst in many ways laudable, this introduces a lot of unhelpful build noise in projects which define macros. Fortunately there is an SBT incremental build option to disable this warning. The PR adds that option to the Cats build.

More info on the warning can be found [here](https://github.com/sbt/sbt/issues/2654).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -51,6 +51,7 @@ lazy val kernelSettings = Seq(
 ) ++ warnUnusedImport
 
 lazy val commonSettings = Seq(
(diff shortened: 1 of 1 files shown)


johnynek:
👍

adelbertc:
👍

PR #1298, opened 2016-08-19
fix botBuild value on travis

kailuowang:
As pointed out here https://github.com/typelevel/cats/pull/1025#issuecomment-218288666, the `botBuild` value on travis isn't set correctly, which might be causing the scalaJS build issues we saw on travis previously. This change should fix the setting. @ceedubs you have time to look at this one?


johnynek:
seems good to me, but I really don't know travis well. I am assuming they always set the `TRAVIS` envvar?

kailuowang:
@johnynek that's right, according to this [doc](https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables)

johnynek:
👍

non:
👍

non:
Let's give it a shot!

PR #1299, opened 2016-08-19
Add ContravariantCartesian typeclass and for Tuple2 add Monad and FlatMap

johnynek:
These new instances should be isomorphic to `WriterT[Id, A, B]` instances, but there may be cases where you already have a tuple2 and want to use a monad.

This approach can be generalized to `TupleN`, but I am not sure it yet justifies code-gen or macros for this purpose.

Interestingly, one could make a `Monad` for a general case-class that is generic in at least the last parameter, but that may be going too far.

I also added `Cartesian[Eq]` to make the tests easier (otherwise I needed to plumb some new implicits everywhere).

### DIFF
--- a/core/src/main/scala/cats/instances/tuple.scala
+++ b/core/src/main/scala/cats/instances/tuple.scala
@@ -1,9 +1,12 @@
 package cats
 package instances
 
+import data.Xor
+import scala.annotation.tailrec
+
 trait TupleInstances extends Tuple2Instances with cats.kernel.instances.TupleInstances
 
-sealed trait Tuple2Instances {
+sealed trait Tuple2Instances extends Tuple2Instances1 {
   implicit val catsStdBitraverseForTuple2: Bitraverse[Tuple2] =
     new Bitraverse[Tuple2] {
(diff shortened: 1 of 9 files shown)


striation:
LGTM 👍

core/src/main/scala/cats/instances/tuple.scala
```diff
@@ -1,9 +1,12 @@
 package cats
 package instances
 
+import data.Xor
```
adelbertc: Either! :D

adelbertc:
https://github.com/typelevel/cats/pull/1289 is merged now - looks like this change doesn't have a merge conflict but I suspect if merged it will break the build :)

johnynek:
I'll update, get the coverage positive and try again.

johnynek:
I think this is green now, except for the PR failing because of the `Xor => Either`. Since that is being reverted, this should be green after #1305 goes through.

adelbertc:
:+1:

kailuowang:
👍

Issue #1290, opened 2016-08-15
Map monoid performance


travisbrown:
Cats's monoid instance for maps is extremely slow for the common use case where you're `foldMap`-ing over e.g. a collection of documents. For example, here's a simple benchmark where we add up 1000 singleton maps with 500 distinct keys:

```scala
import cats.instances.list._, cats.instances.int._, cats.instances.map._
import org.openjdk.jmh.annotations.{ Benchmark, Scope, State }
import scalaz.std.anyVal._, scalaz.std.list._, scalaz.std.map._

@State(Scope.Benchmark)
class MapMonoidBench {
  val words: List[String] = for {
    c <- List("a", "b", "c", "d", "e")
    t <- 1 to 100
  } yield c * t

  val maps: List[Map[String, Int]] = (words ++ words).map(s => Map(s -> 1))

  @Benchmark def combineAllCats: Map[String, Int] =
    cats.Monoid[Map[String, Int]].combineAll(maps)

  @Benchmark def combineCats: Map[String, Int] = maps.foldLeft(Map.empty[String, Int]) {
    case (acc, m) => cats.Monoid[Map[String, Int]].combine(acc, m)
  }

  @Benchmark def combineScalaz: Map[String, Int] = maps.foldLeft(Map.empty[String, Int]) {
    case (acc, m) => scalaz.Monoid[Map[String, Int]].append(acc, m)
  }

  @Benchmark def combineNeither: Map[String, Int] = maps.foldLeft(Map.empty[String, Int]) {
    case (acc, m) => m.foldLeft(acc) {
      case (m, (k, v)) => m.updated(k, v + m.getOrElse(k, 0))
    }
  }

  @Benchmark def foldMapCats: Map[String, Int] = cats.Foldable[List].foldMap(maps)(identity)
  @Benchmark def foldMapScalaz: Map[String, Int] = scalaz.Foldable[List].foldMap(maps)(identity)
}
```

And Scalaz has ~100 times the throughput:

```
[info] Benchmark                       Mode  Cnt     Score     Error  Units
[info] MapMonoidBench.combineAllCats  thrpt   20    68.481 ±   1.521  ops/s
[info] MapMonoidBench.combineCats     thrpt   20    54.264 ±   3.205  ops/s
[info] MapMonoidBench.combineNeither  thrpt   20  8742.891 ± 125.203  ops/s
[info] MapMonoidBench.combineScalaz   thrpt   20  6624.394 ±  68.323  ops/s
[...]

PR #1300, opened 2016-08-19
Drastically speed up Monoid[Map[K, V]] instance.

non:
This commit fixes a number of problems I introduced with cats-kernel.

 1. Map's monoid was not overriding combineAll
 2. It was also doing the wrong thing for combine
 3. Monoid should have overridden combineAllOption

The performance for map's `combine` is 150x faster after this commit,
and `combineAll` is 780x faster. Rather than focusing on how good
these speed ups are, I want to apologize for how bad the
implementations were previously. The new `combine` is competitive with
other implementations, and `combineAll` is about x5 faster than the
equivalent code using combine.

(All figures using before/after numbers for the given benchmark.)

Thanks to Adelbert and Travis for noticing this problem. Also thanks
to Travis for producing this benchmark! Finally, thanks to Oscar for
helping me look at and fix this issue.

### DIFF
--- a/kernel/src/main/scala/cats/kernel/instances/StaticMethods.scala
+++ b/kernel/src/main/scala/cats/kernel/instances/StaticMethods.scala
@@ -5,34 +5,12 @@ import scala.collection.mutable
 
 object StaticMethods {
 
-  def initMutableMap[K, V](m: Map[K, V]): mutable.Map[K, V] = {
-    val result = mutable.Map.empty[K, V]
-    m.foreach { case (k, v) => result(k) = v }
-    result
-  }
-
-  def wrapMutableMap[K, V](m: mutable.Map[K, V]): Map[K, V] =
-    new WrappedMutableMap(m)
-
   private[kernel] class WrappedMutableMap[K, V](m: mutable.Map[K, V]) extends Map[K, V] {
     override def size: Int = m.size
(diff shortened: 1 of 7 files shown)


non:
This fixes #1290.

johnynek:
👍

travisbrown:
:+1:, thanks!

kernel/src/main/scala/cats/kernel/instances/StaticMethods.scala
```diff
@@ -5,15 +5,6 @@ import scala.collection.mutable
 
 object StaticMethods {
 
-  def initMutableMap[K, V](m: Map[K, V]): mutable.Map[K, V] = {
-    val result = mutable.Map.empty[K, V]
-    m.foreach { case (k, v) => result(k) = v }
-    result
-  }
-
-  def wrapMutableMap[K, V](m: mutable.Map[K, V]): Map[K, V] =
```
johnynek: if you remove this, it becomes unusable from the outside. That was the reason for it, I think, to allow using it from say `algebra`.

non: Ah, right! I'll add them back so we can use them in Algebra (and beyond).

non:
Two sign-offs, I'm merging.

PR #1304, opened 2016-08-20
Remove references to Monad from CoflatMap doc

kcsongor:
Apologies in advance for this, if I had the ability to label this PR, it would definitely be something along the lines of 'nitpick'.

The doc for `coflatMap` states that the value is in a Monadic context, which can be confusing, because:
- there is no Monad constraint on the context (the example being a Monad is incidental) 
- the context is *more likely* to be a Comonad


non:
Makes sense to me. 👍 

We'll merge this after 0.7.0 is released. Thanks!

kcsongor:
Thanks for the quick response!

PR #1305, opened 2016-08-20
Roll-back #1301 and #1289 to allow 0.7.0 release.

non:
Due to a communications error we merged some PRs which we had been
planning to keep out of master until after 0.7.0. Since one of them
(#1289) was large and long-running, it was difficult to undo this
merge in an automated way.

This commit represents the following:

 1. Downloaded patches for PRs 1301 and 1289 from Github
 2. Tried to reverse-apply them (-R)
 3. Fixed rejections/failures manually
 4. Got code compiling and tests passing

The code is now different than it was before the PRs were merged,
but things *seem* to be in a reasonable state. I'm keeping all these
changes in one commit so that it will be easier to "reverse" this
once 0.7.0 is relaesed.


non:
FYI, I'm going to be rebasing as I fix things to keep this to a single commit.

adelbertc:
:+1:

adelbertc:
Looks like there's `tut` issues

johnynek:
I'm actually a little confused here. 0.6.1 didn't have FlatMapRec. So changing it to take Either rather than Xor is not going to break anyone.

The other change is to make Either more usable, and that also sounds nice.

Why don't we want the Either based tailRecM now rather than introduce it just to change the signature in the next release?

non:
@johnynek Due to the way things were merged it ended up being easier to do it this way. I don't have a principled argument here but I also don't really want to try to change that part.

adelbertc:
@non I think @johnynek is pointing out that the only changes that need to be reverted are those which affect existing APIs in 0.6.1. For instance, `FlatMapRec` did not exist so introducing it with `Xor` in 0.7.x only to replace with `Either` in 0.8.x would be strange. Similarly adding syntax for `Either` and adding `EitherT` does not affect any previous functionality.

johnynek:
So what do we n d to rollback if not the Either stuff?

I'm a little nervous about the API churn and the deep stack of libraries I have been using that use Cats. If breakage is painful, it is going to lock me on 0.6.1 for a while.

johnynek:
Yes, @adelbertc, that is my point exactly.

non:
Alright, as-per discussions on Gitter it seems like we've decided to hold off doing a full-scale revert and instead just revert the existing API changes (to minimize 0.6.1 -> 0.7.0 breakage). This will make the future 0.7.x -> 0.8.0 upgrade path less painful.

@adelbertc agreed to take point on this. In the meantime, please don't merge this PR unless things change.

non:
Fixed in #1306 closing this PR.

PR #1306, opened 2016-08-20
Partial revert of #1289

adelbertc:
All APIs that are present in 0.6.x continue to use `Xor`. New additions, notably FlatMapRec stuff, use `Either`.

### DIFF
--- a/core/src/main/scala/cats/Foldable.scala
+++ b/core/src/main/scala/cats/Foldable.scala
@@ -207,7 +207,7 @@ import simulacrum.typeclass
   /**
    * Behaves like traverse_, but uses [[Unapply]] to find the
    * [[Applicative]] instance for `G` - used when `G` is a
-   * type constructor with two or more parameters such as [[scala.util.Either]]
+   * type constructor with two or more parameters such as `scala.util.Either`
    *
    * {{{
    * scala> import cats.implicits._
@@ -251,7 +251,7 @@ import simulacrum.typeclass
   /**
    * Behaves like sequence_, but uses [[Unapply]] to find the
    * [[Applicative]] instance for `G` - used when `G` is a
-   * type constructor with two or more parameters such as [[scala.util.Either]]
+   * type constructor with two or more parameters such as `scala.util.Either`
    *
    * {{{
    * scala> import cats.implicits._
--- a/free/src/main/scala/cats/free/Free.scala
+++ b/free/src/main/scala/cats/free/Free.scala
@@ -4,6 +4,7 @@ package free
 import scala.annotation.tailrec
 
 import cats.arrow.FunctionK
+import cats.data.Xor
 
 /**
  * A free operational monad for some functor `S`. Binding is done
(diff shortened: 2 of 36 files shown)


core/src/main/scala/cats/Foldable.scala
```diff
@@ -251,7 +251,7 @@ import simulacrum.typeclass
   /**
    * Behaves like sequence_, but uses [[Unapply]] to find the
    * [[Applicative]] instance for `G` - used when `G` is a
-   * type constructor with two or more parameters such as [[scala.util.Either]]
+   * type constructor with two or more parameters such as `scala.util.Either`
```
adelbertc: For some reason `validate` was giving me "Could not find any members to link for "scala.util.Either" so I went for the easy way out.

free/src/main/scala/cats/free/Free.scala
```diff
@@ -80,8 +81,8 @@ sealed abstract class Free[S[_], A] extends Product with Serializable {
    */
   final def runM[M[_]](f: S[Free[S, A]] => M[Free[S, A]])(implicit S: Functor[S], M: Monad[M]): M[A] = {
     def runM2(t: Free[S, A]): M[A] = t.resume match {
-      case Left(s) => Monad[M].flatMap(f(s))(runM2)
-      case Right(r) => Monad[M].pure(r)
+      case Xor.Left(s) => Monad[M].flatMap(f(s))(runM2)
```
johnynek: Isn't this a recursive flatMap? Why isn't this using tailRecM?

core/src/main/scala/cats/Foldable.scala
non: I totally endorse this. I may-or-may-not have made a snarky tweet about Scaladoc today related to this.

free/src/main/scala/cats/free/Free.scala
adelbertc: Yeah it is - I just went around changing stuff from `Either` to `Xor`, didn't really look at implementations at all.

I think I know how to change this to use `tailRecM` but since you and Eric are currently reviewing and I'd like to keep this all in one commit, I'll let you guys finish reviewing, fix it, have you give it a glance, and push.

non: Given that this was there already I say let's open an issue and not hold this PR up for this.

non: (We can still fix it today but I want to get *master* out of the DANGER DANGER state as soon as possible.)

adelbertc: Sounds good

johnynek: I'll try to have a branch hot and ready to fix this, but yes, let's merge this first.

johnynek:
👍 
I did a grep of `Either` on this branch. It seems correct to me. All the remaining Either in core appear to be expects (tailRecM, EitherT, either syntax).

Thanks @adelbertc !

non:
👍

PR #1312, opened 2016-08-21
Remove superfluous Comonad constraint from Free's run

kcsongor:
At the moment, Free's run method requires a `Comonad` constraint, which I think captures the wrong abstraction.
What is actually needed, is a `forall X. S[X] => X` function (which Comonad's `extract` incidentally is), but this is much more general than a Comonad, and therefore the current constraint is too restrictive.

A possible way of existentially quantifying the `X` is by requiring a natural transformation (or `FunctionK`) `S ~> Id`, which of course expands to the required signature. What this means is that now there is an explicit argument that we require for `run`, instead of the implicit Comonad instance.

I'm not sure if this is the best way of solving this problem, so please let me know if there is a nicer way - but I definitely think that requiring `Comonad` is unnecessary in this context.

### DIFF
--- a/bench/src/main/scala/cats/bench/TrampolineBench.scala
+++ b/bench/src/main/scala/cats/bench/TrampolineBench.scala
@@ -20,9 +20,13 @@ class TrampolineBench {
       y <- Eval.defer(evalFib(n - 2))
     } yield x + y
 
+  private def fun0natId =
+    new (Function0 ~> Id) {
+      def apply[A](fa: Function0[A]): A = fa.extract
+    }
 
   @Benchmark
-  def trampoline(): Int = trampolineFib(N).run
+  def trampoline(): Int = trampolineFib(N).run(fun0natId)
 
   def trampolineFib(n: Int): Trampoline[Int] =
     if (n < 2) Trampoline.done(n) else for {
--- a/free/src/main/scala/cats/free/Free.scala
+++ b/free/src/main/scala/cats/free/Free.scala
@@ -69,11 +69,13 @@ sealed abstract class Free[S[_], A] extends Product with Serializable {
   }
 
   /**
-   * Run to completion, using the given comonad to extract the
-   * resumption.
+   * Run to completion, using the given natural transformation to extract the
+   * resumption. We use a natural transformation from S to Id, because it
+   * is essentially a function `S[X] => X forall X`, and we want the `X`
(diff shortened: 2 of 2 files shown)


ceedubs:
Thanks @kcsongor. I agree that the `Comonad` constraint is overly restrictive, but I thought of it as a helper method for when you have a `Comonad` handy. Is there a reason that you couldn't use `foldMap` for your use-case?

kcsongor:
@ceedubs Right, I see what you mean, and there isn't a reason why I couldn't use foldMap, in fact, the `go` function would work as well. This being said, maybe `go` could require a natural transformation `S[X] => X`, instead of the specialised `S[Free[S, A]] => Free[S, A]`?

It just felt really out of place that there was a Comonad constraint there, and I wasn't aware that it was meant to be a special case.
However, I can't really think of it as a general theme that you want to plug a Comonad into Free, and it feels incidental that `Function0` (used in `Trampoline`) is a Comonad, and I'm not sure that is a very useful abstraction.

What do you think - can you think of other examples where you have a comonad as an underlying functor of a free monad?
Also, what are your thoughts on making `go`'s type signature more parametric by swapping the function with the natural transformation?

kcsongor:
As an aside, I think it would make sense to define a type synonym for `F ~> Id` and call it `Algebra` (which of course corresponds to the F-algebra `F[X] => X`).

My reasoning is that a free monad is obtained by a free algebra for some functor, and we can 'interpret' this free monad by applying some algebra over the free monad - this is what the `Algebra` type synonym corresponds to.
If I understand *co*rrectly, *co*monad's *co*unit doesn't *co*rrespond to this *co*nstruction. (<- lot of 'co's in that sentence, it wasn't intentional!) It's a different arrow in a different setting, with the same type - although I might be wrong here, couldn't find much on this topic, so this is just my intuition.

ceedubs:
@kcsongor I'm sorry this one slipped through the cracks.

I think that `run`, `go`, and `foldMap` are all useful methods that have their own use-cases. I'm sorry but I don't understand the motivation for specializing `go` to take an `S ~> Id` when this would just be a less general form of `foldMap`, which you could already pass your `S ~> Id` into.

As far as `Algebra`, I think that [matryoshka](https://github.com/slamdata/matryoshka) has taken over as the library that you want to use if you are interested in these sort of algebras.

kailuowang:

PR #1313, opened 2016-08-21
Fix date in 0.7.0 release notes

travisbrown:
As @non pointed out on Gitter, I forgot to update the date here this morning.


kailuowang:
:+1:

PR #1319, opened 2016-08-22
Topic/kernel big decimal

### LINKED ISSUE #1295: BigDecimal Order & CommutativeGroup instances Missing
Currently BigDecimal Order and Group typeclasses are missing from the kernel.

non:
Add order and commutative group instances for `BigDecimal`.

This fixes #1295.


kailuowang:
👍  LGTM, we had some problem with BigDecimal precision handling in scala 2.10, but I think in this context, having an `Arbitrary[Double]` parse into an `Arbitrary[BigDecimal]` should be fine (if the build goes green 😉  ).

johnynek:
can you update this to `0.7.0` while we are at it?

https://github.com/typelevel/cats/blob/master/build.sbt#L210

johnynek:
👍

non:
I'm going to merge with two sign-offs. Thanks everyone!

PR #1322, opened 2016-08-23
fix doctest plugin / use same command for release and travis build

kailuowang:
1. disabled doctest plugin on all js projects which was enabled by this [commit])https://github.com/typelevel/cats/commit/67eacc7c3aa2f6ebebcb67eb6ee122103fdce4b1)
2. use the same commands for travis and release
3. added freeJS test to `validateJS`
4. try use aggregated project `test` as much as possible (will see if travis can handle)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -396,7 +396,7 @@ addCommandAlias("validateKernelJS", "kernelLawsJS/test")
 
 addCommandAlias("validateFreeJS", "freeJS/test") //separated due to memory constraint on travis
 
-addCommandAlias("validate", ";clean;validateJS;validateFreeJS;validateJVM")
+addCommandAlias("validate", ";clean;validateJS;validateKernelJS;validateFreeJS;validateJVM")
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 // Base Build Settings - Should not need to edit below this line.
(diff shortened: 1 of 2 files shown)


tkawachi:
Hi @kailuowang, sbt-doctest requires `JvmPlugin` so that it's not enabled automatically in ScalaJS project. Is it enabled without disablePlugin() in this case?

https://github.com/tkawachi/sbt-doctest/blob/master/src/main/scala/com/github/tkawachi/doctest/DoctestPlugin.scala#L34-L35

kailuowang:
@tkawach yes, it's enabled without the disablePlugin, I am not sure if requiring `JvmPlugin` disable it for js projects. See scalajs plugin requires `JvmPlugin` as well.
 https://github.com/scala-js/scalajs/blob/master/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSPlugin.scala#L29

non:
I'd love to see this fixed, so let's give this a shot. 👍 

Thanks for working on this @kailuowang!

fthomas:
In refined I added `doctestGenTests := Seq.empty` to the [common settings for JS subprojects](https://github.com/fthomas/refined/blob/9f0067f3f369a2f4aaed05bce7ac9c7d3a382f5f/build.sbt#L167). This might save us from adding all those `disablePlugins(DoctestPlugin)`.

kailuowang:
@fthomas that worked, thanks!

fthomas:
:+1: 

I think you can now also remove `import com.github.tkawachi.doctest.DoctestPlugin`, right?

kailuowang:
@fthomas good catch. Fixed.

Issue #1321, opened 2016-08-22
Either combine only use `Semigroup[B]`


johnynek:
note `combine(a, b) = a` is a semigroup. We are basically forcing this "left-semigroup" on the left side.

I don't think this is a great idea at all. I would rather have:

``` scala
def combine(that: Either[A, B])(implicit sa: Semigroup[A], sb: Semigroup[B]): Either[A, B] =
  (this, that) match {
    case (Right(lb), Right(rb)) => Right(sb.combine(lb, rb))
    case (Left(la), Left(ra)) => Left(sa.combine(la, ra))
    case (Right(_), l) => l
    case (r, Right(_)) => r
  }
```

johnynek:
#1034 and #888 are related.

johnynek:
so, it originally worked this way, I would say with `Either` we should revert this decision.

If you want a `left` semigroup, you can always make one:
```scala
implicit val leftSg: Semigroup[A] = new Semigroup[A] {
  def combine(a: A, b: A): A = a
}
```
But if you want to use `Either` for something other than error propagation, this is pretty terrible.

I have used either for approximation Monoids where there is also an exact algorithm. For small sizes, I use the `Right` side, but eventually I switch to `Left` to approximate (say using a Hyperloglog on the Left and a Set on the right). That case can be handled with a custom wrapper type on Either (and indeed must be or have an orphan since you need some logic to convert `Right` to `Left`).

But I still think if you want to only keep the left, you should set up a semigroup to do so.

non:
If we wanted to go this way we could provide something like:

```scala
object ThrowableSemigroup extends Semigroup[Throwable] {
  def combine(x: Throwable, y: Throwable): Throwable = x
}
```

That way this change would be source compatible in almost all cases.

yilinwei:
I'm not sure about this. I feel that `Either` doesn't accumulate errors - if I wanted to accumulate errors I would use a `Validated`.

It also feels strange because then it becomes inconsistent with `ap`- `combine = ap(that) { _ |+| _ }` currently, which I think ought to be the default behaviour of combine.

ceedubs:
@johnynek at this point it seems like it would be pretty disruptive to make this change. Are you okay with closing this issue out? I'm not sure about your specific use-case, but I think that the support for `parTraverse` has made moving between accumulating and non-accumulating sum types easier for most of my use-cases.

ceedubs:
[...]

PR #1324, opened 2016-08-23
Fix many cats-kernel instances.

### LINKED ISSUE #1320: Move relevant function instances into kernel
Right now we have a `Monoid` for `Function1` values in *cats-core*. I think it makes sense to move the definition to *cats-kernel* since that's where monoids are defined -- this will allow *algebra* and other kernel-dependents to share these instances.

(More generally, I think where possible all orderings, monoids, etc. should live in *cats-kernel*.)

non:
This commit does a bunch of things:

 1. Move relevant either/function instnaces into kernel
 2. Add missing Eq/PartialOrder instances for Either
 3. Add many missing instances for Function0/Function1
 4. Add instances for BitSet
 5. Remove code duplication between collection monoids
 6. Improve test coverage of function instances
 7. Add some missing package objects in kernel
 8. A few stylistic changes

This is a relatively large commit but I think almost all of this is
stuff we'd like to get done before we freeze cats-kernel."

This fixes #1320.

### DIFF
--- a/core/src/main/scala/cats/instances/either.scala
+++ b/core/src/main/scala/cats/instances/either.scala
@@ -60,18 +60,25 @@ trait EitherInstances extends cats.kernel.instances.EitherInstances {
         }
 
       def traverse[F[_], B, C](fa: Either[A, B])(f: B => F[C])(implicit F: Applicative[F]): F[Either[A, C]] =
-        fa.fold(
-          a => F.pure(Left(a)),
-          b => F.map(f(b))(Right(_))
-        )
+        fa match {
+          case left @ Left(_) => F.pure(left)
+          case Right(b) => F.map(f(b))(Right(_))
+        }
 
       def foldLeft[B, C](fa: Either[A, B], c: C)(f: (C, B) => C): C =
-        fa.fold(_ => c, f(c, _))
+        fa match {
+          case Left(_) => c
+          case Right(b) => f(c, b)
+        }
 
       def foldRight[B, C](fa: Either[A, B], lc: Eval[C])(f: (B, Eval[C]) => Eval[C]): Eval[C] =
-        fa.fold(_ => lc, b => f(b, lc))
+        fa match {
+          case Left(_) => lc
+          case Right(b) => f(b, lc)
+        }
 
-      override def attempt[B](fab: Either[A, B]): Either[A, Either[A, B]] = Right(fab)
+      override def attempt[B](fab: Either[A, B]): Either[A, Either[A, B]] =
+        Right(fab)
       override def recover[B](fab: Either[A, B])(pf: PartialFunction[A, B]): Either[A, B] =
         fab recover pf
       override def recoverWith[B](fab: Either[A, B])(pf: PartialFunction[A, Either[A, B]]): Either[A, B] =
@@ -91,9 +98,10 @@ trait EitherInstances extends cats.kernel.instances.EitherInstances {
 
   implicit def catsStdShowForEither[A, B](implicit A: Show[A], B: Show[B]): Show[Either[A, B]] =
     new Show[Either[A, B]] {
-      def show(f: Either[A, B]): String = f.fold(
-        a => s"Left(${A.show(a)})",
-        b => s"Right(${B.show(b)})"
-      )
+      def show(x: Either[A, B]): String =
+        x match {
+          case Left(a) => "Left(" + A.show(a) + ")"
+          case Right(b) => "Right(" + B.show(b) + ")"
+        }
     }
 }
--- a/core/src/main/scala/cats/instances/either.scala
+++ b/core/src/main/scala/cats/instances/either.scala
@@ -48,9 +48,13 @@ trait EitherInstances extends cats.kernel.instances.EitherInstances {
       @tailrec
       def tailRecM[B, C](b: B)(f: B => Either[A, Either[B, C]]): Either[A, C] =
         f(b) match {
-          case left @ Left(_)  => left.rightCast[C]
-          case Right(Left(b1)) => tailRecM(b1)(f)
-          case Right(right) => right.leftCast[A]
+          case left @ Left(_) =>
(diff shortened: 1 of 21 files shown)


non:
(Another change I failed to mention was that I made several of the `Show` instances more efficient.)

non:
Review by @johnynek, plus anyone else who is interested.

non:
We may want to start using things `Helpers._` in our other tests -- it makes it a lot easier to be sure you're exercising all the different possible type class constructors.

kailuowang:
👍  looks great to me. I wish we had that `Helper` before. Thanks!

core/src/main/scala/cats/instances/either.scala
```diff
@@ -81,51 +81,19 @@ trait EitherInstances extends EitherInstances1 {
...
 
-private[instances] sealed trait EitherInstances1 extends EitherInstances2 {
-  implicit def catsStdPartialOrderForEither[A, B](implicit A: PartialOrder[A], B: PartialOrder[B]): PartialOrder[Either[A, B]] =
-    new PartialOrder[Either[A, B]] {
-      def partialCompare(x: Either[A, B], y: Either[A, B]): Double = x.fold(
-        a => y.fold(A.partialCompare(a, _), _ => -1),
-        b => y.fold(_ => 1, B.partialCompare(b, _))
+  implicit def catsStdShowForEither[A, B](implicit A: Show[A], B: Show[B]): Show[Either[A, B]] =
+    new Show[Either[A, B]] {
+      def show(f: Either[A, B]): String = f.fold(
```
johnynek: we have biased to `match` for alleged performance reasons. What do you think of it here?

core/src/main/scala/cats/instances/list.scala
```diff
@@ -84,6 +84,7 @@ trait ListInstances extends cats.kernel.instances.ListInstances {
 
   implicit def catsStdShowForList[A:Show]: Show[List[A]] =
     new Show[List[A]] {
-      def show(fa: List[A]): String = fa.map(_.show).mkString("List(", ", ", ")")
+      def show(fa: List[A]): String =
+        fa.iterator.map(_.show).mkString("List(", ", ", ")")
```
johnynek: nice catch

core/src/main/scala/cats/instances/map.scala
```diff
@@ -6,11 +6,11 @@ import scala.annotation.tailrec
...
   implicit def catsStdShowForMap[A, B](implicit showA: Show[A], showB: Show[B]): Show[Map[A, B]] =
-    Show.show[Map[A, B]] { m =>
-      val body = m.map { case (a, b) =>
-        s"${showA.show(a)} -> ${showB.show(b)})"
-      }.mkString(",")
-      s"Map($body)"
+    new Show[Map[A, B]] {
+      def show(m: Map[A, B]): String =
+        m.iterator
+          .map { case (a, b) => showA.show(a) + " -> " + showB.show(b) }
```
johnynek: why not `s"..."` here? Is + just as good? (compiled with a StringBuilder?)

kernel/src/main/scala/cats/kernel/instances/either.scala
```diff
@@ -0,0 +1,71 @@
...
+            case Right(_) => -1
+          }
+          case Right(xx) => y match {
+            case Left(_) => 1
+            case Right(yy) => B.compare(xx, yy)
+          }
+        }
+    }
+
+  implicit def catsDataMonoidForEither[A, B](implicit B: Monoid[B]): Monoid[Either[A, B]] =
```
johnynek: still feel #1321 is an issue.

non: I thought of that but I didn't want to make this PR big *and* contentious.

core/src/main/scala/cats/instances/map.scala
non: I thought the compiler would rewrite it to us a string builder, but maybe that's only Java? I can rewrite this using interpolation (it just ends up looking uglier IMO).

kernel/src/main/scala/cats/kernel/instances/either.scala
johnynek: no semigroup here? I think we can have with `Semigroup[B]`.

core/src/main/scala/cats/instances/either.scala
```diff
@@ -61,7 +61,7 @@ trait EitherInstances extends cats.kernel.instances.EitherInstances {
 
       def traverse[F[_], B, C](fa: Either[A, B])(f: B => F[C])(implicit F: Applicative[F]): F[Either[A, C]] =
         fa match {
-          case left @ Left(_) => F.pure(left)
+          case Left(a) => F.pure(Left(a))
```
johnynek: maybe: https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/syntax/either.scala#L301

```scala
case left @ Left(_) => F.pure(left.rightCast[C])
```
save an allocation?

core/src/main/scala/cats/instances/either.scala
```diff
@@ -48,7 +48,7 @@ trait EitherInstances extends cats.kernel.instances.EitherInstances {
       @tailrec
       def tailRecM[B, C](b: B)(f: B => Either[A, Either[B, C]]): Either[A, C] =
         f(b) match {
-          case Left(a)         => Left(a)
+          case left @ Left(_)  => left.rightCast[C]
           case Right(Left(b1)) => tailRecM(b1)(f)
           case Right(Right(c)) => Right(c)
```
johnynek: `case Right(r) => r.leftCast[A]` should work, right?

johnynek:
one last allocation to save, then I'm +1

non:
@johnynek I think this is it right?

johnynek:
👍 Thank you for this work!

Issue #1329, opened 2016-08-24
TailRecM does not make sense for lists / streams


alexandru:
The new [tailRecM](https://github.com/typelevel/cats/blob/v0.7.0/core/src/main/scala/cats/FlatMap.scala#L108) in `FlatMap` probably does not make sense for lists / streams.

I promise to [read the paper](http://functorial.com/stack-safety-for-free/index.pdf) (have some difficulty parsing Haskell/Purescript code), however this operation only makes sense for monads for which this can be implemented in a tail-recursive way (i.e. constant stack AND heap).

Looking for example at the [List implementation](https://github.com/typelevel/cats/blob/v0.7.0/core/src/main/scala/cats/instances/list.scala#L29) the call stack is effectively replaced with a heap buffer that balloons in size with each iteration. This may be stack safe, but it is not tail recursive, which I think misses the point. Unless I'm not reading it right, the growth is exponential. And actually, the possibility of such a space leak is why we have stack-overflow errors (the idea being to crash a single thread instead of the whole process or system) and I cannot see the utility of such a dangerous operation on lists / streams.

And what this operation does is enforced by the [tailRecMConsistentFlatMap](https://github.com/typelevel/cats/blob/v0.7.0/laws/src/main/scala/cats/laws/FlatMapLaws.scala#L36) law, because of which we cannot provide an alternative interpretation. Because I tried doing basically a `.take(1)` on all streams emitted by `f`.

My current problem is that in Monix I have to provide a `Monad` instance for `Observable`. And naturally I noticed the law-tests going from 2 seconds to 60 seconds, because for each observable emitted by `f`, we get an arbitrarily fresh observable that's then flat-mapped in that loop and observables are made of arbitrary lists of numbers and so on. And to give an idea of why this happens, I fired up a profiler and saw that the `defaultTailRecM` implementation was called over 4.5 million times, with over 35 million elements being streamed by the `FlatMapLaws` tests, which along with the tricks I'm doing to test this synchronously and cross-platform, it gets to be freakishly slow.

But besides this ... given its current laws, given that it is unsafe, I do NOT want to provide this function for any lists or streams I can think of. And I can't just provide the default implementation because that default is untestable. I cannot have a test that takes 60 seconds and reconfiguring ScalaCheck to emit smaller lists of numbers or to do less checks isn't OK at all, because I actually cared about the flatMap/monad tests.

alexandru:
[...]

PR #1331, opened 2016-08-25
Reduce the cost of the tailRecM test

johnynek:
This is to address: #1329 where the test cost is too high for some libraries that depend on this.

This weakening should not really weaken the test too much if we know that `map` and `flatMap` are already lawful, which we are testing independently.


kailuowang:
:+1:

non:
LGTM 👍

PR #1333, opened 2016-08-25
Bringing back Xor and XorT - fixes #1332

### LINKED ISSUE #1332: Bring back Xor for 0.8.0 and remove in 0.9.0
Gitter chat: https://gitter.im/typelevel/cats?at=57bf20c05bdd197c1cbe93a0

Actual removal: https://github.com/typelevel/cats/pull/1310

Proposal (if I understand) is to bring back just `Xor` and `XorT` (not including their uses in other parts of the codebase like it was in `ApplicativeError`) for 0.8.0, mark it with `@deprecated`, and actually remove in 0.9.0.

Pinging people who have expressed an opinion about this: @travisbrown @non @johnynek @milessabin @rossabaker

adelbertc:
*_DO NOT MERGE UNTIL WE GET CONSENUS ON #1332 *_

### DIFF
--- a/core/src/main/scala/cats/data/Xor.scala
+++ b/core/src/main/scala/cats/data/Xor.scala
@@ -1,0 +1,379 @@
+package cats
+package data
+
+import scala.annotation.tailrec
+import scala.reflect.ClassTag
+import scala.util.{Failure, Success, Try}
+
+/** Represents a right-biased disjunction that is either an `A` or a `B`.
+ *
+ * An instance of `A [[Xor]] B` is either a `[[Xor.Left Left]][A]` or a `[[Xor.Right Right]][B]`.
+ *
+ * A common use of [[Xor]] is to explicitly represent the possibility of failure in a result as opposed to
+ * throwing an exception.  By convention, [[Xor.Left Left]] is used for errors and [[Xor.Right Right]] is reserved for successes.
+ * For example, a function that attempts to parse an integer from a string may have a return type of
+ * `NumberFormatException [[Xor]] Int`. However, since there is no need to actually throw an exception, the type (`A`)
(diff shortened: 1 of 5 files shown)


core/src/main/scala/cats/data/Xor.scala
```diff
@@ -0,0 +1,374 @@
...
+ *
+ * `A [[Xor]] B` is isomorphic to `scala.Either[A, B]`, but [[Xor]] is right-biased, so methods such as `map` and
+ * `flatMap` apply only in the context of the "right" case. This right bias makes [[Xor]] more convenient to use
+ * than `scala.Either` in a monadic context. Methods such as `swap`, and `leftMap` provide functionality
+ * that `scala.Either` exposes through left projections.
+ *
+ * Some additional [[Xor]] methods can be found in [[Xor.XorOps XorOps]]. These methods are not defined on [[Xor]] itself because
+ * [[Xor]] is covariant in its types `A` and `B`.
+ */
+sealed abstract class Xor[+A, +B] extends Product with Serializable {
```
adelbertc: I would have liked to add a `@deprecated` annotation here but it causes methods associated with type class instances (down below) to error since they are technically outside of the type.. :/

mpilquist: You may be able to mark those deprecated as well. IIRC, something that's deprecated can reference other things that are deprecated without warnings. Though maybe I'm making that up. :)

adelbertc: Ah yes - looks like I have to do this for every single `*InstancesN` , but it works.

adelbertc:
Oh geez the doctest fails on the `@deprecated` stuff (https://travis-ci.org/typelevel/cats/jobs/155125594) what now?

PR #1338, opened 2016-08-28
WIP: Iterable <=> Foldable interop

johnynek:
This is just a sketch (no tests yet, incomplete) to give interop between standard scala `Iterable[T]` and `Foldable[F]`.

Is see this as having two benefits:
1. education. Seeing that Foldable and Iterable are exactly equivalent in terms of computational power can help teach the value of this type class relative to an inheritance model.
2. this allows easier interop with any iterable class that may exist, or code that accepts iterable. In this case, we can sometimes avoid a copy to create an `Iterable` as needed.

### DIFF
--- a/core/src/main/scala/cats/Foldable.scala
+++ b/core/src/main/scala/cats/Foldable.scala
@@ -2,7 +2,9 @@ package cats
 
 import scala.collection.mutable
 import cats.instances.long._
+import cats.instances.int._
 import simulacrum.typeclass
+import scala.collection.immutable.Iterable
 
 /**
  * Data structures that can be folded to a summary value.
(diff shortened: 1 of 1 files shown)


non:
:+1:

I like the idea. In the past I know there was reluctance about providing implicit instances for arbitrary `Iterable[x]` sub classes due to concerns about mutability. However, as you saw, this interop is useful enough that we already had `iterateRight` defined.

I would be immediately in favor of adding this compatibility without adding an implicit `C[x]=> Foldable[C]` instance. If we wanted that instance to be included like everything else I think we should take some more time to think through the implications (for example we'd need to be careful about priority).

This PR also makes me think that we might want to add something like the dual of foldable (`Unfoldable`?) to take about building up `F` values.

kailuowang:

PR #1342, opened 2016-09-01
Remove MonadFilter, and fix MonadCombineLaws

### LINKED ISSUE #1336: Consider an `EmptyK[F[_]]` type class, or remove `MonadFilter`
This is the type constructor version of #307 (Consider an `Empty[+A]` type class ).

Currently there are two typeclasses that promises in `def emtpy` for `F[_]`: `MonadFilter` and `MonoidK`. (There's also `ApplicativeMonoid`)

Would it make sense to unify them as `EmptyK` and mix them into whatever parent traits that needs empty? For example, we can remove `MonadFilter`.

eed3si9n:
Fixes #1336

The `MonadCombine` typeclass unifies `Monad`, `FunctorFilter`, and
`MonoidK`, bringing three of the lineages together. This uses the
emptiness notion introduced in `MonoidK` and `flatMap` to implement
`mapFilter`, along with `unite`.

Given that `MonadCombine` exists (our equivalent of `MonadPlus`) there's no need for `MonadFilter` to exist.

The laws are migrated over from `MonadFilter` to `MonadCombine`, except these laws would now check  identity laws and the consistencies against the correct `F.empty` from `MonoidK`.


mpilquist:
> Given that MonadCombine exists (our equivalent of MonadPlus) there's no need for MonadFilter to exist.

Can you expand on this? Doesn't this mean we can't define `MonadFilter` instances for monads that don't have lawful / sensible `combine` implementations?

eed3si9n:
> Doesn't this mean we can't define MonadFilter instances for monads that don't have lawful / sensible combine implementations?

Yes. If I want to define `filter` there's now `FunctorFilter`, so what's the use of introducing a notion of emptiness that's different from `combine`?

edmundnoble:
`MonadFilter` was removed in cats-mtl as @eed3si9n says because there are no laws, as well as `MonadCombine` because the laws are all derivable from parametricity and `Alternative`'s laws.

PR #1345, opened 2016-09-01
Bump ScalaCheck to 1.13.1, and fix the fallout.

non:
### Status

 - [x] Get PR building correctly
 - [x] Decide what to do about `MonadCombine` laws
 - [x] Fix cryptic errors in the *js* subproject

### Summary

This commit gets all our law tests working with a new, more robust
ScalaCheck version. It currently relies upon an unreleased Discipline
snapshot, so it definitely won't build in Travis.

There were a bunch of test bugs that got fixed, mostly around
foldRight/reduceRight consistency for non-empty data structures.

One law violation was uncovered: MonadCombine[Option] doesn't pass the
left-distributivity law we wrote. We should figure out whether this
represents a real law violation and remove the instance, or whether
the law is too strict somehow.

One counter-example is:

    def f(x: Int): Option[Int] = if (x == 0) None else Some(x)
    val a = Option(0)
    val b = Option(1)
    (a <+> b).flatMap(f) != (a.flatMap(f) <+> b.flatMap(f))

Review by @adelbertc et al. We will need to wait for a new Discipline
release before merging this, and we'll need to do something about the
MonadCombine laws/instances (currently the MonadCombine[Option] test
is commented out).

### DIFF
--- a/kernel-laws/src/test/scala/cats/kernel/laws/LawTests.scala
+++ b/kernel-laws/src/test/scala/cats/kernel/laws/LawTests.scala
@@ -8,7 +8,7 @@ import cats.kernel.instances.all._
 
 import org.typelevel.discipline.{ Laws }
 import org.typelevel.discipline.scalatest.Discipline
-import org.scalacheck.{ Arbitrary, Gen }
+import org.scalacheck.{ Arbitrary, Cogen, Gen }
 import Arbitrary.arbitrary
 import org.scalatest.FunSuite
 
@@ -24,13 +24,24 @@ class LawTests extends FunSuite with Discipline {
   implicit override val generatorDrivenConfig: PropertyCheckConfiguration =
     PropertyCheckConfig(maxSize = PropMaxSize, minSuccessful = PropMinSuccessful)
 
-  implicit def orderLaws[A: Eq: Arbitrary] = OrderLaws[A]
-  implicit def groupLaws[A: Eq: Arbitrary] = GroupLaws[A]
+  implicit def orderLaws[A: Cogen: Eq: Arbitrary]: OrderLaws[A] = OrderLaws[A]
+  implicit def groupLaws[A: Cogen: Eq: Arbitrary]: GroupLaws[A] = GroupLaws[A]
 
   implicit val arbitraryBitSet: Arbitrary[BitSet] =
     Arbitrary(arbitrary[List[Short]].map(ns => BitSet(ns.map(_ & 0xffff): _*)))
 
-  laws[OrderLaws, Map[String, HasEq[Int]]].check(_.eqv)
+  implicit val cogenBigInt: Cogen[BigInt] =
+    Cogen[Long].contramap(_.toLong)
+
+  implicit val cogenBigDecimal: Cogen[BigDecimal] =
+    Cogen[Double].contramap(_.toDouble)
+
+  {
+    // needed for Cogen[Map[...]]
+    implicit val ohe: Ordering[HasEq[Int]] = Ordering[Int].on(_.a)
+    laws[OrderLaws, Map[String, HasEq[Int]]].check(_.eqv)
+  }
+
   laws[OrderLaws, List[HasEq[Int]]].check(_.eqv)
   laws[OrderLaws, Option[HasEq[Int]]].check(_.eqv)
   laws[OrderLaws, Vector[HasEq[Int]]].check(_.eqv)
@@ -114,6 +125,9 @@ class LawTests extends FunSuite with Discipline {
   implicit val arbitraryComparison: Arbitrary[Comparison] =
     Arbitrary(Gen.oneOf(Comparison.GreaterThan, Comparison.EqualTo, Comparison.LessThan))
 
+  implicit val cogenComparison: Cogen[Comparison] =
+    Cogen[Int].contramap(_.toInt)
+
   laws[OrderLaws, Comparison].check(_.eqv)
 
   test("comparison") {
@@ -168,12 +182,16 @@ class LawTests extends FunSuite with Discipline {
       val order = new Random(seed).shuffle(Vector.range(0, nMax))
       Order.by { (n: N) => order(n.n) }
     })
+    implicit val cogNOrder: Cogen[Order[N]] =
+      Cogen[Unit].contramap(_ => ())
     // The arbitrary `Eq[N]` values are created by mapping N values to random
     // integers.
     implicit val arbNEq: Arbitrary[Eq[N]] = Arbitrary(arbitrary[Int].map { seed =>
       val mapping = new Random(seed).shuffle(Vector.range(0, nMax))
       Eq.by { (n: N) => mapping(n.n) }
     })
+    implicit val cogNEq: Cogen[Eq[N]] =
+      Cogen[Unit].contramap(_ => ())
     // needed because currently we don't have Vector instances
     implicit val vectorNEq: Eq[Vector[N]] = Eq.fromUniversalEquals
     // The `Eq[Order[N]]` instance enumerates all possible `N` values in a
@@ -210,6 +228,8 @@ class LawTests extends FunSuite with Discipline {
       Eq[A].on(_.a)
     implicit def hasEqArbitrary[A: Arbitrary]: Arbitrary[HasEq[A]] =
       Arbitrary(arbitrary[A].map(HasEq(_)))
+    implicit def hasCogen[A: Cogen]: Cogen[HasEq[A]] =
+      Cogen[A].contramap(_.a)
   }
 
   case class HasPartialOrder[A](a: A)
@@ -219,6 +239,8 @@ class LawTests extends FunSuite with Discipline {
       PartialOrder[A].on(_.a)
     implicit def hasPartialOrderArbitrary[A: Arbitrary]: Arbitrary[HasPartialOrder[A]] =
       Arbitrary(arbitrary[A].map(HasPartialOrder(_)))
+    implicit def hasCogen[A: Cogen]: Cogen[HasPartialOrder[A]] =
+      Cogen[A].contramap(_.a)
   }
 
   case class LawChecker[L <: Laws](name: String, laws: L) {
--- a/kernel-laws/src/test/scala/cats/kernel/laws/LawTests.scala
+++ b/kernel-laws/src/test/scala/cats/kernel/laws/LawTests.scala
@@ -30,9 +30,13 @@ class LawTests extends FunSuite with Discipline {
   implicit val arbitraryBitSet: Arbitrary[BitSet] =
     Arbitrary(arbitrary[List[Short]].map(ns => BitSet(ns.map(_ & 0xffff): _*)))
 
+  // this instance is not available in scalacheck 1.13.2.
+  // remove this once a newer version is available.
   implicit val cogenBigInt: Cogen[BigInt] =
     Cogen[Long].contramap(_.toLong)
 
+  // this instance is not available in scalacheck 1.13.2.
+  // remove this once a newer version is available.
   implicit val cogenBigDecimal: Cogen[BigDecimal] =
     Cogen[Double].contramap(_.toDouble)
 
--- a/laws/src/main/scala/cats/laws/discipline/Arbitrary.scala
+++ b/laws/src/main/scala/cats/laws/discipline/Arbitrary.scala
@@ -2,8 +2,10 @@ package cats
 package laws
 package discipline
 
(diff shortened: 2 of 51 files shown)


non:
I should also note that I commented out some other implicits that we can probably remove. I definitely need to do a sweep through this PR before we consider merging it, but I wanted to get it up now to show people how things work.

We also probably want to add `Cogen` instances for all our data types: I did the minimal set necessary to get our current tests running again. We could also stand to do some renaming, since `cats.laws.discipline.arbitrary` is no longer a totally correct name (it also contains `Cogen` instances).

non:
Hit another cryptic issue, with the `js` subproject. It seems to be building against the wrong version of ScalaCheck, and then crashing during one of the JS linking steps.

> [info] Compiling 1 Scala source to /Users/erik/w/cats/js/target/scala-2.11/test-classes...
> [info] Fast optimizing /Users/erik/w/cats/js/target/scala-2.11/cats-js-test-fastopt.js
> [error] Referring to non-existent method org.scalacheck.Arbitrary$.arbFunction1(org.scalacheck.Arbitrary)org.scalachec$
.Arbitrary
> [error]   called from cats.js.tests.FutureTests.<init>()
> [error]   called from cats.js.tests.FutureTests.__exportedInits
> [error]   exported to JavaScript with @JSExport
> [error] involving instantiated classes:
> [error]   cats.js.tests.FutureTests

I was able to work around this, but it's really unnerving. I pushed the workaround to get the PR to turn green, but ideally we should fix this before merging.

I'm going to create a checklist in the PR to try to track these deferred issues.

non:
I was wrong -- all of the tests seem to be broken under *scala.js*. I'm going to put this down and come back to it. I should probably download some ScalaCheck jars to be sure that they look OK.

non:
I still haven't figured out what is happening here. I tried on a different computer and was able to test some of these sub-projects but encountered the error in another place (free). I've tried cleaning (and manually nuking `target`) but none of these have helped.

I also asked in the Scala.js Gitter and no one had any ideas.

I'm a bit stumped. If anyone is looking for a frustrating thing to work on this weekend I'd appreciate some help. Even just checking out this branch and running `validate` yourself might give us another data point.

(I'm going to be out of town for a couple days, so I may not be as responsive as I have been recently.)

durban:
I've sent a pull request, please take a look (direct link to the commit: https://github.com/non/cats/pull/3/commits/191e51f6587e1b623f49660f3b5e6dae2038ed90). With this commit, `validate` succeeds for me.

I'm not exactly sure what's actually happening, but an educated guess is the following:
- yes, JS projects (at least some of them) use scalacheck 1.12.5
- JVM projects doesn't
- both kinds of projects use the doctest sbt plugin
- if `doctestWithDependencies` is true (it seems to be the default), the doctest plugin directly modifies `libraryDependencies`, and _adds scalacheck 1.12.5_ (and I guess this has a higher priority than the transitively used scalacheck 1.13)
- for JVM projects, this setting is `false`
- for JS projects, also setting it to `false` seems to solve the version mismatch
- (I also did some other modifications to make it compile; also there was a typo in build.sbt)

I hope this helps :-)

non:
@durban Thanks so much!!!

I merged this to get Travis started but I'm also going to run it locally. The explanation seems plausible, and might explain why the problems are somewhat unpredictable.

ceedubs:
Thanks a bunch, @non. This is great. I'm wondering if @fthomas has any thoughts on the law violation?

non:
For now I've elected to remove the *left distributivity* law from `MonadCombine`. Would love a review by @fthomas, but I think if no one objects and we can get the sign-offs we should just go for it.

If anyone feels strongly I'm also fine with downgrading `MonadCombine[Option]` to `MonoidK[Option]` and `MonadFilter[Option]`.

non:
(Sorry for all this test noise. I will probably rebase the last few commits here, since they basically just consist of me using Travis to do testing.)

adelbertc:
If I recall @mpilquist had knowledge around weak/strong laws, might the left-distributivity law of `MonadCombine` be related?

non:
I don't think we should worry about the coverage errors here. AFAIK the uncovered parts of the patch are cogen instances I created for some of our data types. I thought they were needed (but maybe weren't used) but I think we should include these anyway.

eed3si9n:
+1 on removing monadCombineLeftDistributivity. A useful law to add might be enforcing `empty` added by `MonadFilter` matches up with that of `MonoidK` if you don't want to remove `MonadFilter` altogether in #1342.

non:
@eed3si9n Agreed on that! I think we get this for free since we're also checking both the `MonadFilter` laws and `MonoidK` laws using the type class. (If it turns out I'm wrong we should add that explicitly.)

eed3si9n:
@non They both introduce `empty`, but I don't think any laws are checking that those values match up for a datatype.

non:
@eed3si9n One of us must be confused. I'll try to lay out my reasoning to figure out where the confusion lies:

1. `MonadCombine` and `MonoidK` are both traits with a `def empty[A]: F[A]` declaration.
2. Instances (e.g. [`MonadCombine[Option]`](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/instances/option.scala#L11)) define a concrete `empty` method.
3. Each of the above type classes has a laws class which requires that certain laws hold for its `empty` member.
4. The laws class for `MonadCombine` extends both of the above law classes, so  it checks that those laws hold for the instance being checked.
5. We check the laws for our instances (e.g. [`MonadCombine[List]`](https://github.com/typelevel/cats/blob/master/tests/src/test/scala/cats/tests/ListTests.scala#L16)) which includes those tests above.

Am I missing something? It seems like we're already ensuring consistency through the law-checking, right?

eed3si9n:
@non For clarification, let's call the `MonodK`'s empty `fzero`.

Here's my logic:

1. Given a datatype `F[_]`, `MonoidK[F]`, `MonadFilter[F]`, and `MonadCombine[F]` can be implemented as distinct stack of traits that ends up as distinct implicit values.
2. Within each instances, the `empty` method (and `fzero` method) can be implemented in accordance of the law.
3. It's conceivable that `MonoidK[F].fzero` and `MonadFilter[F].empty` have different semantics for some datatypes (This [comment](https://github.com/typelevel/cats/issues/1336#issuecomment-242954100) supports this idea; right-bias `Either` could be such example?)
4. Since `MonadCombine` brings two hierarchies together, I think the emptiness should be consistent for datatypes that support it, i.e. `fzero == empty` should be a law.

It's completely possible that I'm confused.

non:
@eed3si9n If the method was called `fzero` then I totally agree with you. My sense was that because the same method name is used, it's not possible to have two different ones that end up getting used for an instance, due to the way method overriding works. Can that happen?

eed3si9n:
So you're saying that the implementation of emptiness would never diverge because no one would define instances like this in real life?

```scala
implicit def instancesForEither[A1: Monoid]: MonadCombine[Either[A1, ?]] with Monad[Either[A1, ?]] with Alternative[Either[A1, ?]] {
  def empty[A2]: Either[A1, A2] = Left(???)
  ....
}
implicit def instancesForEither2[A1]: MonadFilter[Either[A1, ?]] {
  def empty[A2]: Either[A1, A2] = Right(???)
  ....
}
```

non:
@eed3si9n Right. Or at least, defined that way there would be ambiguous implicit problems.

Now I understand the issue you're pointing out. I think it's a valid concern but it's a lot deeper than just this case: for example, what if there were both `Functor` and `Applicative` instances that had different `map` methods? For now, we rely on implementors to ensure that implicit scope remains reasonable.

If you want I can open an issue to think about how best to ensure this doesn't happen (even if we write out own test, there's no guarantee that someone else down the line doesn't set up a bizarre  set of implicits somewhere).

kernel-laws/src/test/scala/cats/kernel/laws/LawTests.scala
```diff
@@ -24,13 +24,24 @@ class LawTests extends FunSuite with Discipline {
...
 
   implicit val arbitraryBitSet: Arbitrary[BitSet] =
     Arbitrary(arbitrary[List[Short]].map(ns => BitSet(ns.map(_ & 0xffff): _*)))
 
-  laws[OrderLaws, Map[String, HasEq[Int]]].check(_.eqv)
+  implicit val cogenBigInt: Cogen[BigInt] =
+    Cogen[Long].contramap(_.toLong)
+
+  implicit val cogenBigDecimal: Cogen[BigDecimal] =
+    Cogen[Double].contramap(_.toDouble)
```
ceedubs: It looks like scalacheck has `Cogen` instances for `BigInt` and `BigDecimal`. Is there a reason we need separate instances here? If so, we probably should document it.

laws/src/main/scala/cats/laws/discipline/Arbitrary.scala
```diff
@@ -2,48 +2,77 @@ package cats
...
-         override def toString = s"<function testing whether input is less than $x>"
-       }))
+  implicit def catsLawsCogenForTry[A](implicit A: Cogen[A]): Cogen[Try[A]] =
+    Cogen((seed, x) => x match {
+      case Success(a) => A.perturb(seed, a)
+      case Failure(e) => Cogen[Throwable].perturb(seed, e)
+    })
+
+  implicit def catsLawsCogenForFunction0[A](implicit A: Cogen[A]): Cogen[Function0[A]] =
+    A.contramap(_())
```
ceedubs: A lot of these instances also look like they are defined in scalacheck. Same question as https://github.com/typelevel/cats/pull/1345/files#r77895807

non: @ceedubs I'll go back through and see which ones I can delete.

kernel-laws/src/test/scala/cats/kernel/laws/LawTests.scala
non: Those are only in *master* and were added (by me) recently in a just-merged PR. I think we still need them until another release is out.

ceedubs: Ah okay.

ceedubs:
👍 LGTM but I think that we should create issues to follow up on the left-distributivity law and the short-lived cogen instances.

ceedubs:
By the way, @non I'm really excited about this, and it's fantastic that it caught an issue!

non:
@ceedubs It turns out one of the instances (`Cogen[Stream[A]]`) was no longer needed, so I removed it. I added the following comment to all the other instances:

> this instance is not available in scalacheck 1.13.2.
> remove this once a newer version is available.

johnynek:
I'm not super excited about losing a law. I'd rather lose an instance unless we can make a strong case as to why the law is overly tight.

non:
@johnynek One reason not to mandate it is that there seems to be disagreement about what laws this structure should obey: https://wiki.haskell.org/MonadPlus

It seems like the options are:

 * Left Distribution: `fa.flatMap(f) <+> fb.flatMap(f)` = `(fa <+> fb).flatMap(f)`
 * Left Catch: `pure(a) <+> fb` = `pure(a)`

Either of these would be valid laws to have. One suggestion would be to have two type classes, each with one of these laws. Supporting the *left catch* style might be nice for types like `Try` (since a *take-left* universal monoid is often what you want when doing things like parsing).

That said, I'm fine with removing it if everyone else agrees that we want to commit to the *left distribution* flavor of `MonadCombine`.

BennyHill:
Hi Guys,

I'm keen to get this PR in if possible so I can (hopefully) put in a PR for 2.12.0-RC1 but it needs the new versions of libs this PR brings in.

If there are some open questions, is it possible that they could be moved to a new issue/PR so I can PR on top of this one?

BennyHill:
Or, I see that this is on the 0.8.0 milestone, so perhaps merge to a 0.8.x branch? (along with https://github.com/typelevel/cats/pull/1333 ;) )

johnynek:
@non what about following Haskell's proposal:
https://wiki.haskell.org/MonadPlus_reform_proposal
So, `MonadCombine` would satisfy left distribution and `MonadOr` would satisfy left-catch.

adelbertc:
I'm 👍 on this since the scope of this already rather large PR was to get us up to date on dependencies. 

That Haskell proposal is interesting though - if folks are OK let's call this good and we can create a ticket to follow up with figuring out what we want to do with MonadPlus?

johnynek:
:+1: on the merge for this along with a ticket that we address the idea of
which laws we/MonadOr before the next release.

PR #1347, opened 2016-09-01
Add consistency laws for .combineAll and .combineAllOption.

### LINKED ISSUE #1346: The combineAll method of MapMonoid always returns an empty map
I was updating my tutorial to Cats 0.7.0 and noticed that combining `Map`s doesn't work as expected. 
The `combine` method works:

```scala
import cats.instances.int._, cats.instances.map._

val totals1 = Monoid[Map[String,Int]].combine(Map("a" -> 12, "b" -> 21), Map("a" -> 10))
println(totals1) // Map(a -> 22, b -> 21)
```

However, the `combineAll` method doesn't work:

```scala
val scores = List(Map("a" -> 12, "b" -> 21), Map("a" -> 10))
val totals2 = Monoid[Map[String,Int]].combineAll(scores)
println(totals2) // Map()
```

I looked at the [combineAll implementation](https://github.com/typelevel/cats/blob/master/kernel/src/main/scala/cats/kernel/instances/map.scala), and it seems that `acc` is initialized as an empty `Map`, but it isn't changed throughout the function. Maybe that's the cause of the problem.

non:
These tests will catch sitautions where the optimized methods are
broken (i.e. become inconsistent with the behavior of .combine).
These tests caught an issue with Monoid[Map[K, V]].combineAll, which
was broken.

Fixes #1346.

### DIFF
--- a/kernel-laws/src/main/scala/cats/kernel/laws/GroupLaws.scala
+++ b/kernel-laws/src/main/scala/cats/kernel/laws/GroupLaws.scala
@@ -1,6 +1,8 @@
-package cats.kernel.laws
+package cats.kernel
+package laws
 
 import cats.kernel._
+import cats.kernel.instances.option._
 
 import org.typelevel.discipline.Laws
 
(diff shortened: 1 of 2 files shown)


adelbertc:
👍

kailuowang:
:+1:

Issue #403, opened 2015-07-13
Curate order of type class list in docs


ceedubs:
http://non.github.io/cats//typeclasses.html has a list of type classes present in Cats. Currently it is dynamically generated in alphabetical order. The fact that it's dynamic and therefore we don't need to remember to update it is great. However, it would probably be nice to introduce the simplest type classes first and progress to the more complicated ones.

For example it almost certainly makes sense to learn Functor before Applicative, but currently Applicative is first in the list.

See this [gitter conversation](https://gitter.im/non/cats?at=55a30fc68e28b0c71ac9585d) for motivation.

InTheNow:
I've been reading through the docs and created a summary for each page that is currently done. For each page I listed  what I considered was a pre-requirement (`requires`) for that page, based on the contents of that page. If a topic was mentioned but also explained, I marked this as `repeats`.

So an interesting example is `Applicative` that is discussed in terms of `Monad`. `Monad ` is then introduced on its own page as "Monad extends the Applicative typeclass". So there is a circular dependency on the readers knowledge. Also, https://github.com/non/cats/issues/413 is also highlighted.

The other reason for this is to see how the structures of each page compares, so perhaps some general guidelines can be introduced - eg all pages have an implicit introduction that describes XX in a simple, non-technical way.

```
requires: [forcomprehensions]
repeats: []
# Getting Started


requires: [apply, functor, monad]
repeats: []
# Applicative
## Applicative Functors & Monads


requires: []
repeats[functor]
# Apply
### map
### compose
### ap
### ap2, ap3, etc
### tuple2, tuple3, etc
## apply builder syntax


requires: [applicative, apply, forgetfulfunctor, freemonad, functor, monad]
# Free Applicative Functor


requires: [arrow, applicative, functor, monad]
repeats: [adt, fixedpointfunctor, forgetfulfunctor, naturaltransform]
# Free Monad <!--- Empty --->
## What is it?
## Using Free Monads
### Study your topic
### Create an ADT representing your grammar
### Free your ADT
#### 1. Create a `Free` type based on your ADT
#### 2. Prove `KVStoreA[_]` has a `Functor`.
#### 3. Create smart constructors using `liftF`
#### 4. Build a program
#### 5. Write a compiler for your program
#### 6. Run your program
#### 7. Use a pure compiler (optional)
## For the curious ones: what is Free in theory?
### For the very curious ones
## Future Work (TODO)
## Credits


requires: []
repeats: [kindprojector]
# Functor
[...]

PR #1348, opened 2016-09-01
Doc fixups

longcao:
This PR attempts to address parts of https://github.com/typelevel/cats/issues/403.

The good bits:

- Fix up local-server jekyll script
- Freshen up links to scala-exercises and herding cats
- Move `tut`-checked docs to `typeclasses` and `datatypes` subfolders
  - Removes 'tut' from URL paths
  - This will change doc urls to be the following path scheme: `http://$BASEURL/cats/typeclasses/*.html` and `http://$BASEURL/cats/datatypes/*.html`
- Group typeclasses as suggested by https://github.com/typelevel/cats/issues/403#issuecomment-146160208 and similar to the [scala-exercises](https://www.scala-exercises.org) list.
- Move `Id` link + docs to data types

This is how it looks on my local machine:
screenshot 2016-09-01 15 17 45

**Note:** I don't have really good/original names for the groupings. Taking suggestions for names (and for the ordering) since Group 1, 2, 3, and Misc are by and large just placeholders.

This is mostly cosmetic but any links to the old docs URLs will break, e.g. http://typelevel.org/cats/tut/kleisli.html. Let me know if this PR is ok.


non:
This looks like a big improvement! Thanks!

I'm fine merging this now and working on groupings as a follow-up. Alternately, I can take some time and try to come up with good groupings to suggest for this PR.

longcao:
@non makes sense. The groupings seem fairly independent so maybe thinking of them as 'tracks' might help as well. Let me know what you might be thinking of.

ceedubs:
Thanks, @longcao! I like these changes.

One thing I'm not entirely sure about is whether we want to try to group like this. No matter how we break down the groups you are going to have some cross-cutting concerns since `Traverse` extends both `Foldable` and `Functor`. Maybe it's fine to ignore this fact when grouping, as I think `Foldable` is the more natural segue into `Traverse`. But I find myself wondering if it would be better to just have a single track of type classes increasing in complexity (I think `Semigroup` is a natural first one). I'm not sure, I could really be convinced either way on this.

longcao:
@ceedubs yeah - and I'm easily convinced of either! I do also agree that `Semigroup` + `Monoid` is an easy starting point for anyone with zero knowledge.

longcao:
I just saw https://github.com/typelevel/cats/pull/1369, which totally looks awesome and seems to supercede/conflict the work done here, so I'm closing this PR in favor of tacking on some follow-on commits with https://github.com/typelevel/cats/pull/1369 after it gets merged.

PR #1352, opened 2016-09-03
Add lift to FunctionK

andyscott:
Quick attempt at providing a `lift` method for `FunctionK`, by means of a macro.

In my local testing, this works for `Free` and other more complicated types if you've enabled some form of the SI-2712 fix.

Thoughts?

### DIFF
--- a/core/src/main/scala/cats/arrow/FunctionK.scala
+++ b/core/src/main/scala/cats/arrow/FunctionK.scala
@@ -3,7 +3,7 @@ package arrow
 
 import cats.data.Coproduct
 
-import reflect.macros.blackbox.Context
+import cats.macros.MacroCompat
 
 trait FunctionK[F[_], G[_]] extends Serializable { self =>
   def apply[A](fa: F[A]): G[A]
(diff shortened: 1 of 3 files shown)


andyscott:
Everything passes except some style assertions, such as using null.

non:
Thanks!

Would you mind writing some Scaladoc to explain what the macro does, how to use it, etc.? I think it will be relatively hard for casual users to look at this and understand how to use it.

kailuowang:
Thanks very much! I like the idea (yah! less boilerplate for `Free`) and the test and doc looks good to me, but I am not too qualified to reviewed the macro part, other than @non, maybe @fthomas ?

andyscott:
Less boilerplate for `Free` was exactly what motivated me to write this! I am considering using free monads on a project where a number of people have worked minimally in Scala, so I've been crafting a demo project that minimizes as many "scary" concepts as possible.

andyscott:
Do I need to fix the code coverage issues for this PR? I'm not entirely sure how I'd be able to provide coverage for some of the "internal" methods to the macro.

non:
@andyscott I think it's fine -- I'm not sure if those error message can be hit, but I'm glad you're providing an error message in case someone does hit them. 👍

johnynek:
👍

PR #1353, opened 2016-09-03
Include generated files in source JARs

### LINKED ISSUE #1335: Generated sources missing from source JAR
It seems that the `.scala` files generated by `project/Boilerplate.scala` are not included in the published source JAR (e.g., http://repo1.maven.org/maven2/org/typelevel/cats-core_2.11/0.7.0/cats-core_2.11-0.7.0-sources.jar). It would be nice to have them included, if possible.

See, e.g., http://stackoverflow.com/questions/33760101/sbt-publish-generated-sources.

durban:
Fixes #1335.


kailuowang:
How about we make a val of these settings and give it an indicative name?

kailuowang:
Thanks very much! :+1:

non:
LGTM 👍 Thanks!

PR #1354, opened 2016-09-04
Add semigroup instance for Ior

Fristi:


adelbertc:
👍

kailuowang:
I am curious why not testing SerializableTests here. ( I noticed that we are not SerializableTest `BiFunctor` , which I am not sure why either. )

Fristi:
I'll add those in the next commit!

non:
Thanks! 👍

PR #520, opened 2015-09-10
Add toNel syntax to List

ceedubs:

PR #1355, opened 2016-09-04
vector.toNev syntax

benhutchison:
For consistency with NonEmptyList and because its useful https://github.com/typelevel/cats/pull/520


non:
Seems reasonable! Thanks! 👍

ceedubs:
👍 thanks!

PR #1356, opened 2016-09-05
append and prepend on NonEmptyVectors

benhutchison:
Append and prepend are everyday operations on Vectors and the need quickly arises when working with NonEmptyVectors. They maintain the NonEmpty invariant.


ceedubs:
👍 thanks, @benhutchison!

non:
Thanks! 👍

PR #1370, opened 2016-09-12
Remove RecursiveTailRecM, move stack safety checks into MonadLaws - f…

### LINKED ISSUE #1283: Make stack safety law part of FlatMapRec as opposed to ad-hoc checks
Move these https://github.com/typelevel/cats/blob/master/tests/src/test/scala/cats/tests/MonadRecInstancesTests.scala to be actual law checks

adelbertc:
…ixes #1278 and fixes #1283 

cc @johnynek @non


adelbertc:
Hm it seems it fails for `Future` in the JS tests since you need to `Await` on the `Future` to test for equality, but that gets tricky in JS land.. is there any way around this?

free/src/main/scala/cats/free/Free.scala
```diff
@@ -138,7 +132,7 @@ sealed abstract class Free[S[_], A] extends Product with Serializable {
    * enough, this will work
    */
   final def foldMapUnsafe[M[_]](f: FunctionK[S, M])(implicit M: Monad[M]): M[A] =
-    foldMap[M](f)(M, RecursiveTailRecM.create)
+    foldMap[M](f)
```
guersam: Should `foldMapUnsafe` be deprecated then?

kailuowang:
Should the unsafe `defaultTailrecM`  be removed?
https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Monad.scala#L24

free/src/main/scala/cats/free/Free.scala
johnynek: this should just be removed if you ask me. I doubt anyone is using it and removing `RecursiveTailRecM` is at least as big of an issue.

non:
@adelbertc in the worst-case you can make the test platform-specific using `if (Platform.isJvm) ...`. This might be a good candidate for that.

I know ScalaTest supports a style of asynchronous testing that works with futures in JS, so you could also look into that.

adelbertc:
@non I looked into the async testing support in ScalaTest but it seems like due to how our tests are constructed it's not really compatible? My understanding is the new async stuff lets you give ScalaTest a `Future[Assertion]` instead of just an `Assertion`, but our laws use `Eq` which want `Boolean` ?

non:
Looks like the build failed because some of our `.tut` files need updating.

johnynek:
Of course, I'm +1. I think asking for stack safety (given it seems always achievable) is fine.

I do know of a Monad that was not made stack safe (yet?) Enumeratee, I think it was in iteratee-io (@travisbrown). I have not looked too hard at making it stack safe.

One issue is that the worst case may require trampolines and some library authors might not like that. I wonder if we could devise a trampoline-like approach that only impacts tailRecM and does not hurt the performance of flatMap.

travisbrown:
@johnynek @adelbertc `io.iteratee.Enumerator` has a monad instance but the `tailRecM` is not (currently) stack-safe. The approach that Cats uses in e.g. the instance for `Vector` should work, but it was non-trivial and recursive binding for enumerators isn't a use case I've ever run into, so I passed on it in the 0.6 release.

kailuowang:
👍

PR #1369, opened 2016-09-11
New Proposal for the Cats Microsite

juanpedromoreno:
This PR brings a new proposal for the [Typelevel Cats website](http://typelevel.org/cats/) microsite. 

It's made thanks to the https://github.com/47deg/sbt-microsites plugin and:

* Provides the `override.css` file and the images resources.
* Removes the versioned [site](https://github.com/typelevel/cats/tree/master/docs/src/site).
* Modifies the src/main/tut resources according to: root files, typeclasses and datatypes.
* Updates the current documentation.
* Other minor tweaks to be able to generate the site successfully.

The generated [site](https://github.com/typelevel/cats/tree/master/docs/src/site) has been removed since is not needed from now on. As you can see in the [contributing.md](https://github.com/47deg/cats/blob/master/CONTRIBUTING.md#generating-the-site) file, the `site` can be generated each time on the fly. 

If you want to test locally you need [Jekyll](https://jekyllrb.com/) locally, depending on your platform, you might do this with:

```bash
yum install jekyll

apt-get install jekyll

gem install jekyll
```

Once you have installed it, you could perform the following steps:

* Build the microsite:

```bash
sbt docs/makeMicrosite
```

* Then, to visualize the website in your browser, in a shell, navigate to the generated site directory in `docs/target/site`
* Start jekyll with `jekyll serve`
* Navigate to http://localhost:4000/cats/

Please, review at your convenience. Thanks!

Join-effort from the 47Deg team :)

PR #1376, opened 2016-09-15
Bump version number to 0.7.2 on the cats website

jhzab:


kailuowang:
shall we wait on #1369?

jhzab:
I honestly see no big reason not to update the site. It's _hopefully_ less than 5 minutes of someones times, but makes it easier for new users and paints a better picture of the project. ;)

PR #1382, opened 2016-09-19
Add scala 2.12.0-RC1

BennyHill:
This PR adds 2.12-0-RC1 support.

Please note that `makesite` has been disabled due to an error in 2.12, see comments


BennyHill:
Re failing build on 2.12.0-RC1 this is during `tut` in:

```sbt
[tut] compiling: ./docs/src/main/tut/either.md
[tut] *** Error reported at ./docs/src/main/tut/either.md:239: failure was asserted but no failure occurred
doApp: scala.util.Either[Object,ServiceValue]
```
So this diff passes in 2.12.0-RC1 only, but is not the solution, just a helper/illustration:

```diff
@@ -235,7 +235,7 @@ object Service {
 Let's say we have an application that wants to do database things, and then take database
 values and do service things. Glancing at the types, it looks like `flatMap` will do it.
 
-```tut:fail:silent
+```tut:silent
 def doApp = Database.databaseThings().flatMap(Service.serviceThings)
 ```

```

This is all clearly due to new `Either` semantics in 2.12 so I'll leave this as is for now.

2.11.8 build failure looks like the old OOM issues, so I'll turn of some instrumentation as I mentioned ^^

BennyHill:
Actually, the 2.11 build failure is identical to https://travis-ci.org/typelevel/cats/jobs/161257117 so I'm done here for now :smile:

magicknot:
@BennyHill you don't need to compile scoverage anymore. You can do something like this:

plugins.sbt:
`addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.4.0")`

build.sbt
`coverageScalacPluginVersion := "1.3.0-RC1"`

travisbrown:
@magicknot Thanks! I just confirmed locally that that works.

BennyHill:
I know - I did most of the changes for scoverage along with @paulp :smile:

BennyHill:
But we _do_ need the more recent sbt plugin as well, `1.5.0-RC1` and so I will use that too.

BennyHill:
Todo: Add new issue with `makesite` error once this is in, with notes on how to replicate. Also add new issue to reintroduce `makesite`, either in full with fix or partial, excluding 2.12

BennyHill:
Ref two scala issues:
- https://issues.scala-lang.org/browse/SI-9923
- https://issues.scala-lang.org/browse/SI-9931

travisbrown:
Mind if I close this, @BennyHill, since these commits are in #1415?

BennyHill:
Closing since these commits are in #1415

Issue #1351, opened 2016-09-02
Need to rename cats.js.std to cats.js.instances


non:
Looks like we missed this when we converted everything else over.

PR #1385, opened 2016-09-21
rename cats.js.std to cats.js.instances

sritchie:
Fixes https://github.com/typelevel/cats/issues/1351!


kailuowang:
👍  on build green. Thanks!

kailuowang:
I'll merge since it's such a straightforward change.

PR #1388, opened 2016-09-22
Add StateT.set and StateT.setF.

tixxit:
This adds `StateT.set` and `StateT.setF` methods on the `StateT` companion object. Their omission seemed to be a bit conspicuous so it may just be that I'm missing something!


non:
Seems reasonable to me. 👍

adelbertc:
👍

Issue #1387, opened 2016-09-21
Consider using Travis-CI Trusty Beta


BennyHill:
ref https://docs.travis-ci.com/user/ci-environment/

The scala.js continues to randomly fail, I suspect due to the limited memory, 4 GB __Max__, available in Travis-CI container builds and hence little available for SBT.

Their Trusty Beta environments have 7.5 GB, so would allow  an increase to 5 or 6 GB [in the options] (https://github.com/typelevel/cats/blob/master/.jvmopts#L4). The downside of the the slower boot time is a small fraction of the overall build time.

kailuowang:
I am going to give this a try.

kailuowang:
fixed by #1389

PR #1389, opened 2016-09-22
try travis trusty beta with higher memory

kailuowang:
as suggested by @BennyHill in #1387


non:
Sounds like it's worth a shot! 👍

adelbertc:
👍

BennyHill:
Just a quick note for future ref.

The build has 3 JVM forks, each of which fire up a new JVM with its own memory requirements. So we need to keep an eye out that these forks are not the cause of the OOM in envs like travis.ci that have limited resources. I'm not saying the forking is right/wrong, just "noting".

PR #1393, opened 2016-09-23
Bad markdown spacing was making subtitle to not render properly

alpeb:


peterneyens:
Thanks @alpeb, but it seems that our new site (after #1369) isn't as picky to the new lines, so it is rendered correctly (see http://typelevel.org/cats/datatypes/kleisli.html#other-uses).

alpeb:
Ok cool 👍

non:
So does that mean we should close this PR? @alpeb feel free to close this if you think it's fixed (and either way thanks for looking into it!)

PR #1394, opened 2016-09-24
Adding sequenceM method to Traverse

zainab-ali:


core/src/main/scala/cats/Traverse.scala
```diff
@@ -84,6 +84,24 @@ import simulacrum.typeclass
...
+   * scala> import cats.implicits._
+   * scala> val x: List[Option[List[Int]]] = List(Some(List(1, 2)), Some(List(3)))
+   * scala> val y: List[Option[List[Int]]] = List(None, Some(List(3)))
+   * scala> x.sequenceM
+   * res0: Option[List[Int]] = Some(List(1, 2, 3))
+   * scala> y.sequenceM
+   * res1: Option[List[Int]] = None
+   * }}}
+   */
+  def sequenceM[G[_], A](fgfa: F[G[F[A]]])(implicit G: Applicative[G], F: FlatMap[F]): G[F[A]] =
```
adelbertc: sequenceA? since applicative

kailuowang: Well this one is corresponding `traverseM`, maybe both should be renamed? (if we ended up doing that we need to add the "breaking change" label. )

zainab-ali: What about `flatTraverse` or `flatSequence`, taking the convention from Scala's `flatMap`?  I realize that this isn't well established in other languages, but both `traverseM` and `sequenceM` aren't as self explanatory

peterneyens: There was a similar conversation when I added `traverseM` in #1020.

kailuowang: Sorry for the delayed response. I like `flatTraverse` and `flatSequence`, they make sense to me.

kailuowang:
Thanks very much! @zainab-ali 👍  LGTM.

travisbrown:
:+1:

peterneyens:
I was not convinced about the name change at first, but after playing around with it I actually like the analogy between `flatMap` and `flatTraverse` : 

```scala
def smallerNaturals(i: Int): List[Int] = 
  List.range(1, Math.max(i, 1))

def smallerNaturalsOpt(i: Int): Option[List[Int]] = 
  if (i <= 1) None else Some(List.range(1, i))

List(2,3,4).flatMap(smallerNaturals)
// List[Int] = List(1, 1, 2, 1, 2, 3)

List(2,3,4).flatTraverse(smallerNaturalsOpt)
// Option[List[Int]] = Some(List(1, 1, 2, 1, 2, 3))
```

So :+1: , thanks @zainab-ali.

PR #1395, opened 2016-09-24
Add Order instance for scala.Symbol

durban:


kernel/src/main/scala/cats/kernel/instances/symbol.scala
```diff
@@ -0,0 +1,15 @@
...
+
+package object symbol extends SymbolInstances
+
+trait SymbolInstances {
+  implicit val catsKernelStdOrderForSymbol: Order[Symbol] = new SymbolOrder
+}
+
+class SymbolOrder extends Order[Symbol] {
+  override def eqv(x: Symbol, y: Symbol): Boolean =
+    x eq y
```
johnynek: Any reason not to use == here? I assume Symbol's equal method is correct and we don't need to worry here that eq is correct.

milessabin: `Symbol`s are interned ... might as well take advantage of that fact.

durban: Yes, that was my intention.

johnynek: yes, I know that `eq` is how it is implemented due to the interning, but `equals` already does this:
https://github.com/scala/scala/blob/2.12.x/src/library/scala/Symbol.scala#L31

so, this, to me, assumes the jit may not do the job, yet each reviewer of the code has to recall this fact about how `Symbol` instances are created.

That said, I'm happy to simply add a comment here to explain this so that the code is clearer to someone unfamiliar (I think `Symbol` is actually rather unused as a scala feature these days).

durban: Ok, comment added in 318e456.

johnynek:
👍

kailuowang:
👍

PR #1398, opened 2016-09-30
Add more syntax

Fristi:
@non As we talked about at Lamba.world :-)

- [x] Compose syntax
- [x] `Monoid.combineAll` which I would like syntax for, is the same thing as `Foldable.fold`or `Foldable.combineAll` which already comes in from `Foldable.Ops` right?

### DIFF
--- a/core/src/main/scala/cats/arrow/Compose.scala
+++ b/core/src/main/scala/cats/arrow/Compose.scala
@@ -8,10 +8,10 @@ import simulacrum.typeclass
  */
 @typeclass trait Compose[F[_, _]] { self =>
 
-  @simulacrum.op(">>>", alias = true)
+  @simulacrum.op(":<<", alias = true)
   def compose[A, B, C](f: F[B, C], g: F[A, B]): F[A, C]
 
-  @simulacrum.op("<<<", alias = true)
+  @simulacrum.op(">>:", alias = true)
   def andThen[A, B, C](f: F[A, B], g: F[B, C]): F[A, C] =
     compose(g, f)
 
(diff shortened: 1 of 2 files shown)


jvican:
Same here @adelbertc | @non, labelling this as hackathon would be cool!

Fristi:
Compiling took long at the conference and drained my apple battery. Therefore I committed some stuff so Travis can do the heavy lifting. Compiled/tested it locally now, it should work and in my opinion a squash merge is fine :-)

core/src/main/scala/cats/arrow/Compose.scala
```diff
@@ -7,8 +7,11 @@ import simulacrum.typeclass
  * Must obey the laws defined in cats.laws.ComposeLaws.
  */
 @typeclass trait Compose[F[_, _]] { self =>
+
+  @simulacrum.op(":<<", alias = true)
```
johnynek: I don't like the asymmetry here. If we have the `:<<` The other should be `>>:`.

Fristi: I also dislike it, but I am getting this compile error when introducing `>>:`, guess it's right associative, even when you put the `:` on the right.

```
[error] /Users/mark/Projects/cats/tests/src/test/scala/cats/tests/SyntaxTests.scala:55: type mismatch;
[error]  found   : x$1.type (with underlying type F[B,C])
[error]  required: F[D,?]
[error]     val a = x >>: y >>: z
[error]
```

johnynek:
Why doesn't `<<<` work again?

Fristi:
@johnynek It works! I thought it wouldn't due operator associativity.

Good to squash merge now I think :-)

kailuowang:
👍

johnynek:
👍

PR #1399, opened 2016-10-01
UUID instances

durban:
- `Order[UUID]`
- `Show[UUID]`

### DIFF
--- a/kernel/src/main/scala/cats/kernel/instances/uuid.scala
+++ b/kernel/src/main/scala/cats/kernel/instances/uuid.scala
@@ -1,0 +1,19 @@
+package cats.kernel
+package instances
+
+import java.util.UUID
+
+package object uuid extends UUIDInstances
+
+trait UUIDInstances {
+  implicit val catsKernelStdOrderForUUID: Order[UUID] = new UUIDOrder
+}
+
+class UUIDOrder extends Order[UUID] {
+
+  override def eqv(x: UUID, y: UUID): Boolean =
+    x == y
+
+  def compare(x: UUID, y: UUID): Int =
+    if (x eq y) 0 else x compareTo y
+}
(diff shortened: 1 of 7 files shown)


kernel/src/main/scala/cats/kernel/instances/uuid.scala
```diff
@@ -0,0 +1,19 @@
...
+
+import java.util.UUID
+
+package object uuid extends UUIDInstances
+
+trait UUIDInstances {
+  implicit val catsKernelStdOrderForUUID: Order[UUID] = new UUIDOrder
+}
+
+class UUIDOrder extends Order[UUID] {
```
johnynek: I wonder if you would be willing to add `fromComparable` to `object Order` similar to the `Show.fromToString`, and use it here. So, it would be a method like:
```scala
object Order {
  def fromComparable[T <: Comparable]: Order[T] = new Order[T] {
   ...
  }
}
```

kernel/src/main/scala/cats/kernel/instances/uuid.scala
```diff
@@ -0,0 +1,19 @@
+package cats.kernel
+package instances
+
+import java.util.UUID
+
+package object uuid extends UUIDInstances
+
+trait UUIDInstances {
+  implicit val catsKernelStdOrderForUUID: Order[UUID] = new UUIDOrder
```
johnynek: if you do the below, you can do `= Order.fromComparable[UUID]`

kernel/src/main/scala/cats/kernel/instances/uuid.scala
durban: OK, hopefully I'll have some time for that during the weekend.

durban: Done in 6bf5a18.

johnynek:
:+1:

peterneyens:
Thanks @durban :+1:

non:
I agree with @peterneyens. Otherwise, LGTM too! Thanks!

Issue #700, opened 2015-11-24
Continuation monads


vikraman:
Should `Cont`, `ContT`, `IndexedContT` be added to cats? What would be the proper place to add them?

erikerlandson:
I'd like to see this

erikerlandson:
This blog post makes a better argument for value of continuation monads than I would have made:
http://www.haskellforall.com/2012/12/the-continuation-monad.html


![continuation_kleisli](https://cloud.githubusercontent.com/assets/259898/19016956/3eba92bc-87de-11e6-8f4a-d04d752871f0.png)

johnynek:
I took a stab at this:
https://gist.github.com/johnynek/cdbb988351b63be58af053b716b33106

I don't see how to make it totally stack safe. I had an idea to use threads but that seems terrible. Anyway, something like my above implementation may actually be useful because stack depth is only proportional to the number of nested continuations, not flatMaps (which are trampolined).

I played with using this for the poor man's concurrency monad, and it seems okay.

I (or anyone) could make a cats PR if we think this would be generally useful.

erikerlandson:
@johnynek that looks very cool!   To the extent that I understand the use cases I've seen (which isn't as much as I'd like), I don't think they would manifest deep stacks.

adelbertc:
Would get a 👍 from me. It would also give me an opportunity to harp on about extensible effects and `Cont` ;)

johnynek:
take a look at the PR: #1400

danslapman:
For all, who needs continuation monad, I just ported IndexedContT from scalaz: https://github.com/danslapman/cats-conts

bblfish:
Very cool. Reading Bartosz Milewski's [Category Theory for Programmers, Scala edition](https://twitter.com/hmemcpy/status/1160870623943561216) I understood that continuations are essentially what the Yoneda Lemma is about. So I also looked around to get a better grip on how these tie in to practical applications. In a way it's not difficult - there are many examples of CPS in JS, etc... The tricky bit is getting to the point that it is crystal clear how the theory and practice tie together, so that one can explain it and think with it in everyday coding.

Many of the blog posts on the topic are in Haskell. The best of them was the [continuation monad](http://www.haskellforall.com/2012/12/the-continuation-monad.html) post referenced by @erikerlandson above.  There is also [The mother of all monads](http://blog.sigfpe.com/2008/12/mother-of-all-monads.html) which I thought very helpful.  
[...]

PR #1400, opened 2016-10-04
Add a Continuation data type

johnynek:
addresses #700 

This does not add `ContinuationT[M[_], O, I]` which can also be implemented for any monad `M[_]`. This should not be too hard, but will add a performance hit relative to the non-transformer version.

I may add a follow up that has `ContinuationT` if we feel that is useful. The main trick is to use `tailRecM` rather than the tailrec loop I have here.

### DIFF
--- a/core/src/main/scala/cats/data/Continuation.scala
+++ b/core/src/main/scala/cats/data/Continuation.scala
@@ -24,6 +24,9 @@ object Continuation {
     def apply[I](i: I): Continuation[O, I] = Const(i)
   }
   def from[I, O](fn: (I => O) => O): Continuation[O, I] = Cont(fn)
+  def from2[I1, I2, O](call: (Function2[I1, I2, O]) => O): Continuation[O, (I1, I2)] = from { fn: (((I1, I2)) => O) =>
+    call { (i1: I1, i2: I2) => fn((i1, i2)) }
+  }
 
   implicit def catsDataContinuationMonad[O]: Monad[Continuation[O, ?]] = new Monad[Continuation[O, ?]] {
     def pure[I](i: I): Continuation[O, I] = Const(i)
--- a/core/src/main/scala/cats/data/Continuation.scala
+++ b/core/src/main/scala/cats/data/Continuation.scala
@@ -1,0 +1,105 @@
+package cats
+package data
+
+import java.io.Serializable
+import scala.annotation.tailrec
+
+sealed abstract class Continuation[O, +I] extends Serializable {
+  def map[I2](fn: I => I2): Continuation[O, I2] =
+    Continuation.Mapped(this, fn)
+
+  def flatMap[I2](fn: I => Continuation[O, I2]): Continuation[O, I2] =
+    Continuation.FlatMapped(this, fn)
+
+  final def apply(fn: I => O): O = {
+    import Continuation._
+    val re = new RunEnv[O]
+    re.eval(this, re.ScalaFn(fn)).get
+  }
+}
+
+object Continuation {
+  def pure[O] = new PureBuilder[O]
+  class PureBuilder[O] {
+    def apply[I](i: I): Continuation[O, I] = Const(i)
+  }
+  def from[I, O](fn: (I => O) => O): Continuation[O, I] = Cont(fn)
+
+  implicit def catsDataContinuationMonad[O]: Monad[Continuation[O, ?]] = new Monad[Continuation[O, ?]] {
+    def pure[I](i: I): Continuation[O, I] = Const(i)
+    override def map[I1, I2](f: Continuation[O, I1])(fn: I1 => I2): Continuation[O, I2] = f.map(fn)
+    def flatMap[I1, I2](f: Continuation[O, I1])(fn1: I1 => Continuation[O, I2]): Continuation[O, I2] = f.flatMap(fn1)
+    /**
+     * flatMap is trampolined, BUT the stack depth grows proportional to the number of `from` calls
+     */
+    def tailRecM[A, B](a: A)(fn: A => Continuation[O, Either[A, B]]): Continuation[O, B] =
+      fn(a).flatMap {
+        case Right(b) => Const(b)
+        case Left(a) => tailRecM(a)(fn)
+      }
+  }
+
+  private case class Const[O, I](i: I) extends Continuation[O, I]
+
+  private case class Mapped[O, I1, I2](
+    c: Continuation[O, I1],
+    fn: I1 => I2) extends Continuation[O, I2]
+
+  private case class FlatMapped[O, I1, I2](
+    c: Continuation[O, I1],
+    fn: I1 => Continuation[O, I2]) extends Continuation[O, I2]
+
+  private case class Cont[O, I](runfn: (I => O) => O) extends Continuation[O, I]
+
+  /**
+   * We hide all our implementation in this class. Note there is only
+   * ever one output, so we put the type here.
+   */
+  private class RunEnv[O] {
+    sealed abstract class Result {
+      def get: O
+    }
+    case class Complete(get: O) extends Result
+    case class ContResult[I](runfn: (I => O) => O, fn: Fn[I]) extends Result {
+      def get: O = fn match {
+        case ScalaFn(sfn) => runfn(sfn)
+        case _ =>
+          // This is not stack safe, we could keep finding continuations
+          // the stack depth will scale as the number of inner continuations,
+          // not flatMaps (which you can see
+          runfn { i: I => eval(Const(i), fn).get }
+      }
+    }
+
+    /**
+     * This evaluates a continuation given an Fn
+     */
+    final def eval[I](c: Continuation[O, I], fn: Fn[I]): Result =
+      loop(c.asInstanceOf[Continuation[O, Any]], fn.asInstanceOf[Fn[Any]])
+
+    /**
+     * This is the tail recursive loop that partially evaluates until
+     * we reach the next continuation. Note, it compiles with the type below:
+     *
+     * final def loop[I](c: Continuation[O, I], fn: Fn[I]): Result = c match {
+     *
+     * but we erase I to Any so scala tailrec optimization can work
+     */
+    @tailrec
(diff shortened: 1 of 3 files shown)


core/src/main/scala/cats/data/Continuation.scala
```diff
@@ -0,0 +1,108 @@
...
+  final def apply(fn: I => O): O = {
+    import Continuation._
+    val re = new RunEnv[O]
+    re.eval(this, re.ScalaFn(fn)).get
+  }
+}
+
+object Continuation {
+  def pure[O] = new PureBuilder[O]
+  class PureBuilder[O] {
```
adelbertc: `final` and `private[Continuation]` ? The former for sure, the latter mostly because I saw that's how we've been doing it: https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/OptionT.scala#L167

core/src/main/scala/cats/data/Continuation.scala
```diff
@@ -0,0 +1,108 @@
+package cats
+package data
+
+import java.io.Serializable
+import scala.annotation.tailrec
+
+sealed abstract class Continuation[O, +I] extends Serializable {
```
adelbertc: `extends Product with Serializable`

Also Scaladoc saying `Continuation[O, I]` is `(I => O) => O`. Seeing the type parameters order reversed confused me for a second before I realized it was probably for the `Monad` instance

core/src/main/scala/cats/data/Continuation.scala
```diff
@@ -0,0 +1,108 @@
...
+     */
+    def tailRecM[A, B](a: A)(fn: A => Continuation[O, Either[A, B]]): Continuation[O, B] =
+      fn(a).flatMap {
+        case Right(b) => Const(b)
+        case Left(a) => tailRecM(a)(fn)
+      }
+  }
+
+  private case class Const[O, I](i: I) extends Continuation[O, I]
+
```
adelbertc: Don't know if also marking these as `final` does anything but maybe for style consistency?

core/src/main/scala/cats/data/Continuation.scala
```diff
@@ -0,0 +1,108 @@
...
+      case FlatMapped(c, next) => loop(c, RunCont(next, fn))
+      case Const(i) => fn match {
+        case ScalaFn(sfn) => Complete(sfn(i))
+        case RunCont(mkC, next0) => loop(mkC(i), next0)
+        case AndThen(sfn, next) => loop(Const(sfn(i)), next)
+      }
+    }
+
+    sealed abstract class Fn[-A]
+    case class ScalaFn[A](fn: A => O) extends Fn[A]
```
adelbertc: `extends Product with Serializable`, `final` case classes

core/src/main/scala/cats/data/Continuation.scala
```diff
@@ -0,0 +1,108 @@
...
+   * We hide all our implementation in this class. Note there is only
+   * ever one output, so we put the type here.
+   */
+  private class RunEnv[O] {
+    sealed abstract class Result {
+      def get: O
+    }
+    case class Complete(get: O) extends Result
+    case class ContResult[I](runfn: (I => O) => O, fn: Fn[I]) extends Result {
+      def get: O = fn match {
```
adelbertc: `extends Product with Serializable`, `final` case classes?

tests/src/test/scala/cats/tests/ContinuationTests.scala
```diff
@@ -0,0 +1,50 @@
...
+
+class ContinuationTests extends CatsSuite {
+
+  implicit def arbContFn[O, I](implicit I: Arbitrary[I], O: Arbitrary[O]): Arbitrary[(I => O) => O] = {
+    def call(i: I): (I => O) => O = { fn => fn(i) }
+    def const(o: O): (I => O) => O = { fn => o }
+
+    Arbitrary(Gen.oneOf(I.arbitrary.map(call(_)), O.arbitrary.map { o => const(o) }))
+  }
+  implicit def arbCont[O, I](implicit I: Arbitrary[I], fn: Arbitrary[(I => O) => O]): Arbitrary[Continuation[O, I]] =
```
adelbertc: Should this go in `cats.laws.discipline.arbitrary`

core/src/main/scala/cats/data/Continuation.scala
erikerlandson: I was also going to suggest a signature of `Continuation[+I, O]`, unless there's some reason for the current order.   Is there a reason it's `O` and not `-O` ?

core/src/main/scala/cats/data/Continuation.scala
erikerlandson: the parent class is `sealed`, so case classes are already final, aren't they?

core/src/main/scala/cats/data/Continuation.scala
adelbertc: Assuming my algebra is right I think `O` could also be made covariant like you said. A separate but related question is if we want this to be variant at all since that comes with all the gotchas associated with variance and subtyping in Scala. I would vote no.

I'm guessing the order is because it makes partial application look a bit cleaner since the `Monad` instance is free in `I` as opposed to `R`. with `Continuation[I, O]` you would have `Monad[Continuation[?, O]]` as opposed to `Monad[Continuation[O, ?]]`.

core/src/main/scala/cats/data/Continuation.scala
adelbertc: Yeah but it still bothers me from a style perspective ;)

core/src/main/scala/cats/data/Continuation.scala
johnynek: no, `O` is invariant. Note that it appears in the output, so it is either invariant or covariant. But it also appears in the input of the function, which makes it invariant or contravariant (see the `apply` method), so it is invariant.

A more direct way to see it: if you have a function `(I => O) => O` and say, `O = String`. Now due to covariance, I want to treat it as `(I => Any) => Any` can I do this? It is not clear at all how since internal to a function `(I => String) => String` we could make use of the String structure, but when I am given a function to Any, I can't do that.

So in fact, `Continuation[O, +I]` is the correct variance.

Now, about ordering of the parameters, I agree it is not the most intuitive order, but due to the the 2712 fix, right to left is the standard order searched when looking for single parameter type classes, so I think we are better off keeping the parameter there.

johnynek: about adding co-variance, I'd really like to keep it. Contravariance causes the most problems, I think, but covariance is very useful and common especially in container code.

Can you point to the problem that covariance causes?

johnynek:
So, it is `-I` twice! so that makes it positive again: `F => O` has `-F`
but `F = I => O` so you have `-(-I) = +I`

johnynek:
@adelbertc can you explain the `Product with Serializable` comments here? It seems you are requesting them only on ephemeral and private classes, so I'd rather not do it unless there is really a gain. Have we written down what this gain normally is and why we want to add the extra ceremony (final + product + serializable).

core/src/main/scala/cats/data/Continuation.scala
```diff
@@ -0,0 +1,149 @@
...
+        case ScalaFn(sfn) => runfn(sfn)
+        case _ =>
+          // This is not stack safe, we could keep finding continuations
+          // the stack depth will scale as the number of inner continuations,
+          // not flatMaps (which you can see
```
tixxit: This comment didn't quite get finished.

core/src/main/scala/cats/data/Continuation.scala
peterneyens: I wonder if this name might be confusing with `cats.data.Const` (which gives a `Const[A, B]` for an `A`) even though it is private ?

DavidGregory084:
@johnynek @adelbertc
Ensuring that traits extend `Product with Serializable` just prevents awkwardness like this:
```scala
scala> sealed trait Foo
defined trait Foo

scala> case object Bar extends Foo
defined object Bar

scala> case object Baz extends Foo
defined object Baz

scala> List(Bar, Baz)
res0: List[Product with Serializable with Foo] = List(Bar, Baz) // Product with Serializable with Foo
```

Marking case classes as final only ensures that they cannot be extended, which is good since this usually breaks compiler-generated method implementations like `.unapply`, `.copy` (among other things).

I'm sure you both knew these things already, but there is no other benefit as far as I'm aware.

I'd be inclined to suggest that `final case class Foo(...)` and `sealed trait Bar extends Product with Serializable` should always be applied to classes that form part of cats' public API but perhaps aren't necessary otherwise.

johnynek:
@DavidGregory084 Ahh, I had not seen the explanation for adding `Product with Serializable` to ADTs, but that is nice.

I agree it is nice for public classes.

13h3r:
@johnynek do you need any help to complete this?

johnynek:
I'm not sure there is any interest in merging. The requested edits are
mechanical, but I don't want to spend time on them if actually we won't
merge.

erikerlandson:
I'm still interested in this, however my interest has become mostly abstract.  I now think that free-monads may be sufficient for my purposes.  In any case, my project is completely experimental (and currently inactive) and I could just as easily include this prototype in my own code if I want to play with it some more.  I don't think my side projects are sufficient justification to merge something that there's ambivalence about supporting  :)

djspiewak:
@johnynek I'm tempted to just close this.  There doesn't seem to be that much interest in it, overall, which is unfortunate but understandable.  Opinion?

polkovnikov-ph:
@djspiewak Don't.

johnynek:
I think we can merge it, maybe with a bit of polish.

If we want to make it fully stack safe, I think we have to introduce a
stack safe function type, and use that as the continuation.

djspiewak:
I think a non stack safe variant is probably ok. I mean, the variant you've implemented (which is basically Free of Cont) is stack safe on binds, just not on suspends. Which is fine, because callcc is not something I would expect in a continuous loop.

PR #1403, opened 2016-10-11
fix homepage docs

pepegar:
Previously, _Getting Started_ and _Motivations_ headers didn't look
correctly because of the spacing.

# Before:
screen shot 2016-10-11 at 09 53 00

# After:
screen shot 2016-10-11 at 09 53 16

### DIFF
--- a/docs/src/site/index.md
+++ b/docs/src/site/index.md
@@ -16,6 +16,7 @@ The name is a playful shortening of the word *category*.
 
 ### <a name="getting-started" href="#getting-started"></a>Getting Started
 
+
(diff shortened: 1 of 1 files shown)


kailuowang:
👍

peterneyens:
Thanks @pepegar.

It seems we have the same problem in the _Efficiency_ section with _Project Structure_ and _Copyright and License_. Would you like to fix those as well in this PR ?

pepegar:
Sure!

peterneyens:
Thanks for the quick addition. :+1:

PR #1404, opened 2016-10-11
small wording change to semigroup.md for clarity

wonk1132:
Small change to semigroup.md for clarity. This wording basically matches that of `semigroupK.md` now.


docs/src/main/tut/semigroup.md
```diff
@@ -83,9 +83,9 @@ two |+| n
 ```
 
 You'll notice that instead of declaring `one` as `Some(1)`, I chose
-`Option(1)`, and I added an explicit type declaration for `n`. This is
-because there aren't type class instances for Some or None, but for
-Option. If we try to use Some and None, we'll get errors:
+`Option(1)`, and I added an explicit type declaration for `n`. This is because
+the  type class instance is defined for `Option`. There is no type class
+defined for `None` or `Some`.
```
kailuowang: you mean `There is no instances defined for` right?

wonk1132: ah right. my bad. I also further clarified and expanded this point.

docs/src/main/tut/semigroup.md
```diff
@@ -83,13 +83,38 @@ two |+| n
...
 You'll notice that instead of declaring `one` as `Some(1)`, I chose
-`Option(1)`, and I added an explicit type declaration for `n`. This is
-because there aren't type class instances for Some or None, but for
-Option. If we try to use Some and None, we'll get errors:
+`Option(1)`, and I added an explicit type declaration for `n`. This is because
+the type class instance for `Semigroup` is defined for `Option`. There is no `Semigroup`
+instance defined for `None` or `Some`--you will get errors if you attempt to combine a
+`Some` or `None` instance:
+
+```tut:scala
```
peterneyens: This should be `tut:book`. Changing this and the same below, should make the build pass.

wonk1132: ah ok. fixed.

peterneyens:
You can drop the `>` from the tut sections, and you could probably use `tut:book:fail` for the one which doesn't compile with `Some` and `None`.

For more info see the [tut project](https://github.com/tpolecat/tut).

wonk1132:
ah! got it, thanks for the info. fixed.

wonk1132:
Thanks for your help @peterneyens. I'm not familiar with the approval process in this repo. Do doc changes require two +1's? is there anything else I need to do?

thx again

peterneyens:
It seems like this PR currently re-adds the file _docs/src/main/tut/semigroup.md_ (which was moved to the _typeclasses_ directory in #1369).

non:
These changes seem great!

I agree with @peterneyens -- it seems like a file got accidentally re-added. @wonk1132, do you think you can fix this?

wonk1132:
oops, yea, messed that up.

moved to /typeclasses

adelbertc:
Hey @wonk1132 , thanks for your PR! I recently took it upon myself to go around the Cats docs and start revamping them a bit. I just did it for `Semigroup` and `Monoid` and would appreciate it your feedback on it: https://github.com/typelevel/cats/pull/1442

I tried to incorporate your changes into the new version but also began wondering if these type inference gotchas belong in a specific type class tutorial like `Semigroup` or in a more general place like the [FAQ](https://github.com/typelevel/cats/blob/master/docs/src/main/tut/faq.md) since these gotchas aren't `Semigroup`-specific. What do you think?

wonk1132:
@adelbertc I think a general blurb in typeclasses.md about this typeclass implementation gotcha and adding it to the FAQ would good. I think that info should be linked to in the specific typeclass document itself as most folks will probably go straight to the .md they want to use without reading the general typeclass.md doc itself.

adelbertc:
@wonk1132 Hm I'm now thinking if we should have a separate page for "gotchas" in type classes - stuff like type inference, `Unapply`/SI-2712 come to mind, perhaps put in FAQ.

Would you mind adding this as a question under FAQ?

kailuowang:
@wonk1132 Thanks for your contribution. I am going to close this one as #1442 (in lieu of this) was merged.

Issue #1344, opened 2016-09-01
Add toNestedValidatedand and toNestedValidatedNel to XorT/EitherT


mkotsbak:
These would be similar to `XorT.toOption(implicit F: Functor[F]): OptionT[F, B]`

kailuowang:
`XorT` is removed since 0.7.0. Do you want this for the replacement `EitherT`?

mkotsbak:
Ah ok, yes please, and to XorT too if there are 0.7 minor releases.

conniec:
Hi @mkotsbak, do you mean something like this? 

```def toNestedValidated(implicit F: Functor[F]): Nested[F, Validated[A, ?], B] ``` ?

mkotsbak:
Yes

peterneyens:
Added in #1408.

PR #1408, opened 2016-10-17
Add toNestedValidated and toNestedValidatedNel

conniec:
Issue: https://github.com/typelevel/cats/issues/1344

This is my first time contributing and am fairly new to these types so any feedback welcome. Would also like to add some examples if the code looks about right.

### DIFF
--- a/core/src/main/scala/cats/data/EitherT.scala
+++ b/core/src/main/scala/cats/data/EitherT.scala
@@ -208,6 +208,23 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
    */
   def toNested: Nested[F, Either[A, ?], B] = Nested[F, Either[A, ?], B](value)
 
+  /**
+    * Transform this `EitherT[F, A, B]` into a `[[Nested]][F, Validated[A, ?], B]` or `[[Nested]][F, ValidatedNel[A, B]`.
+    * Example:
+    * {{{
+    * scala> import cats.data.{Validated, EitherT, Nested}
+    * scala> import cats.implicits._
+    * scala> val f: Int => String = i => (i*2).toString
+    * scala> val r1: EitherT[Option, String, Int => String] = EitherT.right(Some(f))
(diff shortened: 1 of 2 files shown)


non:
These seem good to me.

At a minimum we should have tests for both methods. It doesn't look like we have a tutorial for `EitherT` currently, but if we did have one adding these methods to it would be a good idea.

travisbrown:
:+1:, looks good to me with the tests.

peterneyens:
This looks good, thanks @conniec.

Maybe we can add the following example to `toNestedValidatedNel`, it shows the stereotypical example of error short circuiting versus error accumulation, but using `EitherT` and `Nested` / `ValidatedNel`. What do you think ?

```scala
import cats.data.{Nested, EitherT, ValidatedNel}
import cats.implicits._

type ErrorsOr[x] = ValidatedNel[String, x]
type OptionErrorsOr[x] = Nested[Option, ErrorsOr, x]

val ok = EitherT.right[Option, String, Int](Some(1))
val error = EitherT.left[Option, String, Int](Some("error"))

val okV: OptionErrorsOr[Int] = ok.toNestedValidatedNel
val errorV: OptionErrorsOr[Int] = error.toNestedValidatedNel

(okV |@| okV |@| okV).map(_ + _ + _)
// OptionErrorsOr[Int] = Nested(Some(Valid(3)))

(okV |@| errorV |@| errorV).map(_ + _ + _)
// OptionErrorsOr[Int] = Nested(Some(Invalid(NonEmptyList(error, error))))

(ok |@| error |@| error).map(_ + _ + _)
// EitherT[Option,String,Int] = EitherT(Some(Left(error)))
```

travisbrown:
@peterneyens Maybe we should open an issue for `EitherT` documentation and update the `toNestedValidatedNel` example based on what ends up happening there? I'd be happy to see this merged as-is for now.

peterneyens:
You're right, see #1425.

conniec:
Thanks for the help everyone!

PR #1413, opened 2016-10-19
Add MonadError instance for StateT.

### LINKED ISSUE #1397: Add `MonadError` instance to `StateT`
@non Proposal for ScalaCenter hackathon at Lambda World
If `MonadError[?[_],E]` is available for `F[_]`, then we can give a `MonadError[?[_],E]` instance for `StateT[F,S,?]`, for any F, S, E.

peterneyens:
@adelbertc is doing a lot of work on the MTL type classes, but before #1379 (or something similar) lands, we could add this instance for the time being.

Fixes #1397.


adelbertc:
👍

non:
Seems good to me. 👍

Issue #1030, opened 2016-05-13
lazy `foldM`


ceedubs:
`Foldable.foldM` is a really handy function. For example, the following code will sum the input numbers until the sum reaches 10. Once 10 is reached, it will return a `Left`, or if 10 is never reached, it will return a `Right` with the final sum.

```scala
(1 to 20).toStream.foldM[Long Xor ?, Long](0L){ (acc, a) =>
  val sum = acc + a
  if (sum >= 10) Xor.left(sum) else Xor.right(sum)}
```

Unfortunately, in its current form, it will eagerly evaluate the entire input stream, because it's implemented in terms of `foldLeft`. You can see this if you change `(1 to 20).toStream` to `Stream.continually(1)`.

The initial proposal of `foldM` in #356 was lazy in consuming the input list (good), but wasn't stack safe (bad).

I've played around with this a bit, and I've noticed that you can implement a `foldM` that is both lazy and stack-safe given an `Iterator[A]` and a `Monad` that implements something like `  def flatMapEval[A, B](fa: F[A])(f: A => Eval[F[B]]): Eval[F[B]]`. However, the only way you can implement `flatMapEval` in terms of `flatMap` is by calling `.value` on an `Eval`, so this `foldM` instance ends up not being stack-safe for types that don't override `flatMapEval`.

I suspect that the best answer to this issue is to define a more constrained `Monad` type and require it. It's possible that one with the `flatMapEval` definition I mentioned earlier would work. I suspect something like #616 might be a better solution.

ceedubs:
cc @eed3si9n @TomasMikula

eed3si9n:
> I suspect that the best answer to this issue is to define a more constrained Monad type and require it.

Sounds exciting. `LazyMonad`, `EvalFlatMap`, `EvalT`?

ceedubs:
@eed3si9n that's a good question. I'm currently re-reading http://functorial.com/stack-safety-for-free/index.pdf and thinking about whether/where `Eval` fits into the picture here. Especially since I just moved `free` back into its own module which I'm now hoping wasn't a mistake...

TomasMikula:
I wasn't able to come up with an implementation that is both stack-safe and able to terminate early. What is yours (given that you have `flatMapEval` at hand)?

I was only able to imagine one if the return type of `f` (the accumulation function) indicated that there is no need to process any more elements, for example by returning a `M[B] Xor M[B]`.

[...]

PR #1414, opened 2016-10-19
Lazy foldM for "Iterables"

peterneyens:
If I am understanding correctly, now that `FlatMap` has `tailRecM` we can revisit #1030.

I adapted an [old commit](https://github.com/ceedubs/cats/commit/4d08b603f86db440a5d0db5291fac1294f8f3524) by @ceedubs for a PR to add `MonadRec` by @TomasMikula.

### DIFF
--- a/core/src/main/scala/cats/Foldable.scala
+++ b/core/src/main/scala/cats/Foldable.scala
@@ -1,6 +1,5 @@
 package cats
 
-import cats.data.Xor
 import scala.collection.mutable
 import cats.instances.long._
 import simulacrum.typeclass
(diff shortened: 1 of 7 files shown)


TomasMikula:
Some structures, such as `Vector` or a balanced tree, allocate new memory in `tail`, but usually not in `Iterator.next`. Thus I believe that an implementation that uses an iterator instead of repeatedly calling `tail` can be significantly more efficient in some cases.

core/src/main/scala/cats/Foldable.scala
```diff
@@ -372,4 +378,12 @@ object Foldable {
       Eval.defer(if (it.hasNext) f(it.next, loop()) else lb)
     loop()
   }
+
+  def iteratorFoldM[M[_], A, B](it: Iterator[A], z: B)(f: (B, A) => M[B])(implicit M: Monad[M]): M[B] = {
```
kailuowang: How about we also mention here in the doc the incentive of this special implementation for `Iteratable`s?

non:
The code seems good to me. 👍 Thanks!

core/src/main/scala/cats/Foldable.scala
```diff
@@ -379,11 +379,11 @@ object Foldable {
...
   }
 
-  def iterableFoldM[M[_], A, B](fa: Iterable[A], z: B)(f: (B, A) => M[B])(implicit M: Monad[M]): M[B] = {
-    val go: ((B, Iterable[A])) => M[Either[(B, Iterable[A]), B]] = { case (b, fa) =>
-      if (fa.isEmpty) M.pure(Right(b))
-      else M.map(f(b, fa.head))(b1 => Left((b1, fa.tail)))
+  def iteratorFoldM[M[_], A, B](it: Iterator[A], z: B)(f: (B, A) => M[B])(implicit M: Monad[M]): M[B] = {
+    val go: ((B, Iterator[A])) => M[Either[(B, Iterator[A]), B]] = { case (b, it) =>
+      if (it.hasNext) M.map(f(b, it.next))(b1 => Left((b1, it)))
+      else M.pure(Right(b))
```
TomasMikula: Since `Iterator` is mutable, I would keep it private to the method (otherwise we lose referential transparency). I would keep the signature as you had before, and only use `Iterator` inside.

TomasMikula: Sorry if this was not clear from my previous comment.

TomasMikula:
Not sure what I did wrong that in this conversation it shows my newest comment as outdated. I will repeat here:

Since `Iterator` is mutable, I would keep it private to the method (otherwise we lose referential transparency). I would keep the signature as you had before, and only use `Iterator` inside.

Sorry if this was not clear from my previous comment.

non:
@TomasMikula in the past we've exposed other methods using `Iterator` to make it easier for third-parties to build their instances. We could have the method take something like `() => Iterator[A]` if that makes it seem better.

TomasMikula:
Well, in that case, ship it!

TomasMikula:
Another, perhaps premature, optimization would be to avoid allocating the tuple in each step by accessing the iterator from the outer scope instead of passing it along.

TomasMikula:
LGTM 👍

kailuowang:
👍  thanks very much!

johnynek:
Yeah, I'm concerned everything except List will be slower here.

PR #1415, opened 2016-10-20
Scala 2.12.0-RC2

travisbrown:
This PR includes @BennyHill's changes from #1382 together with some more recent version updates and two other things:

1. I've updated SBT to 0.13.13-RC3 and fixed the deprecated operators.
2. I've changed the docs for `Either` to explain the difference in `flatMap` between 2.12 and the syntax provided by Cats for 2.10 and 2.11. I've also changed the `tut` shed that was failing the build (because it no longer failed to compile on 2.12) to an ordinary `scala` shed.

In addition to the usual review, here's what I see that still needs to be done:

- [x] Publish a new version of catalysts and remove the `-SNAPSHOT` here. I've got a [PR](https://github.com/typelevel/catalysts/pull/7) open for this that I think is ready to go.
- [x] Decide what to do about `bench`. Right now it's just not included in the build, so that you have to run something like `sbt bench/test` if you want to confirm that it's not broken. We could use sbt-doge, or we could have some stuff in the Travis CI build to run that command on 2.10 and 2.11 only, or we could just leave it and wait for sbt-jmh to be updated for 2.12. Given that we don't publish `bench` and that it doesn't have any tests, my vote would be just to wait.
- [x] See about changing `fork in test` back to `true`. I'm not sure why it was changed in #1382, but it seems fine to me to change it back. @BennyHill, can you confirm?
- [x] Have someone who's more familiar with the Travis CI setup confirm that there's nothing else that needs to be done.


BennyHill:
> See about changing fork in test back to true

The clue is the original kept as a comment... so for sure it was an oversight and should be reverted if possible

.travis.yml
```diff
@@ -8,8 +8,17 @@ git:
   depth: 9999
 
 scala:
-- 2.10.6
-- 2.11.8
+  - 2.10.6
+  - 2.11.8
+  - 2.12.0-RC2
```
BennyHill: delete this line as it should be in the matrix

.travis.yml
```diff
@@ -8,8 +8,17 @@ git:
...
+
+jdk:
+  - oraclejdk7
+
+matrix:
+  include:
+  - scala: 2.12.0-RC1
```
BennyHill: And change this to RC2

BennyHill:
For bench, see this https://github.com/ktoso/sbt-jmh/issues/89 issue

BennyHill:
https://issues.scala-lang.org/browse/SI-9923 notes that the "Unused import" warning is fixed in paradise_2.12.0-RC2 so we should remove the fix setttings and methods from the build

BennyHill:
for https://issues.scala-lang.org/browse/SI-9931 we should action @sethtisue 's final comment:

>I would suggest you comment your fix, with a link to this ticket, as it isn't obvious why the type annotations would be necessary. Otherwise it seems fine to me. Except that perhaps you might prefer the "val c = d" style fix from my comment above?

BennyHill:
I'll look at the catalysts PR later today,

ktoso:
Hi there @travisbrown @BennyHill,
feel free to ping me directly on anything related sbt-jmh.
In this specific case you just need to upgrade to 0.2.16 and it'll just work. 
The plugin still uses "old scala" which is fine, because that's what sbt uses, and the extras library is scala version independent (pure java), so no need for cross releasing anything for 2.12 (I confirmed this by running a benchmark on 2.12.0-RC2 right now).

Happy hakking :)

BennyHill:
Thanks @ktoso ! I _think_ what has happened is that I pushed an early PR for 2.12-0-RC1 that had quite a few "hacks" to just get it to compile so that others could see what was what - and I think that was about the same time that you released the new jmh.

But I think we are getting there... :smile_cat:

ktoso:
👍  
Looking forward to cats for 2.12.0-RC2, thanks for working on it! :-)

travisbrown:
Thanks, @BennyHill! I've added a comment about SI-9931, re-enabled fatal warnings and forking in tests, added bench back, and fixed the Travis CI versions. Now I think we're just waiting on Catalysts.

travisbrown:
Thanks, @ktoso—everything looks fine with 0.2.16!

travisbrown:
Okay, just pushed the catalyst update (thanks, @BennyHill). Once this succeeds I'll rebase the branch to get rid of the merge commits and then I think it'll be ready for any last review and merging.

travisbrown:
Okay, tests are passing and I've just rebased, so this is ready for merging after review.

adelbertc:
LGTM! Thanks so much @travisbrown 👍

build.sbt
```diff
@@ -1,37 +1,21 @@
...
-              if e.label == "dependency" && e.child.exists(child => child.label == "groupId" && child.text == "org.scoverage") => Nil
-          case _ => Seq(node)
-
-        }
-
-      }).transform(node).head
-  }
+  coverageMinimum := 60,
+  coverageFailOnMinimum := false,
+  coverageHighlighting := scalaBinaryVersion.value != "2.10"
```
BennyHill: As per gitter, I would expect that this is no longer required

BennyHill:
I think there is a missing .configure(disableScoverage210Jvm) in laws

travisbrown:
Thanks, @BennyHill—you're right, I misunderstood what was temporary vs. permanent there. I've added back that line and it's passing locally.

johnynek:
👍

ktoso:
Great, cheers!

PR #1424, opened 2016-10-25
Make NonEmptyList covariant

edmundnoble:
In which `NonEmptyList` and `ValidationNel` are made covariant.

### DIFF
--- a/core/src/main/scala/cats/data/NonEmptyList.scala
+++ b/core/src/main/scala/cats/data/NonEmptyList.scala
@@ -12,7 +12,7 @@ import scala.collection.mutable.ListBuffer
  * A data type which represents a non empty list of A, with
  * single element (head) and optional structure (tail).
  */
-final case class NonEmptyList[A](head: A, tail: List[A]) {
+final case class NonEmptyList[+A](head: A, tail: List[A]) {
 
   /**
    * Return the head and tail into a single list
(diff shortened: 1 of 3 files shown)


johnynek:
I'm +1 on this. I don't really buy the variance anxiety.

That said, you can always do `.widen[T]` to use `Functor` to get it to the type you want, which should cast, which should basically be free.

I think the cats position on variance is something that deserves a doc. I actually enjoy variance, but that seems to not be a popular opinion. I know subtyping in general is not widely admired due to how it complicates type inference, but I don't know of the concrete problems we have with covariance on data types, and I'd like to see them.

edmundnoble:
I believe attaching variance to data types and not typeclasses is the current cats position; the thing is, variance composes much more easily than functors. Once you get more than one level deep, `Functor[F].compose(Functor[G]).widen(t)` starts to get much worse than just `t`.

ceedubs:
I'm pretty late to the show, but 👍 from me. I think that this makes it consistent with data types like `Validated`. It sounds like we have 👍 from @johnynek, but since variance is a touchy subject I'd maybe like to get an opinion from at least one other person. @typelevel/cats what do you think?

ceedubs:
Also, if we merge this, I think that we should do the same for `NonEmptyVector`.

johnynek:
@non what do you say? I'm still +1 on data types having variance.

non:
I'm 👍 on this. I think a position of *variance on data types is fine, but not type classes* sounds like something that could work and would be consistent.

If there are concrete problems folks can point to with variance on data types then I'm open to revisiting this position.

johnynek:
Note, I think we will need to merge master here since there may have been
some methods added to NonEmptyList since this was created.

edmundnoble:
Rebased, with `NonEmptyVector` added to the party.

johnynek:
👍

kailuowang:
👍

ceedubs:
👍 thanks, @edmundnoble!

PR #1426, opened 2016-10-25
Use friendly name in POM for published artifacts - fixes #1396

### LINKED ISSUE #1396: incorrect <name> entry in pom ("core" instead of "cats-core")
if I do `sbt coreJVM/publishLocal` and inspect the resulting `cats-core_2.11.pom` I see:

```xml
    <name>core</name>
```

instead of the expected `cats-core`.

this then confused dbuild in the context of the Scala 2.12 community build, since the macro-compat project (attn @milessabin) also has a `org.typelevel#core` and the duplication caused dbuild to complain.

I am not a pom/maven/ivy jockey at all and I have basically zero idea what the implications of the `<name></name>` part of a pom are. but it seems plausible to me that this could cause other sorts of tooling type problems in other contexts.

I didn't check to see if other subprojects are affected.

after `set name in coreJVM := "cats-core"` or with

```diff
-  .settings(moduleName := "cats-core")
+  .settings(moduleName := "cats-core", name := "cats-core")
```

a correct pom is generated. what the difference between `moduleName` and `name` is I have no idea

adelbertc:
cc @SethTisue @BennyHill @travisbrown

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -201,7 +201,7 @@ lazy val catsJS = project.in(file(".catsJS"))
 
 
 lazy val macros = crossProject.crossType(CrossType.Pure)
-  .settings(moduleName := "cats-macros")
+  .settings(moduleName := "cats-macros", name := "Cats macros")
   .settings(catsSettings:_*)
   .jsSettings(commonJsSettings:_*)
   .jvmSettings(commonJvmSettings:_*)
@@ -213,7 +213,7 @@ lazy val macrosJS = macros.js
 
 lazy val kernel = crossProject.crossType(CrossType.Pure)
   .in(file("kernel"))
-  .settings(moduleName := "cats-kernel")
+  .settings(moduleName := "cats-kernel", name := "Cats kernel")
   .settings(kernelSettings: _*)
   .settings(buildSettings: _*)
   .settings(publishSettings: _*)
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -274,7 +274,7 @@ lazy val lawsJS = laws.js
 
 lazy val free = crossProject.crossType(CrossType.Pure)
   .dependsOn(macros, core, tests % "test-internal -> test")
-  .settings(moduleName := "cats-free")
+  .settings(moduleName := "cats-free", name := "Cats Free")
```
travisbrown: Any reason "Free" is capitalized here (unlike the others)?

adelbertc: I just did it because it was named after the Free monad which I usually see capitalized. I can make it lower case if folks prefer that.

SethTisue:
let's see what the community build thinks: https://scala-ci.typesafe.com/job/scala-2.12.x-integrate-community-build/853/consoleFull (404 til Jenkins catches up)

build.sbt
travisbrown: Fair enough. I don't have a strong opinion either way (but would prefer lowercase).

SethTisue:
```
[cats] Resolving: https://github.com/adelbertc/cats.git#4522b6117574d24c41ce870a72ffc8257afd2e67 ...
...
[info] Project cats--------------------: SUCCESS (project rebuilt ok)
```

community build run is green, so this LGTM

adelbertc:
@SethTisue Awesome possum

@travisbrown Is that a 👍 from you or would you prefer I change the capitalization?

travisbrown:
Sure, as it is is fine. :+1:

non:
Seems reasonable to me. Thanks @adelbertc! :+1:

PR #1412, opened 2016-10-18
0.8.0 time!

travisbrown:
I'd like to get an 0.8.0 release out sooner rather than later (see discussion on [Gitter](https://gitter.im/typelevel/cats?at=58068e502189c68f42fb6df9)), so I thought I'd get the ball rolling with a first draft of the release notes.

I'll also start a checklist for things that need to happen before the release. I'm only including PRs that other people have expressed an interest in seeing in this release (e.g. #1382) or that look ready-to-go and non-controversial, but we can add others if there's other stuff people want in 0.8.

- [x] Make a decision on #1334 (rename `Bifunctor` to `Functor2`, etc.)
- [x] Merge #1417 (`stackUnsafeMonad` laws, has two :+1:s)
- [x] Merge #1416 (type class hierarchy diagram, two :+1:s)
- [x] Merge #1399 (instances for `UUID`, has two :+1:s)
- [x] Merge #1394 (`flatTraverse` and `flatSequence`, has two :+1:s)
- [x] Merge #1415 (Scala 2.12.0-RC2 support, three :+1:s)
- [x] Review and merge #1408 (`toNestedValidated` for `EitherT`, one :+1:)
- [x] Review and merge #1402 (more `Prod` instances, two 👍 s)
- [x] Review and merge #1411 (`Free.foldMap` and `compile`)
- [x] Add `Xor`-to-`Either` migration guide
- [x] Review and merge #1428 (minor clarification in FAQ about `Xor` and `Either`)
- [x] Review FAQ entry about required stack safety of `tailRecM` (#1427)
- [x] Finalize release notes

I'll take care of finalizing the release notes, and could take a stab at the migration guide and `tailRecM` notes, but someone else might be able to get to those two tasks before me.

PR #1427, opened 2016-10-25
Add FAQ entry for tailRecM

travisbrown:
This is a draft of a FAQ entry about the (newly mandatory) stack safety of `tailRecM`, as suggested by @kailuowang in #1412.

### DIFF
--- a/docs/src/main/tut/faq.md
+++ b/docs/src/main/tut/faq.md
@@ -139,16 +139,13 @@ More about the history of machinist and how it works can be discovered at the [p
 
 The `FlatMap` type class has a `tailRecM` method with the following signature:
 
-```tut:fail:silent
+```scala
 def tailRecM[A, B](a: A)(f: A => F[Either[A, B]]): F[B]
-}
 ```
 
 When you are defining a `FlatMap` instance, its `tailRecM` implementation must have two properties in order for the instance to be considered lawful. The first property is that `tailRecM` must return the same result that you would get if you recursively called `flatMap` until you got a `Right` value (assuming you had unlimited stack space—we'll get to that in a moment). In other words, it must give the same result as this implementation:
 
 ```tut:silent
-import cats.Monad
-
 trait Monad[F[_]] {
   def pure[A](x: A): F[A] = ???
   def flatMap[A, B](fa: F[A])(f: A => F[B]): F[B] = ???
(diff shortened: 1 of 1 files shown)


non:
Looks good to me. Thanks @travisbrown! 👍

kailuowang:
Looks great! Thanks! @travisbrown 👍

docs/src/main/tut/faq.md
```diff
@@ -134,6 +135,63 @@ Cats defines a wealth of type classes and type class instances. For a number of
 
 More about the history of machinist and how it works can be discovered at the [project page](https://github.com/typelevel/machinist), or [this article on the typelevel blog](http://typelevel.org/blog/2013/10/13/spires-ops-macros.html).
 
+## <a id="tailrecm" href="#tailrecm"></a>What is `tailRecM`?
+
+The `FlatMap` type class has a `tailRecM` method with the following signature:
+
+```tut:fail:silent
+def tailRecM[A, B](a: A)(f: A => F[Either[A, B]]): F[B]
+}
```
peterneyens: You have a rogue `}` here.

docs/src/main/tut/faq.md
```diff
@@ -134,6 +135,63 @@ Cats defines a wealth of type classes and type class instances. For a number of
 
 More about the history of machinist and how it works can be discovered at the [project page](https://github.com/typelevel/machinist), or [this article on the typelevel blog](http://typelevel.org/blog/2013/10/13/spires-ops-macros.html).
 
+## <a id="tailrecm" href="#tailrecm"></a>What is `tailRecM`?
+
+The `FlatMap` type class has a `tailRecM` method with the following signature:
+
+```tut:fail:silent
```
adelbertc: Should we just use `scala` modifier here since it's only failing because it's not implemented?

docs/src/main/tut/faq.md
adelbertc: Removveeeeeee

docs/src/main/tut/faq.md
```diff
@@ -134,6 +135,63 @@ Cats defines a wealth of type classes and type class instances. For a number of
...
+def tailRecM[A, B](a: A)(f: A => F[Either[A, B]]): F[B]
+}
+```
+
+When you are defining a `FlatMap` instance, its `tailRecM` implementation must have two properties in order for the instance to be considered lawful. The first property is that `tailRecM` must return the same result that you would get if you recursively called `flatMap` until you got a `Right` value (assuming you had unlimited stack space—we'll get to that in a moment). In other words, it must give the same result as this implementation:
+
+```tut:silent
+import cats.Monad
```
peterneyens: I don't think we need this import, right ?

travisbrown:
Thanks, @peterneyens, @adelbertc—fixed now.

adelbertc:
👍 thanks!

travisbrown:
I'm going to merge this since it's green, is docs, has the votes, and is going to trigger another looooong rebuild for #1430.

PR #1428, opened 2016-10-25
Clarify some details about the life and death of Xor

travisbrown:
These changes are minor, but to my ear "up to Cats 0.6.x" sounds like it excludes 0.6.x, and in any case `Xor` wasn't actually gone in 0.7. Since we're pointing to this FAQ entry in the 0.8.0 release / upgrade notes, I thought we should make the timing a little clearer. /cc @adelbertc


kailuowang:
👍

adelbertc:
👍 on Travis green

PR #1433, opened 2016-10-26
Rename interpret and replace ~>

travisbrown:
Not that I'm impatient or anything, but here are the changes suggested by @non in @johnynek's #1411, which is currently the last thing we've said we want for the 0.8.0 release.

I went with `FunctionK` instead of `~>` because there are a lot of `FunctionK`s in cats-free and no `~>`s outside of FreeT.scala.

### DIFF
--- a/free/src/main/scala/cats/free/FreeT.scala
+++ b/free/src/main/scala/cats/free/FreeT.scala
@@ -3,6 +3,8 @@ package free
 
 import scala.annotation.tailrec
 
+import cats.arrow.FunctionK
+
 /**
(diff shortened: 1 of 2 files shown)


non:
👍

johnynek:
👍

johnynek:
after this is merged, I'll update #1411

travisbrown:
@johnynek This PR's against that branch, so if you click merge that'll happen automatically. :)

johnynek:
overlooked that!

PR #1435, opened 2016-10-26
Fix broken documentation links

liff:
`/tut/` leads to a 404.

### DIFF
--- a/core/src/main/scala/cats/data/OptionT.scala
+++ b/core/src/main/scala/cats/data/OptionT.scala
@@ -10,7 +10,7 @@ import cats.syntax.either._
  *
  * It may also be said that `OptionT` is a monad transformer for `Option`.
  *
- * For more information, see the [[http://typelevel.org/cats/tut/optiont.html documentation]].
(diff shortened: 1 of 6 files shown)


peterneyens:
HI @liff, thanks for spotting this.

`grep "tut/" -r --exclude-dir "*target*" cats/` gives us a lot of similar errors, but I will open another issue for this.

liff:
I can add fixes for the other broken links in this PR if that's OK?

peterneyens:
Of course, that would be great.

liff:
Here you go.

There's one more [broken link to Arrow.scala](https://github.com/typelevel/cats/blame/master/docs/src/main/tut/faq.md#L198) in the FAQ but I'm not sure where it should point to exactly.

peterneyens:
I think we lost `{{ site.sources }}` with the new site (It was in this [removed file](https://github.com/typelevel/cats/pull/1369/files#diff-5834e29d7dd3bb9c2cbd367847a9d4ecL10)).

We can probably fix that in another PR.

non:
👍 Thanks for fixing this!

peterneyens:
:+1:

PR #1442, opened 2016-10-27
Revamp Semigroup doc

adelbertc:


docs/src/main/tut/typeclasses/monoid.md
```diff
@@ -7,70 +7,138 @@ scaladoc: "#cats.kernel.Monoid"
...
- 
 ```tut:book
-Monoid[Map[String,Int]].combineAll(List(Map("a" -> 1, "b" -> 2), Map("a" -> 3)))
-Monoid[Map[String,Int]].combineAll(List())
+Monoid[Int].combine(x, Monoid[Int].empty)
+
+Monoid[Int].combine(Monoid[Int].empty, x)
+```
+
+# Exploiting laws: associativity and identity
```
philwills: I don't really see how this sub-heading fits with the text under it.

docs/src/main/tut/typeclasses/monoid.md
```diff
@@ -7,70 +7,138 @@ scaladoc: "#cats.kernel.Monoid"
 ---
 # Monoid
 
-`Monoid` extends the [`Semigroup`](semigroup.html) type class, adding an 
-`empty` method to semigroup's `combine`. The `empty` method must return a 
-value that when combined with any other instance of that type returns the 
-other instance, i.e.
+`Monoid` extends the power of `Semigroup` by providing an additional `empty` value.
```
philwills: This reads much more cleanly. 👍

docs/src/main/tut/typeclasses/semigroup.md
```diff
@@ -7,92 +7,164 @@ scaladoc: "#cats.kernel.Semigroup"
...
+
+Semigroup[Int].combine(Semigroup[Int].combine(x, y), z)
+```
+
+Infix syntax is also available for types that have a `Semigroup` instance.
+
+```tut:book
+import cats.syntax.semigroup._
+
+1 |+| 2
```
philwills: Addition always seems like a poor motivating example for this syntax to me. Maybe `Option[Int]` might make the case better.

adelbertc: The issue I had with introducing the `Option[A]` `Semigroup` this early was there are some nuances to it with regards to how it handles `None`. Since the `Semigroup` instance is sort of incidental because it really is a `Monoid` instance, I opted to introduce it in the `Monoid` section instead.

 I could perhaps show an early `Map` example and say details to come later in this post. As I type that out I think that's a good idea, I'll go ahead and do that.

docs/src/main/tut/typeclasses/semigroup.md
```diff
@@ -7,92 +7,164 @@ scaladoc: "#cats.kernel.Semigroup"
...
-two |+| n
+val sumLeft = left.foldLeft(0)(_ |+| _)
+val sumRight = right.foldLeft(0)(_ |+| _)
+val result = sumLeft |+| sumRight
 ```
 
-You'll notice that instead of declaring `one` as `Some(1)`, I chose
-`Option(1)`, and I added an explicit type declaration for `n`. This is
-because there aren't type class instances for `Some` or `None`, but for
-`Option`. If we try to use `Some` and `None`, we'll get errors:
```
adelbertc: I think we're going to move these "gotchas" to a more generic location, being handled at this PR: https://github.com/typelevel/cats/pull/1404

docs/src/main/tut/typeclasses/monoid.md
adelbertc: Ah I had a different intent for this section initially but went with a different direction and forgot to change it, oops :-)

docs/src/main/tut/typeclasses/monoid.md
```diff
@@ -7,70 +7,138 @@ scaladoc: "#cats.kernel.Monoid"
...
 
-```tut:silent
-import cats._
-import cats.implicits._
+```
+combine(x, empty) = combine(empty, x) = x
 ```
 
-Examples.
+Many types that form a `Semigroup` also form a `Monoid`, such as `Int`s (with `0`) and `Strings` (with `""`).
```
longcao: Probably tangential to this PR, but I've read discussions on the variations of

`T`:
* has a
* is a
* forms a

`Typeclass[T]`, maybe this is something nice to have a sentence about? I don't care which phrase is used though (maybe a note that they're saying the same thing), and I'm guessing somewhere on http://typelevel.org/cats/typeclasses.html would be appropriate.

longcao: @adelbertc 👍

PR #1444, opened 2016-10-27
Remove `defaultTailRecM` from monad doc

guersam:
Replaced `defaultTailRecM` with a handwritten example.


adelbertc:
Hmm how was this passing tut before?

non:
@adelbertc Maybe it was imported? Anyway, LGTM 👍

kailuowang:
@non that method should've been already deleted a while ago. So it's a puzzle to me as well how tut passed...

guersam:
The example on the site is rendered as markdown, not scala and it seems related to the suspicious tut passing. http://typelevel.org/cats/typeclasses/monad.html#composition

PR #1445, opened 2016-10-28
Fix tut in monad doc

guersam:
Not sure of the underlying machinary, but it turned out that the missing closing brace has been causing mysterious tut passing found on #1444.


adelbertc:
I've noticed this in my own tut files from time to time.. hopefully this is the last of them. @tpolecat do you know what might cause the old file to still pass tut?

tpolecat:
@adelbertc yeah Cody discovered that a while ago and I haven't fixed it yet, obviously. https://github.com/tpolecat/tut/issues/65

adelbertc:
👍

PR #298, opened 2015-05-02
Added Reducible[_] type class.

non:
This commit does a lot of related things. Here's a list:

 1. Removes existing `foldRight`, renames `foldLazy` to `foldRight`
 2. Makes `Fold[_]` instances a bit more powerful to work with
 3. Removes `Comonad[OneAnd[?, F]]` instance.
 4. Add some documentation.
 5. Some whitespace and formatting changes in a few files.

These changes are all possibly contentious.

(1) was intended to address the confusing bestiary of various
fold methods: `foldLeft`, `foldRight`, and `foldLazy`. Scala's eager
`foldRight` method is mostly-useless, which is why it seemed
reasonable to just remove it and replace it with our version.
I would be open to a different name -- I became convinced
`foldLazy` was a bad name but maybe I'm wrong?

(2) adds the `.imap` and `.complete` methods to `Fold[_]`. In
particular, .complete is something that almost anyone working
with `Fold[_]` will probably want.

(3) addresses a problem I noticed awhile ago: the `Comonad`
instance we had for `OneAnd` only dealt with the `.head`, dropping
the `.tail` on the floor. This is definitely wrong, although
we don't have a law to encode this yet. I noticed it while
working to have `OneAnd` implement `Reducible`.

(4) just adds a bit of documentation that I felt was missing.

(5) is arguably bad, but I tried to clean up some files that
I happened to be editing. I am torn between wanting to leave
lines alone (so that git blame is more useful and to preserve
the original author's intentions) and standardizing on a more
approachable style.

Issue #145, opened 2015-02-09
Add something like Foldable1 / Traverse1 ?


refried:
These are nice.

non:
I think instead of `Foldable1` we should call it `Reducible` to parallel the relationship between `foldLeft` and `reduceLeft`.

ceedubs:
I definitely would like to see these type classes.

I do like that Reducible would draw a parallel between foldLeft and reduceLeft. However, in the grand scheme of things, I'm wondering if Foldable1 would be more consistent/clear. Assuming we would have Traverse1/Traverse, the relationship between Foldable1/Foldable might be more clear. We have already chosen to use MonoidK and SemigroupK instead of the scalaz names to express their relationships with Monoid/Semigroup (respectively).

Also people hopefully have a fear of std lib's reduceLeft since it blows up on empty collections ;)

non:
Fixed in #298.

PR #1446, opened 2016-10-28
Cofree comonad

edmundnoble:
Basic Cofree Comonad functionality, including a couple of instances. I tried to implement a few other things (`mapBranching(F ~> G)(Cofree[F, A]): Cofree[G, A]`, `decompile(F ~> W)(W[A]): Cofree[F, A]`) but I couldn't make them stack-safe. This implementation is inspired by the one in Scalaz.

### DIFF
--- a/free/src/main/scala/cats/free/Cofree.scala
+++ b/free/src/main/scala/cats/free/Cofree.scala
@@ -1,0 +1,151 @@
+package cats
+package free
+
+/**
+  * A free comonad for some branching functor `S`. Branching is done lazily using [[Eval]].
+  * A tree with data at the branches, as opposed to [[Free]] which is a tree with data at the leaves.
+  * Not an instruction set functor made into a program monad as in [[Free]], but an instruction set's outputs as a
+  * functor made into a tree of the possible worlds reachable using the instruction set.
+  *
+  * This Scala implementation of `Cofree` and its usages are derived from
+  * [[https://github.com/scalaz/scalaz/blob/series/7.3.x/core/src/main/scala/scalaz/Cofree.scala Scalaz's Cofree]],
+  * originally written by Rúnar Bjarnason.
+  */
+final case class Cofree[S[_], A](head: A, tail: Eval[S[Cofree[S, A]]]) {
+
+  /** Evaluates and returns the tail of the computation. */
+  def tailForced: S[Cofree[S, A]] = tail.value
+
+  /** Applies `f` to the head and `g` to the tail. */
+  def transform[B](f: A => B, g: Cofree[S, A] => Cofree[S, B])(implicit S: Functor[S]): Cofree[S, B] =
+    Cofree[S, B](f(head), tail.map(S.map(_)(g)))
+
+  /** Map over head and inner `S[_]` branches. */
+  def map[B](f: A => B)(implicit S: Functor[S]): Cofree[S, B] =
+    transform(f, _.map(f))
+
+  /** Transform the branching functor at the root of the Cofree tree. */
+  def mapBranchingRoot(nat: S ~> S)(implicit S: Functor[S]): Cofree[S, A] =
+    Cofree[S, A](head, tail.map(nat(_)))
+
+  /** Transform the branching functor, using the S functor to perform the recursion. */
+  def mapBranchingS[T[_]](nat: S ~> T)(implicit S: Functor[S]): Cofree[T, A] =
+    Cofree[T, A](head, tail.map(v => nat(S.map(v)(_.mapBranchingS(nat)))))
+
+  /** Transform the branching functor, using the T functor to perform the recursion. */
+  def mapBranchingT[T[_]](nat: S ~> T)(implicit T: Functor[T]): Cofree[T, A] =
+    Cofree[T, A](head, tail.map(v => T.map(nat(v))(_.mapBranchingT(nat))))
+
+  /** Map `f` over each subtree of the computation. */
+  def coflatMap[B](f: Cofree[S, A] => B)(implicit S: Functor[S]): Cofree[S, B] =
+    Cofree[S, B](f(this), tail.map(S.map(_)(_.coflatMap(f))))
+
+  /** Replace each node in the computation with the subtree from that node downwards */
+  def coflatten(implicit S: Functor[S]): Cofree[S, Cofree[S, A]] =
+    Cofree[S, Cofree[S, A]](this, tail.map(S.map(_)(_.coflatten)))
+
+  /** Alias for head. */
+  def extract: A = head
+
+  /** Evaluate just the tail. */
+  def forceTail: Cofree[S, A] =
+    Cofree[S, A](head, Eval.now(tail.value))
+
+  /** Evaluate the entire Cofree tree. */
+  def forceAll(implicit S: Functor[S]): Cofree[S, A] =
+    Cofree[S, A](head, Eval.now(tail.map(S.map(_)(_.forceAll)).value))
+
+}
+
+object Cofree extends CofreeInstances {
+
+  /** Cofree anamorphism, lazily evaluated. */
+  def unfold[F[_], A](a: A)(f: A => F[A])(implicit F: Functor[F]): Cofree[F, A] =
+    Cofree[F, A](a, Eval.later(F.map(f(a))(unfold(_)(f))))
+
+  /**
+    * A stack-safe algebraic recursive fold out of the cofree comonad.
+    */
+  def cata[F[_], A, B](cof: Cofree[F, A])(folder: (A, F[B]) => Eval[B])(implicit F: Traverse[F]): Eval[B] =
+    F.traverse(cof.tailForced)(cata(_)(folder)).flatMap(folder(cof.head, _))
+
+  /**
+    * A monadic recursive fold out of the cofree comonad into a monad which can express Eval's stack-safety.
+    */
+  def cataM[F[_], M[_], A, B](cof: Cofree[F, A])(folder: (A, F[B]) => M[B])(inclusion: Eval ~> M)(implicit F: Traverse[F], M: Monad[M]): M[B] = {
+    def loop(fr: Cofree[F, A]): Eval[M[B]] = {
+      val looped: M[F[B]] = F.traverse[M, Cofree[F, A], B](fr.tailForced)(fr => M.flatten(inclusion(Eval.defer(loop(fr)))))
+      val folded: M[B] = M.flatMap(looped)(fb => folder(fr.head, fb))
+      Eval.now(folded)
+    }
+    M.flatten(inclusion(loop(cof)))
+  }
+
+}
+
+sealed private[free] abstract class CofreeInstances2 {
+  implicit def catsReducibleForCofree[F[_] : Foldable]: Reducible[Cofree[F, ?]] =
+  new CofreeReducible[F] {
+    def F = implicitly
+  }
+}
+
+sealed private[free] abstract class CofreeInstances1 extends CofreeInstances2 {
+  implicit def catsTraverseForCofree[F[_] : Traverse]: Traverse[Cofree[F, ?]] =
+  new CofreeTraverse[F] {
+    def F = implicitly
+  }
+}
+
+sealed private[free] abstract class CofreeInstances extends CofreeInstances1 {
+  implicit def catsFreeComonadForCofree[S[_] : Functor]: Comonad[Cofree[S, ?]] = new CofreeComonad[S] {
+    def F = implicitly
+  }
+}
+
+private trait CofreeComonad[S[_]] extends Comonad[Cofree[S, ?]] {
+  implicit def F: Functor[S]
+
+  override final def extract[A](p: Cofree[S, A]): A = p.extract
+
+  override final def coflatMap[A, B](a: Cofree[S, A])(f: Cofree[S, A] => B): Cofree[S, B] = a.coflatMap(f)
+
+  override final def coflatten[A](a: Cofree[S, A]): Cofree[S, Cofree[S, A]] = a.coflatten
+
+  override final def map[A, B](a: Cofree[S, A])(f: A => B): Cofree[S, B] = a.map(f)
+}
+
+private trait CofreeReducible[F[_]] extends Reducible[Cofree[F, ?]] {
+  implicit def F: Foldable[F]
+
+  override final def foldMap[A, B](fa: Cofree[F, A])(f: A => B)(implicit M: Monoid[B]): B =
+    M.combine(f(fa.head), F.foldMap(fa.tailForced)(foldMap(_)(f)))
+
+  override final def foldRight[A, B](fa: Cofree[F, A], z: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B] =
+    f(fa.head, fa.tail.flatMap(F.foldRight(_, z)(foldRight(_, _)(f))))
+
+  override final def foldLeft[A, B](fa: Cofree[F, A], z: B)(f: (B, A) => B): B =
+    F.foldLeft(fa.tailForced, f(z, fa.head))((b, cof) => foldLeft(cof, b)(f))
+
+  override final def reduceLeftTo[A, B](fa: Cofree[F, A])(z: A => B)(f: (B, A) => B): B =
+    F.foldLeft(fa.tailForced, z(fa.head))((b, cof) => foldLeft(cof, b)(f))
+
+  override def reduceRightTo[A, B](fa: Cofree[F, A])(z: A => B)(f: (A, Eval[B]) => Eval[B]): Eval[B] = {
+    foldRight(fa, Eval.now((None: Option[B]))) {
+      case (l, e) => e.flatMap {
+        case None => Eval.now(Some(z(l)))
+        case Some(r) => f(l, Eval.now(r)).map(Some(_))
+      }
+    }.map(_.getOrElse(sys.error("reduceRightTo")))
+  }
+
+}
+
+private trait CofreeTraverse[F[_]] extends Traverse[Cofree[F, ?]] with CofreeReducible[F] with CofreeComonad[F] {
+  implicit def F: Traverse[F]
+
+  override final def traverse[G[_], A, B](fa: Cofree[F, A])(f: A => G[B])(implicit G: Applicative[G]): G[Cofree[F, B]] =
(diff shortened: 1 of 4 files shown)


free/src/main/scala/cats/free/Cofree.scala
```diff
@@ -0,0 +1,124 @@
+package cats
+package free
+
+/**
+  * A free comonad for some branching functor `S`. Branching is done lazily using Eval.
+  * A tree with data at the branches, as opposed to Free which is a tree with data at the leaves.
+  * Not an instruction set functor made into a program monad as in Free, but an instruction set's outputs as a
+  * functor made into a tree of the possible worlds reachable using the instruction set.
+  */
```
adelbertc: Was this derived/based on Scalaz's [Cofree](https://github.com/scalaz/scalaz/blob/series/7.3.x/core/src/main/scala/scalaz/Cofree.scala) or from scratch? If former could you add attribution to Scalaz, similar to what we have [here](https://github.com/typelevel/cats/blob/master/free/src/main/scala/cats/free/FreeT.scala#L14) ?

free/src/test/scala/cats/free/CofreeTests.scala
```diff
@@ -0,0 +1,118 @@
...
+class CofreeTests extends CatsSuite {
+
+  import CofreeTests._
+
+  implicit val iso1 = CartesianTests.Isomorphisms.invariant[Cofree[Option, ?]]
+  implicit val iso2 = CartesianTests.Isomorphisms.invariant[Cofree[List, ?]]
+
+  checkAll("Cofree[Option, ?]", ComonadTests[Cofree[Option, ?]].comonad[Int, Int, Int])
+  checkAll("Cofree[Option, ?]", FoldableTests[Cofree[Option, ?]].foldable[Int, Int])
+  checkAll("Cofree[Option, ?]", TraverseTests[Cofree[Option, ?]].traverse[Int, Int, Int, Int, Option, Option])
```
adelbertc: I think Traverse checks Foldable already

adelbertc: Check Serializability of the Traverse instance?

free/src/test/scala/cats/free/CofreeTests.scala
```diff
@@ -0,0 +1,118 @@
...
+  checkAll("Comonad[Cofree[Option, ?]]", SerializableTests.serializable(Comonad[Cofree[Option, ?]]))
+  checkAll("Cofree[List, ?]", ComonadTests[Cofree[List, ?]].comonad[Int, Int, Int])
+  checkAll("Cofree[List, ?]", FoldableTests[Cofree[List, ?]].foldable[Int, Int])
+  checkAll("Cofree[List, ?]", TraverseTests[Cofree[List, ?]].traverse[Int, Int, Int, Int, Option, Option])
```
adelbertc: Do we need to test these given the tests above? Seems the only difference is s/Option/List

edmundnoble: If we're comfy with not testing these, absolutely. I don't see any precedent in other tests in doing the same law checks on different instances, so I'll go ahead and remove.

free/src/test/scala/cats/free/CofreeTests.scala
edmundnoble: I don't need to check that the Traverse and Foldable instances are valid separately? They are defined separately. I will add that serializable check.

free/src/main/scala/cats/free/Cofree.scala
edmundnoble: Totally derived from the one at Scalaz. I did not know there was an example of such an attribution somewhere in the code to crib from, I'll add it in.

free/src/test/scala/cats/free/CofreeTests.scala
adelbertc: Are they? Where did you see that? Seem to be together https://github.com/typelevel/cats/blob/master/laws/src/main/scala/cats/laws/discipline/TraverseTests.scala#L9

edmundnoble:
Is this alright for inclusion or should it perhaps have more operators?

adelbertc:
Could you write some functions for the other functions, maybe like `unfold` and `unfoldStrategy` ?

Also looks like the `Foldable`-specific instance isn't being tested because the test is getting the `Foldable` through the `Traverse` test. You can see an example [here](https://github.com/typelevel/cats/blob/master/tests/src/test/scala/cats/tests/NestedTests.scala#L107) of how we test these kinds of "if `F[_]: TC` then `MT[F, ?]: TC` law checks.

Other than that, LGTM 👍 Thanks @edmundnoble !

free/src/main/scala/cats/free/Cofree.scala
```diff
@@ -0,0 +1,128 @@
...
+
+  /** Cofree anamorphism, with customizable evaluation strategy. */
+  def unfoldStrategy[F[_], A](a: A)(f: A => F[A],
+                                    strategy: (=> F[Cofree[F, A]]) => Eval[F[Cofree[F, A]]])(implicit F: Functor[F]): Cofree[F, A] =
+    Cofree[F, A](a, strategy(F.map(f(a))(unfold(_)(f))))
+
+}
+
+sealed abstract class CofreeInstances2 {
+  /** low priority `Foldable1` instance */
```
xuwei-k: This is a `Foldable`, not `Foldable1`. There is `scalaz.Foldable1`. but cats does not have `Foldable1`. 

https://github.com/typelevel/cats/issues/145

free/src/main/scala/cats/free/Cofree.scala
```diff
@@ -0,0 +1,128 @@
...
+  implicit def catsFoldableForCofree[F[_] : Foldable]: Foldable[Cofree[F, ?]] =
+  new CofreeFoldable[F] {
+    def F = implicitly
+  }
+}
+
+sealed abstract class CofreeInstances1 extends CofreeInstances2 {
+  /** low priority `Traverse1` instance */
```
xuwei-k: This is a `Traverse`, not `Traverse1`

free/src/main/scala/cats/free/Cofree.scala
adelbertc: @xuwei-k We have `Foldable1` but we call it `Reducible` https://github.com/typelevel/cats/pull/298. No `Traverse1` though like you said below.

dwhitney:
Bump. This looks interesting and it looks like @edmundnoble has addressed @adelbertc's concerns. Any reason not to merge?

free/src/main/scala/cats/free/Cofree.scala
```diff
@@ -0,0 +1,135 @@
...
+  /** Replace each node in the computation with the subtree from that node downwards */
+  def coflatten(implicit S: Functor[S]): Cofree[S, Cofree[S, A]] =
+    Cofree[S, Cofree[S, A]](this, tailEval.map(S.map(_)(_.coflatten)))
+
+  /** Alias for head. */
+  def extract: A = head
+
+  /** Evaluate just the tail. */
+  def forceTail: Cofree[S, A] =
+    Cofree[S, A](head, Eval.now(tailEval.value))
```
johnynek: this is not tested, can we add a test?

free/src/main/scala/cats/free/Cofree.scala
```diff
@@ -0,0 +1,135 @@
...
+
+  /** Evaluate the entire Cofree tree. */
+  def forceAll(implicit S: Functor[S]): Cofree[S, A] =
+    Cofree[S, A](head, Eval.now(tailEval.map(S.map(_)(_.forceAll)).value))
```
johnynek: this is not tested, can we add a test?

free/src/main/scala/cats/free/Cofree.scala
```diff
@@ -0,0 +1,135 @@
...
+  override final def foldRight[A, B](fa: Cofree[F, A], z: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B] =
+    f(fa.head, fa.tailEval.flatMap(F.foldRight(_, z)(foldRight(_, _)(f))))
+
+  override final def foldLeft[A, B](fa: Cofree[F, A], z: B)(f: (B, A) => B): B =
+    F.foldLeft(fa.tailForced, f(z, fa.head))((b, cof) => foldLeft(cof, b)(f))
+
+  override final def reduceLeftTo[A, B](fa: Cofree[F, A])(z: A => B)(f: (B, A) => B): B =
+    F.foldLeft(fa.tailForced, z(fa.head))((b, cof) => foldLeft(cof, b)(f))
+
+  override def reduceRightTo[A, B](fa: Cofree[F, A])(z: A => B)(f: (A, Eval[B]) => Eval[B]): Eval[B] = {
```
johnynek: why did you do this vs something like:

```scala
f(ta.head, fa.tailEval.map { reduceRightTo(_)(z)(f) })
```

free/src/test/scala/cats/free/CofreeTests.scala
```diff
@@ -0,0 +1,102 @@
...
+import org.scalacheck.{Arbitrary, Cogen, Gen}
+
+class CofreeTests extends CatsSuite {
+
+  import CofreeTests._
+
+  implicit val iso = CartesianTests.Isomorphisms.invariant[Cofree[Option, ?]]
+
+  checkAll("Cofree[Option, ?]", ComonadTests[Cofree[Option, ?]].comonad[Int, Int, Int])
+  locally {
```
adelbertc: woah what's this `locally` thing do?

kailuowang: http://stackoverflow.com/questions/3237727/what-does-predef-locally-do-and-how-is-it-different-from-predef-identity

free/src/main/scala/cats/free/Cofree.scala
edmundnoble: Try doing that yourself. Doesn't typecheck as far as I can see; you're expecting `tailEval` to hold an `Eval[Cofree[F, A]]` instead of an `Eval[F[Cofree[F, A]]]`.

edmundnoble:
I've added tests for the methods @johnynek pointed out and fixed the tests for `mapBranchingS`/`mapBranchingT`/`mapBranchingRoot`/`unfold` which weren't running (but looked to be).

free/src/main/scala/cats/free/Cofree.scala
```diff
@@ -0,0 +1,135 @@
...
+
+object Cofree extends CofreeInstances {
+
+  /** Cofree anamorphism, lazily evaluated. */
+  def unfold[F[_], A](a: A)(f: A => F[A])(implicit F: Functor[F]): Cofree[F, A] =
+    Cofree[F, A](a, Eval.later(F.map(f(a))(unfold(_)(f))))
+
+}
+
+sealed abstract class CofreeInstances2 {
```
peterneyens: Maybe you could make the `CofreeInstancesX` classes and the `CofreeComonad`, `CofreeReducible` and `CofreeTraversable` traits package private (`private[free]`) (mostly so they don't show up in the ScalaDoc) ?

edmundnoble: Very right. One more feature to add to the eventual pedantic-scala tool that will come out once scala.meta has a semantic API ;)

adelbertc:
LGTM 👍

free/src/main/scala/cats/free/Cofree.scala
```diff
@@ -0,0 +1,153 @@
...
+sealed private[free] abstract class CofreeInstances2 {
+  /** low priority `Reducible` instance */
+  implicit def catsReducibleForCofree[F[_] : Foldable]: Reducible[Cofree[F, ?]] =
+  new CofreeReducible[F] {
+    def F = implicitly
+  }
+}
+
+sealed private[free] abstract class CofreeInstances1 extends CofreeInstances2 {
+  /** low priority `Traverse` instance */
```
xuwei-k: I wrote the comments in scalaz.Cofree because there are two methos in scalaz.

- https://github.com/scalaz/scalaz/commit/a05ab2c0053d54b37474
- https://github.com/scalaz/scalaz/blob/v7.2.8/core/src/main/scala/scalaz/Cofree.scala#L128

```scala
  /** low priority `Traverse1` instance */
  implicit def cofreeTraverse[F[_]: Traverse]: Traverse1[Cofree[F, ?]] =
```

- https://github.com/scalaz/scalaz/blob/v7.2.8/core/src/main/scala/scalaz/Cofree.scala#L155

```scala
  /** high priority `Traverse1` instance. more efficient */
  implicit def cofreeTraverse1[F[_]: Traverse1]: Traverse1[Cofree[F, ?]] =
```

Cats have only one instance `Traverse[Cofree[F, ?]]`

adelbertc:
@xuwei-k Looks like @edmundnoble removed it - what do you think?

free/src/main/scala/cats/free/Cofree.scala
```diff
@@ -0,0 +1,151 @@
...
+  * A free comonad for some branching functor `S`. Branching is done lazily using Eval.
+  * A tree with data at the branches, as opposed to Free which is a tree with data at the leaves.
+  * Not an instruction set functor made into a program monad as in Free, but an instruction set's outputs as a
+  * functor made into a tree of the possible worlds reachable using the instruction set.
+  *
+  * This Scala implementation of `Cofree` and its usages are derived from
+  * [[https://github.com/scalaz/scalaz/blob/series/7.3.x/core/src/main/scala/scalaz/Cofree.scala Scalaz's Cofree]],
+  * originally written by Rúnar Bjarnason.
+  */
+final case class Cofree[S[_], A](head: A, tailEval: Eval[S[Cofree[S, A]]]) {
```
travisbrown: Could we just call this `tail`? In general it seems like the convention is to add the suffix only when it's needed for disambiguation (although afaik that's not documented anywhere).

free/src/main/scala/cats/free/Cofree.scala
```diff
@@ -0,0 +1,151 @@
+package cats
+package free
+
+/**
+  * A free comonad for some branching functor `S`. Branching is done lazily using Eval.
```
travisbrown: This is super nitpicky but could you put `[[]]` around `Eval` and the `Free` in the next line? When I'm reading Scaladocs I always get pissed off if the authors make me search to follow a reference to another class.

free/src/main/scala/cats/free/Cofree.scala
```diff
@@ -0,0 +1,151 @@
...
+    Cofree[S, A](head, Eval.now(tailEval.map(S.map(_)(_.forceAll)).value))
+
+}
+
+object Cofree extends CofreeInstances {
+
+  /** Cofree anamorphism, lazily evaluated. */
+  def unfold[F[_], A](a: A)(f: A => F[A])(implicit F: Functor[F]): Cofree[F, A] =
+    Cofree[F, A](a, Eval.later(F.map(f(a))(unfold(_)(f))))
+
```
tpolecat: It would be nice to have the monadic ana as well.

```scala
def unfoldM[M[_]: Monad, F[_]: Traverse, A](a: A)(f: A => M[F[A]]): M[Cofree[F, A]]
```

free/src/main/scala/cats/free/Cofree.scala
```diff
@@ -0,0 +1,151 @@
...
+  /**
+    * A stack-safe algebraic recursive fold out of the cofree comonad.
+    */
+  def cata[F[_], A, B](cof: Cofree[F, A])(folder: (A, F[B]) => Eval[B])(implicit F: Traverse[F]): Eval[B] =
```
tpolecat: I am unhappy that `Eval` forces `Traverse[F]` when the strict version would only need `Functor`.

free/src/main/scala/cats/free/Cofree.scala
edmundnoble: Yeppers.

free/src/main/scala/cats/free/Cofree.scala
edmundnoble: The suffix is there to disambiguate it with `tailForced`, but I think it's reasonable for `tail` to be the safe version.

free/src/main/scala/cats/free/Cofree.scala
edmundnoble: Sure :)

free/src/main/scala/cats/free/Cofree.scala
edmundnoble: Let the record show that I am also made unhappy by this.

edmundnoble:
All cleaned up @tpolecat and @travisbrown, thanks for the reviews :)

free/src/main/scala/cats/free/Cofree.scala
```diff
@@ -0,0 +1,162 @@
...
+}
+
+object Cofree extends CofreeInstances {
+
+  /** Cofree anamorphism, lazily evaluated. */
+  def unfold[F[_], A](a: A)(f: A => F[A])(implicit F: Functor[F]): Cofree[F, A] =
+    Cofree[F, A](a, Eval.later(F.map(f(a))(unfold(_)(f))))
+
+  /** Cofree monadic anamorphism, lazily evaluated. */
+  def unfoldM[F[_], M[_], A](a: A)(f: A => M[F[A]])(implicit F: Traverse[F], M: Monad[M]): M[Cofree[F, A]] = {
```
tpolecat: Can you convince me this is stacksafe? I don't quite see it.

tpolecat: (Corrected: I meant to say lazy.)

edmundnoble: I cannot for the life of me figure out how to make it lazy. I have removed it for now just because it's unsafe otherwise, but I'd like to add it back soon.

dwhitney:
unfoldM is correct ;)

free/src/main/scala/cats/free/Cofree.scala
edmundnoble: :O

edmundnoble:
@dwhitney I tried running it and it looped forever eagerly. It's correct in that it produces the right answer, but it's incorrect in that it makes `Cofree`'s laziness useless.

peterneyens:
Thanks @edmundnoble !

Issue #1447, opened 2016-10-29
Validate.orElse behavior


oscar-stripe:
If we look at validated orElse:
https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/Validated.scala#L51

``` scala
  def orElse[EE, AA >: A](default: => Validated[EE, AA]): Validated[EE, AA] =
    this match {
      case v @ Valid(_) => v
      case Invalid(_) => default
    }
```

Unlike other uses of Validated, if both are Invalid, we only get the right here. It would be nice to have a combinator like:

``` scala
  def or[EE: Semigroup, AA >: A](default: => Validated[EE, AA]): Validated[EE, AA] =
    this match {
      case v @ Valid(_) => v
      case Invalid(e1) => default match {
        case v @ Valid(_) => v
        case Invalid(e2) => Invalid(Semigroup[EE].combine(e1, e2))
      }
    }
```

In fact, I would have thought that is what `orElse` does. It is a bummer to lose the left error in this case when both have errored.

non:
I support the `or` combinator.

I guess `orElse` is intended to be used for some kinds of error-recovery, where the left error isn't an error so much as a signal to try something else? I agree it's a bit weird (but does match the behavior of types like `Option`).

johnynek:
Well, I would say that the `or` above is also like `Option.orElse`: you get an error only if both are errors. The question is *what* error do you get?

I think it is strange that in the `orElse` case we use the `a + b == b` semigroup, but in the product case we use the `Semigroup[EE]`.

Like, if you want the `orElse` behavior, you could go to `Either` to get it. I thought the whole point of `Validated` was to accumulate as many errors as possible in the error case.

non:
So I have no idea what people in the world want. I can imagine situations where you do want this kind of monadic behavior (first check A, if it's fine you're done, otherwise check B ignoring A's failure).

I would be fine changing `orElse`, but I'd like to get wider commentary first, especially since this is a subtle change -- code will continue compiling but do something different. Introducing `or` is an easy way to support both behaviors (assuming people are relying on the current behavior, which I assume they are).

non:
(Also you probably want to update your code example for `or` to add the `Semigroup[EE]` implicit parameter.)

travisbrown:
@johnynek Isn't your `or` just `<+>`?

```scala
scala> import cats.data.{ Validated, ValidatedNel }
import cats.data.{Validated, ValidatedNel}

[...]

PR #1448, opened 2016-11-02
Add or combinator to Validated

mgzuber:


kailuowang:
This is being discussed in #1447

mgzuber:
Ah I see, should I close this PR then, or wait for a decision from that discussion and modify the PR accordingly?

kailuowang:
Your PR, your call. I just wanted to connect this PR and you to that discussion, in case you might want to take part in.

johnynek:
I'd rather this be the default behavior, but if others really think that will cause hard bugs I'm okay with `or`.

One issue: can we comment on each linking to the other? In the comment on orElse contrast it with or and likewise for the opposite?

mgzuber:
Yup, I added that.

travisbrown:
I'd also vote for accumulating `orElse` and a new name for the alternative. We can document it and make the change really prominent in the release notes, but I'd rather not keep around something we all seem to agree is broken just for the sake of consistency with Scalaz or our own (pre-1.0) versions.

non:
Maybe `orThen` for the *lossy* version?

mgzuber:
I think that's probably the right call, I've changed the code to reflect that.

johnynek:
👍

kailuowang:
How about adding some tests?

mgzuber:
Added a couple tests and cleaned up the doc on `orThen`!

travisbrown:
:+1: but I'd vote to hold off on merging this until 0.8.1 is published so that we can keep things simple by publishing off of master. (Even though this would be binary compatible it's a big enough change in behavior that I don't think it should happen in a patch release.)

johnynek:
+1 to not changing in a patch release.

That said, I do want to point out something here: if the `E = Int` and you were trying to count the number of failures, what would you want? In that case an or-style combinator may want to not combine (because you only had 1 failure in this case that blocked the success, you just had two path either of which were acceptable, that blocked).

So, if you wanted to count failures, you would use `.orThen` not `orElse`.

Something in the back of my mind makes me think `orElse` may actually be good as is because it really is `or`.

What we are talking about here is like `or` on the value side, but if that can't succeed, `and` on the error side. I'm starting to wonder if the name `orElse` is really appropriate for that.

I know, I pushed strongly for this earlier, and the motivation there was consistency: always use the semigroup when you can.

However, the error counting example seems to give me a bit of pause.

Maybe `.oneOrCombine`? Maybe too literal? Anyway, just wanted to share some second thoughts.

travisbrown:
@johnynek For me both the consistency and the expectation that `<+>` and `orElse` do the same thing (which is true for all other types that have both in Cats) are a strong enough argument for the `orElse` here.

Even in the case of error counting, I think the accumulating behavior at least arguably makes sense. You actually did try both sides (which may have included attempting to decode something, etc.), and if I'm using `Validated` I think I'd tend to expect that _all_ attempts are taken into account on the failure side. If I want just the failures on the path to a result, I'd use `Either`.

(Also I was wrong about this being binary compatible in my previous comment, since there's the new `Semigroup` requirement.)

johnynek:
@travisbrown good point. Consistency is a strong argument. +1.

travisbrown:
Hmm, I just ran into a fact that's making me question my original position (in favor of accumulating `orElse`).

In circe decoders can be run in either fail-fast or error-accumulating mode, and there's a requirement that simply dropping all errors except the first in error-accumulating mode should always give the same result as fail-fast mode.

This holds for all decoders defined in the library _except_ for ones built with `or` (which I recently changed to have the same behavior as the new `orElse` here).

I'm not sure this is much of an argument one way or the other, but it does feel a little suspicious.

johnynek:
I'm actually not sure `orElse` and `<+>` consistency is the right thing. For instance, if I defined:

```scala
class List[T] {
  def orElse[U >: T](that: List[U]): List[U] =
    if (nonEmpty) this
    else that
}
```
we would not want consistency with `<+>` and `orElse`. So, maybe we would say this is a bad method name, but presuming `orElse` == `<+>` is putting a strong semantic bias on what `<+>` means since `orElse` calls out to be an analog of `a | b` while `<+>` does not really.

mgzuber:
Is there a stronger consensus here? I can change what I've got if the opinion is that the `orElse` signature shouldn't change.

travisbrown:
@mgzuber At this point I don't feel strongly either way—I'm no longer convinced that either implementation is more consistent or elegant, so as long as they're both available I don't think the names matter all that much.

johnynek:
a very clear approach would be to add `def orBothInvalid[B >: A](that: Validated[E, B])(implicit sg: Semigroup[E]): Validated[E, B]`

or something like that. The name should make it clear that at least this one gives both invalids.

mgzuber:
@johnynek that would be in tune with the scalaz approach (the method name is `findSuccess`). I don't totally agree with it and I'm inclined to say that `orElse` seems more correct because it resembles what I would expect going from `Option` to `Validated`. 

That being said, something like `findValid` might make it easier for people to migrate over from scalaz to cats.

johnynek:
So, in the absence of a strong argument one way or the other, I think the fallback of API stability may win the day in my book.

So, I would say +1 to `findValid` to at least evoke scalaz.

mgzuber:
Done

kailuowang:
👍  Thanks very much!

johnynek:
👍

PR #1451, opened 2016-11-04
Fix Arbitrary instances in Free tests for ScalaCheck 1.13.3

travisbrown:
A couple of weeks ago [I noticed](https://gitter.im/typelevel/cats?at=58080b6d48292577613d18e9) that updating to ScalaCheck 1.13.3 was causing weird ScalaTest errors. I "fixed" this by not updating to ScalaCheck 1.13.3.

Staying on 1.13.2 is no longer an option now that we're updating to 2.12.0, and I've confirmed that the same thing happens with 1.13.4 and 2.12.0. 

The problem isn't actually ScalaTest calling `getSimpleName`—that's just making the error message uselessly obscure. Instead the problem seems to be some new bounds checking for `Gen.Choose` that was introduced in ScalaCheck 1.13.3. This PR fixes the issue by ensuring we don't call `Gen.chooseNum` with `min = 1` and `max = 0`.

This seems sensible but someone whose familiar with the logic of these `Arbitrary` instances should confirm that it doesn't mess things up.

Note that we can't update to 1.13.4 yet, since that's not published for 2.12.0-RC2. This PR will make that possible once Catalysts is ready, though.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -94,7 +94,7 @@ lazy val includeGeneratedSrc: Setting[_] = {
 
 lazy val catsSettings = buildSettings ++ commonSettings ++ publishSettings ++ scoverageSettings ++ javadocSettings
 
-lazy val scalaCheckVersion = "1.13.2"
+lazy val scalaCheckVersion = "1.13.3"
 lazy val scalaTestVersion = "3.0.0"
 lazy val disciplineVersion = "0.7.1"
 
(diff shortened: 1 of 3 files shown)


non:
This looks good to me. (And the bounds-checking exceptions were exactly to help people identify these issues, although it sounds like that experience wasn't fantastic.) 👍

travisbrown:
@non Well, the unpleasantness was entirely ScalaTest's fault… :smile:

adelbertc:
👍

ceedubs:
Thanks @travisbrown! I'm pretty sure that I introduced these bad generators.

PR #1452, opened 2016-11-05
Add `Foldable.foldMapM` and `Reducible.reduceMapM`

peterneyens:
This adds 

- `Foldable.foldMapM` which is similar to `foldM` but using a `Monoid` instance
- `Reducible.reduceMapM` which is similar to `reduceLeftM` but using a `Semigroup` instance.


johnynek:
:+1:

travisbrown:
👍, this looks great!

PR #1454, opened 2016-11-06
Either .left, .right syntax

edmundnoble:


adelbertc:
👍

johnynek:
Seems fine, but I'm slightly concerned that `Either.right` and `.left` which exist and are a bit different from this. I wonder if more verbose, yet clearer might be better. `.asLeft` or `.toLeft`?

edmundnoble:
@johnynek I can't see how `Either.right` and `.left` aren't the exact same as this, the same way `Xor.right` and `.left` were the same as their enriched syntax equivalents.

johnynek:
Either instances already have right/left methods. The object does not. So,
if we do: x.right.right the first and second mean different things. Also we
can't nest an either inside an either using .right since the enrichment
won't be triggered on the second call.

To me, this is less clear than it needs to be. .toRight or .toLeft would
resolve the ambiguity I think.

core/src/main/scala/cats/syntax/either.scala
```diff
@@ -306,6 +308,14 @@ final class RightOps[A, B](val right: Right[A, B]) extends AnyVal {
   def leftCast[C]: Either[C, B] = right.asInstanceOf[Either[C, B]]
 }
 
+final class EitherIdOps[A](val obj: A) extends AnyVal {
```
kailuowang: Let's make this `private val obj: A` , otherwise everybody will have an `obj` that point to itself.

edmundnoble: I'll do it to all the ops classes in here, for consistency.

edmundnoble:
@johnynek Sorry for the miscommunication, I was referring to the `Either.right` and `.left` methods we enriched on to the `Either` companion object and not the projection methods defined on the values themselves. From that angle, I can certainly see where you're coming from and think `asLeft` and `asRight` could work.

core/src/main/scala/cats/syntax/either.scala
edmundnoble: @kailuowang I checked and it looks like we don't use private vals in constructors for ops classes elsewhere in cats. Should I perhaps keep it not private for now for consistency, then submit a PR making all of them private?

kailuowang: But this is the only one that applies to all types. So I would favor a bit inconsistency over having a surprising obj on all types.

kailuowang: Ah, don't know how I missed that. I will submit a PR  to fix OptionIdOps as soon as I get back to my laptop. Thanks for pointing it out.  I still don't think being consistent with previous mistakes provides  much value here.

edmundnoble: You will also have to fix `ApplicativeIdOps`, `ValidatedIdSyntax`, and `WriterIdSyntax`. I don't think e.g. attaching an `oa` member to every `Option[A]` is particularly helpful either; are you sure we shouldn't make them all private?

edmundnoble: @kailuowang I have a PR (#1456) making the ops classes consistent in this regard.

johnynek:
:+1: Looks good to me!

core/src/main/scala/cats/syntax/either.scala
kailuowang: Thanks!

kailuowang:
👍 
Thanks very much!

PR #1456, opened 2016-11-06
Consistency for ops classes

edmundnoble:
1. Extend AnyVal when possible, unless there are *only* macro ops
1a. Don't take implicit params in the ops class param list, again unless
    there are macro ops, or they can't extend AnyVal.
(there is one exception to this and a comment explaining it; it's `EqSyntax`)
2. Make all conversions to ops classes final. They are all declared in
   traits, so users with their own syntax hierarchy have no chance of
   inlining unless they are marked final. Plus, it's consistent.
3. Make all ops class constructor fields private by default, to avoid
   polluting the namespace of the type.
Edit: 3 is not possible for Scala 2.10, so it was removed from this PR.

### DIFF
--- a/core/src/main/scala/cats/syntax/cartesian.scala
+++ b/core/src/main/scala/cats/syntax/cartesian.scala
@@ -2,7 +2,7 @@ package cats
 package syntax
 
 private[syntax] trait CartesianSyntax1 {
-  implicit def catsSyntaxUCartesian[FA](fa: FA)(implicit U: Unapply[Cartesian, FA]): CartesianOps[U.M, U.A] =
+  implicit final def catsSyntaxUCartesian[FA](fa: FA)(implicit U: Unapply[Cartesian, FA]): CartesianOps[U.M, U.A] =
     new CartesianOps[U.M, U.A] {
       val self = U.subst(fa)
       val typeClassInstance = U.TC
@@ -10,7 +10,7 @@ private[syntax] trait CartesianSyntax1 {
 }
 
 trait CartesianSyntax extends CartesianSyntax1 {
-  implicit def catsSyntaxCartesian[F[_], A](fa: F[A])(implicit F: Cartesian[F]): CartesianOps[F, A] =
+  implicit final def catsSyntaxCartesian[F[_], A](fa: F[A])(implicit F: Cartesian[F]): CartesianOps[F, A] =
     new CartesianOps[F, A] {
       val self = fa
       val typeClassInstance = F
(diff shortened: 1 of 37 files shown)


johnynek:
Can we add these style rules to a doc somewhere? Ideally, we could tool something like wart-remover to check them.

johnynek:
In spirit, I like this, but I do hope sooner or later we get serious about preserving some binary compatibility. This has a large binary incompatibility cross-section.

kailuowang:
@edmundnoble thanks very much for this! Sorry about raising the val in Ops class issue which probably wasted quite some time of yours.I suspected that 2.10 might be the cause for not making those private at the first place but didn't get the chance to verify. 
I agree there is value in documenting these style rules including issue that 2.10 preventing the self val to be private.

edmundnoble:
I added a lil something in this PR for the ops classes guidelines, while I was fixing the merge conflict. If it's better somewhere else I can take it right back out.

kailuowang:
LGTM, 👍  Thanks again

johnynek:
I added breaking change to this since it breaks binary compatibility (not source in most cases). I think that is appropriate.

ceedubs:
Given that we are merging other binary-incompatible changes, I think that I give this a 👍. Notably this is different than Simulacrum, which doesn't use value classes (if I recall correctly). @mpilquist do you think that value classes are appropriate in these cases?

@johnynek are you okay with us merging the binary-incompatible change? I've been MIA for a while, so to be honest I don't really know what current plans are on Cats releases/compatibility.

edmundnoble:
Rebased.

johnynek:
Two things:

1. binary compatibility is not binary (pun intended). The more incompatible changes we make, the greater the chance we cause an issue for a user.

2. we should investigate using `abstract class` much more than trait, since it is much friendlier for binary compatibility. For linear hierarchies which we often use for priorities, we could probably use more abstract classes.

core/src/main/scala/cats/syntax/cartesian.scala
```diff
@@ -2,27 +2,27 @@ package cats
...
 abstract class CartesianOps[F[_], A] extends Cartesian.Ops[F, A] {
-  def |@|[B](fb: F[B]): CartesianBuilder[F]#CartesianBuilder2[A, B] =
+  final def |@|[B](fb: F[B]): CartesianBuilder[F]#CartesianBuilder2[A, B] =
     new CartesianBuilder[F] |@| self |@| fb
 
-  def *>[B](fb: F[B])(implicit F: Functor[F]): F[B] = F.map(typeClassInstance.product(self, fb)) { case (a, b) => b }
+  final def *>[B](fb: F[B])(implicit F: Functor[F]): F[B] = F.map(typeClassInstance.product(self, fb)) { case (a, b) => b }
 
-  def <*[B](fb: F[B])(implicit F: Functor[F]): F[A] = F.map(typeClassInstance.product(self, fb)) { case (a, b) => a }
+  final def <*[B](fb: F[B])(implicit F: Functor[F]): F[A] = F.map(typeClassInstance.product(self, fb)) { case (a, b) => a }
```
johnynek: codecov says this method is untested.

core/src/main/scala/cats/syntax/monoid.scala
```diff
@@ -3,10 +3,10 @@ package syntax
...
   // TODO: use simulacrum instances eventually
-  implicit def catsSyntaxMonoid[A: Monoid](a: A): MonoidOps[A] =
+  implicit final def catsSyntaxMonoid[A: Monoid](a: A): MonoidOps[A] =
     new MonoidOps[A](a)
 }
 
-final class MonoidOps[A: Monoid](lhs: A) {
-  def isEmpty(implicit eq: Eq[A]): Boolean = Monoid[A].isEmpty(lhs)(eq)
+final class MonoidOps[A](val lhs: A) extends AnyVal {
+  def isEmpty(implicit A: Monoid[A], eq: Eq[A]): Boolean = A.isEmpty(lhs)(eq)
```
johnynek: this method is untested.

kailuowang:
@edmundnoble would you like to resolve conflicts and merge (it has two sign-offs)?

edmundnoble:
Gotcha. Will do today.

PR #1450, opened 2016-11-03
Build for Scala 2.12.0

adamnfish:
Hopefully this can be used to track 2.12 progress. The following dependencies are still missing builds for Scala 2.12, I've updated others where builds are now available.

* catalysts
* simulacrum
* spire

PR #1457, opened 2016-11-07
Update to Scala 2.12.0

travisbrown:
This PR includes @adamnfish's changes from #1450, together with the rest of the stuff we need for 2.12.0.

Note that because of [an issue](https://github.com/scoverage/sbt-scoverage/pull/201) in the 1.5.0 release of the sbt-scoverage plugin, it necessary to set the Scoverage compiler plugin version explicitly for the build:

```scala
coverageScalacPluginVersion := "1.3.0"
```

The issue should be fixed in the next sbt-scoverage release, and then it should be possible to remove that line.


kailuowang:
Looks like there is a problem with the catalysts release. I am working on it now.

travisbrown:
Thanks, @kailuowang! Looks like the missing artifact is there now, so hopefully this time it'll work.

mandubian:
👍

kailuowang:
:+1:

alexandru:
👍

adamnfish:
Thanks, @travisbrown!

gslowikowski:
`coverageScalacPluginVersion := "1.3.0"` setting can be removed from `build.sbt`.

travisbrown:
@gslowikowski How? Did you overwrite the 1.5.0 release on Maven Central? (Is that even possible?)

travisbrown:
@gslowikowski In any case since this only affects the build and is the same behavior you'd get without the line, I'd say let's get this PR merged asap and remove the line in a follow-up PR.

non:
👍

gslowikowski:
@travisbrown sure, I just wanted to let you know, that I fixed the snapshot problem before syncing to Maven central

travisbrown:
@gslowikowski This is getting off-topic for this PR (now that it's merged it probably doesn't matter), but did you figure out why the artifacts made it to Sonatype releases but hadn't been synced to MC after more than a day?

gslowikowski:
I didn't ask. I was happy, that because of this delay I could ask Sonatype team to break this buggy release. I've had similar problem in the past. That was something in their internals.

PR #1458, opened 2016-11-08
Add Validated.fromBoolean

markus1189:
This adds a new method `fromBoolean` to the `Validated` object.

The intention is that this allows an easy conversion of boolean conditions into a `Valid` or `Invalid`.

Essentially it is a shorthand for writing an `if` that returns a `Validated.valid(())` on success and otherwise some error.

Example use case:

```scala
case class Person(first: String, last: String)
def validatePerson(p: Person): ValidatedNel[String,Unit] = {
  Validated.fromBoolean(p.first.nonEmpty, "first name must not be empty").toValidatedNel |+|
    Validated.fromBoolean(p.last.nonEmpty, "last name msut not be empty").toValidatedNel
}
```

### DIFF
--- a/core/src/main/scala/cats/data/Validated.scala
+++ b/core/src/main/scala/cats/data/Validated.scala
@@ -220,7 +220,7 @@ sealed abstract class Validated[+E, +A] extends Product with Serializable {
     fold(_ => this, a => if (f(a)) this else Validated.invalid(onFailure))
 }
 
-object Validated extends ValidatedInstances with ValidatedFunctions{
+object Validated extends ValidatedInstances with ValidatedFunctions {
   final case class Valid[+A](a: A) extends Validated[Nothing, A]
   final case class Invalid[+E](e: E) extends Validated[E, Nothing]
 }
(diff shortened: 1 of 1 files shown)


johnynek:
👍

kailuowang:
I am not sure how I feel about fixing the right type to `Unit`. 
  Using your example `p.valid.ensure(_.first.nonEmpty)` gives you a `Validated[String, Person]`. If we want a `Unit`, `().valid.ensure(p.first.nonEmpty)` gives you exactly the thing you need. 
What do we gain from having `Validated.fromBoolean` over `().valid.ensure`?

markus1189:
Huh fair enough I didn't consider your example. Previously I used
`Validated.valid(()).ensure(???)` or an if which was quite cumbersome.
The reason for the unit on the right is that the semigroup instance with
unit on the right is handy if you only do validation and you can still use
`.as(???)` afterwards to fill it.

Now I am not quite sure if it is worth the addition, maybe because it is a
little easier to find than the version you suggested.

kailuowang:
if accumulating errors is your use case, this [PR](https://github.com/typelevel/cats/pull/1448/files) makes the `orElse` method congregating on the error side.
Then you can just do 
```Scala 
p.valid.ensure( "first present")(_.first.nonEmpty).toValidatedNel orElse
   p.valid.ensure("second present")(_.second.nonEmpty).toValidatedNel
```

markus1189:
Thanks for the suggestions. I'll try those tomorrow and then probably close
the pr, as it seems good enough to me. ;)

kailuowang:
Forgot to clarify that one is not merged yet.

markus1189:
I think I still prefer the `fromBoolean` approach, but I also see if you think it does not add enough value though.  I'll leave it up to you guys whether to merge or close.

PR #1464, opened 2016-11-13
Adding get for Foldable

yilinwei:


core/src/main/scala/cats/Foldable.scala
```diff
@@ -152,6 +152,26 @@ import simulacrum.typeclass
   def size[A](fa: F[A]): Long = foldMap(fa)(_ => 1)
 
   /**
+    * Get the element at the index of the `Foldable`.
+    *
+    * The default implementation of this is based on `foldLeft`, and thus will
+    * always fold across the entire structure.
+    */
+  def get[A](fa: F[A])(idx: Long): Option[A] =
```
edmundnoble: Is is not possible to implement this with `foldRight`?

johnynek: If we are going to add this, I think we need to implement the faster thing for Vector, NonEmpty*, List, etc...

core/src/main/scala/cats/Foldable.scala
```diff
@@ -152,6 +152,26 @@ import simulacrum.typeclass
...
+        case i if i == idx  => (i + 1, Some(a))
+        case i if i < idx => (i + 1, None)
+        case _ => b
+      }
+    }._2
+
+  /**
+    * Get the element at the index of the `Foldable`
+    */
+  def apply[A](fa: F[A])(idx: Long): A = get(fa)(idx).get
```
johnynek: Not crazy about adding methods like this without an `unsafe` warning. Something like `getUnsafe`? (I can't recall what our unsafe convention is: prefix or suffix).

core/src/main/scala/cats/instances/vector.scala
```diff
@@ -73,6 +73,8 @@ trait VectorInstances extends cats.kernel.instances.VectorInstances {
 
       override def size[A](fa: Vector[A]): Long = fa.size.toLong
 
+      override def get[A](fa: Vector[A])(idx: Long): Option[A] = if (fa.size > idx && idx > 0) Some(fa(idx.toInt)) else None
```
johnynek: Ahh, I see you have the `Vector` here, but not yet List (which can be made faster than a full fold) or NonEmpties.

core/src/main/scala/cats/Foldable.scala
yilinwei: I'm not quite sure that I agree with this, since the `apply`/`get` method is named the same as the scala collection methods. I quite like the behavior that in the particular case that we have a concrete type it would `defer` to that behavior.

yilinwei: I would agree for functions which aren't obviously unsafe - the `unsafeRun` in the `Stream` in fs2 comes to mind. However, for getting any index or key in a `Map` or `List` using `apply` in scala is already implied shorthand for `get.get`? 

Since I would argue that users should already be familiar with this behaviour, I can't see much benefit in calling it `getUnsafe`.

I don't feel particularly strongly enough to not change it, but I don't think to signal it's an unsafe method to users is a good enough reason.

core/src/main/scala/cats/instances/vector.scala
yilinwei: Added for `List` and `NonEmpties`

core/src/main/scala/cats/Foldable.scala
yilinwei: Just to be clear - I'll change the name since any change to this ought to change the rest of the names for the sake of consistency but I thought it would be good to raise these points.

jedws: Appealing to terrible decisions in the stdlib should not be an argument for a disciplined FP library where exceptions and other side-effects should be explicitly called out – and better, handled as values. The thing called `apply` should _not_ exist at all, even with an appropriately scary name. As pointed out, if someone really wants to they can call `.get.get`. A disciplined FP library should not provide shortcuts for bad behaviour.

yilinwei: @jedws I disagree with the assertion that only completely safe FP functions should exist in cats. If that were the case, then no unlawful instances (see `Seq`, `Try`) should exist in cats at all, and  cats ought to not dependent on the stdlib for _any_ of its collections because the functions on them such as `map` and `flatMap` are already a mess.

I also don't think it's necessarily bad behaviour to call `apply`; in the case that you know for a fact that a key exists at compile time, why should a user not call `get.get`? Is there any real point in adding extra barriers when the user is clearly aware that it's unsafe and calling it anyway?

adelbertc: I would actually prefer to not have `apply` and just have anyone who wants the behavior to call `.get.get`, as @jedws said. I would much prefer to see only safe FP functions exist in Cats.

Afaik we don't have instances for `Seq` - `Try` is a bit sketchy. I believe the argument was if you only use pure functions, then `Try` is lawful, it basically becomes a glorified `Either[Throwable, A]`.

non: @yilinwei I think `.get` is a good method to have.

I think I agree with others that `apply` is a bit harder to justify. If we have a concrete type (e.g. `Vector`) then the caller can use the vector's own `.apply` method to index into it. rather than needing to rely on `Foldable[F].apply`.

When dealing with a totally generic type `F[A]` with `Foldable[F]` I think it's harder to be sure that a given index will be present.

core/src/main/scala/cats/instances/list.scala
```diff
@@ -70,6 +70,16 @@ trait ListInstances extends cats.kernel.instances.ListInstances {
           G.map2Eval(f(a), lglb)(_ :: _)
         }.value
 
+      @tailrec
+      override def get[A](fa: List[A])(idx: Long): Option[A] =
+        if (idx < 0 || fa.isEmpty) {
+          None
+        } else if (idx < 0) {
```
johnynek: this should be `idx > 0` right?

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -216,6 +216,9 @@ private[data] sealed trait NonEmptyVectorInstances {
       override def foldRight[A, B](fa: NonEmptyVector[A], lb: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B] =
         fa.foldRight(lb)(f)
 
+      override def get[A](fa: NonEmptyVector[A])(idx: Long): Option[A] =
+        fa.get(idx.toInt)
```
johnynek: we might want to check that `idx` fits in a `Int` else return `None`, right?

non: In general I wonder if it's worth just using `Int` as the index. I can't think of many structures you can index into with `Long` values that you'd also want to fold over. (But that doesn't mean they don't exist!)

core/src/main/scala/cats/Foldable.scala
non: But I do agree with @yilinwei that we have some unsafe corners in cats (see `Monoid[Double]`) and I think we should be willing to consider things like that if they have sufficient usefulness to justify their danger.

jedws: I thoroughly agree that unlawful instances should not exist, and significantly complicate and reduce the utility of the library. If you must have unsafe corners you should (a) make them very obvious, and (b) make them difficult to use.

It is very disturbing that prior decisions to allow the existence of non-safety in a disciplined FP lib are being used as justification to add more. Is there a commitment to doing this right or not?

jedws: @non I realise, but the point is larger, there are quite a few other decisions such as unlawful instances that are complicating factors for users of the library. They are usually justified as being "handy" or "pragmatic" or some such, but they really don't have a place in a library that is aiming to be a foundational disciplined FP standard lib.

adelbertc: @jedws Can you point to specific ones in Cats? I would guess `Try` is in that set, which is sketchy in that it's existence was motivated by using `Throwable`, but if users just treat it as a glorified `A => Either[Throwable, A]` I think it is lawful. It certainly breaks in the presence of exceptions in that `Try`'s constructor captures those, but the counter-argument to that is at the point where you're throwing then all bets are off anyways. I agree though that the instance is sketchy.

I did a search for `unsafe` in the codebase and the only ones I see are on the `NonEmpty` stuff that lets you index into a `NonEmptyVector` unsafely. While I personally don't like it, it may there for performance reasons, maybe?

On a related note, for stuff like `Functor[Set]` we kick those out to a [separate lib](https://github.com/non/alleycats) for those who want to make that gamble.

jedws: @adelbertc I do not pay enough attention to have a definitive list. An obvious example would be the fact that _all_ monads are stacksafe:

https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/FlatMap.scala#L102

adelbertc: @jedws Ah there was quite a bit of talk on the location of that particular function during this past weekend at Scala by the Bay :-)

I'd like to avoid having such a discussion on this particular ticket, but if it's OK with you perhaps we can chat somewhere else about this? I'd be very interested in hearing your thoughts about it. I could ping you on IRC, or if you're on Gitter that works too.

core/src/main/scala/cats/data/NonEmptyVector.scala
yilinwei: I used a `Long` because the `size` is a `Long`. It seems strange that your collection can be a certain `size` without allowing you to get elements close to the end.

core/src/main/scala/cats/instances/list.scala
yilinwei: Yes! Will change this.

core/src/main/scala/cats/Foldable.scala
adelbertc: Going back to this line in particular, I would prefer not having this method, especially on something as abstract as `Foldable`.

yilinwei: @adelbertc I presume that you mean regardless of the name? Should `getUnsafe` be removed from corresponding classes for the sake of consistency?

core/src/main/scala/cats/instances/vector.scala
```diff
@@ -73,6 +73,8 @@ trait VectorInstances extends cats.kernel.instances.VectorInstances {
 
       override def size[A](fa: Vector[A]): Long = fa.size.toLong
 
+      override def get[A](fa: Vector[A])(idx: Long): Option[A] = if (idx < Int.MaxValue && fa.size > idx && idx >= 0) Some(fa(idx.toInt)) else None
```
johnynek: can you put a `\n` after `=` to put this code under `def get`?

core/src/main/scala/cats/data/NonEmptyVector.scala
edmundnoble: I don't think it makes sense to offer size as a long and get with an `idx: Int`. :+1: to checking if it fits if an `Int` and returning `None` otherwise.

edmundnoble:
@yilinwei Do you have time to address the code review comments? This is *very* useful.

core/src/main/scala/cats/Foldable.scala
edmundnoble: The implementation now uses `foldM` over `Either`, looks good to me. :+1:

yilinwei:
@edmundnoble Yes; I'll update/rebase this tomorrow.

core/src/main/scala/cats/instances/list.scala
```diff
@@ -70,6 +70,16 @@ trait ListInstances extends cats.kernel.instances.ListInstances {
           G.map2Eval(f(a), lglb)(_ :: _)
         }.value
 
+      @tailrec
+      override def get[A](fa: List[A])(idx: Long): Option[A] =
```
yilinwei: done.

edmundnoble:
Hey, sorry for dragging this out; could you also override `get` for all other `Foldable`/`Traverse` instances? `Option`, `Either`, `Try`, `Id`, `Set`, `Map`, `EitherKTraverse`, `OneAnd`, `Validated`, `Const`, `Ior`, `Stream`, `TupleN` are all the ones I can find.

yilinwei:
@edmundnoble I've done all except for `EitherKTraverse` and `ComposedTraverse` because I'd pretty much write the same thing as the current default implementation.

edmundnoble:
@yilinwei I thought about it some more, and `ComposedTraverse` is hard to override efficiently because to avoid making extra passes you have to `get` on the inner `G` *and* get the *size* of the `G` at once. `EitherKTraverse`'s though looks like:
```scala
  override def get[A](fa: EitherK[F, G, A])(idx: Long): Option[A] =
    fa.run.fold(F.get(_)(idx), G.get(_)(idx))
```

yilinwei:
@edmundnoble I arrived at the same conclusion for `Composed` but I'm ashamed to say I temporarily forgot how subclassing worked for `EitherK`...

ceedubs:
The 2.10 build had timed out, but after a restart it worked fine. I suspect that it won't be an issue with split out of the JS builds that has recently happened on `master`.

👍 from me. Thanks @yilinwei!

core/src/main/scala/cats/data/OneAnd.scala
```diff
@@ -117,6 +117,9 @@ private[data] sealed trait OneAndInstances extends OneAndLowPriority2 {
     new NonEmptyReducible[OneAnd[F, ?], F] {
       override def split[A](fa: OneAnd[F, A]): (A, F[A]) = (fa.head, fa.tail)
 
+      override def get[A](fa: OneAnd[F, A])(idx: Long): Option[A] =
+        if (idx == 0L) Some(fa.head) else F.get(fa.tail)(idx - 1L)
```
peterneyens: I think we can add this implementation to `NonEmptyReducible` (in _Reducible.scala_) as well,

core/src/main/scala/cats/data/Validated.scala
```diff
@@ -404,6 +404,9 @@ private[data] sealed abstract class ValidatedInstances2 {
       override def size[A](fa: Validated[E, A]): Long =
         fa.fold(_ => 0L, _ => 1L)
 
+      override def get[A](fa: Validated[E, A])(idx: Long): Option[A] =
+        if (idx == 0L) fa.fold(_ => None, Some(_)) else None
```
peterneyens: We can use `Validated#toOption`.

core/src/main/scala/cats/instances/try.scala
```diff
@@ -91,6 +91,13 @@ trait TryInstances extends TryInstances1 {
       override def reduceRightOption[A](fa: Try[A])(f: (A, Eval[A]) => Eval[A]): Eval[Option[A]] =
         Now(fa.toOption)
 
+      override def get[A](fa: Try[A])(idx: Long): Option[A] =
+        fa match {
+          case Failure(_) => None
+          case Success(a) =>
+            if (idx == 0L) Some(a) else None
+        }
```
peterneyens: Nitpicking: maybe like in `Validated` :

```scala
if (idx == 0L) fa.toOption else None
```

tests/src/test/scala/cats/tests/FoldableTests.scala
```diff
@@ -235,3 +241,44 @@ class FoldableValidatedCheck extends FoldableCheck[Validated[String, ?]]("valida
...
+
+class FoldableConstCheck extends FoldableCheck[Const[Int, ?]]("const") {
+  def iterator[T](const: Const[Int, T]) = None.iterator
+}
+
+class FoldableTuple2Check extends FoldableCheck[(Int, ?)]("tuple2") {
+  def iterator[T](tuple: (Int, T)) = Some(tuple._2).iterator
+}
+
+class FoldableOneAndCheck extends FoldableCheck[OneAnd[List, ?]]("oneAnd") {
```
peterneyens: This is sort of already checked through `ReducibleNonEmptyStreamCheck` in _OneAndTests.scala_

tests/src/test/scala/cats/tests/FoldableTests.scala
```diff
@@ -235,3 +241,44 @@ class FoldableValidatedCheck extends FoldableCheck[Validated[String, ?]]("valida
...
+  def iterator[T](oneAnd: OneAnd[List, T]) = (oneAnd.head :: oneAnd.tail).iterator
+}
+
+object FoldableComposedCheck {
+  implicit val composedListOptionFoldable: Foldable[λ[α => List[Option[α]]]] = Foldable[List].compose[Option]
+}
+
+import FoldableComposedCheck._
+
+class FoldableComposedCheck extends FoldableCheck[λ[α => List[Option[α]]]]("composed") {
```
peterneyens: We can maybe check this with `Nested[List, Option, ?]`, that way we don't need the `FoldableComposedCheck` companion object. WDYT?

core/src/main/scala/cats/data/Ior.scala
```diff
@@ -209,6 +209,11 @@ private[data] sealed abstract class IorInstances0 {
     def foldRight[B, C](fa: A Ior B, lc: Eval[C])(f: (B, Eval[C]) => Eval[C]): Eval[C] =
       fa.foldRight(lc)(f)
 
+    override def size[B](fa: A Ior B): Long = fa.fold(_ => 0L, _ => 1L, (_, _) => 1L)
+
+    override def get[B](fa: A Ior B)(idx: Long): Option[B] =
+      if (idx == 0L) fa.fold(_ => None, Some(_), (_, b) => Some(b)) else None
```
peterneyens: We can use `Ior#toOption` instead of the `fold`.

tests/src/test/scala/cats/tests/FoldableTests.scala
yilinwei: :+1: I thought this existed but couldn't remember the name of it.

tests/src/test/scala/cats/tests/FoldableTests.scala
yilinwei: I agree there is some duplication, but I don't think it should be tackled in this issue and should probably a different issue where we discuss what to do with tests for a typeclass which aren't part of the discipline laws.

core/src/main/scala/cats/data/OneAnd.scala
yilinwei: Done. I know there isn't a test currently; but I think that if such a test exists it should be comparing the default implementation to any overrides which exist.

Do you mind if we put that on a different issue because I think that the consistency of the methods is important.

core/src/main/scala/cats/Reducible.scala
```diff
@@ -226,6 +226,9 @@ abstract class NonEmptyReducible[F[_], G[_]](implicit G: Foldable[G]) extends Re
     1 + G.size(tail)
   }
 
+  override def get[A](fa: F[A])(idx: Long): Option[A] =
+    if (idx == 0L) Some(split(fa)._1) else G.get(split(fa)._2)(idx)
```
ceedubs: Shouldn't this be `else G.get(split(fa)._2)(idx - 1L)`?

It's a bit funny that the only line of your PR that isn't covered by a unit test is the one with a bug in it 😜

peterneyens:
Merging with 3 :+1:, thanks @yilinwei.

ceedubs:
@peterneyens @yilinwei oops I forgot to hit the "submit review" button after making this review comment earlier :(

core/src/main/scala/cats/Reducible.scala
yilinwei: :cry: You're right.

peterneyens: Shit, looked over that one too quickly as well. Thanks @ceedubs for paying attention.

PR #1466, opened 2016-11-15
Reduce stack depth in StateT

johnynek:
I hit some stack overflows using traverse with StateT. I don't think this will always solve the issue, but I tried to minimize stack depth in a number of cases.

### DIFF
--- a/core/src/main/scala/cats/data/StateT.scala
+++ b/core/src/main/scala/cats/data/StateT.scala
@@ -30,14 +30,27 @@ final class StateT[F[_], S, A](val runF: F[S => F[(S, A)]]) extends Serializable
   def map[B](f: A => B)(implicit F: Functor[F]): StateT[F, S, B] =
     transform { case (s, a) => (s, f(a)) }
 
-  def product[B](sb: StateT[F, S, B])(implicit F: FlatMap[F]): StateT[F, S, (A, B)] =
+  def map2[B, Z](sb: StateT[F, S, B])(fn: (A, B) => Z)(implicit F: FlatMap[F]): StateT[F, S, Z] =
     StateT.applyF(F.map2(runF, sb.runF) { (ssa, ssb) =>
       ssa.andThen { fsa =>
         F.flatMap(fsa) { case (s, a) =>
-          F.map(ssb(s)) { case (s, b) => (s, (a, b)) }
+          F.map(ssb(s)) { case (s, b) => (s, fn(a, b)) }
         }
       }
     })
+
+  def map2Eval[B, Z](sb: Eval[StateT[F, S, B]])(fn: (A, B) => Z)(implicit F: FlatMap[F]): Eval[StateT[F, S, Z]] =
+    F.map2Eval(runF, sb.map(_.runF)) { (ssa, ssb) =>
+      ssa.andThen { fsa =>
+        F.flatMap(fsa) { case (s, a) =>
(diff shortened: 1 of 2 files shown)


johnynek:
looks like a flakey test (Futures), restarting.

core/src/main/scala/cats/data/StateT.scala
```diff
@@ -70,9 +77,12 @@ final class StateT[F[_], S, A](val runF: F[S => F[(S, A)]]) extends Serializable
    * Like [[map]], but also allows the state (`S`) value to be modified.
    */
   def transform[B](f: (S, A) => (S, B))(implicit F: Monad[F]): StateT[F, S, B] =
```
johnynek: this only needs `Functor[F]` in this formulation.

core/src/main/scala/cats/data/StateT.scala
```diff
@@ -12,25 +12,32 @@ import cats.syntax.either._
 final class StateT[F[_], S, A](val runF: F[S => F[(S, A)]]) extends Serializable {
 
   def flatMap[B](fas: A => StateT[F, S, B])(implicit F: Monad[F]): StateT[F, S, B] =
```
johnynek: this only needs `FlatMap[F]` in this formulation, but I didn't want to change binary compatibility for a negligible gain.

core/src/main/scala/cats/data/StateT.scala
```diff
@@ -12,25 +12,32 @@ import cats.syntax.either._
...
+    StateT.applyF(F.map(runF) { sfsa =>
+      sfsa.andThen { fsa =>
+        F.flatMap(fsa) { case (s, a) =>
           fas(a).run(s)
         }
-      })
+      }
+    })
 
   def flatMapF[B](faf: A => F[B])(implicit F: Monad[F]): StateT[F, S, B] =
```
johnynek: only needs `FlatMap[F]` in the current formulation.

core/src/main/scala/cats/data/StateT.scala
```diff
@@ -12,25 +12,32 @@ import cats.syntax.either._
...
-          F.map(faf(a))((s, _))
-        }
+    StateT.applyF(F.map(runF) { sfsa =>
+      sfsa.andThen { fsa =>
+        F.flatMap(fsa) { case (s, a) => F.map(faf(a))((s, _)) }
       }
-    )
+    })
 
   def map[B](f: A => B)(implicit F: Monad[F]): StateT[F, S, B] =
```
johnynek: if we fix `transform` to only use `Functor[F]` then this only needs `Functor[F]`, which is fitting for `map`.

core/src/main/scala/cats/data/StateT.scala
non: I think that makes sense for now. Maybe add a comment to that effect? If we ever want to tighten things up later that will be handy.

non:
These changes seem positive to me.

I think we should either document all the places where type class bounds could be tightened, or just do the change now. While breaking compatibility is a drag, better now than at 1.0.

👍 whichever way you go.

johnynek:
Okay, I lowered everything about as much as I think you can. Nice to only need `Functor[F]` in many places, which makes also worthwhile to add a `Functor[StateT[F, S, ?]]` instance.

non:
👍

johnynek:
@adelbertc what do you think?

adelbertc:
LGTM 👍

PR #1468, opened 2016-11-16
Functor tupleLeft and tupleRight

edmundnoble:
`strengthL` and `strengthR` for `Functor`s. Tuple regular values with effectful ones.

### DIFF
--- a/core/src/main/scala/cats/Functor.scala
+++ b/core/src/main/scala/cats/Functor.scala
@@ -52,6 +52,16 @@ import simulacrum.typeclass
    */
   def as[A, B](fa: F[A], b: B): F[B] = map(fa)(_ => b)
 
+  /**
+    * Tuples the `A` value in `F[A]` with the supplied `B` value, with the `B` value on the left.
+    */
+  def tupleLeft[A, B](fa: F[A], b: B): F[(B, A)] = map(fa)(a => (b, a))
+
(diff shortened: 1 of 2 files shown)


adelbertc:
👍

peterneyens:
I have been thinking about adding these myself, but I wonder if we can't come up with some easier names (even though that's what they are called in scalaz and there seems to be a `strength` function in Haskell as well). Maybe something like `tupleLeft` / `tupleRight` ?

If nobody shares my concern, I am fine adding these as they stand.

edmundnoble:
@peterneyens I am inclined to agree, but I don't at all feel strongly.

johnynek:
.tuple for F[A], B to F[(A, B)] and .tupleSwap for the the other variant
would be my naming instinct, but I like non-insider names geared at
discoverability and and ease of remembering.

adelbertc:
Since we have `Cartesian#product` maybe `productLeft` and `productRight` ?

edmundnoble:
I've switched it around to `tupleLeft` and `tupleRight`, just because I don't think presuming an order of tupling is intuitive @johnynek and I don't want to confuse this with `Cartesian` ops because this requires `Functor` @adelbertc. I hope this is agreeable.

johnynek:
👍

PR #1470, opened 2016-11-19
Phantom of the functors

edmundnoble:
Phantom-variant functors! As @adelbertc has noted with his "functor diamond", phantom functors are what happen when a functor equivalently either:
1. Is both covariant and contravariant, or
2. Contains no values of the type param (e.g. `F[A]` has no `A` values)
This pattern comes up very frequently in user code dealing with phantom types.

The canonical example of a Phantom-variant functor is `Const[A, ?]`. Phantom-variant functors, though they are both covariant and contravariant, cannot express this fact exactly through our typeclass encoding because a) diamond inheritance is forbidden in simulacrum typeclasses and b) typeclass laws can only have one parent. That's why Phantom only extends `Contravariant` and provides a derivable covariant Functor, and only the `Contravariant` tests are run on Phantom functors.

This brought up a couple of issues in cats, notably:
1. The limits of our typeclass encoding, and
2. The only test for `Contravariant.narrow` tests that it is equivalent to `contramap(identity)`, but the `Contravariant` functor used is a phantom functor (`Const`). This means that actually the only test being performed is that `Contravariant.narrow` is equivalent to `contramap(f) forall f: A => B`. Not a giant issue, but effectively means we do not have real coverage for `Contravariant.narrow`.

This PR is not binary-compatible because it changes the name of the `Const` `Contravariant` instance, because it's now a `Phantom` instance. If that is an issue please let me know and I can add it separately in another `Instances` trait.

### DIFF
--- a/core/src/main/scala/cats/functor/Phantom.scala
+++ b/core/src/main/scala/cats/functor/Phantom.scala
@@ -1,0 +1,21 @@
+package cats
+package functor
+
+import simulacrum.typeclass
+
+/**
+  * Phantom-variant Functor.
+  * Values of type A must not appear in F[A].
+  * Natural result of a functor being both covariant and contravariant.
+  */
+@typeclass trait Phantom[F[_]] extends Contravariant[F] {
+  def functor: Functor[F] = new Functor[F] {
+    override def map[A, B](fa: F[A])(f: (A) => B): F[B] = pmap(fa)
+  }
+
+  def pmap[A, B](fa: F[A]): F[B]
+
+  override def imap[A, B](fa: F[A])(f: A => B)(g: B => A): F[B] = pmap[A, B](fa)
+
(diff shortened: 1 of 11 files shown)


core/src/main/scala/cats/functor/Phantom.scala
```diff
@@ -0,0 +1,21 @@
...
+
+import simulacrum.typeclass
+
+/**
+  * Phantom-variant Functor.
+  * Values of type A must not appear in F[A].
+  * Natural result of a functor being both covariant and contravariant.
+  */
+@typeclass trait Phantom[F[_]] extends Contravariant[F] {
+  def functor: Functor[F] = new Functor[F] {
```
johnynek: I don't follow why this can't extend `Functor[F]`. If the tests have limitations, I think we should fix those (I think other test classes have multiple parents).

I know we have thought about different encodings of typeclasses, including this style. I would rather us not have one lone typeclass following a different approach.

Issue #1471, opened 2016-11-20
A strange lambda in the FreeApplicative example


denisftw:
Some people pointed out that there is a strange lambda in the example of [FreeApplicative](http://typelevel.org/cats/datatypes/freeapplicative.html):

```scala
val compiler =
   λ[FunctionK[ValidationOp, FromString]] { fa =>
      str =>
        fa match {
          case Size(size) => str.size >= size
          case HasNumber  => str.exists(c => "0123456789".contains(c))
        }
   }
```

Not sure what it does here, but it doesn't compile. Shouldn't it be:

```scala
val compiler = new FunctionK[ValidationOp, FromString] {
      def apply[A](fa: ValidationOp[A]): FromString[A] = str =>
          fa match {
            case Size(size) => str.size >= size
            case HasNumber  => str.exists(c => "0123456789".contains(c))
          }
      }
```

A similar thing happens with other compilers.

peterneyens:
It uses the [kind projector](https://github.com/non/kind-projector) compiler plugin (more specifically its syntax for [polymorphic lambas](https://github.com/non/kind-projector/#polymorphic-lambda-values)).

It is good that you raise the issue, it was mentioned before that we should probably use the regular syntax in the documentation. We would love a PR with these changes, if you are interested.

denisftw:
Ah, right. For some reason, I though that kind projector only uses the question marks syntax.

Thanks!

peterneyens:
This seems to be a copy of #1449.

denisftw:
I suppose it's different. Rob is referring to the `~>` thing, which is just an alternative syntax for `FunctionK`. If I remember correctly, this type had alias `NaturalTranformation` several months ago, but then it was changed to `FunctionK`. My changes don't affect this, because `FunctionK` was already there.

ceedubs:
Resolved by #1472

PR #1472, opened 2016-11-21
Using regular syntax in the FreeApplicative tutorial (#1471)

denisftw:
As suggested in #1471, this PR adjusts examples in the FreeApplicative tutorial to use regular syntax instead of kind projector's lambdas.


peterneyens:
Thanks @denisftw :+1:

kailuowang:
I think it makes sense to use the regular syntax in the example. But [polymorphic lambdas](https://github.com/non/kind-projector/#polymorphic-lambda-values) was very handy in such situations. And if I remember correctly, such application in Free was one of the incentives of introducing polymorphic lambda in kind projector. We also use it in cats code. It might be worthwhile to mention polymorphic lambda here and use one of the `FunctionK` instance as an example? Something like, 
```
or with  [polymorphic lambdas](https://github.com/non/kind-projector/#polymorphic-lambda-values) provided by kind-projector, you can write compiler as 
 val compiler = λ[FunctionK[ValidationOp, FromString]] { fa => 
    ....
```
 @denisftw @peterneyens what do you guys think?

peterneyens:
It is indeed a good idea to mention the polymorphic lambda syntax and include such an example.

We could probably also add something about the polymorphic lambda syntax to the [section about kind projector](https://github.com/typelevel/cats/blob/master/docs/src/main/tut/faq.md#what-do-types-like--and-%CE%BB-mean) in the FAQ.

non:
Sounds good to me. 👍

johnynek:
This one fell through the cracks.

I'm +1 on what has been said:

1. for beginners, use the verbose syntax.
2. also explain that it could have been written using the compiler plugin with the lambda syntax.

ceedubs:
Thanks @denisftw! I really like this change. I'm sorry that it fell between the cracks.

I'm 👍 on merging this, and there are now `FunctionK` docs that describe using the polymorphic lambda syntax, so I don't think that we need to duplicate it in the `FreeApplicative` docs.

Unfortunately, there is a merge conflict. Would you want to resolve that? If not, I can take your work and turn it into a new PR.

denisftw:
Sure, I'll take a look tomorrow.

ceedubs:
Thanks @denisftw!

Issue #1473, opened 2016-11-23
stack unsafety in traverse


johnynek:
It seems that traverse is easy to blow the stack. Here is an example:

```scala
package cats

import cats.data.StateT
import implicits._

object TravTest {
  def test(size: Int) = {
    type E[T] = Either[String, T]
    type S[T] = StateT[E, Set[Int], T]
    val statev = (0 until size).toVector.traverse[S, Int] { i =>
      for {
        si <- StateT.pure[E, Set[Int], Int](i)
        _ <- StateT.set[E, Set[Int]](Set(i))
      } yield si
    }
    statev.run(Set.empty)
  }
}

```
By adding a `def traverseM` to applicative, we could side-step this because we could implement `traverseM` with `foldM` which can use tailRecM. For instance, in vector:

```scala
      override def traverseM[G[_], A, B](fa: Vector[A])(f: A => G[B])(implicit G: Monad[G]): G[Vector[B]] =
        foldM(fa, Vector.empty[B]) { (vec, a) => G.map(f(a))(vec :+ _) }

```
This is stack safe, and allows us to traverse to any size.

I'm trying to think of some other ways around this, but, especially for `StateT` it seems hard to make a sufficiently lazy `map2Eval` for this to work.

Interestingly, it seems that if you have `F[Eval[T]] => Eval[F[T]]` then you can solve the problem. This seems to keep coming up as a useful construct (it is also sufficient to implement a tailRecM).

johnynek:
cc @non

diesalbla:
I have just tested this code snippet using several versions of `cats`, in an Ammonite REPL. I run `TravTest.test(N)` with N being 1_000, 10_000, or 100_000

Using version 2.0.0, we still get a stack overflow. However, it seems that at least since version 2.2.0, this problem has been solved.

rossabaker:
Thanks, @diesalbla. For posterity, do we have any idea how it was fixed?  I don't see a linked ticket.

johnynek:
I fixed it in #3521

Forgot to link here. Sorry.

rossabaker:
:+1: Thanks for retesting, and thanks for digging up how it was fixed.

PR #1476, opened 2016-11-25
Add TraverseM for stack-safe traverse on collections

johnynek:
This is related to #1473 

The idea here is to make an optional method, `traverseM` that large structures can override which requires a Monad solely to have access to `tailRecM`.

This can make a traversal of for instance `StateT` across a large list actually work, which is nice.

An open question is should we just do this instead:
```scala
val a: Applicative[F] = ...

a match {
  case m: Monad[_] =>
    val mf = m.asInstanceOf[Monad[F]] // should be a safe cast, I think
    // use tailRecM in list, stream, map, vector, nonemptylist, nonemptyvector
  case other =>
    // use the map2 approach
}
```

The above will make more code just work because users don't modify their code at all, but it does have the cast, which I *think* is always safe. I don't think you can be a subclass of `Applicative[F]` and `Monad[G]` if `G != F`. I wish the scala compiler knew this, if true. Can anyone think of a counter example?

### DIFF
--- a/core/src/main/scala/cats/Composed.scala
+++ b/core/src/main/scala/cats/Composed.scala
@@ -69,6 +69,9 @@ private[cats] trait ComposedTraverse[F[_], G[_]] extends Traverse[λ[α => F[G[�
 
   override def traverse[H[_]: Applicative, A, B](fga: F[G[A]])(f: A => H[B]): H[F[G[B]]] =
     F.traverse(fga)(ga => G.traverse(ga)(f))
+
+  override def traverseM[H[_]: Monad, A, B](fga: F[G[A]])(f: A => H[B]): H[F[G[B]]] =
+    F.traverseM(fga)(ga => G.traverseM(ga)(f))
 }
 
 private[cats] trait ComposedTraverseFilter[F[_], G[_]] extends TraverseFilter[λ[α => F[G[α]]]] with ComposedTraverse[F, G] {
--- a/core/src/main/scala/cats/Traverse.scala
+++ b/core/src/main/scala/cats/Traverse.scala
@@ -33,6 +33,16 @@ import simulacrum.typeclass
   def traverse[G[_]: Applicative, A, B](fa: F[A])(f: A => G[B]): G[F[B]]
 
   /**
+   * This is exactly like traverse, except that it requires a Monad[G].
+   * It does this so that can use tailRecM to avoid stack overflows.
+   *
+   * combinators and potentially large data-structures should ideally
+   * override this.
+   */
+  def traverseM[G[_]: Monad, A, B](fa: F[A])(f: A => G[B]): G[F[B]] =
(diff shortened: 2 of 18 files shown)


non:
@johnynek I can confirm that it's impossible to extend `Monad[F]` and `Applicative[G]` unless `F` and `G` are the same type. I think the only reason why Scala can't prove this is that pattern matching has to worry about erasure, so the type parameters aren't known post-match.

non:
I'm 👍 on this method, but I also don't mind hiding the pattern match somewhere internal to give people extra stack-safety. I don't think we should encourage random casting, but up-casting an `Applicative` to a monad when possible to gain stack safety seems worth it.

johnynek:
yeah, I'm leaning to upcasting in the collections so we get the win without changing code. Will update PR.

johnynek:
okay, actually this does not work for list and vector. The code type-checks, but it does not agree with traverse. I don't yet know if this is salvageable or not (yet again giving lie to the notion that if it compiles, it's correct).

pchiusano:
@johnynek There's an approach to safely implementing sequence / traverse without this tailRecM stuff (which is pretty ad hoc and IMO doesn't belong as a requirement on all monads). Just do a balanced fold rather than a right fold to make stack depth logarithmic:

https://gist.github.com/pchiusano/7667597

Since traverse is "shape preserving", suspect it's even possible to have this work for all traversables by having them "go through" list. As in, convert to list, do sequencing there in stack safe way, then zip results back into the structure. Or something like that...

pchiusano:
Elaborating on this a bit - https://gist.github.com/pchiusano/d0bec7196e76c472c659d8037fabdafd

edmundnoble:
@pchiusano Are we sure that's stack-safe? Is there no practical example of it being impossible to build such a balanced tree, or it being too deep to be safe? Also, see this guy for inspiration: https://github.com/scalaz/scalaz/pull/1023.

johnynek:
@pchiusano added a comment: https://gist.github.com/pchiusano/d0bec7196e76c472c659d8037fabdafd#gistcomment-2036105

pchiusano:
Cool, I replied there...

pchiusano:
@edmundnoble on stack safety, yes. Even if you have as many elements in your list as there are atoms in the universe (10^80 or so), stack depth of balanced sequencing is only like 275 or something. :)

There may be other cases where `tailRecM` is really useful. Though I still think it belongs in a separate class.

johnynek:
Empirically we saw that without putting it on the main class, people tend to ignore it even when it could have easily been implemented, and then stack overflows become big pain points for people working with bigger inputs (I'm probably biased in that regard).

To me, while certainly not fundamental to Monads, tailRecM is a very important method on the JVM which is where almost all scala code runs.

kailuowang:

PR #1479, opened 2016-11-28
Hide traits in data package

peterneyens:
Hide
- traits for type class instances: `Coproduct`, `Func` and `Prod`.
- traits for functions: `EitherT`, `Ior`, `Validated` and `WriterT`.

See #1160.


non:
I'm guessing this is to make the Scaladoc a bit nicer?

Seems reasonable to me. 👍

PR #1480, opened 2016-11-30
Add FunctionK and

peterneyens:
Combine a `FunctionK[F, G]` and a `FunctionK[F, H]` to a `FunctionK[F, Prod[G, H, ?]]`.

This won't see as much use as `FunctionK#or`, but see the `FreeApplicative` tut example I changed for an example where it is useful.


non:
LGTM, thanks! 👍

PR #1481, opened 2016-11-30
Add EitherT.cond

andyscott:
Adds `EitherT.cond`, which behaves just like `Either.cond` but lifted into the applicative functor `F`.

I thought about calling this `fromBoolean`, but I wanted to be consistent with the existing cond method on Either.

### DIFF
--- a/core/src/main/scala/cats/data/EitherT.scala
+++ b/core/src/main/scala/cats/data/EitherT.scala
@@ -293,6 +293,28 @@ private[data] trait EitherTFunctions {
     def apply[E, A](opt: Option[A], ifNone: => E)(implicit F: Applicative[F]): EitherT[F, E, A] =
       EitherT(F.pure(Either.fromOption(opt, ifNone)))
   }
+
+  /**  If the condition is satisfied, return the given `A` in `Right`
+    *  lifted into the specified `Applicative`, otherwise, return the
+    *  given `E` in `Left` lifted into the specified `Applicative`.
+    *
+    * {{{
+    * scala> import cats.Id
+    * scala> import cats.data.EitherT
+    * scala> val userInput = "hello world"
+    * scala> EitherT.cond[Id](
+    *      |   userInput.forall(_.isDigit) && userInput.size == 10,
+    *      |   userInput,
+    *      |   "The input does not look like a phone number")
+    * res0: EitherT[Id, String, String] = EitherT(Left(The input does not look like a phone number))
+    * }}}
+    */
+  final def cond[F[_]]: CondPartiallyApplied[F] = new CondPartiallyApplied
+
+  final class CondPartiallyApplied[F[_]] private[EitherTFunctions] {
+    def apply[E, A](test: Boolean, right: => A, left: => E)(implicit F: Applicative[F]): EitherT[F, E, A] =
+      EitherT(F.pure(Either.cond(test, right, left)))
+  }
 }
 
 private[data] abstract class EitherTInstances extends EitherTInstances1 {
--- a/tests/src/test/scala/cats/tests/EitherTTests.scala
+++ b/tests/src/test/scala/cats/tests/EitherTTests.scala
@@ -156,6 +156,12 @@ class EitherTTests extends CatsSuite {
     }
   }
 
+  test("cond") {
+    forAll { (cond: Boolean, s: String, i: Int) =>
+      Either.cond(cond, s, i) should === (EitherT.cond[Id](cond, s, i).value)
+    }
+  }
+
   test("isLeft negation of isRight") {
     forAll { (eithert: EitherT[List, String, Int]) =>
       eithert.isLeft should === (eithert.isRight.map(! _))


kailuowang:
👍  thanks!

core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -293,6 +293,27 @@ private[data] trait EitherTFunctions {
     def apply[E, A](opt: Option[A], ifNone: => E)(implicit F: Applicative[F]): EitherT[F, E, A] =
       EitherT(F.pure(Either.fromOption(opt, ifNone)))
   }
+
+  /**  If the condition is satisfied, return the given `A` in `Right`
+    *  lifted into the specified `Applicative`, otherwise, return the
+    *  given `E` in `Left` lifted into the specified `Applicative`.
+    *
+    * {{{
+    * scala> val userInput = "hello world"
```
peterneyens: I think we need to import `cats.data.EitherT` and (for the `Applicative[Future]`) `cats.implicits._`, `scala.concurrent.Future` and an `ExecutionContext`. Newlines also need a `|` at the start of the line (see for example the `toNested` example).

It seems that this `cond` example wasn't tested by travis (if I haven't missed it), maybe because the format wasn't right ?

The easiest way I found to quickly test a doctest example is to paste it in a fresh `console` (otherwise it is easy to forget some of the imports).

Apart of the example :+1:

andyscott: I had the doc example code running correctly, but the travis build was spitting out errors that the imports weren't needed. I actually went back and removed the imports just for the sake of appeasing Travis!

I'll push a new commit to see what happens. If it all goes well, I can squash before a merge.

andyscott: I pushed a new commit, but when I run the tests locally I get:
```
> test:compileIncremental
[info] Compiling 26 Scala sources to /Users/andy/git/cats/core/.jvm/target/scala-2.12/test-classes...
[error] /Users/andy/git/cats/core/.jvm/target/scala-2.12/src_managed/test/cats/data/EitherTDoctest.scala:217: Unused import
[error]     import cats.data.EitherT
[error]                      ^
[error] /Users/andy/git/cats/core/.jvm/target/scala-2.12/src_managed/test/cats/data/EitherTDoctest.scala:219: Unused import
[error]     import cats.implicits._
[error]                           ^
[error] /Users/andy/git/cats/core/.jvm/target/scala-2.12/src_managed/test/cats/data/EitherTDoctest.scala:221: Unused import
[error]     import scala.concurrent.Future
[error] /Users/andy/git/cats/core/.jvm/target/scala-2.12/src_managed/test/cats/data/EitherTDoctest.scala:223: Unused import
[error]     import scala.concurrent.ExecutionContext.Implicits.global
[error]                                                        ^
[error] four errors found
[error] (coreJVM/test:compileIncremental) Compilation failed
[error] Total time: 12 s, completed Dec 6, 2016 9:09:21 AM
>
```

core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -299,13 +299,17 @@ private[data] trait EitherTFunctions {
...
-    *   userInput.forall(_.isDigit) && userInput.size == 10,
-    *   userInput,
-    *   s"The input does not look like a phone number")
-    * res0: EitherT[Future,String,String] =
-    *   EitherT(Future(Success(Left(The input hello world does not look like a phone number))))
+    *    | userInput.forall(_.isDigit) && userInput.size == 10,
+    *    | userInput,
+    *    | s"The input does not look like a phone number")
+    * res0: cats.data.EitherT[scala.concurrent.Future,String,String] =
+    *   EitherT(Future(Success(Left(The input does not look like a phone number))))
```
peterneyens: I think the new line here is possibly the reason.

This compiled for me : 

```scala
scala> import cats.data.EitherT
scala> import scala.concurrent.Future
scala> import scala.concurrent.ExecutionContext.Implicits.global
scala> val userInput = "hello world"
scala> EitherT.cond[Future](
     |   userInput.forall(_.isDigit) && userInput.size == 10,
     |   userInput,
     |   "The input does not look like a phone number")
res0: EitherT[Future,String,String] = EitherT(Future(Success(Left(The input does not look like a phone number))))
```

Otherwise the generated test was  : 

```scala
include(new org.scalacheck.Properties("cond") {
  import cats.data.EitherT

  import cats.implicits._

  import scala.concurrent.Future

  val userInput = "hello world"
})
```

Given how easy it seems for an example to not turn out exactly as we expect, I think the chances are pretty high that we currently have other examples/doctests which are not properly tested.

peterneyens:
Thanks, sorry for the sbt-doctest / example issues you had to solve. :+1:

andyscott:
No worries. It really was just all spacing issues. And then there was another issue with Futures failing when the scalacheck tests ran for the doctest-- so I just switched to `Id`.

PR #1483, opened 2016-12-04
enable mima binary compatibility check for cats-kernel

sritchie:
This wasn't running. I added the command to `validateJVM`.

Once we publish a 2.12 version we'll want to remove my switch that excludes mima for the 2.12 build - anything that changes should get caught in the meantime.

I'll also make a strong push for enabling this plugin for more of our modules, and being okay with minor version bumps if we do want to move fast. Life's tough downstream when you can't trust the version number.

### DIFF
--- a/scripts/travis-publish.sh
+++ b/scripts/travis-publish.sh
@@ -31,7 +31,7 @@ js="$sbt_cmd validateJS"
 kernel_js="$sbt_cmd validateKernelJS"
 free_js="$sbt_cmd validateFreeJS"
 jvm="$sbt_cmd coverage validateJVM coverageReport && codecov"
-  		  
+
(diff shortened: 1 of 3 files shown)


scripts/travis-publish.sh
```diff
@@ -31,7 +31,7 @@ js="$sbt_cmd validateJS"
 kernel_js="$sbt_cmd validateKernelJS"
 free_js="$sbt_cmd validateFreeJS"
 jvm="$sbt_cmd coverage validateJVM coverageReport && codecov"
-
```
sritchie: whitespace change from my editor

johnynek:
👍

johnynek:
this was there in the past, got removed because we wanted to fix something, but now that spire and algebird depend on kernel, we need to get this locked down and work around with binary compatible changes in the future (for kernel at least).

kailuowang:
👍

PR #1484, opened 2016-12-04
Show instance for scala.Symbol

durban:
This also makes sure, that cats.instances.AllInstances contains the Symbol instances.

BTW, there are other instances, which are not in AllInstances (and as a result, are not in scope after doing `import cats.implicits._`), e.g., Eq[BitSet]. Is this intentional?


johnynek:
👍 

I don't think any instances being missing from `AllInstances` is intentional. If so, we should add a comment to that effect (probably both at `AllInstances` and the instance in question).

kailuowang:
👍

PR #1487, opened 2016-12-08
Apply syntax for tuples

DavidGregory084:
Move apply syntax for tuples under `cats.syntax.apply._` (previously `cats.syntax.tuple._`). Methods are now suffixed by `N` rather than the tuple's arity:

```scala
(fa, fb).mapN { case (a, b) =>  }
(fa, fb, fc).contramapN { case (a, b, c) =>  } 
```

### DIFF
--- a/.jvmopts
+++ b/.jvmopts
@@ -1,4 +1,3 @@
-# see https://weblogs.java.net/blog/kcpeppe/archive/2013/12/11/case-study-jvm-hotspot-flags
 -Dfile.encoding=UTF8
 -Xms1G
 -Xmx6G
@@ -6,7 +5,5 @@
 -XX:ReservedCodeCacheSize=250M
 -XX:+TieredCompilation
 -XX:-UseGCOverheadLimit
-# effectively adds GC to Perm space
 -XX:+CMSClassUnloadingEnabled
-# must be enabled for CMSClassUnloadingEnabled to work
 -XX:+UseConcMarkSweepGC
--- a/core/src/main/scala/cats/data/EitherT.scala
+++ b/core/src/main/scala/cats/data/EitherT.scala
@@ -175,7 +175,7 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
    * scala> val v1: Validated[NonEmptyList[Error], Int] = Validated.Invalid(NonEmptyList.of("error 1"))
    * scala> val v2: Validated[NonEmptyList[Error], Int] = Validated.Invalid(NonEmptyList.of("error 2"))
    * scala> val eithert: EitherT[Option, Error, Int] = EitherT(Some(Either.left("error 3")))
-   * scala> eithert.withValidated { v3 => (v1 |@| v2 |@| v3.leftMap(NonEmptyList.of(_))).map{ case (i, j, k) => i + j + k } }
+   * scala> eithert.withValidated { v3 => (v1, v2, v3.leftMap(NonEmptyList.of(_))).mapN { case (i, j, k) => i + j + k } }
    * res0: EitherT[Option, NonEmptyList[Error], Int] = EitherT(Some(Left(NonEmptyList(error 1, error 2, error 3))))
    * }}}
    */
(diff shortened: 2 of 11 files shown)


kailuowang:
blocked by  #1073

peterneyens:
Now that we have enabled the partial unification fix (#1583) and dropped `Unapply` (#1679), I think we can move forward on this..

I would really like to see this in 1.0.0-MF. I think the sooner we add this, the better.

I think it might be nice to keep the cartesian builder (`|@|`) for at least until the next version (currently that probably will be 1.0.0-RC1), but deprecate it. Opinions?

Do you still have time to work on this @DavidGregory084? Otherwise I am happy to pick this up.

kailuowang:
@peterneyens I am leaning slightly towards removing the cartesian builder (|@|) in 1.0.0-MF. Mainly because that 1.0.0-MF is a "proposal" for the final API  in 1.0.0, it would be a bit weird to have a deprecated thing in the proposal. 
In the past, we have been taking the "let's break things sooner rather than later" route. Maybe we do the same this time?

peterneyens:
Makes sense. Just though that having a deprecation cycle before 1.0.0 might make things easier, especially for something like `|@|` which gets a lot of use.

Paging @johnynek who raised some upgrade/adoption concerns earlier in https://github.com/typelevel/cats/issues/1635#issuecomment-297208436.

.jvmopts
```diff
@@ -1,12 +1,9 @@
...
 -Xms1G
 -Xmx6G
 -XX:MaxPermSize=512M
 -XX:ReservedCodeCacheSize=250M
 -XX:+TieredCompilation
 -XX:-UseGCOverheadLimit
-# effectively adds GC to Perm space
 -XX:+CMSClassUnloadingEnabled
-# must be enabled for CMSClassUnloadingEnabled to work
 -XX:+UseConcMarkSweepGC
```
DavidGregory084: Comments in here cause `Could not find or load main class: #` on Windows

DavidGregory084:
@peterneyens @kailuowang Sure, I have updated the PR. I restored Cartesian builders, let's see how the discussion about deprecating them goes once everyone who's interested has weighed in.

kailuowang:
If we keep cartesian, should we mark them as deprecated and update the docs to use the new syntax?

kailuowang:
@DavidGregory084 I am just curious if you have time to continue on this, I think we reached a consensus that we just need to mark CartesianBuilder stuff as deprecated and update the docs to use new tupple syntax. Sorry for dragging so long and now trying to push it with release date past due.  I'll gladly complete the rest if you want.

DavidGregory084:
@kailuowang Ah sorry, I wasn't clear that we had reached a consensus. :) I should have some time to work on this, but probably not until the end of the week, so if you think you are going to have time earlier than that please feel free to take this on.

kailuowang:
@DavidGregory084 thanks so much! I was just checking if I can find something to work on. Will leave this one to you then.

kailuowang:
@DavidGregory084 still have the bandwidth to finish up this one?  I'll be more than happy to help if you don't.

DavidGregory084:
Thanks @kailuowang, I'm struggling a bit to find the time so it would be great if you could 👍

kailuowang:
no problem @DavidGregory084 , #1745 is submitted to finish up this one. Closing this for now.

johnynek:
I missed where we reached consensus to drop CartesianBuilder. Can we link to that here?

I think this is a bit unfortunate because I know of a few code bases using @ and this seems like a minor stylistic change that will put work on a lot of folks.

Removing Xor kept us on an old version of cats for a while at Stripe because the activation energy to go remove it everywhere was pretty high. I'm afraid this could be similar (though smaller in scope).

kailuowang:
@johnynek do you mind continue the discussion at #1745? I'll reply to you there.

gabro:
> I think this is a bit unfortunate because I know of a few code bases using @ and this seems like a minor stylistic change that will put work on a lot of folks.

Hey @johnynek I was going through the PRs that introduced breaking changes in 1.0.0 and I stumbled upon your comment. Did you have a change to test the [Scalafix rewrites](https://github.com/typelevel/cats/tree/master/scalafix)? They should solve (or at least mitigate greatly) that specific problem :)

PR #1489, opened 2016-12-08
String interpolator for Show

TomasMikula:
Before:

```scala
s"Foo ${bar.show} baz ${qux.show}"
```

After:

```scala
show"Foo $bar baz $qux"
```

### DIFF
--- a/core/src/main/scala/cats/Show.scala
+++ b/core/src/main/scala/cats/Show.scala
@@ -25,6 +25,13 @@ object Show {
     def show(a: A): String = a.toString
   }
 
+  def doShow[A](a: A)(implicit ev: Show[A]): String =
+    ev.show(a)
+
+  implicit class ShowInterpolation(val ctx: StringContext) extends AnyVal {
+    def show(args: Any*): String = macro cats.macros.ShowInterpolator.show
+  }
+
   implicit val catsContravariantForShow: Contravariant[Show] = new Contravariant[Show] {
     def contramap[A, B](fa: Show[A])(f: B => A): Show[B] =
       show[B](fa.show _ compose f)
--- a/tests/src/test/scala/cats/tests/ShowTests.scala
+++ b/tests/src/test/scala/cats/tests/ShowTests.scala
@@ -9,4 +9,32 @@ import cats.laws.discipline.eq._
 class ShowTests extends CatsSuite {
   checkAll("Contravariant[Show]", ContravariantTests[Show].contravariant[Int, Int, Int])
   checkAll("Contravariant[Show]", SerializableTests.serializable(Contravariant[Show]))
+
+  test("show string interpolator") {
+    import Show._
+
+    case class Cat(name: String)
+    object Cat {
+      implicit val showCat: Show[Cat] = new Show[Cat] {
+        def show(cat: Cat): String = cat.name
+      }
+    }
+
+    sealed trait TimeOfDay
+    case object Morning extends TimeOfDay
+    object TimeOfDay {
+      implicit val showTimeOfDay: Show[TimeOfDay] = new Show[TimeOfDay] {
+        def show(tod: TimeOfDay): String = tod match {
+          case Morning => "morning"
+        }
+      }
+    }
+
+    val tod: TimeOfDay = Morning
+    val cat = Cat("Whiskers")
+
+    val greeting = show"Good $tod, $cat!"
+
+    assertResult("Good morning, Whiskers!")(greeting)
+  }
 }
--- a/macros/src/main/scala/cats/macros/ShowInterpolator.scala
+++ b/macros/src/main/scala/cats/macros/ShowInterpolator.scala
@@ -1,0 +1,49 @@
+package cats
+package macros
+
+import scala.reflect.macros.blackbox.Context
+
+/**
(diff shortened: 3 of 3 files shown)


core/src/main/scala/cats/Show.scala
```diff
@@ -25,6 +25,13 @@ object Show {
     def show(a: A): String = a.toString
   }
 
+  def doShow[A](a: A)(implicit ev: Show[A]): String =
```
johnynek: I'm not crazy about "doShow". Sorry for the bike shed. Can we use callShow, showString, showOf or something? "do" sounds like an effect to me (maybe because of Haskell).

tests/src/test/scala/cats/tests/ShowTests.scala
```diff
@@ -9,4 +9,32 @@ import cats.laws.discipline.eq._
...
+      implicit val showTimeOfDay: Show[TimeOfDay] = new Show[TimeOfDay] {
+        def show(tod: TimeOfDay): String = tod match {
+          case Morning => "morning"
+        }
+      }
+    }
+
+    val tod: TimeOfDay = Morning
+    val cat = Cat("Whiskers")
+
```
johnynek: Can we exercise this with something like:

`show"${List(cat).head}"` to show we can use and logic inside braces as long as we return something showable.

tests/src/test/scala/cats/tests/ShowTests.scala
```diff
@@ -9,4 +9,32 @@ import cats.laws.discipline.eq._
...
+    val greeting = show"Good $tod, $cat!"
+
+    assertResult("Good morning, Whiskers!")(greeting)
```
johnynek: I think there is a "does not compile" macro we can use to test some negative examples. That might be good too.

travisbrown:
👍 from me, although yesterday I was talking to @dwijnand about this PR and he pointed out that it could be implemented more simply with an implicit conversion to a simple `Shown` string wrapper. I prefer the macro version over introducing an otherwise meaningless type and an implicit conversion, but the latter would be a lot more maintainable.

dwijnand:
The non-macro way is a lot less verbose, and I think composes a lot better than a macro approach.

I'm happy to propose it as an alternate implementation if there's any interest in it.

TomasMikula:
I sure am interested in seeing a non-macro way (I couldn't figure out one
myself).

core/src/main/scala/cats/Show.scala
TomasMikula: I'm not liking `doShow` either. I wanted to call it simply `show`, but that's taken for the smart constructor:

```scala
def show[A](f: A => String): Show[A]
```

An analogous smart constructor for `Eq` is called [`instance`](https://github.com/typelevel/cats/blob/65433840283da5513a6d3ea5cce082596d9246c9/kernel/src/main/scala/cats/kernel/Eq.scala#L86), so we would only make naming more consistent if we renamed current `show` to `instance` and free up the `show` name for `doShow`.

Otherwise, I could go with `callShow`.

tests/src/test/scala/cats/tests/ShowTests.scala
TomasMikula: Will do, but I will wait until the macro vs. non-macro discussion is resolved.

ceedubs:
In case anyone is interested, a long time ago I threw together a shapeless-based implementation of this: https://github.com/ceedubs/shapeless-cats/blob/master/core/src/main/scala/ShowInterpolation.scala

I think that this macro version (or potentially the alternate approach suggested by @dwijnand) is the way to go. But I just thought people might be interested in other approaches to the same problem if they like more dependencies and worse performance 😛. I guess it could potentially be an argument that this shouldn't be baked into Cats itself since there are several possible approaches. But at the end of the day, people are probably happier if it just works for them; I don't know.

dwijnand:
For those subscribed to this PR and not watching the whole repository, I've submitted a macro-less alternative: #1495.

dwijnand:
Btw @TomasMikula this failed due to macro API differences in Scala 2.10.

TomasMikula:
@dwijnand Yep, noticed that, but deferred investigating workarounds until I get some feedback.

It hasn't occurred to me that arguments to `ShowInterpolation(ctx: StringContext).show(args*)` can be other than `Any*`. I now prefer your macro-free solution.

TomasMikula:
@ceedubs Your solutions is also pretty neat!

> I guess it could potentially be an argument that this shouldn't be baked into Cats itself since there are several possible approaches.

I think on the use site one doesn't really care about the implementation.

TomasMikula:
For who might be interested, I'm also [inquiring](https://gitter.im/propensive/contextual?at=584b0af7bb7d528222dab407) whether [contextual](http://co.ntextu.al/) could be used to reduce the macro boilerplate.

johnynek:
I'm -1 on any additional dependencies if we can do this with a macro or implicit conversion.

For this I think we should do the highest performance version. A macro calling java string builder, I would think, would be the best.

johnynek:
@TomasMikula thanks for getting the ball rolling here. I'm going to close since we merged #1495 but reopen if closing is premature.

TomasMikula:
@johnynek 👍

PR #1490, opened 2016-12-10
Use scala versions define in .travis.yml

dwijnand:
Using sbt-travisci `crossScalaVersions` and `scalaVersion` are defined in terms of the scala versions defined in .travis.yml. Specifically, for `scalaVersion` the last scala version is used (therefore either re-order them or override it in build.sbt if it's not the default you want).

Also, define `organization in ThisBuild` once instead of defining it in `buildSettings` and appending it to every single project (in project scope).

Finally, remove use of `%%` inside `addSbtPlugin`, it's overridden anyways by `addSbtPlugin`.


kailuowang:
👍

dwijnand:
The irony that this conflicts with #1499, which is a 2 commit PR, first updating build.sbt and then .travis.yml :-) (and then mentioning sbt-travisci)

adelbertc:
Hurrah! Thanks @dwijnand ! And thanks to @djspiewak for making me aware of this plugin

dwijnand:
Excellent. Thanks @adelbertc.

PR #1493, opened 2016-12-11
Add roll, cata and cataM to Free

edmundnoble:
Stack-safe recursion, stack-safe monadic recursion, and easy `roll` constructor for those of us who use `Free` to build trees.


johnynek:
by the way, I think `roll` here is definitely worth adding.

I'm a bit skeptical of `cata` because the `Traverse` constraint seems pretty high. Like, I don't know how often I will be using `Free` on a `Traverse: Monad`.

edmundnoble:
As @johnynek says this is not very useful practically.

PR #1495, opened 2016-12-12
String interpolator for Show, without macros

dwijnand:
Non-macro alternative to #1489.

Happy to iterate - feedback welcome.

### DIFF
--- a/core/src/main/scala/cats/Show.scala
+++ b/core/src/main/scala/cats/Show.scala
@@ -30,7 +30,7 @@ object Show {
     implicit def mat[A](x: A)(implicit z: Show[A]): Shown = Shown(z show x)
   }
 
-  implicit class ShowInterpolator(val _sc: StringContext) extends AnyVal {
+  final case class ShowInterpolator(_sc: StringContext) extends AnyVal {
     def show(args: Shown*): String = _sc s (args: _*)
   }
 
--- a/core/src/main/scala/cats/Show.scala
+++ b/core/src/main/scala/cats/Show.scala
@@ -25,6 +25,15 @@ object Show {
     def show(a: A): String = a.toString
   }
 
+  final case class Shown(override val toString: String) extends AnyVal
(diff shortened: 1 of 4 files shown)


core/src/main/scala/cats/Show.scala
```diff
@@ -25,6 +25,15 @@ object Show {
     def show(a: A): String = a.toString
   }
 
+  final case class Shown(override val toString: String) extends AnyVal
+  object Shown {
+    implicit def mat[A](x: A)(implicit z: Show[A]): Shown = Shown(z show x)
+  }
+
+  implicit class ShowInterpolator(val _sc: StringContext) extends AnyVal {
+    def show(args: Shown*): String = StringContext(_sc.parts: _*) s (args: _*)
```
TomasMikula: `StringContext(_sc.parts: _*)` can be replaced by `_sc`, am I right?

TomasMikula:
I would exercise the principle of least power here and go with this solution over the macro one. It is much more transparent to the casual reader what is going on.

peterneyens:
This looks great :+1:

That implicit conversion to `Shown` is a nice trick.

Would  `syntax.show` be a better place ? So you can get it using `import cats.syntax.show._` (or `cats.implicits._`) instead of `import cats.Show._` ?

dwijnand:
Probably. Or it's own syntax object. Happy to go either way.

travisbrown:
👍 to supporting both `cats.syntax.show._` and `cats.implicits._` if we go this route (which I'm still not sure I prefer over #1489).

adelbertc:
👍

dwijnand:
`cats.syntax.show._`, `cats.syntax.all._` and `cats.implicits._` are all supported now.

dwijnand:
Looking for another maintainer to :+1: this.

kailuowang:
👍

PR #1499, opened 2016-12-13
updated scala 2.12.1

kailuowang:


johnynek:
👍

adelbertc:
👍  may be worth looking into using https://github.com/dwijnand/sbt-travisci

dwijnand:
@adelbertc maybe worth looking into approving #1490 :-P

PR #1500, opened 2016-12-14
Add biased operators to Ior

edmundnoble:
Ior is sometimes used in a very biased way. These tools make for convenient use in that case.

### DIFF
--- a/core/src/main/scala/cats/data/Ior.scala
+++ b/core/src/main/scala/cats/data/Ior.scala
@@ -28,6 +28,11 @@ sealed abstract class Ior[+A, +B] extends Product with Serializable {
     case Ior.Both(a, b) => fab(a, b)
   }
 
+  final def putLeft[C](left: C): C Ior B =
+    fold(_ => Ior.left(left), Ior.both(left, _), (_, b) => Ior.both(left, b))
+  final def putRight[C](right: C): A Ior C =
+    fold(Ior.both(_, right), _ => Ior.right(right), (a, _) => Ior.both(a, right))
+
   final def isLeft: Boolean = fold(_ => true, _ => false, (_, _) => false)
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/data/Ior.scala
```diff
@@ -91,6 +96,10 @@ sealed abstract class Ior[+A, +B] extends Product with Serializable {
 
   final def merge[AA >: A](implicit ev: B <:< AA, AA: Semigroup[AA]): AA =
     fold(identity, ev.apply, (a, b) => AA.combine(a, b))
+  final def mergeLeft[AA >: A](implicit ev: B <:< AA): AA =
+    fold(identity, ev.apply, (a, _) => a)
```
johnynek: I don't think you need `ev.apply` here. I *think* `B <:< AA` is already a function `B => AA` so you can pass it directly without allocating another closure.

Although, it would be interesting to know if scalac avoids allocating a closure when you do `fn.apply` and instead just passes `fn`.

edmundnoble: Good point, fixed it in `merge` too.

johnynek:
👍

travisbrown:
👍, looks reasonable to me.

core/src/main/scala/cats/data/Ior.scala
```diff
@@ -28,6 +28,11 @@ sealed abstract class Ior[+A, +B] extends Product with Serializable {
     case Ior.Both(a, b) => fab(a, b)
   }
 
+  final def putLeft[AA >: A](left: AA): AA Ior B =
```
johnynek: actually, why does this need to be `AA >: A` and not just `AA`? Aren't we fully replacing? Same with `putRight`?

edmundnoble:
Thanks @johnynek, fixed.

PR #1502, opened 2016-12-20
add Thunk type alias

jozic:


core/src/main/scala/cats/instances/function.scala
```diff
@@ -13,18 +13,18 @@ private[instances] sealed trait Function0Instances {
...
 
-      def flatMap[A, B](fa: () => A)(f: A => () => B): () => B =
-        () => f(fa())()
+      def flatMap[A, B](fa: Thunk[A])(f: A => Thunk[B]): Thunk[B] =
+        Thunk(f(fa())())
 
-      def tailRecM[A, B](a: A)(fn: A => () => Either[A, B]): () => B =
-        () => {
+      def tailRecM[A, B](a: A)(fn: A => Thunk[Either[A, B]]): Thunk[B] =
+        Thunk {
```
johnynek: this is going to be an extra allocation right? before we make a `Function0` with a given body. With `Thunk(foo: => A): Thunk[A] = () => foo` means this doesn't it: `def desugar(foo: () => A): Thunk[A] = () => foo()`, so we are going to incur an extra layer of wrapping no?

edmundnoble:
I'm with @johnynek on this one. I can't see why it's useful, some examples might help.

core/src/main/scala/cats/instances/function.scala
jozic: not sure about allocation, but this will indeed add a method call (apply), unless it's inlined by JIT

jozic:
@edmundnoble 
i just thought that something like 
`def flatMap[A, B](fa: Thunk[A])(f: A => Thunk[B]): Thunk[B] = Thunk(f(fa())())`
looks a bit more readable than
`def flatMap[A, B](fa: () => A)(f: A => () => B): () => B = () => f(fa())()`
which is very subjective, I admit. this is where it's "useful"

but as pointed by @johnynek Thunk.apply adds a little of overhead, which probably makes it not worth doing it
I guess `@inline final ` can be added, but as far as I understand there's no guarantees that scalac will inline it for real

PR #1504, opened 2016-12-22
Use Monad if available in Traverse

johnynek:
This is an alternate take on the same problem as #1476. Here we just try to fix the existing traverse for potentially large collections making use of tailRecM.

### DIFF
--- a/core/src/main/scala/cats/instances/map.scala
+++ b/core/src/main/scala/cats/instances/map.scala
@@ -17,12 +17,26 @@ trait MapInstances extends cats.kernel.instances.MapInstances {
   implicit def catsStdInstancesForMap[K]: TraverseFilter[Map[K, ?]] with FlatMap[Map[K, ?]] =
     new TraverseFilter[Map[K, ?]] with FlatMap[Map[K, ?]] {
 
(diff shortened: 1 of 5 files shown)


ceedubs:
This is kind of dirty, but I think that I like it :)

@johnynek sorry but you've got some merge conflicts. Maybe we should see if others are on board with this change before you take the time to fix the conflict. @typelevel/cats any thoughts?

core/src/main/scala/cats/instances/map.scala
```diff
@@ -33,6 +47,18 @@ trait MapInstances extends cats.kernel.instances.MapInstances {
         G.map(gbb)(_.toMap)
       }
 
+      private def traverseMFilter[G[_], A, B](fa: Map[K, A])(f: A => G[Option[B]])(implicit G: Monad[G]): G[Map[K, B]] = {
```
kailuowang: hmmm, is this method used?

johnynek:
I'm a little nervous about the PR still. Let me try to prove that this is actually equivalent. If it is, it will be a win and fix an issue I had with Traverse on a large list with StateT.

LukaJCB:
Hey @johnynek did you ever find a good conclusion? If not, I think we should maybe expose `traverseM` so users could "opt-in" to traverse using `tailRecM` if they have a `Monad`. WDYT?

kailuowang:

PR #1506, opened 2016-12-29
Partially revert 4530d3aa93131ec28096968a3c903ed1016dbf1b. Add back v…

marcin-rzeznicki:
…ariance on methods in EitherT and Either syntax.
Current signatures make it impossible to write, for example, for-comprehensions with 'lefts' not being of the same type. The following code does not compile with current signatures, but it compiles after the change (compiler is able to infer the common supertype):

```scala
sealed abstract class AppError
case object Error1 extends AppError
case object Error2 extends AppError

val either1: Future[Either[Error1.type , String]] = Future.successful(Right("hi"))
val either2: Future[Either[Error2.type , String]] = Future.successful(Right("bye"))

for {
  s1 <- EitherT(either1)
  s2 <- EitherT[Future, AppError, String](either2)
} yield s1 ++ s2
```

> Error:(17, 7) type mismatch;
 found   : cats.data.EitherT[scala.concurrent.Future,Error2.type,String]
 required: cats.data.EitherT[scala.concurrent.Future, Error1.type,?]
  s2 <- EitherT(either2)
     ^

This change has been discussed on gitter with @travisbrown and @edmundnoble


travisbrown:
:+1: from me. We'll need to fix the docs ([for example](https://github.com/typelevel/cats/blob/master/docs/src/main/tut/datatypes/either.md#either-in-the-small-either-in-the-large)) to match, and I'd prefer to see that happen in this PR, I guess, but it could be a follow-up.

core/src/main/scala/cats/syntax/either.scala
```diff
@@ -124,51 +124,51 @@ final class EitherOps[A, B](val eab: Either[A, B]) extends AnyVal {
...
     case Right(b1) =>
       that match {
         case Left(_)   => 1
-        case Right(b2) => B.compare(b1, b2)
+        case Right(b2) => BB.compare(b1, b2)
       }
   }
 
-  def partialCompare(that: Either[A, B])(implicit A: PartialOrder[A], B: PartialOrder[B]): Double = eab match {
+  def partialCompare[AA >: A, BB >: B](that: Either[AA, BB])(implicit AA: PartialOrder[AA], BB: PartialOrder[BB]): Double = eab match {
```
johnynek: this method is untested. Is it too much to ask to write a test of some kind during this PR?

core/src/main/scala/cats/syntax/either.scala
```diff
@@ -215,20 +215,20 @@ final class EitherOps[A, B](val eab: Either[A, B]) extends AnyVal {
...
       case left @ Left(_) => left
-      case Right(b2) => Right(B.combine(b1, b2))
+      case Right(b2) => Right(BB.combine(b1, b2))
     }
   }
 
-  def show(implicit A: Show[A], B: Show[B]): String = eab match {
-    case Left(a)  => s"Left(${A.show(a)})"
-    case Right(b) => s"Right(${B.show(b)})"
+  def show[AA >: A, BB >: B](implicit AA: Show[AA], BB: Show[BB]): String = eab match {
```
johnynek: this is untested. Can we add a simple test?

core/src/main/scala/cats/syntax/either.scala
```diff
@@ -215,20 +215,20 @@ final class EitherOps[A, B](val eab: Either[A, B]) extends AnyVal {
...
+    case Left(a)  => s"Left(${AA.show(a)})"
+    case Right(b) => s"Right(${BB.show(b)})"
   }
 
-  def ap[C](that: Either[A, B => C]): Either[A, C] = (new EitherOps(that)).flatMap(this.map)
+  def ap[AA >: A, BB >: B, C](that: Either[AA, BB => C]): Either[AA, C] = new EitherOps(that).flatMap(this.map)
```
johnynek: this is untested. Can we add a simple test?

core/src/main/scala/cats/syntax/either.scala
marcin-rzeznicki: Not at all. Will do

core/src/main/scala/cats/syntax/either.scala
marcin-rzeznicki: Of course

marcin-rzeznicki:
I'll take care of docs update as well, Thanks for pointing out.

core/src/main/scala/cats/syntax/either.scala
marcin-rzeznicki: Actually this one should be covered by order laws at https://github.com/typelevel/cats/blob/master/tests/src/test/scala/cats/tests/EitherTests.scala#L44 , right?

johnynek:
My codecov plugin seems to say that codecov thinks it is uncovered.

johnynek:
Maybe we aren't exercising this call site. The laws can't be calling this
one since PartialOrder is invariant.

marcin-rzeznicki:
Missing tests added in 5c86c44

marcin-rzeznicki:
@travisbrown It looks like the whole 'Either in the small...' chapter can be removed from the docs since `flatMap` you get from enrichment is aligned with what you get on 2.12. WDYT?

travisbrown:
@marcin-rzeznicki Yep, that makes sense to me.

docs/src/main/tut/datatypes/either.md
```diff
@@ -212,132 +212,6 @@ magic("123") match {
 }
 ```
 
-## Either in the small, Either in the large
-Once you start using `Either` for all your error-handling, you may quickly run into an issue where
```
adelbertc: Why is this being removed?

travisbrown: @adelbertc The idea is that all the stuff about the left sides needing to be the same will be a lot less relevant (or not at all relevant, in the form here) once the syntax `flatMap` for `Either` behaves the same as the real `flatMap` method on `Either` in 2.12.

adelbertc: This section talks about using local ADTs for local errors and wrapping as you go wider and wider in your application. Variance is tangentially relevant to Solution 1 but does not help with Solution 2.

travisbrown: Okay, you've convinced me the removal should be more surgical. :)

docs/src/main/tut/datatypes/either.md
```diff
@@ -212,132 +212,6 @@ magic("123") match {
 }
 ```
 
-## Either in the small, Either in the large
```
adelbertc: Please be more careful about removing this section - variance does not subsume the concerns here.

docs/src/main/tut/datatypes/either.md
marcin-rzeznicki: Nonetheless, the whole section was presented as a 'solution' - but the problem being solved does not exist anymore. To quote the relevant part: 'If you're on Scala 2.12, this line will compile and work as expected (...)'. Everything that followed seemed to have been written as a guide to deal with issues on previous versions of Scala. It of course does not invalidate ideas contained therein. Obviously, organizing your errors as ADTs is a valid and useful technique. But a rather generic one and not really tied to usage of `Either` in cats. Of course I'm open to any suggestions, but I must admit that I simply *do not know* what should be there instead.

adelbertc: We should strive to present certain patterns (for lack of a better word) for using these data types - in this case I wrote it to show how one could use `Either` throughout their code to represent possibly failing computations.

docs/src/main/tut/datatypes/either.md
marcin-rzeznicki: @adelbertc I've come up with a new version of docs. Only some parts that had been no longer true in presence of variance were rewritten. Some other parts were rearranged. Please check if you like it. ( Since Contributors' Guide discourages squashing commits, it may be awkward to compare :-( )

marcin-rzeznicki:
@adelbertc I've come up with a new version of docs. Only some parts that had been no longer true in presence of variance were rewritten. Some other parts were rearranged. Please check if you like it. ( Since Contributors' Guide discourages squashing commits, it may be awkward to compare :-( )

adelbertc:
Cool @marcin-rzeznicki , LGTM, thanks!

PR #1507, opened 2016-12-29
make stack-safety-test for tailRecM in MonadTests only fail when checking the test

fkz:
Previously, already calling `MonadTests[Signal].monad[Int, Int, Int].all` would cause a StackOverflowError, now this is only caused when executing the test and can be captured e.g. by using `check`.

In our case, this solved a ScalaTest test suite from aborting and thus not executing any tests to just a failing test.


kailuowang:
👍  on travis green

johnynek:
:+1:

fkz:
thanks for the quick merge! can we backport this to 0.8 or somehow otherwise make some release which includes this soonishly?

PR #1510, opened 2017-01-02
Make catchNonFatal on Future async

### LINKED ISSUE #1482: add abilitity to instantiate a non-blocking Future from MonadError
When abstracting functions to return `MonadError`, to use `Future` in production and `Try` in unit testing, it's impossible to get an async `Future` from any function of the `Future` implementation of `MonadError` - either `Future.successful` or `Future.failed` is called.

In cats versions before 0.7.x. `Monad.pureEval` would call `Future.apply`, as this function was removed, there is no way to use either `MonadError` or `Monad` in an abstract way to create an async `Future`, which makes it hard to wrap synchronised code blocks (e.g. Java, etc.) into abstract Monads.

I would expect `MonadError.catchNonFatal` to show this behaviour but it's implemented as `try pure(a) catch {...}` (in `ApplicativeError`) calling `Future.successful`. @alexknvl suggested on Gitter to specialise this function in `FutureInstances ` as:

```def catchNonFatal[A](a: => A)(implicit ev: Throwable <:< E): Future[A] = Future.apply(a)```

This would makes sense to me. This issue is probably related to #1316.

johnynek:
closes #1482

### DIFF
--- a/core/src/main/scala/cats/instances/future.scala
+++ b/core/src/main/scala/cats/instances/future.scala
@@ -35,6 +35,10 @@ trait FutureInstances extends FutureInstances1 {
       override def recoverWith[A](fa: Future[A])(pf: PartialFunction[Throwable, Future[A]]): Future[A] = fa.recoverWith(pf)
 
       override def map[A, B](fa: Future[A])(f: A => B): Future[B] = fa.map(f)
+
+      override def catchNonFatal[A](a: => A)(implicit ev: Throwable <:< Throwable): Future[A] = Future(a)
+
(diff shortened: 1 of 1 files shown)


travisbrown:
I think this is reasonable, but the [docs for `catchNonFatal`](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/ApplicativeError.scala#L80) should probably be changed not to mention `pure`.

> Here we try to call pure or catch and raise.

non:
👍 with or without doc change

PR #1512, opened 2017-01-03
Implement NonEmptyList#FilterNot

xavier-fernandez:
`NonEmptyList` is implementing the `filter` method, but not the `filterNot`, which is available at `scala.collection.immutable.List`. This method is useful in order to avoid negating a filter condition.


johnynek:
👍

PR #1513, opened 2017-01-04
Revamp Functor, Applicative, Traverse doc

adelbertc:


docs/src/main/tut/typeclasses/applicative.md
```diff
@@ -6,45 +6,277 @@ source: "core/src/main/scala/cats/Applicative.scala"
...
+
+Such an `Applicative` must obey three laws:
+
+* Associativity: No matter the order in which you product together three values, the result is isomorphic
+    * `fa.product(fb).product(fc) ~ fa.product(fb.product(fc))`
+    * With `map`, this can be made into an equality with `fa.product(fb).product(fc) = fa.product(fb.product(fc)).map { case (a, (b, c)) => ((a, b), c) }`
+* Left identity: Zipping a value on the left with unit results in something ismorphic to the original value
+    * `pure(()).product(fa) ~ fa`
+    * As an equality: `pure(()).product(fa).map(_._2) = fa`
+* Right identity: Zipping a value on the right with unit results in something ismorphic to the original value
```
edmundnoble: s/ismorphic/isomorphic, same 3 lines up

adelbertc: Fixed!

docs/src/main/tut/typeclasses/functor.md
```diff
@@ -6,137 +6,106 @@ source: "core/src/main/scala/cats/Functor.scala"
...
-method that exists on many classes in the Scala standard library, for
-example:
+A `Functor` instance must obey two laws:
 
-```tut:book
-Option(1).map(_ + 1)
-List(1,2,3).map(_ + 1)
-Vector(1,2,3).map(_.toString)
-```
+* Composition: Mapping with `f` and then again with `f` is the same as mapping once with the composition of `f` and `g`
```
adelbertc: oops there should be "then again with `g`" here

docs/src/main/tut/typeclasses/functor.md
```diff
@@ -6,137 +6,106 @@ source: "core/src/main/scala/cats/Functor.scala"
...
-holes. In the example above, we took a type which normally has two type holes, 
-`Function1[?,?]` and constrained one of the holes to be the `In` type, 
-leaving just one hole for the return type, resulting in `Function1[In,?]`. 
-Without kind-projector, we'd have to write this as something like 
-`({type F[A] = Function1[In,A]})#F`, which is much harder to read and understand.
+Taking this view, we can view `Functor` as the ability to work with a **single** effect - we can apply a pure
+function to a single effectful value without needing to "leave" the effect.
 
-## Using Functor
+## Functors compose
```
jozic: now this section looks like this

## Functors Compose
we can compose either with compose or with nested
compose example
nested example
compose notes
nested notes

how about this instead?
## Functors Compose
we can compose either with compose or with nested
### compose 
compose example
compose notes
### nested
nested example
nested notes

docs/src/main/tut/typeclasses/traverse.md
```diff
@@ -5,235 +5,138 @@ section: "typeclasses"
...
-Moreover, the fact that our aggregate job is not tied to any specific `SparkContext` allows us
-to pass in a `SparkContext` pointing to a production cluster, or (using the exact same job) pass
-in a test `SparkContext` that just runs locally across threads. This makes testing our large
-job nice and easy.
-
-Finally, this encoding ensures that all the jobs for each topic run on the exact same cluster.
-At no point do we manually pass in or thread a `SparkContext` through - that is taken care for us
-by the (applicative) effect of `Reader` and therefore by `traverse`.
-
-## Sequencing
```
kailuowang: Since the `Traverse` tutorial starts with `Future.sequence`, is it still worth mentioning the `sequence` method in `Traverse`?

docs/src/main/tut/typeclasses/traverse.md
```diff
@@ -5,235 +5,138 @@ section: "typeclasses"
...
-Going back to our `Future` example, we can write an `Applicative` instance for
-`Future` that runs each `Future` concurrently. Then when we traverse a `List[A]`
-with an `A => Future[B]`, we can imagine the traversal as a scatter-gather.
-Each `A` creates a concurrent computation that will produce a `B` (the scatter),
-and as the `Future`s complete they will be gathered back into a `List`.
+```tut:book:silent
+implicit val applicativeForId: Applicative[Id] = new Applicative[Id] {
+  def ap[A, B](ff: Id[A => B])(fa: Id[A]): Id[B] = Id(ff.value(fa.value))
 
-### Playing with `Reader`
```
kailuowang: I found this section is another good example of `Reader`, how about we move it the [kleisli.md](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Traverse.scala) instead of losing it?

docs/src/main/tut/typeclasses/functor.md
adelbertc: Good idea, will do

docs/src/main/tut/typeclasses/traverse.md
adelbertc: Will do

docs/src/main/tut/typeclasses/traverse.md
adelbertc: Hm how about I go back to this when I (or someone else) gets around to revisiting the Reader tut? I don't want to just tack on this section at the end since one of the reasons I'm revisiting these docs is to try to make them more cohesive.

tpolecat:
This looks great to me. 👍

adelbertc:
Thanks for the review all! Merging now.

PR #1514, opened 2017-01-04
Implement NonEmptyVector#filterNot

xavier-fernandez:
As a follow up for the PR #1512, `NonEmptyVector` is also implementing the filter method, but not the `filterNot`, which is available at `scala.collection.immutable.Vector`. This method is useful in order to avoid negating a filter condition.


ceedubs:
👍 LGTM. Thanks!

kailuowang:
👍  Thanks!

PR #1515, opened 2017-01-04
Implement NonEmptyVector#size

xavier-fernandez:
Scala `Vector` instance haves both `length` and `size` methods implemented and working in the same way, this commit adds this missing method to Cats `NonEmptyVector`, so the class becomes more consistent with its Scala counterpart.

### DIFF
--- a/core/src/main/scala/cats/data/NonEmptyVector.scala
+++ b/core/src/main/scala/cats/data/NonEmptyVector.scala
@@ -166,7 +166,29 @@ final class NonEmptyVector[A] private (val toVector: Vector[A]) extends AnyVal {
   def show(implicit A: Show[A]): String =
     s"NonEmpty${Show[Vector[A]].show(toVector)}"
 
+  /**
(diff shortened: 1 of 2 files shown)


travisbrown:
I definitely don't think we need both of these—the duplication in the standard library is an annoying accident and we have a chance to do better.

More generally, I don't think `NonEmptyList` or `NonEmptyVector` either are or should be aiming for API parity with their standard library counterparts. Methods like `reverse` that maintain the non-emptiness definitely belong, and I can see an argument for some other useful methods (including probably `size`), but I don't think "the standard library has it" should automatically be a reason for inclusion here.

xavier-fernandez:
All the collections from the standard library has a method `size`, but only a few have a method `length`, which is (I suppose) an inheritance from Java arrays.

As an alternative, I would propose deprecating the `length` method with:

```scala
 @deprecated(since = "0.9", message = "Vector#length is deprecated, use Vector#size instead")
```

PR #1516, opened 2017-01-05
Implement NonEmptyList#Collect

xavier-fernandez:
- "collect" method is really useful when wanting to do a transformation in some of the elements from a collection, applying a partial function. This PR implements this method both for `NonEmptyVector` and `NonEmptyList`.

### DIFF
--- a/core/src/main/scala/cats/data/NonEmptyList.scala
+++ b/core/src/main/scala/cats/data/NonEmptyList.scala
@@ -67,6 +67,30 @@ final case class NonEmptyList[+A](head: A, tail: List[A]) {
   }
 
   /**
+    * Builds a new `List` by applying a partial function to
+    * all the elements from this `NonEmptyList` on which the function is defined
+    *
+    * {{{
+    * scala> import cats.data.NonEmptyList
+    * scala> val nel = NonEmptyList.of(1, 2, 3, 4, 5)
+    * scala> nel.collect { case v if v < 3 => v }
+    * res0: scala.collection.immutable.List[Int] = List(1, 2)
+    * scala> nel.collect {
+    *      |  case v if v % 2 == 0 => "even"
+    *      |  case _ => "odd"
+    *      | }
+    * res1: scala.collection.immutable.List[String] = List(odd, even, odd, even, odd)
+    * }}}
+    */
+  def collect[B](pf: PartialFunction[A, B]) : List[B] = {
+    if (pf.isDefinedAt(head)) {
+      pf.apply(head) :: tail.collect(pf)
+    } else {
+      tail.collect(pf)
+    }
+  }
+
+  /**
    * Append another NonEmptyList
    */
   def concat[AA >: A](other: NonEmptyList[AA]): NonEmptyList[AA] =
--- a/tests/src/test/scala/cats/tests/NonEmptyListTests.scala
+++ b/tests/src/test/scala/cats/tests/NonEmptyListTests.scala
@@ -88,6 +88,13 @@ class NonEmptyListTests extends CatsSuite {
     }
   }
 
+  test("NonEmptyList#collect is consistent with List#collect") {
+    forAll { (nel: NonEmptyList[Int], pf: PartialFunction[Int, String]) =>
+      val list = nel.toList
+      nel.collect(pf) should === (list.collect(pf))
+    }
+  }
+
   test("NonEmptyList#find is consistent with List#find") {
     forAll { (nel: NonEmptyList[Int], p: Int => Boolean) =>
       val list = nel.toList


johnynek:
I'm weakly -1 on adding methods which are identical to doing `.toList.foo` or `toVector.foo`.

Methods that preserve nonemptiness are more interesting to me.

xavier-fernandez:
What about returning: `Option[NonEmptyList[B]]`?

kailuowang:
@xavier-fernandez, unless I am missing something, a `Option[NonEmptyList[B]]` is as powerful as a `List[B]` but less convenient. So you do want to return a `List[B]`. And if you going to lose the nonemptiness anyway why not `x.toList.collect`? To me it makes the lost of nonemptiness more explicit. Thus I favor that than `x.collect` that returns a `List`. 
The `filter` method we have there is an existing example of such methods that don't preserve nonemptines. @ceedubs do you happen to remember the incentive of adding that filter method?

xavier-fernandez:
I should close the PR?

ceedubs:
The main arguments that I can think of for having the `filter` method is that for `NonEmptyList` it can be implemented every so slightly more efficiently than `toList.filter` and that it may be slightly more convenient for users. I do recall at least one time that a coworker has asked me how they should filter a scalaz `NonEmptyList`, as it isn't necessarily obvious that you should go through `List`. I'm pretty neutral as to whether or not we include the `List`-like methods that don't offer anything specific to being non-empty.

xavier-fernandez:
`NonEmptyList.collect` will have a better performance than `toList.collect`...

On the other hand,`NonEmptyVector.collect` will not yield any performance improvement.

johnynek:
Okay. On this argument I'm +1 on NonEmptyList.collect but not
NonEmptyVector.

xavier-fernandez:
If `NonEmptyList` is having a collect method, it would not be better that `NonEmptyVector` too, for consistency?

johnynek:
I don't think so. Otherwise we'll have some other class justifying why we
should have yet another method.

Frankly, if consistency is the main driver I suggest we go back to only
adding methods if they preserve non-emptiness.

xavier-fernandez:
Modified the PR so the `collect` method is only implemented on `NonEmptyList`

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -67,6 +67,30 @@ final case class NonEmptyList[+A](head: A, tail: List[A]) {
   }
 
   /**
+    * Builds a new collection by applying a partial function to
```
peterneyens: Should we say "Builds a new `List` ..." ?

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -67,6 +67,30 @@ final case class NonEmptyList[+A](head: A, tail: List[A]) {
...
+    *
+    * {{{
+    * scala> import cats.data.NonEmptyList
+    * scala> val nel = NonEmptyList.of(1, 2, 3, 4, 5)
+    * scala> nel.collect { case v if v < 3 => v }
+    * res0: scala.collection.immutable.List[Int] = List(1, 2)
+    * scala> nel.collect {
+    *   case v if v % 2 == 0 => "even"
+    *   case _ => "odd"
+    *   }
```
peterneyens: I think this should be : 

```scala
/**
 * ...
 * scala> nel.collect {
 *      |   case v if v % 2 == 0 => "even"
 *      |   case _ => "odd"
 *      | }
 * ...
 */
```

for the doctest to be generated correctly.

xavier-fernandez:
@peterneyens amend done :)

johnynek:
👍

peterneyens:
You need to use exactly 
```
     | 
```
after a `scala> ` for the test to be picked up.

Right now the generated doctest is 

```scala
    property("example at line 203: nel.collect { case v if v < 3 => v }") = org.scalacheck.Prop.secure {
      sbtDoctestTypeEquals(nel.collect { case v if v < 3 => v })((nel.collect { case v if v < 3 => v }): scala.collection.immutable.List[Int])
      val actual = sbtDoctestReplString(nel.collect { case v if v < 3 => v })
      val expected = "List(1, 2)"
      (actual == expected) :| s"'$actual' is not equal to '$expected'"
    }
```

If you change it to 

```scala
   /**
    * scala> nel.collect {
    *      |   case v if v % 2 == 0 => "even"
    *      |   case _ => "odd"
    *      | }
    */
```

It will include this one as well : 

```scala
    property("example at line 205: nel.collect { ...") = org.scalacheck.Prop.secure {
      sbtDoctestTypeEquals(nel.collect {
  case v if v % 2 == 0 => "even"
  case _ => "odd"
})((nel.collect {
  case v if v % 2 == 0 => "even"
  case _ => "odd"
}): scala.collection.immutable.List[Int])
      val actual = sbtDoctestReplString(nel.collect {
  case v if v % 2 == 0 => "even"
  case _ => "odd"
})
      val expected = "List(odd, even, odd, even, odd)"
      (actual == expected) :| s"'$actual' is not equal to '$expected'"
    }
```

It would be handy if `sbt-doctest` gave some kind of error message if it is ignoring lines.

xavier-fernandez:
@peterneyens will fix tomorrow morning, thank you :)

xavier-fernandez:
Did an ammend fixing the DocTest generation :)

kailuowang:
👍  merging with 2 sign-offs.

PR #1517, opened 2017-01-05
Add NonEmptyList.zipWithIndex

cranst0n:


core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -165,6 +165,26 @@ final case class NonEmptyList[A](head: A, tail: List[A]) {
...
+   * {{{
+   * scala> import cats.data.NonEmptyList
+   * scala> val nel = NonEmptyList.of("a", "b", "c")
+   * scala> nel.zipWithIndex
+   * res0: cats.data.NonEmptyList[(String, Int)] = NonEmptyList((a,0), (b,1), (c,2))
+   * }}}
+   */
+  def zipWithIndex: NonEmptyList[(A, Int)] = {
+    @tailrec
+    def go(rest: List[A], ix: Int, acc: List[(A, Int)]): List[(A, Int)] =
```
johnynek: while this is a nice implementation, I think `ListBuilder` can break the rules and mutate `List`s and be faster. So, something like;
```scala
val bldr = List.newBuilder[(A, Int)]
var idx = 1
val it = tail.iterator
while (it.hasNext) {
  bldr += (it.next, idx)
  idx += 1
}
NonEmptyList((head, 0), bldr.result)
```
will be faster, which is nice when we are talking about common libraries.

cranst0n: Sure. I can update with your recommendations. For the record, here's the output of a quick benchmark I can with the 2 different approaches.

```
[info] Benchmark                                   (listSize)   Mode  Cnt         Score         Error  Units
[info] NelZipWithIndexBench.zipWithIndexBuilder             1  thrpt   20  81866809.100 ± 1827269.422  ops/s
[info] NelZipWithIndexBench.zipWithIndexBuilder            10  thrpt   20   8397494.282 ±  265179.762  ops/s
[info] NelZipWithIndexBench.zipWithIndexBuilder           100  thrpt   20    837125.798 ±   19766.234  ops/s
[info] NelZipWithIndexBench.zipWithIndexBuilder          1000  thrpt   20     65129.596 ±    1794.893  ops/s
[info] NelZipWithIndexBench.zipWithIndexBuilder         10000  thrpt   20      6648.203 ±     103.004  ops/s
[info] NelZipWithIndexBench.zipWithIndexBuilder        100000  thrpt   20       540.067 ±      14.993  ops/s
[info] NelZipWithIndexBench.zipWithIndexBuilder       1000000  thrpt   20        16.204 ±       5.125  ops/s
[info] NelZipWithIndexBench.zipWithIndexBuilder      10000000  thrpt   20         0.205 ±       0.082  ops/s
[info] NelZipWithIndexBench.zipWithIndexRecursive           1  thrpt   20  88856828.471 ± 2776243.282  ops/s
[info] NelZipWithIndexBench.zipWithIndexRecursive          10  thrpt   20   2548195.555 ±   92940.797  ops/s
[info] NelZipWithIndexBench.zipWithIndexRecursive         100  thrpt   20    244793.012 ±    6162.114  ops/s
[info] NelZipWithIndexBench.zipWithIndexRecursive        1000  thrpt   20     21696.738 ±     632.243  ops/s
[info] NelZipWithIndexBench.zipWithIndexRecursive       10000  thrpt   20      4267.619 ±     805.001  ops/s
[info] NelZipWithIndexBench.zipWithIndexRecursive      100000  thrpt   20       356.116 ±       9.618  ops/s
[info] NelZipWithIndexBench.zipWithIndexRecursive     1000000  thrpt   20        10.396 ±       2.741  ops/s
[info] NelZipWithIndexBench.zipWithIndexRecursive    10000000  thrpt   20         0.111 ±       0.014  ops/s
```

non: That's great! Thanks for benchmarking.

👍

johnynek: Way to be rigorous! So, if I'm reading this right, if the `tail != Nil` (listSize > 1) this is ~ 2-3x faster to use the builder. I would have assumed 2x or so because you skip the reverse.

johnynek:
👍

kailuowang:
👍

PR #1518, opened 2017-01-05
Implement NonEmptyList#mkShowString methods

xavier-fernandez:


kailuowang:
We already have a `Show` instance, which is an overlap with the `mkString()`.  
Do we need the two other overloads in `NonEmptyList`? I am not sure. If we could design from scratch I would probably leave those in a type class.  I am leaning slightly towards to not adding them here inside the `NonEmptyList` but I am biased mainly due to my distaste against the scala standard collection library design.

johnynek:
Yeah I'm -1 on mkString without using Show. Not the intent of cats. I would be +1 on a mkShowString or something that uses Show instead of toString on the intermediate parts. Maybe only the three argument version is needed.

non:
I'd support something like this, FWIW:

```scala
def intercalate(sep: String)(implicit ev: A =:= String): String =
  ev(head) + sep + tail.mkString(sep)
```

I agree that we shouldn't implement this method using `.toString`, but it wouldn't be a bad idea to support cases where you already have strings.

xavier-fernandez:
Thank you @non, will try and push an amend as soon as I have it working :)

Should use the name `mkString` or `intercalate`? (In order to be consistent with the rest of the library?)

travisbrown:
I'm -1 on either version of this—both just seem much too specific to deserve inclusion when the syntax / runtime cost of `.toList.mkString` is going to be pretty small.

I think this is another specific case of the "is this library focused on abstractions or just convenient stuff" question.

xavier-fernandez:
Following the feedback from @kailuowang , @johnynek and @non I updated the PR adding an implementation using the `Show` typeclass, instead of using the method `toString`.

xavier-fernandez:
Following the feedback from @kailuowang , @johnynek and @non I updated the PR adding an implementation using the Show typeclass, instead of using the method toString.

xavier-fernandez:
Rebased so this changes are compatible with: #1424

johnynek:
Given the number of allocations mkString already requires, I think
toList.mkString would be my vote and not add this method.

xavier-fernandez:
@johnynek The one without input parameters?

johnynek:
actually, I am concerned about many ways to do something here.

As of #1520 if we wanted a mkString on NonEmptyList we could do:
```scala
import cats.data.NonEmptyList
import cats.implicits._

val ne = NonEmptyList.of(1, 2, 3)
ne.map(_.toString).intercalate(", ")
//
or
ne.map(_.show).intercalate(", ")
// or
ne.toList.mkString(", ")
```
I'm not sure we really need to add another way to do this.

PR #1520, opened 2017-01-06
Add Foldable.intercalate

peterneyens:
Add an element between all the existing elements while folding.

### DIFF
--- a/core/src/main/scala/cats/Foldable.scala
+++ b/core/src/main/scala/cats/Foldable.scala
@@ -393,8 +393,12 @@ import simulacrum.typeclass
    * scala> import cats.implicits._
    * scala> Foldable[List].intercalate(List("a","b","c"), "-")
    * res0: String = a-b-c
+   * scala> Foldable[List].intercalate(List("a"), "-")
+   * res1: String = a
+   * scala> Foldable[List].intercalate(List.empty[String], "-")
+   * res2: String = ""
    * scala> Foldable[Vector].intercalate(Vector(1,2,3), 1)
-   * res1: Int = 8
+   * res3: Int = 8
    * }}}
    */
   def intercalate[A](fa: F[A], a: A)(implicit A: Monoid[A]): A =
--- a/core/src/main/scala/cats/Foldable.scala
+++ b/core/src/main/scala/cats/Foldable.scala
@@ -386,6 +386,22 @@ import simulacrum.typeclass
   def nonEmpty[A](fa: F[A]): Boolean =
     !isEmpty(fa)
 
+  /**
+   * Intercalate/insert an element between the existing elements while folding.
+   *
+   * {{{
+   * scala> import cats.implicits._
+   * scala> Foldable[List].intercalate(List("a","b","c"), "-")
+   * res0: String = a-b-c
+   * scala> Foldable[Vector].intercalate(Vector(1,2,3), 1)
+   * res1: Int = 8
+   * }}}
+   */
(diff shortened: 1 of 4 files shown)


ceedubs:
Thanks @peterneyens. This looks good to me. The string example is the main way I could see myself using this in the real world. Out of curiosity, did you have any other use-cases in mind?

If we add this, should we add something like `intercalate1` (to follow the naming convention of `traverse1_` and `sequence1_`) to `Reducible` that requires only a `Semigroup` as opposed to a `Monoid`?

peterneyens:
- Mostly as an alternative for #1518 (for strings) indeed. 
- Another use case when I was thinking about something like `intercalate` was pretty printing (specifically [here](https://github.com/47deg/fetch/blob/master/debug/shared/src/main/scala/debug.scala#L29) in fetch).
- I think @tpolecat asked once if cats had `intercalate`, maybe he had another use case ?

I added `Reducible.intercalate1`, forgot last night that we could lower the requirement to `Semigroup` for `Reducible`.

ceedubs:
@peterneyens would you be willing to add a test for `intercalate1`? It'd probably be nice to hit the overridden `intercalate` in `Reducible` too.

ceedubs:
Actually I guess the sbt-doctest examples are pretty legitimate tests. 👍 from me.

ceedubs:
Good points @johnynek.

tpolecat:
@peterneyens yeah I need it for generating SQL … I also did a [more general thing](https://github.com/tpolecat/doobie/blob/series/0.3.x/yax/core/src/main/scala/doobie/util/foldable.scala#L22) that takes start/end brackets as well as the delimiter, which generalizes `mkString` and is useful but it's hard to think of a good name.

In any case 👍 for this, with test improvements as suggested above.

adelbertc:
👍 after test improvements.

peterneyens:
Thanks for the pointers @johnynek 

- I used `mkString` and `F[String]` in the tests instead.
- I added a test for `intercalate1` (it was already tested before through the overridden `intercalate` which is tested because `ReducibleCheck` extends `FoldableCheck`, but is probably better to be explicit).
- I added a doctest for `intercalate` and `intercalate1` with a one item foldable. The one element foldable case is already covered by the property tests, right ?

johnynek:
+1 on this.

adelbertc:
Let's do this

johnynek:
one bummer about this: using it to write `mkString` will be `O(N^2)` in cost. This is one of the problems that `combineAll` was designed to deal with. Not sure the best way to go here. For integers it is a waste, but for a lot of strings a huge win.

The simplest thing would be to use `Foldable.toList` to build a list, then intercalate only on the list, then call combineAll on the list. I guess that might be the right way to go actually.

peterneyens:
@johnynek, so something like the following ?

```scala
def intercalate[A](fa: F[A], a: A)(implicit A: Monoid[A]): A =
  combineAll(intersperseList(toList(fa), a))

private def intersperseList[A](xs: List[A], x: A): List[A] = {
  val bld = List.newBuilder[A]
  val it = xs.iterator
  if (it.hasNext) {
    bld += it.next
    while(it.hasNext) {
      bld += x
      bld += it.next
    }
  }
  bld.result
}

// Reducible 
def intercalate1[A](fa: F[A], a: A)(implicit A: Semigroup[A]): A =
  reduce(intersperseList(toList(fa), a))
```

johnynek:
Exactly.

Issue #1184, opened 2016-07-11
Propose stop publishing `cats` bundle 


benhutchison:
(Channeling @travisbrown) I propose getting rid of the cats artifact. The only real role it plays is bundling core, free, and the laws / tests packages, and it's just as easy to explain how and when users need each of those individual pieces.
#1181 is example of confusion caused

benhutchison:
Im actually having second thoughts myself..

@ceedubs clarified the thinking behind the cats artifact: "I think that there's a natural progression of people starting with that and potentially later adding free or laws as needed for their use-case. I think that @non liked the idea of recommending the "batteries included" version and letting people trim off pieces once they were ready for production usage. " 

The batteries included idea resonates with me if/when cats grows lots of modules; it worked for Haskell. And, remembering how annoying it seemed as a Scalaz newbie when things I was using suddenly disappeared and I didn't have context on why, Im think there's value in keeping existing `cats` deps working for users.

What about instead of removing it I write a paragraph for the readme explaining when to use `cats` vs `cats-core`?

travisbrown:
I don't really like the "here's how you do this thing (but remember not to ever _actually_ do this thing, because weird stuff will happen)" model of pedagogy. :smile: 

In particular I'm not sure there's ever a good reason to have a non-test dependency on cats-laws or cats-tests, and this necessarily makes them a non-test dependency. Apart from them there's just cats-free, and if you get a compiler error for `import cats.free._` that's arguably a lot easier to track down than a bunch of unexpected stuff showing up in your builds.

benhutchison:
Agree on removing cats-{laws/test} from `cats` bundle.

Still see value in "batteries included" though. Right now there's only cats-free but more could be added, eg there's some talk of alleycats coming into cats. If `cats` is "everything you might need for runtime use", then there isn't such a clear need to trim it later.

adelbertc:
I would like to see a `cats` that includes all modules except test-related things, and then perhaps a `cats-testing` or similar for laws and tests.

benhutchison:
I started a PR to try close this, that'd remove test-related things from `cats`, but got stuck on some [SBT voodoo](https://gitter.im/sbt/sbt?at=5795a2a21b9de56c0ee44c43) in [build.sbt](https://github.com/typelevel/cats/blob/master/build.sbt#L179). 

If anyone knows too much about Ivy Configurations and could educate & unblock me, I'll be very grateful and send a PR to close this issue in return ;)

[...]

PR #1521, opened 2017-01-07
Don't publish empty jar of root modules

Taig:
The `catsJVM` and `catsJS` root/aggregate modules are currently producing empty `*.jar` files. This is unnecessary and causes minor build issues on the Android toolchain.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -182,6 +182,7 @@ lazy val cats = project.in(file("."))
 lazy val catsJVM = project.in(file(".catsJVM"))
   .settings(moduleName := "cats")
   .settings(catsSettings)
+  .settings(noSourcesSettings)
   .settings(commonJvmSettings)
   .aggregate(macrosJVM, kernelJVM, kernelLawsJVM, coreJVM, lawsJVM, freeJVM, testsJVM, jvm, docs, bench)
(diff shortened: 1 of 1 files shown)


Taig:
I just noticed that the `catsJS` JAR contains a file `JS_DEPENDENCIES` which is probably crucial. Would it make more sense to only apply `noSourcesSettings` to `catsJVM`?

kailuowang:
I think publishing to Sonatype requires the standard set of jars. It might've changed, but I don't really expect that.

Taig:
I recently applied the same change to a library of mine which is handled just fine: https://oss.sonatype.org/#nexus-search;gav~io.taig~communicator_2.11~3.0.1~~

travisbrown:
I think I'd prefer not to complicate the build for the sake of the grab-bag `cats` artifact, which should probably never be used in production anyway (because of the runtime ScalaTest dependency, etc.), and in my view shouldn't even exist, so I'm a weak :-1: on this.

kailuowang:
@travisbrown do we have an issue discussing the grab-bag cats artifact? I thought we have one but couldn't find it.

travisbrown:
@kailuowang Yep, it's #1184.

kailuowang:
we no longer publish all inclusive bundle now.

PR #1522, opened 2017-01-09
Foldable/Reducible intercalate update

peterneyens:
Follow up on #1520, after @johnynek 's comment.

This should reduce the complexity for `intercalate` from `O(N^2)` to `O(N)`.
I also overrode `toList` in some `Foldable` instances.

### DIFF
--- a/core/src/main/scala/cats/Reducible.scala
+++ b/core/src/main/scala/cats/Reducible.scala
@@ -171,10 +171,11 @@ import simulacrum.typeclass
    * }}}
    */
   def intercalate1[A](fa: F[A], a: A)(implicit A: Semigroup[A]): A =
-    reduceLeft(fa)((acc, aa) => A.combine(acc, A.combine(a, aa)))
-
-  override def intercalate[A](fa: F[A], a: A)(implicit A: Monoid[A]): A =
-    intercalate1(fa, a)
+    toNonEmptyList(fa) match {
+      case NonEmptyList(hd, Nil) => hd
+      case NonEmptyList(hd, tl) =>
+        Reducible[NonEmptyList].reduce(NonEmptyList(hd, a :: intersperseList(tl, a)))
+    }
 }
 
 /**
--- a/core/src/main/scala/cats/instances/list.scala
+++ b/core/src/main/scala/cats/instances/list.scala
@@ -83,6 +83,8 @@ trait ListInstances extends cats.kernel.instances.ListInstances {
       override def foldM[G[_], A, B](fa: List[A], z: B)(f: (B, A) => G[B])(implicit G: Monad[G]): G[B] =
         Foldable.iteratorFoldM(fa.toIterator, z)(f)
 
+      override def fold[A](fa: List[A])(implicit A: Monoid[A]): A = A.combineAll(fa)
+
       override def toList[A](fa: List[A]): List[A] = fa
     }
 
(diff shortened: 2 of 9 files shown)


core/src/main/scala/cats/Reducible.scala
```diff
@@ -170,11 +170,14 @@ import simulacrum.typeclass
...
    * }}}
    */
-  def intercalate1[A](fa: F[A], a: A)(implicit A: Semigroup[A]): A =
-    reduceLeft(fa)((acc, aa) => A.combine(acc, A.combine(a, aa)))
+  def intercalate1[A](fa: F[A], a: A)(implicit A: Semigroup[A]): A = {
+    val nel = toNonEmptyList(fa) match {
+      case one @ NonEmptyList(_, Nil) => one
+      case NonEmptyList(hd, tl) => NonEmptyList(hd, a :: intersperseList(tl, a))
+    }
+    Reducible[NonEmptyList].reduce(nel)
```
peterneyens: I could probably skip the unnecessary `reduce` step for `one`.

```scala
  toNonEmptyList(fa) match {
    case NonEmptyList(hd, Nil) => hd
    case NonEmptyList(hd, tl) => 
      Reducible[NonEmptyList].reduce(NonEmptyList(hd, a :: intersperseList(tl, a)))
  }
```

johnynek:
👍

travisbrown:
:+1:

johnynek:
Actually it looks like Foldable.combineAll is just an alias for fold which does not call Monoid.combineAll. I suggest we make Foldable.combineAll be Monoid.combineAll(fa.toList)

peterneyens:
Updated to use the `combineAll` from `Monoid`.

It seems we where already using `Semigroup.combineAllOption` to implement `Reducible#reduce` for `NonEmptyVector`, so I went ahead and did the same for `NonEmptyList` and overrode the `fold` method of the `Foldable` instances of `List`, `Vector`, `Set`, `Stream` and `Map`. (Hmm, this might have deserved its own PR.)

johnynek:
👍

core/src/main/scala/cats/instances/list.scala
```diff
@@ -82,6 +82,10 @@ trait ListInstances extends cats.kernel.instances.ListInstances {
 
       override def foldM[G[_], A, B](fa: List[A], z: B)(f: (B, A) => G[B])(implicit G: Monad[G]): G[B] =
         Foldable.iteratorFoldM(fa.toIterator, z)(f)
+
+      override def fold[A](fa: List[A])(implicit A: Monoid[A]): A = A.combineAll(fa)
```
kailuowang: I probably missed something but I am curious what gain we get from overriding this? I mean the 
In `Monoid.combineAll` it's 
```scala 
as.foldLeft(empty)(combine)
```
In foldable.fold it's
```Scala
 foldLeft(fa, A.empty) { (acc, a) =>
      A.combine(acc, a)
    }
```

peterneyens: I was confused about that too.

When you run the [`MapMonoidBench`](https://github.com/typelevel/cats/blob/master/bench/src/main/scala/cats/bench/MapMonoidBench.scala#L24-L30), you can see that the first using `Monoid.combineAll` is considerably faster than the second.

That's because `Monoid.combineAll` is overridden in some instances, eg [`Map`](https://github.com/typelevel/cats/blob/master/kernel/src/main/scala/cats/kernel/instances/map.scala#L41-L51) and [more general `TraversableOnce`](https://github.com/typelevel/cats/blob/master/kernel/src/main/scala/cats/kernel/instances/StaticMethods.scala#L78-L81).

kailuowang: got it. that's what I suspected, but I only checked the `List` instance.

kailuowang:
👍

PR #1524, opened 2017-01-15
0 9 0 post release

adelbertc:


docs/src/main/tut/index.md
```diff
@@ -81,20 +80,6 @@ we can without making unnecessary sacrifices of purity and
 usability. Where sacrifices have to be made, we will strive to make
 these obvious, and will keep them well documented.
 
-
-### <a name="project-structure" href="#project-structure"></a>Project Structure
```
adelbertc: I removed this section because it was redundant (and actually out of date) to the identical section we have above.

kailuowang:
👍

PR #1525, opened 2017-01-16
Remove link to apply.html from menu

Leammas:
After PR #1513 there's no such page. 
http://typelevel.org/cats/typeclasses/apply.html returns 404.


johnynek:
👍

PR #1526, opened 2017-01-24
inject for free programs

tpolecat:
This adds an `inject` operation for `Free` that lifts programs rather than constructors. This allows you to use existing `Free` programs (like doobie `ConnectionIO`) in coproducts without having to write implicit modules of lifted constructors.

I'll add a test if people think this is useful.


ceedubs:
I think that this is a handy method. But it looks like you have a compile error :)

The implementation is fairly trivial, but I think at least an sbt-doctest example (if not also a separate test) would be great.

tpolecat:
Doh, I was compiling core instead of free.

ceedubs:
👍 LGTM

adelbertc:
👍  LGTM

PR #1527, opened 2017-01-24
Add a few type classes to generated tuple instances

edmundnoble:


peterneyens:
Should we add instances for `CommutativeSemigroup`, `CommutativeGroup` and `BoundedSemilattice` while we are at it?

johnynek:
We're going to need generators for our generators.

Are we getting the priorities correct? We want `CommutativeX` to be a higher priority than `X`.

It would be interesting if we had some better way to generalize over this.

johnynek:
I often wonder how long this approach is the right one and whether a shapeless based approach would be almost as fast and much cleaner to maintain.

edmundnoble:
@johnynek What about Monoid and Semigroup for example, shouldn't they be different priorities too?

@peterneyens I added the two suggested as well.

Note this breaks bincompat, and so needs to be for 1.0.0.

kailuowang:
If there is concerns over breaking bin comp for kernel we can add these instances in cats.core right?

johnynek:
these are fine from algebird's point of view. We don't use these traits directly.

johnynek:
@edmundnoble yes, I think the priority issue might also be a problem for Semigroup. I think it was an oversight. We should probably have resolution tests to make sure things work as expected. Just using Tuple2 is probably enough.

kailuowang:
@edmundnoble I wonder if you have time to finish up this PR? Looks like we are close (1. fix the priority, and 2. add mima exception)

kailuowang:
Continued in #1967

PR #1530, opened 2017-01-30
add a favicon for sbt-microsite

larsrh:
Turns out there is a hardcoded path to a favicon in sbt-microsite, which is currently a broken link.

With this link, users browsing the Cats page should hopefully see a :cat: in the tab bar.


kailuowang:
👍

larsrh:
Fun fact: this is my first pull request to Cats. *late to the party*

kailuowang:
Now you are on the contributor list :)

larsrh:
Success!

screenshot_20170131_090320

PR #1531, opened 2017-02-02
Add piecemeal import for MonadError

peterneyens:
@jackcviers and @Wogan pointed out on our gitter channel that there is no piecemeal syntax import for `MonadError`.


johnynek:
👍

PR #1532, opened 2017-02-02
Override Foldable methods

peterneyens:
- Override `Foldable` methods in a lot of instances (including `Reducible` and `NonEmptyReducible`)
- Override `MonoidK#algebra` for `List`, `Vector`, etc to get kernel `Monoid` instance
- Add `Reducible` for `Tuple2`

I might have gone a bit overboard with overriding in the "one element `Foldable`" instances (`Option`, `Either`, ...).

### DIFF
--- a/core/src/main/scala/cats/Reducible.scala
+++ b/core/src/main/scala/cats/Reducible.scala
@@ -114,7 +114,7 @@ import simulacrum.typeclass
    * Traverse `F[A]` using `Apply[G]`.
    *
    * `A` values will be mapped into `G[B]` and combined using
-   * `Applicative#map2`.
+   * `Apply#map2`.
    *
    * This method is similar to [[Foldable.traverse_]]. There are two
    * main differences:
@@ -176,6 +176,16 @@ import simulacrum.typeclass
       case NonEmptyList(hd, tl) =>
         Reducible[NonEmptyList].reduce(NonEmptyList(hd, a :: intersperseList(tl, a)))
     }
+
+  override def isEmpty[A](fa: F[A]): Boolean = false
+
+  override def nonEmpty[A](fa: F[A]): Boolean = true
+
+  override def minimumOption[A](fa: F[A])(implicit A: Order[A]): Option[A] =
+    Some(minimum(fa))
+
+  override def maximumOption[A](fa: F[A])(implicit A: Order[A]): Option[A] =
+    Some(maximum(fa))
 }
 
 /**
@@ -210,4 +220,55 @@ abstract class NonEmptyReducible[F[_], G[_]](implicit G: Foldable[G]) extends Re
         case None => Later(f(a))
       }
     }
+
+  override def size[A](fa: F[A]): Long = {
+    val (_, tail) = split(fa)
+    1 + G.size(tail)
+  }
+
+  override def fold[A](fa: F[A])(implicit A: Monoid[A]): A = {
+    val (a, ga) = split(fa)
+    A.combine(a, G.fold(ga))
+  }
+
+  override def find[A](fa: F[A])(f: A => Boolean): Option[A] = {
+    val (a, ga) = split(fa)
+    if (f(a)) Some(a) else G.find(ga)(f)
+  }
+
+  override def exists[A](fa: F[A])(p: A => Boolean): Boolean = {
+    val (a, ga) = split(fa)
+    p(a) || G.exists(ga)(p)
+  }
+
+  override def forall[A](fa: F[A])(p: A => Boolean): Boolean = {
+    val (a, ga) = split(fa)
+    p(a) && G.forall(ga)(p)
+  }
+
+  override def toList[A](fa: F[A]): List[A] = {
+    val (a, ga) = split(fa)
+    a :: G.toList(ga)
+  }
+
+  override def toNonEmptyList[A](fa: F[A]): NonEmptyList[A] = {
+    val (a, ga) = split(fa)
+    NonEmptyList(a, G.toList(ga))
+  }
+
+  override def filter_[A](fa: F[A])(p: A => Boolean): List[A] = {
+    val (a, ga) = split(fa)
+    val filteredTail = G.filter_(ga)(p)
+    if (p(a)) a :: filteredTail else filteredTail
+  }
+
+  override def takeWhile_[A](fa: F[A])(p: A => Boolean): List[A] = {
+    val (a, ga) = split(fa)
+    if (p(a)) a :: G.takeWhile_(ga)(p) else Nil
+  }
+
+  override def dropWhile_[A](fa: F[A])(p: A => Boolean): List[A] = {
+    val (a, ga) = split(fa)
+    if (p(a)) G.dropWhile_(ga)(p) else a :: G.toList(ga)
+  }
 }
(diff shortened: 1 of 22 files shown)


core/src/main/scala/cats/Reducible.scala
```diff
@@ -210,4 +220,55 @@ abstract class NonEmptyReducible[F[_], G[_]](implicit G: Foldable[G]) extends Re
...
+    val (a, ga) = split(fa)
+    p(a) && G.forall(ga)(p)
+  }
+
+  override def toList[A](fa: F[A]): List[A] = {
+    val (a, ga) = split(fa)
+    a :: G.toList(ga)
+  }
+
+  override def toNonEmptyList[A](fa: F[A]): NonEmptyList[A] = {
```
johnynek: this method code path is untested

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -251,6 +250,21 @@ private[data] sealed trait NonEmptyVectorInstances {
         NonEmptyVector.fromVectorUnsafe(buf.result())
       }
 
+      override def fold[A](fa: NonEmptyVector[A])(implicit A: Monoid[A]): A =
+        fa.reduce
+
+      override def foldM[G[_], A, B](fa: NonEmptyVector[A], z: B)(f: (B, A) => G[B])(implicit G: Monad[G]): G[B] =
+        Foldable.iteratorFoldM(fa.toVector.toIterator, z)(f)
```
johnynek: code path is untested.

core/src/main/scala/cats/data/Validated.scala
```diff
@@ -323,7 +324,39 @@ private[data] sealed abstract class ValidatedInstances extends ValidatedInstance
           case v @ Validated.Valid(_) => v
         }
       def raiseError[A](e: E): Validated[E, A] = Validated.Invalid(e)
+
+      override def reduceLeftToOption[A, B](fa: Validated[E, A])(f: A => B)(g: (B, A) => B): Option[B] =
```
johnynek: next 5 methods untested.

core/src/main/scala/cats/data/Validated.scala
```diff
@@ -323,7 +324,39 @@ private[data] sealed abstract class ValidatedInstances extends ValidatedInstance
...
+        fa.toOption.filter(f)
+
+      override def exists[A](fa: Validated[E, A])(p: A => Boolean): Boolean =
+        fa.exists(p)
+
+      override def forall[A](fa: Validated[E, A])(p: A => Boolean): Boolean =
+        fa.forall(p)
+
+      override def toList[A](fa: Validated[E, A]): List[A] =
+        fa.fold(_ => Nil, _ :: Nil)
```
johnynek: untested

core/src/main/scala/cats/instances/either.scala
```diff
@@ -83,12 +83,47 @@ trait EitherInstances extends cats.kernel.instances.EitherInstances {
...
       override def recover[B](fab: Either[A, B])(pf: PartialFunction[A, B]): Either[A, B] =
         fab recover pf
+
       override def recoverWith[B](fab: Either[A, B])(pf: PartialFunction[A, Either[A, B]]): Either[A, B] =
         fab recoverWith pf
+
       override def ensure[B](fab: Either[A, B])(error: => A)(predicate: B => Boolean): Either[A, B] =
         fab.ensure(error)(predicate)
+
+      override def reduceLeftToOption[B, C](fab: Either[A, B])(f: B => C)(g: (C, B) => C): Option[C] =
```
johnynek: next 5 method untested.

core/src/main/scala/cats/instances/either.scala
```diff
@@ -83,12 +83,47 @@ trait EitherInstances extends cats.kernel.instances.EitherInstances {
...
+        fab.fold(_ => None, r => if (f(r)) Some(r) else None)
+
+      override def exists[B](fab: Either[A, B])(p: B => Boolean): Boolean =
+        fab.right.exists(p)
+
+      override def forall[B](fab: Either[A, B])(p: B => Boolean): Boolean =
+        fab.right.forall(p)
+
+      override def toList[B](fab: Either[A, B]): List[B] =
+        fab.fold(_ => Nil, _ :: Nil)
```
johnynek: untested.

core/src/main/scala/cats/instances/option.scala
```diff
@@ -71,12 +71,46 @@ trait OptionInstances extends cats.kernel.instances.OptionInstances {
       override def filter[A](fa: Option[A])(p: A => Boolean): Option[A] =
         fa.filter(p)
 
+      override def reduceLeftToOption[A, B](fa: Option[A])(f: A => B)(g: (B, A) => B): Option[B] =
```
johnynek: next 5 methods untested.

core/src/main/scala/cats/instances/option.scala
```diff
@@ -71,12 +71,46 @@ trait OptionInstances extends cats.kernel.instances.OptionInstances {
...
+
       override def exists[A](fa: Option[A])(p: A => Boolean): Boolean =
         fa.exists(p)
 
       override def forall[A](fa: Option[A])(p: A => Boolean): Boolean =
         fa.forall(p)
 
+      override def toList[A](fa: Option[A]): List[A] = fa.toList
+
+      override def filter_[A](fa: Option[A])(p: A => Boolean): List[A] =
```
johnynek: next 5 methods untested.

peterneyens:
Improved the test coverage, only the methods `reduceLeftToOption` and `reduceRightToOption` are not yet tested.

core/src/main/scala/cats/instances/list.scala
```diff
@@ -86,6 +86,19 @@ trait ListInstances extends cats.kernel.instances.ListInstances {
...
+
+      override def find[A](fa: List[A])(f: A => Boolean): Option[A] = fa.find(f)
+
+      override def filter_[A](fa: List[A])(p: A => Boolean): List[A] = fa.filter(p)
+
+      override def takeWhile_[A](fa: List[A])(p: A => Boolean): List[A] = fa.takeWhile(p)
+
+      override def dropWhile_[A](fa: List[A])(p: A => Boolean): List[A] = fa.dropWhile(p)
+
+      override def algebra[A]: Monoid[List[A]] = new kernel.instances.ListMonoid[A]
```
kailuowang: should it be `val` or `lazy val`?

kailuowang:
I think the code is reasonable, especially now that we have full test coverage. 
I just want to share my thoughts. To me, whether to override a default implementation provided by the type class is a trade-off between performance and complexity. The complexity cost comes in in the sense of extra code to maintain and test against, and in some cases, extra laws.  Very often the performance gain outweighs the complexity cost. I feel that it's a pity we don't have a really good gauge on some of the performance gain here, e.g. the one element `Foldable`s. On the other hand, the complexity cost isn't very substantial either. 
For now, I think this PR is good for my 👍. 
Hopefully, in the future, we'll have some benchmarking to give us more data when making such performance/complexity trade-off decisions.

peterneyens:
I concur that some benchmarking would be nice to make these performance/complexity decisions.

kailuowang:
merging with two thumb ups.

Issue #1539, opened 2017-02-19
Adding Ior syntax to cats.syntax package object


leandrob13:
I had a practical use of Ior recently and I had the need of some useful syntax like there is for validated or option. The purpose of this would be to have useful syntax for Ior as well to add into other syntaxes functions `leftIor`, `rightIor` or `both` in validated, option and either syntaxes.

Is this something that would be worth working on? I might have an example to add to the tut docs.

peterneyens:
Added in #1540.

PR #1540, opened 2017-02-19
Ior syntax

leandrob13:
Added syntax for Ior like mentioned on issue #1539 

This pull request includes ior syntax support for:

 - Validated
 - Option
 - Ior


core/src/main/scala/cats/syntax/ior.scala
```diff
@@ -0,0 +1,158 @@
...
+    * Example:
+    * {{{
+    * scala> import cats.data.Ior
+    * scala> import cats.implicits._
+    *
+    * scala> "hello".rightIor[String]
+    * res0: Ior[String, String] = Right(hello)
+    * }}}
+    */
+  def rightIor[B]: Ior[B, A] = Ior.right(a)
```
ChristopherDavenport: Context in Either instances uses `asRight` so perhaps `asRightIor`. Depends on what sort of syntax maintainers would like to see but this makes more sense to me.

tests/src/test/scala/cats/tests/ValidatedTests.scala
```diff
@@ -187,6 +187,18 @@ class ValidatedTests extends CatsSuite {
     }
   }
 
+  test("fromIor consistent with Ior.toValidated"){
+    forAll { (i: Ior[String, Int], s: String) =>
```
leandrob13: @ChristopherDavenport  That just slipped, I will remove it. Thanks!

core/src/main/scala/cats/data/package.scala
```diff
@@ -3,6 +3,7 @@ package cats
 package object data {
   type NonEmptyStream[A] = OneAnd[Stream, A]
   type ValidatedNel[+E, +A] = Validated[NonEmptyList[E], A]
+  type IorNel[+B, +A] = Ior[NonEmptyList[B], NonEmptyList[A]]
```
leandrob13: I am not sure about this definition now. Maybe it should be similar to ValidatedNel

```scala
type IorNel[+B, +A] = Ior[NonEmptyList[B], A]
```

I decided to define it with a NonEmptyList on the right also because I saw some utility on the combine case (or use of Ior's append function to be exact). But when comparing the combine for Either and Validated I think I got it wrong. Ior's flatmap does accumulate errors only if the case is Ior.Both, so it seems appropiate to leave it similar to ValidatedNel's type definition. I would love some feedback on this.

edmundnoble: Personally I cannot see a use-case for `IorNel` as it was before; did you have one in mind? That might help to decide.

leandrob13: @edmundnoble I am going to change it to `type IorNel[+B, +A] = Ior[NonEmptyList[B], A]` to keep consistency with ValidatedNel. I do have a use case for my original proposal but that is when you get to combine Iors with the |+| operator. My main argument to change it is that in the flatMap you get tu accumulate errors on the left side of `Both` so it is the natural usage of it.

core/src/main/scala/cats/syntax/ior.scala
```diff
@@ -0,0 +1,163 @@
...
+    * Example:
+    * {{{
+    * scala> import cats.data.Ior
+    * scala> import cats.implicits._
+    *
+    * scala> "hello".putRightIor("error")
+    * res0: Ior[String, String] = Both(error,hello)
+    * }}}
+    */
+  def putRightIor[B](left: B): Ior[B, A] = Ior.both(left, a)
```
kailuowang: I am not sure about the necessity of these syntax: 
`putRightIor` `putLeftIor` `rightIorNel` `leftIorNel` `putRightIorNel` `putLeftIorNel`
I am just not sure if the convenience of these methods overrides the cost of having them on everything.

core/src/main/scala/cats/syntax/ior.scala
```diff
@@ -0,0 +1,163 @@
...
+    *
+    * scala> "I got it wrong".putLeftIorNel[String]("hello")
+    * res0: IorNel[String, String] = Both(NonEmptyList(I got it wrong),hello)
+    * }}}
+    */
+  def putLeftIorNel[B](right: B): IorNel[A, B] = Ior.bothNel(a, right)
+}
+
+
+final class IorNelListOps[A, B](val list: List[IorNel[A, B]]) extends AnyVal {
```
kailuowang: I think that this can be further abstracted. I am also not sure having syntax for such a specific type.

core/src/main/scala/cats/syntax/list.scala
```diff
@@ -1,12 +1,71 @@
...
+    * scala> val result1: List[String] = List("error 1", "error 2")
+    * scala> result1.toIorNel(1)
+    * res0: IorNel[String, Int] = Left(NonEmptyList(error 1, error 2))
+    *
+    * scala> val result2: List[String] = List.empty[String]
+    * scala> result2.toIorNel(1)
+    * res1: IorNel[String, Int] = Right(1)
+    * }}}
+    */
+  def toIorNel[B](ifEmpty: => B): IorNel[A, B] =
```
kailuowang: The intention of this isn't very clear. I mean if you read a code `result1.toIorNel(1)`, it's not obvious that the `result1` is the error list while `1` is the valid result. Users will have to read the code here to understand the intention.  I am not sure about such syntax addition.

core/src/main/scala/cats/syntax/list.scala
```diff
@@ -1,12 +1,71 @@
...
+    * scala> result1.toValidatedNel(1)
+    * res0: ValidatedNel[String, Int] = Invalid(NonEmptyList(error 1, error 2))
+    *
+    * scala> val result2: List[String] = List.empty[String]
+    * scala> result2.toValidatedNel(1)
+    * res1: ValidatedNel[String, Int] = Valid(1)
+    * }}}
+    */
+  def toValidatedNel[B](ifEmpty: => B): ValidatedNel[A, B] =
+    toNel.fold[ValidatedNel[A, B]](Valid(ifEmpty))(Invalid(_))
```
kailuowang: Same as above.

kailuowang:
thanks very much @leandrob13, I like most of the PR, there are some syntax that I am not sure about. But I am more a minimalist when it comes to API design. So if other people see values in those syntax that I am not sure about I can be persuaded. Or if you want to separate those to another PR I'll gladly give my thumbs up on this one.

core/src/main/scala/cats/syntax/ior.scala
leandrob13: @kailuowang in this case, I wanted to offer in the syntax the put functions that helped build Both type just for convenience. I really found useful the invalidNel syntax of validated so I wanted to be able to use something similar when I was working with Ior, that is why I added rightIorNel-leftIorNel. This means that you also object to Ior.rightNel-Ior.leftNel functions on the companion object?

core/src/main/scala/cats/syntax/list.scala
leandrob13: @kailuowang you are right, I didn't make it clear. This is motivated due to a use case I find frequently. If there is some validations made and were returned as a List, meaning that if it is empty everything went well and if it is not then there are errors present, I had to do something like this:

```scala
errorList match {
  case Nil => Valid("this is valid")
  case err :: errs => Invalid(NoneEmptyList(err, errs))
}
```


I found convenient to have a syntax that helped me with this for ValidatedNel and IorNel.

I do agree with you that if there is not consensus on this I can roll it back and maybe try it again on another pull request when I can make an example illustrating the use of what I proposed. I'll wait for your thoughts on this.

leandrob13:
@kailuowang I agree with you, my specific comments are above. Waiting for your feedback to decide what to do next.

leandrob13:
@kailuowang Thanks! I will follow your advice and prepare another PR to illustrate better the proposals I made. I was actually thinking of a reduceable for the list of iors reduce proposal and thanks to your hint I might work on that too.

peterneyens:
Merging with two :+1:

PR #1541, opened 2017-02-21
Switch from CrossVersion.full to CrossVersion.patch for TLS compatibi…

milessabin:
…lity.

To use Typelevel Scala versions which are not exactly aligned with the corresponding Lightbend Scala version (ie. because they include and additional `-bin-patch-nnn` suffix in their version) we need to modify project builds which use `CrossVersion.full` (which includes the suffix) to use `CrossVersion.patch` (which doesn't) where appropriate (eg. for macro-paradise and other compiler plugins).

See https://github.com/typelevel/scala/issues/135


johnynek:
:+1:

kailuowang:
:+1:

PR #1543, opened 2017-02-26
MonadError instance for Kleisli

### LINKED ISSUE #1542: MonadError instance for Kleisli
Is there any reason for not having a `MonadError` instance for `Kleisli`? Since `MonadError` seems to be simply `Monad` and `ApplicativeError`, and `Kleisli` has instances of both, it seems like it could work. Although prioritizing the instance could be a little tricky, I guess...

durban:
Fixes #1542.

### DIFF
--- a/core/src/main/scala/cats/data/Kleisli.scala
+++ b/core/src/main/scala/cats/data/Kleisli.scala
@@ -118,6 +118,29 @@ private[data] sealed abstract class KleisliInstances extends KleisliInstances0 {
 }
 
 private[data] sealed abstract class KleisliInstances0 extends KleisliInstances1 {
+
+  implicit def catsDataMonadErrorForKleisli[F[_], A, E](
+    implicit me: MonadError[F, E]
+  ): MonadError[Kleisli[F, A, ?], E] with MonadReader[Kleisli[F, A, ?], A] = {
+    new KleisliApplicativeError[F, A, E]
+        with KleisliMonadReader[F, A]
+        with MonadError[Kleisli[F, A, ?], E] {
+      override def M: Monad[F] = me
+      override def AF: ApplicativeError[F, E] = me
+    }
+  }
+}
+
+private[data] sealed abstract class KleisliInstances1 extends KleisliInstances2 {
+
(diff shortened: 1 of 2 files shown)


kailuowang:
`MonadError` is more specific than `ApplicativeError` and thus should be placed in higher priority, no?

durban:
I'm not aware of such a rule (and if there is one, it is already broken by `Applicative` and `MonadReader` for `Kleisli`). Is this documented somewhere? Can it cause any problems having these with a lower priority?

kailuowang:
@durban, you are right, I believe `MonadReader` is in the wrong priority according to this rule (mind to fix this on the way as well?). I don't think this rule is documented anywhere here, it's not a cats specific rule, it's more a best practice to my knowledge. My understanding is that this pratice has two benefits: 
1) having an implicit sequence is better not having one.
2) if you assign the most specific one with the highest priority, it takes less implicit search, so it's slightly more compilation efficient.  And sometimes the most specific (and thus powerful) instance is also more performant.
I might've missed other benefits.

core/src/main/scala/cats/data/Kleisli.scala
```diff
@@ -275,6 +279,20 @@ private trait KleisliApplicativeError[F[_], A, E] extends KleisliApplicative[F,
...
+
+  def flatMap[B, C](fa: Kleisli[F, A, B])(f: B => Kleisli[F, A, C]): Kleisli[F, A, C] =
+    fa.flatMap(f)
+
+  def map[B, C](fa: Kleisli[F, A, B])(f: B => C): Kleisli[F, A, C] =
+    fa.map(f)
+
+  def tailRecM[B, C](b: B)(f: B => Kleisli[F, A, Either[B, C]]): Kleisli[F, A, C] =
+    Kleisli[F, A, C]({ a => FlatMap[F].tailRecM(b) { f(_).run(a) } })
+}
```
peterneyens: Like this we have the some code twice:

- here in `KleisliFlatMap`
- and in `catsDataMonadReaderForKleisli`

Maybe we could create a `KleisliMonad` which can be extended by a `KleisliMonadReader` and a `KleisliMonadError` (similar as how it is done in `StateT`) ?

durban:
@kailuowang @peterneyens I've amended my commit to address your comments. (I've also rebased to the current master.)

PR #1545, opened 2017-03-01
fix scaladoc for Eval methods Unit, True, False, Zero, One

sullivan-:


johnynek:
👍

core/src/main/scala/cats/Eval.scala
```diff
@@ -206,15 +206,43 @@ object Eval extends EvalInstances {
     new Eval.Call[A](a _) {}
 
   /**
-   * Static Eval instances for some common values.
+   * Static Eval instances for common value `Unit`.
```
kailuowang: `instances` -> `instance`

sullivan-: oops! fixed it

kailuowang:
Thanks very much！

PR #1546, opened 2017-03-01
add to Monad ifM example

sullivan-:
I found this section pretty confusing, but cleared up my misunderstanding quickly in the console.
This extra line in the example might help make it clear to people what is going on.

### DIFF
--- a/docs/src/main/tut/typeclasses/monad.md
+++ b/docs/src/main/tut/typeclasses/monad.md
@@ -111,7 +111,7 @@ statement into the monadic context.
 ```tut:book
 import cats.implicits._
 
-Monad[List].ifM(List(true, false, true))(List(1, 2), List(3, 4))
+Monad[List].ifM(List(true, false, true))(ifTrue = List(1, 2), ifFalse = List(3, 4))
 ```
 
 ### Composition


docs/src/main/tut/typeclasses/monad.md
```diff
@@ -112,6 +112,8 @@ statement into the monadic context.
 import cats.implicits._
 
 Monad[List].ifM(List(true, false, true))(List(1, 2), List(3, 4))
+
+Monad[List].ifM(List(true, false, false))(List(1, 2), List(3, 4))
```
jozic: how about also adding names to parameters?
like 

`Monad[List].ifM(List(true, false, true))(ifTrue = List(1, 2), ifFalse = List(3, 4))`

sullivan-: If we did that I think we could leave out the extra example. I will add parameter names. Do you want me to pull the second example?

sullivan-:
I added param names, left the second example in. happy to adjust further.

I did a `push -f` here so a merge would only bring in one commit

docs/src/main/tut/typeclasses/monad.md
```diff
@@ -111,7 +111,9 @@ statement into the monadic context.
 ```tut:book
 import cats.implicits._
 
-Monad[List].ifM(List(true, false, true))(List(1, 2), List(3, 4))
+Monad[List].ifM(List(true, false, true))(ifTrue = List(1, 2), ifFalse = List(3, 4))
+
+Monad[List].ifM(List(true, false, false))(ifTrue = List(1, 2), ifFalse = List(3, 4))
```
kailuowang: I agree that with the parameter name, the second example looks redundant.

jozic: tend to agree, parameter names make it clear imo
i also was confused by the original example

sullivan-: thanks, i got rid of the extra line

johnynek:
👍

PR #1547, opened 2017-03-01
fix ref to non-existent dir in contributing

sullivan-:
it seems as if there are no static pages any more in the website.

I added a path to the menu file instead, as it might help to give people a lead on how the menus are
constructed.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -171,8 +171,8 @@ Write about checking laws
 ### source for the documentation
 The documentation for this website is stored alongside the source, in the [docs subproject](https://github.com/typelevel/cats/tree/master/docs).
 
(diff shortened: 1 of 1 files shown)


kailuowang:
👍  thanks!

vxzfbhgfhjgdhjd:
@tel

vxzfbhgfhjgdhjd:
fix ref to non-existent dir in contributing
it seems as if there are no static pages any more in the website.

I added a path to the menu file instead, as it might help to give people a lead on how the menus are
constructed.

vxzfbhgfhjgdhjd:
vasyagufik900@mail.ru

sullivan-:
Something about this merge looks a little suspicious to me...

PR #1548, opened 2017-03-08
MonadError instance for Ior

### LINKED ISSUE #1544: MonadError instance for Ior
I noticed there is no MonadError instance for Ior. Is there a reason for this?

leandrob13:
Fixes #1544 and #1553


core/src/main/scala/cats/data/Ior.scala
```diff
@@ -153,47 +153,64 @@ private[data] sealed abstract class IorInstances extends IorInstances0 {
...
-          fn(b) match {
-            case Ior.Left(aa) => Ior.left(Semigroup[A].combine(a, aa))
-            case Ior.Both(aa, x) => loop(Ior.both(Semigroup[A].combine(a, aa), x))
-            case Ior.Right(x) => loop(Ior.both(a, x))
-          }
+  // scalastyle:off method.length
+  implicit def catsDataMonadErrorForIor[A: Semigroup]: MonadError[Ior[A, ?], A] with Traverse[Ior[A, ?]] =
+    new MonadError[Ior[A, ?], A] with Traverse[Ior[A, ?]] {
+
+      def raiseError[B](e: A): Ior[A, B] = Ior.left(e)
```
kailuowang: This line is not tested

core/src/main/scala/cats/data/Ior.scala
```diff
@@ -153,47 +153,64 @@ private[data] sealed abstract class IorInstances extends IorInstances0 {
...
+
+      def handleErrorWith[B](fa: Ior[A, B])(f: (A) => Ior[A, B]): Ior[A, B] =
+        fa match {
+          case Ior.Left(e) => f(e)
```
kailuowang: This line is not tested

core/src/main/scala/cats/data/Ior.scala
```diff
@@ -153,47 +153,64 @@ private[data] sealed abstract class IorInstances extends IorInstances0 {
...
+          case r @ Ior.Right(_) => r
+          case Ior.Both(e, _) => f(e)
```
kailuowang: This line is not tested

leandrob13: @kailuowang Thanks for pointing them out. I will get it fixed this week.

core/src/main/scala/cats/data/Ior.scala
```diff
@@ -153,47 +153,64 @@ private[data] sealed abstract class IorInstances extends IorInstances0 {
...
 
   implicit def catsDataBifunctorForIor: Bifunctor[Ior] =
     new Bifunctor[Ior] {
       override def bimap[A, B, C, D](fab: A Ior B)(f: A => C, g: B => D): C Ior D = fab.bimap(f, g)
     }
 }
 
-private[data] sealed abstract class IorInstances0 {
-
-  implicit def catsDataTraverseFunctorForIor[A]: Traverse[A Ior ?] with Functor[A Ior ?] = new Traverse[A Ior ?] with Functor[A Ior ?] {
```
kailuowang: not sure why you decided to merge this into the `MonadError` instance, but this one does not require `A` having a `Semigroup` instance. So if we don't have good reason, we should probably keep as is.

leandrob13: Wouldn't I need it when I define a type like this:

```scala
type IorNel[A, B] = Ior[NonEmptyList[A], B]
```

If I try to use the traverse instance? I saw the mixin of Traverse and ApplicativeError on Validated so I thought it would be equivalent, or in this case is not relevant? I am considering this since I already have a PR #1540 with a proposition on that IorNel.

johnynek: +1 to keep. Traverse should not require a Semigroup.

kailuowang: @leandrob13 for a `IorNel` whether to require the left side being a `Semigroup` is irrelevant because the left side `NonEmptyList` is one. But there are other types of `Ior` in which the left side is not a `Semigroup`. We don't want to add unnecessary requirements/restriction to our instances. 
I think the mixin in `Validated` is not optimal either in the exact same sense. The `Traverse` and `Functor` shouldn't need a `Semigroup`. Care to fix that one as well?

leandrob13: @kailuowang understood! I will fix the Validated instance as well. Thanks for the feedback.

tests/src/test/scala/cats/tests/IorTests.scala
```diff
@@ -63,6 +65,12 @@ class IorTests extends CatsSuite {
     }
   }
 
+  test("valueOr consistent with leftMap") {
+    forAll { (i: Int Ior String, f: Int => String) =>
+      i.valueOr(f) should === (i.leftMap(f).fold(identity, identity, _ + _))
+    }
+  }
+
```
kailuowang: Thanks for adding coverage.

kailuowang:
👍  LKTM, thanks very much @leandrob13 for all the efforts to address the feedbacks!

leandrob13:
@kailuowang you are welcome. I have learned a lot.

core/src/main/scala/cats/data/Validated.scala
```diff
@@ -357,28 +388,6 @@ private[data] sealed abstract class ValidatedInstances extends ValidatedInstance
 
       override def isEmpty[A](fa: Validated[E, A]): Boolean = fa.isInvalid
     }
-    // scalastyle:on method.length
```
peterneyens: We may want to keep this `// scalastyle:off method.length` comment.

At the same time I think we can remove these comments around the `ApplicativeError` instance (`catsDataApplicativeErrorForValidated`).

leandrob13: @peterneyens I'm on it!

leandrob13: @peterneyens done with requested changes.

tests/src/test/scala/cats/tests/IorTests.scala
```diff
@@ -1,11 +1,11 @@
 package cats
 package tests
 
-import cats.data.{EitherT, Ior}
 import cats.kernel.laws.GroupLaws
 import cats.laws.discipline.{BifunctorTests, CartesianTests, MonadErrorTests, SerializableTests, TraverseTests}
+import cats.data.{Ior, NonEmptyList, EitherT}
+import cats.kernel.laws.GroupLaws
```
peterneyens: You had this import already on line 4 it seems.

leandrob13: @peterneyens darn it. That one must have slipped when I merged. I'm on it.

Issue #1277, opened 2016-08-09
Semantics of `Free.suspend` changed in 0.7.0


mpilquist:
I'm not sure if this is an issue or not but we changed semantics of Free.suspend in 0.7:

```scala
scala> val x = Free.suspend { println("Computing"); Free.pure(()) }
x: cats.free.Free[S,Unit] = Free(...)

scala> x.step
Computing
res0: cats.free.Free[S,Unit] = Free(...)
````

Whereas in 0.6.1:

```scala
@ val x = Free.suspend[List,Unit] { println("Computing"); Free.pure(()) }
x: Free[List, Unit] = Gosub(Suspend(List(())),<function1>)
@ x.step
res3: Free[List, Unit] = Gosub(Suspend(List(())),<function1>)
```

This is because step [eagerly evaluates `FlatMapped(Pure(x), f)` when stepping](https://github.com/typelevel/cats/blob/97fc7625923076180a5a2dd7f03798882c85c26d/free/src/main/scala/cats/free/Free.scala#L39), which used to be safe in presence of `suspend`, b/c [`suspend` included a call to `liftF` and hence, a `Suspend` constructor](https://github.com/typelevel/cats/blob/v0.6.1/free/src/main/scala/cats/free/Free.scala#L28).

I was bitten by this same bug in FS2, where suspend has always been defined as `pure(()).flatMap { _ => fa }`. In FS2, I removed the eager evaluation of `FlatMapped(Pure(x), f)` b/c it impacts semantics of effect capture and that's half of what FS2 is about. Not sure what the right thing for Cats is though...

Some options:

(1) Restore old definition, which means suspend requires an `Applicative[F]` for pure
(2) Define both constructors, one which is guaranteed not to be evaluated while stepping (names TBD)
(3) Remove the eager application of `FlatMapped(Pure(x), f)` from step
(4) Ignore this since it is only visible in presence of side effects

Does anyone know the genesis of the eager evaluation in `step` (maybe @d6)? As far as I can tell, this is a performance enhancement to avoid some trampolining.

There was [some discussion on Gitter](https://gitter.im/typelevel/cats?at=57a8fa6d46610f173949bf46) that may be interesting. 

/cc @tpolecat @adelbertc

mikejcurry:
@mpilquist - I think you meant @non as opposed to `d6`? - apologies if I got that wrong.

mpilquist:
@mikejcurry Indeed, thanks, and my apologies to `d6`. :)

[...]

PR #1552, opened 2017-03-09
Issue #1316: Add the MonadDefer type-class

### LINKED ISSUE #1316: Bring back `Applicative.pureEval` or provide equivalent
I see that `pureEval` was removed in [this commit](https://github.com/typelevel/cats/commit/94fcf03f774841806add287a8d5f43e6e1bbc6d2) and deployed as part of version `0.7.0`.

The explanation is fine, however it needs to be brought back or it needs an equivalent. There are two reasons for it:
1. the [point operation in Scalaz](https://github.com/scalaz/scalaz/blob/series/7.3.x/core/src/main/scala/scalaz/Applicative.scala#L25) takes that parameter by name, which means Cats is no longer providing an equivalent to Scalaz's `point`
2. I understand there were concerns about controlling of side-effects, but that's not the whole story. A lazy `pure` can also be used for actual lazy behavior, like when building a lazy generic data-structure; think of `Streaming` (that was moved in Dogs) or checkout this [Enumerator](https://github.com/monixio/monix/blob/1bcab7f/monix-eval/shared/src/main/scala/monix/eval/Enumerator.scala#L202) type that I've been building. In other words, without a lazy `pure`, the `Applicative` type becomes insufficient for problems such as building a generic lazy list like `Cons[F,A](head: A, tail: F[List[A]])` type

alexandru:
This pull request is about issue #1316. We are adding one new type class:

`MonadDefer`: for monadic types which are necessarily for lazy, as it *repeats side-effects* on each evaluation and have a stack safe `flatMap`. Implementing types: `cats.Eval`, `IO`, `Task`, etc.

This is needed for abstracting over applicatives and monads that can capture effects (e.g. `Eval`, `Task`) and libraries are doing their own thing:

1. FS2 has [Suspendable](https://github.com/functional-streams-for-scala/fs2/blob/series/1.0/core/shared/src/main/scala/fs2/util/Suspendable.scala)
2. Doobie has [Capture](https://github.com/tpolecat/doobie/blob/v0.4.1/yax/core/src/main/scala/doobie/util/capture.scala)
3. Monix is [evolving](https://github.com/monix/monix/blob/wip-streams/monix-types/shared/src/main/scala/monix/types/Applicative.scala#L52) to just do `pure(()).map(_ => a)` and `pure(()).flatMap(_ => fa)` and hope for the best

Please checkout the described laws.

**UPDATE:** Originally this PR was introducing a second type class which was removed, leaving the description here for historical purposes:

`ApplicativeEval`, which would add an `eval` operation that lifts values and evaluates side-effects in the `F[_]` context and is *optionally* lazy. This type-class does not require implementing types to be lazy, therefore it can be implemented for eager evaluation as well: `Try`, `Future`, etc.

See the below discussion for why `ApplicativeEval` was removed.


kailuowang:
I will dive into the code later, but meanwhile, do you think adding a brief doc here https://github.com/typelevel/cats/tree/master/docs/src/main/tut/typeclasses with some example use cases to demonstrate the motivation would be helpful? 
Of course that doesn't necessarily have to happen in this PR.

tpolecat:
@alexandru will you be at NEScala? I'd like to get a group together to discuss this. I don't think it should be part of cats because fs2 has its own and scalaz doesn't have one. Would be nice for this to be its own thing so we can all use it once and for all.

core/src/main/scala/cats/ApplicativeEval.scala
```diff
@@ -0,0 +1,25 @@
...
+   */
+  def eval[A](a: Eval[A]): F[A]
+
+  /**
+   * Lifts the given by-name value in the `F[_]` context, with
+   * optional laziness.
+   *
+   * Alias for `eval(always(a))`.
+   */
+  final def delay[A](a: => A): F[A] = eval(always(a))
```
oscar-stripe: should this be `final`? Many types, have their own implementation of this.

core/src/main/scala/cats/ApplicativeEval.scala
```diff
@@ -0,0 +1,25 @@
...
+ * A type class that allows lifting any value into the applicative
+ * context, with its evaluation being controlled by [[Eval]] and
+ * supporting optional laziness.
+ */
+@typeclass trait ApplicativeEval[F[_]] extends Applicative[F] {
+  /**
+   * Lifts any value into the `F[_]` applicative context, where the
+   * evaluation is controlled by [[Eval]] and can be optionally lazy.
+   */
+  def eval[A](a: Eval[A]): F[A]
```
oscar-stripe: what about `fromEval`?

laws/src/main/scala/cats/laws/ApplicativeEvalLaws.scala
```diff
@@ -0,0 +1,24 @@
+package cats
+package laws
+
+import cats.Eval.{always, now}
+import cats.syntax.all._
+
+trait ApplicativeEvalLaws[F[_]] extends ApplicativeLaws[F] {
+  implicit override def F: ApplicativeEval[F]
+
```
oscar-stripe: don't we need something like:
```scala
var i = 0
val asF = F.eval(later { i += 1; i })
i == 0 // we have not incremented i yet.
```

alexandru:
@tpolecat FS2 has its own [Monad](https://github.com/functional-streams-for-scala/fs2/blob/series/1.0/core/shared/src/main/scala/fs2/util/Monad.scala) too and this `MonadDefer` is basically their `Suspendable`. In building this, I took a look at both FS2 and Doobie. Should be in Cats because it is foundational IMO.

Yes, I'll probably be at NEScala - tried to submit a talk too, but some fine folks came ahead of me :-)

laws/src/main/scala/cats/laws/MonadDeferLaws.scala
```diff
@@ -0,0 +1,62 @@
...
+
+  def evalSequenceConsistentWithPureMap[A, B](a: A, b: B, f: (A, B) => A): IsEq[F[A]] = {
+    def tr(s: StatefulBox[A]) = s.transform(a => f(a, b))
+    val state1 = new StatefulBox(a)
+    val state2 = new StatefulBox(a)
+
+    val lh = F.delay(tr(state1))
+    val rh = F.pure(state2).map(tr)
+
+    lh.flatMap(_ => lh) <-> rh.flatMap(_ => rh)
```
oscar-stripe: can you comment on the intent of running these effects each twice and comparing them? I'd rather have an effectful thing on one side, and a non-effectful thing on the other so I can more easily read intent.

tests/src/test/scala/cats/tests/TryTests.scala
```diff
@@ -25,6 +24,9 @@ class TryTests extends CatsSuite {
   checkAll("Try", MonadTests[Try].monad[Int, Int, Int])
   checkAll("Monad[Try]", SerializableTests.serializable(Monad[Try]))
 
+  checkAll("Try[Int]", ApplicativeEvalTests[Try].applicativeEvalWithError[Int, Int, Int])
+  checkAll("ApplicativeEval[Try]", SerializableTests.serializable(ApplicativeEval[Try]))
```
oscar-stripe: can we add a negative test to show that `ApplicativeEval[Try]` and `MonadDefer[Try]` will fail the laws?

I think Try should not be lawful if this typeclass is really representing anything. Since any `Applicative` can do it if `Try` can, no?

tpolecat:
@alexandru but fs2 is not going to add a cats dependency, which means there will need to be another shim to support this. It's propagating the exact problem it's trying to solve.

core/src/main/scala/cats/ApplicativeEval.scala
alexandru: `eval` and `delay` are redundant, imo we should have either one or the other. I added `delay` basically for familiarity purposes.

oscar-stripe: yeah, but it is a shame to have to box to `Eval` just to get a method `Task` already has.

laws/src/main/scala/cats/laws/ApplicativeEvalLaws.scala
alexandru: No, because the purpose of `ApplicativeEval` is evaluation in the `F[_]` context, but it doesn't necessarily have to be lazy, such that `Try` and `Future` can implement it.

What you're thinking of is the other type class called `MonadDefer`.

laws/src/main/scala/cats/laws/MonadDeferLaws.scala
alexandru: The triggered code is side-effectul and thought it would be a good idea to test that `delay(a)` is equivalent with `pure.map(_ => a)` in the presence of side-effects. This doesn't necessarily have to be so. Consider ...

```scala
implicit object EvalInstances extends MonadDefer[Eval] {
  def eval[A](a: Eval[A]) = a.memoize
}
```

And now this law is violated.

oscar-stripe: maybe it helps to add negative tests to be clear exactly what we are ruling out.

tests/src/test/scala/cats/tests/TryTests.scala
alexandru: I also added this law (notice `applicativeEvalWithError`):

```scala
eval(always(throw ex)) == raiseError(ex)
```

It's an *optional* law that happens in case `F[_]` also has an `ApplicativeError[A, Throwable]`. Mentioning this because `Try` follows this law, whereas `Eval` does not.

alexandru:
@tpolecat I think the problem here is that people want to support both Cats and Scalaz, in order to not upset anybody, I have the same problem in @Monix, including my own `Monad`. It sucks, but eventually one will win.

I think it should be in Cats also because it needs `Applicative`, `ApplicativeError` and `Monad` to describe some laws. I know you're a fan of laws, would appreciate some feedback on those.

But anyway, we can further discuss this at NEScala. I wanted to do this PR for quite some time and finally got it out of my system :-)

core/src/main/scala/cats/ApplicativeEval.scala
oscar-stripe: would `pure(a).map(_.value)` be a legit implementation of this? If not, can you add a comment to help people understand the purpose of the type.

If this is not legit, can we point to exactly what law rules it out?

alexandru: If there's consensus, sure, though I like `eval` because it's just one verb.

core/src/main/scala/cats/ApplicativeEval.scala
alexandru: I would have preferred `eval` to be described in terms of `delay`. This can go either way.

edmundnoble:
:+1: goddamn it, *finally*. Maybe this means stack-safety work can continue.

core/src/main/scala/cats/MonadDefer.scala
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * A [[Monad monad]] that allows for arbitrarily delaying the
+ * evaluation of an operation, triggering its execution on each run.
+ *
+ * @see [[ApplicativeEval]] for capturing effects in an `F[_]`
+ *     applicative context, but without the repeating side-effects
+ *     requirement.
+ */
+@typeclass trait MonadDefer[F[_]] extends Monad[F] with ApplicativeEval[F] {
```
etorreborre: Is it worth providing a default instance of `MonadDefer` implemented as `pure(()).map(_ => a)` and `pure(()).flatMap(_ => fa)` for any `Monad` `M`? This way we could have a `MonadDefer` instance even for `Id` which can be useful.

alexandru: So I've added laws that `eval` should indeed be equivalent with `pure(()).map(_ => a)` and `defer` should be equivalent with `pure(()).flatMap(_ => fa)`.

However `Id` cannot implement `MonadDefer`, because for this type-class we are specifying that any *side effects* should be *repeated on each evaluation*, with `defer` building a factory of `F[A]`.

This means that `MonadDefer` can be implemented for `cats.Eval`, `Task`, `IO`, `Coeval`, `Observable` or other monads that handle side-effects, but cannot be implemented for `Id`.  It would fail the specified laws - tried to come up with something meaningful there, would appreciate feedback on those.

`Id` can implement `ApplicativeEval`, which doesn't have such a restriction. `Try` and `Future` are other candidates for `ApplicativeEval` and that can't implement `MonadDefer`. I find `MonadDefer` cool because it describes laziness.

johnynek:
I'm a bit confused: what Applicative can't implement ApplicativeEval? Why
not just add this method to Applicative?

alexandru:
@johnynek because you took it out :-)  See: https://github.com/typelevel/cats/issues/1150

Indeed, any `Applicative` can be an `ApplicativeEval`, although I've added an *optional* law for `ApplicativeError` that says: `eval(always(throw ex)) <-> raiseError(ex)`

Note that not any `Monad` can be a `MonadDefer`. This one specifies laziness.

johnynek:
Indeed, but the argument was that we couldn't find a law other than
something you can already get without it, but the Eval seemed to imply
something different.

Why should the Applicative function not be lazy, but Monad be lazy?

We definitely shouldn't add a typeclass if everything can satisfy it,
otherwise literally no one should only implement Applicative, right?

johnynek:
To be clear: if we want to add a synonym for pure(()).map(_ => a) I guess
that's fine, but there seemed to be less agreement last time around that we
knew and agreed what these functions meant (specifically with regard to
when a side effect may be observed).

alexandru:
> Why should the Applicative function not be lazy, but Monad be lazy?

I can think of two reasons:

1. Because monads are used to handling side-effects, as side-effects need ordering / sequencing and the monad is in fact the type-class that models sequencing due to creating a data dependency between the source and its continuation. E.g. `Task`, `IO`, `Eval`, `Free` are all monads.
2. We need `Monad` in order to make `MonadDefer` lawful, otherwise we can't specify those laws, because we can't ensure ordering of operations. I actually tried specifying those laws with `map2` and it doesn't work. 

> if we want to add a synonym for pure(()).map(_ => a) I guess that's fine

It's not a perfect synonym for `pure(()).map(_ => a)` because `ApplicativeErrorLaws` does not specify `pure(()).map(_ => throw ex) <-> raiseError(ex)`.

This type-class specifies that if there is an `ApplicativeError[F, Throwable]` defined, then exceptions should also get caught in the `F[_]` context.

core/src/main/scala/cats/MonadDefer.scala
etorreborre: I agree on `Id` not having a `MonadDefer` instance. It wouldn't be really lazy and stackoverflow errors would happen (just got one while trying to use `Id` as my "simplest MonadDefer monad").

mpilquist:
Note that in FS2, we optimize `pure(a).flatMap(f)` patterns in to `f(a)` to avoid unnecessary trampolines. Hence, `pure(()).map(_ => a)` is not a safe way to delay evaluation and similarly, `pure(()).flatMap(_ => fa)` is not a safe way to suspend evaluation.

`cats.free.Free` does the same optimization, which means you can't safely use `Free.suspend` for effect capture.

In FS2, we solve this by implementing `suspend` in a way that's distinct from `pure(()).flatMap(_ => fa)`. The former is never eagerly stepped whereas the latter is always eagerly stepped.

johnynek:
Okay, now I see more clearly. Can we separate into two PRs possibly?

1. @alexandru why do we need the ApplicativeEval over catchNonFatal on ApplicativeError: https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/ApplicativeError.scala#L83 ? It seems duplicative to me in that it seems exactly like the law you are proposing is the current implementation. Should we just add more laws? I'm concerned about adding optional laziness. When do I want a type-class to optionally have a behavior? If I need laziness, I probably need it. So why not have that required or not?

2. The MonadDefer I think it totally fine. If it were me, I would lift the methods from ApplicativeEval into MonadDefer and require that they all be lazy or in the by-name versions have always-like behavior.

I think we are actually talking about two or more concerns here:

a. a typeclass about exception capture into a monad (more than ApplicativeError since we want some laws around `throw`).
b. one or more typeclasses about deferred evaluation. I think we want Applicative and Monad flavors of this, but we can certainly start with Monad where we are almost sure we want a flavor of this.

alexandru:
@johnynek 

> why do we need the ApplicativeEval over catchNonFatal on ApplicativeError? It seems duplicative to me

It's not. `ApplicativeEval` is meant to let `F[_]` evaluate the expression, instead of passing a strict parameter in `pure`. This means that it unifies these functions:

- `Eval.apply` (`Eval.always`)
- `Task.delay` (could be `Task.apply`)
- `Try.apply`
- `Future.apply`

What do they have in common? I can see 2 things:

1. All of them take a by-name parameter, which gets evaluated as they see fit
2. All of them are equivalent with `pure(()).map(_ => a)`

We had this in `Applicative.pureEval` but [it was removed](https://github.com/typelevel/cats/commit/94fcf03f774841806add287a8d5f43e6e1bbc6d2).

And here I must mention that the Scalaz equivalent [Applicative.point](https://github.com/scalaz/scalaz/blob/series/7.3.x/core/src/main/scala/scalaz/Applicative.scala#L25) takes a by-name parameter and so it doesn't have this problem. In other words, in Scalaz we can abstract over the above `apply` functions, but with Cats we cannot.

As for `ApplicativeError`, the`catchNonFatal` function I feel that it is missing the point. When you evaluate an expression, you want `F[_]` to be in charge of the evaluation. What happens next, error handling or not, is up to the user. And I know this is totally subjective, but I'll never use `catchNonFatal` to abstract over `Future.apply` simply because it has a scary and long name.
And if these 2 operations really are redundant, then imo `catchNonFatal` is the one that should go.

johnynek:
My concern is that these four things are not similar in any way except that they take a by-name parameter.

1. Eval: does not catch exceptions. `Eval.always, later` have different semantics. Used to create a trampolined `Id[T]`
2. Task: does (I believe) catch exceptions and has `always` semantics? Eventual execution is async.
3. Future.apply( ) catches exceptions. In scala futures async execution using an ExecutionContext, has `later` semantics (not always).
4. Try.apply catches exceptions except with `.now` like semantics.
5. Twitter Future.apply exactly like Try except in a Future (not async, catches exceptions).

So, if we have a typeclass, the idea, to me, is that it has some usable property that I only need to depend on. I don't see what it is here since there are several things it could be:

1. catching exceptions in evaluation
2. always-like semantics (side-effects will be re-run).
3. later-like semantics (we know something will be run 0 or 1 times)
4. async evaluation

I think we should be clear what we want. But right now we seem to be saying, I want to unify the above 4 properties such that at least one of them is true. I don't see how I can usefully program with that.

alexandru:
@johnynek Yes, they take a by-name parameter AND they are consistent with `pure` for pure expressions.

If you're looking that much into it, you might reach the same concerns about `map` and `flatMap`. After all `Try.flatMap` is eager, whereas `Task.flatMap` is lazy, has `always` semantics and is possibly async.

At this point I think we should talk use-cases. Here's the kind of code I'm dealing with right now:

```scala
sealed abstract class Iterant[F[_], +A] { self =>
  // ..
  final def foldLeftL[S](seed: => S)(op: (S,A) => S)(implicit F: Monad[F]): F[S] = {
    def loop(self: Iterant[F,A], state: S): F[S] = {
      try self match {
        case Next(a, rest, stop) =>
          // Side-effects and exceptions possible here!!!
          val newState = op(state, a)
          rest.flatMap(loop(_, newState))
        case Halt(None) =>
          F.pure(state)
        case Halt(Some(ex)) =>
          throw ex
      }
      catch {
        case NonFatal(ex) =>
          // Closing resources!
          earlyStop.flatMap(_ => (throw ex) : F[S])
      }
    }

    val init = F.eval {
      // Exception handling for the seed
      try seed catch {
        case NonFatal(ex) =>
          earlyStop.flatMap(_ => (throw ex) : F[S])
      }
    }

    init.flatMap(loop(self, _))
  }
}
```

For this piece of code I don't want to work with `ApplicativeError` because it would make it unnecessarily restrictive. Without `ApplicativeError` I can work with `Eval` just fine. I also don't care much about strict or lazy semantics, or about synchronous versus async evaluation. Maybe I have a use-case for using `Future`.

The whole point of higher-kinded polymorphism, the way I see it, is the ability to plug-in an `F[_]` that changes the nature of your operation according to F's capabilities. This sample does that, if only we had an `eval`.

So how can this operation be described with the current Cats type-classes and what should the solution look like?

johnynek:
Thank you for taking the time to make a concrete example. I think it helps not talk past one another.

First, Functor and Monad have well defined laws (composition of map and associativity of bind, etc...). While it is true some have lazy semantics, you get no promises and can't rely on it.

To me, it looks like you do want MonadError since you explicitly want to handle the possibility of non fatal exceptions in java/scala code. You also appear to want an "always"-like pure, but maybe you don't care if it is always, but could actually be Now.

Can you explain why you don't want MonadError? You are reimplementing catchNonFatal nearly exactly here.

Lastly, I suspect you may say you don't want MonadError because you want to support Eval, but as your code demonstrates you can nearly hack Eval to make it support MonadError. What is missing is recover. I think we can add a recover node to Eval without hurting performance or stack safety of existing code. As you point out, on the JVM an exception may come at any time, so currently evaluating Eval may throw. It would be nice to have a method like:

```scala
Eval.handleErrorWith[A](e: Eval[A])(fn: Throwable => Eval[A]): Eval[A]
```

If Eval supported MonadError, would you feel that MonadError covered your use case?

By the way, iteratee.io has many very similar cases and seems to use MonadError without it being a problem.

I think there is still the separate suspend use case which must be lazy (or maybe have always semantics). That seems like something we don't have currently. But I still don't see a new typeclass that includes Try, Eval and Task other than Applicative.

Maybe we could have a weaker thing than ApplicativeError where you can raise but not handle errors. Like ApplicativeRaise. Eval and Try can both be ApplicativeRaise with an error type of Throwable, it's just that Eval can't (yet?) handle the errors.

I think we could maybe make a `case class TryT[F[_], A](unwrap: F[Try[A]])` that we could use with Eval to make a lazy Try also.

edmundnoble:
The reason that `MonadDefer` needs to exist is to abstract over free monad implementations that provide stack-safety through some kind of deferred evaluation. These free monad implementations include `Eval`, `Free[Function0, A]`, and every kind of `Task`. Without MonadDefer, it is impossible to safely perform monadic recursion and still abstract over the monad in question. This is the killer usecase of `MonadDefer`. Injections from `Eval` are *not* sufficient to implement `MonadDefer`. Just try it yourself: given a function that yields a `Task[A]` and recurses on itself, you need to convert the Task *into* an `Eval` so it can be suspended and then back *out* of `Eval` to return. That's not possible.

johnynek:
To be clear, I am all for MonadDefer as long as we can be clear about the laws. There seem to be at least two threads:

1. we need defer for stack safe recursion
2. we need defer to control when side-effects occur (i.e. we call some side-effecting method in a `defer( ... pure( ))`.

I don't know if both of those cases should be covered by a single typeclass.

edmundnoble:
@johnynek My position on this is that we should provide a separate side-effect capture typeclass, potentially in a different PR. I don't think it necessarily needs any superclasses. What we will provide to users there is purely an abstraction that allows people to easily *switch* side-effect types, by pushing the decision to the call site. This is *invaluable* for libraries performing side effects that *should not* need to provide clones of their methods for different cats-compatible effect-capture libraries. I think the only law that is necessarily useful there is that whatever is suspended is *not* evaluated eagerly.

alexandru:
Hi guys,

So the current consensus is that we need `MonadDefer`, but for `ApplicativeEval` the need isn't immediately clear, because `ApplicativeError` is fine for the use-cases I had in mind, am I correct?

If this is so, then I'll create another PR with just `MonadDefer`. I think some of us happen to be at NEScala as well and maybe we'll also talk about this.

alexandru:
The history of this issue is interesting, for now I have not created a new PR. Changes:

- dropped `ApplicativeEval` from this PR
- removed the "optional" laws for `ApplicativeError`
- added law for `flatMap` stack safety

Please re-review. Thanks!

johnynek:
Can Free and FreeT implement this? It seems like they could.

Also I think EitherT can if the F there can implement it (maybe any transformer can, ReaderT, WriterT, ... )

edmundnoble:
@johnynek Free and FreeT can only implement this if `F =:= Function0`. Otherwise yeah, transformers preserve it.

johnynek:
@edmundnoble I don't follow why Free.suspend is not a valid defer. What am I missing?

edmundnoble:
@johnynek try Free[Id, A]. Then suspend does nothing for stack safety.

johnynek:
Wait, why not? Free is trampolined. How is stack unsafe? Id has a good
tailRecM so running the Free is also safe. I don't see the problem.

johnynek:
In fact Free[Id, ?] should be equivalent (but presumably less efficient) to
Eval, unless I'm missing something.

mpilquist:
@johnynek Re: `Free.suspend` and `MonadDefer` -- https://github.com/typelevel/cats/issues/1277

johnynek:
Okay. I recall something around this before. So after one evaluation it
will not be repeated with Free.

Is this also the case with FreeT do you know off hand.

mpilquist:
I haven't used `FreeT` so I can't say without investigating. Note that we fixed it in FS2 by introducing a flag on the internal `Pure` constructor indicating whether it can be eagerly stepped. We could certainly do the same in cats: https://github.com/functional-streams-for-scala/fs2/blob/46a1281813f487064aab5d8ad2747bb299e28eeb/core/shared/src/main/scala/fs2/util/Free.scala#L52-L59

alexandru:
@johnynek btw, the current definition for `catchNonFatal` is this:

```scala
  def catchNonFatal[A](a: => A)(implicit ev: Throwable <:< E): F[A] =
    try pure(a)
    catch {
      case NonFatal(e) => raiseError(e)
    }
```

And in the instances defined in Cats, the *only* override I'm seeing is in `FutureInstances`. In other words, `catchNonFatal` really is not what I need, because:

1. it's an afterthought that people don't override much
2. I don't want to impose the `ApplicativeError` restriction when `Applicative` would be fine with both Scalaz and Haskell

I remember mentioning this before, can't find the comment, but Cats is the only one with the problem because all the others have lazy behavior in `pure`.

kailuowang:
[further discussion on gitter](https://gitter.im/typelevel/cats?at=58e65428f22385553d15a496)
[Related issue at typelvel/general](https://github.com/typelevel/general/issues/66)

alexandru:
I have included a [Deferrable](https://github.com/alexandru/schrodinger/blob/master/shared/src/main/scala/schrodinger/Deferrable.scala) type-class and an [Evaluable](https://github.com/alexandru/schrodinger/blob/master/shared/src/main/scala/schrodinger/Evaluable.scala) type-class in my proposal for the [Typelevel Schrodinger](https://github.com/typelevel/general/issues/66) project.

This addresses two concerns:

1. this project does not depend on `cats-core` and these type-classes do not inherit from `Applicative` or `Monad`, hence avoiding MTL issues
2. the provided laws are describing side-effects and so they are kept out of Cats, where purity is preferred

If everybody agrees, I would like to close this pull request and this new project instead.

alexandru:
Hello, I feel this issue has reached a deadlock. I don't want this to stop progress on alternative proposals, therefore I'm closing it.

Issue #1505, opened 2016-12-22
Diverging Inject instances


notxcain:
Given this specific case when a type constructor being injected is the right-right part of coproduct and defined in object.
```scala
trait Foo[A]
trait Bar[A]
object Baz {
  trait F[A]
}
object Repro {
  Inject[Baz.F, Coproduct[Foo, Coproduct[Bar, Baz.F, ?], ?]]
}
```

It does not compile and I get

```
Error:(12, 9) diverging implicit expansion for type cats.free.Inject[Baz.F,[γ$0$]cats.data.Coproduct[Foo,[γ$1$]cats.data.Coproduct[Bar,Baz.F,γ$1$],γ$0$]]
starting with method catsFreeReflexiveInjectInstance in class InjectInstances
  Inject[Baz.F, Coproduct[Foo, Coproduct[Bar, Baz.F, ?], ?]]
```

I compiles well if move `F` out of `Baz`. But sadly for my case it's not an option.

travisbrown:
There seems to be a lot of weirdness here. The following compiles on 2.12.1 if you paste it as individual lines into the REPL:

```scala
trait Foo[A]
trait Bar[A]
object Baz { trait F[A] }
import cats.data.Coproduct, cats.free._
type BarBaz[x] = Coproduct[Bar, Baz.F, x]
type FooBarBaz[x] = Coproduct[Foo, BarBaz, x]
Inject[Baz.F, FooBarBaz]
```

But not with `:paste`, not if the import comes first, and not if the `cats.free._` import is replaced with `cats.free.Inject` (which seems like it should work just fine).

There's weirdness even without the nesting—e.g. none of the `Inject`s here compile:

```scala
import cats.data.Coproduct, cats.free.Inject

trait Foo[A]
trait Bar[A]
trait Baz[A]

Inject[Foo, Coproduct[Foo, Coproduct[Bar, Baz, ?], ?]]
Inject[Bar, Coproduct[Foo, Coproduct[Bar, Baz, ?], ?]]
Inject[Baz, Coproduct[Foo, Coproduct[Bar, Baz, ?], ?]]
```

I think it's likely that we could do a better job of working around the compiler issues here to minimize the weirdness for users, but I'd say we kick that down the road past the 0.9 release, which could otherwise happen this week.

romainreuillon:
I ran into the same pb while trying to stack several free monads:
```scala
  import cats._
  //import cats.data._
[...]

PR #1557, opened 2017-03-13
Improvements to Inject.

sellout:
- move it back to cats-core (near `Coproduct`)
- make `inj`/`prj` natural transformations
- add `apply`/`unapply` (lowered `inj`/`prj`)
- define `Free.roll`
- move `injectRoll` (née `Inject.inject`) and `match_` to `Free`
- reintroduce "null identity" test (written by @edmundnoble)


sellout:
These are largely changes needed to ease Quasar’s migration.

tests/src/test/scala/cats/tests/InjectTests.scala
```diff
@@ -0,0 +1,81 @@
+package cats
+
+// import cats.arrow.FunctionK
```
adelbertc: woops

core/src/main/scala/cats/Inject.scala
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Inject type class as described in "Data types a la carte" (Swierstra 2008).
+ *
+ * @see [[http://www.staff.science.uu.nl/~swier004/publications/2008-jfp.pdf]]
+ */
+sealed abstract class Inject[F[_], G[_]] {
+  def inj: FunctionK[F, G]
+
+  def prj: FunctionK[G, λ[α => Option[F[α]]]]
```
adelbertc: Should we call these `inject` and `project` ?

tests/src/test/scala/cats/tests/InjectTests.scala
sellout: Also, I apparently broke tut. Ok, fixes in the morning 😅

core/src/main/scala/cats/Inject.scala
sellout: I’m fine with either. The current ones match the paper and Scalaz. Also, Matryoshka defines `project`, so I don’t mind the lack of collision 😁

notxcain:
@sellout did you see https://github.com/typelevel/cats/issues/1505 in the wild? If so, how do you workaround?

sellout:
@notxcain Sorry, I haven’t seen that.

core/src/main/scala/cats/Inject.scala
```diff
@@ -0,0 +1,47 @@
+package cats
+
+import cats.arrow.FunctionK
+import cats.data.Coproduct
+import cats.syntax.either._
```
kailuowang: unused import here breaking the build

sellout: Thanks. I have a false memory of seeing a green build, so this would have sat for a while before I noticed.

kailuowang:
Sorry my bad. The either syntax is used in scala 2.11 and 2.10 withi this [`toOption`](https://github.com/sellout/cats/blob/d9566a215f3cf8846d79fd58b87dab19ac954a58/core/src/main/scala/cats/Inject.scala#L33). It's unused in Scala 2.12 because `Either` is right biased and `toOption` runs on the right projection. So now the build on 2.10 and 2.11 is broken. A fix off top of my head is to write `_.run.swap.right.toOption` instead. There may be a better way to support the different syntax in different Scala? cc @adelbertc

sellout:
@kailuowang Yeah – I brought this up on the gitter channel. @paulp mentioned ghik/silencer, but that doesn’t support 2.10. I’ll try the `swap` approach. Thanks.

kailuowang:
👍  LGTM, since this is a breaking change, we might want to wait for a third maintainer sign-off.

kailuowang:
merging with three sign-offs

PR #1558, opened 2017-03-13
Add basic exception-handling to Eval.

non:
These methods provide the basis for a `MonadError[Eval, Throwable]`
instance, without adding any overhead for people who don't wish to use
these (admittedly limited) capabilities.

Since `Eval` defers evaluation, we can take advantage of this to defer
additional recovery logic to the eventual evaluation. Critically, it doesn't
add any kind of *error node* to the `Eval` AST, or any extra overhead to
the default evaluation paths.

This came up when I was trying to catch up on https://github.com/typelevel/cats/pull/1552.
I realized there was no real reason why we couldn't have a
`MonadError[Eval, Throwable]` instance.

Review by @johnynek, @alexandru, et al.

### DIFF
--- a/tests/src/test/scala/cats/tests/EvalTests.scala
+++ b/tests/src/test/scala/cats/tests/EvalTests.scala
@@ -89,16 +89,9 @@ class EvalTests extends CatsSuite {
       else if (i % 100 == 0) loop(i - 1, e.map(n => n).handleErrorWith(_ => Now(-999)))
       else loop(i - 1, e.map(n => n))
 
-    // we expect to use 1k stack frames of error handling
+    // we expect to use ~1k stack frames of error handling, which
+    // should be safe.
     loop(100000, Now(6)).value should === (6)
-
-    // we expect to use at least 100k stack frames and fail
-    try {
-      loop(10000000, Now(7)).value
-      fail("we expected a StackOverflowError but didn't get one")
-    } catch { case t: StackOverflowError =>
-      succeed
-    }
   }
 
   test("Eval.raise(t).handleErrorWith(f) = f(t)") {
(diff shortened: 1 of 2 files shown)


non:
I did a bunch of tests to convince myself that these methods preserve stack safety, and that they didn't consume additional stack unnecessarily.

(One exception I can think of is if the caller writes a `recovery` function which calls `.value` on the instance in question, and then nests these calls recursively. But this isn't something we can guard against easily and seems very unlikely in practice.)

tests/src/test/scala/cats/tests/EvalTests.scala
```diff
@@ -86,7 +86,7 @@ class EvalTests extends CatsSuite {
     checkAll("Eval[Int]", MonadTests[Eval].monad[Int, Int, Int])
   }
   checkAll("Bimonad[Eval]", SerializableTests.serializable(Bimonad[Eval]))
-  checkAll("Monad[Eval]", SerializableTests.serializable(Monad[Eval]))
+  checkAll("MonadError[Eval, Throwable]", SerializableTests.serializable(MonadError[Eval, Throwable]))
```
johnynek: this is only the serializable test, not the actual law test.

tests/src/test/scala/cats/tests/EvalTests.scala
```diff
@@ -86,7 +86,7 @@ class EvalTests extends CatsSuite {
     checkAll("Eval[Int]", MonadTests[Eval].monad[Int, Int, Int])
```
johnynek: note here is where the MonadTests are being run.

tests/src/test/scala/cats/tests/EvalTests.scala
non: Oh, good catch. I'll fix it now!

johnynek:
👍 This tackles at least one use case @alexandru had.

non:
I'm pushing some extra tests. Also, unfortunately I determined that my previous tests were wrong. The `.handleErrorWith` call does use one extra stack frame during `.value` evaluation. This is probably not a huge problem in most cases but does mean that if you nest e.g. 10K of these calls you'll hit a `StackOverflowException`.

I don't see a good way to remediating that issue without impacting the overall `Eval` performance. I'll push my extra tests and comments and we can decide then.

non:
I see two sign-offs so I'm going ahead and merging. Thanks folks!

PR #1564, opened 2017-03-18
Adds an As class which represents subtyping relationships (Liskov)

stew:
This is a direct port of the Liskov class from scalaz.  I named it `As` to be
similar to our new `Is` class representing type equality, but there are aliases
named `<~<` and `Liskov`


core/src/main/scala/cats/evidence/As.scala
```diff
@@ -0,0 +1,268 @@
...
+sealed abstract class As[-A, +B] {
+  def apply(a: A): B = As.witness(this)(a)
+
+  /**
+   * Use this subtyping relationship to replace B with a value of type
+   * A in a contravariant context.  This would commonly be the input
+   * to a function, such as F in: `type F[-B] = B => String`. In this
+   * case, we could use A As B to turn an F[B] Into F[A].
+   */
+  def subst[F[-_]](p: F[B]): F[A]
```
johnynek: We use substitute in Is.

core/src/main/scala/cats/evidence/As.scala
```diff
@@ -0,0 +1,268 @@
...
+ * that expects a `B`.  (e.g. if you could pass an `A` into any
+ * function that expects a `B` as input.)
+ * 
+ * This code was ported directly from scalaz to cats using this version from scalaz:
+ * https://github.com/scalaz/scalaz/blob/a89b6d63/core/src/main/scala/scalaz/As.scala
+ * 
+ *  The original contribution to scalaz came from Jason Zaugg
+ */
+sealed abstract class As[-A, +B] {
+  def apply(a: A): B = As.witness(this)(a)
```
johnynek: I think we called this coerce in Is. Can we use the same for both?

core/src/main/scala/cats/evidence/As.scala
```diff
@@ -0,0 +1,268 @@
...
+  def subst[F[-_]](p: F[B]): F[A]
+
+  final def andThen[C](that: As[B, C]): As[A, C] = As.trans(that, this)
+
+  final def compose[C](that: As[C, A]): As[C, B] = As.trans(this, that)
+
+  /**
+   * Use this relationship to widen the output type of a Function1
+   */
+  def onF[X](fa: X => A): X => B = As.co2_2[Function1, B, X, A](this)(fa)
```
johnynek: Why is this one on the instance but generally the rest are on the object?

core/src/main/scala/cats/evidence/As.scala
```diff
@@ -0,0 +1,268 @@
...
+}
+
+sealed abstract class AsInstances {
+  import As._
+
+  /*
+   * Subtyping forms a category 
+   */
+  implicit val liskov: Category[<~<] = new Category[<~<] {
```
johnynek: I don't think we have a category for Is. we should be consistent.

core/src/main/scala/cats/evidence/As.scala
```diff
@@ -0,0 +1,268 @@
...
+  /**
+   * We can witness the relationship by using it to make a substitution *
+   */
+  implicit def witness[A, B](lt: A <~< B): A => B = 
+    lt.subst[-? => B](identity)
+
+  /**
+   * Subtyping is reflexive 
+   */
+  implicit def refl[A]: (A <~< A) =
```
johnynek: Why do we need this and isa?

johnynek: Also, in Is we use a singleton to avoid allocations.

core/src/main/scala/cats/evidence/As.scala
```diff
@@ -0,0 +1,268 @@
...
+    new (A <~< A) {
+      def subst[F[-_]](p: F[A]): F[A] = p
+    }
+
```
johnynek: We don't have a method to materialize a `<:<` as needed. I think we should. Also a similar unsafe method to go from `<:<` to As.

core/src/main/scala/cats/evidence/As.scala
```diff
@@ -0,0 +1,268 @@
...
+    a.subst[λ[`-α` => T[α] <~< T[A2]]](refl)
+
+  // Similarly, we can do this any time we find a covariant type
+  // parameter. Here we provide the proof for what we expect to be the
+  // most common shapes.
+
+  def co2[T[+_, _], Z, A, B](a: A <~< Z): T[A, B] <~< T[Z, B] =
+    a.subst[λ[`-α` => T[α, B] <~< T[Z, B]]](refl)
+
+  /**
```
johnynek: If we have all these methods below I think we should have the analogous for Is.

core/src/main/scala/cats/evidence/package.scala
```diff
@@ -2,4 +2,24 @@ package cats
 
 package object evidence {
   type Leibniz[A, B] = cats.evidence.Is[A, B]
```
johnynek: Why do we use cats.evidence here but not below?

core/src/main/scala/cats/evidence/As.scala
```diff
@@ -0,0 +1,268 @@
...
+    type c[-X] = F[A2, B2, C2, D2, R2] <~< F[A, B, X, D2, R2]
+    type d[-X] = F[A2, B2, C2, D2, R2] <~< F[A, B, C, X, R2]
+    type r[-X] = F[A2, B2, C2, D2, X] <~< F[A, B, C, D, R2]
+    r.subst[r](d.subst[d](c.subst[c](b.subst[b](a.subst[a](refl)))))
+  }
+
+  /**
+   * Unsafely force a claim that A is a subtype of B 
+   */
+  def force[A, B]: A <~< B =
```
johnynek: I don't think we should have this. You can call isa and cast if you want to do something unsafe. The cast at least shows you it is unsafe. Burying an unsafe cast in a library function seems like a bad choice to me.

core/src/main/scala/cats/evidence/package.scala
sellout: Maybe also add a `===` alias to parallel `<~<`?

core/src/main/scala/cats/evidence/As.scala
```diff
@@ -0,0 +1,201 @@
...
+   * Subtyping is transitive
+   */
+  def compose[A, B, C](f: B <~< C, g: A <~< B): A <~< C =
+    g.substitute[λ[`-α` => α <~< C]](f)
+
+  /**
+   * Lift Scala's subtyping relationship
+   */
+  @inline def unsafeFromPredef[A, B >: A]: (A As B) =
+    reflAny.asInstanceOf[A As B]
```
johnynek: I think this is actually safe (you could allocate A As B here). It is going from `A <:< B` that I think is correct but can't be proven via the types.

johnynek: So, start with refl, then do to covariance you have `A As A <: A As B` so it can be returned. So I think returning `refl[A]` here will compile.

tests/src/test/scala/cats/tests/AsTests.scala
```diff
@@ -0,0 +1,24 @@
+package cats
+package tests
+
+class AsTests extends CatsSuite {
+  import evidence._
+
```
johnynek: Can we test resolution explicitly of a few types? `Int As Any` `(Int, Long) As (Any, Any)`

stew:
@johnynek I *think* I've covered all your requests now

alexknvl:
I think `unsafeFromPredef` should be renamed as `isa` or `reify` and implemented in terms of `refl`. True `unsafeFromPredef` should have signature "A <:< B => A <~< B`.

johnynek:
I agree with @alexknvl 's comments. I don't see how what we have currently called `unsafeFromPredef` is unsafe. If it is, can you comment?

Also, the coverage of this PR is pretty low. I love using this:

https://chrome.google.com/webstore/detail/codecov-extension/keefkhehidemnokodkdkejapdgfjmijf?hl=en-US

to see which methods have no tests. Very helpful (even a fun game to get everything green).

sellout:
I _think_ this is the last thing that needs to be merged to get a Monocle-Cats PR submitted.

alexknvl:
May I also suggest not adding `Liskov` alias? I think it might be a good idea to reserve some name space in case bounded versions of `Is` and `As` are added (they could be named `Leibniz` and `Liskov` accordingly).

kailuowang:
@stew it looks to me that, luckly, the compilation error on scalajs on 2.10.6 can be fixed by upgrading scalajs to 0.6.16, which is the version on master. So if you merge in master, it should probably build. 
@johnynek it looks like that the test coverage is decent now. Wanna take another look?

kailuowang:
@stew I found out the cause, the type inference in `toMap` lin `AsTests` line 7 is a bit too much for scalajs on 2.10.6 to handle. A little help like the following gets it pass. 
```scala
  def toMap[A, B, X](fa: List[X])(implicit ev: X <~< (A,B)): Map[A,B] = {
    type RequiredFunc = (Map[A, B], X) => Map[A, B]
    type GivenFunc = (Map[A, B], (A, B)) => Map[A, B]
    val subst: GivenFunc <~< RequiredFunc = As.contra2_3(ev) //introduced because inference failed on scalajs on 2.10.6
    fa.foldLeft(Map.empty[A,B])(subst(_ + _))
  }
```
@stew do you have time to update the PR with this change? @johnynek do you mind taking another look?

tests/src/test/scala/cats/tests/AsTests.scala
```diff
@@ -0,0 +1,101 @@
...
+    val contra3_3: EatF33[Unit, Unit, Top] As EatF33[Unit, Unit, Bottom] = As.contra3_3(cAsA)
+  }
+
+  test("we can widen a function1") {
+    val f: Any => Bottom = _ => Bottom()
+    val cAsA: Bottom As Top = implicitly
+    val f2: Any => Top = As.onF(cAsA)(f)
+  }
+
+  test("we can simultaneously narrow the input and widen the ouptut of a Function1") {
```
kailuowang: looks like we are missing at test here?

core/src/main/scala/cats/evidence/As.scala
```diff
@@ -0,0 +1,170 @@
...
+
+  /**
+   * We can lift subtyping into any covariant type constructor
+   */
+  def co[T[+_], A, A2] (a: A As A2): (T[A] As T[A2]) =
+    a.substitute[λ[`-α` => T[α] As T[A2]]](refl)
+
+  // Similarly, we can do this any time we find a covariant type
+  // parameter. Here we provide the proof for what we expect to be the
+  // most common shapes.
```
kailuowang: should this go into the scala doc of the following method?

kailuowang:
@stew got a bit time to finish this up? If not I would be more than happy to work on it.

kailuowang:
@johnynek I continued this PR with a new one #1728. cc @sellout

PR #1571, opened 2017-03-24
added whileM, untilM, iterateWhile, etc to Monad

### LINKED ISSUE #1569: Lack of whileM, whileM_, untilM and untilM_ methods for Monad
These methods present in `scalaz.Monad`. Should we add them in cats as well?

kailuowang:
Continuation on #1216. I didn't add the other methods yet because I want to make sure I get the tests right first. There are no tests present in either original PR or Scalaz . 
cc @tpolecat @johnynek 

Fixes #1569.

### DIFF
--- a/core/src/main/scala/cats/Monad.scala
+++ b/core/src/main/scala/cats/Monad.scala
@@ -24,8 +24,8 @@ import syntax.either._
     val b = Eval.later(body)
     tailRecM[G[A], G[A]](G.empty)(xs => ifM(p)(
       ifTrue = {
-        map(b.value) { b =>
-          Left(G.combineK(G.pure(b), xs))
+        map(b.value) { bv =>
+          Left(G.combineK(G.pure(bv), xs))
         }
       },
       ifFalse = pure(xs.asRight[G[A]])
--- a/core/src/main/scala/cats/Monad.scala
+++ b/core/src/main/scala/cats/Monad.scala
@@ -19,6 +19,8 @@ import syntax.either._
    * Execute an action repeatedly as long as the given `Boolean` expression
    * returns `true`. The condition is evaluated before the loop body.
    * Collects the results into an arbitrary `Alternative` value, such as a `Vector`.
+   * This implementation uses append on each evaluation result,
+   * so avoid data structures with non-constant append performance, e.g. `List`.
   */
   def whileM[G[_], A](p: F[Boolean])(body: => F[A])(implicit G: Alternative[G]): F[G[A]] = {
     val b = Eval.later(body)
@@ -53,6 +55,8 @@ import syntax.either._
    * Execute an action repeatedly until the `Boolean` condition returns `true`.
    * The condition is evaluated after the loop body. Collects results into an
    * arbitrary `Alternative` value, such as a `Vector`.
+   * This implementation uses append on each evaluation result,
+   * so avoid data structures with non-constant append performance, e.g. `List`.
    */
   def untilM[G[_], A](f: F[A])(cond: => F[Boolean])(implicit G: Alternative[G]): F[G[A]] = {
     val p = Eval.later(cond)
@@ -75,7 +79,7 @@ import syntax.either._
   def iterateWhile[A](f: F[A])(p: A => Boolean): F[A] = {
     flatMap(f) { i =>
       tailRecM(i) { a =>
-        if(p(a))
+        if (p(a))
           map(f)(_.asLeft[A])
         else pure(a.asRight[A])
       }
@@ -89,7 +93,7 @@ import syntax.either._
   def iterateUntil[A](f: F[A])(p: A => Boolean): F[A] = {
     flatMap(f) { i =>
       tailRecM(i) { a =>
-        if(p(a))
+        if (p(a))
           pure(a.asRight[A])
         else map(f)(_.asLeft[A])
       }
--- a/core/src/main/scala/cats/Monad.scala
+++ b/core/src/main/scala/cats/Monad.scala
@@ -21,9 +21,10 @@ import syntax.either._
    * Collects the results into an arbitrary `MonadCombine` value, such as a `List`.
   */
   def whileM[G[_], A](p: F[Boolean])(body: => F[A])(implicit G: MonadCombine[G]): F[G[A]] = {
+    val b = Eval.later(body)
     tailRecM[G[A], G[A]](G.empty)(xs => ifM(p)(
       ifTrue = {
-        map(body) { b =>
+        map(b.value) { b =>
           Left(G.combineK(G.pure(b), xs))
         }
       },
@@ -39,9 +40,10 @@ import syntax.either._
   def whileM_[A](p: F[Boolean])(body: => F[A]): F[Unit] = {
     val continue: Either[Unit, Unit] = Left(())
     val stop: F[Either[Unit, Unit]] = pure(Right(()))
+    val b = Eval.later(body)
     tailRecM(())(_ => ifM(p)(
       ifTrue = {
-        map(body)(_ => continue)
+        map(b.value)(_ => continue)
       },
       ifFalse = stop
     ))
--- a/core/src/main/scala/cats/Monad.scala
+++ b/core/src/main/scala/cats/Monad.scala
@@ -17,10 +17,10 @@ import syntax.either._
 
   /**
    * Execute an action repeatedly as long as the given `Boolean` expression
-   * returns `true`. The condition is evalated before the loop body.
-   * Collects the results into an arbitrary `MonadCombine` value, such as a `List`.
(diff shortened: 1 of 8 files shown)


core/src/main/scala/cats/Monad.scala
```diff
@@ -14,4 +14,36 @@ import simulacrum.typeclass
...
+      ifFalse = pure(xs.asRight[G[A]])
+    ))
+  }
+
+  /**
+   * Execute an action repeatedly as long as the given `Boolean` expression
+   * returns `true`. The condition is evaluated before the loop body.
+   * Discards results.
+   */
+  def whileM_[A](p: F[Boolean], body: => F[A]): F[Unit] = {
```
johnynek: why not use the same style as above`(p: F[Boolean])(body: => F[A])`?

core/src/main/scala/cats/Monad.scala
```diff
@@ -14,4 +14,36 @@ import simulacrum.typeclass
 @typeclass trait Monad[F[_]] extends FlatMap[F] with Applicative[F] {
   override def map[A, B](fa: F[A])(f: A => B): F[B] =
     flatMap(fa)(a => pure(f(a)))
+
+  /**
+   * Execute an action repeatedly as long as the given `Boolean` expression
+   * returns `true`. The condition is evalated before the loop body.
+   * Collects the results into an arbitrary `MonadCombine` value, such as a `List`.
+  */
+  def whileM[G[_], A](p: F[Boolean])(body: => F[A])(implicit G: MonadCombine[G]): F[G[A]] = {
```
edmundnoble: We don't use by-name params in cats.

kailuowang: @johnynek I think the intention here is to avoid creating the body `F[A]` if `p` is never true. Thanks to you, I noticed that I missed the memoization of body. I just implemented using a `Later`.
@edmundnoble internally we don't use by-name params, but we do have plenty of public APIs using it for the sake of ease to use when we don't really want to give user the control over evaluation. 
In this particular case, having user passing in a by-name and then memoize it might be misleading. I am open to change to take in a `Later[A]` instead. Or maybe it's okay to give user the control and just take in an `Eval`. What do you guys think?

kailuowang: @johnynek I think the intention here is to avoid creating the body `F[A]` if `p` is never true. Thanks to you, I noticed that I missed the memoization of body. I just implemented using a `Later`.
@edmundnoble internally we don't use by-name params, but we do have plenty of public APIs using it for the sake of ease to use when we don't really want to give user the control over evaluation. 
In this particular case, having user passing in a by-name and then memoize it might be misleading. I am open to change to take in a `Later[A]` instead. Or maybe it's okay to give user the control and just take in an `Eval`. What do you guys think?

edmundnoble: @kailuowang Oh, I didn't know it was solely an optimization to avoid evaluating the body. I don't see any reason to let the user supply an Eval because it will be evaluated more than once; otherwise makes sense to me.

core/src/main/scala/cats/Monad.scala
```diff
@@ -14,4 +14,86 @@ import simulacrum.typeclass
...
+   * Execute an action repeatedly as long as the given `Boolean` expression
+   * returns `true`. The condition is evaluated before the loop body.
+   * Collects the results into an arbitrary `Alternative` value, such as a `Vector`.
+  */
+  def whileM[G[_], A](p: F[Boolean])(body: => F[A])(implicit G: Alternative[G]): F[G[A]] = {
+    val b = Eval.later(body)
+    tailRecM[G[A], G[A]](G.empty)(xs => ifM(p)(
+      ifTrue = {
+        map(b.value) { bv =>
+          Left(G.combineK(xs, G.pure(bv)))
```
kailuowang: Notice here that to maintain the order of evaluation results I have to append the latest result to the end of the `Alternative`, which means the performance will be horrible for `List`. 
I see 2 options here: 
1. in the API document, point this out and warn people against data structures with non-constant append performance. 
2. use prepend here and return the `G` with results in the reverse order, i.e. last result first. 
3. maintain an internal `List` in the reverse order and construct the `G` at last in the right order. 

@tpolecat and @Atry, what would be your preference here?

core/src/main/scala/cats/Monad.scala
```diff
@@ -14,4 +14,90 @@ import simulacrum.typeclass
...
+
+  /**
+   * Execute an action repeatedly as long as the given `Boolean` expression
+   * returns `true`. The condition is evaluated before the loop body.
+   * Discards results.
+   */
+  def whileM_[A](p: F[Boolean])(body: => F[A]): F[Unit] = {
+    val continue: Either[Unit, Unit] = Left(())
+    val stop: F[Either[Unit, Unit]] = pure(Right(()))
+    val b = Eval.later(body)
```
johnynek: why the later vs always semantics? For most monads it shouldn't matter, but for something like `Future` or `Try` where there is possibly a side-effect going on, it would.

kailuowang: Do you think these methods would be of practical use for `Future` or `Try`? Neither a `Try[Boolean]` nor `Future[Boolean]` will ever change the underneath value, and they are call by value.
It seems to me that the original incentive of introducing these methods are **only** for data types that track side effects such as `IO`/`Task`/`StateT`. When it comes to other data types like `Option` or `List`, they don't make sense to me, at least with current API design.

core/src/main/scala/cats/Monad.scala
```diff
@@ -14,4 +14,90 @@ import simulacrum.typeclass
...
+   * Execute an action repeatedly until the `Boolean` condition returns `true`.
+   * The condition is evaluated after the loop body. Collects results into an
+   * arbitrary `Alternative` value, such as a `Vector`.
+   * This implementation uses append on each evaluation result,
+   * so avoid data structures with non-constant append performance, e.g. `List`.
+   */
+  def untilM[G[_], A](f: F[A])(cond: => F[Boolean])(implicit G: Alternative[G]): F[G[A]] = {
+    val p = Eval.later(cond)
+    flatMap(f)(x => map(whileM(map(p.value)(!_))(f))(xs => G.combineK(G.pure(x), xs)))
+  }
```
tpolecat: So, is `val p = Eval.later(cond)` what we're doing instead of `lazy val p = cond`? Is there a discussion somewhere?

kailuowang: Yes. I picked `Eval.later(cond)` over `lazy val` because of the optimization in `Later`, and the fact that an equivalent `Need` is used in scalaz. But I don't have proof that it is worth the extra allocation. 
I am not aware of any discussion regarding choosing between these two in cats.

tpolecat: Ok, I don't feel strongly about it. Was just curious.

johnynek: Later just uses lazy val under the hood: https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Eval.scala#L186 

The only reason I can think of to use `Later` vs `lazy val` is maybe the thunk can be GC'd in the Later case? cc @non

core/src/main/scala/cats/Monad.scala
johnynek: That's fine, I was mostly asking because of the `Eval.later`. If we used `Eval.always` then you could use `Try` or `Future` by sneaking some effect into the call-by-name param, but I'm not a fan of that.

If we want to stick with `Eval.later` over lazy val, I'd like us to document the reason. The only reason I can see is that it allows the thunk in the call-by-name to be GC'ed after we run, which can be nice. We should comment why because someone will forget and revert since Later is only more work (it uses lazy val + some work).

tpolecat:
Actually this jogs my memory. See https://github.com/scalaz/scalaz/issues/1144

core/src/main/scala/cats/Monad.scala
non: There's also a locking issue -- lazy vals lock their enclosing class, `Later` only locks itself.

tpolecat: Yeah see https://github.com/scalaz/scalaz/issues/1144 for discussion (moving my comment up).

core/src/main/scala/cats/Monad.scala
kailuowang: I agree we should document these not-so-obvious tech decisions somewhere (another example in my mind is typeclass instance priority). Maybe in the [contributing](http://typelevel.org/cats/contributing.html)?

core/src/main/scala/cats/Monad.scala
johnynek: okay. Makes sense to me. 👍

Issue #1384, opened 2016-09-21
Symbols page not on site


ceedubs:
We have this handy symbols guide: https://github.com/typelevel/cats/blob/master/docs/src/main/tut/symbols.md

But as far as I can tell it's not being published to the docs site. I think that it may be because it's missing the yaml/jekyll front matter at the top of the file. We should also make sure that we link to this page so people can find it.

zainab-ali:
I'd be happy to add this, but I'm unsure which section it would go in.  Definitely not typeclasses or data types - perhaps it should be a subsection under the FAQ?

kailuowang:
fixed by #1573

PR #1537, opened 2017-02-16
Add FlatMap.forEffect

cranst0n:
See #1535.

PR #1573, opened 2017-03-25
Add symbols to FAQ

zainab-ali:
Add symbols to FAQ, as discussed in #1384


docs/src/main/tut/faq.md
```diff
@@ -191,6 +192,37 @@ If you're having trouble figuring out how to implement `tailRecM` lawfully, you
 
 In some cases you may decide that providing a lawful `tailRecM` may be impractical or even impossible (if so we'd like to hear about it). For these cases we provide a way of testing all of the monad laws _except_ for the stack safety of `tailRecM`: just replace `MonadTests[F].monad[A, B, C]` in your tests with `MonadTests[F].stackUnsafeMonad[A, B, C]`.
 
+
+## <a id="symbol" href="#symbol"></a>What does this symbol mean?
+
+Below is a list of symbols used in cats.
+
+The `~>`, `⊥` and `⊤` symbols can be imported with `import cats._`.
```
kailuowang: `:<:` and `:≺:` was added to cats package in #1557

kailuowang:
@zainab-ali do you plan to continue this one?

zainab-ali:
Yes - I think I've got all the symbols, but can't find a definitive way of checking.  Any ideas are welcome.

rossabaker:
Would nicknames be useful here?  I've heard "scream" or "cinnabon" for `|@|` and "spaceship" for `<=>` and condescension toward those who don't know.  I don't want to encourage the jargon or especially the condescension, but maybe a slang column here could assist communication of these operators.

zainab-ali:
I like the idea - especially as part of the FAQ.  I think @fommil had some ideas on names earlier.

fommil:
nice! I'm going to print this out and keep it on my desk

zainab-ali:
@rossabaker where / what is `<=>`?  I just realized I missed it from the list

rossabaker:
I might have just had a Perl flashback with `<=>`.

There is a `<->` in [cats-laws](https://github.com/typelevel/cats/blob/f73752ccc92d6693de27572ffa508d2dd32fe09a/laws/src/main/scala/cats/laws/package.scala#L5).

kailuowang:
@zainab-ali got any time to continue this one, if not I'll be more than happy to take over.

zainab-ali:
Sorry for taking so long on this one.

I've added the cinnabon / scream nickname, but am having trouble thinking of others.
GitHub has suddenly decided that the `|` inside the code block for `|@|` is a table delimiter, hence the odd rendering.  Jekyll renders the correct output, so this shouldn't be a problem on the site.

kailuowang:
@rossabaker , would you mind take another look?

rossabaker:
This is nice.  The table works for me on GitHub if I escape the pipes in `|@|` with a backslash.  I'm not sure what this does to Jekyll.

Only missing ones I can think of:
* [`<<<` and `>>>`](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/arrow/Compose.scala) from Compose
* [<->](https://github.com/typelevel/cats/blob/f73752ccc92d6693de27572ffa508d2dd32fe09a/laws/src/main/scala/cats/laws/package.scala#L5) from cats-laws.

I thought we'd have more nicknames when I suggested it, but Cats doesn't have left fish (`<=<`), right fish (`>=>`), or spaceship (`<=>`), so I guess that column is going to be a bit sparse.  That probably speaks well of Cats. :smile:

zainab-ali:
Jekyll didn't like the backslashes for me.  This [post](http://stackoverflow.com/questions/17319940/how-to-escape-a-pipe-char-in-a-code-statement-in-a-markdown-table) seems to do the trick, though it's somewhat of a mess. It works with both Jekyll and GitHub.

docs/src/main/tut/faq.md
```diff
@@ -202,6 +203,37 @@ If you're having trouble figuring out how to implement `tailRecM` lawfully, you
...
+| <code>fa &#124;@&#124; fb</code> | Cartesian builder      | Cinnabon, scream | `Cartesian[F[_]]`       | <code>&#124;@&#124;(fa: F[A])(fb: F[B]): F[(A, B)]</code> |
+| `fa *> fb`                       | right apply            |                  | `Cartesian[F[_]]`       | `*>(fa: F[A])(fb: F[B]): F[A]`                            |
+| `fa <* fb`                       | left apply             |                  | `Cartesian[F[_]]`       | `<*(fa: F[A])(fb: F[B]): F[B]`                            |
+| `x === y`                        | equals                 |                  | `Eq[A]`                 | `eqv(x: A, y: A): Boolean`                                |
+| `x =!= y`                        | not equals             |                  | `Eq[A]`                 | `neqv(x: A, y: A): Boolean`                               |
+| `fa >>= f`                       | flatMap                |                  | `FlatMap[F[_]]`         | `flatMap(fa: F[A])(f: A => F[B]): F[B]`                   |
+| `fa >> fb`                       | followed by            |                  | `FlatMap[F[_]]`         | `followedBy(fa: F[A])(fb: F[B]): F[B]`                    |
```
peterneyens: As of #1537 there is also `<<` (`FlatMap#forEffect`).

peterneyens:
Sorry for the late feedback. If you want to add `>>>` , `<<<` and `<<`, I think this is good to go :+1:

docs/src/main/tut/faq.md
zainab-ali: I've now added that

zainab-ali:
I've added `>>>`, `<<<` and `<<`

zainab-ali:
The build is failing with a python pip version error - is this a transient issue?

edmundnoble:
Is it perhaps worth it to include more fine-grained imports for the symbols than `cats.implicits._`?

kailuowang:
@edmundnoble I agree. This PR is already a big improvement from what we have. Shall we merge it for now and work on the fine-grained imports in another PR?

edmundnoble:
SGTM @kailuowang. Let's get this in.

PR #1576, opened 2017-03-27
Optimize Eval to use the stack most of the time

edmundnoble:
`Eval` *should* be able to avoid jumping on the trampoline most of the time, instead jumping on the stack until it's necessary to retreat (at some arbitrary depth). Preliminary (non-JMH) benchmarks indicate that the performance is about the same, but the GC pressure is about half. This makes `Eval` use more stack-space; up to 75 frames, so unsafe usage of `Eval` is more unsafe. Inspired by http://blog.higher-order.com/blog/2015/06/18/easy-performance-wins-with-scalaz/ (which makes for the sixth time I've visited that blog post).

### DIFF
--- a/core/src/main/scala/cats/Eval.scala
+++ b/core/src/main/scala/cats/Eval.scala
@@ -74,30 +74,46 @@ sealed abstract class Eval[+A] extends Serializable { self =>
   def flatMap[B](f: A => Eval[B]): Eval[B] =
     this match {
       case c: Eval.Compute[A] =>
-        new Eval.Compute[B] {
-          type Start = c.Start
-          // See https://issues.scala-lang.org/browse/SI-9931 for an explanation
-          // of why the type annotations are necessary in these two lines on
-          // Scala 2.12.0.
-          val start: () => Eval[Start] = c.start
-          val run: Start => Eval[B] = (s: c.Start) =>
-            new Eval.Compute[B] {
-              type Start = A
-              val start = () => c.run(s)
-              val run = f
-            }
+        // magic constant
+        if (c.count < 75) {
+          // pay a stack frame
+          new Eval.Compute[B] {
+            // See https://issues.scala-lang.org/browse/SI-9931 for an explanation
+            // of why the type annotations are necessary in these two lines on
+            // Scala 2.12.0.
+            type Start = c.Start
+            val start: () => Eval[Start] = c.start
+            val run: Start => Eval[B] = f.compose(c.run(_).value)
+            override private[cats] val count = c.count + 1
+          }
+        } else {
+          // pay a trampoline jump
+          new Eval.Compute[B] {
+            type Start = c.Start
+            val start: () => Eval[Start] = c.start
+            val run: Start => Eval[B] = (s: c.Start) =>
+              new Eval.Compute[B] {
+                type Start = A
+                val start = () => c.run(s)
+                val run = f
+                override private[cats] val count = 0
+              }
+            override private[cats] val count = 0
+          }
         }
       case c: Eval.Call[A] =>
         new Eval.Compute[B] {
           type Start = A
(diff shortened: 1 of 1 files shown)


alexandru:
Just a note, I've tried this optimisation in the initial version of the Monix `Task` and I've ultimately decided against doing it, because I ended with the conclusion that you can't avoid bumping into stack overflows.

Question: what do you think the call stack depth is going to be in this case?

```scala
def plus3(x: Int) = plus2(x) +1
def plus2(x: Int) = plus1(x) + 1
def plus1(x: Int) = x + 1

def loop(n: Int, acc: Int) = 
  Eval.defer {
    if (n <= 0) Eval.now(acc) else
      loop(n - 1, plus3(acc))
  }
```

Well, in case of a regular trampoline it is going to be about ~ 3, plus whatever the trampoline itself does, so lets say 4 or 5 frames.

But in your case it's going to be at least `75 * 3` (75 here is whatever magic constant you're going to use) + whatever your implementation takes.  And this multiplying factor is a problem. You see, it's not just your `Eval` trampoline taking up space, it's everything else as well that's done per each loop cycle.

edmundnoble:
I don't quite understand; I'm trying to measure the stack depth consumed, and not having much luck reproducing that:
```scala
def consumeStackFrames[A](n: Int, f: A => A): A => A =
  if (n == 0) f else { (identity[A] _) andThen consumeStackFrames(n - 1, f) }

def freshEx(maxStackDepth: Int): Int = 
  Math.max(maxStackDepth, new Exception().getStackTrace().length)

def loop(n: Int, stackDepth: Int): Eval[Int] = Eval.defer {
  if (n <= 0) Eval.now(stackDepth)
  else loop(n - 1, consumeStackFrames(100, freshEx)(stackDepth))
}

scala> loop(2000, 0).value
res8: Int = 146

```

It seems (to me) like it performs additively. But again, I am not sure I'm measuring correctly.

djspiewak:
@alexandru @edmundnoble This PR seems really cool.  Bumping to avoid having it die on the vine…

I spent some time trying to understand your (Alex) example, and I'm actually coming up empty.  As with Edmund, I'm pretty sure it performs additively.  But critically, you wouldn't really formulate it in that way if you were using `Eval` as a trampoline.  Something more like this seems more realistic:

```scala
def plus3(x: Int) = plus2(x).map(1 +)
def plus2(x: Int) = plus1(x).map(1 +)
def plus1(x: Int) = Eval.now(x + 1)

def loop(n: Int) = 
  if (n <= 0) plus3(0) else loop(n - 1).flatMap(plus3)
```

And that should have a maximum stack depth of 75.

johnynek:
I really like anything that can improve performance but I hope we can find the time to benchmark it before we merge.

edmundnoble:
@johnynek (disclaimer: I did not use JMH) It very much depends on the loop count; at the (very conservative, 75) count I have in this PR, there is no decrease in time spent; increasing it to 200 got me up to 30% faster. However, even with this loop count, the GC pressure is one 75th of what it was before, and that's one of the biggest issues with Eval.

djspiewak:
This gets a 👍 from me for merge, barring serious ongoing objections.  It's been measured, and I'm still having trouble replicating Alex's objection from earlier.

Obviously build failures need to be fixed though.

edmundnoble:
@djspiewak I am not convinced this approach is safe, unfortunately. The `StateT` test failure looks to be legitimate. It may be worth just adding `Steque` to cats and using that to make `Eval` create a balanced tree.

djspiewak:
I have an evil way to fix the problem if you're interested. :-P

edmundnoble:
It turned out that the evil way, despite being wholly evil, was not evil enough to work. I'm retiring this PR for now, and I'll keep looking at optimizations for Eval.

PR #1578, opened 2017-03-27
add last, sortBy and sorted to NonEmptyList

julien-truffaut:


core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -217,6 +219,38 @@ final case class NonEmptyList[+A](head: A, tail: List[A]) {
...
+    * scala> import cats.instances.int._
+    * scala> val nel = NonEmptyList.of(('a', 4), ('z', 1), ('e', 22))
+    * scala> nel.sortBy(_._2)
+    * res0: cats.data.NonEmptyList[(Char, Int)] = NonEmptyList((z,1), (a,4), (e,22))
+    * }}}
+    */
+  def sortBy[B: Order](f: A => B): NonEmptyList[A] = {
+    toList.sortBy(f)(Order[B].toOrdering) match {
+      case x :: xs => NonEmptyList(x, xs)
+      case Nil     => sys.error("absurd")
```
julien-truffaut: I am not particularly proud of this, suggestions welcome!

julienrf: This is not really a solution but you can “disable” exhaustivity checking instead:

~~~ scala
match {
  case List(x, xs @ _*) => NonEmptyList(x, xs)
}
~~~

I would add a comment, though, to emphasize that we can not get an empty list since we got our list by converting the `NonEmptyList` into a `List`

johnynek: I think the current thing is fine except I would say: `=> sys.error("unreachable: reduced the number of items in a NonEmptyList after a sort")` or something that is somewhat more of a comment.

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -217,6 +219,38 @@ final case class NonEmptyList[+A](head: A, tail: List[A]) {
...
+    * {{{
+    * scala> import cats.data.NonEmptyList
+    * scala> import cats.instances.int._
+    * scala> val nel = NonEmptyList.of(12, 4, 3, 9)
+    * scala> nel.sorted
+    * res0: cats.data.NonEmptyList[Int] = NonEmptyList(3, 4, 9, 12)
+    * }}}
+    */
+  def sorted[AA >: A](implicit AA: Order[AA]): NonEmptyList[AA] =
+    sortBy(a => a: AA)
```
johnynek: I'd rather not pay the cost to apply the identity function here. Can we instead directly implement?

core/src/main/scala/cats/data/NonEmptyList.scala
julien-truffaut: sure, I will update the PR

julien-truffaut:
rebased

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -19,6 +19,8 @@ final case class NonEmptyList[+A](head: A, tail: List[A]) {
    */
   def toList: List[A] = head :: tail
 
+  def last: A = tail.lastOption.getOrElse(head)
```
julien-truffaut: sure

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -219,6 +224,40 @@ final case class NonEmptyList[+A](head: A, tail: List[A]) {
...
+    * {{{
+    * scala> import cats.data.NonEmptyList
+    * scala> import cats.instances.int._
+    * scala> val nel = NonEmptyList.of(('a', 4), ('z', 1), ('e', 22))
+    * scala> nel.sortBy(_._2)
+    * res0: cats.data.NonEmptyList[(Char, Int)] = NonEmptyList((z,1), (a,4), (e,22))
+    * }}}
+    */
+  def sortBy[B: Order](f: A => B): NonEmptyList[A] =
+    toList.sortBy(f)(Order[B].toOrdering) match {
```
peterneyens: I think we generally use an implicit parameter (compared to a context bound) when we actually use the the type class instance?

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -219,6 +224,40 @@ final case class NonEmptyList[+A](head: A, tail: List[A]) {
...
+    * {{{
+    * scala> import cats.data.NonEmptyList
+    * scala> import cats.instances.int._
+    * scala> val nel = NonEmptyList.of(12, 4, 3, 9)
+    * scala> nel.sorted
+    * res0: cats.data.NonEmptyList[Int] = NonEmptyList(3, 4, 9, 12)
+    * }}}
+    */
+  def sorted[AA >: A](implicit AA: Order[AA]): NonEmptyList[AA] =
+    toList.sorted(Order[AA].toOrdering) match {
```
peterneyens: Should we do `AA.toOrdering`?

core/src/main/scala/cats/data/NonEmptyList.scala
julien-truffaut: let's be consistent

PR #1579, opened 2017-03-30
Add Edmund Noble to maintainers.

non:
I've already polled the current maintainers and they were unanimous in wanting to add Edmund to the list. I also spoke to Edmund and he agreed, so this PR is just a formality at this point.


edmundnoble:
Thanks very much :+1:

julien-truffaut:
👍

PR #1583, opened 2017-04-01
Enable SI-2712 fix in cats / Remove unapply machinery

### LINKED ISSUE #1073: Figure out how to handle SI-2712 fix

kailuowang:
fixes #1073
I also added an FAQ item to help people running into SI-2712


tests/src/test/scala/cats/tests/RegressionTests.scala
```diff
@@ -85,7 +85,7 @@ class RegressionTests extends CatsSuite {
 
   test("#500: foldMap - traverse consistency") {
     assert(
-      List(1,2,3).traverseU(i => Const(List(i))).getConst == List(1,2,3).foldMap(List(_))
+      List(1,2,3).traverse(i => Const.of[List[Int]](List(i))).getConst == List(1,2,3).foldMap(List(_))
```
kailuowang: for some reason the `traverse` version can no longer infer the type correctly. Thus I had to pass the type into `Const`'s apply method. For convenience, I added an `of`  factory method to `Const`. Please shout out if you know a better solution.

core/src/main/scala/cats/data/Const.scala
```diff
@@ -39,6 +39,12 @@ final case class Const[A, B](getConst: A) {
 object Const extends ConstInstances {
   def empty[A, B](implicit A: Monoid[A]): Const[A, B] =
     Const(A.empty)
+
+  class partialApply[B] {
```
adelbertc: Perhaps change this to something like `OfPartiallyApplied` to be consistent with https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/EitherT.scala#L275

tests/src/test/scala/cats/tests/RegressionTests.scala
edmundnoble: That's not nice. Perhaps `Const` needs its type params reversed?

kailuowang: would that affect partial unification?

djspiewak: Partial unification is strictly left to right. If Const varies on the left (can't look right now), then it wouldn't work with partial unification.as a rule of thumb, think about how you would declare the type constructor in Haskell.

kailuowang: @djspiewak it varies on the right. @edmundnoble any other suggestion in terms what to do with this?

adelbertc: Might it have something to do with Scala inferring `Nothing` on the unused type parameter which can make the compiler freak out since it treats `Nothing` specially?

ceedubs: I think that specifying the type parameter here is an acceptable tradeoff for getting rid of the `Unapply` machinery, but paging @milessabin to see if he has any thoughts on this popping up.

kailuowang: Here is the minimised version in case anyone is interested. 
Here is the minimized version
```Scala
    def trav[F[_], A, B](f: A => F[B]): F[B] = ???
    case class C[A, B](a: A)
    val c: C[Int, Int] = trav(C(_:Int))   //compiles
    val a: Int  = trav(C(_:Int)).a    //doesn’t compile
```
> [error]  found   : Int => C[Int,Nothing]
[error]  required: Int => C[Int,B]
[error]     val a: Int  = trav(C(_:Int)).a

We might have to live with this as is.

djspiewak: What happens if you make `C` covariant in `B`?

djspiewak: Nevermind, I just tried it myself.  It resolves the issue.

kailuowang:
Thanks to @djspiewak, the type inference glitch is resolved. Does any maintainer have time to give it another look?

ceedubs:
I'm not sure about making `B` covariant in `Const` to get around the inference. Since it's a phantom type it could just as easily be treated as contravariant. I'm hesitant to choose one over the other in the name of (questionable) inference.

There's some more information about phantom types and variance [on the typelevel blog](http://typelevel.org/blog/2016/09/19/variance-phantom.html) about phantom types in scala for anyone who is interested.

ceedubs:
cc @djspiewak about my last message. You had good insight that making the parameter covariant  might get around the inference issue. What are your thoughts on actually doing this for `Const` though?

kailuowang:
@ceedubs is it a phantom type? I'm literally not sure. It appears in both covariant and contravariant positions (as part of type `Const[A, B]`).

I agree it that feels hackish when the only reason to chose this variance is for type inference. On the other hand, for `Const`, I am not aware of any need of subtyping a `Const[A, B]`(since there is `retag`?), so I am not sure about the downside of choosing a variance either.

djspiewak:
@ceedubs It could certainly be either covariant or contravariant, and so by that argument would *correctly* be invariant.  However, Scala's type inference biases downward, not upward and not "in place".  It always *tries* to drive things to `Nothing`.  For that reason, covariant is the correct default in any phantom position where inference is relevant.

Though I agree it seems awfully arbitrary.

@kailuowang As long as there's no need to subtype `Const`, I think it's safe to just toss the `+` on there and call it a day.  We should make it clear in the release notes for RC1 that this was done, so people who use `Const` can take a crack at it and see if their own particular niche use-case is broken.

milessabin:
I share everyone's discomfort, but I think this is the right thing to do.

ceedubs:
Thanks for the input @djspiewak and @milessabin. @djspiewak you raise some good points, and I don't think that I'm opposed to the `B` becoming covariant in `Const`.

@kailuowang thank you I'm very excited about this change!

I assume that kittens will have to be updated in a similar way (especially since `Unapply` has been removed from Cats)? I'd be really interested in seeing how much simpler the heterogeneous `HList` sequencing becomes without the need for `Unapply`. It might be nice to give that a spin against this version of Cats before merging this PR just to make sure that everything still works smoothly. I won't have time today, but I could give that a try tomorrow or this weekend if nobody beats me to it.

I think that it's outside of the scope of this PR, but I wonder if it would be helpful to have a FAQ item for "Why doesn't the code example in the Cats docs compile for me?" which could have an answer that links to both the kind-projector and si-2712 items.

ceedubs:
@kailuowang sorry but it looks like there are merge conflicts :(

kailuowang:
Resolved the conflicts and squashed. 
I also brought back the `Unapply` class. It has ueses other than overcoming SI-2712. Given the experience in kittens, I believe there is good chance that some users are using it to help resolve implicit search.
I also tried kittens with this branch of code without using `Unapply` though. It worked out fine. Here is the [PR](https://github.com/milessabin/kittens/pull/36)

ceedubs:
> I also brought back the Unapply class. It has ueses other than overcoming SI-2712. Given the experience in kittens, I believe there is good chance that some users are using it to help resolve implicit search.

@kailuowang I don't see any `Unapply` references in the current kittens code. Is there something in particular that you are referring to? I was pretty happy about the prospect of dropping `Unapply` from Cats :)

kailuowang:
@ceedubs, sorry I should've been more specific, it was late when I wrote the last update. `Unapply` is used in a situation where you are given specific type 'F`, but you wanna try find out if this is actually a `G[T]` and there is an instance of `H[G[_]]`. You can achieve this using a different trick like [here in kittens](https://github.com/milessabin/kittens/blob/master/core/src/main/scala/cats/sequence/sequence.scala#L98), but I suspect it will take user some time to discover that. So I am not sure if we should simply remove it. Maybe we can deprecate it instead?

ceedubs:
@kailuowang hmm, I wonder if there are many places where code outside of Cats is using the Cats `Unapply`. I'm a bit inclined to think that after the SI-2712 fix this isn't really something that belongs in Cats but could potentially live elsewhere if libraries really still need it. But I'm definitely open to other viewpoints. Does anyone else want to weigh in on this?

Aside from the question of whether there should still be an `Unapply` in Cats, 👍 on this PR.

ceedubs:
Oops I'm bad at the GitHub review system. I don't think I submitted https://github.com/typelevel/cats/pull/1583#discussion_r111368823 until just now.

johnynek:
So, the side effect of this change is that `Const[Int, Int]` is going to be usable where a `Const[Int, Any]` is needed. This seems a bit weird to me, since the type tag has now become an upper bound tag, not the exact value.

I imagine this will break some use cases.

It is certainly the case that in principle a phantom type can be set to be covariant, but the meaning of the type is considerably changed, isn't it?

kailuowang:
@ceedubs @johnynek I reverted the change of making `Const` covariant on the phantom type and created a separate issue #1608. I think that is a separate issue from removing `Unapply` machinery. 
As for the `Unapply` class, I also created a new ticket #1607. I am all for moving it out of cats, but creating a new repo and release take some time and this PR touches a lot of files and is blocking other PRs. So maybe we can address it in a separate PR?

ceedubs:
It's been a while and there have been some changes since @adelbertc's review, so I'm going to wait for another review or two before merging.

core/src/main/scala/cats/data/Const.scala
```diff
@@ -6,6 +6,7 @@ import cats.functor.Contravariant
 /**
  * [[Const]] is a phantom type, it does not contain a value of its second type parameter `B`
  * [[Const]] can be seen as a type level version of `Function.const[A, B]: A => B => A`
+ * B is set covariant to help type inference.
```
notxcain: You meant `invariant`?

core/src/main/scala/cats/data/Const.scala
```diff
@@ -6,6 +6,7 @@ import cats.functor.Contravariant
...
  */
 final case class Const[A, B](getConst: A) {
```
notxcain: Oh, I saw a discussion below, I think you forgot to add `+B` here

core/src/main/scala/cats/data/Const.scala
notxcain: Oh, I saw a discussion below, I think you forgot to make `B` covariant in `Const` definition

kailuowang: we reverted that in this PR and created a new Issue #1608 for it, but I forgot to remove this sentence. Thanks for spotting it!

docs/src/main/tut/faq.md
```diff
@@ -47,13 +49,22 @@ There are a few minor mismatches between `Xor` and `Either`. For example, in som
 
 Similarly, `cats.data.XorT` has been replaced with `cats.data.EitherT`, although since this is a type defined in Cats, you don't need to import syntax or instances for it (although you may need imports for the underlying monad).
 
+## <a id="si-2712" href="#si-2712"></a>Why is the compiler having trouble with types with more than one type parameters?
```
edmundnoble: s/parameters/parameter

ceedubs:
👍 LGTM.

@johnynek would you want to take another look after the most recent changes?

johnynek:
👍

PR #1584, opened 2017-04-03
Move arbitrary instance of StateT to laws

### LINKED ISSUE #1274: Add Arbitrary instances for all data types

kailuowang:
This should be the only data type missing `Arbitrary` instance in laws. 
fixes #1274

### DIFF
--- a/tests/src/test/scala/cats/tests/KleisliTests.scala
+++ b/tests/src/test/scala/cats/tests/KleisliTests.scala
@@ -16,7 +16,7 @@ class KleisliTests extends CatsSuite {
     Eq.by[Kleisli[F, A, B], A => F[B]](_.run)
 
   implicit def readerEq[A, B](implicit A: Arbitrary[A], FB: Eq[Id[B]]): Eq[Reader[A, B]] =
-    Eq.by[Reader[A, B], A => Id[B]](_.run)
+    kleisliEq
 
   implicit val eitherTEq = EitherT.catsDataEqForEitherT[Kleisli[Option, Int, ?], Unit, Int]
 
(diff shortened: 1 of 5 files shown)


peterneyens:
As we are providing `Arbitrary` instances for `Writer` and `State`, should we also provide one for `Reader`?

kailuowang:
good catch @peterneyens , added.

kailuowang:
I added an `Arbitrary` for reader `Reader` by simply delegating to `arbitraryForKleisli`, but there is not tests for `Reader` to use it, hence the coverage miss. I think that should be okay?

ceedubs:
@kailuowang I'm hesitant to add it without any tests that use it. Especially because this is the sort of thing that can easily lead to ambiguous instances. I'd be inclined to add a test and even see if we could get by without the instances that are specific to `Id` (though I suspect that implicit search won't be that kind to us).

On a separate note, our instances for function-like types like `Kleisli` should probably be using `Cogen` so they don't just ignore the input and return a static value. But that is probably outside of the scope of this PR. I'll create an issue for it.

kailuowang:
@ceedubs I added a test for `Reader` to hit the arbitrary instance created for it. I also verified that we can't get by without this instance specific to `Id`, i.e. removing the instance failed the compilation of the test.

tests/src/test/scala/cats/tests/KleisliTests.scala
```diff
@@ -15,6 +15,9 @@ class KleisliTests extends CatsSuite {
   implicit def kleisliEq[F[_], A, B](implicit A: Arbitrary[A], FB: Eq[F[B]]): Eq[Kleisli[F, A, B]] =
     Eq.by[Kleisli[F, A, B], A => F[B]](_.run)
 
+  implicit def readerEq[A, B](implicit A: Arbitrary[A], FB: Eq[Id[B]]): Eq[Reader[A, B]] =
+    Eq.by[Reader[A, B], A => Id[B]](_.run)
```
ceedubs: Minor, but this could probably just delegate to the `kleisliEq` couldn't it?

ceedubs:
@peterneyens do you want to take another look after the most recent changes?

PR #1586, opened 2017-04-04
Add Applicative.unit

alexandru:
Adding a short-hand to `Applicative` for `F.pure(())`.

```scala
@typeclass trait Applicative[F[_]] extends Apply[F] { self =>
  /**
   * Returns an `F[Unit]` value, equivalent with `pure(())`.
   *
   * A useful shorthand, also allowing implementations to optimize the
   * returned reference (e.g. it can be a `val`).
   */
  def unit: F[Unit] = pure(())
```

Reasons:

1. I find myself doing `F.pure(()).map(_ => a)` a lot (due to `pure` not taking a by-name param and having no way to do that otherwise, but that's another discussion)
2. `F[Unit]` is often used to indicate callbacks that are going to trigger side-effects, like for resource clean-ups - for example in `Task.doOnFinish` and we often need to create empty `F[Unit]` instances that don't do anything; `F.pure(())` is more noisy versus `F.unit`
3. `Applicative` implementations can optimize this to always return the same reference, instead of rebuilding `pure(())` over and over again

Prior art:

```scala
scala.concurrent.Future.unit
```

### DIFF
--- a/core/src/main/scala/cats/Applicative.scala
+++ b/core/src/main/scala/cats/Applicative.scala
@@ -22,6 +22,14 @@ import simulacrum.typeclass
    */
   def pure[A](x: A): F[A]
 
+  /**
+   * Returns an `F[Unit]` instance, equivalent with `pure(())`.
+   *
+   * A useful shorthand, also allowing implementations to optimize the
+   * returned reference (e.g. it can be a `val`).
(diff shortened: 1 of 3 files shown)


mpilquist:
:+1:

PR #1589, opened 2017-04-05
Rename prod coproduct

### LINKED ISSUE #1582: rename Prod / Coproduct to ProductK / CoproductK
The problem is that we already have `scala.Product` and `shapeless.Coproduct` and that means it is easy to get the wrong import.

In addition, it is very unfortunate that `cats` and `shapeless` (the foundations of this FP stack) should have a name conflict on something so simple.

I understand that the `K` postfix is appropriate for a higher kinded thing as a workaround until https://github.com/scala/scala/pull/5538 become ubiquitous. I suppose `M` would also be appropriate?

kailuowang:
fixes #1582
Note that this should be a fairly straightforward PR to review. However, github failed to render the diff in renamed files correctly - it displays the diff as a whole file removed and a new file. To overcome this, you might want to review the two commits separately.


johnynek:
thanks for doing this! I think this is a win for consistency and readability prior to 1.0.

👍

adelbertc:
Bikeshedding, but `Tuple2K` feels weird even though it's correct since it's a higher-order `Tuple2`, but what about just `TupleK` ?

kailuowang:
@adelbertc I don't dislike `TupleK`. I picked `Tuple2K` exactly due to the correctness, and `TupleK` sounds like it's arity free?

adelbertc:
yeahhhhh i just don't like having numbers in names for some reason.. but oh well 👍

edmundnoble:
@adelbertc I thought `PairK` might be nice for that reason, but that might drive people to use `scala.Predef.Pair`.

PR #1592, opened 2017-04-07
added instances of BitSet to allInstances

### LINKED ISSUE #1485: Instances missing from AllInstances
Not all type class instances are included in `cats.instances.AllInstances` (and thus they are also missing when `import cats.implicits._`). One example is the `Eq[BitSet]` instance. (See also the discussion at #1484.)

The reason is probably that they're included in `cats.kernel.instances.AllInstances`, but not in `cats.instances.AllInstances`. (Note, that the latter doesn't extend the former, but instead every separate `*Instances` trait tend to extend its pair in `kernel`. For example, `cats.instances.StringInstances` extends `cats.kernel.instances.StringInstances`.)

kailuowang:
Fixes #1485 , I double checked. I believe that `BitSet` is the only missing one.

### DIFF
--- a/core/src/main/scala/cats/instances/bitSet.scala
+++ b/core/src/main/scala/cats/instances/bitSet.scala
@@ -1,8 +1,6 @@
 package cats.instances
 
 import scala.collection.immutable.BitSet
-import int.catsStdShowForInt
-import cats.syntax.show._
 import cats.Show
 
 trait BitSetInstances extends cats.kernel.instances.BitSetInstances {
--- a/core/src/main/scala/cats/instances/bitSet.scala
+++ b/core/src/main/scala/cats/instances/bitSet.scala
@@ -6,8 +6,5 @@ import cats.syntax.show._
 import cats.Show
 
 trait BitSetInstances extends cats.kernel.instances.BitSetInstances {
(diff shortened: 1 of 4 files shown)


core/src/main/scala/cats/instances/bitSet.scala
```diff
@@ -0,0 +1,13 @@
...
+
+import scala.collection.immutable.BitSet
+import int.catsStdShowForInt
+import cats.syntax.show._
+import cats.Show
+
+trait BitSetInstances extends cats.kernel.instances.BitSetInstances {
+  implicit def catsStdShowForBitSet: Show[BitSet] = new Show[BitSet] {
+    def show(fa: BitSet): String =
+      fa.toIterator.map(_.show).mkString("BitSet(", ", ", ")")
```
johnynek: why would we use show here vs `fa.toString`? There is nothing generic about BitSet and the current implementation is just going to be the same as `toString` no?

kailuowang: duh, I wasn't thinking when I mechanically copied from set instances.

johnynek:
👍

PR #1594, opened 2017-04-07
Add NonEmptyList#fromFoldable

markus1189:
Adds a new method `NonEmptyList#fromFoldable`.  This generalizes `NonEmptyList.fromList` such that you can also create a `NonEmptyList` from anything that has a `Foldable` instance, e.g., `Vector`, `Set`, ...

### DIFF
--- a/core/src/main/scala/cats/data/NonEmptyList.scala
+++ b/core/src/main/scala/cats/data/NonEmptyList.scala
@@ -313,6 +313,9 @@ object NonEmptyList extends NonEmptyListInstances {
       case h :: t => NonEmptyList(h, t)
     }
 
+  def fromFoldable[F[_], A](fa: F[A])(implicit F: Foldable[F]): Option[NonEmptyList[A]] =
+    fromList(F.toList(fa))
+
   def fromReducible[F[_], A](fa: F[A])(implicit F: Reducible[F]): NonEmptyList[A] =
     F.toNonEmptyList(fa)
 }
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -313,6 +313,9 @@ object NonEmptyList extends NonEmptyListInstances {
       case h :: t => NonEmptyList(h, t)
     }
 
+  def fromFoldable[F[_], A](fa: F[A])(implicit F: Foldable[F]): Eval[Option[NonEmptyList[A]]] =
+    F.reduceRightToOption(fa)(x => NonEmptyList.of(x))((e,acc) => Later(e :: acc.value))
```
peterneyens: What about `NonEmptyList.fromList(F.toList(fa))` ?

The `Foldable#toList` implementation using a `mutable.ListBuffer`, which is overridden in most `Foldable` instances (eg `List` just returns the list itself, `Vector` calls `toList`, etc), should be more efficient than prepending a `NonEmptyList`.

markus1189:
Fine with me, I thought that going through a `List` is probably worse than folding it into an `NonEmptyList`

ceedubs:
Sorry about that I accidentally clicked the close button.

There's a small argument for adding this method to the `Foldable` type class so that the `NonEmptyList` `Foldable` instance could override it for efficiency. However, I guess you probably don't want to bloat the `Foldable` type class with a bunch of methods to create specific collection types. And also the implementation for `NonEmptyList` will still be quite efficient anyway.

👍 thanks!

PR #1596, opened 2017-04-08
Rename Inject to InjectK

andyscott:
Rename `Inject` to `InjectK` to be consistent with names/renames for `EitherK`, `FunctionK`, `Tuple2K`, and friends. See #1589.

This is a breaking API and breaking binary change.

Additionally, this opens the door to  "re-introduce" `Inject` for `Either`:

```scala
sealed abstract class Inject[A, B] {
  def inj: A => B
  def prj: B => Option[A]
  def apply(a: A): B = inj(a)
  def unapply(b: B): Option[A] = prj(b)
}

private[cats] sealed abstract class InjectInstances {
  implicit def catsReflexiveInjectInstance[A]: Inject[A, A] =
    new Inject[A, A] {
      val inj = identity
      val prj = Some(_)
    }

  implicit def catsLeftInjectInstance[A, B]: Inject[A, Either[A, B]] =
    new Inject[A, Either[A, B]] {
      val inj = Left(_)
      val prj = _.left.toOption
    }

  implicit def catsRightInjectInstance[A, B, C](implicit I: Inject[A, B]): Inject[A, Either[C, B]] =
    new Inject[A, Either[C, B]] {
      val inj = a => Right(I.inj(a))
      val prj = _.right.toOption.flatMap(I.prj)
    }

}

object Inject extends InjectInstances {
  def apply[A, B](implicit I: Inject[A, B]): Inject[A, B] = I
}
```

Inject for Either would be useful when combining error algebras across domains with `Free`.

### DIFF
--- a/core/src/main/scala/cats/InjectK.scala
+++ b/core/src/main/scala/cats/InjectK.scala
@@ -4,11 +4,12 @@ import cats.arrow.FunctionK
 import cats.data.EitherK
 
 /**
- * Inject type class as described in "Data types a la carte" (Swierstra 2008).
+ * The injection type class as described in "Data types a la carte"
+ * (Swierstra 2008).
  *
  * @see [[http://www.staff.science.uu.nl/~swier004/publications/2008-jfp.pdf]]
  */
-sealed abstract class Inject[F[_], G[_]] {
+sealed abstract class InjectK[F[_], G[_]] {
   def inj: FunctionK[F, G]
 
   def prj: FunctionK[G, λ[α => Option[F[α]]]]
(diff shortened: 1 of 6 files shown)


kailuowang:
Makes sense to me. 👍  Can we also create a ticket for the new `Inject`? It would be a bit awkward to have `InjectK` but not `Inject`.

andyscott:
Sure, and I'll put up a PR up for Inject if this one gets merged.

ceedubs:
Thanks @andyscott this looks good to me. I only have one minor suggestion: there are a couple places that now say "InjectK type class as described in "Data types a la carte". This might be slightly confusing to readers, since that paper doesn't mention an `InjectK`. Maybe we could just change this text to "The injection type class as described..." or something?

andyscott:
@ceedubs Good call, I'll address that now

Issue #1599, opened 2017-04-09
Provide an Inject for Either


andyscott:
See https://github.com/typelevel/cats/pull/1596.

If `Inject` is renamed to `InjectK`, an an Inject can be provided for `Either`.

kailuowang:
remove from project since there is already a PR to represent this work .

kailuowang:
fixed by #1600

PR #1600, opened 2017-04-09
Inject for either

andyscott:
Re-introduces `Inject`, but for `Either`. I also went ahead and tried to document both `Inject` and `InjectK`.

Todo:
- [x] Add Inject
- [x] Scaladoc for Inject and InjectK
- [x] Add tests

Associated issue: https://github.com/typelevel/cats/issues/1599

### DIFF
--- a/core/src/main/scala/cats/Inject.scala
+++ b/core/src/main/scala/cats/Inject.scala
@@ -1,0 +1,54 @@
+package cats
+
+/**
+  * Inject is a type class providing an injection from type `A` into
+  * type `B`. An injection is a function `inj` which does not destroy
+  * any information: for every `b: B` there is at most one `a: A` such
+  * that `inj(a) = b`.
+  *
+  * Because of this all injections admit partial inverses `prj` which
+  * pair a value `b: B` back with a single value `a: A`.
+  *
+  * @since 1.0
+  * @note Prior to cats 1.0, Inject handled injection for type
+  * constructors. For injection of type constructors, use [[InjectK]].
+  *
+  * @see [[InjectK]] for injection for [[cats.data.EitherK]]
+  */
+abstract class Inject[A, B] {
+  def inj: A => B
+
+  def prj: B => Option[A]
+
+  final def apply(a: A): B = inj(a)
+
+  final def unapply(b: B): Option[A] = prj(b)
+}
+
+private[cats] sealed abstract class InjectInstances {
+  implicit def catsReflexiveInjectInstance[A]: Inject[A, A] =
+    new Inject[A, A] {
+      val inj = identity(_: A)
+
+      val prj = Some(_: A)
+    }
+
+  implicit def catsLeftInjectInstance[A, B]: Inject[A, Either[A, B]] =
+    new Inject[A, Either[A, B]] {
+      val inj = Left(_: A)
+
+      val prj = (_: Either[A, B]).left.toOption
+    }
+
+  implicit def catsRightInjectInstance[A, B, C](implicit I: Inject[A, B]): Inject[A, Either[C, B]] =
+    new Inject[A, Either[C, B]] {
+      val inj = (a: A) => Right(I.inj(a))
+
+      val prj = (_: Either[C, B]).right.toOption.flatMap(I.prj)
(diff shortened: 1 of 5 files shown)


andyscott:
I'll rebase/finish this after settling #1725

core/src/main/scala/cats/Inject.scala
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * @see [[http://www.staff.science.uu.nl/~swier004/publications/2008-jfp.pdf]]
+ * @see [[InjectK]] for injection for [[cats.data.EitherK]]
+ */
+abstract class Inject[A, B] {
+  def inj: A => B
+
+  def prj: B => Option[A]
+
+  def apply(a: A): B = inj(a)
```
kailuowang: shall we make these two final as well?

andyscott: Good catch!

core/src/main/scala/cats/Inject.scala
```diff
@@ -0,0 +1,54 @@
+package cats
+
+/**
+ * Inject is a type class to inject a value of type `A` into a
+ * coproduct of types `B` when `A :≺: B`. The coproduct `B` is
+ * modelled with `Either`.
```
kailuowang: It's true that it is the original incentive for introducing this class, but now that we made a type class, this is just one specific usage/instances of `Inject` and `InjectK` right? 
We could have an `Inject` between `A` and `List[A]` or `Int` and `Double`, right?

andyscott: Yes, that's correct. I suppose it could be reworded to indicate that that's just the default behavior?
```
Inject is a type class to inject a value of type `A` into another type `B`.
The default instances provided by cats injects `A` into coproduct of types
`B` when `A :≺: B`. The coproduct `B` is modelled with `Either`.
```

Also in this explanation, `:≺:` is used in a conceptual sense. `:≺:` exists as a type alias for `InjectK` which could be confusing. Thoughts?

kailuowang: My understanding of the `Inject` type class is that it represents an injective function between type `A` and `B`.  If we treat types as sets, and a function between the two types are just pairs of elements in the two sets `A` and `B`. Each pair has an "input" element in set `A` and an "output" element of `B`.  For an injective function, for every element in `A` there is exactly one pair. And for every element in `B` there is either one pair or no pair at all. 

A total pure scala function is an injective function, but it only specifies pairs for all elements in `A`, i.e. given any `A` you can get a `B`. It doesn't specify pairs for elements in `B`, this `Inject` type class does that through `proj`.

edmundnoble: Agreed with @kailuowang; I don't think this is particularly informative as to what `Inject` does, especially because we're likely going to provide these other instances in the future from cats.  I'd pick something like:

Inject is a type class providing an injection `f` from type `A` into another type `B`.
An injection is a function which does not destroy any information, and thus for every `b: B` there is at most one `a: A` such that `f(a) = b`.

Because of this all injections admit partial inverses, functions `B => Option[A]`, which pair a value `b: B` with the single value `a: A` that `f` sends to `b` if it exists.

andyscott: @edmundnoble I like your explanation. I tweaked it a tad-- how about:
```
/**
 * Inject is a type class providing an injection from type `A` into
 * another type `B`.  An injection is a function `inj` which does not
 * destroy any information: for every `b: B` there is at most one
 * `a: A` such that `inj(a) = b`.
 *
 * Because of this all injections admit partial inverses `prj` which
 * pair a value `b: B` back with a single value `a: A`.
 *
 * @since 1.0
 * @note Prior to cats 1.0, Inject handled injection for type
 * constructors. For injection of type constructors, use [[InjectK]].
 *
 * @see [[InjectK]] for injection for [[cats.data.EitherK]]
 */
abstract class Inject[A, B] // ...
```

I'll give the same treatment to `InjectK` once proper doc is decided.

kailuowang:
merge with 2 sign-offs

PR #1601, opened 2017-04-10
Recover Kleisli from F[_] errors

leandrob13:
Added functions for recovering from errors from F[_]. 

The functions recover, recoverWith and recoverWithF need an ApplicativeError instance for F[_]. 

```scala
val k = Kleisli[Either[String, ?], Int, Int] { i =>
  if (i < 0) Left("negative") else Right(i)
}
val r = k.recover[String] { case s => 0 }
val rw = k.recoverWith[String] { case s => Kleisli.lift[Either[String, ?], Int, Int](Right(0)) }
val rwf = k.recoverWithF[String] { case s => Right(0) }
val res = (r.run(-2), rw.run(-2), rwf.run(-2)) // should be (Right(0), Right(0), Right(0))
```

### DIFF
--- a/core/src/main/scala/cats/data/Kleisli.scala
+++ b/core/src/main/scala/cats/data/Kleisli.scala
@@ -60,6 +60,15 @@ final case class Kleisli[F[_], A, B](run: A => F[B]) { self =>
   def second[C](implicit F: Functor[F]): Kleisli[F, (C, A), (C, B)] =
     Kleisli{ case (c, a) => F.map(run(a))(c -> _)}
 
+  def recover[C](pf: PartialFunction[C, B])(implicit ae: ApplicativeError[F, C]): Kleisli[F, A, B] =
+    Kleisli( a => ae.recover(run(a))(pf))
+
+  def recoverWith[C](pf: PartialFunction[C, Kleisli[F, A, B]])(implicit ae: ApplicativeError[F, C]): Kleisli[F, A, B] =
+    Kleisli( a => ae.recoverWith(run(a))(pf andThen(_.run(a))))
+
+  def recoverWithF[C](pf: PartialFunction[C, F[B]])(implicit ae: ApplicativeError[F, C]): Kleisli[F, A, B] =
+    Kleisli( a => ae.recoverWith(run(a))(pf) )
+
   def apply(a: A): F[B] = run(a)
 }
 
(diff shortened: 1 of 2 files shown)


kailuowang:
I might missed something but the [`ApplicativeError` instance for `Kleisli`](https://github.com/leandrob13/cats/blob/adc8534b60960ba6a8b2be0b13122904e768ceff/core/src/main/scala/cats/data/Kleisli.scala#L125) should've provided these, right?

leandrob13:
@kailuowang you are right. I rewrote it as this:

```scala
def recover[E](pf: PartialFunction[E, B])(implicit aek: ApplicativeError[Kleisli[F, A, ?], E]): Kleisli[F, A, B] =
    aek.recover[B](this)(pf)
```
That should do it. I'll fix the other two.

peterneyens:
I agree with @kailuowang. You get `recover` and `recoverWith` from the `ApplicativeError` syntax.

You might need to use a Scala version which support `Ypartial-unification` or the SI 2712 compiler plugin, to get the syntax or use a type alias like `type K[A] = Kleisli[Either[String, ?], String, A]` to be able to get the syntax.

Merging #1583 will definitely help in this case as well.

leandrob13:
@kailuowang @peterneyens made the changes, didn't have problems running those tests. Just had to add the implicit instances for Kleisli with either and Kleisli with validated.

kailuowang:
@leandrob13 what we are saying is that the methods you added are redundant, they are already available through `ApplicativeError` syntax. Did you try enable this sbt plugin (https://github.com/fiadliel/sbt-partial-unification) and see if you can get those methods out of the box (cats 0.9.0) already?

leandrob13:
@kailuowang I just thought that they could be convenient. If you would want to recover a Kleisli you would have to invoke:

```scala
ApplicativeError[Kleisli[Either[String, ?], Int, ?], String].recover(k) { ... }
``

Am I right? I just thought it would be useful to do add a function that does this directly on Kleisli.

```scala
k.recover { ... }
```

Right now I am setting up the test for a use case that may have issues with the partial unification, I will try the plugin you suggest.

kailuowang:
@leandrob13,  you don't have to write `ApplicativeError[Kleisli[Either[String, ?], Int, ?], String].recover(k) { ... }`


```Scala
val k = Kleisli((i:Int) => if(i == 0) Left(1) else Right("big") ) 

val res13 = k.recover{ case 1 => "haha" } 

res13.run(3) 
// res14: Either[Int, String] = Right("big")

res13.run(0) 
// res16: Either[Int, String] = Right("haha")
```

For this to compile, you need partial-unification enabled. But usually the methods on typeclasses are "injected" directly onto the instance through the typeclass's syntax implicit conversion.

leandrob13:
@kailuowang I had no idea you could do that. Thanks, this is going to be of great help! I will close the PR.

PR #1602, opened 2017-04-12
Stack-safe Coyoneda

edmundnoble:
Makes Coyoneda's `map` stack-safe, by keeping a list of functions and reversing it before applying them.
Breaking change.

### DIFF
--- a/free/src/main/scala/cats/free/Coyoneda.scala
+++ b/free/src/main/scala/cats/free/Coyoneda.scala
@@ -8,6 +8,7 @@ import cats.arrow.FunctionK
  * This is isomorphic to `F` as long as `F` itself is a functor.
  * The homomorphism from `F[A]` to `Coyoneda[F,A]` exists even when
  * `F` is not a functor.
+ * Implemented using a List of functions for stack-safety.
  */
 sealed abstract class Coyoneda[F[_], A] extends Serializable { self =>
 
@@ -17,10 +18,13 @@ sealed abstract class Coyoneda[F[_], A] extends Serializable { self =>
   /** The underlying value. */
   val fi: F[Pivot]
 
-  /** The transformer function, to be lifted into `F` by `run`. */
-  val k: Pivot => A
+  /** The list of transformer functions, to be composed and lifted into `F` by `run`. */
+  private[cats] val ks: List[Any => Any]
 
-  import Coyoneda.{Aux, apply}
+  /** The list of transformer functions composed into a single function, to be lifted into `F` by `run`. */
+  final def k: Pivot => A = Function.chain(ks.reverse)(_).asInstanceOf[A]
+
+  import Coyoneda.{Aux, unsafeApply}
 
   /** Converts to `F[A]` given that `F` is a functor */
   final def run(implicit F: Functor[F]): F[A] = F.map(fi)(k)
@@ -36,10 +40,10 @@ sealed abstract class Coyoneda[F[_], A] extends Serializable { self =>
    * the underlying `F`.
    */
   final def map[B](f: A => B): Aux[F, B, Pivot] =
-    apply(fi)(f compose k)
+    unsafeApply(fi)(f.asInstanceOf[Any => Any] :: ks)
 
   final def transform[G[_]](f: FunctionK[F, G]): Aux[G, A, Pivot] =
-    apply(f(fi))(k)
+    unsafeApply(f(fi))(ks)
 
 }
 
@@ -53,11 +57,17 @@ object Coyoneda {
   /** `F[A]` converts to `Coyoneda[F,A]` for any `F` */
   def lift[F[_], A](fa: F[A]): Coyoneda[F, A] = apply(fa)(identity[A])
 
-  /** Like `lift(fa).map(_k)`. */
+  /** Like `lift(fa).map(k0)`. */
   def apply[F[_], A, B](fa: F[A])(k0: A => B): Aux[F, B, A] =
+    unsafeApply(fa)(k0.asInstanceOf[Any => Any] :: Nil)
+
+  /** Creates a `Coyoneda[F, A]` for any `F`, taking an `F[A]`
+   * and a list of [[Functor.map]]ped functions to apply later
+   */
+  private[cats] def unsafeApply[F[_], A, B](fa: F[A])(ks0: List[Any => Any]): Aux[F, B, A] =
     new Coyoneda[F, B] {
       type Pivot = A
-      val k = k0
+      val ks = ks0
       val fi = fa
     }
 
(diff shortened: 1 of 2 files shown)


free/src/main/scala/cats/free/Coyoneda.scala
```diff
@@ -17,29 +18,32 @@ sealed abstract class Coyoneda[F[_], A] extends Serializable { self =>
   /** The underlying value. */
   val fi: F[Pivot]
 
-  /** The transformer function, to be lifted into `F` by `run`. */
-  val k: Pivot => A
+  /** The transformer functions, to be lifted into `F` by `run`. */
+  val k: List[Any => Any]
```
adelbertc: Given its signature should this be exposed? Seems like we should hide it.

free/src/main/scala/cats/free/Coyoneda.scala
```diff
@@ -53,14 +57,18 @@ object Coyoneda {
   /** `F[A]` converts to `Coyoneda[F,A]` for any `F` */
   def lift[F[_], A](fa: F[A]): Coyoneda[F, A] = apply(fa)(identity[A])
 
-  /** Like `lift(fa).map(_k)`. */
-  def apply[F[_], A, B](fa: F[A])(k0: A => B): Aux[F, B, A] =
+  /** unsafe mang */
```
adelbertc: Maybe a better comment than this ;)

edmundnoble: Omg I was removing it you *fast* person

free/src/main/scala/cats/free/Coyoneda.scala
edmundnoble: I would say so, but the user should have the power to concatenate the lists IMO. Perhaps a scarier name?

edmundnoble: Never mind, that doesn't sound useful. Making it `private[cats]` just in case we change it in future.

free/src/main/scala/cats/free/Coyoneda.scala
```diff
@@ -53,14 +57,20 @@ object Coyoneda {
   /** `F[A]` converts to `Coyoneda[F,A]` for any `F` */
   def lift[F[_], A](fa: F[A]): Coyoneda[F, A] = apply(fa)(identity[A])
 
-  /** Like `lift(fa).map(_k)`. */
-  def apply[F[_], A, B](fa: F[A])(k0: A => B): Aux[F, B, A] =
+  /** Creates a `Coyoneda[F, A]` for any `F`, taking an `F[A]`
+   * and a list of [[Functor.map]]ped functions to apply later
+   */
+  def unsafeApply[F[_], A, B](fa: F[A])(k0: List[Any => Any]): Aux[F, B, A] =
```
wedens: I don't think it (or anything with `Any` in it) should be public.

edmundnoble: My bad, I was on my way to making this private. Not because of Any, but because it fixes List.

free/src/main/scala/cats/free/Coyoneda.scala
```diff
@@ -17,29 +18,32 @@ sealed abstract class Coyoneda[F[_], A] extends Serializable { self =>
   /** The underlying value. */
   val fi: F[Pivot]
 
-  /** The transformer function, to be lifted into `F` by `run`. */
-  val k: Pivot => A
+  /** The transformer functions, to be lifted into `F` by `run`. */
+  private[cats] val k: List[Any => Any]
 
-  import Coyoneda.{Aux, apply}
+  def execute(a: Pivot): A =
```
kailuowang: Does this one need to be public?

johnynek:
I wonder if we want to just add code like:
```scala
sealed trait StackSafeFn[A, B] {
  def apply(a: A): B = StackSafeFn.run(this, a)
}
object StackSafeFn {
  private case class FromScala[A, B](fn: A => B) extends StackSafeFn[A, B]
  private case class Compose[A, B, C](first: StackSafeFn[A, B], second: StackSafeFn[B, C]) extends StackSafeFn[A, C]

  private def run[A, B](fn: StackSafeFn[A, B], a: A): B = ...
}

```

free/src/main/scala/cats/free/Coyoneda.scala
edmundnoble: Good question: it should in my opinion be public because a) it's as safe as exposing `k` previously and b) `k` was public before, which means users have had time to create use-cases that require using the function inside a `Coyo` on other values.

edmundnoble:
@johnynek What we need is `Steque`, because it's exactly that data structure (but you need `Empty` as well, and it also deals with things other than `A => B`). The only issue in your suggestion is that you're missing a notion of "stack-safe value"; a value accompanied by functions which *will* be applied to it later. Adding that in makes it equivalent to `Free[Id, ?]` if we implement `Free` using `Steque`.

free/src/main/scala/cats/free/Coyoneda.scala
```diff
@@ -17,29 +18,32 @@ sealed abstract class Coyoneda[F[_], A] extends Serializable { self =>
   /** The underlying value. */
   val fi: F[Pivot]
 
-  /** The transformer function, to be lifted into `F` by `run`. */
-  val k: Pivot => A
+  /** The transformer functions, to be lifted into `F` by `run`. */
+  private[cats] val k: List[Any => Any]
```
johnynek: what if we called this `ks` and keep `k: Pivot => A = execute _` so we don't break binary compatibility here?

free/src/main/scala/cats/free/Coyoneda.scala
```diff
@@ -17,29 +18,32 @@ sealed abstract class Coyoneda[F[_], A] extends Serializable { self =>
...
 
   /** Converts to `F[A]` given that `F` is a functor */
-  final def run(implicit F: Functor[F]): F[A] = F.map(fi)(k)
+  final def run(implicit F: Functor[F]): F[A] = F.map(fi)(execute)
 
   /** Converts to `Yoneda[F,A]` given that `F` is a functor */
   final def toYoneda(implicit F: Functor[F]): Yoneda[F, A] =
     new Yoneda[F, A] {
-      def apply[B](f: A => B): F[B] = F.map(fi)(k andThen f)
+      def apply[B](f: A => B): F[B] = F.map(fi)(execute _ andThen f)
```
johnynek: if we keep `k` we can not change this line.

free/src/main/scala/cats/free/Coyoneda.scala
edmundnoble: :+1: Done, got it down as `final def k` and added a doc.

johnynek:
:+1:

kailuowang:
merging with 3 sign-offs

PR #1606, opened 2017-04-17
Use Cogen for Kleisli and Func Arbitrary instances

### LINKED ISSUE #1605: Use Cogen for function-like type Arbitrary instances

ceedubs:
I think that this resolves #1605. It may have merge conflicts with #1584
and I think that one should probably be merged first.

I noticed a slight discrepancy between how I've implemented this (taking
implicit `Cogen` and `Arbitrary` instances) vs how the current `StateT`
arbitrary instance works (it takes an implicit `Arbitrary[S => F[(S, A)]]]`). I'm not really sure whether one approach is better than the other. We should probably align on one of the two, though.


ceedubs:
Thanks for the feedback @edmundnoble. I'm going to go ahead and close out this PR and open a new one that uses the approach that you suggested and that doesn't have merge conflicts.

PR #1609, opened 2017-04-18
Include Id docs in the menu

ceedubs:
I've also moved the Id doc from the typeclasses dir to the datatypes
dir, which seems like a better fit to me.

This was inspired by [a prompt](https://gitter.im/typelevel/cats?at=58f5de53ad849bcf4295d321) in the Cats Gitter channel from @BenFradet. @BenFradet could you please review?


BenFradet:
shouldn't the section be changed to `datatypes` though?

ceedubs:
@BenFradet good catch. I was having some trouble trying to build the site locally to test this :(. Change incoming.

kailuowang:
LGTM. thanks

PR #1611, opened 2017-04-18
Add Traverse1 typeclass (#1577)

### LINKED ISSUE #1577: Add Traversable1 Typeclass.
Add a Traversable1 typeclass for `OneAnd` types , that would operate on `Semigroup` and `Apply` instead of `Monoid` and `Applicative` respectively.

It would extend both `Reducible` and `Traverse`. I have a PR coming up soon as discussed on the Gitter channel :)

LukaJCB:
Resolves #1577. Looking for feedback on this one, not sure if there's an easier way to implement it, or if there's general things to improve.


core/src/main/scala/cats/Traverse1.scala
```diff
@@ -0,0 +1,133 @@
...
+
+import cats.data.NonEmptyList
+import simulacrum.typeclass
+
+@typeclass trait Traverse1[F[_]] extends Traverse[F] with Reducible[F] {
+
+  def traverse1[G[_]: Apply, A, B](fa: F[A])(f: A => G[B]): G[F[B]]
+
+  def sequence1[G[_]: Apply, A](fga: F[G[A]]): G[F[A]] =
+    traverse1(fga)(identity)
```
edmundnoble: This method is untested.

core/src/main/scala/cats/Traverse1.scala
```diff
@@ -0,0 +1,133 @@
...
+
+  def flatTraverse1[G[_], A, B](fa: F[A])(f: A => G[F[B]])(implicit G: Apply[G], F: FlatMap[F]): G[F[B]] =
+    G.map(traverse1(fa)(f))(F.flatten)
```
edmundnoble: Ditto for this, and `flatSequence`, `sequence`, `traverse1U`, `sequence1U` in the same class.

core/src/main/scala/cats/Traverse1.scala
```diff
@@ -0,0 +1,133 @@
...
+
+  override def sequence[G[_] : Applicative, A](fga: F[G[A]]): G[F[A]] =
+    traverse1(fga)(identity)
+
+
+
+  def traverse1U[A, GB](fa: F[A])(f: A => GB)(implicit G: Unapply[Apply, GB]): G.M[F[G.A]] =
+    traverse1(fa)(G.subst.compose(f))(G.TC)
+
+  def sequenceU1[GA](fga: F[GA])(implicit U: Unapply[Apply, GA]): U.M[F[U.A]] =
```
edmundnoble: Should this perhaps be `sequence1U`?

core/src/main/scala/cats/Traverse1.scala
```diff
@@ -0,0 +1,133 @@
...
+}
+
+/**
+  * This class defines a `Reducible[F]` in terms of a `Foldable[G]`
+  * together with a `split` method, `F[A]` => `(A, G[A])`.
+  *
+  * This class can be used on any type where the first value (`A`) and
+  * the "rest" of the values (`G[A]`) can be easily found.
+  */
+abstract class NonEmptyTraverse1[F[_], G[_]](implicit G: Foldable[G] with Traverse[G]) extends Traverse1[F] {
```
edmundnoble: Tests?

core/src/main/scala/cats/data/OneAnd.scala
```diff
@@ -202,10 +206,22 @@ private[data] trait OneAndLowPriority1 extends OneAndLowPriority0 {
 }
 
 private[data] trait OneAndLowPriority2 extends OneAndLowPriority1 {
-  implicit def catsDataTraverseForOneAnd[F[_]](implicit F: Traverse[F]): Traverse[OneAnd[F, ?]] =
-    new Traverse[OneAnd[F, ?]] {
-      def traverse[G[_], A, B](fa: OneAnd[F, A])(f: (A) => G[B])(implicit G: Applicative[G]): G[OneAnd[F, B]] = {
-        G.map2Eval(f(fa.head), Always(F.traverse(fa.tail)(f)))(OneAnd(_, _)).value
+  implicit def catsDataTraverseForOneAnd[F[_]](implicit F: Traverse[F], F2: MonadCombine[F]): Traverse1[OneAnd[F, ?]] =
```
edmundnoble: I'm curious; why doesn't this use `NonEmptyTraverse1`?

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -323,9 +325,9 @@ object NonEmptyList extends NonEmptyListInstances {
 private[data] sealed trait NonEmptyListInstances extends NonEmptyListInstances0 {
 
   implicit val catsDataInstancesForNonEmptyList: SemigroupK[NonEmptyList] with Reducible[NonEmptyList]
-      with Comonad[NonEmptyList] with Traverse[NonEmptyList] with Monad[NonEmptyList] =
+      with Comonad[NonEmptyList] with Traverse1[NonEmptyList] with Monad[NonEmptyList] =
     new NonEmptyReducible[NonEmptyList, List] with SemigroupK[NonEmptyList] with Comonad[NonEmptyList]
-      with Traverse[NonEmptyList] with Monad[NonEmptyList] {
+      with Traverse1[NonEmptyList] with Monad[NonEmptyList] {
```
edmundnoble: This one as well, why not NonEmptyTraverse1 it up?

ceedubs:
Thanks @LukaJCB! This is good stuff.

If you would like to move forward with this, could you please address Edmund's review comments? Also I think that you should be able to implement a `compose` similar to the one defined for `Traverse`.

LukaJCB:
Yeah, I'm sorry I've been so inactive, have been quite busy the past few weeks, but I'll get to this in the next couple of days! :)

core/src/main/scala/cats/Traverse1.scala
```diff
@@ -0,0 +1,133 @@
+package cats
+
+import cats.data.NonEmptyList
+import simulacrum.typeclass
+
+@typeclass trait Traverse1[F[_]] extends Traverse[F] with Reducible[F] {
```
kailuowang: Can we also add some brief scaladocs there?

core/src/main/scala/cats/Traverse1.scala
```diff
@@ -0,0 +1,133 @@
...
+
+  override def traverse[G[_] : Applicative, A, B](fa: F[A])(f: (A) => G[B]): G[F[B]] =
+    traverse1(fa)(f)
+
+  override def sequence[G[_] : Applicative, A](fga: F[G[A]]): G[F[A]] =
+    traverse1(fga)(identity)
+
+
+
+  def traverse1U[A, GB](fa: F[A])(f: A => GB)(implicit G: Unapply[Apply, GB]): G.M[F[G.A]] =
```
kailuowang: all unapplied versions are no longer needed

core/src/main/scala/cats/syntax/traverse1.scala
```diff
@@ -0,0 +1,19 @@
+package cats
+package syntax
+
+private[syntax] trait Traverse1Syntax1 {
+  implicit def catsSyntaxUTraverse1[FA](fa: FA)(implicit U: Unapply[Traverse1, FA]): Traverse1.Ops[U.M, U.A] =
```
kailuowang: We don't need this unapply version anymore.

core/src/main/scala/cats/syntax/reducible.scala
```diff
@@ -1,11 +1,12 @@
 package cats
 package syntax
 
-trait ReducibleSyntax extends Reducible.ToReducibleOps {
-  implicit final def catsSyntaxNestedReducible[F[_]: Reducible, G[_], A](fga: F[G[A]]): NestedReducibleOps[F, G, A] =
-    new NestedReducibleOps[F, G, A](fga)
+private[syntax] trait ReducibleSyntax1 {
+  implicit def catsSyntaxUReducible[FA](fa: FA)(implicit U: Unapply[Reducible, FA]): Reducible.Ops[U.M, U.A] =
+    new Reducible.Ops[U.M, U.A] {
```
kailuowang: we can use a doctest in the type class to cover this.

laws/src/main/scala/cats/laws/Traverse1Laws.scala
```diff
@@ -0,0 +1,73 @@
+package cats.laws
+
+
+import cats.{Apply, Id, Semigroup, Traverse1}
+import cats.data.{Const, Nested}
+import cats.syntax.traverse1._
+import cats.syntax.reducible._
+
+trait Traverse1Laws[F[_]] extends TraverseLaws[F] with ReducibleLaws[F] {
```
kailuowang: None of the laws and test were actually hit. We need to add one law test in a data type that has a vanilla `Traverse1` instance and one for a data type that uses the `NonEmptyTraverse1`.

core/src/main/scala/cats/data/OneAnd.scala
LukaJCB: Hey, I'm not sure what exactly you mean by this comment, if you could clarify, that'd be great :)

ceedubs: @LukaJCB I think that @edmundnoble is asking why the `Traverse1` instance for `OneAnd` isn't using the `NonEmptyTraverse1` abstract class that you've submitted with this PR.

core/src/main/scala/cats/Reducible.scala
```diff
@@ -187,3 +187,93 @@ import simulacrum.typeclass
...
 }
+
+/**
+ * This class defines a `Reducible[F]` in terms of a `Foldable[G]`
+ * together with a `split` method, `F[A]` => `(A, G[A])`.
+ *
+ * This class can be used on any type where the first value (`A`) and
+ * the "rest" of the values (`G[A]`) can be easily found.
+ */
+abstract class NonEmptyReducible[F[_], G[_]](implicit G: Foldable[G]) extends Reducible[F] {
```
LukaJCB: Unfortunately adding `val` didn't seem to fix the issue, I'll have to look into it further 🤔

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -352,7 +354,12 @@ private[data] sealed trait NonEmptyListInstances extends NonEmptyListInstances0
 
       def extract[A](fa: NonEmptyList[A]): A = fa.head
 
-      def traverse[G[_], A, B](fa: NonEmptyList[A])(f: A => G[B])(implicit G: Applicative[G]): G[NonEmptyList[B]] =
+      def traverse1[G[_] : Apply, A, B](as: NonEmptyList[A])(f: A => G[B]): G[NonEmptyList[B]] = {
+        as.map(a => Apply[G].map(f(a))(NonEmptyList.of(_)))
+          .reduceLeft((acc, a) => (acc |@| a).map( _ |+| _ ))
+      }
```
peterneyens: This is not really the best implementation of `traverse1`.

A better implementation : 

```scala
def traverse1[G[_], A, B](nel: NonEmptyList[A])(f: A => G[B])(implicit G: Apply[G]): G[NonEmptyList[B]] =
  Foldable[List].reduceRightToOption[A, G[List[B]]](nel.tail)(a => G.map(f(a))(_ :: Nil)) { (a, lglb) =>
  }.map { 
    case None => G.map(f(nel.head))(NonEmptyList(_, Nil))
    case Some(gtail) => G.map2(f(nel.head), gtail)(NonEmptyList(_, _))
  }.value
```

`Traverse[Vector].traverse` also uses prepending (`+:`), so I think this should work for `NonEmptyVector` as well. For `OneAnd` we need to come up with something else if we want to improve that one as well.

A [benchmark](https://gist.github.com/peterneyens/befa719c6b9c9ee92434106910734619) comparing your `traverse1` to mine above and a copy of yours using `NonEmptyList(_, Nil)` instead of `NonEmptyList.of`.

```
[info] Benchmark                          Mode  Cnt      Score     Error  Units
[info] NelTraverseBench._traverse1noOf   thrpt   10  10977.183 ± 240.703  ops/s
[info] NelTraverseBench._traverse1prime  thrpt   10  89787.940 ± 849.006  ops/s
[info] NelTraverseBench.traverse1        thrpt   10   3929.498 ±  72.846  ops/s
```
The difference of just replacing `NonEmptList.of` is already quite big, so I am plugging #1707 here again.

peterneyens:
Thanks for the quick changes @LukaJCB!
I think you may need to fix the conflicts in the _typeclasses.md_ file before travis runs again.

core/src/main/scala/cats/data/OneAnd.scala
```diff
@@ -205,19 +209,16 @@ private[data] trait OneAndLowPriority1 extends OneAndLowPriority0 {
 }
 
 private[data] trait OneAndLowPriority2 extends OneAndLowPriority1 {
-  implicit def catsDataTraverseForOneAnd[F[_]](implicit F: Traverse[F]): Traverse[OneAnd[F, ?]] =
-    new Traverse[OneAnd[F, ?]] {
-      def traverse[G[_], A, B](fa: OneAnd[F, A])(f: (A) => G[B])(implicit G: Applicative[G]): G[OneAnd[F, B]] = {
-        G.map2Eval(f(fa.head), Always(F.traverse(fa.tail)(f)))(OneAnd(_, _)).value
-      }
+  implicit def catsDataTraverse1ForOneAnd[F[_]](implicit F: Traverse[F], F2: MonadCombine[F]): Traverse1[OneAnd[F, ?]] =
```
peterneyens: Can we override `traverse` using the implementation originally used in `Traverse`?
We probably want to override the `traverse` for `NonEmptyVector` as well using the previous implementation, come to think of it.

This implementation using `MonadCombine` also means that you cannot get a `Traverse[OneAnd[F, ?]]` instance anymore if `F` has a `Traverse` but no `MonadCombine`.

LukaJCB: Great idea!

tests/src/test/scala/cats/tests/OneAndTests.scala
```diff
@@ -16,6 +16,7 @@ class OneAndTests extends CatsSuite {
   checkAll("OneAnd[Stream, Int]", OrderLaws[OneAnd[Stream, Int]].eqv)
 
   checkAll("OneAnd[Stream, Int] with Option", TraverseTests[OneAnd[Stream, ?]].traverse[Int, Int, Int, Int, Option, Option])
+  checkAll("OneAnd[Stream, Int] with Option", Traverse1Tests[OneAnd[Stream, ?]].traverse1[Option, Int, Int, Int, Int, Option, Option])
```
peterneyens: The old `Traverse[OneAnd[F, ?]]` instance is not tested anymore.

Can you add something like

```scala
{
  implicit val traverse = ListWrapper.traverse
  checkAll("OneAnd[ListWrapper, Int] with Option", TraverseTests[OneAnd[LisTWrapper, ?]].traverse[Int, Int, Int, Int, Option, Option])
  checkAll("Traverse[OneAnd[ListWrapper, A]]", SerializableTests.serializable(Traverse[OneAnd[ListWrapper, ?]]))
}
```

Can you also add `SerializableTests` for `Traverse1` here (`OneAnd`) and for `NonEmptyList` and `NonEmptyVector`?

You can also remove the tests for `Traverse` (line 18 and 20 here) and for `NonEmptyList` and `NonEmptyVector` as those tests are covered by the `Traverse1Tests` you wrote as well.

laws/src/main/scala/cats/laws/discipline/Traverse1Tests.scala
```diff
@@ -0,0 +1,59 @@
...
+import Prop.forAll
+import cats.{Applicative, Eq, Monoid, Traverse1}
+import cats.laws.Traverse1Laws
+
+
+trait Traverse1Tests[F[_]] extends TraverseTests[F] with ReducibleTests[F] {
+  def laws: Traverse1Laws[F]
+
+  def traverse1[G[_]: Applicative, A: Arbitrary, B: Arbitrary, C: Arbitrary, M: Arbitrary, X[_]: Applicative, Y[_]: Applicative](implicit
+                                                                                                             ArbFA: Arbitrary[F[A]],
```
peterneyens: I know indentation preferences are a bit personal (and we don't really have style guidelines), but this seems a bit excessive :smiley: 

If you base yourself `TraverseTests` and `TraverseLaws` as guidelines for indentation, `Traverse1Tests` and `Traverse1Laws` will be consistent with the rest.

core/src/main/scala/cats/data/OneAnd.scala
```diff
@@ -221,4 +225,18 @@ private[data] trait OneAndLowPriority2 extends OneAndLowPriority1 {
...
+private[data] trait OneAndLowPriority3 extends OneAndLowPriority2 {
+  implicit def catsDataTraverse1ForOneAnd[F[_]](implicit F: Traverse[F], F2: MonadCombine[F]): Traverse1[OneAnd[F, ?]] =
+    new NonEmptyReducible[OneAnd[F, ?], F] with Traverse1[OneAnd[F, ?]] {
+      def traverse1[G[_], A, B](fa: OneAnd[F, A])(f: (A) => G[B])(implicit G: Apply[G]): G[OneAnd[F, B]] = {
+        import cats.syntax.cartesian._
+
+        fa.map(a => Apply[G].map(f(a))(OneAnd(_, F2.empty[B])))(F)
+          .reduceLeft(((acc, a) => (acc |@| a).map((x: OneAnd[F, B], y: OneAnd[F, B]) => x.combine(y))))
+      }
+
```
peterneyens: Can you add the implementation of `traverse` used in the `Traverse` instance above here as well?

peterneyens:
Sorry to ask you one more thing, but can you add a `Traverse1#compose` method (to create a `Traverse1[λ[α => F[G[α]]]]` instance), like @ceedubs asked [here](https://github.com/typelevel/cats/pull/1611#issuecomment-301268178)?

See:
 - [`Traverse#compose`](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Traverse.scala#L86-L90)
 - [`ComposedTraverse`](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Composed.scala#L66-L72)
 - `Traverse` instance for [`Nested`](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/Nested.scala#L42-L45)
 - Testing `compose` and `Nested` instance in [`NestedTests`](https://github.com/typelevel/cats/blob/master/tests/src/test/scala/cats/tests/NestedTests.scala#L113-L118)

We could also merge as is and do this in follow up PR.

LukaJCB:
Don't be! I'm on mobile right now, but I'll do it in the morning tomorrow :)

core/src/main/scala/cats/data/Nested.scala
```diff
@@ -133,13 +133,20 @@ private[data] sealed abstract class NestedInstances9 extends NestedInstances10 {
...
 
-private[data] sealed abstract class NestedInstances10 {
+private[data] sealed abstract class NestedInstances10 extends NestedInstances11 {
   implicit def catsDataInvariantForNestedContravariant[F[_]: Invariant, G[_]: Contravariant]: Invariant[Nested[F, G, ?]] =
     new NestedInvariant[F, G] {
       val FG: Invariant[λ[α => F[G[α]]]] = Invariant[F].composeContravariant[G]
     }
 }
 
+private[data] sealed abstract class NestedInstances11 {
```
LukaJCB: Not sure if making a new class here was the best thing to do, feedback very welcome :)

peterneyens: I think you can put the `catsDataTraverse1ForNested` inside `NestedInstances`.

As you can never have a `TraverseFilter` and a `Traverse1` for a data type at the same time, I don't think that the composition `Traverse` and `TraverseFilter` versus the composition of `Traverse1` and `Traverse1` can be ambiguous.

peterneyens:
Seems that the `Reducible[Nested[NonEmptyList, NonEmptyVector, ?]]` instance isn't tested any more after we moved the `Traverse1` instance.

I think replacing the `ReducibleTests[Nested[NonEmptyList, NonEmptyVector, ?]]` by 

```scala
{
  implicit val foldable = ListWrapper.foldable
  checkAll("Nested[NonEmptyList, OneAnd[ListWrapper, ?], ?]", ReducibleTests[Nested[NonEmptyList, OneAnd[ListWrapper, ?], ?]].reducible[Option, Int, Int])
  checkAll("Reducible[Nested[NonEmptyList, OneAnd[ListWrapper, ?], ?]]", SerializableTests.serializable(Reducible[Nested[NonEmptyList, OneAnd[ListWrapper, ?], ?]]))
}
```

should fix this.

kailuowang:
I agree that `Traverse1` probably is not a good enough name. I can live with `NonEmptyTraverse` as suggested by @LukaJCB [here](https://github.com/typelevel/cats/issues/1297#issuecomment-304613675)

peterneyens:
What do we do in that case with the methodsuffix `1`?
Appending `NonEmpty` becomes a bit unwieldy (`traverseNonEmpty`, `flatSequenceNonEmpty`, ...).

LukaJCB:
I think I would prefer `traverseNonEmpty`, but yeah, your point still holds, `nonEmptyFlatSequence` is quite a bit. Can't think of anything better right now, but I'll try to brainstorm later.

kailuowang:
I vote for going with `NonEmpty`, it's not the best, but better than 1

LukaJCB:
We currently also have `traverse1_` and `sequence1_` in reducible. I suggest renaming them both and since we'd be breaking compatibility anyway, I also suggest we find better names for these two. I realize the underscore comes from the Haskell equivalent, but I do not believe it conveys its intent. 

I suggest something along the lines of `traverseVoid` or `traverseToUnit`.

andyscott:
I like `NonEmpty` as a prefix. It's consistent with the naming for `NonEmptyList`. Additionally, I prefer `nonEmpty` as a prefix for the associated methods.

peterneyens:
I don't think we will be able to come up with anything better than `NonEmptyTraverse` and `nonEmptyTraverse` / `nonEmptySequence`, so :+1:.

I think it is best to leave the `_` suffix in this PR and create a new issue for a potential rename.

laws/src/main/scala/cats/laws/discipline/NonEmptyTraverseTests.scala
```diff
@@ -0,0 +1,59 @@
...
+        EqXFB.eqv(x._1, y._1) && EqYFB.eqv(x._2, y._2)
+    }
+    new RuleSet {
+      def name: String = "nonEmptyTraverse"
+      def bases: Seq[(String, RuleSet)] = Nil
+      def parents: Seq[RuleSet] = Seq(traverse[A, B, C, M, X, Y], reducible[G, A, B])
+      def props: Seq[(String, Prop)] = Seq(
+        "nonEmptyTraverse identity" -> forAll(laws.nonEmptyTraverseIdentity[A, C] _),
+        "nonEmptyTraverse sequential composition" -> forAll(laws.traverseSequentialComposition[A, B, C, X, Y] _),
+        "nonEmptyTraverse parallel composition" -> forAll(laws.traverseParallelComposition[A, B, X, Y] _),
```
kailuowang: should we add tests for  `laws.nonEmptyTraverseParallelComposition` and `laws.nonEmptyTraverseSequentialComposition` ?

LukaJCB: Yeah, not sure what happened there 😅

LukaJCB:
There's currently one last thing, in `Reducible` we have the `intercalate1` method. As with `traverse1_` and `sequence1_`, I'm going to rename this to `nonEmptyIntercalate` if no one objects.

PR #1612, opened 2017-04-18
Add ensureWith to Validated and Either (#1550)

### LINKED ISSUE #1550: Add Validated.ensure[EE >: E](onFailure: A => EE)(f: A => Boolean): Validated[EE, A]
Add a `cats.data.Validated.ensure(...)` with this signature: 

`ensure[EE >: E](onFailure: A => EE)(f: A => Boolean): Validated[EE, A]`. 

To be able to extract informations from `A` to construct the `EE` in the failure case with the `onFailure` function.

```scala
val personValidated = ???

personValidated.ensure(p => s"Too young (age: ${p.age} < 18)")(_.age >= 18)
```

LukaJCB:
Resolves #1550. Also looking for feedback on naming here, or maybe I should be adding an additional test apart from the doc test. Would be much appreciated :)


kailuowang:
I think additional tests would be nice for these logic.

ceedubs:
It looks like ScalaDoc isn't happy about the `$s` in your message, because it's trying to interpret it as a ScalaDoc variable. You can change it to `\$s` or you can use traditional string concatenation instead of string interpolation to work around it.

LukaJCB:
Thank you @ceedubs! CI seems to be happy now :)

tests/src/test/scala/cats/tests/ValidatedTests.scala
```diff
@@ -248,4 +248,20 @@ class ValidatedTests extends CatsSuite {
...
+
+  test("ensureWith on Invalid is identity") {
+    forAll { (x: Validated[Int,String], f: String => Int, p: String => Boolean) =>
+      if (x.isInvalid) {
+        x.ensureWith(f)(p) should === (x)
+      }
+    }
+  }
+
+  test("ensureWith should fail if predicate not satisfied") {
```
peterneyens: Should we add this test to `EitherTests` as well?

LukaJCB: Done!

ceedubs:
Thanks @LukaJCB! This LGTM. The only thing that I'm not sure about is the name `ensureWith`. We have `recoverWith` defined in `ApplicativeError`, and the `With` there means something a little different. Having said that, I don't have a better name in mind and ultimately I'm okay with `ensureWith` if nobody has a better proposal.

It doesn't necessarily need to be done in this PR, but I think that `ensureWith` could also be added as a derived method alongside `ensure` in the `MonadError` type class.

LukaJCB:
Thanks for your feedback! :)
+1 for keeping it open to see if there's a better name! I can create another PR for the same method in `MonadError` soon! :)

kailuowang:
Given that we are close to 1.0.0-MF, I kind of prefer having the addition to `MonadError` and these in a single PR. I noticed that the `MonadError` instance for `EitherT` did not override the `ensure` method. So there are probably some optimizations can be done. 
As for the name, I am probably the last person to give a better proposal but how about `ensureOr`? It's not more intuitive than `ensureWith` but at least it does not conflict with `recoverWith`.

kailuowang:
This is going to be really useful. I constantly need this feature. 
 I would like to make the following suggestion. 
1. add the two new laws to `MonadErrorLaws` something like 

```scala
def monadErrorEnsureConsistency[A](fa: F[A], e: E, p: A => Boolean): IsEq[F[A]] =
    F.ensure(fa)(e)(p) <-> F.flatMap(fa)(a => if (p(a)) F.pure(a) else F.raiseError(e))

def monadErrorEnsureOrConsistency[A](fa: F[A], e: A => E, p: A => Boolean): IsEq[F[A]] =
    F.ensure(fa)(e)(p) <-> F.flatMap(fa)(a => if (p(a)) F.pure(a) else F.raiseError(e(a)))
```

2. override the `ensureOr` method in the `MonadError` instances for `Either` and `EitherT` , override the `ensure` method in the `MonadError` instance of `EitherT`, just like [here](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/instances/either.scala#L93)

3. remove the `ensure` and `ensureOr` tests in `EitherTests` and `EitherTTests`, since they will be covered by the law tests.

LukaJCB:
So implemented the changes, but removing the tests seems to have reduced the coverage, will investigate soon :D

kailuowang:
The newly added laws need to go into the tests in the discipline package.

LukaJCB:
Oops, yeah, that should've been obvious to me, fixed!

core/src/main/scala/cats/syntax/either.scala
```diff
@@ -63,6 +63,11 @@ final class EitherOps[A, B](val eab: Either[A, B]) extends AnyVal {
     case Right(b) => if (f(b)) eab else Left(onFailure)
   }
 
+  def ensureOr[AA >: A](onFailure: B => AA)(f: B => Boolean): Either[AA, B] = eab match {
```
ceedubs: It looks like this is following a precedent, but I find myself wondering if this `Either` syntax should really be providing syntax for methods that should already be available via type class syntax (in this case `MonadError`). Let's table that for a separate issue though.

kailuowang:
thanks so much @LukaJCB , sorry about the delay. LGTM. 👍 
@peterneyens any other thoughts/feedback?

peterneyens:
There seem to be some merge conflicts, otherwise this looks OK :+1: , thanks @LukaJCB !

I wonder if we maybe should implement `ensure` in terms of `ensureOr` that way just overriding `ensureOr` gives you a more efficient `ensure` as well.

kailuowang:
I took the liberty of resolving the conflict using github's merging too. Will merge if build green.

PR #1614, opened 2017-04-18
added leftT and improved existing lift API for EitherT

kailuowang:
using PartiallyApplied to improve the API
added a `leftT` and `rightT` alias

### DIFF
--- a/core/src/main/scala/cats/data/EitherT.scala
+++ b/core/src/main/scala/cats/data/EitherT.scala
@@ -89,7 +89,7 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
   def map[D](f: B => D)(implicit F: Functor[F]): EitherT[F, A, D] = bimap(identity, f)
 
   def semiflatMap[D](f: B => F[D])(implicit F: Monad[F]): EitherT[F, A, D] =
-    flatMap(b => EitherT.right[F, A, D](f(b)))
+    flatMap(b => EitherT.right(f(b)))
 
   def leftMap[C](f: A => C)(implicit F: Functor[F]): EitherT[F, C, B] = bimap(f, identity)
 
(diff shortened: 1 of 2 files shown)


diego-alvarez-hs:
Thank you for this @kailuowang

core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -236,11 +236,78 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
 object EitherT extends EitherTInstances with EitherTFunctions
 
 private[data] trait EitherTFunctions {
-  final def left[F[_], A, B](fa: F[A])(implicit F: Functor[F]): EitherT[F, A, B] = EitherT(F.map(fa)(Either.left))
 
-  final def right[F[_], A, B](fb: F[B])(implicit F: Functor[F]): EitherT[F, A, B] = EitherT(F.map(fb)(Either.right))
+  final class LeftPartiallyApplied[B] private[EitherTFunctions] {
```
edmundnoble: I am curious. Is Scalaz's trick here (add a dummy parameter to this class to make it an `AnyVal`; see [here](https://github.com/scalaz/scalaz/blob/926dd8c380bab865afed8ac1a2d9da2bde488a00/core/src/main/scala/scalaz/EitherT.scala)) necessary to make this zero-cost?

kailuowang: That seems a nice trick. Do we want to use it here first in this PR, or create another PR applying the pattern globally?

edmundnoble: I'm okay with another PR to apply it globally.

core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -362,6 +362,19 @@ private[data] trait EitherTFunctions {
...
+   * {{{
+   * scala> import cats.implicits._
+   * scala> val o: Option[Int] = None
+   * scala> EitherT.fromOptionF(List(o), "Answer not known.")
+   * res0: EitherT[List, String, Int]  = EitherT(List(Left(Answer not known.)))
+   * scala> EitherT.fromOptionF(List(Option(42)), "Answer not known.")
+   * res1: EitherT[List, String, Int] = EitherT(List(Right(42)))
+   * }}}
+   */
+  final def fromOptionF[F[_], E, A](fopt: F[Option[A]], ifNone: => E)(implicit F: Functor[F]): EitherT[F, E, A] =
```
peterneyens: Just wondering, is this better than `OptionT(fopt).toRight(ifNone)` ?

kailuowang: Didn't know about the `OptionT.toRight`, Looks like this one has slightly less function calls and an allocation to `OptionT`?

kailuowang: @peterneyens anything else in this PR?

PR #1615, opened 2017-04-18
Fix doc for InvariantMonoidal

BenFradet:
InvariantMonoidal was not generated because of a typo and not present on the left-side menu, this PR solves both issues.

Related to #1609.


kailuowang:
👍

PR #1618, opened 2017-04-19
Override some methods in Kleisli instances

peterneyens:
- Switch the order of the traits so that some methods are overridden by the implementation in other type class instances in its hierarchy

  For example:
  - `extends Monad[Kleisli[F, A, ?]] with KleisliApplicative[F, A]`
  - instead of `extends KleisliApplicative[F, A] with Monad[Kleisli[F, A, ?]]`)

  We were already overriding `ap` in `Monad`, this also overrides `product` (which was the initial reason for this PR).
  For the rest we can just loose a couple of explicit overrides (eg `super[KleisliSplit].split(f, g)`)

- Add a `MonadReader` test for `Reader`, which improves some code coverage.

### DIFF
--- a/core/src/main/scala/cats/data/Kleisli.scala
+++ b/core/src/main/scala/cats/data/Kleisli.scala
@@ -82,10 +82,10 @@ private[data] sealed trait KleisliFunctions {
 
 private[data] sealed abstract class KleisliInstances extends KleisliInstances0 {
 
-  implicit def catsDataMonoidForKleisli[F[_], A, B](implicit M: Monoid[F[B]]): Monoid[Kleisli[F, A, B]] =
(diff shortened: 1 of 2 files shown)


ceedubs:
Thanks @peterneyens!

This generally looks really good to me. One minor question: was there any particular motivation for changing a bunch of implicit parameter names to something like `ev`? I was thinking that in most other places in cats we were using implicit parameter names like `F`, `M`, `AE` etc, but admittedly I haven't looked around much to see if that's true.

peterneyens:
Most of the parameters here in `Kleisli` were called `ev`, but it is probably best to be consistent in general, so I switched them around.

PR #1619, opened 2017-04-21
Allow non-constant arbitrary functions for (Co)Kleisli, State, Func

### LINKED ISSUE #1605: Use Cogen for function-like type Arbitrary instances

ceedubs:
Resolves #1605.

This is a replacement for #1606.

Now that `Arbitrary` instances for functions have been improved with the introduction of `Cogen` in scalacheck, we can pick up better arbitrary functions than the constant ones we were using.

### DIFF
--- a/laws/src/main/scala/cats/laws/discipline/Arbitrary.scala
+++ b/laws/src/main/scala/cats/laws/discipline/Arbitrary.scala
@@ -77,8 +77,8 @@ object arbitrary extends ArbitraryInstances0 {
       (a, b) => A.perturb(B.perturb(seed, b), a)))
 
 
-  implicit def catsLawsArbitraryForCokleisli[F[_], A, B](implicit B: Arbitrary[B]): Arbitrary[Cokleisli[F, A, B]] =
-    Arbitrary(B.arbitrary.map(b => Cokleisli[F, A, B](_ => b)))
+  implicit def catsLawsArbitraryForCokleisli[F[_], A, B](implicit FAB: Arbitrary[F[A] => B]): Arbitrary[Cokleisli[F, A, B]] =
+    Arbitrary(FAB.arbitrary.map(Cokleisli(_)))
 
   implicit def catsLawsArbitraryForOptionT[F[_], A](implicit F: Arbitrary[F[Option[A]]]): Arbitrary[OptionT[F, A]] =
     Arbitrary(F.arbitrary.map(OptionT.apply))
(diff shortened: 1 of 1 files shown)


ceedubs:
Requesting a review from @non. As you are the one who introduced `Cogen` into Scalacheck, I'm interested in your input on the decision that was made in https://github.com/typelevel/cats/pull/1606#pullrequestreview-33132927 to require implicit arbitrary instances for functions as opposed to requiring implicit `Gen` and `Cogen` instances and deriving the arbitrary functions ourselves.

ceedubs:
Oops there's a compile error in the free module. I'll fix it.

ceedubs:
@edmundnoble hmm actually I guess this compile error is one potential reason to favor the direct `Cogen`/`Arbitrary` implicit parameters as opposed to the function `Arbitrary` instance. In some cases the implicit search for the function `Arbitrary` instance fails while the `Arbitrary[A]` and `Cogen[A]` implicit search succeeds. I also kind of like the idea that conceptually the `Cogen` and `Arbitrary` instances are the fundamental things that we need to provide our instance and the fact that we get there through scalacheck's `Function1` instance is an implementation detail, but I suppose one could argue this in the opposite direction.

Now I'm second-guessing this approach :\

edmundnoble:
Huh. I can't see why that'd be the case, the implicit error that is. TBH I am more inclined toward the argument in the other direction; users can provide their own `Arbitrary` function instances that way, and the `Function1` instance is what I'd see people choosing to use or not use. Having different `Cogen` and output `Arbitrary` instances in scope will already feed into the `Function1` instance generator. Also... is there not an error in the Kleisli monoid in the tests? https://travis-ci.org/typelevel/cats/jobs/224346639#L1658

ceedubs:
@edmundnoble hmm I guess that the implicit divergence only is happening in the scala.js build:

```
[info] Compiling 7 Scala sources to /home/travis/build/typelevel/cats/free/.js/target/scala-2.10/test-classes...
[error] /home/travis/build/typelevel/cats/free/src/test/scala/cats/free/FreeTTests.scala:211: diverging implicit expansion for type org.scalacheck.Arbitrary[Int => cats.Eval[(Int, A)]]
[error] starting with method arbContainer2 in trait ArbitraryLowPriority
[error]   implicit def intStateArb[A: Arbitrary]: Arbitrary[IntState[A]] = catsLawArbitraryForState[Int, A]
[error]                                                                                            ^
[error] one error found
[error] (freeJS/test:compileIncremental) Compilation failed
```

And yeah you are right, it is reporting a problem with the Kleisli Monoid. That's concerning. I'll try to take a look at this this evening (Eastern US time).

ceedubs:
Closing in favor of #1666. @edmundnoble sorry I'm departing from your preferred approach, but I seem to need to in order for this to compile in the scala.js build 😬

PR #1620, opened 2017-04-21
[WIP] adding `FunctorK`

kailuowang:
Submitted to solicit feedbacks.  
The incentive for adding this is so that later we can transform final tagless encoded algebra with `FunctionK`s if we can have `FunctorK` instances automatically derived by kittens. 
E.g.

```Scala 
trait Algebra[F[_]] extends Product with Serializable {
   def parseInt: String => F[Int] 
   def parseFloat: String => F[Float] 
}
case class OptionInterpeter (
   val passInt: String => Option[Int] = Try((_:String).toInt).toOption
   val passFloat: String => Option[Float] = Try((_:String).toFloat).toOption
) extends Algebra[Option] {
```
Assume that Kittens can auto derive a `FunctorK` instance for `Algebra[F[_]]`, then we can do
```Scala
val transformed: Algebra[List] = OptionInterpeter().mapK(λ[Option ~> List](_.toList))
```

### DIFF
--- a/core/src/main/scala/cats/FunctorK.scala
+++ b/core/src/main/scala/cats/FunctorK.scala
@@ -1,0 +1,10 @@
+package cats
+
+import arrow.FunctionK
+
+import simulacrum.typeclass
+
+
+@typeclass trait FunctorK[F[_[_]]] {
+  def mapK[G[_], H[_]](fg: F[G])(f: FunctionK[G, H]): F[H]
+}
--- a/core/src/main/scala/cats/data/Kleisli.scala
+++ b/core/src/main/scala/cats/data/Kleisli.scala
@@ -82,6 +82,10 @@ private[data] sealed trait KleisliFunctions {
 
 private[data] sealed abstract class KleisliInstances extends KleisliInstances0 {
 
+  implicit def catsDataFunctorKForKleisli[A, B]: FunctorK[Kleisli[?[_], A, B]] = new FunctorK[Kleisli[?[_], A, B]]{
+    def mapK[G[_], H[_]](fg: Kleisli[G, A, B])(f: G ~> H): Kleisli[H, A, B] = fg.transform(f)
+  }
+
   implicit def catsDataMonoidForKleisli[F[_], A, B](implicit FB0: Monoid[F[B]]): Monoid[Kleisli[F, A, B]] =
     new KleisliMonoid[F, A, B] { def FB: Monoid[F[B]] = FB0 }
 
--- a/core/src/main/scala/cats/data/WriterT.scala
+++ b/core/src/main/scala/cats/data/WriterT.scala
@@ -66,6 +66,10 @@ object WriterT extends WriterTInstances with WriterTFunctions {
 }
 
 private[data] sealed abstract class WriterTInstances extends WriterTInstances0 {
+  implicit def catsDataFunctorKForWriterT[L, V]: FunctorK[WriterT[?[_], L, V]] = new FunctorK[WriterT[?[_], L, V]]{
+    def mapK[G[_], H[_]](fg: WriterT[G, L, V])(f: G ~> H): WriterT[H, L, V] = WriterT(f(fg.run))
+  }
+
 
   implicit def catsDataMonadForWriterTId[L:Monoid]: Monad[WriterT[Id, L, ?]] =
     catsDataMonadWriterForWriterT[Id, L]
--- a/core/src/main/scala/cats/syntax/all.scala
+++ b/core/src/main/scala/cats/syntax/all.scala
@@ -19,6 +19,7 @@ trait AllSyntax
     with FlatMapSyntax
     with FoldableSyntax
     with FunctorSyntax
+    with FunctorKSyntax
     with FunctorFilterSyntax
     with GroupSyntax
     with InvariantSyntax
--- a/core/src/main/scala/cats/syntax/functorK.scala
+++ b/core/src/main/scala/cats/syntax/functorK.scala
@@ -1,0 +1,4 @@
+package cats
+package syntax
+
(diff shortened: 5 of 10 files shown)


johnynek:
I think this is interesting. Can we find more instances in cats? I guess WriterT also has an instance?

kailuowang:
@johnynek, yup, `WriterT` probably has one. I'll add later.

johnynek:
some more: Free, FreeT, TupleK, EitherK (in fact the last three have a BifunctorK I guess). 

Also, I guess `Reducible[F].contramapK(fn: FunctionK[G, F]): Reducible[G]` is a thing. So, I think that means Reducible and Foldable are probably ContravariantK.

kailuowang:
As @sellout suggested in gitter. There are several existing implementations of a `FunctorK[F[_[_], _]]` targeting type `F[_[_], _]`. 
```scala 
trait FunctorK[F[_[_], _]] {
   def mapK[G[_], H[_], A](fga: F[G, A])(f: G ~> H): F[H, A]
}
```
This is also the definition of `HFunctor` in [Haskell](http://hackage.haskell.org/package/category-extras-0.53.4/docs/Control-Functor-HigherOrder.html)
I think this might be a more natural `FunctorK` type class for the data types in cats such as `Kleisli`, `Free`, `EitherK`. They are close to `F[_[_], _]` in the sense that they all have at least one other type parameter aside the `_[_]` which in some way, determines the usage of the `_[_]` type constructor inside the data type. In another word, the usage of the `_[_]` inside the data type `F` is parametric polymorphic. The target type in this `FunctorK`, however, is `F[_[_]]`, which means that the usage of the `_[_]` type constructor has to be fixed, just as indicated in the example algebra in the PR description. So, when defining the instances of this `FunctorK` for a datatype where the usage of the `_[_]` is parametric, the parametricity is provided through the creation of the instance e.g.
```scala
implicit def catsDataFunctorKForWriterT[L, V]: FunctorK[WriterT[?[_], L, V]] = new FunctorK[WriterT[?[_], L, V]]{
    def mapK[G[_], H[_]](fg: WriterT[G, L, V])(f: G ~> H): WriterT[H, L, V] = WriterT(f(fg.run))
  }
```
This is not as natural as if the `FunctorK`'s target type is a `F[_[_], _]`. 

So if we want a `FunctorK` for the datatypes in cats, we should probably make it target type `F[_[_], _]`. But this typeclass won't be of much use when transforming final tagless interpreters (the example in the PR description). If there is no other immediate need for this `FunctorK[F[_[_], _]`, maybe we should be too eager to add it into cats?
 If transforming the final tagless algebra is the only use case for a `FunctorK[F[_[_]]`, then maybe we don't need a typeclass, just some specialized shapeless or meta magic would probably be more straightforward. 
So I am leaning slightly towards closing this PR for now.

edmundnoble:
I can put the transformer-focused one into `cats-mtl`.

julienrf:
> Assume that Kittens can auto derive a `FunctorK` instance for `Algebra[F[_]]`, then we can do
> 
> ~~~ scala
> val transformed: Algebra[List] = OptionInterpeter().mapK(λ[Option ~> List](_.toList))
> ~~~

What’s the advantage over `λ[Option ~> List](_.toList).apply(OptionInterpreter())`?

kailuowang:
@julienrf I am not sure I follow you. You cannot pass an `OptionInterpreter()` directly into `λ[Option ~> List](_.toList).apply` which it takes a `Option[X]`, right?

julienrf:
Oh, right, I misunderstood.

kailuowang:
Just did a POC for transforming final tagless interpreter with shapeless. It was surprisingly easy (although I used case class rather than trait for encoding algebras)
Here is the example code that can also be found at this [henkan tut doc](https://github.com/kailuowang/henkan/blob/poc-final-tagless/docs/src/main/tut/FinalTagless.md)
```scala
import util.Try
// import util.Try

import shapeless._
// import shapeless._

import shapeless.ops.hlist.Mapper
// import shapeless.ops.hlist.Mapper

import cats.~>
// import cats.$tilde$greater

case class Algebra[F[_]](
  parseInt: String ⇒ F[Int],
  parseFloat: String ⇒ F[Float]
)
// warning: there was one feature warning; for details, enable `:setting -feature' or `:replay -feature'
// defined class Algebra

object Algebra1 {
  type F[T] = Try[T]

  val intepreter = Algebra[F](
    parseInt = (s: String) ⇒ Try(s.toInt),
    parseFloat = (s: String) ⇒ Try(s.toFloat)
  )
}
// defined object Algebra1

object FunctorK {

  object PolyMapK extends Poly1 {
    implicit def caseF[A, F[_], G[_], B](implicit n: F ~> G) = at[A ⇒ F[B]] { f ⇒ (a: A) ⇒
      n(f(a))
    }
  }

  def mapK[A[_[_]], F[_], G[_], LAF <: HList, LAG <: HList](af: A[F])(
    implicit
    genF: Generic.Aux[A[F], LAF],
    genG: Generic.Aux[A[G], LAG],
    mapper: Mapper.Aux[PolyMapK.type, LAF, LAG]
  ): A[G] = {
    genG.from(mapper(genF.to(af)))
  }
}
// warning: there were 6 feature warnings; for details, enable `:setting -feature' or `:replay -feature'
// defined object FunctorK

implicit val f: Try ~> Option = λ[Try ~> Option](_.toOption)
// f: scala.util.Try ~> Option = $anon$1@e80846a

val b: Algebra[Option] = FunctorK.mapK(Algebra1.intepreter)
// b: Algebra[Option] = Algebra(FunctorK$PolyMapK$$$Lambda$5117/560757496@581f743a,FunctorK$PolyMapK$$$Lambda$5117/560757496@4336ce76)

println(b.parseInt("3"))
// Some(3)

println(b.parseFloat("btr"))
// None
```

PR #1598, opened 2017-04-08
Implement a ReaderWriterStateT data type

iravid:
Resolves #1597.

Still missing:
- [x] Method parity with ReaderT
- [x] MonadWriter, MonadReader instances
- [x] tests and laws
- [x] scaladoc

I'd love to get some feedback on the structure; this is mostly just a re-implementation of methods from StateT, WriterT and ReaderT, which was great for understanding how everything is wired up and is probably more efficient, but I do wonder about the maintainability of this approach.

PR #1621, opened 2017-04-22
Replace TransLift  with MonadTrans

### LINKED ISSUE #1613: Unable to use TransLift as a typeclass
With `TransLift` I'm unable to use it as a typeclass for `liftT` function for any transformer.

With `MonadTrans` I was able to do it. For example to avoid boilerplate in defining typeclass instances for transformers:
```scala
def mkTrans[F[_]: Monad, MT[_[_], _]: MonadTrans](implicit F: MyTypeclass[F]): MyTypeclass[MT[F, ?]] = 
  new MyTypeclass[MT[F, ?]] {
    def op = F.op.liftT[MT]
    def op2(a: Int) = F.op2(a).liftT[MT]
  }

implicit def eitherTMyTypeclass[F[_]: Monad: MyTypeclass, E]: MyTypeclass[EitherT[F, E, ?]] = mkTrans[F, EitherT[?[_], E, ?]]
```
or for `liftT` as  a natural transformation:
```scala
def liftTTrans[F[_]: Monad, G[_[_], _]: MonadTrans] =  λ[F ~> G[F, ?]](_.liftT[G])
```
it can be defined with `TransLift` as `def liftTT[F[_]: Monad, G[_[_], _], H[_[_]]](implicit G: TransLift.Aux[G, H], M: H[F]) = λ[F ~> G[F, ?]](x => G.liftT(x))` but this extra type parameter `H` makes it useless.

For `TransLift` to be useful in syntax it uses some scary thing [TLExtract](https://github.com/typelevel/cats/blob/19c3a3702f3eb995fa0972a9cc65a5da5d0fc22a/core/src/main/scala/cats/syntax/transLift.scala#L15). I hope it's not expected to be used in other places where `TransLift` required as a typeclass.

wedens:
Resolves #1613.

Pros:
* `MonadTrans` is lawful (well... almost). There are identity (`liftT . pure = pure`) and composition (`liftT (m >>= f) = liftT m >>= (liftT . f)`) laws. I think for `MonadTrans` to be really lawful it should extend (in current encoding) `Monad`. But it will make it unusable for the same reason `MonadError` et al is (diamonds 💎 )
* Machinery like [TLExtract](https://github.com/typelevel/cats/blob/01711b0063b7949de8a503b833777fafba6c3a9b/core/src/main/scala/cats/syntax/transLift.scala#L14) is not required
* Can be used as a typeclass without dealing with varying typeclass (#1613 is about it)

Cons:
* `Monad` constraint is imposed on for any base functor. For some transformers (e.g. `Kleisli`) `liftT` can be implemented with weaker constraint

/cc @djspiewak @ceedubs


djspiewak:
This gets a 👍 from me.  I think recent efforts really prove that we weren't gaining anything from the flexibility of `TransLift`, and if anything we were losing a lot due to the insane complexity of getting the types to work out.  The laws point is also a kicker.  It was a really neat experiment, but I think it's fair to say that it failed.  I would rather have `MonadTrans`.

wedens:
CI on 2.11.8 fails due to "The job exceeded the maximum time limit for jobs, and has been terminated.". Using [coursier](https://github.com/coursier/coursier) may chop off some build time.

core/src/main/scala/cats/MonadTrans.scala
```diff
@@ -0,0 +1,16 @@
+package cats
+
+/**
+  * A type class which abstracts over the ability to lift an M[A] into a
+  * MonadTransformer
+  */
```
kailuowang: nit here: the scala doc style in cats is 
```
/**
 *
 */
``` 
i.e. all stars aligned with the first star.

kailuowang: other than that it's a 👍  from me.

wedens: fixed

core/src/main/scala/cats/MonadTrans.scala
```diff
@@ -0,0 +1,16 @@
+package cats
+
+/**
+ * A type class which abstracts over the ability to lift an M[A] into a
+ * MonadTransformer
+ */
+trait MonadTrans[MT[_[_], _]] {
+  /**
+    * Lift a value of type M[A] into a monad transformer MT[M, A]
+    */
```
kailuowang: And this one.  also maybe a blank line above?

wedens: haven't noticed it ;)

ceedubs:
Does anyone know of any data structures that require `Monad` to lift as opposed to just `Applicative` (or weaker)? I don't see any examples currently inside Cats, which makes me wonder if it should be `ApplicativeTrans`. However, I guess the downside with that is that you lose the composition law?

wedens:
@ceedubs `ContT`, I think

djspiewak:
Apparently `Codensity` does as well.  @sellout and I had this conversation last week.

sellout:
@wedens I don’t think `ContT`, but I’m far from certain.

@ceedubs Is there anything other than `Codensity` that requires more than `Functor`? I ask because `FunctorK[F[_[_], _]]` (or `HFunctor`, or whatever) is very similar to Scalaz’s `Hoist` and `PointedK` (or `HPointed`) is very similar to `MonadTrans`, _except_ that the `K` (or `H`) variants only allow a `Functor` constraint (I think), so I thought that perhaps we could have
```scala
@typeclass trait PointedK[F[_[_], _]] extends MonadTrans[F] {
  def pointk[G[_]: Functor]: G ~> F[G, ?]

  def liftT[M[_]: Monad, A](ma: M[A]) = pointk[M].apply(ma)
}
```
(ok, maybe we have to deal with the natural transformation somehow, but you get the gist). Right now, `Codensity` is the only thing that I’m pretty sure can implement `MonadTrans` but not `PointedK`.

And the underlying need for `FunctorK`, etc. is that with mutually-recursive recursion schemes you end up with a _lot_ of `G[_]`s that aren’t monads.

djspiewak:
@sellout I'm pretty sure you can't define `runCont` without a `Monad` for the constituent effect of `ContT`.

sellout:
@djspiewak Maybe (I think – again, not sure – you only need Applicative), but that doesn’t mean you can’t `liftT` with a lesser constraint.

djspiewak:
@sellout Oh, [you're right](https://github.com/djspiewak/cont-exp/blob/master/src/main/scala/cont/ContT.scala#L25).  The implementation of `runCont` doesn't even need `Applicative`.  `ContT` could be done without any constraints at all.

sellout:
I’m going to frame that comment.

core/src/main/scala/cats/syntax/monadTrans.scala
```diff
@@ -0,0 +1,11 @@
+package cats
+package syntax
+
+trait MonadTransSyntax {
+  implicit final def catsSyntaxMonadTrans[F[_], A](fa: F[A]): MonadTransOps[F, A] = new MonadTransOps(fa)
+}
+
+final class MonadTransOps[F[_], A](val fa: F[A]) extends AnyVal {
+  def liftT[MT[_[_], _]](implicit F: Monad[F], MT: MonadTrans[MT]): MT[F, A] =
```
kailuowang: How about a doctest here? It will provide coverage for both here and the syntax line above.

core/src/main/scala/cats/Trivial.scala
```diff
@@ -1,23 +0,0 @@
...
- * The "Unit type class".  The only instance of `Trivial` is given by
- * `Trivial.manifest`, and this instance is guaranteed to be in the
- * implicit scope.  Several convenience type aliases are provided in
- * companion object, covering a few common use cases and avoiding the
- * need for unnecessary lambdas (e.g. if you want a trivial type class
- * instance for a type constructor, you should use `Trivial.PH1`).
- */
-sealed trait Trivial
-
-object Trivial {
```
alexknvl: I understand that as a result of this PR `Trivial` is no longer needed in cats, but it is still a very useful type constraint. Is there a good reason to remove `Trivial`?

djspiewak: I thought about making the same objection, but my conclusion was that "less is more" in terms of API surface area.  Shapeless already has a `Trivial`, and it is very likely to be on the classpath of anyone who needs this sort of thing.  I don't think there's a need for cats to keep it.

kailuowang: We are also going to remove `unapply` to a separate module. Maybe we should copy trivial there too?

core/src/main/scala/cats/syntax/monadTrans.scala
wedens: There is a test in `SyntaxTests`.

core/src/main/scala/cats/syntax/all.scala
```diff
@@ -39,7 +39,7 @@ trait AllSyntax
     with ShowSyntax
     with SplitSyntax
     with StrongSyntax
-    with TransLiftSyntax
+    with MonadTransSyntax
```
mpilquist: Keep in alphabetical order?

wedens: ok

core/src/main/scala/cats/MonadTrans.scala
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * A type class which abstracts over the ability to lift an M[A] into a
+ * MonadTransformer
+ */
+trait MonadTrans[MT[_[_], _]] {
+
+  /**
+   * Lift a value of type M[A] into a monad transformer MT[M, A]
+   */
+  def liftT[M[_]: Monad, A](ma: M[A]): MT[M, A]
```
adelbertc: One thing Edward Kmett noted in a talk he gave on [monad homomorphisms](https://youtu.be/YTaNkWjd-ac) is that `MonadTrans` provides no constraint that the resulting `MT[M, A]` has a `Monad` instance. I believe in his talk he fixes this by adding a function taking the `Monad[M]` constraint and providing a `Monad[MT[M, ?]]` constraint. In Scala it might look something like:

```scala
implicit def mtMonad[M[_]: Monad]: Monad[MT[M, A]]
```

Would we want something like that here?

wedens: scalaz [does it](https://github.com/scalaz/scalaz/blob/series/7.3.x/core/src/main/scala/scalaz/MonadTrans.scala#L16). Not sure if it's useful. But it makes sense in theory.

adelbertc: This would communicate the law: For every lawful `M[_]: Monad` there exists a lawful `MT[M, ?]: Monad`

tpolecat: Can we check such a law? Would be cool if we could rule out the naïve `ListT` for instance.

adelbertc: A simple way would be to pick a fixed `M`, e.g. `Option`, get the `Monad[MT[Option, ?]]` from this method, and then law-check that.

A fancier way would be to use existentials or something to be able to vary `M` and do law-checking across all these :-)

peterneyens:
Should we merge this as is and add an eventual `Monad[MT[M, ?]]` law in a follow up PR?

kailuowang:
Note:  https://github.com/milessabin/kittens/pull/37 to move `Trivial` to kittens.

wedens:
is there something holding this PR?

kailuowang:
@wedens I just thumbed up @peterneyens's last comment. @adelbertc what do you think? If you have no issue with that, I think this one is good to merge.

ceedubs:
While it took a while, I'm sold on the fact that `TransLift` was a failed experiment.

One thing that I didn't realize at first and that I haven't seen come up in comments (other than the original PR description) is that here, `MonadError` doesn't extend `Monad` (citing diamonds as the issue). It sounds like there are some plans to provide mechanisms to attach laws to this, but I just wanted to call that out and make sure that people were aware and don't see it being an issue. cc @djspiewak @adelbertc @kailuowang @sellout

kailuowang:
@ceedubs I just created the issue for adding such a law. 
https://github.com/typelevel/cats/issues/1674
As for this PR, I think we are good to merge?

core/src/main/scala/cats/syntax/monadTrans.scala
kailuowang: I pointed out because codecov reported test miss here. Now I see that `SyntaxTest` is only a compilation check. I think I prefer a doctest in the type class which actually run the code (thus coverage) and provide a concrete example there. But we can certainly address that in a different PR.

ceedubs:
@tpolecat @edmundnoble @djspiewak are you still happy with this PR in response to https://github.com/typelevel/cats/pull/1621#issuecomment-301257576 ? I'm definitely not opposed to this PR being merged, but I respect others' opinion on this much more than my own, because I only rarely use `MonadTrans`.

peterneyens:
I think we may need to change the `Translift` of `ReaderWriterStateT` to `MonadTrans` after #1598.

djspiewak:
@ceedubs 

> One thing that I didn't realize at first and that I haven't seen come up in comments (other than the original PR description) is that here, MonadError doesn't extend Monad (citing diamonds as the issue). 

Uh, am I missing something?  `MonadError` isn't touched in this PR.

ceedubs:
@djspiewak oops I had `MonadError` on the brain from the `Eval` discussion. I meant `MonadTrans`.

ceedubs:
Wait that makes sense that `MonadTrans` wouldn't extend `Monad`. Ignore my previous comments.

ceedubs:
Okay this has 3 signoffs. I'm going to go ahead and merge it.

ceedubs:
Oops I had forgotten about https://github.com/typelevel/cats/pull/1621#issuecomment-302232935 . I can't take care of it at the moment. I should be able to do it this evening if nobody beats me to it.

wedens:
@ceedubs I've fixed it #1680

edmundnoble:
@wedens Just so you know, cats-mtl has its own formulation of `MonadTrans` which is not compatible with this one. So I will have to remove it when I remove the MTL type classes from cats.

PR #1623, opened 2017-04-22
Improve test coverage `IdT`

peterneyens:
I am not sure how often `IdT` is actually used, but this PR should improve its test coverage.

I think it makes sense to merge #1621 first, I will replace `TransLift` with `MonadTrans` in this PR afterwards.


peterneyens:
Updated to use `MonadTrans` and added some missing serializable tests for `MonadTrans` instances.

kailuowang:
:+1:

PR #1626, opened 2017-04-24
WIP: Attempt to reduce build time on CI

wedens:
- [x] cache `pip` artifacts (codecov and deps)
- [x] cache `gem` artifacts (jekyll and deps)
- [x] enable [coursier](https://github.com/coursier/coursier) for faster dependencies resolution (only on CI)


wedens:
looks like it's not helping much :/

PR #1627, opened 2017-04-24
improve EitherT inference for use in for-comprehension

jtjeferreira:


kailuowang:
The original signature is to enable for comprehension for EitherT with different left type, see  https://github.com/typelevel/cats/pull/1506. We've been back and forth on this one. Should probably add a FAQ entry.

jtjeferreira:
@kailuowang I asked first in https://gitter.im/typelevel/cats?at=58f8c38708c00c092a9bdb7c. Maybe a unit test would be enough...

kailuowang:
@jtjeferreira to be clear I am 👎  on dropping support for for-comprehension with `EitherT` with different Left types. I think its benefit justifies the inconvenience of the imperfect type inference.

jtjeferreira:
I understood. Was just suggesting that instead of adding something to the FAQ, just add that example from #1506 to a unit test. I can create a PR with that...

PR #1628, opened 2017-04-24
add init and size methods to NonEmptyList

jtjeferreira:


core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -17,15 +17,57 @@ final case class NonEmptyList[+A](head: A, tail: List[A]) {
...
+   *
+   * {{{
+   * scala> import cats.data.NonEmptyList
+   * scala> val nel = NonEmptyList.of(1, 2, 3, 4, 5)
+   * scala> nel.init
+   * res0: scala.collection.immutable.List[Int] = List(1, 2, 3, 4)
+   * }}}
+   */
+  def init: List[A] = tail match {
+    case Nil => List.empty
```
peterneyens: Shouldn't this be `head :: Nil`?

jtjeferreira: If the tail is empty the last element is the head, so the result is the empty list

jtjeferreira: @peterneyens I added tests for `init` and `size`

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -17,15 +17,57 @@ final case class NonEmptyList[+A](head: A, tail: List[A]) {
...
+   * {{{
+   * scala> import cats.data.NonEmptyList
+   * scala> val nel = NonEmptyList.of(1, 2, 3, 4, 5)
+   * scala> nel.last
+   * res0: Int = 5
+   * }}}
+   */
   def last: A = tail.lastOption match {
     case None    => head
     case Some(a) => a
```
kailuowang: `tail.lastOption.getOrElse(head)` might be more performant.

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -17,15 +17,57 @@ final case class NonEmptyList[+A](head: A, tail: List[A]) {
...
+   * Selects all elements except the last
+   *
+   * {{{
+   * scala> import cats.data.NonEmptyList
+   * scala> val nel = NonEmptyList.of(1, 2, 3, 4, 5)
+   * scala> nel.init
+   * res0: scala.collection.immutable.List[Int] = List(1, 2, 3, 4)
+   * }}}
+   */
+  def init: List[A] = tail match {
```
jtjeferreira: @kailuowang any preference regarding init performance? `toList.init` is also an alternative

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -17,15 +17,54 @@ final case class NonEmptyList[+A](head: A, tail: List[A]) {
...
+   * Selects the last element
+   * {{{
+   * scala> import cats.data.NonEmptyList
+   * scala> val nel = NonEmptyList.of(1, 2, 3, 4, 5)
+   * scala> nel.last
+   * res0: Int = 5
+   * }}}
+   */
+  def last: A = tail.lastOption.getOrElse(head)
+
```
johnynek: Please don't change this implementation. The one above avoids the allocation of the closure and was the motivation for writing it that way.

jtjeferreira: ok. will revert

core/src/main/scala/cats/data/NonEmptyList.scala
kailuowang: I am going to stop pretend that I know anything about performance. I don't know if much difference can be made here.

core/src/main/scala/cats/data/NonEmptyList.scala
kailuowang: oops, I forgot that `getOrElse` is by-name. 👅  we need a strict `getOrElese`.

ceedubs:
👍 from me after https://github.com/typelevel/cats/pull/1628/files#r112997682 is addressed.

core/src/main/scala/cats/data/NonEmptyList.scala
edmundnoble: It (should) be more performant this way I think, because there's one less branch. But only very slightly.

core/src/main/scala/cats/data/NonEmptyList.scala
johnynek: he got the exact opposite advice from me a few days ago. Why do you say this? My understanding is that case class matching is very efficient and this also avoids the closure allocation (since getOrElse is by-name).

kailuowang: @johnynek yeah, I forgot that getOrElse is by-name, and corrected myself in the other comment https://github.com/typelevel/cats/pull/1628#discussion_r113015121

johnynek:
👍

PR #1629, opened 2017-04-24
add unit test for variance on methods in EitherT

jtjeferreira:
related to #1627 and #1506

### DIFF
--- a/tests/src/test/scala/cats/tests/EitherTTests.scala
+++ b/tests/src/test/scala/cats/tests/EitherTTests.scala
@@ -392,10 +392,19 @@ class EitherTTests extends CatsSuite {
       s2 <- EitherT[Id, AppError, String](either2)
     } yield s1 ++ s2
 
+    for {
+      s1 <- EitherT(either1)
+      s2 <- EitherT.right[AppError]("1".pure[Id])
+    } yield s1 ++ s2
 
     for {
       s1 <- EitherT(either1)
-      s2 <- EitherT.pure[Id, AppError, String]("1")
+      s2 <- EitherT.left[String](Error1.pure[Id])
+    } yield s1 ++ s2
+
+    for {
+      s1 <- EitherT(either1)
+      s2 <- EitherT.pure[Id, AppError]("1")
     } yield s1 ++ s2
   }
 }
--- a/tests/src/test/scala/cats/tests/EitherTTests.scala
+++ b/tests/src/test/scala/cats/tests/EitherTTests.scala
@@ -378,4 +378,24 @@ class EitherTTests extends CatsSuite {
       }
     }
   }
+
+  test("inference works in for-comprehension") {
+    sealed abstract class AppError
+    case object Error1 extends AppError
+    case object Error2 extends AppError
+
+    val either1: Id[Either[Error1.type , String]] = Right("hi").pure[Id]
+    val either2: Id[Either[Error2.type , String]] = Right("bye").pure[Id]
+
+    for {
+      s1 <- EitherT(either1)
+      s2 <- EitherT[Id, AppError, String](either2)
+    } yield s1 ++ s2
+
+
+    for {
+      s1 <- EitherT(either1)
+      s2 <- EitherT.right[Id, AppError, String]("1".pure[Id])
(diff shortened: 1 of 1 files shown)


tests/src/test/scala/cats/tests/EitherTTests.scala
```diff
@@ -378,4 +378,24 @@ class EitherTTests extends CatsSuite {
...
+
+    for {
+      s1 <- EitherT(either1)
+      s2 <- EitherT[Id, AppError, String](either2)
+    } yield s1 ++ s2
+
+
+    for {
+      s1 <- EitherT(either1)
+      s2 <- EitherT.right[Id, AppError, String]("1".pure[Id])
```
kailuowang: you should be able to write `EitherT.pure[Id, AppError]("1")` here

jtjeferreira: done

ceedubs:
@jtjeferreira thanks!

It looks like there's a compile error in the test:

```
[error] /home/travis/build/typelevel/cats/tests/src/test/scala/cats/tests/EitherTTests.scala:398: wrong number of type parameters for method pure of type [F[_], A]=> cats.data.EitherT.PurePartiallyApplied[F,A]
[error]       s2 <- EitherT.pure[Id, AppError, String]("1")
```

jtjeferreira:
@ceedubs that's very weird because I cant reproduce it locally...

kailuowang:
@jtjeferreira your local code is probably outdated.

ceedubs:
@jtjeferreira here's how I'm able to reproduce it locally:

```
❯ sbt
...
> coreJVM/doc
[info] Main Scala API documentation to /Users/cody/code/cats/core/.jvm/target/scala-2.12/api...
model contains 525 documentable templates
[warn] /Users/cody/code/cats/core/src/main/scala/cats/data/Validated.scala:248: Variable s undefined in comment for method ensureWith in class Validated
[warn]     * scala> Validated.valid("ab").ensureWith(s => new IllegalArgumentException(s"Must be longer than 3, provided '$s'"))(_.length > 3)
[warn]                                                                                                                     ^
[warn] one warning found
[info] Main Scala API documentation successful.
[success] Total time: 46 s, completed Apr 25, 2017 12:45:43 PM
```

I think that we have the fatal warnings flag enabled, so this ends up being a fatal error. Note that it doesn't complain if you use `++2.10.6`, I think because we've disabled scaladoc creation for Scala 2.10 due to some other issues.

jtjeferreira:
my local branch was outdated, rebased and fixed. waiting for Travis...

tests/src/test/scala/cats/tests/EitherTTests.scala
```diff
@@ -378,4 +378,33 @@ class EitherTTests extends CatsSuite {
...
+
+    for {
+      s1 <- EitherT(either1)
+      s2 <- EitherT.left[String](Error1.pure[Id])
+    } yield s1 ++ s2
+
+    for {
+      s1 <- EitherT(either1)
+      s2 <- EitherT.pure[Id, AppError]("1")
+    } yield s1 ++ s2
```
kailuowang: I might've missed something but it's not very clear to me why there are all 4 tests needed here. The types of the two `EitherT`s in the first, second and last test are exactly the same, aren't they?

jtjeferreira: @kailuowang hi sorry for the late reply. These tests are just to prevent regressions like the one I was going to introduce in #1627 so for the sake of completion I covered the different use cases...

jtjeferreira: on the other hand this was more problematic issue before #1583 was merged...

ceedubs: Thanks @jtjeferreira I think that there is some value in testing the inference here, and I certainly don't think the extra unit tests hurt anything. 👍 from me.

kailuowang: yeah, I guess I was just nitpicking.

PR #1631, opened 2017-04-24
make all PartialApplied class AnyVal to achieve zero cost

kailuowang:
A trick from scalaz thanks to @edmundnoble from this [comment](https://github.com/typelevel/cats/pull/1614#discussion_r112397323)

### DIFF
--- a/core/src/main/scala/cats/data/Const.scala
+++ b/core/src/main/scala/cats/data/Const.scala
@@ -40,7 +40,7 @@ object Const extends ConstInstances {
   def empty[A, B](implicit A: Monoid[A]): Const[A, B] =
     Const(A.empty)
 
-/** Uses the [[http://typelevel.org/cats/guidelines.html#partially-applied-type Partially Applied Type technique]] for ergonomics. */
+/** Uses the [[http://typelevel.org/cats/guidelines.html#partially-applied-type-params Partially Applied Type Params technique]] for ergonomics. */
   private[data] final class OfPartiallyApplied[B](val dummy: Boolean = true ) extends AnyVal {
     def apply[A](a: A): Const[A, B] = Const(a)
   }
--- a/core/src/main/scala/cats/data/Const.scala
+++ b/core/src/main/scala/cats/data/Const.scala
@@ -40,7 +40,7 @@ object Const extends ConstInstances {
   def empty[A, B](implicit A: Monoid[A]): Const[A, B] =
     Const(A.empty)
 
-  private[data] final class OfPartiallyApplied[B](val dummy: Boolean) extends AnyVal {
+  private[data] final class OfPartiallyApplied[B](val dummy: Boolean = true ) extends AnyVal {
     def apply[A](a: A): Const[A, B] = Const(a)
   }
 
@@ -52,7 +52,7 @@ object Const extends ConstInstances {
    * res0: Const[String, Int] = Const(a)
    * }}}
    */
-  def of[B]: OfPartiallyApplied[B] = new OfPartiallyApplied(true)
+  def of[B]: OfPartiallyApplied[B] = new OfPartiallyApplied
 }
 
 private[data] sealed abstract class ConstInstances extends ConstInstances0 {
--- a/core/src/main/scala/cats/data/Const.scala
+++ b/core/src/main/scala/cats/data/Const.scala
@@ -40,7 +40,9 @@ object Const extends ConstInstances {
   def empty[A, B](implicit A: Monoid[A]): Const[A, B] =
     Const(A.empty)
 
-/** Uses the [[http://typelevel.org/cats/guidelines.html#partially-applied-type-params Partially Applied Type Params technique]] for ergonomics. */
+  /**
+   * Uses the [[http://typelevel.org/cats/guidelines.html#partially-applied-type-params Partially Applied Type Params technique]] for ergonomics.
+   */
   private[data] final class OfPartiallyApplied[B](val dummy: Boolean = true ) extends AnyVal {
     def apply[A](a: A): Const[A, B] = Const(a)
   }
--- a/core/src/main/scala/cats/data/Const.scala
+++ b/core/src/main/scala/cats/data/Const.scala
@@ -40,6 +40,9 @@ object Const extends ConstInstances {
   def empty[A, B](implicit A: Monoid[A]): Const[A, B] =
     Const(A.empty)
 
+  /**
+   * @param dummy is introduced solely for the sake of making this a value class and thus zero allocation cost.
+   */
   private[data] final class OfPartiallyApplied[B](val dummy: Boolean = true ) extends AnyVal {
     def apply[A](a: A): Const[A, B] = Const(a)
   }
--- a/core/src/main/scala/cats/data/Const.scala
+++ b/core/src/main/scala/cats/data/Const.scala
@@ -40,7 +40,7 @@ object Const extends ConstInstances {
   def empty[A, B](implicit A: Monoid[A]): Const[A, B] =
     Const(A.empty)
 
-  final class OfPartiallyApplied[B] {
+  private[data] final class OfPartiallyApplied[B](val dummy: Boolean) extends AnyVal {
     def apply[A](a: A): Const[A, B] = Const(a)
   }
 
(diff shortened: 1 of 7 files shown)


johnynek:
this is great!

👍

fthomas:
If you define `dummy` with a default value (`val dummy: Boolean = true`), you don't need to change the call sites.

kailuowang:
@fthomas thanks. done

ceedubs:
Scala is silly sometimes 😄

I'm in favor of this change, but I think that it would be good to leave `@param dummy <explanation>` scaladoc comments wherever this appears, because someone encountering this could reasonably be pretty bewildered by the dummy parameter.

I also always forget that you can make classes package-private but still statically return their type from a public method. Should there be any concern about doing this?

kailuowang:
@ceedubs I added the docs. As for making them package private, I couldn't think of any downside either.

sellout:
At SlamData, the `PartiallyApplied` variant we use looks like https://github.com/quasar-analytics/quasar/wiki/Scala%5BQuasar%5D#partially-applied-type-parameters
1. we scope the `PartiallyApplied` class inside the object, which means we don’t need unique names and
2. is there some alternative to `(val dummy: Boolean = true)` that won’t trigger the `DefaultArguments` wart? (so we can steal it at SD, too)

I’d also be inclined to use `(val dummy: Unit = ())`, just to make it clearer that there’s no info in the argument.

Finally, I think there should just be one place where this trick is documented, and maybe link to it from the use sites, because there is far more than `dummy` that is confusing as hell about this technique (as useful as it is).

sellout:
To answer Q2, I guess I could not use the default argument, and instead use `new PartiallyApplied[W](())`. That seems fine.

sellout:
And, as you people probably already know, `Unit` causes the typer to throw a fit, so `Boolean` it is …

ceedubs:
> we scope the PartiallyApplied class inside the object, which means we don’t need unique names

@sellout are you suggesting that we should have something like `private[this] final class OfPartiallyApplied` or `private[Const] final class OfPartiallyApplied` instead of the current `private[data] final class OfPartiallyApplied`? I don't see any reason that we need it scoped to the package as opposed to just the object, so that seems like a decent suggestion to me if that's what you are saying.

I agree that it makes sense to document this pattern somewhere and link to it in the relevant call sites. I'm fine with that either happening in this PR or creating an issue to follow up on it, since I think that this PR still represents progress and these classes weren't documented before anyway.

sellout:
@ceedubs I meant instead of
```scala
  private[data] final class OfPartiallyApplied[B](val dummy: Boolean = true ) extends AnyVal {
    def apply[A](a: A): Const[A, B] = Const(a)
  }
  def of[B]: OfPartiallyApplied[B] = new OfPartiallyApplied
```
having
```scala
  object of {
    def apply[B]: PartiallyApplied[B] = new PartiallyApplied
    private[of] final class PartiallyApplied[B](val dummy: Boolean = true) extends AnyVal {
      def apply[A](a: A): Const[A, B] = Const(a)
    }
  }
```
I put the tight `private[of]` in there only because of your comment, but it definitely wasn’t what I meant. I meant turning the non-private `def` into an `object` with an `apply`, then moving the private class inside it.

And this isn’t a “my way is better” thing, just a “I’ve been doing it this way, and I have no idea what the tradeoffs are, and I’d generally like to do what Cats/Typelevel thinks is best” (especially if it’s documented somewhere that I am not maintaining 😄).

The downside of my approach, I think, is that the Scaladoc gets even worse. Yours at least looks like a `def` and has _some_ (the required) type parameters on it. Mine shows up as an object with no type params at all (see [`transPrepro`](https://oss.sonatype.org/service/local/repositories/releases/archive/com/slamdata/matryoshka-core_sjs0.6_2.12/0.19.0/matryoshka-core_sjs0.6_2.12-0.19.0-javadoc.jar/!/matryoshka/Birecursive$$Ops.html#transPrepro) vs [`cond`](http://typelevel.org/cats/api/cats/data/EitherT$.html#cond[F[_]]:EitherTFunctions.this.CondPartiallyApplied[F])). The fact that the non-nested approach is so much clearer there makes me inclined to switch to it.

And yes – these comments are less suggestions for this PR and more that it raised questions about how we do the same thing at SlamData.

So, I think my new approach is basically “do it like in Cats, except for the default argument.” I’ll update my docs, and hopefully it’ll eventually end up in some official Typelevel contributor doc.

kailuowang:
@sellout @ceedubs thanks for all the input. I added an entry to the guideline document and link to it from the use sites. I agree it's quite necessary.

docs/src/main/tut/guidelines.md
```diff
@@ -11,17 +11,64 @@ All guidelines in cats should have clear justifications. There is no room for tr
...
 
-### Ops Classes
+### <a id="ops-classes" href="#ops-classes"></a> Ops Classes
 
 Ops classes should be marked final and extend AnyVal, to take full advantage of inlining and prevent unnecessary allocations.
 
 The most notable exception is the case where all of the ops in the class are provided by zero-cost macros anyway,
 for example with Simulacrum.
 
+### <a id="partially-applied-type" href="#partially-applied-type"></a> Partially-Applied Type
```
edmundnoble: Is this perhaps more accurately referred to as "partially applied type params"?

docs/src/main/tut/guidelines.md
```diff
@@ -11,17 +11,64 @@ All guidelines in cats should have clear justifications. There is no room for tr
...
+
+In Scala, when there are multiple type parameters in a function, either scalac infers all type parameters or the user has to
+specify all of them. Often we have functions where there are one or more types that are inferable but not all of them. For example, there is helper function in `OptionT` that creates an `OptionT[A]` from an `A`. It could be written as:
```
edmundnoble: s/`OptionT[A]`/`OptionT[F, A]`?

docs/src/main/tut/guidelines.md
```diff
@@ -11,17 +11,64 @@ All guidelines in cats should have clear justifications. There is no room for tr
...
+```
+```tut:book
+def pure[F[_], A](a: A)(implicit F: Applicative[F]): OptionT[F, A] =
+  OptionT(F.pure(Some(a)))
+
+
+pure[List, Int](1)
+```
+
+Note that the type `A` should've been given by the `a: A` argument, but since scalac cannot infer `F[_]`, user still have to specify all type params.
```
edmundnoble: s/user/the user, s/have/has?

docs/src/main/tut/guidelines.md
kailuowang: I used the term "Partially-Applied Type" to refer to the intermediate class as a new type with type parameters partially applied. Now that I said it out loud, It's quite a mouthful. I'll change it.

kailuowang:
@ceedubs, @fthomas do either of you have time to take another look?

core/src/main/scala/cats/data/Const.scala
```diff
@@ -40,7 +40,8 @@ object Const extends ConstInstances {
   def empty[A, B](implicit A: Monoid[A]): Const[A, B] =
     Const(A.empty)
 
-  final class OfPartiallyApplied[B] {
+/** Uses the [[http://typelevel.org/cats/guidelines.html#partially-applied-type-params Partially Applied Type Params technique]] for ergonomics. */
+  private[data] final class OfPartiallyApplied[B](val dummy: Boolean = true ) extends AnyVal {
```
ceedubs: Minor, unnecessarily picky comment: the alignment of the scaladoc comment is a little off here.

free/src/main/scala/cats/free/Free.scala
```diff
@@ -221,8 +221,9 @@ object Free {
 
   /**
    * Pre-application of an injection to a `F[A]` value.
+   * @param dummy is introduced solely for the sake of making this a value class and thus zero allocation cost.
```
ceedubs: Should this have the same comment with link that the other places have?

kailuowang:
@ceedubs comments addressed.

kailuowang:
ping @ceedubs again

ceedubs:
@kailuowang sorry for the long delay and on such trivial nitpicky comments 😲. This looks great to me! 👍

fthomas:
For posterity: the reason why the `dummy` parameter is of type `Boolean` instead of `Unit` is https://github.com/scala/bug/issues/9240

PR #1632, opened 2017-04-24
upgraded to scala 2.12.2 and 2.11.11 and scalaJs

kailuowang:


johnynek:
do we need to set: `-Ypartial-unification` or is that no longer needed?

kailuowang:
it's taken care of by the plugin sbt-partial-unification

build.sbt
```diff
@@ -571,3 +570,12 @@ lazy val update2_12 = Seq(
     }
   }
 )
+
+lazy val xlint = Seq(
+  scalacOptions += {
+    CrossVersion.partialVersion(scalaVersion.value) match {
+      case Some((2, 12)) => "-Xlint:-unused,_"
```
sellout: What’s the motivation for this?

I guess in 2.11, `unused` wasn’t part of `lint`, but is there a particular problem with it (asking because I’m in the process of enabling it on all my projects!)

kailuowang: This now warns on unused parameters, vals and implicits according to `scalac -xlint:help`

sellout: Right, but that sounds like a good thing to me. Where does that cause problems?

kailuowang: I got quite a few seemingly related to simulacrum:
>[error] /home/travis/build/typelevel/cats/core/src/main/scala/cats/Apply.scala:10: parameter value ev$macro$3 in method ap is never used
[error] @typeclass(excludeParents = List("ApplyArityFunctions"))

ceedubs: Ah I see. We have some evidence parameters that we don't directly use.

There's some related discussion [here](https://github.com/scala/scala/pull/5402#issuecomment-295261637). It sounds like scala doesn't offer great options for selectively ignoring these. Unless someone sees a better option, I'm fine with this change coming in as-is.

kailuowang:
@ceedubs @johnynek do you have more feedbacks on this one?

build.sbt
DavidGregory084: @kailuowang @ceedubs perhaps we should keep an eye on [this PR](https://github.com/scala/scala/pull/5876) as it might help with Simulacrum-related warnings

kailuowang: that's good to know. Thanks @DavidGregory084

project/plugins.sbt
```diff
@@ -7,7 +7,7 @@ addSbtPlugin("pl.project13.scala"  % "sbt-jmh"               % "0.2.16")
 addSbtPlugin("org.scalastyle"      % "scalastyle-sbt-plugin" % "0.8.0")
 addSbtPlugin("org.scoverage"       % "sbt-scoverage"         % "1.5.0")
 addSbtPlugin("com.typesafe.sbt"    % "sbt-git"               % "0.8.5")
-addSbtPlugin("org.scala-js"        % "sbt-scalajs"           % "0.6.14")
+addSbtPlugin("org.scala-js"        % "sbt-scalajs"           % "0.6.15")
```
edmundnoble: Why not 0.6.16?

kailuowang: just updated. 0.6.15 was the latest when this PR was created.

PR #1636, opened 2017-04-25
added some category theory into FunctionK document

kailuowang:
I think this will be helpful because it's often referred as Natural ~~Transformer~~ Transformation in other context.


wedens:
I think it's known as "natural transformation", not "transformer" :)

kailuowang:
@wedens thanks for pointing out. Corrected.

docs/src/main/tut/datatypes/functionk.md
```diff
@@ -121,3 +121,25 @@ type ErrorOr[A] = Either[String, A]
 val errorOrFirst: FunctionK[List, ErrorOr] =
   λ[FunctionK[List, ErrorOr]](_.headOption.toRight("ERROR: the list was empty!"))
 ```
+
+## Natural Transformation
+
+In category theory, a [Natural Transformation](https://en.wikipedia.org/wiki/Natural_transformation) provides a morphism between Functors while preserving the internal structures. It's one of the most fundamental notions of category theory.
```
SystemFw: > while preserving the internal structures

I think _the internal structure_ is slightly better. It refers to the composition of morphisms (similarly to what a functor does)

docs/src/main/tut/datatypes/functionk.md
```diff
@@ -121,3 +121,25 @@ type ErrorOr[A] = Either[String, A]
...
+(fa:F[A]) => fK(F.map(fa)(f))
+```
+and
+```Scala
+(fa: F[A]) => G.map(fK(fa))(f)
+````
+
+We don't need to write a law to test the implementation of the `fk` for the above to be true. It's automatically given by its parametric polymorphism.
+
+This is why a parametric polymorphic function `FunctionK[F, G]` is sometime referred as a Natural Transformation.
```
SystemFw: Pedantry warning (sorry): it should be `sometimes`, not `sometime`.  :)

docs/src/main/tut/datatypes/functionk.md
```diff
@@ -121,3 +121,25 @@ type ErrorOr[A] = Either[String, A]
...
+However, they are two different concepts. `FunctionK` is a stronger and stricter construction than Natural Transformation.
+
```
SystemFw: I'm not entirely sure this is worth including unless we can give an explanation as to how do they differ

docs/src/main/tut/datatypes/functionk.md
```diff
@@ -121,3 +121,25 @@ type ErrorOr[A] = Either[String, A]
...
   λ[FunctionK[List, ErrorOr]](_.headOption.toRight("ERROR: the list was empty!"))
 ```
+
+## Natural Transformation
+
+In category theory, a [Natural Transformation](https://en.wikipedia.org/wiki/Natural_transformation) provides a morphism between Functors while preserving the internal structures. It's one of the most fundamental notions of category theory.
+
+If we have two `Functor`s: `F` and `G`, then, being parametric polymorphic, `FunctionK[F, G]` is automatically a Natural Transformation between them. That is, if we have a `fk: F ~> G`, then for any combination of `A`, `B` and function `f: A => B`, the following two functions are equivalent:
+```Scala
+(fa:F[A]) => fK(F.map(fa)(f))
```
alexknvl: Minor inconsistency: `fK` and `fk`.

kailuowang:
@ceedubs I am interested in hearing your thoughts on this.

docs/src/main/tut/datatypes/functionk.md
```diff
@@ -121,3 +121,24 @@ type ErrorOr[A] = Either[String, A]
...
   λ[FunctionK[List, ErrorOr]](_.headOption.toRight("ERROR: the list was empty!"))
 ```
+
+## Natural Transformation
+
+In category theory, a [Natural Transformation](https://en.wikipedia.org/wiki/Natural_transformation) provides a morphism between Functors while preserving the internal structure. It's one of the most fundamental notions of category theory.
+
+If we have two `Functor`s: `F` and `G`, then, being parametric polymorphic, `FunctionK[F, G]` is automatically a Natural Transformation between them. That is, if we have a `fk: F ~> G`, then for any combination of `A`, `B` and function `f: A => B`, the following two functions are equivalent:
+```Scala
+(fa:F[A]) => fK(F.map(fa)(f))
```
ChristopherDavenport: This has `fK` where is is defined above as `fk`

docs/src/main/tut/datatypes/functionk.md
```diff
@@ -121,3 +121,24 @@ type ErrorOr[A] = Either[String, A]
...
+(fa:F[A]) => fk(F.map(fa)(f))
+```
+and
+```Scala
+(fa: F[A]) => G.map(fk(fa))(f)
+````
+
+We don't need to write a law to test the implementation of the `fk` for the above to be true. It's automatically given by its parametric polymorphism.
+
+This is why a parametric polymorphic function `FunctionK[F, G]` is sometimes referred as a Natural Transformation. However, they are two different concepts.
```
ceedubs: Minor: is there a particular reason that you capitalized "Natural Transformation" here?

ceedubs:
Thanks @kailuowang. I really like linking this back to the category theory (and probably also providing some insight to people who are used to the `NaturalTransformation` name in scalaz). I also like that it's added as a separate section at the end so people can get into the theory if they want to but can stop short of it if they prefer. 👍

ceedubs:
@adelbertc as someone who has written a lot of docs and read up on your category theory, does this look good to you?

docs/src/main/tut/datatypes/functionk.md
```diff
@@ -121,3 +121,24 @@ type ErrorOr[A] = Either[String, A]
 val errorOrFirst: FunctionK[List, ErrorOr] =
   λ[FunctionK[List, ErrorOr]](_.headOption.toRight("ERROR: the list was empty!"))
 ```
+
+## Natural Transformation
+
+In category theory, a [Natural Transformation](https://en.wikipedia.org/wiki/Natural_transformation) provides a morphism between Functors while preserving the internal structure. It's one of the most fundamental notions of category theory.
```
adelbertc: ncatlab might be a better reference than Wikipedia for such a subject :-) https://ncatlab.org/nlab/show/natural+transformation

docs/src/main/tut/datatypes/functionk.md
```diff
@@ -121,3 +121,24 @@ type ErrorOr[A] = Either[String, A]
...
+
+If we have two `Functor`s: `F` and `G`, then, being parametric polymorphic, `FunctionK[F, G]` is automatically a Natural Transformation between them. That is, if we have a `fk: F ~> G`, then for any combination of `A`, `B` and function `f: A => B`, the following two functions are equivalent:
```
adelbertc: Perhaps drop the colon, and use 'parametricity', e.g.

If we have two `Functor`s `F` and `G`, `FunctionK[F, G]` is a natural transformation via parametricity. That is, given `fk: FunctionK[F, G]`, for all functions `A => B` the following are equivalent:

docs/src/main/tut/datatypes/functionk.md
```diff
@@ -121,3 +121,24 @@ type ErrorOr[A] = Either[String, A]
...
+```Scala
+(fa:F[A]) => fk(F.map(fa)(f))
+```
+and
+```Scala
+(fa: F[A]) => G.map(fk(fa))(f)
+````
+
+We don't need to write a law to test the implementation of the `fk` for the above to be true. It's automatically given by its parametric polymorphism.
```
adelbertc: by parametricity

docs/src/main/tut/datatypes/functionk.md
adelbertc: Probably don't need to capitalize

adelbertc: This line could use some elaboration.. maybe say something about how `FunctionK` is more powerful than natural transformations, or natural transformations can be implemented in terms of `FunctionK` (I prefer the latter)

adelbertc:
@ceedubs you flatter me :-)

docs/src/main/tut/datatypes/functionk.md
kailuowang: We had a discussion about this on gitter, and @edmundnoble pointed out that natural transformation could be deemed as more powerful in some ways. My intuition is that `FunctionK` is more powerful in one category but natural transformation is a concept that could involve two categories. Maybe when we restrict natural transformation to endofunctors we can say something about it. In conclusion we feel it's too complex a thing to include here.

PR #1637, opened 2017-04-25
show(f:T) to show(t:T)

PeterPerhac:
f could be misleading. usually f parameter is used for functions


kailuowang:
merged with one sign-off since this is a simple typo correction.

Issue #1643, opened 2017-04-30
Add recoverF and mapF to EitherT


leandrob13:
I have been dealing with use cases where there is an `EitherT[Future, X, Y]` and there is the need to recover from a failed future.

```scala
def someFuture(x: String): Future[Int] = Future(x.toInt)
def fe(s: String): Future[Either[String, Int]] = someFuture(s).map { x =>
  if (x == 0) Right(x)
  else Left("Error")
}

EitherT.right[Future, String, Int](someFuture("1")) //EitherT(Future(Success(Right(1))))
EitherT.right[Future, String, Int](someFuture("not an int")) //EitherT(Future(Failure(java.lang.NumberFormatException: For input string: "not an int")))
EitherT(fe("0")) //EitherT(Future(Success(Right(0))))
EitherT(fe("1")) //EitherT(Future(Success(Left(Error))))
EitherT(fe("not an int")) //EitherT(Future(Failure(java.lang.NumberFormatException: For input string: "not an int")))
```

If I want to recover from the future that failed and keep the EitherT context, would have to do:
```scala
EitherT(EitherT(fe("not an int")).value.recoverWith { ... })
```
I want to be able to just do `EitherT(fe("not an int")).recoverWith { ... }` obviously not only for futures but with any F that has a ApplicativeError instance.

Is it a good idea to add a recoverF and recoverFWith functions to EitherT? I have been considering adding a mapF also.

edmundnoble:
Sounds good to me so far. :+1: to making a PR.

leandrob13:
@edmundnoble I'm on it.

djspiewak:
I take it that the `MonadError` instance for `EitherT[F, L, ?]` where `F` also has a `MonadError` is not sufficient to implement this?

leandrob13:
It has a Monad instance, not a MonadError for F so there is no recovery for F. I just checked:

```scala
implicit def catsDataMonadErrorForEitherT[F[_], L](implicit F0: Monad[F]): MonadError[EitherT[F, L, ?], L] =
    new EitherTMonadError[F, L] { implicit val F = F0 }

private[data] trait EitherTMonadError[F[_], L] extends MonadError[EitherT[F, L, ?], L] with EitherTMonad[F, L] {
  def handleErrorWith[A](fea: EitherT[F, L, A])(f: L => EitherT[F, L, A]): EitherT[F, L, A] =
    EitherT(F.flatMap(fea.value) {
      case Left(e) => f(e).value
[...]

PR #1644, opened 2017-05-01
Add MonadError instance for EitherT that recovers from F[_] errors.

leandrob13:
Adds functions for recovery and mapping functions for context F. 

Addressed in #1643


core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -88,11 +88,22 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
 
   def map[D](f: B => D)(implicit F: Functor[F]): EitherT[F, A, D] = bimap(identity, f)
 
+  def mapF[N[_], D](fe: F[Either[A, B]] => N[Either[A, D]]): EitherT[N, A, D] = EitherT(fe(value))
```
kailuowang: this and the `leftMapF` feels weird to me. The type signature doesn't prevent the user from mapping on the other side.

core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -88,11 +88,22 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
...
+
+  def recoverF[E](pf: PartialFunction[E, Either[A, B]])(implicit ae: ApplicativeError[F, E]): EitherT[F, A, B] =
```
kailuowang: This also feels weird to me because you can "recover" to the left side, which is a different semantics from the original `recover`. I suspect that these two different sets of `recover`s and `recoverWith`s should probably best be organized into two different derivations of `MonadError` instances. That way the code organization might be clearer and more importantly, we might be able to test with `MonadError` laws.

leandrob13: @kailuowang that is exactly what I intended, maybe the names should be changed so the semantics are right? The problem is that if you have an `val x = EitherT(Future.failed(new Exception("Error")))` you can recover it with `x.recoverFWith[Throwable](ex => Future.successfull(Either.left(ex.getMessage)))` or if you want with a Right too. I didn't add a MonadError implicitly because Validated doesn't have one.

What do you mean by organizing them into two different derivations of MonadError instances?

As I discussed in #1643, there is a limitation on recovering for monad transformers like EitherT because you can't recover from F (my example is with a Future). My intention is to keep it practical and since EitherT has its own implementation of recover and recoverWith which are used for overriding the corresponding function of its MonadError instace, I thought we could add these functions without including them in MonadError or some other typeclass.

core/src/main/scala/cats/data/EitherT.scala
leandrob13: @kailuowang I recognize your point. It is mapping F to N. Whatever is inside, could be a Left or a Right (mapped or not) and it is still mapping the Either. If I left the F[_] to N[_] without mapping the Either inside it would be a transform (F[_] ~> N[_]) but in EitherT transform is like 

```scala
def transform[C, D](f: Either[A, B] => Either[C, D])(implicit F: Functor[F]): EitherT[F, C, D] =
    EitherT(F.map(value)(f))
```
Should it be a transformF? 

```scala
def transformF[N[_], C, D](fe: F[Either[A, B]] => N[Either[C, D]]): EitherT[F, A, D] = 
  EitherT(fe(value))
```

Should we make standard what to do in a mapF? In Kleisli it looks like this:
```scala
def mapF[N[_], C](f: F[B] => N[C]): Kleisli[N, A, C] =
    Kleisli(run andThen f)
``` 
That is why I thought I could add a mapping function from F[_] to N[_] in EitherT. Also, transform in Kleisli means a transformation from F to N and also maps what is inside of F. Should we fix those semantics in the monad transformers?

kailuowang: This transform sounds like we might need a new higher order type class? like 
```scala
trait HFunctor[F[_[_], _] {
  def hmap[A[_], B[_], T](fa: F[A, T])(fk: A ~> B): F[B, T]
}
```
There is some discussion in this [gist](https://gist.github.com/kailuowang/ddf4f99a3bd97f150277a0c614e0a6f2). It maybe too theoretical, but I think we can see its practial usage here.

core/src/main/scala/cats/data/EitherT.scala
kailuowang: So as @djspiewak pointed out in #1643, there are two ways to provide a `MonadError` instance for a `EitherT`, the existing one ignores the error in `F` and recovers from the left side of the Either. We can have a new `MonadError` instance if the`F` already has `MonadError` instance and delegate to that, which is basically what you are doing with `recoverF`. These two instances might be able to coexist since the Error type is different, which should be true as long as the E in `F: MonadError` and the `EitherT`'s left type are different, right?

leandrob13: @kailuowang right! Let me work it out then, thanks!

core/src/main/scala/cats/data/EitherT.scala
leandrob13: @kailuowang I will look into that!

leandrob13: @kailuowang taken your advice I opened #1713 and came to the conclusion to add this to EitherT

```scala
def transformF[G[_]](f: FunctionK[F, G]): EitherT[G, A, B] = EitherT(f(value))
```

There is something confusing about EitherT and OptionT using `transform` as a bimap/map of the parametrized types respectively. I called it transformF just to avoid the confusion. The issue is opened for the evaluation of your proposal and to address the confusion around `transform` semantics.

leandrob13:
@djspiewak @kailuowang  just to validate, I ended up defining the new MonadError instance for EitherT as:

```scala
implicit def catsDataMonadErrorFForEitherT[F[_], E, L](implicit FE0: MonadError[F, E]): MonadError[EitherT[F, L, ?], E] =
    new EitherTMonadErrorF[F, E, L] { implicit val F = FE0 }

private[data] trait EitherTMonadErrorF[F[_], E, L] extends MonadError[EitherT[F, L, ?], E] with EitherTMonad[F, L] {
  implicit val F: MonadError[F, E]

  def handleErrorWith[A](fea: EitherT[F, L, A])(f: E => EitherT[F, L, A]): EitherT[F, L, A] =
    EitherT(F.handleErrorWith(fea.value)(f(_).value))

  override def handleError[A](fea: EitherT[F, L, A])(f: E => A): EitherT[F, L, A] =
    EitherT(F.handleError(fea.value)(f andThen(Right(_))))

  def raiseError[A](e: E): EitherT[F, L, A] = EitherT.left(F.raiseError(e)) // I am not so sure about this one

  override def recover[A](fla: EitherT[F, L, A])(pf: PartialFunction[E, A]): EitherT[F, L, A] =
    fla.recoverF(pf)

  override def recoverWith[A](fla: EitherT[F, L, A])(pf: PartialFunction[E, EitherT[F, L, A]]): EitherT[F, L, A] =
    fla.recoverFWith(pf.andThen(_.value))
}
```

kailuowang:
@leandrob13 I was thinking you don't need to override methods, and thus you don't need `recoverF`. 
```scala
private[data] trait EitherTMonadErrorF[F[_], E, L] extends MonadError[EitherT[F, L, ?], E] with EitherTMonad[F, L] {
  implicit val F: MonadError[F, E]

  def handleErrorWith[A](fea: EitherT[F, L, A])(f: E => EitherT[F, L, A]): EitherT[F, L, A] =
    EitherT(F.handleErrorWith(fea.value)(f(_).value))

  def raiseError[A](e: E): EitherT[F, L, A] = EitherT(F.raiseError(e)) //shouldn't be left or right here I think, `F.raiseError(e)` should return a `F[Either[L, A]]`.   
}
```

tests/src/test/scala/cats/tests/EitherTTests.scala
```diff
@@ -46,16 +48,22 @@ class EitherTTests extends CatsSuite {
...
+    implicit val me = EitherT.catsDataMonadErrorFForEitherT[Option, Unit, String](option.catsStdInstancesForOption)
 
     Functor[EitherT[ListWrapper, String, ?]]
     Applicative[EitherT[ListWrapper, String, ?]]
     Monad[EitherT[ListWrapper, String, ?]]
     MonadTrans[EitherT[?[_], String, ?]]
 
     checkAll("EitherT[ListWrapper, String, Int]", MonadErrorTests[EitherT[ListWrapper, String, ?], String].monadError[Int, Int, Int])
+    checkAll("EitherT[Option, String, Int]", MonadErrorTests[EitherT[Option, String, ?], Unit].monadError[Int, Int, Int])
     checkAll("MonadError[EitherT[List, ?, ?]]", SerializableTests.serializable(MonadError[EitherT[ListWrapper, String, ?], String]))
```
kailuowang: we need to law tests the new instance.

leandrob13: I really did that by copying and it seems to give the coverage to the new MonadError instance. If it is not this way, can you please give me a hint?

leandrob13: @kailuowang I still think that it is useful to have the recoverF/With functions but if it is a blocker for the PR to not be approved then I could remove them and keep the new MonadError instance and the transformF function.

kailuowang: I apologize, @leandrob13, I meant serializable test the instance. Not sure what I was thinking.

leandrob13: @kailuowang no problem, got it!

core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -91,6 +92,14 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
 
   def map[D](f: B => D)(implicit F: Functor[F]): EitherT[F, A, D] = bimap(identity, f)
 
+  def transformF[G[_]](fe: FunctionK[F, G]): EitherT[G, A, B] = EitherT(fe(value))
```
LukaJCB: In accordance with the corresponding method on `Kleisli` or `ReaderT` being called only `transform`, maybe we should name it this way also? See #1724

kailuowang:
I am thinking that the MonadError instance change is straightforward, the transform thing seems a more complex problem. How about we separate the transform part out of this PR?

leandrob13:
@LukaJCB @kailuowang agreed. I didn't call it transform because there is one already defined for EitherT, so I agree to tackle this issue in another PR so we can agree on the standard semantics for transform along MonadT's.

leandrob13:
@kailuowang BTW, the build is only failing for scala 2.10 and I don't know how to fix it.

LukaJCB:
Looks like you have diverging implicits. Don't have time to look in detail right now, but can you build the tests with the `-Xlog-implicits` flag?

leandrob13:
@LukaJCB sure thing, thanks!

leandrob13:
@LukaJCB I solved the error. I wrongly defined the Eq instance for the MonadError EitherT test.

tests/src/test/scala/cats/tests/EitherTTests.scala
```diff
@@ -339,6 +362,17 @@ class EitherTTests extends CatsSuite {
     }
   }
 
+  test("collectRight with Id consistent with Either foldRight") {
+    val et: EitherT[Option, String, Int] = EitherT.pure(1)
+    et.collectRight should === (Some(1))
```
LukaJCB: Shouls this use `Id` instead? Or should you rename the test?

leandrob13: @LukaJCB I am changing the name and the test. I forgot to fix this!

LukaJCB:
LGTM! :)

tests/src/test/scala/cats/tests/EitherTTests.scala
```diff
@@ -46,17 +47,24 @@ class EitherTTests extends CatsSuite {
 
   {
     //if a Monad is defined
+    import cats.instances.option
```
kailuowang: I wonder why this line is needed. I think `CatsSuite` should provide all the instance. What happens if you remove this line?

kailuowang:
@leandrob13 regarding `recoverF` and `recoverFWith`, I still don't quite understand we needed them. These two methods are already implemented by the new MonadError instance you added, aren't they. Calling `(et: EitherT[F, L, R]).recoverF[E]` should be the same as calling `(et: EitherT[F, L, R]).recover[E]` right? Seems to me that reimplementing them is unnecessary and bears the risk of inconsistent implementation. 
One incentive to provide these alternatives could be when `E` and `L` is the same (an example I can think of is `EitherT[Future, Throwable, ?]`).  In that case the monadError instance you get out of box is the one operating on the `Either`. But if that's the reasoning, why we only provide a `recoverF` and `recoverFWith`, what about all other methods on `MonadError`? It seems to me that the right thing to do in this situation is to explicitly choose the lower priority `MonadError` instance (that operates on `Future`) to use.

leandrob13:
@kailuowang I finally understood your point. I will change again the description of the PR and make the changes.

leandrob13:
@kailuowang This `(et: EitherT[F, L, R]).recover[E]` is with the help of the applicative error syntax, right? What i actually intended is for it to be straight forward in the DataType. And yes, there can be that problem with `EitherT[Future, Throwable, ?] ` but this recoverF and recoverFWith functions are property of the EitherT. There is no complete implementation of the MonadError functions natively in any data type that has an instance of it (like future or option). What I intended was to have it natively as part of the features of the EitherT since it was a necessity detected from practice. 

I see the need of the extra MonadError instance but I think it will be useful. Another example is that the recover and recoverWith functions on the MonadError instance that recovers from the Either are overriden for the original implementation of EitherT's recover and recoverWith, that is why I thought I could do the same.

What do you think? I just wish this things could be done more straight forward in the data types instead of having a syntax imported for something that should be native. I understand in cases like option and either that are not part of cats, but for those that we have control we can provide that straightforwardness. My argument was similar for Kleisli recovery too, given that we had to define an alias for the Kleisli type in order to get the MonadError instance to work.

I can leave the PR only about the addition of the new ME instance but I hope you can consider these arguments.

kailuowang:
@leandrob13  IMO, in a canonical FP way, when you use an `EitherT`, you shouldn't even need to know this concrete type, instead, all you need to know is that it is `MonadError`. 
For example, say you have method that takes an `EitherT`, 
```scala
def f[F[_], A, B](e: EitherT[F, A, B]):  XXX
```
If you don't care about the left side, then you don't have to take an EitherT right? You could just do this, making the type more generic (and thus parametric, explained in detail [here](https://www.youtube.com/watch?v=D0Fnzr15BAU))
```scala
def f[F[_]: Monad, T](e: F[T]): XXX
```
Now the only reason you care about the left side is that you want to handle the error, in which case, you can do 
```scala
def f[F[_], E, T](e: F[T])(implicit F: MonadError[F, E]): XXX
```
This signature is telling everyone that there is only one way to handle the error from `F`, that is the `MonadError` instance. They don't need to learn what's native on `F`. More importantly, there is only one possible error sencrio you need to handle. If we have two sets of methods of error handling on `F` (one native, one provided by the `MonadError`), then it will be significanlty more complex to deal with.  Are you going to handle the error on the `EitherT` or the `F` or both? 
In my projects, whenever I was given a `Future[Either[E, T]` (usually by a 3rd party library), I converts it into an `EitherT[Future, E, T]` with something like the following
```scala
  def throwableToE(t: Throwable) : E = ???
   EitherT(f.recover {
        case e: Throwable =>throwableToE(e).asLeft[T]
   })
```
Then from this point on, if we stick with the never-throw-exception FP paradigm, we don't have to worry about the `Future` errors. 
Maybe what we should also add is a factory method on the EitherT companion method (obviously doesn't have to be in this PR):
```scala
private[data] final class FromMonadErrorPartiallyApplied[E](val dummy: Boolean = true) extends AnyVal {
    def apply[F[_], A, B](fa: F[Either[A, B]])(recover : E => A)(implicit F: MonadError[F, E]): EitherT[F, A, B] =
     EitherT(fa.handleError(fa)(e => recover(e).asLeft[B] ))
}
def fromMonadError[E] : FromMonadErrorPartiallyApplied[E] 
```

leandrob13:
@kailuowang understood and I like the last idea. I will delete those and leave only the new instance. Thanks for taking the time to explain.

core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -483,9 +483,13 @@ private[data] abstract class EitherTInstances1 extends EitherTInstances2 {
     new EitherTBitraverse[F] {
       val F0: Traverse[F] = F
     }
+
+  implicit def catsDataMonadErrorFForEitherT[F[_], E, L](implicit FE0: MonadError[F, E]): MonadError[EitherT[F, L, ?], E] =
```
kailuowang: nitpick: would you add a comment doc here to indicate that this instance delegate to the `MonadError` of `F` instead of using `Either`?

tests/src/test/scala/cats/tests/EitherTTests.scala
```diff
@@ -46,17 +47,23 @@ class EitherTTests extends CatsSuite {
...
+    implicit val eq3 = EitherT.catsDataEqForEitherT[EitherT[Option, String, ?], Unit, String](eq2)
+    implicit val me = EitherT.catsDataMonadErrorFForEitherT[Option, Unit, String](catsStdInstancesForOption)
 
     Functor[EitherT[ListWrapper, String, ?]]
     Applicative[EitherT[ListWrapper, String, ?]]
     Monad[EitherT[ListWrapper, String, ?]]
     MonadTrans[EitherT[?[_], String, ?]]
 
     checkAll("EitherT[ListWrapper, String, Int]", MonadErrorTests[EitherT[ListWrapper, String, ?], String].monadError[Int, Int, Int])
+    checkAll("EitherT[Option, String, String]", MonadErrorTests[EitherT[Option, String, ?], Unit].monadError[String, String, String])
```
kailuowang: another nitpick, would you add a comment indicating that this is testing the `catsDataMonadErrorFForEitherT` and the above one is testing the `catsDataMonadErrorForEitherT`?

core/src/main/scala/cats/data/EitherT.scala
leandrob13: @kailuowang sure, I'm on it.

kailuowang:
Thanks so much for this valuable contribution and being so patient with me.

leandrob13:
@kailuowang thanks! No problem, I am learning a lot with this conversations.

tests/src/test/scala/cats/tests/EitherTTests.scala
```diff
@@ -46,17 +47,25 @@ class EitherTTests extends CatsSuite {
...
     Applicative[EitherT[ListWrapper, String, ?]]
     Monad[EitherT[ListWrapper, String, ?]]
     MonadTrans[EitherT[?[_], String, ?]]
-
+    // Tests for catsDataMonadErrorForEitherT, for recovery on errors on Either.
     checkAll("EitherT[ListWrapper, String, Int]", MonadErrorTests[EitherT[ListWrapper, String, ?], String].monadError[Int, Int, Int])
     checkAll("MonadError[EitherT[List, ?, ?]]", SerializableTests.serializable(MonadError[EitherT[ListWrapper, String, ?], String]))
+    // Tests for catsDataMonadErrorFForEitherT instance, for recovery on errors of F.
+    checkAll("EitherT[Option, String, String]", MonadErrorTests[EitherT[Option, String, ?], Unit].monadError[String, String, String])
+    checkAll("MonadError[EitherT[Option, ?, ?]]", SerializableTests.serializable(MonadError[EitherT[Option, String, ?], Unit]))
```
peterneyens: I think we should separate these `MonadError` tests using `Option` in a separate section.

```scala
{
  // If a MonadError is defined for F
  val eq = ...
  
  checkAll("EitherT[Option, String, String]", ...)
}
```

leandrob13: @peterneyens no problem, I can do that.

leandrob13:
@peterneyens completed your request.

kailuowang:
@peterneyens  got some time to re-review this one?

tests/src/test/scala/cats/tests/EitherTTests.scala
```diff
@@ -54,15 +56,32 @@ class EitherTTests extends CatsSuite {
...
+    // Tests for catsDataMonadErrorForEitherT, for recovery on errors on Either.
     checkAll("EitherT[ListWrapper, String, Int]", MonadErrorTests[EitherT[ListWrapper, String, ?], String].monadError[Int, Int, Int])
     checkAll("MonadError[EitherT[List, ?, ?]]", SerializableTests.serializable(MonadError[EitherT[ListWrapper, String, ?], String]))
+    // Tests for MonadTrans instance.
     checkAll("EitherT[ListWrapper, String, Int]]", MonadTransTests[EitherT[?[_], String, ?]].monadTrans[ListWrapper, Int, Int])
     checkAll("MonadTrans[EitherT[?[_], String, ?]]", SerializableTests.serializable(MonadTrans[EitherT[?[_], String, ?]]))
   }
 
   {
     //if a Monad is defined
```
kailuowang: just noticed here, should it be `if a MonadError is defined"?

tests/src/test/scala/cats/tests/EitherTTests.scala
```diff
@@ -54,15 +56,32 @@ class EitherTTests extends CatsSuite {
...
+    // Tests for catsDataMonadErrorFForEitherT instance, for recovery on errors of F.
```
peterneyens: Minor, but should we say "if a MonadError is defined"?

leandrob13:
@peterneyens @kailuowang done with the comment fix. Thanks guys.

kailuowang:
merge with 3 sign-offs

PR #1651, opened 2017-05-06
Add apply method to MonadState

oskoi:


oskoi:
@edmundnoble 
a) You're right. Changed the name of method.
b) Sorry, but I did not understand you. We have method `raiseError` in MonadError. Explain, please, what did you mean

edmundnoble:
To clarify and correct myself, in `ApplicativeError`, a method `def either[A](x: E Either A): F[A]`, which just calls `_.fold(raiseError, pure)`.

oskoi:
@edmundnoble done.

core/src/main/scala/cats/MonadState.scala
```diff
@@ -17,6 +17,9 @@ package cats
  * }}}
  */
 trait MonadState[F[_], S] extends Monad[F] {
+  def state[A](f: S => (S, A)): F[A] =
+    flatMap(get)(s => f(s) match { case (s, a) => inspect(_ => a)})
```
peterneyens: Shouldn't this use `set` to set the `S` resulting from `f(s)`? As is this will just return the initial `S`, right?

oskoi: @peterneyens Right. Forgot to update the state ( Fixed this!

kailuowang:
We might want to add some doctests.

oskoi:
@kailuowang done

PR #1652, opened 2017-05-08
Fix typo in FreeApplicative doc.

fan-t-sakai:
It seems like we use `validator` as a variable name for a similar value in the previous chapter.
And beside of that, this value is made from `ParValidator` type.


fan-t-sakai:
I'm sorry I made this Pull Request by a wrong user.
I'll make an exact same one soon.

PR #1658, opened 2017-05-10
Add Validated.validNel

edmundnoble:
Present as a syntax enrichment already.


ceedubs:
Thanks @edmundnoble. 👍 to this change, but it looks like there are merge conflicts. Also, while this change is trivial, we should probably hit it from a code coverage perspective. Maybe just an sbt-doctest example of using it?

edmundnoble:
Oof. What *is* in my `master` branch? I'll add a doctest and fix the conflicts.

edmundnoble:
@ceedubs Should be good now, after CI.

Issue #1657, opened 2017-05-10
Kernel instances are in typeclasses which are subclasses


yilinwei:
~~Both `Order` and `PartialOrder` need a `Contravariant` instance.~~

Will submit a PR for it shortly.

EDIT: #1659 has proper explanation of the issue.

ceedubs:
@yilinwei I think that those exist here: https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/functor/Contravariant.scala#L42-L58

peterneyens:
Fixed in #1659.

PR #1659, opened 2017-05-11
move instances into separate trait

yilinwei:
#1657 

The current instances sit in the object of `Contravariant`.

This means that you can't get an implicit of a subclass - `implicitly[Invariant[Order]]` doesn't work because the instance is not in the companion object of `Order` or `Invariant` and the methods such as `imap` won't work either.

I would argue that for the behaviour to be consistent with the other instances, these should be moved into separate traits.

Doing a quick grep for `Kernel...Instances`

- [x] Cartesian
- [x] InvariantMonoidal
- [x] ContravariantCartesian
- [x] Contravariant
- [x] Invariant

are affected.


ceedubs:
@yilinwei thanks! I think that you raise a good point about `implicitly[Invariant[Order]]` not working.

While `import cats.implicits._` is the approach that we recommend to people, we also support piecemeal imports such as `cats.instances.string._`. I wonder if we should provide something similar for these instances? If so I'm not quite sure how one should do that. `cats.instances.kernel.order._` or something? It seems kind of awkward. And it makes me wonder if instead of `trait EqInstances` you should have `trait KernelEqInstances` or something.

edmundnoble:
I like `cats.instances.order` and the approach in this PR. We provide instances originating in kernel from `instances._` already with mixins (see `instances.bitSet`); I think that fits well with providing instances for types provided from kernel.

yilinwei:
@ceedubs 

I forgot about the piecemeal imports; I think we should definitely include them. I also agree with @edmundnoble that they should sit under `instances` rather than `instances.kernel`.

ceedubs:
Thanks for adding the piecemeal imports @yilinwei. This looks good to me.

It's a little bit of a head-scratcher to think about the fact that `MonoidInstances` and `import cats.instances.monoid._` don't actually include any `Monoid[A]` instances. Instead they include functor instances _for_ monoids :). However, I do think that this setup is consistent.

codecov is legitimately reporting a slight code coverage drop on this PR. While it's pretty trivial, it may be good to hit the new piecemeal imports in unit tests to make sure that they work as intended. I also think that it would be good to have the unit tests check some implicit resolutions for things like the `Invariant[Order]` that was the motivation for this PR.

yilinwei:
@ceedubs I feel those tests ought to be autogenerated, ideally something along the lines of,

`canResolve[F, TC] {
  import ...
}`

should reify into `implicitly[TC1]; implicitly[TC2] ....` for the whole of the hierarchy.

wdyt?

yilinwei:
@ceedubs 

I've committed an example of what I mean;

The macro generates an `implicitly` for each typeclass in the hierarchy with the import and any implicit vals in scope.
 
There are two problems at the moment though;

1. I can't quite work out how to do test->test dependencies since I'm not too familiar with the javascript build
2. More importantly, scalac flags the import as unused when it's definitely being used. I've disabled the flag for now; but I'd really like to be able to turn it back on if anyone has any ideas.

peterneyens:
I like the idea of testing the resolution of all the type class instances in the hierarchy.
It would be nice to replace code like [this](https://github.com/typelevel/cats/blob/master/tests/src/test/scala/cats/tests/WriterTTests.scala#L315-L324).

yilinwei:
I've simplified the macro to just do the `implictly` expansion because scalac was doing funny things with the imports.

The names are `implicitlyH` which stands for `implicitlyHierarchy` and `implicitlyHK` which stands for `implicitlyHierarchyKind`. I'm not particularly happy with the name because `H` usually stands for heterogenous  and could be confusing.

I still need to deal with project structures, but the scalac flag problem should go away.

@peterneyens I've added an example for the `WriterT` to show it works correctly.

kailuowang:
My original concern is that we might have incentive to go the other direction, i.e. move these instance into the companion objects for maintainability of binary compatibility. https://github.com/typelevel/cats/issues/1683
After some inspection of the instances being moved in this PR. It doesnt seem to me that there is a good chance of changing these traits. So 👍  from me. As for automatically generated tests, can we attack them in a separate PR?

yilinwei:
@kailuowang Sure; I was more hoping for feedback on whether people like this approach before taking it any further.

Regardless I'll separate the changes from the original ones tomorrow with the additional tests suggested by @ceedubs.

djspiewak:
👍 from me.  So this is good for merge once @yilinwei is finished with it.

kailuowang:
👍  modular build green

tests/src/test/scala/cats/tests/EqTests.scala
```diff
@@ -0,0 +1,24 @@
...
+package tests
+
+import org.scalatest._
+
+import cats.functor._
+
+class EqTests extends FunSuite {
+  {
+    import cats.implicits._
+    implicitly[Invariant[Eq]]
```
peterneyens: Just nitpicking, but we don't need `implicitly` in all these tests as simulacrum generates an `apply` in the companion object. Should we just write `Invariant[Eq]`, `Invariant[Monoid]`, ...?

core/src/main/scala/cats/instances/monoid.scala
```diff
@@ -0,0 +1,26 @@
...
+    def product[A, B](fa: Monoid[A], fb: Monoid[B]): Monoid[(A, B)] = new Monoid[(A, B)] {
+      val empty = fa.empty -> fb.empty
+      def combine(x: (A, B), y: (A, B)): (A, B) = fa.combine(x._1, y._1) -> fb.combine(x._2, y._2)
+    }
+
+    def imap[A, B](fa: Monoid[A])(f: A => B)(g: B => A): Monoid[B] = new Monoid[B] {
+      val empty = f(fa.empty)
+      def combine(x: B, y: B): B = f(fa.combine(g(x), g(y)))
+      override def combineAll(bs: TraversableOnce[B]): B =
+        f(fa.combineAll(bs.map(g)))
```
kailuowang: This line was not covered in the original code, which is a bit surprising to me. We can probably address it in a different PR. Just want to keep a note here.

core/src/main/scala/cats/instances/semigroup.scala
```diff
@@ -0,0 +1,24 @@
...
+
+  implicit val catsInvariantMonoidalSemigroup: InvariantMonoidal[Semigroup] = new InvariantMonoidal[Semigroup] {
+    def product[A, B](fa: Semigroup[A], fb: Semigroup[B]): Semigroup[(A, B)] = new Semigroup[(A, B)] {
+      def combine(x: (A, B), y: (A, B)): (A, B) = fa.combine(x._1, y._1) -> fb.combine(x._2, y._2)
+    }
+
+    def imap[A, B](fa: Semigroup[A])(f: A => B)(g: B => A): Semigroup[B] = new Semigroup[B] {
+      def combine(x: B, y: B): B = f(fa.combine(g(x), g(y)))
+      override def combineAllOption(bs: TraversableOnce[B]): Option[B] =
+        fa.combineAllOption(bs.map(g)).map(f)
```
kailuowang: This line was not covered in the original code, which is a bit surprising to me. We can probably address it in a different PR. Just want to keep a note here.

core/src/main/scala/cats/instances/semigroup.scala
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    def pure[A](a: A): Semigroup[A] = new Semigroup[A] {
+      def combine(x: A, y: A): A = a
+      override def combineAllOption(as: TraversableOnce[A]): Option[A] =
+        if (as.isEmpty) None else Some(a)
```
kailuowang: This line was not covered in the original code. We can probably address it in a different PR. Just want to keep a note here.

tests/src/test/scala/cats/tests/EqTests.scala
yilinwei: changed

peterneyens:
Merging with three thumbs up. Thanks @yilinwei!

Issue #1661, opened 2017-05-11
Group[Eval[A]] is not lawful


djspiewak:
Specifically, the inverse laws do not hold, for the same reason that they would not hold for `Group[IO[A]]`.  The problem is that `Eval` can (now) encode failure, meaning that it has values which cannot be eliminated.  Simple counter-example:

```scala
val zero = Group[Eval[Int]].empty
val example = Eval.raise(new Exception)

example.inverse |+| example === zero   // nope!
```

In the above, `example.inverse |+| example` will be equal to `Eval.raise(new Exception)`, which is clearly incorrect.

Basically, `Group` and `ApplicativeError` are mutually exclusive.  You can't have both of them at once.  If you want `Eval` to implement `MonadError`, then it can be at most a `Monoid`, but not a `Group`.

**Edit** See below.  `Group` is the least of the problems.

djspiewak:
As a sidebar, and perhaps worthy of another issue, the reason this unlawfulness is not caught by the current tests is [the generators](https://github.com/typelevel/cats/blob/master/laws/src/main/scala/cats/laws/discipline/Arbitrary.scala#L95-L99) do not include `Eval.raise`, meaning that we're just not checking the full space of `Eval` instances.  This should *also* be addressed, and will force a resolution to the OP.

ceedubs:
Good catch @djspiewak. I don't think that we should have an `ApplicativeError` instance for `Eval` (and probably not even a `raise` method) without including `Eval.raise` instances in `Eval` generators. I wouldn't be surprised if doing so would identify more instances that aren't lawful, and that would leave us with a choice as to whether we want to drop those instances or drop the error-capturing component of `Eval`. Personally I'm not sold on the error-capturing behavior (especially now with the existence of cats-effect), but I've voiced that concern before, so I won't hammer on it :)

djspiewak:
I'd be fine with or without the error capturing behavior, though `cats.effect.Sync[Eval]` will not be lawful if it doesn't capture errors.  We just can't have that behavior *and* have `Group[Eval[A]]`, so it's a tradeoff that needs to be weighed.

edmundnoble:
For the record I'm also in favor of removing the error capturing behavior, especially seeing how it would interact with the stack-usage optimization I have; sacrificing individual stack frames becomes sacrificing `n` stack frames. The idea that an `Eval` I'm holding might carry an `Exception` makes it much harder from a library perspective to use it. 

[...]

PR #1663, opened 2017-05-12
Removed MonadError[Eval, Throwable]

djspiewak:
Fixes a couple outstanding issues.  Also obsoletes #1634.  See #1661 for an in-depth discussion of why this is necessary.


adelbertc:
👍

kailuowang:
Shall we address @alexandru's [objection here](https://github.com/typelevel/cats/issues/1661#issuecomment-301039355) first?

ceedubs:
I'm in favor of this change. However, I think that we should give a little more time for discussion on #1661. I propose that we allow a week for discussion and go with what the general consensus seems to be after that amount of time.

I also think it would be great if @non could weigh in, since I think that he originally added this code.

tests/src/test/scala/cats/tests/EvalTests.scala
```diff
@@ -83,60 +82,13 @@ class EvalTests extends CatsSuite {
...
 
-    {
-      // we need exceptions which occur during .value calls to be
-      // equal to each other (equivalent behavior).
-      implicit def eqWithTry[A: Eq]: Eq[Eval[A]] =
-        Eq[Try[A]].on((e: Eval[A]) => Try(e.value))
-
-      checkAll("Eval[Int]", MonadErrorTests[Eval, Throwable].monadError[Int, Int, Int])
-    }
+    checkAll("Eval[Int]", MonadTests[Eval].monad[Int, Int, Int])
```
ceedubs: Aren't these `MonadTests` just a subset of the `BimonadTests` that are happening above?

Also it seems a little off to me that the `SerializableTests` for the `Bimonad` instance are below in a separate scope. This was the case before your PR I think, but I don't know if there's a good reason for it.

kailuowang:
quick note, the conflict is due to the merge of #1634.

djspiewak:
@kailuowang Apparently I had already fixed the conflict, but forgot to push.  Pushed now!

kailuowang:
Thanks @djspiewak I think we've allowed 10 days of time for people to weigh in. Looks like that the only thing holding is this [comment](https://github.com/typelevel/cats/pull/1663#discussion_r117142358) by @ceedubs ?

tests/src/test/scala/cats/tests/EvalTests.scala
djspiewak: You're correct that `MonadTests` should just be a subset of `BimonadTests`.  Similar to the `SerializableTests`, I'm sort of puzzled as to why this was the case before.  Mostly I was just carrying it along.  If we're all in agreement that it looks fishy though, I'll just zap this line.

kailuowang: I 👍  especially if it's confirmed that we don't have a coverage drop after.

ceedubs:
There seems to be general (though unfortunately not quite unanimous) consensus on this in #1661, after allowing 10 days or so for discussion. Merging with 4 approvals.

PR #1666, opened 2017-05-13
Use Cogen for arbitrary instances

### LINKED ISSUE #1605: Use Cogen for function-like type Arbitrary instances

ceedubs:
Resolves #1605.

This is a replacement for #1619. The approach taken in that PR led to
implicit resolution failures in scala.js.

### DIFF
--- a/laws/src/main/scala/cats/laws/discipline/Arbitrary.scala
+++ b/laws/src/main/scala/cats/laws/discipline/Arbitrary.scala
@@ -76,9 +76,8 @@ object arbitrary extends ArbitraryInstances0 {
       B.perturb(seed, _),
       (a, b) => A.perturb(B.perturb(seed, b), a)))
 
-
-  implicit def catsLawsArbitraryForCokleisli[F[_], A, B](implicit B: Arbitrary[B]): Arbitrary[Cokleisli[F, A, B]] =
-    Arbitrary(B.arbitrary.map(b => Cokleisli[F, A, B](_ => b)))
+  implicit def catsLawsArbitraryForCokleisli[F[_], A, B](implicit AFA: Arbitrary[F[A]], CFA: Cogen[F[A]], B: Arbitrary[B]): Arbitrary[Cokleisli[F, A, B]] =
+    Arbitrary(Arbitrary.arbitrary[F[A] => B].map(Cokleisli(_)))
 
   implicit def catsLawsArbitraryForOptionT[F[_], A](implicit F: Arbitrary[F[Option[A]]]): Arbitrary[OptionT[F, A]] =
     Arbitrary(F.arbitrary.map(OptionT.apply))
@@ -104,11 +103,11 @@ object arbitrary extends ArbitraryInstances0 {
   implicit def catsLawsArbitraryForTuple2K[F[_], G[_], A](implicit F: Arbitrary[F[A]], G: Arbitrary[G[A]]): Arbitrary[Tuple2K[F, G, A]] =
     Arbitrary(F.arbitrary.flatMap(fa => G.arbitrary.map(ga => Tuple2K[F, G, A](fa, ga))))
 
-  implicit def catsLawsArbitraryForFunc[F[_], A, B](implicit F: Arbitrary[F[B]]): Arbitrary[Func[F, A, B]] =
-    Arbitrary(F.arbitrary.map(fb => Func.func[F, A, B](_ => fb)))
+  implicit def catsLawsArbitraryForFunc[F[_], A, B](implicit AA: Arbitrary[A], CA: Cogen[A], F: Arbitrary[F[B]]): Arbitrary[Func[F, A, B]] =
+    Arbitrary(Arbitrary.arbitrary[A => F[B]].map(Func.func))
 
-  implicit def catsLawsArbitraryForAppFunc[F[_], A, B](implicit F: Arbitrary[F[B]], FF: Applicative[F]): Arbitrary[AppFunc[F, A, B]] =
-    Arbitrary(F.arbitrary.map(fb => Func.appFunc[F, A, B](_ => fb)))
+  implicit def catsLawsArbitraryForAppFunc[F[_], A, B](implicit AA: Arbitrary[A], CA: Cogen[A], F: Arbitrary[F[B]], FF: Applicative[F]): Arbitrary[AppFunc[F, A, B]] =
+    Arbitrary(Arbitrary.arbitrary[A => F[B]].map(Func.appFunc(_)))
 
   implicit def catsLawsArbitraryForWriter[L:Arbitrary, V:Arbitrary]: Arbitrary[Writer[L, V]] =
     catsLawsArbitraryForWriterT[Id, L, V]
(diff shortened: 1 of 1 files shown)


ceedubs:
@edmundnoble that surprised me too.

kailuowang:
merge with two sign-off and no objections.

kailuowang:
after merging build fails on 
> - Kleisli[Option, Int, String].monoid.isEmpty *** FAILED ***
[info]     Falsified after 1 successful property evaluations.
[info]       arg0 = Kleisli(<function1>)
[info]     )
[info]       (true ?== false) failed

I'm investigating.

ceedubs:
@kailuowang I have been looking into this too. So far I've only seen it failing in the JS build and scala versions less than 2.12.

ceedubs:
@kailuowang one other thing that I noticed is that the `isId` method that is called in the `isEmpty` tests may be a little troublesome conceptually when it comes to `Kleisli`/function instances. It will generate random `Kleisli`s and ensure that we get the same result from calling `isEmpty` on that `Kleisli` instance and using the `Kleisli` `Eq` instance to compare the generated `Kleisli` with the result of calling `empty` on the `Kleisli` monoid. I think that it's possible (though I suspect extremely unlikely) that we could get inconsistent results on those two calls based on the 50 random values that we happen to pass into the kleisli instances when checking for "equality".

ceedubs:
Could this be some sort of race condition heisenbug in scala.js? I can't seem to recreate it once I put in printlns to try to track down values that it's failing on.

ceedubs:
Sorry for the live stream of my thought process while debugging this.

I just realized that a difference here is that for the JVM we are using 50 values to test function equality and for JS we are only using 5. So I think that the real issue here is just what I brought up with https://github.com/typelevel/cats/pull/1666#issuecomment-301533524 and it's not as unlikely as I thought when I posted that comment since we are only using 5 values in this case.

So essentially the problem here is that with our testing `Eq` instance that we use for `Kleisli`, there's a chance that the `Monoid` instance for `Kleisli` has a probability of returning `true` for `isEmpty`, even if the `Kleisli` doesn't _always_ return an `empty` value. This is an issue with our tests and not the `Kleisli` `Monoid` itself, but I don't know of a particularly good solution to it.

kailuowang:
@ceedubs the law is basically testing `isEmpty` being consistent with `empty`.  In `Monoid` the `isEmpty` is using `Eq[A]`, for Kleisli, it means that it requires an `Eq` instance for `Kleisli`, i.e. an `Eq` for `A => F[B]` , there isn't true `Eq` for Functions, so I think the `isEmpty` method is simply not available for `Monoid[Kleisli[F, A, B]`. We were able to test this because we have a proximate `Eq` for Functions, but I think we shouldn't test it against laws. Thus we should be testing this law against `Kleisli`. WDYT?

ceedubs:
> Thus we should be testing this law against Kleisli. 

@kailuowang sorry I don't think that I followed you there at the end. We _are_ testing this law against `Kleisli` aren't we?

kailuowang:
I meant if simply put if there is no Eq for kleisli, we shouldn't be testing isEmpty which isn't available for kleisli.

ceedubs:
@kailuowang I see. It's currently part of the monoid laws, so you would need to remove it from there to not test it for `Kleisli`. You can't call the monoid laws without an `Eq` instance, so I guess it kind of makes sense that they would test `isEmpty` as well.

Another potential thing that you could do is change the `isEmpty` law to check that `Monoid[A].isEmpty(Monoid[A].empty)` is `true` without passing arbitrary `A` values in and performing the comparison. This would change the semantics a bit in that it would no longer enforce that there's only "one" empty value for a monoid.

kailuowang:
I was just thinking should empty be unique?

edmundnoble:
I think `empty` has to be unique, by contradiction, as @alexknvl stated in the gitter:
`e1 === e1 combine e2 === e2`.

kailuowang:
I agree that `empty` has to be unique by law. 
After further inspection of the tests, it seems to me that the test is basically testing the two separatedly generated instances of `Eq[A => F[B]]` being consistent with each other. This `Eq` in tests is [implemented by passing a number of arbitrary input and comparing the outputs](https://github.com/typelevel/cats/blob/master/laws/src/main/scala/cats/laws/discipline/Eq.scala#L15-L21). 
On scalajs, the number is only 5, which makes it subject to false positive, i.e. two different functions having a chance to return the same result for 5 inputs and thus become a false positive equal. So the test fails when one Eq instance report false equal while the other doesn't. I verified this hypothesis by turning the number to 3, which made the test fail consistently. 
For a short-term solution, I propose we increase the number in scalajs to 10. I just created a PR https://github.com/typelevel/cats/pull/1671 to test how it behave on travis.

edmundnoble:
In the long term, `Eq` for `Function1` is giving us false positives, which translate to false negatives on the left side of an implication. Should we discourage negation of the `Eq` instance for this reason, or remove it altogether? I don't think false negatives are acceptable given property-based testing's already mammoth issues with reproducibility.

ceedubs:
Thanks @kailuowang. I think that bumping up the number of checks for the JS build is a pretty reasonable solution, at least for the short-to-medium- term.

@edmundnoble I agree that this isn't great. Just to make sure that we are on the same page, you know that the `Eq` instance only exists in the test scope, right? The problem with removing it is that then we wouldn't be able to test laws on our `Function1`/`Kleisli`/`State` instances at all, which seems worse to me than longer build times and possible occasional build failures due to false positives. I would love to find a more elegant solution, though.

edmundnoble:
Yeah, I'm aware it's only in test scope. My only issue is with negating the `Eq` causing false positives to translate to false negatives. The issue is with our usage of the `Eq` instance.

ceedubs:
@edmundnoble could you please expand upon what you mean by "negating the Eq"? You've said that a couple of times now, and I don't think that I understand it.

alexknvl:
When you check
```scala
P(f, g) = f === g
```
using `Eq[A => B]`, it's the same as just adding an extra parameter like this:
```scala
P(f, g, a) = f(a) === g(a)
```
At worst you might get some false positives (`f` is not the same as `g`, but on your test set they are the same).

Now consider
```scala
P(f, g) = !(f === g)
```
this will instead produce false negatives (`f` is not the same as `g`, but on your test set they coincide)

kailuowang:
Thanks @alexknvl

ceedubs:
Thanks @alexknvl. @edmundnoble is this something that we are doing in our tests? Also I guess the hope would be that with enough checks over enough rng seeds (a new one each build), false negatives would eventually be uncovered.

kailuowang:
To help facilitate the discussion. Here is the code 
The test is `Rules.isId("isEmpty", A.empty)(A.isEmpty)`
which translate to 
```scala
forAll {( x: A) => Eq[A].eqv(x, A.empty) ?== A.isEmpty(x) }
```
The default implementation of `A.isEmpty` is
```scala
Eq[A].eqv(a, A.empty)
```
So for instances that don't override the default implementation this test is equvilent to 
```scala
forAll { ( x: A) => 
 Eq[A].eqv(x, A.empty) ?== Eq[A].eqv(x, A.empty)
}
```

PR #1669, opened 2017-05-14
Remove implicit params from ops classes (WIP)

edmundnoble:
I missed this during the consistent ops classes PR. *Some* implicit conversions to ops classes take type class instances implicitly still, despite the fact that they don't and can't use them, adding to compile time.

There is a cost I hadn't predicted, however. Type inference does not function as well. `ApplicativeErrorTests` in particular fails to compile, because the `handle` lambdas do not have annotated parameter types so knowing `F[_]` and `A` is not enough for the compiler to know `E`.

Keeping the implicit parameter on the conversion means that with missing instances users get `handle is not a method of F[A]` instead of `cannot find implicit ApplicativeError[F, E]`. This also has the effect that type class ops could conflict less often, but AFAIK we do not have type class ops with duplicate names and never will.

Removing the implicit parameter from the conversion has the effect of users being able to use several `ApplicativeError` (or analogous) instances with syntax, provided that the type parameters on the syntax methods allow the user to disambiguate the instances.

Actually *using* the implicit parameters being passed in the body of the ops class means that ops classes are unable to extend `AnyVal`, because they close over the parameter.

 From my perspective we have three options to fix this:
a) Remove all implicit parameters from non-macro ops class conversions. **This PR implements this**. Makes type inference worse in exchange for the most consistent and readable missing-instance error messages and more flexible instance selection. Also doesn't worsen the inconsistency between macro and non-macro ops classes.
b) For *multi-parameter* type classes, keep the unused implicit parameter to fix type inference, and fix the type parameter of the ops class using it. The (IMO) least consistent and hardest to explain option, has the best type inference and in most cases best missing-instance error messages (except for multi-parameter type classes, where they revert to `is not a method`).
c) For *all* type classes, actually *use* the implicit parameter. The least efficient option, with the worst missing-instance error messages, but it's the most consistent.

Option b) would mean we have *three* flavors of ops class conversions: some that take implicit params and use them, but have ops classes implemented with zero-cost macros; some that are multi-parameter and take implicit params just to fix type parameters and don't use them; and some that take no implicit params at all. This is why I think it's the least consistent and hardest to explain.

I'm looking for feedback, but leaning toward a) and fixing `ApplicativeErrorTests`.

### DIFF
--- a/core/src/main/scala/cats/syntax/applicativeError.scala
+++ b/core/src/main/scala/cats/syntax/applicativeError.scala
@@ -7,8 +7,8 @@ trait ApplicativeErrorSyntax {
   implicit final def catsSyntaxApplicativeErrorId[E](e: E): ApplicativeErrorIdOps[E] =
     new ApplicativeErrorIdOps(e)
 
-  implicit final def catsSyntaxApplicativeError[F[_], E, A](fa: F[A])(implicit F: ApplicativeError[F, E]): ApplicativeErrorOps[F, E, A] =
-    new ApplicativeErrorOps[F, E, A](fa)
+  implicit final def catsSyntaxApplicativeError[F[_], A](fa: F[A]): ApplicativeErrorOps[F, A] =
+    new ApplicativeErrorOps[F, A](fa)
 }
 
 final class ApplicativeErrorIdOps[E](val e: E) extends AnyVal {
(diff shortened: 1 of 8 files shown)


kailuowang:
It seems to me that there might be one downside of this approach. if you have two type classes (or op classes) with the exact same op methods, you get a conflict. 
```scala

scala> trait A[T] { def op: Int }
defined trait A

scala> trait B[T] { def op: Int }
defined trait B

scala> implicit class aOps[T](self: T) { def op(implicit at: A[T]) = at.op } 
defined class aOps

scala> implicit class bOps[T](self: T) { def op(implicit at: B[T]) = at.op } 
defined class bOps

scala> implicit val aString : A[String] = new A[String] {val op = 3}
aString: A[String] = $anon$1@7de843ef

scala> "332".op
<console>:15: error: type mismatch;
 found   : String("332")
 required: ?{def op: ?}
Note that implicit conversions are not applicable because they are ambiguous:
 both method aOps of type [T](self: T)aOps[T]
 and method bOps of type [T](self: T)bOps[T]
 are possible conversion functions from String("332") to ?{def op: ?}
       "332".op
       ^
<console>:15: error: value op is not a member of String
       "332".op

```

edmundnoble:
Agreed @kailuowang, I skimmed over that in the third paragraph. My question in that case is twofold:
- if you have *both* an `A` and a `B` instance for a type, are you not already forbidden from calling `op`?
- do we plan to ever have duplicate-named ops on type classes? Haskell is forbidden from this, so it seems to me unlikely that we would for any ops copied from Haskell.

mpilquist:
`traverse` & `sequence` are defined on both `Applicative` and `Traverse`

edmundnoble:
They are defined in *one* ops class.

kailuowang:
@edmundnoble I definitely see the benefits of making this change. I was thinking about the scenarios in which type class A and B is completely irrelevant, thus unlikely having shared instance data types, they may even be from different libraries.

edmundnoble:
@kailuowang Good point. It's not always avoidable, and we are polluting the namespace in a more harmful way with this change. Do you think that's worth the error messages? Or is b) perhaps a better option?

kailuowang:
@edmundnoble I am leaning towards b). Maybe adding some doc to the guidelines would help too.

PR #1670, opened 2017-05-14
Add Order->Ordering implicit conversion to implicits, instances

edmundnoble:
Does... anyone know where I would put a test? If it's worth testing? Also, kernel is open for changes for 1.0, right?

### DIFF
--- a/kernel/src/main/scala/cats/kernel/Order.scala
+++ b/kernel/src/main/scala/cats/kernel/Order.scala
@@ -139,6 +139,15 @@ abstract class OrderFunctions[O[T] <: Order[T]] extends PartialOrderFunctions[O]
     ev.max(x, y)
 }
 
+trait OrderToOrderingConversion {
+  /**
+   * Implicitly derive a `scala.math.Ordering[A]` from a `Order[A]`
+   * instance.
+   */
+  implicit def catsKernelOrderingForOrder[A](implicit ev: Order[A]): Ordering[A] =
+    ev.toOrdering
+}
+
 object Order extends OrderFunctions[Order] {
   /**
    * Access an implicit `Order[A]`.
--- a/core/src/main/scala/cats/instances/order.scala
+++ b/core/src/main/scala/cats/instances/order.scala
@@ -3,7 +3,7 @@ package instances
 
 import cats.functor.Contravariant
 
-trait OrderInstances {
+trait OrderInstances extends cats.kernel.OrderToOrderingConversion {
 
   implicit val catsFunctorContravariantForOrder: Contravariant[Order] =
     new Contravariant[Order] {
@@ -14,3 +14,4 @@ trait OrderInstances {
       def contramap[A, B](fa: Order[A])(f: B => A): Order[B] = fa.on(f)
     }
 }
+
--- a/tests/src/test/scala/cats/tests/OrderTests.scala
+++ b/tests/src/test/scala/cats/tests/OrderTests.scala
@@ -17,4 +17,18 @@ class OrderTests extends FunSuite {
     Invariant[Order]
     Contravariant[Order]
   }
+
+  // ambiguity test:
+  // the Ordering instance from the Order instance should be trumped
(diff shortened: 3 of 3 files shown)


kernel/src/main/scala/cats/kernel/Order.scala
```diff
@@ -139,7 +139,16 @@ abstract class OrderFunctions[O[T] <: Order[T]] extends PartialOrderFunctions[O]
     ev.max(x, y)
 }
 
-object Order extends OrderFunctions[Order] {
+trait OrderToOrderingConversion {
+  /**
+   * Implicitly convert a `Order[A]` to a `scala.math.Ordering[A]`
```
johnynek: This is not an implicit conversion which is something else. Can we say something like "provide an Ordering instance equivalent to the Order".

It does convert, and so I know I'm being pedantic, but I really don't like implicit conversions except for method enrichment, and in any case an implicit conversion has exactly one non-implicit parameter.

edmundnoble: I agree, but I changed this comment. I am not seeing how it showed up here :stuck_out_tongue:

edmundnoble: Ah sorry. Forgot to push.

kernel/src/main/scala/cats/kernel/Order.scala
```diff
@@ -139,7 +139,16 @@ abstract class OrderFunctions[O[T] <: Order[T]] extends PartialOrderFunctions[O]
...
+trait OrderToOrderingConversion {
+  /**
+   * Implicitly derive a `scala.math.Ordering[A]` from a `Order[A]`
+   * instance.
+   */
+  implicit final def catsKernelOrderingForOrder[A](implicit ev: Order[A]): Ordering[A] =
+    ev.toOrdering
+}
+
+object Order extends OrderFunctions[Order] with OrderToOrderingConversion {
```
kailuowang: I haven't researched this, but I am wondering if we place the method directly in the object instead of a trait, will that help binary comp going forward (on scala 2.11.x)?

johnynek:
for those following at home, the mima check failed on kernel.

edmundnoble:
I know that this isn't binary compatible. Is code duplication an acceptable fix?

johnynek:
@edmundnoble code duplication is an acceptable fix (i.e. they don't need to share the trait).

kernel/src/main/scala/cats/kernel/Order.scala
ceedubs: Is there any benefit to having `implicit final def catsKernelOrderingForOrder[A](implicit ev: Order[A]): Ordering[A]` inside the `Order` companion object? I don't think that scala will look here when it is searching for an `Ordering` instance. I could see a minor benefit of having a non-`implicit` version of this method inside the companion object. As it stands it seems like it's just a liability as I would expect you to get ambiguous implicits if you've imported both `cats.implicits._` and `cats.kernel.Order._`.

kailuowang: Now I don't know why I thought that would be a valid point to make. It has to be a trait for `cats.implicits._`. Sorry about the noise.

edmundnoble:
Binary compatible, now.

johnynek:
I wonder if this is going to create ambiguous implicits.

Can we have a test where we do `import cats.implicits._` in a scope and make sure that we can still do `implicitly[Ordering[String]]` etc... for some types that already have scala Orderings?

edmundnoble:
Ambiguity test added; it passes, I'm fairly sure the reason is that the `Ordering` companion object is firmly below imports in implicit scope.

edmundnoble:
@johnynek Any more feedback? Just rebased on master and squashed.

johnynek:
👍

edmundnoble:
Merged with two sign-offs.

PR #1671, opened 2017-05-16
Fixing `Eq[Function1]` in testsJS; break JS build to separate matrix build. 

kailuowang:
This PR breaks out the scalajs builds to separate travis matrix build to cut down build time which was exceeding the 50 minute hard limit set by traivs. 
see [#1666](https://github.com/typelevel/cats/pull/1666#issuecomment-301656528) for further discussion.

### DIFF
--- a/laws/src/test/scala/cats/laws/discipline/EqTests.scala
+++ b/laws/src/test/scala/cats/laws/discipline/EqTests.scala
@@ -1,5 +1,0 @@
-package cats.laws.discipline
-
-class EqTests {
-
-}
--- a/laws/src/test/scala/cats/laws/discipline/EqTests.scala
+++ b/laws/src/test/scala/cats/laws/discipline/EqTests.scala
@@ -1,0 +1,5 @@
+package cats.laws.discipline
+
+class EqTests {
+
+}
--- a/scripts/travis-publish.sh
+++ b/scripts/travis-publish.sh
@@ -20,18 +20,24 @@ export publish_cmd="publishLocal"
 if [[ $TRAVIS_PULL_REQUEST == "false" && $TRAVIS_BRANCH == "master" && $(cat version.sbt) =~ "-SNAPSHOT" ]]; then
   export publish_cmd="publish gitSnapshots publish"
   # temporarily disable to stabilize travis
-  #if [[ $TRAVIS_SCALA_VERSION = "2.11.8" ]]; then
-  #  export publish_cmd="$publish_cmd ghpagesPushSite"
+  #if [[ $TRAVIS_SCALA_VERSION = "2.11.11" ]]; then
+  #  export publish_cmd="publishMicrosite"
   #fi
 fi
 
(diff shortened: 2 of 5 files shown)


laws/src/test/scala/cats/laws/discipline/EqTests.scala
```diff
@@ -0,0 +1,5 @@
+package cats.laws.discipline
```
ceedubs: What's this file?

scripts/travis-publish.sh
```diff
@@ -20,18 +20,24 @@ export publish_cmd="publishLocal"
 if [[ $TRAVIS_PULL_REQUEST == "false" && $TRAVIS_BRANCH == "master" && $(cat version.sbt) =~ "-SNAPSHOT" ]]; then
   export publish_cmd="publish gitSnapshots publish"
   # temporarily disable to stabilize travis
-  #if [[ $TRAVIS_SCALA_VERSION = "2.11.8" ]]; then
-  #  export publish_cmd="$publish_cmd ghpagesPushSite"
+  #if [[ $TRAVIS_SCALA_VERSION = "2.11.11" ]]; then
```
ceedubs: Kind of tangential to this PR, and it looks like the code is commented out anyway, but we may want to change this to something like `if [[ $TRAVIS_SCALA_VERSION =~ ^2\.11\. ]]`. My knowledge on bash regular expressions is a bit foggy, so this may not be quite right.

laws/src/test/scala/cats/laws/discipline/EqTests.scala
kailuowang: Oops that's an  accident. Will remove after I get to a computer

scripts/travis-publish.sh
kailuowang: I verified that the code works.

kailuowang:
@ceedubs feedback addressed.

ceedubs:
Thanks @kailuowang! This looks good to me -- just waiting for the build to finish.

It looks like we are only using 1 worker in the scalatest/scalacheck configuration. Do you remember whether we've tried to increase that to reduce build times?

kailuowang:
> It looks like we are only using 1 worker in the scalatest/scalacheck configuration. Do you remember whether we've tried to increase that to reduce build times?

I don't, but maybe @BennyHill does?

BennyHill:
From memory:

On scala.js, you only have one thread, so setting more workers will have no difference.

On jvm, it may help, if you have _lots_ of threads. But sbt tends to grab lots of threads first, and besides, on Travis-CI we do not have many threads to start with - and I believe we run tests in a separate JVM as well

So it still _might_ be worth looking, but I would rather spend my time elsewhere

ceedubs:
@BennyHill thanks for the input. Since it's really easy to try bumping the number of workers, I'll give it a try in a PR once this one is merged, but I won't get my hopes up :)

PR #1672, opened 2017-05-16
Add Daniel Spiewak to maintainers

kailuowang:
The current maintainers are polled, all feedbacks are positive. Daniel himself also agreed to help out. So this PR will make it formal.


peterneyens:
Should we add @djspiewak and @edmundnoble to the pom file as well?

kailuowang:
@peterneyens good catch, updated.

adelbertc:
👍

PR #1673, opened 2017-05-16
Use 2 workers in JVM build

ceedubs:
As far as I can tell, our CI build has 2 cores available on Travis. So I
suspect that 2 workers might be a sweet spot for our JVM testing. As I
understand it, the JS runtime is single-threaded, so we probably won't
benefit from additional workers on it (as our tests should be
CPU-bound).

I could be totally wrong about these assumptions, but I'm interested in
how the build times for this PR compare.


ceedubs:
In the first build, this seems to reduce the JVM build from ~35min to ~30 min. It's tough to infer based on a sample size of 1 though.

ceedubs:
I'm going to restart the build to try to see whether the times are consistent.

ceedubs:
There seem to be similar results (or even a bit better) on the second build.

Issue #1675, opened 2017-05-17
Dedicated "Dev" channel. 


kailuowang:
Problem - development related communication is suboptimal
1. It's getting more difficult to call a certain maintainer's attention. Given that we all work on Github, each maintainer gets a load of github notifications and the majority of them do not require attent. A mention notification often gets buried in them and missed. 
2. the existing gitter room is vibrant which is good. We have a lot of support discussion going on in that channel which is great and what it supposed to be. But it also makes it harder to follow the important discussion around the development of cats. I've been spending 2-4 hours a day on cats on average in the last couple of months, and I can hardly keep up with the gitter. 
3. There are over 1300 people in the current gitter room, again a great thing, but also makes it a bit inappropriate to use a @/all to make dev only announcements. With the urgency of 1.0.0 increasing, we need the ability to quickly gather opinions from as many maintainers and contributors as possible. 

Proposed solution:
A dedicated public gitter chat room named `cats-dev`. Maintainers and active contributors are encouraged to join this room and discuss cats development only topics here. The discussion here should be much easier to follow on a regular basis. And we can more directly communicate to each other here (rather than through github Email notifications).

tpolecat:
👍

dwijnand:
I'd recommend "contrib" or "contributors" over "dev" given that cats users are also developers.

longcao:
Agree - as more of a user, it was nice seeing dev chatter on the same channel but since it's grown, it's getting harder to follow. It would be good to have `cats` be the de facto 'user' channel and a `cats-dev` or `cats-contributors` channel for longer discussions on library development, similar to many mailing list structures.

edmundnoble:
:+1: from me, scala does this itself.

ceedubs:
👍 for `cats-contributors`. I'd say `cats-contrib` but I feel like there's probably a library somewhere with this name already...

dwijnand:
Slashes is also an option, btw: `typelevel/cats/contrib`.

kailuowang:
With unanimous affirmative feedbacks, I went ahead created the channel
https://gitter.im/typelevel/cats-contrib
Will submit a PR to update the readme soon.

ceedubs:
Closing this out, as it has been resolved.

PR #1677, opened 2017-05-18
Update readme with the new dev channel.

kailuowang:
Per #1675, I created the cats-contrib channel and this is the PR to add it to README.


adelbertc:
👍

PR #1680, opened 2017-05-18
MonadTrans instance for RWST and make MonadTrans serializable

wedens:


tests/src/test/scala/cats/tests/ReaderWriterStateTTests.scala
```diff
@@ -344,6 +344,13 @@ class ReaderWriterStateTTests extends CatsSuite {
     checkAll("ReaderWriterStateT[ListWrapper, String, Int, String, Int]",
       SemigroupKTests[ReaderWriterStateT[ListWrapper, String, Int, String, ?]].semigroupK[Int])
   }
+
+  {
+    implicit def F = ListWrapper.monad
```
ceedubs: Minor: I think that this could be a `val` instead of a `def`.

tests/src/test/scala/cats/tests/StateTTests.scala
```diff
@@ -235,7 +235,7 @@ class StateTTests extends CatsSuite {
 
     checkAll("StateT[ListWrapper, Int, Int]", MonadTests[StateT[ListWrapper, Int, ?]].monad[Int, Int, Int])
     checkAll("Monad[StateT[ListWrapper, Int, ?]]", SerializableTests.serializable(Monad[StateT[ListWrapper, Int, ?]]))
-    checkAll("MonadTrans[EitherT[?[_], String, ?]]", MonadTransTests[EitherT[?[_], String, ?]].monadTrans[ListWrapper, Int, Int])
+    checkAll("MonadTrans[StateT[?[_], Int, ?]]", MonadTransTests[StateT[?[_], String, ?]].monadTrans[ListWrapper, Int, Int])
```
ceedubs: Oops good catch!

tests/src/test/scala/cats/tests/ReaderWriterStateTTests.scala
ceedubs: Should we be checking the serialization laws here too?

ceedubs:
Wow that was fast. Thanks @wedens!

tests/src/test/scala/cats/tests/ReaderWriterStateTTests.scala
wedens: copy/pasted it from some code above, so it's not the only place where it should be fixed :)

wedens:
fixed

wedens:
I'm not sure how to fix the build. Fails only on non-js 2.12.2

```
[info] - MonadTrans[ReaderWriterStateT[?[_], String, Int, String, ?]].serializable.can serialize and deserialize *** FAILED *** (45 milliseconds)
[info]   NotSerializableException was thrown during property evaluation.
[info]     Message: cats.data.RWSTInstances$$anon$12
[info]   
[info]     )
```

ceedubs:
Oh @wedens I think that we probably want the `MonadTrans` trait to extend `Any with Serializable`. I think that this is being done automatically for our type classes that use simulacrum. See `Semigroup` for an example of a type class that isn't using simulacrum.

Issue #1684, opened 2017-05-19
Consistency tests for override instances


yilinwei:
Currently we have overrides for specific methods for instances which are more efficient.

It's important that the output of the overriden method should be consistent with the output of the derived method and should be tested.

kailuowang:
The strategy we had in the past is to add it to the laws, (example)[https://github.com/typelevel/cats/blob/master/laws/src/main/scala/cats/laws/FlatMapLaws.scala#L21-L25]. The issue with this strategy is that these tests are not the same as the more fundamental laws that usually test against abstract methods. Shall we setup a new organization to make a distinction between these two types of laws?

peterneyens:
FWIW, [this](https://github.com/typelevel/cats/pull/1532#issuecomment-277326220) is what I played around with after we talked about creating some sort of benchmark for overridden/default implementations.

In essence it boils down to a type class specific wrapper that implements only the fundamental type class operations using the actual instance of the data type it wraps.

```scala
case class FoldableWrapper[F[_], A](val inner: F[A])(implicit F: Foldable[F])

object FoldableWrapper {
  implicit def foldableWrapperInstance[F[_]](implicit F: Foldable[F]): Foldable[FoldableWrapper[F, ?]] =
    new Foldable[FoldableWrapper[F, ?]] {
      def foldLeft[A, B](fa: FoldableWrapper[F, A], b: B)(f: (B, A) => B): B =
        F.foldLeft(fa.inner, b)(f)
    
      def foldRight[A, B](fa: FoldableWrapper[F, A], lb: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B] =
        F.foldRight(fa.inner, lb)(f)
    } 
}
```

Maybe we can use something like this somehow (for consistency and/or benchmarking) ?

yilinwei:
@peterneyens I like this approach and I think we can/should generate these tests.

On a side note it's like the issue described in #1659, in that we have a behaviour we expect but we enforce it via manual checking rather than in some automated fashion.

PR #1685, opened 2017-05-19
Fix issue in Reducible

yilinwei:
courtesy of @ceedubs finding the bug.

### DIFF
--- a/core/src/main/scala/cats/Reducible.scala
+++ b/core/src/main/scala/cats/Reducible.scala
@@ -227,7 +227,7 @@ abstract class NonEmptyReducible[F[_], G[_]](implicit G: Foldable[G]) extends Re
   }
 
   override def get[A](fa: F[A])(idx: Long): Option[A] =
-    if (idx == 0L) Some(split(fa)._1) else G.get(split(fa)._2)(idx)
+    if (idx == 0L) Some(split(fa)._1) else G.get(split(fa)._2)(idx - 1L)
 
   override def fold[A](fa: F[A])(implicit A: Monoid[A]): A = {
     val (a, ga) = split(fa)
--- a/tests/src/test/scala/cats/tests/ReducibleTests.scala
+++ b/tests/src/test/scala/cats/tests/ReducibleTests.scala
@@ -23,6 +23,15 @@ class ReducibleTestsAdditional extends CatsSuite {
       if (a === goal) Now(true) else lb
     }
 
+  test("Reducible[NonEmptyList] default get/size implementation") {
+    val R = new NonEmptyReducible[NonEmptyList, List] {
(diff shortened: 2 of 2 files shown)


ceedubs:
👍 LGTM but maybe this is a sign that we should have a unit test for this :)

yilinwei:
@ceedubs point taken....

kailuowang:
I think in principle every method that is not a simple delegation should be covered.  This [chrome extension](https://chrome.google.com/webstore/detail/codecov-extension/keefkhehidemnokodkdkejapdgfjmijf?hl=en-US) is really helpful in checking that.

peterneyens:
I think the problem here is that all the instances of `Reducible` override `get`. Another method I can think of with the same issue is ` Foldable#size`.

It would be nice if we could somehow check the consistency (#1684) and benchmark (something we talked about in #1532, see https://github.com/typelevel/cats/pull/1532#issuecomment-27732622) overridden and the default implementation of type class methods.

kailuowang:
How about we add a doctest to cover this and address the overrides test coverage in #1684? That way we can merge this fix sooner.

peterneyens:
@kailuowang, do you mean something like : 

```scala
scala> import cats.data.NonEmptyList

scala> val R = new NonEmptyReducible[NonEmptyList, List] {
     |   def split[A](nel: NonEmptyList[A]): (A, List[A]) = (nel.head, nel.tail)
     | }

scala> val nel = NonEmptyList.of(1,2,3)

scala> R.get(nel, 1L)
res0: Option[Int] = Some(2)

scala> R.get(nel, 4L)
res1: Option[Int] = None
```

kailuowang:
@peterneyens yup. that would work.

core/src/main/scala/cats/Reducible.scala
```diff
@@ -227,7 +227,7 @@ abstract class NonEmptyReducible[F[_], G[_]](implicit G: Foldable[G]) extends Re
   }
 
   override def get[A](fa: F[A])(idx: Long): Option[A] =
-    if (idx == 0L) Some(split(fa)._1) else G.get(split(fa)._2)(idx)
+    if (idx == 0L) Some(split(fa)._1) else G.get(split(fa)._2)(idx - 1L)
```
johnynek: this is not actually tested according to codecov.

Can we write a test to explicitly exercise this?

yilinwei: @johnynek done.

yilinwei:
Failure in the test is due to the `Kleisli` `Eq` presumably not working as intended; given that this PR doesn't touch that are we good to merge?

djspiewak:
Merging with two sign-offs.

Issue #1686, opened 2017-05-19
Where is cats.data.Tuple2K?


ctongfei:
The maven 0.9.0 jar does not seem to contain this type.

peterneyens:
It is called `cats.data.Prod` in 0.9.0.

ctongfei:
Got it. Thanks!

kailuowang:
it's my bad, I think I published the website with unreleased class name, but I think this is the only one.

PR #1687, opened 2017-05-19
Added some note for an unreleased class name `Tuple2K` fixes 

kailuowang:
related #1686


kailuowang:
closing this as fixed in #1688

PR #1691, opened 2017-05-20
Build JVM before JS on travis

peterneyens:
If there are **no other builds** (for other PRs, branches, ...), switching the JVM/JS order to JS/JVM can be 10-15 minutes quicker.

There are now 6 Travis jobs for every build (3 Scala versions and JVM / JS) and only 5 are executing at the same time. By switching the order, the three longer JVM jobs are started first and when the 2.11 JS job finishes, the 2.12 JS job can start (which doesn't take as long as the 2.12 JVM job which is now started last).  

I ran Travis [twice](http://imgur.com/a/dPLRz) resulting in total build times of 31 and 39 minutes while the last master builds took around 50 minutes.

This is just a minor improvement of the Travis matrix build change by @kailuowang in #1671.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,8 +47,8 @@ notifications:
 
 env:
   matrix:
-    - JS_BUILD=true
     - JS_BUILD=false
+    - JS_BUILD=true
   global:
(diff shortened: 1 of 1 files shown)


djspiewak:
> This is great. Thanks! I plan to research how to have more than 5 concurrent executions, but I guess you already did and it's not possible.

It's part of their commercial offering.

PR #1693, opened 2017-05-20
Clean up EitherT doctests

peterneyens:
With the partial unification fix in place, we can clean up some `EitherT` doctests a bit.


djspiewak:
👍 from me as well.  Merging with three sign-offs.

PR #1695, opened 2017-05-22
update ETA for 1.0.0

kailuowang:
Didn't notice that we had an ETA before.


kailuowang:
merging with 1 sign-off as very simple doc modification.

PR #1696, opened 2017-05-22
Make syntax.show extend ShowSyntax instead of Show.ToShowOps

edmundnoble:
Nobody can use the `show` interpolator otherwise with a-la-carte imports 😝

Edit: Thanks @alexknvl for pointing this out.


ceedubs:
@edmundnoble it looks like this broke `show` syntax usage in a test.

edmundnoble:
@ceedubs Fixed by removing the explicit import, it's odd that it would work this way but I think it's because it's shadowed by importing it when it's already in scope (through `CatsSuite`).

PR #1699, opened 2017-05-24
Link to sbt-partial-unification plugin

Blaisorblade:
Plugin argued for in https://github.com/typelevel/general/issues/69.
EDIT: it seems this is needed because of https://github.com/typelevel/cats/pull/1679. This particular PR is just one way to solve it.


adelbertc:
👍

PR #1705, opened 2017-05-26
Revert "up the sample number for `Eq` of function1 for JS"

kailuowang:
This reverts commit 80d56e649c6d2688bf57a3e717f415fcc8fc23c6 which was commit by mistake. Sorry.


kailuowang:
If two+ maintainers explicitly thumb up the original change (i.e. bump sample number from 10 -> 20 in scalaJs) then I think we can just close this PR.

edmundnoble:
@ceedubs You referring to the reversion or the original PR? For the record I'm thumbs up on the original.

ceedubs:
@edmundnoble oh I don't know....

I guess I'm just giving a 👍 for whatever the value is that doesn't produce false positives in the tests but doesn't make our builds take too long 😬

peterneyens:
It doesn't seem this slows the JS builds down (much), so :+1: 
Closing with 3 thumbs up (that's a first :smiley:),

PR #1706, opened 2017-05-27
Clean up ReaderWriterStateT

peterneyens:
- Check if type class instances are serializable
- Move Arbitrary instance to cats-laws
- Move unambiguous type class instances

---

I also find the ordering of the type parameters a bit confusing. Now it is `E` (`Reader`) -> `S` (`State`) -> `L` (`Writer`), while the name `ReaderWriterStateT` suggests that `Writer` comes before `State`. The reason is probably the `Bifunctor` instance for the `Writer` part.

I would drop the `Bifunctor` instance and switch the order of `S` and `L` so it aligns with the name. We could also leave as it is or change to `ReaderStateWriterT` :smiley:. If we want to change something, it should probably be done before 1.0.0-MF. Opinions?

### DIFF
--- a/core/src/main/scala/cats/data/ReaderWriterStateT.scala
+++ b/core/src/main/scala/cats/data/ReaderWriterStateT.scala
@@ -42,18 +42,6 @@ final class ReaderWriterStateT[F[_], E, S, L, A](val runF: F[(E, S) => F[(L, S, 
     transform { (l, s, a) => (f(l), s, a) }
 
   /**
-   * Combine this computation with `rwsb` using `fn`. The state will be be threaded
-   * through the computations and the log values will be combined.
-   */
-  def map2[B, Z](rwsb: ReaderWriterStateT[F, E, S, L, B])(fn: (A, B) => Z)(
-    implicit F: FlatMap[F], L: Semigroup[L]): ReaderWriterStateT[F, E, S, L, Z] =
-    flatMap { a =>
-      rwsb.map { b =>
-        fn(a, b)
-      }
-    }
-
-  /**
    * Modify the result of the computation by feeding it into `f`, threading the state
    * through the resulting computation and combining the log values.
    */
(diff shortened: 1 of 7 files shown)


core/src/main/scala/cats/data/ReaderWriterStateT.scala
```diff
@@ -42,18 +42,6 @@ final class ReaderWriterStateT[F[_], E, S, L, A](val runF: F[(E, S) => F[(L, S,
...
-   * Combine this computation with `rwsb` using `fn`. The state will be be threaded
-   * through the computations and the log values will be combined.
-   */
-  def map2[B, Z](rwsb: ReaderWriterStateT[F, E, S, L, B])(fn: (A, B) => Z)(
-    implicit F: FlatMap[F], L: Semigroup[L]): ReaderWriterStateT[F, E, S, L, Z] =
-    flatMap { a =>
-      rwsb.map { b =>
-        fn(a, b)
-      }
-    }
```
peterneyens: You can get the same from `Apply#map2`

The version from `Apply#map2` would look like : 

```scala
map(flatMap(fa)(a => map(fb)(b => (a, b)))) { case (a, b) => f(a, b) }
``` 

If we want the simpler implementation used here in `RWST`, we could override `map2` in `FlatMap`.

tests/src/test/scala/cats/tests/ReaderWriterStateTTests.scala
```diff
@@ -279,53 +280,77 @@ class ReaderWriterStateTTests extends CatsSuite {
...
 
   {
     implicit val LWM: Monad[ListWrapper] = ListWrapper.monad
 
     checkAll("ReaderWriterStateT[ListWrapper, String, Int, String, Int]",
       MonadReaderTests[ReaderWriterStateT[ListWrapper, String, Int, String, ?], String].monadReader[String, String, String])
+
+    // check serializable using Option
+    checkAll("MonadReader[ReaderWriterStateT[Option, String, Int, String, ?], String]",
+      SerializableTests.serializable(MonadReader[ReaderWriterStateT[Option, String, Int, String, ?], String]))
```
peterneyens: If I used `ListWrapper here I got

```
Message: cannot assign instance of scala.collection.immutable.List$SerializationProxy to field cats.tests.ListWrapper.list of type scala.collection.immutable.List in instance of cats.tests.ListWrapper
```

So I used `Option` like the analogous test for `Kleisli`.

kailuowang:
Thanks for doing this @peterneyens! I agree we should address these before 1.0.0-MF

Issue #1635, opened 2017-04-25
Why does Eval use `defer` while Free uses `suspend`?


alexandru:
It's inconsistent naming, we should standardize.

Also see https://github.com/typelevel/cats/pull/1634

johnynek:
+1 to standardizing, but please keep the deprecated method around to ease adoption. Too much churn makes it difficult to upgrade large code bases.

kailuowang:
fixed by #1709

PR #1709, opened 2017-05-27
Rename `suspend` to `defer` (or vice versa ??)

peterneyens:
I chose `defer` over `suspend` because both `Free` and `FreeT` have a `Suspend` which is used by `liftF` (not by `suspend`), so this may reduce that confusion as well.

This meant replacing three methods though, where `Eval.defer` -> `Eval.suspend` would be only one change. Cats-effect also seems to have chosen `suspend` (in `Sync`).

So this isn't set in stone, I'll happily switch it around if that is what we prefer. Just thought that a PR may help us to decide, since #1635 has been a bit quiet. So, what do we think?


johnynek:
I like defer over suspend as well (shorter and in my mind clearer).

kailuowang:
👍  on `defer` and the code change here.

kailuowang:
merge with three thumbs up;

Issue #1567, opened 2017-03-21
SplitLaw doesn't hold in general


peterneyens:
When you run our current `SplitLaw` for `Kleisli[Either[String, ?], ?, ?]` it fails the `splitInterchange` law.
@fthomas already mentioned in #232 that he wasn't sure that it was sound.

If you add the following to `KleisliTests` you can see it fail.
```scala
{
  type StringOr[A] = Either[String, A]
  
  // SI 2712 workarounds
  implicit val EqFAA     = implicitly[Eq[Kleisli[StringOr,Int,Int]]]
  implicit val EqFACB    = implicitly[Eq[Kleisli[StringOr,(Int, Int),Int]]]
  implicit val EqFACBC   = implicitly[Eq[Kleisli[StringOr,(Int, Int),(Int, Int)]]]
  implicit val EqFACDBCD = implicitly[Eq[Kleisli[StringOr,((Int, Int), Int),(Int, (Int, Int))]]]
  
  implicit val invariant = Kleisli.catsDataFunctorForKleisli[StringOr, Int]
  implicit val iso = CartesianTests.Isomorphisms.invariant[Kleisli[StringOr, Int, ?]]
  
  implicit val ArbFAB = implicitly[org.scalacheck.Arbitrary[Kleisli[StringOr,Int,Int]]]
  
  implicit val catsDataArrowForKleisli = Kleisli.catsDataSplitForKleisli[StringOr]
  checkAll("Kleisli[Either[String, ?], Int, Int]", SplitTests[Kleisli[StringOr, ?, ?]].split[Int, Int, Int, Int, Int, Int])
}
```

I looked at it with @diesalbla and the law only seems to hold for commutative arrows.

diesalbla:
[...]

Issue #334, opened 2015-06-08
Typeclasses that ONLY define laws.


woparry:
What do people think of adding typeclasses that only add new laws (ie. have no new functionality, but restrict which implementations are valid.)

I'm specifically thinking of something like CommutativeMonad, which would add a commutativity restriction to Monad (so that e.g. a distributive law can be derived which would allow composition, see #332 and the related discussion re: Traverse).

Since we have discipline, the extra laws can be checked by implementers fairly easily, and it opens up new avenues for writing slightly more general combinators in cats.

ceedubs:
Sounds good to me.

PR #1719, opened 2017-06-04
Cats-1567 Replace Split with Commutative Arrow. Introduces Commutative Monad. 

diesalbla:
This PR carries the work to solve issue #1567 following the direction agreed in [this comment](https://github.com/typelevel/cats/issues/1567#issuecomment-288121078). 

* We remove the `Split` type-class, which was only declaring a `split` method. This method is now integrated into `Arrow`, with a default implementation. We also remove the (flawed) `SplitLaws`, its `SplitTests`.
* We add a new method-less type-class `CommutativeArrow`, which restricts the `Arrow` type-class with an extra law, the commutativity of the `split` operation. We also create its laws and tests. The use of type-classes with laws but without operations was suggested in #334. 
* We remove the instances of `Split` for `Kleisli` and `CoKleisli`; the special implementation for `split` is now integrated into their respective `Arrow` instances.

We found that this would unable some tests of the arrow-commutativity laws for certain instances of `Arrow[Kleisli[F,?,?]]`, such as `F=Id` and `F=Option`, in which the law holds. To keep these tests: 

* We introduce a sub-type-class of `Monad`, `CommutativeMonad`, in which the `flatMap` operation is commutative. We change the sets of instances of `Monad[Option]` and `Monad[Id]` to `CommutativeMonad`. 
* We create some instances of `CommutativeArrow[Kleisli[F,?,?]]`` for those `F[_]` that are `CommutativeMonad`.

I can mention a couple of design choices that may be open to discussion: 
* We could as well create  a type-class of commutative _Comonads_, and their use for `Cokleisli`, but I have not found any existing tests for those so there seems to be no urgent need at present. 
* We may consider if _commutativity_ should be available not just for `Arrow` and `Monad`, but perhaps for `Compose` and `Apply` as well.

PS (Edited): The instances of `CommutativeMonad` for `Option` and `Id` are the ones needed to keep existing tests. There may be other types `F[_]` for which a `Monad[F]` is provided in `cats` that is also a `CommutativeMonad[F]`.


diesalbla:
@edmundnoble @fthomas @kailuowang @peterneyens

core/src/main/scala/cats/arrow/Arrow.scala
```diff
@@ -20,6 +26,25 @@ import simulacrum.typeclass
...
+   * scala> f((3, 4.0f))
+   * res0: (Long, Double) = (3,4.0)
+   * }}}
+   *
+   * Note that the arrow laws do not guarantee the non-interference between the _effects_ of
+   *  `f` and `g` in the context of F. This means that `f *** g` may not be equivalent to `g *** f`.
+   */
+  @simulacrum.op("***", alias = true)
+  def split[A, B, C, D](f: F[A, B], g: F[C, D]): F[(A, C), (B, D)] =
     andThen(first(f), second(g))
```
kailuowang: quick note, codecov somehow mistakenly thinks this line is uncovered.

core/src/main/scala/cats/data/Cokleisli.scala
```diff
@@ -107,18 +107,15 @@ private trait CokleisliArrow[F[_]] extends Arrow[Cokleisli[F, ?, ?]] with Coklei
   override def dimap[A, B, C, D](fab: Cokleisli[F, A, B])(f: C => A)(g: B => D): Cokleisli[F, C, D] =
     super[CokleisliProfunctor].dimap(fab)(f)(g)
 
-  override def split[A, B, C, D](f: Cokleisli[F, A, B], g: Cokleisli[F, C, D]): Cokleisli[F, (A, C), (B, D)] =
-    super[CokleisliSplit].split(f, g)
+  override def split[A1, B1, A2, B2](f: Cokleisli[F, A1, B1], g: Cokleisli[F, A2, B2]): Cokleisli[F, (A1, A2), (B1, B2)] =
+    Cokleisli(fac => f.run(F.map(fac)(_._1)) -> g.run(F.map(fac)(_._2)))
```
kailuowang: Since we removed the `split` from the laws, this line is not tested at all (same in `Kleisli`). It looks like parametricity guarantees the right implementation, but to be consistent, shall we still have some test coverage (I am fine with a quick doc test)?

core/src/main/scala/cats/data/Kleisli.scala
```diff
@@ -163,15 +166,15 @@ private[data] sealed abstract class KleisliInstances5 {
...
 
   def lift[A, B](f: A => B): Kleisli[F, A, B] =
     Kleisli(a => F.pure(f(a)))
-}
-
-private trait KleisliSplit[F[_]] extends Split[Kleisli[F, ?, ?]] with KleisliCompose[F] {
-  implicit def F: FlatMap[F]
 
   override def split[A, B, C, D](f: Kleisli[F, A, B], g: Kleisli[F, C, D]): Kleisli[F, (A, C), (B, D)] =
     Kleisli{ case (a, c) => F.flatMap(f.run(a))(b => F.map(g.run(c))(d => (b, d))) }
```
kailuowang: same situation here as in `Cokleisli`.
An alternative approach we can take is to add `split` to the arrowLaws that simply tests whatever override is consistent with  `andThen(first(f), second(g))`. Unfortunately, we don't have a consistent strategy over how to tests these overrides in instances yet.  So for now, I'd vote for just adding a doctest.

core/src/main/scala/cats/data/Cokleisli.scala
diesalbla: The type does imply that in each output value `(b,d): (B,D)`, to obtain the `B` (resp. `D`) you need to execute `f` (resp `g`) on the input component of type `A` (resp. `C`), but that leaves space to at least two possible implementations: one in which `f` runs before `g`, and one in which `g` runs before `f`. These implementations are not equivalent if the effects interfere, for instance, if they each one echoes and prints its input into `stdout`. That is why `split` is only really bound, apart from its definition, by commutativity. We can test this line if we write the _dual_ of the commutative monad, and use it to derive instances as done for `Kleisli`. 

The conventional implementation (Haskell) is to run `f` _first_, which I guess is because `f` refers to the _first_ components of input and output.

kailuowang: Right.  I was saying that the `split` method is not directly tested in the commutativity law, so it's not tested for both `Kleisli` and `Cokleisli`. I agree that it's probably fine since the commutativity law is the only **relevant** law to it and that together with parametricity probably it is enough to ensure its correctness (although, as you pointed out, in the noncommutative case the correctness is a bit arbitrary). 
I don't feel strong about it but adding a doctest (or just a regular unit test) to provide some codecov coverage wouldn't hurt right? I mean right now you can replace the implementation of these two `split` impl with `???`s without breaking any test, which is a bit uneasy to me.

edmundnoble:
It may be out of scope for this PR, but I'd also like to see `CommutativeCartesian`, `CommutativeApply` and `CommutativeApplicative`.

diesalbla:
@edmundnoble I am afraid that the issue is a bit out of scope for this PR, whose original intent was just to fix a problem in the `arrow` package. It can easily be done in a separate PR, concurrently from this one. I suppose that the laws for such type-classes would be as follows: 

- Commutative Cartesian: for any `a: F[A]` and `b : F[B]` , it holds that 
```product(a,b) <-> product(b,a).map(swap)```.
- Commutative Apply: for any `a: F[A]` and `f: F[A => B]`, 
```Scala
def flip$(x: A)(f: A => B): B = f(x)  // Haskell `flip ($)`
ap(f,a) <-> ap( a.map(flip$), f)
```
- Commutative Applicative: no special laws, since `Applitcative` only gives the Unit. Just subtyping and inheritance.

edmundnoble:
@diesalbla Agreed, though I think `CommutativeApply` has no extra laws because that one is derivable from `CommutativeCartesian`'s.

core/src/main/scala/cats/data/Cokleisli.scala
```diff
@@ -45,37 +45,24 @@ object Cokleisli extends CokleisliInstances {
...
 
 private[data] sealed abstract class CokleisliInstances extends CokleisliInstances0 {
-  implicit def catsDataArrowForCokleisli[F[_]](implicit ev: Comonad[F]): Arrow[Cokleisli[F, ?, ?]] =
-    new CokleisliArrow[F] { def F: Comonad[F] = ev }
-
-  implicit def catsDataMonadForCokleisli[F[_], A]: Monad[Cokleisli[F, A, ?]] = new Monad[Cokleisli[F, A, ?]] {
-    def pure[B](x: B): Cokleisli[F, A, B] =
-      Cokleisli.pure(x)
+  implicit def catsDataCommutativeArrowForCokleisli[F[_]](implicit ev: CommutativeComonad[F]): CommutativeArrow[Cokleisli[F, ?, ?]] =
+    new CokleisliCommutativeArrow[F] { def F: CommutativeComonad[F] = ev }
```
kailuowang: Looks like that the build wasn't happy because you forgot to give this extra help scalac needed. 
```scala
  implicit val catsDataCommutativeArrowForCokleisliId: CommutativeArrow[Cokleisli[Id, ?, ?]] =
    catsDataCommutativeArrowForCokleisli[Id]
```

kailuowang:
@diesalbla just curious if you have time to finish this one up. If not I would be more than happy to help. I think it's very close, just need to fix the build according to the comment I made above.

kailuowang:
looks like one of the new tests failed
```
 NonEmptyList[Int].commutative comonad.coflatmap commutativity *** FAILED *** (32 milliseconds)
 GeneratorDrivenPropertyCheckFailedException was thrown during property evaluation.
  (Discipline.scala:14)
   Falsified after 3 successful property evaluations.
   Location: (Discipline.scala:14)
   Occurred when passed generated values (
     arg0 = NonEmptyList(-1),
     arg1 = NonEmptyList(890208370, -2147483648),
     arg2 = org.scalacheck.GenArities$$Lambda$5347/112471456@45377d5e
   )
   Label of failing property:
     (NonEmptyList(NonEmptyList(1, 1)) ?== NonEmptyList(NonEmptyList(1), NonEmptyList(1))) failed
```

kailuowang:
It looks like that the current `NonEmptyList.coflatmap` isn't commutative.

kailuowang:
@diesalbla I've created a PR on your branch which should fix the conflicts and build error. 
https://github.com/diesalbla/cats/pull/1

kailuowang:
continued with #1766

diesalbla:
This PR were continued in PR #1766, which has been merged. I am closing this Pull Request.

PR #1720, opened 2017-06-05
 type classes with only defined laws to guideline

### LINKED ISSUE #334: Typeclasses that ONLY define laws.

kailuowang:
This shall fix #334


kailuowang:
assigning @ceedubs since you thumbed up on #334

ceedubs:
@kailuowang thanks for this! Sorry, but I didn't see it until too late, and it has merge conflicts :(

kailuowang:
@ceedubs sorry for the delayed response. Conflicts resolved

Issue #1645, opened 2017-05-01
Add laws for InjectK typeclass


andyscott:
InjectK has (tentatively) been converted to a typeclass and deserves the full treatment by introducing laws.

See #1642

andyscott:
I'd assign myself to this, but I can't! So feel free to assign me.

PR #1725, opened 2017-06-09
Add InjectK laws

andyscott:
This addresses issue #1645.

### DIFF
--- a/laws/src/main/scala/cats/laws/InjectKLaws.scala
+++ b/laws/src/main/scala/cats/laws/InjectKLaws.scala
@@ -1,0 +1,20 @@
+package cats
+package laws
+
+trait InjectKLaws[F[_], G[_]] {
+  def injectK: InjectK[F, G]
+
+  def injectKRoundTripInj[A](fa: F[A]): IsEq[Option[F[A]]] =
+    (injectK.prj compose injectK.inj).apply(fa) <-> Some(fa)
+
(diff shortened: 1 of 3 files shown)


laws/src/main/scala/cats/laws/InjectKLaws.scala
```diff
@@ -0,0 +1,12 @@
...
+package laws
+
+trait InjectKLaws[F[_], G[_]] {
+  def injectK: InjectK[F, G]
+
+  def injectKRoundTripInj[A](fa: F[A]): IsEq[Option[F[A]]] =
+    (injectK.prj compose injectK.inj).apply(fa) <-> Some(fa)
+
+  def injectKRoundTripPrj[A](ga: G[A]): IsEq[Option[G[A]]] =
+    injectK.prj(ga).map(fa => injectK.inj(fa)) <-> Some(ga)
```
kailuowang: I think it's lawful for `prj` to return a `None`, isn't it?

andyscott: Yes! This commit is goofy-- I cherry picked it from ages ago and I'm reworking it now.

That being said, how do you check this kind of behavior with discipline? Would it suffice to just have the `injectKRoundTripInj` and nix `injectKRoundTripPrj`?

kailuowang: I'd say when it returns a `Some` you should check the roundtrip right?

PR #1728, opened 2017-06-12
Finishing up #1564 (add As a.k.a liskov)

kailuowang:
This is an effort to finish up #1564 
complete the missing test with 2 new helper methods for `Function1`

Closes #1564


johnynek:
looks great!

Thanks a ton @stew and @kailuowang (and prior scalaz authors).

alexknvl:
Can `Leibniz` and `Liskov` type aliases be removed? I am not seeing the rationale behind having extra symbolic type aliases. They could be used for bounded versions of `As` and `Is` in the future.

johnynek:
Can you describe what you mean in more detail?

Bounded how?

alexknvl:
Similar to https://github.com/alexknvl/leibniz/blob/master/src/main/scala/leibniz/Leibniz.scala
As it is `Is[A, B]` can not be used to substitute into bounded type constructors such as `F[A >: L <: H]` even if you know that both `A >: L <: H` and `B >: L <: H`.

johnynek:
I'm confused why you want such a thing rather than just several As
instances. Won't As[L, A], Is[A, B] and As[B, H] be the same as your
Leibniz?

alexknvl:
No. You can't go to `L <: A` knowing only `L <~< A`, at least not with the available functionality of `As` (there are some tricks that you can do involving existential types, but it won't help with `F[A >: L <: H]`). The compiler won't let you write `F[A]` for `F[X >: L <: H]` because it has to know that such type constructor makes sense. You can't even `asInstanceOf` it since `F[A]` (or `F[B]`) will be a semantic error.

alexknvl:
Having `A === B` and knowing `A >: L <: H` and `B >: L <: H` is enough to justifiably coerce `F[A]` to `F[B]`, using `asInstanceOf`. But unless you have a bounded version of `Is`, you can't do it without `asInstanceOf`.

alexknvl:
Similarly, having just `As[A, H]` is not enough to be able to call `def foo[X <: H](x: X): Option[X]`, you need something along the lines of [`As1`](https://github.com/alexknvl/leibniz/blob/master/src/main/scala/leibniz/As1.scala).

Issue #1729, opened 2017-06-13
Add environment comonad


andyscott:
Would a PR to add `EnvT` (https://hackage.haskell.org/package/comonad-5.0.1/docs/Control-Comonad-Trans-Env.html) be a reasonable addition?

adelbertc:
Hm given Matryoshka is planning on moving to Cats (ping @sellout ) do we want this in Cats, and is this "generally useful" enough to be in something like cats-core?

edmundnoble:
We already have this, without its own datatype, in `instances.tuple`. My questions are: do we want it as its own type like `Reader`? Do we want the transformer? And what will we do with comonad transformers in general?

sellout:
Yeah, I dunno – Haskell does alias `Env` to `EnvT Id`, but I would generally just use tuples. In Matryoshka we use `EnvT` two ways – for its structure as the pattern functor of `Cofree` and then in `gzygo`, which actually uses it as a transformer. I suppose that given some MTL-style type class, I would use that instead, which hadn’t even occurred to me until now.

But I have no problem keeping it all in Matryoshka until/if it’s useful for someone else, or if someone wants to build the dual of cats-mtl or whatever. But if it’s in Cats, I’ll definitely get rid of my definition of it.

PR #1730, opened 2017-06-14
Add Env/Coreader/Product comonad

andyscott:
Adds Env/Coreader/Product comonad.

This is missing tests and documentation. I'd like to get a feel for whether or not this is a welcomed addition to cats before I go ahead and implement the rest. I filed an issue as well, see #1729. Thanks!

### DIFF
--- a/core/src/main/scala/cats/data/EnvT.scala
+++ b/core/src/main/scala/cats/data/EnvT.scala
@@ -1,0 +1,57 @@
+package cats
+package data
+
+final case class EnvT[F[_], E, A](ask: E, lower: F[A]) {
+
+  def map[B](f: A => B)(implicit F: Functor[F]): EnvT[F, E, B] =
+    EnvT[F, E, B](ask, F.map(lower)(f))
+
+  def coflatten(implicit F: CoflatMap[F]): EnvT[F, E, EnvT[F, E, A]] =
+    EnvT[F, E, EnvT[F, E, A]](ask, F.coflatMap(lower)(EnvT[F, E, A](ask, _)))
+
+  def coflatMap[B](f: EnvT[F, E, A] => B)(implicit F: CoflatMap[F]): EnvT[F, E, B] =
+    coflatten.map(f)
+
+  def extract(implicit F: Comonad[F]): A = F.extract(lower)
+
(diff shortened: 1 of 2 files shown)


kailuowang:
@andyscott what are your thoughts on this PR given the feedbacks in #1729?

andyscott:
I have also been using `EnvT` as a pattern functor for `Cofree` in some private code. I'm happy to finish the PR if we'd like to get this merged into cats.

In general, Cats has missing odds and ends-- but I'd hate to see the core become bloated by adding them all. I think we ultimately need these odds and ends, and ideally we want them in maintained in a modular fashion so other libraries can use them.

With that in mind, I'd argue there should ultimately be auxiliary modules for cats (either as a submodule in this repo, or in a separate repo).

But again, I'm happy to finish this PR for the core module :).

kailuowang:
> In general, Cats has missing odds and ends-- but I'd hate to see the core become bloated by adding them all. I think we ultimately need these odds and ends, and ideally we want them in maintained in a modular fashion so other libraries can use them.

I have the same thoughts. Also locking down `cats-core` bincomp probably means that practically new additions like this would have to be added as separate modules anyway. 
We could start a new module and label it as "experimental" which means no backward compat guarantee until we remove the label. 
The question in this particular matter is that does this EnvT justify a new module? If so how shall we name the module?

andyscott:
A few notes on potentially adding a new module:

- For the name of a new module, I would prefer to avoid the "experimental" moniker anywhere in the path. For me it has a slightly negative association, especially with regards to using it in production software. I think something along the lines of `cats-addendum` or `cats-auxiliary` would be nice. Or `cats-adjunct` if you want a fun name.

- Items added to the new module could eventually be moved to the core cats module if they become cornerstone.

- Would this new module live in this repo? Or should it be a new repo? (ideally under TL).

- `EnvT` on its own right doesn't justify a whole new module. But I think we could easily assemble some odds and ends in order to justify the new module.

- How would we package items added to the new module? Would `cats.data.aux.EnvT`, for example, be a safe path?

If we go this route, I am happy to jump start it.

kailuowang:
We might want to open a new issue to bikeshed about this new module.

PR #1731, opened 2017-06-16
Add scalafmt

### LINKED ISSUE #1206: Consider adding scalafmt
As @mpilquist pointed out in #1195, [scalafmt](https://olafurpg.github.io/scalafmt/) can both check that javadoc-style comments are used _and_ automatically format them for you. We may want to consider incorporating it into the project in some way.

ceedubs:
Resolves #1206

A few notes:
- Currently this is using sbt-scalafmt-coursier, which is a project
that's in beta status but seems to be working.
- Currently this is set up so that no code is automatically formatted
(you have to run the `scalafmt` command), but if there is a formatting
issue, the `validateJVM` command will fail.
- There weren't that many formatting changes required, which probably
means some combination of Cats having consistent style and the default
formatting rules being fairly lenient.


ceedubs:
Other things to consider:

- There may be rules that are useful that aren't enabled by default. I haven't really looked into this.
- I think that there's still benefit in having scalastyle for some of its rules, but there may be some that are made redundant with scalafmt. This probably is fine, though.

kailuowang:
👍.  I also vote for turning automatical formatting on. I have auto format enabled for all of my projects and never encountered a problem.

djspiewak:
I'm generally in favor of enabling auto-formatting, but only after we bikeshed the settings just a little bit.  Some of the changes in the diff make me twitch violently, so I'd like the chance to at least make my case as to why it's a bad rule before we universally enforce it.

ceedubs:
@djspiewak I just used the default scalafmt configuration because personally I value consistency but am fairly ambivalent on the details. I imagine that a lot of projects have(and will) adopted the default scalafmt settings, so I see a bit of benefit in using them, but I'm also completely fine with an alternative configuration being used. Feel free to submit a PR to my branch or to open a separate PR if you will get to this soon. If you don't see yourself making a PR in the near future, then it may be best for us to merge this for the sake of consistency and we can always make changes in the future (but formatting changes tend to cause merge conflicts).

djspiewak:
@ceedubs I expect to loop back 'round to this by Saturday.  If you don't have a PR from me by then, consider it an implicit :+1: to merge as-is.

ceedubs:
@djspiewak I don't think that merging as-is is an option anyway. There are build errors in scala 2.10 and 2.11. I'm not sure that I completely understand them. I seem to have a recollection of scalafmt not working in scala 2.10? Guidance appreciated 😬

kailuowang:
my theory is that scalafmt is built on scalameta which doesn't support scala 2.10

easel:
I'd love to see this merged. Given the small number of reformatting that introducing scalafmt with the default settings added out of the gate, I vote to dispense with the bike shedding of settings and add a .scalafmt file referencing the default settings.

easel:
@ceedubs You should have a PR from me with this branch rebased and updated to the latest scalafmt. I'm reasonably hopeful it will "just work" now, it's evolved a bunch since mid 2017.

ceedubs:
Closing out as this branch is stale and @easel looks to be picking up this effort!

easel:
Yeah can do, just didn't want to create a duplicate PR while yours was open. Thanks

Issue #1568, opened 2017-03-22
No Ops for `tailRecM`


Atry:
I guess we can create a `FlatMapIdOps` that contains `tailRecM` method, like what we had in `ApplicativeIdOps`

diesalbla:
@Atry `tailRecM` is a method in the `FlatMap` type class, but AFAIK it mostly attends an implementation concern, i.e., that of avoiding stack overflows when using, say, a `traverse` method. I am not entirely sure as to whether a specific syntax for it would be warranted. 

¿Could you provide a code example where such syntax would be useful?

edmundnoble:
Monadic tail recursion is just as useful as ordinary tail recursion, especially if at the use site you don't know what monad it is. People use it in user code all the time.

Atry:
@diesalbla When I was porting [ThoughtWorks Each](https://github.com/ThoughtWorksInc/each) to cats, I looked for `whileM_` https://github.com/typelevel/cats/issues/1569#issuecomment-288590918 . Then I found there is a similar method `tailRecM`.

I guess I will use `whileM_` if it presents.

PR #1732, opened 2017-06-16
Add tailRecM syntax for FlatMap

aaronlevin:
resolves https://github.com/typelevel/cats/issues/1568

### DIFF
--- a/core/src/main/scala/cats/syntax/flatMap.scala
+++ b/core/src/main/scala/cats/syntax/flatMap.scala
@@ -8,6 +8,9 @@ trait FlatMapSyntax extends FlatMap.ToFlatMapOps {
 
   implicit final def catsSyntaxIfM[F[_]: FlatMap](fa: F[Boolean]): IfMOps[F] =
     new IfMOps[F](fa)
+
+  implicit final def catsSyntaxFlatMapIdOps[A](a: A): FlatMapIdOps[A] =
+    new FlatMapIdOps[A](a)
 }
 
 final class FlattenOps[F[_], A](val ffa: F[F[A]]) extends AnyVal {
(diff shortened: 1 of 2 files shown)


aaronlevin:
Thanks for the `low-hanging-fruit` flag! Wouldn't have found this issue without it.

core/src/main/scala/cats/syntax/flatMap.scala
```diff
@@ -50,3 +52,17 @@ final class IfMOps[F[_]](val fa: F[Boolean]) extends AnyVal {
...
+   * scala> import cats.implicits._
+   *
+   * scala> val a: Int = 10
+   * scala> a.tailRecM[Option,String](i => if (i == 20) Some(Right("done")) else Some(Left(i+1)))
+   * res0: Option[String] = Some(done)
+   *}}}
+   */
+final class FlatMapIdOps[A](val a: A) extends AnyVal {
+  def tailRecM[F[_],B](f: A => F[Either[A,B]])(implicit F: FlatMap[F]): F[B] = F.tailRecM(a)(f)
+}
```
peterneyens: I think you need to add a new line hear to appease scalastyle.

ceedubs: Thanks, @aaronlevin!

Minor: it might be more helpful to have this scaladoc comment on the `tailRecM` method instead of the `FlatMapIdOps` class. If someone hovers over the `tailRecM` in something like `10.tailRecM(...)` in their IDE, this scaladoc might pop up if it's defined on the method as opposed to the class. Also conceivably there could be more methods that we add to this class.

aaronlevin:
Thanks, I'll address these two comments right away (on mobile right now).

PR #1714, opened 2017-06-01
StateT violates laws

ceedubs:
As pointed out [here](https://github.com/typelevel/cats/issues/1640#issuecomment-298125906)

This reveals law violations in `StateT`. It looks like `flatMap`/`ap`
consistency does not hold, as well as `MonadCombine`
right-distributivity. It's not immediately clear to me how to fix this,
so I'm opening this up in case somebody else gets to it first.

PR #1735, opened 2017-06-20
StateT no longer violates laws

djspiewak:
Closes #1714 

Fixes the issues uncovered by @ceedubs.  After a fair bit of digging, I'm relatively confident that the issues in question are stemming from the inconsistencies (but not unlawfulness) of deriving `Applicative` instances from nested type constructors (inconsistent relative to analogous monadic derivation).  My arguments on #1714 were overbroad and unsupported, and further digging showed that they don't hold in general (which is good!).

So I think we're good with `StateT` in its current formulation.  At least, in so far as I can tell.  The tests are now passing.  This does break the API though, since certain functions are removed from `StateT` and now once again only available implicitly.


ceedubs:
Thanks @djspiewak! I was feeling very uneasy about this.

It seems like it would be nice to have a `map2Eval` implementation that were lazier (for `F` types that support laziness), but I guess I don't even know whether that's possible. And (assuming the build passes), this definitely seems like an improvement over unlawful implementations. 👍

djspiewak:
Merging with two sign-offs.

PR #1736, opened 2017-06-20
Added StackSafeMonad mixin

djspiewak:
I'm tired of defining the same `tailRecM` over and over again when the underlying monad is already known to be stack-safe.  This mixin provides a more convenient way of achieving this goal.  Obviously, the implementation is only valid if the underlying monad has a stack-safe `flatMap`, but *if* it has a stack-safe `flatMap`, then this is almost certainly going to be the implementation.  May as well reduce some of the boilerplate in that case.


djspiewak:
Merging with two sign-offs.

PR #1739, opened 2017-06-21
Add `onError` and `adaptError` to Applicative/MonadError

SystemFw:
PR:
As per [this](https://gitter.im/typelevel/cats?at=592d9206fcbbe1891c5cd4ac) Gitter conversation with @djspiewak, this PR adds two functions:

`ApplicativeError#onError`, to execute a callback on certain errors and rethrow:
```scala
onError[A](fa: F[A])(pf: PartialFunction[E, F[Unit]]): F[A]
```
with laws:
```scala
pure(a).onError(f) <-> pure(a)
raiseError(e).onError(err => fb) <-> fb *> raiseError(e)
```
Use case: error handling that respects module boundaries, e.g. rollback at the db layer, and then send a failed response at the http layer

`MonadError#adaptError`, to transform certain errors
```scala
def adaptError[A](fa: F[A])(pf: PartialFunction[E,E]): F[A] 
```
with laws:
```scala
pure(a).adaptError(f) <-> pure(a)
raiseError(e).adaptError(f) <-> raiseError(f(e))
```
Use case: transforming errors (especially Throwable) between different layers of an application.  The type is not as good as `leftMap` but the best we can have I think (except maybe constraining to Throwable).

A few things I'm not sure about code-wise:
- Lack of `syntax` imports and trying to adhere to the existing style resulted in fairly "lispy" code
- I couldn't get an `Arbitrary[F[Unit]]` in the laws, had to settle for `F[B].void`
- Haven't added non-discipline tests (yet?), they all seem to be based on Option, for which none of these two functions happen to make sense
- Should I add specialised implementations where relevant?

### DIFF
--- a/core/src/main/scala/cats/syntax/applicativeError.scala
+++ b/core/src/main/scala/cats/syntax/applicativeError.scala
@@ -34,4 +34,7 @@ final class ApplicativeErrorOps[F[_], E, A](val fa: F[A]) extends AnyVal {
 
   def recoverWith(pf: PartialFunction[E, F[A]])(implicit F: ApplicativeError[F, E]): F[A] =
     F.recoverWith(fa)(pf)
+
+  def onError(fa: F[A])(pf: PartialFunction[E, F[Unit]])(implicit F: ApplicativeError[F, E]): F[A] =
+    F.onError(fa)(pf)
 }
--- a/core/src/main/scala/cats/syntax/applicativeError.scala
+++ b/core/src/main/scala/cats/syntax/applicativeError.scala
@@ -35,6 +35,6 @@ final class ApplicativeErrorOps[F[_], E, A](val fa: F[A]) extends AnyVal {
   def recoverWith(pf: PartialFunction[E, F[A]])(implicit F: ApplicativeError[F, E]): F[A] =
     F.recoverWith(fa)(pf)
 
-  def onError(fa: F[A])(pf: PartialFunction[E, F[Unit]])(implicit F: ApplicativeError[F, E]): F[A] =
+  def onError(pf: PartialFunction[E, F[Unit]])(implicit F: ApplicativeError[F, E]): F[A] =
     F.onError(fa)(pf)
 }
--- a/core/src/main/scala/cats/syntax/monadError.scala
+++ b/core/src/main/scala/cats/syntax/monadError.scala
@@ -12,4 +12,7 @@ final class MonadErrorOps[F[_], E, A](val fa: F[A]) extends AnyVal {
 
   def ensureOr(error: A => E)(predicate: A => Boolean)(implicit F: MonadError[F, E]): F[A] =
     F.ensureOr(fa)(error)(predicate)
+
+  def adaptError(fa: F[A])(pf: PartialFunction[E, E])(implicit F: MonadError[F, E]): F[A] =
+    F.adaptError(fa)(pf)
 }
--- a/core/src/main/scala/cats/syntax/monadError.scala
+++ b/core/src/main/scala/cats/syntax/monadError.scala
@@ -13,6 +13,6 @@ final class MonadErrorOps[F[_], E, A](val fa: F[A]) extends AnyVal {
   def ensureOr(error: A => E)(predicate: A => Boolean)(implicit F: MonadError[F, E]): F[A] =
     F.ensureOr(fa)(error)(predicate)
 
-  def adaptError(fa: F[A])(pf: PartialFunction[E, E])(implicit F: MonadError[F, E]): F[A] =
+  def adaptError(pf: PartialFunction[E, E])(implicit F: MonadError[F, E]): F[A] =
     F.adaptError(fa)(pf)
 }
--- a/laws/src/main/scala/cats/laws/ApplicativeErrorLaws.scala
+++ b/laws/src/main/scala/cats/laws/ApplicativeErrorLaws.scala
@@ -40,11 +40,11 @@ trait ApplicativeErrorLaws[F[_], E] extends ApplicativeLaws[F] {
   def attemptFromEitherConsistentWithPure[A](eab: Either[E, A]): IsEq[F[Either[E, A]]] =
     F.attempt(F.fromEither(eab)) <-> F.pure(eab)
 
-  def onErrorPure[A, B](a: A, f: E => F[B]): IsEq[F[A]] =
-    F.onError(F.pure(a)){case err => F.void(f(err))} <-> F.pure(a)
+  def onErrorPure[A](a: A, f: E => F[Unit]): IsEq[F[A]] =
+    F.onError(F.pure(a))(PartialFunction(f)) <-> F.pure(a)
 
-  def onErrorRaise[A, B](fa: F[A], e: E, fb: F[B]): IsEq[F[A]] =
-    F.onError(F.raiseError[A](e)){case err => F.void(fb)} <-> F.map2(fb, F.raiseError[A](e))((_, b) => b)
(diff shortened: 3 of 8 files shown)


kailuowang:
assigning @edmundnoble as well since he is removing `ApplicativeError`

SystemFw:
I'm waiting for #1751 to be resolved to update this PR (probably just moving stuff to `MonadError`

core/src/main/scala/cats/syntax/applicativeError.scala
```diff
@@ -34,4 +34,7 @@ final class ApplicativeErrorOps[F[_], E, A](val fa: F[A]) extends AnyVal {
 
   def recoverWith(pf: PartialFunction[E, F[A]])(implicit F: ApplicativeError[F, E]): F[A] =
     F.recoverWith(fa)(pf)
+
+  def onError(fa: F[A])(pf: PartialFunction[E, F[Unit]])(implicit F: ApplicativeError[F, E]): F[A] =
+    F.onError(fa)(pf)
```
johnynek: this is untested.

core/src/main/scala/cats/syntax/monadError.scala
```diff
@@ -12,4 +12,7 @@ final class MonadErrorOps[F[_], E, A](val fa: F[A]) extends AnyVal {
 
   def ensureOr(error: A => E)(predicate: A => Boolean)(implicit F: MonadError[F, E]): F[A] =
     F.ensureOr(fa)(error)(predicate)
+
+  def adaptError(fa: F[A])(pf: PartialFunction[E, E])(implicit F: MonadError[F, E]): F[A] =
+    F.adaptError(fa)(pf)
```
johnynek: this is untested.

core/src/main/scala/cats/syntax/applicativeError.scala
SystemFw: Yes, I mentioned this in the description :)
> Haven't added non-discipline tests (yet?), they all seem to be based on Option, for which none of these two functions happen to make sense

Should I add a more complicated transformer to the tests to test this?

core/src/main/scala/cats/syntax/monadError.scala
SystemFw: Same as above

SystemFw:
@edmundnoble @kailuowang @djspiewak Any news on this one? I'm still looking for advice on the testing, but apart from that, this would be nice to have for use with cats-effect IO, now that ApplicativeError seems to be staying in cats.

core/src/main/scala/cats/syntax/applicativeError.scala
kailuowang: hmmm, since this is a simple delegation what about adding a simple doctest (maybe using `Either`) to the method in the type class?

SystemFw: Bear in mind that tests are missing for the implementation, not just for the syntax. I have added laws and discipline tests, but the reason I still haven't added unit tests is that all the unit tests I could find are based on Option, whereas testing this would require something like 
`EitherT[Writer, ....`, I don't have a problem with it, just wanting to know if it's a good idea

laws/src/main/scala/cats/laws/ApplicativeErrorLaws.scala
```diff
@@ -39,6 +39,12 @@ trait ApplicativeErrorLaws[F[_], E] extends ApplicativeLaws[F] {
 
   def attemptFromEitherConsistentWithPure[A](eab: Either[E, A]): IsEq[F[Either[E, A]]] =
     F.attempt(F.fromEither(eab)) <-> F.pure(eab)
+
+  def onErrorPure[A, B](a: A, f: E => F[B]): IsEq[F[A]] =
+    F.onError(F.pure(a)){case err => F.void(f(err))} <-> F.pure(a)
```
kailuowang: Not quite sure I follow here, why not ` def onErrorPure[A, B](a: A, f: E => F[Unit])....`?

kailuowang:
Also I think #1751 is close to merge. I think we can safely assume that `ApplicativeError` is going to stay.

core/src/main/scala/cats/syntax/applicativeError.scala
kailuowang: I assumed that the discipline tests provides enough coverage for the implementation, no? Anyway I am not against adding a `EitherT[Writer,...` to the unit tests

laws/src/main/scala/cats/laws/ApplicativeErrorLaws.scala
SystemFw: I've mentioned this in my comments, I couldn't get a `Gen[F[Unit]]`. However, it was my first time dealing with all the discipline infrastructure, and never had to to this particular thing in my own experience with scalacheck, so glad to be proven wrong :)

core/src/main/scala/cats/syntax/applicativeError.scala
SystemFw: Well, my idea was exactly that discipline tests might be enough, which is way I submitted this in its current form. Looking for advice from the maintainers to take the final decision

kailuowang: yeah, I think @johnynek and I were saying that ideally we also want to cover the delegation in the syntax, and one way to do that easily is to add a doctest in the type class method declaration, which uses the syntax and can also serve as an example in the scala doc.

laws/src/main/scala/cats/laws/ApplicativeErrorLaws.scala
kailuowang: just add a  `ArbFU: Arbitrary[F[Unit]]` param to the `MonadErrorTests` and `ApplicativeErrorTest` methods should do the trick,

SystemFw: I think this is exactly what I did and iirc it didn't compile. Let me give it another try though :)

kailuowang: compiled for me, EitherTests passes. running full now.

SystemFw: Yes you're right. I hadn't realised I needed to add it to the MonadErrorTests as well (because of `parent`). Thank you! (I'll update the PR tonight with the change)

kailuowang: You are welcome. Thank you for contributing!

core/src/main/scala/cats/syntax/applicativeError.scala
SystemFw: I can't see any doctests for the other methods on ApplicativeError, apart from a scaladoc example for `fromEither`.  Would you happen to have an example at hand of how syntax forwarders are tested in cats via doctests  ?

kailuowang: Sure, here is one https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Apply.scala#L54

SystemFw: I'll add these today, especially since there's a bug :grin: (the `fa` should not be a parameter, shadows the implicit class parameter)

SystemFw:
@kailuowang I've simplified the discipline tests to take `F[Unit]`, and added doctests for `onError` and `applicativeError`, let me know if there's any more feedback

edmundnoble:
Merging with two approvals.

PR #1740, opened 2017-06-23
remove iteratorFoldM fixes #1716

### LINKED ISSUE #1716: `Foldable` instance for `List` breaks substitution
The `Foldable` instance for `List` breaks substitution, as the following program demonstrates. The value of `result.value` differs between the first and subsequent calls.

```scala
import cats.instances.list._
import cats.Eval
import cats.syntax.foldable._

val result  = List.range(0,10).foldM(List.empty[Int])((accum, elt) => Eval.always(elt :: accum))

result.value
// res: List[Int] = List(9, 8, 7, 6, 5, 4, 3, 2, 1, 0)

result.value
// res: List[Int] = List(0)
```

I haven't had time to debug but I suspect it relates to the use of a mutable `ListBuffer` in the implementation of `tailRecM`.

kailuowang:
also related https://github.com/typelevel/cats/pull/1117/files#r123765689

### DIFF
--- a/core/src/main/scala/cats/data/NonEmptyList.scala
+++ b/core/src/main/scala/cats/data/NonEmptyList.scala
@@ -431,6 +431,9 @@ private[data] sealed trait NonEmptyListInstances extends NonEmptyListInstances0 
       override def fold[A](fa: NonEmptyList[A])(implicit A: Monoid[A]): A =
         fa.reduce
 
+      override def foldM[G[_], A, B](fa: NonEmptyList[A], z: B)(f: (B, A) => G[B])(implicit G: Monad[G]): G[B] =
+        G.flatMap(f(z, fa.head))(bnext => Foldable[List].foldM(fa.tail, bnext)(f))
+
       override def find[A](fa: NonEmptyList[A])(f: A => Boolean): Option[A] =
         fa find f
 
--- a/core/src/main/scala/cats/data/NonEmptyList.scala
+++ b/core/src/main/scala/cats/data/NonEmptyList.scala
@@ -431,9 +431,6 @@ private[data] sealed trait NonEmptyListInstances extends NonEmptyListInstances0 
       override def fold[A](fa: NonEmptyList[A])(implicit A: Monoid[A]): A =
         fa.reduce
 
-      override def foldM[G[_], A, B](fa: NonEmptyList[A], z: B)(f: (B, A) => G[B])(implicit G: Monad[G]): G[B] =
-        G.flatMap(f(z, fa.head))(bnext => Foldable[List].foldM(fa.tail, bnext)(f))
-
       override def find[A](fa: NonEmptyList[A])(f: A => Boolean): Option[A] =
         fa find f
 
--- a/core/src/main/scala/cats/data/NonEmptyList.scala
+++ b/core/src/main/scala/cats/data/NonEmptyList.scala
@@ -431,9 +431,6 @@ private[data] sealed trait NonEmptyListInstances extends NonEmptyListInstances0 
       override def fold[A](fa: NonEmptyList[A])(implicit A: Monoid[A]): A =
         fa.reduce
 
-      override def foldM[G[_], A, B](fa: NonEmptyList[A], z: B)(f: (B, A) => G[B])(implicit G: Monad[G]): G[B] =
-        Foldable.iteratorFoldM(fa.toList.toIterator, z)(f)
-
       override def find[A](fa: NonEmptyList[A])(f: A => Boolean): Option[A] =
         fa find f
 
--- a/core/src/main/scala/cats/data/NonEmptyList.scala
+++ b/core/src/main/scala/cats/data/NonEmptyList.scala
@@ -428,11 +428,11 @@ private[data] sealed trait NonEmptyListInstances extends NonEmptyListInstances0 
         NonEmptyList.fromListUnsafe(buf.result())
       }
 
-      override def fold[A](fa: NonEmptyList[A])(implicit A: Monoid[A]): A =
-        fa.reduce
-
       override def foldM[G[_], A, B](fa: NonEmptyList[A], z: B)(f: (B, A) => G[B])(implicit G: Monad[G]): G[B] =
         G.flatMap(f(z, fa.head))(bnext => Foldable[List].foldM(fa.tail, bnext)(f))
+
+      override def fold[A](fa: NonEmptyList[A])(implicit A: Monoid[A]): A =
+        fa.reduce
 
       override def find[A](fa: NonEmptyList[A])(f: A => Boolean): Option[A] =
         fa find f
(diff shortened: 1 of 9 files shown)


tests/src/test/scala/cats/tests/FoldableTests.scala
```diff
@@ -214,17 +209,23 @@ class FoldableTestsAdditional extends CatsSuite {
   }
 
   test(".foldLeftM short-circuiting optimality") {
-    // test that no more elements are evaluated than absolutely necessary
+    // test that no more than 1 elements are evaluated than absolutely necessary
```
peterneyens: Should we look if we can solve this additional evaluation with @TomasMikula's [suggestion](https://github.com/typelevel/cats/pull/1117/files#r123775020)?

kailuowang: updated.

kailuowang:
weird, for some reason github doesn't allow me to request @TomasMikula to review.

tests/src/test/scala/cats/tests/FoldableTests.scala
peterneyens: You forgot to update this comment.

TomasMikula:
👍

However, it's still the case that most data structures can have a more efficient implementation of `foldM` than the default one from `Foldable`.

kailuowang:
@TomasMikula yes, it's nice that you also mentioned it in the comment of `foldM`. we'll get to it in a later PR(s).

core/src/main/scala/cats/instances/list.scala
```diff
@@ -90,9 +90,6 @@ trait ListInstances extends cats.kernel.instances.ListInstances {
 
       override def filter[A](fa: List[A])(f: A => Boolean): List[A] = fa.filter(f)
 
-      override def foldM[G[_], A, B](fa: List[A], z: B)(f: (B, A) => G[B])(implicit G: Monad[G]): G[B] =
-        Foldable.iteratorFoldM(fa.toIterator, z)(f)
```
johnynek: can't we maybe do this one?

```scala
def step(in: (List[A], B])): G[Either[(List[A], B), A]] = in match {
  case (Nil, b) => G.pure(Right(b))
  case (a :: tail, b) => G.map(f(a, b)) { bnext => Left((tail, bnext)) }
} 

G.tailRecM((fa, z))(step)
```
right? If we had this, we could cover the common case for `List` and `NonEmptyList`, no?

core/src/main/scala/cats/instances/stream.scala
```diff
@@ -118,9 +118,6 @@ trait StreamInstances extends cats.kernel.instances.StreamInstances {
 
       override def collect[A, B](fa: Stream[A])(f: PartialFunction[A, B]): Stream[B] = fa.collect(f)
 
-      override def foldM[G[_], A, B](fa: Stream[A], z: B)(f: (B, A) => G[B])(implicit G: Monad[G]): G[B] =
-        Foldable.iteratorFoldM(fa.toIterator, z)(f)
```
johnynek: we could copy the `List` approach above here:
```scala
def step(in: (Stream[A], B])): G[Either[(Stream[A], B), A]] = in match {
  case (Stream.empty, b) => G.pure(Right(b))
  case (a #:: tail, b) => G.map(f(a, b)) { bnext => Left((tail, bnext)) }
} 

G.tailRecM((fa, z))(step)
```

core/src/main/scala/cats/instances/vector.scala
```diff
@@ -90,9 +90,6 @@ trait VectorInstances extends cats.kernel.instances.VectorInstances {
 
       override def collect[A, B](fa: Vector[A])(f: PartialFunction[A, B]): Vector[B] = fa.collect(f)
 
-      override def foldM[G[_], A, B](fa: Vector[A], z: B)(f: (B, A) => G[B])(implicit G: Monad[G]): G[B] =
-        Foldable.iteratorFoldM(fa.toIterator, z)(f)
```
johnynek: `Vector.tail` is so slow, we are probably better off doing `fa.toList.foldM` here if we do the above.

core/src/main/scala/cats/instances/stream.scala
kailuowang: Patten match Stream into `a #:: tail` will force an evaluation of the head of the `tail`, so I used vanilla `if`

kailuowang:
@johnynek thanks for helping me with these, it was too lazy of me.

core/src/main/scala/cats/instances/stream.scala
```diff
@@ -118,8 +119,17 @@ trait StreamInstances extends cats.kernel.instances.StreamInstances {
...
+      override def foldM[G[_], A, B](fa: Stream[A], z: B)(f: (B, A) => G[B])(implicit G: Monad[G]): G[B] = {
```
TomasMikula: This is good, but now again the default implementation of `foldM` in `Foldable` is not tested.

kailuowang: oops. I'll correct the test.

kailuowang: updated.

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -431,6 +431,9 @@ private[data] sealed trait NonEmptyListInstances extends NonEmptyListInstances0
       override def fold[A](fa: NonEmptyList[A])(implicit A: Monoid[A]): A =
         fa.reduce
 
+      override def foldM[G[_], A, B](fa: NonEmptyList[A], z: B)(f: (B, A) => G[B])(implicit G: Monad[G]): G[B] =
+        G.flatMap(f(z, fa.head))(bnext => Foldable[List].foldM(fa.tail, bnext)(f))
+
```
peterneyens: `NonEmptyList` sort of gets this implementation already through `NonEmptyReducible`, but I guess this saves us the `split`.

core/src/main/scala/cats/instances/stream.scala
peterneyens: Here we would profit if had some machinery mentioned in #1684.

core/src/main/scala/cats/instances/stream.scala
```diff
@@ -118,8 +119,17 @@ trait StreamInstances extends cats.kernel.instances.StreamInstances {
...
+        def step(in: (Stream[A], B)): G[Either[(Stream[A], B), B]] = {
+          val (s, b)  = in
+          if(s.isEmpty)
```
peterneyens: Scalastyle seems to want a space between the `if` and the `(`.

core/src/main/scala/cats/data/NonEmptyList.scala
kailuowang: ah, forgot about that. A single split call probably doesn't justify the code duplication. I am going to remove this.

core/src/main/scala/cats/instances/stream.scala
johnynek: @TomasMikula I think we are still tested with `Map` and `Set` but I agree tests for overrides would be good.

kailuowang: yup. I thought about that too.

johnynek:
👍

kailuowang:
@peterneyens any other issues/feedback?

PR #1744, opened 2017-06-27
Make links link to the .html files instead of .md

LukaJCB:


LukaJCB:
Not sure why CI fails here, probably a mistake on Travis side, can someone restart the job?

PR #1745, opened 2017-06-27
Deprecate CartesianBuilder, finish up #1487

kailuowang:
This PR is to finish up #1487. I deprecated the `CartesianBuilder`, merged the master and one small correction in the build file.

### DIFF
--- a/.jvmopts
+++ b/.jvmopts
@@ -1,4 +1,3 @@
-# see https://weblogs.java.net/blog/kcpeppe/archive/2013/12/11/case-study-jvm-hotspot-flags
 -Dfile.encoding=UTF8
 -Xms1G
 -Xmx6G
@@ -6,7 +5,5 @@
 -XX:ReservedCodeCacheSize=250M
 -XX:+TieredCompilation
 -XX:-UseGCOverheadLimit
-# effectively adds GC to Perm space
 -XX:+CMSClassUnloadingEnabled
-# must be enabled for CMSClassUnloadingEnabled to work
 -XX:+UseConcMarkSweepGC
--- a/build.sbt
+++ b/build.sbt
@@ -67,7 +67,7 @@ lazy val commonJsSettings = Seq(
   scalaJSStage in Global := FastOptStage,
   parallelExecution := false,
   requiresDOM := false,
-  jsEnv := NodeJSEnv().value,
+  jsEnv := new org.scalajs.jsenv.nodejs.NodeJSEnv(),
   // Only used for scala.js for now
   botBuild := scala.sys.env.get("TRAVIS").isDefined,
   // batch mode decreases the amount of memory needed to compile scala.js code
--- a/core/src/main/scala/cats/data/EitherT.scala
+++ b/core/src/main/scala/cats/data/EitherT.scala
@@ -175,7 +175,7 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
    * scala> val v1: Validated[NonEmptyList[Error], Int] = Validated.invalidNel("error 1")
    * scala> val v2: Validated[NonEmptyList[Error], Int] = Validated.invalidNel("error 2")
    * scala> val eithert: EitherT[Option, Error, Int] = EitherT.leftT[Option, Int]("error 3")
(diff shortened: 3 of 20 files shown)


build.sbt
```diff
@@ -67,7 +67,7 @@ lazy val commonJsSettings = Seq(
   scalaJSStage in Global := FastOptStage,
   parallelExecution := false,
   requiresDOM := false,
-  jsEnv := NodeJSEnv().value,
+  jsEnv := new org.scalajs.jsenv.nodejs.NodeJSEnv(),
```
kailuowang: old one is deprecated.

.jvmopts
```diff
@@ -1,12 +1,9 @@
...
 -Dfile.encoding=UTF8
 -Xms1G
 -Xmx6G
 -XX:MaxPermSize=512M
 -XX:ReservedCodeCacheSize=250M
 -XX:+TieredCompilation
 -XX:-UseGCOverheadLimit
-# effectively adds GC to Perm space
 -XX:+CMSClassUnloadingEnabled
-# must be enabled for CMSClassUnloadingEnabled to work
```
kailuowang: @DavidGregory084 pointed out that this causes problem on Windows.

core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -175,7 +175,7 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
    * scala> val v1: Validated[NonEmptyList[Error], Int] = Validated.invalidNel("error 1")
    * scala> val v2: Validated[NonEmptyList[Error], Int] = Validated.invalidNel("error 2")
    * scala> val eithert: EitherT[Option, Error, Int] = EitherT.leftT[Option, Int]("error 3")
-   * scala> eithert.withValidated { v3 => (v1 |@| v2 |@| v3.toValidatedNel).map { case (i, j, k) => i + j + k } }
+   * scala> eithert.withValidated { v3 => (v1, v2, v3.leftMap(NonEmptyList.of(_))).mapN { case (i, j, k) => i + j + k } }
```
peterneyens: I think we can keep the `toValidatedNel` (instead of `.leftMap(NonEmptyList.of(_))`)

core/src/main/scala/cats/data/Tuple2K.scala
```diff
@@ -160,7 +160,7 @@ private[data] sealed trait Tuple2KTraverse[F[_], G[_]] extends Traverse[λ[α =>
   def G: Traverse[G]
 
   override def traverse[H[_]: Applicative, A, B](fa: Tuple2K[F, G, A])(f: A => H[B]): H[Tuple2K[F, G, B]] =
-    (F.traverse(fa.first)(f) |@| G.traverse(fa.second)(f)).map(Tuple2K(_, _))
+    (F.traverse(fa.first)(f), G.traverse(fa.second)(f)).mapN(Tuple2K(_, _))
```
peterneyens: Maybe we can just write `H.map2` instead of the syntax version?

core/src/main/scala/cats/data/EitherT.scala
kailuowang: missed that during the merge. will fix.

kailuowang:
@johnynek please allow me to reply you here since any further change is probably going to be through this PR. I think @non first suggested the possibility of deprecating the `|@|` syntax once the tuple syntax is proven [here](https://github.com/typelevel/cats/pull/1112#issuecomment-226190365). There wasn't any objections ATM.  The cartesian syntax on tuple has been released since 0.7.0 for almost a year, I think it's about time to deprecate the `|@|` syntax. Unlike `Xor`, this change should be rather straightforward and mechanical (I just did it in cats code base, took me 10 minutes). If we still want to keep it, for how long?

johnynek:
Okay. Thanks for linking to the relevant discussion.

kailuowang:
@peterneyens I think I addressed all your comments, have time to take another look?

kailuowang:
merge with 2 sign-offs given this has been discussed more in the previous related PRs.

fosskers:
If `|@|` is being deprecated, is the tuple thing the new recommended syntax for Applicative-style?

kailuowang:
@fosskers yes. I gave some examples in the migration doc. Let me know if that works for you.  https://github.com/kailuowang/cats/blob/8df835fe4700455135140c2e3d02b6555328985c/CHANGES.md

PR #1746, opened 2017-06-27
Add EitherNel type alias for Either[NonEmptyList[E], A]

andyscott:
Often when doing validations using `ValidatedNel`, the result is converted to `Either` when returning to a higher level where steps are sequenced. It'd be convenient to represent this type as `EitherNel` instead of `Either[NonEmptyList[...`


johnynek:
👍

kailuowang:
merge with 2 sign-offs

PR #1751, opened 2017-06-30
MTL is dead, long live MTL

edmundnoble:
This PR removes the types ApplicativeError, FunctorFilter, MonadCombine, MonadFilter, MonadReader, MonadState, MonadTrans, MonadWriter, TraverseFilter.

It also replaces MonadCombine's operations with equivalents on Alternative (no laws for them anyway), moves methods on ApplicativeError to MonadError; all other functionality will be moved to cats-mtl if it is already not present.

Sorry about the length. There were some missing Alternative instances I had to add as well to make users of MonadCombine less unhappy about the change.


SystemFw:
Just  a thought: in the proposed design, the `ApplicativeError` methods will no longer be available for `Validated`. There are specialised variants, but `F: MonadError` will no longer be able to unify with `Validated`, which is unfortunate.

johnynek:
I'm -1 on removing ApplicativeError.

What is the motivation for this?

johnynek:
yeah, and we could have a `ValidatedT` I think, which could give us an `ApplicativeError` on any `Applicative` (though I haven't bothered to try to implement to be sure there is not some corner case I'm overlooking.

adelbertc:
What would `ValidatedT` give over `Nested[F, Validated[E, ?], ?]` ?

johnynek:
Nothing. Other than a smaller type I guess. I just don't use Nested much so
it wasn't on the tip of my tongue.

johnynek:
wait, can't we make an `ApplicativeError` with a `Validated[E, A]` and an `Applicative[A]`?

seems like you can, but I didn't write all the methods. Why do you think you can't? Anything concrete?

johnynek:
okay, I guess you can't:
```scala
case class ValidatedT[F[_], E, A](f: F[Validated[E, A]])

class ValidatedTApplicativeError[F[_]: Applicative, E] extends ApplicativeError[ValidatedT[F, E, ?]] {
  def pure[A](a: A): ValidatedT(Applicative[F].pure(Validated.valid(a)))
  def raiseError[A](e: E): ValidateT(Applicative[F].pure(Validated.invalid(e)))
  def handleErrorWith[A](fa: ValidatedT[F, E, A])(f: E => ValidateT[F, E, A]): ValidatedT[F, E, A] =
     // here we can't look inside F[Validate[E, A]] to see if have an error
     // we need a monad on F to do this, it seems.
}
```
`Validated[E, F[A]]` can be an applicativeError:
```scala
class ValidatedTApplicativeError[F[_]: Applicative, E] extends ApplicativeError[Lambda[T => Validated[E, F[T]]]] {
  def pure[A](a: A): Validated.valid(Applicative[F].pure(a))
  def raiseError[A](e: E) = Validated.invalid(e)
  def handleErrorWith[A](fa: Validated[E, F[A]])(f: E => Validated[E, F[A]]): Validated[E, F[A]] =
     fa match {
       case v@Validated.Valid(_) =>
       case Validated.Invalid(e) => f(e)
     }
}
```
so, you can't hide the error inside the F.

kailuowang:
It didn't come to my mind that the removal of `ApplicativeError` will impact `Validated` usage. I am curious about the motivation too. Is it still a good trade-off?

edmundnoble:
You've hit the nail on the head here, basically. `ApplicativeError` is too large an abstraction to lift through a transformer stack. Thus, as an abstraction, the only use it has is writing code that works with an `EitherT`-containing applicative transformer stack *or* just a single `Validated` on the outside with the call site deciding.

This is why cats-mtl splits it into two classes: `ApplicativeHandle` and `ApplicativeRaise`, the latter of which is possible to lift through arbitrary `Applicative` transformers.

@kailuowang I have already moved the methods to `MonadError`; the only step remaining to glue over `ApplicativeError` is to add them to `Validated`.

My motivation for removing `ApplicativeError` is that cats-mtl includes Applicative-based transformer type classes, which puts it on equal footing with `MonadError` when it comes to removal; except `MonadError` has subtypes, and `ApplicativeError` doesn't.

It's not a large issue to keep `ApplicativeError` IMO though; it's just odd that we'd treat errors differently than other MTL classes.

edmundnoble:
To avoid bogging this PR down, I'm taking a vote here: :+1: to keep `ApplicativeError`, :-1: to take it out.

core/src/main/scala/cats/data/WriterT.scala
```diff
@@ -184,14 +184,21 @@ private[data] sealed abstract class WriterTInstances8 extends WriterTInstances9
...
+private[data] sealed abstract class WriterTInstances9 extends WriterTInstances10 {
   implicit def catsDataMonadErrorForWriterT[F[_], L, E](implicit F: MonadError[F, E], L: Monoid[L]): MonadError[WriterT[F, L, ?], E] =
     new WriterTMonadError[F, L, E] {
       implicit val F0: MonadError[F, E] = F
       implicit val L0: Monoid[L] = L
     }
 }
 
+private[data] sealed abstract class WriterTInstances10 {
+  implicit def catsDataFunctorForWriterT[F[_], L](implicit F: Functor[F]): Functor[WriterT[F, L, ?]] =
```
kailuowang: Not sure why this is needed. The [`Coflatmap` instance](https://github.com/edmundnoble/cats/blob/1f460fab573ad4150885743bfbd5e2f1b7eea466/core/src/main/scala/cats/data/WriterT.scala#L177-L180) should surfice, no?

kailuowang:
If it's not big deal, I vote for keeping the `ApplicativeError` which seems less friction and less burden for user migration. 

As a related note, I don't have a strong argument for it but I start to feel that the error effect might be worth some special treatment. We did that for `MonadError`.

edmundnoble:
I don't particularly feel myself like errors deserve any special treatment, however I observe that MonadError does not add any laws to ApplicativeError and it is okay to special-case *one* MTL class and still avoid ambiguity.

edmundnoble:
@andyscott See https://github.com/edmundnoble/cats-mtl/issues/10.

core/src/main/scala/cats/data/WriterT.scala
edmundnoble: Fixed.

free/src/main/scala/cats/free/FreeT.scala
```diff
@@ -204,39 +193,36 @@ private[free] sealed trait FreeTInstances2 extends FreeTInstances3 {
...
 
-private[free] sealed trait FreeTInstances extends FreeTInstances0 {
-  implicit def catsFreeMonadCombineForFreeT[S[_], M[_]: Alternative]: MonadCombine[FreeT[S, M, ?]] =
-    new MonadCombine[FreeT[S, M, ?]] with FreeTCombine[S, M] with FreeTMonad[S, M] {
-      override def M = implicitly
-      override def M1 = implicitly
-
-      override def empty[A] = FreeT.liftT[S, M, A](MonoidK[M].empty[A])(M)
+private[free] sealed trait FreeTInstances3 {
+  implicit def catsFreeSemigroupKForFreeT[S[_], M[_]: Applicative: MonoidK]: SemigroupK[FreeT[S, M, ?]] =
```
kailuowang: `: Applicative: SemigroupK` ?

free/src/main/scala/cats/free/FreeT.scala
```diff
@@ -204,39 +193,36 @@ private[free] sealed trait FreeTInstances2 extends FreeTInstances3 {
...
 }
 
-private[free] sealed trait FreeTInstances0 extends FreeTInstances1 {
-  implicit def catsFreeMonadForFreeT[S[_], M[_]](implicit M0: Applicative[M]): Monad[FreeT[S, M, ?]] =
-    new FreeTMonad[S, M] {
-      def M = M0
-    }
+private[free] sealed trait FreeTInstances2 extends FreeTInstances3 {
+  implicit def catsFreeAlternativeForFreeT[S[_], M[_]: Alternative: Monad]: Alternative[FreeT[S, M, ?]] =
+    new Alternative[FreeT[S, M, ?]] with FreeTMonad[S, M] with FreeTSemigroupK[S, M] {
```
kailuowang: `with FreeTSemigroupK` -> `with FreeTMonoidK` ?

free/src/main/scala/cats/free/FreeT.scala
```diff
@@ -204,39 +193,36 @@ private[free] sealed trait FreeTInstances2 extends FreeTInstances3 {
...
+      override def M = implicitly[Alternative[M]]
+      override def M1 = implicitly[Alternative[M]]
 
-  implicit def catsFreeCombineForFreeT[S[_], M[_]: Applicative: SemigroupK]: SemigroupK[FreeT[S, M, ?]] =
-    new FreeTCombine[S, M] {
-      override def M = implicitly
-      override def M1 = implicitly
+      def ap[A, B](fa: FreeT[S, M, A])(ff: FreeT[S, M, A => B]): FreeT[S, M, B] = fa.flatMap(a => ff.map(_(a))(M))
+      override final def empty[A]: FreeT[S, M, A] = FreeT.liftT[S, M, A](M1.empty[A])(M)
```
kailuowang: Not needed if extend `with FreeTMonoidK`

free/src/test/scala/cats/free/FreeTTests.scala
```diff
@@ -27,15 +27,15 @@ class FreeTTests extends CatsSuite {
   }
 
   {
-    implicit val freeTSemigroupK: SemigroupK[FreeTOption] = FreeT.catsFreeCombineForFreeT[Option, Option]
-    checkAll("FreeT[Option, Option, Int]", SemigroupKTests[FreeTOption].semigroupK[Int])
+    implicit val freeTSemigroupK: SemigroupK[FreeTOption] = FreeT.catsFreeSemigroupKForFreeT[Option, Option]
+    checkAll("FreeT[Option, Option, Int]", MonoidKTests[FreeTOption].monoidK[Int])
```
kailuowang: `SemigroupKTests` here?

edmundnoble:
@kailuowang Comments addressed, thanks.

kailuowang:
thanks @edmundnoble I haven't gone through all of it yet. Will continue tomorrow.

tests/src/test/scala/cats/tests/ApplicativeTests.scala
```diff
@@ -4,7 +4,7 @@ package tests
 import cats.Applicative
 
 
-class ApplicativeTests extends CatsSuite {
+class ApplicativeCheck extends CatsSuite {
```
kailuowang: Not sure why this rename? I noticed that `ApplicativeErrorCheck` is also using the `Check` suffix. But that looks like an outlier? The vast majority have the suffix `Tests`.  There is also `MonadErrorSuite`, `BitraverseTest`, `CategoryTest`, `ComposeTest`, `ContravariantTest`, `FoldableCheck`, `ReducableCheck`, `FunctorTest`, `MonadTest`. If you leave  this one alone, I can submit another PR to unify them.

tests/src/test/scala/cats/tests/OneAndTests.scala
```diff
@@ -180,6 +174,13 @@ class OneAndTests extends CatsSuite {
       got should === (expected)
     }
   }
+
+  test("filter includes elements based on a predicate") {
+    forAll { (nes: NonEmptyStream[Int]) =>
+      nes.filter(_ % 2 == 0) should ===(nes.unwrap.filter(_ % 2 == 0))
+    }
+  }
+
```
kailuowang: Just curious, was the arbitrary function in the original test causing problems?

tests/src/test/scala/cats/tests/ApplicativeTests.scala
edmundnoble: Will do, good point. I didn't know about `MonadErrorSuite` or the rest of the `Check`-style suites.

tests/src/test/scala/cats/tests/OneAndTests.scala
edmundnoble: I don't quite know what you mean. The original test in this case is `MonadCombineTests`, which has been removed, but I did needlessly use `_ % 2 == 0` instead of an arbitrary function; that's been corrected now.

kailuowang: Sorry I didn't make it clear. "needlessly use _ % 2 == 0 instead of an arbitrary function" is exactly what I intended to point out.

edmundnoble: Ah makes sense :)

alexandru:
👍 on this pull request.

I do have a concern about `cats-mtl`: projects are going to block for it and given that it's a totally new project that hasn't been tested, I wouldn't introduce experimental functionality, only the essentials, as in, only what's being removed here + eventual changes to encoding in order to ensure forward compatibility.

core/src/main/scala/cats/data/WriterT.scala
```diff
@@ -127,8 +109,8 @@ private[data] sealed abstract class WriterTInstances1 extends WriterTInstances2
 }
 
 private[data] sealed abstract class WriterTInstances2 extends WriterTInstances3 {
-  implicit def catsDataMonadWriterForWriterT[F[_], L](implicit F: Monad[F], L: Monoid[L]): MonadWriter[WriterT[F, L, ?], L] =
-    new WriterTMonadWriter[F, L] {
+  implicit def catsDataMonadForWriterT[F[_], L](implicit F: Monad[F], L: Monoid[L]): Monad[WriterT[F, L, ?]] =
```
peterneyens: This can be moved to `WriterTInstances0`, which means that the `Alternative`, `Applicative`, `FlatMap`, ... instances can be moved up as well.

free/src/main/scala/cats/free/FreeT.scala
```diff
@@ -182,18 +182,7 @@ object FreeT extends FreeTInstances {
...
-      override def M = implicitly
-      override def get =
-        FreeT.liftT(M1.get)
-      override def set(s: E) =
-        FreeT.liftT(M1.set(s))
-    }
-}
-
-private[free] sealed trait FreeTInstances2 extends FreeTInstances3 {
+sealed trait FreeTInstances extends FreeTInstances0 {
```
peterneyens: Should we leave this as `private[free]` ?

free/src/main/scala/cats/free/FreeT.scala
```diff
@@ -204,39 +193,35 @@ private[free] sealed trait FreeTInstances2 extends FreeTInstances3 {
...
 
-private[free] sealed trait FreeTInstances0 extends FreeTInstances1 {
-  implicit def catsFreeMonadForFreeT[S[_], M[_]](implicit M0: Applicative[M]): Monad[FreeT[S, M, ?]] =
-    new FreeTMonad[S, M] {
-      def M = M0
-    }
+private[free] sealed trait FreeTInstances2 extends FreeTInstances3 {
+  implicit def catsFreeAlternativeForFreeT[S[_], M[_]: Alternative: Monad]: Alternative[FreeT[S, M, ?]] =
+    new Alternative[FreeT[S, M, ?]] with FreeTMonad[S, M] with FreeTMonoidK[S, M] {
+      override def M = implicitly[Alternative[M]]
```
peterneyens: I think we prefer `Alternative[M]` over `implicitly[Alternative[M]]`.

free/src/main/scala/cats/free/Yoneda.scala
```diff
@@ -35,7 +35,7 @@ object Yoneda {
   /**
    * `Yoneda[F, _]` is a functor for any `F`.
    */
-  implicit def catsFreeFunctorForYoneda[F[_]]: Functor[Yoneda[F, ?]] =
+  implicit def catsCofreeFunctorForYoneda[F[_]]: Functor[Yoneda[F, ?]] =
```
peterneyens: I think the `Free` here is the package (like we have `catsDataFunctorForKleisli`.

tests/src/test/scala/cats/tests/NestedTests.scala
```diff
@@ -18,7 +18,7 @@ class NestedTests extends CatsSuite {
 
   {
     // Invariant composition
-    implicit val instance = ListWrapper.invariant
+    implicit val instance = Nested.catsDataInvariantForNested(ListWrapper.invariant, ListWrapper.invariant)
```
peterneyens: Is there a reason that we need to create these instances for `Nested` (here and below) explicitly?

core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -535,6 +539,18 @@ private[data] trait EitherTFunctor[F[_], L] extends Functor[EitherT[F, L, ?]] {
   override def map[A, B](fa: EitherT[F, L, A])(f: A => B): EitherT[F, L, B] = fa map f
 }
 
+private[data] trait EitherTApply[F[_], L] extends EitherTFunctor[F, L] with Apply[EitherT[F, L, ?]] {
```
peterneyens: I think we need to drop these instances before we merge, all my other comments are things which can be cleaned up.

edmundnoble: This is *very* interesting, and will likely lead to some corrections in cats-mtl as it stands.

core/src/main/scala/cats/data/Kleisli.scala
```diff
@@ -158,11 +151,21 @@ private[data] sealed abstract class KleisliInstances4 extends KleisliInstances5
...
 }
 
-private[data] sealed abstract class KleisliInstances5 {
+private[data] sealed abstract class KleisliInstances5 extends KleisliInstances6 {
   implicit def catsDataFunctorForKleisli[F[_], A](implicit F0: Functor[F]): Functor[Kleisli[F, A, ?]] =
     new KleisliFunctor[F, A] { def F: Functor[F] = F0 }
 }
 
+private[data] sealed abstract class KleisliInstances6 extends KleisliInstances7 {
+  implicit def catsDataApplicativeErrorForKleisli[F[_], E, A](implicit F0: ApplicativeError[F, E]): ApplicativeError[Kleisli[F, A, ?], E] =
```
edmundnoble: :+1:

core/src/main/scala/cats/data/ReaderWriterStateT.scala
```diff
@@ -478,14 +434,21 @@ private[data] sealed trait RWSTSemigroupK[F[_], E, S, L] extends SemigroupK[Read
     }
 }
 
-private[data] sealed trait RWSTMonadCombine[F[_], E, S, L]
-    extends MonadCombine[ReaderWriterStateT[F, E, S, L, ?]] with RWSTMonad[F, E, S, L]
-    with RWSTSemigroupK[F, E, S, L] with RWSTMonadTrans[E, S, L] {
+private[data] sealed trait RWSTAlternative[F[_], E, S, L]
+    extends Alternative[ReaderWriterStateT[F, E, S, L, ?]]
+    with RWSTSemigroupK[F, E, S, L] {
```
edmundnoble: :+1:

core/src/main/scala/cats/data/ReaderWriterStateT.scala
```diff
@@ -374,32 +362,14 @@ private[data] sealed trait RWSTInstances2 extends RWSTInstances3 {
...
-  implicit def catsDataMonadWriterForRWST[F[_], E, S, L](
-    implicit F0: Monad[F], L0: Monoid[L]): MonadWriter[ReaderWriterStateT[F, E, S, L, ?], L] =
-    new RWSTMonadWriter[F, E, S, L] {
-      implicit def F: Monad[F] = F0
-      implicit def L: Monoid[L] = L0
-    }
-}
-
-private[data] sealed trait RWSTInstances5 extends RWSTInstances6 {
   implicit def catsDataMonadForRWST[F[_], E, S, L](implicit F0: Monad[F], L0: Monoid[L]): Monad[ReaderWriterStateT[F, E, S, L, ?]] =
```
edmundnoble: :+1:

free/src/main/scala/cats/free/FreeT.scala
edmundnoble: :+1:

free/src/main/scala/cats/free/FreeT.scala
edmundnoble: :+1:

free/src/main/scala/cats/free/Yoneda.scala
edmundnoble: :+1:

tests/src/test/scala/cats/tests/NestedTests.scala
edmundnoble: I'll check again, I had a coverage drop which was fixed by this but I think there was an underlying problem I fixed.

edmundnoble:
All feedback should be addressed.

tests/src/test/scala/cats/tests/WriterTTests.scala
```diff
@@ -355,9 +310,9 @@ class WriterTTests extends CatsSuite {
   }
 
   {
-    // F has an ApplicativeError and L has a Monoid
+    // F has an Applicative and L has a Monoid
     implicit val L: Monoid[ListWrapper[Int]] = ListWrapper.monoid[Int]
-    implicit val appErr = WriterT.catsDataApplicativeErrorForWriterT[Validated[String, ?], ListWrapper[Int], String]
+    implicit val appErr = WriterT.catsDataApplicativeForWriterT[Validated[String, ?], ListWrapper[Int]]
```
peterneyens: Minor, should we call this `app` instead of `appErr` ?

free/src/main/scala/cats/free/FreeT.scala
```diff
@@ -182,61 +182,46 @@ object FreeT extends FreeTInstances {
...
+  implicit def catsFreeAlternativeForFreeT[S[_], M[_]: Alternative: Monad]: Alternative[FreeT[S, M, ?]] =
+    new Alternative[FreeT[S, M, ?]] with FreeTMonad[S, M] with FreeTMonoidK[S, M] {
+      override def M = Alternative[M]
+      override def M1 = Alternative[M]
 
-  implicit def catsFreeCombineForFreeT[S[_], M[_]: Applicative: SemigroupK]: SemigroupK[FreeT[S, M, ?]] =
-    new FreeTCombine[S, M] {
-      override def M = implicitly
-      override def M1 = implicitly
+      def ap[A, B](fa: FreeT[S, M, A])(ff: FreeT[S, M, A => B]): FreeT[S, M, B] = fa.flatMap(a => ff.map(_(a))(M))
```
peterneyens: I don't think we need this, right?

With `override` this should give a compiler error because `Apply.ap` has the parameters in the other order (first `ff` then `fa`)
```scala
```

core/src/main/scala/cats/data/StateT.scala
```diff
@@ -268,11 +250,29 @@ private[data] sealed trait StateTSemigroupK[F[_], S] extends SemigroupK[StateT[F
...
+
+  override def ap[A, B](ff: StateT[F, S, A => B])(fa: StateT[F, S, A]): StateT[F, S, B] =
+    StateT[F, S, B]((s: S) =>
+      F.flatMap(ff.run(s)) { sab =>
+        val (sn, f) = sab
+        F.map(fa.run(sn)) { case (snn, a) => (snn, f(a)) }
+      }
+    )
+
+  override def map[A, B](fa: StateT[F, S, A])(f: A => B): StateT[F, S, B] = fa.map(f)(F)
```
peterneyens: We can drop this line now that you've added `with StateTFunctor`.

free/src/main/scala/cats/free/FreeT.scala
edmundnoble: Fixed.

tests/src/test/scala/cats/tests/WriterTTests.scala
edmundnoble: Fixed.

free/src/main/scala/cats/free/Coyoneda.scala
```diff
@@ -29,6 +29,10 @@ sealed abstract class Coyoneda[F[_], A] extends Serializable { self =>
   /** Converts to `F[A]` given that `F` is a functor */
   final def run(implicit F: Functor[F]): F[A] = F.map(fi)(k)
 
+  /** Converts to `F[A]` given that `F` is a functor */
```
peterneyens: "Converts to `G[A]` given that `G` is a functor".

edmundnoble: Fixed.

edmundnoble:
Merging with two approvals.

alexandru:
🍻💕

PR #1752, opened 2017-06-30
More coverage

edmundnoble:
Calculated hail mary to increase coverage, to somewhat balance out the MTL removal coverage decrease.

### DIFF
--- a/tests/src/test/scala/cats/tests/EitherTests.scala
+++ b/tests/src/test/scala/cats/tests/EitherTests.scala
@@ -10,6 +10,7 @@ class EitherTests extends CatsSuite {
   implicit val iso = CartesianTests.Isomorphisms.invariant[Either[Int, ?]]
 
   checkAll("Either[String, Int]", GroupLaws[Either[String, Int]].monoid)
+  checkAll("Semigroup[Either[String, Int]]", SerializableTests.serializable(Semigroup[Either[String, Int]]))
 
   checkAll("Either[Int, Int]", CartesianTests[Either[Int, ?]].cartesian[Int, Int, Int])
   checkAll("Cartesian[Either[Int, ?]]", SerializableTests.serializable(Cartesian[Either[Int, ?]]))
@@ -27,6 +28,9 @@ class EitherTests extends CatsSuite {
 
   checkAll("Either[ListWrapper[String], ?]", SemigroupKTests[Either[ListWrapper[String], ?]].semigroupK[Int])
   checkAll("SemigroupK[Either[ListWrapper[String], ?]]", SerializableTests.serializable(SemigroupK[Either[ListWrapper[String], ?]]))
+
+  checkAll("Either[ListWrapper[String], Int]", GroupLaws[Either[ListWrapper[String], Int]].semigroup)
+  checkAll("Semigroup[Either[ListWrapper[String], Int]]", SerializableTests.serializable(Semigroup[Either[ListWrapper[String], Int]]))
 
   val partialOrder = catsStdPartialOrderForEither[Int, String]
   val order = implicitly[Order[Either[Int, String]]]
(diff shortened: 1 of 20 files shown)


peterneyens:
Seems there is a bit of overlap with what I did in #1623.

tests/src/test/scala/cats/tests/IdTTests.scala
```diff
@@ -1,7 +1,8 @@
 package cats
 package tests
 
-import cats.data.IdT
+import cats.{Foldable, Functor, Monad, NonEmptyTraverse, Traverse}
```
peterneyens: Is there a reason we add these imports explicitly?

tests/src/test/scala/cats/tests/MonoidTests.scala
```diff
@@ -1,22 +1,29 @@
...
 package tests
 
-import org.scalatest._
-
 import cats.functor._
 
-class MonoidTests extends FunSuite {
+class MonoidTests extends CatsSuite {
   {
-    import cats.implicits._
```
peterneyens: Should we maybe include a comment specifying that the instances below should be found when all the instances are in scope?
It might not be as obvious what we are doing here, now that we moved to `CatsSuite`.

tests/src/test/scala/cats/tests/EitherTests.scala
```diff
@@ -10,6 +10,7 @@ class EitherTests extends CatsSuite {
   implicit val iso = CartesianTests.Isomorphisms.invariant[Either[Int, ?]]
 
   checkAll("Either[String, Int]", GroupLaws[Either[String, Int]].monoid)
+  checkAll("Semigroup[Either[String, Int]]", SerializableTests.serializable(Semigroup[Either[String, Int]]))
```
peterneyens: Maybe check `Monoid[Either[String, Int]]` here?
The instance tested should be the same, but might be clearer given that we are testing `GroupLaws.monoid` above?

tests/src/test/scala/cats/tests/IdTTests.scala
edmundnoble: I didn't have the multiple package decls in a previous version of the file, IntelliJ magic.

tests/src/test/scala/cats/tests/MonoidTests.scala
edmundnoble: I'm not sure, it seems like all other transformers do not have or need such a comment.

tests/src/test/scala/cats/tests/WriterTTests.scala
```diff
@@ -335,11 +335,10 @@ class WriterTTests extends CatsSuite {
     implicit val FLV: Monoid[ListWrapper[(Int, Int)]] = ListWrapper.monoid[(Int, Int)]
 
     Monoid[WriterT[ListWrapper, Int, Int]]
-    Semigroup[WriterT[ListWrapper, Int, Int]]
     checkAll("WriterT[ListWrapper, Int, Int]", kernel.laws.GroupLaws[WriterT[ListWrapper, Int, Int]].monoid)
 
-    Monoid[WriterT[Id, Int, Int]]
-    Semigroup[WriterT[Id, Int, Int]]
+    Monoid[Writer[Int, Int]]
+    checkAll("Writer[Int, Int]", kernel.laws.GroupLaws[Writer[Int, Int]].monoid)
```
edmundnoble: Added.

tests/src/test/scala/cats/tests/EitherTests.scala
edmundnoble: Done.

tests/src/test/scala/cats/tests/PartialOrderTests.scala
```diff
@@ -1,20 +1,31 @@
...
-class PartialOrderTests extends FunSuite {
+class PartialOrderTests extends CatsSuite {
   {
-    import cats.implicits._
     Invariant[PartialOrder]
     Contravariant[PartialOrder]
   }
 
-  {
+  test("companion object syntax") {
```
johnynek: seems like this could be a scalacheck.

edmundnoble: Excellent point, corrected.

edmundnoble:
Merging with two sign-offs, assuming my latest edit was sufficient for the last comment.

PR #1756, opened 2017-07-05
Add piecemeal import guide

### LINKED ISSUE #1668: Piecemeal import guide
Currently the [imports guide](http://typelevel.org/cats/typeclasses/imports.html) only really describes the "uber" import (`cats.implicits._`). It may be handy to have a detailed guide of piecemeal imports, such as `cats.instances.string._` and `cats.syntax.functor._` as an additional section for people who prefer this approach.

LukaJCB:
Still a WIP, but looking for some early feedback. Should resolve #1668

### DIFF
--- a/docs/src/main/tut/typeclasses/imports.md
+++ b/docs/src/main/tut/typeclasses/imports.md
@@ -40,4 +40,29 @@ val o: Option[String] = None
 o.orEmpty
 ```
 
-**Note**: if you import `cats.implicits._` (the preferred method), you should _not_ also use imports like `cats.syntax.option._` or `cats.instances.either._`. This can result in ambiguous implicit values that cause bewildering compile errors.
+If you'd like to import à la carte, you can do so, by importing from `cats.instances` for the type class instances and `cats.syntax` for syntax enrichment.
+For example, if you'd like to import the `Semigroup` instance for `String` and the corresponding syntax:
+```tut:book
+import cats.instances.string._
+import cats.syntax.semigroup._
+
+"Hello, " |+| "World!"
+```
+The first import pulls the `Semigroup` instance for String into the scope, while the second import adds the `|+|` syntax.
+
+You can also import all syntax or all instances by importing `cats.instances.all._` or `cats.syntax.all._` respectively.
+
+**Note**: Beware that if you import a type class instance or its syntax twice, you will receive conflicting implicits with a less than helpful error message.
+This usually happens when importing different type classes in the same hierarchy or when importing syntax enrichment for all type classes using `cats.syntax.all._` or `cats.implicits._` together with a more specific import like `cats.syntax.option._` or `cats.instances.either._`.
+Below is an example of this phenomenon:
+```tut:fail
+import cats.instances.all._
+import cats.syntax.semigroup._
+val x = -2 |+| 1
+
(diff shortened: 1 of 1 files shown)


kailuowang:
Thanks! looking good to me so far.

docs/src/main/tut/typeclasses/imports.md
```diff
@@ -40,4 +40,29 @@ val o: Option[String] = None
...
+For example, if you'd like to import the `Semigroup` instance for `String` and the corresponding syntax:
+```tut:book
+import cats.instances.string._
+import cats.syntax.semigroup._
+
+"Hello, " |+| "World!"
+```
+The first import pulls the `Semigroup` instance for String into the scope, while the second import adds the `|+|` syntax.
+
+You can also import all syntax or all instances by importing `cats.instances.all._` or `cats.syntax.all._` respectively.
```
durban: Isn't this mixed up? I.e., it should be something like "(syntax or instances) ... (`cats.syntax.all._` or `cats.instances.all._`) respectively". (Although english is not my first language.)

ceedubs:
🎉 so happy to see this finally land. It's something that I started on multiple times but never completed. Thanks @LukaJCB!

docs/src/main/tut/typeclasses/imports.md
```diff
@@ -40,4 +40,40 @@ val o: Option[String] = None
 o.orEmpty
 ```
 
-**Note**: if you import `cats.implicits._` (the preferred method), you should _not_ also use imports like `cats.syntax.option._` or `cats.instances.either._`. This can result in ambiguous implicit values that cause bewildering compile errors.
+If you'd like to import à la carte, you can do so, by importing from `cats.instances` for the type class instances and `cats.syntax` for syntax enrichment.
+For example, if you'd like to import the `Monoid` instance for `String` and the corresponding syntax:
+```tut:book
+import cats.instances.string._
+import cats.syntax.monoid._
```
peterneyens: I was surprised this actually worked and we didn't need `import cats.syntax.semigroup._`, but it seems that `MonoidSyntax` extends `SemigroupSyntax`.

It seems that all the kernel syntax extends their "super syntax traits", which is another inconsistency between cats-kernel and cats-core (even though this syntax is in cats-core).

docs/src/main/tut/typeclasses/imports.md
```diff
@@ -40,4 +40,40 @@ val o: Option[String] = None
...
+This usually happens when importing different type classes in the same hierarchy or when importing syntax enrichment for all type classes using `cats.syntax.all._` or `cats.implicits._` together with a more specific import like `cats.syntax.option._` or `cats.instances.either._`.
+Below is an example of this phenomenon:
+```tut:silent
+import cats.instances.all._
+import cats.syntax.semigroup._
+val x = -2 |+| 1
+
+//now we also need access to isEmpty from Monoid
+import cats.syntax.monoid._
+(x |+| 1).isEmpty //error: value |+| is not a member of Int
```
peterneyens: If `MonoidSyntax` didn't extend `SemigroupSyntax`, we wouldn't have this issue.

You would only have this issue when you use `cats.implicits._` or `cats.xxx.all._` in combination with an a la carte import.

PR #1758, opened 2017-07-06
stop publishing cats all bundle , start to publish cats-testkit

### LINKED ISSUE #1682: Cats 1.0.0 packaging
This is an issue consolidating all the packaging/publish related issues #1656, #1184, #472 and #592. 
There are three separate decisions that are sort of related. It may make sense to decide on all of them all together. Here is the revised version of @BennyHill's summarization in this [comment](https://github.com/typelevel/cats/issues/1184#issuecomment-298991494) (I added more description for context)

1. Shall we continue publishing the current cats bundle which includes cats-laws and cats-free?
The cats-laws includes a scalacheck dependency which [surprises people.](https://github.com/typelevel/cats/issues/1184#issuecomment-285366498)

Options are:
a) Keep as-is
b) Keep as-is but minus test libs
c) Rename as-is
d) Rename as-is but minus test libs
e) Remove

2. Shall we include alleycats into the cats repo as submodule? detailed discussion in #472

Options are: 
a) Keep as-is, i.e. separate repo
b) move in as cats submodule. It will keep using a separate top `alleycats` namespace and will not be included in the cats bundle shall we decide to keep one. As a submodule, its releases will be part of cats release, but unlike cats-core, it is allowed to have binary breaking changes between patch versions for alley-cats, at least for now, since it's API hasn't reached the stable status yet.

3. Shall we publish cats-testkit? discussion in #1656

a) Create a new module called `cats-testkit` that includes the following utils from `cats-tests`: [Helpers](https://github.com/typelevel/cats/blob/master/tests/src/test/scala/cats/tests/Helpers.scala), 
 [CatsSuite](https://github.com/typelevel/cats/blob/master/tests/src/test/scala/cats/tests/CatsSuite.scala) and [CatsEquality](https://github.com/typelevel/cats/blob/master/tests/src/test/scala/cats/tests/CatsEquality.scala)
Note that these classes are not required for users to write cats style tests - `cats-laws/discipline` is the one that is required. `cats-testkit` will depend on `cats-laws` 
b) keep as-is 
c) simply publish `cats-tests` ( but not included in the cats bundle shall we have one. )

We've done extensive discussion around these for quite a while. So I propose we are ready to collect votes and see if we can reach consensus.

kailuowang:
fixes #1682


edmundnoble:
Merging with two (plus two) approvals.

Issue #1757, opened 2017-07-06
Some tests are really slow


kailuowang:
- Cokleisli[NonEmptyList, Int, Int].arrow.compose associativity (23 seconds, 144 milliseconds)
- Cokleisli[NonEmptyList, Int, Int].monoidK.semigroupK associative (22 seconds, 496 milliseconds)
- Cokleisli[NonEmptyList, Int, Int].semigroupK.semigroupK associative (26 seconds, 5 milliseconds)
- Traversing with ReaderWriterState is stack-safe (1 minute, 47 seconds)
- StateT[ListWrapper, Int, Int].monadState.apply composition (32 seconds, 992 milliseconds)
- StateT[ListWrapper, Int, Int].monadState.cartesian associativity (31 seconds, 908 milliseconds)
- StateT[ListWrapper, Int, Int].monadState.flatMap associativity (20 seconds, 123 milliseconds)
- StateT[ListWrapper, Int, Int].monad.apply composition (47 seconds, 880 milliseconds)
- StateT[ListWrapper, Int, Int].monad.cartesian associativity (24 seconds, 700 milliseconds)
- StateT[ListWrapper, Int, Int].monad.flatMap associativity (27 seconds, 404 milliseconds)
- StateT[ListWrapper, Int, Int].monadCombine.apply composition (21 seconds, 939 milliseconds)
- StateT[ListWrapper, Int, Int].monadCombine.cartesian associativity (28 seconds, 935 milliseconds)
- StateT[ListWrapper, Int, Int].monadCombine.flatMap associativity (24 seconds, 395 milliseconds)

peterneyens:
For the `StateT[ListWrapper, ...]` tests, maybe we can try limiting the size of the `List`s which is what I think we are doing in [`WriterTTests`](https://github.com/typelevel/cats/blob/master/tests/src/test/scala/cats/tests/WriterTTests.scala#L15-L19).

peterneyens:
Limiting the size range in `StateTTests` cuts the total time for `StateTTests` from around 4 minutes to around 20 seconds.

```scala
  implicit override val generatorDrivenConfig: PropertyCheckConfiguration =
    checkConfiguration.copy(sizeRange = 5)
````

PR #1759, opened 2017-07-06
Faster tests by reducing the size of lists

peterneyens:
(Partial) fix for #1757.

This doesn't help to speed up the associativity laws in `CokleisliTests` (or not very much).

### DIFF
--- a/tests/src/test/scala/cats/tests/CokleisliTests.scala
+++ b/tests/src/test/scala/cats/tests/CokleisliTests.scala
@@ -12,7 +12,9 @@ import org.scalacheck.Arbitrary
 class CokleisliTests extends SlowCatsSuite {
 
   implicit override val generatorDrivenConfig: PropertyCheckConfiguration =
-    slowCheckConfiguration.copy(sizeRange = slowCheckConfiguration.sizeRange.min(5))
+    slowCheckConfiguration.copy(
+      sizeRange = slowCheckConfiguration.sizeRange.min(5),
+      minSuccessful = slowCheckConfiguration.minSuccessful.min(20))
 
   implicit def cokleisliEq[F[_], A, B](implicit A: Arbitrary[F[A]], FB: Eq[B]): Eq[Cokleisli[F, A, B]] =
     Eq.by[Cokleisli[F, A, B], F[A] => B](_.run)
--- a/tests/src/test/scala/cats/tests/CokleisliTests.scala
+++ b/tests/src/test/scala/cats/tests/CokleisliTests.scala
@@ -11,6 +11,9 @@ import org.scalacheck.Arbitrary
 
 class CokleisliTests extends SlowCatsSuite {
 
+  implicit override val generatorDrivenConfig: PropertyCheckConfiguration =
+    slowCheckConfiguration.copy(sizeRange = slowCheckConfiguration.sizeRange.min(5))
+
   implicit def cokleisliEq[F[_], A, B](implicit A: Arbitrary[F[A]], FB: Eq[B]): Eq[Cokleisli[F, A, B]] =
     Eq.by[Cokleisli[F, A, B], F[A] => B](_.run)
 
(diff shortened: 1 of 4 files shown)


tests/src/test/scala/cats/tests/CokleisliTests.scala
```diff
@@ -34,30 +37,14 @@ class CokleisliTests extends SlowCatsSuite {
   checkAll("Cokleisli[Option, Int, Int]", ContravariantTests[Cokleisli[Option, ?, Int]].contravariant[Int, Int, Int])
   checkAll("Contravariant[Cokleisli[Option, ?, Int]]", SerializableTests.serializable(Contravariant[Cokleisli[Option, ?, Int]]))
 
-  {
-    // Ceremony to help scalac to do the right thing, see also #267.
-    type CokleisliNEL[A, B] = Cokleisli[NonEmptyList, A, B]
```
peterneyens: We don't need this ceremony anymore.

tests/src/test/scala/cats/tests/CokleisliTests.scala
```diff
@@ -34,30 +37,14 @@ class CokleisliTests extends SlowCatsSuite {
...
-    }
-
-    {
-      implicit val cokleisliSemigroupK = Cokleisli.catsDataSemigroupKForCokleisli[NonEmptyList]
-      checkAll("Cokleisli[NonEmptyList, Int, Int]", SemigroupKTests[CokleisliNELE].semigroupK[Int])
-      checkAll("SemigroupK[λ[α => Cokleisli[NonEmptyList, α, α]]]", SerializableTests.serializable(cokleisliSemigroupK))
-    }
-  }
+  checkAll("Cokleisli[List, Int, Int]", SemigroupKTests[λ[α => Cokleisli[List, α, α]]].semigroupK[Int])
+  checkAll("SemigroupK[λ[α => Cokleisli[List, α, α]]]", SerializableTests.serializable(SemigroupK[λ[α => Cokleisli[List, α, α]]]))
```
peterneyens: `Cokleisli`'s `SemigroupK` instance just needs `CoflatMap` so I replaced `NonEmptyList` with `List`.

tests/src/test/scala/cats/tests/WriterTTests.scala
```diff
@@ -16,7 +16,7 @@ class WriterTTests extends CatsSuite {
   // Scalacheck to calm down a bit so we don't hit memory and test duration
   // issues.
   implicit override val generatorDrivenConfig: PropertyCheckConfiguration =
-    PropertyCheckConfiguration(minSuccessful = 20, sizeRange = 5)
+    checkConfiguration.copy(sizeRange = 5)
```
peterneyens: Keeping `minSuccessful` as it is set in `CatsSuite` only seems to add around half a second to the total test time of `WriterTTests`.

tests/src/test/scala/cats/tests/CokleisliTests.scala
```diff
@@ -11,6 +11,9 @@ import org.scalacheck.Arbitrary
 
 class CokleisliTests extends SlowCatsSuite {
 
+  implicit override val generatorDrivenConfig: PropertyCheckConfiguration =
+    slowCheckConfiguration.copy(sizeRange = slowCheckConfiguration.sizeRange.min(5))
```
peterneyens: Reducing `minSuccessful` would reduce the time for `CokleisliTests`:

- 50, around 1m20s (now, default value `CatsSuite` for JVM)
- 25, around 50s
- 20, around 20s

Opinions?

kailuowang: 20 sounds safe enough to me. 1m is a big difference for people who wants to work on cokleisli.

kailuowang:
looks like it reduced the build time by about 5-10 min

PR #1761, opened 2017-07-07
Add index related helpers to Traverse

andyscott:
This adds `traverseWithIndex`, `mapWithIndex` and `zipWithIndex` to `Traverse`.
- `traverseWithIndexM` is akin to `traverse`, but includes the index of the value in the structure `F`. This requires the effect to be monadic.
- `mapWithIndex` is akin to `map`, but includes the index of the value in the structure `F`.
- `zipWithIndex` returns F, with values tupled with their indices, just like Scala collection's zipWithIndex.

Instances for `List`, `Vector`, and `Stream` have overridden `mapWithIndex` and `zipWithIndex` that delegate to related underlying collections methods. `mapWithIndex` on `List` and `Vector` uses `.iterator` to avoid building an intermediate collection when first zipping with indices.

### DIFF
--- a/core/src/main/scala/cats/Traverse.scala
+++ b/core/src/main/scala/cats/Traverse.scala
@@ -1,6 +1,7 @@
 package cats
 
 import cats.data.State
+import cats.data.StateT
 
 import simulacrum.typeclass
 
@@ -111,8 +112,18 @@ import simulacrum.typeclass
     mapWithIndex(fa)((a, i) => (a, i))
 
   /**
-   * Akin to [[map]], but also provides the value's index in structure `F`.
+   * Akin to [[map]], but also provides the value's index in structure
+   * F when calling the function.
    */
   def mapWithIndex[A, B](fa: F[A])(f: (A, Int) => B): F[B] =
-    traverse(fa)(a => State((s: Int) => (s + 1, f(a, s)))).runA(0).value
+    traverse(fa)(a =>
+      State((s: Int) => (s + 1, f(a, s)))).runA(0).value
+
+  /**
+   * Akin to [[traverse]], but also provides the value's index in
+   * structure F when calling the function.
+   */
+  def traverseWithIndex[G[_], A, B](fa: F[A])(f: (A, Int) => G[B])(implicit G: Monad[G]): G[F[B]] =
+    traverse(fa)(a =>
+      StateT((s: Int) => G.map(f(a, s))(b => (s + 1, b)))).runA(0)
 }
--- a/core/src/main/scala/cats/Traverse.scala
+++ b/core/src/main/scala/cats/Traverse.scala
@@ -1,5 +1,7 @@
 package cats
 
+import cats.data.State
+
 import simulacrum.typeclass
 
 /**
@@ -97,4 +99,14 @@ import simulacrum.typeclass
 
   override def map[A, B](fa: F[A])(f: A => B): F[B] =
     traverse[Id, A, B](fa)(f)
+
+  /**
+   * Traverses through the structure F, pairing the values with
+   * assigned indices.
+   *
+   * The behavior is consistent with the Scala collection library's
+   * `zipWithIndex` for collections such as `List`.
+   */
+  def indexed[A](fa: F[A]): F[(A, Int)] =
+    traverse(fa)(a => State((s: Int) => (s + 1, (a, s)))).runA(0).value
 }
--- a/core/src/main/scala/cats/Traverse.scala
+++ b/core/src/main/scala/cats/Traverse.scala
@@ -108,5 +108,11 @@ import simulacrum.typeclass
    * `zipWithIndex` for collections such as `List`.
    */
   def indexed[A](fa: F[A]): F[(A, Int)] =
-    traverse(fa)(a => State((s: Int) => (s + 1, (a, s)))).runA(0).value
+    mapWithIndex(fa)((a, i) => (a, i))
+
+  /**
+   * Akin to [[map]], but also provides the value's index in structure `F`.
+   */
+  def mapWithIndex[A, B](fa: F[A])(f: (A, Int) => B): F[B] =
+    traverse(fa)(a => State((s: Int) => (s + 1, f(a, s)))).runA(0).value
 }
--- a/core/src/main/scala/cats/Traverse.scala
+++ b/core/src/main/scala/cats/Traverse.scala
@@ -102,16 +102,6 @@ import simulacrum.typeclass
     traverse[Id, A, B](fa)(f)
 
   /**
-   * Traverses through the structure F, pairing the values with
-   * assigned indices.
-   *
-   * The behavior is consistent with the Scala collection library's
-   * `zipWithIndex` for collections such as `List`.
-   */
-  def indexed[A](fa: F[A]): F[(A, Int)] =
-    mapWithIndex(fa)((a, i) => (a, i))
-
-  /**
    * Akin to [[map]], but also provides the value's index in structure
    * F when calling the function.
    */
@@ -126,4 +116,14 @@ import simulacrum.typeclass
   def traverseWithIndex[G[_], A, B](fa: F[A])(f: (A, Int) => G[B])(implicit G: Monad[G]): G[F[B]] =
     traverse(fa)(a =>
       StateT((s: Int) => G.map(f(a, s))(b => (s + 1, b)))).runA(0)
+
+  /**
+   * Traverses through the structure F, pairing the values with
+   * assigned indices.
+   *
+   * The behavior is consistent with the Scala collection library's
+   * `zipWithIndex` for collections such as `List`.
+   */
+  def zipWithIndex[A](fa: F[A]): F[(A, Int)] =
+    mapWithIndex(fa)((a, i) => (a, i))
 }
--- a/core/src/main/scala/cats/Traverse.scala
+++ b/core/src/main/scala/cats/Traverse.scala
@@ -112,8 +112,12 @@ import simulacrum.typeclass
   /**
    * Akin to [[traverse]], but also provides the value's index in
    * structure F when calling the function.
+   *
+   * This performs the traversal in a single pass but requires that
+   * effect G is monadic. An applicative traveral can be performed in
+   * two passes using [[zipWithIndex]] followed by [[traverse]].
    */
-  def traverseWithIndex[G[_], A, B](fa: F[A])(f: (A, Int) => G[B])(implicit G: Monad[G]): G[F[B]] =
+  def traverseWithIndexM[G[_], A, B](fa: F[A])(f: (A, Int) => G[B])(implicit G: Monad[G]): G[F[B]] =
     traverse(fa)(a =>
       StateT((s: Int) => G.map(f(a, s))(b => (s + 1, b)))).runA(0)
 
--- a/core/src/main/scala/cats/instances/vector.scala
+++ b/core/src/main/scala/cats/instances/vector.scala
@@ -83,7 +83,7 @@ trait VectorInstances extends cats.kernel.instances.VectorInstances {
         }.value
 
       override def mapWithIndex[A, B](fa: Vector[A])(f: (A, Int) => B): Vector[B] =
-        fa.view.zipWithIndex.map(ai => f(ai._1, ai._2)).toVector
+        fa.iterator.zipWithIndex.map(ai => f(ai._1, ai._2)).toVector
 
       override def zipWithIndex[A](fa: Vector[A]): Vector[(A, Int)] =
         fa.zipWithIndex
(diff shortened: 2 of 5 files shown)


core/src/main/scala/cats/Traverse.scala
```diff
@@ -33,6 +35,17 @@ import simulacrum.typeclass
...
   /**
+   * Traverse with a function returning state.
+   *
+   * Given a function which returns state, thread the state through
+   * the running of this function on all the values in F, returning
+   * F[B] in the state context.
+   */
+  def stateTraverse[S, A, B](fa: F[A])(f: A => State[S, B]): State[S, F[B]] =
+    State[S, F[B]](s =>
+      traverse[State[S, ?] , A, B](fa)(f).run(s).value)
```
peterneyens: We don't need the additional `State` wrapping of `run(s).value`, right?

This is equal to `traverse[State[S, ?] , A, B](fa)(f)` and with SI 2712 fixed, we don't even need to specify the types, so it is just `traverse(fa)(f)`.

ceedubs: To follow up on @peterneyens comment, since this is just equivalent to `traverse`, is there any reason to have a separate `stateTraverse`?

johnynek: yeah, I'm confused why this is needed.

tests/src/test/scala/cats/tests/ListTests.scala
```diff
@@ -42,4 +42,10 @@ class ListTests extends CatsSuite {
       l.show should === (l.toString)
     }
   }
+
+  test("indexed consistent with zipWithIndex") {
+    forAll { (fa: List[String]) =>
+      Traverse[List].indexed(fa) should === (fa.zipWithIndex)
```
johnynek: can we try this for a very long list (say 500k elements)? I'm concerned State with traverse can fail for us. I think I have seen issues like that before. I don't want to leave a landmine for someone.

I think we may be able to work around by using tailRecM on State though, but let's see if we need to.

core/src/main/scala/cats/Traverse.scala
andyscott: It isn't!

PSA: Don't code while deliriously tired.

tests/src/test/scala/cats/tests/ListTests.scala
andyscott: Added. Went ahead and created a dedicated file for testing `Traverse` instead of piggybacking on the tests in `ListTests.scala`.

johnynek:
👍

kailuowang:
seems problematic on scalaJS
> Security context: 0x1784390b4629 <JS Object>
    1: flatMap__F1__Lcats_Eval [/home/travis/build/typelevel/cats/tests/.js/target/scala-2.10/cats-tests-test-fastopt.js:~49091] [pc=0x323fb4f2641] (this=0x3b070c22ee61 <a $c_Lcats_Eval$$anon$5 with map 0x14937c2e6199>,f=0xe6fa70d8d09 <a $c_sjsr_AnonFunction1 with map 0x14937c2a97e1>)
    2: map__F1__Lcats_Eval [/home/travis/build/typelevel/cats/tests/.js/target/scala-2.10/cats-tests-test-fasto...
[error] FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - process out of memory

andyscott:
I did some quick testing. On the JVM, everything is fine. However, on JS the performance of `.indexed` is nonlinear. The tests run reasonably quick for < 100,000 items in the collection. After 300,000 items, the tests get extremely slow or fail.

Currently I haven't looked for an underlying problem. For the time being, I have adjusted the tests so that JS is uses a smaller collection.

kailuowang:
Do we know that 100,000 isn't enough to test against stack safety for JVM? Now the 3 tests takes more than 30 seconds, not sure if that's necessary
> [info] - Traverse[List].indexed stack safety (12 seconds, 943 milliseconds)
[info] - Traverse[Stream].indexed stack safety (15 seconds, 11 milliseconds)
[info] - Traverse[Vector].indexed stack safety (12 seconds, 585 milliseconds)

edmundnoble:
@andyscott @kailuowang Afaik you actually need at most 70000 (see my latest merged PR).

andyscott:
@edmundnoble Aah! Right now it's been switched to 100000. I'll switch it to 70000 :).

edmundnoble:
Can you also give me something like indexed which takes a function `(A, Int) => B` and produces an `F[B]`?

andyscott:
something like `mapWithIndex[B](fa: F[A])(f: (A, Int) => B): F[B] = ...`?

edmundnoble:
Yeah sure.

andyscott:
Would it be preferable to implement `indexed` in terms of `mapWithIndex`?

```scala
  def indexed[A](fa: F[A]): F[(A, Int)] =
    mapWithIndex(fa)((a, i) => (a, i))

  def mapWithIndex[A, B](fa: F[A])(f: (A, Int) => B): F[B] =
    traverse(fa)(a => State((s: Int) => (s + 1, f(a, s)))).runA(0).value
```

andyscott:
Any objections to also adding `traverseWithIndex`?

```scala
def traverseWithIndex[G[_]: Monad, A, B](fa: F[A])(f: (A, Int) => G[B]): G[F[B]] =
      traverse(fa)(a => StateT((s: Int) => f(a, s).map(b => (s + 1, b)))).runA(0)
```

edmundnoble:
That's fine with me but I think in the long run it could be better, as soon as we have a `Sum` new type for that `Monoid` you can change the constraint to `G[_]: Applicative` and use `WriterT`.

core/src/main/scala/cats/Traverse.scala
```diff
@@ -97,4 +100,30 @@ import simulacrum.typeclass
...
     traverse[Id, A, B](fa)(f)
+
+  /**
+   * Traverses through the structure F, pairing the values with
+   * assigned indices.
+   *
+   * The behavior is consistent with the Scala collection library's
+   * `zipWithIndex` for collections such as `List`.
+   */
+  def indexed[A](fa: F[A]): F[(A, Int)] =
```
johnynek: any reason not to call this `zipWithIndex`?

tests/src/test/scala/cats/tests/TraverseTests.scala
```diff
@@ -0,0 +1,53 @@
...
+abstract class TraverseCheck[F[_]: Traverse](name: String)(implicit ArbFInt: Arbitrary[F[Int]]) extends CatsSuite with PropertyChecks {
+
+  test(s"Traverse[$name].indexed") {
+    forAll { (fa: F[Int]) =>
+      fa.indexed.toList should === (fa.toList.zipWithIndex)
+    }
+  }
+
+  test(s"Traverse[$name].mapWithIndex") {
+    forAll { (fa: F[Int]) =>
```
johnynek: can we take `((Int, Int)) => Int` as a function and do:
```scala
fa.mapWithIndex((a, i) => fn((a, i))).toList should === (fa.toList.zipWithIndex.map(fn))
```

tests/src/test/scala/cats/tests/TraverseTests.scala
```diff
@@ -0,0 +1,53 @@
...
+      fa.mapWithIndex((a, i) => (a, i)).toList should === (fa.toList.zipWithIndex)
+    }
+  }
+
+  test(s"Traverse[$name].traverseWithIndex") {
+    forAll { (fa: F[Int]) =>
+      fa.traverseWithIndex((a, i) => Option((a, i))).map(_.toList) should === (Option(fa.toList.zipWithIndex))
```
johnynek: could we have a stronger test, like take `fn: ((Int, Int)) => Option[Int]` and then
```scala
val left = fa.traverseWithIndex((a, i) => fn((a, i))).map(_.toList)
val right = (fa.toList.zipWithIndex match {
  case Nil => Some(Nil)
  case h :: tail => tail.foldM(h :: Nil) { case (l, ai) => fn(ai).map(_ :: l) }.map(_.reverse)
})
left should === right
```

core/src/main/scala/cats/Traverse.scala
```diff
@@ -97,4 +100,30 @@ import simulacrum.typeclass
...
+    mapWithIndex(fa)((a, i) => (a, i))
+
+  /**
+   * Akin to [[map]], but also provides the value's index in structure
+   * F when calling the function.
+   */
+  def mapWithIndex[A, B](fa: F[A])(f: (A, Int) => B): F[B] =
```
johnynek: can we override this for `List`, `Vector` and `Stream`?

core/src/main/scala/cats/Traverse.scala
johnynek: can we override this for `List` `Vector` and `Stream`?

core/src/main/scala/cats/Traverse.scala
andyscott: Sure thing

core/src/main/scala/cats/Traverse.scala
andyscott: Yep, will do.

tests/src/test/scala/cats/tests/TraverseTests.scala
andyscott: Update incoming. I switched from `Option` to `Tuple2` to make the stronger test more straightforward.

andyscott:
@johnynek I think I've addressed everything.

Using views seemed to be the best way to override `mapWithIndex`, as to avoid creating an intermediate collection for `List` and `Vector` traverse instances.

tests/src/test/scala/cats/tests/TraverseTests.scala
```diff
@@ -0,0 +1,76 @@
...
+}
+
+object TraverseCheck {
+  // forces testing of the underlying implementation (avoids overridden methods)
+  abstract class Underlying[F[_]: Traverse](name: String)(implicit ArbFInt: Arbitrary[F[Int]])
+      extends TraverseCheck(s"$name (underlying)")(proxyTraverse[F], ArbFInt)
+
+  // proxies a traverse instance so we can test default implementations
+  // to achieve coverage using default datatype instances
+  private def proxyTraverse[F[_]: Traverse]: Traverse[F] = new Traverse[F] {
```
johnynek: this is nice.

core/src/main/scala/cats/Traverse.scala
```diff
@@ -97,4 +100,30 @@ import simulacrum.typeclass
...
+    traverse(fa)(a =>
+      State((s: Int) => (s + 1, f(a, s)))).runA(0).value
+
+  /**
+   * Akin to [[traverse]], but also provides the value's index in
+   * structure F when calling the function.
+   */
+  def traverseWithIndex[G[_], A, B](fa: F[A])(f: (A, Int) => G[B])(implicit G: Monad[G]): G[F[B]] =
```
johnynek: my only slight reservation with the name here is that `traverse` takes an `Applicative[G]` and implies often some parallelism, where this is sequential and requires `Monad[G]` since the applicative for `StateT` still requires `Monad[G]` (I think).

andyscott: I don't really like that I had to require `Monad`. I suppose this could also be implemented with just `Applicative` if it's done with two passes on the underlying structure-- `traverse(zipWithIndex(f))(ai => fa(ai._1, ai._2))

edmundnoble: Disregard my previous comment that you can't do this. You can use the Group[Int] and WriterT instead of State.

johnynek: if we can relax to `Applicative` that would be ideal.

andyscott: I can't sort out how to make it work with `WriterT`. What did you have in mind, @edmundnoble?

core/src/main/scala/cats/instances/vector.scala
```diff
@@ -82,6 +82,12 @@ trait VectorInstances extends cats.kernel.instances.VectorInstances {
           G.map2Eval(f(a), lgvb)(_ +: _)
         }.value
 
+      override def mapWithIndex[A, B](fa: Vector[A])(f: (A, Int) => B): Vector[B] =
+        fa.view.zipWithIndex.map(ai => f(ai._1, ai._2)).toVector
```
johnynek: can we benchmark with `fa.iterator.zipWithIndex.map(...`, I think `.iterator` may be faster than `.view`

andyscott: An official benchmark in the benchmark submodule? Or will a quick local benchmark work, updating the code according?

core/src/main/scala/cats/Traverse.scala
edmundnoble: My bad, I can't find a way to do this. I can calculate the length though :P.

andyscott:
@johnynek I renamed `traverseWithIndex` to `traverseWithIndexM` and add a note about traversing with index using applicative. I also switched `.view` to `.iterator` after a bit of investigation and more trivial benchmarking locally.

kailuowang:
merged with 3 sign-offs

Issue #1733, opened 2017-06-18
Kliesli is not stack-safe on left-associated binds


djspiewak:
```scala
    val unit = Kleisli.pure[Eval, Unit, Unit](())

    val result = (0 until 10000).foldLeft(unit) { (acc, _) =>
      acc.flatMap(_ => unit)
    }

    result <-> unit
```

This law fails.  The reason this law fails is here:

```scala
  def flatMap[C](f: B => Kleisli[F, A, C])(implicit F: FlatMap[F]): Kleisli[F, A, C] =
    Kleisli((r: A) => F.flatMap[B, C](run(r))((b: B) => f(b).run(r)))
    //                                ^^^^^^ this bit here!
```

A chain of right-associated binds will push inside the lambda argument to `flatMap`, meaning it will be subject to the stack safety of the constituent monad `F[_]`.  A chain of left-associated binds will result in a proportionally long chain of `run(r)`.  In order to solve this, we would need to use a similar technique as to what is done in `StateT`.

ceedubs:
In years of scalaz usage, the stack unsafety of `StateT` bit me on a few occasions. But I never ran into the same with `Kleisli`. I'm curious whether anyone has ever had problems in real-world code because of this.

I think that safety (no runtime surprises) is a major argument for reformulating `Kleisli` to be stack-safe. I think that consistency with `StateT` is a minor argument for it. I think that performance and beginner-friendliness are minor arguments against it.

I'm assigning this to  @adelbertc and @edmundnoble, who have done a lot of work with monad transformers. I'm also curious whether @johnynek has any thoughts on this.

edmundnoble:
> A chain of right-associated binds will push inside the lambda argument to `flatMap`, meaning it will be subject to the stack safety of the constituent monad `F[_]`

I don't believe this is correct, because your constituent monad is `Eval` and it's still not stack-safe. It is less stack-safe than the constituent monad; the `A => F[B]` function composition is the issue here. I have a potential quick solution in mind: `FreeT[A => ?, F, B]` is equivalent to `Kleisli[F, A, B]` but the former is stack-safe. Should we:

1. Use the `StateT` trick (which would work out as `F[A => F[B]]`),
2. Use `FreeT[A => ?, F, B]`,
3. Use a hand-inlined `2` for performance, with custom operations.

To my mind the issue is important enough to merit 3 eventually, but 2 seems like an alright first approximation. I think the `StateT` trick is deleterious enough to performance to avoid in future.

[...]

PR #1764, opened 2017-07-09
FreeT as stack-safe Kleisli

edmundnoble:
Part of a tentative solution to #1733. If this is merged another issue should be created for the hand-inlined version of this, if we end up wanting it.

### DIFF
--- a/free/src/main/scala/cats/free/package.scala
+++ b/free/src/main/scala/cats/free/package.scala
@@ -4,4 +4,6 @@ package object free {
   /** Alias for the free monad over the `Function0` functor. */
   type Trampoline[A] = Free[Function0, A]
   object Trampoline extends TrampolineFunctions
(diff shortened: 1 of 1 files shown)


free/src/main/scala/cats/free/package.scala
```diff
@@ -4,4 +4,6 @@ package object free {
   /** Alias for the free monad over the `Function0` functor. */
   type Trampoline[A] = Free[Function0, A]
   object Trampoline extends TrampolineFunctions
+  type StacksafeKleisli[F[_], A, B] = FreeT[A => ?, F, B]
```
andyscott: The "S" in "Safe" is already capitalized in `StackSafeMonad`, so it should probably be capitalized here for consistency.

edmundnoble:
@johnynek Has a better solution in `Free[Kleisli[F, E, ?], A]`.

Issue #1364, opened 2016-09-08
PSA: Be more careful about type classes included in cats(-core)


adelbertc:
Filing this as an issue just to have a central place to talk about this. It seems Cats is getting larger and larger with a lot of perhaps niche type classes. One of the goals of Cats was to be more modular/not grow into this gigantic project with everything included.

Examples that come to mind:

- FunctorFlatten https://github.com/typelevel/cats/pull/1337
- Cocartesian https://github.com/typelevel/cats/issues/1362
- Reducer https://github.com/typelevel/cats/issues/1360
- LeftFoldable https://github.com/typelevel/cats/issues/1339

I'm not saying we shouldn't include these, or that the current set of type classes we already have are blessed - I just want to bring up the fact I think we should **really** think about if a type class is necessary/general enough to be in a library like `cats-core` before including it.

cc-ing some folks who look to be active on those tickets @non @ceedubs @johnynek

eed3si9n:
+1 on keeping things lean esp if Cats 1.0 is going to have multi-year bincompat story.

johnynek:
I think this is a good thing to consider for sure. I don't think we want to add every type-class imaginable.

For what it's worth, I submitted mine only because it is very common in work I do (more so, than the less general): https://github.com/typelevel/cats/pull/1225/files (that said, I'm starting to wonder if it is not just a method on Alternative).

PR #1766, opened 2017-07-13
Replace Split with Commutative Arrow. Continuing #1719

kailuowang:
per conversation with @diesalbla, this it the PR attempting to complete #1719


edmundnoble:
My review is at the moment to use `locally` on line 42 of `CokleisliTests`.

core/src/main/scala/cats/data/Kleisli.scala
```diff
@@ -91,11 +91,11 @@ private[data] sealed abstract class KleisliInstances extends KleisliInstances0 {
...
-    new KleisliArrow[F] { def F: Monad[F] = M }
+  implicit def catsDataCommutativeArrowForKleisli[F[_]](implicit M: CommutativeMonad[F]): CommutativeArrow[Kleisli[F, ?, ?]] =
+    new KleisliCommutativeArrow[F] {def F: CommutativeMonad[F] = M }
 
-  implicit val catsDataArrowForKleisliId: Arrow[Kleisli[Id, ?, ?]] =
-    catsDataArrowForKleisli[Id]
+  implicit val catsDataCommutativeArrowForKleisliId: CommutativeArrow[Kleisli[Id, ?, ?]] =
+    catsDataCommutativeArrowForKleisli[Id]
 
   implicit def catsDataMonadReaderForKleisliId[A]: MonadReader[Kleisli[Id, A, ?], A] =
```
edmundnoble: This is seemingly a commutative monad now too, given `F[_]: Commutative`.

laws/src/main/scala/cats/laws/CommutativeCoflatMapLaws.scala
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Laws that must be obeyed by any `CommutativeCoflatMap`.
+ */
+trait CommutativeCoflatMapLaws[F[_]] extends CoflatMapLaws[F] {
+  implicit override def F: CommutativeCoflatMap[F]
+
+  def coflatmapCommutative[A, B, C](fa: F[A], fb: F[B], g: (F[A], F[B]) => C): IsEq[F[F[C]]] =
+    F.coflatMap(fa)(x => F.coflatMap(fb)(y => g(x, y))) <->
+    F.coflatMap(fb)(y => F.coflatMap(fa)(x => g(x, y)))
```
LukaJCB: Tiny nitpick, but should we name these `a` and `b` instead of `x` and `y`?

kailuowang:
@edmundnoble you are right, both are `CommutativeMonad`, instances added.

peterneyens:
Do we now if there are other `CommutativeCoflatMap` and/or `CommutativeComonad` instances than for `Id`? Otherwise I am not sure if these are worth adding ... ?

We can also make the `Arrow[Function1]` instance a `CommutativeArrow[Function1]`, right?

tests/src/test/scala/cats/tests/NonEmptyListTests.scala
```diff
@@ -2,8 +2,9 @@ package cats
 package tests
 
 import cats.kernel.laws.{GroupLaws, OrderLaws}
+
 import cats.data.{NonEmptyList, NonEmptyVector}
-import cats.laws.discipline.{ComonadTests, MonadTests, NonEmptyTraverseTests, ReducibleTests, SemigroupKTests, SerializableTests}
+import cats.laws.discipline.{SemigroupKTests, MonadTests, SerializableTests, NonEmptyTraverseTests, ReducibleTests, ComonadTests}
```
peterneyens: Maybe we can leave these alphabetically?

tests/src/test/scala/cats/tests/KleisliTests.scala
```diff
@@ -238,9 +254,8 @@ class KleisliTests extends CatsSuite {
     MonadReader[Reader[Int, ?], Int]
     Monoid[Reader[Int, String]]
     MonoidK[λ[α => Reader[α, α]]]
-    Arrow[Reader[?, ?]]
+    CommutativeArrow[Reader[?, ?]]
```
peterneyens: We can leave `Arrow[Reader[?, ?]]` here (move it below `CommutativeArrow`).

kailuowang:
@peterneyens I made function1 also commutative and correct the tests. 
The `CommutativeComonad` was introduced because `Reader` is a `CommutativeArrow` thanks to `Id` being a `CommutativeComonad`. Other than that I am not aware any other instances of `CommutativeComonad`, @diesalbla thought `NonEmptyList` is one but it turns out it isn't (or at least the law doesn't think so). I think there is value for `Reader` to be `CommutativeArrow`. We could remove `CommutativeComonad` and make a dedicated `CommutativeArrow` instance for `Reader`, not sure if that's a better solution though.

peterneyens:
The `CommunativeComonad` is used in the `CommutativeArrow` instance of `Cokleisli`, right?
`Reader` has a `CommutativeArrow` because `Id` has a `CommutativeMonad`.

I was just thinking that if we can't come up with other instances, adding `CommunativeCoflatMap` and `CommutativeComonad` might not be worth it, given that it currently only gives us a `CommutativeArrow` for `Cokleisli[Id, ?, ?]` (also see #1364).

I think this PR is good to be merged :+1: (thanks @diesalbla and @kailuowang) ; I just wanted to hear other opinions if `CommutativeCoflatMap` and `CommutativeComonad` are useful enough to merit inclusion.

edmundnoble:
Agreed with @peterneyens, I'd need at least one example of a commutative comonad/coflatmap before adding that but otherwise :+1: for merging.

kailuowang:
@peterneyens you are right, it was `Cokleisli` that is benefiting from it not `Reader`, I agree with you and @edmundnoble , will remove `CommutativeComonad` and `CommutativeCoflatmap` for now

tests/src/test/scala/cats/tests/CokleisliTests.scala
```diff
@@ -33,17 +33,20 @@ class CokleisliTests extends SlowCatsSuite {
...
 
   checkAll("Cokleisli[NonEmptyList, Int, Int]", MonoidKTests[λ[α => Cokleisli[NonEmptyList, α, α]]].monoidK[Int])
   checkAll("MonoidK[λ[α => Cokleisli[NonEmptyList, α, α]]]", SerializableTests.serializable(MonoidK[λ[α => Cokleisli[NonEmptyList, α, α]]]))
+  locally {
+    implicit def cokleisliIdEq[A, B](implicit A: Arbitrary[A], FB: Eq[B]): Eq[Cokleisli[Id, A, B]] =
+      Eq.by[Cokleisli[Id, A, B], A => B](_.run)
+
+    checkAll("Cokleisli[Id, Int, Int]", CommutativeArrowTests[Cokleisli[Id, ?, ?]].commutativeArrow[Int, Int, Int, Int, Int, Int])
+    checkAll("CommutativeArrow[Cokleisli[Id, ?, ?]]", SerializableTests.serializable(CommutativeArrow[Cokleisli[Id, ?, ?]]))
+  }
```
peterneyens: I think we need to add a test for `Cokleisli[F, ?, ?]` as well (maybe the existing one using `NonEmptyList`)?

Nitpicking, can we put these `CommutativeArrow` tests not between the `MonoidK` and `SemigroupK` tests?

kailuowang: addressed.

diesalbla:
Regarding `CommutativeCoflatmap` and `CommutativeComonad`, the only reason I introduced these classes in the first case was to preserve the similarity between the `Arrow` (and relatives) instances for `Kleisli` and `Cokleisli`. I could not come up with too many instances for them.

kailuowang:
merging with 2 sign-offs

PR #1767, opened 2017-07-19
Add traverseWithStateM to Traverse

andyscott:
Add `traverseWithStateM` and `traverseWithStateMA` to Traverse.

This allows you to carry a value through a computation without having to deal with the state monad directly. I have been using this perform multiple stateful computations at once, such as carrying the index value while also marking duplicates. The test I added demonstrates marking for duplicates.

For method naming, I chose `traverseWithStateM` because the "traverse" prefix seems most important. My only minor concern is that the "state" suffix implies the state monad being exposed.

### DIFF
--- a/core/src/main/scala/cats/Traverse.scala
+++ b/core/src/main/scala/cats/Traverse.scala
@@ -130,4 +130,38 @@ import simulacrum.typeclass
    */
   def zipWithIndex[A](fa: F[A]): F[(A, Int)] =
     mapWithIndex(fa)((a, i) => (a, i))
+
+  /**
+   * Statefully traverse through the structure F while carrying and updating a
+   * value of type S.
+   *
+   * Value `s0` is used as the initial state and function `fs` is used
+   * to update the state after each application of `f` through the structure.
+   *
+   * Function `fs` can update the state using the current traversal value, the
+   * previous state, and the output of the current application of `f`.
+   *
+   * Both the final state value and the traversed result are returned.
+   *
+   * The parameters are grouped in this order for better type inferencing.
+   *
+   * @see [[traverseWithStateMA]] if the final state value isn't needed.
+   */
+  def traverseWithStateM[G[_], A, S, B](fa: F[A])(s0: S)(f: (A, S) => G[B])(fs: (A, S, B) => S)(implicit G: Monad[G]): G[(S, F[B])] =
+    internalTraverseWithStateM(fa)(s0)(f)(fs).run(s0)
+
+  /**
+   * Statefully traverse through the structure F while carrying and updating a
+   * value of type S.
+   *
+   * This is identical to [[traverseWithStateM]] except it does not
+   * include the final state value in the result.
+   */
+  def traverseWithStateMA[G[_], A, S, B](fa: F[A])(s0: S)(f: (A, S) => G[B])(fs: (A, S, B) => S)(implicit G: Monad[G]): G[F[B]] =
+    internalTraverseWithStateM(fa)(s0)(f)(fs).runA(s0)
+
+  private[this] def internalTraverseWithStateM[G[_], A, S, B](
+    fa: F[A])(s0: S)(f: (A, S) => G[B])(fs: (A, S, B) => S)(implicit G: Monad[G]
+  ): StateT[G, S, F[B]] =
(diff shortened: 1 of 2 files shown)


kailuowang:
What do you think about adding a version that returns the state as well? like ` G[F[(S, B)]]`

andyscott:
You can mostly achieve that with the current function just by pairing your result with the input state.
```
scala> val fa = List(1, 2, 3, 4, 2, 4, 6)
fa: List[Int] = List(1, 2, 3, 4, 2, 4, 6)

scala> fa.traverseWithStateM(Set.empty[Int])((a, s) => if (s.contains(a)) Eval.now("dupe" -> s) else Eval.later(a.toString -> s))((a, s, _) => s + a).value                
res0: List[(String, Set[Int])] = List((1,Set()), (2,Set(1)), (3,Set(1, 2)), (4,Set(1, 2, 3)), (dupe,Set(1, 2, 3, 4)), (dupe,Set(1, 2, 3, 4)), (6,Set(1, 2, 3, 4)))                                                                 
```

kailuowang:
@andyscott right. thanks for coming up with an example.

andyscott:
Sure thing. While the incremental state can be retained with the current function, the final output state is inaccessible. I don't know if it's worth it, but I could add a function that returns `(S, G[F[B]])` instead of `G[F[B]]` by calling `run` instead of `runA`.

kailuowang:
I am leaning towards adding the `run` version, it seems useful to me and we won't be able to add it when we lock binary compat.

andyscott:
This method isn't really for me!

It mainly struck me as a bit cleaner to use and easier to work with for newer functional programmers. In this particular case, it removes a mental dependency on the state monad for the end user.

```scala
// note: haven't actually compiled this
// ... and it may not be the best example...

case class Record(id: Long, /* ... */)
case class S(index: Int, ids: Set[Long])

def doWork(in: Record): F[Result] = // ...

val ins: List[Record] = // ...

// with helper method
val outs1: F[List[Result]] =
  ins.statefulTraverse(S(0, Set.empty))(
    (rec, s) =>
      if (s.ids.contains(rec.id)) F.pure(DuplicateResult)
      else doWork(rec, s.index) )(
    (rec, s, _) => s.copy(
      index = s.index + 1,
      ids   = s.ids + rec.id) )

// without helper method
val outs2: F[List[Result]] =
  ins.traverse(rec => StateT[F, S, String](s => {
    val fb =
      if (s.ids.contains(rec.id)) F.pure(DuplicateResult)
      else doWork(rec, s.index))
    fb.map(_ => s.copy(
      index = s.index + 1,
      ids   = s.ids + rec.id) -> b)
  }).runA(S(0, Set.empty))
```

Some of the recent index related additions to `Traverse` could be implemented using this method.

iravid:
Sorry to butt in, but wouldn't this be formulated more clearly with `foldM`?

iravid:
E.g.
```scala
ins.foldLeftM(S(0, Set.empty)) { (s, rec) =>
    val fb =
      if (s.ids.contains(rec.id)) F.pure(DuplicateResult)
      else doWork(rec, s.index))
    fb.map(_ => s.copy(
      index = s.index + 1,
      ids   = s.ids + rec.id) -> b)
}
```

(sorry, was on mobile yesterday)

andyscott:
The posted code example with `foldLeftM` doesn't work, as you need to fold into some single type `B` in monad `G`. You've got the state type as input but the computation type as output.

With `foldLeftM`, I think the best you could do is mash up the state and computation values into a single target type `B`-- for example `B: (Set[Int], F[List[String]])`. You'd have to handle flatmapping the computations in the fold, and you'd wind up with a `F[(Set[Int], F[List[String]])`. I suppose you could `.map(_._2).flatten` this to get the final result.

iravid:
That's true. Thanks for the correction.

andyscott:
Sure thing. You forced me to think about the problem a bit more, which is always a good thing 😄 👍.

kailuowang:
@johnynek is @andyscott's argument [above](https://github.com/typelevel/cats/pull/1767#issuecomment-316752324) convincing enough for you? If not, I am going to unschedule it from 1.0-RC1, we can add it through extension in a separate module or project.

andyscott:
I'm not too concerned about this anymore (I'm indifferent if we merge it or
not).

PR #1777, opened 2017-08-01
Add reducible docs 

LukaJCB:
Looking for some early feedback on this, still mostly unstructured :)


docs/src/main/tut/typeclasses/reducible.md
```diff
@@ -0,0 +1,62 @@
...
+source: "core/src/main/scala/cats/Reducible.scala"
+scaladoc: "#cats.Reducible"
+---
+
+# Reducible
+
+`Reducible` extends the `Foldable` type class with additional `reduce` methods.
+
+You may have come by one of the `reduce`, `reduceLeft` or `reduceOption` defined in Scala's standard collections.
+`Reducible` offers exactly these methods with the guarantee, that the collection won't throw an exception at runtime, without having to reduce to an `Option`.
```
ceedubs: Minor: I don't think that you need a comma after "guarantee".

ceedubs: I guess that years of Java have made me paranoid. Maybe "the collection won't throw an exception at runtime" is a bit of a strong guarantee. Maybe we should just say that these methods won't throw an exception due to a collection being empty? :)

kailuowang:
Seems history got messed up?

LukaJCB:
Fixed!

kailuowang:
Sorry I dropped ball on this one. @LukaJCB for some reason Github conflict merge tool is not working for us. Do yo mind update this one and merge?

LukaJCB:
I think it's fine actually, no? You guys just need to re-review to merge :)

PR #1780, opened 2017-08-04
Bump machinist version to 0.6.2

shokohara:
I updated machinist version to 0.6.2.


kailuowang:
👍  thanks very much! merge with one sign-off as a simple dependency update.

shokohara:
I'll try to continue to send some PRs like this.

PR #1781, opened 2017-08-04
promoting the ecosystem in readme a bit and fixed a typo

kailuowang:
also updated cats-mtl url

### DIFF
--- a/docs/src/main/tut/index.md
+++ b/docs/src/main/tut/index.md
@@ -100,7 +100,6 @@ language", and integrate with each other with ease.
 
 #### General purpose libraries to support pure functional programming
 
- * [Mouse](https://github.com/benhutchison/mouse): a small companion to cats that provides convenient syntax (aka extension methods)
  * [Dogs](https://github.com/stew/dogs): pure functional collections and data structures.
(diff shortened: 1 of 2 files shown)


milessabin:
I suggest changing "**Generic** libraries to support pure functional programming" to " **General purpose** libraries to support pure functional programming".

kailuowang:
thanks @milessabin ,updated

docs/src/main/tut/index.md
```diff
@@ -98,7 +98,7 @@ Many projects integrate with cats. By sharing the same set of
 type classes, instances and data types, projects can speak the same "cats
 language", and integrate with each other with ease. 
 
-#### Generic libraries to support pure functional programming
+#### General purpose libraries to support pure functional programming
 
  * [Mouse](https://github.com/benhutchison/mouse): a small companion to cats that provides convenient syntax (aka extension methods)
```
andyscott: Mouse can be removed from this list, as it's listed above as one of the cats extra modules. Also, it's not on this list in the primary README anyway.

kailuowang: Good catch! Fixed.

johnynek:
👍

PR #1782, opened 2017-08-04
Bump scalacheck version to 1.13.5

shokohara:
I updated scalacheck version to 1.13.5.


kailuowang:
👍  another simple dependency update, merged with 1 sign-off

PR #1784, opened 2017-08-04
Add `existsM` and `forallM` to Foldable

refried:
I was missing these utility functions.  I've been out of the loop for a while so let me know if I didn't contrib right, and I'll be happy to update.

### DIFF
--- a/core/src/main/scala/cats/Foldable.scala
+++ b/core/src/main/scala/cats/Foldable.scala
@@ -327,6 +327,56 @@ import simulacrum.typeclass
     }.value
 
   /**
+    * Check whether at least one element satisfies the predicate.
+    *
+    * If there are no elements, the result is `false`.
+    *
+    * All effects are sequenced, i.e. later effects aren't skipped even if a `true` is encountered.
+    *
+    * For example:
+    *
+    * {{{
+    * scala> import cats.implicits._
+    * scala> val F = Foldable[List]
+    * scala> F.existsM(List(1,2,3,4))(n => Option(n <= 4))
+    * res0: Option[Boolean] = Some(true)
+    *
+    * scala> F.existsM(List(1,2,3,4))(n => if (n <= 2) Option(true) else Option(false))
+    * res1: Option[Boolean] = Some(true)
+    *
+    * scala> F.existsM(List(1,2,3,4))(n => if (n <= 2) Option(true) else None)
+    * res2: Option[Boolean] = None
+    * }}}
+    */
+  def existsM[G[_], A](fa: F[A])(p: A => G[Boolean])(implicit G: Monad[G]): G[Boolean] =
+    foldM(fa, false)((z, a) => G.map(p(a))(b => b || z))
+
+  /**
+    * Check whether all elements satisfy the predicate.
+    *
+    * If there are no elements, the result is `true`.
+    *
+    * All effects are sequenced, i.e. later effects aren't skipped even if a `false` is encountered.
+    *
+    * For example:
+    *
+    * {{{
+    * scala> import cats.implicits._
+    * scala> val F = Foldable[List]
+    * scala> F.forallM(List(1,2,3,4))(n => Option(n <= 4))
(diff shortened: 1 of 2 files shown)


kailuowang:
I think if you use Fordable.Source and `tailRecM` directly rather than `foldM`, you should be able to shortcircuit
e.g. 
```scala
  def existsM[G[_], A](fa: F[A])(p: A => G[Boolean])(implicit G: Monad[G]): G[Boolean] = {
    val src = Foldable.Source.fromFoldable(fa)(self)
    G.tailRecM(src) { srcc => srcc.uncons match {
      case Some((a, src)) => G.map(p(a))(bb => if (bb) Right(true) else Left(src.value))
      case None => G.pure(Right(false))
    }}
  }
```

core/src/main/scala/cats/Foldable.scala
```diff
@@ -327,6 +327,56 @@ import simulacrum.typeclass
...
+    * For example:
+    *
+    * {{{
+    * scala> import cats.implicits._
+    * scala> val F = Foldable[List]
+    * scala> F.existsM(List(1,2,3,4))(n => Option(n <= 4))
+    * res0: Option[Boolean] = Some(true)
+    *
+    * scala> F.existsM(List(1,2,3,4))(n => if (n <= 2) Option(true) else Option(false))
+    * res1: Option[Boolean] = Some(true)
```
kailuowang: might be better to have a negative test case here, e.g.
```
   * scala> F.existsM(List(1,2,3,4))(n => if (n > 4) Option(true) else Option(false))
   * res1: Option[Boolean] = Some(false)
```

refried:
@kailuowang Yes, but do we want it to short-circuit and skip effects?  I could imagine an argument either way.  For my use case, short-circuiting would probably be fine, but I didn’t want to presume that for everyone.  IIRC scalaz's intentionally doesn’t shortcircuit, but I don’t know the rationale.

kailuowang:
I think there is definitely a use case for short-circuiting implementation. If we have to pick from one of the two, I would argue we should pick the short-circuit one, because the other one isn't that cumbersome to write on use-site if needed. How about we ask for more opinions?

refried:
100% agreement.  I meant to go back and say that short-circuiting would probably be more efficient in my use case; since the effects typically involve extra throwaway computation.

refried:
@kailuowang I was mistaken when I said scalaz doesn't short-circuit; it does:
```
scala> import scalaz._, Scalaz._
import scalaz._
import Scalaz._

scala> List(Some(true), None).anyM(identity)
res0: Option[Boolean] = Some(true)

scala> List(Some(true), None).allM(identity)
res1: Option[Boolean] = None
```
I'll redo the PR when I get a chance.

kailuowang:
err. not sure if you need to close this PR. If you replace with the implementation I gave it shortcircuit and pass the tests.

refried:
Ok, I wasn't sure if I should create a new one, or rebase or what.

core/src/main/scala/cats/Foldable.scala
```diff
@@ -327,6 +327,79 @@ import simulacrum.typeclass
...
+    *
+    * scala> F.forallM(List(1,2,3,4))(n => if (n <= 2) Option(false) else None)
+    * res3: Option[Boolean] = Some(false)
+    *
+    * scala> F.forallM(List(1,2,3,4))(n => if (n <= 2) None else Option(false))
+    * res4: Option[Boolean] = None
+    * }}}
+    */
+  def forallM[G[_], A](fa: F[A])(p: A => G[Boolean])(implicit G: Monad[G]): G[Boolean] = {
+    val src = Foldable.Source.fromFoldable(fa)(self)
```
johnynek: This isn't being used right.

refried: I meant to delete line 393, but do you want to elaborate?

refried:
I will add an example to the `tut` docs.  Sorry I didn't think of it sooner.

PR #1788, opened 2017-08-05
Add Eq docs

LukaJCB:
Adds some basic documentation for `Eq`. I'm not sure if this is enough, or if more would complicate things and turn off users, would love to hear some feedback :)

### DIFF
--- a/docs/src/main/tut/typeclasses/eq.md
+++ b/docs/src/main/tut/typeclasses/eq.md
@@ -67,4 +67,16 @@ implicit val eqFoo: Eq[Foo] = Eq.fromUniversalEquals
 Foo(10, "") === Foo(10, "")
 ```
 
+You can even go one step further and make use of the fact, that every case class will extend `scala.Product`, by creating an `Eq` instance for it.
+This means, you'll get an instance for all your case classes, but you'll have to define them for regular classes and objects yourself.
+```tut:book
+
+case class Bar(a: Double, b: Int)
+
+implicit def eqProduct[A <: Product]: Eq[A] = Eq.fromUniversalEquals
+
+Bar(2, 0) === Bar(2, 0)
+```
+
+
 For an example using Kittens check out the [kittens repo](https://github.com/milessabin/kittens).
--- a/docs/src/main/tut/typeclasses/eq.md
+++ b/docs/src/main/tut/typeclasses/eq.md
@@ -67,15 +67,5 @@ implicit val eqFoo: Eq[Foo] = Eq.fromUniversalEquals
 Foo(10, "") === Foo(10, "")
 ```
 
-You can even go one step further by making this implicitly available for all `scala.Product`s (which include case classes).
-```tut:book
-
-case class Bar(a: Double, b: Int)
-
-implicit def eqProduct[A <: Product]: Eq[A] = Eq.fromUniversalEquals
-
-Bar(2, 0) === Bar(2, 0)
-```
-
 
 For an example using Kittens check out the [kittens repo](https://github.com/milessabin/kittens).
--- a/docs/src/main/tut/typeclasses/eq.md
+++ b/docs/src/main/tut/typeclasses/eq.md
@@ -48,4 +48,23 @@ import cats.implicits._
 ```
 
 Implementing `Eq` instances yourself for every data type might seem like huge drawback compared to only slight gains of typesafety.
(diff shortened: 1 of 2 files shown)


docs/src/main/tut/typeclasses/eq.md
```diff
@@ -0,0 +1,51 @@
...
+
+```tut:book
+import cats.implicits._
+
+1 === 1
+
+"Hello" =!= "World"
+```
+
+Implementing `Eq` instances yourself for every data type might seem like huge drawback compared to only slight gains of typesafety.
```
kailuowang: we could mention [`fromUniversalEquals`](https://github.com/typelevel/cats/blob/master/kernel/src/main/scala/cats/kernel/Eq.scala#L97) which should work fine for case classes.

LukaJCB: Done :)

gabro: not sure this should go in the documentation, but I recently found this useful:

```scala
implicit def productEq[A <: Product]: Eq[A] = Eq.fromUniversalEquals
```

unless I'm missing something (and if so, please let me know!) `fromUniversalEqual` should work fine with any `Product`. This makes incredibly easy to provide a reasonable `Eq` instance for every case class.

LukaJCB: That's pretty interesting, seems like it would work for all case classes! Any known downsides?

gabro: not that I know. We're using it in a project since a few weeks and so far so good.

That said, I'm far from being an expert :)

kailuowang: I agree it's worth mentioning in the documentation.

LukaJCB: Added :)

docs/src/main/tut/typeclasses/eq.md
```diff
@@ -0,0 +1,82 @@
...
+case class Foo(a: Int, b: String)
+
+
+implicit val eqFoo: Eq[Foo] = Eq.fromUniversalEquals
+
+
+Foo(10, "") === Foo(10, "")
+```
+
+You can even go one step further and make use of the fact, that every case class will extend `scala.Product`, by creating an `Eq` instance for it.
```
kailuowang: I think this mechanism is probably good enough for all `Product`, so it doesn't limit to case classes. How about:

You can even go one step further by making this implicitly available for all `scala.Product` (which include case classes). 
```tut:book
implicit def eqProduct[A <: Product]: Eq[A] = Eq.fromUniversalEquals
```

docs/src/main/tut/typeclasses/eq.md
```diff
@@ -0,0 +1,81 @@
...
+layout: docs
+title:  "Eq"
+section: "typeclasses"
+source: "kernel/src/main/scala/cats/kernel/Eq.scala"
+scaladoc: "#cats.kernel.Eq"
+---
+
+# Eq
+
+Show is an alternative to the standard Java `equals` method.
```
kailuowang: s/Show/`Eq` ?

LukaJCB: Whooops!

docs/src/main/tut/typeclasses/eq.md
```diff
@@ -0,0 +1,81 @@
...
+
+Foo(10, "") === Foo(10, "")
+```
+
+You can even go one step further by making this implicitly available for all `scala.Product`s (which include case classes).
+```tut:book
+
+case class Bar(a: Double, b: Int)
+
+implicit def eqProduct[A <: Product]: Eq[A] = Eq.fromUniversalEquals
```
ceedubs: Sorry to disagree with a couple of other commenters, but personally I wouldn't be too inclined to include this example in the docs. A couple of significant difference between this and kittens derivation is that kittens will only derive the `Eq` instance when all types within the case class have `Eq` instances, and it will be consistent with those individual `Eq` instances, while this relies on universal equality so it won't necessarily.

I like the `fromUniversalEquals` method and use it on occasion, but I think it's a big leap to add this instance into implicit scope for _all_ `Product` types. I don't want to be a stick in the mud though, so if others want to go forward with this in the docs, that's okay.

kailuowang: that's a really good point. It didn't come to my mind that fromUniversalEquals wont recursively use `Eq` for fields. Agree we should remove this example usage.

PR #1789, opened 2017-08-05
Add documentation for `Show`

LukaJCB:
This is still WIP as, `kittens` doesn't currently support `Show` (I hope to create PR soonish, https://github.com/milessabin/kittens/issues/40) @kailuowang also expressed interest :)

### DIFF
--- a/docs/src/main/tut/typeclasses/show.md
+++ b/docs/src/main/tut/typeclasses/show.md
@@ -37,5 +37,38 @@ def show[A](f: A => String): Show[A]
 def fromToString[A]: Show[A]
 ```
 
+These can be used like this:
+
+```tut:book
+import cats.Show
+
+case class Person(name: String, age: Int)
+
+implicit val showPerson: Show[Person] = Show.show(person => person.name)
+
+case class Department(id: Int, name: String)
+
+implicit val showDep: Show[Department] = Show.fromToString
+```
+
+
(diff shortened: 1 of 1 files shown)


docs/src/main/tut/typeclasses/show.md
```diff
@@ -7,3 +7,35 @@ scaladoc: "#cats.Show"
...
+```scala
+/** creates an instance of Show using the provided function */
+def show[A](f: A => String): Show[A]
+
+/** creates an instance of Show using object toString */
+def fromToString[A]: Show[A]
+```
+
+This still may not seem useful to you, because case classes already automatically implement `toString`, while `show` would have to be implemented manually for each case class.
+Thankfully with the help of a small library called [kittens](https://github.com/milessabin/kittens) a lot of type class instances including `Show` can be derived automatically!
```
kailuowang: So I guess your PR to add `show` to kittens is almost ready? BTW, in case you haven't seen this, miles wrote a shapeless example deriving a `Show` https://github.com/milessabin/shapeless/blob/master/examples/src/main/scala/shapeless/examples/shows.scala

LukaJCB: Oh wow, I didn't know about LabelledTypeClass, thank you so much! :)

kailuowang:
kittens 1.0.0-RC0 is released with `Show` derivation (even added the example to the README).

kailuowang:
👍

PR #1790, opened 2017-08-06
Implement EitherT#leftFlatMap and EitherT#leftSemiflatMap

vendethiel:
Modeled after the right versions.

### DIFF
--- a/core/src/main/scala/cats/data/EitherT.scala
+++ b/core/src/main/scala/cats/data/EitherT.scala
@@ -96,6 +96,18 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
 
   def leftMap[C](f: A => C)(implicit F: Functor[F]): EitherT[F, C, B] = bimap(f, identity)
 
+  def leftFlatMap[BB >: B, D](f: A => EitherT[F, D, BB])(implicit F: Monad[F]): EitherT[F, D, BB] =
+    EitherT(F.flatMap(value) {
+      case Left(a) => f(a).value
+      case r@Right(_) => F.pure(r.leftCast)
+    })
+
+  def leftSemiflatMap[D](f: A => F[D])(implicit F: Monad[F]): EitherT[F, D, B] =
+    EitherT(F.flatMap(value) {
+      case Left(a) => F.map(f(a)) { d => Left(d) }
+      case r@Right(_) => F.pure(r.leftCast)
+    })
+
   def compare(that: EitherT[F, A, B])(implicit o: Order[F[Either[A, B]]]): Int =
     o.compare(value, that.value)
 
--- a/tests/src/test/scala/cats/tests/EitherTTests.scala
+++ b/tests/src/test/scala/cats/tests/EitherTTests.scala
@@ -434,4 +434,16 @@ class EitherTTests extends CatsSuite {
     } yield s1 ++ s2
   }
 
+  test("leftFlatMap") {
+    forAll { (eithert: EitherT[List, String, Int], f: String => String) =>
+      eithert.leftFlatMap(v => EitherT.left[Int](List(f(v)))) should ===(eithert.leftMap(f))
+    }
+  }
+
+  test("leftSemiflatMap") {
+    forAll { (eithert: EitherT[List, String, Int], f: String => String) =>
+      eithert.leftSemiflatMap(v => List(f(v))) should ===(eithert.leftMap(f))
+    }
+  }
+
 }
--- a/tests/src/test/scala/cats/tests/EitherTTests.scala
+++ b/tests/src/test/scala/cats/tests/EitherTTests.scala
@@ -438,11 +438,17 @@ class EitherTTests extends CatsSuite {
     forAll { (eithert: EitherT[List, String, Int], f: String => String) =>
       eithert.leftFlatMap(v => EitherT.left[Int](List(f(v)))) should ===(eithert.leftMap(f))
     }
+    forAll { (eithert: EitherT[List, String, Int], f: String => EitherT[List, String, Int]) =>
+      eithert.leftFlatMap(f) should ===(eithert.swap.flatMap(a => f(a).swap).swap)
+    }
   }
 
   test("leftSemiflatMap") {
     forAll { (eithert: EitherT[List, String, Int], f: String => String) =>
       eithert.leftSemiflatMap(v => List(f(v))) should ===(eithert.leftMap(f))
+    }
+    forAll { (eithert: EitherT[List, String, Int], f: String => List[String]) =>
+      eithert.leftSemiflatMap(f) should ===(eithert.swap.semiflatMap(a => f(a)).swap)
     }
   }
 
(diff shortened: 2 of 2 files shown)


kailuowang:
thanks very much! 👍

core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -96,6 +96,15 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
...
   def leftMap[C](f: A => C)(implicit F: Functor[F]): EitherT[F, C, B] = bimap(f, identity)
 
+  def leftFlatMap[BB >: B, D](f: A => EitherT[F, D, BB])(implicit F: Monad[F]): EitherT[F, D, BB] =
+    EitherT(F.flatMap(value) {
+      case Left(a) => f(a).value
+      case r@Right(_) => F.pure(r.leftCast)
+    })
+
+  def leftSemiflatMap[D](f: A => F[D])(implicit F: Monad[F]): EitherT[F, D, B] =
+    leftFlatMap(a => EitherT.left(f(a)))
```
johnynek: this is going to allocate the `EitherT` just to call `.value` above. Why not just inline the implementation:
```scala
EitherT(F.flatMap(value) {
  case Left(a) => f(a).map { d => Left(d) }
  case r@Right(_) => F.pure(r.leftCast)
})
```

vendethiel: Should I rewrite `leftFlatMap` in terms of `leftSemiflatMap` then?

vendethiel:
@johnynek Back from holidays. Did just that.

jrduncans:
Seems like there should be a `leftFlatMap` for `f: (A) ⇒ F[Either[D, BB]]` to fully match `flatMap`.

vendethiel:
You mean `leftFlatMapF`?

jrduncans:
Oh, yes, that's correct.

non:
I'd like to have laws for these in terms of `swap` and the right-biased methods, e.g.

```scala
et.leftFlatMap(a => f(a)) = et.swap.flatMap(a => f(a).swap)
```

Does that seem reasonable?

vendethiel:
@non I tried to do it at first, but I missed one swap. It seems to work with one extra:

```scala
et.leftFlatMap(a => f(a)) <-> et.swap.flatMap(a => f(a).swap).swap
```

non:
@vendethiel Let's add these sorts of tests in addition to what you already have. I'd go ahead and write tests like this for all the left-biased methods that have a right-biased equivalent.

vendethiel:
Okay, what about `leftFlatMapF`, do we want that?

core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -96,6 +96,18 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
 
   def leftMap[C](f: A => C)(implicit F: Functor[F]): EitherT[F, C, B] = bimap(f, identity)
 
+  def leftFlatMap[BB >: B, D](f: A => EitherT[F, D, BB])(implicit F: Monad[F]): EitherT[F, D, BB] =
+    EitherT(F.flatMap(value) {
+      case Left(a) => f(a).value
+      case r@Right(_) => F.pure(r.leftCast)
```
LukaJCB: Tiny nitpick, but can we change this to `case r: Right => ...` since we don't actually want to extract anything?

core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -96,6 +96,18 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
...
+    })
+
+  def leftSemiflatMap[D](f: A => F[D])(implicit F: Monad[F]): EitherT[F, D, B] =
+    EitherT(F.flatMap(value) {
+      case Left(a) => F.map(f(a)) { d => Left(d) }
+      case r@Right(_) => F.pure(r.leftCast)
```
LukaJCB: Ditto, as above :)

core/src/main/scala/cats/data/EitherT.scala
adelbertc: While in practice I don't think it matters, the use of `:` implies type casing which is generally used in contexts where parametricity is violated: https://typelevel.org/blog/2014/11/10/why_is_adt_pattern_matching_allowed.html

I could have sworn there were some cases where `:` was allowed and ADT patmat wasn't but I can't remember.

LukaJCB: Very interesting article, thanks a lot! :) 
I always thought that scalac transforms `case t: Type` into an `isInstanceOf` check whereas `case Type(_)` would be turned into `Type.unapply(...)`, but that article shows that my assumption was false. 
It seems then, there's no real benefit to my suggestion, so disregard.

kailuowang:
@vendethiel if you waiting for an answer for `leftFlatMapF` I'd say let's address that in a different PR.

vendethiel:
Okay, thanks. Then I think the PR is ready?

kailuowang:
I still think the tests @non suggested are valuable to add.

tests/src/test/scala/cats/tests/EitherTTests.scala
```diff
@@ -434,4 +434,22 @@ class EitherTTests extends CatsSuite {
     } yield s1 ++ s2
   }
 
+  test("leftFlatMap") {
+    forAll { (eithert: EitherT[List, String, Int], f: String => String) =>
```
LukaJCB: Could we name this one "leftFlatMap consistent with leftMap"?

tests/src/test/scala/cats/tests/EitherTTests.scala
```diff
@@ -434,4 +434,22 @@ class EitherTTests extends CatsSuite {
...
+      eithert.leftFlatMap(v => EitherT.left[Int](List(f(v)))) should ===(eithert.leftMap(f))
+    }
+    forAll { (eithert: EitherT[List, String, Int], f: String => EitherT[List, String, Int]) =>
```
LukaJCB: How about naming this one "leftFlatMap consistent with swap and then flatMap"?

tests/src/test/scala/cats/tests/EitherTTests.scala
```diff
@@ -434,4 +434,22 @@ class EitherTTests extends CatsSuite {
...
+      eithert.leftFlatMap(f) should ===(eithert.swap.flatMap(a => f(a).swap).swap)
+    }
+  }
+
+  test("leftSemiflatMap") {
```
LukaJCB: Same as above for these two tests :)

vendethiel:
Somehow missed the notifications! Sorry.

LukaJCB:
Don't be, it's all good! Thanks very much for your contribution! 😊

kailuowang:
👍  Thanks very much indeed!

vendethiel:
Thanks! 🍰

Issue #1791, opened 2017-08-06
Automate migration to 1.0 with Scalafix


gabro:
Now that the 1.0 is close, I got a chance to take a look at the breaking changes it introduces and I noticed that the vast majority of them are well within the possibilities of Scalafix.

So, well, why not just automate the migration away?
I started experimenting on it and it's promising! So far I got:

- removal of `Unapply` machinery (just a bunch of symbol renames, but, hey, I wanted to start easy)
- removal of `CartesianBuilder` syntax (this is trickier, but the rewrite is still around 50 lines of code)

Would you be interested in this?

If so, I think it would be really cool to ship the rewrites with the final release of 1.0.0 so that anyone can access them easily simply by having cats as a dependency. I'll push my work in progress soon, but in the meantime here's some diff I was able to fully automate:

```diff
   import cats.instances.option._
-  import cats.syntax.cartesian._
+  import cats.syntax.apply._
   val o1: Option[Int] = Some(42)
   val o2: Option[String] = Some("hello")
   val o3: Option[Int] = Some(2)
-  (o1 |@| o2).map((i: Int, s: String) => i.toString ++ s)
-  (o1 |@| o2).tupled
-  (o1 |@| o2 |@| o3).map(_ + _ + _)
+  (o1, o2).mapN((i: Int, s: String) => i.toString ++ s)
+  (o1, o2).tupled
+  (o1, o2, o3).mapN(_ + _ + _)

   import cats.{Semigroup, Eq}
   import cats.instances.string._
   import cats.instances.double._
   case class Foo(a: String, c: List[Double])

-  (Semigroup[String] |@| Semigroup[List[Double]]).imap(Foo.apply)(Function.unlift(Foo.unapply))
-
-  (Eq[Double] |@| Eq[String]).contramap { (a: Foo) => (2, "bar") }
+  (Semigroup[String], Semigroup[List[Double]]).imapN(Foo.apply)(Function.unlift(Foo.unapply))
+  (Eq[Double], Eq[String]).contramapN { (a: Foo) => (2, "bar") }
```

A couple of more things to notice:
- the rewrites are semantic, meaning they're using the semantic information not just the syntactic ones (e.g. when renaming `map` we're looking for `_root_.cats.syntax.CartesianBuilder#CartesianBuilder2.map.`, not just any `map`)
- I'm currently targeting the 0.9.0 -> 1.0.0 migration, although this would probably be easy to extend to prior versions

kailuowang:
[...]

PR #1793, opened 2017-08-06
Add Scalafix rewrites for 1.0.0

gabro:
This is a WIP for #1791.

It's very preliminary, but feel free to share your thoughts, try this out and add more edge cases to the tests!


kailuowang:
This is awesome. I suggest that we make it available to more user sooner. 
If we can add a quick step by step instruction on how to use this, we can merge it now and link it from the migration guide. It doesn't have to include migrations for all the breaking changes. We can add those incrementally - that would also allow more people to contribute. We just need to communicate that this is an incomplete auto-migration. 
What do you think?

gabro:
I pushed another easy one (`RenameFreeSuspend`) and refactored some code a bit.

@kailuowang I agree that merging soon would be a good idea. About the strategy to make this available to the users I would like @olafurpg's opinion, since I think this is the first real-world scenario of making a rewrite available with a library and things are still a bit sketchy.

olafurpg:
Great work @gabro!

The simplest way to share a rewrite at the moment is to put it in a single source file under `scalafix/rewrites/src/main/scala/Cats_v1.scala` and users can run it from the sbt shell with `> scalafix github:typelevel/cats/v1` with the sbt-scalafix plugin. It's possible to download the file locally and run `> scalafix file:path/to/Cats_v1.scala` if people prefer that.

Alternatively, it's possible to publish the rewrites to Maven. As long as the artifacts are on the same classpath as scalafix then users can run the rewrites with `> scalafix scala:fix.Cats_v1` (assuming rewrite fqn is `fix.Cats_v1`).

The sbt plugin is a thin shim over the scalafix cli so it should be possible to integrate with other build tools than sbt. Pants has a scalafix integration, but I'm not aware of plugins for gradle or maven.

I hope to release scalafix 0.5.0 later this week, I addressing a few issues in the sbt plugin first.

kailuowang:
thanks @gabro! I agree `sbt-scalafix` running using github src is the easiest.

gabro:
I just had a chat with @olafurpg about the best strategy for this, and we came up with the following solution (using scalafix 0.5):

- we'll keep all the rewrites separated, so that people can run them in isolation in case anything goes wrong
- I'll generate a single file called `Cats_v1_0_0.scala` (or something like that) that's just the concatenation of all the single rewrites in a single file. This is currently the most convenient way of running all the rewrites with scalafix, although it involves duplicating the source (single rewrites + generated file)

This will allow users to run all the rewrites using the scalafix sbt plugin: `sbt scalafix github:typelevel/cats/v1_0_0` 🎉 

We also drafted a solution for future versions of scalafix for making the process much simpler for both the rewrite authors and the users. @olafurpg will provide more details soon :)

gabro:
My bad, I got some parts of that mixed up. There won't be any duplication, just a single `Cats_v1_0_0.scala` file containing all the rewrites. Coming soon :)

olafurpg:
Any recommendations for OSS repos where we can try out the migration?

kailuowang:
https://github.com/stew/dogs might be a good candidate.

gabro:
@kailuowang I've covered a few corner case of the available rewrites and added some instructions on how to run the fixes. Take a look ;)

gabro:
Also, once this gets merged I would suggest to use the original issue as an aggregation point for bug reports and general feedback. I expect *a lot* of corner cases will be found.

kailuowang:
Looking great to me, thanks so much! :+1:

kailuowang:
Would love to get this merge and start to benefit users and iterate.

milessabin:
LGTM. I'd like to see this merged ASAP :-)

scalafix/project/plugins.sbt
```diff
@@ -0,0 +1,5 @@
+addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.5.0-M1")
```
pepegar: This line is repeated :)

scalafix/project/plugins.sbt
```diff
@@ -0,0 +1,5 @@
...
+addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.5.0-M1")
```
peterneyens: Once is good enough, right?

gabro: most definitely! Good catch, thanks :)

scalafix/project/plugins.sbt
gabro: good catch, thank you!

gabro:
Updated, I've removed the duplicated line in plugins and bumped scalafix to the latest 0.5.0-M2 version.

PR #1796, opened 2017-08-07
add Foldable.existsM/forallM to docs

refried:
Not sure if these examples are unnecessarily complicated?  I considered using `List[Option[Boolean]]` and `identity` instead, but I thought that might be confusing too.  Thanks for any feedback.


kailuowang:
Straightforward docs addition, merged with one sign off

PR #1797, opened 2017-08-08
Add detail to documentation for scala.cats.Foldable#foldRight.

aholland:
IMHO documentation of scala.cats.Foldable#foldRight was not sufficiently detailed for users of the API confidently to implement a lazy right fold. I hope my changes remedy that.

### DIFF
--- a/core/src/main/scala/cats/Foldable.scala
+++ b/core/src/main/scala/cats/Foldable.scala
@@ -39,6 +39,30 @@ import simulacrum.typeclass
    * Eval[B]` to support laziness in a stack-safe way. Chained
    * computation should be performed via .map and .flatMap.
    *
+   * Because this fold is designed to handle a potentially infinite
+   * sequence, it does not construct the result by starting with the last element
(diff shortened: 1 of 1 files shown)


core/src/main/scala/cats/Foldable.scala
```diff
@@ -39,6 +39,32 @@ import simulacrum.typeclass
    * Eval[B]` to support laziness in a stack-safe way. Chained
    * computation should be performed via .map and .flatMap.
    *
+   * Because this fold is designed to handle a potentially infinite
+   * sequence, it does not construct the result by starting with the last element
+   * and working back towards the head of the sequence (as a regular right fold would).
```
ceedubs: I think that this statement is a bit misleading. In general right-associated folds can support infinite collections while left-associated folds cannot, so I don't think that there would be any thing "regular" about a right fold that starts with the last element.

easel:
This is valuable additional detail to the docs for `foldRight`, @aholland any chance you can update this taking into account the feedback from @ceedubs ?

kailuowang:

Issue #1786, opened 2017-08-05
Add transformS() in ReaderWriterStateT


ronanM:
Add `transformS()` in `ReaderWriterStateT` like `cats.data.StateT.transformS`

`def transformS[R](f: R => S, g: (R, S) => R)(implicit F: Functor[F]): StateT[F, R, A]`.

kailuowang:
related #1775

ronanM:
> related #1775

Does `ReaderWriterStateT` will automatically gain `transformS()` from `IndexedStateT` ?

iravid:
@ronanM see https://github.com/typelevel/cats/pull/1803; added there

PR #1803, opened 2017-08-08
Convert ReaderWriterStateT to IndexedReaderWriterStateT

### LINKED ISSUE #1774: Provide an IndexedReaderWriterStateT data type

iravid:
Resolves #1774, #1786.

In favour of consistency with #1775, the following has changed:
- `transformS` has been added with a similar signature
- `IRWST.contramap` now transforms the input state and so does the `Contravariant` instance, accordingly
- `IRWST.dimap` now transforms the input and output states, and so does the `Profunctor` instance
- `IRWST.bimap` and a `Bifunctor` instance has been added to transform the resulting output state and output value


iravid:
All feedback appreciated, folks

non:
One small thing: I think we should rename the file this is in to match its name (e.g. rename `ReaderWriterStateT.scala` to `IndexedReaderWriterStateT.scala`).

iravid:
Thanks for the comment - I’ll do that once review finishes, as it makes the diffs much harder to read :-)

core/src/main/scala/cats/data/ReaderWriterStateT.scala
```diff
@@ -318,78 +412,91 @@ private[data] abstract class RWSFunctions {
     ReaderWriterStateT.tell(l)
 }
 
-private[data] sealed trait RWSTInstances extends RWSTInstances1 {
-  implicit def catsDataProfunctorForRWST[F[_], L, S](implicit F0: Functor[F]): Profunctor[ReaderWriterStateT[F, ?, L, S, ?]] =
-    new RWSTProfunctor[F, L, S] {
+private[data] sealed trait IRWSTInstances extends IRWSTInstances1 {
```
kailuowang: per #1879 do you mind change all the instance traits here into abstract class? I am creating a PR to change other classes, but don't want to touch this one to avoid conflict.

iravid: Done

iravid:
Ping. This has been sitting here for a while. Any chance we could move this forward?

core/src/main/scala/cats/data/ReaderWriterStateT.scala
```diff
@@ -1,59 +1,84 @@
...
- * an accumulated log `L` and a result `A`.
+ * Represents a stateful computation in a context `F[_]`, from state `SA` to state `SB`,
+ *  with an initial environment `E`, an accumulated log `L` and a result `A`.
  *
  * In other words, it is a pre-baked stack of `[[ReaderT]][F, E, A]`, `[[WriterT]][F, L, A]`
- * and `[[StateT]][F, S, A]`.
+ * and `IndexedStateT[F, SA, SB, A]`.
  */
-final class ReaderWriterStateT[F[_], E, L, S, A](val runF: F[(E, S) => F[(L, S, A)]]) extends Serializable {
+final class IndexedReaderWriterStateT[F[_], E, L, SA, SB, A](val runF: F[(E, SA) => F[(L, SB, A)]]) extends Serializable {
```
johnynek: it feels like this could have an `Arrow[SA, SB]` instance now. Am I wrong?

I don't know if it is worth it, but it if it can have an instance, it would be nice to add (same may be true for `IndexedStateT`).

iravid: Hmm, we can definitely do `Compose` and `Strong`. 

`Category` (`id`) and `Arrow` (`lift`) are a bit more problematic. We need to conjure up the value somehow. That could work for `V: Monoid`.

I'll add `Compose`+`Strong` and `Arrow` as separate traits. Thanks for the suggestion

iravid: @johnynek Added the instances both here and in the IndexedStateT PR.

johnynek:
👍

kailuowang:
LGTM, I guess it's time to change the filenames.

iravid:
Hmm, I see that coverage has decreased somehow. I'll fix that.

iravid:
Ok, 100% diff coverage, we should be good to go :-)

kailuowang:
Thanks so much @iravid. Sorry it dragged so long.

iravid:
Thanks @kailuowang, looking forward to using this when 1.0.0-RC1 is out :-)

PR #1804, opened 2017-08-08
workaround for a possible scala bug in show for value class. 

### LINKED ISSUE #1802: java.lang.ClassFormatError: Duplicate method name&signature when upgrading to cats 1.0.0-MF
Hi all,

When @sellout tried to upgrade newts (https://github.com/julien-truffaut/newts) from cats 0.9.0 to 1.0.0-MF, we started to experience `java.lang.ClassFormatError` when calling `show` on case class using `AnyVal` e.g.:

```
final case class First[A](getFirst: A) extends AnyVal
object First {
  implicit def showInstance[A](implicit ev: Show[A]): Show[First[A]] = new Show[First[A]] {
    override def show(f: First[A]): String = s"First(${ev.show(f.getFirst)})"
  }
}

First("aString").show
```

here is the exact error:
```
[info] newts.FirstTest *** ABORTED ***
[info]   java.lang.ClassFormatError: Duplicate method name&signature in class file newts/First$$anon$1
[info]   at java.lang.ClassLoader.defineClass1(Native Method)
[info]   at java.lang.ClassLoader.defineClass(ClassLoader.java:760)
```

Now what's interesting is that the error disappear if you:

- remove `Anyval` or
- change uses `Show.show` e.g. 
```
implicit def showInstance[A](implicit ev: Show[A]): Show[First[A]] =  Show.show(a =>
    s"First(${Show[A].show(a.getFirst)})"
  )
```

For precision, everything worked fine in 0.9.0

kailuowang:
fixes #1802

### DIFF
--- a/core/src/main/scala/cats/Show.scala
+++ b/core/src/main/scala/cats/Show.scala
@@ -13,11 +13,12 @@ import cats.functor.Contravariant
 @typeclass trait Show[T] extends Show.ContravariantShow[T] {
   // duplicated so simulacrum knows it requires an instance of this trait
   def show(t: T): String
+  def _show(t: T): String = show(t)
 }
 
 object Show {
   trait ContravariantShow[-T] {
-    def show(t: T): String
+    def _show(t: T): String
   }
 
   /** creates an instance of [[Show]] using the provided function */
@@ -32,7 +33,7 @@ object Show {
 
   final case class Shown(override val toString: String) extends AnyVal
   object Shown {
-    implicit def mat[A](x: A)(implicit z: ContravariantShow[A]): Shown = Shown(z show x)
+    implicit def mat[A](x: A)(implicit z: ContravariantShow[A]): Shown = Shown(z _show x)
   }
 
   final case class ShowInterpolator(_sc: StringContext) extends AnyVal {
--- a/core/src/main/scala/cats/Show.scala
+++ b/core/src/main/scala/cats/Show.scala
@@ -13,12 +13,15 @@ import cats.functor.Contravariant
 @typeclass trait Show[T] extends Show.ContravariantShow[T] {
   // duplicated so simulacrum knows it requires an instance of this trait
   def show(t: T): String
-  def _show(t: T): String = show(t)
+  def cshow(t: T): String = show(t)
 }
 
 object Show {
   trait ContravariantShow[-T] {
-    def _show(t: T): String
+    /**
+     * using a different name than `show` as a work around of scala bug https://github.com/scala/bug/issues/10458
+     */
+    def cshow(t: T): String
   }
 
   /** creates an instance of [[Show]] using the provided function */
@@ -33,7 +36,7 @@ object Show {
 
   final case class Shown(override val toString: String) extends AnyVal
   object Shown {
-    implicit def mat[A](x: A)(implicit z: ContravariantShow[A]): Shown = Shown(z _show x)
+    implicit def mat[A](x: A)(implicit z: ContravariantShow[A]): Shown = Shown(z cshow x)
   }
 
   final case class ShowInterpolator(_sc: StringContext) extends AnyVal {
--- a/core/src/main/scala/cats/Show.scala
+++ b/core/src/main/scala/cats/Show.scala
@@ -1,6 +1,5 @@
 package cats
 
-import simulacrum.typeclass
 import cats.functor.Contravariant
 
 /**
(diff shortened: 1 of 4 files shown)


dwijnand:
Minimised and reported upstream at https://github.com/scala/bug/issues/10458, if you want to leave a reference in the code.

kailuowang:
Thanks very much! @dwijnand

kailuowang:
Now I run into https://issues.scala-lang.org/browse/SI-6260 which is not fixed on 2.10.6 
@dwijnand any suggestions?

dwijnand:
Drop 2.10 support? :trollface:

Otherwise I suggest seeing if simulacrum can be taught to see the show inherited from ContravariantShow.

core/src/main/scala/cats/Show.scala
```diff
@@ -13,11 +13,12 @@ import cats.functor.Contravariant
 @typeclass trait Show[T] extends Show.ContravariantShow[T] {
   // duplicated so simulacrum knows it requires an instance of this trait
   def show(t: T): String
+  def _show(t: T): String = show(t)
```
johnynek: should this be `final`?

kailuowang: I moved away from this solution.

kailuowang:
I can't get simulacrum to recoganize the super class. So i decided to hand roll the type class boilerplate by simulacrum

core/src/main/scala/cats/Show.scala
```diff
@@ -10,16 +9,32 @@ import cats.functor.Contravariant
...
  * explicitly provided one.
  */
-@typeclass trait Show[T] extends Show.ContravariantShow[T] {
-  // duplicated so simulacrum knows it requires an instance of this trait
-  def show(t: T): String
-}
+trait Show[T] extends Show.ContravariantShow[T]
 
+/**
+ * Hand rolling the type class boilerplate due to https://issues.scala-lang.org/browse/SI-6260 and scala/bug#10458
```
dwijnand: SI-6260 is scala/bug#6260

kailuowang:
so I got rid of the simulacrum and thus the duplicated `show` in `Show`, but I still can't get around SI-6260 in scala 2.10. Not quite sure about it at the moment.  I only see 2 options
1. leave everything as is, i.e. close this PR. The work around isn't that terrible, instead of creating the anonymous class instance of `Show`, use `Show.show` 
2. use the current fix, but give up on 2.10, which means I probably need to remove the regression test to get build passing. 

What do you guys think? @johnynek @julien-truffaut @sellout @edmundnoble @dwijnand

dwijnand:
3. use the current fix, but restrict the test to 2.11+ to avoid SI-6260.

kailuowang:
@dwijnand what would be the best way to restrict the test to 2.11+? I can think of using `sbt-buildinfo` to make scala version available in test code. Is there any other better ways?

dwijnand:
```scala
unmanagedSourceDirectories in Test ++= {
  val ss = (scalaSource in Test).value
  if (CrossVersion.partialVersion(scalaVersion.value) exists (_._2 >= 11))
    Seq(file(ss.getPath + "-2.11+"))
  else
    Nil
}
```

```
> show test:unmanagedSourceDirectories
[info] * /s/t-test211/src/test/scala-2.12
[info] * /s/t-test211/src/test/scala
[info] * /s/t-test211/src/test/java
[info] * /s/t-test211/src/test/scala-2.11+
```

kailuowang:
thanks @dwijnand in scalajs enabled projects the `scalaSource` is 
`.jvm/src/test/scala`, which means that the settings you gave added `/cats/tests/.jvm/src/test/scala-2.11+` instead of 
`/cats/tests/src/test/scala-2.11+`. I am not sure if this test should be jvm specific. 
I wonder if there is a key pointing to `/cats/tests/src/test/scala` to use instead of `scalaSource`?

dwijnand:
Hmm, that makes things more complicated :)

```scala
unmanagedSourceDirectories in Test ++= {
  val bd = baseDirectory.value
  if (CrossVersion.partialVersion(scalaVersion.value) exists (_._2 >= 11))
    CrossType.Pure.sharedSrcDir(bd, "test").toList map (f => file(f.getPath + "-2.11+"))
  else
    Nil
}
```

```
> show tJVM/test:unmanagedSourceDirectories
[info] * /s/t-test211/.jvm/src/test/scala-2.12
[info] * /s/t-test211/.jvm/src/test/scala
[info] * /s/t-test211/.jvm/src/test/java
[info] * /s/t-test211/src/test/scala-2.12
[info] * /s/t-test211/src/test/scala
[info] * /s/t-test211/src/test/scala-2.11+

> show tJS/test:unmanagedSourceDirectories
[info] * /s/t-test211/.js/src/test/scala-2.12
[info] * /s/t-test211/.js/src/test/scala
[info] * /s/t-test211/.js/src/test/java
[info] * /s/t-test211/src/test/scala-2.12
[info] * /s/t-test211/src/test/scala
[info] * /s/t-test211/src/test/scala-2.11+
```

kailuowang:
Thanks! @dwijnand it would've taken me weeks to figure out. fix pushed.

edmundnoble:
Does this mean that 2.10 users won't be able to use `Show`?

kailuowang:
@edmundnoble , it means that they will use `Show.show` when writing a `Show` instance for a value class

PR #1809, opened 2017-08-09
Add iterateWhileM and iterateUntilM

drbild:
* Adds `iterateWhileM` and `iterateUntilM` to complement the existing `iterateWhile` and `iterateUntil` methods.
* Reimplements `iterateWhile` and `iterateUntil` in terms of the new `iterateWhileM` and `iterateUntilM`.

### DIFF
--- a/core/src/main/scala/cats/Monad.scala
+++ b/core/src/main/scala/cats/Monad.scala
@@ -100,4 +100,23 @@ import simulacrum.typeclass
     }
   }
 
+  /**
+    * Apply a monadic function iteratively until its result fails
+    * to satisfy the given predicate and return that result.
+    */
+  def iterateWhileM[A](init: A)(f: A => F[A])(p: A => Boolean): F[A] =
+    tailRecM(init) { a =>
+      if (p(a))
+        map(f(a))(Left(_))
+      else
+        pure(Right(a))
+    }
+
+  /**
+    * Apply a monadic function iteratively until its result satisfies
+    * the given predicate and return that result.
+    */
+  def iterateUntilM[A](init: A)(f: A => F[A])(p: A => Boolean): F[A] =
+    iterateWhileM(init)(f)(!p(_))
+
 }
--- a/core/src/main/scala/cats/Monad.scala
+++ b/core/src/main/scala/cats/Monad.scala
@@ -76,29 +76,19 @@ import simulacrum.typeclass
    * Execute an action repeatedly until its result fails to satisfy the given predicate
    * and return that result, discarding all others.
    */
-  def iterateWhile[A](f: F[A])(p: A => Boolean): F[A] = {
+  def iterateWhile[A](f: F[A])(p: A => Boolean): F[A] =
     flatMap(f) { i =>
-      tailRecM(i) { a =>
-        if (p(a))
-          map(f)(Left(_))
-        else pure(Right(a))
-      }
+      iterateWhileM(i)(_ => f)(p)
     }
-  }
 
   /**
    * Execute an action repeatedly until its result satisfies the given predicate
    * and return that result, discarding all others.
    */
-  def iterateUntil[A](f: F[A])(p: A => Boolean): F[A] = {
+  def iterateUntil[A](f: F[A])(p: A => Boolean): F[A] =
     flatMap(f) { i =>
-      tailRecM(i) { a =>
-        if (p(a))
-          pure(Right(a))
-        else map(f)(Left(_))
-      }
+      iterateUntilM(i)(_ => f)(p)
     }
-  }
 
   /**
     * Apply a monadic function iteratively until its result fails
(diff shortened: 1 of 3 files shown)


kailuowang:
looks great! Thanks! 👍  pending build pass.

peterneyens:
Just wondering, these `iterateUntil` / `iterateWhile` methods only make sense for something like `Task` or `IO`, right ?
Are there cases where the `f: A => F[A]` would actually return something different every time, so you couldn't use `f(a).iterateUntil(pred)` instead of `a.iterateUntilM(f)(pred)` ?

drbild:
@peterneyens It's a common idiom in Haskell where `A` represents some state and `A => F[A]` is a side-effecting update function and you want to iterate until some terminal state is reached.   

Simple games are a frequent example.  `A` is the current state of the world. `A => IO[A]` is the step function that reads user input, updates the state accordingly, and refreshes the UI.  `A => Boolean` checks for a termination condition like the player dying.

So the main loop becomes something like
```scala
def initialWorld: World = ??? // initial world state

def stepWorld(old): IO[World] = getUserInput >>= updateWorld(old) >>! refreshUI

def keepRunning(current): Boolean = isPlayerAlive(current)

def loop: IO[World] = initialWorld.iterateWhileM(stepWorld)(keepRunning)
```

This example could all be rewritten using `StateT[IO, World, ?]` so that
```scala
def loop = stepWorld.iterateWhile(keepRunning).run(initialWorld)
```
would work.   But pulling in `StateT` often seems like overkill.

drbild:
I think of `iterateWhile` as a specialized version of `iterateWhileM`:
```
iterateWhile : iterateWhileM :: followedBy : flatMap
```

kailuowang:
I agree that some less trivial tests/examples would be nice.

tests/src/test/scala/cats/tests/MonadTest.scala
```diff
@@ -74,4 +74,28 @@ class MonadTest extends CatsSuite {
     result should ===(50000)
   }
 
+  test("iterateWhileM") {
+    forAll(Gen.posNum[Int]) { (max: Int) =>
```
johnynek: might this not run a super long time? any reason why `max` shouldn't be something from like: `Gen.choose(0, 500000)`?

core/src/main/scala/cats/Monad.scala
```diff
@@ -92,12 +88,29 @@ import syntax.either._
...
-      }
+      iterateUntilM(i)(_ => f)(p)
     }
   }
 
+  /**
+    * Apply a monadic function iteratively until its result fails
+    * to satisfy the given predicate and return that result.
+    */
+  def iterateWhileM[A](init: A)(f: A => F[A])(p: A => Boolean): F[A] = {
```
johnynek: no need for the outer `{ }` here.

core/src/main/scala/cats/Monad.scala
```diff
@@ -92,12 +88,29 @@ import syntax.either._
...
+      else
+        pure(a.asRight[A])
+    }
+  }
+
+  /**
+    * Apply a monadic function iteratively until its result satisfies
+    * the given predicate and return that result.
+    */
+  def iterateUntilM[A](init: A)(f: A => F[A])(p: A => Boolean): F[A] = {
```
johnynek: no need for the outer `{ }`

core/src/main/scala/cats/Monad.scala
```diff
@@ -76,28 +76,37 @@ import syntax.either._
...
+  /**
+    * Apply a monadic function iteratively until its result fails
+    * to satisfy the given predicate and return that result.
+    */
+  def iterateWhileM[A](init: A)(f: A => F[A])(p: A => Boolean): F[A] =
+    tailRecM(init) { a =>
+      if (p(a))
+        map(f(a))(_.asLeft[A])
+      else
+        pure(a.asRight[A])
```
peterneyens: Can we replace the `asLeft` and `asRight` with just `Left(..)` and `Right(..)` (in `whileM` as well) ?

kailuowang:
LGTM, @peterneyens @johnynek looks like that the comments were addressed. any other feedback?

drbild:
I've left the `Id`-based example in the scaladoc, because nothing with a non-trivial Monad but simple enough for a scaladoc has come to mind.  I could drop the scaladoc entirely for now.  Some simple iterative algorithm - like Euclid's GCD - but with a possible failure condition would be good. I just haven't been able come up with one.

I'm also working on some additions to the `Monad` doc, using some real examples - finding all solutions of a subset-sum variant and all solutions of the n-queens problem, both in the `List` monad.  Those can go in a different PR though.

kailuowang:
I love the doc ideas, they sound awesome. I'd say we can have them in a separate PR so that we can have the binary breaking change merged in asap.

non:
Looks good to me. 👍

kailuowang:
since all feedback seems addressed and I am going to merge with 3 sign-offs.

PR #1813, opened 2017-08-10
Add RenameInjectProdAndCoproduct, RenameTupleApplySyntax and RemoveSplit Scalafix rewrites

gabro:
This PR adds these rewrites:

- `RenameInjectProdAndCoproduct`
- `RenameTupleApplySyntax`
- `RemoveSplit`

Two of them are super easy thanks to a feature recently landed in Scalafix (`replaceSymbol`). For this reason I also upgraded Scalafix to the latest milestone release for 0.5.0.

### DIFF
--- a/scalafix/input/src/main/scala/fix/v1_0_0/RemoveSplit.scala
+++ b/scalafix/input/src/main/scala/fix/v1_0_0/RemoveSplit.scala
@@ -1,0 +1,17 @@
+/*
+rewrite = "scala:fix.v1_0_0.RemoveSplit"
+ */
+package fix
+package to1_0_0
+
+import cats.implicits._
+import cats.arrow.Split
+import cats.syntax.split._
+
+object RemoveSplitTests {
+  val toLong: Int => Long = _.toLong
+  val toDouble: Float => Double = _.toDouble
+  val f: ((Int, Float)) => (Long, Double) =
+    Split[Function1].split(toLong, toDouble)
+  f((3, 4.0f))
+}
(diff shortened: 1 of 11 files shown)


scalafix/input/src/main/scala/fix/v1_0_0/RemoveSplit.scala
```diff
@@ -0,0 +1,17 @@
+/*
+rewrite = "scala:fix.v1_0_0.RemoveSplit"
+ */
+package fix
+package to1_0_0
+
+import cats.implicits._
+import cats.arrow.Split
+import cats.syntax.split._
```
peterneyens: The `split` syntax isn't actually used below, and if it was used it, this import would clash with `cats.implicits._`.

If you want to test the split syntax you would have

```scala
import cats.syntax.split._ // afterwards => import cats.syntax.arrow._
import cats.instances.function._

val f = toLong split toDouble
```

gabro: ah, good catch, thanks. That's a half-baked attempt that I accidentally left in there.
I just pushed a proper version of that fix.

By the way, it'd be nice if scalafix tests were run by the CI (so that these kind of mistake would be caught). Do you think it's possible to add them to the build?

kailuowang: I think we should add them to the build. We can tackle it in another PR.

PR #1814, opened 2017-08-10
remove outdated TODOs

kailuowang:
related type classes are removed.


kailuowang:
Merged with one signoff as simple dead to-do comment removal

Issue #1713, opened 2017-05-30
Transform typeclass


leandrob13:
There has been some discussion about adding a typeclass for dealing with `F ~> G` transformations in #1644.  The original proposal was made by @kailuowang in this [gist](https://gist.github.com/kailuowang/ddf4f99a3bd97f150277a0c614e0a6f2) and I materialized it by the name of `Transform`. The example below includes only instances for EitherT and Kleisli.

```scala
import cats.arrow.FunctionK
import cats.data.{EitherT, Kleisli}
import cats.~>

trait Transform[H[_[_], _]] {
  def map[F[_], G[_], A](ha: H[F, A])(f: F ~> G): H[G, A]

  def lift[F[_], G[_]](f: F ~> G): H[F, ?] ~> H[G, ?] =
    new FunctionK[H[F, ?], H[G, ?]] {
      def apply[A](fa: H[F, A]): H[G, A] = map(fa)(f)
    }
}

object Transform {

  implicit def eitherTTransform[E]: Transform[EitherT[?[_], E, ?]] =
    new Transform[EitherT[?[_], E, ?]] {
      def map[F[_], G[_], A](ha: EitherT[F, E, A])(f: F ~> G): EitherT[G, E, A] =
        EitherT[G, E, A](f(ha.value)) //This could be added to EitherT as the F ~> G operation
    }

  implicit def kleisliTransform[E]: Transform[Kleisli[?[_], E, ?]] =
    new Transform[Kleisli[?[_], E, ?]] {
      def map[F[_], G[_], A](ha: Kleisli[F, E, A])(f: F ~> G): Kleisli[G, E, A] =
        ha.transform(f)
    }
}
```
The pattern is apparent in Kleisli since it already implements transform. For the rest of the monad transformers it is not defined as a `F ~> G` and it is confusing given that for `EitherT` and `OptionT` transform mean to bimap/map the inner types. This might be an opportunity to make standard the `F ~> G` operations in MonadTs and solve the confusion for the actual use of `transform`.

edmundnoble:
I can empathize with this requirement, and I would approve a PR that renames the methods to only be `transform` for higher-kinded mapping. However, cats-mtl has something very close to this in the form of `cats.mtl.monad.TransFunctor` and `cats.mtl.monad.LayerFunctor`.

wedens:
This typeclass is better known as `Hoist`.

leandrob13:
[...]

PR #1815, opened 2017-08-10
added `TFunctor` 

### LINKED ISSUE #1812: TFunctor for monad transformers
As a continuation of #1713 and #1492. I am working on a PR

```scala
/**
 * This is an endofunctor in the category of endofunctors in `Skal`.
 *
 * `Skal` is the category of scala types. Functors in `Skal`
 * is encoded as `Functor`. The functors in `Skal` themselves forms
 * a category, let's denote it as `F[Skal]`.
 * In `F[Skal]`, functors of Skal, e.g. `Option[_]` and `Either[E, _]`, are objects,
 * while natural transformations, e.g. `Option ~> Either[E, ?]`, are arrows.
 * A endofunctor in `F[Skal]` maps one set of functors of `Skal` to another
 * set of functors of `Skal` while preserving the structures.
 *
 * For `TFunctor`, the domain is `F[_]`, the codomain is `H[F, _]`, both are
 * functors in `Skal`. The `TFunctor` provides a mapping from the arrows between
 * `F[_]` and `G[_]`, i.e. `F ~> G` to arrows between `H[F, _]` and `H[G, _]`,
 * i.e. `H[F, ?] ~> H[G, ?]`. The `lift` method makes this intention clear.
 *
 * In `cats.core`, examples of such `TFunctor`s are monad transformers such
 * as `OptionT`, `EitherT`
 *
 */
@typeclass trait TFunctor[H[_[_], _]] {
  def map[F[_], G[_], A](h: H[F, A])(f: F ~> G): H[G, A]

  def lift[F[_], G[_]](f: F ~> G): H[F, ?] ~> H[G, ?] =
    λ[H[F, ?] ~> H[G, ?]](hf => map(hf)(f))
}
```

kailuowang:
added instances for 

- `WriterT`
-  `Kleisli`
-  `OptionT`
-  `EitherT`
-  `IdT`
-  `EitherK`
-  `Tuple2K`
-  `Free`
-  `FreeT`
-  `Func`
-  `Nested`
-  `OneAnd`

Fixes #1812, #1713 and #1492. also sort of a continuation of #1620
Any instances I miss? 
~~Question, do we need an instance for `Free`, not sure if there is a use case in the wild.~~

Update: Added instance for `Free`, `EitherK`, `TupleK`, `Func`, `Nested` and `OneAnd`

### DIFF
--- a/core/src/main/scala/cats/TFunctor.scala
+++ b/core/src/main/scala/cats/TFunctor.scala
@@ -1,0 +1,31 @@
+package cats
+
+import simulacrum.typeclass
+
+/**
+ * This is an endofunctor in the category of endofunctors in `Skal`.
+ *
+ * `Skal` is the category of scala types. Functors in `Skal`
+ * is encoded as `Functor`. The functors in `Skal` themselves forms
+ * a category, let's denote it as `F[Skal]`.
+ * In `F[Skal]`, functors of Skal, e.g. `Option[_]` and `Either[E, _]`, are objects,
+ * while natural transformations, e.g. `Option ~> Either[E, ?]`, are arrows.
+ * A endofunctor in `F[Skal]` maps one set of functors of `Skal` to another
+ * set of functors of `Skal` while preserving the structures.
+ *
+ * For `TFunctor`, the domain is `F[_]`, the codomain is `T[F, _]`, both are
+ * functors in `Skal`. The `TFunctor` provides a mapping from the arrows between
+ * `F[_]` and `G[_]`, i.e. `F ~> G` to arrows between `T[F, _]` and `T[G, _]`,
+ * i.e. `T[F, ?] ~> T[G, ?]`. The `lift` method makes this intention clear.
+ *
+ * In `cats.core`, examples of such `TFunctor`s are monad transformers such
+ * as `OptionT`, `EitherT`
+ *
+ */
+@typeclass trait TFunctor[T[_[_], _]] {
+  def mapNT[F[_], G[_], A](h: T[F, A])(f: F ~> G): T[G, A]
+
+  def liftNT[F[_], G[_]](f: F ~> G): T[F, ?] ~> T[G, ?] =
+    λ[T[F, ?] ~> T[G, ?]](hf => mapNT(hf)(f))
+}
+
--- a/core/src/main/scala/cats/TFunctor.scala
+++ b/core/src/main/scala/cats/TFunctor.scala
@@ -25,6 +25,17 @@ import simulacrum.typeclass
 @typeclass trait TFunctor[T[_[_], _]] {
   def mapNT[F[_], G[_], A](h: T[F, A])(f: F ~> G): T[G, A]
 
+  /**
+   * Lift a `F ~> G` to a `T[F, ?] ~> T[G, ?]`.
+   *
+   * {{{
+   * scala> import cats.implicits._, cats.data.OptionT
+   * scala> val lv: List ~> Vector = λ[List ~> Vector](_.toVector)
+   * scala> val olv: OptionT[List, ?] ~> OptionT[Vector, ?] = TFunctor[OptionT].liftNT(lv)
+   * scala> olv(OptionT.liftF(List(1)))
+   * res0: OptionT[Vector, Int] = OptionT(Vector(Some(1)))
+   * }}}
+   */
   def liftNT[F[_], G[_]](f: F ~> G): T[F, ?] ~> T[G, ?] =
     λ[T[F, ?] ~> T[G, ?]](hf => mapNT(hf)(f))
 }
--- a/core/src/main/scala/cats/data/Func.scala
+++ b/core/src/main/scala/cats/data/Func.scala
@@ -31,6 +31,11 @@ object Func extends FuncInstances {
 }
 
 private[data] abstract class FuncInstances extends FuncInstances0 {
+
+  implicit def catsDataTFunctorFoFunc[A]: TFunctor[Func[?[_], A, ?]] = new TFunctor[Func[?[_], A, ?]] {
+    def mapNT[F[_], G[_], B](h: Func[F, A, B])(f: F ~> G): Func[G, A, B] = Func.func(h.run.andThen(f.apply))
+  }
+
   implicit def catsDataApplicativeForFunc[F[_], C](implicit FF: Applicative[F]): Applicative[λ[α => Func[F, C, α]]] =
     new FuncApplicative[F, C] {
       def F: Applicative[F] = FF
--- a/core/src/main/scala/cats/data/Func.scala
+++ b/core/src/main/scala/cats/data/Func.scala
@@ -32,7 +32,7 @@ object Func extends FuncInstances {
 
 private[data] abstract class FuncInstances extends FuncInstances0 {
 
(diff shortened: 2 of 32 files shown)


iravid:
Not sure about the laws but this could work for StateT as well, no?

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

kailuowang:
@iravid `StateT`'s inner data has a nested `F[_]`, i.e. `F[S => F[(S, A)]]`, I don't know how to transform that to a `G[S => G[(S, A)]]`using just `F~>G`, any ideas?

iravid:
Ah, you're right - I missed the part where `StateT.transformF` would need to impose more constraints on `T`. Oh well :-)

core/src/main/scala/cats/data/Func.scala
```diff
@@ -31,6 +31,11 @@ object Func extends FuncInstances {
 }
 
 private[data] abstract class FuncInstances extends FuncInstances0 {
+
+  implicit def catsDataTFunctorFoFunc[A]: TFunctor[Func[?[_], A, ?]] = new TFunctor[Func[?[_], A, ?]] {
```
peterneyens: Missing "r" in "foFunc", similarly in "foNested".

edmundnoble:
I don't think we need to provide this type class to have a uniform API. We can just have a uniform API. Type classes are abstractions; unless there's a useful way to abstract over this, I think we should not include it. This is the standard for all other type classes in cats if I'm not mistaken.

kailuowang:
@edmundnoble
The law tests provided by the type class is more thorough than the existing test coverage (e.g. `FreeT ` and `Kleisili`) - also plus many of the instances I added doesn't have a `mapNT` equivalent yet, not to mention the tests, so I would need to duplicate the same tests code for 12 instances.

I think the type class might be able to provide reusable code other than `liftNT` . I am thinking `compose`. I will try implement that and maybe try find more after vacation.

core/src/main/scala/cats/TFunctor.scala
```diff
@@ -0,0 +1,42 @@
...
+ * For `TFunctor`, the domain is `F[_]`, the codomain is `T[F, _]`, both are
+ * functors in `Skal`. The `TFunctor` provides a mapping from the arrows between
+ * `F[_]` and `G[_]`, i.e. `F ~> G` to arrows between `T[F, _]` and `T[G, _]`,
+ * i.e. `T[F, ?] ~> T[G, ?]`. The `lift` method makes this intention clear.
+ *
+ * In `cats.core`, examples of such `TFunctor`s are monad transformers such
+ * as `OptionT`, `EitherT`
+ *
+ */
+@typeclass trait TFunctor[T[_[_], _]] {
```
johnynek: do we need `T[_[_], _]` or would `T[_[_]]` do? I'm not sure what the second parameter is buying us. Maybe I'm missing it. Can't we just ignore the inner value type? Also, wouldn't we have more flexibility only acting on `T[_[_]]`?

edmundnoble: Agreed here. If we had `T[_[_]]` I could come up with a lot more usecases. Maybe call it `FunctorK` then. We'd have to wrap the transformers in `Forall` though, which may mean our unified transform syntax fails to be found implicitly.

johnynek:
+1 for FunctorK. Seems more consistent with cats naming styles.

core/src/main/scala/cats/TFunctor.scala
kailuowang: The main reason I chose `T[_[_], _]` here, is that given any `Functor` `F[_]`,  `T[F, _]` is also a `Functor`. So `T[_[_], _]` can be deemed as a *functor* between the `F[_]` `Functor` and the `T[F, _]` `Functor`. i.e. a *endofunctor* in the category of *endofunctor*s, thus more like a higher kinded `Functor`. 
If you use `T[_[_]]`, then given any `Functor` `F[_]`, `T[F]` isn't a `Functor`, it's a kind * type, not a type constructor. So `T[_[_]]` is a *functor* between `F[_]` `Functor` and kind * type `T[F]`s. That makes it less a high kinded `Functor`. However, you can work around with the parametricity in instance definitions, e.g. for `OptionT`, 
```scala
def functorKInstance[A]: FunctorK[OptionT[?[_], A]]
```
With parametric `A`, we can have a `FunctorK` instance forall `A`s.  In some sense, we may choose to see this set of `FunctorK` instances as a single truely higher kinded *functor*.

Practical use wise, I can't think of any downside of using `T[_[_]]`. The only downside is its awkward encoding of the higher kinded `Functor` - it relies on instance definition. It may have more usecases, one that immediately jumps out is `FunctionK`. So maybe practically it's more useful. 

Now, the fact that we aren't sure about this thing makes me think that maybe we shouldn't place this in `cats-core`. I already have `FunctorK[T[_]]` defined in mainecoon, I could add all these instances there. Or we can create a new `cats-extra` module for this type of less commonly used type class.

sellout: With the example of `FunctionK`, my first thought is that that’s what’s defined incorrectly. We really want `FunctionK[F[_], G[_], A]` and `Forall[FunctionK[F, G, ?]]` for a natural transformation. Then you _do_ have a `FunctorK[FunctionK[F, ?[_], ?]]`. But that might just be crazy talk.

edmundnoble: I'm not sure if "higher-kinded X" is a really rigid term. In my mind, higher-kinded functor could equally well mean a functor from (endofunctors in Scal) to Scal, or a functor from endofunctors in Scal to other endofunctors in Scal. This is my main problem with the `-K` naming scheme; it gets even worse when type classes have multiple type parameters.

As well, the encoding proposed above does not reify that the `A` can vary with the instance remaining valid. The consumer of the instance needs a `Forall[Lambda[A =>FunctorK[OptionT[?[_], A]]]]` to have that fact in hand, and that makes it much more unwieldy, though a type alias could help.

To my mind if we're going to do a cats-extras project we'll want this there.

LukaJCB: I agree with @kailuowang and @edmundnoble, I don't think it's all that useful in cats-core. I'd like to see conformity in Monad Transformers for defining a `mapK` method, but it doesn't need a type class in core IMO.

sellout: > I'm not sure if "higher-kinded X" is a really rigid term.

Absolutely. I don’t think it helps at all. But if `Functor` is really “endofunctor in the category of Scal”, I’d expect `FunctorK` to mean the same thing as `K` in other instances, which basically adds “endofunctors” to the description – i.e., “endofunctor in the category of endofunctors in Scal”.

I’d also be happy with `Functor` being renamed to `Endofunctor` and having `EndofunctorK` and something like `LowerFunctor` (for a functor from the category of endofunctors to the category of Skal).

Also agreed that none of this should be in cats-core. I would like to find some time to work with `-Ykind-polymorphism` and to define these things as specializations of kind-polymorphic endofunctors, etc.

This sort of thing has been started in a few different libraries already – khats (which is too homophonically-named to be useful), Griffins (which never really got off the ground and mostly lives inside a Matryoshka branch), and I think one other that I have forgotten 😕

kailuowang:
Closing this for now since 1.0 already released

Issue #1703, opened 2017-05-25
SOE in Eval.defer


mpilquist:
We encountered a stack overflow when using `Eval.defer` in FS2 with cats 0.9:

```
[info] fs2.SegmentSpec *** ABORTED *** (5 seconds, 363 milliseconds)
[info]   java.lang.StackOverflowError:
[info]   at fs2.Segment.$anonfun$stage$1(Segment.scala:16)
[info]   at fs2.Segment.$anonfun$stage$1$adapted(Segment.scala:15)
[info]   at fs2.Segment$$anon$6.$anonfun$stage0$33(Segment.scala:233)
[info]   at cats.Eval$Call$$anon$10.$anonfun$start$3(Eval.scala:240)
[info]   at cats.Eval$Compute.loop$1(Eval.scala:279)
[info]   at cats.Eval$Compute.value(Eval.scala:293)
[info]   at cats.Eval$Call.value(Eval.scala:229)
[info]   at cats.Eval$Compute.loop$1(Eval.scala:289)
[info]   at cats.Eval$Compute.value(Eval.scala:293)
[info]   at cats.Eval$Call.value(Eval.scala:229)
[info]   at cats.Eval$Compute.loop$1(Eval.scala:289)
[info]   at cats.Eval$Compute.value(Eval.scala:293)
[info]   at cats.Eval$Call.value(Eval.scala:229)
[info]   at cats.Eval$Compute.loop$1(Eval.scala:289)
[info]   at cats.Eval$Compute.value(Eval.scala:293)
[info]   at cats.Eval$Call.value(Eval.scala:229)
[info]   at cats.Eval$Compute.loop$1(Eval.scala:289)
[info]   at cats.Eval$Compute.value(Eval.scala:293)
[info]   at cats.Eval$Call.value(Eval.scala:229)
[info]   at cats.Eval$Compute.loop$1(Eval.scala:289)
[info]   at cats.Eval$Compute.value(Eval.scala:293)
[info]   at cats.Eval$Call.value(Eval.scala:229)
[info]   at cats.Eval$Compute.loop$1(Eval.scala:289)
[info]   at cats.Eval$Compute.value(Eval.scala:293)
[info]   at cats.Eval$Call.value(Eval.scala:229)
[info]   at cats.Eval$Compute.loop$1(Eval.scala:289)
[info]   at cats.Eval$Compute.value(Eval.scala:293)
[info]   at cats.Eval$Call.value(Eval.scala:229)
[...]

PR #1816, opened 2017-08-11
Add Eval documentation

LukaJCB:
Adds some eval documentation with very basic examples, not sure if there are better examples around that can better showcase the usefulness of `Eval`. Either way, looking forward to feedback :)

### DIFF
--- a/docs/src/main/tut/datatypes/eval.md
+++ b/docs/src/main/tut/datatypes/eval.md
@@ -1,0 +1,108 @@
+---
+layout: docs
+title:  "Eval"
+section: "data"
+source: "core/src/main/scala/cats/Eval.scala"
+scaladoc: "#cats.Eval"
+---
+# Eval
+
+Eval is a data type for controlling synchronous evaluation.
+Its implementation is designed to provide stack-safety at all times using a technique called trampolining.
+
+There are two different factors that play into evaluation: memoization and laziness.
+
+Memoized evaluation evaluates an expression only once and then remembers (memoizes) that value.
+Lazy evaluation refers to when the expression is evaluated.
+We talk about eager evaluation if the expression is immediately evaluated when defined and about lazy evaluation if the expression is evaluated when it's first used.
+
+For example, in Scala, a `lazy val` is both lazy and memoized, a method definition `def` is lazy, but not memoized, since the body will be evaluated on every call.
+A normal `val` evaluates eagerly and also memoizes the result.
+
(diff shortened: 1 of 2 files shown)


johnynek:
This does not seem to mention that some uses of nested `.value` calls can be unsafe and kill stack safety.

Ideally, you never call `.value` inside something that could be itself inside an Eval. I don't fully grok, myself rules for when it might be safe to call `.value` inside another Eval, but @non seems to have some rules of thumb. I think without this discussion we could be setting someone up for confusion stack overflows.

johnynek:
note for instance #1703 which I think may be currently broken still.

johnynek:
Can we add an example of using `Eval.defer` to do stack safe recursion? That's one of the main motivators of Eval (along with giving a type to an optionally lazy value).

another note: it would be really nice to have some, even non-rigorous, benchmarks on Eval. In fact, Eval is MUCH slower than using standard function calls. So if you are doing it to get stack safe recursion, if you know the recursion depth will be shallow, you may be much better off with normal on-stack recursion.

cc @non

docs/src/main/tut/datatypes/eval.md
```diff
@@ -0,0 +1,108 @@
...
+We can run the computation using the given evaluation strategy anytime by using the `value` method.
+
+```tut:book
+eager.value
+
+```
+
+#### Eval.later
+
+If we want lazy evaluation, we can use `Eval.later`:
```
kailuowang: Might be worthy mentioning here that there are two advantages of `Later` over `lazy val` 
1. `Later` allows the thunk being GC'd after evaluation and
2. `Later`'s synchronization lock locks itself while lazy vals lock their enclosing class.

kailuowang:
I agree with @ceedubs that having this is much better than having none, @LukaJCB what do you think? Do you think those points are super quick to address or do you want to iterate fast? 
BTW, there is an awesome talk by @non on this topic. https://www.youtube.com/watch?v=XDDy-QD0aVU maybe worth mentioning

LukaJCB:
I did address @kailuowangs points and I also added a small note on `Eval.defer` :) 
I was going to wait for @non to chime in here, about stack-safe recursion, since I'm not super knowledgable on that, but I'm fine with merging now, and making a new PR with `Eval.defer`.

ceedubs:
Merging with 2 approvals. I do think that @kailuowang's suggestion to link to Erik's talk on `Eval` would be a nice addition in a future PR.

PR #1817, opened 2017-08-11
Add seals to related projects

durban:
@kailuowang As discussed in https://github.com/durban/seals/issues/62.


kailuowang:
simple doc addition, merge with 1 sign off

PR #1819, opened 2017-08-11
Added more implementations of map2Eval

johnynek:
I hit a case where I wanted laziness with map2Eval and realized in many
cases where we could have it, we do not. There were all the cases I saw
that we missed, which I hope our existing tests cover.


johnynek:
@kailuowang @edmundnoble can you gentlemen review this? I think it should be ready.

PR #1820, opened 2017-08-16
Fix links

n4to4:
These seem to be redirected


peterneyens:
Thanks @n4to4.

Would you mind changing those in _docs/src/main/tut/index.md_ as well?

n4to4:
done @peterneyens

PR #1822, opened 2017-08-17
Add docs for Ior

LukaJCB:
Some basic usage so far. Might make sense to also document more of the methods available on `Ior`:)

### DIFF
--- a/docs/src/main/tut/datatypes/ior.md
+++ b/docs/src/main/tut/datatypes/ior.md
@@ -60,7 +60,7 @@ case class User(name: Username, pw: Password)
 
 def validateUsername(u: String): Failures Ior Username = {
   if (u.isEmpty)
-    Nel.one("Can't be empty").leftIor
+    Ior.leftNel("Can't be empty")
   else if (u.contains("."))
     Ior.both(Nel.one("Dot in name is deprecated"), Username(u))
   else
@@ -69,7 +69,7 @@ def validateUsername(u: String): Failures Ior Username = {
 
 def validatePassword(p: String): Failures Ior Password = {
   if (p.length < 8)
-    Nel.one("Password too short").leftIor
+    Ior.leftNel("Password too short")
   else if (p.length < 10)
     Ior.both(Nel.one("Password should be longer"), Password(p))
   else
--- a/docs/src/main/tut/datatypes/ior.md
+++ b/docs/src/main/tut/datatypes/ior.md
@@ -21,7 +21,7 @@ def map[B, C](fa: A Ior B)(f: B => C): A Ior C
 We can create `Ior` values using `Ior.left`, `Ior.right` and `Ior.both`:
 
 ```tut
-import cats.data.Ior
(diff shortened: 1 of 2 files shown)


docs/src/main/tut/datatypes/ior.md
```diff
@@ -0,0 +1,114 @@
...
+layout: docs
+title:  "Ior"
+section: "data"
+source: "core/src/main/scala/cats/data/Ior.scala"
+scaladoc: "#cats.data.Ior"
+---
+# Ior
+
+`Ior` represents an inclusive-or relationship between two data types.
+This makes it very similar to the [`Either`](either.html) data type, which represents an `Xor` relationship.
```
ceedubs: I'm a bit hesitant about this `Xor` reference. I think that this might be confusing, as Cats no longer has an `Xor` type. For consistency with the previous sentence, it might be better to write this as "which represents an exclusive-or" relationship.

In logic/circuits, this is often written as `XOR`, which we could potentially use, but I think that exclusive-or is more straightforward.

peterneyens:
Something which would be nice as well is to show the difference between the `Applicative` behaviour and the `Semigroup` behaviour, potentially in comparison with `Either` and `Validated`.

Something like : 

```scala
import cats.data.{ IorNel, NonEmptyList, ValidatedNel }
import cats.implicits._

val oddOrEvenEither: Int => Either[Int, Int] = 
  n => Either.cond(n % 2 == 0, n, n)

val oddOrEvenValidatedNel: Int => ValidatedNel[Int, Int] =
  oddOrEven(_).toValidated.toValidatedNel

val oddOrEvenIorNel: Int => IorNel[Int, Int] =
  oddOrEven(_).toIor.leftMap(NonEmptyList.one)

type OddEither[A] = Either[Int, A]
type OddNelVal[A] = ValidatedNel[Int, A]
type OddNelIor[A] = IorNel[Int, A]

val nel = NonEmptyList.of(1, 2, 3, 4, 5)

// Ior short circuits like Either

nel.traverse[OddEither, Int](oddOrEvenEither)       // Left(1)
nel.traverse[OddNelVal, Int](oddOrEvenValidatedNel) // Invalid(NonEmptyList(1, List(3, 5)))
nel.traverse[OddNelIor, Int](oddOrEvenIorNel)       // Left(NonEmptyList(1, List()))

// for Either and Validated, Semigroup same result as with applicative
// but for Ior gives both sides

nel.reduceMap(oddOrEvenEither)        // Left(1)
nel.reduceMap(oddOrEvenValidatedNel)  // Invalid(NonEmptyList(1, List(3, 5)))
nel.reduceMap(oddOrEvenIorNel)        // Both(NonEmptyList(1, List(3, 5)), 6)
```

I can add this in a follow up PR if needed.

ceedubs:
LGTM. @peterneyens are there any pending items you wanted to see make it into this before merging?

peterneyens:
No, I'll follow up with another PR.

PR #1836, opened 2017-08-21
Add fromFuture and fromRecoverable to EitherT companion object

leandrob13:
This PR adds the capability of creating EitherT's instances with a default recovery for errors in `Future`, also adds `fromRecoverable` that allows create an instance with an error recovery in `F[_]` given a `ApplicativeError`.

### DIFF
--- a/core/src/main/scala/cats/data/EitherT.scala
+++ b/core/src/main/scala/cats/data/EitherT.scala
@@ -5,6 +5,8 @@ import cats.functor.Bifunctor
 import cats.instances.either._
 import cats.syntax.either._
 
+import scala.concurrent.{ExecutionContext, Future}
+
 /**
  * Transformer for `Either`, allowing the effect of an arbitrary type constructor `F` to be combined with the
  * fail-fast effect of `Either`.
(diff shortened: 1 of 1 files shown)


peterneyens:
Isn't this `fromFuture` similar to `ApplicativeError[Future, Throwable].attemptT`?

leandrob13:
@peterneyens it seems so. Should I define it from an `implicit ae: ApplicativeError[Future, Throwable]`?

peterneyens:
I don't think it makes sense to add `EitherT.fromFuture(fut)` if you could do `fut.attemptT` as well (you probably need `-Ypartial-unification` for the `ApplicativeError` syntax to work).

johnynek:
Future does have a MonadError, which implies ApplicativeError.

core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -335,6 +337,50 @@ object EitherT extends EitherTInstances {
...
+    * scala> EitherT.fromRecoverable[Option, Unit](o2) { case _ => 1 }
+    * res1: EitherT[Option, Nothing, Int] = EitherT(Some(Right(1)))
+    * }}}
+    */
+  final def fromRecoverable[F[_], E]: FromRecoverablePartiallyApplied[F, E] = new FromRecoverablePartiallyApplied
+
+  /**
+    * Uses the [[http://typelevel.org/cats/guidelines.html#partially-applied-type-params Partially Applied Type Params technique]] for ergonomics.
+    */
+  private[data] final class FromRecoverablePartiallyApplied[F[_], E](val dummy: Boolean = true) extends AnyVal {
```
johnynek: does `val unit: Unit = ()` not work here? The Boolean bothers me slightly.

core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -335,6 +337,50 @@ object EitherT extends EitherTInstances {
...
+    def apply[A, B](f: F[B])(pf: PartialFunction[E, B])(implicit me: MonadError[F, E]): EitherT[F, A, B] =
```
johnynek: can we change this to `ApplicativeError[F, E]`

core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -335,6 +337,50 @@ object EitherT extends EitherTInstances {
...
+    * res0: Either[Throwable, Int] = Right(1)
+    * scala> val f2: Future[Int] = Future.failed(new Throwable("error"))
+    * scala> Await.result(EitherT.fromFuture(f2).value, Duration.Inf)
+    * res1: Either[Throwable, Int] = Left(java.lang.Throwable: error)
+    * }}}
+    */
+  final def fromFuture[B](future: Future[B])(implicit ec: ExecutionContext): EitherT[Future, Throwable, B] =
+    EitherT {
+      future.map(Right(_)).recoverWith {
+        case ex => Future.successful(Left(ex))
```
adelbertc: Do we want to catch everything or just `NonFatal(_)`? or does the Future API only give you `NonFatal` anyways. If so maybe a comment clarifying.

core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -335,6 +337,50 @@ object EitherT extends EitherTInstances {
...
+
+  /**
+    * Uses the [[http://typelevel.org/cats/guidelines.html#partially-applied-type-params Partially Applied Type Params technique]] for ergonomics.
+    */
+  private[data] final class FromRecoverablePartiallyApplied[F[_], E](val dummy: Boolean = true) extends AnyVal {
+    def apply[A, B](f: F[B])(pf: PartialFunction[E, B])(implicit me: MonadError[F, E]): EitherT[F, A, B] =
+      EitherT.right[A](me.recover(f)(pf))
+  }
+
+  /** Creates an `EitherT` from a Future, recovers its error into a Either.Left.
```
adelbertc: Can we add a comment saying this will result in a Future that always succeeds but contains the potential error (and value) in the success. This has been a source of confusion with some of my colleagues.

core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -335,6 +337,50 @@ object EitherT extends EitherTInstances {
...
+    * scala> import scala.concurrent.ExecutionContext.Implicits.global
+    * scala> val f1: Future[Int] = Future.successful(1)
+    * scala> Await.result(EitherT.fromFuture(f1).value, Duration.Inf)
+    * res0: Either[Throwable, Int] = Right(1)
+    * scala> val f2: Future[Int] = Future.failed(new Throwable("error"))
+    * scala> Await.result(EitherT.fromFuture(f2).value, Duration.Inf)
+    * res1: Either[Throwable, Int] = Left(java.lang.Throwable: error)
+    * }}}
+    */
+  final def fromFuture[B](future: Future[B])(implicit ec: ExecutionContext): EitherT[Future, Throwable, B] =
```
adelbertc: Can this just delegate to `fromRecoverable`?

core/src/main/scala/cats/data/EitherT.scala
fthomas: @johnynek That does not work here because of https://github.com/scala/bug/issues/9240

core/src/main/scala/cats/data/EitherT.scala
leandrob13: @johnynek sure thing!

core/src/main/scala/cats/data/EitherT.scala
leandrob13: @adelbertc will do!

core/src/main/scala/cats/data/EitherT.scala
leandrob13: @adelbertc yes, it is just `NonFatal`, according to the Promise implementation of `transform` which is used by `recover`:

```scala
override def transform[S](f: Try[T] => Try[S])(implicit executor: ExecutionContext): Future[S] = {
    val p = new DefaultPromise[S]()
    onComplete { result => p.complete(try f(result) catch { case NonFatal(t) => Failure(t) }) }
    p.future
  }
```

core/src/main/scala/cats/data/EitherT.scala
leandrob13: @adelbertc what do you mean by delegate?

core/src/main/scala/cats/data/EitherT.scala
leandrob13: @johnynek btw, I thought `ApplicativeError` was going away. That is why I chose `MonadError`.

core/src/main/scala/cats/data/EitherT.scala
LukaJCB: I think he means implementing in terms of `fromRecoverable`, i.e. something along the lines of
```scala
def fromFuture[B](future: Future[B])(implicit ec: ExecutionContext): EitherT[Future, Throwable, B] =
  EitherT.fromRecoverable[Future, Throwable](future)
```

larsrh:
I'm closing this stale PR for many years of inactivity. Sorry about that! If there's any renewed interest, please open a fresh one.

PR #1837, opened 2017-08-21
Add Parallel type class

### LINKED ISSUE #1830: Add a Parallel type class
In [PureScript](https://pursuit.purescript.org/packages/purescript-parallel/3.0.0/docs/Control.Parallel.Class) there is a `Parallel` type class, which introduces the idea of a Monad that supports parallel composition with a related Applicative instance.  This could work for `IO` in cats-effect and a theoretical `ParIO`, but also for `Either` and `Validated`. 
We could also include instances for Kleisli (ReaderT) and WriterT.  

I think it could really be useful, especially with regards to `parTraverse`.

LukaJCB:
Should resolve #1830. Only Syntax and type class for now, but more to follow :)


core/src/main/scala/cats/syntax/parallel.scala
```diff
@@ -0,0 +1,32 @@
...
+  def self : T[A]
+
+  implicit val T = typeClassInstance
+
+  def parTraverse[M[_]: Monad, F[_], B]
+  (f: A => M[B])(implicit P: Parallel[M, F]): M[T[B]] =
+    Parallel.parTraverse(self)(f)
+
+  def parSequence[M[_]: Monad, F[_], B](implicit ev: A <:< M[B], P: Parallel[M, F]): M[T[B]] =
+    Parallel.parSequence(self.asInstanceOf[T[M[B]]])
```
LukaJCB: Anyone know a better solution? Feels a little bit dirty.

LukaJCB:
Summary of what this entails:

Instances for:
- Semigroup[E] => Either[E, ?] <-> Validated[E,?] 
- Parallel[M, F] => OptionT[M, ?] <-> Nested[F, Option, ?]
- Parallel[M, F] => EitherT[M, E, ?] <-> Nested[F, Validated[E, ?]
- Parallel[M, F] => WriterT[M, A, ?] <-> WriterT[F, A, ?]
- Parallel[M, F] => Kleisli[M, A, ?] <-> Kleisli[F, A, ?]

Syntax for:
- `parTraverse`
- `parSequence`
- Tuple syntax `parMapN`

LukaJCB:
### Summary of what this entails:

#### Instances for:

- Semigroup[E] => Either[E, ?] <-> Validated[E,?]
- Parallel[M, F] => OptionT[M, ?] <-> Nested[F, Option, ?]
- Parallel[M, F] => EitherT[M, E, ?] <-> Nested[F, Validated[E, ?]
- Parallel[M, F] => WriterT[M, A, ?] <-> WriterT[F, A, ?]
- Parallel[M, F] => Kleisli[M, A, ?] <-> Kleisli[F, A, ?]

#### Syntax for:

parTraverse
parSequence
Tuple syntax parMapN

#### Other notes:

We could also add an instance for List with an Applicative instance that zips, there currently is a newtype `ZipList` in newts, so either it would have to be defined there, or we would have to pull it inside cats.

edmundnoble:
I will also transport cats-mtl type classes across parallel. 

In cats, you should also have (`Parallel[M,F], MonadError[M]`) => `ApplicativeError[F]`.

LukaJCB:
@edmundnoble I've added the instance for it, but now I'm getting failing Serializable Tests, any pointers there? 🤔

LukaJCB:
Another open question is about an instance for `Future`. Currently the standard `Applicative` instance already executes in parallel, which is certainly nice for a lot of people, but also inconsistent with it's `Monad` instance. I'm not sure how strict we are with inconsistent Monad/Applicative instances, but changing the current `Applicative` instance would be a massive breaking change, so I'm thinking we should not do it anytime soon and thus a `Parallel` instance doesn't really make all that much sense right now. Opinions? :)

core/src/main/scala/cats/Parallel.scala
```diff
@@ -0,0 +1,100 @@
...
+  * The Parallel type class allows us to represent this relationship.
+  */
+trait Parallel[M[_], F[_]] {
+  /**
+    * The applicative instance for F[_]
+    */
+  def applicative: Applicative[F]
+
+  /**
+    * Natural Transformation from the sequential Monad M[_] to the parallel Applicative F[_].
```
kailuowang: the other way? `from  the parallel Applicative F[_] to the sequential Monad M[_]`

core/src/main/scala/cats/Parallel.scala
```diff
@@ -0,0 +1,100 @@
...
+    */
+  def sequential(implicit M: Monad[M]): F ~> M
+
+  /**
+    * Natural Transformation from the parallel Applicative F[_] to the sequential Monad M[_].
```
kailuowang: also in reverse order

core/src/main/scala/cats/Parallel.scala
```diff
@@ -0,0 +1,100 @@
+package cats
+
+/**
+  * Some types that form a Monad, are also capable of forming an Applicative that supports parallel composition.
+  * The Parallel type class allows us to represent this relationship.
+  */
+trait Parallel[M[_], F[_]] {
```
kailuowang: you want `extends Serializable`.

LukaJCB:
Thank you @kailuowang! That fixed the failing tests. :)

kailuowang:
My opinion over `Future` is that we shouldn't try to provide more functionalities over it than we currently have. I think users who are already coding at this level of pure FP probably abandoned `Future` long ago, especially now that we have so many good options for replacement in the cats ecosystem.

LukaJCB:
Totally agree with you @kailuowang.

CI does not pass the Validated tut tests right now because of ambiguous implicits as I feared before. I 'm not sure why it doesn't fail in the same way in the WriterT and ValidatedTests, but right now importing from `cats.implicits._` results in ambiguous implicits for `ApplicativeError` for `Validated. 

We can delete the derived `ApplicativeError` instance, or somehow find a way to give the "real" instance higher priority. Although I don't really know how we could right now.

edmundnoble:
The Applicative instance for Future is actually not responsible for the evaluation strategy, seeing as all Futures are already evaluated. That is to say, even if you changed Future's Applicative operations to use `flatMap`, Applicative is not lazy in its first or second parameter, so it's not even possible to make a sequential Applicative. This is fine if you make the frequently-made assumption that if you're using cats with Future you aren't using side effects, because with Future that's exactly what they are, side effects (not suspendable into Future), and cats doesn't morally care about side effects like we don't care about null.

kailuowang:
@LukaJCB to avoid the ambiguous implicits, an option that I can think of is to not implicitly provide the `ApplicativeX` instances through `Parallel` (which I am not sure about the use cases are anyway), rather, make them available through explicit methods on `Parallel`. Users will need to explicitly use them if they need parallel execution.

LukaJCB:
That's an excellent idea, will do! :)

LukaJCB:
CI is happy again, I'd be fine with merging this without further functionality and deal with e.g. a List/ZipList instance afterwards :)

LukaJCB:
Now that #1885 is merged, I went ahead and added instances for NEL and NEV :)

LukaJCB:
I also added an instance `EitherT[M, E, ?] <-> Nested[M, Validated[E,?]]` which basically allows us to treat `EitherT` as a theoretical `ValidatedT` without requiring there to be a `Parallel` instance for `M`.

core/src/main/scala/cats/Parallel.scala
```diff
@@ -0,0 +1,139 @@
...
+trait Parallel[M[_], F[_]] extends Serializable {
+  /**
+    * The applicative instance for F[_]
+    */
+  def applicative: Applicative[F]
+
+  /**
+    * Natural Transformation from the parallel Applicative F[_] to the sequential Monad M[_].
+    */
+  def sequential(implicit M: Monad[M]): F ~> M
```
johnynek: Why do we need to accept this implicit? Why not make `def monad: Monad[M]` on Parallel?

core/src/main/scala/cats/Parallel.scala
```diff
@@ -0,0 +1,139 @@
...
+
+  /**
+    * Natural Transformation from the sequential Monad M[_] to the parallel Applicative F[_].
+    */
+  def parallel(implicit M: Monad[M]): M ~> F
```
johnynek: Same question about why we want M here and not as a method.

core/src/main/scala/cats/Parallel.scala
```diff
@@ -0,0 +1,139 @@
...
+    parAp(Monad[M].map(ma)(a => (b: B) => (a, b)))(mb)
+
+  /**
+    * Like `Applicative[F].ap2`, but uses the applicative instance
+    * corresponding to the Parallel instance instead.
+    */
+  def parAp2[M[_]: Monad, F[_], A, B, Z](ff: M[(A, B) => Z])
+                                        (ma: M[A], mb: M[B])
+                                        (implicit P: Parallel[M, F]): M[Z] =
+    Monad[M].map(parProduct(ma, parProduct(mb, ff))) { case (a, (b, f)) => f(a, b) }
```
johnynek: Seems weird to reimplement these rather than call P.applicative.ap2. Can't we deoptimize an existing Applicative by doing this?

core/src/main/scala/cats/Parallel.scala
LukaJCB: I initially only did this to make instances of `Parallel` shorter, but honestly, I really don't feel strongly about it either way.

core/src/main/scala/cats/Parallel.scala
```diff
@@ -0,0 +1,139 @@
...
+
+  /**
+    * Provides an `ApplicativeError[F, E]` instance for any F, that has a `Parallel[M, F]`
+    * and a `MonadError[M, E]` instance.
+    * I.e. if you have a type M[_], that supports parallel composition through type F[_],
+    * then you can get `ApplicativeError[F, E]` from `MonadError[M, E]`.
+    */
+  def applicativeError[M[_], F[_], E]
```
kailuowang: when I made the suggestion, I was thinking having this method on Parallel instance. I.e. 
```
trait Parallel[M[_], F[_]] extends Serializable {
    ....
   def parallelError[E](implicit ME: MonadError[M, E]):  ApplicativeError[F, E] = ...
}
```
which feels a natural addition to the existing API. Your API has benefits too. What do you think? Maybe we have both API?

core/src/main/scala/cats/Parallel.scala
```diff
@@ -0,0 +1,139 @@
...
+
+  /**
+    * Like `Traverse[A].sequence`, but uses the applicative instance
+    * corresponding to the Parallel instance instead.
+    */
+  def parSequence[T[_]: Traverse, M[_]: Monad, F[_], A]
+  (tma: T[M[A]])(implicit P: Parallel[M, F]): M[T[A]] = {
+    implicit val F = P.applicative
+    val fta: F[T[A]] = Traverse[T].traverse(tma)(P.parallel.apply)
+    P.sequential.apply(fta)
```
kailuowang: if we remove the `Monad` implicit arg (and make it a implicit def in the typeclass), we can do `P.sequential(fta)` here right?  That would be a better API.

LukaJCB:
So after a large refactor session, I did change the API, like @johnynek suggested :)

core/src/main/scala/cats/Parallel.scala
LukaJCB: Done!

core/src/main/scala/cats/Parallel.scala
```diff
@@ -97,9 +107,13 @@ object Parallel extends ParallelArityFunctions {
     */
   def parAp2[M[_]: Monad, F[_], A, B, Z](ff: M[(A, B) => Z])
                                         (ma: M[A], mb: M[B])
-                                        (implicit P: Parallel[M, F]): M[Z] = P.sequential.apply(
-    P.applicative.ap2(P.parallel.apply(ff))(P.parallel.apply(ma), P.parallel.apply(mb))
-  )
+                                        (implicit P: Parallel[M, F]): M[Z] = {
+    implicit val M = P.monad
+    P.sequential.apply(
```
edmundnoble: `P.sequential()`? Here and in other places, and with `P.parallel`.

johnynek: +1

core/src/main/scala/cats/Parallel.scala
```diff
@@ -0,0 +1,164 @@
...
+  /**
+    * Like `Applicative[F].ap`, but uses the applicative instance
+    * corresponding to the Parallel instance instead.
+    */
+  def parAp[M[_], F[_], A, B](mf: M[A => B])
+                                    (ma: M[A])
+                                    (implicit P: Parallel[M, F]): M[B] = {
+    implicit val F = P.applicative
+    implicit val M = P.monad
+    val fb = Applicative[F].ap(P.parallel(mf))(P.parallel(ma))
```
johnynek: what about `P.applicative.ap()` rather than setting up the implicit to use it once.

Also, I don't see you using `val M` do we need it?

core/src/main/scala/cats/Parallel.scala
```diff
@@ -0,0 +1,164 @@
...
+    P.sequential(fb)
+  }
+
+  /**
+    * Like `Applicative[F].product`, but uses the applicative instance
+    * corresponding to the Parallel instance instead.
+    */
+  def parProduct[M[_]: Monad, F[_], A, B](ma: M[A], mb: M[B])
+                                         (implicit P: Parallel[M, F]): M[(A, B)] =
+    parAp(Monad[M].map(ma)(a => (b: B) => (a, b)))(mb)
```
johnynek: what about `P.sequential(P.applicative.produce(P.parallel(ma), P.parallel(mb)))`

core/src/main/scala/cats/Parallel.scala
```diff
@@ -0,0 +1,164 @@
...
+
+  /**
+    * Like `Applicative[F].ap2`, but uses the applicative instance
+    * corresponding to the Parallel instance instead.
+    */
+  def parAp2[M[_]: Monad, F[_], A, B, Z](ff: M[(A, B) => Z])
+                                        (ma: M[A], mb: M[B])
+                                        (implicit P: Parallel[M, F]): M[Z] = {
+    implicit val M = P.monad
```
johnynek: where are we using this?

core/src/main/scala/cats/Parallel.scala
```diff
@@ -0,0 +1,164 @@
...
+  def parallel: M ~> F
+
+  /**
+    * Provides an `ApplicativeError[F, E]` instance for any F, that has a `Parallel[M, F]`
+    * and a `MonadError[M, E]` instance.
+    * I.e. if you have a type M[_], that supports parallel composition through type F[_],
+    * then you can get `ApplicativeError[F, E]` from `MonadError[M, E]`.
+    */
+  def applicativeError[E](implicit E: MonadError[M, E]): ApplicativeError[F, E] =
+    Parallel.applicativeError[M, F, E](this, E)
```
johnynek: can we reverse this?

If you make this call the `Parallel.appicativeError` then if I override this method if you call the object version we don't get the override.

Either it should be final here, or the object should call this method.

core/src/main/scala/cats/instances/future.scala
```diff
@@ -35,6 +35,11 @@ trait FutureInstances extends FutureInstances1 {
 private[instances] sealed trait FutureInstances1 extends FutureInstances2 {
   implicit def catsStdMonoidForFuture[A: Monoid](implicit ec: ExecutionContext): Monoid[Future[A]] =
     new FutureMonoid[A]
+
+  implicit def catsStdParallelForFuture[A](implicit ec: ExecutionContext): Parallel[Future, Future] = {
+    implicit val M: Monad[Future] = cats.instances.future.catsStdInstancesForFuture
+    Parallel.identity[Future]
```
johnynek: what is your thinking on this?

edmundnoble: I think this is about as valid as the `Monad[Future]`; you get parallel execution using the `Applicative` operators on `Future` anyway, because none of them are lazy, whereas `flatMap` has to be.

johnynek: If you zip two Futures, if I'm not mistaken, it does not complete until the left one completes even if it fails. If the right one fails first, you still have to wait for the left. I can imagine an applicative (with an appropriately loose notion of Exception equality) that fails as soon as either side fails.

LukaJCB: That's a very interesting idea! Sorry for not responding earlier, I'll give this a shot soon, seems pretty useful :)

LukaJCB: I created a newtype `FailFastFuture` that behaves in the way you specified, not sure if my implementation is the best, but couldn't think of another way to implement it more succinctly.

kailuowang:
Just curious do you have the codecov chrome extension?

LukaJCB:
I just downloaded it, that is very helpful! Thank you so much :)

core/src/main/scala/cats/Parallel.scala
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  /**
+    * Like `Applicative[F].ap`, but uses the applicative instance
+    * corresponding to the Parallel instance instead.
+    */
+  def parAp[M[_], F[_], A, B](mf: M[A => B])
+                                    (ma: M[A])
+                                    (implicit P: Parallel[M, F]): M[B] =
+    P.sequential(P.applicative.ap(P.parallel(mf))(P.parallel(ma)))
```
kailuowang: this method seems not tested.

LukaJCB: Fixed :)

kailuowang:
Given the amount of new features included, are we sure if it's a good idea to include this in 1.0 which is going to lock its API.

LukaJCB:
Feedback from gitter and the likes has been overwhelmingly positive, so I think it'd be a disappointment if we don't add this in 1.0. I'd rather rip out some not essential features then delay the whole PR :)

kailuowang:
@LukaJCB it's not the usefulness that I doubt, it's the maturity. It's the fact that this API is not battle tested and yet we are about to lock it for a long time.

 An alternative could be we create a new `cats.experiment` module and place this there alone with some of the other novel features. We might consider adding another `cats.extra` module where we can place in non-core but battle tested features. `cats.experiment`  as the name suggests, has no guarantee of backward compatibility going forward, features can graduate from here to either `core` or `cats.extra`. We would strongly recommend against libraries with downstream dependencies from using this module. 
`cats.extra` on the other hand can probably have the same back compat promise as `core`. 
WDYT?

alexandru:
Adding my 2 cents ...

I would be very pleased to see `Parallel` in `cats-core` because it solves an important problem that people have with `Task` and `IO` implementations.

People expect to do `map2` and `ap` in parallel and are then surprised that the default `Applicative` instance doesn't do that.

So you have to teach them to either use a type alias, with some of the available `newtype` hacks floating around, all of which are hacks that don't work well. This is something that Scalaz 7 is doing. Scalaz 7 also has a type class called `Nondeterminism`, that I think we can agree that it wasn't a good idea. Or you have to tell them to use an import for a different implicit in scope, overriding the default, which then lands you in Scala's implicit resolution limbo and what we have in Monix.

And then this extends to all abstractions that are doing parametric polymorphism. For example FS2 or the new `Iterant` data type coming in Monix 3.0. And thus there's no way to express for example a [zip operation](https://github.com/monix/monix/blob/v3.0.0-M1/monix-tail/shared/src/main/scala/monix/tail/Iterant.scala#L1485) that fetches tuples in parallel, without having the user to rely on a weird and error prone import.

As for the proposal to do this in a `cats.experiment` module, the real trouble is that the `Parallel` type class is middleware. And there's no way I can depend on `cats-experiment`, or a `cats-extra` for that matter. In Monix depending on `cats-core` and `cats-effect` is already above the threshold that some of the userbase can tolerate.

SystemFw:
Seems like @alexandru already wrote most of what I wanted to write. I also think `Parallel` is a great improvement, and important enough in scope that it should be in `core`. `Either` and `Validated` is another _very_ common use case that people stumble on

kailuowang:
Alright, look like the benefit of including it in core and lock the API outweigh the risk.  As discussed in [gitter](https://gitter.im/typelevel/cats-dev?at=59ce4d6197cedeb04829fa35), let's reduce the non-essential parts to minimize the risk.

LukaJCB:
Okay this PR includes the following instances:

- `Semigroup[E] => Either[E, ?] <-> Validated[E,?]`
- `(Semigroup[E], Monad[M]) => EitherT[M, E, ?] <-> Nested[M, Validated[E,?]]`
- `Parallel[M, F] => OptionT[M, ?] <-> Nested[F, Option, ?]`
- `Parallel[M, F] => WriterT[M, A, ?] <-> WriterT[F, A, ?]`
- `Parallel[M, F] => Kleisli[M, A, ?] <-> Kleisli[F, A, ?]`
- `(Parallel[M, F], Semigroup[E]) => EitherT[M, E, ?] <-> Nested[F, Validated[E, ?]`

The other instances will be included in another PR :)

core/src/main/scala/cats/Parallel.scala
```diff
@@ -0,0 +1,220 @@
...
+trait NonEmptyParallel[M[_], F[_]] extends Serializable {
+  /**
+    * The Apply instance for F[_]
+    */
+  def applicative: Apply[F]
+
+  /**
+    * The FlatMap instance for M[_]
+    */
+  def monad: FlatMap[M]
```
johnynek: could we call this `def flatMap: FlatMap[M]` or something? I don't like returning `FlatMap` with a `monad`.

core/src/main/scala/cats/Parallel.scala
```diff
@@ -0,0 +1,220 @@
...
+
+/**
+  * Some types that form a FlatMap, are also capable of forming an Apply that supports parallel composition.
+  * The NonEmptyParallel type class allows us to represent this relationship.
+  */
+trait NonEmptyParallel[M[_], F[_]] extends Serializable {
+  /**
+    * The Apply instance for F[_]
+    */
+  def applicative: Apply[F]
```
johnynek: can we call this `def apply` or something?

core/src/main/scala/cats/Parallel.scala
```diff
@@ -0,0 +1,220 @@
...
+      parallel(MonadError[M, E].raiseError(e))
+
+    def handleErrorWith[A](fa: F[A])(f: (E) => F[A]): F[A] = {
+      val ma = MonadError[M, E].handleErrorWith(sequential(fa))(f andThen sequential.apply)
+      parallel(ma)
+    }
+
+    def pure[A](x: A): F[A] = applicative.pure(x)
+
+    def ap[A, B](ff: F[(A) => B])(fa: F[A]): F[B] = applicative.ap(ff)(fa)
```
johnynek: I think we should override basically everything here or we could de-optimize an existing `applicative`.

core/src/main/scala/cats/Parallel.scala
```diff
@@ -0,0 +1,220 @@
...
+
+  /**
+    * A Parallel instance for any type `M[_]` that supports parallel composition through itself.
+    * Can also be used for giving `Parallel` instances to types that do not support parallel composition,
+    * but are required to have an instance of `Parallel` defined,
+    * in which case parallel composition will actually be sequential.
+    */
+  def identity[M[_]: Monad]: Parallel[M, M] = new Parallel[M, M] {
+
+    def monad: Monad[M] = implicitly[Monad[M]]
```
johnynek: I think all these `defs` can be `val` and be a bit nicer on allocations.

project/Boilerplate.scala
```diff
@@ -222,6 +248,44 @@ object Boilerplate {
...
+      val tpesString = tpes mkString ", "
+
+      val tuple = s"Tuple$arity[$tpesString]"
+      val tupleTpe = s"t$arity: $tuple"
+      val tupleArgs = (1 to arity) map { case n => s"t$arity._$n" } mkString ", "
+
+      val n = if (arity == 1) { "" } else { arity.toString }
+
+      val parMap =
+        if (arity == 1) s"def parMap[F[_], Z](f: (${`A..N`}) => Z)(implicit p: NonEmptyParallel[M, F]): M[Z] = p.flatMap.map($tupleArgs)(f)"
```
kailuowang: is the arity=1 version tested?

project/Boilerplate.scala
```diff
@@ -192,6 +194,30 @@ object Boilerplate {
...
+      val fparams = (fargs zip tpes) map { case (v,t) => s"$v:$t"} mkString ", "
+      val fargsS = fargs mkString ", "
+
+      val nestedProducts = (0 until (arity - 2)).foldRight(s"Parallel.parProduct(m${arity - 2}, m${arity - 1})")((i, acc) => s"Parallel.parProduct(m$i, $acc)")
+      val `nested (a..n)` = (0 until (arity - 2)).foldRight(s"(a${arity - 2}, a${arity - 1})")((i, acc) => s"(a$i, $acc)")
+
+      block"""
+         |package cats
+         |trait ParallelArityFunctions {
+        -  def parMap$arity[M[_], F[_], ${`A..N`}, Z]($fparams)(f: (${`A..N`}) => Z)(implicit p: NonEmptyParallel[M, F]): M[Z] =
```
kailuowang: are these `parMap`1~22 tested?

tests/src/test/scala/cats/tests/ListTests.scala
```diff
@@ -1,7 +1,7 @@
 package cats
 package tests
 
-import cats.data.NonEmptyList
+import cats.data.{NonEmptyList}
```
kailuowang: nit, no need for the braces here.

project/Boilerplate.scala
LukaJCB: They're not. Related to #1968

laws/src/main/scala/cats/laws/NonEmptyParallelLaws.scala
```diff
@@ -0,0 +1,21 @@
+package cats
+package laws
+
+
+/**
+  * Laws that must be obeyed by any `cats.NonEmptyParallel`.
+  */
+trait NonEmptyParallelLaws[M[_], F[_]] {
+  def P: NonEmptyParallel[M, F]
+
```
johnynek: Should we say that the Functors should be the same. If not it seems pretty hard to reason about how they are related.

LukaJCB: Yes that's a good idea, I'll add a law!

SystemFw:
Just want to reiterate that this is awesome, great work @LukaJCB

alexandru:
Just wanted to drop a line to say that I'm really glad to see this PR merged 🤗

johnynek:
This with the opaque types SIP will be great.

PR #1838, opened 2017-08-21
Sync NEL and NEV methods (fixes #1832)

### LINKED ISSUE #1832: NonEmptyVector.last
`NonEmptyVector` could have a total `last` method (unlike `Vector`). I think it would be useful.

(`NonEmptyList` could also have this, but I don't know how useful would it be, since it would need to walk the whole list...)

durban:
There is one notable change: NEL previously had a `concat` which accepted another NEL. To be similar to NEV, I changed this (and added `concatNel`).

The following differences are (intentionally) not addressed:
- `groupBy`: NEL has one, but I didn't add it to NEV, because I think that it should have an `Eq` constraint, however implementing that would need a `Map` which works with `Eq`...
- `length`/`size`: NEV has a `length: Int` and a `size: Long` (as an extension method). NEL had a `size: Int` and I've added `length: Int`. (Possibly `size` should be changed to `Long`, or simply removed, to be an extension method.)
- a few `Vector`-specific methods, e.g., `updated`.

This fixes #1832 (cc @johnynek).

### DIFF
--- a/core/src/main/scala/cats/data/NonEmptyList.scala
+++ b/core/src/main/scala/cats/data/NonEmptyList.scala
@@ -80,6 +80,10 @@ final case class NonEmptyList[+A](head: A, tail: List[A]) {
   def concat[AA >: A](other: List[AA]): NonEmptyList[AA] =
     NonEmptyList(head, tail ::: other)
 
+  @deprecated("Use concatNel", since = "1.0.0-RC1")
+  def concat[AA >: A](other: NonEmptyList[AA]): NonEmptyList[AA] =
+    concatNel(other)
+
   /**
    * Append another NonEmptyList
    */
--- a/core/src/main/scala/cats/data/NonEmptyList.scala
+++ b/core/src/main/scala/cats/data/NonEmptyList.scala
@@ -66,6 +66,8 @@ final case class NonEmptyList[+A](head: A, tail: List[A]) {
    */
   def size: Int = 1 + tail.size
 
+  def length: Int = size
+
   /**
    *  Applies f to all the elements of the structure
    */
@@ -73,12 +75,24 @@ final case class NonEmptyList[+A](head: A, tail: List[A]) {
     NonEmptyList(f(head), tail.map(f))
 
   def ++[AA >: A](l: List[AA]): NonEmptyList[AA] =
-    NonEmptyList(head, tail ++ l)
+    concat(l)
+
+  def concat[AA >: A](other: List[AA]): NonEmptyList[AA] =
+    NonEmptyList(head, tail ::: other)
+
+  /**
+   * Append another NonEmptyList
+   */
+  def concatNel[AA >: A](other: NonEmptyList[AA]): NonEmptyList[AA] =
+    NonEmptyList(head, tail ::: other.toList)
 
   def flatMap[B](f: A => NonEmptyList[B]): NonEmptyList[B] =
     f(head) ++ tail.flatMap(f andThen (_.toList))
 
   def ::[AA >: A](a: AA): NonEmptyList[AA] =
+    prepend(a)
+
+  def prepend[AA >: A](a: AA): NonEmptyList[AA] =
     NonEmptyList(a, head :: tail)
 
   /**
@@ -136,12 +150,6 @@ final case class NonEmptyList[+A](head: A, tail: List[A]) {
       tail.collect(pf)
     }
   }
-
-  /**
-   * Append another NonEmptyList
-   */
-  def concat[AA >: A](other: NonEmptyList[AA]): NonEmptyList[AA] =
-    NonEmptyList(head, tail ::: other.toList)
 
   /**
    * Find the first element matching the predicate, if one exists
@@ -405,7 +413,7 @@ private[data] sealed abstract class NonEmptyListInstances extends NonEmptyListIn
       with Monad[NonEmptyList] with NonEmptyTraverse[NonEmptyList] {
 
       def combineK[A](a: NonEmptyList[A], b: NonEmptyList[A]): NonEmptyList[A] =
-        a concat b
+        a concatNel b
 
       override def split[A](fa: NonEmptyList[A]): (A, List[A]) = (fa.head, fa.tail)
 
--- a/core/src/main/scala/cats/data/NonEmptyVector.scala
+++ b/core/src/main/scala/cats/data/NonEmptyVector.scala
@@ -269,6 +269,9 @@ private[data] sealed abstract class NonEmptyVectorInstances {
       override def traverse[G[_], A, B](fa: NonEmptyVector[A])(f: (A) => G[B])(implicit G: Applicative[G]): G[NonEmptyVector[B]] =
         G.map2Eval(f(fa.head), Always(Traverse[Vector].traverse(fa.tail)(f)))(NonEmptyVector(_, _)).value
 
+      override def zipWithIndex[A](fa: NonEmptyVector[A]): NonEmptyVector[(A, Int)] =
+        fa.zipWithIndex
+
       override def foldLeft[A, B](fa: NonEmptyVector[A], b: B)(f: (B, A) => B): B =
         fa.foldLeft(b)(f)
 
--- a/core/src/main/scala/cats/data/NonEmptyVector.scala
+++ b/core/src/main/scala/cats/data/NonEmptyVector.scala
@@ -36,6 +36,10 @@ final class NonEmptyVector[+A] private (val toVector: Vector[A]) extends AnyVal 
 
   def tail: Vector[A] = toVector.tail
 
+  def last: A = toVector.last
+
+  def init: Vector[A] = toVector.init
+
   /**
     * Remove elements not matching the predicate
     *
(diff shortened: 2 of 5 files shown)


core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -67,19 +67,33 @@ final case class NonEmptyList[+A](head: A, tail: List[A]) {
...
-    NonEmptyList(head, tail ++ l)
+    concat(l)
+
+  def concat[AA >: A](other: List[AA]): NonEmptyList[AA] =
+    NonEmptyList(head, tail ::: other)
+
+  /**
+   * Append another NonEmptyList
+   */
+  def concatNel[AA >: A](other: NonEmptyList[AA]): NonEmptyList[AA] =
```
johnynek: why can't this just be `concat`? `NonEmptyList` and `List` are different types. I'm not sure why we took this route and if there is not a good reason I'd rather change it before 1.0.

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -139,12 +153,6 @@ final case class NonEmptyList[+A](head: A, tail: List[A]) {
   }
 
   /**
-   * Append another NonEmptyList
-   */
-  def concat[AA >: A](other: NonEmptyList[AA]): NonEmptyList[AA] =
```
johnynek: I like this naming better. We should add `concat` to NonEmptyVector in my opinion, not remove here.

durban:
@johnynek I guess the `concat`/`concatNel`/`concatNev` is to avoid overloading. I don't feel very strong about it, I just followed NEV. I could live with it either way. (However, I think there are people who really don't like overloading, e.g., there is a wart in WartRemover, which forbids it. Maybe we should ask them...)

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -184,6 +190,18 @@ final class NonEmptyVector[+A] private (val toVector: Vector[A]) extends AnyVal
 
     NonEmptyVector(head, buf.result())
   }
+
+  def reverse: NonEmptyVector[A] =
+    new NonEmptyVector(toVector.reverse)
+
+  def zipWithIndex: NonEmptyVector[(A, Int)] =
```
peterneyens: You could use this to override `Traverse#zipWithIndex` for `NonEmptyVector`.

durban: Done.

durban:
@johnynek Here is the [issue](https://github.com/typelevel/cats/issues/1204) and [PR](https://github.com/typelevel/cats/pull/1212) which removed concat overloading from NEV. This [PR](https://github.com/typelevel/cats/pull/492) also has some discussion about overloading in general.

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -237,6 +255,9 @@ private[data] sealed trait NonEmptyVectorInstances {
       override def traverse[G[_], A, B](fa: NonEmptyVector[A])(f: (A) => G[B])(implicit G: Applicative[G]): G[NonEmptyVector[B]] =
         G.map2Eval(f(fa.head), Always(Traverse[Vector].traverse(fa.tail)(f)))(NonEmptyVector(_, _)).value
 
+      override def zipWithIndex[A](fa: NonEmptyVector[A]): NonEmptyVector[(A, Int)] =
+        fa.zipWithIndex
```
johnynek: this is untested, so I guess that means zipWithIndex is currently lawless.

LukaJCB: We could write a test to check if `zipWithIndex` is consistent with `zipWith(Range(0, size))` :)

LukaJCB:
Hey @durban, are you still working on this? Or would you like some help? :)

LukaJCB:
I agree there's not much left :) 
However could you address the review comments? Some of the extra code isn't tested, you can check this in the  codecov link above! 
Also, you currently have some conflicts, so a rebase would be great :)

durban:
I've rebased and also added more tests for `zipWithIndex`. Please let me know if I've missed anything else.

LukaJCB:
LGTM (pending concensus on `concat`) thank you very much! :)

kailuowang:
👍  I am fine with the`concat` as in this PR. We should probably not go back and forth on overrides.   @johnynek WDYT

LukaJCB:
I've thought a bit more about this and I think having `concat` concatenate a `NEL` and a `List` is fine for the most part, mainly because most concatenation will probably done using the `Semigroup` instance and `|+|` is as short as it gets. 
However, I disagree on keeping `++` as to me, it implies symmetry between the two types. I'd expect something with the operator `++` to take two arguements of the same type on either side, so I think we should just get rid of `++`. People can use `concat` instead.

I'd also be fine with having `concat` be the concatenation between two same types and using `concatList` and `concatVector` accordingly.

kailuowang:
@durban do you still have time to work on this? if you want, I am happy to continue it for you.

durban:
I'd argue that `++` only accepting a NEL/NEV would make it less useful. `++` as it is, has a less "restrictive" signature (it's possible to make a List/Vector from a NEL/NEV; the other direction is not possible). (Please also note, that I absolutely did not touch `++` in this PR.)

kailuowang:
Here is my final proposal, it looks that we have to make some trade off or we are going to have to break either NEL or NEV. The trade off I am willing to make here is to allow overload in this particular case.

So in NEL we are going to have 
```scala
concat(l: List) //new
concat(nel: NEL)
concatNel(nel: NEL) //new
++(l: List)
```

the `concat(nel: NEL)` is the overload to allow backward comp.
in NEV we are going to have  
```scala
concat(l: List)
concat(nev: NEV)  //new
concatNev(nev: NEV)
++(l: Vector)
```
the `concat(nev: NEV)` is the overload to maintain symmetry.

In PR #1949 
we will add to NEL
```
def :::(other: NEL)
```
and add to NEV
```
def ++:(other: NEV)
```
This might mitigate the @LukaJCB 's concern over an asymmetric symbol concatenation. 
WDYT? @johnynek @durban @LukaJCB

jcranky:
For NEV, you mean `concat(v: Vector)`, not `concat(l: List)`, right?

LukaJCB:
I think that's a good idea :+1:

durban:
@kailuowang I'm ok with your proposal. If others also agree, I can add it to this PR.

kailuowang:
@LukaJCB the idea is to avoid unnecessary breaking changes. `++:` and `:::` is kind of inherited from `Vector` and `List`. We keep `++` the same to maintain backward compat. I can see arguments on both side for `++` to take a `NEL` or `List`. So I am not convinced that we need a breaking change to chose one over the other. If we want `++` to take `NEL` and `NEV` I incline to add an override in another PR.

LukaJCB:
I'm good with asymmetry when it means as little breaking changes as possible 👍

durban:
@kailuowang This is not gonna work:
```
[error] def concat[AA >: A](other: Vector[AA]): cats.data.NonEmptyVector[AA] at line 77 and
[error] def concat[AA >: A](other: cats.data.NonEmptyVector[AA]): cats.data.NonEmptyVector[AA] at line 79
[error] have same type after erasure: (other: scala.collection.immutable.Vector)scala.collection.immutable.Vector
```

(For NEL it would work, because that isn't a value class, but at that point we'd lose the consistency.)

There are various ugly workarounds for this, but I think they're more ugly than only having `concatNel`/`concatNev`.

ceedubs:
In previous PRs we've discussed overloading and the general consensus seemed to be that the downsides outweigh the benefits. Personally I'm 👍 for `concatNel` and `concatNev`.

kailuowang:
Okay, can we add `concat(nel: NEL)` back as a deprecated overload.
We will need a scalafix  migration. `concat(nel)` -> `concatNel(nel)`, doesn't need to be this PR though, unless if you are interested in trying out the scalafix stuff.

durban:
@kailuowang Done. I'd skip learning scalafix for now.

tests/src/test/scala/cats/tests/NonEmptyListTests.scala
```diff
@@ -12,6 +12,7 @@ import cats.data.{NonEmptyList, NonEmptyVector}
 import cats.laws.discipline.arbitrary._
 import cats.laws.discipline.{ComonadTests, NonEmptyTraverseTests, MonadTests, ReducibleTests, SemigroupKTests, SerializableTests}
 
+@deprecated("to be able to test deprecated methods", since = "1.0.0-RC1")
```
kailuowang: Sorry I hate to drag on this but I wouldn't deprecate the whole suite for one test. I'd rather we duplicated the test suite class with just the `concat` test in it and deprecate that one.

durban: Done.

LukaJCB:
Hey @durban, sorry to bother you again, but we just need to fix those conflicts, and then we're good to merge! :) If you're too busy, or don't want to invest more time, that's completely okay!

kailuowang:
haha I beat you @LukaJCB

PR #1839, opened 2017-08-21
Update scalastyle-sbt-plugin to 1.0.0

fthomas:
This version is compatible with sbt 1.0.0 and allows to get rid of
a workaround for https://github.com/scalastyle/scalastyle-sbt-plugin/issues/47.

It also complained about new issues which I've ignored for now. The same
issues already have been silenced in other places of the library.


non:
LGTM 👍

Issue #1833, opened 2017-08-21
`Traverse[Free[F, ?]]` is missing.


sellout:
The `Cofree` instance is there, but not one for `Free`, so I’m guessing it’s an oversight.

aaronlevin:
I can take a stab at adding this.

aaronlevin:
We can close this as the above PR was merged. /cc @sellout

PR #1840, opened 2017-08-21
Add Foldable and Traversable instances for Free

aaronlevin:
Addresses https://github.com/typelevel/cats/issues/1833 .

### DIFF
--- a/free/src/main/scala/cats/free/Free.scala
+++ b/free/src/main/scala/cats/free/Free.scala
@@ -55,6 +55,20 @@ sealed abstract class Free[S[_], A] extends Product with Serializable {
   }
 
   /**
+   * A combination of step and fold.
+   */
+  private[free] final def foldStep[B](
+    onPure: A => B,
+    onSuspend: S[A] => B,
+    onFlatMapped: ((S[X], X => Free[S, A]) forSome { type X }) => B
+  ): B = this.step match {
+    case Pure(a) => onPure(a)
+    case Suspend(a) => onSuspend(a)
+    case FlatMapped(Suspend(fa), f) => onFlatMapped((fa, f))
+    case _ => sys.error("FlatMapped should be right associative after step")
+  }
+
+  /**
    * Run to completion, using a function that extracts the resumption
    * from its suspension functor.
    */
@@ -161,7 +175,7 @@ sealed abstract class Free[S[_], A] extends Product with Serializable {
     "Free(...)"
 }
 
-object Free {
+object Free extends FreeInstances {
 
   /**
    * Return from the computation with the given value.
@@ -248,5 +262,58 @@ object Free {
       def pure[A](a: A): Free[S, A] = Free.pure(a)
       override def map[A, B](fa: Free[S, A])(f: A => B): Free[S, B] = fa.map(f)
       def flatMap[A, B](a: Free[S, A])(f: A => Free[S, B]): Free[S, B] = a.flatMap(f)
+    }
+}
+
+private trait FreeFoldable[F[_]] extends Foldable[Free[F, ?]] {
+
+  implicit def F: Foldable[F]
+
+  override final def foldLeft[A, B](fa: Free[F, A], b: B)(f: (B, A) => B): B =
+    fa.foldStep(
+      a => f(b, a),
+      fa => F.foldLeft(fa, b)(f),
+      { case (fx, g) => F.foldLeft(fx, b)((bb, x) => foldLeft(g(x), bb)(f)) }
+    )
+
+  override final def foldRight[A, B](fa: Free[F, A], lb: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B] =
+    fa.foldStep(
+      a => f(a, lb),
+      fa => F.foldRight(fa, lb)(f),
+      { case (fx, g) => F.foldRight(fx, lb)( (a, lbb) => foldRight(g(a), lbb)(f)) }
+    )
+}
+
+private trait FreeTraverse[F[_]] extends Traverse[Free[F, ?]] with FreeFoldable[F] {
+  implicit def TraversableF: Traverse[F]
+
+  def F: Foldable[F] = TraversableF
+
+  override final def traverse[G[_], A, B](fa: Free[F, A])(f: A => G[B])(implicit G: Applicative[G]): G[Free[F, B]] =
+    fa.resume match {
+      case Right(a) => G.map(f(a))(Free.pure(_))
+      case Left(ffreeA) => G.map(TraversableF.traverse(ffreeA)(traverse(_)(f)))(Free.roll(_))
+    }
+
+  // Override Traverse's map to use Free's map for better performance
+  override final def map[A, B](fa: Free[F, A])(f: A => B): Free[F, B] = fa.map(f)
+}
+
+sealed private[free] abstract class FreeInstances {
+
+  implicit def catsFreeFoldableForFree[F[_]](
+    implicit
+    foldableF: Foldable[F]
+  ): Foldable[Free[F, ?]] =
+    new FreeFoldable[F] {
+      val F = foldableF
+    }
+
+  implicit def catsFreeTraverseForFree[F[_]](
+    implicit
+    traversableF: Traverse[F]
+  ): Traverse[Free[F, ?]] =
+    new FreeTraverse[F] {
+      val TraversableF = traversableF
     }
 }
--- a/free/src/test/scala/cats/free/FreeTests.scala
+++ b/free/src/test/scala/cats/free/FreeTests.scala
@@ -3,7 +3,7 @@ package free
 
 import cats.arrow.FunctionK
 import cats.data.EitherK
-import cats.laws.discipline.{CartesianTests, MonadTests, SerializableTests}
+import cats.laws.discipline.{CartesianTests, FoldableTests, MonadTests, SerializableTests, TraverseTests}
 import cats.laws.discipline.arbitrary.catsLawsArbitraryForFn0
 import cats.tests.CatsSuite
 
@@ -17,6 +17,19 @@ class FreeTests extends CatsSuite {
 
   checkAll("Free[Option, ?]", MonadTests[Free[Option, ?]].monad[Int, Int, Int])
   checkAll("Monad[Free[Option, ?]]", SerializableTests.serializable(Monad[Free[Option, ?]]))
+
+  locally {
+    implicit val instance = Free.catsFreeFoldableForFree[Option]
+
+    checkAll("Free[Option, ?]", FoldableTests[Free[Option,?]].foldable[Int,Int])
+    checkAll("Foldable[Free[Option,?]]", SerializableTests.serializable(Foldable[Free[Option,?]]))
+  }
+
+  locally {
+    implicit val instance = Free.catsFreeTraverseForFree[Option]
+    checkAll("Free[Option,?]", TraverseTests[Free[Option,?]].traverse[Int, Int, Int, Int, Option, Option])
+    checkAll("Traverse[Free[Option,?]]", SerializableTests.serializable(Traverse[Free[Option,?]]))
+  }
 
   test("toString is stack-safe") {
     val r = Free.pure[List, Int](333)


aaronlevin:
Hmm, the `JS` build failed. I'll look into this.

free/src/main/scala/cats/free/Free.scala
```diff
@@ -250,3 +250,40 @@ object Free {
...
+  implicit def FunctorF: Functor[F]
+
+  override final def foldLeft[A,B](fa: Free[F,A], b: B)(f: (B,A) => B): B =
+    fa.fold(f(b,_), F.foldLeft(_,b)((bb,freeA) => foldLeft(freeA,bb)(f)))
+
+  override final def foldRight[A,B](fa: Free[F,A], lb: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B] =
+    fa.fold(f(_, lb), F.foldRight(_,lb)( (freeA,eb) => foldRight(freeA,eb)(f)))
+}
+
+private trait FreeTraverse[F[_]] extends Traverse[Free[F,?]] with FreeFoldable[F] {
```
peterneyens: You can override `map` here to use `Free#map` which is more efficient than the default `map` for `Traverse` (`traverse` with `Id`).

free/src/main/scala/cats/free/Free.scala
```diff
@@ -250,3 +250,40 @@ object Free {
...
+
+sealed private[free] abstract class FreeInstances {
+
+  implicit def catsFoldableForFree[F[_] : Functor : Foldable]: Foldable[Free[F,?]] =
+    new FreeFoldable[F] {
+      def F = implicitly
+      def FunctorF = implicitly
+    }
+
+  implicit def catsTraverseForFree[F[_] : Functor : Traverse]: Traverse[Free[F, ?]] =
```
peterneyens: `Traverse` is a `Functor` as well, so you can drop the `Functor` requirement.

I think we generally prefer "explicit implicits" (is that a thing?), compared to context bounds.

free/src/main/scala/cats/free/Free.scala
```diff
@@ -250,3 +250,40 @@ object Free {
...
+    new FreeTraverse[F] {
+      def F = implicitly
+      def FunctorF = implicitly
```
peterneyens: These `def`s can be `val`s.

free/src/main/scala/cats/free/Free.scala
aaronlevin: Good point, I'll do that!

free/src/main/scala/cats/free/Free.scala
aaronlevin: Hah, I was copying the style from `Cofree`, but I'm happy to add explicit implicits.

aaronlevin:
One question I had regarded which `Foldable[F]` instance to use in `Traverse[Free[F,?]]`. I'm assuming this _must_ be the `Foldable[F]` which corresponds to the `Traverse[F]`. I tried to check the `Traverse` laws but didn't find anything online stating how `Traverse` relates to `Foldable`.

If it's not the case that the `Foldable[F]` used in `Traverse[Free[F,?]]` does not need to correspond to the `Traverse[F]` instance, then I think the selection should be pushed down to the `catsTraverseForFree` (effectively adding a third implicit for `Foldable[F]`). 

But my gut tells me how its implemented now is correct (fwiw both styles pass the tests).

free/src/main/scala/cats/free/Free.scala
```diff
@@ -250,3 +250,53 @@ object Free {
...
+      ffreeA => G.map(TraversableF.traverse(ffreeA)(traverse(_)(f)))(Free.roll(_))
+    )
+
+  // Override Traverse's map to use Free's map for better performance
+  override final def map[A,B](fa: Free[F, A])(f: A => B): Free[F, B] = fa.map(f)
+}
+
+sealed private[free] abstract class FreeInstances {
+
+  implicit def catsFoldableForFree[F[_]](
```
peterneyens: It is not really documented outside issue https://github.com/typelevel/cats/issues/1061 itself (it should probably be added to [these guidelines](https://typelevel.org/cats/guidelines.html)): the implicit values in Cats are prefixed with the package.

So these implicits would be `catsFreeFoldableForFree` and `catsFreeTraverseForFree`.

free/src/main/scala/cats/free/Free.scala
```diff
@@ -250,3 +250,53 @@ object Free {
...
+    }
+
+  implicit def catsTraverseForFree[F[_]](
+    implicit
+    functorF: Functor[F],
+    traversableF: Traverse[F]
+  ): Traverse[Free[F, ?]] =
+    new FreeTraverse[F] {
+      val TraversableF = traversableF
+      val FunctorF = functorF
```
peterneyens: You might have missed my previous comment, but `Traverse[F]` is also a `Functor[F]`.
So you could write : 

```scala
implicit def catsTraverseForFree[F[_]](implicit F: Traverse[F]): Traverse[Free[F, ?]] =
  new FreeTraverse[F] {
    val TraversableF = F
    val FunctorF = F
  }
```

peterneyens:
Generally if a datatype implements multiple typeclasses in a hierarchy, the the implicits are overridden by a more powerful instance.

For example in `OptionT` : 
- [`Foldable`](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/OptionT.scala#L280) and [`Traverse`](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/OptionT.scala#L290).
- [`Eq`](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/OptionT.scala#L320), [`PartialOrder`](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/OptionT.scala#L326) and [`Order`](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/OptionT.scala#L332).

free/src/main/scala/cats/free/Free.scala
aaronlevin: Ah, yes, I did miss that. Thanks!

free/src/main/scala/cats/free/Free.scala
aaronlevin: cool. Thanks, I've updated.

aaronlevin: I also created an issue for this: https://github.com/typelevel/cats/issues/1841

aaronlevin:
FYI, I'll squash and rebase when the tests pass and I've incorporated all your feedback. Thanks again, @peterneyens !!

aaronlevin:
The tests passed, so I've squashed. Should be ready to merge.

peterneyens:
Given that these methods are recursive, we should probably use `resume` and pattern match instead of `fold` to reduce some stack space.

I think we can make `foldRight` stack safe as well :
```scala
fa.resume match {
  case Left(fFreeA) =>
    F.foldRight(fFreeA, lb)( (freeA, eb) => 
      eb.flatMap(b => foldRight(freeA, Eval.now(b))(f))
    )
  case Right(a) => f(a, lb)
}
```

peterneyens:
Hmm, it seems `Foldable[Free[F, ?]]` as it stands also requires `Functor[F]` as it uses `fold` (or `resume`), which means that there is only a `Foldable` instance for `Free` given a `Traverse[F]` (`Foldable` + `Functor`).

It seems that Scalaz uses a `foldStep` method which uses the underlying representation of `Free` (which doesn't need a `Functor[F]` in `Free.liftF`) to support `Foldable[Free[F, ?]]` for just `Foldable[F]`.

So either we have to add something similar to `foldStep` or we only provide the `Traverse` instance.

aaronlevin:
@peterneyens ok, I'll add `foldStep` and see about implementing `Foldable` without `Functor` (I double-checked the Haskell implementation and indeed the `Functor[F]` constraint is not present).

aaronlevin:
(I'll also see about making `foldRight` stack-safe)

aaronlevin:
@peterneyens OK, so I've implemented `foldStep` and removed the `Functor` constraint on `Foldable[Free[F,?]]`. My one issue is I don't really know how / what to test for `foldStep`. It's kind of a weird function.

peterneyens:
I think we probably want to hide `foldStep` as it would expose the actual implementation of `Free` compared to the one provided by `resume` / `fold` / ...

If we hide `foldStep` I don't think it needs tests, `step` also doesn't have tests itself. 

As the name suggest `foldStep` is a combination of `step` and a  `fold` and I think it could have been implemented using `step` as well : 

```scala
this.step match {
  case Pure(a) => onPure(a)
  case Suspend(a) => onSuspend(a)
  case FlatMapped(Suspend(fa), f) => onFlatMapped((fa, f))
  case _ => sys.error("FlatMapped should be right associative after step") // :(
}
```

aaronlevin:
@peterneyens ok, I've implemented `foldStep` in terms of `step` as you suggested. I've also rebased and (I hope) fixed all the build issues.

peterneyens:
Not sure if you know, but you can run scalastyle locally as well in sbt using `scalastyle`.

aaronlevin:
ah, hah. I was doing that before, but I wasn't prefixing it with the project (e.g. `scalastyle` vs `freeJVM/scalastyle`) and I was confused because the errors in travis were happening on the `JS` project. _anyway_, got it now and that will be helpful for sure!

everything should be good now.

free/src/main/scala/cats/free/Free.scala
```diff
@@ -55,6 +55,20 @@ sealed abstract class Free[S[_], A] extends Product with Serializable {
   }
 
   /**
+   * A combination of step and fold.
+   */
+  final def foldStep[B](
```
peterneyens: I think we need to make this `private[free]` to not expose the actual representation of `Free` compared to `Either[F[Free[F, A]]], A]` which is exposed by `resume` and `fold`.

free/src/main/scala/cats/free/Free.scala
```diff
@@ -250,3 +264,57 @@ object Free {
...
+      a => f(b, a),
+      fa => F.foldLeft(fa, b)(f),
+      { case (fx, g) => F.foldLeft(fx, b)((bb, x) => foldLeft(g(x), bb)(f)) }
+    )
+
+  override final def foldRight[A, B](fa: Free[F, A], lb: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B] =
+    fa.foldStep(
+      a => f(a, lb),
+      fa => F.foldRight(fa, lb)(f),
+      { case (fx, g) => F.foldRight(fx, lb)( (a, lbb) => foldRight(g(a), lbb)(f)) }
```
peterneyens: I think that `lbb.flatMap(bb => foldRight(g(a), Eval.now(bb))(f))` instead of `foldRight(g(a), lbb)(f)` should make this stack safe.

We should probably add a test to check this. Something like (haven't compiled this):

```scala
val n = 50000
val freeOption: Int => Free[Option, Int] = Free.pure(x)
val free = (0 to n).foldLeft(freeOption(0))((r, _) => r.flatMap(n => freeOption(n + 1)))
free.foldRight(0)((a, lb) => lb.map(_ + a)) should (n)
```

free/src/main/scala/cats/free/Free.scala
aaronlevin: done!

free/src/main/scala/cats/free/Free.scala
aaronlevin: i've added the test!

aaronlevin:
@peterneyens I'm not savvy about understanding `Free`'s stack usage, but two comments about that test:

1. as the test is written, both my definition and your definition pass (i.e. I did not observe the stack overflow)
2. I also used the same test as above to test `foldLeft` and it was also stack safe.

Is this expected.

peterneyens:
I am sorry, it seems my comments about `foldRight` don't apply any more after you introduced `foldStep`.

Feel free to revert back to your original `foldRight` implementation. Maybe you can adapt the `foldRight` stack safety test to check `foldLeft` as well?

aaronlevin:
@peterneyens ok, I've updated and added the `foldLeft` tests. I'm not quite sure if the tests actually test the stack safety as I've not yet seen a failing case, but I can try to find the old, pre-`foldStep` implementation and check that it breaks the stack.

aaronlevin:
@peterneyens I just tried with the old implementations of `foldLeft` and `foldRight` using `Free.fold` (and adding the `Functor[F]` constraint), and the stacks did not explode.

So I'm not convinced that example is actually testing the stack. I'm also unsure if there is a problem with the stacks. Perhaps the implementation of `foldLeft` and `foldRight` are ok?

peterneyens:
It seems you are right, not sure what I tried yesterday which let to a stack overflow. Sorry for the confusion.

free/src/main/scala/cats/free/Free.scala
```diff
@@ -250,3 +264,57 @@ object Free {
...
+      val F = foldableF
+    }
+
+  implicit def catsFreeTraverseForFree[F[_]](
+    implicit
+    traversableF: Traverse[F]
+  ): Traverse[Free[F, ?]] =
+    new FreeTraverse[F] {
+      val TraversableF = traversableF
+      val FunctorF = traversableF
```
peterneyens: This `val FunctorF` isn't used any more right?

aaronlevin: fixed.

aaronlevin:
@peterneyens ok, I've just squashed everything. so I think this is good. Who can merge this?

PR #1845, opened 2017-08-23
Adding the origami project to the list of Typelevel projects using cats

etorreborre:
This is a proposal to add `origami` to the list of Typelevel projects using cats. However if I remember correctly one of the criteria for being part of the system is to have a compiler-checked documentation which doesn't exist anymore (it was there on a previous version of the library). I will add it and maybe then we can merge in this PR? Is anything else needed?


fthomas:
Having compiler-checked documentation is a requirement for being a Typelevel project. I don't think we have such strict requirements for the list of projects that use Cats data types and type classes.

Therefore :+1:

etorreborre:
Thanks, I guess my next PR will be on `https://github.com/typelevel/typelevel.github.com` then :-).

PR #1846, opened 2017-08-23
Add `grafter` to the list of Typelevel projects using cats

etorreborre:
This is for your consideration. Grafter uses `Reader` instances for component dependencies and `Eval` for actions starting-up the application graph.


fthomas:
:+1:

PR #1847, opened 2017-08-23
Add right and left functor to BiFunctor

### LINKED ISSUE #1834: Access to implied `Functor`s from `Bifunctor` (and `Bifoldable`, `Bitraverse`)
Scalaz adds `leftFunctor` and `rightFunctor` members to the instance. I can imagine other approaches, but it seems like we should have some way to get these from the `Bi` implementations.

LukaJCB:
Quick and dirty, should resolve #1834.


non:
Could we use a `private abstract class` (or even `private case class`) instead of `trait`?

Classes have a better binary compatibility story on 2.10 and 2.11 and I don't think we'll ever want to use these traits as mix-ins.

LukaJCB:
Done :)

non:
Thanks! LGTM 👍

PR #1852, opened 2017-08-24
Remove sbt-ghpages because it is pulled in by sbt-microsites

fthomas:
sbt-ghpages is a dependency of sbt-microsites and therefore does
not need to be listed explicitly as plugin. We have the same situation
with tut whose keys we are using in our `build.sbt` but do not explicitly
list in `plugins.sbt`.


tpolecat:
👍

PR #1854, opened 2017-08-25
Add EitherT docs

Technius:
I've been using `EitherT` recently, so I thought it would be a good idea to add some docs for it.

### DIFF
--- a/docs/src/main/tut/datatypes/eithert.md
+++ b/docs/src/main/tut/datatypes/eithert.md
@@ -14,12 +14,10 @@ following program:
 
 ```tut:book
 import scala.util.Try
+import cats.syntax.either._
 
 def parseDouble(s: String): Either[String, Double] =
-  Try(s.toDouble).toEither match {
-    case Left(_) => Left(s"$s is not a number")
-    case Right(n) => Right(n)
-  }
+  Try(s.toDouble).map(Right(_)).getOrElse(Left(s"$s is not a number"))
 
 def divide(a: Double, b: Double): Either[String, Double] =
   Either.cond(b != 0, a / b, "Cannot divide by zero")
@@ -66,7 +64,7 @@ def divisionProgramAsync(inputA: String, inputB: String): Future[Either[String, 
   }
 ```
 
-Clearly, the updated code is less readible and more verbose: the details of the
+Clearly, the updated code is less readable and more verbose: the details of the
 program are now mixed with the error handling. In addition, as more `Either`s
 and `Futures` are included, the amount of boilerplate required to properly
 handle the errors will increase dramatically.
@@ -94,8 +92,8 @@ divisionProgramAsync("4", "2").value // Future(Right(2.0))
 divisionProgramAsync("a", "b").value // Future(Left("a is not a number"))
 ```
 
-Note that since `EitherT` is a monad, monadic combinators such as `flatMap` can
-be used to compose `EitherT` values.
+Note that when `F` is a monad, then `EitherT` will also form a monad, allowing
+monadic combinators such as `flatMap` to be used in composing `EitherT` values.
 
 ## From `A` or `B` to `EitherT[F, A, B]`
 
@@ -110,7 +108,7 @@ val error: EitherT[Option, String, Int] = EitherT.leftT("Not a number")
 
 ## From `F[A]` or `F[B]` to `EitherT[F, A, B]`
 
-Similary, use `EitherT.left` and `EitherT.right` to convert a `F[A]` or a `F[B]`
+Similary, use `EitherT.left` and `EitherT.right` to convert an `F[A]` or an `F[B]`
 into an `EitherT`. It is also possible to use `EitherT.liftT` as an alias for
 `EitherT.right`.
 
@@ -125,7 +123,7 @@ val error: EitherT[Option, String, Int] = EitherT.left(errorO)
 ## From `Either[A, B]` or `F[Either[A, B]]` to `EitherT[F, A, B]`
 
 Use `EitherT.fromEither` to a lift a value of `Either[A, B]` into `EitherT[F, A, B]`.
-A `F[Either[A, B]]` can be converted into `EitherT` using the `EitherT` constructor.
+An `F[Either[A, B]]` can be converted into `EitherT` using the `EitherT` constructor.
 
 ```tut:silent
 val numberE: Either[String, Int] = Right(100)
@@ -139,7 +137,7 @@ val numberFET: EitherT[List, String, Int] = EitherT(numberFE)
 
 ## From `Option[B]` or `F[Option[B]]` to `EitherT[F, A, B]`
 
-An `Option[B]` or a `F[Option[B]]`, along with a default value, can be passed to
+An `Option[B]` or an `F[Option[B]]`, along with a default value, can be passed to
 `EitherT.fromOption` and `EitherT.fromOptionF`, respectively, to produce an
 `EitherT`.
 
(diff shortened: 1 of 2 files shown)


docs/src/main/tut/datatypes/eithert.md
```diff
@@ -0,0 +1,160 @@
...
+To obtain a left version or a right version of `EitherT` when given an `A` or a
+`B`, use `EitherT.leftT` and `EitherT.rightT` (which is an alias for
+`EitherT.pure`), respectively.
+
+```tut:silent
+val number: EitherT[Option, String, Int] = EitherT.rightT(5)
+val error: EitherT[Option, String, Int] = EitherT.leftT("Not a number")
+```
+
+## From `F[A]` or `F[B]` to `EitherT[F, A B]`
```
iravid: Missing a comma in `EitherT[F, A*,* B]`

docs/src/main/tut/datatypes/eithert.md
```diff
@@ -0,0 +1,162 @@
...
+      
+      parseResult match {
+        case Right((a, b)) => divideAsync(a, b)
+        case l@Left(err) => Future.successful(Left(err))
+      }
+    }
+  }
+```
+
+Clearly, the updated code is less readible and more verbose: the details of the
```
LukaJCB: "readable"

docs/src/main/tut/datatypes/eithert.md
```diff
@@ -0,0 +1,162 @@
...
+    b <- EitherT(parseDoubleAsync(inputB))
+    result <- EitherT(divideAsync(a, b))
+  } yield result
+
+divisionProgramAsync("4", "2").value // Future(Right(2.0))
+divisionProgramAsync("a", "b").value // Future(Left("a is not a number"))
+```
+
+Note that since `EitherT` is a monad, monadic combinators such as `flatMap` can
+be used to compose `EitherT` values.
```
LukaJCB: Maybe clarify here, that `EitherT` forms a Monad, only if `F[_]` is also a Monad :)

LukaJCB:
Also your Tut examples seem to fail on 2.11.x and 2.10.x. You can execute `+ tut` in the sbt console to test on all Scala versions :)

docs/src/main/tut/datatypes/eithert.md
```diff
@@ -0,0 +1,162 @@
...
+val myOptionET = EitherT.fromOption[Future](myOption, "option not defined")
+val myOptionListET = EitherT.fromOptionF(myOptionList, "option not defined")
+```
+
+## Extracting an `F[Either[A, B]]` from an `EitherT[F, A, B]`
+
+Use the `value` method defined on `EitherT` to retrieve the underlying `F[Either[A, B]]`:
+
+```tut:book
+val errorT: EitherT[Option, String, Int] = EitherT.leftT("foo")
```
LukaJCB: I know it's just an example, but an `EitherT[Option, String, Int]` doesn't really seem that useful, and I'm not sure if we should include those in the official docs. Maybe change it to `EitherT[Future, String, Int]` instead? :)

Technius: Yeah, that's much better actually.

docs/src/main/tut/datatypes/eithert.md
```diff
@@ -0,0 +1,160 @@
...
+# EitherT
+
+`Either` can be used for error handling in most situations. However, when
+`Either` is placed into effectful types such as `Option` or`Future`, a large
+amount of boilerplate is required to handle errors. For example, consider the
+following program:
+
+```tut:book
+import scala.util.Try
+import cats.syntax.either._
```
peterneyens: We decide to use the uber import `import cats.implicits._` in the documentation, see issue #1026.

docs/src/main/tut/datatypes/eithert.md
```diff
@@ -0,0 +1,160 @@
...
+
+def divisionProgramAsync(inputA: String, inputB: String): EitherT[Future, String, Double] =
+  for {
+    a <- EitherT(parseDoubleAsync(inputA))
+    b <- EitherT(parseDoubleAsync(inputB))
+    result <- EitherT(divideAsync(a, b))
+  } yield result
+
+divisionProgramAsync("4", "2").value // Future(Right(2.0))
+divisionProgramAsync("a", "b").value // Future(Left("a is not a number"))
```
peterneyens: If you want to show the output, we should move these comments to a "non silent tut block", that way nobody can forget to update the comments if the code is changed.

docs/src/main/tut/datatypes/eithert.md
```diff
@@ -0,0 +1,160 @@
...
+def divideAsync(a: Double, b: Double): Future[Either[String, Double]] =
+  Future.successful(divide(a, b))
+
+def divisionProgramAsync(inputA: String, inputB: String): Future[Either[String, Double]] =
+  parseDoubleAsync(inputA) flatMap { eitherA =>
+    parseDoubleAsync(inputB) flatMap { eitherB =>
+      val parseResult = for {
+        a <- eitherA
+        b <- eitherB
+      } yield (a, b)
```
peterneyens: You could write `(eitherA, eitherB).tupled`.

docs/src/main/tut/datatypes/eithert.md
Technius: Using `tut:book`, it shows
```scala
divisionProgramAsync("4", "2").value
// res4: scala.concurrent.Future[Either[String,Double]] = Future(<not completed>)

divisionProgramAsync("a", "b").value
// res5: scala.concurrent.Future[Either[String,Double]] = Future(<not completed>)
```
which is why I opted to use the comments.

docs/src/main/tut/datatypes/eithert.md
Technius: That syntax seems obscure to me, so I'm just replacing the `parseResult` snippet with a pattern match on the eithers.

docs/src/main/tut/datatypes/eithert.md
```diff
@@ -51,14 +51,10 @@ def divideAsync(a: Double, b: Double): Future[Either[String, Double]] =
...
-        a <- eitherA
-        b <- eitherB
-      } yield (a, b)
-      
-      parseResult match {
-        case Right((a, b)) => divideAsync(a, b)
-        case l@Left(err) => Future.successful(Left(err))
+      (eitherA, eitherB) match {
+        case (Right(a), Right(b)) => divideAsync(a, b)
+        case (l@Left(err), _) => Future.successful(Left(err))
```
LukaJCB: `l` is unused here, right? Why not just make it `Left(err)`? :)

docs/src/main/tut/datatypes/eithert.md
kailuowang: IMO, doesn't hurt to do an `Await.result` here.

kailuowang:
LGTM, I left a minor comment there but could be addressed by another PR.

docs/src/main/tut/datatypes/eithert.md
```diff
@@ -0,0 +1,156 @@
...
+`Either` is placed into effectful types such as `Option` or`Future`, a large
+amount of boilerplate is required to handle errors. For example, consider the
+following program:
+
+```tut:book
+import scala.util.Try
+import cats.implicits._
+
+def parseDouble(s: String): Either[String, Double] =
+  Try(s.toDouble).map(Right(_)).getOrElse(Left(s"$s is not a number"))
```
ceedubs: Super nitpicky and unreasonable comment: `s` could still technically be a number that can't be represented as a `Double`. It might be best to leave it simple and just say "`$s is not a valid double`.

Issue #1743, opened 2017-06-26
partitionE for Nel


fommil:
any interest in this?

```scala
  // a partitionE for Nel
  implicit class PartitionE[E](trav: NonEmptyList[E]) {
    @SuppressWarnings(Array("org.wartremover.warts.Throw"))
    def partitionE[A, B](
        f: E => Either[A, B]
      ): (Ior[NonEmptyList[A], NonEmptyList[B]]) = {
      val (lefts, rights) = trav.toList.map(f).partition(_.isLeft)
      Ior
        .fromOptions(
          Nel(lefts.collect { case Left(left) => left }),
          Nel(rights.collect { case Right(right) => right })
        )
        .getOrElse {
          throw new IllegalStateException("impossible")
        }
    }
  }
````

edmundnoble:
Love it, :+1:

LukaJCB:
Why was this closed? I still think this is a valuable addition 🤔

fommil:
I don't have plans to contribute it.

PR #1858, opened 2017-08-26
Add NonEmptyList#partitionE

LukaJCB:
As per #1743 (now closed). Also open for alternative names (`partitionIor` maybe?)

### DIFF
--- a/core/src/main/scala/cats/Alternative.scala
+++ b/core/src/main/scala/cats/Alternative.scala
@@ -22,21 +22,6 @@ import simulacrum.typeclass
     (as, bs)
   }
 
-  /**
-    * Separate this Foldable into a Tuple by a separating function `A => Either[B, C]`
-    */
-  def mapSeparate[A, B, C](fa: F[A])(f: A => Either[B, C])(implicit F: Foldable[F], A: Alternative[F]): (F[B], F[C]) = {
-    import cats.instances.tuple._
-
-    implicit val mb: Monoid[F[B]] = A.algebra[B]
-    implicit val mc: Monoid[F[C]] = A.algebra[C]
-
-    F.foldMap(fa)(a => f(a) match {
-      case Right(c) => (empty[B], pure(c))
-      case Left(b) => (pure(b), empty[C])
-    })
-  }
-
   override def compose[G[_]: Applicative]: Alternative[λ[α => F[G[α]]]] =
     new ComposedAlternative[F, G] {
       val F = self
--- a/core/src/main/scala/cats/Alternative.scala
+++ b/core/src/main/scala/cats/Alternative.scala
@@ -22,6 +22,18 @@ import simulacrum.typeclass
     (as, bs)
   }
 
+  def mapSeparate[A, B, C](fa: F[A])(f: A => Either[B, C])(implicit F: Foldable[F], A: Alternative[F]): (F[B], F[C]) = {
+    import cats.instances.tuple._
+
+    implicit val mb: Monoid[F[B]] = Alternative[F].algebra[B]
+    implicit val mc: Monoid[F[C]] = Alternative[F].algebra[C]
+
+    F.foldMap(fa)(a => f(a) match {
+      case Right(c) => (empty[B], pure(c))
+      case Left(b) => (pure(b), empty[C])
+    })
+  }
+
   override def compose[G[_]: Applicative]: Alternative[λ[α => F[G[α]]]] =
     new ComposedAlternative[F, G] {
       val F = self
--- a/core/src/main/scala/cats/Alternative.scala
+++ b/core/src/main/scala/cats/Alternative.scala
@@ -25,8 +25,8 @@ import simulacrum.typeclass
   def mapSeparate[A, B, C](fa: F[A])(f: A => Either[B, C])(implicit F: Foldable[F], A: Alternative[F]): (F[B], F[C]) = {
     import cats.instances.tuple._
 
-    implicit val mb: Monoid[F[B]] = Alternative[F].algebra[B]
-    implicit val mc: Monoid[F[C]] = Alternative[F].algebra[C]
+    implicit val mb: Monoid[F[B]] = A.algebra[B]
+    implicit val mc: Monoid[F[C]] = A.algebra[C]
 
     F.foldMap(fa)(a => f(a) match {
       case Right(c) => (empty[B], pure(c))
--- a/core/src/main/scala/cats/Alternative.scala
+++ b/core/src/main/scala/cats/Alternative.scala
@@ -22,6 +22,9 @@ import simulacrum.typeclass
     (as, bs)
   }
 
+  /**
+    * Separate this Foldable into a Tuple by a separating function `A => Either[B, C]`
+    */
   def mapSeparate[A, B, C](fa: F[A])(f: A => Either[B, C])(implicit F: Foldable[F], A: Alternative[F]): (F[B], F[C]) = {
     import cats.instances.tuple._
 
--- a/core/src/main/scala/cats/Foldable.scala
+++ b/core/src/main/scala/cats/Foldable.scala
@@ -407,6 +407,21 @@ import simulacrum.typeclass
     }.toList
 
   /**
+    * Separate this Foldable into a Tuple by a separating function `A => Either[B, C]`
+    */
+  def partitionEither[A, B, C](fa: F[A])(f: A => Either[B, C])(implicit F: Foldable[F], A: Alternative[F]): (F[B], F[C]) = {
+    import cats.instances.tuple._
+
+    implicit val mb: Monoid[F[B]] = A.algebra[B]
+    implicit val mc: Monoid[F[C]] = A.algebra[C]
+
+    F.foldMap(fa)(a => f(a) match {
+      case Right(c) => (A.empty[B], A.pure(c))
+      case Left(b) => (A.pure(b), A.empty[C])
+    })
+  }
+
+  /**
    * Convert F[A] to a List[A], only including elements which match `p`.
    */
   def filter_[A](fa: F[A])(p: A => Boolean): List[A] =
--- a/core/src/main/scala/cats/Foldable.scala
+++ b/core/src/main/scala/cats/Foldable.scala
@@ -409,13 +409,13 @@ import simulacrum.typeclass
   /**
     * Separate this Foldable into a Tuple by a separating function `A => Either[B, C]`
     */
-  def partitionEither[A, B, C](fa: F[A])(f: A => Either[B, C])(implicit F: Foldable[F], A: Alternative[F]): (F[B], F[C]) = {
+  def partitionEither[A, B, C](fa: F[A])(f: A => Either[B, C])(implicit A: Alternative[F]): (F[B], F[C]) = {
     import cats.instances.tuple._
 
     implicit val mb: Monoid[F[B]] = A.algebra[B]
     implicit val mc: Monoid[F[C]] = A.algebra[C]
 
-    F.foldMap(fa)(a => f(a) match {
+    foldMap(fa)(a => f(a) match {
       case Right(c) => (A.empty[B], A.pure(c))
       case Left(b) => (A.pure(b), A.empty[C])
     })
--- a/core/src/main/scala/cats/Foldable.scala
+++ b/core/src/main/scala/cats/Foldable.scala
@@ -408,6 +408,16 @@ import simulacrum.typeclass
 
   /**
     * Separate this Foldable into a Tuple by a separating function `A => Either[B, C]`
+    * Equivalent to `Functor#map` and then `Alternative#separate`.
+    *
+    * {{{
+    * scala> import cats.implicits._
+    * scala> val list = List(1,2,3,4)
+    * scala> Foldable[List].partitionEither(list)(a => if (a % 2 == 0) Left(a.toString) else Right(a))
+    * res0: (List[String], List[Int]) = (List(2, 4),List(1, 3))
+    * scala> Foldable[List].partitionEither(list)(a => Right(a * 4))
+    * res1: (List[Nothing], List[Int]) = (List(),List(4, 8, 12, 16))
+    * }}}
     */
   def partitionEither[A, B, C](fa: F[A])(f: A => Either[B, C])(implicit A: Alternative[F]): (F[B], F[C]) = {
     import cats.instances.tuple._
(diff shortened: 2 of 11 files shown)


kailuowang:
Would it be more useful to add it to `NonEmptyTraverse` with a slightly different signature? 
```scala
def partitionIor[B, C](f: A => Ior[B, C]): Ior[NonEmptyList[B], NonEmptyList[C]] = {
```

LukaJCB:
You mean generalizing it to `NonEmptyTraverse`?
 
```scala
def partitionIor[F[_]: NonEmptyTraverse, A, B, C](fa: F[A])(f: A => Ior[B, C]): Ior[F[B], F[C]]
```

Not sure, if that works, but would be cool :)

kailuowang:
@LukaJCB i didn't mean tp generalize the result `NonEmptyList` part (not sure if that's doable). I.e.
for `NonEmptyVector` it would still return `Ior[NonEmptyList[B], NonEmptyList[C]]`

johnynek:
You don't need traverse do you? Reducible will do, no?

LukaJCB:
@johnynek is correct! I was able to generalize it to Reducible :)

kailuowang:
looks good. is it just me or does anyone else feel a bit weird that we have a `partitionE` on `Reducible` but not a `partition` on `Foldable` (on which it would probably be returning `(List[A], List[B])` 

like how about this 
```scala
  def partitionEither[A, B, C](fa: F[A])(f: A => Either[B, C])(implicit F: MonoidK[F], A: Applicative[F]): (F[B], F[C]) =
    foldRight(fa, Eval.later((F.empty[B], F.empty[C]))) { (a, e) =>
      e.map {
        case (fbs, fcs) => f(a).fold(
          b => (F.combineK(A.pure(b), fbs), fcs),
          c => (fbs, F.combineK(A.pure(c), fcs))
        )
      }
    }.value
}
```
Note that it requires `Applicative` and `MonoidK` on `F`, maybe we can reduce that requirements?

LukaJCB:
Pretty sure, we could reduce the requirements by returning `(List[A], List[B])` instead of `(F[B], F[C])`

LukaJCB:
I don't think we can reduce the requirements, because either side of the tuple might be empty, so MonoidK is a must, and for Applicative to go, we'd need a way to go from `Either[B,C]` to `(F[B], F[C])`, so either we'd keep those two, or we'd just return `(List[A], List[B])` as you said.

kailuowang:
I am fine with keeping those requirements. My guess is that for most of the use cases they are going to have those two instances anyway.

peterneyens:
That `Foldable#partitionEither` would be the same as a `map` followed by `Alternative#separate`, right?
Maybe we can call it `Alternative#mapSeparate` ?

LukaJCB:
That's a good idea! It also just dawned on me that `MonoidK` + `Applicative` is `Alternative` 🙈. I think the implementation could be even easier:

```scala
def partitionEither[A, B, C](fa: F[A])(f: A => Either[B, C])(implicit A: Alternative[F]): (F[B], F[C]) =
  foldMap(fa)(a => f(a) match {
    case Right(c) => (A.empty[B], A.pure(c))
    case Left(b) => (A.pure(b), A.empty[C])
  })
```

kailuowang:
Right I keep forgetting `Alternative`

LukaJCB:
It seems the above didn't work, because there's no `Monoid` instance only `MonoidK`. It'd need to be `foldMapK`.

kailuowang:
can `MonoidK#algebra` help?

LukaJCB:
`algebra` works like a charm, thanks! Should I add this to the current PR, or make a new one? :)

kailuowang:
current PR works for me.

core/src/main/scala/cats/Alternative.scala
```diff
@@ -22,6 +22,18 @@ import simulacrum.typeclass
     (as, bs)
   }
 
+  def mapSeparate[A, B, C](fa: F[A])(f: A => Either[B, C])(implicit F: Foldable[F], A: Alternative[F]): (F[B], F[C]) = {
+    import cats.instances.tuple._
+
+    implicit val mb: Monoid[F[B]] = Alternative[F].algebra[B]
```
kailuowang: nit: could be `A.algebra` here?

core/src/main/scala/cats/Alternative.scala
```diff
@@ -22,6 +22,18 @@ import simulacrum.typeclass
...
+    implicit val mc: Monoid[F[C]] = Alternative[F].algebra[C]
+
+    F.foldMap(fa)(a => f(a) match {
+      case Right(c) => (empty[B], pure(c))
+      case Left(b) => (pure(b), empty[C])
+    })
+  }
+
```
kailuowang: I think @peterneyens was suggesting that 
you can just write 
```scala
  import cats.instances.either._
  def mapSeparate[A, B, C](fa: F[A])(f: A => Either[B, C])(implicit FM: Monad[F]): (F[B], F[C]) =
    separate(map(fa)(f))
```
So instead of requiring `F` to be `Foldable`, it requires to be a `Monad`. 
Or we can use your implementation, which requires `Foldable`, which might be more performant? (traverse once instead three times?)

I still think we should name it `partitionEither` or something, the idea is to provide a familiar api to users. `map` and then `separate` probably don't justify a special delegate method also named `mapSeparate`.

LukaJCB: I think I prefer the `Foldable` way, for the reasons you mentioned. I'm not married to the name, but it makes sense in the same way `flatMap` makes sense, I guess. 

I think I'd call the method on `Foldable` `partitionEither` as you suggested and renamed the one on `Reducible` to `partitionIor` Maybe? 
`partitionE` isn't a great name IMO, because there's no real precedence for the "E" and I think if we want to stay a beginner friendly library we should have as little of these one-letter-abbreviations as possible.

peterneyens: The current implementation is better than `map` + `separate` indeed. Just wanted to point out the similarity.

`separate` is available for all `Bifoldable`s while we currently have to use `Either` with `mapSeparate`:
- With `separate` you could also split `Ior` and `Tuple2` (similar to `unzip`).
-  `mapSeparate` + `toEither` will probably be faster than `map(toEither).separate` as you fold once instead of twice.

I agree that `partitionE` isn't a great name.

core/src/main/scala/cats/Reducible.scala
```diff
@@ -177,6 +176,34 @@ import simulacrum.typeclass
...
+    * res0: cats.data.Ior[cats.data.NonEmptyList[String],cats.data.NonEmptyList[Int]] = Both(NonEmptyList(2, 4),NonEmptyList(1, 3))
+    * scala> Reducible[NonEmptyList].partitionE(nel)(a => Right(a * 4))
+    * res1: cats.data.Ior[cats.data.NonEmptyList[Nothing],cats.data.NonEmptyList[Int]] = Right(NonEmptyList(4, 8, 12, 16))
+    * }}}
+    */
+  def partitionE[A, B, C](fa: F[A])(f: A => Either[B, C]): Ior[NonEmptyList[B], NonEmptyList[C]] = {
+    import cats.syntax.either._
+
+    def g(a: A, eval: Eval[Ior[NonEmptyList[B], NonEmptyList[C]]]): Eval[Ior[NonEmptyList[B], NonEmptyList[C]]] = {
+      val ior = eval.value
```
peterneyens: We should probably do: 
```scala
eval.map { ior =>
  (f(a), ior) match {
    // ...
  }
}
```

core/src/main/scala/cats/Alternative.scala
kailuowang: my 2 cents 
1. since we are going to use `Foldable`, it seems to make more sense to be on `Foldable` than `Alternative`.  Having a pair of partition methods on `Foldable` and `Reducible` is what I was looking for. 
1.  both methods should be named `partitionEitherXX` since they both use a function that returns an Either for the actual partition logic. Maybe we can call `Reducible#partitionEitherNE` and `Foldable#partitionEither`
1. maybe we should mention the `map` + `separate` option in the scaladoc in `Foldable#partitionEither`, I think it might worth some promotion.

LukaJCB: I'm +1 on renaming to `Foldable#partitionEither` and mentioning the `map` + `separate` option in the scaladoc. Disagree on `partitionEitherNE` though, for similar reasons as `partitionE`. 
How about `nonEmptyPartition`? Would go in line with `nonEmptyIntercalate` etc. 🤔

LukaJCB:
Thanks for your review @peterneyens! I addressed your comments :)

core/src/main/scala/cats/Alternative.scala
kailuowang: I am fine with nonEmptyPartition

LukaJCB: Renamed to `Foldable#partitionEither` and `Reducible#nonEmptyPartition` :)

tests/src/test/scala/cats/tests/FoldableTests.scala
```diff
@@ -25,6 +25,40 @@ abstract class FoldableCheck[F[_]: Foldable](name: String)(implicit ArbFInt: Arb
     }
   }
 
+  test("Alternative#partitionEither retains size") {
```
kailuowang: titles of the tests should be `Foldable#partitionEither xxxx`

core/src/main/scala/cats/Foldable.scala
```diff
@@ -407,6 +407,21 @@ import simulacrum.typeclass
     }.toList
 
   /**
+    * Separate this Foldable into a Tuple by a separating function `A => Either[B, C]`
+    */
+  def partitionEither[A, B, C](fa: F[A])(f: A => Either[B, C])(implicit F: Foldable[F], A: Alternative[F]): (F[B], F[C]) = {
```
kailuowang: you are already in `Foldable`, so no need for the `F: Foldable[F]` here

tests/src/test/scala/cats/tests/FoldableTests.scala
```diff
@@ -25,6 +25,40 @@ abstract class FoldableCheck[F[_]: Foldable](name: String)(implicit ArbFInt: Arb
...
+
+      val withG = Foldable[List].partitionEither(list)(g)._2
+      withG should === (list.map(f))
+
+      val withH = Foldable[List].partitionEither(list)(h)._1
+      withH should === (list.map(f))
+    }
+  }
+
+  test("Foldable#mapSeparate remains sorted") {
```
peterneyens: You forgot to rename it here.

core/src/main/scala/cats/Foldable.scala
```diff
@@ -407,6 +407,31 @@ import simulacrum.typeclass
     }.toList
 
   /**
+    * Separate this Foldable into a Tuple by a separating function `A => Either[B, C]`
+    * Equivalent to `Functor#map` and then `Alternative#separate`.
```
peterneyens: If we did define it in `Alternative`, we could add a consistency law checking this equality, but hey ... you can't always get what you want :smiley:

LukaJCB: True, I don't really mind either way tbh.
We could add a test that checks if it's consistent with List#partition though 😄

fommil:
@LukaJCB since I wrote this, I'd appreciate a mention in the method :wink: recall that my original implementation carried no license with it.

fommil:
hmm, on second thoughts, your impl is sufficiently different that I don't consider it a derivative works, no attribution required.

LukaJCB:
Hey @fommil sorry for any confusion, didn't mean any harm!

fommil:
@LukaJCB none at all taken :smile: I'm glad to see that somebody else is interested in the functionality. The ticket kinda died so I lost all motivation.

kailuowang:
@peterneyens anything other feedback?

core/src/main/scala/cats/Reducible.scala
```diff
@@ -177,6 +176,34 @@ import simulacrum.typeclass
...
+    * {{{
+    * scala> import cats.data.NonEmptyList
+    * scala> val nel = NonEmptyList.of(1,2,3,4)
+    * scala> Reducible[NonEmptyList].nonEmptyPartition(nel)(a => if (a % 2 == 0) Left(a.toString) else Right(a))
+    * res0: cats.data.Ior[cats.data.NonEmptyList[String],cats.data.NonEmptyList[Int]] = Both(NonEmptyList(2, 4),NonEmptyList(1, 3))
+    * scala> Reducible[NonEmptyList].nonEmptyPartition(nel)(a => Right(a * 4))
+    * res1: cats.data.Ior[cats.data.NonEmptyList[Nothing],cats.data.NonEmptyList[Int]] = Right(NonEmptyList(4, 8, 12, 16))
+    * }}}
+    */
+  def nonEmptyPartition[A, B, C](fa: F[A])(f: A => Either[B, C]): Ior[NonEmptyList[B], NonEmptyList[C]] = {
```
johnynek: couldn't we do this considerably more efficiently for NonEmptyList and NonEmptyVector? Can we add methods to those datatypes that does so? Eval is kind of a perf killer, sadly, which this method hides.

LukaJCB: I just had the same idea :D I already implemented a O(n) solution for `partitionEither` on `List` and will override nonemptyPartition as well!

LukaJCB: Added overrides for NEL and NEV :)

tests/src/test/scala/cats/tests/ReducibleTests.scala
```diff
@@ -95,4 +95,37 @@ abstract class ReducibleCheck[F[_]: Reducible](name: String)(implicit ArbFInt: A
...
+
+      val withG = fi.nonEmptyPartition(g).right.getOrElse(NonEmptyList.one(""))
+      withG should === (Reducible[F].toNonEmptyList(fi).map(f))
+
+      val withH = fi.nonEmptyPartition(h).left.getOrElse(NonEmptyList.one(""))
+      withH should === (Reducible[F].toNonEmptyList(fi).map(f))
+    }
+  }
+
+  test("Reducible#nonEmptyPartition remains sorted") {
```
johnynek: this is a bit weird. The only `F[_]` part of it `toNonEmptyList`. After that, we are testing `nonEmptyPartition` on NonEmptyList, no?

LukaJCB: Yes that's true, I wrote this test when originally `nonEmptyPartition` was part of `NonEmptyList`. I could move the test there, if you'd like :) Otherwise, I'm not sure it's possible to sort an `F[_]` with just `Reducible`

johnynek:
👍

kailuowang:
Great contribution. Thanks @LukaJCB !

PR #1867, opened 2017-08-27
Remove unused parameters from methods

BennyHill:
see also https://github.com/typelevel/cats/pull/1632#pullrequestreview-34456144
https://github.com/scala/scala/pull/5402#issuecomment-295261637
scala/scala#5876

See https://github.com/typelevel/cats/pull/1669 for related work on `MonadError` and `ApplicativeError`

### DIFF
--- a/core/src/main/scala/cats/syntax/either.scala
+++ b/core/src/main/scala/cats/syntax/either.scala
@@ -24,7 +24,7 @@ object EitherSyntax {
    * Uses the [[http://typelevel.org/cats/guidelines.html#partially-applied-type-params Partially Applied Type Params technique]] for ergonomics.
    */
   private[syntax] final class CatchOnlyPartiallyApplied[T](val dummy: Boolean = true ) extends AnyVal {
-    def apply[A](f: => A)(implicit CT: ClassTag[T], NT: NotNull[T]): Either[T, A] =
+    def apply[A](f: => A)(implicit CT: ClassTag[T]): Either[T, A] =
       try {
         Right(f)
       } catch {
--- a/core/src/main/scala/cats/syntax/partialOrder.scala
+++ b/core/src/main/scala/cats/syntax/partialOrder.scala
@@ -8,7 +8,7 @@ trait PartialOrderSyntax extends EqSyntax {
     new PartialOrderOps[A](a)
 }
 
-final class PartialOrderOps[A](lhs: A)(implicit A: PartialOrder[A]) {
+final class PartialOrderOps[A](lhs: A) {
   def >(rhs: A): Boolean = macro Ops.binop[A, Boolean]
   def >=(rhs: A): Boolean = macro Ops.binop[A, Boolean]
   def <(rhs: A): Boolean = macro Ops.binop[A, Boolean]
--- a/core/src/main/scala/cats/data/Validated.scala
+++ b/core/src/main/scala/cats/data/Validated.scala
@@ -278,7 +278,7 @@ object Validated extends ValidatedInstances with ValidatedFunctions{
    * Uses the [[http://typelevel.org/cats/guidelines.html#partially-applied-type-params Partially Applied Type Params technique]] for ergonomics.
    */
   private[data] final class CatchOnlyPartiallyApplied[T](val dummy: Boolean = true ) extends AnyVal{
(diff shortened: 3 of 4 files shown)


BennyHill:
The failing coverage test is : 

```scala

private[this] val singleton: NotNull[Any] = new NotNull[Any] {}
implicit def catsNotNullForA[A]: NotNull[A] = singleton.asInstanceOf[NotNull[A]]
```

So, iff it's OK to change the API in this PR, the next question is whether to remove these two lines or add tests for them

core/src/main/scala/cats/syntax/either.scala
```diff
@@ -24,7 +24,7 @@ object EitherSyntax {
    * Uses the [[http://typelevel.org/cats/guidelines.html#partially-applied-type-params Partially Applied Type Params technique]] for ergonomics.
    */
   private[syntax] final class CatchOnlyPartiallyApplied[T](val dummy: Boolean = true ) extends AnyVal {
-    def apply[A](f: => A)(implicit CT: ClassTag[T], NT: NotNull[T]): Either[T, A] =
+    def apply[A](f: => A)(implicit CT: ClassTag[T]): Either[T, A] =
```
ceedubs: While the `NT` parameter isn't referenced at runtime, it serves an important compile-time role. Consider that you wanted to do something like the following:

```scala
scala> Either.catchOnly[NumberFormatException]("foo".toInt)
res2: Either[NumberFormatException,Int] = Left(java.lang.NumberFormatException: For input string: "foo")
```

If you forget to add the type parameter before this PR you get a compile-time error:

```scala
scala> Either.catchOnly("foo".toInt)
<console>:18: error: ambiguous implicit values:
 both method If you are seeing this, you probably need to add an explicit type parameter somewhere, because Null is being inferred. in object NotNull of type => cats.NotNull[Null]
 and method catsAmbiguousNotNullNull2 in object NotNull of type => cats.NotNull[Null]
 match expected type cats.NotNull[Null]
       Either.catchOnly("foo".toInt)
```

However, with this PR you get a runtime error because `Null` is inferred as the type parameter:

```scala
scala> Either.catchOnly("foo".toInt)
java.lang.NumberFormatException: For input string: "foo"
  at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
  at java.lang.Integer.parseInt(Integer.java:580)
  at java.lang.Integer.parseInt(Integer.java:615)
  at scala.collection.immutable.StringLike.toInt(StringLike.scala:301)
  at scala.collection.immutable.StringLike.toInt$(StringLike.scala:301)
  at scala.collection.immutable.StringOps.toInt(StringOps.scala:29)
  at .$anonfun$res0$1(<console>:15)
  at scala.runtime.java8.JFunction0$mcI$sp.apply(JFunction0$mcI$sp.java:12)
  at cats.syntax.EitherSyntax$CatchOnlyPartiallyApplied$.apply$extension(either.scala:29)
  ... 39 elided
```


I think that the correct change to make here is to document why this `NotNull` parameter exists. The same goes for the `Validated` case.

core/src/main/scala/cats/syntax/partialOrder.scala
```diff
@@ -8,7 +8,7 @@ trait PartialOrderSyntax extends EqSyntax {
     new PartialOrderOps[A](a)
 }
 
-final class PartialOrderOps[A](lhs: A)(implicit A: PartialOrder[A]) {
+final class PartialOrderOps[A](lhs: A) {
```
ceedubs: Can someone who knows machinist better than I do weigh in on whether this change is fine (I suspect so, but I want to make sure). cc @non @johnynek

johnynek: I have no idea about the macro expansion in machinist.

I am not sure that is what is going on here anyway. I think it may be to prevent this class from being allocated without a PartialOrder.

We could I'm not sure if this is even tested sadly.

diesalbla: This seems no longer relevant, since macros have been now removed in preparation for Dotty.

larsrh:
Closing this because the two proposed changes are not desired or now outdated.

PR #1873, opened 2017-08-29
Add pureconfig and finch to ecosystem project list

kailuowang:
approval acquired at https://github.com/pureconfig/pureconfig/issues/298 and https://github.com/finagle/finch/issues/818

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,9 @@ language", and integrate with each other with ease.
  * [decline](https://github.com/bkirwi/decline): A composable command-line parser
  * [seals](https://github.com/durban/seals): tools for schema evolution and language-integrated schemata
  * [grafter](https://github.com/zalando/grafter): dependency-injection library using the `Reader` pattern
+ * [finch](https://github.com/finagle/finch): Scala combinator library for building Finagle HTTP services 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -110,6 +110,9 @@ language", and integrate with each other with ease.
  * [decline](https://github.com/bkirwi/decline): A composable command-line parser
  * [seals](https://github.com/durban/seals): tools for schema evolution and language-integrated schemata
  * [grafter](https://github.com/zalando/grafter): dependency-injection library using the `Reader` pattern
+ * [finch](https://github.com/finagle/finch): Scala combinator library for building Finagle HTTP services 
+ * [pureconfig](https://github.com/pureconfig/pureconfig): A boilerplate-free library for loading configuration files
```
peterneyens: Do we need to make a difference between libraries which use cats and those which provide an `-cats` extension like pureconfig and frameless (and refined, typelevel/paiges, ...) ?

kailuowang: Adding a "through `-cats extension`" note might be useful. Just wasn't sure how important it is to justify the time (also those may subject to change).

kailuowang: either way can we address those in a different PR?

PR #1875, opened 2017-08-30
Fix typo in SemigroupK scaladoc

LukaJCB:


kailuowang:
not a native speaker, but according to dictionary, both valid spelling of the word?

johnynek:
seems both are legit, but this change moves to the more common usage on mathscinet (but only a 2 to 1 ratio):

https://english.stackexchange.com/questions/79847/parametrise-or-parameterise-a-curve

fine with me to make the change.

👍

fthomas:
If we do this, we should change all occurrences of "parameterized":
https://github.com/typelevel/cats/search?utf8=%E2%9C%93&q=parameterized

LukaJCB:
Huh, didn't know both versions are valid, TIL 😄 
I'll go ahead and change the other occurrences

johnynek:
👍

PR #1878, opened 2017-08-31
Add some instances we were missing.

### LINKED ISSUE #1853: Contravariant functor instance for scala.math.Ordering
There doesn't seem to be a `Contravariant` instance for `scala.math.Ordering`.

non:
This commit adds support for some types:

 - Ordering, PartialOrdering, and Equiv
 - scala.concurrent.duration.Duration
 - scala.collection.immutable.Queue

The ordering types should be relatively uncontroversial.

I chose to support Duration even though arguably FiniteDuration is a
bit better behaved. This isn't the worst group we support, and I feel
like it's possible that folks will really benefit from these instances.

Queue is probably the most interesting immutable data structure we
didn't already support. I based its instances on those for List,
rewriting things to use .isEmpty and .dequeue instead of
pattern-matching.

We could potentially try to support some of the other concrete
collection types (e.g. HashMap, TreeMap, IntMap, SortedMap, etc.)
which might not be a bad idea. I haven't done that here.

I added tests for Queue and Duration, but have not (yet) tested the
pretty trivial contravariant instances.

Fixes #1853.


non:
Looks like this is failing a binary compatibility check.

kailuowang:
There are a few other tickets blocked by kernel bin compat, (#1712, #1527). @denisrosset suggested in [gitter](https://gitter.im/non/algebra?at=5976eae3c101bc4e3aba5e99) that it's no big deal for the spire community to break in the 1.0 release, I just asked in the algebird gitter room. @johnynek suggested that he'd rather not break. Is that something we can reevaluate? If not what's the alternative? (move all these to cats.core or a new module?)

non:
@kailuowang I think if we leave `DurationInstances` and `QueueInstances` out of the kernel's `all.scala` that should fix the kernel issues, no? I'll try that here.

non:
So with that commit we maintain binary compatibility for kernel.

kailuowang:
so shall we have a `cats.instances.extra` to place all the new instances helpers or something?

non:
@kailuowang Let's wait and see how many there are. For now I think we should just add them to `all.scala` commented out, and then make a decision just before the 1.0 release.

(Looks like last error was formatting, pushing a fix now.)

non:
In cases where we are adding instances for a new type, the approach I used should be good I think. For adding instances to an existing type, let's just put them in `core` with a comment that they could be in kernel except for bincompat.

What do you think @johnynek?

kailuowang:
@non 👍  on deciding later.  I just created a dedicated issue for it.

core/src/main/scala/cats/instances/ordering.scala
```diff
@@ -0,0 +1,16 @@
...
+import cats.functor.Contravariant
+
+trait OrderingInstances {
+
+  implicit val catsFunctorContravariantForOrdering: Contravariant[Ordering] =
+    new Contravariant[Ordering] {
+      /** Derive an `Ordering` for `B` given an `Ordering[A]` and a function `B => A`.
+       *
+       * Note: resulting instances are law-abiding only when the functions used are injective (represent a one-to-one mapping)
+       */
```
kailuowang: On a related note (not suggesting we should change anything here), we do have an `Inject` with laws, user could use it to test if a function `f: A => B` is injective, by creating a `prj: B => Option[A]` and an `Inject[A, B]` using these two, then testing the law against this `Inject` instance. 
```scala
{
  implicit val inject = new Inject[A, B] { 
     val inj = f 
     val prj : B => Option[A] = ???
  }
 checkAll("Inject[A, B]", InjectTests[A, B].inject)
}
```
If you think this would be a good idea, we could potentially added to the `cats-testkit`.

core/src/main/scala/cats/instances/partialOrdering.scala
```diff
@@ -0,0 +1,19 @@
+package cats
+package instances
+
+import cats.functor.Contravariant
+
+trait PartialOrderingInstances {
+  implicit val catsFunctorContravariantForPartialOrdering: Contravariant[PartialOrdering] =
```
kailuowang: Note sure if this instance is tested.

core/src/main/scala/cats/instances/queue.scala
```diff
@@ -0,0 +1,145 @@
...
+
+import cats.syntax.show._
+
+import scala.annotation.tailrec
+import scala.collection.immutable.Queue
+import scala.util.Try
+
+trait QueueInstances extends cats.kernel.instances.QueueInstances {
+
+  implicit val catsStdInstancesForQueue: Traverse[Queue] with Alternative[Queue] with Monad[Queue] with CoflatMap[Queue] =
```
kailuowang: A bunch of these overrides are not tested. I am not sure how big a deal it is since they don't seem to be tested for the `list` instance either.

non:
@kailuowang I can probably get us test coverage on all of this if we're willing to wait until later today or tonight.

kernel-laws/src/test/scala/cats/kernel/laws/LawTests.scala
```diff
@@ -93,6 +136,7 @@ class LawTests extends FunSuite with Discipline {
   laws[OrderLaws, List[String]].check(_.order)
   laws[OrderLaws, Vector[Int]].check(_.order)
   laws[OrderLaws, Stream[Int]].check(_.order)
+  laws[OrderLaws, Queue[Int]].check(_.order)
```
kailuowang: I couldn't find `OrderLaws` test with `Duration`?

core/src/main/scala/cats/instances/ordering.scala
```diff
@@ -0,0 +1,16 @@
+package cats
+package instances
+
+import cats.functor.Contravariant
+
+trait OrderingInstances {
+
+  implicit val catsFunctorContravariantForOrdering: Contravariant[Ordering] =
+    new Contravariant[Ordering] {
```
johnynek: can't this be `ContravariantCartesian` as well? Just use left to right ordering?

core/src/main/scala/cats/instances/partialOrdering.scala
```diff
@@ -0,0 +1,19 @@
...
+    new Contravariant[PartialOrdering] {
```
johnynek: can't this be contravariantCartesian using left to right ordering?

non:
Added a bunch more tests, and ended up doing some refactors to other parts of the tests to remove some boilerplate.

It's sort of terrifying to realize that we're closing in on 1.0 and there's still so much potential polish we could do. 😛

core/src/main/scala/cats/instances/partialOrdering.scala
non: Yup. Initially I was uncertain about adding this but ended up going for it.

core/src/main/scala/cats/instances/ordering.scala
non: Yup. Initially I was uncertain about adding this but ended up going for it.

laws/src/main/scala/cats/laws/discipline/Arbitrary.scala
```diff
@@ -133,21 +133,34 @@ object arbitrary extends ArbitraryInstances0 {
...
+  implicit def catsLawsArbitraryForEquiv[A: Arbitrary]: Arbitrary[Equiv[A]] =
+    Arbitrary(getArbitrary[Eq[A]].map(Eq.catsKernelEquivForEq(_)))
 
   implicit def catsLawsArbitraryForPartialOrder[A: Arbitrary]: Arbitrary[PartialOrder[A]] =
-    Arbitrary(Gen.oneOf(
-      PartialOrder.from[A]((_: A, _: A) => Double.NaN),
-      PartialOrder.from[A]((_: A, _: A) => -1.0),
-      PartialOrder.from[A]((_: A, _: A) => 0.0),
-      PartialOrder.from[A]((_: A, _: A) => 1.0)))
+    Arbitrary(getArbitrary[Int => Double].map(f => new PartialOrder[A] {
```
kailuowang: this makes a lot more sense

kailuowang:
@non I wonder if you and/or @ceedubs have time to quickly skim through all the code in `cats-core` to spot all the obvious polish needed before 1.0 release?

non:
The previous test failure was legit: by testing some of the `Foldable` and `Traversable` methods against their references, I found two instances (`Map[K, ?]` and `NonEmptyVector`) which were treating negative indices incorrectly.

non:
I think this is finally ready-to-go test coverage-wise.

johnynek:
👍 

Thanks for tackling this before 1.0!

Issue #1272, opened 2016-08-06
Make cats-kernel law organization consistent with cats-core


adelbertc:

kailuowang:
see comment here https://github.com/typelevel/cats/issues/1341#issuecomment-243899002

LukaJCB:
I know, I just wrote the law testing guide that documented this inconsistency, but going forward I'd love to have these be consistent. So we are going to break bin-compat one last time before potentially a year or something, so now seems to be a good time, right?

kailuowang:
On the other hand it's less big a deal to break bin compat in the laws module.

LukaJCB:
If so, I'm going to try to make a PR early next week :)

johnynek:
I think keeping kernel *very* compatible is still important. Laws less so in my opinion. You don't transitively pull testing packages generally.

PR #1880, opened 2017-08-31
Add law testing guide

### LINKED ISSUE #1585: Document how to test typeclass instances with respective laws.
From [gitter discussion](https://gitter.im/typelevel/cats?at=58e14e54ad849bcf4244a81e)
We can add a section into the [typeclass page](https://github.com/typelevel/cats/blob/master/docs/src/main/tut/typeclasses/typeclasses.md). 

This would be
> A huge win for cats  - @djspiewak

LukaJCB:
Should resolve #1585. I'm thinking of creating another short section on why Laws are actually important, but for now this should be a good start I think.

Also I modified the build, so that `docs` depends on `testkit` and `laws`. Not sure, if this was the right way to do it, so feedback is very welcome :)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -181,7 +181,7 @@ lazy val docs = project
   .settings(noPublishSettings)
   .settings(docSettings)
   .settings(commonJvmSettings)
-  .dependsOn(coreJVM, freeJVM)
+  .dependsOn(coreJVM, freeJVM, kernelLawsJVM, lawsJVM, testkitJVM)
 
 lazy val cats = project.in(file("."))
   .settings(moduleName := "root")
--- a/build.sbt
+++ b/build.sbt
@@ -181,7 +181,6 @@ lazy val docs = project
   .settings(noPublishSettings)
   .settings(docSettings)
   .settings(commonJvmSettings)
-  .settings(libraryDependencies += "com.github.alexarchambault" %% "scalacheck-shapeless_1.13" % "1.1.5" % Compile)
   .dependsOn(coreJVM, freeJVM, kernelLawsJVM, lawsJVM, testkitJVM)
 
 lazy val cats = project.in(file("."))
--- a/build.sbt
+++ b/build.sbt
@@ -181,6 +181,7 @@ lazy val docs = project
   .settings(noPublishSettings)
   .settings(docSettings)
   .settings(commonJvmSettings)
+  .settings(libraryDependencies += "com.github.alexarchambault" %% "scalacheck-shapeless_1.13" % "1.1.5" % Compile)
   .dependsOn(coreJVM, freeJVM, kernelLawsJVM, lawsJVM, testkitJVM)
 
 lazy val cats = project.in(file("."))
--- a/docs/src/main/tut/typeclasses/lawtesting.md
+++ b/docs/src/main/tut/typeclasses/lawtesting.md
@@ -64,11 +64,23 @@ Before we do so, however,
 we will have to bring our instances into scope as well as the derived `Arbitrary` instances from `scalacheck-shapeless`:
 
 
-```tut:book
+```tut:silent
 import Tree._
-
+```
+```scala
 import org.scalacheck.Shapeless._
+```
 
+```tut:invisible
+import org.scalacheck.{Arbitrary, Gen}
+implicit def arbFoo[A: Arbitrary]: Arbitrary[Tree[A]] =
+  Arbitrary(Gen.oneOf(Gen.const(Leaf), (for {
+      e <- Arbitrary.arbitrary[A]
+    } yield Node(e, Leaf, Leaf).asInstanceOf[Tree[A]]))
+  )
+```
+
+```tut:book
 import cats.laws.discipline.FunctorTests
 
 class TreeLawTests extends CatsSuite {
--- a/docs/src/main/tut/typeclasses/lawtesting.md
+++ b/docs/src/main/tut/typeclasses/lawtesting.md
@@ -1,6 +1,6 @@
 # Law testing
 
-Laws are an important part of cats.
+[Laws](https://typelevel.org/cats/typeclasses.html#laws) are an important part of cats.
 They help us enforce constraints that we usually assume to be true and give us guarantees that our code does what we think it does.
 Manually testing each data type to adhere to each of its type class instances can be quite cumbersome.
 Thankfully cats comes with the `cats-testkit`, which makes it a lot easier to write tests for type class instances using laws.
--- a/docs/src/main/tut/typeclasses/lawtesting.md
+++ b/docs/src/main/tut/typeclasses/lawtesting.md
@@ -1,9 +1,8 @@
 # Law testing
 
 [Laws](https://typelevel.org/cats/typeclasses.html#laws) are an important part of cats.
-They help us enforce constraints that we usually assume to be true and give us guarantees that our code does what we think it does.
-Manually testing each data type to adhere to each of its type class instances can be quite cumbersome.
-Thankfully cats comes with the `cats-testkit`, which makes it a lot easier to write tests for type class instances using laws.
+Cats uses `catalysts` and `discipline` to help test instances with laws.
+To make things easier, cats ships with `cats-testkit`, which makes use of `catalysts` and `discipline` and exposes `CatsSuite` based on ScalaTest.
 
 
 ## Getting started
--- a/docs/src/main/tut/typeclasses/lawtesting.md
+++ b/docs/src/main/tut/typeclasses/lawtesting.md
@@ -1,0 +1,146 @@
+# Law testing
+
+Laws are an important part of cats.
+They help us enforce constraints that we usually assume to be true and give us guarantees that our code does what we think it does.
+Manually testing each data type to adhere to each of its type class instances can be quite cumbersome.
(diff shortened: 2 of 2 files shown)


docs/src/main/tut/typeclasses/lawtesting.md
```diff
@@ -0,0 +1,146 @@
+# Law testing
+
+Laws are an important part of cats.
```
kailuowang: Laws are explained [here](https://typelevel.org/cats/typeclasses.html), maybe make that section a linkable anchor and link from here?

LukaJCB: Markdown embeds `id`s so, we don't even need an anchor 🎉. Great suggestion, will do!

docs/src/main/tut/typeclasses/lawtesting.md
```diff
@@ -0,0 +1,146 @@
...
+They help us enforce constraints that we usually assume to be true and give us guarantees that our code does what we think it does.
+Manually testing each data type to adhere to each of its type class instances can be quite cumbersome.
+Thankfully cats comes with the `cats-testkit`, which makes it a lot easier to write tests for type class instances using laws.
```
kailuowang: well, strictly speaking, `catalysts` and `discipline` are the two libraries to make it easier to "write tests for type class instances using laws.", `cats-testkit` is just a tiny lib on top of them to make it slightly more convenient to test against cats type classes using scalatests.  relevant discussion https://gitter.im/typelevel/sbt-catalysts?at=59a85afdee5c9a4c5f1d45fb

LukaJCB: That's true, what do you think of rewording to 
"Thankfully cats comes with the `cats-testkit`, which is based on `catalysts` and `discipline`, both of which make it a lot easier to write law tests for type class instances."
? :)

docs/src/main/tut/typeclasses/lawtesting.md
```diff
@@ -0,0 +1,146 @@
...
+```tut:silent
+import Tree._
+```
+```scala
+import org.scalacheck.Shapeless._
+```
+
+```tut:invisible
+import org.scalacheck.{Arbitrary, Gen}
+implicit def arbFoo[A: Arbitrary]: Arbitrary[Tree[A]] =
```
kailuowang: so this is the one that is actually in use, not the one generated by scalacheck-shapeless?

LukaJCB: Yeah, I realize it's not an optimal solution, but it's the best I came up with. If you have a better idea, please do tell :)

kailuowang: i think the spirit of our documentation is that people can just copy code and it runs. This breaks that assumption. why scalacheck-shapeless didn't work?

LukaJCB: I wouldn't know how to include a foreign module for a tut doc 🤔. I guess it would be possible to add it as a dependency, but not sure if that's a good idea for just one part of the documentation?

docs/src/main/tut/typeclasses/lawtesting.md
kailuowang: I think, it's probably more accurate to break it into two sentences. 
catalysts and discipline are the two libraries to make it easier to "write tests for type class instances using laws. cats-testkit makes it more convenient to test against cats type classes using scalatests. paging @BennyHill for better wording.

docs/src/main/tut/typeclasses/lawtesting.md
```diff
@@ -0,0 +1,146 @@
+# Law testing
+
+[Laws](https://typelevel.org/cats/typeclasses.html#laws) are an important part of cats.
+They help us enforce constraints that we usually assume to be true and give us guarantees that our code does what we think it does.
+Manually testing each data type to adhere to each of its type class instances can be quite cumbersome.
```
kailuowang: hmm, I think the wording here could use some polish.  e.g. it's not testing data types, it's testing instances.  I'd say for the sake of conciseness and time, how about we just say 
```
[Laws](https://typelevel.org/cats/typeclasses.html#laws) are an important part of cats.
Cats uses catalysts and dicipline to help test instances with laws. 
```
then introduce cats-testkit

docs/src/main/tut/typeclasses/lawtesting.md
kailuowang: I am inclined to not pretend to use it in the documentation. Make the arbitrary instance visible and add a note that you can use scalacheck-shapeless to auto derive this instance. 
BTW, could we try avoid the `asInstance` in the arbitrary instance?

LukaJCB: I got it to work by adding this single line to `docs` in the build.sbt:

```
.settings(libraryDependencies += "com.github.alexarchambault" %% "scalacheck-shapeless_1.13" % "1.1.5" % Compile)
```

tut then compiles the whole thing without a hitch.

tpolecat:
Hey I just saw this and it looks great. I had to figure this out a few weeks ago and it was non-obvious so this will be very helpful for people. 👍

build.sbt
```diff
@@ -181,7 +181,8 @@ lazy val docs = project
   .settings(noPublishSettings)
   .settings(docSettings)
   .settings(commonJvmSettings)
-  .dependsOn(coreJVM, freeJVM)
+  .settings(libraryDependencies += "com.github.alexarchambault" %% "scalacheck-shapeless_1.13" % "1.1.5" % Compile)
```
BennyHill: I would strongly discourage adding a third party dependency unless it is _really_ essential. cats is quite low down the compile chain so the fewer dependencies it has that could get in the way of upgrades the better. And for thing like testing new scala compilers (eg 2.13.0.M1, native,etc) it is not the dependencies that a third party library brings into cats that is the issue, rather the dependencies needed to build that 3rd party library in the first place.

So I'm not saying it should not have any at all, but that we have to be careful.

BennyHill:
Sorry for the tardiness of the comment, my bad

adelbertc:
Semi-unrelated to this commit but is there a reason why our `kernel` tests are structured differently from `core` and friends? I remember having an issue ticket about this a while ago but can't seem to find it at the moment. cc @johnynek 

Other than that, guide looks good!

build.sbt
BennyHill: As a starting suggestion/discussion.....

The same is true for `testkit`, that I have suggested we split into two: a `testkit` that has no dependencies other than pure cats, and another `testkit-scalatest` that adds the scalatest dependency. The idea here is that we make the the base testkit independent to the testing frameworks. But what about specs2, should we also add a `testkit-specs2`, `testkit-utest` and so on? 

And what about, say, [cats-check](https://github.com/non/cats-check). For sure we _can't_ add that as a dependency.

So maybe now is a good time to address this. Maybe we need some kind of uber-cats repo that brings these things together?

BennyHill:
@adelbertc see https://github.com/typelevel/cats/issues/1272

docs/src/main/tut/typeclasses/lawtesting.md
```diff
@@ -0,0 +1,133 @@
...
+
+```tut:book
+import cats.laws.discipline.FunctorTests
+import cats.kernel.laws.GroupLaws
+
+class TreeLawTests extends CatsSuite {
+  checkAll("Tree[Int].MonoidLaws", GroupLaws[Tree[Int]].semigroup)
+  checkAll("Tree.FunctorLaws", FunctorTests[Tree].functor[Int, Int, String])
+}
+```
```
BennyHill: Note that with catalysts you can do something similar to how algebra/kernel calls the laws, which is something like:

```scala
...
import catalysts.macros.TypeTagM

// this can be added to the general test suite
implicit def groupLaws[A: Eq: Arbitrary]: GroupLaws[A] = GroupLaws[A]

....
laws[GroupLaws, Tree[Int]].check(_.semigroup)
```

build.sbt
LukaJCB: You raise a good point, I was hesitant to add the dependency before. I can revert the last commit and make the `Arbitrary` instance explicit with a mention to scalacheck-shapeless.

kailuowang: I agree we should be very cautious about adding dependencies to the project. In this particular case, it's a dependency added to the docs generation project, so no downstream impact. That's why I didn't object it. However, now I realized that it does impact the docs build in scala 2.13.

 "Making the `Arbitrary` instance explicit with a mention to scalacheck-shapeless" would make more sense to me.

docs/src/main/tut/typeclasses/lawtesting.md
kailuowang: I still prefer making this visible and mentioned the alternative solution from `scalacheck-shapeless` in a note.

docs/src/main/tut/typeclasses/lawtesting.md
LukaJCB: Do you think this is worth mentioning? I'm not really sure if I can see the benefit 🤔

ceedubs:
Do we need to add this to https://github.com/typelevel/cats/blob/master/docs/src/main/resources/microsite/data/menu.yml and/or anywhere else?

kailuowang:
We may also want to link it from FAQ.

Issue #1879, opened 2017-08-31
Decide on a solution for adding new features to cats-kernel avoiding breaking bin-compat.


kailuowang:
related #1878 #1712 #1527

kailuowang:
Also we'll probably quickly run into the same question IRT `cats-core` after 1.0.0. So hopefully the solution we come up with kernel will be a good reference.

kailuowang:
good point @johnynek , I am creating a PR to change existing "instance" trait to abstract class.

kailuowang:
removed from RC1 scope, since all breaking changes are okayed by @johnynek representing the Algebird community.

PR #1882, opened 2017-08-31
convert trait into abstract class for better bin compact

kailuowang:
related #1879


ceedubs:
Merging since this is a minor documentation fix.

ceedubs:
Oops this isn't the PR that I meant to merge 😬. I think that it's probably fine though. If anyone has any concerns, please let me know.

kailuowang:
I agree it should be fine.

PR #1884, opened 2017-09-01
Rename typeclass => type class in Readme.md

LukaJCB:
I think we have settled on "type class" being two words in all of our docs, so should be the same way here :)


kailuowang:
👍

PR #1885, opened 2017-09-01
Add zipWith to NonEmptyList and NonEmptyVector

### LINKED ISSUE #1870: Add `zipWith` to NonEmptyList
There's currently no way to zip two (or more) NELs, I think it is a very common use case and would be a worthwhile addition. We could even maybe generalize it to `Reducible`. Will have to look into that. What do you think?

LukaJCB:
Should resolve #1870

### DIFF
--- a/core/src/main/scala/cats/data/NonEmptyVector.scala
+++ b/core/src/main/scala/cats/data/NonEmptyVector.scala
@@ -184,6 +184,20 @@ final class NonEmptyVector[+A] private (val toVector: Vector[A]) extends AnyVal 
 
     NonEmptyVector(head, buf.result())
   }
+
+  /**
+    * Zips this `NonEmptyVector` with another `NonEmptyVector` and applies a function for each pair of elements.
+    *
+    * {{{
+    * scala> import cats.data.NonEmptyVector
+    * scala> val as = NonEmptyVector.of(1, 2, 3)
+    * scala> val bs = NonEmptyVector.of("A", "B", "C")
+    * scala> as.zipWith(bs)(_ + _)
+    * res0: cats.data.NonEmptyVector[String] = NonEmptyVector(1A, 2B, 3C)
+    * }}}
+    */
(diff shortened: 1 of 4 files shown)


core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -184,6 +184,20 @@ final class NonEmptyVector[+A] private (val toVector: Vector[A]) extends AnyVal
...
+    * {{{
+    * scala> import cats.data.NonEmptyVector
+    * scala> val as = NonEmptyVector.of(1, 2, 3)
+    * scala> val bs = NonEmptyVector.of("A", "B", "C")
+    * scala> as.zipWith(bs)(_ + _)
+    * res0: cats.data.NonEmptyVector[String] = NonEmptyVector(1A, 2B, 3C)
+    * }}}
+    */
+  def zipWith[B, C](b: NonEmptyVector[B])(f: (A, B) => C): NonEmptyVector[C] =
+    NonEmptyVector.fromVectorUnsafe((toVector, b.toVector).zipped.map(f))
```
johnynek: I think this will be maybe faster if we do:
```scala
toVector.iterator.zip(b.toVector.iterator).map(f).toVector
```
I think the current code has to allocate a NonEmptyVector after `zipped` and then another one after `map`.

But I'm just guessing since I don't know this tuple enrichment. Can you comment as to why this is incorrect if I am wrong?

LukaJCB: Yes of course! I'm on my phone right now, but if you look at the source for zipped https://github.com/scala/scala/tree/v2.12.3/src/library/scala/runtime/Tuple2Zipped.scala#L1 you'll see that it does not actually iterate over either of the vectors until you call map or any of the other methods :)

johnynek: ahh nice.

johnynek:
👍

PR #1888, opened 2017-09-02
 Add a stack safety stress test for Eval.

### LINKED ISSUE #1703: SOE in Eval.defer

non:
The basic idea is to create very deep chains of operations to try to expose any
stack problems. We know that for things like map/flatMap Eval seems to work, so
let's just randomly construct deeply-nested Eval expressions to see what
happens.

This test exposed a weakness with .memoize which is fixed.

As part of this commit, I noticed that our Arbitrary[Eval[A]] instances were
somewhat weak, so I upgraded them.

I was hoping to find something involving `Eval.defer` to fix #1703 which @mpilquist raised, but I couldn't find any weakness besides the things here involving `.memoize`.


core/src/main/scala/cats/Eval.scala
```diff
@@ -253,8 +253,15 @@ object Eval extends EvalInstances {
    * they will be automatically created when needed.
    */
   sealed abstract class Call[A](val thunk: () => Eval[A]) extends Eval[A] {
-    def memoize: Eval[A] = new Later(() => value)
-    def value: A = Call.loop(this).value
+
+    def memoize: Eval[A] =
+      new Call[A](thunk) {
+        override def memoize: Eval[A] = this
+        override lazy val value: A = Call.loop(this).value
```
johnynek: Won't this be lost if you flatMap? Isn't it ignore in that case. Is that okay?

non:
Maybe my change to use "real" `Arbitrary[Eval[A]]` instances is killing Travis? Not sure what's going on.

core/src/main/scala/cats/Eval.scala
```diff
@@ -294,12 +301,19 @@ object Eval extends EvalInstances {
...
     val run: Start => Eval[A]
 
-    def memoize: Eval[A] = Later(value)
+    def memoize: Eval[A] =
+      new Compute[A] {
+        type Start = self.Start
+        val start: () => Eval[Start] = self.start
+        val run: Start => Eval[A] = self.run
+        override def memoize: Eval[A] = this
+        override lazy val value: A = self.value
```
johnynek: Same, this lazy val is lost on flatMap no?
Maybe memoise only applies to an outer flatMap?

I have used it when forking into two paths so I don't evaluate something twice, but this wouldn't actually work with this change I think.

johnynek:
Can we have a test something like this:

```scala
forAll { (e: Eval[Int], fn: Int => Eval[Int]) =>
  var cnt = 0

  val action = e.flatMap { i = >
    cnt += 1
    fn(i)
  }.memoize

  val res = for {
    i1 <- action
    i2 <- action
  } yield i1 == i2

  assert(res.value == true)
  assert(cnt == 1) // memoize means we don't build what is up that tree more than once.
}
```

non:
@johnynek This is a great observation.

Much like the (now-removed) exception-handling, I think if we want to support this kind of intermediate memoization in the current approach, we have to sacrifice constant stack per intermediate memoization.

The alternative would be to use a more complicated structure instead of a queue of functions, and then to "update" the various nodes as their results are complete.

I'm happy to revert to the previous memoization implementation, especially since I don't think the code @mpilquist was worried about was using `memoize`.

johnynek:
So, the contract on `memoize` then is that it only applies to a final `value`? If that's true, why have it at all?

you can just do `lazy val myValue = eval.value` and get the result right? Or `Eval.lazy(eval.value)` if you want to return an Eval still. Seems like it is a bit dishonest if it can be discarded.

That said, I don't quite see why it is impossible, just that this implementation does not have it. For instance, what if we add a `case class Memoized[A](of: Eval[A]) extends Eval[A]` node and change the evaluation loop. Maybe that hits similar issues as raise did, but I can imagine if we had a side `mutable.Map[Eval[_], _]` where we cached the value, it might be okay, but making sure we update the map in a safe way might be a challenge

non:
@johnynek Your `Memoized` suggestion is what I mean -- we can totally do it but we'll need to complicate the *compute* logic a fair bit.

I wasn't disagreeing with you -- I actually think that we should not take the change I made, and should either stick with what we have for `memoize` or pursue a more radical alternative.

johnynek:
Here's what I mean @non 

https://github.com/typelevel/cats/commit/bce5de8e37c4cd0d23515781f73547d3a2d2e888

non:
@johnynek Nice!

I don't think you even need the `mutable.Map` right? Once you update `var result` in the continuation, that should be good enough, no? (Since unwinding these things is synchronous, there's no real opportunity for a race.)

johnynek:
indeed, that's true, but using the Map means if you do:
```scala
val foo: Eval[A] = getFoo
val foo1 = foo.memoize
val foo2 = foo.memoize
```
and they wind up in the same Eval DAG, then you can can still only evaluate once.

Maybe that is such a corner case it isn't worth it, but it is pretty cheap to have the map (although maybe not worth it since you have to do the map updates for everything).

Would you be open to adding something like this to your change?

non:
Yes! In fact, I"m already integrating it (without the `Map` for now). There was on bug with your thing that I think I've fixed. I'm cleaning it up now.

johnynek:
this looks great to me! Thanks for tackling this. Seems like a strictly better situation now (assuming perf didn't tank).

johnynek:
minor suggestion (also made offline): what about renaming `Call` => `Defer` and `Compute` => `FlatMap` which will make it easier to read this code and recall what each are for.

core/src/main/scala/cats/Eval.scala
```diff
@@ -246,56 +246,60 @@ object Eval extends EvalInstances {
...
-      case compute: Eval.Compute[A] =>
-        new Eval.Compute[A] {
+      case call: Eval.Defer[A] =>
+        advance(call.thunk())
+      case compute: Eval.FlatMap[A] =>
+        new Eval.FlatMap[A] {
           type Start = compute.Start
           val start: () => Eval[Start] = () => compute.start()
-          val run: Start => Eval[A] = s => doCall1(compute.run(s))
+          val run: Start => Eval[A] = s => advance1(compute.run(s))
```
johnynek: Why do we need this here? Seems like it will be run later without beating the function call here.

johnynek:
👍 

@kailuowang can you also review?

core/src/main/scala/cats/Eval.scala
```diff
@@ -246,83 +246,130 @@ object Eval extends EvalInstances {
...
+   * Advance until we find a non-deferred Eval node.
+   *
+   * Often we may have deep chains of Defer nodes; the goal here is to
+   * advance through those to find the underlying "work" (in the case
+   * of FlatMap nodes) or "value" (in the case of Now, Later, or
+   * Always nodes).
+   */
+  @tailrec private def advance[A](fa: Eval[A]): Eval[A] =
+    fa match {
+      case call: Eval.Defer[A] =>
```
kailuowang: totally nitpick: `call` can be renamed to `defer` and the `compute` below. No need to address in this PR, I can do in a separate one.

core/src/main/scala/cats/Eval.scala
```diff
@@ -246,83 +246,130 @@ object Eval extends EvalInstances {
...
+            case xx =>
+              loop(c.run(xx.value), fs)
+          }
+        case call: Defer[_] =>
+          loop(advance(call), fs)
+        case m@Memoize(eval) =>
+          m.result match {
+            case Some(a) =>
+              fs match {
+                case f :: fs => loop(f(a), fs)
```
kailuowang: This line isn't tested. I am curious how come the random stack safety stress test didn't hit it, is it expected?

kailuowang:
not sure if @non has time to work on this PR. I am okay with merging it now and if necessary address the two minor issues later. WDYT @johnynek

johnynek:
sounds good.

I'll merge. I think we are in a good shape.

Issue #1818, opened 2017-08-11
Add a ValidatedT to give ApplicativeError to any Applicative.


johnynek:
currently it only has MonadError when `F[_]: Monad` but I think you can have ApplicativeError when `F[_]: Applicative`.

peterneyens:
I think this again is because the `ap` defined using `Applicative[F]` is inconsistent with the one using `Monad[F]`. That is the reason why there is no `Apply` and `Applicative` for `EitherT`, which implies that `ApplicativeError` would suffer from the same problem, right? 
See https://github.com/typelevel/cats/issues/1467#issuecomment-305853570 and the linked issues.

LukaJCB:
I think for the same reason Haskell also only has
`Monad m => Applicative (EitherT e m)`

johnynek:
I didn't know it was inconsistent. It's not obvious to me why we can't make it consistent, but no huge deal. We can convert to ValidatedT.

johnynek:
actually, there is no ValidatedT.

edmundnoble:
There is no `ValidatedT`? How so? Applicatives compose. `ValidatedT[F, E, ?]` would just be a wrapper on top of `Nested[F, Validated[E, ?], ?]`, I believe.

johnynek:
The point of `ValidatedT`, I think, is to get ApplicativeError from Applicative. You can't do that currently with Nested.

Maybe we want ApplicativeError on `Nested[F, G, ?]` if you have `F[_]: Applicative` and `G[_]: ApplicativeError`. Seems like you could also do the opposite perhaps: `F[_]: ApplicativeError` and `G[_]: Applicative` If true, it is not 100% clear which should be preferred.

I think an explicit ValidatedT offers some usability wins when the point is to add error handling to an Applicative.

adelbertc:
I think there is no ValidatedT because that name invokes the idea of a monad transformer, which Validated cannot have. Of course as Edmund mentions you can have other effects composed with Validated through Nested.

Assuming the definition is general I could see such a Nested type class instance for ApplicativeError being useful. I wouldn't want to see a ValidatedT though since it would just be a special case of Nested and it would seem like we're privileging Validated. Maybe a type alias though if this is a common enough thing people want to do.

johnynek:
There are a lot of utility methods, similar to what we have in EitherT that you would want. Like mapInvalid, isValid, etc... just using Nested means we don't have those.

edmundnoble:
[...]

PR #1893, opened 2017-09-07
Add ApplicativeError for Nested

LukaJCB:
Adds an instance for `ApplicativeError[Nested[F, G, ?], E]` given `ApplicativeError[F, E]` and `Applicative[G]` as discussed in #1818.

### DIFF
--- a/core/src/main/scala/cats/data/Nested.scala
+++ b/core/src/main/scala/cats/data/Nested.scala
@@ -81,6 +81,16 @@ private[data] sealed abstract class NestedInstances3 extends NestedInstances4 {
 }
 
 private[data] sealed abstract class NestedInstances4 extends NestedInstances5 {
+  implicit def catsDataApplicativeErrorForNested[F[_]: ApplicativeError[?[_], E], G[_]: Applicative, E]: ApplicativeError[Nested[F, G, ?], E] =
+    new NestedApplicativeError[F, G, E] {
+      val G: Applicative[G] = Applicative[G]
+
+      val AEF: ApplicativeError[F, E] = ApplicativeError[F, E]
+    }
+
+}
+
+private[data] sealed abstract class NestedInstances5 extends NestedInstances6 {
   implicit def catsDataApplicativeForNested[F[_]: Applicative, G[_]: Applicative]: Applicative[Nested[F, G, ?]] =
     new NestedApplicative[F, G] {
       val FG: Applicative[λ[α => F[G[α]]]] = Applicative[F].compose[G]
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/data/Nested.scala
```diff
@@ -81,6 +81,16 @@ private[data] sealed abstract class NestedInstances3 extends NestedInstances4 {
 }
 
 private[data] sealed abstract class NestedInstances4 extends NestedInstances5 {
+  implicit def catsDataApplicativeErrorForNested[F[_]: ApplicativeError[?[_], E], G[_]: Applicative, E]: ApplicativeError[Nested[F, G, ?], E] =
```
johnynek: couldn't you do it both ways? Put the `F[_]: ApplicativeError` or the `G[_]: ApplicativeError`?

Like: `Nested[Validated[String, ?], Option, ?]` or `Nested[Option, Validate[String, ?], ?]` You could have put the error on the inside or out, no?

johnynek: could we make two implementations at different priorities?

johnynek: I see your comment about reversing being hard unless you have `F[_]: Monad`. Indeed.

I wonder if it is worth adding that instance (`F[_]: Monad` and `G[_]: ApplicativeError`).

LukaJCB: Generally, I don't think we often have a constellation like that, but yeah for something like `Nested[Option, Validated[String, ?], ?]` it would work. However, I don't think it's enough to implement it on `Nested`. In my comment, I used `F[Validated[E, A]]` (aka `ValidatedT[F, E, A]`), but that has the extra property of being able to fold the `Validated` inside.

If all we have is `ApplicativeError`, we can't get there, because we don't have a function from `E => G[A]`, we only have `E => F[G[A]]`

```scala
// We only have this
def handleErrorWith(ga: G[A])(f: E => G[A]): G[A]

// but we need to implement this:
def handleErrorWith(fa: Nested[F, G, A])(f: E => Nested[F, G, A]): Nested[F, G, A]
```
We can't use the `f: E => Nested[F, G, A]` for the inner `ApplicativeError`, since we have no way to turn a `Nested[F, G, A]` into a `G[A]`

 The only way to do that, would be to give F and G `Traverse` instances or F an extra `Comonad` instance. Both of which would make this whole thing not really worth it IMO.
 
We can do it with `Validated`, because it allows us to `fold` it, but that's not true for the generic `ApplicativeError`.

Of course, I might be completely wrong, so point me in the right direction if I made a wrong assumption somewhere. 😄

johnynek:
👍

PR #1895, opened 2017-09-07
trying to fix contributing page

kailuowang:


kailuowang:
I will try republish the website shortly (we should be fine, no breaking change so far)

PR #1897, opened 2017-09-07
add bin compat goal

### LINKED ISSUE #1233: Document version compatibility goals
I think that we should document what our goals are with respect to binary/source compatibility as we reach 1.0.

I'd be inclined to propose the following:
- A change to only the patch version will always be binary compatible. For example `1.0.2` is binary compatible with `1.0.0` and `1.0.1`.
- A change to the minor version number may not be binary or source compatible. For example, `1.1.0` is not necessarily binary compatible with `1.0.0`.
- For incompatible changes introduced in a new version, we will try to make the transition as painless as possible. Before methods or types are removed, they will go through a deprecation cycle. If a change may silently lead to different behavior, we will try to highlight it in the release notes.
- The `cats-kernel` module will remain binary-compatible across minor versions. Binary-incompatible changes to this module should only occur alongside a major version bump (which will be few and far-between).
- We plan to support minor versions for at least ...some amount of time. Personally I'd be inclined to try to support minor versions for the greater of (2 years after their release, 1 year after the release of the next minor version). By support I mean backporting and publishing bugfixes and significant performance improvements.

What do people think about the ones I have? Are there others that we should add?

One thing that I don't really know how to quantify is source-compatibility. In general I want us to strive for source-compatible changes, but as long as they aren't too intrusive, I think it's less of a concern than binary compatibility.

cc @johnynek and @mpilquist who I think are both interested in compatibility.

kailuowang:
fixes #1233

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ into larger problems.
 If you are being harassed, please contact one of [us](#maintainers)
 immediately so that we can support you.
 
-### Binary compatibility goal
+### Binary compatibility and versioning
 
 After `1.0.0` release, we [decided](https://github.com/typelevel/cats/issues/1233) 
 to use *MAJOR.MINOR.PATCH* [Semantic Versioning 2.0.0](http://semver.org/)
--- a/README.md
+++ b/README.md
@@ -158,7 +158,7 @@ immediately so that we can support you.
 
 After `1.0.0` release, we [decided](https://github.com/typelevel/cats/issues/1233) 
 to use *MAJOR.MINOR.PATCH* [Semantic Versioning 2.0.0](http://semver.org/)
-going forward, which is different from the *EPIC.MAJOR.MINOR* schema commonly in
+going forward, which is different from the *EPOCH.MAJOR.MINOR* schema commonly in
 Java and Scala libraries (including the Scala lang). 
 
 Cats' strives to provide a solid and stable foundation for an ecosystem of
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -154,6 +154,33 @@ into larger problems.
 If you are being harassed, please contact one of [us](#maintainers)
 immediately so that we can support you.
 
+### Binary compatibility and versioning
+
+After `1.0.0` release, we [decided](https://github.com/typelevel/cats/issues/1233) 
+to use *MAJOR.MINOR.PATCH* [Semantic Versioning 2.0.0](http://semver.org/)
+going forward, which is different from the *EPIC.MAJOR.MINOR* schema commonly in
```
sjrd: EPIC -> EPOCH ;)

kailuowang: thanks! corrected.

README.md
```diff
@@ -154,6 +154,33 @@ into larger problems.
...
+application still runs thanks to this backward binary compatibility.  
+
+Also worth noting is that according to semantic versioning, 
+*MINOR* version Y (x.Y.z | x > 0) MUST be incremented 
+if new, backwards compatible functionality is introduced to the public API. 
+It MUST be incremented if any public API functionality is marked as deprecated.
+
+Any binary breaking changes will require a *MAJOR* version bump, which we will be very
+cautious about. We will also consider using `organization` and package name for major 
+versioning in the future. But that decision is yet to be made.
```
eed3si9n: 👍  Thanks for this inclusion.

PR #1898, opened 2017-09-07
Update various sbt plugins

fthomas:


kailuowang:
Merged with 1 sign off as simple dependency bump

PR #1899, opened 2017-09-07
Fix symbol signature for right and left apply in faq.md

suhasgaddam:
I think the right and left apply signatures should be switched, if I am reading https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/syntax/cartesian.scala#L20
 and https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/syntax/cartesian.scala#L23
 correctly.


kailuowang:
merge with 1 sign off as simple typo fix

PR #1900, opened 2017-09-08
add newts to related projects

kailuowang:


kailuowang:
merged with 1 sign off as a simple documentation update

PR #1902, opened 2017-09-09
Upgrade Scalafix to 0.5.0-RC2

gabro:
Since the last milestone there have been several breaking changes to the 0.5.0 API, which is now final!

Tests still pass ;)

/cc @olafurpg


gabro:
Thanks for the quick merge guys!

Issue #1678, opened 2017-05-18
Beginner-friendliness of Validated doc


ceedubs:
There has been some feedback on Gitter that the `Validated` doc isn't as beginner-friendly as it could be. Pretty early in the doc it delves into writing a `parallelValidate` method, discussing `Apply`, etc. This offers some nice insight into _deriving_ `Validated` methods, but it's not necessarily obvious that a user doesn't need to _write_ these things in their code. It would likely be better for users if the docs first showed simple examples of using `Validated` to accumulate errors and _then_ in an advanced section at the end covered more advanced concepts and derivation.

yilinwei:
Me and @zainab-ali have been discussing this problem since we've run several coding dojos/workshops over the past few months on typelevel libraries.

In my view, people who are interested in cats fall roughly into 3 categories

1. People who want to understand how cats works and why it works that way. These people have usually gone through the red book and gain most from a discussion of the datatypes, the typeclasses required for them to exist as well as a one liner about how you should use them.
2. People who just want to use cats to make there jobs a little easier. These people typically will only care about the bare minimum amount of syntax and types needed to achieve the task and a few nice examples of usages to make there lives easier (for validated it would be something like the `mapN` methods or something like `.traverse`).
3. People who are complete beginners to FP. These people require a build up from something a bit more basic (for example constraining the accumulated error to a `NonEmptyList[Throwable]`) and need some hand holding through some of the combinators that groups 1 & 2 would be familiar with such as `map`.

The documentation in cats can't cater to all 3 groups at once. It's never been clear to me who the docs are for.

zainab-ali:
IMO I don't think the cats docs should cater to complete beginners to FP.  The red book, already listed under [resources for learners](http://typelevel.org/cats/resources_for_learners.html), does a much better job of that.

I think it's possible to cater to people who want to use it, and also to people who want to understand how it works.  A good way to do this would be to:
1. Pose a common problem that `Validated` can solve.  The current docs do this using config parsing.
2. Show how `Validated` solves that problem without going into detail about how it works
3. Explain a little bit about how `Validated` works 

I'd go as far as saying that most data type docs should have this kind of format.

I don't think the docs need to go into detail about the implementations of typeclass instances.  A nod to the code should be enough for people who want to investigate those.

[...]

PR #1903, opened 2017-09-09
Validated beginners doc

AlejandroME:
This is an approach to address #1678.

Being my first PR, I pushed to my fork the first draft of this a week ago and @zainab-ali gladly helped me with a revision.
This PR contains fixes to the suggestions made in the revision, grammar fixes and the addition of `toValidated`/`toValidatedNel` combinators in the example.

I'm aware of the discussion made in the issue about the detail level of this doc. I tried to document this as clearly as possible without delving into too much detail.

### DIFF
--- a/docs/src/main/tut/datatypes/validated.md
+++ b/docs/src/main/tut/datatypes/validated.md
@@ -76,8 +76,8 @@ case object AgeIsInvalid extends DomainValidation {
 
 We have our `RegistrationData` case class that will hold the information the user has submitted, alongside the definition of the error model that we'll be using it for displaying the possible errors of every field. Now, let's explore the proposed implementation:
 
-```scala
-trait FormValidator{
+```tut:silent
+sealed trait FormValidator{
   private def validateUserName(userName: String): Either[DomainValidation, String] =
     if (userName.matches("^[a-zA-Z0-9]+$")) Right(userName) else Left(UsernameHasSpecialCharacters)
 
@@ -108,6 +108,8 @@ trait FormValidator{
 
 }
 
+object FormValidator extends FormValidator
+
 ```
 
 The logic of the validation process is as follows: **check every individual field based on the established rules for each one of them. If the validation is successful, then return the field wrapped in a `Right` instance; If not, then return a `DomainValidation` with the respective message, wrapped in a `Left` instance**.
@@ -118,7 +120,7 @@ Well, yes, but the error reporting part will have the downside of showing only t
 
 Let's look in detail this part:
 
-```scala
+```tut:silent:fail
     for {
       validatedUserName <- validateUserName(username)
       validatedPassword <- validatePassword(password)
@@ -134,12 +136,8 @@ A for-comprehension is _fail-fast_. If some of the evaluations in the `for` bloc
 
 If we run our code:
 
-```scala
-scala> object FormValidator extends FormValidator
-defined object FormValidator
-
-scala> FormValidator.validateForm("fakeUs3rname", "password", "John", "Doe", 15)
-res0: Either[DomainValidation,RegistrationData] = Left(PasswordDoesNotMeetCriteria)
+```tut:book
+FormValidator.validateForm("fakeUs3rname", "password", "John", "Doe", 15)
 ```
 
 We should have gotten another `DomainValidation` with the invalid age.
@@ -150,9 +148,10 @@ Time to do some refactoring! We're going to try a `Validated` approach:
 
 ```scala
 import cats.data._
+import cats.data.Validated._
 import cats.implicits._
 
-trait FormValidator{
+sealed trait FormValidatorValidated{
   private def validateUserName(userName: String): Validated[DomainValidation, String] =
     if (userName.matches("^[a-zA-Z0-9]+$")) Valid(userName) else Invalid(UsernameHasSpecialCharacters)
 
@@ -166,7 +165,7 @@ trait FormValidator{
   private def validateLastName(lastName: String): Validated[DomainValidation, String] =
     if (lastName.matches("^[a-zA-Z]+$")) Valid(lastName) else Invalid(LastNameHasSpecialCharacters)
 
-  private def validateAge(age: Int): Either[DomainValidation, Int] =
+  private def validateAge(age: Int): Validated[DomainValidation, Int] =
     if (age >= 18 && age <= 75) Valid(age) else Invalid(AgeIsInvalid)
 
   def validateForm(username: String, password: String, firstName: String, lastName: String, age: Int): Validated[DomainValidation, RegistrationData] = {
@@ -182,6 +181,8 @@ trait FormValidator{
   }
 
 }
+
+object FormValidatorValidated extends FormValidatorValidated
 ```
 
 Looks similar to the first version. What we've done here was to use `Validated` instead of `Either`. Please note that our `Right` is now a `Valid` and `Left` is an `Invalid`.
@@ -195,7 +196,7 @@ value flatMap is not a member of cats.data.Validated[DomainValidation,String]
 
 The problem lies here:
 
-```scala
+```tut:silent:fail
  for {
       validatedUserName <- validateUserName(username)
       validatedPassword <- validatePassword(password)
@@ -214,11 +215,12 @@ So, how do we do here?
 
 We have to look into another direction: a for-comprehension plays well in a fail-fast scenario, but the structure in our previous example was designed to catch one error at a time, so, our next step is to tweak the implementation a bit.
 
-```scala
+```tut:silent
 import cats.data._
+import cats.data.Validated._
 import cats.implicits._
 
-trait FormValidatorNel {
+sealed trait FormValidatorNel {
 
   type ValidationResult[A] = Validated[NonEmptyList[DomainValidation], A]
 
@@ -239,14 +241,16 @@ trait FormValidatorNel {
     if (age >= 18 && age <= 75) age.validNel else AgeIsInvalid.invalidNel
 
   def validateForm(username: String, password: String, firstName: String, lastName: String, age: Int): ValidationResult[RegistrationData] = {
-    (validateUserName(username) |@|
-    validatePassword(password) |@|
-    validateFirstName(firstName) |@|
-    validateLastName(lastName) |@|
-    validateAge(age)).map(RegistrationData)
+    (validateUserName(username),
+    validatePassword(password),
+    validateFirstName(firstName),
+    validateLastName(lastName),
+    validateAge(age)).mapN(RegistrationData)
   }
 
 }
+
+object FormValidatorNel extends FormValidatorNel
 ```
 
 Let's see what changed here:
@@ -254,21 +258,30 @@ Let's see what changed here:
 1. In this new implementation, we're using a `NonEmptyList`, a data structure that guarantees that at least one element will be present. In case that multiple errors arise, you'll get a list of `DomainValidation`.
 2. We've declared the type alias `ValidationResult` that conveniently express the return type of our validation.
 3. `.validNel` and `.invalidNel` combinators let us _lift_ the success or failure in their respective container (either a `Valid` or `Invalid[NonEmptyList[A]]`).
-4. The [applicative](../typeclasses/applicative.html) syntax `|@|` provides us a way to accumulatively apply the validation functions and yield a cartesian product with their successful result or the accumulated errors in the `NonEmptyList`.
+4. The [applicative](../typeclasses/applicative.html) syntax `(a, b, c, ...).mapN(...)` provides us a way to accumulatively apply the validation functions and yield a product with their successful result or the accumulated errors in the `NonEmptyList`. Then, we transform that product with `mapN` into a valid instance of `RegistrationData`.
+
+**Deprecation notice:** since cats `1.0.0-MF` the cartesian syntax `|@|` for applicatives is deprecated. If you're using `0.9.0` or less, you can use the syntax: `(a |@| b |@| ...).map(...)`.
 
 Note that, at the end, we expect to lift the result of the validation functions in a `RegistrationData` instance. If the process fails, we'll get our `NonEmptyList` detailing what went wrong.
 
 For example:
 
-```scala
-scala> object FormValidator extends FormValidatorNel
-defined object FormValidator
+```tut:book
+FormValidatorNel.validateForm(
+            username = "Joe",
+            password = "Passw0r$1234",
+            firstName = "John",
+            lastName = "Doe",
+            age = 21
+          )
 
-scala> FormValidator.validateForm("Joe", "Passw0r$1234", "John", "Doe", 21)
-res0: FormValidator.ValidationResult[RegistrationData] = Valid(RegistrationData(Joe,Passw0r$1234,John,Doe,21))
-
-scala> FormValidator.validateForm("Joe%%%", "password", "John", "Doe", 21)
-res1: FormValidator.ValidationResult[RegistrationData] = Invalid(NonEmptyList(UsernameHasSpecialCharacters, PasswordDoesNotMeetCriteria))
+FormValidatorNel.validateForm(
+            username = "Joe%%%",
+            password = "password",
+            firstName = "John",
+            lastName = "Doe",
+            age = 21
+          )
 ```
 
 Sweet success! Now you can take your validation process to the next level!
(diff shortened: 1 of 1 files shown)


LukaJCB:
Hi @AlejandroME, this looks really good and thank you for your contribution! 
However, we want our code to be checked by the scala compiler and use [tut](https://github.com/tpolecat/tut) for that, can you change your example code to use `tut`? :)

AlejandroME:
@LukaJCB Thank you for your feedback! :)
Sure, I'll take a look at tut and then I'll reach you out again.

AlejandroME:
@LukaJCB It's done! :)

I have a couple of points to state:

1. In the example of `Validated` with a `for`-comprehension, I had to leave it annotated with `scala` because `tut:silent:fail` doesn't work (it is throwing an error with the imports expecting that they'll also fail). I found [this](https://github.com/tpolecat/tut/issues/57) and I see that is the correct behavior of `tut`. Also, I left the signatures of `Validated`, `Valid` and `Invalid` annotated with `scala` for preventing collisions with the rest of the code.

2. For my dummy tests of the example, I was using `0.9.0` and I've found that in `1.0.0-MF`, `|@|` is deprecated. So I've changed the example to use the suggested `.mapN` syntax and I left a _deprecation notice_ stating this. As I see, these docs reflect the _global_ behavior of the library. Because of this, I added the warning.
If you find that this should not be here please let me know and I fix this.

Thank you!

docs/src/main/tut/datatypes/validated.md
```diff
@@ -12,7 +12,310 @@ Response comes back saying your username can't have dashes in it, so you make so
...
+  private def validateFirstName(firstName: String): ValidationResult[String] =
+    if (firstName.matches("^[a-zA-Z]+$")) firstName.validNel else FirstNameHasSpecialCharacters.invalidNel
+
+  private def validateLastName(lastName: String): ValidationResult[String] =
+    if (lastName.matches("^[a-zA-Z]+$")) lastName.validNel else LastNameHasSpecialCharacters.invalidNel
+
+  private def validateAge(age: Int): ValidationResult[Int] =
+    if (age >= 18 && age <= 75) age.validNel else AgeIsInvalid.invalidNel
+
+  def validateForm(username: String, password: String, firstName: String, lastName: String, age: Int): ValidationResult[RegistrationData] = {
```
LukaJCB: Can you perhaps put all the `validateN` functions into another snippet that's checked by tut, and then only have the `validateForm` go without `tut`? I don't think you need the surrounding trait btw :)

LukaJCB: Meant to put this on the snippet above 😅

AlejandroME: Got it! Working on it and also on the support of <2.11.x `Either` for-comprehensions (Travis complained about `flatMap` on this versions).

Thank you for your patience with this! :)

LukaJCB: Just import `cats.syntax.either._` to get that working :)

Thank you, for you help with the docs!

AlejandroME:
@LukaJCB I've fixed the checking with `tut` as you've suggested: I've split into two parts that example: the `validateN`, checked as `tut:silent` and the for-comprehension part, checked as `tut:book:fail` for getting the output of the compilation.

I'm waiting for the Travis build outcome and further suggestions :)

docs/src/main/tut/datatypes/validated.md
```diff
@@ -12,7 +12,298 @@ Response comes back saying your username can't have dashes in it, so you make so
...
+
+## A first approach
+
+You'll note firsthand that `Validated` is very similar to `Either` because it also has two possible values: errors on the left side or successful computations on the right side.
+
+Signature of the structure is as follows:
+
+```scala
+sealed abstract class Validated[+E, +A] extends Product with Serializable {
+	// Implementation elided
```
LukaJCB: Tiny nitpick, but can we reduce the indentation here?

docs/src/main/tut/datatypes/validated.md
```diff
@@ -12,7 +12,298 @@ Response comes back saying your username can't have dashes in it, so you make so
...
+import cats.data._
+import cats.data.Validated._
+import cats.implicits._
+
+sealed trait FormValidatorNel {
+
+  type ValidationResult[A] = Validated[NonEmptyList[DomainValidation], A]
+
+  private def validateUserName(userName: String): ValidationResult[String] =
+    if (userName.matches("^[a-zA-Z0-9]+$")) userName.validNel else UsernameHasSpecialCharacters.invalidNel
```
LukaJCB: Maybe we could give a very quick heads-up on the use of `.validNel` and `.invalidNel`? :)

LukaJCB: Nevermind, I see you're doing it afterwards 😄

docs/src/main/tut/datatypes/validated.md
```diff
@@ -12,7 +12,298 @@ Response comes back saying your username can't have dashes in it, so you make so
...
+But this approach won't compile, as you can see in the previous snippet. Why?
+
+Without diving into details about monads, a for-comprehension uses the `flatMap` method for composition. Monads like `Either` can be composed in that way, but the thing with `Validated` is that it isn't a monad, but an [_Applicative Functor_](../typeclasses/applicativetraverse.html).
+That's why you see the message: `error: value flatMap is not a member of cats.data.Validated[DomainValidation,String]`.
+
+So, how do we do here?
+
+### Meeting applicative
+
+We have to look into another direction: a for-comprehension plays well in a fail-fast scenario, but the structure in our previous example was designed to catch one error at a time, so, our next step is to tweak the implementation a bit.
```
LukaJCB: Maybe we should mention somewhere here, that in order to get the `Applicative` instance for `Validated` the left side needs to have a `Semigroup` instance :)

docs/src/main/tut/datatypes/validated.md
```diff
@@ -12,7 +12,298 @@ Response comes back saying your username can't have dashes in it, so you make so
...
+
+### Coming from `Either`-based validation
+
+cats offer you a nice set of combinators to transform your `Either` based approach to a `Validated` one and vice-versa.
+Please note that, if you're using an `Either`-based approach as seen in our first example, you're constrained to the fail-fast nature of `Either`, but you're gaining a broader set of features with `Validated`.
+
+To do this, you'll need to use either `.toValidated` or `.toValidatedNel`. Let's see an example:
+
+```tut:book
+FormValidator.validateForm(
```
LukaJCB: I'm not sure how good this example really is, the fail-fast behaviour of `Either` is already done at this point, so converting to Validated does not somehow enable the Error accumulation we might want, as you probably know. :)
Maybe we could include an example that doesn't drop the error accumulation?

docs/src/main/tut/datatypes/validated.md
AlejandroME: Is there a standard used by cats? Spaces, tabs?

docs/src/main/tut/datatypes/validated.md
AlejandroME: Working on it! :)

docs/src/main/tut/datatypes/validated.md
AlejandroME: I was thinking the same. I've done this little example for making a use of `.invalid`/`.invalidNel` but nothing, apart from this snippet came to my mind. I'll think about this and I'll reach you out again :)

docs/src/main/tut/datatypes/validated.md
LukaJCB: two spaces is standard :)

AlejandroME: Got it! I'll fit it in all the code snippets.

AlejandroME:
@LukaJCB Some notes here:

1. I've Fixed indentation in all the snippets, as suggested :)
2. I've added a new section (_Short detour_) in where I explain the role of a `Semigroup` in the accumulation and the possibility of use of other structures different to `NonEmptyList`. I haven't mentioned the `Applicative`, so, If you see that we need to detail a little bit about it or If I've made some mistake in the definition, please let me know :) (PS: I think that the Wikipedia quote should fit well here and also the reference to the `Semigroup` docs).
3. I definitely don't know how to implement the example of `.toValidated/.toValidatedNel` in a fail-slow manner. I've tweaked a bit that section (renamed to _Going back and forth_) adding the `.toEither` combinator. If you have a suggestion for the fail-slow example, please let me know.
If not, I can delete that part.

LukaJCB:
About your 3rd point, why don't you just convert some of your `Either` example using `toValidatedNel`, e.g. `validateUserName`. That should work, right? :)

docs/src/main/tut/datatypes/validated.md
```diff
@@ -256,47 +259,97 @@ For example:
...
 ```
 
 Sweet success! Now you can take your validation process to the next level!
 
-### Coming from `Either`-based validation
+### A short detour
+
+Typically, you'll see that `Validated` will be accompanied by a `NonEmptyList` when it comes to accumulation. The thing here is that you can define your own accumulative data structure and you're not limited to the aforementioned construction.
+
+For doing this, you have to provide a `Semigroup` instance. `NonEmptyList`, by definition has its own `Semigroup`. For those who don't know what a `Semigroup` is, let's see a simple example.
```
LukaJCB: This is really great, thank you! However, you didn't really include an example?

docs/src/main/tut/datatypes/validated.md
```diff
@@ -256,47 +259,97 @@ For example:
...
+
+#### Accumulative Structures
+
+According to [Wikipedia](https://en.wikipedia.org/wiki/Semigroup):
```
LukaJCB: Instead of linking to wikipedia, I think the link to cats documentation should be enough :) 
Maybe we could just say " For those who don't know what a `Semigroup` is, you can find out more [here](../typeclasses/semigroup.html)."

What do you think? :)

AlejandroME: Sounds good! I'll fix it.

docs/src/main/tut/datatypes/validated.md
AlejandroME: It was more about how `ap` works for `Invalid`, but I can also provide a simple example of `Semigroup`. What do you think about giving the example first and then point to the actual behavior, as it is written now?

AlejandroME:
@LukaJCB Yes! Sorry, I didn't figure it out! I'll work on this :)
Thanks for the suggestion!

docs/src/main/tut/datatypes/validated.md
```diff
@@ -12,7 +12,351 @@ Response comes back saying your username can't have dashes in it, so you make so
...
+Time to do some refactoring! We're going to try a `Validated` approach:
+
+```tut:silent
+import cats.data._
+import cats.data.Validated._
+import cats.implicits._
+
+def validateUserName(userName: String): Validated[DomainValidation, String] = {
+  if (userName.matches("^[a-zA-Z0-9]+$")) Valid(userName) else Invalid(UsernameHasSpecialCharacters)
+}
```
peterneyens: Maybe reuse the `Either` methods here and call `toValidated` ?

docs/src/main/tut/datatypes/validated.md
```diff
@@ -12,7 +12,351 @@ Response comes back saying your username can't have dashes in it, so you make so
...
+So, how do we do here?
+
+### Meeting applicative
+
+We have to look into another direction: a for-comprehension plays well in a fail-fast scenario, but the structure in our previous example was designed to catch one error at a time, so, our next step is to tweak the implementation a bit.
+
+```tut:silent
+import cats.data._
+import cats.data.Validated._
+import cats.implicits._
```
peterneyens: You don't need to import these again.

docs/src/main/tut/datatypes/validated.md
```diff
@@ -12,7 +12,351 @@ Response comes back saying your username can't have dashes in it, so you make so
...
+
+sealed trait FormValidatorNel {
+
+  type ValidationResult[A] = Validated[NonEmptyList[DomainValidation], A]
```
peterneyens: Since you are using `validNel` and `invalidNel` below, I think you can use `ValidatedNel` here as well.

docs/src/main/tut/datatypes/validated.md
```diff
@@ -12,7 +12,351 @@ Response comes back saying your username can't have dashes in it, so you make so
...
+def ap[EE >: E, B](f: Validated[EE, A => B])(implicit EE: Semigroup[EE]): Validated[EE, B] =
+  (this, f) match {
+    // ...
+    case (Invalid(e1), Invalid(e2)) => Invalid(EE.combine(e2, e1))
+    // ...
+  }
+}
+```
+
+We've omitted the complete implementation because our focus here is the case in where you need to append (that's the function of this method) two failures. Note the `implicit EE: Semigroup[EE]` parameter and the usage of its `.combine` operation. In the case of `NonEmptyList`, we're talking about a `List`, with certain properties that allow us to _combine_ (append) more than one element to it. That's because, apart from the fact that it is a `List`, it also has an instance of `Semigroup`, telling it how to operate with the accumulation.
```
peterneyens: I find this section a bit confusing.

Maybe something like : 

> We've omitted the complete implementation and only show the part where two failures are combined using the `combine` method of the [`Semigroup`](https://typelevel.org/cats/typeclasses/semigroup.html) instance of `EE`.

I don't think we need the part about `NonEmptyList`. Maybe you can add an example of `NonEmptyList`'s `Semigroup` to the Accumulative Structures section ?

Something like : 

```scala
NonEmptyList.one("error 1") |+| NonEmptyList("error 2", "error 3")

"error 1".invalidNel[Int] |+| "error 2".invalidNel
("error 1".invalidNel[Int], "error 2".invalidNel[Int]).mapN(_ + _)
```

docs/src/main/tut/datatypes/validated.md
```diff
@@ -12,7 +12,351 @@ Response comes back saying your username can't have dashes in it, so you make so
...
+
+As we've said before: if you need another data type for processing the failures, you can use it, providing an instance of a `Semigroup` with the `.combine` logic.
+
+### Going back and forth
+
+cats offer you a nice set of combinators to transform your `Validated` based approach to an `Either` one and vice-versa.
+Please note that, if you're using an `Either`-based approach as seen in our first example and you choose to convert it to a `Validated` one, you're constrained to the fail-fast nature of `Either`, but you're gaining a broader set of features with `Validated`.
```
peterneyens: What do you mean with 
> you're gaining a broader set of features with `Validated`.

docs/src/main/tut/datatypes/validated.md
```diff
@@ -12,7 +12,351 @@ Response comes back saying your username can't have dashes in it, so you make so
...
+final case class Valid[+A](a: A) extends Validated[Nothing, A]
+final case class Invalid[+E](e: E) extends Validated[E, Nothing]
+```
+
+Before diving into `Validated`, let's take a look at an `Either` based first approach to address our validation necessity.
+
+Our data will be represented this way:
+
+```tut:silent
+case class RegistrationData(username: String, password: String, firstName: String, lastName: String, age: Int)
```
peterneyens: Maybe add `final` here. Probably good to show best practices in the documentation?

docs/src/main/tut/datatypes/validated.md
```diff
@@ -12,7 +12,351 @@ Response comes back saying your username can't have dashes in it, so you make so
...
+```
+
+We have our `RegistrationData` case class that will hold the information the user has submitted, alongside the definition of the error model that we'll be using it for displaying the possible errors of every field. Now, let's explore the proposed implementation:
+
+```tut:silent
+import cats.syntax.either._
+
+sealed trait FormValidator{
+  private def validateUserName(userName: String): Either[DomainValidation, String] =
+    if (userName.matches("^[a-zA-Z0-9]+$")) Right(userName) else Left(UsernameHasSpecialCharacters)
```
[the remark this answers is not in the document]
AlejandroME: This is new for me! I'll change it in all the conditions :)

docs/src/main/tut/datatypes/validated.md
AlejandroME: Good suggestion. I'll take a look at it!

docs/src/main/tut/datatypes/validated.md
AlejandroME: This is the section in where I have doubts about it. I was trying to express that, coming from `Either` to `Validated`, the person will gain more _combinators_ (_Features_), even if you transfer the `Either` behavior (fail-fast) to `Validated`. 

With your previous comments, I have an idea for simplifying this section (or even delete it). Let me work on it and I'll reach you out again :)

AlejandroME:
Hello, @LukaJCB and @peterneyens!

I've worked on your suggestions. Some notes:

1. I've deleted the ambiguous part about the conversion between `Validated` and `Either`, losing the fail-slow feature but _gaining_ some other combinators. With the suggestions of @peterneyens, since the beginning of the docs I've used `.toValidated` for reusing the validation methods and then at the end I've used `.toEither` on the `ValidatedNel` example.
2. I've followed the advice of Peter about the `Semigroup` example and I just used the example that he provided about `NonEmptyList`, but with the validations developed through this docs.
3. I've fixed some typos and fixed some other things.

I'll stay tuned for your comments about this.

Thanks!

kailuowang:
Looks like all comments are addressed. LGTM thanks so much @AlejandroME !!

AlejandroME:
Thank you so much for all the feedback received! I'm very happy to contribute to the project! :+1:

PR #1905, opened 2017-09-10
s/rewrites/rules and update scalafix version in the README

gabro:
This a follow up of #1902, in which I forgot to:

- rename the directory from `rewrites` to `rules` (this made the github shorthand syntax fail)
- update the version in the README

Sorry for the double PR!


gabro:
🤔 mm tests don't seem to pass anymore, I'm investigating...

gabro:
ok, everything works correctly now. I also took the opportunity to use the new `SymbolMatcher` API which makes the use of Symbols a bit more solid and convenient.

PR #1910, opened 2017-09-13
Add NonEmptyList.ofInitLast

eddsteel:
I mentioned this on gitter (thanks @edmundnoble for the name suggestion). I needed to build a NEL from an init list and a last element, and didn't like using `unsafeFromList`. Not sure if it's generally useful but seems pretty natural given List's `init` and `last` operations.

No worries if you think it's out of scope and don't want to include it.


johnynek:
:+1:

Issue #1904, opened 2017-09-10
many kernel algebraic typeclasses should be `Invariant`


johnynek:
If you have `Invariant[M]` and an isomorphism (or bijection) `A => B, B => A` then you can go between `M[A] <=> M[B]`. This can be very useful when combined with shapeless which can generate iso morphisms between case classes and tuples, etc...

However, Semigroup, Semilattice, Monoid, Group, etc... are left out of that as far as I can see.

It would be nice to have instances for those in `object Invariant` so they could be used without import.

kailuowang:
Shall we schedule this in RC1?

johnynek:
I would like to. I hate when we have obvious instances for our own types that we don't ship.

LukaJCB:
Hey @johnynek, can you elaborate a bit on what you mean by this?  :)
Do you mean instances like `Invariant[Group]` etc.? 
We do already have `InvariantMonoidal` which does extend `Invariant`, so we can already do 
```scala
Monoid[A].imap(aToB)(bToA)
```
I don't think we have these for any of the other kernel classes apart from `Semigroup` and `Monoid` though.

johnynek:
Concretely:

1. we usually put instances for cats owned types into companions. For Monoid we have an import. Should not need that. We should just put that into `Invariant` or `InvariantMonodial` object.
2. we are missing many instances: e.g. all the ones I mentioned above. I'm not sure all are `InvariantMonoidal`. Monoid is, but some of the others may not (pure may or may not work for all cases).

LukaJCB:
I like the idea of having everything in `object Invariant`, but we'd break the API, no? 
We could do something like `object InvariantInstances extends MonoidInstances with SemigroupInstances`, WDYT?

johnynek:
I think these are so rarely used, I don't mind breaking the API to make it consistent current cats practice before 1.0

LukaJCB:
I'll submit a PR soon :)

LukaJCB:
Fixed by #1914 :)

Issue #1908, opened 2017-09-12
ApplicativeError: catchNonFatal, catchNonFatalEval and fromTry are not tested


alexandru:
There are currently no laws in `ApplicativeErrorLaws` that cover:

- `catchNonFatal`
- `catchNonFatalEval`
- `fromTry`

When I check the associated laws for a type-class, I expect to get full code coverage given good enough `Arbitrary` values.

kailuowang:
> When I check the associated laws for a type-class, I expect to get full code coverage given good enough Arbitrary values.

Unfortunately that's the not the case for most of type classes when it comes to implemented methods. Some of them are covered by tests for the type classes themselves, some are not covered  at all.  @non recently [added a bunch of such tests](https://github.com/typelevel/cats/pull/1878) to laws and name them with a "Ref" suffix. Maybe we should make it a convention.

johnynek:
+1 to make it a convention: if there is a default implementation that may be slow, we should add a law that every implementation matches that but may have better ways to implement.

peterneyens:
Maybe we can use something like I mentioned [here](https://github.com/typelevel/cats/issues/1684#issuecomment-302723134).

In essence create a wrapper for every type class which uses the implementations of the fundamental operations of that type class of the data type it wraps. We can than compare the actual implementation (eg `ApplicativeError[Foo, E]`) is equal to the default implementation (`ApplicativeError[ApplicativeErrorDefaultWrapper[Foo, E, ?], E]`).

This way the consistency laws wouldn't need to reimplement the default implementation.
I'll try to create a more concrete proof of concept.

kailuowang:
@peterneyens yeah, we dropped the ball on that one. It's possible to generate check methods using macro so that all future methods with default implementation can be automatically tested. It maybe even possible to generate the wrappers, but that seems trickier.

PR #1914, opened 2017-09-15
Add Invariant instances for kernel type classes

LukaJCB:
Adds `InvariantMonoidal` to most of the kernel type classes as discussed in #1904.


kailuowang:
close and reopen to trigger build again (somehow failed on codecov)

LukaJCB:
Should mention that this is an API breaking change, as imports from `cats.instances.monoid` and `cats.instances.semigroup` no longer work.

kailuowang:
I probably missed something but I had the impression that we want to include all these instances into the companion objects of `Invariant` and `InvariantMonoidal`, so that no imports is needed at all. Is that not possible or having significant cons?

LukaJCB:
Oh, I may have misunderstood then. The only con, I can think about is that these won't get imported when importing all instances which might be confusing to some?

kailuowang:
If you put them in the companion, they never need to be imported right?

LukaJCB:
They don't? 🤔 Maybe I'm missing something here, but what would happen if I want to do something like:

```scala
val semiB: Semigroup[B] = Semigroup[A].imap(...)
```

I'd need to import the implicit instance from somewhere, right? 
So it'd have be something along the lines of `import cats.functor.Invariant.

kailuowang:
@LukaJCB for your example, it should work without import if the `Invariant` instance for `Semigroup` is defined in `Invariant`'s companion object. Actually what I said previously isn't accurate, we should put these `Invariant` and `InvariantMonoidal` instances into the companion object of `Invariant`

demo
```scala
trait TCParent[F[_]]

trait TCChild[F[_]] extends TCParent[F]

trait A[T]

object TCParent {
  implicit val tcA: TCChild[A] = null
}

object Test {

  implicit class TCParentOps[F[_], T](f: F[T])(implicit tcF: TCParent[F]) {
    def e1: String = "bah"
  }

  implicit class TCOps[F[_], T](f: F[T])(implicit tcF: TCChild[F]) {
    def e: String = "bah"
  }

  val a: A[String] = ???
  a.e
  a.e1
}

```

LukaJCB:
Ah okay, I think I've found the issue, we currently have tests, that check whether `Invariant`, `Cartesian` and `InvariantMonoidal` instances can be summoned and if we put it in `object Invariant` it can't find those `Cartesian` instances. However, if we put them in `object InvariantMonoidal` it can't find `Invariant` or `Cartesian` instances. 
I guess the `Cartesian` instances aren't that important, but I'm not sure, WDYT?

kailuowang:
ah didn't realize about the `Cartesian`, Crap, we can't put them in both then we have to put them an `invariant` object for people to import then.  @johnynek WDYT.

johnynek:
we could have a fallback possibly that looks in the other places.

So if `A extends B` we could have a low priority implicit that looks in A to provide an implicit of B. That said, that makes everything have a circular dependency.

Not really sure that's a great idea.

core/src/main/scala/cats/instances/invariant.scala
```diff
@@ -0,0 +1,168 @@
...
+      def combine(x: A, y: A): A = a
+      override def combineAllOption(as: TraversableOnce[A]): Option[A] =
+        if (as.isEmpty) None else Some(a)
+    }
+  }
+
+  implicit val catsInvariantMonoidalMonoid: InvariantMonoidal[Monoid] = new InvariantMonoidal[Monoid] {
+    def product[A, B](fa: Monoid[A], fb: Monoid[B]): Monoid[(A, B)] = new Monoid[(A, B)] {
+      val empty = fa.empty -> fb.empty
+      def combine(x: (A, B), y: (A, B)): (A, B) = fa.combine(x._1, y._1) -> fb.combine(x._2, y._2)
```
johnynek: shame this destroys the combineAllOption optimization. Maybe we can add a ticket to address this after we merge.

In algebird, we buffer 1000 items internally, then pass over the block with each of the underlying monoids.

core/src/main/scala/cats/instances/invariant.scala
```diff
@@ -0,0 +1,168 @@
...
+
+    def imap[A, B](fa: CommutativeMonoid[A])(f: A => B)(g: B => A): CommutativeMonoid[B] = new CommutativeMonoid[B] {
+      val empty = f(fa.empty)
+      def combine(x: B, y: B): B = f(fa.combine(g(x), g(y)))
+    }
+
+    def pure[A](a: A): CommutativeMonoid[A] = new CommutativeMonoid[A] {
+      val empty = a
+      def combine(x: A, y: A): A = a
+      override def combineAll(as: TraversableOnce[A]): A = a
```
johnynek: can we only override `combineAllOption`? If we override only `combineAll`, then calls to `combineAllOption` still get the slow path. `combineAll` calls back to `combineAllOption` (combineAll should probably be final, actually).

core/src/main/scala/cats/instances/invariant.scala
```diff
@@ -0,0 +1,168 @@
+package cats.instances
+
+import cats.kernel._
+import cats.InvariantMonoidal
+
+trait InvariantInstances {
```
johnynek: looks like these are all `InvariantMonoidal` not just `Invariant`. Should we put them in a different trait name?

LukaJCB:
I've added an override for `combineAllOption` to all the instances :)

core/src/main/scala/cats/instances/invariant.scala
LukaJCB: I did override in all of the instances, but can you comment on how this is useful? I really can't come up with a way 🤔

tests/src/test/scala/cats/tests/AlgebraInvariantTests.scala
```diff
@@ -22,21 +22,157 @@ class AlgebraInvariantTests extends CatsSuite {
...
+  implicit val arbBoundedSemilatticeSetInt: Arbitrary[BoundedSemilattice[Set[Int]]] =
+    Arbitrary(genBoundedSemilatticeSetInt)
+
+  implicit val arbGroupInt: Arbitrary[Group[Int]] =
+    Arbitrary(genCommutativeGroupInt)
+
+  implicit val arbCommutativeGroupInt: Arbitrary[CommutativeGroup[Int]] =
+    Arbitrary(genCommutativeGroupInt)
+
+  test("Semigroup combineAllOption after imap with id is consistent"){
```
kailuowang: I propose we add this to the Semigroup law as a "Ref" law like these ones 
https://github.com/typelevel/cats/blob/master/laws/src/main/scala/cats/laws/FoldableLaws.scala#L122-L143

LukaJCB: Ah, I didn't know we had such a convention, good to know :)

kailuowang: We dont' really, but I'd like to establish one, see discussion [here](https://github.com/typelevel/cats/issues/1908#issuecomment-328899290). @peterneyens has some interesting ideas too.

tests/src/test/scala/cats/tests/AlgebraInvariantTests.scala
```diff
@@ -22,21 +23,159 @@ class AlgebraInvariantTests extends CatsSuite {
...
+  implicit val arbBoundedSemilatticeSetInt: Arbitrary[BoundedSemilattice[Set[Int]]] =
+    Arbitrary(genBoundedSemilatticeSetInt)
+
+  implicit val arbGroupInt: Arbitrary[Group[Int]] =
+    Arbitrary(genCommutativeGroupInt)
+
+  implicit val arbCommutativeGroupInt: Arbitrary[CommutativeGroup[Int]] =
+    Arbitrary(genCommutativeGroupInt)
+
+  def combineAllOption_Ref[F[_] : InvariantMonoidal, A: Eq](name: String, a: A)(implicit ev: F[A] <:< Semigroup[A]): Unit = {
```
kailuowang: Hmm, I was suggesting adding this to the `SemigroupLaw` but I just realized that the consolidation PR #1922 isn't merged yet. Let's wait until that one gets merged, shall we?

LukaJCB: I agree we should wait, however, this one isn't really about `combineAllOption` for any given semigroup, but for the ones created by `InvariantMonoidal[F].pure`. `SemigroupLaws` already has a check for `combineAllOption` here: https://github.com/typelevel/cats/blob/master/kernel-laws/src/main/scala/cats/kernel/laws/GroupLaws.scala#L33 :)

kailuowang: ah I see.  so, even without this methods and the tests that use it, the `combineAllOption` should be covered once you used `GroupLaws` to test the new instances? So for these `pure` given instance, why not test them with the `GroupLaws` again?

LukaJCB: I think we could do that, but I didn't think the `pure` ones were actually lawful instances.

kailuowang: I'd be curious to find out.

LukaJCB: Okay so the `Semigroup` instance is not lawful, because of the `combineAllOption`, as defined  [here](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/instances/semigroup.scala#L20), is not consistent with `reduceOption(_ |+|
 _)`, because when you pass a collection with only one value `reduceOption` will only turn that into an `Option`. One could argue that this doesn't necessarily break the Semigroup laws though.

In addition the `Monoid` instance fails to fulfil the left and right identity laws. 
Also these instances aren't able to abide by the idempotency law. 

Does this mean `Monoid` is not a valid `InvariantMonoidal`? I'm not sure I know enough about that.

kailuowang: if the pure instance is not lawful, I am not sure if it's still a `InvariantMonoidal`, maybe just a `Invariant`? maybe worthwhile to find out how it's defined in Haskell.

johnynek: I was surprised at the `pure`. I thought these two were only `Invariant`.

By the way, I think `Invariant` and `Cartesian` are the two ways you actually want to compose these things.

LukaJCB: So, I think we should get rid of all Invariant Monoidal except for Semigroup and Commutative Semigroup and Turn the others into Invariant instances

kailuowang: 👍  BTW, for invariance instances, they can be moved the companion object of `Invariant` right?

LukaJCB:
To summarize the breaking changes introduced in this PR for later use:

- The `InvariantMonoidal` instance from `cats.instances.semigroup` was moved to `cats.instances.invariant`
- The `InvariantMonoidal` instance from `cats.instances.monoid` was removed as it was not lawful. In its place there's an `Invariant` instance, located in `object Invariant`.

kailuowang:
do you mind also update the doc of `InvariantMonoidal` https://github.com/typelevel/cats/blob/master/docs/src/main/tut/typeclasses/invariantmonoidal.md#L20
Line 20 mentioned `Monoid` as an `InvariantMonoidal`

kailuowang:
@johnynek got any more feedback for this PR?

tests/src/test/scala/cats/tests/MonoidTests.scala
```diff
@@ -6,8 +6,6 @@ package tests
 class MonoidTests extends CatsSuite {
   {
     Invariant[Monoid]
-    Semigroupal[Monoid]
```
johnynek: why did we lose this? I don't see it?

LukaJCB: Good catch, I falsely assumed, that because `Monoid` wasn't a valid `InvariantMonoidal`, that it also wasn't a lawful `Semigroupal`. Turns out, it is, and I added it back! Thanks! :)

kailuowang: I missed that too. yeah, what broke the law was the `pure` method added in `InvariantMonoidal`

johnynek:
👍

johnynek:
@LukaJCB Thank you so much for carrying this over the line!

PR #1917, opened 2017-09-18
Don't use package object convention for object source path

travisbrown:
The `cats.implicits` object is just an ordinary Scala object, not a package object, but its source path follows the convention for a package object. This causes problems on [the latest version of Bazel](https://github.com/bazelbuild/rules_scala/pull/279), where the Scala compiler gets confused by the presence of the `-sources` jar and complains that "package cats contains object and package with same name: implicits".

Thanks to @ianoc-stripe for tracking this down, and for proposing a fix on the Bazel side. There's not really any reason for Cats to do this in the first place, though, and this PR simply moves the file so that the source jar doesn't end up with a `cats/implicits` directory. This is 100% compatible and shouldn't have any affect on anything except build tools that for whatever reason feed source jars to the Scala compiler.


kailuowang:
Thank you @travisbrown. No idea why it was set up that way.

johnynek:
👍

Issue #1841, opened 2017-08-22
Document implicit naming convention


aaronlevin:
We should update the contributing guidelines to specify implicit naming convention. I believe the convention is to prefix the name of the implicit with a camel-cased version of the package namespace. See quote from @peterneyens on a ticket I was working on (adding `Traverse` instances in the `cats.free` package):

> It is not really documented outside issue #1061 itself (it should probably be added to [these guidelines](https://typelevel.org/cats/guidelines.html)): the implicit values in Cats are prefixed with the package.
>
> So these implicits would be `catsFreeFoldableForFree` and `catsFreeTraverseForFree`.

tbje:
Hey @aaronlevin, did #1920 fix this issue?

kailuowang:
fixed by #1920 thanks again @tbje

PR #1920, opened 2017-09-20
Document naming implicits according to @non s comment in #1061

tbje:
- fixing #1841.

### DIFF
--- a/docs/src/main/tut/guidelines.md
+++ b/docs/src/main/tut/guidelines.md
@@ -74,10 +74,14 @@ to make this intermediate class a [Value Class](http://docs.scala-lang.org/overv
 In a widely-used library it's important to minimize the chance that the names of implicits will be used by others and
 therefore name our implicits according to the following rules:
 
-- Implicits should start with "cats" -- a third-party will less likely do this.
+- Implicits should start with "cats" followed by the package name (where the instance is defined).
+- If the package contains `instances` leave `instances` out.
 - The type and the type class should be mentioned in the name.
+- If the instance is for multiple type classes, use `InstancesFor` instead of a type class name.
+- If the instance is for a standard library type add `Std` after the package. i.e. `catsStdShowForVector` and `catsKernelStdGroupForTuple`.
 
-As an example, for `Monoid[List[A]]`, the name of the implicit should be `catsMonoidForList`.
+As an example, for `cats.kernel.Monoid[List[A]]`, the name of the implicit should be `catsKernelMonoidForList`.
+This rule is relatively flexible. Use what you see appropriate. The goal is to maintain uniqueness and avoid conflicts.
 
 #### TODO:
 Once we drop 2.10 support, AnyVal-extending class constructor parameters can be marked as private.
--- a/docs/src/main/tut/guidelines.md
+++ b/docs/src/main/tut/guidelines.md
@@ -13,7 +13,7 @@ All guidelines in cats should have clear justifications. There is no room for tr
 
 ### <a id="implicit-syntax-conversions" href="#implicit-syntax-conversions"></a> Composing Implicit Conversions in Traits
 
-Implicit syntax conversions provided in publicly-exposed traits should be marked final 
+Implicit syntax conversions provided in publicly-exposed traits should be marked final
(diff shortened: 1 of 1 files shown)


docs/src/main/tut/guidelines.md
```diff
@@ -69,6 +69,15 @@ The user doesn't need to specify the type `A` which is given by the parameter.
...
+### <a id="ops-classes" href="#implicit-naming"></a> Implicit naming
 
-#### TODO: 
+In a widely-used library it's important to minimize the chance that the names of implicits will be used by others and
+therefore name our implicits according to the following rules:
+
+- Implicits should start with "cats" -- a third-party will less likely do this.
+- The type and the type class should be mentioned in the name.
+
+As an example, for `Monoid[List[A]]`, the name of the implicit should be `catsMonoidForList`.
```
kailuowang: Thanks so much! We actually also include the package name in the name now. 
`catsPackageNameTypeclassNameForDataStruct`

tbje: Thank you for letting me know. Would the following be correct ?

In a widely-used library it's important to minimize the chance that the names of implicits will be used by others and therefore name our implicits according to the following rules:

- Implicits should start with "cats" followed by the package name.
- The type and the type class should be mentioned in the name.

As an example, for `cats.kernel.Monoid[List[A]]`, the name of the implicit should be `catsKernelMonoidForList`.

peterneyens: We should probably mention that in `cats.instances` we don't follow this rule and use `catsStd....`.

kailuowang: The package used in the name is package where the instance is defined, your example gives the impression that it's the package where the type class is defined. 
BTW, we are not really strictly following this rule, for instance, in `cats.instances` we named an instances as 
`catsStdInstancesForOption`
https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/instances/option.scala#L8

here `Std` is to represent the package `instances` which used to be called `std`, we didn't change the name here because `catsInstancesInstancesForOption` would be weird.
the next word `Instances` in the name is used here because this instance is an instance for multiple type classes. 
So we do have some flexibility here.

kailuowang: My recommendation would be just 
```
In a widely-used library it's important to minimize the chance that the names of implicits will be used by others and therefore name our implicits according to the following rules:

* Implicits should start with "cats" followed by the package (where instance is defined) name.
* The type and the type class should be mentioned in the name.
* If the instance is for multiple type classes, use `InstancesFor` instead 
* This rule is relatively flexible. Use what you see appropriate. The goal is to maintain the uniqueness and avoid conflicts. 
```

tbje: It seems like Std is used when the instance is for a Scala standard library type.

edmundnoble: I've gone through all of the instances with @tbje and as far as I can tell the current revision is correct in all cases.

docs/src/main/tut/guidelines.md
```diff
@@ -69,6 +69,19 @@ The user doesn't need to specify the type `A` which is given by the parameter.
...
+In a widely-used library it's important to minimize the chance that the names of implicits will be used by others and
+therefore name our implicits according to the following rules:
+
+- Implicits should start with "cats" followed by the package name (where the instance is defined).
+- If the package contains `instances` leave `instances` out.
+- The type and the type class should be mentioned in the name.
+- If the instance is for multiple type classes, use `InstancesFor` instead of a type class name.
+- If the instance is for a standard library type add `Std` after the package. i.e. `catsStdShowForVector` and `catsKernelStdGroupForTuple`.
+
+As an example, for `cats.kernel.Monoid[List[A]]`, the name of the implicit should be `catsKernelMonoidForList`.
```
kailuowang: This example may give the false impression that the package name is the package of the type class. How about 
```
As an example, an implicit instance of `Monoid` for `List` defined in the package `Kernel` should be named as `catsKernelMonoidForList`
```

tbje:
Thank you for guiding me !

PR #1748, opened 2017-06-28
Stack-safe FreeAp

edmundnoble:
Port from JS, from @safareli's implementation. It's stack-safe, I just need to benchmark and comment it.

Fixes #1151.

PR #1921, opened 2017-09-20
Optimize FreeApplicative.product

peterneyens:
Follow up after #1748 

`Ap` now contains two `FreeApplicative`s (instead of just one before), so we can do some more optimisation as we are currently doing. (The new implementation should also result in the same `FreeApplicative` as the original implementation).

Currently the result of `FreeApplicative`'s `map2` is one of the following two :
```scala
Pure(f(a, b))
Ap({ case (a, b) => f(a, b) }, Ap(Ap(Pure(a => b => (a, b)), fa), fb))
```

With this change it can be one of :
```scala
Pure(f(a, b))
Ap(Pure(f(a, _)), fb)
Ap(Pure(f(_, b)), fa)
Ap(Ap(Pure(a => b => f(a, b))), fa), fb)
```

### DIFF
--- a/free/src/main/scala/cats/free/FreeApplicative.scala
+++ b/free/src/main/scala/cats/free/FreeApplicative.scala
@@ -32,6 +32,16 @@ sealed abstract class FreeApplicative[F[_], A] extends Product with Serializable
     }
   }
 
+  final def map2[B, C](fb: FA[F, B])(f: (A, B) => C): FA[F, C] =
+    this match {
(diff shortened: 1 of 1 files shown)


edmundnoble:
This is awesome, it's causing less stack jumps in https://github.com/typelevel/cats/blob/master/free/src/main/scala/cats/free/FreeApplicative.scala#L99 by creating more single-argument functions. :+1: from me.

PR #1922, opened 2017-09-20
Make kernel laws consistent with core laws

### LINKED ISSUE #1272: Make cats-kernel law organization consistent with cats-core

LukaJCB:
Should fix #1272, still a WIP, but compiles and runs so looking for some feedback on some of the decisions made when porting between styles :)


LukaJCB:
I moved most of the machinery from `cats.laws.*` to `cats.kernel.laws.*`, including `IsEq`, `SerializableLaws` and `SerializableTests`. By doing so, I removed the seperate `IsSerializable` object, as I considered the test inside to be equivalent to the one inside `SerializableLaws` (Please correct me here if I'm wrong)

Otherwise I'm pretty much done with this :)

johnynek:
note algebra-laws do depend on this, so we will need to fix algebra before it can be updated after this is published:

https://github.com/typelevel/algebra/blob/master/build.sbt#L143

LukaJCB:
I'll create a PR :)

kernel-laws/src/main/scala/cats/kernel/laws/discipline/BoundedSemilatticeTests.scala
```diff
@@ -0,0 +1,25 @@
...
+package discipline
+
+import org.scalacheck.{Arbitrary, Prop}
+
+trait BoundedSemilatticeTests[A] extends CommutativeMonoidTests[A] with SemilatticeTests[A] {
+
+  def laws: BoundedSemilatticeLaws[A]
+
+  def boundedSemilattice(implicit arbA: Arbitrary[A], eqA: Eq[A]): RuleSet =
+    new RuleSet {
```
kailuowang: nit: most of these `new RuleSet {` where `bases` is empty could be just `new DefaultRuleSet(` right like the one you used in `BandTests`? No big deal though, just that it could be slightly more concise.

kernel-laws/src/main/scala/cats/kernel/laws/Rules.scala
```diff
@@ -1,104 +0,0 @@
-package cats.kernel
-package laws
-
-import org.scalacheck.Prop
-import org.scalacheck.Prop._
-import org.scalacheck.{Arbitrary, Prop}
-import cats.kernel.instances.boolean._
-
-object Rules {
```
kailuowang: This setup is indeed less readable than the cats.core way.

tests/src/test/scala/cats/tests/CategoryTests.scala
```diff
@@ -1,7 +1,7 @@
 package cats
 package tests
 
-import cats.kernel.laws.GroupLaws
+import cats.kernel.laws.discipline.{MonoidTests => MonoidTypeclassTests}
```
kailuowang: Another nit: Maybe we can call these `MonoidLawTests`, because the other `MonoidTest` is also testing the Type Class.

kernel-laws/src/main/scala/cats/kernel/laws/MonoidLaws.scala
```diff
@@ -0,0 +1,28 @@
+package cats.kernel.laws
+
+import cats.kernel.Monoid
+
+trait MonoidLaws[A] extends SemigroupLaws[A] {
```
kailuowang: I couldn't find the replacement for `Rules.isId("isEmpty", A.empty)(A.isEmpty)`?

kernel-laws/src/main/scala/cats/kernel/laws/discipline/BoundedSemilatticeTests.scala
```diff
@@ -0,0 +1,25 @@
...
+      val name: String = "boundedSemilattice"
+      val bases: Seq[(String, RuleSet)] = Nil
+      val parents: Seq[RuleSet] = Seq(commutativeSemigroup, band)
```
kailuowang: should this be `communtativeMonoid, semilattice` instead?

kernel-laws/src/main/scala/cats/kernel/laws/discipline/BoundedSemilatticeTests.scala
LukaJCB: That doesn't work for Typeclasses with more than one parent though, no?

kailuowang: I didn't realize that. disregard my comment then. Maybe we can create one that does in `discipline` later.

kernel-laws/src/main/scala/cats/kernel/laws/PartialOrderLaws.scala
```diff
@@ -0,0 +1,55 @@
+package cats.kernel.laws
+
+import cats.kernel.PartialOrder
+
+trait PartialOrderLaws[A] extends EqLaws[A] {
+  override implicit def E: PartialOrder[A]
+
+  def reflexitivity(x: A): IsEq[Boolean] =
+    E.gteqv(x, x) <-> true
```
kailuowang: the original test uses `?<=`, i.e. `E.lteqv`,  I am not sure if it matters. (mabye we can have both `lteqv` and `gteqv`?)

kernel-laws/src/main/scala/cats/kernel/laws/PartialOrderLaws.scala
```diff
@@ -0,0 +1,55 @@
...
+  def partialCompare(x: A, y: A): IsEq[Boolean] = {
+    val c = E.partialCompare(x, y)
+    (((c < 0) == E.lt(x, y)) && ((c == 0) == E.eqv(x, y)) && ((c > 0) == E.gt(x, y))) <-> true
+  }
+
+  def pmin(x: A, y: A): IsEq[Option[A]] = {
+    val c = E.partialCompare(x, y)
+    val m = E.pmin(x, y)
+    if (c < 0) m <-> Option(x)
+    else if (c == 0) Option(x) <-> Option(y)
```
kailuowang: hmm, this is a bit tricky, by using `IsEq[Option[A]]` we lost the test of `pmin` when the two are equal. I see two of options here,
1.  one change it to `IsEq[Boolean]` just like the old test
2.  change it to `IsEq[(Option[A], Option[A])]`, then you have to do `(m, m) <-> (Option(y), Option(y))`, `(m, m) <-> (Option(x), Option(y))`, etc. you can create a nested helper method to make the duplicated case  more readable
```scala
def isEqualTo(a: Option[A]) =  (m, m) <-> (a, a)
```
It seems to me that 1 is good enough and easier.

kernel-laws/src/main/scala/cats/kernel/laws/PartialOrderLaws.scala
```diff
@@ -0,0 +1,55 @@
...
+    else if (c > 0) m <-> Option(y)
+    else m <-> None
+  }
+
+  def pmax(x: A, y: A): IsEq[Option[A]] = {
```
kailuowang: same as above.

kernel-laws/src/main/scala/cats/kernel/laws/OrderLaws.scala
```diff
@@ -1,122 +1,40 @@
...
-      else if (c > 0) m ?== Some(x)
-      else m ?== None
-    }
-  )
-  // scalastyle:on cyclomatic.complexity
+  def min(x: A, y: A): IsEq[A] = {
+    val c = E.compare(x, y)
+    val m = E.min(x, y)
+    if (c < 0) m <-> x
+    else if (c == 0) x <-> y
```
kailuowang: same issue  as in `PartialOrder`, `min` isn't tested when x == y

kailuowang:
It's looking great. I left a couple of comments. mostly about the min/max tests. Thanks so much for this epic refactor @LukaJCB

tests/src/test/scala/cats/tests/FunctionTests.scala
```diff
@@ -45,7 +44,6 @@ class FunctionTests extends CatsSuite {
   checkAll("Function0[Eqed]", EqLawTests[Function0[Eqed]].eqv)
   checkAll("Function0[POrd]", PartialOrderLawTests[Function0[POrd]].partialOrder)
   checkAll("Function0[Ord]", OrderLawTests[Function0[Ord]].order)
-  checkAll("Function0[Hsh]", HashLaws[Function0[Hsh]].hash)
```
kailuowang: Hmm, was this test removed intentionally?

LukaJCB: Good catch, I removed it initially for erroring for erroring out, as with the new setup, all the laws are tested and you can't single out one specific law to run. I wanted to replace it with a manual test, but forgot, will fix :)

LukaJCB:
I added another MiMa exception as I had to reshuffle the option instances :)

tests/src/test/scala/cats/tests/FunctionTests.scala
```diff
@@ -54,6 +55,13 @@ class FunctionTests extends CatsSuite {
   checkAll("Function0[Grp]", GroupLawTests[Function0[Grp]].group)
   checkAll("Function0[CGrp]", CommutativeGroupTests[Function0[CGrp]].commutativeGroup)
 
+  test("Function0[Hsh]") {
```
kailuowang: just curious why not just test the hash law? won't pass?

tests/src/test/scala/cats/tests/FunctionTests.scala
```diff
@@ -42,19 +42,25 @@ class FunctionTests extends CatsSuite {
...
+  checkAll("Function0[CSemi]", CommutativeSemigroupTests[Function0[CSemi]].commutativeSemigroup)
+  checkAll("Function0[Bnd]", BandTests[Function0[Bnd]].band)
+  checkAll("Function0[SL]", SemilatticeTests[Function0[SL]].semilattice)
+  checkAll("Function0[BSL]", BoundedSemilatticeTests[Function0[BSL]].boundedSemilattice)
+  checkAll("Function0[Mono]", MonoidLawTests[Function0[Mono]].monoid)
+  checkAll("Function0[CMono]", CommutativeMonoidTests[Function0[CMono]].commutativeMonoid)
+  checkAll("Function0[Grp]", GroupLawTests[Function0[Grp]].group)
+  checkAll("Function0[CGrp]", CommutativeGroupTests[Function0[CGrp]].commutativeGroup)
+
+  test("Function0[Hsh]") {
```
kailuowang: Just curious, why not just test the full HashLaw? won't pass?

LukaJCB: Yeah, it's not the same as universal hash and not the same as `scala.util.Hashing`

kernel/src/main/scala/cats/kernel/instances/option.scala
```diff
@@ -3,17 +3,19 @@ package instances
 
 package object option extends OptionInstances
 
-trait OptionInstances extends OptionInstances1 {
+trait OptionInstances extends OptionInstances2 {
```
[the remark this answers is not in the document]
LukaJCB: I renamed the numbers :) Btw, I think this break is an oversight from the Hash PR, as `PartialOrder` and `Hash` instances had the same priority and both extend `Eq`, so it led to ambiguous implicits when requiring `Eq` for `Option`

kailuowang:
> filter with: ProblemFilters.exclude[UpdateForwarderBodyProblem]("cats.kernel.instances.OptionInstances0.catsKernelStdEqForOption")
filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("cats.kernel.instances.OptionInstances0.catsKernelStdPartialOrderForOption")
filter with: ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("cats.kernel.instances.OptionInstances1.catsKernelStdHashForOption")
 filter with: ProblemFilters.exclude[MissingTypesProblem]("cats.kernel.instances.OptionInstances1")

LukaJCB:
Thanks @kailuowang! Fixed it :)

kailuowang:
sorry, more merge needed.

LukaJCB:
No worries, done :)

kernel-laws/src/main/scala/cats/kernel/laws/IsSerializable.scala
```diff
@@ -1,38 +0,0 @@
...
-import org.scalacheck.Prop
-import org.scalacheck.Prop._
-import scala.util.control.NonFatal
-import scala.util.DynamicVariable
-
-/**
- * Object with a dynamic variable that allows users to skip the
- * serialization tests for certain instances.
- */
-private[laws] object IsSerializable {
```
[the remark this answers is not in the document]
kailuowang: No, it's replaced by the `SerializableLaws` moved from `cats.tests`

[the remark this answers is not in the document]
oscar-stripe: 👍

adelbertc:
Oh damn just saw this @LukaJCB - you the man!

Issue #1801, opened 2017-08-08
Documentation megathread


LukaJCB:
I'm trying to compile a thorough collection of everything currently documented and everything we still want to document. I'd also like to talk about a few other things other people brought up. This is sort of meant as a follow up to #1443. Another question that arrises, is if we really need documentation for every single one of these, or if some don't really make sense to have their own page. It might make sense to combine some of these as well (e.g. NonEmpty{List, Vector}).

## Type classes:

- Kernel
  - [ ] Eq #1788 
  - [ ] PartialOrder
  - [ ] Order
  - [x] Semigroup
  - [x] Monoid
  - [ ] Group
  - [ ] CommutativeSemigroup
  - [ ] Band
  - [ ] CommutativeMonoid
  - [ ] Semilattice
  - [ ] CommutativeGroup
  - [ ] BoundedSemilattice

- Functor
  - [x] Invariant
  - [x] Contravariant
  - [ ] BiFunctor
  - [ ] ProFunctor
  - [ ] Strong 

- Core 
  - [x] Functor
  - [x] Apply
  - [x] Applicative
  - [ ] FlatMap
  - [x] Monad
  - [x] Foldable
  - [x] Traverse
  - [ ] Reducible #1777 
  - [ ] NonEmptyTraverse #1787 
  - [ ] CoFlatMap
  - [ ] CoMonad
  - [ ] BiMonad
  - [ ] BiFoldable
  - [ ] BiTraverse
  - [ ] ApplicativeError
  - [ ] MonadError
  - [ ] CommutativeFlatMap
  - [ ] CommutativeMonad
  - [ ] InvariantMonoidal
  - [x] SemigroupK
  - [x] MonoidK
  - [ ] Alternative

  - [ ] Show #1789 
  - [ ] Inject
  - [ ] InjectK
  - [ ] NotNull

- Arrow
  - [ ] Arrow
  - [ ] Category
  - [ ] Choice
  - [ ] CommutativeArrow
  - [ ] Compose

## Data types  

- [x] Const
- [x] Kleisli
- [ ] Cokleisli
- [ ] Id
- [ ] EitherK
- [ ] Tuple2K
- [ ] EitherT
- [x] OptionT
- [x] State
- [ ] StateT
- [x] OneAnd
- [ ] NonEmptyList
- [ ] NonEmptyVector
- [x] Validated
- [x] FunctionK
- [x] Either
- [ ] Ior
- [ ] WriterT
- [ ] ReaderWriterStateT
[...]

PR #1924, opened 2017-09-20
Add docs for Arrow

zliu41:
issue: #1801

Initial draft for `Arrow` documentation. Did this as part of the Scala World open source spree.


docs/src/main/tut/typeclasses/arrow.md
```diff
@@ -0,0 +1,119 @@
+---
+layout: docs
+title:  "Arrow"
+section: "typeclasses"
+source: "core/src/main/scala/cats/arrow/Arrow.scala"
+scaladoc: "#cats.arrow.Arrow"
+---
+# Arrow
+
+`Arrow` is a useful type class for modeling something that behaves like functions, such as `Function1`: `A => B`, `Kleisli`: `A => F[B]` (also known as `ReaderT`), `Cokleisli`: `F[A] => B`, etc. So useful, that Haskell provides special syntax (the `proc` notation) for composing and combining Arrows, similar as the `do` notation for sequencing monadic operations.
```
LukaJCB: I think "similar to" would be better than similar as :)

docs/src/main/tut/typeclasses/arrow.md
```diff
@@ -0,0 +1,119 @@
...
+
+def id[A]: F[A, A]
+
+def compose[A, B, C](f: F[B, C], g: F[A, B]): F[A, C]
+
+def first[A, B, C](fa: F[A, B]): F[(A, C), (B, C)]
+```
+
+Once a type `F` has an `Arrow` instance, it gets for free a number of methods for composing and combining with other `Arrow`s. You will be able to do things like:
+- Composing `fab: F[A, B]` and `fbc: F[B, C]` into `fac: F[A, C]` with `fab >>> fbc`. If `F` is `Function1` then `>>>` becomes alias of `andThen`.
```
LukaJCB: I think it should be "becomes an alias for" :)

docs/src/main/tut/typeclasses/arrow.md
```diff
@@ -0,0 +1,119 @@
...
+    b :: runList(ff2, t)
+  case _ => List()
+}
+```
+
+Now we can try it out:
+
+```tut:book:silent
+runList(accum[Int, Int](0)(_ + _), List(6, 5, 4, 3, 2, 1))
+// res1: List[Int] = List(6, 11, 15, 18, 20, 21)
```
LukaJCB: Can we not make this non-silent? Then we could remove the comment, or am I missing something?

docs/src/main/tut/typeclasses/arrow.md
```diff
@@ -0,0 +1,119 @@
...
+
+def avg: FancyFunction[Int, Double] =
+  combine(sum[Int], count[Int]) >>> Arrow[FancyFunction].lift{case (x, y) => x.toDouble / y}
+```
+
+And try it out:
+
+```tut:book:silent
+runList(avg, List(1, 10, 100, 1000))
+// res2: List[Int] = List(1.0, 5.5, 37.0, 277.75)
```
LukaJCB: Same as above

docs/src/main/tut/typeclasses/arrow.md
```diff
@@ -0,0 +1,119 @@
...
+
+As an example, let's create a fancy version of `Function1` that is capable of maintaining states. Let's call it `FancyFunction`:
+
+```tut:book:silent
+case class FancyFunction[A, B](run: A => (FancyFunction[A, B], B))
+```
+
+That is, given an `A`, it not only returns a `B`, but also returns a new `FancyFunction[A, B]`. This sounds similar as the `State` monad (which returns a result and a new `State` from an initial `State`), and indeed, `FancyFunction` can be used to perform stateful transformations.
+
+Here's an `Arrow` instance for `FancyFunction`:
```
julienrf: Would you mind giving an example of use of `FancyFunction` before showing the `Arrow` instance implementation?

docs/src/main/tut/typeclasses/arrow.md
zliu41: Fixed. I can be pretty bad at choosing the right preposition :)

docs/src/main/tut/typeclasses/arrow.md
zliu41: You are right, I'm not super familiar with tut.

docs/src/main/tut/typeclasses/arrow.md
zliu41: Yep. There's now an example of how to use `avg`, before showing the Arrow instance for `FancyFunction` and explaining how to implement `avg`.

docs/src/main/tut/typeclasses/arrow.md
```diff
@@ -0,0 +1,118 @@
+---
+layout: docs
+title:  "Arrow"
+section: "typeclasses"
+source: "core/src/main/scala/cats/arrow/Arrow.scala"
+scaladoc: "#cats.arrow.Arrow"
+---
+# Arrow
+
+`Arrow` is a useful type class for modeling something that behaves like functions, such as `Function1`: `A => B`, `Kleisli`: `A => F[B]` (also known as `ReaderT`), `Cokleisli`: `F[A] => B`, etc. So useful, that Haskell provides special syntax (the `proc` notation) for composing and combining arrows, similar to the `do` notation for sequencing monadic operations.
```
ceedubs: I think that we should avoid bringing Haskell into the explanation this early in the docs (if at all). I know that for people who are interested in functional programming, Haskell is the go-to language to describe concepts, but in my general experience people who already know haskell will already be familiar with concepts such as arrows, and mentioning Haskell won't be helpful for people who don't already know it.

docs/src/main/tut/typeclasses/arrow.md
```diff
@@ -0,0 +1,118 @@
...
+
+To create an `Arrow` instance for a type `F[A, B]`, the following abstract methods need to be implemented:
```
ceedubs: It might be helpful to show a few examples of using `Arrow` methods on instances that already exist (such as `Function1` before diving into what's needed to create your own custom `Arrow` instance. It's very rare that you'll be creating your own custom `Arrow` instance, but more common that people will want to take advantage of some of the methods on existing types.

docs/src/main/tut/typeclasses/arrow.md
```diff
@@ -0,0 +1,118 @@
...
+def id[A]: F[A, A]
+
+def compose[A, B, C](f: F[B, C], g: F[A, B]): F[A, C]
+
+def first[A, B, C](fa: F[A, B]): F[(A, C), (B, C)]
+```
+
+Once a type `F` has an `Arrow` instance, it gets for free a number of methods for composing and combining multiple `Arrow`s. You will be able to do things like:
+- Composing `fab: F[A, B]` and `fbc: F[B, C]` into `fac: F[A, C]` with `fab >>> fbc`. If `F` is `Function1` then `>>>` becomes an alias for `andThen`.
+- Taking two arrows `fab: F[A, B]` and `fbc: F[B, C]` and combining them into `F[(A, C) => (B, D)]` with `fab *** fbc`. The resulting arrow takes two inputs and processes them with two arrows, one for each input.
```
ceedubs: This is related to the previous comment, but I think that some concrete examples might help these explanations.

docs/src/main/tut/typeclasses/arrow.md
zliu41: Good point.

zliu41:
Updated. Added examples for `Function1` and `Kleisli` in addition to the original examples :)

docs/src/main/tut/typeclasses/arrow.md
```diff
@@ -0,0 +1,179 @@
+---
+layout: docs
+title:  "Arrow"
+section: "typeclasses"
+source: "core/src/main/scala/cats/arrow/Arrow.scala"
+scaladoc: "#cats.arrow.Arrow"
+---
+# Arrow
+
+`Arrow` is a type class for modeling something that behaves like functions. For example, `scala.Function1`(`A => B`), `cats.data.Kleisli`(wrapping an `A => F[B]`, also known as `ReaderT`), and `cats.data.Cokleisli`(wrapping an `F[A] => B`), have `Arrow` instances.
```
kailuowang: Sorry I am late in the game.  This is kind of a nit but I am not so sure about this very first sentence. 

> `Arrow` is a type class for modeling something that behaves like functions.

I kind of feel that this is not the original intention of the type class. `Arrow` is, well, arrow in category theory. function is just one of the possible definitions.  So a more precise way to put it might be 
```markdown
`Arrow` is a type class for modeling composable relationship between two types. One example of such composable relationship is function `A => B`, other examples includes `cats.data.Kleisli`(wrapping an `A => F[B]`, also known as `ReaderT`), and `cats.data.Cokleisli`(wrapping an `F[A] => B`), have `Arrow` instances.
```
WDYT?

docs/src/main/tut/typeclasses/arrow.md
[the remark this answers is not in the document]
zliu41: This was fixed in the 2nd commit. First time using tut..

docs/src/main/tut/typeclasses/arrow.md
zliu41: sgtm. Thanks for the correction.

zliu41:
Rewritten the first paragraph based on @kailuowang's suggestion.

zliu41:
Would like to thank @LukaJCB, @julienrf, @ceedubs and @kailuowang for the feedback, and @peterneyens for helping me get started at Scala World. Looking forward to making more contributions to Cats in the future.

PR #1925, opened 2017-09-21
temporarily disable MiMa check on kernel

kailuowang:
temporarily disable to allow a couple of approved PRs (#1712 #1527) to pass build, should reenable after 1.0.0-RC1.
Also added `QueueInstances` and `DurationInstances` to `all` imports

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -231,7 +231,18 @@ lazy val macros = crossProject.crossType(CrossType.Pure)
 lazy val macrosJVM = macros.jvm
 lazy val macrosJS = macros.js
 
-val binaryCompatibleVersion = "1.0.0-RC1"
+val binaryCompatibleVersion = "0.8.0"
+
+val binaryCompatibleExceptions = {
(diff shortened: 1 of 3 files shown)


johnynek:
can we not totally disable, and instead as people to add expections?

This prevents us from going hog-wild breaking things.

You can find notes:
https://github.com/typesafehub/migration-manager/wiki/Sbt-plugin#advanced-usage-filtering-binary-incompatibilities

kailuowang:
ah, then it probably makes more sense for those two PRs to add their own exceptions. I am going to add the exception for the `all` instance here.

johnynek:
yeah, I think each PR should add its own exception.

kailuowang:
updated. @johnynek

johnynek:
👍 

This is great!

kailuowang:
@LukaJCB do yo mind take a quick look at this one? would be nice to merge so other PRs can follow example without conflicts in the build.sbt.

PR #1919, opened 2017-09-19
update to sbt 1.0

kailuowang:
fixes #1848 
~I had to disable coursier~
~it's giving me~
>sbt:Cats core> compile
[error] scala.MatchError: Patch() (of class sbt.librarymanagement.Patch)
[error] 	at coursier.FromSbt$.sbtCrossVersionName(FromSbt.scala:31)
[error] 	at coursier.FromSbt$.sbtModuleIdName(FromSbt.scala:21)
[error] 	at coursier.FromSbt$.moduleVersion(FromSbt.scala:47)
[error] 	at coursier.FromSbt$.dependencies(FromSbt.scala:63)
...

Update: Coursier fixed.

PR #1926, opened 2017-09-23
Update coursier to version 1.0.0-RC12

mxl:
I can't build cats because of https://github.com/coursier/coursier/issues/655.


ceedubs:
Thanks @mxl!

This was also done in #1919, but mixed with an upgrade to SBT 1.0 which currently presents some issues with building on scala 2.13. It might make sense to merge this change here while we figure out the right path forward on #1919. WDYT @kailuowang @fthomas?

Issue #1831, opened 2017-08-21
Foldable.toList potentially inconsistent for Map


julien-truffaut:
```
test("toList") {
    import cats.Foldable
    val x = Map(1 -> "abc", 2 -> "cde")
    val y = Map(2 -> "cde", 1 -> "abc")

    x === y should be (true)

    val F = Foldable[Map[Int, ?]]

    F.toList(x) should be (List("abc", "cde"))
    F.toList(y) should be (List("abc", "cde")) // fail, it returns List("cde", "abc")
  }
```

The problem comes from `Map.toList` which is used by cats. As far as I know scalaz side step this issue because it requires an `Order` on `Map` keys.

You can also see this problem when you use traverse:

```scala
test("traverse - toList") {
    import cats.{Foldable, Traverse, Id}
    val x = Map(1 -> "abc", 2 -> "cde")

    val F = Foldable[Map[Int, ?]]
    val T = Traverse[Map[Int, ?]]

    F.toList(T.traverse[Id, String, String](x)(_.reverse)) should be (List("cba", "edc"))  // fail, it returns List("cde", "abc")
  }
```

johnynek:
This same issue would show up for `Foldable[Set]` too, no?

julien-truffaut:
@johnynek I suppose so

tpolecat:
I don't know how we can possibly get congruence here since you can trivially distinguish values that are ostensibly equal. Universal properties FTL yet again. Do we punt and say yeah that's the risk you take with non-parametric data types, or remove the instances for `Map` and `Set` and only provide them for `Tree*`? Or something else behind door number 3?

I'd like to push past this as it seems to be the only thing holding up the Monocle port.

adelbertc:
I have a selfish preference for the `Order` constraint since I have found the `Traverse[Map[K, ?]]` instance useful several times, and really anything you use as keys in a `Map` should have an `Order`.. right?

Though I guess we can put it in `alleycats`.

tpolecat:
Are you suggesting we constrain the instance to `K: Order` and pre-sort the traversal? That's an expensive operation to sweep under the rug … I think I'd prefer to add instances for `TreeMap` and punt this one to alleycats, although subtyping is going to make the instances ambiguous I think.

:-\

adelbertc:
Oh I totally forgot about `TreeMap`. I like that.

julien-truffaut:
[...]

PR #1927, opened 2017-09-24
Add CommutativeApply and CommutativeApplicative

LukaJCB:
Had a quick discussion about this on gitter. Also adds an instance for `Validated[E: CommutativeSemigroup, ?]` :)


core/src/main/scala/cats/CommutativeApplicative.scala
```diff
@@ -0,0 +1,14 @@
...
+/**
+  * Commutative Applicative.
+  *
+  * Further than an Applicative, which just allows composition of independent effectful functions,
+  * in a Commutative Applicative those functions can be composed in any order, which guarantees
+  * that their effects do not interfere.
+  *
+  * Must obey the laws defined in cats.laws.CommutativeApplicativeLaws.
+  */
+@typeclass trait CommutativeApplicative[F[_]] extends Applicative[F] with CommutativeApply[F]
```
LukaJCB: For some reason Codecov doesn't think this is being hit, does anyone know why?

johnynek:
This seems like a very interesting property (one that Option springs to mind as a canonical example), but chat can I do with an instance?

Can we not write a method or two to show the value of the abstraction. I tend to think to make it into cats an abstraction needs to pay its way. I haven’t yet seen that here, but probably because I am unfamiliar.

tpolecat:
`Future` is another effect with this property. The order in which you *construct* futures is significant if you're side-effecting, but the order of applicative composition doesn't matter. Also the "sbt effect": the rewrite macro pulls the out the `.value` in an arbitrary order and it doesn't matter.

You could use this property to properly constrain general rewrite systems like monadless (which preserves equational reasoning only for commutative monads).

None of these observations leads us closer to an easy example though. @edmundnoble mentioned that it might be handy for optimizing free applicatives since you could re-order effects.

edmundnoble:
@johnynek Given the ability to reorder computations in FreeApplicative, you can batch them safely.

`httpGet(url1) *> httpGet(url2) *> httpGet(url1)`
====>
`httpGet(url1) <* httpGet(url2)`

Also, unordered containers can be traversed safely with `CommutativeApplicative`, e.g. `Set` and `Map`. Right now, if you want to traverse them, you are forced into observing the implementation-defined ordering, and violating the contract of equality.

johnynek:
those are good examples.

Could we add the methods to this PR?

LukaJCB:
Hmm, so as I see we could either create some sort of `traverseUnordered` as syntax sugar on `Set` and `Map`, or we could want to create a type class for such unordered containers? 

Leaning towards the former, as the latter might open another larger can of worms (Would a `TraverseUnordered` inherit from a theoretical non-structure preserving `Functor`?)

LukaJCB:
Okay, maybe `Future` is not a valid `CommutativeApplicative` because the ordering matters if one of the Futures is a `Failure`, so it only works for Successful cases.

tpolecat:
@LukaJCB What was the issue with `Future`? It should be a perfectly cromulent instance as long as you're not side-effecting.

LukaJCB:
Yeah, but only if the result is successful:

```
scala> val a: Future[Int] = Future.failed(new Exception("A"))
a: scala.concurrent.Future[Int] = Future(Failure(java.lang.Exception: A))

scala> val b: Future[Int] = Future.failed(new Exception("B"))
b: scala.concurrent.Future[Int] = Future(Failure(java.lang.Exception: B))

scala> a *> b
res0: scala.concurrent.Future[Int] = Future(Failure(java.lang.Exception: A))

scala> b *> a
res1: scala.concurrent.Future[Int] = Future(Failure(java.lang.Exception: B))
```

tpolecat:
@LukaJCB hell. Can't ever win with `Future`. Good counterexample, thanks!

LukaJCB:
You really can't 😄 . Want to weigh in on syntax sugar vs type class? :)

LukaJCB:
Another cool thing is that `CommutativeApplicative` also composes using `Nested` and `Tuple2K`

core/src/main/scala/cats/data/Nested.scala
```diff
@@ -135,13 +135,27 @@ private[data] sealed abstract class NestedInstances9 extends NestedInstances10 {
...
-private[data] sealed abstract class NestedInstances10 {
+private[data] sealed abstract class NestedInstances10 extends NestedInstances11 {
   implicit def catsDataInvariantForNestedContravariant[F[_]: Invariant, G[_]: Contravariant]: Invariant[Nested[F, G, ?]] =
     new NestedInvariant[F, G] {
       val FG: Invariant[λ[α => F[G[α]]]] = Invariant[F].composeContravariant[G]
     }
 }
 
+private[data] sealed abstract class NestedInstances11 extends NestedInstances12 {
+  implicit def catsDataCommutativeApplyForNestedContravariant[F[_]: CommutativeApply, G[_]: CommutativeApply]: CommutativeApply[Nested[F, G, ?]] =
```
kailuowang: Usually we place more specific instance higher priority, i.e. `CommutativeApply` and `CommutativeApplicative` should have higher priority than `Apply` and `Applicative`. 
`CommutativeApplicative` should have higher priority than `CommutativeApply`. 

Also line 53 the instance name was wrong `catsDataContravariantForNested` -> `catsDataFunctorForNested` do you mind fix that in this PR as well?

core/src/main/scala/cats/CommutativeApplicative.scala
```diff
@@ -0,0 +1,41 @@
...
+
+
+object CommutativeApplicative {
+
+  def traverseUnordered[F[_]: CommutativeApplicative, A, B](sa: Set[A])(f: A => F[B]): F[Set[B]] =
```
johnynek: Shouldn’t we put these on the trait on the chance they can be optimized for some instances?

LukaJCB: Good point!

core/src/main/scala/cats/data/Validated.scala
```diff
@@ -374,6 +375,20 @@ private[data] sealed abstract class ValidatedInstances1 extends ValidatedInstanc
       def combine(x: Validated[A, B], y: Validated[A, B]): Validated[A, B] = x combine y
     }
 
+  implicit def catsDataCommutativeApplicativeForValidated[E: CommutativeSemigroup]: CommutativeApplicative[Validated[E, ?]] =
```
kailuowang: Is it possible to extract this and `ApplicativeError` to a common super trait?

LukaJCB: You mean adding a `CommutativeApplicativeError` typeclass? Or just reshuffling the instances?

kailuowang: not a new type class, just a new private trait for duplicated code in the two instances.

LukaJCB: Ah of course, will do! :)

kailuowang:
LGTM except for that nit comment.

core/src/main/scala/cats/CommutativeApplicative.scala
```diff
@@ -0,0 +1,36 @@
...
+/**
+  * Commutative Applicative.
+  *
+  * Further than an Applicative, which just allows composition of independent effectful functions,
+  * in a Commutative Applicative those functions can be composed in any order, which guarantees
+  * that their effects do not interfere.
+  *
+  * Must obey the laws defined in cats.laws.CommutativeApplicativeLaws.
+  */
+@typeclass trait CommutativeApplicative[F[_]] extends Applicative[F] with CommutativeApply[F] {
```
johnynek: I wonder if we really want an `UnorderedFoldable` and `UnorderedTraverse` that instead of `toList` has `toSet` and has a `foldMap` that takes a `CommutativeMonoid` and for `sequence/traverse` uses `CommutativeApplicative`.

That seems cleaner to me than putting these methods here now that I think about it.

I think we can construct sane laws for `UnorderedFoldable` that would make sense. If we added a method like `insert[T](f: F[T], item: T): F[T]` then I think you could make some law.

LukaJCB: I agree that an extra type class would feel a lot cleaner than baking these 4 operations into the `CommutativeApplicative` type class. Related thoughts [here](https://github.com/typelevel/cats/issues/1831#issuecomment-337221498).
I've played around with some laws, and we wouldn't have to get rid of too many laws too make it fit. I also like the `insert`, although I'd have to think about it more. We could let `Foldable` extend `UnorderedFoldable`, if that makes any sense. I'll play around with it and update the PR accordingly :)

LukaJCB: This opens up a can of worms though. We could probably also define something like `unorderedFoldM` with `CommutativeMonad`, a `UnorderedReducible`, `NonEmptyUnordered` and all kinds of others. I came up with something crazy a while ago on gitter https://gitter.im/typelevel/cats-dev?at=59cf81a5210ac269208e0bcf but seems I've only scratched the surface :smile: 

I'll try to define this as minimal as possible and go from there. Might make a good blog post though.

LukaJCB: I had to remove `insert` for now, because it's not possible to implement it for `Map` :/

kailuowang:
I have a proposal:
 `CommunitiveApply` and `CommutativeApplicative` are obviously useful, straightforward and clearly defined. Can we limit this PR to these two type classes (without the `unorderedTraverseXXX` methods) and move all the unordered traverse stuff to a separate follow up PR?

LukaJCB:
Yeah I can do that :)

tests/src/test/scala/cats/tests/MapSuite.scala
```diff
@@ -15,6 +16,28 @@ class MapSuite extends CatsSuite {
   checkAll("Map[Int, Int] with Option", TraverseTests[Map[Int, ?]].traverse[Int, Int, Int, Int, Option, Option])
   checkAll("Traverse[Map[Int, ?]]", SerializableTests.serializable(Traverse[Map[Int, ?]]))
 
+  test("traverseUnordered identity") {
```
kailuowang: you forgot to remove these tests

LukaJCB:
Moved the unordered stuff to #1981

kailuowang:
thanks so much! @LukaJCB

kailuowang:
@johnynek this PR is simplified with only the Communtative type classes. Any other comments/feedbacks? Shall we merge this and then we can move to #1981

PR #1933, opened 2017-09-26
Add syntax for ApplicativeError.fromEither

tpolecat:
This adds syntax for `ApplicativeError.fromEither` as [discussed on gitter](https://gitter.im/typelevel/cats?at=59caa574614889d47540eba7).


johnynek:
nit: I would probably have rather this been `pureOrRaise` since the `Or` is usually used in to represent the failure case (at least thinking of `Option[T]`).

Issue #1930, opened 2017-09-25
Restructure `arrow` and `functor` package. 


kailuowang:
1. move `arrow` to a separate module.
2. Move `Profunctor` and `Strong` to this module `arrow` 
3. Move `Bifunctor`, `Invariant` and `Contravariant` to `core`

kailuowang:
Please vote if you, like me, think it's mostly bikeshedding. But if you have stronger feeling to put down, please do, it will be helpful for sure.

tpolecat:
If you agree with my organizing principle for the [infographic](https://github.com/tpolecat/cats-infographic) the outliers are pretty evident I think.

kailuowang:
looks like the decision can be made. I am moving this issue to "ready"

kailuowang:
working on a PR.

kailuowang:
`arrow.FunctionK` depends on `cats.core`, I incline to keep it in cats.core but it means that I need to move it to the `cats` root package.  cc @sellout.  Speaking of `FunctionK` its documentation is in the data types section, but I am not sure if that's accurate.

sellout:
@kailuowang Having `FunctionK` in core sounds good to me.

LukaJCB:
I think adding it to data does make some sense, as we have mostly type classes in `cats` . Don't have any strong opinion either way though.

kailuowang:
the part I am not sure about is if `FunctionK` is a data type. Anyway moving arrow to a module is more complicated than I thought, since `Category` and `Compose` depends on `SemigroupK` and `MonoidK`, so more triage needed for that one. I'll submit one PR for restructuring the `functor` package first.

LukaJCB:
Well, I'd say it's as much a data type as Kleisli and cokleisli, no? I think we could put everything that's not a type class into data :D

sellout:
@kailuowang 

1. I think having `FunctionK` in core/data is the right place
2. It should be fine for the `arrow` module to depend on the `core` module, no?

But yes, get rid of `functor` first 👍🏾

kailuowang:
If I am not mistaken, Core depends on arrow, so circular intermodule dependency? I am okay with functionK going too data

sellout:
@kailuowang Ah, yeah – it looks like the syntax stuff depends on arrow. That should be pulled out too if arrow is a separate module. But then I guess we’d have `cats.arrow.implicits`?

[...]

PR #1934, opened 2017-09-26
Restruct functor

kailuowang:
Big but rather mechanical changes according to #1930
1. Move `Profunctor` and `Strong` to the `arrow` package
1. Move `Bifunctor`, `Invariant` and `Contravariant` to root package


tests/src/test/scala/cats/tests/EitherKTests.scala
```diff
@@ -1,9 +1,8 @@
 package cats.tests
 
-import cats._
+import cats.{Contravariant, _}
```
edmundnoble: ?

PR #1942, opened 2017-09-29
add traverseN to cartesian syntax

julien-truffaut:


kailuowang:
Shall we change title to "add traverseN to cartesian syntax", which is what it does. `CartesianBuilder` is deprecated.

julien-truffaut:
I am not sure what does it mean:
```
[error] /home/travis/build/typelevel/cats/tests/src/test/scala/cats/tests/SyntaxTests.scala:147: discarded non-Unit value
[error]     (fa, fb, fc) traverseN ft
```

tests/src/test/scala/cats/tests/SyntaxTests.scala
```diff
@@ -137,6 +140,11 @@ object SyntaxTests extends AllInstances with AllSyntax {
 
     val fga = mock[F[G[A]]]
     val gunit: G[F[A]] = fga.sequence
+
+    val ft = mock[(A, B, C) => G[Z]]
+
+    tfabc traverseN ft
+    (fa, fb, fc) traverseN ft
```
kailuowang: Either add a `()` at the end of the method  or make it a statement by adding a `val blah = (fa, fb, fc) traverseN ft` which is a ()

PR #1946, opened 2017-10-03
Give higher priority to partial-unification fix

LukaJCB:
After answering questions about relating to partial-unification, I think we should reprioritize its standing in the Readme.md. This PR mentions the fix up front instead of a side-note further down below. WDYT? :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,11 @@ For more detail about Cats' motivations, go [here](http://typelevel.org/cats/mot
 
 Cats is currently available for Scala 2.10, 2.11 and 2.12, and [Scala.js](http://www.scala-js.org/).
 
-To get started with SBT, simply add the following to your `build.sbt`
+In order for Cats to work correctly with regards to type inference you'll sometimes need support for improved type inference.
+We strongly recommend you to enable improved type inference using `partial-unification`. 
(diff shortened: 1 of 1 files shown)


tpolecat:
This is an improvement but I think I would probably still miss it ... when I look at a README like this I tend to gloss over everything other than the code blocks. Don't know if anyone else does this. I think I would be more likely to notice something like


> Cats relies on improved type inference via the fix for [SI-2112](https://issues.scala-lang.org/browse/SI-2712), which is not enabled by default. For **Scala 2.11.9 or later** you should add the following to your build:
> 
>     scalacOptions += "-Ypartial-unification"
>
> **Or**, if you need to support older versions of Scala you can use the [sbt-partial-unification](https://github.com/fiadliel/sbt-partial-unification#sbt-partial-unification) plugin which extends support back through **Scala 2.10.6 or later**:
>
>     addSbtPlugin("org.lyranthe.sbt" % "partial-unification" % "1.1.0")

kailuowang:
Good idea @tpolecat

LukaJCB:
Thanks @tpolecat, I definitely agree with your suggestion! :) 
I added something along the lines, WDYT?

PR #1949, opened 2017-10-04
Add ::: to NonEmptyList

jcranky:


kailuowang:
if you want, you can add a doctest to make the test coverage happy.

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -82,6 +82,9 @@ final case class NonEmptyList[+A](head: A, tail: List[A]) {
   def ::[AA >: A](a: AA): NonEmptyList[AA] =
     NonEmptyList(a, head :: tail)
 
+  def :::[AA >: A](other: NonEmptyList[AA]): NonEmptyList[AA] =
```
kailuowang: Actually to maintain symmetry, can we add `++:` to `NonEmptyVector` as well?

jcranky: Sure :)

jcranky:
Regarding doctest, where should I look at? I new to the cats codebase.

LukaJCB:
Doc test is a test in the Scaladoc right above the method itself :) our build pipeline we automatically check if the result is the same.

core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -66,6 +66,11 @@ final class NonEmptyVector[+A] private (val toVector: Vector[A]) extends AnyVal
   def ++[AA >: A](other: Vector[AA]): NonEmptyVector[AA] = concat(other)
 
   /**
+    * Append this NEV to another NEV, producing a new `NonEmptyVector`.
+    */
+  def ++:[AA >: A](other: NonEmptyVector[AA]): NonEmptyVector[AA] = other.concatNev(this)
```
LukaJCB: Shouldn't these be the other way around? `++` for two `NEV`s  and `++:` for an `NEV` and a `Vector`?

jcranky: I'm not sure, I went with this https://github.com/typelevel/cats/pull/1838#issuecomment-334344911 - perhaps we should have both for both, since overloading is happening already anyway?

jcranky:
@LukaJCB thanks for the explanation on doc test, I'll have a look

jcranky:
Tests added. This doctest thing is really cool, can you point me to somewhere where I can understand better how it works? Thanks!

LukaJCB:
Check out [sbt-doctest](https://github.com/tkawachi/sbt-doctest) :)

jcranky:
@LukaJCB thanks!

core/src/main/scala/cats/data/NonEmptyVector.scala
kailuowang: This is to mimic the API in `Vector` and `List`.  Let's address #1838 first

jcranky:
In the meanwhile, please let me know if I should do something else here :)

jcranky:
I guess we can now merge this as well? :)

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -83,6 +83,19 @@ final case class NonEmptyList[+A](head: A, tail: List[A]) {
...
+    *
+    * {{{
+    * scala> import cats.data.NonEmptyList
+    * scala> val nel = NonEmptyList.of(1, 2, 3)
+    * scala> nel ::: NonEmptyList.of(4, 5)
+    * res0: cats.data.NonEmptyList[Int] = NonEmptyList(1, 2, 3, 4, 5)
+    * }}}
+    */
+  def :::[AA >: A](other: NonEmptyList[AA]): NonEmptyList[AA] =
+    other.concat(this)
```
kailuowang: Shall we use concatNel here instead?  concat is deprecated

jcranky: sure

jcranky:
@kailuowang done

jcranky:
@kailuowang anything I should do here ? :)

kailuowang:
The build is broken

> /home/travis/build/typelevel/cats/core/src/main/scala/cats/data/NonEmptyList.scala:102: The link target "concat" is ambiguous. Several members fit the target

Instead of linking it using `[[]]`, just use `concat` should be fine.

jcranky:
@kailuowang ok, my bad, I was blind and didn't see the build was broken. Fixed :)

kailuowang:
@jcranky no worries at all. i probably should've pointed it out earlier.

jcranky:
@kailuowang now the build is ok :+1:  :dagger:

tpolecat:
It's not obvious to me what the use case is here. `a ::: b`/`a ++: b` are always equivalent to `a ++ b` right?

jcranky:
@tpolecat my main reasoning for this was to keep symmetry  with `List`. Plus, `++` / `++:` are not there, or are they?

jcranky:
(I found myself doing `.toList` a few times juts to use `:::`)

kailuowang:
@tpolecat  `++` was added to concat with `List` rather than NEL, for maintaining compatibility we didn't want to change that.

jcranky:
thanks :)

PR #1953, opened 2017-10-07
Add a more direct tailRecM law

johnynek:
I noticed you can implement flatMap from tailRecM and map. This gives a more direct law for flatMap rather than the limited `A, A => F[A]` iterated law that we have currently (since we can take `F[A]` and `A => F[B]` we have less structure on the input, so it is a stronger law.

I'd like to add this even though it shouldn't increase test coverage or currently find any bugs, it is a nice documentation of thinking about tailRecM which I continue to think of as a very important method for FlatMap/Monad when thinking about implementations which use bounded stack such as the scala on the jvm.


LukaJCB:
Merging with two sign-offs :)

PR #1955, opened 2017-10-07
Deprecate FlatMap's >> and << 

### LINKED ISSUE #1954: Deprecate `>>` and `<<` in favor of `*>` and `<*`
The former two don't really serve a purpose and are strictly the same when the Monad in question is consistent with the type's Applicative.

LukaJCB:
To do so, I moved `*>` and `<*` from `Cartesian` syntax to `Cartesian` type class. I also moved `followedBy` and `forEffect` to `Apply`. :)

Should fix #1954

### DIFF
--- a/core/src/main/scala/cats/Apply.scala
+++ b/core/src/main/scala/cats/Apply.scala
@@ -19,6 +19,14 @@ trait Apply[F[_]] extends Functor[F] with Cartesian[F] with ApplyArityFunctions[
   override def product[A, B](fa: F[A], fb: F[B]): F[(A, B)] =
     ap(map(fa)(a => (b: B) => (a, b)))(fb)
 
+  /** Sequentially compose two actions, discarding any value produced by the first. */
+  def followedBy[A, B](fa: F[A])(fb: F[B]): F[B] =
+    map(product(fa, fb)) { case (_, b) => b }
+
+  /** Sequentially compose two actions, discarding any value produced by the second. */
+  def forEffect[A, B](fa: F[A])(fb: F[B]): F[A] =
+    map(product(fa, fb)) { case (a, _) => a }
+
   /**
    * ap2 is a binary version of ap, defined in terms of ap.
    */
--- a/core/src/main/scala/cats/Apply.scala
+++ b/core/src/main/scala/cats/Apply.scala
@@ -19,13 +19,21 @@ trait Apply[F[_]] extends Functor[F] with Cartesian[F] with ApplyArityFunctions[
   override def product[A, B](fa: F[A], fb: F[B]): F[(A, B)] =
     ap(map(fa)(a => (b: B) => (a, b)))(fb)
 
-  /** Sequentially compose two actions, discarding any value produced by the first. */
+  /** Compose two actions, discarding any value produced by the first. */
   def followedBy[A, B](fa: F[A])(fb: F[B]): F[B] =
-    map(product(fa, fb)) { case (_, b) => b }
+    map2(fa, fb)((_, b) => b)
 
-  /** Sequentially compose two actions, discarding any value produced by the second. */
+  /** Alias for [[followedBy]]. */
+  @inline final def *>[A, B](fa: F[A])(fb: F[B]): F[B] =
+    followedBy(fa)(fb)
+
+  /** Compose two actions, discarding any value produced by the second. */
   def forEffect[A, B](fa: F[A])(fb: F[B]): F[A] =
-    map(product(fa, fb)) { case (a, _) => a }
+    map2(fa, fb)((a, _) => a)
+
+  /** Alias for [[forEffect]]. */
+  @inline final def <*[A, B](fa: F[A])(fb: F[B]): F[A] =
+    forEffect(fa)(fb)
 
   /**
    * ap2 is a binary version of ap, defined in terms of ap.
--- a/core/src/main/scala/cats/Cartesian.scala
+++ b/core/src/main/scala/cats/Cartesian.scala
@@ -14,12 +14,6 @@ import simulacrum.typeclass
  */
 @typeclass trait Cartesian[F[_]] {
   def product[A, B](fa: F[A], fb: F[B]): F[(A, B)]
-
-  @inline final def *>[A, B](fa: F[A])(fb: F[B])(implicit F: Functor[F]): F[B] =
-    F.map(product(fa, fb)) { case (_, b) => b }
-
-  @inline final def <*[A, B](fa: F[A])(fb: F[B])(implicit F: Functor[F]): F[A] =
-    F.map(product(fa, fb)) { case (a, _) => a }
 }
 
 object Cartesian extends CartesianArityFunctions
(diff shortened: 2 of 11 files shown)


core/src/main/scala/cats/Apply.scala
```diff
@@ -19,6 +19,14 @@ trait Apply[F[_]] extends Functor[F] with Cartesian[F] with ApplyArityFunctions[
   override def product[A, B](fa: F[A], fb: F[B]): F[(A, B)] =
     ap(map(fa)(a => (b: B) => (a, b)))(fb)
 
+  /** Sequentially compose two actions, discarding any value produced by the first. */
+  def followedBy[A, B](fa: F[A])(fb: F[B]): F[B] =
+    map(product(fa, fb)) { case (_, b) => b }
```
johnynek: Shouldn’t we use map2 here since that can sometimes be more efficient than product + map

core/src/main/scala/cats/Cartesian.scala
```diff
@@ -14,6 +14,12 @@ import simulacrum.typeclass
  */
 @typeclass trait Cartesian[F[_]] {
   def product[A, B](fa: F[A], fb: F[B]): F[(A, B)]
+
+  @inline final def *>[A, B](fa: F[A])(fb: F[B])(implicit F: Functor[F]): F[B] =
+    F.map(product(fa, fb)) { case (_, b) => b }
```
johnynek: I don’t like this. If you have a Functor for Cartesian you have an Apply, no?

I’m -1 in accepting other type classes on F in methods for a typeclass of F in most cases.

johnynek: Also, I don’t see why these aren’t exactly symbols for the methods you moved to Apply.

core/src/main/scala/cats/syntax/cartesian.scala
```diff
@@ -17,10 +17,4 @@ abstract class CartesianOps[F[_], A] extends Cartesian.Ops[F, A] {
   final def |@|[B](fb: F[B]): CartesianBuilder[F]#CartesianBuilder2[A, B] =
     new CartesianBuilder[F] |@| self |@| fb
 
-  final def *>[B](fb: F[B])(implicit F: Functor[F]): F[B] =
```
johnynek: Ahh. I see someone put them here originally. This was an error in my view. They should be on Apply.

docs/src/main/tut/faq.md
```diff
@@ -215,8 +215,6 @@ All other symbols can be imported with `import cats.implicits._`
 | `x === y`                        | equals                 |                  | `Eq[A]`                 | `eqv(x: A, y: A): Boolean`                                |
 | `x =!= y`                        | not equals             |                  | `Eq[A]`                 | `neqv(x: A, y: A): Boolean`                               |
 | `fa >>= f`                       | flatMap                |                  | `FlatMap[F[_]]`         | `flatMap(fa: F[A])(f: A => F[B]): F[B]`                   |
-| `fa >> fb`                       | followed by            |                  | `FlatMap[F[_]]`         | `followedBy(fa: F[A])(fb: F[B]): F[B]`                    |
-| `fa << fb`                       | for effect             |                  | `FlatMap[F[_]]`         | `forEffect(fa: F[A])(fb: F[B]): F[A]`                     |
```
johnynek: Should we not add them back to another place place in the doc?

free/src/test/scala/cats/free/FreeTests.scala
```diff
@@ -189,7 +189,7 @@ class FreeTests extends CatsSuite {
     forAll { (x: Int, y: Int) =>
       val expr1: Free[T, Int] = Free.injectRoll[T, Test1Algebra, Int](Test1(x, Free.pure))
       val expr2: Free[T, Int] = Free.injectRoll[T, Test2Algebra, Int](Test2(y, Free.pure))
-      val res = distr[T, Int](expr1 >> expr2)
+      val res = distr[T, Int](expr1 >>= (_ => expr2))
```
johnynek: Why won’t `*>` work here?

core/src/main/scala/cats/syntax/cartesian.scala
LukaJCB: Yeah +1 on that :)

core/src/main/scala/cats/Cartesian.scala
LukaJCB: I'll add these as aliases for those in apply :)

core/src/main/scala/cats/Apply.scala
LukaJCB: Yes that's a great idea

johnynek: also I think "sequentially" is confusing now. It is sequential if we have a monad, if we have an applicative it is not necessarily "sequential"

free/src/test/scala/cats/free/FreeTests.scala
LukaJCB: There was something odd going on with the tests so that simply replacing >> with *> did not work.

docs/src/main/tut/faq.md
LukaJCB: I added them at the bottom with a deprecated notice

johnynek:
Looks good to me except for coverage. I’m on the phone now but can we exercise each of the lines to make sure there isn’t any issue that we don’t notice?

Also, I wonder if this is really the last example where we have methods on the wrong type classes. Hard to catch except by audit.

core/src/main/scala/cats/syntax/flatMap.scala
```diff
@@ -11,6 +11,18 @@ trait FlatMapSyntax extends FlatMap.ToFlatMapOps {
 
   implicit final def catsSyntaxFlatMapIdOps[A](a: A): FlatMapIdOps[A] =
     new FlatMapIdOps[A](a)
+
+  implicit final def catsSyntaxFlatMapOps[F[_]: FlatMap, A](fa: F[A]): FlatMapOps[F, A] =
+    new FlatMapOps[F, A](fa)
```
LukaJCB: These are untested, because of deprecation

core/src/main/scala/cats/Apply.scala
```diff
@@ -19,6 +19,22 @@ trait Apply[F[_]] extends Functor[F] with Cartesian[F] with ApplyArityFunctions[
...
+  @inline final def *>[A, B](fa: F[A])(fb: F[B]): F[B] =
+    followedBy(fa)(fb)
+
+  /** Compose two actions, discarding any value produced by the second. */
+  def forEffect[A, B](fa: F[A])(fb: F[B]): F[A] =
+    map2(fa, fb)((a, _) => a)
+
+  /** Alias for [[forEffect]]. */
+  @inline final def <*[A, B](fa: F[A])(fb: F[B]): F[A] =
+    forEffect(fa)(fb)
```
LukaJCB: This is untested because `<*` and `<<` were before this PR.

johnynek:
:+1:

LukaJCB:
Merging with two sign-offs

PR #1956, opened 2017-10-08
Mention the right issue number of doom.

hrhino:
I'm as much a fan of Rush as the next person, but 2712 should be a famous number in its own right.

Also, scala uses GitHub now.

I was intending on making a cooler first PR, but I skimmed the README while IntelliJ did what it does and now I can't un-notice this.


adelbertc:
Doc change, merging with 1 👍

PR #1958, opened 2017-10-09
Add FlatMap#flatTap, a more principled version of the kestrel combinator.

hrhino:
It's a useful way of adding side-effects (as in, effects on the side) of a monadic computation.

I almost called it `flatTap`; I welcome opinions for or against either name.

### DIFF
--- a/core/src/main/scala/cats/FlatMap.scala
+++ b/core/src/main/scala/cats/FlatMap.scala
@@ -127,4 +127,24 @@ import simulacrum.typeclass
    * Implementations of this method should use constant stack space relative to `f`.
    */
   def tailRecM[A, B](a: A)(f: A => F[Either[A, B]]): F[B]
+
+  /**
+    * Apply a monadic function and discard the result while keeping the effect.
+    *
+    * {{{
+    * scala> import cats._, implicits._
+    * scala> Option(1).flatTap(_ => None)
+    * res0: Option[Int] = None
+    * scala> Option(1).flatTap(_ => Some("123"))
+    * res1: Option[Int] = Some(1)
+    * scala> def nCats(n: Int) = List.fill(n)("cat")
+    * nCats: (n: Int)List[String]
+    * scala> List[Int](0).flatTap(nCats)
+    * res2: List[Int] = List()
+    * scala> List[Int](4).flatTap(nCats)
+    * res3: List[Int] = List(4, 4, 4, 4)
+    * }}}
+    */
+  def flatTap[A, B](fa: F[A])(f: A => F[B]): F[A] =
+    flatMap(fa)(a => map(f(a))(_ => a))
 }


core/src/main/scala/cats/Monad.scala
```diff
@@ -109,4 +109,25 @@ import simulacrum.typeclass
...
+    * res1: Option[Int] = Some(1)
+    * scala> def nCats(n: Int) = List.fill(n)("cat")
+    * nCats: (n: Int)List[String]
+    * scala> List[Int](0).tapM(nCats)
+    * res2: List[Int] = List()
+    * scala> List[Int](4).tapM(nCats)
+    * res3: List[Int] = List(4, 4, 4, 4)
+    * }}}
+    */
+  def tapM[A, B](fa: F[A])(f: A => F[B]): F[A] =
```
johnynek: can't this be on `FlatMap` not `Monad`?

hrhino:
On October 9, 2017 4:21:20 PM EDT, "P. Oscar Boykin" <notifications@github.com> wrote:
>johnynek commented on this pull request.
>
>
>
>> +    * {{{
>+    * scala> import cats._, implicits._
>+    *
>+    * scala> Option(1).tapM(_ => None)
>+    * res0: Option[Int] = None
>+    * scala> Option(1).tapM(_ => Some("123"))
>+    * res1: Option[Int] = Some(1)
>+    * scala> def nCats(n: Int) = List.fill(n)("cat")
>+    * nCats: (n: Int)List[String]
>+    * scala> List[Int](0).tapM(nCats)
>+    * res2: List[Int] = List()
>+    * scala> List[Int](4).tapM(nCats)
>+    * res3: List[Int] = List(4, 4, 4, 4)
>+    * }}}
>+    */
>+  def tapM[A, B](fa: F[A])(f: A => F[B]): F[A] =
>
>can't this be on `FlatMap` not `Monad`?
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/typelevel/cats/pull/1958#pullrequestreview-68096884

Oh. Yeah, I guess it can. Thanks!

edmundnoble:
I vote for `flatTap` :D

core/src/main/scala/cats/FlatMap.scala
```diff
@@ -66,8 +66,18 @@ import simulacrum.typeclass
...
+    * {{{
+    * scala> import cats._, implicits._
+    * scala> List(1) forEffect Nil
+    * res0: List[Int] = List()
+    * scala> IO('a') forEffect IO(print("b\n")) unsafePerformIO ()
+    * b
+    * res1: Char = 'a'
+    * }}}
+    * */
+  def forEffect[A, B](fa: F[A])(fb: F[B]): F[A] = tapM(fa)(_ => fb)
```
johnynek: There is another PR in flight to move this to a Apply. I don’t think we need tapM here. Can we revert this change?

hrhino:
Cool. That's two for the ~~better~~ other name and zero opponents who have noticed that it might be changed.

PR #1960, opened 2017-10-10
Harmonize naming of discipline test classes in cats-kernel-laws.

denisrosset:
The naming scheme of the discipline test classes is inconsistent: `SemigroupLawTests` but `CommutativeSemigroupTests`, etc...

This is due to a name clash in the tests subproject. However, having consistent naming in `cats-kernel-laws` should take precedence, as we can always fix the imports by renaming them.

This is done in this PR.


kailuowang:
Thanks for fixing this.
@LukaJCB this is actually the way I meant to propose originally, you took the other direction but I didn't feel strong enough about it to drag the PR further.

peterneyens:
Thanks @denisrosset.
Should we actually rename them everywhere, instead of the import renames?

denisrosset:
Hi @peterneyens,

I'm going to move `spire`, and by extension `algebra` to the type of law descriptions used by cats. As `cats-kernel-laws` is going to be imported by these libraries, it is better to have a consistent naming scheme, while the `tests` package of cats is not going to be exported for external use.

As I'm not very much involved with cats, I have no opinion of what should be done outside of the `cats-kernel` project!

kailuowang:
I thought about renaming all  `discipline.XXXXTests` to `XXXXLawTests`, but it's kind of redundant with the `laws` in the package name. Another approach is to rename all the `tests.XXXXXTests` to something else, maybe `tests.XXXXXAuxTests` or something. BTW, we lack a consistent naming scheme for tests in `cats.tests` anyway, so maybe this can be addressed together with an effort of making them consistent.

For this PR,  it's merely to bring the tests in `kernel.laws` consistent within and with `cats.laws`, so I incline to merge this as is and address the conflict between `laws.discpline.XXXTests` and `tests.XXXTests` in a separate PR. What do you think @peterneyens ?

denisrosset:
Why not use e.g. `tests,MonoidSuite` instead of `MonoidAuxTests` or `MonoidTests`?

PR #1963, opened 2017-10-11
Move followedBy/forEffect tests to ApplyTests

peterneyens:
Move the `followedBy` and `forEffect` consistency tests from `FlatMapTests` to `ApplyTests`.

Small follow up on #1955.


johnynek:
:+1:

PR #1967, opened 2017-10-13
Add a few type classes to generated tuple instances (continued) 

kailuowang:
This is to finish up #1527 by 
a) reorder the instance according to priority
b) added Mima rule exceptions, these changes were approved by @johnynek


kailuowang:
Yes @edmundnoble I tested in REPL . There is a separated issue for adding tests for generated coded #1968. We'll tackle the tests there.

PR #1971, opened 2017-10-15
First approach to `NonEmptyList` docs

AlejandroME:
This is the first draft for `NonEmptyList` docs, as required by #1801.

I'm requesting feedback about this. What should we have to explore in this docs? I've seen the API of `NonEmptyList` and I find a lot of similarities with plain old `List` that I don't find useful to address here.

I appreciate your comments about this WIP and suggestions about what should we put in here.

Thanks!


AlejandroME:
@kailuowang and @LukaJCB: can you give me some advice on this? :)

LukaJCB:
Hey @AlejandroME Sorry, for not getting back to you earlier, I'm going to give you a review later today :)

docs/src/main/tut/datatypes/NonEmptyList.md
```diff
@@ -0,0 +1,62 @@
+---
+layout: docs
+title:  "NonEmptyList"
+section: "data"
+source: "core/src/main/scala/cats/data/NonEmptyList.scala"
+scaladoc: "#cats.data.NonEmptyList"
+---
+# NonEmptyList
+
+If you have had the opportunity of taking a look to [Validated](validated.html) or [IOR](ior.html), you'll find that one of the most common case is to use `NonEmptyList` with one of these data structures.
```
rossabaker: s/one of the most/a/

docs/src/main/tut/datatypes/NonEmptyList.md
```diff
@@ -0,0 +1,62 @@
...
+
+Why? Because it fits nicely in the error reporting cases. As its stated by its name, `NonEmptyList` is a _specialized_ data type that will have, at least, one element. You'll find its behaviour like a list, with the aforementioned constraint. Think of it as a `List` wrapper.
```
rossabaker: s/As its stated/As stated/

docs/src/main/tut/datatypes/NonEmptyList.md
```diff
@@ -0,0 +1,62 @@
...
+
+The signature of the structure is as follows:
+
+```scala
+final case class NonEmptyList[+A](head: A, tail: List[A]) {
+	// Implementation elided
+}
+```
+
+The `head` of the will be _non empty_, meanwhile, the `tail` can have zero or more elements.
```
rossabaker: s/non empty, meanwhile/non-empty. Meanwhile/

docs/src/main/tut/datatypes/NonEmptyList.md
```diff
@@ -0,0 +1,62 @@
...
+For example:
+
+```tut:book
+import cats.data.NonEmptyList
+
+val list = NonEmptyList.of("Hello")
+list.head
+```
+
+You've seen the creation of a `NonEmptyList` with one `String` element (_"Hello"_), using `.of` combinator, but also you can create a `NonEmptyList` from a regular `List` like this:
```
rossabaker: I wouldn't call `.of` a combinator.

It would be nice to show that `.of` takes a variadic tail, too: `NonEmptyList.of(1, 2, 3)`

docs/src/main/tut/datatypes/NonEmptyList.md
```diff
@@ -0,0 +1,62 @@
...
+
+```tut:book
+val regularList = List("a", "b", "c", "d")
+val nel = NonEmptyList.fromList(regularList)
+```
+
+But, what happen if you try to convert an empty `List` into a `NonEmptyList`? Let's see:
```
rossabaker: s/happen/happens/

docs/src/main/tut/datatypes/NonEmptyList.md
```diff
@@ -0,0 +1,62 @@
...
+```tut:book
+val emptyList = List.empty[String]
+val nel = NonEmptyList.fromList(emptyList)
+```
+
+As you can see, `.fromList` is type safe, giving you a `None` if you try to do the conversion. There's also an unsafe version of that combinator: `.fromListUnsafe` that gives you the `NonEmptyList` if the conversion is successful or an `IllegalArgumentException` if you give it an empty `List`.
+
+## All is about safety
+
+`List` has a lot of unsafe methods because it doesn't enforce any constraint (other than its type). If you do this:
```
rossabaker: Are we assuming that people know what "unsafe" means?

docs/src/main/tut/datatypes/NonEmptyList.md
```diff
@@ -0,0 +1,62 @@
...
+```tut:silent
+val emptyList = List.empty[String]
+```
+```tut:book:fail
+emptyList.head
+```
+
+You can see that you get a `NoSuchElementException` because you're trying to get the head of an empty `List`. The same behaviour applies with methods like `.last` or `.tail`.
+
+Using `NonEmptyList` you'll have the type safety in compile-time for this data type.
```
rossabaker: I think I might talk more about totality than type safety.  Neither `List[Orange].head` nor `NonEmptyList[Orange].head` will ever give you an apple.  It's just that the NEL will always give you an Orange.

docs/src/main/tut/datatypes/NonEmptyList.md
AlejandroME: Perfect! I'll expand this section

docs/src/main/tut/datatypes/NonEmptyList.md
AlejandroME: Can you explain me in more depth? I don't understand you when you say "totality".

docs/src/main/tut/datatypes/NonEmptyList.md
AlejandroME: Thank you. I'll fix all the typos that you've found.

docs/src/main/tut/datatypes/NonEmptyList.md
AlejandroME: Good point! I can use an example for illustrate the concept.

docs/src/main/tut/datatypes/NonEmptyList.md
LukaJCB: Total functions are functions that return a value for every input given. Partial functions (like `List#head`) don't have that guarantee and aren't defined for certain inputs.

AlejandroME: Ah, ok, got it! So, based on @rossabaker feedback, should I rephrase this section to address totality or should I leave this as it is and expand it with totality?

rossabaker: I think totality instead of type safety.

AlejandroME:
This is the second iteration of the documentation based on @rossabaker and @LukaJCB feedback.
Apart from the suggested improvements, I don't know what else to address in these docs, so, if there's a lack of some concept important to detail here, please let me know.

Thanks! :)

docs/src/main/tut/datatypes/NonEmptyList.md
```diff
@@ -0,0 +1,80 @@
...
+title:  "NonEmptyList"
+section: "data"
+source: "core/src/main/scala/cats/data/NonEmptyList.scala"
+scaladoc: "#cats.data.NonEmptyList"
+---
+# NonEmptyList
+
+If you have had the opportunity of taking a look to [Validated](validated.html) or [IOR](ior.html), you'll find that a common case is to use `NonEmptyList` with one of these data structures.
+
+Why? Because it fits nicely in the error reporting cases. As stated by its name, `NonEmptyList` is a _specialized_ data type that will have, at least, one element. You'll find that its behavior is like a list, with the aforementioned constraint. Think of it as a `List` wrapper.
```
rethab: What does specialized mean here? When I first read this, I thought about `@specialized` but I guess it's not? I think this might be confusing.

docs/src/main/tut/datatypes/NonEmptyList.md
```diff
@@ -0,0 +1,80 @@
...
+```
+
+The `head` of the `NonEmptyList` will be _non-empty_. Meanwhile, the `tail` can have zero or more elements.
+
+For example:
+
+```tut:book
+import cats.data.NonEmptyList
+
+val list = NonEmptyList.of("Hello")
```
rethab: Have you thought about adding types to your examples? I think this would demonstrate the difference between `of` and `fromList` more clearly:

```scala
val list: NonEmptyList[String] = NonEmptyList.of("Hello")

val nel: Option[NonEmptyList[String]] = NonEmptyList.fromList(regularList)
```

rethab:
As you said in the intro of this PR, `NonEmptyList` supports many operations that a conventional `List` does. Perhaps add a comment about that.

Also, you could add an example about `cats.syntax.list`. I like `groupByNel` in particular, because for me, the regular groupBy was a typical case where I knew it would only produce non-empty lists, but the types did not express that.

More ideas:
- reference/relate `NonEmptyTraverse`
- reference/relate `NonEmptyVector`

docs/src/main/tut/datatypes/NonEmptyList.md
AlejandroME: Sorry, I mean that is a _specific-purpose_ data type (compared to a regular `List`). I'll rephrase this.

AlejandroME:
Thank you @rethab for your review. I'll work on your suggestions and refine a little bit this docs.

kailuowang:
completed through #2107

PR #1972, opened 2017-10-16
Add `SortedMap` and `SortedSet` instances/Move `Set` and `Map` instances to Alleycats

### LINKED ISSUE #1966: Add Apply and Traverse for SortedMap
see #1964 for more discussion.

LukaJCB:
Should fix #1966


core/src/main/scala/cats/instances/sortedMap.scala
```diff
@@ -0,0 +1,173 @@
...
+      override def map2[A, B, Z](fa: SortedMap[K, A], fb: SortedMap[K, B])(f: (A, B) => Z): SortedMap[K, Z] =
+        if (fb.isEmpty) SortedMap.empty(Order[K].toOrdering) // do O(1) work if fb is empty
+        else ??? //fa.flatMap { case (k, a) => fb.get(k).map(b => (k, f(a, b))) }
+
+      override def map2Eval[A, B, Z](fa: SortedMap[K, A], fb: Eval[SortedMap[K, B]])(f: (A, B) => Z): Eval[SortedMap[K, Z]] =
+        if (fa.isEmpty) Eval.now(SortedMap.empty(Order[K].toOrdering)) // no need to evaluate fb
+        else fb.map(fb => map2(fa, fb)(f))
+
+      override def ap[A, B](ff: SortedMap[K, A => B])(fa: SortedMap[K, A]): SortedMap[K, B] =
+        ??? //fa.flatMap { case (k, a) => ff.get(k).map(f => (k, f(a))) }(scala.collection.breakOut)
```
kailuowang: still WIP, same below

LukaJCB: Yes, sorry. I meant to state so in the description 😅

core/src/main/scala/cats/instances/sortedMap.scala
```diff
@@ -0,0 +1,173 @@
...
+
+class SortedMapHash[K, V](implicit V: Hash[V], O: Order[K]) extends SortedMapEq[K, V]()(V, O) with Hash[SortedMap[K, V]] {
+  // adapted from [[scala.util.hashing.MurmurHash3]],
+  // but modified standard `Any#hashCode` to `ev.hash`.
+  import scala.util.hashing.MurmurHash3._
+  def hash(x: SortedMap[K, V]): Int = {
+    var a, b, n = 0
+    var c = 1;
+    x foreach { case (k, v) =>
+      // use the default hash on keys because that's what Scala's Map does
```
johnynek: I don't think we want to do that here actually. `Map[K, V]` needs to do this because the hash in question in a `Map[K, V]` is the `.hashCode`. But I think we can be unconstrained on a `SortedMap` since it is only using our Order typeclass.

core/src/main/scala/cats/instances/sortedMap.scala
```diff
@@ -0,0 +1,173 @@
...
+      override def ap2[A, B, Z](f: SortedMap[K, (A, B) => Z])(fa: SortedMap[K, A], fb: SortedMap[K, B]): SortedMap[K, Z] =
+        f.flatMap { case (k, f) =>
+          for { a <- fa.get(k); b <- fb.get(k) } yield (k, f(a, b))
+        }
+
+      def foldLeft[A, B](fa: SortedMap[K, A], b: B)(f: (B, A) => B): B =
+        fa.foldLeft(b) { case (x, (k, a)) => f(x, a)}
+
+      def foldRight[A, B](fa: SortedMap[K, A], lb: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B] =
+        Foldable.iterateRight(fa.values.iterator, lb)(f)
```
johnynek: I thought this was actually unsafe since it captures a mutable iterator? I think to make it safe, it needs to have ` () => Iterator[A]` which it can call to get a new one each time (which is another way of saying the `Iterable[A]` and it calls `.iterator` internally.

We had some bug related to this in the past. Not sure why we didn't fix `Foldable.iterateRight`. Actually, it was related, but not exactly the same. See #1716.

LukaJCB:
Damn there's no mutable `SortedMap` pre-Scala 2.12

LukaJCB:
So the tests fail for 2.11 and earlier for some reason. Will have to investigate

kailuowang:
@LukaJCB I submitted a PR against your branch https://github.com/LukaJCB/cats/pull/2
let me know if that sounds a good idea to you.

core/src/main/scala/cats/instances/sortedSet.scala
```diff
@@ -0,0 +1,101 @@
...
+
+  override def eqv(s1: SortedSet[A], s2: SortedSet[A]) = {
+    implicit val x = Order[A].toOrdering
+    s1.toStream.corresponds(s2.toStream)(Order[A].eqv)
+  }
+}
+
+class SortedSetHash[A: Order: Hash] extends Hash[SortedSet[A]] {
+  // TODO replace
+  def hash(x: SortedSet[A]): Int = x.hashCode()
```
johnynek: seems like we want an in order traversal and mix the values together.

tests/src/test/scala/cats/tests/FoldableSuite.scala
```diff
@@ -218,14 +218,6 @@ class FoldableSuiteAdditional extends CatsSuite {
     checkFoldMStackSafety[Vector](_.toVector)
   }
 
-  test("Foldable[Set].foldM stack safety") {
-    checkFoldMStackSafety[Set](_.toSet)
-  }
```
kailuowang: shall we replace these two foldMStackSafety tests with SortedSet and SortedMap ones?

tests/src/test/scala/cats/tests/ParallelTests.scala
```diff
@@ -30,7 +30,7 @@ class ParallelTests extends CatsSuite with ApplicativeErrorForEitherTest {
   }
 
   test("ParTraverse_ identity should be equivalent to parSequence_") {
-    forAll { es: Set[Either[String, Int]] =>
+    forAll { es: List[Either[String, Int]] =>
```
kailuowang: why not use `SortedSet`? (`Either` has `Order` instance)

tests/src/test/scala/cats/tests/RegressionSuite.scala
```diff
@@ -102,11 +101,6 @@ class RegressionSuite extends CatsSuite {
     Stream(1,2,6,8).traverse(validate) should === (Either.left("6 is greater than 5"))
     checkAndResetCount(3)
 
-    type StringMap[A] = Map[String, A]
-    val intMap: StringMap[Int] = Map("one" -> 1, "two" -> 2, "six" -> 6, "eight" -> 8)
-    intMap.traverse(validate) should === (Either.left("6 is greater than 5"))
-    checkAndResetCount(3)
```
kailuowang: Not sure what this regression test is about, do you know? If neither of us are sure maybe we should replace it with a sortedMap one?

LukaJCB: Apparently it's about https://github.com/typelevel/cats/issues/513. I had to adjust this test slightly, because the order of the Strings matter now, so it wouldn't work if we just change the type of the collection :)

LukaJCB:
Everything should be addressed now 🎉 Thanks everyone!

core/src/main/scala/cats/instances/map.scala
```diff
@@ -14,16 +14,8 @@ trait MapInstances extends cats.kernel.instances.MapInstances {
     }
 
   // scalastyle:off method.length
-  implicit def catsStdInstancesForMap[K]: Traverse[Map[K, ?]] with FlatMap[Map[K, ?]] =
-    new Traverse[Map[K, ?]] with FlatMap[Map[K, ?]] {
-
-      def traverse[G[_], A, B](fa: Map[K, A])(f: A => G[B])(implicit G: Applicative[G]): G[Map[K, B]] = {
```
johnynek: Can we add some law that this violates? I hate to remove it if we don’t actually have a law that it violates. I think we can write some (the original ticket had an example). I fear we could have other lawless instances without a law.

core/src/main/scala/cats/instances/set.scala
```diff
@@ -1,56 +1,17 @@
...
-  implicit val catsStdInstancesForSet: Foldable[Set] with MonoidK[Set] =
-    new Foldable[Set] with MonoidK[Set] {
+  implicit val catsStdInstancesForSet: MonoidK[Set] =
+    new MonoidK[Set] {
 
       def empty[A]: Set[A] = Set.empty[A]
 
       def combineK[A](x: Set[A], y: Set[A]): Set[A] = x | y
 
-      def foldLeft[A, B](fa: Set[A], b: B)(f: (B, A) => B): B =
```
johnynek: Same concern here about adding a law that this violates. Maybe a == b implies toList(a) == toList(b)

core/src/main/scala/cats/instances/map.scala
LukaJCB: Done :)

core/src/main/scala/cats/instances/set.scala
LukaJCB: We're going to need to spend some more time to find something. That one won't work, because that way e.g
`Either.Left(123) =!= Either.Left(-30)` would imply `List() =!= List()`

LukaJCB: I'm not sure if it's the best solution but I worked around it, by only considering the positive case. I.e `a == b` implies `toList(a) == toList(b)` but not `a != b` implies `toList(a) != toList(b)`

johnynek: This is what I meant (which is why I didn’t say (a == b) == (a.toList == b.toList))

johnynek:
This looks good to me. One question: do you know if this test would actually have caught Map/Set problems? I guess it must have since small maps/sets are just tuples and created in the order added.

LukaJCB:
Yes I tested these on master and they failed for set/map.

LukaJCB:
Although to be fair, I ran into some problems because it might not actually generate a situation where the tests fail every time because it's rare that they generate two different collections with the same values but different orderings.

kailuowang:
I couldn't think of a way to write the law so that will consistently fail on unordered data types either.  Another option we have is to simply remove the `orderedConsistency` test and simply document the fact that `Foldable` only supports ordered `F[_]`, for unordered `F[_]` use the new `UnorderedFoldable` and `UnorderedTraverse`, which we should probably do anyway.

kailuowang:
thanks so much for pushing this through

Issue #1973, opened 2017-10-16
Foldable.iterateRight is not sound


johnynek:
similar to #1716 I think `Foldable.iterateRight` will break if `Eval.always` are always used.

I think the fix is that we must accept an `Iterable[A]` rather than `Iterator[A]` and call `.iterator` internally. Basically add a `Eval.always(iter.iterator).flatMap { ... }` and I think it should be safe.

Of course, we need to demonstrate the problem with a failing test first.

LukaJCB:
Found a test that doesn't work:

```scala
forAll { (fa: List[Int]) =>
  val eval = Foldable.iterateRight(fa, Eval.later(0)) { (a, eb) =>
    Eval.always(a + eb.value)
  }
  eval.value should === (fa.sum)
  //Repeat here so the result is evaluated again
  eval.value should === (fa.sum)
}
```

LukaJCB:
Fixed by #1980

Issue #1800, opened 2017-08-08
Consider adding some form of newtypes to cats


LukaJCB:
We have https://github.com/alexknvl/newtypes, but we could also look at other options :)

alexknvl:
Once https://github.com/scala/scala/pull/6074 is merged in and there is a scalameta release with https://github.com/scalameta/paradise/pull/207 (maybe M10 already has it), that technique will be more usable. There are also some issues with IJ and newtypes with existential parameters at the moment. https://github.com/scala/scala/pull/6088 could improve the performance a bit (?).

LukaJCB:
Maybe we should wait out this: http://docs.scala-lang.org/sips/opaque-types.html

kailuowang:
Options that I see so far. 
1. depends on https://github.com/alexknvl/newtypes
1. depends on an adaption of https://github.com/estatico/scala-newtype which doesn't support type parameter as of now. @alexknvl [might be able to do some experiments with it.](https://gitter.im/typelevel/cats-dev?at=59c98b95cfeed2eb654d12c9)
1. fold a newtype project (one of above) into a cats module (could be cats.kernel or a new one) that other cats modules depends. 
1. wait for [opaque-types](http://docs.scala-lang.org/sips/opaque-types.html), basically it means no newtype for cats 1.0

kailuowang:
I am okay with either 3 or 4. 
2 is still a question mark. but generally speaking, if `cats.core` has depend on something it has to be something that we have control over binary compat, otherwise cats.core's bin compat promise becomes rather useless. so neither 1 or 2 should on the table.

edmundnoble:
I prefer 3. I'm not really okay with 4 myself. I think we're going to get stuck with a lot of things we don't want to be stuck with if we wait that long.

oscar-stripe:
I prefer 4 since we are in the midst of this currently unfolding.

Also, can we have some concrete proposals of where we want newtype in cats? For things like `EitherT` or `OptionT` or something? Can we actually make a benchmark that shows some win first? Usually one more allocation in a stack 2-deep of monads is not actually going to be a big problem.

I don't want to complicate things, nor consider dropping 2.10, if we don't have a concrete win to look at.

I added this in gitter:
```
the newtype macro is very new.
it would be safe IF:
1) the encoding is binary compatible from release to release. Which it may or may not be.
and
2) it leaves no runtime calls in the generated code that make visible changes to the class files.
[...]

PR #1974, opened 2017-10-16
Getting ready for RC1

### LINKED ISSUE #1832: NonEmptyVector.last

kailuowang:
RC1 is the only planned RC release. 1.0.0 would probably be within a month after (unless we found more bugs that neccessets 

Outstanding items (ones currently scheduled with RC1)

## PRs

Most of them only 1 maintainer approval short. 

- [x] Add SortedMap instances #1972  
- [x] Add a few type classes to generated tuple instances (continued)   #1967
- [x] Add CommutativeApply and CommutativeApplicative   #1927  
- [x] Add Invariant instances for kernel type classes #1914
- [x] Sync NEL and NEV methods (fixes #1832)   #1838
- [x] Add Parallel type class   #1837
- [x] Welcome, Alleycats #1984

## Issues

- [x] Foldable.toList potentially inconsistent for Map #1831
  - [x] Move traverse, foldable instance for Map and Set to alleycats, add laws (no ticket yet, pending #1984)
- [x] Foldable.iterateRight is not sound API  #1973 
- [ ] Kliesli is not stack-safe on left-associated binds #1733
- ~[ ] ??? Consider adding some form newtypes to cats #1800~

??? If we can't reach consensus on it, we probably have to release without it. 

## Nice to have

- [x] Reconsider the use of fold in Validated #1951 

## Please add here if you notice any missing items.  
Update: removed  #1800 from RC1 scope


iravid:
Would be good to include https://github.com/typelevel/cats/issues/1733 in this list. `Sync` for `ReaderT[IO, E, ?]` seems very important.

CHANGES.md
```diff
@@ -1,3 +1,163 @@
...
+* [#1934](https://github.com/typelevel/cats/pull/1934) Restruct `functor`  by @kailuowang
+* [#1803](https://github.com/typelevel/cats/pull/1803) Convert `ReaderWriterStateT` to `IndexedReaderWriterStateT`  by @iravid
+* [#1775](https://github.com/typelevel/cats/pull/1775) Convert `StateT` to `IndexedStateT`  by @iravid
+* [#1098](https://github.com/typelevel/cats/pull/1098) Add a different `MonoidK` and `SemigroupK` instance for `Kleisli`  by @peterneyens
+* [#1922](https://github.com/typelevel/cats/pull/#922) Make kernel laws consistent with core laws  by @LukaJCB
+
+To migrate from 1.0.0-MF.
+
+* The rename of `Cartesian` to `Semigroupal` and `EitherT.liftT` to `EitherT.liftF` can be done automatically through scalafix we provide. See instructions [here](https://github.com/typelevel/cats/blob/master/scalafix/README.md)
+* For `FlatMap`'s `>>` and `<<`, use `Apply`'s `>*` and `<*` instead
```
iravid: Typo - `>*` instead of `*>`

kailuowang:
@iravid for #1733, @johnynek 's solution seems good, does it necessitate an API addition to in cats (vs. letting users hand roll themselves)? If so, would you propose the new API in the issue?

iravid:
I unfortunately cannot comment intelligently on the proposed solution (other than that it would necessitate a `cats-core -> cats-free` dependency) as I don't understand how the usage of flatMap should look.

ceedubs:
> Reconsider the use of fold in Validated #1951

Since this is a change that could be made in a completely compatible way, I don't see much of a reason that a 1.0 release should be blocked by it.

kailuowang:
@ceedubs that's true, I had the same thought, I included it because the performance penalty seems significant and the fix is simple (if no-one else, I can give it an attack this week), I am going to move it out of the list though.

LukaJCB:
What about #1831? I think we should address that in some way maybe

CHANGES.md
```diff
@@ -1,3 +1,185 @@
...
+* [#1955](https://github.com/typelevel/cats/pull/1955) Deprecate `FlatMap`'s `>>` and `<<`  by @LukaJCB
+* [#1947](https://github.com/typelevel/cats/pull/1947) Rename `EitherT.liftT` to `EitherT.liftF`  by @aeons
+* [#1934](https://github.com/typelevel/cats/pull/1934) Restruct `functor`  by @kailuowang
+* [#1803](https://github.com/typelevel/cats/pull/1803) Convert `ReaderWriterStateT` to `IndexedReaderWriterStateT`  by @iravid
+* [#1775](https://github.com/typelevel/cats/pull/1775) Convert `StateT` to `IndexedStateT`  by @iravid
+* [#1098](https://github.com/typelevel/cats/pull/1098) Add a different `MonoidK` and `SemigroupK` instance for `Kleisli`  by @peterneyens
+* [#1922](https://github.com/typelevel/cats/pull/1922) Make kernel laws consistent with core laws  by @LukaJCB
+* [#1838](https://github.com/typelevel/cats/pull/1838) Sync `NonEmptyList` and `NonEmptyVector` methods by @durban
+* [#1914](https://github.com/typelevel/cats/pull/1914) Add `Invariant` instances for kernel type classes by @LukaJCB 
+* [#1980](https://github.com/typelevel/cats/pull/1980) Make `iterateRight` in `Foldable` sound  by @LukaJCB
```
johnynek: this is also a bug fix. I would either list it twice or move it there (I doubt any user will be hit by it as an incompatibility).

LukaJCB: My PR approval still stands if you decide to add this @kailuowang :) So go ahead and merge afterwards 🎉

PR #1976, opened 2017-10-16
reduced usage of `fold` in `Validated`

### LINKED ISSUE #1951: Reconsider the use of fold in Validated
A lot of the method implementations in `Validated` are extremely inefficient, which isn't ideal for projects like circe that rely on it extensively and are aiming for reasonable performance. For example, given this simplified definition (where `eqF` is the current `===` and `eqP` is a lightly optimized version):

```scala
import cats.kernel.Eq

sealed abstract class Validated[+E, +A] extends Product with Serializable {
  import Validated.{Invalid, Valid}

  def fold[B](fe: E => B, fa: A => B): B = this match {
    case Invalid(e) => fe(e)
    case Valid(a) => fa(a)
  }

  def eqF[EE >: E, AA >: A](that: Validated[EE, AA])
    (implicit EE: Eq[EE], AA: Eq[AA]): Boolean = fold(
      a => that.fold(EE.eqv(a, _), _ => false),
      b => that.fold(_ => false, AA.eqv(b, _))
    )

  def eqP[EE >: E, AA >: A](that: Validated[EE, AA])
    (implicit EE: Eq[EE], AA: Eq[AA]): Boolean = this match {
      case Invalid(e) => that match {
        case Invalid(ee) => EE.eqv(e, ee)
        case _ => false
      }
      case Valid(a) => that match {
        case Valid(aa) => AA.eqv(a, aa)
        case _ => false
      }
    }
}

object Validated {
  final case class Valid[+A](a: A) extends Validated[Nothing, A]
  final case class Invalid[+E](e: E) extends Validated[E, Nothing]
}
```
And a benchmark like this:

```scala
import cats.kernel.instances.all._
import java.util.concurrent.TimeUnit
import org.openjdk.jmh.annotations._

@State(Scope.Thread)
class ValidatedBenchmark {
  val a: Validated[String, Int] = Validated.Valid(1)
  val b: Validated[String, Int] = Validated.Valid(2)
  val c: Validated[String, Int] = Validated.Invalid("bad")
  val d: Validated[String, Int] = Validated.Invalid("worse")

  @Benchmark def eqF: Boolean =
    a.eqF(a) && !a.eqF(b) && !a.eqF(c) && c.eqF(c) && !c.eqF(d) && !c.eqF(a)

  @Benchmark def eqP: Boolean =
    a.eqP(a) && !a.eqP(b) && !a.eqP(c) && c.eqP(c) && !c.eqP(d) && !c.eqP(a)
}
```

The pattern matching implementation gets over four times the throughput on 2.12:

```
Benchmark                Mode  Cnt          Score         Error  Units
ValidatedBenchmark.eqF  thrpt   40   27131124.981 ±  164113.218  ops/s
ValidatedBenchmark.eqP  thrpt   40  118711782.519 ± 1166236.166  ops/s
```
The situation is even worse on 2.11, where the difference is a little larger and the `fold` version also involves a lot of unnecessary allocations:

```
Benchmark                                        Mode  Cnt          Score         Error   Units
ValidatedBenchmark.eqF:gc.alloc.rate.norm       thrpt    5        144.000 ±       0.001    B/op
ValidatedBenchmark.eqP:gc.alloc.rate.norm       thrpt    5         ≈ 10⁻⁵                  B/op
```

For most use cases these differences are unlikely to matter at all, but it seems like a shame to impose this extra cost on everyone just for the sake of slightly more concise implementations.

(There are similar issues in `IndexedStateT`, etc., but `Validated` is the one piece I personally really care that much about, because of how it affects circe.)

kailuowang:
fixes #1951

### DIFF
--- a/core/src/main/scala/cats/data/Validated.scala
+++ b/core/src/main/scala/cats/data/Validated.scala
@@ -14,33 +14,55 @@ sealed abstract class Validated[+E, +A] extends Product with Serializable {
       case Valid(a) => fa(a)
     }
 
-  def isValid: Boolean = fold(_ => false, _ => true)
-  def isInvalid: Boolean = fold(_ => true, _ => false)
+  def isValid: Boolean = this match {
+    case Invalid(_) => false
+    case _ => true
+  }
+
+  def isInvalid: Boolean = this match {
+    case Invalid(_) => true
+    case _ => false
+  }
 
   /**
    * Run the side-effecting function on the value if it is Valid
    */
-  def foreach(f: A => Unit): Unit = fold(_ => (), f)
+  def foreach(f: A => Unit): Unit = this match {
+    case Valid(a) => f(a)
+    case _ => ()
+  }
 
   /**
    * Return the Valid value, or the default if Invalid
    */
-  def getOrElse[B >: A](default: => B): B = fold(_ => default, identity)
+  def getOrElse[B >: A](default: => B): B = this match {
+    case Valid(a) => a
+    case _ => default
+  }
 
   /**
     * Return the Valid value, or the result of f if Invalid
     */
-  def valueOr[B >: A](f: E => B): B = fold(f, identity)
+  def valueOr[B >: A](f: E => B): B = this match {
+    case Invalid(e) => f(e)
+    case Valid(a) => a
+  }
 
   /**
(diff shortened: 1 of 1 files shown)


johnynek:
👍

Issue #1943, opened 2017-10-01
Monad Transformer for Ior


frroliveira:
I have put together a proposal mostly based on what EitherT does in here
https://gist.github.com/frroliveira/5d4fbdfa1fd465e8728615f6107c2716

* named it `IorT`
* left out `ensure`, `ensureOr`, `compare` and `partialCompare` as `Ior` does not implement them

PR #1977, opened 2017-10-17
Add Ior Monad Transformer

frroliveira:
Initial work for #1943. Happy to add docs and more methods (that EitherT has), if this seems ok

### DIFF
--- a/core/src/main/scala/cats/data/IorT.scala
+++ b/core/src/main/scala/cats/data/IorT.scala
@@ -406,18 +406,18 @@ private[data] abstract class IorTInstances1 extends IorTInstances2 {
   implicit def catsDataFoldableForIorT[F[_], A](implicit F: Foldable[F]): Foldable[IorT[F, A, ?]] =
     new IorTFoldable[F, A] { val F0: Foldable[F] = F }
 
-  implicit def catsDataMonadErrorFForIorT[F[_], A, E](implicit FE: MonadError[F, E], A: Semigroup[A]): MonadError[IorT[F, A, ?], E] =
-    new IorTMonadErrorF[F, A, E] {
-      val A0: Semigroup[A] = A
-      val F0: MonadError[F, E] = FE
-    }
-}
-
-private[data] abstract class IorTInstances2 extends IorTInstances3 {
   implicit def catsDataMonadErrorForIorT[F[_], A](implicit F: Monad[F], A: Semigroup[A]): MonadError[IorT[F, A, ?], A] =
     new IorTMonadError[F, A] {
       val A0: Semigroup[A] = A
       val F0: Monad[F] = F
+    }
+}
+
+private[data] abstract class IorTInstances2 extends IorTInstances3 {
+  implicit def catsDataMonadErrorFForIorT[F[_], A, E](implicit FE: MonadError[F, E], A: Semigroup[A]): MonadError[IorT[F, A, ?], E] =
+    new IorTMonadErrorF[F, A, E] {
+      val A0: Semigroup[A] = A
+      val F0: MonadError[F, E] = FE
     }
 
   implicit def catsDataEqForIorT[F[_], A, B](implicit F: Eq[F[Ior[A, B]]]): Eq[IorT[F, A, B]] =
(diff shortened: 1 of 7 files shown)


kailuowang:
@frroliveira thanks very much for this PR. To give you a heads up, at the moment we are focusing on pushing across the RC1  finish line. This PR since is completely backward compatible, although useful, is not the highest on the priority list right now. But we will get back to it later.

frroliveira:
@kailuowang No problem. Thanks for letting me know

core/src/main/scala/cats/data/IorT.scala
```diff
@@ -0,0 +1,490 @@
...
+      case l @ Ior.Left(_) => F.pure(l.asInstanceOf[Ior[AA, D]])
+      case Ior.Right(b) => f(b).value
+      case Ior.Both(a, b) => F.map(f(b).value) {
+        case Ior.Left(a1) => Ior.Left(AA.combine(a, a1))
+        case Ior.Right(d) => Ior.Both(a, d)
+        case Ior.Both(a1, d) => Ior.Both(AA.combine(a, a1), d)
+      }
+    })
+
+  def flatMapF[AA >: A, D](f: B => F[Ior[AA, D]])(implicit F: Monad[F], AA: Semigroup[AA]): IorT[F, AA, D] =
```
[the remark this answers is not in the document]
frroliveira: Current implementation needs it to call `flatMap`. I couldn't find a way to use it here, however `collectRight` can be changed to `F: FlatMap[F]`

core/src/main/scala/cats/data/IorT.scala
```diff
@@ -0,0 +1,490 @@
...
+    flatMap(f andThen IorT.apply)
+
+  def subflatMap[AA >: A, D](f: B => Ior[AA, D])(implicit F: Monad[F], AA: Semigroup[AA]): IorT[F, AA, D] =
```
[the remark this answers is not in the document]
frroliveira: This one can actually be `Functor`

frroliveira:
Do you guys prefer docs to be merged with code or with separate PR?

kailuowang:
@frroliveira it would be great to include docs in this PR as well.

frroliveira:
Sorry it took me some time to get the docs.
Meanwhile, I have added an extra method `condF`. It would probably be useful in `EitherT` as well.

Regarding another PR that has been merged. Should I keep the order of the MonadError instances? Or switch them to be consistent with `EitherT`?

frroliveira:
Ah, and I forgot. The apt package manager does not install the latest version of jekyll by default, so I updated the Contributing file

core/src/main/scala/cats/data/IorT.scala
```diff
@@ -0,0 +1,509 @@
...
+}
+
+private[data] abstract class IorTInstances1 extends IorTInstances2 {
+  implicit def catsDataSemigroupForIorT[F[_], A, B](implicit F: Semigroup[F[Ior[A, B]]]): Semigroup[IorT[F, A, B]] =
+    new IorTSemigroup[F, A, B] { val F0: Semigroup[F[Ior[A, B]]] = F }
+
+  implicit def catsDataFoldableForIorT[F[_], A](implicit F: Foldable[F]): Foldable[IorT[F, A, ?]] =
+    new IorTFoldable[F, A] { val F0: Foldable[F] = F }
+
+  implicit def catsDataMonadErrorFForIorT[F[_], A, E](implicit FE: MonadError[F, E], A: Semigroup[A]): MonadError[IorT[F, A, ?], E] =
```
kailuowang: yeah, +1 on switching the order of these two MonadError instance to be consistent with EitherT

core/src/main/scala/cats/data/IorT.scala
```diff
@@ -0,0 +1,509 @@
...
+   * scala> import cats.implicits._
+   * scala> val o: Option[Int] = Some(3)
+   * scala> val n: Option[Int] = None
+   * scala> IorT.liftF(o)
+   * res0: cats.data.IorT[Option,Nothing,Int] = IorT(Some(Right(3)))
+   * scala> IorT.liftF(n)
+   * res1: cats.data.IorT[Option,Nothing,Int] = IorT(None)
+   * }}}
+   */
+  final def liftF[F[_], A, B](fb: F[B])(implicit F: Applicative[F]): IorT[F, A, B] = right(fb)
```
kailuowang: do you mind also add `liftK` like this one https://github.com/SystemFw/cats/blob/4830e326acab64223fa642a2b39111665e590763/core/src/main/scala/cats/data/EitherT.scala#L365 ? If you don't have the time it's fine too, I can add one in a follow up PR.

frroliveira: sure, no problem

PR #1978, opened 2017-10-18
Helper method to convert Kleisli[F, A, B] to Kleisli[Id, A, F[B]]

cb372:
I'm not sure how common a use case this is but I find myself needing to do it occasionally.

The test I added feels a bit pointless. Can delete if people agree.

## A motivating example

Say I'm using `Kleisli[Either[String, ?], Context, ?]` in the core of my program:

```scala
case class Context()
case class User()
case class Thing()

def getUser(userId: String): Kleisli[Either[String, ?], Context, User] = ???
def getThing(user: User): Kleisli[Either[String, ?], Context, Thing] = ???

def getThingByUserId(userId: String): Kleisli[Either[String, ?], Context, Thing] =
  for {
    user  <- getUser(userId)
    thing <- getThing(user)
  } yield thing
```

Then on the edge of the program, say in an HTTP controller, I want to match on the `Either` in order to build an HTTP response:

```scala
def buildHttpResult(errorOrThing: Either[String, Thing]) = Kleisli[Id, Context, Result] { ctx =>
  errorOrThing match {
    case Left(e)      => Results.InternalServerError(s"oh no: $e")
    case Right(thing) => Results.Ok("yay")
  }
}
```

The new `toReader` method allows me to pass along the `Either` (rather than the value inside the `Either`) when doing monadic composition:

```scala
def action(userId: String): Result = {
  val context = Context()

  val kleisli: Kleisli[Id, Context, Result] =
    for {
      errorOrThing <- getThingByUserId(userId).toReader
      result       <- buildHttpResult(errorOrThing)
    } yield result

  kleisli.run(context)
}
```

### DIFF
--- a/core/src/main/scala/cats/data/Kleisli.scala
+++ b/core/src/main/scala/cats/data/Kleisli.scala
@@ -68,8 +68,7 @@ final case class Kleisli[F[_], A, B](run: A => F[B]) { self =>
   def tapWith[C](f: (A, B) => C)(implicit F: Functor[F]): Kleisli[F, A, C] =
     Kleisli(a => F.map(run(a))(b => f(a, b)))
 
-  def toReader: Kleisli[Id, A, F[B]] =
-    mapF[Id, F[B]](scala.Predef.identity)
+  def toReader: Kleisli[Id, A, F[B]] = Kleisli[Id, A, F[B]](run)
 
   def apply(a: A): F[B] = run(a)
 }
--- a/core/src/main/scala/cats/data/Kleisli.scala
+++ b/core/src/main/scala/cats/data/Kleisli.scala
@@ -1,7 +1,7 @@
 package cats
 package data
 
-import cats.Contravariant
+import cats.{Contravariant, Id}
 import cats.arrow._
 
 /**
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/data/Kleisli.scala
```diff
@@ -68,6 +68,9 @@ final case class Kleisli[F[_], A, B](run: A => F[B]) { self =>
   def tapWith[C](f: (A, B) => C)(implicit F: Functor[F]): Kleisli[F, A, C] =
     Kleisli(a => F.map(run(a))(b => f(a, b)))
 
+  def toReader: Kleisli[Id, A, F[B]] =
```
johnynek: a generalization of this is `Kleisli[Lambda[x => F[G[x]]], A, B] => Kleisli[F, A, G[B]]`. I wonder if it is worth adding a method like that to the companion object:
```scala
def rotate[A, B, F[_], G[_]](k: Klesli[Lambda[x => F[G[x]]], A, B]): Klesli[F, A, G[B]] =
  k.mapF[F, G[B]](identity)
```
I don't know if it worth having the method or not to save one `mapF` call. `toReader` looks useful enough to me.

core/src/main/scala/cats/data/Kleisli.scala
```diff
@@ -68,6 +68,9 @@ final case class Kleisli[F[_], A, B](run: A => F[B]) { self =>
...
+    mapF[Id, F[B]](scala.Predef.identity)
```
kailuowang: at least we can save the identity call right? 
```scala
def toReader: Kleisli[Id, A, F[B]] = Kleisli[Id, A, F[B]](run)
```

cb372: Good point! Updated.

kailuowang:
So sorry, I have one more nit, shall we make the return type `Reader[A, F[B]]`?  just reads better, that's all.

LukaJCB:
Thank you! :)

PR #1981, opened 2017-10-19
Add UnorderedFoldable and UnorderedTraverse 

### LINKED ISSUE #1831: Foldable.toList potentially inconsistent for Map

LukaJCB:
As discussed in #1927, together with #1972 and #1966 this should fix #1831.
In this PR, I remove the `Traverse` and `Foldable` instances from `Map` and `Set` and replace them with `UnorderedTraverse` and `UnorderedFoldable`. I think we could let `Foldable` and `Traverse` extend these.

This PR depends on #1927.


kailuowang:
I like the idea of introducing these two new type classes.
I am not sure they should be introduced to the `Traverse`/`Foldable` hierarchy. It's not clear to me that for `Traverse` and `Foldable` and their instances, what benefit them for inheriting `UnorderedFoldable` and `UnorderedTraverse`. They have to override most methods and they got two additional unorderred methods that are not needed by their instances. Is there a concrete example in which such inheritance arrangement would be beneficial?

LukaJCB:
Good question! I don't think there's a huge benefit apart from deduplication.

kailuowang:
You probably guessed what I would say next. They are useful but
1) not sure how useful as an abstraction to be included in core. If not we could just add them as `Map` and `Set` syntax extension. 
2) not sure if, as abstractions, they are mature enough to be locked down in RC1. 

Sorry, I am biased towards limiting scope in 1.0.

LukaJCB:
Understandable, don't be sorry :) I agree that scope should be limited as we want to remain lightweight. I fairly like these two classes though and think they'd be a worthwhile abstraction. 
Maybe we should put this one up for a vote :)

core/src/main/scala/cats/UnorderedFoldable.scala
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+  * `UnorderedFoldable` is like a `Foldable` for unordered containers.
+  */
+@typeclass trait UnorderedFoldable[F[_]] {
+
+  /**
+    * Left associative fold on 'F' using the function 'f'.
+    */
+  def foldLeft[A, B](fa: F[A], b: B)(f: (B, A) => B): B
```
johnynek: How can we have this in an unordered way? I’d like to remove this.

core/src/main/scala/cats/UnorderedFoldable.scala
```diff
@@ -0,0 +1,66 @@
...
+
+  def unorderedFoldMap[A, B: CommutativeMonoid](fa: F[A])(f: A => B): B =
```
johnynek: This should be the core method in my view.

core/src/main/scala/cats/UnorderedFoldable.scala
```diff
@@ -0,0 +1,66 @@
...
+    foldLeft(fa, Monoid[B].empty)((b, a) => Monoid[B].combine(b, f(a)))
+
+  def unorderedFold[A: CommutativeMonoid](fa: F[A]): A =
+    unorderedFoldMap(fa)(identity)
+
+  def toSet[A](fa: F[A]): Set[A] =
```
johnynek: This can be implemented with foldMap

core/src/main/scala/cats/UnorderedFoldable.scala
```diff
@@ -0,0 +1,66 @@
...
+    foldLeft(fa, mutable.ListBuffer.empty[A]) { (buf, a) =>
+      buf += a
+    }.toSet
+
+
+  /**
+    * Returns true if there are no elements. Otherwise false.
+    */
+  def isEmpty[A](fa: F[A]): Boolean =
```
johnynek: This can be implemented with foldMap (map to true, do the commutative “or” monoid

core/src/main/scala/cats/UnorderedFoldable.scala
```diff
@@ -0,0 +1,66 @@
...
+    foldLeft(fa, Option.empty[A]) { (lb, a) =>
+      if (f(a)) Some(a) else lb
+    }
+
+  /**
+    * Check whether at least one element satisfies the predicate.
+    *
+    * If there are no elements, the result is `false`.
+    */
+  def exists[A](fa: F[A])(p: A => Boolean): Boolean =
```
johnynek: This can be implemented with unorderedFoldMap

core/src/main/scala/cats/UnorderedFoldable.scala
```diff
@@ -0,0 +1,66 @@
...
+    foldLeft(fa, false) { (lb, a) =>
+      if (p(a)) true else lb
+    }
+
+  /**
+    * Check whether all elements satisfy the predicate.
+    *
+    * If there are no elements, the result is `true`.
+    */
+  def forall[A](fa: F[A])(p: A => Boolean): Boolean =
```
johnynek: This can be implemented with foldMap using and monoid.

kailuowang:
If they are not part of the `Traverse`/`Foldable` hierarchy, they can be added in a binary compatible way, meaning they can be added post RC1 or 1.0.0.  Personally, I think these two new classes need more discussions and use case scenarios for those discussions. Only then we can confidently decide on whether they should go `cats.core` or `cats.more` or somewhere else. 
In the meantime, I am not sure if there are users who would really need this to hold up 1.0 release.

LukaJCB:
Adding a superclass should be a API compatible change, no? To be clear, the API breaking change I thought about in this PR was the removal of the `Foldable` and `Traverse` instances for `Set` and ` Map`.

LukaJCB:
I could move the part where we remove the instances to a different PR so we could include that in RC1

kailuowang:
I am talking about binary compatibility. It has to be a binary compatible change to be introduced post 1.0 release. 
I am moving alleycats in so that we can move the `travese` instance there.

kailuowang:
Yeah, alleycats move in PR is in. After that is merged in we can move the traverse and foldable instances there and add tests to expose the inconsistency.

johnynek:
I like this PR. I think it gives a principled way to use Map/Set rather than just using the law violating ones in alleycats.

In my experience, it is hard to be sure when a law violation is actually going to burn you later.

That said, with a typeclass we should have several instances for it generally. It would be good to think of more instances than just hash-(map|set).

I guess spark `RDD[T]` or `TypedPipe[T]` in scalding are almost unorderedFoldable (they each require side-effects to do the fold, but both are unordered).

kailuowang:
Can we make these two separate from the `Traverse` and `Foldable`. The way I see it they are parallel rather than hierarchical.  `Traverse` and `Foldable` are for ordered data types, while these two are for unordered ones.

kailuowang:
Can we rebase master? I am curious if this one is bin compat with 1.0.0-RC1, Also I still feel pretty strongly that we should not have Traverse and Foldable inherit from them.

core/src/main/scala/cats/UnorderedTraverse.scala
```diff
@@ -0,0 +1,14 @@
+package cats
+
+import simulacrum.typeclass
+
+/**
+  * `UnorderedTraverse` is like a `Traverse` for unordered containers. In addition to the traverse and sequence
+  * methods it provides nonEmptyTraverse and nonEmptySequence methods which require an `Apply` instance instead of `Applicative`.
```
hrhino: This line of docs seems inaccurate?

core/src/main/scala/cats/UnorderedFoldable.scala
```diff
@@ -0,0 +1,58 @@
...
+    unorderedFoldMap(fa)(p)(UnorderedFoldable.orMonoid)
+
+  /**
+    * Check whether all elements satisfy the predicate.
+    *
+    * If there are no elements, the result is `true`.
+    */
+  def forall[A](fa: F[A])(p: A => Boolean): Boolean =
+    unorderedFoldMap(fa)(p)(UnorderedFoldable.andMonoid)
+}
```
hrhino: This could implement the `size` method from `Foldable` as well, I think.

laws/src/main/scala/cats/laws/UnorderedFoldableLaws.scala
```diff
@@ -0,0 +1,46 @@
...
+        (F.isEmpty(fa) || F.exists(fa)(p))
+    } else true // can't test much in this case
+  }
+
+  /**
+    * If `F[A]` is empty, forall must return true.
+    */
+  def forallEmpty[A](fa: F[A], p: A => Boolean): Boolean = {
+    !F.isEmpty(fa) || F.forall(fa)(p)
+  }
```
hrhino: Could it be made to be such that `existsLazy` and `forallLazy` are laws for `UnorderedFoldable` and not just `Foldable`?

laws/src/main/scala/cats/laws/UnorderedTraverseLaws.scala
```diff
@@ -0,0 +1,62 @@
+package cats
+package laws
+
+import cats.data.Nested
+
+trait UnorderedTraverseLaws[F[_]] extends UnorderedFoldableLaws[F] {
```
hrhino: This could/should extend `FunctorLaws[F]` and assert some things about the compatibility of `UnorderedTraverse` and `Functor`; for instance, I would expect `unorderedSequence` to  be `unorderedTraverse` on identity (which assumes that `Eq` for these instances respects unorderedness if it's not also a `Traverse`, I think)

laws/src/main/scala/cats/laws/UnorderedFoldableLaws.scala
LukaJCB: I think lazyness would require `foldRight`, no?

laws/src/main/scala/cats/laws/UnorderedTraverseLaws.scala
LukaJCB: Hmmm, I'm not sure. That would imply that `Set` would be a valid `Functor` which I believe it is not. We could add a function like `unorderedMap` that's just `unorderedTraverse` with `Id`.

Thanks for your review by the way, it's been very helpful! :)

laws/src/main/scala/cats/laws/UnorderedFoldableLaws.scala
hrhino: I believe you can do it with `Eval`: see my duplicative [attempt](https://github.com/hrhino/cats/blob/5638c658bafcac4e06a841138d1bb5f6fa71617c/core/src/main/scala/cats/UnorderedFoldable.scala#L94-L99) at this type class.

```scala
  def exists[A](fa: F[A])(p: A => Boolean): Boolean =
    foldMapUnordered(fa)(a => Eval.later(p(a)))(new CommutativeMonoid[Eval[Boolean]] {
      def empty = Eval.False
      def combine(x: Eval[Boolean], y: Eval[Boolean]): Eval[Boolean] =
        x.flatMap(xv => if (xv) Eval.True else y)
    }).value
```

and dually for `forall`.

laws/src/main/scala/cats/laws/UnorderedTraverseLaws.scala
hrhino: No, you're right; I got a little bit overenthusiastic about drawing these parallels. 

I think `unorderedMap` is kinda dubious, because the difference isn't even lack of order (which is unobservable in `Functor#map` already!), but that the operation can "drop" values, losing fusion, and therefore isn't a (categorical) `Functor`. I don't know what laws for `unorderedMap` would be, other than "it's the same as traversing with `Id`.

laws/src/main/scala/cats/laws/UnorderedFoldableLaws.scala
LukaJCB: You're absolutely right, this is great! Thank you so much!

core/src/main/scala/cats/UnorderedFoldable.scala
johnynek: I wonder if actually we should take an Order here. This is using universal hashing and i hate to bake that into a new typeclass.

LukaJCB: If we take an Order we could make it a `SortedSet`, not sure if that really makes sense. I agree with your sentiment, but not sure how to deal with it exactly. Maybe take a `Hash[A]`? I dunno.

kailuowang: I agree that having `UnorderedFoldable` taking a `Order` doesn't make much sense. 
The whole point of this PR is to support a principled functionalities for `Set` and `Map`, which are based off universal hashing I am not sure if we can get away from it. 
If we really want `UndorderedFoldable` as a principled abstraction,  we should probably only support the `Set` and `Map` in dogs.

johnynek: we could make this `toList` with the only law being about `contains` and `Eq[A]`. It would be `O(N^2)` to check the law, but that is probably fine.

I just hate sneaking universal hashing or universal equality in. 

I don't see the use or `Order[A]` as not making sense, since it is the *collection* that is unordered here, not the type `A`. So, I don't want to get those two confused.

laws/src/main/scala/cats/laws/UnorderedFoldableLaws.scala
```diff
@@ -0,0 +1,64 @@
...
+
+  def toSetRef[A](fa: F[A]): IsEq[Set[A]] =
+    F.unorderedFoldMap(fa)(a => Set(a)) <-> F.toSet(fa)
```
johnynek: this could be
```scala
val lst1 = F.unorderedFoldMap(fa)(a => List(a))
val lst2 = F.toList(fa)
def contains[A: Eq](a: A): Boolean = lst1.exists(Eq[A].eqv(_, a))
val allCont = list2.map(contains)
val allTrue = lst.map(_ => true)
allCont <-> allTrue
```
In this way, we have a law that only needs Eq, not unusual, and the typeclass does not require any Order or Hash.
if we have an `Order[A]`. So we use an `Order` to check the law more quickly, or we could

core/src/main/scala/cats/UnorderedFoldable.scala
kailuowang: +1 on this solution.

LukaJCB: I'm not sure `toList` really makes sense, given the fact that we only want to fold using a `CommutativeMonoid` which `List` clearly is not. Maybe we just get rid of this method all together? I really wish we had a `HashMap`.

kailuowang: I agree. At the very least this is not stable enough to be included in 1.0.0

PR #1984, opened 2017-10-20
Welcome, Alleycats

kailuowang:
This PR moves alleycats into cats. We had some back and forth on this issue, but more and more evidences show it's more beneficial to have it inside cats repo. 
It still has its own root package as `alleycats` and is only available when user explicitly add it as a dependency. So that users shall be careful when using instances defined here. 

In the process, I removed the dependencies from alleycats to algebra, since 1) this should be solely for outlaw TCs in cats and 2) algebra depends on cats.kernel.
I believe it is fine since kittens (which I co-maintain) is the only project that uses alleycats and it doesn't use the part in alleycats that has those dependencies. 

The immediate incentive is for us to move traverse and foldable instances for map and set in here.


kailuowang:
not sure if anyone wants to review all the alleycats code. I think only the `build.sbt` needs some review.

alleycats-laws/src/main/scala/alleycats/laws/discipline/FlatMapRecTests.scala
```diff
@@ -0,0 +1,29 @@
...
+
+import cats._
+import cats.laws.FlatMapLaws
+import cats.laws.discipline._
+import org.scalacheck.Arbitrary
+import org.scalacheck.Prop._
+import org.typelevel.discipline.Laws
+
+
+trait FlatMapRecTests[F[_]] extends Laws {
```
johnynek: this looks duplicative.

alleycats-core/src/main/scala/alleycats/std/iterable.scala
```diff
@@ -0,0 +1,38 @@
...
+@exports
+object IterableInstances {
+  @export(Orphan)
+  implicit val exportIterableFoldable: Foldable[Iterable] =
+    new Foldable[Iterable] {
+      override def foldLeft[A, B](fa: Iterable[A], b: B)(f: (B, A) => B): B = fa.foldLeft(b)(f)
+
+      // based upon foldRight of List in Cats
+      override def foldRight[A, B](fa: Iterable[A], lb: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B] = {
+        def loop(as: Iterable[A]): Eval[B] =
```
johnynek: I think this can actually be faster and get code-reuse if we use `Foldable.iterateRight`: https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Foldable.scala#L508

alleycats-core/src/main/scala/alleycats/std/iterable.scala
```diff
@@ -0,0 +1,38 @@
...
+import export._
+
+@reexports(IterableInstances)
+object iterable extends LegacyIterableInstances
+
+@exports
+object IterableInstances {
+  @export(Orphan)
+  implicit val exportIterableFoldable: Foldable[Iterable] =
+    new Foldable[Iterable] {
```
johnynek: I don't actually know why this shouldn't be in cats-core. What is the problem exactly? `Foldable` is lawless to begin with.

alleycats-core/src/main/scala/alleycats/syntax/foldable.scala
```diff
@@ -0,0 +1,14 @@
...
+
+import cats.Foldable
+import cats.syntax.foldable._
+
+object foldable extends FoldableSyntax
+
+trait FoldableSyntax {
+  implicit class ExtraFoldableOps[F[_]: Foldable, A](fa: F[A]) {
+    def foreach(f: A => Unit): Unit =
+      fa.foldLeft(()) { (unit, a) => f(a); unit }
```
johnynek: why aren't we returning `f(a)` here? I don't understand? It is `Unit` right?

alleycats-core/src/main/scala/alleycats/std/set.scala
```diff
@@ -0,0 +1,87 @@
...
+  // Since iteration order is not guaranteed for sets, folds and other
+  // traversals may produce different results for input sets which
+  // appear to be the same.
+  @export(Orphan)
+  implicit val setTraverse: Traverse[Set] =
+    new Traverse[Set] {
+      def foldLeft[A, B](fa: Set[A], b: B)(f: (B, A) => B): B =
+        fa.foldLeft(b)(f)
+      def foldRight[A, B](fa: Set[A], lb: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B] =
+        Foldable.iterateRight(fa.iterator, lb)(f)
```
LukaJCB: This one needs to be 
```scala
Foldable.iterateRight(fa, lb)(f)
```

alleycats-laws/src/main/scala/alleycats/laws/discipline/FlatMapRecTests.scala
kailuowang: I believe this is the only way to test the one specific rule out of the rule set, in this case `tailRecMConsistentFlatMap`

alleycats-core/src/main/scala/alleycats/std/iterable.scala
LukaJCB: AFAIK `Iterable`'s only method is `iterator` which isn't exactly referentially transparent, no?

BennyHill:
👍 Better late than never 😛

alleycats-core/src/main/scala/alleycats/std/iterable.scala
johnynek: But that is the only method you need to implement Foldable. And I don’t see why creating a new Iterator is not referentially transparent. It certainly is for all instances of Iterable is the standard library AFAIK.

johnynek: Doesn’t Stream have a Foldable instance in cats already? What prevents it from being Foldable?

kailuowang: you are right, `Stream` is `Foldable`. Yeah, I don't see why we couldn't move `Foldable` of`Iterable` into core.

PR #1986, opened 2017-10-21
Remove Edmund Noble as maintainer and author

edmundnoble:
In light of the recent admission of attempted career sabotage by @larsrh on Twitter, I'm resigning as maintainer on all typelevel and typelevel incubator projects. It was nice working with all of you, and I hope to work with you again in the future.


larsrh:
> In light of the recent admission of attempted career sabotage

For the record: I did not admit to any such thing, because it never happened. If you had read the thread, you'd have seen that I even offered to provide evidence to the contrary.

PR #1993, opened 2017-10-25
Added examples for Cokleisli

raymondtay:
Thought it might be helpful to include examples on `dimap`, `lmap` and `rmap` for learners of the library


core/src/main/scala/cats/data/Cokleisli.scala
```diff
@@ -11,9 +11,31 @@ import scala.annotation.tailrec
...
 
+  /**
+    * Example:
+    * {{{
+    * scala> import cats.Id, cats.implicits._
+    * scala> val x : Id[Int] = 42
+    * scala> def before(x: Int) = x + 1
+    * scala> def after(x: Int) = x - 1
+    * scala> val example : Cokleisli[Id,Int,Int] = Cokleisli((f: Id[Int]) => f.extract)
+    * scala> example.dimap(before)(after) == 42
```
LukaJCB: Can you add what this returns here and below?

raymondtay: Thanks @LukaJCB for the reminder, appreciate it.

raymondtay: @johnynek thanks for the review and agree with you. Following on your lead, i've revised the examples to use `NonEmptyList` instead, let me know what you think when you look at them?

raymondtay:
@johnynek thanks for the review and i agree with you. Following your suggestion, i've changed it examples to using `NonEmptyList` instead of `Id`. Let me know what you think of those changes?

core/src/main/scala/cats/data/Cokleisli.scala
```diff
@@ -14,12 +14,13 @@ final case class Cokleisli[F[_], A, B](run: F[A] => B) { self =>
...
     * {{{
-    * scala> import cats.Id, cats.implicits._
-    * scala> val x : Id[Int] = 42
-    * scala> def before(x: Int) = x + 1
-    * scala> def after(x: Int) = x - 1
-    * scala> val example : Cokleisli[Id,Int,Int] = Cokleisli((f: Id[Int]) => f.extract)
-    * scala> example.dimap(before)(after) == 42
+    * scala> import cats._, data._
+    * scala> val f = Cokleisli((xs: NonEmptyList[Int]) => xs.reverse.head)
+    * f: cats.data.Cokleisli[cats.data.NonEmptyList,Int,Int] = Cokleisli(<function1>)
```
rossabaker: I'd omit these toString renderings, just like you omitted the output of the defs.  The other examples in Cats tend to omit the trivial outputs and focus on the setup and the res0.

raymondtay: @rossabaker gotcha :)

core/src/main/scala/cats/data/Cokleisli.scala
```diff
@@ -11,15 +11,47 @@ import scala.annotation.tailrec
  */
 final case class Cokleisli[F[_], A, B](run: F[A] => B) { self =>
 
+  /**
+    * Example:
+    * {{{
+    * scala> import cats._, data._
+    * scala> val f = Cokleisli((xs: NonEmptyList[Int]) => xs.reverse.head)
+    * scala> def before(x: String) = x.toInt
```
johnynek: I hate to be a jerk, but can we use a function that won’t throw. I know this example is safe but using toInt in general is dangerous and I hate to set the example in docs.

raymondtay: Agree

Issue #1995, opened 2017-10-29
Remove `Eq.and`, `Eq.or`, `Eq/PartialOrder/Order.on`, `PartialOrder/Order.reverse`


denisrosset:
These methods are quite unusual in the sense that `Eq`, `PartialOrder` and `Order` are (to my knowledge) the only typeclasses to have direct methods to derive instances. IMHO, these methods belong to the companion objects.

In particular, `on` is a nightmare to override consistently in hierarchies that inherit from `Eq`. For example, currently `(h: Hash[A]).on[B]` returns `Eq[B]`, not `Hash[B]`, while `Hash.by` does the right thing.

And the methods `Eq.and`, `Eq.or` are confusing when `Eq` should be actually a coherent typeclass.

Thus, I propose to remove them. If compatibility is needed, they can be provided by enrichment methods.

Edit: if there is interest, I will propose a PR, but I don't know where `cats-kernel` stands binary compatibility wise.

kailuowang:
+1 on moving them to companion

johnynek:
I don’t think actually have a policy that we assume coherent instances and as such don’t want combinator methods on typeclasses.

I personally would love if scala has coherent typeclasses you can only define in companions of either the type or the typeclass but it doesn’t and in the mean time things like Order.reverse are useful in sorting applications.

The subclass argument is a good one to me, but I would move the method into the companion so it is still easy to contramap these types.

ctongfei:
This is the problem of branching subtypeclasses.
Hash <: Eq; Order <: Eq.

If we have `Hash[B]#on[B](f: B=>A): Hash[B]`, then something that extends `Hash` and `Order` will have to override `on` because there is already `Order[A]#on[B](f: B=>A): Order[B]`.

I agree that `on` methods could probably be removed, because once we import `cats.implicits.all` we should have the injected `contramap` method because of their existing `Contravariant` instances.

Should `Eq` be a coherent typeclass? Sometimes we can have different equivalence relations endowed on the same type. I don't think that making `Eq` coherent is a good idea.

kailuowang:
So @johnynek  will it be okay to break kernel back compat with this change (i.e. moving them to companion)?

johnynek:
@kailuowang from the algebird side, we are using the algebraic typeclasses. These equality ones, not at all yet.

LukaJCB:
Fixed by #1997! Thanks @denisrosset! :)

PR #1997, opened 2017-10-30
Minimizing typeclass surface in cats-kernel

denisrosset:
As per #1995


LukaJCB:
Can you rebase with master? :)

kailuowang:
this wil fail Mima test , need to add some exceptions [here](https://github.com/typelevel/cats/blob/master/build.sbt#L236)

kailuowang:
> [info] Compiling 78 Scala sources to /home/travis/build/typelevel/cats/tests/.js/target/scala-2.11/test-classes...
[error] /home/travis/build/typelevel/cats/tests/src/test/scala/cats/tests/SortedSetSuite.scala:17: value reverse is not a member of cats.kernel.PartialOrder[scala.collection.immutable.SortedSet[Int]]
[error]   checkAll("PartialOrder[SortedSet[Int]].reverse", PartialOrderTests(PartialOrder[SortedSet[Int]].reverse).partialOrder)
[error]                                                                                                   ^
[error] /home/travis/build/typelevel/cats/tests/src/test/scala/cats/tests/SortedSetSuite.scala:18: value reverse is not a member of cats.kernel.PartialOrder[scala.collection.immutable.SortedSet[Int]]
[error]   checkAll("PartialOrder[SortedSet[Int]].reverse.reverse", PartialOrderTests(PartialOrder[SortedSet[Int]].reverse.reverse).partialOrder)
[error]                                                                                                           ^
[error] two errors found
[error] (testsJS/test:compileIncremental) Compilation failed

denisrosset:
@kailuowang Rebased with the latest master, should solve the SortedSetSuite problem.

PS: my git-fu is minimal, how should I combine the above commits to avoid polluting the history?

denisrosset:
We should also remove the tests corresponding to the old `.reverse.reverse` test, as we no longer perform the optimization that the reverse of the reverse is the original object.

Note: I could restore this optimization by pattern matching on the type of derived `Order`, at the price of adding new classes.

kailuowang:
Its okay to have multiple trivial commits in a PR. We will squash merge into one commit. I myself don't see it necessary to optimize .reverse.reverse. Testing it against the law doesn't hurt though right?

denisrosset:
It doesn't hurt, but it's unnecessary, as the `.reverse.reverse` test was covering an optimization that is no longer there, so we are basically testing that `-(-x) = x` as the `Order.reverse` method cannot be overloaded. Your call.

PR #1998, opened 2017-10-30
Add Validated.cond and Validated.condNel

andyscott:
Since `cond` is already available on `Either` and `EitherT` it seems reasonable to have it on `Validated` as well.


johnynek:
👍

PR #2019, opened 2017-11-07
WIP: Proposal for Parallel.mirror

alexandru:
Proposal for adding the following capability:

```scala
trait NonEmptyParallel[M[_], F[_]] extends Serializable {
   // ...
  def mirror: NonEmptyParallel[M, M] = ???
}

trait Parallel[M[_], F[_]] extends NonEmptyParallel[M, F] {
   // ...
  def mirror: Parallel[M, M] = ???
}
```

Such that we can do:

```scala
// Source
val p1: Parallel[IO, IO.Par] = Parallel[IO, IO.Par]

// Mirrored, to remove the need for IO.Par
val p2: Parallel[IO, IO] = m1.mirror
```

The reason is that:

1. there are cases in which people just want to do `parMapN` or `parSequence` on `IO` or `Task` without having to carry an extra type parameter around, as in Scala such type parameters suck
2. sometimes you want to reuse code where the only difference made is that of the `Applicative` implementation being used — and yes, type class instances need to be *coherent*, but for people that know what they are doing, for internal things, I don't see the problem

The use-case for Monix is that I'd rather not have the [ParallelApplicative](https://github.com/monix/monix/blob/v3.0.0-RC1/monix-eval/shared/src/main/scala/monix/eval/instances/ParallelApplicative.scala#L25) in its repository, advising people to use it, and yet I do.

The concrete use case is this:

```scala
import cats.Parallel
import cats.effect.Sync

sealed trait Iterant[F[_], A] {

  final def parZipMap[G[_], B, C](rhs: Iterant[F, B])(f: (A, B) => C)
    (implicit F: Sync[F], P: Parallel[F, G]): Iterant[F, C] = ???

  final def zipMap[B, C](rhs: Iterant[F, B])(f: (A, B) => C)
    (implicit F: Sync[F]): Iterant[F, C] = ???
}
```

Here we have an actual use-case of `Parallel` in an API — we are describing a `parZipMap` that can do stuff in parallel. And ... our `G[_]` data type is ignored, it simply doesn't matter after solving that implicit instance, it's not used in the result of type `Iterant[F, C]` (in both cases), being an implementation detail (internally a different `map2` is used than the one provided by the `Sync` instance).

And when you get to implementing something like this, you want *code reuse*. Also internally, the reusable implementation looks like:

```scala
def zipMap[F[_], A, B, C](lh: Iterant[F, A], rh: Iterant[F, B])(f: (A, B) => C)
    (implicit F: Sync[F], A: Applicative[F]): Iterant[F, C]
```

So naturally we are just passing an extra `Applicative` around, which is easy and makes sense. However we need an `Applicative[F]`, but with a `Parallel[F, G]` what we have is an `Applicative[G]`, oops!

So the `ParallelApplicative` instance is like:

```scala
private[monix] final class ParallelApplicative[F[_], G[_]](implicit P: Parallel[F, G])
  extends Applicative[F] {

  override def pure[A](x: A): F[A] =
    P.monad.pure(x)
  override def unit: F[Unit] =
    P.monad.unit
  override def map[A, B](fa: F[A])(f: (A) => B): F[B] =
    P.monad.map(fa)(f)

  override def ap[A, B](ff: F[A => B])(fa: F[A]): F[B] =
    P.sequential(P.applicative.ap(P.parallel(ff))(P.parallel(fa)))
    P.sequential(P.applicative.product(P.parallel(fa), P.parallel(fb)))
  override def map2[A, B, Z](fa: F[A], fb: F[B])(f: (A, B) => Z): F[Z] =
    P.sequential(P.applicative.map2(P.parallel(fa), P.parallel(fb))(f))
}
```

Besides this being really annoying to build, I also hate the transformations `F ~> G` and `G ~> F` being used in case a `Parallel[Task, Task]` is in use, since those calls are totally unnecessary.

And as an optimization, if `mirror` was available on `cats.Parallel`, we could simply do this, avoiding any overhead:

```scala
class CatsParallelForTask extends Parallel[Task, Task.Par] {
  // ...
  def mirror = this.asInstanceOf[Parallel[Task, Task]]
}
```

If OOP is not good for such optimizations, then it's good for nothing.


alexandru:
Bump ...

Wanted to ask if this is of interest.
I think it needs tests.

The ability to work with `Parallel[F, F]`, given a `Parallel[F, G]`, is a useful optimization IMO, especially in Scala where the proliferation of type parameters and working with newtypes is not that good.

DavidGregory084:
@alexandru I would also have had use for something like this in an existing codebase where an error-accumulating map2 had been implemented for a monadic type

alexandru:
Also recent discussion with @gvolpe on Gitter: [March 21, 2018 2:07 PM](https://gitter.im/typelevel/cats-effect?at=5ab24b1a5f188ccc15de2308)

oleg-py:
It's a fair call that most uses of `Parallel` don't need the second type parameter, and using it generically complexes type signatures with another type parameter and explicit implicit argument (as compared to using context bound)

I don't like losing the second type completely, however. Bothers me to have both `Monad[F]` and `Applicative[F]` inside a same class, as well as non-descriptive transformations `parallel: F ~> F` and `sequential: F ~> F`. (How does it work? If I compose `parallel` with `parallel`, does it become doubly-paralel?).

Would rather have second type parameter of `Parallel` turned into to a type member and have `Parallel.Aux` for the few cases where it is really needed (although this won't happen in 1.x, but we might have `Parallel1` or something like that until next major release.

kailuowang:
another thing is that we can't add any methods to typeclass without breaking binary compatibility. So this will at least have to wait for a cats 2.0 which is binary compatible with 1.x only on scala 2.12

johnynek:
This seems fine to me. I think we can add it as an enrichment, since we really don't want a subclass giving us special behavior. Or we could add `def mirror` to the companion object.

alexandru:
@kubukoz TBH it doesn't address the issue.

When code sharing is involved between parallel and non-parallel operations, you really want an `Applicative[F]`, not a `Parallel[F, G]` or a `Parallel[F]` or an `Applicative[G]`.

alexandru:
I do understand if this issue doesn't make it due to type class coherency concerns. But type class coherency concerns won't stop users like myself from defining their own `mirror`, for as long as that's possible.

diesalbla:
@alexandru @LukaJCB  Given that the current `Parallel` and `NonEmptyParallel` only take one HK-type parameter, it seems as if changes in this PR will not be merged.

LukaJCB:
Indeed, closing this :)

PR #2020, opened 2017-11-07
Add `foldl` and `foldr` aliases to `Foldable`

felixmulder:
This PR adds aliases for Foldable's `foldRight` and `foldLeft` ops so that the callee can be sure that the correct method is being invoked.

Because of syntax extensions the end signature looks like:

```scala
def foldRight[B](b: Eval[B])(f: (Eval[B], A) => Eval[B]): Eval[B]
```

This can then lead to some confusion when folding over things which already have the `foldRight` operation, namely pretty much all collections.

Even though we do:

```scala
import cats.implicits._

List
  .range(0, 10000000)
  .foldRight(Eval.now(0))((n, e) => e.map(_ + n))
```

the fold won't be evaluated lazily, using `Foldable`'s definition of `foldRight` instead the stdlib's version will be chosen.

To avoid this pitfall, this PR introduces these aliases:

- `foldr <=> foldRight`
- `foldl <=> foldLeft`

these methods can be used safely with syntax extensions.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -188,7 +188,14 @@ lazy val docSettings = Seq(
 lazy val binaryCompatibleVersion = "1.0.0-RC1"
 
 def mimaSettings(moduleName: String) = Seq(
-    mimaPreviousArtifacts :=  Set("org.typelevel" %% moduleName % binaryCompatibleVersion))
+  mimaPreviousArtifacts := Set("org.typelevel" %% moduleName % binaryCompatibleVersion),
+  // TODO: remove this post-release of 1.0.0
+  mimaBinaryIssueFilters += {
+    import com.typesafe.tools.mima.core._
+    import com.typesafe.tools.mima.core.ProblemFilters._
+    exclude[ReversedMissingMethodProblem]("cats.syntax.FoldableSyntax.catsSyntaxFoldOps")
+  }
+)
 
 lazy val docs = project
   .enablePlugins(MicrositesPlugin)
--- a/core/src/main/scala/cats/Foldable.scala
+++ b/core/src/main/scala/cats/Foldable.scala
@@ -30,6 +30,27 @@ import simulacrum.typeclass
 
   /**
    * Left associative fold on 'F' using the function 'f'.
+   *
+   * Example:
+   * {{{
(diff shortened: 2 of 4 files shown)


kailuowang:
seems you were on a code base prior to 1.0.0-MF. These alias should probably be added to the type class , but they will break bin compat with 1.0RC1 so it would be helpful if there are more justification in the PR.

felixmulder:
Sure, I'll update the initial comment with a more detailed explanation.

kailuowang:
as expected, this fails the binary compat check
> Cats core: found 1 potential binary incompatibilities while checking against org.typelevel:cats-core_2.10:1.0.0-RC1  
> * method catsSyntaxFoldOps(java.lang.Object,cats.Foldable)java.lang.Object in trait cats.syntax.FoldableSyntax is present only in current version
filter with: ProblemFilters.exclude[ReversedMissingMethodProblem("cats.syntax.FoldableSyntax.catsSyntaxFoldOps")

we can add an exception filter [here](https://github.com/typelevel/cats/blob/master/build.sbt#L191)
[Here is the documentation on how](https://github.com/lightbend/migration-manager/wiki/Sbt-plugin#advanced-usage-filtering-binary-incompatibilities)
please also add a comment that this filter shall be removed post 1.0 release.

felixmulder:
Cool, thanks for having a look, will update this tomorrow :+1:

felixmulder:
@kailuowang - since the syntax suite doesn't actually run, I guess codecov doesn't pick it up. Is there somewhere I should add a runtime test for this so as not to affect codecov?

kailuowang:
@felixmulder a good way to add test coverage for such aliases is doctest. How about we mention these aliases and provide some doctests in scala docs for  `foldRight` and `foldLeft` in `Foldable`. Examples of doctests can be found also in the file [here](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Foldable.scala#L63-L87)

felixmulder:
@kailuowang - added the doctest :+1:

felixmulder:
NP, thanks for the pointers :)

PR #2021, opened 2017-11-09
[WIP] Add an UnorderedFoldable that we can give instances for Set/Map/&c

hrhino:
`Set`, e.g., is not a lawful `Foldable`, since iteration order between multiple sets can be observed via `toList`. If we promise never to observe the order in which elements are folded, by strengthening the `Monoid` required by `foldMap` to a `CommutativeMonoid`, we can have a broader type class `UnorderedFoldable` that contains such unordered collections.

I mentioned this on gitter and no one said no. It's a work in progress: 
- The name is ~~bikesheddable~~ debatable
- I need to add more tests, and possibly more instances
- We presumably need an `UnorderedFoldableSuite` somewhere in there
- Does this imply a ~~`TraverseRandomly`~~ `UnorderedTraverse`?

but I'm making a PR so people can say what they think about the idea _before_ I put any more effort into it, rather than after.

### DIFF
--- a/core/src/main/scala/cats/CommutativeApplicative.scala
+++ b/core/src/main/scala/cats/CommutativeApplicative.scala
@@ -1,5 +1,6 @@
 package cats
 
+import cats.kernel.CommutativeMonoid
 import simulacrum.typeclass
 
 /**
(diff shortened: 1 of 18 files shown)


kailuowang:
Are you aware of #1981?

hrhino:
Mm... Now I am. Sorry about that!

kailuowang:
@hrhino no worries at all. If you are interested in the topic, a review of #1981 from you will be greatly appreciated.

PR #2024, opened 2017-11-10
Optimize foldMap implementations with combineAll

carymrobbins:


alleycats-core/src/main/scala/alleycats/std/set.scala
```diff
@@ -68,6 +68,9 @@ object SetInstances {
       def foldRight[A, B](fa: Set[A], lb: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B] =
         Foldable.iterateRight(fa, lb)(f)
 
+      override def foldMap[A, B](fa: Set[A])(f: A => B)(implicit B: Monoid[B]): B =
+        B.combineAll(fa.map(f))
```
johnynek: `fa.iterator.map(f)` will be faster (not allocate a set) also maybe more correct. `fa.map(f)` if `f = { _ => 1 }` will create a set of 1 thing. If you do `foldMap` what do we mean? I actually don't know if this should return the count of items (which I think it should, which means we should do `fa.iterator.map`), or if it should return 0 if empty and 1 otherwise.

I guess since this is lawless especially we need a comment as to the behavior.

kailuowang:
Do you guys mind to provide some enlightenment here?  I must've missed something, I have trouble understanding how this optimize over the default implementation. 
 `combineAll(as)` is doing `as.foldLeft(empty)(combine)`, replace `as` with `as.iterator.map(f)`, then you get  `as.iterator.map(f).foldLeft(B.empty)(B.combine)`,
Default implementation of `as.foldMap` is doing 
`as.foldLeft(fa, B.empty)((b, a) => B.combine(b, f(a)))`
Why doing `map` first then `foldLeft` with `combine` is better than `foldLeft` with `combine`ing with `f(a)` on going?

tpolecat:
`combineAll` is optimized for some types, which fixes some N^2 cases (like `String` via `StringBuilder` or always concatenating lists right to left).

kailuowang:
thanks @tpolecat .

alleycats-core/src/main/scala/alleycats/std/set.scala
carymrobbins: That's a good point, I originally avoided using `.iterator` to maintain the same semantics, but I forgot to confirm that was the case.

```scala
@ alleycats.std.SetInstances.setTraverse.foldMap(Set(1,2,3))(_ => "a")
res13: String = "aaa"
```

Seems _not_ using `.iterator` actually would be a breaking change, so I'll switch to use it. Nice catch!

johnynek:
👍

PR #2026, opened 2017-11-13
Correctly close a tut:silent block in faq

vendethiel:
Here's how it looks like for me currently:

image


LukaJCB:
Thanks for contributing! :)

PR #2027, opened 2017-11-13
Rename Validation to Validated in Validated docs

Ttcao:
I'm learning Validated from the docs and think this may be a typo. 
Please let me know otherwise :)


LukaJCB:
Thanks for your contribution @Ttcao :)

PR #2029, opened 2017-11-14
make sure that EitherT  MonadError syntax works the old way

### LINKED ISSUE #2022: Can't use MonadError syntax for EitherT
With introduction (https://github.com/typelevel/cats/pull/1644/) of prioritised `MonadError` instance for `F: MonadError` in `EitherT[F, E, A]`, it's now impossible to use `MonadError` syntax for handling `E`. 
It makes it difficult to upgrade to 1.0.0-RC1 as this syntax is used quite a bit in a project.

```scala
implicit val merr: MonadError[IO, Throwable] = ???
type M[A] = EitherT[IO, Error, A]
val a: M[A] = ???
a.ensure(Error)(predicate) // type mismatch: found Error, required Throwable
```

kailuowang:
flipped the priority sequence between the two `MonadError` instances for `EitherT` so that the syntax goes back to the behavior prior to 1.0-RC1. If user need the `MonadError` syntax with the  `MonadError` instance that deals with the Error type on the inner `F`, they would need to create a local implicit instance that takes higher precedence, see the `RegressionSuite` for example
fixes #2022


tests/src/test/scala-2.11+/cats/tests/RegressionTests.scala
```diff
@@ -1,25 +0,0 @@
-package cats
-package tests
-
-
-import ExtraRegressionSuite._
-
-class ExtraRegressionSuite extends CatsSuite {
```
LukaJCB: Why are we deleting this? 🤔

kailuowang: oops I did the renamed but didn't include the renamed file in the git. just pushed. good catch.

kailuowang:
@LukaJCB, @johnynek I added an mima exception to get it pass the build, either of you has time to give a quick re-approval?

PR #2030, opened 2017-11-16
 added `collectFirst` and  `collectFirstSome` to `Foldable`

kailuowang:


core/src/main/scala/cats/Foldable.scala
```diff
@@ -215,6 +215,33 @@ import simulacrum.typeclass
...
+   * Like `collectFirst` from `scala.collection.Traversable` but takes `A => Option[B]`
+   * instead of `PartialFunction`s.
+   *
+   * To avoid conflict with the original collectFirst on many data types, there is a
+   * `collectFst` alias
+   *
+   * {{{
+   * scala> import cats.implicits._
+   * scala> val numbers = List(2,4,5,6,8,10)
+   * scala> numbers.collectFst(i => if(i % 2 == 1) Some(i.toDouble / 2d) else None)
```
LukaJCB: scalastyle wants a space between `if` and the parantheses.

core/src/main/scala/cats/Foldable.scala
```diff
@@ -215,6 +215,33 @@ import simulacrum.typeclass
...
+   * res0: Option[Double] = Some(2.5)
+   * scala> List(2, 4, 6).collectFst(i => if(i % 2 == 1) Some(i.toDouble / 2d) else None)
```
LukaJCB: Same as above :)

rossabaker:
I find it confusing to introduce a `collectFirst` that is different from the standard `collectFirst` that will shadow it on so many `Foldable`s.  I imagine only wanting to call this via the alias.

kailuowang:
@rossabaker that's a valid point. We had the same issue with `foldRight` as well, in which case we still keep the `foldRight` name and provide a `foldR` alias. I don't feel strongly about i, just try to be consistent. If you feel strongly about it I can remove the `collectFirst` and just keep `collectFst`

rossabaker:
My argument is diminished by the fact I've never noticed the `foldRight` clash.  I still think it's confusing, but I feel less strongly than I did two comments ago in light of that.

core/src/main/scala/cats/Foldable.scala
```diff
@@ -215,6 +215,33 @@ import simulacrum.typeclass
...
+   * }}}
+   */
+  def collectFirst[A, B](fa: F[A])(f: A => Option[B]): Option[B] =
+    foldRight(fa, Eval.now(Option.empty[B])) { (a, lb) =>
+      val fa = f(a)
+      if (fa.isDefined) Eval.now(fa) else lb
+    }.value
+
+  /**
+   * alias for collectFirst
```
johnynek: can we explain why this is here also (not just in the above comment).

core/src/main/scala/cats/Foldable.scala
```diff
@@ -215,6 +215,33 @@ import simulacrum.typeclass
       }
 
   /**
+   * Like `collectFirst` from `scala.collection.Traversable` but takes `A => Option[B]`
+   * instead of `PartialFunction`s.
```
johnynek: I think partial functions are very nice for this case. why not use them?

kailuowang: For a `PartialFunction`, when `isDefinedAt` and `apply` relies on a shared computation, that computation will have to be repeated for the hit case? For example, giving a list of ids, you want to find the first id with which you can find a record in a DB. A `PartialFunction` would course one extra DB query. Also `Id => Option[Record]` is just very nature, making it a `PartialFunction` would be awkward right?

rossabaker: :+1: Reviewing some old code, I've implemented it this way on a couple pre-cats projects.  I'd rather this suggestion than dropping it altogether.

rossabaker: `TraversableOnce` does [an unspeakable thing](https://github.com/scala/scala/blob/v2.12.4/src/library/scala/collection/TraversableOnce.scala#L141-L148) to avoid the double evaluation.

kailuowang: @rossabaker wow, that is unspeakable.  That certainly removed one of the incentives of adding this method. Just need a opposite version of `PartialFunction#lift`, maybe in mouse.

rossabaker: [Function.unlift](https://www.scala-lang.org/api/2.12.4/scala/Function$.html#unlift[T,R](f:T=%3EOption[R]):PartialFunction[T,R])?

kailuowang: done! closing this one.

rossabaker: Oh, I thought you were talking about removing the incentive for `collectFst`.  What about `Foldable` without `collectFirst`?  I've needed it on `NonEmptyList` before.

kailuowang: My mistake, my only incentive is for a A=>Option[B] collectFirst. I am reopening it. I wonder if we should add `collectFirst: PartialFunction[A, B]` and `collectFirstO: A => Option[B]`
For the default `collectFirst` implementation (which is going to be overriden for actual `TraversableOnce` data types)  shall we do the same trick as `TraversableOnce`?

johnynek:
I’m +1 on two methods. I’d be more verbose. collectFirstSome or something. Which is what we are doing in the Option returning case.

I’m also +1 on ugly hacks in libraries that speed things up for many users if they are safe. I haven’t studied this particular hack in detail however.

kailuowang:
Just made the change, I can't use the hacks because I couldn't get applyOrElse API to work with foldRight, however I did override it in our instances.

core/src/main/scala/cats/Foldable.scala
```diff
@@ -214,6 +214,31 @@ import simulacrum.typeclass
         case Right(_) => None
       }
 
+  def collectFirst[A, B](fa: F[A])(pf: PartialFunction[A, B]): Option[B] =
```
rossabaker: Is it worth mentioning that a more efficient implementation is often possible?  We do this on `size` and `foldM`, but I suppose it would be easy to get carried away with this.  I don't feel strongly.

rossabaker: My battery is going to die before I can run the tests, but I think this might be right:

```scala
    foldRight(fa, Eval.now(Option.empty[B])) { (a, lb) =>
      var result: Eval[Option[B]] = lb
      pf.runWith(b => result = Eval.now(Some(b)))(a)
      result
    }.value
```

rossabaker: This is the `TraversableOnce` trick in `foldRight` form.  `TraversableOnce` claims it's faster than `runWith`, but I have not benchmarked it in this context.  Tests pass even after I remove the overrides.  `sentinel` could be moved to a private member of the companion.

```scala
    val sentinel: Function1[A, Any] = new scala.runtime.AbstractFunction1[A, Any]{ def apply(a: A) = this }
    foldRight(fa, Eval.now(Option.empty[B])) { (a, lb) =>
      val x = pf.applyOrElse(a, sentinel)
      if (x.asInstanceOf[AnyRef] ne sentinel) Eval.now(Some(x.asInstanceOf[B]))
      else lb
    }.value
```

kailuowang: thanks a lot @rossabaker , I made the change.

core/src/main/scala/cats/Foldable.scala
```diff
@@ -214,6 +214,36 @@ import simulacrum.typeclass
...
+   * scala> keys.collectFirstSome(map.get)
+   * res0: Option[String] = Some(Four)
+   * scala> val map2 = Map(6 -> "Six", 7 -> "Seven")
+   * scala> keys.collectFirstSome(map2.get)
+   * res1: Option[String] = None
+   * }}}
+   */
+  def collectFirstSome[A, B](fa: F[A])(f: A => Option[B]): Option[B] =
+    foldRight(fa, Eval.now(Option.empty[B])) { (a, lb) =>
+      val fa = f(a)
```
johnynek: nit: can we not shadow `fa` with `f(a)` which is actually ` Option[B]` and not a `F[A]`?

kailuowang: oops. fixed.

kailuowang:
@rossabaker I didnt move it because it was parameterized to `A`, I just realized that it doesn't have to. So I moved it.

PR #2031, opened 2017-11-17
Add parallel docs

LukaJCB:
Adds some documentation for the new `Parallel` class. Right now it depends on #1938. We should be able to give more examples once we have a `Parallel` instance for `cats.effect.IO` :)

### DIFF
--- a/docs/src/main/tut/typeclasses/parallel.md
+++ b/docs/src/main/tut/typeclasses/parallel.md
@@ -1,0 +1,118 @@
+---
+layout: docs
+title:  "Parallel"
+section: "typeclasses"
+source: "core/src/main/scala/cats/Parallel.scala"
+scaladoc: "#cats.Parallel"
+---
+# Parallel
+
+When browsing the various `Monads` included in cats,
+you may have noticed that some of them have data types that are actually of the same structure,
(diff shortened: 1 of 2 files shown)


kailuowang:
Sorry I squashed the other pr, you might need to Cherry pick this one.

LukaJCB:
Yeah I'll do that soonish :)

docs/src/main/tut/typeclasses/parallel.md
```diff
@@ -0,0 +1,118 @@
...
+We can define a different `ap` for most of them and use the `parMapN` syntax for that:
+
+```tut
+(List(1, 2, 3), List(4, 5, 6)).parMapN(_ + _)
+```
+
+## NonEmptyParallel - a weakened Parallel
+
+If you recall that some types cannot form a `Monad` or an `Applicative` because it's not possible to implement the `pure` function for them.
+However, these types usually can have instances for `FlatMap` or `Apply`.
```
ceedubs: 2 minor comments:

- This might read a little more cleanly if you drop the "If you recall that". That is, it could start with "Some types cannot...".
- "usually" might be an overly strong assertion here. Maybe just say "often"?

LukaJCB: Done :)

PR #2032, opened 2017-11-20
Add TraverseM (Continuation of #1476)

LukaJCB:
Just merged this with Master for now. Not sure if we need other changes before merging. :)


kailuowang:
also need mima exception

johnynek:
So, I couldn’t get a test to pass before. It seems like a good idea, but I think even for List, which is an interesting Traverse case because it can eat very big, I didn’t get agreement.

There is also the naming concern. It is a bit junky to have traverse and traverseM. I may be in favor of punting this until we could find something similar to tailRecM for Applicative that would allow us to have stack safety for traverse.

LukaJCB:
Maybe of interest: https://github.com/paf31/purescript-safely

LukaJCB:
Going to close this for now :)

PR #2033, opened 2017-11-20
standardise on `liftF` and add `liftK` to transformers

SystemFw:
This PR deprecates `lift` on all transformers in favour of `liftF`, and adds a FunctionK version named `liftK` (for use with `mapK`).

I'm leaving a possible scalafix to a separate PR.

### DIFF
--- a/core/src/main/scala/cats/data/IndexedReaderWriterStateT.scala
+++ b/core/src/main/scala/cats/data/IndexedReaderWriterStateT.scala
@@ -266,6 +266,10 @@ private[data] sealed trait CommonIRWSTConstructors {
   /**
    * Return an effectful `a` and an empty log without modifying the input state.
    */
+  def liftF[F[_], E, L, S, A](fa: F[A])(implicit F: Applicative[F], L: Monoid[L]): IndexedReaderWriterStateT[F, E, L, S, S, A] =
+    IndexedReaderWriterStateT((_, s) => F.map(fa)((L.empty, s, _)))
+
+  @deprecated("Use liftF instead", "1.0.0-RC2")
   def lift[F[_], E, L, S, A](fa: F[A])(implicit F: Applicative[F], L: Monoid[L]): IndexedReaderWriterStateT[F, E, L, S, S, A] =
     IndexedReaderWriterStateT((_, s) => F.map(fa)((L.empty, s, _)))
 
@@ -570,7 +574,7 @@ private[data] sealed abstract class RWSTAlternative[F[_], E, L, S]
       G.combineK(x.run(e, sa), y.run(e, sa))
     }
 
-  def empty[A]: ReaderWriterStateT[F, E, L, S, A] = ReaderWriterStateT.lift(G.empty[A])
+  def empty[A]: ReaderWriterStateT[F, E, L, S, A] = ReaderWriterStateT.liftF(G.empty[A])
 
   def pure[A](a: A): ReaderWriterStateT[F, E, L, S, A] = ReaderWriterStateT.pure[F, E, L, S, A](a)
 
@@ -584,7 +588,7 @@ private[data] sealed abstract class RWSTMonadError[F[_], E, L, S, R]
 
   implicit def F: MonadError[F, R]
 
-  def raiseError[A](r: R): ReaderWriterStateT[F, E, L, S, A] = ReaderWriterStateT.lift(F.raiseError(r))
+  def raiseError[A](r: R): ReaderWriterStateT[F, E, L, S, A] = ReaderWriterStateT.liftF(F.raiseError(r))
 
   def handleErrorWith[A](fa: ReaderWriterStateT[F, E, L, S, A])(f: R => ReaderWriterStateT[F, E, L, S, A]): ReaderWriterStateT[F, E, L, S, A] =
     ReaderWriterStateT { (e, s) =>
--- a/core/src/main/scala/cats/data/IndexedReaderWriterStateT.scala
+++ b/core/src/main/scala/cats/data/IndexedReaderWriterStateT.scala
@@ -284,7 +284,7 @@ private[data] sealed trait CommonIRWSTConstructors {
       def apply[A](fa: F[A]): IndexedReaderWriterStateT[F, E, L, S, S, A] = IndexedReaderWriterStateT.liftF(fa)
     }
 
-  @deprecated("Use liftF instead", "1.0.0-RC2")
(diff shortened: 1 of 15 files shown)


core/src/main/scala/cats/data/IndexedReaderWriterStateT.scala
```diff
@@ -266,6 +266,25 @@ private[data] sealed trait CommonIRWSTConstructors {
...
+   * scala> b.value.runEmpty(true).value
+   * res0: (List[String], String, Option[Int]) = (List(),"",Some(1))
+   * }}}
+   */
+  def liftK[F[_], E, L, S](implicit F: Applicative[F], L: Monoid[L]): F ~> IndexedReaderWriterStateT[F, E, L, S, S, ?] =
+    new (F ~> IndexedReaderWriterStateT[F, E, L, S, S, ?]) {
+      def apply[A](fa: F[A]): IndexedReaderWriterStateT[F, E, L, S, S, A] = IndexedReaderWriterStateT.liftF(fa)
+    }
+
+  @deprecated("Use liftF instead", "1.0.0-RC2")
```
kailuowang: There probably won't be a rc2, the next one is 1.0

SystemFw: Yeah, wasn't quite sure. I'll fix tomorrow

SystemFw:
Fixed the `RC` deprecation warning. I'm adding the scalafix now (changed my mind, I'll do it in this PR)

kailuowang:
also need some mima exclusion
> [0m[[31merror[0m] [0m   filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("cats.data.CommonIRWSTConstructors.liftF")[0m
[0m[[31merror[0m] [0m * method liftK(cats.Applicative,cats.kernel.Monoid)cats.arrow.FunctionK in trait cats.data.CommonIRWSTConstructors is present only in current version[0m
[0m[[31merror[0m] [0m   filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("cats.data.CommonIRWSTConstructors.liftK")[0m
[0m[[31merror[0m] [0m * method liftF(java.lang.Object)cats.data.Kleisli in trait cats.data.KleisliFunctions is present only in current version[0m
[0m[[31merror[0m] [0m   filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("cats.data.KleisliFunctions.liftF")[0m
[0m[[31merror[0m] [0m * method liftK()cats.arrow.FunctionK in trait cats.data.KleisliFunctions is present only in current version[0m
[0m[[31merror[0m] [0m   filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("cats.data.KleisliFunctions.liftK")[0m
[0m[[31merror[0m] [0m * method liftF(java.lang.Object,cats.Applicative)cats.data.IndexedStateT in trait cats.data.CommonStateTConstructors is present only in current version[0m
[0m[[31merror[0m] [0m   filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("cats.data.CommonStateTConstructors.liftF")[0m
[0m[[31merror[0m] [0m * method liftK(cats.Applicative)cats.arrow.FunctionK in trait cats.data.CommonStateTConstructors is present only in current version[0m
[0m[[31merror[0m] [0m   filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("cats.data.CommonStateTConstructors.liftK")[0m

SystemFw:
@kailuowang Any quick reference to how you exclude something from MiMa or should I RTFM? :)

LukaJCB:
@SystemFw You can add them right here: https://github.com/typelevel/cats/blob/master/build.sbt#L212 :)

SystemFw:
Thanks a lot! @LukaJCB

aeons:
This is great. I've had to write the `OptionT` variant several times lately.

Is there a reason you're not using the kind-projector syntax?

```scala
def liftK[F[_]](implicit F: Functor[F]): F ~> OptionT[F, ?] = 
  λ[F ~> OptionT[F, ?]](OptionT.liftF(_))
```
(or `Lambda` instead of `λ`)

SystemFw:
@aeons Yes, http4s was the initial use case :)

There's no reason for not using it, I might change it as I fix the other things

SystemFw:
I think I'm mostly done, except my scalafix test doesn't compile ( it's telling me that `RWST` is not a member of `cats.data` ).
This happens because the scalafix is running on 0.9.0, and RWST was introduced later. How should I proceed?

P.S. the error in the travis seems spurious

kailuowang:
the migration is for migrating from 0.9. ideally we need a new set of migration that migrates from 1.0 RC1, although I am not sure a simple rename in RWST justify that. I propose we just remove RWST from the test.

SystemFw:
I think the errors are spurious and I should be done. The scalafix has fixes for both 0.9 and 1.0RC, the RC ones are not tested as discussed above

SystemFw:
Some of the builds fail with 

> Could not find any member to link for "mapK"

should I address this?

LukaJCB:
That seems odd, seems like an error during scaladoc compilation. Maybe someone else has an idea?

SystemFw:
Yeah, I reference `[[mapK]]` in a scaladoc, might be pretty stupid. Can't really test locally since MiMa fails before on my machine ("versions not set" so it's my setup, it succeeds on Travis)

LukaJCB:
I think it might be because `mapK` is on the instance while `liftF` is on the companion object, so they need to be referred to differently. Not sure how exactly though.

SystemFw:
I think we agree on what the problem is then :)
I also think we agree in not knowing what the solution is :P
I'll investigate (this sort of search and replace PRs always end up taking ages for me)

kailuowang:
Linking reference in scaladoc has always been problematic, the solutions people have been doing in the past, and I propose doing here, is to just remove the link.

SystemFw:
so just `mapK`  rather than [[[mapK]]`? Sounds good to me :)

SystemFw:
I need to fix the doctest for kleisli. It's the only bit of codecov that actually spotted a problem. Doing so right now

frroliveira:
If you're still interested in getting the scaladoc references working, you can prefix `mapK` with the class name, e.g. `EitherT.mapK`. That's what I used in `IorT.liftK`

kailuowang:
Close and reopen to trigger build

Issue #1940, opened 2017-09-29
[Proposal] new `cats.more`  module


kailuowang:
Relevant discussion can be found [here](https://gitter.im/typelevel/cats-dev?at=59ce495f19147ac323178653)

I am proposing creating a `cats.more` module (could be a separate repo) where we can place in non-core, no-so-complex features. Possible candidates for going into this module includes #1935, #1812, #1767,  #1755, #1730,  #1400, #921, #324 
(We shall still decide separately for each of these cases IRT where they should actually go)

It's for stuff that we can’t justify a single module for it and can’t find where it belongs. The API should still be stable to be included in this module though. We may or may not want to provide the same backward compatibility promise with this module as with `core`

Update: Renamed from "extra" to "more"

tpolecat:
I'm not convinced this is justifiable for any technical reason, but there may be a benefit in a "restrained" core just to keep the intimidation level down. However if we do this I would like it to be clear that it's not a bunch of random crap, but rather additional features that are less-commonly used. As for naming I kind of like `cats-core` and `cats-more` because it doesn't have the junk drawer connotation.

kailuowang:
I love the idea of naming it `cats-more`

LukaJCB:
I'll also refer to #1364, as it also has some type classes that seem like they could go into `cats-more` :)

Issue #1935, opened 2017-09-27
Add Divisible and Decideable (contravariant Applicative and co.)


adelbertc:
https://hackage.haskell.org/package/contravariant-1.4/docs/Data-Functor-Contravariant-Divisible.html

Example use case: `divide` can be seen as a way of deriving behavior by splitting a "larger" type in two

iravid:
Yes please! Very useful for encoders

LukaJCB:
Can you expand a bit on the example use case? I'm not really seeing its usefulness right now

iravid:
`Divide` is to `Applicative` as `Contravariant` is to `Functor`.

Let's say that we have an encoder trait:
```scala
trait Encoder[T] {
  def encode(t: T): String
}
```

Encoder has a contravariant functor as it is isomorphic to `? => String`. 

Given this definition of divide:
```
trait Divide[F[_]] extends Contravariant[F[_]] {
  def contramap2[A, B, C](fa: F[A], fb: F[B])(f: C => (A, B): F[C]
}
```

I could derive an encoder given two other encoders:
```scala
case class Record(i: Int, s: String)
val recEnc: Encoder[Record] = Divide[Record].contramap2(Encoder[Int], Encoder[String])((rec: Record) => (rec.i, rec.s))
```

LukaJCB:
Thanks @iravid, that indeed seems like a useful addition. However, I'm not sure that we'd have a lot of instances. The only one I can come up with right now would be `Show` and `Order`. I'm not really sure these are common enough to warrant adding them to core.

Maybe we could add them in some other module? WDYT?

iravid:
I wouldn't mind that, as long as they're available :-)

adelbertc:
Looking at https://hackage.haskell.org/package/contravariant-1.4/docs/Data-Functor-Contravariant-Divisible.html we can define instances for `Const`, the usual transformer "induction"/"back chaining", `Nested`, `Tuple2K`

stephen-lazaro:
Would anyone object if I were to try my hand at this? I'd be happy to give it a shot.

kailuowang:
This one is more likely to goto cats.more.

LukaJCB:
This can be closed now after #2034

LukaJCB:
Actually I closed to early as we do have an analogue to `Divisible`, but not to `Decidable`, so I'll reopen this :)

LukaJCB:
[...]

PR #2034, opened 2017-11-22
Add ContravariantMonoidal

stephen-lazaro:
Adds the `Divisible` typeclass as discussed in https://github.com/typelevel/cats/issues/1935.
I went ahead and opened this against `core` though it was suggested it might belong more in `more`, just because there seems to be thoughts and discussion around https://github.com/typelevel/cats/issues/1940 still. This way, we can get something that's relatively ready to merge when/if `more` is a thing. I'd be happy to move things over at that time.

Anyway, some notes:
- I switched the names from `divide` and `conquer` to `contramap2` and `unit` respectively, since that seems a little more in line with the naming standards in cats. Welcome to bike shedding on that front.

- While implementing, I noticed that the only real contribution `Divisible` gives is the presence of a left and right `unit` along the diagonal, since in point of fact `contramap2` is perfectly well implementable on `ContravariantSemigroupal` as:
```scala
def contramap2[A, B, C](fb: F[B], fc: F[C])(f: A => (B, C)): F[A] =
  contramap(product(fb, fc))(f)
```
I could see the impedance of adding another typeclass being high enough that maybe it'd be more reasonable just to add some enrichment to `ContravariantSemigroupal`. I could also see having both, or just `Divisible`. Not sure what the most desirable outcome would be here.

- I did not add an instance for `Show` because `unit` seems a little against the spirit of the thing, for that particular instance. The only lawful instance AFAICT is the constantly empty string `unit` induced by the `A => B` where `B : Monoid`.

- The branch name is a bit of a misnomer because I decided not to implement `Decideable` cause honestly I had trouble parsing the laws on that typeclass in the primary source here:
https://hackage.haskell.org/package/contravariant-1.4/docs/Data-Functor-Contravariant-Divisible.html#t:Divisible
I'll probably take another stab at it unless someone beats me to it.

- Added some `tut` documentation, but am happy to drop it entirely if its unwanted.

- I couldn't figure out how to the higher arity stuff works in `Apply` so I didn't take a shot implementing a dual here. If someone points me in the right direction and it is desired, I'd be happy to try.

- I added an attribution to `ekmett` in the Scala doc for `Divisible`. Hopefully, it satisfies that requirement, but if we have a format I've not met, please do let me know.

First time contributing to this repo so I'm sure I missed something. Thanks preemptively for your solicitous review, as well as your forbearance. :)


julienrf:
> I did not add an instance for Show because unit seems a little against the spirit of the thing, for that particular instance.

Indeed, for some typeclasses there is no sensible `unit` implementation. I think this is why we should have both `Divisible` and `ContravariantSemigroupal`.

Also, since we don’t use the names `divide` and `conquer` I’m not sure the `Divisible` name makes sense. To me it is more intuitive to think of it as a `ContravariantMonoidal` (since it extends `ContravariantSemigroupal` with a `unit` method).

core/src/main/scala/cats/Divisible.scala
```diff
@@ -0,0 +1,52 @@
...
+   */
+  def unit[A]: F[A]
+
+  /**
+   * `contramap2`
+   *
+   * Given two values in the Divisible context, and a function producing a value of both types,
+   * yields an element of the domain of the function lifted into the context.
+   */
+  def contramap2[A, B, C](fb: F[B], fc: F[C])(f: A => (B, C)): F[A]
```
julienrf: Why not provide a default implementation in terms of `contramap` and `product`? (like we do for covariant semigroupal: https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Apply.scala#L44-L50 )

stephen-lazaro: Sure, that makes sense, consistency is pretty valuable. I mostly went this approach based on the primary document.

stephen-lazaro:
Renamed to `ContravariantMonoidal` and used `product` and `contramap` to implement `contramap2` on `ContravariantSemigroupal`.

laws/src/main/scala/cats/laws/ContravariantMonoidalLaws.scala
```diff
@@ -0,0 +1,38 @@
...
+    (F.unit).contramap2(fa)(delta[A]) <-> fa
+
+  def contravariantMonoidalContramap2CompatibleContramapLeft[A, B, C](fa: F[A], f: B => (A, C)): IsEq[F[B]] =
+    fa.contramap2(F.unit)(f) <-> fa.contramap(f andThen (_._1))
+
+  def contravariantMonoidalContramap2CompatibleContramapRight[A, B, C](fa: F[A], f: C => (B, A)): IsEq[F[C]] =
+    (F.unit).contramap2(fa)(f) <-> fa.contramap(f andThen (_._2))
+
+  def contravariantMonoidalContramap2DiagonalAssociates[A](m: F[A], n: F[A], o: F[A]): IsEq[F[A]] =
+    (m.contramap2(n)(delta[A])).contramap2(o)(delta[A]) <-> m.contramap2(n.contramap2(o)(delta[A]))(delta[A])
```
julienrf: I think this law could go to `ContravariantSemigroupal`

core/src/main/scala/cats/Composed.scala
```diff
@@ -113,6 +113,19 @@ private[cats] trait ComposedContravariantCovariant[F[_], G[_]] extends Contravar
...
+  def F: Applicative[F]
+  def G: ContravariantMonoidal[G]
+
+  override def unit[A]: F[G[A]] = F.pure(G.unit)
+
+  override def contramap[A, B](fa: F[G[A]])(f: B => A) =
+    F.map(fa)(G.contramap(_)(f))
+
+  override def product[A, B](fa: F[G[A]], fb: F[G[B]]) =
+    F.map2(fa, fb)(G.product(_,_))
```
LukaJCB: We need a space after the comma here :)

core/src/main/scala/cats/Composed.scala
```diff
@@ -113,6 +113,19 @@ private[cats] trait ComposedContravariantCovariant[F[_], G[_]] extends Contravar
     F.contramap(fga)(gb => G.map(gb)(f))
 }
 
+private[cats] trait ComposedApplicativeContravariantMonoidal[F[_], G[_]] extends ContravariantMonoidal[λ[α => F[G[α]]]] { outer =>
+  def F: Applicative[F]
+  def G: ContravariantMonoidal[G]
+
+  override def unit[A]: F[G[A]] = F.pure(G.unit)
+
+  override def contramap[A, B](fa: F[G[A]])(f: B => A) =
```
LukaJCB: Can you add type annotations here and for `product`? :)

docs/src/main/tut/typeclasses/contravariantsemigroupal.md
```diff
@@ -0,0 +1,80 @@
...
+implicit val contravariantSemigroupalPredicate: ContravariantSemigroupal[Predicate] =
+  new ContravariantMonoidal [Predicate] {
+    def unit[A]: Predicate[A] = Predicate[A](Function.const(true))
+
+    def contramap2[A, B, C](fb: Predicate[B], fc: Predicate[C]
+      )(f: A => (B, C)): Predicate [A] = Predicate(
+      (a: A) => f(a) match {
+        case (b, c) => fb.run(b) && fc.run(c)
+      })
+  }
```
LukaJCB: tut fails here, because `contramap2` is missing an override modifier and `contramap` and `product` are unimplemented.

core/src/main/scala/cats/ContravariantSemigroupal.scala
```diff
@@ -12,4 +12,7 @@ import simulacrum.typeclass
       def F = self
       def G = Functor[G]
     }
+
+  def contramap2[A, B, C](fb: F[B], fc: F[C])(f: A => (B, C)): F[A] =
```
LukaJCB: I'm not exactly sure if we need this definition here, as we already have `contramap2`- `contramap22` in the auto-generated `SemigroupalArityFunctions`. WDYT?

LukaJCB: We could mix in the `SemigroupalArityFunctions` into the companion object of `ContravariantSemigroupal` maybe?

docs/src/main/tut/typeclasses/contravariantsemigroupal.md
```diff
@@ -0,0 +1,80 @@
+---
+layout: docs
+title: "Contravariant Semigroupal"
+section: "typeclasses"
+source: "core/src/main/scala/cats/ContravariantSemigroupal.scala"
+scaladoc: "#cats.ContravariantSemigroupal"
+---
+# Contravariant Semigroupal
```
LukaJCB: We're kind of mixing the `Monoidal` and the `Semigroupal` in this doc. Maybe we should fix it all to be `ContravariantMonoidal`?

docs/src/main/tut/typeclasses/contravariantsemigroupal.md
stephen-lazaro: Good catch, thank you. Nice to have `tut` catch these things.

core/src/main/scala/cats/ContravariantSemigroupal.scala
stephen-lazaro: Ah, yes! I agree. I didn't realize they were in `SemigroupalArityFunctions`. I'll do that.

docs/src/main/tut/typeclasses/contravariantsemigroupal.md
stephen-lazaro: Sounds good to me, I'll pin it all to `ContravariantMonoidal`.

stephen-lazaro:
Switched everything over to use the `SemigroupalArityFunctions` and the `TupleSemigroupalSyntax`.
Thanks for reviewing, both @julienrf and @LukaJCB.

LukaJCB:
Seems there's still some binary compatibly issues:

```
Cats core: found 2 potential binary incompatibilities while checking against org.typelevel:cats-core_2.10:1.0.0-RC1  (filtered 177)[0m
[0m[[31merror[0m] [0m * abstract method catsSyntaxContravariantSemigroupal(java.lang.Object,cats.ContravariantSemigroupal)cats.ContravariantSemigroupal#Ops in trait cats.syntax.ContravariantSemigroupalSyntax is inherited by class AllSyntax in current version.[0m
[0m[[31merror[0m] [0m   filter with: ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("cats.syntax.ContravariantSemigroupalSyntax.catsSyntaxContravariantSemigroupal")[0m
[0m[[31merror[0m] [0m * abstract method catsSyntaxContravariantMonoidal(java.lang.Object,cats.ContravariantMonoidal)cats.syntax.ContravariantMonoidalOps in trait cats.syntax.ContravariantMonoidalSyntax is inherited by class AllSyntax in current version.[0m
[0m[[31merror[0m] [0m   filter with: ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("cats.syntax.ContravariantMonoidalSyntax.catsSyntaxContravariantMonoidal")
```

If you could add those two to the build.sbt, I think we're done. Thanks again!

core/src/main/scala/cats/ContravariantSemigroupal.scala
```diff
@@ -13,3 +13,5 @@ import simulacrum.typeclass
       def G = Functor[G]
     }
 }
+object ContravariantSemigroupal extends SemigroupalArityFunctions {
```
LukaJCB: nitpick, but can we get rid of these curly braces? :)

LukaJCB:
I wonder, can we create instances for EitherK in the way we can for Applicative?

docs/src/main/tut/typeclasses/contravariantmonoidal.md
```diff
@@ -0,0 +1,110 @@
...
+---
+# Contravariant Monoidal
+
+The `ContravariantMonoidal` type class is for [`Contravariant`](contravariant.html) functors that can define a
+`product` function and a `unit` function.
+
+```tut:book:silent
+import cats.Contravariant
+
+trait ContravariantMonoidal[F[_]] extends Contravariant[F] {
```
LukaJCB: Maybe we shouldn't use `tut` here as it shadows the real `ContravariantMonoidal`.

core/src/main/scala/cats/ContravariantSemigroupal.scala
stephen-lazaro: Hah, yeah, for sure.

stephen-lazaro:
Yeah, that seems like a no brainer. I'll do that. Thanks for taking care of the binary compatibility issues. I think I'm not running the check correctly locally, because it keeps telling me it passes 🤔

LukaJCB:
Yeah, you probably need to switch to scala 2.11.11 to see the breakage.

docs/src/main/tut/typeclasses/contravariantmonoidal.md
kailuowang: we could still use `tut` here then add a `tut:reset` after ([explained here](https://github.com/tpolecat/tut#modifiers)) so that it has no effect after.

stephen-lazaro:
No problem! I've realized why code coverage dropped a little and think I can fix it.

By the way, would it be preferred that I interactively rebase this down, or do you guys just tend to squash merge things in?

kailuowang:
we squash merge, did you see my comment on using `tut:reset`?

stephen-lazaro:
Just did, will do :+1: Thanks for the pointer.

docs/src/main/tut/typeclasses/contravariantmonoidal.md
```diff
@@ -0,0 +1,110 @@
...
+section: "typeclasses"
+source: "core/src/main/scala/cats/ContravariantMonoidal.scala"
+scaladoc: "#cats.ContravariantMonoidal"
+---
+# Contravariant Monoidal
+
+The `ContravariantMonoidal` type class is for [`Contravariant`](contravariant.html) functors that can define a
+`product` function and a `unit` function.
+
+```tut:reset
```
kailuowang: Sorry I didn't make it clear. I was suggesting using a `tut:silent` here and `tut:reset:silent` for the next block at line 41.

stephen-lazaro: Ah, I see, I was just starting to debug this. Ok, cool, sorry about this, I'm not used to testing against multiple versions of the compiler. I'll fix shortly.

stephen-lazaro: Ok, checked it against 2.10.6, 2.11.11, and 2.12.3 and the docs all compile with that recommendation. So we should be GTG.

LukaJCB:
I was sort of negative on this being in cats-core before, but I think it makes a lot of sense especially how we formulated it. It really makes the relationship between Semigroups/Monoids and Semigroupals/Monoidals/Applicatives much clearer for me. So kudos to all! 🎉

stephen-lazaro:
Awesome! Thanks for the help all involved. Was fun to try my hand at something like this.

core/src/main/scala/cats/ContravariantMonoidal.scala
```diff
@@ -0,0 +1,32 @@
...
+   * `unit` produces an instance of `F` for any type `A`
+   * that is trivial with respect to `contramap2` along
+   * the diagonal
+   */
+  def unit[A]: F[A]
+
+  def liftContravariant[A, B](f: A => B): F[B] => F[A] =
+    ContravariantMonoidal.contramap2(unit[B], _: F[B])(((b: B) => (b, b)) compose f)(self, self)
+
+  // Technically, this is not correct, as the Applicative is composed with the ContravariantMonoidal, not the other way around
```
kailuowang: Would you mind elaborate a bit more (or point to a more detailed discussion/explanation) here?

stephen-lazaro: Sure, I can also just change the name and move this possibly. The point here is mostly that I didn't want to add anything to `Applicative` about `ContravariantMonoidal` since this might live in `cats.more`, so I ended up having this composition live here in `ContravariantMonoidal` even though the composition is `Applicative compose ContravariantMonoidal`, so the name and the signature are a little out of sync with what I saw in the rest of the repository.

kailuowang: ah, I propose we rename and move it to `Applicative`

stephen-lazaro: Got it, will do. Thanks.

core/src/main/scala/cats/data/Kleisli.scala
```diff
@@ -165,6 +165,9 @@ private[data] sealed abstract class KleisliInstances1 extends KleisliInstances2
 private[data] sealed abstract class KleisliInstances2 extends KleisliInstances3 {
   implicit def catsDataAlternativeForKleisli[F[_], A](implicit F0: Alternative[F]): Alternative[Kleisli[F, A, ?]] =
     new KleisliAlternative[F, A] { def F: Alternative[F] = F0 }
+
+  implicit def catsDataContravariantMonoidalForKleisli[F[_], A](implicit F0: ContravariantMonoidal[F]): ContravariantMonoidal[Kleisli[F, A, ?]] =
```
kailuowang: The convention in cats is that instances that are more specific should be at higher priority. I.E. instance of `ContravariantMonoidal` should be lower in the inheritance chain than `Contravariant` and `ContravriantSemigroupal` and `Semigroupal`.

stephen-lazaro: Gotcha, will fix.

kailuowang:
The code look good for me. But I don't know the area well enough, so I am not comfortable for my sign-off to be the final one for merge. @iravid @julienrf @adelbertc would anyone of you guys give a quick review? (You guys can focus just on the API the laws)

laws/src/main/scala/cats/laws/ContravariantMonoidalLaws.scala
```diff
@@ -0,0 +1,38 @@
...
+    (F.unit[A], fa).contramapN(delta[A]) <-> fa
+
+  def contravariantMonoidalContramap2CompatibleContramapLeft[A, B, C](fa: F[A], f: B => (A, C)): IsEq[F[B]] =
+    (fa, F.unit[C]).contramapN(f) <-> fa.contramap(f andThen (_._1))
+
+  def contravariantMonoidalContramap2CompatibleContramapRight[A, B, C](fa: F[A], f: C => (B, A)): IsEq[F[C]] =
+    (F.unit[B], fa).contramapN(f) <-> fa.contramap(f andThen (_._2))
+
+  def contravariantMonoidalContramap2DiagonalAssociates[A](m: F[A], n: F[A], o: F[A]): IsEq[F[A]] =
+    ((m, n).contramapN(delta[A]), o).contramapN(delta[A]) <-> (m, (n, o).contramapN(delta[A])).contramapN(delta[A])
```
julienrf: Why can’t we move this law to `ContravariantSemigroupal`? It seems that we don’t need `unit` here.

stephen-lazaro: Ah, we probably could. Looks like I overlooked your prior comment, sorry about that. I can move that over.

core/src/main/scala/cats/instances/order.scala
```diff
@@ -2,21 +2,31 @@ package cats
...
+            fa.compare(f(x), f(y))
+        }
 
       def product[A, B](fa: Order[A], fb: Order[B]): Order[(A, B)] =
         new Order[(A, B)] {
-          def compare(x: (A, B), y: (A, B)): Int = {
-            val z = fa.compare(x._1, y._1)
-            if (z == 0) fb.compare(x._2, y._2) else z
-          }
+          def compare(x: (A, B), y: (A, B)): Int =
```
iravid: Same comment re allocations

core/src/main/scala/cats/instances/order.scala
```diff
@@ -2,21 +2,31 @@ package cats
...
+       * Provides trivial order
+       */
+      def unit[A]: Order[A] = Order.from[A]((x: A, y: A) => 0)
       /** Derive an `Order` for `B` given an `Order[A]` and a function `B => A`.
        *
        * Note: resulting instances are law-abiding only when the functions used are injective (represent a one-to-one mapping)
        */
-      def contramap[A, B](fa: Order[A])(f: B => A): Order[B] = Order.by[B, A](f)(fa)
+      def contramap[A, B](fa: Order[A])(f: B => A): Order[B] =
+        new Order[B] {
```
iravid: Same re delegation to `Order.by`

core/src/main/scala/cats/instances/ordering.scala
```diff
@@ -2,20 +2,28 @@ package cats
...
        * Note: resulting instances are law-abiding only when the functions used are injective (represent a one-to-one mapping)
        */
-      def contramap[A, B](fa: Ordering[A])(f: B => A): Ordering[B] = fa.on(f)
+
+      def unit[A]: Ordering[A] = new Ordering[A] {
+        def compare(l: A, r: A): Int = 0
+      }
+
+      def contramap[A, B](fa: Ordering[A])(f: B => A): Ordering[B] =
+        new Ordering[B] {
```
iravid: Same re delegation to `on`

core/src/main/scala/cats/instances/ordering.scala
```diff
@@ -2,20 +2,28 @@ package cats
...
+          def compare(x: B, y: B) = fa.compare(f(x), f(y))
+        }
 
       def product[A, B](fa: Ordering[A], fb: Ordering[B]): Ordering[(A, B)] =
         new Ordering[(A, B)] {
-          def compare(x: (A, B), y: (A, B)): Int = {
-            val z = fa.compare(x._1, y._1)
-            if (z == 0) fb.compare(x._2, y._2) else z
+          def compare(x: (A, B), y: (A, B)): Int = (x, y) match {
+            case ((aL, bL), (aR, bR)) => {
```
iravid: Same re allocations

core/src/main/scala/cats/instances/eq.scala
```diff
@@ -2,11 +2,30 @@ package cats
...
+       * Note: resulting instances are law-abiding only when the functions used are injective (represent a one-to-one mapping)
+       */
+      def contramap[A, B](fa: Eq[A])(f: B => A): Eq[B] =
+        Eq.instance { (l: B, r: B) =>
+          fa.eqv(f(l), f(r))
+        }
 
       def product[A, B](fa: Eq[A], fb: Eq[B]): Eq[(A, B)] =
-        Eq.instance { (left, right) => fa.eqv(left._1, right._1) && fb.eqv(left._2, right._2) }
+        Eq.instance { (l, r) =>
```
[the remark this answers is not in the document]
stephen-lazaro: I guess that's true, I sort of forgot that aspect of things. :+1:

stephen-lazaro:
Thanks @iravid! I'll take care of those tonight.

kailuowang:
Thanks so much @stephen-lazaro for this awesome contribution!

stephen-lazaro:
:tada: Thanks everyone for the assistance!

PR #2036, opened 2017-11-23
Clean up applicative syntax doc

bkirwi:
The code was updated to use the new syntax, but the text no longer matched up with the code. This brings them in sync once again.


johnynek:
Thanks Ben!

PR #2037, opened 2017-11-23
Tests: MonadCombine->Alternative, add missing ones

vendethiel:
Wrt `guard` test... This one doesn't seem amazing, but I wasn't too sure what else to test.


vendethiel:
I guess they're [technically went through here](https://github.com/typelevel/cats/blob/6863f3fb36d17cf2e9548ec2ab6f52b46f9c2530/tests/src/test/scala/cats/tests/SyntaxSuite.scala#L291-L297).

tests/src/test/scala/cats/tests/AlternativeSuite.scala
```diff
@@ -2,6 +2,14 @@ package cats
 package tests
 
 class MonadCombineSuite extends CatsSuite {
```
LukaJCB: How about renaming the class as well? :)

tests/src/test/scala/cats/tests/AlternativeSuite.scala
```diff
@@ -11,4 +19,10 @@ class MonadCombineSuite extends CatsSuite {
       Alternative[List].separate(list) should === (expected)
     }
   }
+
+  test("guard") {
+    forAll { (b: Boolean) =>
+      Alternative[Option].guard(b).isDefined should === (b)
```
LukaJCB: This test case will probably run 100 times, but only two values are actually tested. I'd rather generate different valid `Alternative` instances.

tests/src/test/scala/cats/tests/AlternativeSuite.scala
vendethiel: Oops, nice catch!

tests/src/test/scala/cats/tests/AlternativeSuite.scala
vendethiel: I'm not sure how to do that with `forall`. Do you have an example testcase I can look into and copy the behavior of?

johnynek: You don’t need forall. Just write a def for the law, then call it for two inputs.

vendethiel:
Pushed this version. I can add more tests that work the way @LukaJCB wanted if someone can give me some pointers :).

vendethiel:
Thanks!

PR #2038, opened 2017-11-24
Add &> and <& as syntax for Parallel

### LINKED ISSUE #2015: Add a parallel *> and <*
Probably two very useful operations, however giving these a good "operator name" might be fairly difficult. For now, I'll suggest something like `!>` and `<!`, but bikeshedding very welcome!

LukaJCB:
Should fix #2015. I chose `&>` and `<&` seemingly at random. I wanted to use `!>` before, but it appears that's illegal, as it's XML or something. If someone has better names for these operations, I'd be glad to hear! :)


johnynek:
👍

PR #2040, opened 2017-11-24
Remove ApplicativeError#catchNonFatalEval (fixes #2006)

### LINKED ISSUE #2006: Remove ApplicativeError#catchNonFatalEval
[`ApplicativeError#catchNonFatalEval`](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/ApplicativeError.scala#L128) uses `Eval` for effect capture. However, as far as I can tell, `Eval` is not for effect capture. For related discussion, see https://github.com/typelevel/cats-effect/issues/78 (possibly that discussion should be continued, instead of starting a parallel one here).

/cc @djspiewak @alexandru

durban:
Remove ApplicativeError#catchNonFatalEval (fixes #2006)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -209,7 +209,11 @@ def mimaSettings(moduleName: String) = Seq(
       exclude[ReversedMissingMethodProblem]("cats.Foldable.collectFirstSome"),
       exclude[ReversedMissingMethodProblem]("cats.Foldable.collectFirst"),
       exclude[ReversedMissingMethodProblem]("cats.Foldable#Ops.collectFirstSome"),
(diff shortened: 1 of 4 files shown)


diesalbla:
@ceedubs It seems that you already closed the original issue that requested this change, but did not close this PR.

ceedubs:
@diesalbla  thanks good catch.

Quite a while ago I inadvertently gave up my own cats repo write permissions when I was making some changes to my GitHub settings. It's probably for the best though, as I'm not active in cats maintenance these days. Perhaps one of the maintainers such as @travisbrown would want to close this issue out?

PR #2043, opened 2017-11-25
Removed deprecation of >> and changed its param to be a by-name

mpilquist:
In FS2, we rely heavily on recursive functions that return values of type `Pull[F,O,R]`. These functions typically take a form similar to:

```scala
def foo: Pull[F,O,R] =
  if (guard) Pull.done else doSomeWork *> foo
```

Recently, @pchlupacek noticed that FS2 defines `*>` with a by-name param whereas cats defines `*>` with a strict param. Such recursive programs are broken if we use a strict param. Hence, we are going to change the FS2 definition of `*>` to be strict and reintroduce `>>` with a by-name param and then use `>>` in all recursive pulls.

Ideally, cats would have the same definition for `>>`, so in this PR I removed the deprecation of `>>` and defined it with a by-name param, making it a true alias for `flatMap(_ => fb)`. Note that we already have `followedByEval` and `forEffectEval`, though the syntax overhead of using those is way too high IMO.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -247,7 +247,15 @@ def mimaSettings(moduleName: String) = Seq(
       exclude[ReversedMissingMethodProblem]("cats.Foldable#Ops.collectFirst"),
       exclude[ReversedMissingMethodProblem]("cats.NonEmptyParallel.parForEffect"),
       exclude[ReversedMissingMethodProblem]("cats.NonEmptyParallel.parFollowedBy"),
-      exclude[ReversedMissingMethodProblem]("cats.syntax.ParallelSyntax.catsSyntaxParallelAp")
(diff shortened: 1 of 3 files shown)


LukaJCB:
I'm generally +1 on this, but should we maybe move it to `Apply`?

mpilquist:
I thought about that but settled on leaving `>>` on `FlatMap` since the guarantee is that `fb` won't be computed until `fa` produces an `A` (precisely what `flatMap` describes).

LukaJCB:
I see, yeah I think that makes sense! :)

LukaJCB:
I think you'll need to add a mima exception

mpilquist:
Yeah, just pushed a mima fix.

PR #2045, opened 2017-11-25
Fix scalafix testing instructions

kubukoz:


ChristopherDavenport:
I've always used `sbt "; ...;..."` when running multiple commands. I'm surprised that either of the above work well.

kubukoz:
Well, it doesn't work on Mac as it is in the repo. I'm still figuring out the correct way, though ;)

kubukoz:
In fact, I don't see the project `coreFree` at all...

kubukoz:
`sbt coreJVM/publishLocal freeJVM/publishLocal` should do the job if I'm not mistaken in thinking that `freeJVM` is the correct replacement for `coreFree`. I'll update the PR once it's confirmed.

cc @LukaJCB, by chance?

kubukoz:
Ahh, we need `macrosJVM/publishLocal` and `kernelJVM/publishLocal` as well

kubukoz:
With all these subprojects we need to publish, maybe there's a way to just publish all things JVM? :)

kailuowang:
On Travis it runs 
```
sbt ;coreJVM/publishLocal;freeJVM/publishLocal
cd scalafix
sbt tests/test
cd ..
```

kubukoz:
Sorry, I didn't see your comment before.

Well, running `sbt ;coreJVM/publishLocal;freeJVM/publishLocal` doesn't seem to work with either zsh or bash on Mac (iterm2 / terminal are both affected). It just opens the sbt shell...

It seems right that for `tests/test` we only need coreJVM and freeJVM, so this should be portable enough, and I suggest to change the readme to:
`sbt coreJVM/publishLocal freeJVM/publishLocal`

the other modules were needed for the whole `test` task (and most likely for correct depedency resolution in IDEA). We might want to mention that in the readme too...

kailuowang:
thanks very much! since this is very straightforward documentation correction, I am going to merge with 1 sign-off

PR #2046, opened 2017-11-26
Add distributive typeclass and some instances

### LINKED ISSUE #2014: Add Distributive typeclass
There is a typeclass call `Distributive` that is the dual of `Traverse`

```scala
trait Distributive[F[_]] extends Functor[F] {
  def distributel[G[_]:Functor,A,B](fa: G[A])(f: A => F[B]): F[G[B]]
}
```

The feature that's really handy with this encoding is that it's stack safe to distribute over a `Functor` of arbitrary "size". For instance:

```scala
scala> val f = {x:Int => x +1}
f: Int => Int = $$Lambda$1271/1964149457@6dbabd33

scala> val funcs = List.fill(10000)(f)
funcs: List[Int => Int] = List($$Lambda$1271/1964149457@6dbabd33, $$Lambda$1271/1964149457@6dbabd33, $$Lamb...)

scala> funcs.distribute(identity).apply(1)
res15: List[Int] = List(2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2 ...)

scala> funcs.sequence.apply(1)  //Stack overflow
```

Would this typeclass be welcome in cats?

coltfred:
- Started working on this to fix #2014

Things still left: 

- [x] Laws and Tests
  - [x] Function0 Test 
  - [x] Function1 Test 
  - [x] Nested Test 
  - [x] Kleisli Test 
  - [x] Tuple2K Test
- [x] Syntax that makes sense
- [x] More instances?

I started on it this weekend and wanted to push it up to start getting feedback as well as help on the syntax.  Distributive is weird in that it's adding syntax to the `Functor`, not to the type that has the `Distributive` instance. 

We want to add `distribute[G[_]: Distributive, A, B](f: A => G[B]): G[F[B]]` to the functor `F`. This means that we should be able to turn `List[Int => Int]` into `Int => List[Int]`, but I don't know how to add the syntax to the Functor that's passed to it.


LukaJCB:
Hey Colt, for the syntax you can do something like this:

```scala
trait DistributiveSyntax extends Distributive.ToDistributiveOps {
  implicit final def catsSyntaxDistributiveOps[F[_]: Functor, A](fa: F[A]): DistributiveOps[F, 
 A] =
    new DistributiveOps[F, A](fa)
}

final class DistributiveOps[F[_], A](val fa: F[A]) extends AnyVal {
  def distribute[G[_]: Distributive, A, B](f: A => G[B]): G[F[B]] = /* ... */
}
```

LukaJCB:
I think we can also do instances for Tuple2K :)

core/src/main/scala/cats/Composed.scala
```diff
@@ -1,6 +1,14 @@
 package cats
 
 
+private [cats] trait ComposedDistributive[F[_], G[_]] extends Distributive[λ[α => F[G[α]]]] with ComposedFunctor[F, G] { outer =>
```
LukaJCB: Scalastyle doesn't like the space before `[cats]` here.

core/src/main/scala/cats/syntax/DistributiveSyntax.scala
```diff
@@ -0,0 +1,14 @@
...
+package syntax
+
+import cats.evidence.===
+
+trait DistributiveSyntax extends Distributive.ToDistributiveOps {
+  implicit final def catsSyntaxDistributiveOps[F[_]: Functor, A](fa: F[A]): DistributiveOps[F, A] = new DistributiveOps[F, A](fa)
+}
+
+// Add syntax to functor as part of importing distributive syntax.
+final class DistributiveOps[F[_]: Functor, A](val fa: F[A]) {
```
LukaJCB: Can we make this an AnyVal? :)

core/src/main/scala/cats/syntax/DistributiveSyntax.scala
```diff
@@ -0,0 +1,14 @@
+package cats
+package syntax
+
+import cats.evidence.===
+
+trait DistributiveSyntax extends Distributive.ToDistributiveOps {
+  implicit final def catsSyntaxDistributiveOps[F[_]: Functor, A](fa: F[A]): DistributiveOps[F, A] = new DistributiveOps[F, A](fa)
```
LukaJCB: Maybe we should add another class for `cosequence`, we wouldn't need `cats.evidence` then:

```scala
final class CosequenceOps[F[_]: Functor, G[_]: Distributive, A](val fga: F[G[A]]) extends AnyVal {
  def cosequence: G[F[A]] = G.cosequence(fga)
}
```

core/src/main/scala/cats/syntax/DistributiveSyntax.scala
coltfred: No, because the class has the Functor constraint.

core/src/main/scala/cats/syntax/DistributiveSyntax.scala
coltfred: Hmmm, not sure I see a reason why that's better? The evidence constraint is on just the one method and it saves the extra class for syntax. Is there a reason it should be preferred?

core/src/main/scala/cats/Composed.scala
coltfred: fixed.

coltfred:
Laws and tests are all that's needed. Sorry I haven't been able to hammer those out it's been a busy week. I'll try and get the laws written tomorrow.

core/src/main/scala/cats/syntax/DistributiveSyntax.scala
LukaJCB: Ah yes, you're right! We should move the `Functor` constraint to the method instead

core/src/main/scala/cats/syntax/DistributiveSyntax.scala
LukaJCB: Maybe you're right, disregard me! :D

coltfred:
@LukaJCB I think everything should be in there, now. Let me know if you have problems with anything.

LukaJCB:
Hey Colt, looks really great, thanks again! I think it's ready to merge, but you'll need to add some MiMa Exceptions :)

core/src/main/scala/cats/data/Kleisli.scala
```diff
@@ -228,11 +228,16 @@ private[data] sealed abstract class KleisliInstances6 extends KleisliInstances7
...
 }
 
-private[data] sealed abstract class KleisliInstances7 {
+private[data] sealed abstract class KleisliInstances7 extends KleisliInstances8 {
   implicit def catsDataFunctorForKleisli[F[_], A](implicit F0: Functor[F]): Functor[Kleisli[F, A, ?]] =
     new KleisliFunctor[F, A] { def F: Functor[F] = F0 }
 }
 
+private[data] sealed abstract class KleisliInstances8 {
+  implicit def kleisliDistributive[F[_], R](implicit F0: Distributive[F]): Distributive[Kleisli[F, R, ?]] =
```
kailuowang: The Cats convention is that instances that are more specific should have higher priority. So we should probably swap the `Functor` and `Distributive`

core/src/main/scala/cats/data/Tuple2K.scala
```diff
@@ -108,20 +108,36 @@ private[data] sealed abstract class Tuple2KInstances6 extends Tuple2KInstances7
...
 
   implicit def catsDataFunctorForTuple2K[F[_], G[_]](implicit FF: Functor[F], GG: Functor[G]): Functor[λ[α => Tuple2K[F, G, α]]] = new Tuple2KFunctor[F, G] {
     def F: Functor[F] = FF
     def G: Functor[G] = GG
   }
 }
 
+private[data] sealed abstract class Tuple2KInstances8 {
+
+  implicit def catsDataDistributiveForTuple2K[F[_], G[_]](implicit FF: Distributive[F], GG: Distributive[G]): Distributive[λ[α => Tuple2K[F, G, α]]] = new Tuple2KDistributive[F, G] {
```
kailuowang: Same as in `Kleisli`, need to swap the two instances.

core/src/main/scala/cats/data/Nested.scala
```diff
@@ -167,6 +167,11 @@ private[data] sealed abstract class NestedInstances12 {
     new NestedInvariant[F, G] {
       val FG: Invariant[λ[α => F[G[α]]]] = Invariant[F].composeContravariant[G]
     }
+
+  implicit def catsDataDistributiveForNested[F[_]: Distributive, G[_]: Distributive]: Distributive[Nested[F, G, ?]] =
```
kailuowang: It needs to go above the `Functor` instance

core/src/main/scala/cats/data/Tuple2K.scala
kailuowang: Same as `Kleisli`

core/src/main/scala/cats/data/Kleisli.scala
kailuowang: According to  naming convention of implicits, it should be `catsDataDistributiveForKleisli`

core/src/main/scala/cats/instances/function.scala
```diff
@@ -92,11 +101,16 @@ private[instances] sealed trait Function1Instances extends Function1Instances0 {
...
 
-
 private[instances] sealed trait Function1Instances0 {
   implicit def catsStdContravariantForFunction1[R]: Contravariant[? => R] =
     new Contravariant[? => R] {
       def contramap[T1, T0](fa: T1 => R)(f: T0 => T1): T0 => R =
         fa.compose(f)
     }
+
+  implicit def functior1Distributive[T1]: Distributive[T1 => ?] = new Distributive[T1 => ?] {
```
kailuowang: `functior1Distributive` -> `catsStdDistributiveForFunction1`

core/src/main/scala/cats/data/Kleisli.scala
coltfred: Ok.

core/src/main/scala/cats/data/Tuple2K.scala
coltfred: ok.

core/src/main/scala/cats/data/Nested.scala
coltfred: ok.

tests/src/test/scala/cats/tests/KleisliSuite.scala
```diff
@@ -113,6 +113,11 @@ class KleisliSuite extends CatsSuite {
     checkAll("Functor[Kleisli[Option, Int, ?]]", SerializableTests.serializable(Functor[Kleisli[Option, Int, ?]]))
   }
 
+  {
+    checkAll("Kleisli[Function0, Int, ?]", DistributiveTests[Kleisli[Function0, Int, ?]].distributive[Int, Int, Int, Option, Id])
+    checkAll("Distributive[Kleisli[Option, Int, ?]]", SerializableTests.serializable(Distributive[Kleisli[Function0, Int, ?]]))
```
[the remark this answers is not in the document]
coltfred: ok.

tests/src/test/scala/cats/tests/NestedSuite.scala
```diff
@@ -153,4 +153,10 @@ class NestedSuite extends CatsSuite {
     checkAll("Nested[ListWrapper, Option, ?]", MonoidKTests[Nested[ListWrapper, Option, ?]].monoidK[Int])
     checkAll("MonoidK[Nested[ListWrapper, Option, ?]]", SerializableTests.serializable(MonoidK[Nested[ListWrapper, Option, ?]]))
   }
+
+  {
+    //Distributive composition
+    checkAll("Nested[Function0, Function0, ?]", DistributiveTests[Nested[Function0, Function0, ?]].distributive[Int, Int, Int, Option, Function0])
```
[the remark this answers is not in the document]
coltfred: I ran into an error about Eq instances because `distributive` in the Tests needs to be able to compare. I gave up and thought this was a reasonable compromise.

kailuowang:
scalastyle
>  /home/travis/build/typelevel/cats/core/src/main/scala/cats/data/Nested.scala:142:0: Whitespace at end of line

tests/src/test/scala/cats/tests/NestedSuite.scala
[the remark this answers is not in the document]
kailuowang: Ah, that's due to the Eq instance for Function1 isn't provided by default
You can do the following 

```scala
    import cats.laws.discipline.eq._
    //Distributive composition
    checkAll("Nested[Function1[Int, ?], Function0, ?]", DistributiveTests[Nested[Function1[Int, ?], Function0, ?]].distributive[Int, Int, Int, Option, Function0])

```
That passes.

PR #2047, opened 2017-11-26
CommutativeMonoid instance for SortedMap

alonsodomin:
This PR adds a `CommutativeMonoid` instance for `SortedMap` when there is a `CommutativeSemigroup` instance for the value type.

Besides adding a test for `CommutativeMonoid[SortedMap[String, Int]]`, I also modified the test for `Monoid[SortedMap[String, Int]]` to be a `Monoid[SortedMap[String, String]]`, such that I can ensure that the `Monoid` instance can still be resolved when there is not a `CommutativeSemigroup` for the value type.


LukaJCB:
You'll need to add an exception Mima exception, but otherwise this looks great, thank you!

LukaJCB:
Awesome, thank you!

Issue #2005, opened 2017-11-01
Change scaladoc link target to https://typelevel.org/cats/api/cats/


arosien:
Currently the API documentation (scaladoc) link points to http://typelevel.github.io/cats/api/, which resolves to the `root` package of the scaladoc, even though cats is entirely under package `cats`. It might be less confusing to directly link to https://typelevel.org/cats/api/cats/, which would avoid requiring the viewer to immediately click the `cats` link, which is the only option available.

PR #2050, opened 2017-11-27
Link Directly to Cats Package in ScalaDoc

stephen-lazaro:
This resolves https://github.com/typelevel/cats/issues/2005 providing link directly to the `cats` package in the ScalaDoc.

Rider: Fix some Scala source links in microSite sources that were killed when `Contravariant` and `Invariant` were moved out of `functor` package.


kailuowang:
Thanks very much!

PR #2052, opened 2017-11-29
Add labels to prop produced from IsEq

nigredo-tori:
For failed `IsEq` produce labels with both values to simplify debugging.

The implementation was borrowed from `org.scalacheck.Prop.?=` with minimal changes. No changes to `*Test` traits were made, so the default pretty-printer will be used in most cases, which shouldn't cause any problems. Changes to pass a pretty-printer through would be trivial, but would require a lot of small changes (and would be fragile unless we convert `IsEq` to `Prop` explicitly everywhere, since we always have implicit pretty-printer for `Any`).

### DIFF
--- a/kernel-laws/src/main/scala/cats/kernel/laws/discipline/package.scala
+++ b/kernel-laws/src/main/scala/cats/kernel/laws/discipline/package.scala
@@ -2,8 +2,15 @@ package cats.kernel.laws
 
 import cats.kernel.Eq
 import org.scalacheck.Prop
+import org.scalacheck.util.Pretty
 
 package object discipline {
-  implicit def catsLawsIsEqToProp[A](isEq: IsEq[A])(implicit ev: Eq[A]): Prop =
-    ev.eqv(isEq.lhs, isEq.rhs)
+  implicit def catsLawsIsEqToProp[A](isEq: IsEq[A])(implicit ev: Eq[A], pp: A => Pretty): Prop =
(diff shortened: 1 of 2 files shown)


kernel-laws/src/main/scala/cats/kernel/laws/discipline/package.scala
```diff
@@ -2,8 +2,15 @@ package cats.kernel.laws
...
 import cats.kernel.Eq
 import org.scalacheck.Prop
+import org.scalacheck.util.Pretty
 
 package object discipline {
-  implicit def catsLawsIsEqToProp[A](isEq: IsEq[A])(implicit ev: Eq[A]): Prop =
-    ev.eqv(isEq.lhs, isEq.rhs)
+  implicit def catsLawsIsEqToProp[A](isEq: IsEq[A])(implicit ev: Eq[A], pp: A => Pretty): Prop =
+    isEq match { case IsEq(x, y) =>
+      if(ev.eqv(x, y)) Prop.proved else Prop.falsified :| {
```
kailuowang: This fails the scalastyle check with
> no space after token if

kailuowang:
This is super cool! Thanks!

ceedubs:
I guess that this resulted in a slight code coverage drop. It would probably be nice to have a test case for respecting `Pretty` instances, but it's not really a huge deal.

PR #2053, opened 2017-11-30
Fix #2051, Remove superfluous implicit

### LINKED ISSUE #2051: EqTests#eqv requires an Eq[A]
`cats.kernel.laws.discipline.EqTests#eqv` requires an implicit `Eq[A]`, while it already has one, in the `EqLaws[A]` instance it has. Thus the implicit `Eq[A]` parameter is unnecessary and could be removed.

rsoeldner:
First contribution to cats :tada: , let me know if something is missing.


johnynek:
👍 

Thank you!

LukaJCB:
Awesome thanks! :)

PR #2056, opened 2017-12-01
Functor.fmap

fosskers:
This PR adds a `.fmap` method to `Functor` as an alias of `.map`. This is in the spirit of `Foldable.combineAll` which is an alias for `.fold`, as many types have a built-in `.fold` with different behaviour, preventing `.fold` from being injected.

#### PROs

- `.fmap` is injectable, exposing lawful Functorness to implementing classes (i.e. `Map` has a built-in `.map` that breaks Functor laws, but its injected `.fmap` would be lawful). This also benefits authors of new Functor instances.
- Haskellers get a warm, fuzzy feeling of familiarity.

#### CONs

- Increased API surface area for a method that is not strictly necessary (but then you'd have to say the same for `.combineAll`)
- Potential confusion with `.flatMap` for non-Haskellers.


fosskers:
`final` so it can't be overridden, yeah?

fosskers:
`Foldable.combineAll` is included directly in the typeclass.

SystemFw:
I don't have any strong opinions on this, and avoid having to do `Functor[Blah].map` seems useful enough, however wrt "haskellers get a warm-fuzzy feeling of familiarity", this is a case where actually (as a Haskeller), things got _fixed_ here.
Haskell used to have `fmap` be called just `map`. Then a `map` specialised to Lists was added in the early days to avoid complicating things for beginners that don't know about typeclasses. However, after the FTP proposal has brought polymorphism in the Prelude, Functor `fmap` and a  special `map` for List stick out like a sore thumb.

LukaJCB:
Hi @fosskers, can you make a mima exception, please? :)

fosskers:
@SystemFw The `fmap` / `map` situation in Haskell is weird, yes. Hopefully we can settle on something nice there too. For Cats, it's a problem of "Well `.map` can't be injected. Okay let's make an alias. What to call it? Okay, `.fmap` because of prescedent and intuition: fmap = Functor map."

SystemFw:
Sure, I agree that the underlying problem of underlying methods taking precedence over typeclass ones is there.
I guess the question is how many instances of this are out there for the `map` case? And the other question is how big of a problem it actually is. Unless people _always_ call `fmap` , you still need to know whether the `map` you're calling is a good one, or a bad one, in which case you can do `Functor[F].map`.

In any case this is making me sound _way_ more opposed to the proposal that I actually am. My only concern is that an advanced user knows that `Functor[F].map` will work, and a beginner might be confused as to why `fmap`  is similar to `flatMap`, especially given that the name `fmap` is not nice in Haskell either. Unfortunately I can think of a better name that isn't `map` or `lift`, both already taken.

fosskers:
Lol wuh github? I didn't dismiss a review.

LukaJCB:
It automatically dismissed reviews when you push, don't worry :)

PR #2057, opened 2017-12-01
Add `Ior.fromEither`

markus1189:
Adds the `Ior.fromEither` method.  I am aware that there is a `toIor` provided via `EitherOps`, but this one is harder to find for newcomers (IMHO :D) and needs the import of the `EitherSyntax`.

So I added the `Ior.fromEither` which does not require any import and is easily discoverable on the `Ior` companion.

### 24 Pull Requests (https://24pullrequests.com) ###
https://24pullrequests.com

### DIFF
--- a/core/src/main/scala/cats/data/Ior.scala
+++ b/core/src/main/scala/cats/data/Ior.scala
@@ -257,4 +257,25 @@ private[data] sealed trait IorFunctions {
         case None => None
       }
     }
+
+  /**
+   * Create an `Ior` from an `Either`.
+   * @param eab an `Either` from which the `Ior` should be created
+   *
+   * @return [[Ior.Left]] if the `Either` was a `Left`,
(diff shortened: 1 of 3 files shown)


core/src/main/scala/cats/data/Ior.scala
```diff
@@ -257,4 +257,25 @@ private[data] sealed trait IorFunctions {
         case None => None
       }
     }
+
+  /**
+   * Create an `Ior` from an `Either`.
+   * @param eab an `Either` from which the `Ior` should be created
+   *
+   * @return [[Ior.Left]] if the `Either` was a `Left`,
```
LukaJCB: You have a whitespace error here :)

markus1189: Nice that cats checks for them!

PR #2059, opened 2017-12-01
Add Parallel instance for Ior

andyscott:
Adds a `Parallel` instance for `Ior`.

The parallel instance is defined as `Parallel[Ior[E, ?], Ior[E, ?]]` without any wrapper type for a "parallel Ior". We could add one, if it's useful.

### DIFF
--- a/core/src/main/scala/cats/data/Ior.scala
+++ b/core/src/main/scala/cats/data/Ior.scala
@@ -205,6 +205,7 @@ private[data] sealed abstract class IorInstances extends IorInstances0 {
       override def bimap[A, B, C, D](fab: A Ior B)(f: A => C, g: B => D): C Ior D = fab.bimap(f, g)
     }
 
+  // scalastyle:off cyclomatic.complexity
   implicit def parallelForIor[E]
     (implicit E: Semigroup[E]): Parallel[Ior[E, ?], Ior[E, ?]] = new Parallel[Ior[E, ?], Ior[E, ?]]
   {
@@ -235,8 +236,10 @@ private[data] sealed abstract class IorInstances extends IorInstances0 {
           }
         }
     }
+
     lazy val monad: Monad[Ior[E, ?]] = Monad[Ior[E, ?]]
   }
+  // scalastyle:on cyclomatic.complexity
 
 
 }
--- a/core/src/main/scala/cats/data/Ior.scala
+++ b/core/src/main/scala/cats/data/Ior.scala
@@ -2,6 +2,7 @@ package cats
 package data
 
 import cats.Bifunctor
+import cats.arrow.FunctionK
 import cats.data.Validated.{Invalid, Valid}
 
 import scala.annotation.tailrec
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/data/Ior.scala
```diff
@@ -203,6 +204,39 @@ private[data] sealed abstract class IorInstances extends IorInstances0 {
     new Bifunctor[Ior] {
       override def bimap[A, B, C, D](fab: A Ior B)(f: A => C, g: B => D): C Ior D = fab.bimap(f, g)
     }
+
+  implicit def parallelForIor[E]
+    (implicit E: Semigroup[E]): Parallel[Ior[E, ?], Ior[E, ?]] = new Parallel[Ior[E, ?], Ior[E, ?]]
+  {
+
+    def parallel: Ior[E, ?] ~> Ior[E, ?] = FunctionK.id
```
johnynek: these could be `val` do avoid the reallocations, right?

tests/src/test/scala/cats/tests/ParallelSuite.scala
```diff
@@ -17,7 +16,7 @@ import scala.collection.immutable.SortedSet
 class ParallelSuite extends CatsSuite with ApplicativeErrorForEitherTest {
```
johnynek: can we test the parallel laws?

https://github.com/typelevel/cats/blob/master/laws/src/main/scala/cats/laws/ParallelLaws.scala

core/src/main/scala/cats/data/Ior.scala
```diff
@@ -203,6 +204,39 @@ private[data] sealed abstract class IorInstances extends IorInstances0 {
...
+            case Ior.Left(e2) => Ior.Left(E.combine(e2, e1))
+          }
+          case Ior.Left(e1) => ff match {
+            case Ior.Right(f) => Ior.Left(e1)
+            case Ior.Both(e2, f) => Ior.Left(E.combine(e2, e1))
+            case Ior.Left(e2) => Ior.Left(E.combine(e2, e1))
+          }
+        }
+    }
+    def monad: Monad[Ior[E, ?]] = Monad[Ior[E, ?]]
```
johnynek: this can be a `val` no?

core/src/main/scala/cats/data/Ior.scala
```diff
@@ -203,6 +204,39 @@ private[data] sealed abstract class IorInstances extends IorInstances0 {
...
     }
+
+  implicit def parallelForIor[E]
+    (implicit E: Semigroup[E]): Parallel[Ior[E, ?], Ior[E, ?]] = new Parallel[Ior[E, ?], Ior[E, ?]]
+  {
+
+    def parallel: Ior[E, ?] ~> Ior[E, ?] = FunctionK.id
+    def sequential: Ior[E, ?] ~> Ior[E, ?] = FunctionK.id
+
+    def applicative: Applicative[Ior[E, ?]] = new Applicative[Ior[E, ?]] {
```
johnynek: can be a `val` no?

andyscott: 👍

core/src/main/scala/cats/data/Ior.scala
andyscott: 👍

core/src/main/scala/cats/data/Ior.scala
andyscott: 👍

tests/src/test/scala/cats/tests/ParallelSuite.scala
andyscott: Will do. 👍

core/src/main/scala/cats/data/Ior.scala
```diff
@@ -203,6 +204,41 @@ private[data] sealed abstract class IorInstances extends IorInstances0 {
...
+            case Ior.Left(e2) => Ior.Left(E.combine(e2, e1))
+          }
+          case Ior.Left(e1) => ff match {
+            case Ior.Right(f) => Ior.Left(e1)
+            case Ior.Both(e2, f) => Ior.Left(E.combine(e2, e1))
+            case Ior.Left(e2) => Ior.Left(E.combine(e2, e1))
+          }
+        }
+    }
+    lazy val monad: Monad[Ior[E, ?]] = Monad[Ior[E, ?]]
```
andyscott: The monad instance isn't used for the `parTraverse/parSequence` use cases on `Ior`, so I've used a lazy val. All the other vals/defs are used.

johnynek:
👍

PR #2060, opened 2017-12-04
Add parallel instance for IorT

andyscott:
The same functionality as https://github.com/typelevel/cats/pull/2059, but for IorT instead of Ior.

### DIFF
--- a/core/src/main/scala/cats/data/IorT.scala
+++ b/core/src/main/scala/cats/data/IorT.scala
@@ -1,6 +1,7 @@
 package cats
 package data
 
+import cats.arrow.FunctionK
 import cats.syntax.either._
 import cats.syntax.option._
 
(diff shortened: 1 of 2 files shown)


kailuowang:
just realized that this one is bin compat, so we can potentially wait after 1.0 but obviously there is no reasons that we have to.

kailuowang:
@LukaJCB shall we merge this for now and add that instance in a followup (if you do merge please change the milestone to 1.0) thanks!

andyscott:
@LukaJCB do you mean something like:
```scala
implicit def catsDataParallelForIorT[M[_], F[_], E]
    (implicit P: Parallel[M, F], E: Semigroup[E]): Parallel[IorT[M, E, ?], IorT[F, E, ?]] = new Parallel[IorT[M, E, ?], IorT[F, E, ?]]
  {
    val parallel: IorT[M, E, ?] ~> IorT[F, E, ?] = λ[IorT[M, E, ?] ~> IorT[F, E, ?]](fm => IorT(P.parallel(fm.value)))
    val sequential: IorT[F, E, ?] ~> IorT[M, E, ?] = λ[IorT[F, E, ?] ~> IorT[M, E, ?]](ff => IorT(P.sequential(ff.value)))

    private[this] val FA: Applicative[F] = P.applicative
    private[this] val IorA: Applicative[Ior[E, ?]] = Parallel[Ior[E, ?], Ior[E, ?]].applicative

    val applicative: Applicative[IorT[F, E, ?]] = new Applicative[IorT[F, E, ?]] {
      def pure[A](a: A): IorT[F, E, A] = IorT.pure(a)(FA)
      def ap[A, B](ff: IorT[F, E, A => B])(fa: IorT[F, E, A]): IorT[F, E, B] =
        IorT(FA.map2(ff.value, fa.value)((f, a) => IorA.ap(f)(a)))
    }

    lazy val monad: Monad[IorT[M, E, ?]] = {
      implicit def underlyingMonadM: Monad[M] = P.monad
      Monad[IorT[M, E, ?]]
    }
  }
```

I think this would work nicely as a higher priority instance over the implementation I already provided. Lots of effect types don't have parallel instances available so the fallback instance for `Parallel[IorT[F, E, ?], IorT[F, E, ?]]` seems useful. This higher priority instances would be parallel over both the effect and the accumulation of errors while the lower priority instance would be sequential over the effect but still accumulate errors.

PR #2063, opened 2017-12-05
Added `merge` (product) to `Arrow` for arrows composition

marcobattaglia:
merge &&&
```scala
scala> import cats.arrow.Arrow
scala> val addEmpty: Int => Int = _ + 0
scala> val multiplyEmpty: Int => Double= _ * 1d
scala> val f: Int => (Int, Double) = addEmpty &&& multiplyEmpty
scala> f(1)
res0: (Int Double) = (1,1.0)
```
combineAndByPass -<

 Create a new computation `F` that apply f andThen biFork the result
     one way is applied to g and the other pass through
     the result is a tuple
```
                 /-----out1 = f(input) -----\
      input ----{                            }----(out1,out2)
                 \-----out2 = g(f(input))---/
```

Example:

```scala
     scala> import cats.arrow.Arrow
     scala> val twoTimes: Int => Double = _ * 2d
     scala> val fiveTimes: Double => Double= _ * 5
     scala> val f: Int => (Double, Double) = twoTimes -< fiveTimes
     scala> f(2)
     res0: (Double Double) = (4.0,20.0)
```

### DIFF
--- a/core/src/main/scala/cats/arrow/Arrow.scala
+++ b/core/src/main/scala/cats/arrow/Arrow.scala
@@ -52,12 +52,11 @@ import simulacrum.typeclass
     * Example:
     * {{{
     * scala> import cats.implicits._
-    * scala> import cats.arrow.Arrow
     * scala> val addEmpty: Int => Int = _ + 0
     * scala> val multiplyEmpty: Int => Double= _ * 1d
     * scala> val f: Int => (Int, Double) = addEmpty &&& multiplyEmpty
     * scala> f(1)
-    * res0: (Int Double) = (1,1.0)
+    * res0: (Int, Double) = (1,1.0)
     * }}}
     *
     * Note that the arrow laws do not guarantee the non-interference between the _effects_ of
@@ -78,12 +77,11 @@ import simulacrum.typeclass
     * Example:
     * {{{
     * scala> import cats.implicits._
-    * scala> import cats.arrow.Arrow
     * scala> val twoTimes: Int => Double = _ * 2d
     * scala> val fiveTimes: Double => Double= _ * 5
     * scala> val f: Int => (Double, Double) = twoTimes -< fiveTimes
     * scala> f(2)
-    * res0: (Double Double) = (4.0,20.0)
+    * res0: (Double, Double) = (4.0,20.0)
     * }}}
     *
     * Note that the arrow laws do not guarantee the non-interference between the _effects_ of
--- a/core/src/main/scala/cats/arrow/Arrow.scala
+++ b/core/src/main/scala/cats/arrow/Arrow.scala
@@ -6,14 +6,15 @@ import simulacrum.typeclass
 /**
  * Must obey the laws defined in cats.laws.ArrowLaws.
  */
-@typeclass trait Arrow[F[_, _]] extends Category[F] with Strong[F] { self =>
+@typeclass trait Arrow[F[_, _]] extends Category[F] with Strong[F] {
+  self =>
 
   /**
-   * Lift a function into the context of an Arrow.
-   *
-   * In the reference articles "Arrows are Promiscuous...", and in the corresponding Haskell
-   * library `Control.Arrow`, this function is called `arr`.
-   */
+    * Lift a function into the context of an Arrow.
+    *
+    * In the reference articles "Arrows are Promiscuous...", and in the corresponding Haskell
+    * library `Control.Arrow`, this function is called `arr`.
+    */
   def lift[A, B](f: A => B): F[A, B]
 
   override def dimap[A, B, C, D](fab: F[A, B])(f: C => A)(g: B => D): F[C, D] =
@@ -21,27 +22,28 @@ import simulacrum.typeclass
 
   override def second[A, B, C](fa: F[A, B]): F[(C, A), (C, B)] = {
     def swap[X, Y]: F[(X, Y), (Y, X)] = lift[(X, Y), (Y, X)] { case (x, y) => (y, x) }
+
     compose(swap, compose(first[A, B, C](fa), swap))
   }
 
   /**
-   * Create a new computation `F` that splits its input between `f` and `g`
-   * and combines the output of each.
-   *
-   * Example:
-   * {{{
-   * scala> import cats.implicits._
-   * scala> import cats.arrow.Arrow
-   * scala> val toLong: Int => Long = _.toLong
-   * scala> val toDouble: Float => Double = _.toDouble
-   * scala> val f: ((Int, Float)) => (Long, Double) = Arrow[Function1].split(toLong, toDouble)
-   * scala> f((3, 4.0f))
-   * res0: (Long, Double) = (3,4.0)
-   * }}}
-   *
-   * Note that the arrow laws do not guarantee the non-interference between the _effects_ of
-   *  `f` and `g` in the context of F. This means that `f *** g` may not be equivalent to `g *** f`.
-   */
+    * Create a new computation `F` that splits its input between `f` and `g`
+    * and combines the output of each.
+    *
+    * Example:
+    * {{{
+    * scala> import cats.implicits._
+    * scala> import cats.arrow.Arrow
+    * scala> val toLong: Int => Long = _.toLong
+    * scala> val toDouble: Float => Double = _.toDouble
+    * scala> val f: ((Int, Float)) => (Long, Double) = Arrow[Function1].split(toLong, toDouble)
+    * scala> f((3, 4.0f))
+    * res0: (Long, Double) = (3,4.0)
+    * }}}
+    *
+    * Note that the arrow laws do not guarantee the non-interference between the _effects_ of
+    * `f` and `g` in the context of F. This means that `f *** g` may not be equivalent to `g *** f`.
+    */
   @simulacrum.op("***", alias = true)
   def split[A, B, C, D](f: F[A, B], g: F[C, D]): F[(A, C), (B, D)] =
     andThen(first(f), second(g))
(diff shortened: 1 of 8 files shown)


kailuowang:
will need to add some mima exceptions
```scala
exclude[ReversedMissingMethodProblem]("cats.arrow.Arrow.merge"),
exclude[ReversedMissingMethodProblem]("cats.arrow.Arrow.combineAndByPass"),
exclude[ReversedMissingMethodProblem]("cats.arrow.Arrow#Ops.merge"),
exclude[ReversedMissingMethodProblem]("cats.arrow.Arrow#Ops.combineAndByPass"),
exclude[ReversedMissingMethodProblem]("cats.arrow.Arrow#Ops.&&&"),
exclude[ReversedMissingMethodProblem]("cats.arrow.Arrow#Ops.-<"),
```
to [here](https://github.com/typelevel/cats/blob/master/build.sbt#L198)

kailuowang:
I took the liberty to update your comment with some formatting. Hope you don't mind.

marcobattaglia:
HI Kailuowang. I have committed the exclude directives.  Do I need to do another pull request or is it ok for you?
Thanks for your better formatting.

kailuowang:
You are welcome! Everything related in one PR would be better.

kailuowang:
some errors in the doctest
> m/home/travis/build/typelevel/cats/core/.jvm/target/scala-2.11/src_managed/test/cats/arrow/ArrowDoctest.scala:46: identifier expected but ')' found.

/home/travis/build/typelevel/cats/core/.jvm/target/scala-2.11/src_managed/test/cats/arrow/ArrowDoctest.scala:50: ')' expected but '}' found.
/home/travis/build/typelevel/cats/core/.jvm/target/scala-2.11/src_managed/test/cats/arrow/ArrowDoctest.scala:65: identifier expected but ')' found.
/home/travis/build/typelevel/cats/core/.jvm/target/scala-2.11/src_managed/test/cats/arrow/ArrowDoctest.scala:69: ')' expected but '}' found.

To run doctest, `sbt coreJVM/test`

core/src/main/scala/cats/arrow/Arrow.scala
```diff
@@ -45,4 +45,50 @@ import simulacrum.typeclass
...
+    *  `f` and `g` in the context of F. This means that `f &&& g` may not be equivalent to `g &&& f`.
+    */
+  @simulacrum.op("&&&", alias = true)
+  def merge[A, B, C](f: F[A, B], g: F[A, C]): F[A, (B, C)] = {
+    andThen(lift((x: A) => (x, x)), split(f, g))
+  }
+
+  /**
+    * Create a new computation `F` that apply f andThen biFork the result
+    * one way is applied to g and the other pass through
```
kailuowang: This description kind of imply that `f` is only applied once, but actually it is applied twice right? the diagram is more accurate.

kailuowang:
This `merge` is basically `Semigroupal.product`, if you just care about `Function` or `Kleisli` they already have `Semigroupal` instance
What this proves, though, is that `Arrow` forms a `Semigroupal`
We could add the following to `Semigroupal`'s companion
```scala
  implicit def catsSemigroupalForArrow[F[_, _], A](implicit F: Arrow[F]): Semigroupal[F[A, ?]] = new Semigroupal[F[A, ?]] {
    def product[B, C](fb: F[A, B], fc: F[A, C]): F[A, (B, C)] =
      F.andThen(F.lift((x: A) => (x, x)), F.split(fb, fc))
  }
```

kailuowang:
I think this PR is good as is.  I thought about renaming `merge` to `product` but I think it might cause syntax  conflicts between `ArrowSyntax` and `SemigroupalSyntax`. 
We can add the `Semigroupal` instance for `F[_, ?]` with `Arrow` in a followup PR after this one is merged.

core/src/main/scala/cats/arrow/Arrow.scala
```diff
@@ -45,4 +45,50 @@ import simulacrum.typeclass
...
+    * scala> val twoTimes: Int => Double = _ * 2d
+    * scala> val fiveTimes: Double => Double= _ * 5
+    * scala> val f: Int => (Double, Double) = twoTimes -< fiveTimes
+    * scala> f(2)
+    * res0: (Double, Double) = (4.0,20.0)
+    * }}}
+    *
+    */
+  @simulacrum.op("-<", alias = true)
+  def combineAndByPass[A, B, C](f: F[A, B], g: F[B, C]): F[A, (B, C)] = {
```
ceedubs: Could you explain the meaning behind this name? Should it be `Bi` instead of `By`? It makes me think of the word [bypass](http://wordnik.com/words/bypass), but I suspect you weren't going for that based on the camelCasing.

ceedubs:
Thanks @marcobattaglia!

I definitely like having the `merge`, even if we decide to separately define a `Semigroupal` for all arrows. The use-cases for the `-<` method seem a little more narrow to me, but my only concerns about it are the textual name (see comment above), and potential bloating of type class methods. If people find it to be useful, I have no objections.

core/src/main/scala/cats/arrow/Arrow.scala
marcobattaglia: Sorry it should  be 'bypass' without camelCase. I'm converting the camel to a dromedary ;-) and the I'll commit.

marcobattaglia:
@ceedubs 
We have often used these methods in our projects, and it happens to compose functions using `-<` . 
We have also same methods to compose 'Kleisli arrows' but I need to study better cats in order to decide if it makes sense to prepare a PR for those . Thank you for your support  and consideration.

kailuowang:
Actually, before I sign off this PR, I start to wonder what would be the difference between 
calling `combineAndBypass(f, g)`  and calling `merge(f, f >>> g)`?

core/src/main/scala/cats/arrow/Arrow.scala
```diff
@@ -45,4 +45,50 @@ import simulacrum.typeclass
...
+    * scala> val fiveTimes: Double => Double= _ * 5
+    * scala> val f: Int => (Double, Double) = twoTimes -< fiveTimes
+    * scala> f(2)
+    * res0: (Double, Double) = (4.0,20.0)
+    * }}}
+    *
+    */
+  @simulacrum.op("-<", alias = true)
+  def combineAndBypass[A, B, C](f: F[A, B], g: F[B, C]): F[A, (B, C)] = {
+    andThen(lift((x: A) => (x, x)), split(f, andThen(f, g)))
```
ceedubs: One downside of this is that it calculates `f(x)` twice. I think that for pretty much any arrow you would want to override this to reuse the result of `f(x)`.

kailuowang: I agree. I think I incline to not adding this one, since it's just an alias to `merge(f, f >>> g)`

marcobattaglia: Yes you are right. It is also the same of this one:
 `andThen(andThen( f, lift((x: B) => (x, x)) ), split(id, g))`
This last one  evicts the double calculation.
 Essentially: input=> f(input) => out1=> (out1,out1)=> (out1,g(out1))=>(out1,out2)
Please consider if it's better to eliminate combineAndBypass because it is possible to use `merge(f, f >>> g)` or if it is good to have combineAndBypass as a sort of shortcut.
I'll prepare the commit following your decision.

marcobattaglia: mmm.... I think that also `merge(f, f >>> g) ` evaluates f twice, instead this one `andThen(andThen( f, lift((x: B) => (x, x)) ), split(id, g))` evaluates f one time, so that it is not properly a simple shortcut althought the result is the same.

core/src/main/scala/cats/arrow/Arrow.scala
```diff
@@ -45,4 +45,50 @@ import simulacrum.typeclass
...
+    * scala> val twoTimes: Int => Double = _ * 2d
+    * scala> val fiveTimes: Double => Double= _ * 5
+    * scala> val f: Int => (Double, Double) = twoTimes -< fiveTimes
+    * scala> f(2)
+    * res0: (Double, Double) = (4.0,20.0)
+    * }}}
+    *
+    */
+  @simulacrum.op("-<", alias = true)
+  def combineAndBypass[A, B, C](f: F[A, B], g: F[B, C]): F[A, (B, C)] = {
```
kailuowang: This can be `andThen(f, merge (id, g))`  right?

marcobattaglia: Right. Do you prefer to eliminate the combineAndBypass or to leave it as a simple shortcut? For me it's the same, I think that it depends on how cats is generally organized. I will commit following your opinion.

marcobattaglia: I've tries to rewrite `andThe(f, merge(d,g))` as something like `andThen(f, catsSemigroupalForArrow.product(id,g))` and I've noticed that `catscatsSemigroupalForArrow.product` is a little bit different than `merge` infact the product change also the types of the resulting function. 
For example the type of 
`catsSemigroupalForArrow.product( ((x:Int) => x * 2d), ((x:Double) => x * 5) )` is Double with Int => (Double, Double) , instead the type of `((x:Int) => x * 2d) merge ((x:Int) => x * 5)' is Int => (Double, Int). My question is: 'product' means also an operation on types, in the sense that it also work on the domain types of the function ?

kailuowang: Your `catsSemigroupalForArrow.product( ((x:Int) => x * 2d), ((x:Double) => x * 5) )` is supprising. I didn't expect that to compile. We can discuss that #2078 

If you don't mind I'd prefer removing `combineAndBypass`. It's not providing much, typing `f >>> (id &&& g) ` isn't too much worse than `f -< g`  which arguably is more obscure to read. And it's a lot harder to remove a method than adding one.

marcobattaglia: I agree. I will prepare the commit to delete it. Thx

marcobattaglia: this capanion object: ` object Arrow {
    implicit def catsSemigroupalForArrow[F[_, _], A](implicit F: Arrow[F]): Semigroupal[F[A, ?]] = new Apply[F[A, ?]] {
      def map[B, C](fb: F[A, B])(f: B => C): F[A, C] = F.rmap(fb)(f)
      def ap[B, C](ff: F[A, B => C])(fb: F[A, B]): F[A, C] =
        F.rmap(F.andThen(F.lift((x: A) => (x, x)), F.split(ff, fb)))(tup => tup._1(tup._2))
    }` compiles.

core/src/main/scala/cats/arrow/Arrow.scala
```diff
@@ -6,42 +6,44 @@ import simulacrum.typeclass
...
-   * Lift a function into the context of an Arrow.
-   *
-   * In the reference articles "Arrows are Promiscuous...", and in the corresponding Haskell
-   * library `Control.Arrow`, this function is called `arr`.
-   */
+    * Lift a function into the context of an Arrow.
+    *
+    * In the reference articles "Arrows are Promiscuous...", and in the corresponding Haskell
+    * library `Control.Arrow`, this function is called `arr`.
+    */
```
kailuowang: I hate to nitpick on this, but this format is unnecessarily changed away from the scala doc convention.

marcobattaglia: Excuse me kailuowang, but I don't understand. What format is changed away? I don't know where I must correct the format, because I don't notice any difference.Please show me better.thx

kailuowang: The indentations of the * :)

kailuowang: again, it's a total nitpick. I think it's ready to merge.

marcobattaglia:
Ops! Space! Sorry, I've committed  just now.

PR #2064, opened 2017-12-05
added `set -e` to travis build

kailuowang:
travis build should exit on error.


kailuowang:
not sure why this fails , I'm closing this one. retry in #2065

PR #2074, opened 2017-12-06
Example for `Validated.ap` and swap combine order

markus1189:
Okay this was strange, why was the order of arguments in the `combine` swapped?

### 24 Pull Requests (https://24pullrequests.com) ###
https://24pullrequests.com

### DIFF
--- a/core/src/main/scala/cats/data/Validated.scala
+++ b/core/src/main/scala/cats/data/Validated.scala
@@ -166,11 +166,24 @@ sealed abstract class Validated[+E, +A] extends Product with Serializable {
   /**
    * From Apply:
    * if both the function and this value are Valid, apply the function
+   *
+   * Example:
+   * {{{
+   * scala> import cats.instances.list._ // semigroup for list
+   * scala> Validated.valid[List[String], Int](1).ap(Validated.valid((x: Int) => x + 1))
+   * res0: Validated[List[String], Int] = Valid(2)
+   * scala> Validated.invalid[List[String], Int](List("Input invalid")).ap(Validated.valid((x: Int) => x + 1))
+   * res1: Validated[List[String], Int] = Invalid(List(Input invalid))
+   * scala> Validated.valid[List[String], Int](1).ap(Validated.invalid[List[String], Int => Int](List("Function invalid")))
+   * res2: Validated[List[String], Int] = Invalid(List(Function invalid))
+   * scala> Validated.invalid[List[String], Int](List("Input invalid")).ap(Validated.invalid[List[String], Int => Int](List("Function invalid")))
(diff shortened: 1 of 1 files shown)


ceedubs:
@markus1189 ah thanks for pointing this out. This is confusing. Strangely enough, this `ap` implementation is "right", but what's strange about it is that it's kind of in the reverse order as the `ap` method defined in `Apply`. See #803 and the chain of related issues/PRs. I would imagine that we have some similar headscratchers for other data types.

One solution to this might be to change the signature to require that the `Validated` instance that you are calling this on is of the form `Validated[E, A => B]`. A possibly more straightforward solution would be to just remove this method from `Validated` and let people access it via type class syntax if they want it. I can't think of a single time that I've ever seen anyone use `ap` directly in application code. I think that it's far more common to use `map2`, `product`, etc.

ceedubs:
Related failures in the build:

```
[info] *** 4 TESTS FAILED ***

[info] ParallelSuite:

[info] 

[info] - ApplicativeError[Validated[String, Int]].applicativeError.ap consistent with product + map *** FAILED ***

[info]     Falsified after 6 successful property evaluations.

[info]       arg0 = Invalid(),

[info]       arg1 = Invalid(᧞줧)

[info]     )

[info]       Expected: Invalid(᧞줧)

[info]   Received: Invalid(᧞줧)

[info] ParallelSuite:

[info] 

[info] - parAp accumulates errors in order *** FAILED ***

[info]   Left("WorldHello") did not equal Left("HelloWorld") (ParallelSuite.scala:98)

[info] NestedSuite:

[info] 

[info] - Nested[Validated[String, ?], ListWrapper, ?].applicativeError.ap consistent with product + map *** FAILED ***

[info]     Falsified after 11 successful property evaluations.

[info]       arg0 = Nested(Invalid(죉玔䒛)),

[info]       arg1 = Nested(Invalid(夒ㄳㇶ))

[info]     )

[info]       Expected: Nested(Invalid(夒ㄳㇶ죉玔䒛))

[info]   Received: Nested(Invalid(죉玔䒛夒ㄳㇶ))

[info] ValidatedSuite:

[info] 

[info] - Validated[String, Int].applicativeError.ap consistent with product + map *** FAILED ***

[info]     Falsified after 27 successful property evaluations.

[info]       arg0 = Invalid(㱈ѥ躨㟫볓),

[info]       arg1 = Invalid(뼊᪯)

[info]     )

[info]       Expected: Invalid(뼊᪯㱈ѥ躨㟫볓)

[info]   Received: Invalid(㱈ѥ躨㟫볓뼊᪯)
```

markus1189:
@ceedubs Yes this was confusing :D  But thinking of it in Haskell terms I can see how the confusion arises.

I would vote for removing the `ap` from `Validated`, it seems like there is no added convenience.

PR #2076, opened 2017-12-06
Add  `Endo` type alias for (`A =>A`)

kailuowang:
`Endo` is necessary when I found that `List[A => A]().foldK` doesn't work, it has to use an `Endo` type alias
the main use case is 
```scala
scala> import cats.implicits._, cats.Endo
scala> val l: List[Endo[Int]] = List(_ * 2, _ + 3)
scala> val f = l.reduceRightK //using `compose` to chain the functions
scala> f(1)
res0: Int = 5
```

~Update: to further clarify, using `reduceRightK`, the functions inside the list is applied from left to right. If we use the existing `foldK` the functions inside the list would be applied from right to left which gives a final result of 8.~ 

further Update: `reduceRightK` is removed. This PR is only about adding `Endo`

### DIFF
--- a/core/src/main/scala/cats/Foldable.scala
+++ b/core/src/main/scala/cats/Foldable.scala
@@ -245,6 +245,24 @@ import Foldable.sentinel
     }
 
   /**
+   * Like `foldK` but from the right side.
+   * One use case would be endo functions (`A => A`) chaining
+   * because the default MonoidK instance for endo functions
+   * is using `compose` rather than the `andThen`
+   * {{{
+   * scala> import cats.implicits._, cats.Endo
+   * scala> val l: List[Endo[Int]] = List(_ * 2, _ + 3)
+   * scala> val f = l.reduceRightK //using `compose` to chain the functions
+   * scala> f(1)
+   * res0: Int = 5
+   * }}}
+   */
+  def reduceRightK[G[_], A](fga: F[G[A]])(implicit G: MonoidK[G]): G[A] =
+    foldRight(fga, Eval.now(G.empty[A])) { (ga, ega) =>
+      ega.map(rga => G.combineK(rga, ga))
+    }.value
+
+  /**
    * Alias for [[fold]].
    */
   def combineAll[A: Monoid](fa: F[A]): A = fold(fa)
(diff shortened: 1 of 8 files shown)


core/src/main/scala/cats/Foldable.scala
```diff
@@ -244,6 +244,24 @@ import Foldable.sentinel
...
+   * is using `compose` rather than the `andThen`
+   * {{{
+   * scala> import cats.implicits._, cats.Endo
+   * scala> val l: List[Endo[Int]] = List(_ * 2, _ + 3)
+   * scala> val f = l.reduceRightK //using `compose` to chain the functions
+   * scala> f(1)
+   * res0: Int = 5
+   * }}}
+   */
+  def reduceRightK[G[_], A](fga: F[G[A]])(implicit G: MonoidK[G]): G[A] =
```
johnynek: seems like `combineAllK` to me.

kailuowang: `combineAllK` is `foldK` right?

johnynek: I see. I misunderstood.

I am very -1 on this unless I am mistaken. reduceRight(Option) is right associative, but does not reverse the effective order that we iterate through. This method does by swapping.

I would call this `reverseCombineAllK`.

Since the `MonoidK` is associative, the result is the same if you right associate or left associate (though there may be efficiency reasons to pick one). This is something else: it is like making a new Monoid that has the property that `combine(a, b) == original.combine(b, a)`

This is actually still a monoid, but it behaves differently for non-commutative cases.

@kailuowang can you explain if I am wrong about this being different from reduceRight?

kailuowang: You are right it's different from reduceRight(Option). `reverseCombineAllK` is probably a better name, but you also got to the bottom of it, we don't need it, we just need a different `Monoid` instance that does the `original.combine(b, a)`.  Maybe adding a `reverse` method to `Monoid` and `MonoidK`.  But I am not sure are they worthwhile to be added? 

What about the `Endo` type alias?

kailuowang:
merge with 2 sign-off.

Issue #2069, opened 2017-12-06
Get `Semigroupal` from `Arrow` for free


kailuowang:
We could add the following to `Semigroupal`'s companion
```scala
  implicit def catsSemigroupalForArrow[F[_, _], A](implicit F: Arrow[F]): Semigroupal[F[A, ?]] = new Semigroupal[F[A, ?]] {
    def product[B, C](fb: F[A, B], fc: F[A, C]): F[A, (B, C)] =
      F.andThen(F.lift((x: A) => (x, x)), F.split(fb, fc))
  }
```
#2063 is adding a `merge` method, so this instance can just delegate to that. 
To test this we might need to create a new `case class MyFunc[A, B](f: A => B)` and an `Arrow` instance for it and then test the Semigroupal laws against it. 

I set the milestone to be 1.1, but it can certainly be done by 1.0

LukaJCB:
Would this be a binary compatible change?

kailuowang:
I think so.

ceedubs:
A relevant question would be whether there are conceivable `F` types that form an `Arrow` but that we _wouldn't_ want to derive this implicit `Semigroupal` for. I can't think of anything that might fall into that realm, but it's probably worth thinking about.

kailuowang:
@ceedubs are you suggesting making it explicit?

ceedubs:
@kailuowang that was one thought that I had. But I don't know if there's really any reason to that. Just my usual paranoia :). I'll ponder this a bit and give others a chance to weigh in over the next day or two, and if nobody can think of a reason that this implicit def shouldn't exist, then maybe it's either good or mostly harmless.

stephen-lazaro:
I can open a PR for this. I think you can reasonable strengthen it to `Apply` as well, after playing with it a bit. Given that that's a fair amount stronger, it might lend more weight to explicitness over implicitness here.

LukaJCB:
Fixed by #2078

PR #2078, opened 2017-12-07
Add Applicative for Arrow

stephen-lazaro:
Adds an `Apply` instance as discussed in https://github.com/typelevel/cats/issues/2069
It can be weakened to `Semigroupal` if we don't buy the `Apply`.
Currently, it is implicit. There was discussion in the issue on whether it should be explicit or implicit. Either answer seems reasonable so I wrote the doctest to be compatible with either way people end up preferring.
Do we need additional tests over the doctest?

### DIFF
--- a/core/src/main/scala/cats/Applicative.scala
+++ b/core/src/main/scala/cats/Applicative.scala
@@ -70,7 +70,7 @@ object Applicative {
     new ApplicativeMonoid[F, A](f, monoid)
 
   /**
-   * Creates a semigroupal functor for `F`, holding domain fixed and combining
+   * Creates an applicative functor for `F`, holding domain fixed and combining
    * over the codomain.
    *
    * Example:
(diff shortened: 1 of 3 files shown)


kailuowang:
This is cool. I start to think `Applicative` is also possible right? 

```scala
def pure[B](a: B): F[A, B] = lift(_ => a) 
```

core/src/main/scala/cats/arrow/Arrow.scala
```diff
@@ -48,3 +48,25 @@ import simulacrum.typeclass
...
+   * scala> import cats.implicits._
+   * scala> import cats.arrow.Arrow.catsSemigroupalForArrow
+   * scala> val toLong: Int => Long = _.toLong
+   * scala> val double: Int => Int = 2*_
+   * scala> val f: Int => (Long, Int) = catsSemigroupalForArrow.product(toLong, double)
+   * scala> f(3)
+   * res0: (Long, Int) = (3,6)
+   * }}}
+   */
+  implicit def catsSemigroupalForArrow[F[_, _], A](implicit F: Arrow[F]): Semigroupal[F[A, ?]] = new Apply[F[A, ?]] {
```
kailuowang: it's probably worthwhile to override `product`.

stephen-lazaro:
You're totally right, that seems to work. We'll have to check the laws but just from intuition it seems right. I have heard `Arrow is Category + Applicative` so maybe we're now figuring out the `Applicative` part ;)
Will do on the moving and adding the tests. :+1:

core/src/main/scala/cats/arrow/Arrow.scala
stephen-lazaro: Good catch, will do.

julienrf:
In case you don’t know this paper, I think it’s a good read about the relationship between arrows and applicatives: http://homepages.inf.ed.ac.uk/slindley/papers/idioms-arrows-monads.pdf

stephen-lazaro:
Cool. Strengthened the instance to `Applicative` and added it to the `Applicative` companion object. Also added a test for a function type (let me know if I misunderstood the suggestion there and you'd prefer a stronger test).

core/src/main/scala/cats/Applicative.scala
```diff
@@ -67,8 +68,84 @@ import simulacrum.typeclass
...
+   * Pure for Arrows lifts the constant function at that value.
+   * Example:
+   * {{{
+   * scala> import cats.implicits._
+   * scala> import cats.Applicative.catsApplicativeForArrow
+   * scala> val g: Int => String = catsApplicativeForArrow.pure("hello")
+   * scala> g(5)
+   * res0: String = hello
+   * }}}
+   */
```
kailuowang: Thanks for writing these doctests. But doctests in an instance are probably not necessary. They won't be included in apidoc, and these methods are already thoroughly tested in law tests of the instance. So, although the more tests the better, having them here has two slight cons, a) more code to read, in an instance definition, people want to quickly see how the instance is the implemented, rather than the API usage. and b) they provide redundant test coverage for these implementations, so if somehow the instance is not test by law tests, the test coverage report won't expose that.

stephen-lazaro: Ah, yeah, that makes sense.

Issue #1113, opened 2016-06-11
Put arity functions into separate scaladoc groups


ceedubs:
Currently certain scaladocs (such as [Apply](http://typelevel.org/cats/api/#cats.Apply)) contain a whole bunch of generated methods of various arities. This makes it really tough to find the "core" methods that people are probably interested in. I think that we should move the arity methods (at least for 3 or 4 and above) into a separate scaladoc `@group`s.

One downside is that methods in groups [cannot be linked to](https://issues.scala-lang.org/browse/SI-8309). However I think not being able to link to these methods is not going to be a problem, and this probably isn't a hard scala bug to fix, so one of us should just fix it :)

cc @mpilquist who may be the only scala dev I know who is good about using scaladoc groups

ceedubs:
I have this working on a local branch that I need to get around to turning into a PR.

ceedubs:
Resolved by #2080.

PR #2080, opened 2017-12-07
Add grouping to scaladoc for arity methods

ceedubs:
This makes it so when you look at the `Applicative` scaladoc you don't
have a bunch of `ap3`, `ap4`, etc noise mixed in with the more
meaningful methods. Instead, the generated higher-arity methods show up
in separate scaladoc groups below.

To see the changes you can pull this branch, run `sbt unidoc` (you may
need to clean first), and open
`cats/docs/target/scala-2.12/unidoc/cats/Applicative.html` in your
browser.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -178,6 +178,7 @@ lazy val docSettings = Seq(
   fork in (ScalaUnidoc, unidoc) := true,
   scalacOptions in (ScalaUnidoc, unidoc) ++= Seq(
     "-Xfatal-warnings",
+    "-groups",
     "-doc-source-url", scmInfo.value.get.browseUrl + "/tree/master€{FILE_PATH}.scala",
     "-sourcepath", baseDirectory.in(LocalRootProject).value.getAbsolutePath,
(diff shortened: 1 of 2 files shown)


ceedubs:
Oops I spotted a typo. Change incoming.

ceedubs:
Resolves #1113

ceedubs:
It seems that only @kailuowang shares my enthusiasm for this change :P

Asking people to pull my branch and generate the scaladoc to check out this change probably had a fairly significant barrier. For anyone who is interested, here are before/after images for part of the scaladocs.

Before this PR:
screen shot 2017-12-18 at 12 03 40 pm

After this PR:
screen shot 2017-12-18 at 12 09 14 pm

Higher-arity methods are grouped below the main methods:
screen shot 2017-12-18 at 12 13 13 pm

ceedubs:
@travisbrown or @mpilquist do either of you want to take a look? Circe and Scodec are two of very few projects that I can think of that use Scaladoc groups.

PR #2082, opened 2017-12-08
Add Parallel instance for free monad/applicative

ceedubs:
This implementation feels a little counter to the usual FP paradigm, as
to convert between`Free[M, ?]` and `FreeApplicative[F, ?]`, you "run"
(`runTailRec`/`fold`) to get an `M`/`A` and then lift it into the
appropriate type. Usually you just transform the free structure until
you are ready to run it at the end of the day. However, here I think
that you need to do this so that the `F` instances are combined with the
`Applicative[F]` `ap` instead of being translated to `Free[M, ?]`
structures that would then be combined with `M`'s sequential `ap`. My
first attempt at this was less eager with calling `fold`/`runTailRec`
and violated the parallel laws.

I _think_ that this implementation is legitimate, but someone else may
know better.

Even if this is legitimate, I don't know if it's all that _useful_. It
works for working with `Free[Either[A, ?], ?]` and `Free[Validation[A,
?], ?]`. But often people use `Free` for something like `Free[MyDsl,
?]`, and they probably won't have a `Parallel` instance for `MyDsl`. I
thought that I would throw this out into the world to see whether or not
anyone else thought that it was useful.

### DIFF
--- a/free/src/main/scala/cats/free/Free.scala
+++ b/free/src/main/scala/cats/free/Free.scala
@@ -334,4 +334,21 @@ sealed private[free] abstract class FreeInstances {
     new FreeTraverse[F] {
       val TraversableF = traversableF
     }
+
+  implicit def catsFreeParallelForFreeFreeApplicative[F[_], G[_]](
+    implicit FG: Parallel[F, G]
+  ): Parallel[Free[F, ?], FreeApplicative[G, ?]] =
+    new Parallel[Free[F, ?], FreeApplicative[G, ?]] {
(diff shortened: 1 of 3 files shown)


kailuowang:
@LukaJCB any thoughts on this one?

LukaJCB:
Tbh, as Cody mentions, I don't think this is super useful, so I'd hesitate to adding this.

ceedubs:
It looks like there isn't much of a use-case for this, so I'm going to go ahead and close it out. I won't delete the branch in case anyone comes looking for this in the future :)

PR #2083, opened 2017-12-08
Change forEffect/followedBy to productL/productR

### LINKED ISSUE #1983: Rename `followedBy` and `effectFor`
The `Apply` typeclass has aliases for `<*` and `*>` as `effectFor` and `followedBy` respectively. This doesn't really make sense, whenever you think about their application and I think the names detract from the actual definitions. 

Consider this example:
```scala
> val x = List(1,2,3) followedBy List(4,5,6) 
x: List[Int] = List(4, 5, 6, 4, 5, 6, 4, 5, 6)
```
The name `followedBy` would seem to imply concatenation. 

`effectFor` is a little better, since it actually expresses the application of effects. However the analogue method being `followedBy` makes things more confusing.

My personal preference is for there to be no named aliases here, but I understand some people don't feel that way.  I can't think of any better names though.  Thoughts?

Jacoby6000:
This should resolve #1983. 

I picked apL and apR based on discussion in that issue.  

The deprecation caused scalac to complain about the code generated by simulacrum, so I moved the deprecated method aliases in to syntax.apply.


Jacoby6000:
Just realized I missed a bunch of other methods.  working on that.

kailuowang:
Thanks! @Jacoby6000, a scalafix migration would be handy. Let me know if you are interested in giving it a try or I can add it in a follow-up PR.  [Here](https://github.com/typelevel/cats/pull/2033/commits/be784fa8d8fd2a36979b3d2507d57e58ab86d7b6) is a good example for such rename migrations.

kailuowang:
also, we won't have RC2, so all the deprecation note should be `1.0.0`

Jacoby6000:
@kailuowang I built the scalafix stuff, but it seems to cause some problems.  forEffect and followedBy were introduced during 1.0.0, and I'm not sure how to split up the fix migrations.

kailuowang:
The problem is that the Scalafix test depends on 0.9 right? I would say for this PR we simply don't test this migration. It should be low risk

ceedubs:
👍 once there is resolution on the scalafix rules.

I don't think that any word-based name is going to be perfect, but `apL` and `apR` seem less likely to cause confusion with certain effect types.

Jacoby6000:
I've added in the scalafix stuff that I drafted.  I can't test it, so please look it over 😅

kailuowang:
for the scalafix stuff, if you include the input and output they will be compiled, which fails due to mis-dependency as we discussed earlier. Sorry I wasn't clear, by "don't test" I mean we shouldn't include them in the input and output of the scalafix tests.

Qi77Qi:
Think [faq](https://github.com/typelevel/cats/blob/master/docs/src/main/tut/faq.md#what-does-this-symbol-mean) page needs to be updated as well?

kailuowang:
close and reopen to trigger build

core/src/main/scala/cats/syntax/apply.scala
```diff
@@ -9,4 +9,19 @@ trait ApplySyntax extends TupleSemigroupalSyntax {
       val self = fa
       val typeClassInstance = F
     }
+
+  implicit final def catsSyntaxApplyOps[F[_], A](fa: F[A]): ApplyOps[F, A] =
+    new ApplyOps(fa)
+}
+
+final class ApplyOps[F[_], A](val fa: F[A]) extends AnyVal {
```
ceedubs: Hmm, why is it that this new syntax is needed? Does simulacrum not generate syntax for deprecated methods?

kailuowang: Generate as deprecated?

johnynek:
I’m -1 on apL and apR the more I think about it. Note we already have ap. And these new methods are not that similar. I would say productR and productL are much better.

kailuowang:
I am fine with `productR` and `productL`, but if people can't agree on it soon, I would propose we give up on bikeshedding this one and just leave as is.

emilypi:
@johnynek @kailuowang these are not products, or creating products, or dealing with products in any way, and naming their methods as such would be misleading and counterintuitive. If you think they are products, or act like products, then you have a flaw in the way you're approaching the problem. 

The fact of the matter is, even my suggestions of `fill(L|R)` would be a poor way of describing the nature of these combinators. `ap` is really the only way to intuitively describe them, and if that means users need to fill their brains with 2 letters and associate it with the given behavior, so be it.

Jacoby6000:
I'm not much of a fan of `productL` and `productR`, mainly because I don't really like `product`, either.   To me, `product` intuitively means tuple.. So, that method makes me think `a -> b -> (a, b)`, not `f a -> f b -> f (a, b)`. 

As for the methods not being related,  I think their relationship becomes more clear if you define `map2` more traditionally, and then `ap2` in terms of that.  If `product` is to stay, then it should probably be rewritten in terms of `map2`, not the other way around as it is now.

```
def apL[F[_]: Apply, A, B](fa: F[A])(fb: F[B]): F[A] =
  map2(fa, fb)((a, _) => a)

def apR[F[_]: Apply, A, B](fa: F[A])(fb: F[B]): F[B] =
  map2(fa, fb)((_, b) => b)

def map2[F[_], A, B, C](fa: F[A], fb: F[B])(f: (A, B) => C)(implicit F: Apply[F]): F[C] = 
  F.ap(F.map(fa)(a => (b: B) => f(a, b)))(fb)
```

core/src/main/scala/cats/syntax/apply.scala
Jacoby6000: Simulacrum does generate the deprecated methods, which makes scalac complain since it references the original deprecated method.

Jacoby6000:
I'm willing to submit and change it to productL and productR to appease the community, I only ever use the proper notation anyway.

Jacoby6000:
Going deeper in to the relationship, and why I think apL and apR are better than productL or productR:
* `<*>` is ap.  
* `*>` is `apR` which applies both effects, as `ap` does, but only retains the value on the right.  
* `<*` is `apL` which applies both effects, as `ap` does, but only retains the value on the left.

I think having names which are as analagous as the symbols does a great deal towards teaching the notation.   If people never use `<*>` or `*>` or `<*`, but instead opt for `ap`, `apR`, and `apL`, they will be better off whenever they move to a system using more traditional notation (like, if they learn haskell for instance). Once they realize that `<*>` is ap, the understanding of `*>` and `<*` will come more easily, and vice versa.

johnynek:
I agree with some of what you say, I feel like you are overlooking that `ap` is dealing with a `A => B` at some point: `F[A => B] <*> F[A]: F[B]`. The name is related to `apply` which is about `(A => B) apply A: B`. So, keeping `ap` about `apply` from `A => B` when there is no function in sight, to me is a bad name.

I hear you that you don't like `product`, but that is the name we have for `F[A], F[B] => F[(A, B)]`. To me, the types there are much closer to what we have at hand.

If you don't like them, I'm pretty happy to keep the current names since I think as long as names don't mislead (and I feel apR and apL do), I don't think they matter too much, so forEffect is just a pointer to the docs which often has a good mnemonic. To me, that's fine.

Jacoby6000:
I'm going to update this PR to use `productL` and `productR`, specifically because cats already has product (which I have said before). Ultimately my agreement is of little consequence.  Besides that, just about anything is better than what is there currently.

kailuowang:
@Jacoby6000 thanks very much! Hate to ask you one more thing, we ended up deciding that we are going to have a quick 1.0.0-RC2 release, so if it's easy for you, would you update the deprecate versions? I'd be happy to change those for you if you don't have the time.

Jacoby6000:
I'll have time in a few hours, if it can wait.

core/src/main/scala/cats/Apply.scala
```diff
@@ -62,7 +63,7 @@ trait Apply[F[_]] extends Functor[F] with Semigroupal[F] with ApplyArityFunction
    * map2 can be seen as a binary version of [[cats.Functor]]#map.
    */
   def map2[A, B, Z](fa: F[A], fb: F[B])(f: (A, B) => Z): F[Z] =
-    map(product(fa, fb)) { case (a, b) => f(a, b) }
+    ap(map(fa)(a => (b: B) => f(a, b)))(fb)
```
kailuowang: This is a sensible improvement if we only consider the default implementation, but many instances overrides `product` for better performance. So we might want to keep the original implementation of `product` and `map2`

kailuowang: actually the law is broken with `Const` instance which overrides `product`
>ConstSuite:
[info] 
[info] - Apply[Const[NonEmptyList[String], Int]].apply.map2/product-map consistency *** FAILED ***
[info]     Falsified after 3 successful property evaluations.
[info]       arg0 = Const(NonEmptyList(ꙫ, 鰓)),
[info]       arg1 = Const(NonEmptyList(, )),
[info]       arg2 = <function2>
[info]     )
[info]       Expected: Const(NonEmptyList(, , ꙫ, 鰓))
[info]   Received: Const(NonEmptyList(ꙫ, 鰓, , ))

Jacoby6000: Ah Yeah, that was a bit short sighted of me. I didn't consider overrides.

kailuowang:
Thanks very much! @Jacoby6000

LukaJCB:
Yes, thanks a lot, Jacob! :)

scalafix/rules/src/main/scala/fix/Cats_v1_0_0.scala
```diff
@@ -242,6 +242,23 @@ case class RenameTransformersLift(index: SemanticdbIndex)
...
+    ctx.replaceSymbols(
+      "_root_.cats.Apply.forEffect." -> "apL",
+      "_root_.cats.Apply.followedBy." -> "apR",
+      "_root_.cats.Apply.Ops.forEffect." -> "apL",
+      "_root_.cats.Apply.Ops.followedBy." -> "apR",
+      "_root_.cats.NonEmptyParallel.parForEffect." -> "parApL",
+      "_root_.cats.NonEmptyParallel.parFollowedBy." -> "parApR",
+      "_root_.cats.FlatMap.forEffectEval." -> "apLEval",
+      "_root_.cats.FlatMap.followedByEval." -> "apREval",
+    )
```
peterneyens: These rewrite rules still need to be updated to `productR`/`productL`, right?

Jacoby6000: Ah yep, they sure do.  I missed those.

kailuowang: fix submitted https://github.com/typelevel/cats/pull/2117

PR #2084, opened 2017-12-08
correct `>>` symbol in FAQ

kailuowang:


docs/src/main/tut/faq.md
```diff
@@ -218,7 +218,7 @@ All other symbols can be imported with `import cats.implicits._`
 | `x === y`                        | equals                   |                  | `Eq[A]`                 | `eqv(x: A, y: A): Boolean`                                          |
 | `x =!= y`                        | not equals               |                  | `Eq[A]`                 | `neqv(x: A, y: A): Boolean`                                         |
 | `fa >>= f`                       | flatMap                  |                  | `FlatMap[F[_]]`         | `flatMap(fa: F[A])(f: A => F[B]): F[B]`                             |
-| `fa >> fb`                       | followed by              |                  | `FlatMap[F[_]]`         | `followedBy(fa: F[A])(fb: F[B]): F[B]`                              |
+| `fa >> fb`                       | followed by              |                  | `FlatMap[F[_]]`         | `>>[B](fb: => F[B]): F[B] = F.flatMap(fa)(_ => fb)`                              |
```
ceedubs: I may be missing something, but it looks like the other methods have the non-symbolic name in the `Signature` column?

kailuowang: This one seems without non-symbolic counterpart when added in #2043

LukaJCB: Maybe we should add one, and `<<` as well? WDYT?

kailuowang: The most sensible name I can think of is `followedBy`, but it's taken by `Applicative` and it won't be released even after the rename. So I don't know. From the discussion #1983 seems that the community's preference is to not have name for such operations. So personally I am fine with this one without name. 
Maybe we should fix the documentation here first?

LukaJCB: Agreed

kailuowang:
@ceedubs WDYT?

PR #2086, opened 2017-12-09
Add doctest examples for Apply

ceedubs:


kailuowang:
What's your opinion over #2083?

ceedubs:
Thanks for pointing that out @kailuowang. I just left a comment on that PR. I think that we should get that one in first since it has the incompatible changes. I can update this PR afterwards.

kailuowang:
#2083 is merged

ceedubs:
I should be able to update this today.

ceedubs:
I've merged in the changes from master and resolved the conflicts.

PR #2088, opened 2017-12-10
Add InvariantSemigroupal and ability to turn Monoidals to Monoids

LukaJCB:
Adds `InvariantSemigroupal` as supertype of `InvariantMonoidal` and just like with `Apply` and `Applicative` we have functions that turn invariant and contravariant `Monoidal` and `Semigroupal` Functors into `Monoids` and `Semigroup`s  respectively. 
I'll try to add tests for these, as well as the ones in `Apply` and `Applicative` soon.


core/src/main/scala/cats/Composed.scala
```diff
@@ -136,6 +136,18 @@ private[cats] trait ComposedSemigroupal[F[_], G[_]] extends ContravariantSemigro
     }
 }
 
+private[cats] trait ComposedInvariantApplySemigroupal[F[_], G[_]] extends InvariantSemigroupal[λ[α => F[G[α]]]] with ComposedInvariantCovariant[F, G] { outer =>
+  def F: InvariantSemigroupal[F]
+  def G: Apply[G]
+
+  def product[A, B](fa: F[G[A]], fb: F[G[B]]): F[G[(A, B)]] =
+    F.imap(F.product(fa, fb)) { case (ga, gb) =>
```
kailuowang: do you plan to add tests for this instance?

core/src/main/scala/cats/InvariantSemigroupal.scala
```diff
@@ -0,0 +1,28 @@
...
+
+import simulacrum.typeclass
+
+/**
+  * [[InvariantSemigroupal]] is nothing more than something both invariant
+  * and Semigroupal. It comes up enough to be useful, and composes well
+  */
+@typeclass trait InvariantSemigroupal[F[_]] extends Semigroupal[F] with Invariant[F] { self =>
+
+   def composeApply[G[_]: Apply]: InvariantSemigroupal[λ[α => F[G[α]]]] =
```
kailuowang: Any plan to test cover this one as well?

core/src/main/scala/cats/InvariantSemigroupal.scala
```diff
@@ -0,0 +1,28 @@
...
+     new ComposedInvariantApplySemigroupal[F, G] {
+       def F = self
+       def G = Apply[G]
+     }
+
+
+}
+
+object InvariantSemigroupal extends SemigroupalArityFunctions {
+  def semigroup[F[_], A](implicit f: InvariantSemigroupal[F], sg: Semigroup[A]): Semigroup[F[A]] =
```
kailuowang: The diff coverage is a bit low, one thing we could do is to add doctests for this `semigroup` method and the one in `ContravariantSemitgroupal`, which could also serve as examples. Not critically important though.

kailuowang:
Overall looking good to me. Could use a couple of more tests for the composed instances.

LukaJCB:
Yeah this is still a WIP, wanted to add a bunch more tests, for Applicative/Apply as well.

LukaJCB:
The more I work with this the more I think `Apply` and `ContravariantSemigroupal` should extend `InvariantSemigroupal` and the same applies to `Applicative` and `ContravariantMonoidal`. 
It would definitely simplify a lot of the compositional things. :)

LukaJCB:
I moved the `semigroup` and `monoid` methods to the type classes instead. 
This gives us the benefit, that we can optimize this for any instance.

LukaJCB:
So the version right now works without breaking changes, but it'd probably be cleaner to define `InvariantMonoidal` in terms of `InvariantSemigroupal` + `def unit: F[Unit]` and deriving `pure` from there. 
Another thing that bothers me a bit is that I think every `Semigroupal` is bound to also be an `Invariant` Functor, no? Maybe I'm wrong about this, but I don't think it's possible. It would also simplify law testing because we can now `imap` over the resulting `F`.

kailuowang:
> InvariantMonoidal in terms of InvariantSemigroupal + def unit: F[Unit] and deriving pure from there.

Then `Applicative` won't be able to inherit from `InvariantMonoidal` right? It would be the otherway around?

LukaJCB:
Nope, that's no problem, in fact if you look at `Applicative` right now, it already derives `unit` from `pure`:

```scala 
trait Applicative[F[_]] extends Apply[F] with InvariantMonoidal[F] {
  def pure[A](a: A): F[A]
  def unit: F[Unit] = pure(())
}
```

kailuowang:
hmm, I admit I am a bit confused, in `ContravariantMonoidal` the `unit` type is 
`def unit[A]: F[A]`. It might not be able to define another `def unit: F[Unit]` ?

LukaJCB:
It would have to be changed or renamed, yeah :/
 
Something like: 
```scala
trait ContravariantMonoidal[F[_]] extends ContravariantSemigroupal[F] with InvariantMonoidal[F] {
  def conquer[A]: F[A]
  def unit: F[Unit] = conquer[Unit] 
}
```

Or we just remove it, since you can `contramap` to any type from an `F[Unit]` by simply doing `val fa: F[A] = F.unit.contramap(_ => ())`

LukaJCB:
Another argument I find compelling is that literature always only refers to `Semigroupal` or `Monoidal` as `Functors`. i.e. `Lax Monoidal Functor`, `Semigroupal Endofunctor`. 
So having a `Semigroupal` that's not a `Functor` doesn't seem to make a lot of sense naming wise, I think.

kailuowang:
did you take a look at the `FreeInvarantMonoidal`?

LukaJCB:
I'm certain `FreeInvariantMonoidal` can stay exactly as is, with just a small change to its `InvariantMonoidal` instance. I have a build that passes the tests with all the changes I mentioned locally. :)

kailuowang:
I think to get to the bottom of this, we need to understand why `Monoidal` has been avoided and why `pure` instead of `unit: F[Unit]` in `InvariantMonoidal`. 
calling @adelbertc @ceedubs @mpilquist

kailuowang:
also calling @OliverBlanvillain for  why `pure(a:A): F[A]` instead of `unit: F[Unit]` in `InvariantMonoidal`.

LukaJCB:
Would also like to hear what @julienrf thinks :)

LukaJCB:
Here's a branch I made where I made the changes outlined.
https://github.com/LukaJCB/cats/commit/111cd15a5821060372485165b3dbc14c2bb63896

core/src/main/scala/cats/Applicative.scala
```diff
@@ -184,11 +172,14 @@ import simulacrum.typeclass
    */
   def whenA[A](cond: Boolean)(f: => F[A]): F[Unit] =
     if (cond) void(f) else pure(())
+
+  override def monoid[A](implicit A: Monoid[A]): Monoid[F[A]] =
```
johnynek: every time we add a method to a trait, we break binary compatibility. Is there any reason to have this here vs the `object`? Can you really customize this based on the `F[_]`?

core/src/main/scala/cats/Apply.scala
```diff
@@ -79,11 +79,14 @@ trait Apply[F[_]] extends Functor[F] with Semigroupal[F] with ApplyArityFunction
       val F = self
       val G = Apply[G]
     }
+
+  override def semigroup[A](implicit A: Semigroup[A]): Semigroup[F[A]] =
```
johnynek: same question about customization.

core/src/main/scala/cats/ContravariantMonoidal.scala
```diff
@@ -11,12 +11,24 @@ import simulacrum.typeclass
...
    * `unit` produces an instance of `F` for any type `A`
    * that is trivial with respect to `contramap2` along
    * the diagonal
    */
   def unit[A]: F[A]
+
+  override def pure[A](a: A): F[A] = unit
+
+  override def monoid[A](implicit A: Monoid[A]): Monoid[F[A]] =
+    ContravariantMonoidal.monoid(this)
```
johnynek: this seems to be breaking the previous pattern: we are calling the object from the trait, but in the other cases we call it the opposite direction.

core/src/main/scala/cats/Applicative.scala
LukaJCB: Hmm, maybe you are right. We had the same discussion with regards to parallel, but I think in this case it might not be as customizable.

OlivierBlanvillain:
It was a long time ago, but if I remember correctly having `pure(a: A): F[A]` instead of `unit: F[Unit]` was just a design choice. For the use case of codecs it might be faster to create `Codec[A]` directly instead of always going thought `Codec[Unit]` + `imap`.

core/src/main/scala/cats/ContravariantMonoidal.scala
LukaJCB: Yes, this is because we can define a `ContravariantMonoidal` regardless of if `A` is a Monoid or not.

julienrf:
Hi, I won’t have time to have deep look at the PR contents, but I think this is great to have this abstraction. I’m using it, for instance, [here](https://github.com/julienrf/endpoints/blob/a4e5e2f55651cabef9e49c6e81c0155a5627abeb/openapi/json-schema/src/main/scala/endpoints/documented/algebra/JsonSchemas.scala#L85-L89), in _endpoints_.

ceedubs:
Regarding `pure` vs `unit`, here is some GitHub history:

- [here](https://github.com/typelevel/cats/issues/817#issuecomment-175625517) people seem to align on having `unit` instead of `pure`.
- [here](https://github.com/typelevel/cats/pull/845#issuecomment-221098061) @OlivierBlanvillain says that they should be equivalent, but `pure` seems to be more elegant.

It does seem like some instances would want to override `pure` to make it more efficient. But it seems to be fairly arbitrary that `pure` was chosen over `unit`.

ceedubs:
I've been a bit out of touch with the various `Semigroupal`/`Monoidal` changes. Would someone (maybe @LukaJCB?) be willing to provide a bit of information about the motivation behind this PR and how it will be useful? Bonus points if it's in documentation as opposed to just GitHub comments :)

LukaJCB:
Hey Cody, I'm on the phone right now, but the gist is that we have `Functor` and `Contravariant` as sub types of `Invariant`. And now we also have semigroupal and monoidal versions of these Functors. I.e every `Applicative` is also an `InvariantMonoidal` and every `ContravariantMonoidal` is as well. The same goes for `Apply` ,`ContravariantSemigroupal` and `InvariantSemigroupal`. So the relationship between these types exist, but aren't represented in our hierarchy right now, which hurts composition. 
The other part of this, is that we can construct a `Monoid (f a)` for every Monoidal Functor `f` and the same applies to Semigroups and semigroupal Functors. I think making these relationships more clear will make things clearer for users. :)

LukaJCB:
To clarify the hierarch now looks like this:

(Semigroupal, Invariant) -> InvariatSemigroupal
InvariantSemigroupal -> Apply

kailuowang:
did we find anything that is `Semigroupal` but not `Invariant` or anything that is `Monoidal` but not `Invariant`?

LukaJCB:
There's nothing in cats, however we had this conversation on gitter:


> LukaJCB @LukaJCB Dec 13 09:51
Is there any type F[_] that isn't also Invariant? I can't come up with anything
My intuition is every F[A] will either have zero or more values of A in contra, Invariant or covariant position. In every case I'm able to define an Invariant simply by applying the two functions given to imap

> Rob Norris @tpolecat Dec 13 15:20
@LukaJCB as long as you use the scalazzi set I think you’re right, but Set for instance isn’t invariant because it violates patametricity.

> Cody Allen @ceedubs Dec 13 15:55
@LukaJCB @tpolecat what about something like this?
```scala
sealed abstract class Op[A]

object Op {
  final case class PutStrLn(s: String) extends Op[Unit]

  implicit val invOp: Invariant[Op] = new Invariant[Op] {
    // if we consider `PutStrLn` to be the only constructor for `Op`, then we
    // can't get an `Op[B]` for all `B`
    def imap[A, B](fa: Op[A])(f: A => B)(g: B => A): Op[B] = ???
  }
}
```

Both of which wouldn't really be able to define a valid `Semigroupal` though.

kailuowang:
We have roughly two basic groups:
1.  `Invariant`, `Contravariant` and `Functor`
2. `Semigroupal` and `Monoidal` (`product` and `unit`)

Then we have the combinations between group 1 and 2. 
`InvariantSemigroupal`, `ContravariantSemigroupal`, `InvariantMonoidal` and `ContravariantMonoidal`

Then we have `Apply` and `Applicative`, It seems to me that
`Apply` is equivalent to `FunctorSemigroupal`
 `Applicative` is equivalent to `FunctorMonoidal`

Right?

LukaJCB:
Yes, that's pretty much exactly what it is

LukaJCB:
Oh I'm sorry, I got the hierarchy wrong, it is like you said, all the monoidals extend their respective semigroupal

core/src/main/scala/cats/ContravariantMonoidal.scala
```diff
@@ -11,12 +11,22 @@ import simulacrum.typeclass
...
-@typeclass trait ContravariantMonoidal[F[_]] extends ContravariantSemigroupal[F] { self =>
+@typeclass trait ContravariantMonoidal[F[_]] extends ContravariantSemigroupal[F] with InvariantMonoidal[F] {
   /**
-   * `unit` produces an instance of `F` for any type `A`
+   * `conquer` produces an instance of `F` for any type `A`
    * that is trivial with respect to `contramap2` along
    * the diagonal
    */
-  def unit[A]: F[A]
+  def conquer[A]: F[A] = contramap(unit)(_ => ())
```
ceedubs: I don't want to bikeshed, but I'm not crazy about `conquer` as a name. I'm guessing it was just a cute name in the context of `Divisible` (`divide`/`conquer`)? Or is there a more meaningful reason behind the name?

Throwing out a few names, none of which I like all that much: `trivial`, `emptyK`, `pureEmpty`

tests/src/test/scala/cats/tests/ListWrapper.scala
```diff
@@ -43,25 +42,27 @@ object ListWrapper {
 
   def eqv[A : Eq]: Eq[ListWrapper[A]] = Eq.by(_.list)
 
-  val traverse: Traverse[ListWrapper] = {
+  val traverse: Traverse[ListWrapper] with InvariantSemigroupal[ListWrapper] = {
```
ceedubs: I'm not sure that this return type should be changed. There may be cases where we are writing tests to ensure that an implicit instance is found when we have only a `Traverse` instance for `F`, and those tests would no longer be valid.

ceedubs: We could always add a separate member if you want want that has this full type returned.

core/src/main/scala/cats/ContravariantMonoidal.scala
LukaJCB: I like trivial!

tests/src/test/scala/cats/tests/ListWrapper.scala
LukaJCB: You're right, I'll change it!

ceedubs:
In general, I think that I like the symmetry of adding this. I left a couple of minor comments.

Also I think that there are a couple of pending comments of @johnynek. I'm inclined to take his advice and put these methods into companion object if we can't think of an instance that might want to override them.

kailuowang:
So, for the sake of symmetry shall we add Monoidal?

LukaJCB:
I don't think it makes sense to define a `Monoidal` without a Functor, we wouldn't have any instances for it and the laws can only really be expressed with some notion of a Functor. 
As Rob said, if we don't violate parametricity, `Semigroupal` and `Monoidal` don't really make much sense without atleast an `Invariant` Functor.  I'm fine with keeping `Semigroupal` around for compatibility reasons, but I don't think adding `Monoidal` makes a lot of sense. :)

core/src/main/scala/cats/instances/invariant.scala
```diff
@@ -1,27 +1,34 @@
 package cats.instances
 
 import cats.kernel._
-import cats.InvariantMonoidal
+import cats.{InvariantMonoidal, Monoid, InvariantSemigroupal}
 
 trait InvariantMonoidalInstances {
 
+  implicit def catsSemigroupalForMonoid: InvariantSemigroupal[Monoid] = new InvariantSemigroupal[Monoid] {
```
kailuowang: I am curious why not move these two to the companion of `Monoid` and `Semigroup` to avoid a required import?

kailuowang:
Left a minor comment, everything else looking great to me. This PR really brings things together. Thank you @LukaJCB

core/src/main/scala/cats/instances/invariant.scala
LukaJCB: Because they're in kernel :/

kailuowang: oh Duh

kailuowang:
@ceedubs , @johnynek it seems all issues addressed?

core/src/main/scala/cats/InvariantSemigroupal.scala
ceedubs: Has this been addressed?

core/src/main/scala/cats/instances/invariant.scala
```diff
@@ -1,27 +1,34 @@
...
-        fa.combineAllOption(bs.map(g)).map(f)
     }
 
-    def pure[A](a: A): Semigroup[A] = new Semigroup[A] {
-      def combine(x: A, y: A): A = a
-      override def combineAllOption(as: TraversableOnce[A]): Option[A] =
-        if (as.isEmpty) None
-        else if (as.size == 1) as.toList.headOption
-        else Some(a)
+    def unit: Semigroup[Unit] = new Semigroup[Unit] {
```
ceedubs: You should be able to return the `Unit` group instance already defined in `kernel`.

core/src/main/scala/cats/instances/invariant.scala
```diff
@@ -32,16 +39,10 @@ trait InvariantMonoidalInstances {
...
-        fa.combineAllOption(bs.map(g)).map(f)
     }
 
-    def pure[A](a: A): CommutativeSemigroup[A] = new CommutativeSemigroup[A] {
-      def combine(x: A, y: A): A = a
-      override def combineAllOption(as: TraversableOnce[A]): Option[A] =
-        if (as.isEmpty) None
-        else if (as.size == 1) as.toList.headOption
-        else Some(a)
+    def unit: CommutativeSemigroup[Unit] = new CommutativeSemigroup[Unit] {
```
ceedubs: Again, this already exists in kernel.

core/src/main/scala/cats/instances/order.scala
```diff
@@ -7,7 +7,7 @@ trait OrderInstances extends cats.kernel.OrderToOrderingConversion {
       /**
        * Provides trivial order
        */
-      def unit[A]: Order[A] = Order.from[A]((x: A, y: A) => 0)
+      def unit: Order[Unit] = Order.from[Unit]((x: Unit, y: Unit) => 0)
```
ceedubs: Exists in kernel :)

core/src/main/scala/cats/instances/ordering.scala
```diff
@@ -8,8 +8,8 @@ trait OrderingInstances {
        * Note: resulting instances are law-abiding only when the functions used are injective (represent a one-to-one mapping)
        */
 
-      def unit[A]: Ordering[A] = new Ordering[A] {
-        def compare(l: A, r: A): Int = 0
+      def unit: Ordering[Unit] = new Ordering[Unit] {
```
ceedubs: Not sure if this exists in kernel, but it probably should.

tests/src/test/scala/cats/tests/ListWrapper.scala
```diff
@@ -61,7 +60,13 @@ object ListWrapper {
 
   val functor: Functor[ListWrapper] = traverse
 
-  val invariant: Invariant[ListWrapper] = functor
+  val invariant: InvariantSemigroupal[ListWrapper] = new InvariantSemigroupal[ListWrapper] {
```
ceedubs: I have the same comment here that I previously left on the `traverse` instance. There may be tests that are checking implicit resolution when `invariant` is not returning anything more specific than `Invariant`. If we need to, we can add a separate method that returns an `InvariantSemigroupal[ListWrapper]`.

ceedubs:
Thanks @LukaJCB! I had a few comments, but I think that it's getting really close.

LukaJCB:
Great, I'll get it fixed in the next hour or so!

core/src/main/scala/cats/InvariantSemigroupal.scala
LukaJCB: Yep!

LukaJCB:
Everything should be adressed :)

core/src/main/scala/cats/Composed.scala
```diff
@@ -144,6 +144,18 @@ private[cats] trait ComposedSemigroupal[F[_], G[_]] extends ContravariantSemigro
...
+      G.map2(ga, gb)(_ -> _)
+    } { g: G[(A, B)] =>
+      (G.map(g)(_._1), G.map(g)(_._2))
```
kailuowang: This line is not tested, we might need to test an instance that is not a covariant.

laws/src/main/scala/cats/laws/InvariantMonoidalLaws.scala
```diff
@@ -4,16 +4,16 @@ package laws
...
  */
-trait InvariantMonoidalLaws[F[_]] extends InvariantLaws[F] with SemigroupalLaws[F] {
+trait InvariantMonoidalLaws[F[_]] extends InvariantSemigroupalLaws[F] {
   override implicit def F: InvariantMonoidal[F]
   import cats.syntax.semigroupal._
   import cats.syntax.invariant._
 
   def invariantMonoidalLeftIdentity[A, B](fa: F[A], b: B): IsEq[F[A]] =
-    F.pure(b).product(fa).imap(_._2)(a => (b, a)) <-> fa
+    F.point(b).product(fa).imap(_._2)(a => (b, a)) <-> fa
```
kailuowang: We can express these laws in `unit` right?  like
```scala
def invariantMonoidalLeftIdentity[A, B](fa: F[A]): IsEq[F[A]] =
 F.unit.product(fa).imap(_._2)(a => ((), a)) <-> fa		 
```
Then we can add a `point` - `unit` consistency law. WDYT?

core/src/main/scala/cats/data/Const.scala
```diff
@@ -136,7 +136,7 @@ private[data] sealed abstract class ConstInstances0 extends ConstInstances1 {
 
 private[data] sealed abstract class ConstInstances1 {
   implicit def catsConstInvariantMonoidal[C: Monoid]: InvariantMonoidal[Const[C, ?]] = new InvariantMonoidal[Const[C, ?]] {
-    def pure[A](a: A): Const[C, A] =
+    def unit: Const[C, Unit] =
       Const.empty
```
kailuowang: This one is not tested. I suspect that the instance was never tested before?

core/src/main/scala/cats/data/IndexedStateT.scala
```diff
@@ -387,11 +387,11 @@ private[data] sealed abstract class IndexedStateTContravariantMonoidal[F[_], S]
   implicit def F: ContravariantMonoidal[F]
   implicit def G: Applicative[F]
 
-  override def unit[A]: IndexedStateT[F, S, S, A]  =
-    IndexedStateT.applyF(G.pure((s: S) => F.unit[(S, A)]))
+  override def unit: IndexedStateT[F, S, S, Unit]  =
+    IndexedStateT.applyF(G.pure((s: S) => F.trivial[(S, Unit)]))
```
kailuowang: It's reported that this instance is not tested. Shall add one?

core/src/main/scala/cats/instances/eq.scala
```diff
@@ -8,7 +8,7 @@ trait EqInstances {
        * Defaults to the trivial equivalence relation
        * contracting the type to a point
        */
-      def unit[A]: Eq[A] = Eq.allEqual
+      def unit: Eq[Unit] = Eq.allEqual
```
kailuowang: not quite sure I understand why this one is reported as not tested either. Any idea?

laws/src/main/scala/cats/laws/InvariantMonoidalLaws.scala
LukaJCB: Agreed!

core/src/main/scala/cats/data/IndexedStateT.scala
kailuowang: the ContravariantMonoidal instance of IndexedStateT requires that F to have both FlatMap and ContravriantMonoidal. If Phantom is the only one that satisfies that I don’t know if it’s worthwhile to have it.

kailuowang: I incline to remove this instance, WDYT @stephen-lazaro ?

stephen-lazaro: Agreed, it seems at best awkward and unuseful.

stephen-lazaro: Agreed, it seems at best awkward and unuseful.

LukaJCB:
Sorry @kailuowang, needed to add a couple more mima exceptions

kailuowang:
close and reopen to trigger build and recheck of conflict

PR #2091, opened 2017-12-10
Simplified `traverse` method for vectors by using `foldLeft`

amitayh:
By doing this, there's no need for the `Eval` wrapper and avoiding the
`foldRight` which could potentially be more efficient


ceedubs:
Thanks for bringing this up, @amitayh.

It could _potentially_ be more efficient, but `foldLeft` is also always going to do a full traversal of the vector. For example consider the code `(1 to 100000).toVector.traverse[Option, Int]{i => println(i); None}`. With the current implementation, it will print `1` and then halt. With your change it will print the full range.

I actually thought that we had some tests that were checking that `traverse` was short-circuiting, but it appears that we don't at least for `Vector`. While there are definitely trade-offs between the two approaches, I'm inclined to think that the current implementation is probably a better default. I think that we should probably add some tests to ensure the traverse is short-circuiting the structure when possible.

ceedubs:
It looks like we have these tests for `List`, `Stream`, and a couple of other types. I'll submit a PR to add them for `Vector`.

johnynek:
Can we add benchmarks when we are making such trade offs?

amitayh:
Is there some tool you're using for benchmarking? I can try to run a few tests...

PR #2092, opened 2017-12-10
Test that Vector's traverse short-circuits

ceedubs:
This adds a test case that the `traverse` implementation for `Vector`
short-circuits the traversal when possible. This would have failed the build
in #2091.

See #1015 for when similar tests were added for similar structures.


LukaJCB:
:+1:

PR #2094, opened 2017-12-11
minor fix for test of `Applicative` from `Arrow`. 

kailuowang:


kailuowang:
this approach doesn't work

PR #2095, opened 2017-12-11
Update guidelines.md with implicit priority

kailuowang:


docs/src/main/tut/guidelines.md
```diff
@@ -84,5 +84,32 @@ As an example, an implicit instance of `Monoid` for `List` defined in the packag
...
+
+...
+object Kleisli extends KleisliInstance0 
+
+abstract class KleisliInstance0 extends KleisliInstance1 {
+  implicit def catsDataMonadForKleisli[F[_], A]: Monad[Kleisli[F, A, ?]] = ...
+}
+
+abstract class KleisliInstance1 {
+  implicit def catsDataFunctorForKleisli[F[_], A]: Functor[Kleisli[F, A, ?]] = ...
```
ceedubs: It's really outside of the point of this example, but my mental compiler complained that you would need `F[_]:Monad` and `F[_]:Functor` constraints to be able to implement them :)

Issue #2067, opened 2017-12-06
Add ArrowChoice


stephen-lazaro:
Would something like the ArrowChoice typeclass from https://hackage.haskell.org/package/base-4.10.1.0/docs/Control-Arrow.html#t:ArrowChoice be of interest in `cats.arrow`?
Every Kleisli arrow is an instance for monads, and I have an application that is not strictly a Kleisli arrow but a sort of wrapped one at my workplace that I contemplated using this for.
I can work on a PR If so.

kailuowang:
Sounds a good addition to me. And we still have time before the Dec 12 deadline (after which we should still be able to add it but won't have the syntax available from `cats.implicits._ `)

stephen-lazaro:
Cool, I'll try to cook something up. Thanks Kailuo.

stephen-lazaro:
Resolved in #2096

PR #2096, opened 2017-12-12
Add Arrow Choice

stephen-lazaro:
Coming in right under the wire, this resolves issue https://github.com/typelevel/cats/issues/2067.
The laws are admittedly hideous, but seem morally correct.  I believe I caught all the binary incompatibilities.
I also updated the symbols table with the new `+++` for `choose` (whose name I feel somewhat undecided about).
If coverage complains, I'll add more doc tests for the `ArrowChoice` methods.


core/src/main/scala/cats/arrow/ArrowChoice.scala
```diff
@@ -0,0 +1,43 @@
...
+  def choose[A, B, C, D](f: F[A, C])(g: F[B, D]): F[Either[A, B], Either[C, D]]
+
+  def left[A, B, C](fab: F[A, B]): F[Either[A, C], Either[B, C]] =
+    choose(fab)(lift(identity[C]))
+
+  def right[A, B, C](fab: F[A, B]): F[Either[C, A], Either[C, B]] =
+    choose(lift(identity[C]))(fab)
+
+  override def choice[A, B, C](f: F[A, C], g: F[B, C]): F[Either[A, B], C] =
+    rmap(choose(f)(g))(_.fold(identity, identity))
```
julienrf: I noticed that `choice` doesn’t have `|||` as a symbolic alias. This is not really related to this PR but that would be a good thing to have.

core/src/main/scala/cats/data/Kleisli.scala
```diff
@@ -147,11 +146,10 @@ private[data] sealed abstract class KleisliInstances0 extends KleisliInstances1
   implicit def catsDataMonadForKleisliId[A]: Monad[Kleisli[Id, A, ?]] =
     catsDataMonadForKleisli[Id, A]
 
-  implicit def catsDataCommutativeArrowForKleisli[F[_]](implicit M: CommutativeMonad[F]): CommutativeArrow[Kleisli[F, ?, ?]] =
-    new KleisliCommutativeArrow[F] {def F: CommutativeMonad[F] = M }
-
-  implicit val catsDataCommutativeArrowForKleisliId: CommutativeArrow[Kleisli[Id, ?, ?]] =
-    catsDataCommutativeArrowForKleisli[Id]
+  implicit def catsDataArrowForKleisli[F[_]](implicit M: Monad[F]): Arrow[Kleisli[F, ?, ?]] =
```
kailuowang: If I am not mistaken, since both `Arrow` and `ArrowChoice` instances have the same requirement (`Monad[F]`), one `ArrowChoice` instance should be fine. You can basically leave `catsDataCommutativeArrowForKleisliId` and `catsDataCommutativeArrowForKleisli` as is.

core/src/main/scala/cats/data/Kleisli.scala
```diff
@@ -237,6 +241,18 @@ private[data] trait KleisliCommutativeArrow[F[_]] extends CommutativeArrow[Kleis
   implicit def F: CommutativeMonad[F]
 }
 
+private[data] trait KleisliArrowChoice[F[_]] extends ArrowChoice[Kleisli[F, ?, ?]] with KleisliArrow[F] {
```
kailuowang: We can fold `KleisliArrow` into `KleisliArrowChoice` right? `KleisliCommutativeArrow` can extend `KleisliArrowChoice` .

core/src/main/scala/cats/data/Kleisli.scala
```diff
@@ -161,6 +159,12 @@ private[data] sealed abstract class KleisliInstances1 extends KleisliInstances2
   implicit def catsDataMonadForKleisli[F[_], A](implicit M: Monad[F]): Monad[Kleisli[F, A, ?]] =
     new KleisliMonad[F, A] { def F: Monad[F] = M }
 
+  implicit def catsDataCommutativeArrowForKleisli[F[_]](implicit M: CommutativeMonad[F]): CommutativeArrow[Kleisli[F, ?, ?]] =
```
kailuowang: These can return `CommutativeArrow with ArrowChoice`

laws/src/main/scala/cats/laws/ArrowChoiceLaws.scala
```diff
@@ -0,0 +1,43 @@
...
+
+import cats.arrow.ArrowChoice
+import cats.syntax.compose._
+
+/**
+ * Laws that must be obeyed by any `cats.arrow.ArrowChoice`.
+ */
+trait ArrowChoiceLaws[F[_, _]] extends ArrowLaws[F] with ChoiceLaws[F] {
+  implicit override def F: ArrowChoice[F]
+  implicit def Function: ArrowChoice[Function1]
```
kailuowang: Are these laws based on the ones defined in [Control-ArrowChoice](https://hackage.haskell.org/package/base-4.10.1.0/docs/Control-Arrow.html)?

stephen-lazaro: Hmm, they are meant to but interestingly the ones in the source and the ones in the Haskell doc do not quite look the same, I'll give them another pass tonight to make sure I transcribed the correct set.

core/src/main/scala/cats/data/Kleisli.scala
stephen-lazaro: Sounds good, will do.

core/src/main/scala/cats/arrow/ArrowChoice.scala
stephen-lazaro: Good call, I can add.

laws/src/main/scala/cats/laws/ArrowChoiceLaws.scala
stephen-lazaro: Yeah, I totally misread them, somehow. How embarrassing! Sorry about that. Fixing.

core/src/main/scala/cats/instances/function.scala
```diff
@@ -76,13 +76,13 @@ private[instances] sealed trait Function1Instances extends Function1Instances0 {
...
-  implicit val catsStdInstancesForFunction1: Choice[Function1] with CommutativeArrow[Function1] =
-    new Choice[Function1] with CommutativeArrow[Function1] {
-      def choice[A, B, C](f: A => C, g: B => C): Either[A, B] => C = {
-        case Left(a)  => f(a)
-        case Right(b) => g(b)
-      }
+  implicit val catsStdInstancesForFunction1: ArrowChoice[Function1] with CommutativeArrow[Function1] =
+    new ArrowChoice[Function1] with CommutativeArrow[Function1] {
+      def choose[A, B, C, D](f: A => C)(g: B => D): Either[A, B] => Either[C, D] =
+        _.fold(
```
kailuowang: using patten match rather than `fold` has performance gain. see https://github.com/typelevel/cats/issues/1951

core/src/main/scala/cats/arrow/ArrowChoice.scala
```diff
@@ -0,0 +1,43 @@
...
+   * }}}
+   */
+  @simulacrum.op("+++", alias = true)
+  def choose[A, B, C, D](f: F[A, C])(g: F[B, D]): F[Either[A, B], Either[C, D]]
+
+  def left[A, B, C](fab: F[A, B]): F[Either[A, C], Either[B, C]] =
+    choose(fab)(lift(identity[C]))
+
+  def right[A, B, C](fab: F[A, B]): F[Either[C, A], Either[C, B]] =
+    choose(lift(identity[C]))(fab)
```
kailuowang: This is never tested by law as I noticed from the law set defined in Haskell. Thus it doesn't have any test coverage. 
Shall we add a law for this something like 
```scala
  def leftRightConsistent[A, B, C](f: A => B): IsEq[F[Either[A, C], Either[B, C]]] =
     F.right[A, B, C](F.lift[A, B](f)) <-> F.left[A, B, C](F.lift[A, B](f)).dimap(_.swap, _.swap)
```

core/src/main/scala/cats/instances/function.scala
stephen-lazaro: Yup, can do. I tend to forget that end of things.

core/src/main/scala/cats/arrow/ArrowChoice.scala
stephen-lazaro: I actually added symmetric versions for all the laws last night for that reason, but it takes the method well over 50 lines. Something like that sounds like a more reasonable way to get similar effect. I'll try it out today, it certainly should work.

stephen-lazaro:
Thanks @kailuowang!

PR #2097, opened 2017-12-12
added distributeIdentityLaw

kailuowang:


kailuowang:
@coltfred no worries. 
I thought about removing `cosequenceIdentity`, but as you suggested it's verifying the consistency between the two.

PR #2098, opened 2017-12-12
Add a CommutativeMonoid for Map

ceedubs:
Since we already have a `Monoid` instance for `Map`, it seems natural to
have a `CommutativeMonoid` when the value type has a commutative
semigroup.

Note: Unfortunately, I believe that this is a binary-incompatible change in cats kernel. cc @johnynek

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -334,7 +334,8 @@ def mimaSettings(moduleName: String) = Seq(
       exclude[InheritedNewAbstractMethodProblem]("cats.instances.Function0Instances0.function0Distributive"),
       exclude[InheritedNewAbstractMethodProblem]("cats.instances.Function1Instances0.catsStdDistributiveForFunction1"),
       exclude[InheritedNewAbstractMethodProblem]("cats.instances.Function1Instances0.catsStdDistributiveForFunction1"),
-      exclude[InheritedNewAbstractMethodProblem]("cats.instances.Function1Instances0.catsStdDistributiveForFunction1")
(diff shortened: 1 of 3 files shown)


denisrosset:
Right now, Spire has its own instances; the proposed definition is sane and sound.

I have a minor complaint: both the commutative and noncommutative tests of `Monoid[Map[K,V]]` use commutative monoids on the values.

Consider changing the `Monoid[Map[String, Int]]` to `Monoid[Map[String, String]]`, so that if the instance does not behave correctly with noncommutative types, the tests can catch it. This problem is currently biting us in Spire.

ceedubs:
@denisrosset thanks for the review. You have a good point regarding the tests. Just to confirm that I understand you, you are suggesting leaving the new test as it is, but changing the existing `MonoidTests[Map[String, Int]]` call to `MonoidTests[Map[String, String]]`, right?

denisrosset:
@ceedubs, exactly. Not a big deal in this case, but a good habit to keep: try to not use for tests an instance that has more structure (i.e. commutativity) than required.

ceedubs:
@denisrosset 👍 that sounds like a good general guideline.

Oh thanks @LukaJCB you beat me to it.

johnynek:
a non commutative semigroup that does not accumulate in memory (like String) does is:
`(a0, a1) + (b0, b1) = (a0, b1)`

This can be for any `(A, B)`.

ceedubs:
@johnynek on that note, wouldn't a simplified version be the "always first" semigroup? `a0 |+| a1 = a0`?

johnynek:
Indeed it would.

PR #2099, opened 2017-12-12
CommutativeMonad and CommutativeFlatMap instances for Tuple2

ceedubs:
I think that these are fairly straightforward/legitimate. Speak up if
that's not true :)


ceedubs:
It should be possible to do the same sort of thing for `Writer`, but `Writer` is a little different in that there are a couple of different `FlatMap` instances defined in a hierarchy. I don't have the time to look into it right at the moment, but someone else can feel free.

ceedubs:
I need to add some Mima exceptions. I'll do that now.

PR #2100, opened 2017-12-13
Add comparison method in Order companion object

ceedubs:
The main goal of this PR was to add syntax tests for the `Order` and
`PartialOrder` syntax (provided via machinist ops) to address [this comment](https://github.com/typelevel/cats/pull/1867#discussion_r138765045).

In the process I noticed that the Order companion object didn't have a
`comparison` function to forward through to the type class instance like
it did for other methods, so I added that.

I did notice that the `Order`/`PartialOrder` syntax does work whether or
not you take the `Order`/`PartialOrder` instances as implicit parameters
to the ops classes. I think that only @non understands quite what's
going on with the machinist syntax macros, so I didn't make the changes
to take away those implicit parameters without knowing the implications.
The tests pass either way, so I think that end-code should work for
users either way.


ceedubs:
Tests passed locally, but during the build one failed with `Message: NaN did not equal NaN` 🤦‍♂️.

I'll come up with a workaround.

PR #2101, opened 2017-12-13
Add Semigroup.instance method

jozic:


core/src/main/scala/cats/instances/invariant.scala
```diff
@@ -19,9 +18,7 @@ trait InvariantMonoidalInstances {
     def pure[A](a: A): Semigroup[A] = new Semigroup[A] {
       def combine(x: A, y: A): A = a
       override def combineAllOption(as: TraversableOnce[A]): Option[A] =
-        if (as.isEmpty) None
-        else if (as.size == 1) as.toList.headOption
-        else Some(a)
+        if (as.size > 1) Some(a) else as.toTraversable.headOption
```
ceedubs: There may be merit to these changes, but they seem to be unrelated to the rest of the PR, and I think that they should be separated out into another PR. I'm not sure if there are types that implement `TraversableOnce` for which an `isEmpty` check would be significantly faster than a `size` check on an empty collection, but it's something that people might not think about when reviewing this PR.

jozic: Np, I will revert these changes, they are indeed not related to `Semigroup.instance`

jozic: @kailuowang that's true, but unfortunately not all of us are in Java 8/Scala 2.12 lands yet

@ceedubs 
I can write jmh benchmarks for both cases - the old way of creating instances  and the new one using `Semigroup.instance` 
 OR
 alternatively I can revert those changes and just leave  addition of `Semigroup.instance` , as this is the main goal here. Then we can try  another PR with benchmarks or just let it go
Please let me know

ceedubs: > I can revert those changes and just leave addition of Semigroup.instance , as this is the main goal here. Then we can try another PR with benchmarks or just let it go

@jozic this sounds like the path of least resistance to me. Just please be sure to add a test for the `Semigroup.instance` method, since I think that currently it's only indirectly tested through the `Semigroup` instances that you are using it to create.

jozic:
@kailuowang @ceedubs 
thanks for your review! 
I just pushed updates to the PR as discussed

2 questions left: 
1. Should I send PR with changes to `catsInvariantMonoidalSemigroup.pure` I excluded from this PR as they were irrelevant? 
2. Should I benchmark creation of default Semigroup instances w/ and w/o usage of `Semigroup.instance` introduced in this PR?

johnynek:
👍

kailuowang:
@jozic thanks! IRT InvarianyMonoidal change, do you mind take a look at #2088?
I would say just switching back to detail creation method inside cats might be easier than benchmarking... This kind of trait instance creation is all over cats. People are so used to read it that there isn't much gain in readability in replacing them with the convenience method, IMHO.

ceedubs:
@kailuowang are you happy with this one? If so, feel free to merge.

PR #2102, opened 2017-12-13
Make instances of `Kleisli` interchangeable with functions

amitayh:
`Kleisli[F[_], A, B]` holds a function `A => F[B]`, and has an `apply` method, which means it can be used as a function:

```scala
val nonZero = Kleisli[Option, Int, Int](a => if (a != 0) Some(a) else None)

nonZero(0) // None
nonZero(1) // Some(1)
```

However, it can't be passed to another function that expects a function:

```scala
def runOne(f: Int => Option[Int]): Option[Int] = f(1)

runOne(nonZero) // Does not compile
```

By changing `Kleisli` to extend `A => F[B]`, the example above should work as expected

### DIFF
--- a/core/src/main/scala/cats/data/Kleisli.scala
+++ b/core/src/main/scala/cats/data/Kleisli.scala
@@ -7,7 +7,8 @@ import cats.arrow._
 /**
  * Represents a function `A => F[B]`.
  */
-final case class Kleisli[F[_], A, B](run: A => F[B]) { self =>
+final case class Kleisli[F[_], A, B](run: A => F[B])
+  extends (A => F[B]) { self =>
 
   def ap[C](f: Kleisli[F, A, B => C])(implicit F: Apply[F]): Kleisli[F, A, C] =
     Kleisli(a => F.ap(f.run(a))(run(a)))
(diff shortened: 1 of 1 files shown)


kailuowang:
Thanks for contributing. I am not sure about this change though. The `Kleisli` data structure is created as a new type of `Function1[A => F[B]]`, (if you are not familiar with the concept new type, it's sort of like like a type alias, but the system treats it as a different type). The purpose of that is so that we can define type class instances for this new type independent of the `Function1` type without conflict. If we have `Kleisli` extending `Function1`, that would beat the purpose.  Having to call `.run` when a `Function` is needed is a small cost we pay for that purpose.

ceedubs:
I tend to agree with @kailuowang's comment here. `Kleisli` has an `andThen` method that looks like `def andThen[C](f: B => F[C])(implicit F: FlatMap[F]): Kleisli[F, A, C]`. This would conflict with the `Function1.andThen` method that would have the signature `def andThen[C](f: F[B] => F[C])(implicit F: FlatMap[F]): Kleisli[F, A, C]`. There is a similar issue with `compose`. By distinguishing between functions like `A => B` and `A => F[B]`, we get the composition semantics that we want.

I also suspect that the contravariance and covariance used in `Function1` might cause some issues in `Kleisli`, but I'm not sure about that.

I think that calling `.run` to get the `Function1` instance is a small price to pay for these benefits.

amitayh:
Thanks for your feedback, closing the PR

ceedubs:
Thanks for the PR, @amitayh! If nothing else, this serves as some helpful documentation if someone else asks why `Kleisli` doesn't extend `Function1`.

PR #2106, opened 2017-12-14
Update to latest patch versions of scala

ceedubs:
These patch versions include a [security update](http://www.scala-lang.org/news/security-update-nov17.html).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,9 +10,9 @@ git:
   depth: 9999
 
 scala:
-  - 2.10.6
-  - 2.11.11
-  - 2.12.3
+  - 2.10.7
+  - 2.11.12
+  - 2.12.4
 
 jdk:
   - oraclejdk7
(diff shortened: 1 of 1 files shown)


johnynek:
missing the 2712 plugin on 2.10:
```
[error] (macrosJS/*:coursierResolutions) coursier.ResolutionException: Encountered 1 error(s) in dependency resolution:

[error]     com.milessabin:si2712fix-plugin_2.10.7:1.2.0:```

ceedubs:
Ah bummer. I'll try to get a PR in for the plugin.

milessabin:
PR merged, but unfortunately I won't be able to publish until this evening.

milessabin:
FWIW, I don't believe there's any reason to wait for 2.10.7 to make the release ... the security issue is with the compiler itself rather then binaries built with it, so releasing with 2.10.6 would be just fine.

milessabin:
Hmm ... and you'll need a build for 2.11.11 as well ... on it now ...

milessabin:
Plugin published for 2.10.7 ...

PR #2108, opened 2017-12-14
Update version of deprecation to 1.0.0-RC2

rossabaker:
This updates the deprecations we previously expected to debut in 1.0.0.


rossabaker:
[Discussion of a potential RC2](https://gitter.im/typelevel/cats-dev?at=5a32ba83232e79134d4d4531).  If that happens, we should get this in.  If that doesn't, we should Close Without Merge™.

LukaJCB:
Thanks for this!

PR #2110, opened 2017-12-15
add `Comparison` to `cats` package

### LINKED ISSUE #2109: cats.kernel.Comparison not exported by the cats package object
Contrast with cats.kernel.Order which is exported. 

Is this on purpose? This means that to pattern match on the result of `v1 comparison v2` one must import `cats.kernel.Comparison` in addition to importing `cats._`.

kailuowang:
fixes #2109

### DIFF
--- a/core/src/main/scala/cats/package.scala
+++ b/core/src/main/scala/cats/package.scala
@@ -81,6 +81,7 @@ package object cats {
 
   type Eq[A] = cats.kernel.Eq[A]
   type PartialOrder[A] = cats.kernel.PartialOrder[A]
+  type Comparison = cats.kernel.Comparison
   type Order[A] = cats.kernel.Order[A]
   type Hash[A] = cats.kernel.Hash[A]
   type Semigroup[A] = cats.kernel.Semigroup[A]
--- a/core/src/main/scala/cats/package.scala
+++ b/core/src/main/scala/cats/package.scala
@@ -90,6 +90,7 @@ package object cats {
   val Eq = cats.kernel.Eq
   val PartialOrder = cats.kernel.PartialOrder
   val Order = cats.kernel.Order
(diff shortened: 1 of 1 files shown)


iravid:
Thanks @kailuowang! Should the type be exported as well?

core/src/main/scala/cats/package.scala
```diff
@@ -90,6 +90,7 @@ package object cats {
   val Eq = cats.kernel.Eq
   val PartialOrder = cats.kernel.PartialOrder
   val Order = cats.kernel.Order
+  val Comparison = cats.kernel.Comparison
```
johnynek: do we not need the `type Comparison`?

kailuowang:
I don't see much of the benefit of aliasing `Comparison` type. Unlike existing aliases, it's not a type class. The original request in #2109 is to do pattern match on comparison without an extra import, the current change should suffice for that requirement right? 
```scala
a comparison b match {
   case Comparison.GreaterThan => ...
   ...
}
```
Of course if we want to pass the comparison result around we would need the type, but how often is that? I was just trying to do the minimum (and minimize namespace pollution) to satisfy the original request.

johnynek:
I think it is inconsistent to be able to do `def getEq: cats.kernel.Comparison = Comparison.Equal` why does one have the kernel, but not the other?

iravid:
For pattern matching accessing the data constructors is indeed sufficient. I don’t have a concrete usecase for passing around a comparison currently but it looks inconsistent with the rest of the kernel re-exports.

Namespace Pollution is an interesting consideration but I’m not familiar with the trade offs here. My desire is to stick with the cats, cats.data and cats.implicits imports :-)

kailuowang:
Consistency is a good argument. I added the type.

LukaJCB:
:+1:

PR #2114, opened 2017-12-17
sbt-coursier 1.0.0

sullis:


fthomas:
Thanks for the PR, @sullis. Could you also update the coursier version in `project/project/plugins.sbt`?

Issue #2093, opened 2017-12-11
Code review: cats-kernel


denisrosset:
Hi all,

as algebra, and by extension spire both depend on cats-kernel, I quickly went through the cats-kernel package. Please find my observations below.

There are a few small inconsistencies that could be corrected before freezing binary compatibility. I also think that, while typeclasses and data types are documented, the motivation for the default instances is sometimes missing.

I'm not asking to correct any of those; I just want to be sure that we commit to the (sometimes implicit) choices we are making for the future.

Denis

# Typeclasses

## Hierarchy

- Two tails in the hierarchy Eq->Hash and Eq->PartialOrder->Order, could lead to ambiguous implicits (#2012)

## Properties

- SemigroupFunctions.isCommutative/isIdempotent: document that a return value of true=property is valid, false=property may be not valid (as it uses internally isInstanceOf).

Alternative 1: rename as knownCommutative, or provenCommutative (as to understand that no does not mean violation of the property)

Suggestion: put the functions directly on the typeclass instance, and override def = true in the respective Band and CommutativeSemigroup traits

- maybeCombine are helper methods on the companion; in general, when should methods be part of the typeclass, when should they be part of the companion functions?

## Exhaustiveness

We do not have magmas (i.e. non associative semigroups), nor quasigroups/loops (which are in scalaz 8 roadmap).

See: types of magma, https://en.wikipedia.org/wiki/Magma_(algebra)

We distinguish bands and semilattices, which are important for algebird/for the meet&join lattices in algebra.

Other classes of semigroups:

https://en.wikipedia.org/wiki/Special_classes_of_semigroups

## Conversion methods

They would behave better on the companion object, which avoids syntax ambiguities when an instance inherits from several typeclasses.

- Semilattice.as[Meet|Join]PartialOrder

- Order.toOrdering, but no PartialOrder.toPartialOrdering

## Automatic conversion to/from Scala instances

- Do we want the implicit conversions (Order->Ordering, PartialOrder->PartialOrdering) to be always available? In Spire, a special import is needed for them to be available.

## Exhaustivity of methods

- Lack of exhaustivity: Ordering->Order conversion is available, but not PartialOrdering->PartialOrder (and no Hash->Hashing)

- No Eq.from and Hash.from, while PartialOrder.from and Order.from are available

# Instances

[...]

PR #2116, opened 2017-12-17
conversion `PartialOrder` to `PartialOrdering` and `Hash` to `Hashing`

kailuowang:
To address some items in #2093
Note that this code change is binary compatible in kernel  (I tested Mima without any of the exceptions I added) but not binary compatible in core.

### DIFF
--- a/kernel/src/main/scala/cats/kernel/PartialOrder.scala
+++ b/kernel/src/main/scala/cats/kernel/PartialOrder.scala
@@ -123,7 +123,7 @@ abstract class PartialOrderFunctions[P[T] <: PartialOrder[T]] extends EqFunction
     ev.gt(x, y)
 }
 
-object PartialOrder extends PartialOrderFunctions[PartialOrder] {
+object PartialOrder extends PartialOrderFunctions[PartialOrder] with PartialOrderToPartialOrderingConversion {
   /**
    * Access an implicit `PartialOrder[A]`.
    */
@@ -154,10 +154,10 @@ object PartialOrder extends PartialOrderFunctions[PartialOrder] {
       def partialCompare(x: A, y: A) = f(x, y)
     }
 
-  /**
-   * Implicitly convert a `PartialOrder[A]` to a
-   * `scala.math.PartialOrdering[A]` instance.
-   */
+}
+
+
+trait PartialOrderToPartialOrderingConversion {
   implicit def catsKernelPartialOrderingForPartialOrder[A](implicit ev: PartialOrder[A]): PartialOrdering[A] =
     new PartialOrdering[A] {
       def tryCompare(x: A, y: A): Option[Int] = ev.tryCompare(x, y)
--- a/kernel/src/main/scala/cats/kernel/PartialOrder.scala
+++ b/kernel/src/main/scala/cats/kernel/PartialOrder.scala
@@ -154,8 +154,9 @@ object PartialOrder extends PartialOrderFunctions[PartialOrder] with PartialOrde
       def partialCompare(x: A, y: A) = f(x, y)
     }
 
-  def fromPartialOrdering[A](implicit ev: PartialOrdering[A]): PartialOrder[A] = from[A] { (a, b) =>
-    ev.tryCompare(a, b).fold(Double.NaN)(_.toDouble)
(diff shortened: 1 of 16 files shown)


kailuowang:
@denisrosset , will you take a look as well?

denisrosset:
Why do both the core instances and the `PartialOrder` companion object extends the trait containing implicit conversion?

kernel/src/main/scala/cats/kernel/PartialOrder.scala
```diff
@@ -154,10 +154,13 @@ object PartialOrder extends PartialOrderFunctions[PartialOrder] {
...
     }
 
-  /**
-   * Implicitly convert a `PartialOrder[A]` to a
-   * `scala.math.PartialOrdering[A]` instance.
-   */
+  def fromPartialOrdering[A](implicit ev: PartialOrdering[A]): PartialOrder[A] = from[A] { (a, b) =>
+    ev.tryCompare(a, b).fold(Double.NaN)(_.toDouble)
+  }
+}
```
denisrosset: I'd prefer extending the `PartialOrder` trait here instead of having so many lambdas (`from` and the `fold`).

kailuowang:
@denisrosset the core instances extends it so that it's available through `import cats.implicits._`, the  `PartialOrder` companion object extends it to maintain compatibility.

denisrosset:
So, if I understand correctly, the conversion will always be available regardless of the `import cats.implicits._` directive?

kailuowang:
no, it will only be available through `import cats.implicits._`. The one in the companion object is of little use.

denisrosset:
> Update: To be clear, the original way it was added was basically wrong - it's added as an implicit but in a companion that won't get searched.

I suggest to document this, implicits are confusing enough. If you don't want to modify this PR, I'll add a comment on the Google doc and will have a pass a it later.

denisrosset:
A couple of comments.

- `Eq` does not have a conversion to `Equiv`.
- `Hash` has an implicit conversion to `Hashing`, but a contradictory comment in the `Hash` typeclass itself.
- No `toHashing`, `toPartialOrdering`, `toEquiv` methods.
- The implicit conversion for `Order` is on `instances.order`, but the one for `PartialOrder` on `instances.partialOrdering`.

This is corrected by the commit mentioned above.

https://github.com/denisrosset/cats/commit/2b0a67fc65aab1860fe396b5f043403be9603d05

LukaJCB:
Awesome @denisrosset! Would you like to PR that to Kai's branch? :)

kailuowang:
@denisrosset I will address your feedbacks. 
In regards to the  `toHashing`, `toPartialOrdering`, `toEquiv` methods though, they will break binary compat for those type classes. If it's okay to break `kernel.implicits` I can add the implicits conversion traits there so that kernel users can have them through `import cats.kernel.implicits._`.  WDYT?

LukaJCB:
As far as I can tell from the source, neither Algebra, nor Spire or Algebird make use of `kernel.implicits`.

kailuowang:
>  `Hash#toHashing` deliberately not implemented since `scala.util.hashing.Hashing` is only		 
compatible with universal equality.

This is probably a good reason for not providing any `Hash` to `Hashing` conversion, no?

kailuowang:
@denisrosset I've added the conversions to `cats.kernel.instances.all` so that they can be available for kernel users through reports. I removed the `Hash` -> `Hashing` conversion due to reason commented above.

denisrosset:
> This is probably a good reason for not providing any Hash to Hashing conversion, no?

Except that the documentation for `Hashing` in the 2.12 api is:

> Note: when using a custom Hashing, make sure to use it with the Equiv such that if any two objects are equal, then their hash codes must be equal.

kailuowang:
@denisrosset is it the same way in 2.11 and 2.10?

denisrosset:
See https://static.javadoc.io/org.scala-lang/scala-library/2.10.0/scala/util/hashing/Hashing.html

kailuowang:
@denisrosset that seems right. I couldn't find from the original [PR](https://github.com/typelevel/cats/pull/1712/) why that comment was added. 
Added the Hashing conversion back.

PR #2117, opened 2017-12-18
fix scalafix rules for productL/R

kailuowang:


peterneyens:
Should we rename the laws/tests as well?

- [`ApplyLaws`](https://github.com/typelevel/cats/blob/master/laws/src/main/scala/cats/laws/ApplyLaws.scala#L24-L28)
- [`ApplyTests`](https://github.com/typelevel/cats/blob/master/laws/src/main/scala/cats/laws/discipline/ApplyTests.scala#L33-L34)

kailuowang:
@peterneyens I will submit a separate PR for law tests rename.

PR #2119, opened 2017-12-18
Fixes type params for OptionT.getOrElse(F)

fcanedo:
Adds a type parameter to getOrElse and getOrElseF to conform to
Option.getOrElse, which allows it to be called without needing a type
hint:

OptionT(Future {
  Option(Right(1))
}).getOrElse(Left("error"))

Instead of:

OptionT(Future {
  Option(Right(1): Either[String, Int])
}).getOrElse(Left("error"))

### DIFF
--- a/core/src/main/scala/cats/data/OptionT.scala
+++ b/core/src/main/scala/cats/data/OptionT.scala
@@ -51,10 +51,10 @@ final case class OptionT[F[_], A](value: F[Option[A]]) {
   def subflatMap[B](f: A => Option[B])(implicit F: Functor[F]): OptionT[F, B] =
     transform(_.flatMap(f))
 
-  def getOrElse(default: => A)(implicit F: Functor[F]): F[A] =
+  def getOrElse[B >: A](default: => B)(implicit F: Functor[F]): F[B] =
     F.map(value)(_.getOrElse(default))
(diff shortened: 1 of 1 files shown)


ceedubs:
It might be nice to add a test for this, since it's functionality that we expect to work.

kailuowang:
Thanks! A doctest would be nice.

fcanedo:
I added some unit tests, but what’s a “doctest”?

kailuowang:
Doctest is really cool! You write code and expected result in Scaladoc, and it runs as a test. You can find plenty of examples here
https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Apply.scala

fcanedo:
I can’t reproduce the errors that CI finds locally; any tips?

kailuowang:
It only happens on scala 2.10 and 2.11. In sbt you can type `++2.11.11` to test in scala 2.11.  You might need to use `"error".asLeft[Int]` to help scala compiler prior to 2.12.

fcanedo:
@kailuowang, your fix seems to solve the problem. Also, I confirmed that the tests fail to compile if I revert the fix. I think this is done.

What needs to happen to get this merged?

PR #2122, opened 2017-12-20
Add doctest example for ApplicativeError.raiseError

ceedubs:


core/src/main/scala/cats/ApplicativeError.scala
```diff
@@ -10,8 +10,27 @@ import scala.util.control.NonFatal
...
   /**
    * Lift an error into the `F` context.
+   *
+   * Example:
+   * {{{
+   * scala> import cats.implicits._
+   *
+   * // integer-rounded division
+   * scala> def divide[F[_]](dividend: Int, divisor: Int)(implicit F: ApplicativeError[F, String]): F[Int] =
+   *      | if (divisor === 0) F.raiseError(s"division by zero: $dividend / 0")
```
kailuowang: > /home/travis/build/typelevel/cats/core/src/main/scala/cats/ApplicativeError.scala:23: Variable dividend undefined in comment for method raiseError in trait ApplicativeError

[error]    *      | if (divisor === 0) F.raiseError(s"division by zero: $dividend / 0")

Scaladocs seems to be parsing `$XXXX` token, the easiest thing to do is probably to let it win and remove it.

ceedubs: Oh right I've dealt with this before. There's a trick to fix it, but the string interpolation isn't adding much to the example anyway so I'll just remove it.

Issue #2113, opened 2017-12-16
Add contains, collectFirst and mkString to NonEmptyList


rsoeldner:
I would like to add these functionality to NonEmptyList, what do you think ?

```scala
 def contains[AA >: A](a: AA): Boolean = exists(_ == a)

  def collectFirst[B](pf: PartialFunction[A, B]): Option[B] =
    pf.lift(head).orElse(tail.collectFirst(pf))

  def mkString(s: String): String = mkString("", s, "")

  def mkString(begin: String, s: String, end: String): String = {
    val builder = new StringBuilder()
    builder.append(begin)
    builder.append(head)
    tail.foreach(elem => builder.append(s"$s$elem"))
    builder.append(end)
    builder.toString()
  }
```

I'm quite unsure about my mkString :smile:

johnynek:
Contains would need an Eq and mkString should take a Show.

rossabaker:
`collectFirst` is already on `Foldable`.  `mkString` isn't, but some use cases are handled by `intercalate`.

I've needed `contains` and `mkString` on `NonEmptyList` before.  My instinct is to provide these for all `Foldable`, but our `Eq`- and `Show`-based versions would have a name clash with the standard library's foldables.

kailuowang:
I would prefer to add contains and mkstring to Foldable too, but we already passed the deadline for non bug bin compat breaking PRs...

rossabaker:
Yeah, I failed to think about that. 😞  

To speak to the usefulness of the request, I have a [syntax extension with the same three methods](https://github.com/http4s/http4s/blob/a3a74eee11152826e7f579615bd459f29dc37b4f/core/src/main/scala/org/http4s/syntax/NonEmptyListSyntax.scala), and feel like a jerk for not creating a PR in time.

tpolecat:
`mkString` can be generalized to any monoid in the same way `intercalate` is. I call it `foldSmash` in doobie and would be pleased to see it in cats with a better name.

johnynek:
@tpolecat actually Semigroups can be generalized in this way: if a+b is monoid, then a+z+b is a semigroup for all z. That’s one way to think about this: you are transforming the Monoid and then doing a combine.

rsoeldner:
I ended up with 
```scala
  def mkString[A](fa: F[A], prefix: String, delim: String, suffix: String)(implicit G: Functor[F], A: Show[A], M: Monoid[String]): String =
[...]

PR #2123, opened 2017-12-20
Add `contains_`, `foldSmash` and `mkString_` to `FoldableOps`

rsoeldner:
#2113, Am I right, test should be added to the `FoldableSuite` ?

### DIFF
--- a/core/src/main/scala/cats/Foldable.scala
+++ b/core/src/main/scala/cats/Foldable.scala
@@ -578,6 +578,15 @@ import Foldable.sentinel
 
   override def unorderedFoldMap[A, B: CommutativeMonoid](fa: F[A])(f: (A) => B): B =
     foldMap(fa)(f)
+
+  def contains[A](fa: F[A])(v: A)(implicit ev: Eq[A]): Boolean =
+    exists(fa)(ev.eqv(_, v))
+
+  def foldSmash[A](fa: F[A])(prefix: A, delim: A, suffix: A)(implicit M: Monoid[A]): A =
+    M.combine(prefix, M.combine(intercalate(fa, delim), suffix))
+
+  def mkString[A: Monoid](fa: F[A])(prefix: String, delim: String, suffix: String)(implicit F: Functor[F], S: Show[A], ev1: Monoid[String]): String =
+    foldSmash(F.map(fa)(S.show))(prefix, delim, suffix)
 }
 
 object Foldable {
--- a/core/src/main/scala/cats/Foldable.scala
+++ b/core/src/main/scala/cats/Foldable.scala
@@ -578,15 +578,6 @@ import Foldable.sentinel
 
   override def unorderedFoldMap[A, B: CommutativeMonoid](fa: F[A])(f: (A) => B): B =
     foldMap(fa)(f)
-
-  def contains[A](fa: F[A])(v: A)(implicit ev: Eq[A]): Boolean =
-    exists(fa)(ev.eqv(_, v))
-
-  def foldSmash[A](fa: F[A])(prefix: A, delim: A, suffix: A)(implicit M: Monoid[A]): A =
-    M.combine(prefix, M.combine(intercalate(fa, delim), suffix))
-
-  def mkString[A: Monoid](fa: F[A])(prefix: String, delim: String, suffix: String)(implicit F: Functor[F], S: Show[A], ev1: Monoid[String]): String =
-    foldSmash(F.map(fa)(S.show))(prefix, delim, suffix)
 }
 
 object Foldable {
--- a/core/src/main/scala/cats/syntax/foldable.scala
+++ b/core/src/main/scala/cats/syntax/foldable.scala
@@ -33,7 +33,7 @@ final class FoldableOps[F[_], A](val fa: F[A]) extends AnyVal {
 
   def foldr[B](b: Eval[B])(f: (A, Eval[B]) => Eval[B])(implicit F: Foldable[F]): Eval[B] =
     F.foldRight(fa, b)(f)
-  
+
    /**
    * test if `F[A]` contains an `A`, named contains_ to avoid conflict with existing contains which uses universal equality
    *
@@ -48,10 +48,9 @@ final class FoldableOps[F[_], A](val fa: F[A]) extends AnyVal {
    * res1: Boolean = false
    * }}}
    */
-  def contains_[A](v: A)(implicit ev: Eq[A], F: Foldable[F]): Boolean =
+  def contains_(v: A)(implicit ev: Eq[A], F: Foldable[F]): Boolean =
     F.exists(fa)(ev.eqv(_, v))
 
-  
    /**
    * Intercalate with a prefix and a suffix
    *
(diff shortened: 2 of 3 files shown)


core/src/main/scala/cats/Foldable.scala
```diff
@@ -578,6 +578,15 @@ import Foldable.sentinel
...
   override def unorderedFoldMap[A, B: CommutativeMonoid](fa: F[A])(f: (A) => B): B =
     foldMap(fa)(f)
+
+  def contains[A](fa: F[A])(v: A)(implicit ev: Eq[A]): Boolean =
+    exists(fa)(ev.eqv(_, v))
+
+  def foldSmash[A](fa: F[A])(prefix: A, delim: A, suffix: A)(implicit M: Monoid[A]): A =
+    M.combine(prefix, M.combine(intercalate(fa, delim), suffix))
+
+  def mkString[A: Monoid](fa: F[A])(prefix: String, delim: String, suffix: String)(implicit F: Functor[F], S: Show[A], ev1: Monoid[String]): String =
```
chuwy: Hey @rsoeldner,

Actually, for `mkString` I believe we can also generalize over `String`, so `foldSmash` would work with `String`. Something like:

```scala
def mkString[A: Monoid, B: Show](fa: F[A])(prefix: B, delim: B, suffix: B)(implicit F: Functor[F], S: Show[A]): String =
  foldSmash(F.map(fa)(S.show))(prefix.show, delim.show, suffix.show)
```

Maybe this is bit a bike-shedding, but probably can help if you want to have unusual `Show[B]` instance in scope.

rsoeldner: Right, I thought about this, but wasn't able to find a use-case. Do you have an Idea ?

chuwy: I guess the main question here is "why I might want to pass different `B`s into `mkString`". I think it has use-case for [pretty printers](https://github.com/typelevel/paiges) to output some structure differently based on environment:

```
structure.mkString(LeftPad(0), LeftPad(5), LeftPad(0))
```

Maybe this is a completely made-up example, but why to constrain it?

rsoeldner: I see your point, should we support different types for _prefix_, _suffix_ and _delim_ ?

chuwy: Don't know, for me it looks like a skew into "over-generalized" approach, but let's see what community says - maybe even suggestion with `Show[B]` is unnecessary.

rsoeldner: :+1:

kailuowang:
Thanks @rsoeldner will you be interested in taking [this approach](https://github.com/typelevel/cats/issues/2142) to avoid breaking binary compatibility?

rsoeldner:
@kailuowang Yes, I'll try come up with a PR tomorrow ;)

core/src/main/scala/cats/Foldable.scala
johnynek: we shouldn't need `Functor[F]`. You can do this 
```scala
def mkString[A:Show, B: Show](fa: F[A])(prefix: B, delim: B, suffix: B): String =
  foldLeft(fa, prefix.show) { (left, a) =>
    left + (delim.show) + a.show
  } + (suffix.show)
```
of course, that is a O(N^2) implementation, but it could be fixed using a `StringBuilder`:
```scala
def mkString[A:Show, B: Show](fa: F[A])(prefix: B, delim: B, suffix: B): String = {
  val bldr = new java.lang.StringBuilder
  bldr.append(prefix.show)
  val b2 = foldLeft(fa, bldr) { (b, a) =>
    b.append((delim.show) + a.show)
  }
  b2.append(suffix.show)
  b2.toString
```

johnynek: that said, I think `B` is overgeneralized. Users should just do `prefix.show` etc... and keep it to be `String`.

rsoeldner: After some time, I think you are right. :+1:

rsoeldner:
Is it possible to reformat the entire project ? I used IJ's format functionality and now everything looks different :sob:

kailuowang:
@rsoeldner sorry I wasn't sure if that's a question. This would be a great addition. Now #2148 provides a good example on how to add functions without breaking bin compat.

kailuowang:
@rsoeldner now I think about it, maybe we should just move the implementations into `FoldableOps`. 
Since we can't let user override them in instances (due to binary compat, we can't add these methods to the trait), we probably shouldn't make them appear to be available from the type class itself anyway. 
Let me know if you are still interested/available in continue on this, I'll be more than happy to continue it for you.

core/src/main/scala/cats/syntax/foldable.scala
```diff
@@ -33,4 +33,54 @@ final class FoldableOps[F[_], A](val fa: F[A]) extends AnyVal {
...
+   * Example:
+   * {{{
+   * scala> import cats.implicits._
+   *
+   * scala> val l: List[Int] = List(1, 2, 3)
+   * scala> l.mkString_("List(", ",", ")")
+   * res0: String = List(1,2,3)
+   * }}}
+   */
+  def mkString_[A: Monoid](prefix: String, delim: String, suffix: String)(implicit F: Functor[F], S: Show[A], ev1: Monoid[String], ev2: Foldable[F]): String =
```
johnynek: Can we not take Functor on this? If we implemented directly from foldLeft we could. Shame to take an additional constraint just to make implementation easier.

kailuowang:
> Elsewhere in cats, namely Traverse, we use the _ to indicate we want the contained type to be Unit

I am aware of that as well. However I also noticed that there are already `filter_`, `takeWhile_`, `dropWhile_` in `Foldable` , so at least in `Foldable` there is already this mini convention of using `_` to avoid conflict. I don't mind renaming them, however `mkShowString`, `containsEq` sounds a bit awkward. How about `makeString` or `mkShow` and `has` ? The downside is that they are harder to discover.

kailuowang:
@johnynek WDYT about my comment above?

rsoeldner:
@kailuowang had a small break :+1: thanks for pushing this forward

core/src/main/scala/cats/syntax/foldable.scala
```diff
@@ -33,4 +33,58 @@ final class FoldableOps[F[_], A](val fa: F[A]) extends AnyVal {
...
+   * scala> l.mkString_("L[", ";", "]")
+   * res0: String = L[1;2;3]
+   * }}}
+   */
+  def mkString_(prefix: String, delim: String, suffix: String)(implicit A: Show[A], F: Foldable[F]): String = {
+    val b = F.foldLeft(fa, new StringBuilder){ (builder, a) =>
+      builder append A.show(a) append delim
+    }
+    if (b.isEmpty)
+      ""
```
johnynek: don't we need the prefix and suffix here? Can we have a test that exercises this against scala's mkString in the same case?

kailuowang: oops. yeah, I meant to change that but slipped. will update now.

tests/src/test/scala/cats/tests/FoldableSuite.scala
```diff
@@ -140,6 +140,12 @@ abstract class FoldableSuite[F[_]: Foldable](name: String)(
       fa.toList should === (iterator(fa).toList)
     }
   }
+  
+  test(s"Foldable[$name] mkString_") {
+    forAll { (fa: F[Int]) =>
+      fa.mkString_("L[", ";", "]") should === fa.mkString("L[", ";", "]")
```
johnynek: now is the right hand side working? don't we need a toList or something?

kailuowang: good catch. Since this is not my repo I can only edit files through github and my mental compiler is clearly not good enough.

PR #2124, opened 2017-12-21
Fixed a typo in the ScalaDoc for the combine method (Semigroup)

raffael-dzikowski:
Fixed a typo in the ScalaDoc for the `combine` method of `Semigroup`.


LukaJCB:
Good catch, thanks!

PR #2128, opened 2017-12-25
Add parTupled methods

iravid:
Pretty sure this isn't binary compatible, but opening anyway to see what MiMa says.


johnynek:
Shame some of these enrichments aren’t abstract classes so adding methods like this is safe.

kailuowang:
We can generate a new trait with this method and add it to cats.implicits and cats.syntax.all in a bin compat way. Interested?

iravid:
Sure! I’ll look into that next week.

--
Itamar

kailuowang:
@iravid still have time/insterest to continue on this?

iravid:
Very interested, but I don't have the time unfortunately to pick this up in the next week or two. Would be great if someone else can continue :-)

barambani:
if no one is interested in this at the moment I would give it a try

LukaJCB:
Go for it @barambani!

iravid:
Superseded by #2183

PR #2134, opened 2018-01-01
Minor scaladoc fix for :::

jcranky:


LukaJCB:
Good catch, thanks Paulo! :)

PR #2135, opened 2018-01-02
Remove duplicate code

akilegaspi:
I saw a duplicate entry on the website namely at [FAQ](https://typelevel.org/cats/faq.html) and saw that it was also duplicated in the repository.


LukaJCB:
Hi @akilegaspi, thanks for contributing! Unfortunately, if you look closely, the two symbols aren't actually the same, one of them is written in unicode:

`≺` vs `<`.

PR #2137, opened 2018-01-04
Remove trailing comma in Scalafix rewrites as a workaround for #2136

### LINKED ISSUE #2136: Trailing comma in rewrites currently breaks Scalafix
When loading the scalafix rewrites with

```
scalafix github:typelevel/cats/v1.0.0
```

they get downloaded, parsed and classloaded by Scalafix.
Unfortunately Scalafix uses the Scala 2.11 dialect, so the rewrites currently don't work due to a parser error.
  
Reported on gitter https://gitter.im/typelevel/cats?at=5a4e26b929ec6ac311b9f025

gabro:
Fixes #2136   

This is a workaround until we fix the problem upstream in Scalafix. See https://github.com/scalacenter/scalafix/issues/543
 
Tested locally with

```
scalafix github:gabro/cats/v1.0.0?sha=trailing-comma-fix
```


vpavkin:
@gabro Thanks for quick response and fix! :)

PR #2138, opened 2018-01-04
To either extensions

frosforever:
Enables writing code like:
```scala
Try(...).toEither

Option(...).toEither(...)
```
instead of
```scala
Either.fromTry(Try(...))

Either.fromOption(Some(...), ...)
```

I find this a bit easier to read and more natural to write.

### DIFF
--- a/core/src/main/scala/cats/syntax/TrySyntax.scala
+++ b/core/src/main/scala/cats/syntax/TrySyntax.scala
@@ -1,0 +1,29 @@
+package cats.syntax
+
+import cats.Applicative
+import cats.data.EitherT
+
+import scala.util.{ Failure, Success, Try }
+
+trait TrySyntax {
+  implicit final def catsSyntaxTry[A](t: Try[A]): TryOps[A] = new TryOps[A](t)
+}
+
+final class TryOps[A](val t: Try[A]) extends AnyVal {
+
+  /**
+    * Converts a `Try[A]` to a `Either[Throwable, A]`.
+    *
+    * @see [[cats.syntax.EitherObjectOps#fromTry]]
+    */
+  def toEither: Either[Throwable, A] = t match {
+    case Failure(e) => Left(e)
+    case Success(v) => Right(v)
+  }
+
+  /**
+    * Converts a `Try[A]` to a `EitherT[F, Throwable, A]`.
+    */
+  def toEitherT[F[_]](implicit F: Applicative[F]): EitherT[F, Throwable, A] =
+    EitherT.fromEither[F](toEither)
+}
--- a/core/src/main/scala/cats/syntax/package.scala
+++ b/core/src/main/scala/cats/syntax/package.scala
@@ -1,7 +1,7 @@
 package cats
 
 package object syntax {
-  object all extends AllSyntax
+  object all extends AllSyntax with TrySyntax
   object alternative extends AlternativeSyntax
   object applicative extends ApplicativeSyntax
   object applicativeError extends ApplicativeErrorSyntax
@@ -45,6 +45,7 @@ package object syntax {
   object show extends ShowSyntax
   object strong extends StrongSyntax
   object traverse extends TraverseSyntax
+  object trySyntax extends TrySyntax
   object nonEmptyTraverse extends NonEmptyTraverseSyntax
   object validated extends ValidatedSyntax
   object vector extends VectorSyntax
--- a/core/src/main/scala/cats/implicits.scala
+++ b/core/src/main/scala/cats/implicits.scala
@@ -1,3 +1,3 @@
 package cats
 
-object implicits extends syntax.AllSyntax with instances.AllInstances
+object implicits extends syntax.AllSyntax with instances.AllInstances with syntax.TrySyntax


core/src/main/scala/cats/syntax/TrySyntax.scala
```diff
@@ -0,0 +1,17 @@
+package cats.syntax
+
+import scala.util.Try
+
+trait TrySyntax {
+  implicit final def catsSyntaxTry[A](t: Try[A]): TryOps[A] = new TryOps[A](t)
+}
+
+final class TryOps[A](val t: Try[A]) {
```
johnynek: shouldn't this `extends AnyVal`

core/src/main/scala/cats/syntax/TrySyntax.scala
```diff
@@ -0,0 +1,17 @@
...
+
+  /**
+    * Converts a `Try[A]` to a `Either[Throwable, A]`.
+    *
+    * @see [[cats.syntax.EitherObjectOps#fromTry]]
+    */
+  def toEither: Either[Throwable, A] = new EitherObjectOps(Either).fromTry(t)
```
johnynek: can we just write:
```scala
t match {
  case Success(a) => Right(a)
  case Failure(ex) => Left(ex)
}
```
since this is so simple?

core/src/main/scala/cats/syntax/option.scala
```diff
@@ -27,6 +27,14 @@ final class OptionIdOps[A](val a: A) extends AnyVal {
 }
 
 final class OptionOps[A](val oa: Option[A]) extends AnyVal {
+  /**
+    * If the `Option` is `Some` return its value in a [[scala.util.Right]].
+    * If the `Option` is `None`, return the provided `B` value in a [[scala.util.Left]].
+    *
+    * @see [[cats.syntax.EitherObjectOps#fromOption]]
+    */
+  def toEither[B](b: => B): Either[B, A] = new EitherObjectOps(Either).fromOption(oa, b)
```
johnynek: similarly can we just do:
```scala
oa match {
  case Some(a) => Right(a)
  case None => Left(b)
}
```

core/src/main/scala/cats/syntax/all.scala
```diff
@@ -43,6 +43,7 @@ trait AllSyntax
     with SemigroupKSyntax
     with ShowSyntax
     with StrongSyntax
+    with TrySyntax
```
kailuowang: Unfortunately this isn't binary compatible scala 2.11-. 
We can still include these syntax, but this TrySyntax won't be imported using `cats.implicits._`, a hackish work around would be adding it to `EitherSyntax` instead.

frosforever: @kailuowang I don't believe that's possible as `try` is a reserved keyword. I've instead added `trySyntax` here https://github.com/typelevel/cats/pull/2138/files#diff-cdc99a8a4f35341b8a5aa3a52482e06fR48

frosforever:
@johnynek & @kailuowang Thanks for taking the time to review!

core/src/main/scala/cats/syntax/option.scala
```diff
@@ -27,6 +27,17 @@ final class OptionIdOps[A](val a: A) extends AnyVal {
 }
 
 final class OptionOps[A](val oa: Option[A]) extends AnyVal {
+  /**
+    * If the `Option` is `Some` return its value in a [[scala.util.Right]].
+    * If the `Option` is `None`, return the provided `B` value in a [[scala.util.Left]].
```
kailuowang: > r] /home/travis/build/typelevel/cats/core/src/main/scala/cats/syntax/option.scala:30: Could not find any member to link for "scala.util.Right".
[error]   /**
[error]   ^
[error] one error found
[error] (docs/scalaunidoc:doc) Scaladoc generation failed


linking to external types fails, I would just use `scala.util.Right/Left` instead.

core/src/main/scala/cats/syntax/package.scala
```diff
@@ -45,6 +45,7 @@ package object syntax {
   object show extends ShowSyntax
   object strong extends StrongSyntax
   object traverse extends TraverseSyntax
+  object trySyntax extends TrySyntax
```
kailuowang: sorry I missed this last time, there is an `all` object at the top of this list, will you make it inherit `TrySyntax` as well?

core/src/main/scala/cats/syntax/option.scala
```diff
@@ -27,6 +27,17 @@ final class OptionIdOps[A](val a: A) extends AnyVal {
 }
 
 final class OptionOps[A](val oa: Option[A]) extends AnyVal {
+  /**
+    * If the `Option` is `Some` return its value in a `scala.util.Right`.
+    * If the `Option` is `None`, return the provided `B` value in a `scala.util.Left`.
+    *
+    * @see [[cats.syntax.EitherObjectOps#fromOption]]
+    */
+  def toEither[B](ifNone: => B): Either[B, A] = oa match {
```
peterneyens: What is the benefit over `.toRight` provided by the standard library?

core/src/main/scala/cats/syntax/TrySyntax.scala
```diff
@@ -0,0 +1,29 @@
...
+}
+
+final class TryOps[A](val t: Try[A]) extends AnyVal {
+
+  /**
+    * Converts a `Try[A]` to a `Either[Throwable, A]`.
+    *
+    * @see [[cats.syntax.EitherObjectOps#fromTry]]
+    */
+  def toEither: Either[Throwable, A] = t match {
```
peterneyens: Just mentioning that this syntax is already provided by [mouse](https://github.com/typelevel/mouse/blob/master/shared/src/main/scala/mouse/try.scala#L17).
There is no real clear line though for what belongs in Cats itself or what in mouse.

core/src/main/scala/cats/syntax/option.scala
frosforever: Doh! what a fantastic point. I can't believe I forgot about that. Thanks!

core/src/main/scala/cats/syntax/TrySyntax.scala
frosforever: good point. I'm happy to close this without merging. I can add `toEitherT` to mouse instead.

frosforever:
Better served in https://github.com/typelevel/mouse/blob/f248d85e43fae304b0d634a095ac2ae9f4ca2ddd/shared/src/main/scala/mouse/try.scala

Issue #27, opened 2015-02-01
Typeclass instance usage convention


ceedubs:
The two implementations are effectively identical:

```scala
implicit def OrShow1[A: Show, B: Show]: Show[A Or B] =
  Show.show[A Or B](_.fold(a => s"LOr(${Show[A].show(a)})", b => s"ROr(${Show[B].show(b)})"))

implicit def OrShow2[A, B](implicit showA: Show[A], showB: Show[B]): Show[A Or B] =
  Show.show[A Or B](_.fold(a => s"LOr(${showA.show(a)})", b => s"ROr(${showB.show(b)})"))
```

In order to make it easier for newcomers to contribute, I think cats should establish and document a convention for which of the two approaches to use.

Here is a proposed convention:
If your function needs to call a method on the type class instance, it should use the latter approach. This prevents the unnecessary calls to `Show.apply`.
If your function just needs the evidence of the type class instances so it can call through to another function that requires them, it should use the former aproach, as it is a bit cleaner.

There are also questions about whether you should use names like `showA: Show[A]` or `ShowA: Show[A]` or `A: Show[A]`, etc. I don't really have a preference on convention here, but I think the more of this we can put into a contributor guide the less people have to worry about not knowing the right way to do it or submitting a pull request that gets nitpicked.

Thoughts?

non:
I agree with pretty much all of this: I think we should prefer `[A: Foo]` if the `Foo[A]` instance is just being routed somewhere else, and `(implicit xyz: Foo[A])` if method's on the `xyz` instance are needed (since even with `@inline` the `Foo.apply` call has overhead).

As far as the names of the evidence parameter, I know @tixxit prefers `(implicit A: Foo[A])`. How do folks feel about that? I often use `ev` but on reflection I don't think there's really a good reason -- it seems a bit cranky. I guess I don't know `showA` but unless there are multiple instances it seems verbose.

non:
(And yes, we do need a contributor guide that spells this out.)

fthomas:
I always liked the idea to view the `A` in `(implicit A: Foo[A])` as the companion for the type parameter. But this is only applicable as long as you only have one instance per type parameter. In the case of two or more instances I guess I'd prefer `(implicit fooA: Foo[A], barA: Bar[A])`.

julien-truffaut:
shall we close this ticket?

non:
Yes, I think we have reached broad agreement.

PR #2141, opened 2018-01-08
Add NonEmptyMap

LukaJCB:
This is basically just a copy-paste from https://github.com/LukaJCB/cats-maps-sets, I still need to add some more scaladoc and maybe a test here or there, but if someone has some early feedback I'd love to hear!


jtjeferreira:
should cats.data.NonEmptyList#groupBy return an NonEmptyMap?

LukaJCB:
@jtjeferreira that's an excellent idea in my opinion, however it might not be feasible to change it though due to backwards compatibility.

jtjeferreira:
maybe add a new method `groupByNem`

core/src/main/scala/cats/data/NonEmptyMap.scala
```diff
@@ -0,0 +1,225 @@
...
+package cats
+package data
+
+import cats.instances.sortedMap._
+import cats.kernel._
+import cats.{Always, Apply, Eval, Foldable, Functor, Later, NonEmptyTraverse, Now, SemigroupK, Show}
+
+import scala.collection.immutable._
+
+final class NonEmptyMap[K, A] private(val value: SortedMap[K, A]) extends AnyVal {
```
jozic: does it have to be a `SortedMap`?

LukaJCB: What would you want it to be?

jozic: just `immutable.Map[K,A]`?
or maybe a better name for the class then `NonEmptySortedMap`?

kailuowang:
I will find some time Friday.

LukaJCB:
Excellent, thank you @kailuowang! No rush at all :)

ceedubs:
@LukaJCB @kailuowang sorry, I'm unlikely to find time for any reviews until Feb 4th or so.

LukaJCB:
No worries at all @ceedubs! No need to rush :)

core/src/main/scala/cats/data/NonEmptyMap.scala
```diff
@@ -0,0 +1,343 @@
...
+    * Returns a `SortedSet` containing all the keys of this map.
+    * {{{
+    * scala> import cats.data.NonEmptyMap
+    * scala> import cats.implicits._
+    * scala> val nem = NonEmptyMap.of(1 -> "A", 2 -> "B")
+    * scala> nem.keys
+    * res0: scala.collection.immutable.SortedSet[Int] = Set(1, 2)
+    * }}}
+    */
+  def keys: SortedSet[K] = value.keySet
```
LukaJCB: This should be a non empty set

core/src/main/scala/cats/data/NonEmptyMap.scala
```diff
@@ -0,0 +1,343 @@
+/*
+ * Copyright (c) 2018 Luka Jacobowitz
```
kailuowang: Shall we add a copyright header to all files in a separate PR (I am thinking maybe we can use sbt-header)

LukaJCB: Sounds like a good idea!

kailuowang: Shall we use copyright maintainers like the website or each file copyright goes to individual creators?

LukaJCB: Copyright maintainers seems best to me, I just copy and pasted this one from my own project :smile:

core/src/main/scala/cats/data/NonEmptyMap.scala
```diff
@@ -23,246 +23,261 @@ import cats.{Always, Apply, Eval, Foldable, Functor, Later, NonEmptyTraverse, No
...
-
-  /**
-    * Removes a key from this map, returning a new SortedMap.
-    */
-  def -(key: K): SortedMap[K, A] = value - key
-
-  /**
-    * Adds a key-value pair to this map, returning a new `NonEmptyMap`.
-    * */
-  def +(ka: (K, A)): NonEmptyMap[K, A] = new NonEmptyMap(value + ka)
```
LukaJCB: I had to get rid of this operator because of the super annoying `any2StringAdd` :(

core/src/main/scala/cats/data/NonEmptyMapImpl.scala
```diff
@@ -317,7 +322,7 @@ private[data] sealed abstract class NonEmptyMapInstances {
   implicit def catsDataShowForNonEmptyMap[K: Show, A: Show]: Show[NonEmptyMap[K, A]] =
     Show.show[NonEmptyMap[K, A]](_.show)
 
-  implicit def catsDataBandForNonEmptyMap[K, A]: Band[NonEmptyMap[K, A]] = new Band[NonEmptyMap[K, A]] {
+  implicit def catsDataSemilatticeForNonEmptyMap[K, A]: Semilattice[NonEmptyMap[K, A]] = new Semilattice[NonEmptyMap[K, A]] {
```
kailuowang: you changed it to semilattice here but not the law testing? I am not sure if `Map` combine is commutative.

LukaJCB: Oh, right, yeah that was dumb, I'll check that one later :)

core/src/main/scala/cats/data/NonEmptyMapImpl.scala
```diff
@@ -0,0 +1,327 @@
...
+  def of[K: Order, A](a: (K, A), as: (K, A)*): NonEmptyMap[K, A] =
+    create(SortedMap(as: _*)(Order[K].toOrdering) + a)
+
+  def one[K: Order, A](k: K, a: A): NonEmptyMap[K, A] =
+    create(SortedMap((k, a))(Order[K].toOrdering))
+
+  implicit def catsNonEmptyMapOps[K, A](value: Type[K, A]): NonEmptyMapOps[K, A] =
+    new NonEmptyMapOps(value)
+
+
```
aeons: Nitpicky, I know, but there are a lot of these double empty lines around.

core/src/main/scala/cats/data/NonEmptyMapImpl.scala
```diff
@@ -0,0 +1,327 @@
...
+  /**
+    * Optionally returns the value associated with the given key.
+    */
+  def lookup(k: K): Option[A] = toSortedMap.get(k)
+
+  /**
+    * Returns a `SortedSet` containing all the keys of this map.
+    */
+  def keys: SortedSet[K] = toSortedMap.keySet
```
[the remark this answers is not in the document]
LukaJCB: Yep, as mentioned before here :) https://github.com/typelevel/cats/pull/2141#discussion_r164121840

core/src/main/scala/cats/data/Newtype2.scala
```diff
@@ -0,0 +1,8 @@
+package cats
+package data
+
+trait Newtype2 { self =>
```
ceedubs: This seems kind of strange. Is this the same as the `Newtype` in the `Set` PR?

core/src/main/scala/cats/data/NonEmptyMapImpl.scala
```diff
@@ -0,0 +1,327 @@
...
+
+  /**
+    * Left-associative reduce using f.
+    */
+  def reduceLeft(f: (A, A) => A): A =
+    reduceLeftTo(identity)(f)
+
+
+  /**
+    * Apply `f` to the "initial element" of this map and lazily combine it
```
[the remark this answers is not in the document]
LukaJCB: A mistake :smile:

core/src/main/scala/cats/data/package.scala
```diff
@@ -11,6 +11,9 @@ package object data {
   def NonEmptyStream[A](head: A, tail: A*): NonEmptyStream[A] =
     OneAnd(head, tail.toStream)
 
+  type NonEmptyMap[K, +A] = NonEmptyMapImpl.Type[K, A]
+  val NonEmptyMap = NonEmptyMapImpl
```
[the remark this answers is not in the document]
LukaJCB: It is its own type aka, `cats.data.NonEmptyMap.type` :)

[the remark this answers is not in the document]
durban: Well, sure ... but that's not what I asked. The inferred type seems to be this:
```scala
private[this] val NonEmptyMap: cats.data.NonEmptyMapImpl.type = NonEmptyMapImpl;
<stable> <accessor> def NonEmptyMap: cats.data.NonEmptyMapImpl.type = `package`.this.NonEmptyMap;
```
Which means that we have a `private[data]` type on a public API.

core/src/main/scala/cats/data/NonEmptyMapImpl.scala
```diff
@@ -0,0 +1,320 @@
...
+  def one[K: Order, A](k: K, a: A): NonEmptyMap[K, A] =
+    create(SortedMap((k, a))(Order[K].toOrdering))
+
+  implicit def catsNonEmptyMapOps[K, A](value: Type[K, A]): NonEmptyMapOps[K, A] =
+    new NonEmptyMapOps(value)
+
+
+}
+
+sealed class NonEmptyMapOps[K, A](val value: NonEmptyMap[K, A]) {
```
durban: Maybe `private val value`?

LukaJCB: These are just the Ops, what do we gain from making the `value` public? :)

durban: Nothing, that's why I suggested making it private :-)
In fact, I think it shouldn't be a `val` at all.

LukaJCB: Oh yeah, I got it switched around in my head :smile: 
Currently all of our Ops Syntax classes look this way and consistency is good the way I see it. 
What would it be if not a `val`?

aeons: Adding `val` makes it a member on the class, omitting it, it's only in the constructor.

LukaJCB: Huh, yeah that makes sense, I wonder why that's the convention for syntax classes in cats, maybe someone else can chime in 🤔

ceedubs: @LukaJCB you have to make it public (ie `val`) if it's a value class, which is why you see it on other syntax classes.

core/src/main/scala/cats/data/NonEmptyMapImpl.scala
```diff
@@ -0,0 +1,320 @@
...
+
+  def fromMap[K: Order, A](as: SortedMap[K, A]): Option[NonEmptyMap[K, A]] =
+    if (as.nonEmpty) Option(create(as)) else None
+
+  def fromMapUnsafe[K: Order, A](m: SortedMap[K, A]): NonEmptyMap[K, A] =
+    if (m.nonEmpty) create(m)
+    else throw new IllegalArgumentException("Cannot create NonEmptyMap from empty map")
+
+  def apply[K: Order, A](head: (K, A), tail: SortedMap[K, A]): NonEmptyMap[K, A] =
+    create(SortedMap(head)(Order[K].toOrdering) ++ tail)
```
ceedubs: This and I think some other methods don't follow the convention that we kind of settled on in the early days of Cats [here](https://github.com/typelevel/cats/issues/27). Admittedly, that never got added to the guidelines, and I'm not sure whether we have been consistent about that in other places.

LukaJCB: I personally don't have a strong opinion on this and didn't know we had such a convention :smile:, you're right we should maybe reopen discussion and add it to the guideline, but for now I'll just change it, so we can avoid calling `apply` on the companion :+1: Thanks!

core/src/main/scala/cats/data/NonEmptyMapImpl.scala
peterneyens: If I remember correctly the reason it must be public is 2.10. For 2.11 and 2.12 `private` would work as well.

core/src/main/scala/cats/data/NonEmptyMapImpl.scala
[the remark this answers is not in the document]
kailuowang: 2141 merged :)

LukaJCB:
@ceedubs everything addressed? :)

PR #2143, opened 2018-01-08
Add NonEmptySet

LukaJCB:
Pretty much the same situation as #2141

### DIFF
--- a/core/src/main/scala/cats/data/Newtype.scala
+++ b/core/src/main/scala/cats/data/Newtype.scala
@@ -1,0 +1,8 @@
+package cats
+package data
+
+trait Newtype { self =>
+  private[data] type Base
+  private[data] trait Tag extends Any
+  type Type[A] <: Base with Tag
+}
--- a/core/src/main/scala/cats/data/Newtype.scala
+++ b/core/src/main/scala/cats/data/Newtype.scala
@@ -5,7 +5,7 @@ package data
   * Helper trait for `newtype`s. These allow you to create a zero-allocation wrapper around a specific type.
   * Similar to `AnyVal` value classes, but never have any runtime overhead.
   */
-trait Newtype { self =>
+private[data] trait Newtype { self =>
   private[data] type Base
   private[data] trait Tag extends Any
   type Type[A] <: Base with Tag
--- a/core/src/main/scala/cats/data/Newtype.scala
+++ b/core/src/main/scala/cats/data/Newtype.scala
@@ -1,6 +1,10 @@
 package cats
 package data
 
+/**
+  * Helper trait for `newtype`s. These allow you to create a zero-allocation wrapper around a specific type.
+  * Similar to `AnyVal` value classes, but never have any runtime overhead.
+  */
 trait Newtype { self =>
   private[data] type Base
   private[data] trait Tag extends Any
--- a/core/src/main/scala/cats/data/NonEmptySet.scala
+++ b/core/src/main/scala/cats/data/NonEmptySet.scala
@@ -33,10 +33,10 @@ private[data] object NonEmptySetImpl extends NonEmptySetInstances with Newtype {
     s.asInstanceOf[SortedSet[A]]
 
 
-  def fromSet[A: Order](as: SortedSet[A]): Option[NonEmptySet[A]] =
+  def fromSet[A](as: SortedSet[A]): Option[NonEmptySet[A]] =
     if (as.nonEmpty) Option(create(as)) else None
 
-  def fromSetUnsafe[A: Order](set: SortedSet[A]): NonEmptySet[A] =
+  def fromSetUnsafe[A](set: SortedSet[A]): NonEmptySet[A] =
     if (set.nonEmpty) create(set)
     else throw new IllegalArgumentException("Cannot create NonEmptySet from empty set")
 
--- a/core/src/main/scala/cats/data/NonEmptySet.scala
+++ b/core/src/main/scala/cats/data/NonEmptySet.scala
@@ -1,0 +1,218 @@
+/*
+ * Copyright (c) 2018 Luka Jacobowitz
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cats
+package data
+
+import cats.instances.sortedSet._
+import cats.kernel._
+import cats.{Always, Eq, Eval, Foldable, Later, Now, Reducible, SemigroupK, Show}
+
+import scala.collection.immutable._
+
+final class NonEmptySet[A] private (val set: SortedSet[A]) extends AnyVal {
+
+  private implicit def ordering: Ordering[A] = set.ordering
+  private implicit def order: Order[A] = Order.fromOrdering
+
+  def +(a: A): NonEmptySet[A] = new NonEmptySet(set + a)
+  def ++(as: NonEmptySet[A]): NonEmptySet[A] = concat(as)
+
+  def concat(as: NonEmptySet[A]): NonEmptySet[A] = new NonEmptySet(set ++ as.set)
+
+  def -(a: A): SortedSet[A] = set - a
+  def map[B: Order](f: A ⇒ B): NonEmptySet[B] =
+    new NonEmptySet(SortedSet(set.map(f).to: _*)(Order[B].toOrdering))
+
+  def toNonEmptyList: NonEmptyList[A] = NonEmptyList.fromListUnsafe(set.toList)
+
+  def head: A = set.head
+  def tail: SortedSet[A] = set.tail
+  def last: A = set.last
+
+  def apply(a: A): Boolean = set(a)
+  def contains(a: A): Boolean = set.contains(a)
+
+
+  def union(as: NonEmptySet[A]): NonEmptySet[Ag] = new NonEmptySet(set.union(as.toSet))
+  def size: Int = set.size
+  def forall(p: A ⇒ Boolean): Boolean = set.forall(p)
+  def exists(f: A ⇒ Boolean): Boolean = set.exists(f)
+  def find(f: A ⇒ Boolean): Option[A] = set.find(f)
+  def collect[B](pf: PartialFunction[A, B]): Set[B] = set.collect(pf)
+  def filter(p: A ⇒ Boolean): SortedSet[A] = set.filter(p)
+  def filterNot(p: A ⇒ Boolean): SortedSet[A] = filter(t => !p(t))
+
+
+  /**
+    * Left-associative fold using f.
+    */
+  def foldLeft[B](b: B)(f: (B, A) => B): B =
+    set.foldLeft(b)(f)
+
+  /**
+    * Right-associative fold using f.
+    */
+  def foldRight[B](lb: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B] =
+    Foldable[SortedSet].foldRight(set, lb)(f)
+
+  /**
+    * Left-associative reduce using f.
+    */
+  def reduceLeft(f: (A, A) => A): A =
+    set.reduceLeft(f)
+
+  def reduceLeftTo[B](f: A => B)(g: (B, A) => B): B = {
+    tail.foldLeft(f(head))((b, a) => g(b, a))
+  }
+
+  /**
+    * Left-associative reduce using f.
+    */
+  def reduceRight(f: (A, Eval[A]) => Eval[A]): Eval[A] =
+    reduceRightTo(identity)(f)
+
+  def reduceRightTo[B](f: A => B)(g: (A, Eval[B]) => Eval[B]): Eval[B] =
+    Always((head, tail)).flatMap { case (a, ga) =>
+      Foldable[SortedSet].reduceRightToOption(ga)(f)(g).flatMap {
+        case Some(b) => g(a, Now(b))
+        case None => Later(f(a))
+      }
+    }
+
+  /**
+    * Reduce using the Semigroup of A
+    */
+  def reduce[AA >: A](implicit S: Semigroup[AA]): AA =
+    S.combineAllOption(set).get
+
+  def toSet: SortedSet[A] = set
+
+  /**
+    * Typesafe stringification method.
+    *
+    * This method is similar to .toString except that it stringifies
+    * values according to Show[_] instances, rather than using the
+    * universal .toString method.
+    */
+  def show(implicit A: Show[A]): String =
+    s"NonEmpty${Show[SortedSet[A]].show(set)}"
+
+  /**
+    * Typesafe equality operator.
+    *
+    * This method is similar to == except that it only allows two
+    * NonEmptySet[A] values to be compared to each other, and uses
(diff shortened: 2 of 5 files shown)


core/src/main/scala/cats/data/NonEmptySet.scala
```diff
@@ -0,0 +1,218 @@
...
+package cats
+package data
+
+import cats.instances.sortedSet._
+import cats.kernel._
+import cats.{Always, Eq, Eval, Foldable, Later, Now, Reducible, SemigroupK, Show}
+
+import scala.collection.immutable._
+
+final class NonEmptySet[A] private (val set: SortedSet[A]) extends AnyVal {
```
johnynek: `AnyVal` boxing can lead to a lot of reboxing when this is used in a generic context, which it would be in cats code since any of the typeclasses use it in generic context. I think we don't actually want `AnyVal`.

core/src/main/scala/cats/data/NonEmptySet.scala
```diff
@@ -0,0 +1,218 @@
...
+
+
+  def union(as: NonEmptySet[A]) = new NonEmptySet(set.union(as.toSet))
+  def size: Int = set.size
+  def forall(p: A ⇒ Boolean): Boolean = set.forall(p)
+  def exists(f: A ⇒ Boolean): Boolean = set.exists(f)
+  def find(f: A ⇒ Boolean): Option[A] = set.find(f)
+  def collect[B](pf: PartialFunction[A, B]): Set[B] = set.collect(pf)
+  def filter(p: A ⇒ Boolean): SortedSet[A] = set.filter(p)
+  def filterNot(p: A ⇒ Boolean): SortedSet[A] = filter(t => !p(t))
```
johnynek: I think we can also add `concatMap[B](A => SortedSet[B]): SortedSet[B]` and I think it would be useful. We can also do `concatMap[B](A => NonEmptySet[B]): NonEmptySet[B]`

core/src/main/scala/cats/data/NonEmptySet.scala
johnynek: nit: should this be `NonEmptySortedSet`? In the future we might want a Hash-based Set so just using `Set` can be ambiguous.

core/src/main/scala/cats/data/NonEmptySet.scala
```diff
@@ -0,0 +1,218 @@
...
+      }
+    }
+
+  /**
+    * Reduce using the Semigroup of A
+    */
+  def reduce[AA >: A](implicit S: Semigroup[AA]): AA =
+    S.combineAllOption(set).get
+
+  def toSet: SortedSet[A] = set
```
johnynek: what about `toSortedSet`?

core/src/main/scala/cats/data/NonEmptySet.scala
```diff
@@ -0,0 +1,218 @@
...
+
+  def fromSetUnsafe[A: Order](set: SortedSet[A]): NonEmptySet[A] =
+    if (set.nonEmpty) new NonEmptySet(set)
+    else throw new IllegalArgumentException("Cannot create NonEmptySet from empty set")
+
+
+  def of[A: Order](a: A, as: A*): NonEmptySet[A] =
+    new NonEmptySet(SortedSet(a)(Order[A].toOrdering) ++ SortedSet(as: _*)(Order[A].toOrdering) + a)
+  def apply[A: Order](head: A, tail: SortedSet[A]): NonEmptySet[A] = new NonEmptySet(SortedSet(head)(Order[A].toOrdering) ++ tail)
+  def one[A: Order](a: A): NonEmptySet[A] = new NonEmptySet(SortedSet(a)(Order[A].toOrdering))
```
johnynek: do we need this one vs `of`? I also like `singleton` or something better.

core/src/main/scala/cats/data/NonEmptySet.scala
```diff
@@ -0,0 +1,218 @@
...
+
+  implicit def catsDataEqForNonEmptySet[A: Order]: Eq[NonEmptySet[A]] =
+    new Eq[NonEmptySet[A]]{
+      def eqv(x: NonEmptySet[A], y: NonEmptySet[A]): Boolean = x === y
+    }
+
+  implicit def catsDataShowForNonEmptySet[A](implicit A: Show[A]): Show[NonEmptySet[A]] =
+    Show.show[NonEmptySet[A]](_.show)
+
+  implicit def catsDataBandForNonEmptySet[A]: Band[NonEmptySet[A]] = new Band[NonEmptySet[A]] {
```
johnynek: this is a semilattice isn't it (it is commutative, no?) Am I missing something?

tests/src/test/scala/cats/tests/NonEmptySetSuite.scala
```diff
@@ -0,0 +1,225 @@
...
+import scala.collection.immutable.SortedSet
+
+class NonEmptySetSuite extends CatsSuite {
+
+  checkAll("NonEmptySet[Int]", SemigroupKTests[NonEmptySet].semigroupK[Int])
+  checkAll("NonEmptySet[Int]", ReducibleTests[NonEmptySet].reducible[Option, Int, Int])
+  checkAll("NonEmptySet[String]", BandTests[NonEmptySet[String]].band)
+  checkAll("NonEmptySet[String]", EqTests[NonEmptySet[String]].eqv)
+
+
```
johnynek: can we test `nes.forall { v => Order[Int].lteq(nes.head, v) }`

core/src/main/scala/cats/data/NonEmptySet.scala
LukaJCB: Good point! Man, we really need opaque types :(

LukaJCB: Hmm, yeah in theory this is true, but that's quite a long name, we could just have `NonEmptySet` and `NonEmptyHashSet` maybe?

core/src/main/scala/cats/data/NonEmptySet.scala
LukaJCB: Yeah, that's no doubt useful.

core/src/main/scala/cats/data/NonEmptySet.scala
LukaJCB: 👍

core/src/main/scala/cats/data/NonEmptySet.scala
LukaJCB: I took the name from `NonEmptyList.one`, but I have no preference wrt naming.

LukaJCB:
Maybe we should implement these with newtypes similar to what Alex did here: https://github.com/typelevel/cats-effect/pull/115/files#diff-c01f002766a28f49dfe20d6b1f0afd50R300

Thoughts?

core/src/main/scala/cats/data/NonEmptySet.scala
kailuowang: I vote for consistent API

kailuowang:
I am also fine of renaming it to `NonEmptySortedSet`, it's a long name, but not a very high cost to avoid potential confusion.

core/src/main/scala/cats/data/NonEmptySet.scala
```diff
@@ -0,0 +1,409 @@
...
+
+import cats.instances.sortedSet._
+import cats.kernel._
+import cats.{Always, Eq, Eval, Foldable, Later, Now, Reducible, SemigroupK, Show}
+
+import scala.collection.immutable._
+
+final class NonEmptySet[A] private (val set: SortedSet[A]) {
+
+  private implicit def ordering: Ordering[A] = set.ordering
```
LukaJCB: These should be `val`s

core/src/main/scala/cats/data/NonEmptySet.scala
```diff
@@ -23,22 +23,54 @@ import cats.{Always, Eq, Eval, Foldable, Later, Now, Reducible, SemigroupK, Show
...
 import scala.collection.immutable._
 
-final class NonEmptySet[A] private (val set: SortedSet[A]) {
+trait Newtype { self =>
+  private[data] type Base
+  private[data] trait Tag extends Any
+  private[cats] type Type[A] <: Base with Tag
+}
+
+private[data] object NonEmptySetImpl extends NonEmptySetInstances with Newtype {
```
kailuowang: shall we override the toString as well?

LukaJCB: We can't really do that, since we can't actually add methods on the type, but only through implicit extension methods. This means that it will always call  the `toString` of the underlying type, we can of course use `show` though.

core/src/main/scala/cats/data/NonEmptySet.scala
```diff
@@ -0,0 +1,415 @@
...
+package cats
+package data
+
+import cats.instances.sortedSet._
+import cats.kernel._
+import cats.{Always, Eq, Eval, Foldable, Later, Now, Reducible, SemigroupK, Show}
+
+import scala.collection.immutable._
+
+trait Newtype { self =>
```
andyscott: Can we move `NewType` to it's own file?

LukaJCB:
@johnynek Care to re-review? :)

core/src/main/scala/cats/data/NonEmptySet.scala
```diff
@@ -0,0 +1,410 @@
...
+  def fromSet[A: Order](as: SortedSet[A]): Option[NonEmptySet[A]] =
+    if (as.nonEmpty) Option(create(as)) else None
+
+  def fromSetUnsafe[A: Order](set: SortedSet[A]): NonEmptySet[A] =
+    if (set.nonEmpty) create(set)
+    else throw new IllegalArgumentException("Cannot create NonEmptySet from empty set")
+
+
+  def of[A: Order](a: A, as: A*): NonEmptySet[A] =
+    create(SortedSet(a)(Order[A].toOrdering) ++ SortedSet(as: _*)(Order[A].toOrdering) + a)
```
ceedubs: Is this adding `a` to the set twice? (Initializing with `SortedSet(a)` and then calling `+ a`.)

ceedubs: This is making two separate calls to `toOrdering`. If the ordering is needed twice, a slight optimization would be to store the result of `toOrdering`. Also, on a more minor note, the `Order[A]` instance is being summoned twice, so if the implicit instance is defined by a `def` this will result in unnecessary allocations.

core/src/main/scala/cats/data/Newtype.scala
```diff
@@ -0,0 +1,8 @@
+package cats
+package data
+
+trait Newtype { self =>
```
ceedubs: Can you add some documentation that motivates this `trait`?

core/src/main/scala/cats/data/NonEmptySet.scala
```diff
@@ -0,0 +1,410 @@
+/*
+ * Copyright (c) 2018 Luka Jacobowitz
```
ceedubs: Is there a reason that this file has this header but `Newtype.scala` doesn't?

core/src/main/scala/cats/data/NonEmptySet.scala
LukaJCB: Yeah, you're absolutely right, this is a really stupid mistake, not sure how it slipped through, probably refactored it one to many times 🙈

core/src/main/scala/cats/data/Newtype.scala
LukaJCB: Yes, good idea :)

core/src/main/scala/cats/data/NonEmptySet.scala
LukaJCB: Nope, just copy and paste, Kai suggested we try to standardize on something in a future PR, possibly with sbt-header or something like that :)

core/src/main/scala/cats/data/NonEmptySet.scala
```diff
@@ -0,0 +1,411 @@
...
+private[data] object NonEmptySetImpl extends NonEmptySetInstances with Newtype {
+
+  private[cats] def create[A](s: SortedSet[A]): Type[A] =
+    s.asInstanceOf[Type[A]]
+
+  private[cats] def unwrap[A](s: Type[A]): SortedSet[A] =
+    s.asInstanceOf[SortedSet[A]]
+
+
+  def fromSet[A: Order](as: SortedSet[A]): Option[NonEmptySet[A]] =
```
ceedubs: This takes an implicit `Order[A]` but then doesn't actually use it, does it? That seems misleading to me.

LukaJCB: Yeah that's true, it assumes there's an `Ordering` for `A`, maybe best to get rid of it?

kailuowang: Is this the last bit that prevents this PR being merged?

LukaJCB: It's done :)

kailuowang:
@ceedubs any other issues?

kailuowang:
@johnynek were all your feedback addressed?

core/src/main/scala/cats/data/NonEmptySet.scala
```diff
@@ -0,0 +1,411 @@
...
+  /**
+    * Returns the first value that matches the given predicate.
+    */
+  def find(f: A ⇒ Boolean): Option[A] = toSortedSet.find(f)
+
+  /**
+    * Returns a new `SortedSet` containing all elements where the result of `pf` is defined.
+    */
+  def collect[B: Order](pf: PartialFunction[A, B]): SortedSet[B] = {
+    implicit val ordering = Order[B].toOrdering
```
ceedubs: Same comment as https://github.com/typelevel/cats/pull/2141/files#r173359573

ceedubs: I think there are some other places that this applies to this PR. I don't necessarily consider this a blocker, but I'd like us to at least decide what we want to do about this before we publish a release that has the name of the implicit parameter fixed (I think that the auto-generated one will be something like `ev1`).

LukaJCB:
I removed the context bounds @ceedubs :)

core/src/main/scala/cats/data/Newtype.scala
```diff
@@ -0,0 +1,12 @@
+package cats
+package data
+
+/**
+  * Helper trait for `newtype`s. These allow you to create a zero-allocation wrapper around a specific type.
+  * Similar to `AnyVal` value classes, but never have any runtime overhead.
+  */
+trait Newtype { self =>
```
ceedubs: Does it make sense for this to be a public trait if `Base` and `Tag` are both `private[data]`?

LukaJCB: Debatable IMO, but to me those two are implementation details, you're never actually going to want to access these, no?

ceedubs: Oh right I guess that in order to create your own type that uses `Newtype` you don't actually have to set `Base` and `Tag`, right?

I guess I'm partially hesitant to expose `Newtype` publicly because there are a lot of approaches to newtypes in the scala world and I'm not sure whether we want to take on the burden of keeping this one around and stable in cats. Especially since it's an internal implementation detail that isn't really used as an abstraction and is trivial enough that people could easily copy/paste it if they wanted it. I'd be inclined to make it private for now and if it works out well and people want it to be exposed publicly it's easier to open it up later than to make it private later. Admittedly I'm sometimes overly paranoid about this sort of thing...

LukaJCB: I'd be okay with making them private :)

guizmaii:
Hi,

Why finally did you choose to name it `NonEmptySet` instead of `NonEmptySortedSet` ?

PR #2146, opened 2018-01-10
added `leftNel` and `rightNel` syntax

kailuowang:
since `EitherNel` is a lot more useful with the `Parallel` TC

### DIFF
--- a/core/src/main/scala/cats/syntax/either.scala
+++ b/core/src/main/scala/cats/syntax/either.scala
@@ -1,7 +1,7 @@
 package cats
 package syntax
 
-import cats.data.{EitherT, Ior, Validated, ValidatedNel}
+import cats.data.{EitherT, Ior, NonEmptyList, Validated, ValidatedNel}
 import scala.reflect.ClassTag
 import scala.util.{Failure, Success, Try}
 import EitherSyntax._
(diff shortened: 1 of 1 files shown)


core/src/main/scala/cats/syntax/either.scala
```diff
@@ -329,6 +329,31 @@ final class EitherIdOps[A](val obj: A) extends AnyVal {
...
   /** Wrap a value in `Right`. */
   def asRight[B]: Either[B, A] = Right(obj)
+
+  /**
+   * Wrap a value to a left EitherNel
+   *
+   * For example:
+   * {{{
+   * scala> import cats.implicits._, cats.data.NonEmptyList
+   * scala> "Err".leftNel[Int]
```
LukaJCB: Whitespace error on the end of this line

core/src/main/scala/cats/syntax/either.scala
```diff
@@ -329,6 +329,31 @@ final class EitherIdOps[A](val obj: A) extends AnyVal {
...
+   */
+  def leftNel[B]: Either[NonEmptyList[A], B] = Left(NonEmptyList.of(obj))
+
+  /**
+   * Wrap a value to a right EitherNel
+   *
+   * For example:
+   * {{{
+   * scala> import cats.implicits._, cats.data.NonEmptyList
+   * scala> 1.rightNel[String]
```
LukaJCB: This line also has an extra space

core/src/main/scala/cats/syntax/either.scala
```diff
@@ -329,6 +329,31 @@ final class EitherIdOps[A](val obj: A) extends AnyVal {
...
+   * Wrap a value to a left EitherNel
+   *
+   * For example:
+   * {{{
+   * scala> import cats.implicits._, cats.data.NonEmptyList
+   * scala> "Err".leftNel[Int]
+   * res0: Either[NonEmptyList[String], Int] = Left(NonEmptyList(Err))
+   * }}}
+   */
+  def leftNel[B]: Either[NonEmptyList[A], B] = Left(NonEmptyList.of(obj))
```
peterneyens: You could use `NonEmptyList.one` here.

LukaJCB: :+1:

kailuowang: Thanks! @peterneyens  addressed.

Issue #2147, opened 2018-01-11
Missing syntax import for `.unorderedTraverse`


andyscott:
The import `cats.syntax.traverse._` doesn't provide the `.unorderedTraverse`. This is inconsistent, as the import `cats.syntax.foldable._` does provide the `unorderedFold` syntax.

Unfortunately I think an immediate fix for this would break binary compatibility. `import cats.UnorderedTraverse.ops._` can be used for the time being.

kailuowang:
`cats.syntax.traverse` can extend an extra trait without breaking bin compat. So I think we should be able to fix this one

andyscott:
If that's the case, I'll submit a PR shortly.

PR #2148, opened 2018-01-11
Add missing UnorderedTraverse syntax

andyscott:
Addresses #2147:
>The import cats.syntax.traverse._ doesn't provide the .unorderedTraverse. This is inconsistent, as the import cats.syntax.foldable._ does provide the unorderedFold syntax.

### DIFF
--- a/core/src/main/scala/cats/syntax/all.scala
+++ b/core/src/main/scala/cats/syntax/all.scala
@@ -1,6 +1,10 @@
 package cats
 package syntax
 
+abstract class AllSyntaxBinCompat
+    extends AllSyntax
+    with AllSyntaxBinCompat0
+
 trait AllSyntax
     extends AlternativeSyntax
     with ApplicativeSyntax
@@ -48,3 +52,6 @@ trait AllSyntax
     with ValidatedSyntax
     with VectorSyntax
     with WriterSyntax
+
+trait AllSyntaxBinCompat0
+    extends UnorderedTraverseSyntax
--- a/core/src/main/scala/cats/implicits.scala
+++ b/core/src/main/scala/cats/implicits.scala
@@ -1,3 +1,6 @@
 package cats
 
-object implicits extends syntax.AllSyntax with instances.AllInstances
+object implicits
+    extends syntax.AllSyntax
+    with syntax.AllSyntaxBinCompat0
+    with instances.AllInstances
--- a/core/src/main/scala/cats/syntax/package.scala
+++ b/core/src/main/scala/cats/syntax/package.scala
@@ -1,7 +1,7 @@
 package cats
 
 package object syntax {
-  object all extends AllSyntax
+  object all extends AllSyntaxBinCompat
   object alternative extends AlternativeSyntax
   object applicative extends ApplicativeSyntax
   object applicativeError extends ApplicativeErrorSyntax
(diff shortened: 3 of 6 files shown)


andyscott:
2.10 and 2.11 JVM failed the binary compatibility check. /cc @kailuowang

kailuowang:
You can't touch an existing trait, you have to have the `traverse` object extend this extra trait instead of through the `TraverseSyntax` trait

andyscott:
Aah yes. I'll get that switched today.

core/src/main/scala/cats/syntax/traverse.scala
```diff
@@ -2,3 +2,4 @@ package cats
 package syntax
 
 trait TraverseSyntax extends Traverse.ToTraverseOps
+trait TraverseSyntaxBinCompat0 extends UnorderedTraverse.ToUnorderedTraverseOps
```
kailuowang: shall we just name this one `UnorderedTraverseSyntax` since we dont' have one?

kailuowang:
also if this this not available in `cats.implicits._` we need to have that one extend this trait as well.

core/src/main/scala/cats/syntax/traverse.scala
andyscott: We could. Is there any merit to keeping it as similar to Foldable/UnorderedFoldable (which doesn't have `UnorderedFoldableSyntax`)?

kailuowang: `Foldable/UnorderedFoldable` is the unconventional one. I am referring to 
`trait FoldableSyntax extends Foldable.ToFoldableOps with UnorderedFoldable.ToUnorderedFoldableOps`
The convention in Cats is that the syntax trait of a TC does NOT extend the syntax of its parent TC. For users to get the parent syntax, they have to either explicitly import the parent syntax, or using the all inclusive import. 
So I think having an `UnorderdTraverseSyntax` trait is more conventional.

kailuowang: sorry about the bike shedding here. The main reason I am discussing more about it is that this would set up a precedence for working around bin compat breakage. although it's a bit atypical

kailuowang:
@andyscott how do you want to proceed with this PR? I'll be happy to make the rename changes for you if you want.

andyscott:
Sorry for the delay. I can take care of it within the next day.

core/src/main/scala/cats/syntax/all.scala
```diff
@@ -48,3 +48,6 @@ trait AllSyntax
     with ValidatedSyntax
     with VectorSyntax
     with WriterSyntax
+
+trait AllSyntaxBinCompat0
```
kailuowang: One more small thing, if we make this one abstract class (and maybe even have it extends AllSyntax) we can add more traits to it going forward.

andyscott: How about a `abstract class AllSyntaxBinCompat extends AllSyntax with AllSyntaxBinCompat0`?

kailuowang: then what's the purpose of AllSyntaxBinCompat0 then? just as a versioning scheme?

andyscott: A downstream consumer can't mix in the abstract class but they could mix in `AllSyntax` and `AllSyntaxBinCompat0` to get roughly the same effect. Without `AllSyntaxBinCompat0`, but then you'd have to remember that `AllSyntax` is missing `UnorderedTraverse` syntax. So it's versioning scheme to facilitate downstream use.

I'm not sure if it's worth it though. It's just an idea.

kailuowang: I see. Makes sense to me

PR #2149, opened 2018-01-11
Add leftFlatMap to Either

rohanshah:
### Motivation

`Either` should have `leftFlatMap` for cases when you want to `leftMap` a function that returns an `Either` and flatten the total result.

### DIFF
--- a/core/src/main/scala/cats/syntax/either.scala
+++ b/core/src/main/scala/cats/syntax/either.scala
@@ -148,6 +148,11 @@ final class EitherOps[A, B](val eab: Either[A, B]) extends AnyVal {
     case Right(b)    => f(b)
   }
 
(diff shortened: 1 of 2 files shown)


kailuowang:
would love to see some tests

rohanshah:
@kailuowang yeah, I was wondering where to put them. The only test file I found for `Either` was [EitherSuite.scala](https://github.com/typelevel/cats/blob/master/tests/src/test/scala/cats/tests/EitherSuite.scala) but that didn't even have a test for `flatMap` in it.

kailuowang:
EitherSuite should be the right place. My guess is that `flatMap` isn't explicitly tested there because  it's covered by the law tests

rohanshah:
@kailuowang added tests!

PR #2150, opened 2018-01-12
contravariant coyoneda

tpolecat:
This is mostly a copy/paste job from `Coyoneda` with the arrows flipped.

### DIFF
--- a/free/src/main/scala/cats/free/ContravariantCoyoneda.scala
+++ b/free/src/main/scala/cats/free/ContravariantCoyoneda.scala
@@ -65,8 +65,8 @@ object ContravariantCoyoneda {
       val fi = fa
     }
 
-  /** `ContravariantCoyoneda[F, ?]` provides a conntravariant functor for any `F`. */
-  implicit def catsFreeCovariantFunctorForCovariantCoyoneda[F[_]]: Contravariant[ContravariantCoyoneda[F, ?]] =
+  /** `ContravariantCoyoneda[F, ?]` provides a contravariant functor for any `F`. */
+  implicit def catsFreeContravariantFunctorForContravariantCoyoneda[F[_]]: Contravariant[ContravariantCoyoneda[F, ?]] =
     new Contravariant[ContravariantCoyoneda[F, ?]] {
       def contramap[A, B](cfa: ContravariantCoyoneda[F, A])(f: B => A): ContravariantCoyoneda[F, B] =
         cfa.contramap(f)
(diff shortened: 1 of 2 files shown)


free/src/main/scala/cats/free/ContravariantCoyoneda.scala
```diff
@@ -0,0 +1,75 @@
...
+   */
+  private[cats] def unsafeApply[F[_], A, B](fa: F[A])(ks0: List[Any => Any]): Aux[F, B, A] =
+    new ContravariantCoyoneda[F, B] {
+      type Pivot = A
+      val ks = ks0
+      val fi = fa
+    }
+
+  /** `ContravariantCoyoneda[F, ?]` provides a conntravariant functor for any `F`. */
+  implicit def catsFreeCovariantFunctorForCovariantCoyoneda[F[_]]: Contravariant[ContravariantCoyoneda[F, ?]] =
```
iravid: Should be `Contravariant` here

free/src/main/scala/cats/free/ContravariantCoyoneda.scala
```diff
@@ -0,0 +1,75 @@
...
+  private[cats] val ks: List[Any => Any]
+
+  /** The composed transformer function, to be lifted into `F` by `run`. */
+  final def k: A => Pivot = Function.chain(ks)(_).asInstanceOf[Pivot]
+
+  /** Converts to `F[A]` given that `F` is a contravariant functor */
+  final def run(implicit F: Contravariant[F]): F[A] = F.contramap(fi)(k)
+
+  /** Converts to `G[A]` given that `G` is a contravariant functor */
+  final def foldMap[G[_]](trans: F ~> G)(implicit G: Contravariant[G]): G[A] =
```
kailuowang: can we add a test for this one as well?

kailuowang:
@tpolecat would you like to bring this one into the upcoming 1.1 release?

tpolecat:
Yeah I'll try to do that test this morning.

tpolecat:
I added a kind of lame test that exercises `foldMap` … I couldn't come up with a contravariant functor that was anything other than some wrapping of `? => A`. Happy to take suggestions but I think this may just be the nature of contravariant functors.

kailuowang:
@iravid any other feedback on this one?

iravid:
Nope 👍🏻

--
Itamar

LukaJCB:
I'm very similar to Kai, the code looks very much okay to me, but I don't quite grasp the use cases, so if we could get one other review, maybe that'd be best.

tpolecat:
As a practical matter the main use case is factoring out the boilerplate of writing a [stack-safe] contravariant functor. Simple example [here](https://github.com/tpolecat/doobie-tsql/blob/master/modules/core/src/main/scala/tsql/Write.scala#L30-L41).

PR #2154, opened 2018-01-14
contravariant.md [reversing roles of A and B in commentary to match the code]

phderome:
that says that B is subtype of A.

See also gitter discussion with @tpolecat on Jan 14 2017.


phderome:
Luka, @tpolecat had additional clarification/suggestion in gitter that I missed, not sure what to do with it:

> Or A <= B implies F[A] => F[B] which is the same statement in a different category. 
> 
> I think we could do better explaining the relationship.

PR #2158, opened 2018-01-23
update travis cache settings

kailuowang:
mostly copied from circe


johnynek:
👍 

seems marginally faster, or at least not slower, and it is also simpler.

PR #2166, opened 2018-02-07
Fix "Documentation" link at the top of "https://typelevel.org/cats"

srp:
Currently the "Documentation" link at the top of "https://typelevel.org/cats" is broken. This fixes it. I have a feeling that this is actually generated by something, but figure this is a good way to raise the issue.

---
This change is [Reviewable](https://reviewable.io/reviews/typelevel/cats/2166)


kailuowang:
This is generated through sbt-microsit using this setting 
https://github.com/typelevel/cats/blob/master/build.sbt#L179
```
docsMappingsAPIDir := "api",
```
This only happens when you visit the cats website without the trailing slash in the url i.e.
"https://typelevel.org/cats". If you use "https://typelevel.org/cats/" it works. I suspect that if we change the settings to "/cats/api" it might work.

srp:
Looks like this was already fixed in 8db3501e2026fdfe94a44affbf5acaa64c433f51, closing.

PR #2167, opened 2018-02-07
Fix Compose documentation

sideeffffect:


kailuowang:
I don't quite grasp the incentive of this change. Could you elaborate a bit?

sideeffffect:
The comment was interfering with the `@simulacrum.op("<<<", alias = true)` annotation:

# Before

image

image

# After

image

PR #2174, opened 2018-02-21
[iffy] Add Ior Monoid instance

andyscott:
Adds an Monoid instance for Ior. It's a bit peculiar-- so I'm sure if it's a good candidate for merging-- but it works nonetheless.

This allows structures of Ior to be combined on both sides simultaneously:

```scala
val in: List[IorNel[String, Int]] = List(
  Ior.right(1),
  Ior.leftNel("foo"),
  Ior.bothNel("bar", 2))

in.map(_.map(_ :: Nil)).combineAll
// Both(NonEmptyList(foo, bar),List(1, 2))
```

### DIFF
--- a/core/src/main/scala/cats/data/Ior.scala
+++ b/core/src/main/scala/cats/data/Ior.scala
@@ -178,9 +178,14 @@ private[data] sealed abstract class IorInstances extends IorInstances0 {
     def show(f: A Ior B): String = f.show
   }
 
-  implicit def catsDataSemigroupForIor[A: Semigroup, B: Semigroup]: Semigroup[Ior[A, B]] = new Semigroup[Ior[A, B]] {
-    def combine(x: Ior[A, B], y: Ior[A, B]) = x.combine(y)
-  }
+  implicit def catsDataMonoidForIor[A: Semigroup, B: Monoid]: Monoid[Ior[A, B]] =
+    new Monoid[Ior[A, B]] {
+      lazy val empty: Ior[A, B] = Ior.right(Monoid[B].empty)
+      def combine(x: Ior[A, B], y: Ior[A, B]) =
+        if (x == empty) y
+        else if (y == empty) x
+        else x.combine(y)
+    }
 
   implicit def catsDataMonadErrorForIor[A: Semigroup]: MonadError[Ior[A, ?], A] =
     new MonadError[Ior[A, ?], A] {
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/data/Ior.scala
```diff
@@ -178,9 +178,14 @@ private[data] sealed abstract class IorInstances extends IorInstances0 {
...
 
-  implicit def catsDataSemigroupForIor[A: Semigroup, B: Semigroup]: Semigroup[Ior[A, B]] = new Semigroup[Ior[A, B]] {
-    def combine(x: Ior[A, B], y: Ior[A, B]) = x.combine(y)
-  }
+  implicit def catsDataMonoidForIor[A: Semigroup, B: Monoid]: Monoid[Ior[A, B]] =
+    new Monoid[Ior[A, B]] {
+      lazy val empty: Ior[A, B] = Ior.right(Monoid[B].empty)
+      def combine(x: Ior[A, B], y: Ior[A, B]) =
+        if (x == empty) y
+        else if (y == empty) x
```
johnynek: I don't think we can use `==` I think you are going to need `B: Eq`.

andyscott: 👍

core/src/main/scala/cats/data/Ior.scala
```diff
@@ -178,9 +178,14 @@ private[data] sealed abstract class IorInstances extends IorInstances0 {
     def show(f: A Ior B): String = f.show
   }
 
-  implicit def catsDataSemigroupForIor[A: Semigroup, B: Semigroup]: Semigroup[Ior[A, B]] = new Semigroup[Ior[A, B]] {
-    def combine(x: Ior[A, B], y: Ior[A, B]) = x.combine(y)
-  }
+  implicit def catsDataMonoidForIor[A: Semigroup, B: Monoid]: Monoid[Ior[A, B]] =
+    new Monoid[Ior[A, B]] {
+      lazy val empty: Ior[A, B] = Ior.right(Monoid[B].empty)
```
johnynek: I think this should be `def` or `val`. If `Monoid[B].empty` is expensive, it is up to `Monoid[B]` to see if it wants to cache it.

core/src/main/scala/cats/data/Ior.scala
johnynek: I don't feel great about this. It is a weird kind of combine in that `Right(_) combine Left(_)` returns `Both` *unless* you have `Right(Monoid[B].empty)` in which case it returns `Left`.

I think if you want to lift it into a monoid, using `Option[Ior[A, B]]` is the better approach than pressing Right into service.

andyscott: Yep. I felt a bit sketchy while writing this but figured it was worth sharing to get feedback.

PR #2179, opened 2018-03-01
added `liftTo` to `Try`, `Either` and `Option`

kailuowang:
simple syntax extensions

### DIFF
--- a/core/src/main/scala/cats/syntax/option.scala
+++ b/core/src/main/scala/cats/syntax/option.scala
@@ -2,6 +2,7 @@ package cats
 package syntax
 
 import cats.data.{Ior, Validated, ValidatedNel}
+import cats.syntax.OptionOps.LiftToPartiallyApplied
 
 trait OptionSyntax {
   final def none[A]: Option[A] = Option.empty[A]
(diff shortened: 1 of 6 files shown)


core/src/main/scala/cats/syntax/option.scala
```diff
@@ -170,4 +171,28 @@ final class OptionOps[A](val oa: Option[A]) extends AnyVal {
...
+    * scala> import cats.implicits._
+    *
+    * scala> Some(1).liftTo[Either[String, ?]]("Empty")
+    * res0: scala.Either[String, Int] = Right(1)
+    *
+    * scala> Option.empty[Int].liftTo[Either[String, ?]]("Empty")
+    * res1: scala.Either[String, Int] = Left(Empty)
+    * }}}
+    */
+  def liftTo[F[_]]: LiftToPartiallyApplied[F, A] = new LiftToPartiallyApplied(oa)
```
tpolecat: I'm surprised this works because `LiftToPartiallyApplied` is marked `private[cats]` … I guess I can have a value of this type but I can't declare or ascribe it. Interesting trick.

kailuowang: this is the fun of working with scala.

PR #2181, opened 2018-03-02
Add Applicative IorT instance

LukaJCB:
This was a missing instances cc @arosien

### DIFF
--- a/core/src/main/scala/cats/data/IorT.scala
+++ b/core/src/main/scala/cats/data/IorT.scala
@@ -482,7 +482,15 @@ private[data] abstract class IorTInstances2 extends IorTInstances3 {
     new IorTEq[F, A, B] { val F0: Eq[F[Ior[A, B]]] = F }
 }
 
-private[data] abstract class IorTInstances3 {
+private[data] abstract class IorTInstances3 extends IorTInstances4 {
+  implicit def catsDataApplicativeForIorT[F[_], A](implicit F: Applicative[F], A: Semigroup[A]): Applicative[IorT[F, A, ?]] =
+    new IorTApplicative[F, A] {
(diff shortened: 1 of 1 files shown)


peterneyens:
`EitherT` and `OptionT` explicitly don't have this instance because it is inconsistent with the `Monad` instance (which is one of the reason for the `toNested` methods on those data types).
I haven't checked it for `IorT`, but I suspect it will have the same issue.

tests/src/test/scala/cats/tests/IorTSuite.scala
```diff
@@ -29,6 +29,13 @@ class IorTSuite extends CatsSuite {
     checkAll("Traverse[IorT[ListWrapper, Int, ?]]", SerializableTests.serializable(Traverse[IorT[ListWrapper, Int, ?]]))
   }
 
+  {
+    implicit val F = ListWrapper.applicative
+
+    checkAll("IorT[ListWrapper, String, Int]", ApplicativeTests[IorT[ListWrapper, String, ?]].applicative[Int, Int, Int])
+    checkAll("Applicative[IorT[List, ?, ?]]", SerializableTests.serializable(Applicative[IorT[ListWrapper, String, ?]]))
```
martin-g: The description says `List` but the code uses `ListWrapper`. Is this intended ?

LukaJCB:
Thanks for pointing that out @peterneyens! I wasn't aware this would lead to inconsistencies, I understand how OptionT and EitherT would differ, but I'm not 100% certain I'd be the same for `IorT` I'll have to investigate some more.

arosien:
@peterneyens @LukaJCB perhaps we could add some kind of scaladoc that would let the user know to call `toNested` to use this in an `Applicative` manner.

PR #2184, opened 2018-03-08
Add Dsl.scala in the cats ecosystem

Atry:


Atry:
See https://static.javadoc.io/com.thoughtworks.dsl/dsl_2.12/1.0.0-RC2/com/thoughtworks/dsl/domains/cats$.html for the usage

kailuowang:
simple doc change, merge with one sign off. Note that this will show up in Readme but not on the website until the next release.

PR #2185, opened 2018-03-10
Fix #1733: make Kleisli.flatMap stack safe

### LINKED ISSUE #1733: Kliesli is not stack-safe on left-associated binds

alexandru:
Close #1733.

This is badly needed in `cats-effect` in order to define a `Sync[Kleisli[F, R, ?]]` implementation.

I'm already pushing one right now, but with a patched `flatMap` implementation that would no longer preserve _coherence_ with the official `flatMap` of Kleisli. So it would be nice if both these PRs happen, see: https://github.com/typelevel/cats-effect/pull/144

The way to fix it is to define a `flatMap` implementation that triggers the `F[_]` context _earlier_. Like this:

```scala
def flatMap[C](f: B => Kleisli[F, A, C])(implicit F: Monad[F]): Kleisli[F, A, C] =
  Kleisli(a => F.pure(a).flatMap(a => F.flatMap[B, C](run(a))((b: B) => f(b).run(a))))
  //                     ^^^^^^^                      ^^^^^^
  //              Suspended execution in F[_], no longer increases the stack size
```

Unfortunately this solution doesn't work directly because we have a `FlatMap` restriction on that operation, not `Monad`. And so the trick is to discriminate between plain `FlatMap` and `Monad` via inheritance and the subtyping encoding that we're all doing, this being a subtitute for `Kleisli.apply`:

```scala
private[data] def shift[F[_], A, B](run: A => F[B])
  (implicit F: FlatMap[F]): Kleisli[F, A, B] = {

  F match {
    case ap: Applicative[F] @unchecked =>
      Kleisli(r => F.flatMap(ap.pure(r))(run))
    case _ =>
      Kleisli(run)
  }
}
```

And now the `flatMap` definition can be:

```scala
def flatMap[C](f: B => Kleisli[F, A, C])(implicit F: FlatMap[F]): Kleisli[F, A, C] =
  Kleisli.shift(a => F.flatMap[B, C](run(a))((b: B) => f(b).run(a)))
```

As proven by the tests with `Eval`, this works. And it will work splendidly with `cats-effect`.

---

## FAQ

In preparation for the upcoming discussions, I can think of the following ...

### Is there any other solution possible?

No, this is the only fix possible for Cats version 1.x, given the `A => F[B]` encoding.

We could end up with a different internal encoding, say `F[A => F[B]]`, but note that such an encoding doesn't guarantee a fix either. In spite of popular belief note that `StateT` isn't currently stack safe for left binds either, see: https://github.com/typelevel/cats-effect/issues/139

### Doesn't this assume that a stack safe `F.pure(a).flatMap(f)` is lazy?

Yes, however it is my firm belief that no stack safe `flatMap` can be defined with strict evaluation, not even when the source is `F.pure`. This is because you can always define a recursive structure that's already evaluated (in memory) and that triggers stack overflows.

In other words if `F[_]` really has a stack safe `flatMap` (that could pass the laws of `Sync` in `cats-effect` for example, or that could describe `tailRecM`), then the behavior of `F.pure(a).flatMap(f)` is necessarily lazy.

For details and a working example: https://github.com/typelevel/cats-effect/issues/92

### Isn't usage of isInstanceOf a hack?

A much more elaborate signature would do something like the [Priority](https://github.com/typelevel/algebra/blob/master/core/src/main/scala/algebra/Priority.scala) pattern in Algebra.

However consider that:

1. we cannot break compatibility
2. this is cleaner, lets not forget the scary `CanBuildFrom` signatures
3. the cases where this will yield problems (i.e. those cases where the `Monad` definition comes from a different place than `FlatMap`) are totally uninteresting; the Scala compiler will pick `Monad` if in scope because it has a preference for sub-types or in case they happen to have the same "score", it yields a conflict anyway

Most importantly is that this PR solves a problem and without it we cannot solve that problem.

Perfect is the enemy of good 😉

### DIFF
--- a/core/src/main/scala/cats/data/Kleisli.scala
+++ b/core/src/main/scala/cats/data/Kleisli.scala
@@ -28,19 +28,19 @@ final case class Kleisli[F[_], A, B](run: A => F[B]) { self =>
     Kleisli[G, A, B](run andThen f.apply)
 
   def flatMap[C](f: B => Kleisli[F, A, C])(implicit F: FlatMap[F]): Kleisli[F, A, C] =
-    Kleisli((r: A) => F.flatMap[B, C](run(r))((b: B) => f(b).run(r)))
+    Kleisli.shift(a => F.flatMap[B, C](run(a))((b: B) => f(b).run(a)))
 
   def flatMapF[C](f: B => F[C])(implicit F: FlatMap[F]): Kleisli[F, A, C] =
-    Kleisli(a => F.flatMap(run(a))(f))
+    Kleisli.shift(a => F.flatMap(run(a))(f))
 
   def andThen[C](f: B => F[C])(implicit F: FlatMap[F]): Kleisli[F, A, C] =
-    Kleisli((a: A) => F.flatMap(run(a))(f))
+    Kleisli.shift(a => F.flatMap(run(a))(f))
 
   def andThen[C](k: Kleisli[F, B, C])(implicit F: FlatMap[F]): Kleisli[F, A, C] =
     this andThen k.run
 
   def compose[Z](f: Z => F[A])(implicit F: FlatMap[F]): Kleisli[F, Z, B] =
-    Kleisli((z: Z) => F.flatMap(f(z))(run))
+    Kleisli.shift((z: Z) => F.flatMap(f(z))(run))
 
   def compose[Z](k: Kleisli[F, Z, A])(implicit F: FlatMap[F]): Kleisli[F, Z, B] =
     this compose k.run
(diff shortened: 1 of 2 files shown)


rossabaker:
A couple practical observations:
* I have never been bitten by stack-unsafe `Kleisli`, and now all our taxes are going up.
* But `Sync[Kleisli[F, A, ?]]` is [useful in practice](https://github.com/typelevel/frameless/blob/master/docs/src/main/tut/Cats.md).

I don't have any benchmarks from real projects, but my gut feeling is that this is a good trade.

alexandru:
@rossabaker thanks.

As said in the issue that this solves, the stack unsafety is for "_left-associated binds_", which for our purposes are less important than "_right-associated binds_", which are those that happen in recursive loops.

I suspect the taxes are going up due to an extra `F.flatMap`, however if our `F` is an `IO` / `Task` data type, then it's probably not a big deal. If there's concern, I guess we can always do a benchmark.

I would be interested about Kleisli usage in the wild and what data types people use for `F`.

rossabaker:
In descending order of `F` frequency for me: `IO`, `OptionT[IO, ?]`, `Id`, `Future`.  Of these, `OptionT` and `Future` have the heavy `flatMap`s.  Where I use `OptionT`, I'm doing much heavier things.  And `Future` is something I'm happy to support but not happy to hold back progress to optimize.

LukaJCB:
These kinds of type system hacks always make me a bit weary (though tbh, I played around with implementing stack-safe function composition by using a big fat `List[Any => Any]` 😅), but from what I can see in this case, it seems to make a lot of sense, some quick benchmarks would be nice of course, but I understand if you don't have the time. After thinking about this for a while I'm :+1:

PR #2187, opened 2018-03-10
Fix #2186: make IndexedStateT stack safe

### LINKED ISSUE #2186: IndexedStateT is stack unsafe for left associative binds
This test will fail with a stack overflow:

```scala
test("flatMap is stack safe on repeated left binds when F is") {
  val unit = StateT.pure[Eval, Unit, Unit](())
  val result = (0 until 100000).foldLeft(unit) { (acc, _) =>
    acc.flatMap(_ => unit)
  }
  result.run(()).value should === (((), ()))
}
```

Note that the usual 10,000 iterations that we've been using for testing stack safety aren't enough, because for `IndexedStateT` the stack size doesn't grow that fast. But the error can be seen for 100,000.

This is a problem for projects like `cats-effect` that need to provide a lawful `Sync[StateT[F, S, ?]]` instance.

NOTE: similar with #1733

alexandru:
Close #2186.

This is very similar in spirit with the PR #2185 that fixes `Kleisli`, but this one is for `IndexedStateT` and with a different solution.

For `IndexedStateT` instead of piggybacking on an `F: Monad` for the stack safety of repeated left binds, this time we are introducing an `AndThen` implementation.

`AndThen` was first introduced in `cats-effect` for describing functions that are stack safe in `andThen` and `compose`, the original `cats.effect.IO` implementation being based on it.

Introduced as an internal helper (thus not exposed to the public), this is now used to provide stack safety for `IndexedStateT`.

Note that an implementation similar to that of PR #2185 is also possible, deferring this responsibility to `F[_]`, except that this version proposed here works for any `F[_]` when it comes to the stack safety of left associative binds.

---

## Benchmarks

I've added a benchmark (see [StateTBench](https://github.com/typelevel/cats/pull/2187/files#diff-cdb9115fe31f9ee2cf733f49d63749fc)) that I used to compare the new implementation with `master`. The results are:

| Benchmark |   Master |   This PR |  Units |
|-------------|----------:|----------:|------:|
| single flatMap | 3924997 | 3832845 | ops/s |
| repeated left binds | 353197 | 343995 | ops/s |
| repeated right binds | 346440 | 337696 | ops/s |

There is a difference but is imo insignificant.

## FAQ

In preparation for the upcoming discussion:

### Could this solution be used for Kleisli?

I do not see a solution, because `andThen` swallows the input and you need it a second time for triggering the instance returned by the mapping function:

```scala
def flatMap[C](f: B => Kleisli[F, A, C])(implicit F: FlatMap[F]): Kleisli[F, A, C] =
  Kleisli(AndThen(run).andThen { b =>
    f(b).run(a)
    //   ^^^^^^
    // We've lost `A` here, so this won't work :-(
  })
```

Unless I'm missing something, the answer is no — but it would have been awesome 😞

### Is `AndThen` reusable?

Yes, we might end up using it for fixing the stack safety of other data types too.

### Should we expose `AndThen`?

It's a pretty useful data type, maybe. Currently it's `private[cats]`.

### DIFF
--- a/core/src/main/scala/cats/data/AndThen.scala
+++ b/core/src/main/scala/cats/data/AndThen.scala
@@ -29,11 +29,27 @@ private[cats] sealed abstract class AndThen[-T, +R]
   final def apply(a: T): R =
     runLoop(a)
 
-  override def compose[A](g: A => T): A => R =
-    composeF(AndThen(g))
+  override def andThen[A](g: R => A): T => A = {
+    // Fusing calls up to a certain threshold, using the fusion
+    // technique implemented for `cats.effect.IO#map`
+    this match {
+      case Single(f, index) if index != 127 =>
+        Single(f.andThen(g), index + 1)
+      case _ =>
+        andThenF(Single(g, 0))
+    }
+  }
 
-  override def andThen[A](g: R => A): T => A =
-    andThenF(AndThen(g))
+  override def compose[A](g: A => T): A => R = {
+    // Fusing calls up to a certain threshold, using the fusion
+    // technique implemented for `cats.effect.IO#map`
+    this match {
+      case Single(f, index) if index != 127 =>
+        Single(f.compose(g), index + 1)
+      case _ =>
+        composeF(Single(g, 0))
+    }
+  }
 
   private def runLoop(start: T): R = {
     var self: AndThen[Any, Any] = this.asInstanceOf[AndThen[Any, Any]]
@@ -42,11 +58,11 @@ private[cats] sealed abstract class AndThen[-T, +R]
 
     while (continue) {
       self match {
-        case Single(f) =>
+        case Single(f, _) =>
           current = f(current)
           continue = false
 
-        case Concat(Single(f), right) =>
+        case Concat(Single(f, _), right) =>
           current = f(current)
           self = right.asInstanceOf[AndThen[Any, Any]]
 
@@ -57,9 +73,9 @@ private[cats] sealed abstract class AndThen[-T, +R]
     current.asInstanceOf[R]
   }
 
-  final def andThenF[X](right: AndThen[R, X]): AndThen[T, X] =
+  private final def andThenF[X](right: AndThen[R, X]): AndThen[T, X] =
     Concat(this, right)
-  final def composeF[X](right: AndThen[X, T]): AndThen[X, R] =
+  private final def composeF[X](right: AndThen[X, T]): AndThen[X, R] =
     Concat(right, this)
 
   // converts left-leaning to right-leaning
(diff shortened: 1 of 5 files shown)


core/src/main/scala/cats/data/AndThen.scala
```diff
@@ -0,0 +1,120 @@
+package cats.data
+
+import java.io.Serializable
+
+/**
+ * Internal API (Cats) — A type-aligned seq for representing
```
johnynek: This actually seems like something to move to kernel since it is not higher kinded very useful.

I see it is private, but lots of use cases could exist for an optimized AndThen.

alexandru: It is not very clear to me what the boundaries of Cats's sub-projects and packages are. 

I can move it to cats-kernel if needed, however I only see a bunch of type classes in cats-kernel, whereas I am seeing data types that aren't type classes in `cats.data`.

alexandru:
@johnynek I've added benchmark results, compared with current master, see the description.

alexandru:
Also let me know what you think about `AndThen` — should we make it public and if yes in what package?

core/src/main/scala/cats/data/AndThen.scala
```diff
@@ -0,0 +1,112 @@
...
+      case _ =>
+        andThenF(AndThen(g))
+    }
+  }
+
+  override def compose[A](g: A => T): AndThen[A, R] = {
+    // Fusing calls up to a certain threshold, using the fusion
+    // technique implemented for `cats.effect.IO#map`
+    this match {
+      case Single(f, index) if index != 127 =>
```
kailuowang: Nitpick, can we make this 127 (and at line 32) a constant with a sensible name?

kailuowang:
`AndThen` definitely looks useful. But my vote is to keep internal for a bit before we release it and lock the binary compatibility, (the same approach we are taking with new type)

johnynek:
If we look at the API of AndThen, from a binary compatibility point of view it is just an alias for Function1. I'm not sure how we could even make a change that would break binary compatibility, yet still be an instance of Function1.

Even the apply method on the companion just takes a Function1 and returns and AndThen.

So as long as we keep the subclasses private and all the support methods private, I think we should make it available. If for no other reason than cats-effect can use it at that point.

kailuowang:
@johnynek this API is basically fixed to Function1. What I was worrying about was that, outside of `IndexedStateT`, is this encoding (i.e. extending the `Function1` type), best suited for public usage for stacksafe compose of functions? That being said, I just realized that if there turns out to be a better encoding user can just switch to that one, this `AndThen` is still needed by the `StateT`. So now I am more neutral on whether to open it up.

johnynek:
+1

core/src/main/scala/cats/data/AndThen.scala
alexandru: Thanks for the suggestion. I've updated the code with a documented constant.

kailuowang:
fixed scalastyle issues, will merge after build green.

PR #2194, opened 2018-03-14
Add ensureP (partial function) method to MonadError

Wogan:
Partial functions are a convenient method for expressing `ensure`.

In ApplicativeError `handleError` has another method `recover` which accepts `PartialFunction`. This is similar but for the reverse direction. If you can think of a better name than `ensureP` I'm open to it - perhaps `reject`, `guard` or `guarantee`.


Wogan:
Oh hang on tests still need attention. I will revisit this after I have fixed them.

kailuowang:
Thanks for this. The method seems useful, but I do agree that `ensureP` might give a false impression given the other `ensure` methods. `reject` might be a good choice. 
The other issue is that, due to our commitment to maintaining binary compatibility, we can't add new methods to any public trait. So this new method might have to go to `MonadErrorOps` as a mere syntax extension 
https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/syntax/monadError.scala

Wogan:
Thanks for taking a look. I've renamed it to `reject` and removed it from `MonadError` directly (leaving only the syntax extension.

Re: binary compatibility, mea cupla, I forgot adding a method to a trait isn't safe prior to Scala 2.12.

core/src/main/scala/cats/syntax/monadError.scala
```diff
@@ -16,6 +16,9 @@ final class MonadErrorOps[F[_], E, A](val fa: F[A]) extends AnyVal {
   def ensureOr(error: A => E)(predicate: A => Boolean)(implicit F: MonadError[F, E]): F[A] =
     F.ensureOr(fa)(error)(predicate)
 
+  def reject(pf: PartialFunction[A, E])(implicit F: MonadError[F, E]): F[A] =
+    F.flatMap(fa)(a => if (pf.isDefinedAt(a)) F.raiseError(pf(a)) else F.pure(a))
```
kailuowang: There is [this trick](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Foldable.scala#L210) that avoids calling `isDefinedAt` and then `apply` of the `PartialFunction`. Also, instead of `F.pure(a)` shall we just return `fa`?

Wogan: I'm thinking of doing `F.flatMap(fa)(a => pf.andThen(F.raiseError[A]).applyOrElse(a, _ => fa))`, though that does create a new "AndThen" object. Is avoiding the allocation important?

kailuowang: that might be a good balance between readability and performance.

PR #2197, opened 2018-03-16
Added Magma, and alter Semigroup so it's based on Magma.

ryanstull:
I'm using cats in a project where I'm building abstract syntax trees (ASTs).  I've run into an issue where it would be nice to be able to differentiate between associative operators and non-associative operators, so you could know if you could rearrange the AST.  Currently with Semigroup being the top of the Group-like structure hierarchy, this is not possible as far as I can tell.

This change will break binary compatibility, but I'm putting it forward in response to the [Code review cats-kernel](https://github.com/typelevel/cats/issues/2093) issue.  We might also want to think about adding Quasi-groups and loops into the mix while we're at it, but I do think since Magma is the most abstract Group-like algebraic structure, we should at least add that.

I'm not super familiar with the entire cats codebase so this is more of a work in progress and will probably need a few more things before be able to be merged.  The good thing is for Magma, there aren't any laws since the only thing about it (having a closed binary operation) can be checked by the type system.

### DIFF
--- a/core/src/main/scala/cats/package.scala
+++ b/core/src/main/scala/cats/package.scala
@@ -84,6 +84,7 @@ package object cats {
   type Comparison = cats.kernel.Comparison
   type Order[A] = cats.kernel.Order[A]
   type Hash[A] = cats.kernel.Hash[A]
+  type Magma[A] = cats.kernel.Magma[A]
   type Semigroup[A] = cats.kernel.Semigroup[A]
   type Monoid[A] = cats.kernel.Monoid[A]
   type Group[A] = cats.kernel.Group[A]
@@ -93,6 +94,7 @@ package object cats {
   val Order = cats.kernel.Order
   val Comparison = cats.kernel.Comparison
   val Hash = cats.kernel.Hash
+  val Magma = cats.kernel.Magma
   val Semigroup = cats.kernel.Semigroup
   val Monoid = cats.kernel.Monoid
   val Group = cats.kernel.Group
--- a/kernel/src/main/scala/cats/kernel/Magma.scala
+++ b/kernel/src/main/scala/cats/kernel/Magma.scala
@@ -1,0 +1,88 @@
+package cats.kernel
+
+import scala.{ specialized => sp }
+import scala.annotation.tailrec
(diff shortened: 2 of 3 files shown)


denisrosset:
Structure can still be built on top of `Magma`, in a way that would parallel the associative hierarchy. So even if magmas do not have laws per se, their inclusion closes the hierarchy at the bottom.

There are quasigroups, loops, ..., but I haven't a use case at hand.

We also have morphisms from the structure X to structure Y, i.e. a function `f: X => Y` such that `f` preserves the structure. I'm planning to define (in Spire) a subtype of `Function[X, Y]` that has a type parameter `F[_]` corresponding to the algebraic structure preserved (magma, semigroup, monoid, group, ring, etc...)

For magmas, `f(a) |+| f(b) == f(a |+| b)` is still an important law.

LukaJCB:
I agree that `Magma` itself is fairly useless. However, I did have a use case recently where I needed a commutative operation that turned out not to be associative, so I couldn't create a valid `Semigroup` instance. If we had something like `CommutativeMagma` I could've used that. That said, though, I'm not sure if it's worth adding a superclass to the whole `Monoid`-hierarchy just for these exotic type classes (another one that comes to mind is the medial magma).

denisrosset:
That would enable other people to extend the hierarchy in unforeseen ways, and still be compatible at a basic level.

(You can still do that now by creating a `Magma` instance, and having implicit conversions from `Semigroup`).

johnynek:
+1 to have people define several uses first, then we can reconsider.

I think `type Magma[A] = Function2[A, A, A]` and an implicit from Semigroup to Magma would be fine for users to start.

After we have some interesting magmas and functions we want to generalize on magmas, we can reconsider.

For instance, I guess octonions are non-associative, but I don't see how I can write a generic function over a magma that I want to use for octonions and say complex numbers.

armanbilge:
So the good news is, I no longer think this change would be binary-breaking in modern Scala.

However, due to lack of use-cases coming up since it was opened 4 years ago, seems unlikely to land. I'm going to close for now.

PR #2198, opened 2018-03-16
1.1.0 Post Release 

kailuowang:


AUTHORS.md
```diff
@@ -155,8 +166,10 @@ possible:
  * Richard Miller
  * Rintcius Blok
  * Rob Norris
+ * Rohan Shah
  * Romain Ruetschi
  * Ross A. Baker
+ * rsoeldner
```
m-sp: Seems rsoeldner is also a couple of lines below in the list

kailuowang: thanks good catch

PR #2200, opened 2018-03-17
Add `lefts` and `rights` functions to `Alternative` typeclass

guizmaii:
Hi,

It's my first PR here. 🙂

Would you be interested in adding these two functions ? 
I invented nothing. Scalaz has these functions and, IMHO, they can be useful.

Let me know.

Jules


LukaJCB:
Hi Jules, thanks for creating this PR, unfortunately its not binary compatible, so we'll either have to add them as simple syntax extensions or wait for Cats 2.0.

guizmaii:
> add them as simple syntax extensions

How do you do that ? `implicit class` ??

kailuowang:
@guizmaii you can add them to this class
https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/syntax/alternative.scala#L35

guizmaii:
Hi @LukaJCB and @kailuowang,

I finished my work here. You can review it. Thanks for your help.

It seems that my `git squash` messed up the git history. I can open a new PR if this is a problem.


Jules

core/src/main/scala/cats/syntax/alternative.scala
```diff
@@ -34,6 +34,38 @@ final class UniteOps[F[_], G[_], A](val fga: F[G[A]]) extends AnyVal {
...
+    * Example:
+    * {{{
+    * scala> import cats.implicits._
+    * scala> val l: List[Either[String, Int]] = List(Right(1), Left("error"))
+    * scala> l.lefts
+    * res0: List[String] = List(error)
+    * }}}
+    */
+  def lefts(implicit
+            F: Monad[F],
```
johnynek: You only need F: FlatMap here right?

guizmaii: Yes, it seems. We could make the same remark for `separate`, no ?

peterneyens:
It seems that for this PR it would be useful to get the "left foldable" / "right foldable" from a `Bifoldable` (see #1834).

We can't add `left` and `right` to `Bifoldable` though, so we would probably need something like

```scala
object Bifoldable {
  def left[G[_, _], A](implicit G: Bifoldable[G]): Foldable[G[?, A]] = ???
  def rigth[G[_, _], A](implicit G: Bifoldable[G]): Foldable[G[A, ?]] = ???
}
```

Both `lefts` and `rights` could then be implemented in terms of `unite`, I think:

```scala
def rights(implicit F: Monad[F], A: Alternative[F], G: Bifoldable[G]): F[B] =
  A.unite[G[A, ?], B](fgab)(F, Bifoldable.right[G, A])

def lefts(implicit F: Monad[F], A: Alternative[F], G: Bifoldable[G]): F[A] =
  A.unite[G[?, B], A](fgab)(F, Bifoldable.left[G, B])
```

guizmaii:
Hi @peterneyens,

I should say that I have no opinion on that subject.

Maybe the other speakers in this PR have some ?

kailuowang:
👍 on @peterneyens' suggestion.

guizmaii:
I'm sorry but I don't have the required knowledge to implement what you're asking. 

Maybe @peterneyens, if you have the time to do this, you can continue my work ? or someone else ?

I close the PR.

PR #2203, opened 2018-03-19
Add support for 2.13.0-M3

BennyHill:
But allow to fail on 2.13.0-M3 see 

- https://github.com/scalaz/scalaz/pull/1632
- https://github.com/scalaz/scalaz/issues/1629
- https://github.com/scala/bug/issues/10706
- etc


BennyHill:
Notes on build failures:
- OOM, perhaps needs a restart? https://travis-ci.org/typelevel/cats/jobs/355615298
- ""Synthetic tree ..." Usually seen in coverage with macos, etc, see https://github.com/scala/bug/issues/10706  https://travis-ci.org/typelevel/cats/jobs/355615302
- "ambiguous implicit" see https://github.com/scalaz/scalaz/pull/1632 and https://github.com/scalaz/scalaz/issues/1629  https://travis-ci.org/typelevel/cats/jobs/355615303

BennyHill:
@tpolecat, eperinan For the build to work, I had to explicitly add the `tut` plugin, otherwise it was trying to find the 0.6.2 version, so something is not quite right here

see https://github.com/typelevel/cats/pull/2203/files#diff-a84f91f20f040218bccd09fed4761fb3R16

BennyHill:
@tpolecat, @eperinan , here's a printout : 

> [warn] Found version conflict(s) in library dependencies; some are suspected to be binary incompatible:
[warn]  * org.tpolecat:tut-plugin:0.6.3 is selected over 0.6.2
[warn]      +- default:catalysts-bh-topic-sbt-catalaysts-0-8-5-build:0.1.0-SNAPSHOT (scalaVersion=2.12, sbtVersion=1.0) (depends on 0.6.3)
[warn]      +- com.47deg:sbt-microsites:0.7.17 (scalaVersion=2.12, sbtVersion=1.0) (depends on 0.6.2)

BennyHill:
Rebased on master, as that was blocked

ceedubs:
@BennyHill why is it that this also includes 1.1 release notes, author updates, etc?

BennyHill:
> @BennyHill why is it that this also includes 1.1 release notes, author updates, etc?

I rebased on master. see before - merge was blocked

ceedubs:
Hmm if the changes were just brought in from master then they shouldn't be showing up in the GitHub file diff though, should they? I'm a little confused as to what's going on here.

BennyHill:
Ok, I'll do a squash in the morning to fix them up

BennyHill:
Hmm, looks the same in https://github.com/typelevel/cats/pull/2200 ?

BennyHill:
Done, ready for review

build.sbt
```diff
@@ -335,8 +346,16 @@ lazy val testkit = crossProject.crossType(CrossType.Pure)
...
-  .settings(
-    libraryDependencies += "org.scalatest" %%% "scalatest" % scalaTestVersion)
+  // 2.13.0-M3 workaround
+  //.settings(libraryDependencies += "org.scalatest" %%% "scalatest" % scalaTestVersion)
+  .settings(libraryDependencies += {
+    CrossVersion.partialVersion(scalaVersion.value) match {
+      case Some((2, 13)) =>
+        "org.scalatest" %%% "scalatest" % scalatest_2_13
+      case _ =>
+       "org.scalatest" %%% "scalatest" % scalaTestVersion
```
ceedubs: This logic is duplicated in two places. Should we create a `def scalatestVersion(scalaVersion: String): String` or something?

.travis.yml
```diff
@@ -13,6 +13,11 @@ scala:
   - 2.10.7
   - 2.11.12
   - 2.12.5
+  - 2.13.0-M3
+
+matrix:
+  allow_failures:
```
ceedubs: It might be helpful to leave a comment here about why this is happening.

build.sbt
BennyHill: They are not quite the same, the first has a the `test` scope. A "real" function looks more like [this](https://github.com/typelevel/sbt-catalysts/blob/master/src/main/scala/sbtcatalysts/CatalystsPlugin.scala#L108-L123) or [as in catalysts](https://github.com/typelevel/catalysts/blob/master/build.sbt#L298-L327). So I opted for a comment instead  i,e `// 2.13.0-M3 workaround`

So as this is commented as an interim workaround, is this OK for now?

.travis.yml
BennyHill: no problem, I'll add a comment

build.sbt
BennyHill: I can also add a link to the comment describing this interim workaround - https://github.com/scalatest/scalatest/issues/1321#issuecomment-365140680

ceedubs: @BennyHill I opened a PR against your repo with the workaround that I was thinking might clean things up a little bit. Feel free to merge if you like it: https://github.com/BennyHill/cats/pull/1

build.sbt
```diff
@@ -401,7 +420,7 @@ lazy val bench = project.dependsOn(macrosJVM, coreJVM, freeJVM, lawsJVM)
   .settings(commonJvmSettings)
   .settings(coverageEnabled := false)
   .settings(libraryDependencies ++= Seq(
-    "org.scalaz" %% "scalaz-core" % "7.2.15"))
+    "org.scalaz" %% "scalaz-core" % "7.2.19"))
```
barambani: `7.2.20` could be used here. Very minor though

BennyHill: Yep, small point but the small points tend to add up over time. Besides, might as well add this whilst doing other changes - thx for catching this :+1:

barambani: Thank you 👍

PR #2205, opened 2018-03-20
Use helper addSbtCoursier in project/plugins.sbt

BennyHill:
ref https://gitter.im/typelevel/cats-dev?at=5ab0392be4d1c636041b5183


kailuowang:
I was assuming that this would allow the sbt plugins and their dependencies to be loaded using coursier as well?

BennyHill:
> I was assuming that this would allow the sbt plugins and their dependencies to be loaded using coursier as well?

No idea, tbh :) I thought I would just PR the change and see if there is any difference. Also, the comments in the chat will be lost otherwise.

BennyHill:
See commit comment https://github.com/coursier/coursier/commit/a743f1183e6412c205b17d52f7bf1d6e3cd09161

Issue #1374, opened 2016-09-14
Use polymorphic lambdas


adelbertc:
https://github.com/non/kind-projector/tree/v0.9.0#polymorphic-lambda-values

PR #2213, opened 2018-03-25
Use poly lambda syntax in FreeApplicative test

ceedubs:
This is the last explicit `new FunctionK` that I saw in the code base
when I looked to close out #1374.


peterneyens:
There are a few more using `~>` in [`CofreeSuite`](https://github.com/typelevel/cats/blob/master/free/src/test/scala/cats/free/CofreeSuite.scala#L154-L167).

ceedubs:
Good catch @peterneyens. Updated.

peterneyens:
Sorry for not linking to them in my previous comment, but there are also `toList` and `inclusion` in `CofreeSuite`. (`inclusion` could be `OptionT.liftK[Eval]` I think).

ceedubs:
@peterneyens yet another good catch. Updated again.

PR #2214, opened 2018-03-25
Check MonadTests for Vector and List

ceedubs:
This is surprising to me, so I may be missing something, but as far as I
can tell, we weren't running the MonadTests for Vector and List.


ceedubs:
Indeed the way that I discovered this was by noticing that `tailRecM` for `Vector` didn't have coverage.

PR #2217, opened 2018-03-25
Add dual categories.

sellout:


LukaJCB:
This looks great to me, thanks! One thing to consider is if we want to `newtype` this instead of wrapping like we did for `NonEmptySet` and `NonEmptyMap`?

PR #2218, opened 2018-03-27
Remove extra 'a' in eithert.md

donaldosalas:


LukaJCB:
Thanks, good catch! :)

Issue #2139, opened 2018-01-04
Update CONTRIBUTING.md


kailuowang:
- [ ]  In the write test section, link to https://typelevel.org/cats/typeclasses/lawtesting.html
- [ ] add a step of checking binary compatible: `++2.11.12` and then `mimaReportBinaryIssues`
- [ ] in the write code section, copy/consolidate the discussion from the linked issues to [Guidelines.md](https://github.com/typelevel/cats/blob/master/docs/src/main/tut/guidelines.md) and then link to it. 
- [ ] remove the "Publishing the site to github" section, this is done by sbt-microsite now. 
- [ ] The doc suggested to use the Waffle.io to find issues to work on, should replace with finding issue with the low-hanging-fruit or help-wanted tag 
- [ ] in the what goes in cats section also introduce mouse, cats-effect and cats-mtl
Not sure if I captured all the outdated bits. But these should be make it by and large update-to-date.

AlejandroME:
@kailuowang Can I take this?

kailuowang:
of course. @AlejandroME Thanks!

Rlucas12:
Is this done ?

AlejandroME:
@Rlucas12 WIP, pushing the first draft between today and tomorrow.

PR #2221, opened 2018-04-05
Update contributors guide

AlejandroME:
This is the first draft of the suggestions proposed by @kailuowang in #2139.

I have to say that I'm not entirely sure of the _Write Code_ section, because I'm not totally familiar with some concepts mentioned in the related issues. So, if there are some mistakes, please help me identify them for fixing.

Thanks!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,11 +33,10 @@ skip these steps and jump straight to submitting a pull request.
 
 ### Find something that belongs in cats
 
-Looking for a way that you can help out? Check out our
-[Waffle.io page](https://waffle.io/typelevel/cats). Choose a card from the
-"Ready" column. Before you start working on it, make sure that it's
-not already assigned to someone and that nobody has left a comment
-saying that they are working on it!
+Looking for a way that you can help out? Check out the [open issues]
+(https://github.com/typelevel/cats/issues). Look for issues tagged as _**help wanted**_ oe _**low-hanging fruit**_. These ones are the easiest way to start contributing, but if you find other issues that catch your eye, you're most welcome to tackle them!
+
+Make sure that it's not already assigned to someone and that nobody has left a comment saying that they are working on it!
 
 (Of course, you can also comment on an issue someone is already
 working on and offer to collaborate.)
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -154,10 +173,23 @@ with [Discipline](https://github.com/typelevel/discipline) for law checking, and
...
+It is important to verify that the feature you are implementing is compatible with Scala 2.11.x and Scala 2.12.x (Scala <2.10.x is not supported). When you submit a PR, Travis makes this check, but it is time-expensive, so you can assure this step beforehand by issuing the command `+2.11.12`, which sets the cats' Scala version to `2.11.12` and then compile the project.
+
+A summary of these steps is as follows:
+
+```
+$ sbt
+> +2.11.12
+> compile
+```
+If you see errors in the compilation process related to the Scala version, you can execute the command `mimaReportBinaryIssues` to see what happened.
```
kailuowang: binary breaking changes won't necessarily cause compilation errors.  So, instead of `compile` run `mimaReportBinaryIssues` should report the binary breaking issues.

CONTRIBUTING.md
```diff
@@ -115,15 +123,26 @@ builds:
 
 ### Write code
 
-TODO
+Here are some suggestions for you while implementing your feature...
```
kailuowang: Type class is well documented in the typeclass.md, I'd say for now just remove the content of this section and provide a link to the https://typelevel.org/cats/guidelines.html
e.g. 
```markdown
[See guidelines](/cats/guidelines.html)
```

CONTRIBUTING.md
```diff
@@ -196,13 +228,6 @@ which appears in the documentation, this ensures us that our examples
...
 staying up-to-date.
 
-### Publishing the site to github.
-
-The `git.remoteRepo` variable in `docs/build.sbt` controls which
-repository you will push to. Ensure that this variable points to a
-repo you wish to push to, and that it is one for which you have push
-access, then run `sbt ghpagesPushSite`
-
 ### Write examples
```
kailuowang: I guess we can write here: 
One of the best ways to provide examples is doctest, here is an [example of doctest examples in cats](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Functor.scala#L19-L33).

kailuowang:
Thanks so much! I left some minor comments.

AlejandroME:
Thank you @kailuowang! I'll address your observations this weekend.

AlejandroME:
@kailuowang I've fixed the sections based on your comments. Please let me know if there's something else.

Thanks!

PR #2225, opened 2018-04-12
Add doctest example for groupByNel

ceedubs:
Mostly stolen from the `groupBy` example on `NonEmptyList`.


kailuowang:
This is a handy one. Thanks for adding the doc @ceedubs !

PR #2231, opened 2018-04-17
Add ErrorControl

### LINKED ISSUE #2229: Add a type class for principled error handling
We currently have `ApplicativeError` and `MonadError`, but as I argue in [this blogpost](https://typelevel.org/blog/2018/04/13/rethinking-monaderror.html) it has some limitations that I'd like to overcome. I'd be happy to lead the charge on this if no one else steps up and I'd be even more happy to get some help on naming things :)

LukaJCB:
Should fix #2229 
Bikeshedding welcome :)


tests/src/test/scala/cats/tests/KleisliSuite.scala
```diff
@@ -39,13 +39,26 @@ class KleisliSuite extends CatsSuite {
...
 
   checkAll("Kleisli[Id, Int, ?]", CommutativeMonadTests[Kleisli[Id, Int, ?]].commutativeMonad[Int, Int, Int])
-  checkAll("CommutativeMonad[Kleisli[Id, Int, ?]]",SerializableTests.serializable(CommutativeMonad[Kleisli[Id, Int, ?]]))
+  checkAll("CommutativeMonad[Kleisli[Id, Int, ?]]", SerializableTests.serializable(CommutativeMonad[Kleisli[Id, Int, ?]]))
+
+  //TODO: Why can't type inference figure this out?
+  implicit val x: ErrorControl[Kleisli[Option, Int, ?], Kleisli[Id, Int, ?], Unit] =
+    Kleisli.catsErrorControlForKleisli[Option, Id, Int, Unit]
+
+  checkAll("Kleisli[Option, Int, ?]", ErrorControlTests[Kleisli[Option, Int, ?], Kleisli[Id, Int, ?], Unit].errorControl[Int])
```
LukaJCB: For some reason Scala can't seem to find an implicit `ErrorControl[Kleisli[Option, Int, ?], Kleisli[Id, Int, ?], Unit]` even though it is defined in the companion object of `Kleisli`. Anyone know a way to help scalac here?

kailuowang: what if you call `Kleisli.catsErrorControlForKleisli[Option, Id, Int, Unit]` directly? does it complain that couldn't find `ErrorControl[Option, Id, Unit]`?

core/src/main/scala/cats/data/IndexedStateT.scala
```diff
@@ -247,6 +246,23 @@ private[data] sealed abstract class IndexedStateTInstances extends IndexedStateT
   implicit def catsDataAlternativeForIndexedStateT[F[_], S](implicit FM: Monad[F],
     FA: Alternative[F]): Alternative[IndexedStateT[F, S, S, ?]] with Monad[IndexedStateT[F, S, S, ?]] =
     new IndexedStateTAlternative[F, S] { implicit def F = FM; implicit def G = FA }
+
+  implicit def catsErrorControlForStateT[F[_], G[_], S, E]
+  (implicit E: ErrorControl[F, G, E], M: Monad[G]): ErrorControl[StateT[F, S, ?], StateT[G, S, ?], E] =
```
kailuowang: `M: Monad[G]` no longer required since it's provided by `ErrorControl` right?

LukaJCB: Yep, will fix, this was needed when G was an Applicative so that `StateT[G, S, A]` could be an Applicative

tests/src/test/scala/cats/tests/KleisliSuite.scala
LukaJCB: Yeah it won't compile without it :/

kailuowang: hmm, since StateT has exactly the same shape but passes with an implicit Monad[G], I wonder what if you add `Monad[G]` to `Kleisli` might help? 
Generally, when I get into this type of error (the implicit method is available and legit but scalac refuses to find it) I'll just throw in a `Lazy` and see if it helps. but that's not an option here. And I suspect that it's not the same scalac bug.

LukaJCB: Nope, I needed to add the extra implicit to WriterT and StateT as well, so it's the same problem.

kailuowang: (╯°□°)╯︵ ┻━┻  time! 
Just to experiment, what if add them to the companion of ErrorControl?

core/src/main/scala/cats/ErrorControl.scala
```diff
@@ -0,0 +1,38 @@
+package cats
+
+import cats.data.EitherT
+
+trait ErrorControl[F[_], G[_], E] extends Serializable {
+  val monadErrorF: MonadError[F, E]
```
johnynek: I don't like using abstract vals on traits since it invites the val init order bug. Can we make these def?

core/src/main/scala/cats/ErrorControl.scala
```diff
@@ -0,0 +1,38 @@
+package cats
+
+import cats.data.EitherT
+
+trait ErrorControl[F[_], G[_], E] extends Serializable {
```
johnynek: I like to use abstract types as "output types". It seems to me we might want `G[_]` to be an abstract type here, since we might often want to write algorithms to work in F go through G, but return an F. Then we could accept an implicit `ErrorControl[F, E]` to do that, which might be nice.

core/src/main/scala/cats/data/OptionT.scala
```diff
@@ -217,6 +217,22 @@ private[data] sealed abstract class OptionTInstances extends OptionTInstances0 {
   implicit def catsDataMonadForOptionT[F[_]](implicit F0: Monad[F]): Monad[OptionT[F, ?]] =
     new OptionTMonad[F] { implicit val F = F0 }
 
+  implicit def catsEndeavorForOptionT[F[_]: Monad, E]: ErrorControl[OptionT[F, ?], F, Unit] =
```
johnynek: should this be catsErrorControlForOptionT?

oleg-py:
> The EitherEff "transformer" would require all existing code using IO or Task to completely change all of their code, whereas this will be a strictly backwards compatible change. 

Anybody who wants to take advantage of new abstraction would have to change their code. I don't see how bifunctor version would break anything (except if we remove MonadError too, which I didn't propose to do).

>Nothing stops us from adding additional bifunctorial error handling abstractions or bifunctorial IO types.

Yes. It's a question whether we should do stuff, not whether we could. Adding something to cats might be easy, but removing it is likely to be nearly impossible. And I do quite like the minimalism in cats API. So I'd rather not add stuff if it will become obsolete (like I said, I'm not sure what approach is the best, so that is a possibility).

>  `F[_], G[_]` is just 3 characters longer than `F[_, _]`

It's more like `[F[_, _] : ErrorControl]` vs `[F[_], G[_]](implicit F: ErrorControl[F, G, _])`

core/src/main/scala/cats/ErrorControl.scala
```diff
@@ -0,0 +1,38 @@
...
+  val monadErrorF: MonadError[F, E]
+  val monadG: Monad[G]
+
+  def controlError[A](fa: F[A])(f: E => G[A]): G[A]
+
```
johnynek: seems like a `def control[A,B](fa: F[A])(f: Either[E, A] => G[B]): G[B]` might be a nice function to have. It is more general than controlError and can also be used to implement trial. I think users might often want this variant personally.

The name might not be great, but I would love to see this and have syntax for it.

core/src/main/scala/cats/ErrorControl.scala
LukaJCB: I like that as well, see my recent issue in parallel, though I'm not 100% sure if it applies here, have to think about it more.

jdegoes:
Given @LukaJCB's cool adapter ("@jdegoes transformer" 😆), I personally prefer the much simpler:

```scala
trait Fallible[B[_, _]] extends Bifunctor[B] {
  def applicative[E]: Applicative[B[E, ?]]
  
  def attempt[E1, E2, A](b: B[E1, A]): B[E2, Either[E1, A]]
  
  def fail[E, A](e: E): B[E, A]
}
```

This is a sufficient basis to implement everything else, and has very precise types and laws.

The only disadvantage, if you consider it such, is that it requires you use an adapter in order to use the type class.

core/src/main/scala/cats/ErrorControl.scala
LukaJCB: I don't think we'd be able to express these https://github.com/typelevel/cats/pull/2231/files#diff-aadaecd5f11de32b7cdacd702c9a5393R38 maybe you know a way?

LukaJCB:
> Anybody who wants to take advantage of new abstraction would have to change their code. I don't see how bifunctor version would break anything (except if we remove MonadError too, which I didn't propose to do).

Right, I don't disagree here, though with `ErrorControl` you can keep using `IO[A]` and make use of more principled error handling by making use of the type class to get to `UIO[A]`. I've been using this for some time and it works well for me. People can keep using their `IO`s and `Task`s, without changing anything. Whereas once you go to `IO[E, A]` as your base type you have to change a LOT. Also bifunctor `IO` seems super promising and I really like the idea, but no one has actually had any real experience with it in practice yet AFAIK.

For existing applications I think this is the better way forward. Cats-1.0.0-RC1 was just released and we aren't going to change the nature of `IO` anytime soon. 

> Yes. It's a question whether we should do stuff, not whether we could. Adding something to cats might be easy, but removing it is likely to be nearly impossible. And I do quite like the minimalism in cats API. So I'd rather not add stuff if it will become obsolete (like I said, I'm not sure what approach is the best, so that is a possibility).

Again, I agree in principal, but I see these two as rather orthogonal (kind of like how `Bifunctor` and `Functor` are orthogonal). I think there's room in the future to add a type class like you're suggesting, without either becoming obsolete.

oleg-py:
> Whereas once you go to `IO[E, A]`

I'm ***not*** proposing to change all existing effect types, it is probably not feasible with current ecosystem state and b/c the best approach is not yet set in stone. Newtyping would be the way to go here, so we will have e.g. `IO[A]` and `StrictIO[E, A]` versus `IO[A]` and `EitherT[UIO, E, A]`.

We will also be able to provide e.g. a `StrictFuture[E, A]` using the same machinery for impure folks :)

It's also more general than unexceptional types. To represent lack of errors, you can use `StrictIO[Nothing, A]` (well, you probably need an alias for `Nothing` for scalac to not complain). To represent possibility of *some* errors with `UIO`, you need `Either`.

> no one has actually had any real experience with it in practice yet AFAIK.

Similar can be said about three-parameter version of `ErrorControl` :) It might be worth just to back off a bit and experiment with different options we have.

LukaJCB:
> I'm not proposing to change all existing effect types, it is probably not feasible with current ecosystem state and b/c the best approach is not yet set in stone. Newtyping would be the way to go here, so we will have e.g. IO[A] and StrictIO[E, A] versus IO[A] and EitherT[UIO, E, A].

👍 Fully agree here.

> It's also more general than unexceptional types. To represent lack of errors, you can use StrictIO[Nothing, A] (well, you probably need an alias for Nothing for scalac to not complain). To represent possibility of some errors with UIO, you need Either.

I agree here as well, however I'd like to argue `UIO` is a special case because it's completely error free. I see it very similar to how `NonEmptyList` is a special case of `List`, both are special because they give very specific guarantees, never halting execution for `UIO`, and having atleast one element for `NonEmptyList`. Both have more general solutions in `BIO[Nothing, A]` and `shapeless.Sized[L <: Nat]`, but they're still very useful because they encode very specific invariants. 
I certainly believe that even if we had full dependent types, something like the `Reducible` or `NonEmptyTraverse` type classes or a `SizedList[> 0]` would be still be very valuable, because of their guaranteed invariants.

> Similar can be said about three-parameter version of ErrorControl :) It might be worth just to back off a bit and experiment with different options we have.

That's fair, I've been experimenting with both approaches for some time now, and concluded for myself that existing code bases are probably better of leveraging unexceptional types, which provide more interoperatability with existing code. 

Of course, that's only my opinion and my experimentation has only been going on for a couple of months, so my conclusion might still be a bit premature. :)

I would certainly love to hear more input from others.

LukaJCB:
On another note, I think we could split this into an `ApplicativeError <-> Applicative` and `MonadError <-> Monad`, but I'm not sure it's worth it.

core/src/main/scala/cats/ErrorControl.scala
```diff
@@ -0,0 +1,98 @@
...
+  def intercept[A](fa: F[A])(f: E => A): G[A] =
+    controlError(fa)(f andThen monadG.pure)
+
+  def absolve[A](gea: G[Either[E, A]]): F[A] =
+    monadErrorF.flatMap(accept(gea))(_.fold(monadErrorF.raiseError, monadErrorF.pure))
+
+  def assure[A](ga: G[A])(error: => E)(predicate: A => Boolean): F[A] =
+    assureOr(ga)(_ => error)(predicate)
+
+  def assureOr[A](ga: G[A])(error: A => E)(predicate: A => Boolean): F[A] =
```
johnynek: seems to me `A => Option[E]` rather than two functions is a nicer way to go. We also only have one invocation. using two functions in one call is usually not super ergonomic in scala.

LukaJCB: I assume you mean `Option[A] => E`? I think we could just get rid of what's currently `assure`, as `assureOr` covers all the use cases. I just kept both for the sake of symmetry in `ensure` and `ensureOr`. I'd rather just have a single function `assure` that takes `A => E` like in the blog post.

wedens:
There was a typeclass [`TransLift`](https://github.com/typelevel/cats/blob/v0.9.0/core/src/main/scala/cats/TransLift.scala) in cats with an abstract member. It required some utterly horrible machinery for syntax: https://github.com/typelevel/cats/blob/v0.9.0/core/src/main/scala/cats/syntax/transLift.scala.

LukaJCB:
Thanks for that link @wedens! 

Another problem is that unlike `Parallel`, you'll often want to access the `G[_]` value:

```scala
def foo[F[_]](fi: F[Int])(implicit EC: ErrorControl[F, String]): EC.G[Int] =
    F.intercept(fi)(_ => 0)
```

kailuowang:
If we do make `G` an abstract, then we probably want to provide the `ErrorControl.Aux` as well. 
```scala
object ErrorControl {
   type Aux[F[_], G0[_], E] = ErrorControl[F, E] { type G = G0 } 
}
```
so that user can write 
```scala
def foo[F[_], G[_]](fi: F[Int])(implicit EC: ErrorControl.Aux[F, G, String]): G[Int] =
```
Then user can basically pick to use this `Aux` or `E.G` whichever fits in the situation.
If `G` is uniquely determined by `F` then abstract type with Aux makes sense. You can get some benefits as @johnynek pointed out. 

The syntax machinery in Translift is a bit horrifying but that's a different kind. `((* - > *) -> * ) -> (* -> *) -> * -> *`, v.s. `( * -> * ) -> (* -> *) -> * -> *` .  I speculate that `ErrorControl.Aux` would be easier to provide syntax to `G[_]`.

LukaJCB:
Right that makes sense, thank you @kailuowang! 

Though now I'm more leaning towards keeping the two type constructors, as it doesn't seem to provide a large benefit and adds complexity for people not familiar with the `Aux` pattern or path dependent types.

LukaJCB:
Anyone have any additional comments? I don't want to rush this, but interested in more feedback :)

kailuowang:
TBH, I'm still not sure how I feel about adding this to `cats.core`.  More discussion would help. The abstraction itself is sound and meaningful, what I don't know is whether practical benefit justifies adding it to Cats. 
The common strategy in FP I've seen so far is to have the effects being handled at the edge. This abstraction allows users to completely remove the error effect in the middle of an application. What are the scenarios in which can we handle *ALL* errors in the middle of an application?  It seems to me that it must satisfy the following conditions:

1. the result data is absolutely optional for further computation, either an optional side effect or Option of something, i.e. F[Option[A]] or F[Unit].
2. More importantly, we know for sure that whatever error occurs in this scenario, it's exactly the same domain meaning as getting an empty result. 

And as soon as the computation involves a sub-computation that does not satisfy the conditions, the whole computation has to maintain the error effect. So it seems to me that the scope of this abstraction would be rather limited within an application. 

A real-world application example would really shed more light on this. Without it, IMO, it would be hard to convince people that this is actually more practical than `BIO`

kubukoz:
I, for one, would *love* to revive this PR. A useful scenario is having `F`=`EitherT[IO, NEL[E], ?]` and `G`=`IO`. You define your http4s router for `G`, handling errors from `F`, e.g. by doing `controlError(errors => BadRequest(errors.asJson))`.

This would make it much easier to work with sealed error ADTs that aren't also Throwables (while working with single-functor IO). While `IO` can still fail, Throwables can be handled on another layer.

LukaJCB:
I'm still quite fond of this little type class, though I admit it's not that strong an abstraction. But from my time with using `UIO` and `IO`, it came in handy :)

kubukoz:
WDYT about publishing it as a small separate lib for now? Just so we don't have to commit to having this in cats in the long term.

kailuowang:
I am +1 on experimenting it with a separate lib or module.

LukaJCB:
An extra module sounds good :+1: though I probably won't have any time to do it anytime soon :(

kubukoz:
I created a repo for this, we can transfer to you later @LukaJCB, hope you don't mind :) I'll try to publish some time this week. [error-control](https://github.com/kubukoz/error-control)

LukaJCB:
Not at all, thanks for running with it :)

PR #2235, opened 2018-04-19
add .option postfix syntax

danbills:


danbills:
Hi @LukaJCB , yes the difference is the treatment of null values, and that is in fact the use case that we worry about.

I'm intrigued by the phrase "justify the extra implicit syntax."  Does implicit syntax impose a cost other than perhaps a cognitive load?

LukaJCB:
It does cost us by "polluting" the implicit scope. If we add this, every user that imports all syntax or all implicits is now going to get `option` suggested as auto-complete for every single value (as this works on a generic `A`). The scala compiler also has to look through all implicits in scope to find what it needs. Overall one method is not going to move the needle much, but accumulated it does make a fairly large difference, so I'm inclined to say no to this change, as it doesn't add much beyond `.some` and is easily replaced by `Option(a)`, which is just one extra keystroke.

johnynek:
I'm -1 on putting almost any implicit on `A` for all `A`. This leads to weird behaviors when the methods collide with existing methods and users get the types wrong.

I'm especially -1 on doing this to deal with null, which we usually discourage very strongly. Making it more ergonomic seems like a low priority.

like `.some.flatMap(Option(_))` does the same thing for the cases people think they may have null. Or they can just stop and capture the value in a val and `Option` that val.

danbills:
So be it.  Thanks for your input.

PR #2237, opened 2018-04-23
Fix #2161, adding MonadError.redeemWith, enhance EitherT

### LINKED ISSUE #2161: Proposal: MonadError.transformWith
So `MonadError` has `attempt` and `handleErrorWith` and for many use-cases one has to introduce both `map` / `flatMap` transformations and `handleErrorWith` handlers in the same "frame".

E.g. what happens for data types implementing `cats-effect` like `cats.effect.IO` and `monix.eval.Task` is something like this

```scala
fa.attempt.flatMap {
  case Right(next) => continueLoop(next)
  case Left(error) => stopLoop(error)
}
```

This is more or less the equivalent of this pattern:

```scala
var streamError = true
try {
  val next = fa()
  streamError = false
  continueLoop(next)
} catch {
  case error if streamError => 
    stopLoop(error)
}
```

Well the problem is that usage of either `attempt` or `handleErrorWith` involves 2 `flatMap` (bind) frames instead of one. And as we all know, the performance of `IO` / `Task` data types, compared with plain function calls, is terrible.

The proposal is to introduce an alternative to `attempt.flatMap` like so:

```scala
def transformWith[A, B](fa: F[A])(bind: A => F[B], recover: Throwable => F[B]): F[B]
```

Just did.a fast [benchmark](https://gist.github.com/alexandru/603e8003fa8da45d5d37f58bf928cb0f) to show what the difference can be:

```
attempt             10000  thrpt   10  2383.878 ±  46.546  ops/s
onErrorHandleWith   10000  thrpt   10  2813.948 ± 176.036  ops/s
transformWith       10000  thrpt   10  4115.213 ±  40.821  ops/s
```

Precedents:

1. [Task.transformWith](https://monix.io/api/2.3/monix/eval/Task.html#transformWith[R](fa:A=%3Emonix.eval.Task[R],fe:Throwable=%3Emonix.eval.Task[R]):monix.eval.Task[R])
2. [Future.transformWith](http://www.scala-lang.org/api/2.12.3/scala/concurrent/Future.html#transformWith[S](f:scala.util.Try[T]=%3Escala.concurrent.Future[S])(implicitexecutor:scala.concurrent.ExecutionContext):scala.concurrent.Future[S])

Note — I'd prefer the version with the two function parameters, because I'd rather avoid the extra `Either` boxing. 

Implementation notes:

1. `cats.effect.IO` is already optimized for this operation
2. `monix.eval.Task` is already optimized for this operation

Basically I'd like to not feel bad about adding error handling.

alexandru:
Addresses #2161 by adding `redeem` and `redeemWith` to `MonadError`.

These are derived by `attempt.map` and `attempt.flatMap` respectively:

```scala
trait MonadError[F[_], E] {
  // ...

  def redeem[A, B](fa: F[A])(recover: E => B, map: A => B): F[B] =
    redeemWith(fa)(recover.andThen(pure), map.andThen(pure))

  def redeemWith[A, B](fa: F[A])(recover: E => F[B], bind: A => F[B]): F[B] =
    flatMap(attempt(fa))(_.fold(recover, bind))
}
```

As mentioned in #2161, this represents an optimization on usage of `attempt` because for certain datatypes (e.g. `IO`, `Task`, `Future`):

1. it avoids the `Either` boxing
2. and even more importantly, it avoids a second `map` / `flatMap` operation

The benchmark results in that issue speak for themselves and note that I've measured `IO` / `Task`, which have fairly efficient `map` and `flatMap` operations. For a data type like `Future` working with `attempt` for error recovery is a disaster.

My use-case for `redeem` / `redeemWith` is in `monix.tail.Iterant` that now makes heavy use of `attempt` for error recovery. This is a streaming abstraction, so `attempt.flatMap` operations can pile up and add quite significant overhead.

---

Given that we have `Future` instances provided by Cats, it would have been a shame to not also provide optimized `redeemWith` and `attempt` operations — because a new, abstract `Future#transformWith` has been introduced in Scala 2.12 as the base function for all other operations (e.g.`flatMap`, `recoverWith`).

The difficulty here is to keep compatibility with Scala 2.11- so this commit introduces `FutureShims` provided with different implementations for different Scala versions. Compatibility is thus maintained even with Scala 2.10.

---

I also went ahead and enhanced `EitherT` with the following operations, straight on the class definition itself:

- `widen`
- `leftWiden`
- `attempt`
- `attemptF`
- `handleError`
- `handleErrorF`
- `handleErrorWith`
- `handleErrorWithF`
- `recoverF`
- `recoverWith`
- `recoverWithF`
- `redeem`
- `redeemF`
- `redeemWith`
- `redeemWithF`

So given `EitherT[F, A, B]`, the functions suffixed with `F` operate on any `F: MonadError[F, E]` in scope, thus recovering from `E` errors thrown in the `F[_]` context.

Two reasons for it:

1. the `F` suffixed methods are provided because working with the provided `MonadError[EitherT[F, L, ?], E]` instance sucks (see #2029)
2. the call sites are too heavy for such common operations, so relying on implicit conversions to provide the necessary syntax creates short-term junk
3. discovery of features for new users is better this way

On performance, the JVM is able to do some limited escape analysis to avoid short-term allocations in certain situations and the garbage collector is usually good at dealing with short-term junk, but irregardless, the JVM has a limited budget for optimizations and the way to think about it is that obvious optimizations at some call sites detract from doing obvious optimizations at other call-sites.

--- 

In order to optimize memory allocations and the byte-code of those functions, I've rewritten some of them to do more explicit pattern matching, with a default `cast _ =>` branch. 

Benchmarks are a marketing scheme and an implementation that's easier on memory and possibly performance beats an implementation that's elegant. Libraries don't have to be elegant, libraries have to be efficient such that the user's code can be elegant.

Have also written ScalaDocs — which triggered the Scalastyle check for "maximum file size", which I had to disable completely, since it makes no sense to have a restriction that discourages the writing of ScalaDocs.

--- 

Unfortunately using a provided `MonadError[EitherT[F, L, ?], E]` that's explicitly imported in scope no longer works for those functions.

E.g. this sample will trigger an error, hence that particular test is now gone:

```scala
test("#2022 EitherT syntax no long works the old way") {
  import data._

  EitherT.right[String](Option(1)).handleErrorWith((_: String) => EitherT.pure(2))
  {
    implicit val me = MonadError[EitherT[Option, String, ?], Unit]
    // Triggers compile-time error!
    EitherT.right[String](Option(1)).handleErrorWith((_: Unit) => EitherT.pure(2))
  }
}
```

This is one reason for why I don't like type-class driven syntax. Such overloads should not be possible or desirable, being much worse than Java-driven overloads which I also prefer to avoid.

Another argument for doing this is that the previous implementation had inconsistent definitions for `recover` and `recoverWith` straight on the `case class` definition, but not for `handleError` and `handleErrorWith`. So you could do the above for `handleError`, but not for `recover`.

This means that this PR has a source incompatibility with 1.0.0, but this is not a binary incompatibility 😉 and is in my opinion a valid fix.

---

We've got the following incompatibilities:

```scala
// Internal classes, moved due to refactoring, not a problem
exclude[MissingClassProblem]("cats.syntax.EitherUtil$"),
exclude[MissingClassProblem]("cats.syntax.EitherUtil"),
// Scala 2.11 only — internal classes, not a problem
exclude[ReversedMissingMethodProblem]("cats.data.EitherTMonadErrorF.redeem"),
exclude[ReversedMissingMethodProblem]("cats.data.EitherTMonadErrorF.redeemWith"),
exclude[ReversedMissingMethodProblem]("cats.data.EitherTMonadErrorF.handleError"),
exclude[ReversedMissingMethodProblem]("cats.data.EitherTMonadErrorF.attempt"),
exclude[ReversedMissingMethodProblem]("cats.data.EitherTMonadError.redeem"),
exclude[ReversedMissingMethodProblem]("cats.data.EitherTMonadError.redeemWith"),
// BREAKAGE — Scala 2.11 only!
exclude[ReversedMissingMethodProblem]("cats.MonadError.redeem"),
exclude[ReversedMissingMethodProblem]("cats.MonadError.redeemWith")
```

Don't know what the current binary compatibility policy is, but due to adding methods with default implementations on traits, these are not backward compatible on Scala 2.11, because it targets Java 7, which doesn't support interfaces with "default implementations".

Scala 2.12 doesn't have such problems though.

But I'm guessing that due to this change, such a PR can only target Cats 2.0, is that correct?

### DIFF
--- a/core/src/main/scala/cats/data/EitherT.scala
+++ b/core/src/main/scala/cats/data/EitherT.scala
@@ -33,11 +33,11 @@ import cats.internals.EitherUtil.{eitherCast, rightBox, leftBox}
  *
  * Note that in this function the error type is part of the signature and
  * is more specific than the customary `Throwable` or `Exception`. You can
- * always "upcast" its error type to `Throwable` later using [[upcastL]]:
+ * always "upcast" its error type to `Throwable` later using [[leftWiden]]:
  *
  * {{{
  *   val num: EitherT[Eval, Throwable, Long] =
- *     parseNum("10210", 10).upcastL
+ *     parseNum("10210", 10).leftWiden
  * }}}
  *
  * The power of `EitherT` is that it combines `F[_]` with the `Either`
@@ -163,12 +163,12 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
    *     EitherT.rightT(10210)
    *
    *   val num2: EitherT[Eval, Throwable, Long] =
-   *     num.upcastL
+   *     num.leftWiden
    * }}}
    *
-   * @see [[upcastR]]
+   * @see [[rightWiden]]
    */
-  @inline def upcastL[AA >: A]: EitherT[F, AA, B] =
+  @inline def leftWiden[AA >: A]: EitherT[F, AA, B] =
     this.asInstanceOf[EitherT[F, AA, B]]
 
   /**
@@ -187,12 +187,12 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
    *     EitherT.rightT(List(10))
    *
    *   val iter: EitherT[Eval, String, Iterable[Int]] =
-   *     list.upcastR
+   *     list.rightWiden
    * }}}
    *
-   * @see [[upcastL]]
+   * @see [[leftWiden]]
    */
-  @inline def upcastR[BB >: B]: EitherT[F, A, BB] =
+  @inline def rightWiden[BB >: B]: EitherT[F, A, BB] =
     this.asInstanceOf[EitherT[F, A, BB]]
 
   /**
--- a/core/src/main/scala/cats/data/EitherT.scala
+++ b/core/src/main/scala/cats/data/EitherT.scala
@@ -1,123 +1,1058 @@
 package cats
 package data
 
-import cats.Bifunctor
 import cats.instances.either._
 import cats.syntax.either._
+import cats.internals.EitherUtil.{eitherCast, rightBox, leftBox}
 
 /**
- * Transformer for `Either`, allowing the effect of an arbitrary type constructor `F` to be combined with the
- * fail-fast effect of `Either`.
+ * Transformer for `Either`, allowing the effect of an arbitrary type
+ * constructor `F` to be combined with the fail-fast effect of `Either`.
  *
- * `EitherT[F, A, B]` wraps a value of type `F[Either[A, B]]`. An `F[C]` can be lifted in to `EitherT[F, A, C]` via `EitherT.right`,
- * and lifted in to a `EitherT[F, C, B]` via `EitherT.left`.
+ * `EitherT[F, A, B]` wraps a value of type `F[Either[A, B]]`.
+ *
+ * For example `EitherT` can be combined with types such as [[Eval]],
+ * `scala.concurrent.Future` or `cats.effect.IO` for principled error
+ * handling:
+ *
+ * {{{
+ *   import cats.Eval
+ *   import cats.Eval.always
+ *   import cats.data.EitherT
+ *   import java.lang.Long.{parseNum => javaParseNum}
+ *
+ *   def parseNum(s: String, radix: Int = 10): EitherT[Eval, NumberFormatException, Long] =
+ *     EitherT(always {
+ *       try
+ *         Right(javaParseNum(s, radix))
+ *       catch { case e: NumberFormatException =>
+ *         Left(e)
+ *       }
+ *     })
+ * }}}
+ *
+ * Note that in this function the error type is part of the signature and
+ * is more specific than the customary `Throwable` or `Exception`. You can
+ * always "upcast" its error type to `Throwable` later using [[upcastL]]:
+ *
+ * {{{
+ *   val num: EitherT[Eval, Throwable, Long] =
+ *     parseNum("10210", 10).upcastL
+ * }}}
+ *
+ * The power of `EitherT` is that it combines `F[_]` with the `Either`
+ * data type, with the result still being a `MonadError`, so you can
+ * comfortably use operations such as [[map]], [[flatMap]], [[attempt]]
+ * and others:
+ *
+ * {{{
+ *   def parseEvenNum(s: String, radix: Int = 10): EitherT[Eval, NumberFormatException, Long] =
+ *     parseNum(s, radix).flatMap { i =>
+ *       if (i % 2 == 0)
+ *         EitherT.rightT(i)
+ *       else
+ *         EitherT.leftT(new NumberFormatException(s"Not an even number: $$i"))
+ *     }
+ * }}}
+ *
+ * Tip: An `F[A]` can be lifted in to `EitherT[F, E, A]` via [[EitherT.right]]
+ * and lifted in to a `EitherT[F, A, B]` via [[EitherT.left]].
+ *
+ * {{{
+ *   def rand(seed: Long): Eval[Int] =
+ *     Eval.always {
+ *       val newSeed = (seed * 0x5DEECE66DL + 0xBL) & 0xFFFFFFFFFFFFL
+ *       (newSeed >>> 16).toInt
+ *     }
+ *
+ *   // Lifting values
+ *   def randPredicate(seed: Long)(p: Int => Boolean): EitherT[Eval, String, Int] =
+ *     EitherT.right(rand(seed)).flatMap { nr =>
+ *       if (p(nr))
+ *         EitherT.leftT("Predicate was false")
+ *       else
+ *         EitherT.pure(nr)
+ *     }
+ * }}}
+ *
+ * @define monadErrorF For example data types like `cats.effect.IO` implement
+ *         [[MonadError]] and can thus express computations that can fail on their
+ *         own, usually via `Throwable`. When working with `EitherT`, if the
+ *         underlying `F` is implementing [[ApplicativeError]], sometimes it is
+ *         necessary to recover from such errors thrown in `F[_]`.
  */
 final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
-  def fold[C](fa: A => C, fb: B => C)(implicit F: Functor[F]): F[C] = F.map(value)(_.fold(fa, fb))
+  /**
+   * Folds the underlying `Either` value, yielding an `F[_]` as result.
+   *
+   * Example:
+   * {{{
+   *   val parsed: Eval[Long] = parseNum(num).fold(_ => 0, x => x)
+   * }}}
+   */
+  def fold[C](fa: A => C, fb: B => C)(implicit F: Functor[F]): F[C] =
+    F.map(value)(_.fold(fa, fb))
 
-  def isLeft(implicit F: Functor[F]): F[Boolean] = F.map(value)(_.isLeft)
+  /**
+   * Queries the underlying `Either` value, returning `true` if it's a `Left`,
+   * or `false` otherwise.
+   *
+   * {{{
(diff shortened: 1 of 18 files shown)


LukaJCB:
We could add redeem and `redeemWith` add syntax on the `SyntaxBinCompat` trait :)

core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -1,123 +1,1058 @@
...
+   *   val num: EitherT[Eval, NumberFormatException, Long] =
+   *     EitherT.rightT(10210)
+   *
+   *   val num2: EitherT[Eval, Throwable, Long] =
+   *     num.upcastL
+   * }}}
+   *
+   * @see [[upcastR]]
+   */
+  @inline def upcastL[AA >: A]: EitherT[F, AA, B] =
```
johnynek: can we call this leftWiden to mirror widen which we normally use here?
https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Bifunctor.scala#L54

core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -1,123 +1,1058 @@
...
+   *   val list: EitherT[Eval, String, List[Int]] =
+   *     EitherT.rightT(List(10))
+   *
+   *   val iter: EitherT[Eval, String, Iterable[Int]] =
+   *     list.upcastR
+   * }}}
+   *
+   * @see [[upcastL]]
+   */
+  @inline def upcastR[BB >: B]: EitherT[F, A, BB] =
```
johnynek: rightWiden?

alexandru:
@LukaJCB 
> We could add redeem and redeemWith add syntax on the SyntaxBinCompat trait :)

Not sure what you mean, I see there is a `AllSyntaxBinCompat0`, don't know what the purpose is.

@johnynek 

I renamed `upcastL` to `leftWiden` and `upcastR` to `widen`. That naming is unfortunate tbh, because the technical term is _upcasting_. It explains for example why I had no idea they existed. But being consistent is more important.

alexandru:
@johnynek actually I removed `widen` / `leftWiden`, as I don't want to defend adding those on the class itself 🙂

LukaJCB:
Sorry about the lack of depth on my last comment, was on mobile and on a train. We can create some `MonadErrorExtraOps` trait put the two functions there and then create another `AllSyntaxBinCompat1` trait just like the one [here](https://github.com/typelevel/cats/blob/ef05c76d92076f7ba0e1589b8a736cb973a9f610/core/src/main/scala/cats/syntax/all.scala#L56). Then we can mix those into the `AllSyntaxBinCompat` trait as well as the `object implicits` found [here](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/implicits.scala). That should give everyone the ability to use these methods whenever they import all syntax.

When cats 2.0 comes around, we can then move them into their respective type classes :)

alexandru:
Actually I want to think about this more, given Cats 2.0's timeline I think that's OK, we've got time. Closing it to prevent merging or PR creep, will reopen if the case.

alexandru:
OK, sorry for reopening it — I've carefully considered possibilities, doing benchmarks with alternatives, like a variant that takes a function with an `Either[Throwable, A]` parameter, plus I modified some Monix code to see how it feels like and I think this approach is the best.

The signature is nice too, it's basically a fold (catamorphism?) that happens in the `F[_]` context.

joan38:
Is this PR scheduled to go in for 2.0? If yes can we add the label?

alexandru:
> Is this PR scheduled to go in for 2.0? 

I hope so 🙂

joan38:
@kailuowang could you add the `2.0` label please ? 😄

joan38:
Hey, Did we forget to merge this for the 2.0 release? Is it still time?

kailuowang:
We didn't forget. Our release plan changed. The original 2.0 release is going to be 2.1, the 2.0 release we are preparing now is different and doesn't allow PRs that break binary compat on Scala 2.11 like this one.

travisbrown:
Removing this from the 2.1.0-RC1 milestone in favor of the subset in #3146. I do think we should aim to get the `EitherT` enhancements merged at some point, so I don't think we should close it yet.

core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -174,6 +245,194 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
...
+   * {{{
+   *   parseNum(s).handleError(_ => 0L)
+   * }}}
+   *
+   * @see [[handleErrorWith]] for recovering errors by mapping them to
+   *      `EitherT` values (aka the equivalent of `flatMap` for errors)
+   *
+   * @see [[handleErrorF]] for recovering errors thrown in `F[_]`
+   */
+  def handleError(f: A => B)(implicit F: Functor[F]): EitherT[F, A, B] =
```
johnynek, line 323: I kind of think this should take a new type parameter for A:

```
def handleError[A1](f: A => B)... : EitherT[F, A1, B]
```
Since we know that there is no A left in the result.

core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -174,6 +245,194 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
...
+   *   // Yields 0L on evaluation
+   *   eio.handleErrorF(_ => 0L)
+   * }}}
+   *
+   * @see [[handleErrorWithF]] for recovering errors by mapping them to
+   *      `EitherT` values (aka the equivalent of `flatMap` for errors)
+   *
+   * @see [[handleError]] for recovering errors expressed via `EitherT`
+   */
+  def handleErrorF[E](f: E => B)(implicit F: ApplicativeError[F, E]): EitherT[F, A, B] =
```
johnynek, line 355: Same comment about A

core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -174,6 +245,194 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
...
+   *       // Rethrowing error, because we don't know what it is
+   *       EitherT.leftT(other)
+   *   }
+   * }}}
+   *
+   * @see [[handleError]] for recovering errors by mapping them to simple values
+   *
+   * @see [[handleErrorWithF]] for recovering errors thrown in `F[_]`
+   */
+  def handleErrorWith(f: A => EitherT[F, A, B])(implicit F: Monad[F]): EitherT[F, A, B] =
```
johnynek, line 394: Seems this could also benefit from a type parameter A1 in f result type.

core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -174,6 +245,194 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
...
+   *       // prefer to treat them as unrecoverable
+   *       EitherT.right(IO.raiseError(other))
+   *   }
+   * }}}
+   *
+   * @see [[handleErrorF]] for recovering errors by mapping them to simple values
+   *
+   * @see [[handleErrorWith]] for recovering errors expressed via `EitherT`
+   */
+  def handleErrorWithF[E](f: E => EitherT[F, A, B])(implicit F: ApplicativeError[F, E]): EitherT[F, A, B] =
```
johnynek, line 433: Some A comment

johnynek (changes requested):

core/src/main/scala/cats/data/EitherT.scala
armanbilge: I guess we can do `A1 >: A` or something.

PR #2238, opened 2018-04-26
Updating the version number on the microsite landing page

dandxy89:


dandxy89:
@kailuowang - the README is correct. So all that required is to push the latest build to 'gh-pages' :)

PR #2245, opened 2018-05-04
clarify law testing's dependency on cats-testkit and scalatest

kailuowang:


docs/src/main/tut/typeclasses/lawtesting.md
```diff
@@ -7,23 +7,25 @@ section: "typeclasses"
 # Law testing
 
 [Laws](https://typelevel.org/cats/typeclasses.html#laws) are an important part of cats.
-Cats uses `catalysts` and `discipline` to help test instances with laws.
-To make things easier, cats ships with `cats-testkit`, which makes use of `catalysts` and `discipline` and exposes `CatsSuite` based on ScalaTest.
+Cats uses `catalysts` and [discipline](https://github.com/typelevel/discipline) to help test instances with laws. To test type class
```
tpolecat: Why is `catalysts` relevant here?

docs/src/main/tut/typeclasses/lawtesting.md
```diff
@@ -50,8 +52,8 @@ For simplicity we'll just use `Eq.fromUniversalEquals`:
 implicit def eqTree[A: Eq]: Eq[Tree[A]] = Eq.fromUniversalEquals
 ```
 
-Then we can begin to write our law tests. Start by creating a new class in your `test` folder and inheriting from `cats.tests.CatsSuite`.
-`CatsSuite` extends the standard ScalaTest `FunSuite` as well as `Matchers`.
+Then we can begin to write our law tests. If you are using `ScalaTest` you can inherit from `cats.tests.CatsSuite` 
+from `cats-testkit`. `CatsSuite` extends the standard ScalaTest `FunSuite` as well as `Matchers`.
```
tpolecat: What do you do if you're not using ScalaTest? Not sure this helps.

kailuowang:
@tpolecat how about now?

docs/src/main/tut/typeclasses/lawtesting.md
```diff
@@ -7,23 +7,28 @@ section: "typeclasses"
 # Law testing
 
 [Laws](https://typelevel.org/cats/typeclasses.html#laws) are an important part of cats.
-Cats uses `catalysts` and `discipline` to help test instances with laws.
-To make things easier, cats ships with `cats-testkit`, which makes use of `catalysts` and `discipline` and exposes `CatsSuite` based on ScalaTest.
+Cats uses [discipline](https://github.com/typelevel/discipline) to define type class laws and 
+the [ScalaCheck](https://github.com/rickynils/scalacheck) tests based on them.
+
+To test type class laws from Cats against your instances, you need to add a `cats-laws` dependency. 
+If you are using `ScalaTests`, Cats also ships with optional `cats-testkit`, which provites a convenient
```
LukaJCB: typo `ScalaTests`

kailuowang:
made some tut compilation corrections. need a quick re-review.

docs/src/main/tut/typeclasses/lawtesting.md
```diff
@@ -43,34 +48,27 @@ object Tree {
   }
 }
 ```
-We will also need to create an `Eq` instance, as most laws will need to compare values of a type to properly test for correctness.
-For simplicity we'll just use `Eq.fromUniversalEquals`:
-
-```tut:book
-implicit def eqTree[A: Eq]: Eq[Tree[A]] = Eq.fromUniversalEquals
+```tut:invisible
+import Tree._ //there is no real companion object is in REPL
```
rossabaker: s/is in/in the/

PR #2248, opened 2018-05-06
Update state.md

rsekulski:
I think functions `open` and `close` were swapped in section about `IndexedStateT`.


rsekulski:
I'm really sorry for such late response. I've added code which tries to run `valid` with both `Open` and `Closed` to demonstrate type safety.

ceedubs:
Thanks @rsekulski!

PR #2249, opened 2018-05-06
Add some FlatMap loops useful for State and Effects

johnynek:
Three cases I have wanted:

1. taking a scalacheck `Gen[Option[A]] => Gen[A]`
2. writing a server that has a single step as `IO[Unit]` then repeating that forever.
3. a server that has some state: `S => IO[S]` that each call updates the state and repeat forever.

This has to be done with syntax backwards compatibility hacks, but we can repair that at cats 2.0

### DIFF
--- a/core/src/main/scala/cats/data/OptionT.scala
+++ b/core/src/main/scala/cats/data/OptionT.scala
@@ -137,14 +137,6 @@ final case class OptionT[F[_], A](value: F[Option[A]]) {
    * }}}
    */
   def toNested: Nested[F, Option, A] = Nested(value)
-
-  /**
-   * This repeats an F until we get defined values. This can be useful
-   * for polling type operations on State (or RNG) Monads, or in effect
-   * monads.
-   */
-  def untilDefined(implicit F: FlatMap[F]): F[A] =
-    (new syntax.FlatMapOptionOps(value)).untilDefinedM
 }
 
 object OptionT extends OptionTInstances {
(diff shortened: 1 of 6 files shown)


core/src/main/scala/cats/data/OptionT.scala
```diff
@@ -137,6 +137,14 @@ final case class OptionT[F[_], A](value: F[Option[A]]) {
    * }}}
    */
   def toNested: Nested[F, Option, A] = Nested(value)
+
+  /**
+   * This repeats an F until we get defined values. This can be useful
+   * for polling type operations on State (or RNG) Monads, or in effect
+   * monads.
+   */
+  def untilDefined(implicit F: FlatMap[F]): F[A] =
```
johnynek: I didn't name this the same as `untilDefinedM` because it is slightly different. `OptionT[F, Option[A]]` should still have access to untilDefinedM. This can possibly be removed since it adds little over `optt.value.untilDefinedM`

tests/src/test/scala/cats/tests/IndexedStateTSuite.scala
```diff
@@ -269,6 +269,41 @@ class IndexedStateTSuite extends CatsSuite {
...
+
+  test("foreverM works") {
+    val step = StateT[Either[Int, ?], Int, Unit] { i =>
+      if (i > 1000) Left(i) else Right((i + 1, ()))
+    }
+    step.foreverM.run(0) match {
+      case Left(big) => big should === (1001)
+      case Right((_, _)) => fail("unreachable code due to Nothing, but scalac won't let us match on it")
+    }
+  }
```
kailuowang: nitpick. The current implementation looks stack safe, but shall we still add a stack safety test for future guard rail?

johnynek:
@kailuowang look good now?

kailuowang:
Thanks for adding the test. @johnynek

PR #2252, opened 2018-05-09
Add flatTapWith to Kleisli

marcodippy:


core/src/main/scala/cats/data/Kleisli.scala
```diff
@@ -75,6 +75,9 @@ final case class Kleisli[F[_], A, B](run: A => F[B]) { self =>
   def tapWith[C](f: (A, B) => C)(implicit F: Functor[F]): Kleisli[F, A, C] =
     Kleisli(a => F.map(run(a))(b => f(a, b)))
 
+  def flatTapWith[C](f: (A, B) => F[C])(implicit F: FlatMap[F]): Kleisli[F, A, C] =
```
johnynek: we have this function `flatTap` on anything with a FlatMap, but it does not seem consistent with this:

https://github.com/typelevel/cats/blob/ff7bd409ea21775087552daba12078b79f3d2abf/core/src/main/scala/cats/FlatMap.scala#L142

it discards the value inside the `F[C]`.

I am concerned we don't have a clear meaning associated with `tap` and `flatTap`.

marcodippy: I like `tapWithF`, thanks for the suggestion @johnynek!
I already pushed the change 👍

PR #2259, opened 2018-05-18
Binested sketch

iravid:
I thought this could be a nice way to abstract over the different ways to compose a two-slot type constructor with two other type constructors.

If this looks useful, I'll add the rest of the million instances available for it. :-)


kailuowang:
Looks useful to me 👍

LukaJCB:
I like it too, thanks very much! :)

iravid:
Added `Bifoldable` and `Bitraverse` instances along with `.binested` syntax in the spirit of #2262. I mixed `AllSyntaxBinCompat` into `SyntaxSuite` to test the bincompat syntax.

I think this is good to merge if all tests pass.

core/src/main/scala/cats/data/Binested.scala
```diff
@@ -0,0 +1,93 @@
...
+    implicit F0: Bifoldable[F], G0: Foldable[G], H0: Foldable[H]): Bifoldable[Binested[F, G, H, ?, ?]] =
+    new BinestedBifoldable[F, G, H] {
+      override implicit def F: Bifoldable[F] = F0
+      override implicit def G: Foldable[G] = G0
+      override implicit def H: Foldable[H] = H0
+    }
+}
+
+trait BinestedInstances0 {
+  implicit def catsDataBitraverseForBinested[F[_, _], G[_], H[_]](
```
kailuowang: `Bitraverse` should take higher priority than `Bifoldable` and `Bifunctor`
https://typelevel.org/cats/guidelines.html#a-idimplicit-priority-hrefimplicit-prioritya-implicit-instance-priority

core/src/main/scala/cats/syntax/all.scala
```diff
@@ -58,5 +58,6 @@ trait AllSyntaxBinCompat0
     extends UnorderedTraverseSyntax
     with ApplicativeErrorExtension
     with TrySyntax
+    with BinestedSyntax
```
kailuowang: hmmm, this might break bin compat with 1.1.0. 
Would you mind also change the bin compat version [here](https://github.com/typelevel/cats/blob/master/build.sbt#L207) to `1.1.0`? That should capture this error.

iravid: Ah cool. I'll rebase on that.

core/src/main/scala/cats/data/Binested.scala
iravid: Right, thanks

iravid:
fingers crossed for the MiMa gods

core/src/main/scala/cats/syntax/all.scala
iravid: I see that sticking it on `AllSyntaxBinCompat1` is fine. Will change to that...

kailuowang:
oops, sorry, I squashed merged the other PR. Would you reset this branch to master and cherry pick your commit 
`git cherry-pick e9f80d988222efec9b7fce11d7acee3727e93333`

iravid:
Yeah no worries.

iravid:
Rebased and resolved conflicts

iravid:
Travis failed 2 builds due to memory limits. Restarted the build...

PR #2260, opened 2018-05-18
Update Scala, sbt, and sbt plugins

fthomas:


ceedubs:
The build is failing with `[error] (bench / Compile / compileIncremental) scala.reflect.internal.MissingRequirementError: object java.lang.Object in compiler mirror not found.`

I feel like `java.lang.Object` is something that should probably be there. 🤔

LukaJCB:
@ceedubs That seems to be an issue in every single build for 2.11, not specific to this PR. No clue what's going on there :smile:

fthomas:
The first error seems to be 
```
[error] error while loading <root>, Error accessing /home/travis/.coursier/cache/v1/https/repo1.maven.org/maven2/org/scalaz/scalaz-core_2.11/7.2.21/scalaz-core_2.11-7.2.21.jar
```
I tried deleting the build cache but this had no effect. I'll investigate further.

fthomas:
Updating Scalaz to 7.2.23 fixed the problem on 2.11 - but don't ask me why.

PR #2264, opened 2018-05-21
update bin compat versions to 1.0.0 and 1.1.0

kailuowang:
somehow I missed that last release

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -204,7 +204,7 @@ lazy val docSettings = Seq(
   includeFilter in Jekyll := (includeFilter in makeSite).value
 )
 
-lazy val binaryCompatibleVersion = "1.0.0"
+lazy val binaryCompatibleVersion = "1.1.0"
 
 def mimaSettings(moduleName: String) = Seq(
   mimaPreviousArtifacts := Set("org.typelevel" %% moduleName % binaryCompatibleVersion)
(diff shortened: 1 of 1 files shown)


tpolecat:
It should stay at 1.0 shouldn't it? We want to guarantee compatibility all the way back.

kailuowang:
but that won't capture bin compat break between 1.1.0 and 1.2.0, e.g. if 1.1.0 introduced a new trait, 1.2.0 could add a method to it while still bin compat with 1.0.

ChristopherDavenport:
@kailuowang  I believe MiMa allows multiple versions to be set. Such that mimaPreviousArtifacts points at both `1.1.0` and `1.0.0`

kailuowang:
@chir thanks for the tip , I'll update that

tpolecat:
I remain confused. What's the value in checking against 1.1.0?

johnynek:
@tpolecat we could actually be compatible with `1.0.0` but not compatible with `1.1.0`. Imagine we remove some of the compatible methods we added in 1.1.0.

We want to be compatible with all previous versions in 1.x.y, and unless mima has bugs (which of course, it does because it is software), checking only the previous should be safe. But we could help detect bugs in mima by also doing the full transitive closure in our CI. If it gets too slow, we can only do a subset of the previous releases.

tpolecat:
Aha, gotcha. Yep 👍

tpolecat:
It would be cool if sbt-release maintained this list the same way it maintains `version`.

PR #2265, opened 2018-05-22
Add Function To Check All Relevant MiMa Versions

ChristopherDavenport:
Removes Need for Manually Setting Binary Compatible Versions

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -231,7 +231,7 @@ def mimaSettings(moduleName: String) = {
         semverBinCompatVersions(major.toInt, minor.toInt, patch.toInt)
           .map{case (maj, min, pat) => s"${maj}.${min}.${pat}"}
       case _ =>
-        List.empty[String]
+        Set.empty[String]
     }
   }
   // Safety Net For Exclusions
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -204,11 +204,35 @@ lazy val docSettings = Seq(
...
+  import sbtrelease.Version
+  def mimaVersions(version: String): List[String] = {
+    def semverBinCompatVersions(major: Int, minor: Int, patch: Int): List[(Int, Int, Int)] = {
+      val majorVersions: List[Int] = List(major)
+      val minorVersions : List[Int] = 
+        if (major >= 1) Range(0, minor).inclusive.toList
+        else List(minor)
+      val patchVersions: List[Int] = 
+        if (minor == 0 || patch == 0) List.empty[Int] 
+        else Range(0, patch - 1).inclusive.toList
```
kailuowang: hmmm, I wonder if this would work. Your patchVersions is derived from the current major.minor.patch range, but you don't know if your previous minor version's patch range right? 
e.g. if the current version is `1.1.0`, you wouldn't check against `1.0.1`? or if the current version is `1.1.3-SNAPSHOT`, your list would include 1.0.2 which doesn't exist. Another issue is that if either if the current patch is 0, you wouldn't check against anything, e.g. `1.2.0-SNAPSHOT` wouldn't check mima, right?

ChristopherDavenport: This is why I knew this would be the place. 😄

Issue #2269, opened 2018-05-24
Adding biSemiFlatMap to EitherT


ericaovo:
Looking into the methods available on an `EitherT` instance I came across the need of having something like the following:

```
def biSemiflatMap[C, D](fa: A => F[C], fb: B => F[D]): EitherT[F, C, D] =
  eitherT.leftSemiflatMap(fa).semiflatMap(fb)
```

Here's an example:
```
import cats.data._, cats.implicits._, cats.Monad

val eitherT: EitherT[List, String, Int] = EitherT(List(Left("abc"), Right(123)))

implicit class EitherTOps[F[_]: Monad, A, B](val eitherT: EitherT[F, A, B]) {
  def biSemiflatMap[C, D](fa: A => F[C], fb: B => F[D]): EitherT[F, C, D] =
    eitherT.leftSemiflatMap(fa).semiflatMap(fb)
}

eitherT.biSemiflatMap(string => List(string.length), int => List(int.toFloat))
// res: cats.data.EitherT[List,Int,Float] = EitherT(List(Left(3), Right(123.0)))
```

Happy to open a PR if it sounds reasonable.

PR #2274, opened 2018-06-01
Adding biSemiflatMap to EitherT (#2269)

ericaovo:
Addresses [#2269](https://github.com/typelevel/cats/issues/2269).

### DIFF
--- a/core/src/main/scala/cats/data/EitherT.scala
+++ b/core/src/main/scala/cats/data/EitherT.scala
@@ -120,6 +120,9 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
       case r@Right(_) => F.pure(r.leftCast)
     })
 
+  def biSemiflatMap[C, D](fa: A => F[C], fb: B => F[D])(implicit F: Monad[F]): EitherT[F, C, D] =
+    leftSemiflatMap(fa).semiflatMap(fb)
+
   def compare(that: EitherT[F, A, B])(implicit o: Order[F[Either[A, B]]]): Int =
     o.compare(value, that.value)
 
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -120,6 +120,9 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
       case r@Right(_) => F.pure(r.leftCast)
     })
 
+  def biSemiflatMap[C, D](fa: A => F[C], fb: B => F[D])(implicit F: Monad[F]): EitherT[F, C, D] =
+    leftSemiflatMap(fa).semiflatMap(fb)
```
kubukoz: Maybe it would be better to implement this in a single walk over `F`?

If `F` is e.g. `List`, this would traverse the list twice, and we should probably not do that ;)

Other than that, I think it's worth mentioning this PR to one of the maintainers, since there hasn't been any activity from them on this or the original issue.

ericaovo: Fair point @kubukoz , so I would refactor it to be:
```
def biSemiflatMap[C, D](fa: A => F[C], fb: B => F[D])(implicit F: Monad[F]): EitherT[F, C, D] =
    EitherT(F.flatMap(value) {
      case Left(a) => F.map(fa(a)) { c => Left(c) }
      case Right(b) => F.map(fb(b)) { d => Right(d) }
    })
```
In this case it will `flatMap` and `map` it once, instead of `flatMap` it twice.

ericaovo:
@kailuowang would you mind reviewing this please?

ericaovo:
Sorry about that, I just had the time to fix it now, thank you! :) Would you mind having another look @LukaJCB @kailuowang ? I've automatically dismissed your approval when I pushed :|

PR #2277, opened 2018-06-03
Add an explanation for the missing Applicative instances for monad transformers

### LINKED ISSUE #2272: Document why monad transformers don't have Applicative instances.
see https://github.com/typelevel/cats/pull/2181#issuecomment-370098157
Probably in the FAQ section, maybe also mention in the guideline.

kubukoz:
Closes #2272.

Just a few lines to start with, hope you have some feedback on what else I can mention/what can be reworded :)

### DIFF
--- a/docs/src/main/tut/faq.md
+++ b/docs/src/main/tut/faq.md
@@ -17,6 +17,7 @@ position: 40
  * [Why is some example code not compiling for me?](#example-compile)
  * [How can I turn my List of `<something>` into a `<something>` of a list?](#traverse)
  * [Where is `ListT`?](#listt)
+ * [Where are `Applicative`s for monad transformers?(#applicative-monad-transformers)
  * [Where is `IO`/`Task`?](#task)
  * [What does `@typeclass` mean?](#simulacrum)
  * [What do types like `?` and `λ` mean?](#kind-projector)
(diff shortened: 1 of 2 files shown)


LukaJCB:
I like it, but can we extract a bit from the issue you linked, I feel like just linking to an issue will cause people to have to wade through a bunch of discussion instead of getting to the actual point. You know what I mean? 🤔

kubukoz:
Sure, I'll look for something that can be there instead of the link.

kubukoz:
Can we just drop the link? The potential inconsistency is already mentioned. WDYT?

LukaJCB:
I think that might be more confusing, then it'd just say "This is inconsistent" without naming a good reason.

LukaJCB:
Maybe add the counter example noted here? https://github.com/typelevel/cats/pull/1106#issuecomment-225067111

Of course would need porting to `EitherT` or `OptionT` instead. :)

kubukoz:
@easel I haven't had much time to work on this recently, but I'll surely extract something before I consider this good to merge :P

kubukoz:
@LukaJCB as part of this week's goal to finish up all open OSS PRs, I updated this one - please take a look :)

kubukoz:
@kailuowang FYI

kailuowang:
Thanks very much @kubukoz !

PR #2278, opened 2018-06-03
Fix typo in Foldable comment

kubukoz:
btw. would it make sense to say more about the trick used?


johnynek:
it would make sense to add to the comment, which is a general trick to avoid calling both isDefined and apply, but also to avoid using lift.

kubukoz:
I'll add a mention of that.

kubukoz:
Finally found a moment to add more info, please review

johnynek:
👍

kubukoz:
:shipit:

PR #2279, opened 2018-06-04
Add Defer typeclass, laws and implementations

### LINKED ISSUE #2273: Proposal: adding a Suspend typeclass
It is often useful to have a way to see if a type constructor supports laziness.

Consider:
```scala
trait Defer[F[_]] {
  def defer[A](f: => F[A]): F[A]
}
```

Note, that cats-effect has this method (called suspend) on Sync: https://github.com/typelevel/cats-effect/blob/master/core/shared/src/main/scala/cats/effect/Sync.scala#L37

In cats, the law would be:
```scala
defer(fa) <-> fa
```
along with the ad-hoc law that:
```scala
var evaluated = false
val deferFa = defer {
  evaluated = true
  fa
}

assert(!evaluated)
deferFa <-> fa
```

Some example instances:
```scala
class DeferFunction0 extends Defer[Function0] {
  def defer[A](fa: => Function0[A]): Function0[A] =
    () => fa()
}

class DeferFunction1[Z] extends Defer[Z => ?] {
  def defer[A](fa: => Function[Z, A]): Function[Z, A] =
    { z: Z => fa(z) }
}

class DeferEval extends Suspend[Eval] {
  def defer[A](fa: => Eval[A]): Eval[A] =
    Eval.defer(fa)
}

class DeferFree[M[_]] extends Defer[Free[M, ?]] {
  def defer[A](fa: => Free[M, A]): Free[M, A] =
    Free.defer(fa)
}
```
and Kleisli, IndexedStateT, EitherT, OptionT, etc... I think all have them if the `F[_]` they are working with has them.

One use for these, for example, is working with `map2Eval` on Apply: 
https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Apply.scala#L194

If we have a `Defer[F]` then we could (arguably) implement:
```scala
  def map2Eval[A, B, Z](fa: F[A], fb: Eval[F[B]])(f: (A, B) => Z): Eval[F[Z]] =
    Eval.later(map2(fa, defer(fb.value))(f))
```

If we added this typeclass, `cats.effect.Sync` could extend `Defer` with `suspend == defer`.

johnynek:
This seems like a little thing, but I was able to use it to implement `ContT[M, A, B]`, a wrapper for `(B => M[A]) => M[A]` and pass all the laws using the naive map/flatMap (with cats.data.AndThen) and the following tailRecM:
```scala
 def tailRecM[M[_], A, B, C](a: A)(fn: A => ContT[M, C, Either[A, B]])(implicit M: Defer[M]): ContT[M, C, B] =
  ContT[M, C, B] { cb: (B => M[C]) => 
     def go(a: A): M[C] =
        fn(a).run {
          case Left(a) => M.defer(go(a))
          case Right(b) => cb(b)
        }
       go(a)
    }
```

I think this is pretty clear that this is an interesting typeclass along with the fact that we have so many implementations.

closes #2273

### DIFF
--- a/core/src/main/scala/cats/data/IndexedReaderWriterStateT.scala
+++ b/core/src/main/scala/cats/data/IndexedReaderWriterStateT.scala
@@ -462,6 +462,12 @@ private[data] sealed abstract class IRWSTInstances extends IRWSTInstances1 {
       implicit def L: Monoid[L] = L0
     }
 
+  implicit def catsDataDefer[F[_], E, L, SA, SB](implicit F: Defer[F]): Defer[IndexedReaderWriterStateT[F, E, L, SA, SB, ?]] =
+    new Defer[IndexedReaderWriterStateT[F, E, L, SA, SB, ?]] {
+      def defer[A](fa: => IndexedReaderWriterStateT[F, E, L, SA, SB, A]): IndexedReaderWriterStateT[F, E, L, SA, SB, A] =
+        IndexedReaderWriterStateT.applyF(F.defer(fa.runF))
+    }
+
 }
 
 private[data] sealed abstract class IRWSTInstances1 extends IRWSTInstances2 {
--- a/core/src/main/scala/cats/data/IorT.scala
+++ b/core/src/main/scala/cats/data/IorT.scala
@@ -435,6 +435,12 @@ private[data] abstract class IorTInstances extends IorTInstances1 {
       Monad[IorT[M, E, ?]]
     }
   }
+
+  implicit def catsDataDeferForIor[F[_], E](implicit F: Defer[F]): Defer[IorT[F, E, ?]] =
+    new Defer[IorT[F, E, ?]] {
+      def defer[A](fa: => IorT[F, E, A]): IorT[F, E, A] =
+        IorT(F.defer(fa.value))
+    }
 }
 
 private[data] abstract class IorTInstances1 extends IorTInstances2 {
--- a/laws/src/main/scala/cats/laws/DeferLaws.scala
+++ b/laws/src/main/scala/cats/laws/DeferLaws.scala
@@ -18,6 +18,14 @@ trait DeferLaws[F[_]] {
     }
     evaluated <-> false
   }
+
+  def deferIsStackSafe[A](fa: Unit => F[A]): IsEq[F[A]] = {
+    def loop(c: Int): F[A] =
+      if (c <= 0) F.defer(fa(()))
+      else F.defer(loop(c - 1))
+
+    loop(1000000) <-> (fa(()))
+  }
 }
 
 object DeferLaws {
--- a/laws/src/main/scala/cats/laws/DeferLaws.scala
+++ b/laws/src/main/scala/cats/laws/DeferLaws.scala
@@ -1,0 +1,26 @@
+package cats
+package laws
+
+/**
+ * Laws that must be obeyed by any `Defer`.
+ */
+trait DeferLaws[F[_]] {
+  implicit def F: Defer[F]
+
+  def deferIdentity[A](fa: Unit => F[A]): IsEq[F[A]] =
+    F.defer(fa(())) <-> fa(())
+
+  def deferDoesNotEvaluate[A](fa: Unit => F[A]): IsEq[Boolean] = {
+    var evaluated = false
+    val deferUnit = F.defer {
+      evaluated = true;
+      fa(())
+    }
+    evaluated <-> false
+  }
+}
+
+object DeferLaws {
+  def apply[F[_]](implicit ev: Defer[F]): DeferLaws[F] =
+    new DeferLaws[F] { def F: Defer[F] = ev }
+}
--- a/core/src/main/scala/cats/Defer.scala
+++ b/core/src/main/scala/cats/Defer.scala
@@ -1,0 +1,33 @@
+package cats
+
+/**
+ * Defer is a typeclass that shows the ability to defer creation
+ * inside of the type constructor F[_].
+ *
+ * This comes up with F[_] types that are implemented with a trampoline
+ * or are based on function application.
+ *
+ * The law is that defer(fa) is equivalent to fa, but not evaluated immediately,
+ * so
+ * {{{
+ * import cats._
+ * import cats.implicits._
+ *
+ * var evaluated = false
+ * val dfa =
+ *   Defer[Eval].defer {
+ *     evaluated = true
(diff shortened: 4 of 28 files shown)


johnynek:
I think this strategy can allow us to extend ContT to any cats.Monad: in the ContT.tailRecM loop, lift M to Free, use defer on Free, then finally run the Free to get back to M at the end.

core/src/main/scala/cats/data/IndexedReaderWriterStateT.scala
```diff
@@ -462,6 +462,12 @@ private[data] sealed abstract class IRWSTInstances extends IRWSTInstances1 {
       implicit def L: Monoid[L] = L0
     }
 
+  implicit def catsDataDefer[F[_], E, L, SA, SB](implicit F: Defer[F]): Defer[IndexedReaderWriterStateT[F, E, L, SA, SB, ?]] =
```
kailuowang: totally trivial nit: the naming convention has a `ForIRWST` suffix. Although I don't think it matters much since there is no risk of naming conflict here.

johnynek: yep, that was a goof up.

laws/src/main/scala/cats/laws/DeferLaws.scala
```diff
@@ -0,0 +1,34 @@
...
+    }
+    evaluated <-> false
+  }
+
+  def deferIsStackSafe[A](fa: Unit => F[A]): IsEq[F[A]] = {
+    def loop(c: Int): F[A] =
+      if (c <= 0) F.defer(fa(()))
+      else F.defer(loop(c - 1))
+
+    loop(1000000) <-> (fa(()))
```
kailuowang: this number might be so too big that all jobs in build ran OOM.

core/src/main/scala/cats/data/IorT.scala
```diff
@@ -435,6 +435,12 @@ private[data] abstract class IorTInstances extends IorTInstances1 {
       Monad[IorT[M, E, ?]]
     }
   }
+
+  implicit def catsDataDeferForIor[F[_], E](implicit F: Defer[F]): Defer[IorT[F, E, ?]] =
```
kubukoz: missing `T` after `catsDataDeferForIor`?

kubukoz:
Why not call it `Suspend` to keep consistent with `Sync`'s `suspend` method? You mentioned yourself that it could be an alias to it, but is there a reason why we can't just make `Sync` extend `Defer`/`Suspend` and provide `delay`?

kailuowang:
@johnynek do you want this in the coming 1.2 release?

johnynek:
@kailuowang I would.

kailuowang:
#2284 adds a FunctionInstancesBinCompat0, shall we merge that one first so that you can add your Function `Defer` instance there?

johnynek:
@kailuowang sounds great.

laws/src/main/scala/cats/laws/DeferLaws.scala
```diff
@@ -0,0 +1,36 @@
...
+      evaluated = true;
+      fa(())
+    }
+    evaluated <-> false
+  }
+
+  def deferIsStackSafe[A](fa: Unit => F[A]): IsEq[F[A]] = {
+    def loop(c: Int): F[A] =
+      if (c <= 0) F.defer(fa(()))
+      else F.defer(loop(c - 1))
```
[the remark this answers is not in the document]
johnynek: That's a good question about making the law stronger: if there is a monad, flatMap must be stack safe. The function cases wouldn't pass this law.

As to what might not be a monad: command line parsing is often done with an applicative, not a monad (see optparse-applicative library in haskell or https://github.com/bkirwi/decline for a cats version). Those types can still have a lawful Defer.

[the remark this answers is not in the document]
alexandru: If we have `Applicative` examples without a Monad definition, I guess that's fine 👍

kailuowang:
@johnynek do you happen to have time to finish this one up? or can this wait until the next release?

johnynek:
@kailuowang what needs to happen to merge?

I didn't want to take the time to cut the counts in half suggested above.

kailuowang:
Do you want to respond to [this feedback](https://github.com/typelevel/cats/pull/2279#discussion_r197781737)?

Also the instance of `Function1` has a place now, thanks to the `FunctionInstancesBinCompat0` added by  #2284

laws/src/main/scala/cats/laws/DeferLaws.scala
```diff
@@ -0,0 +1,36 @@
...
+
+    val cnt = if (Platform.isJvm) 20000 else 2000
```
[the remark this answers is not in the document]
johnynek: I don’t see a compelling reason why this is only 2x too large. Can you give an idea?

laws/src/main/scala/cats/laws/DeferLaws.scala
```diff
@@ -0,0 +1,36 @@
...
+ * Laws that must be obeyed by any `Defer`.
+ */
+trait DeferLaws[F[_]] {
+  implicit def F: Defer[F]
+
+  def deferIdentity[A](fa: Unit => F[A]): IsEq[F[A]] =
+    F.defer(fa(())) <-> fa(())
+
+  def deferDoesNotEvaluate[A](fa: Unit => F[A]): IsEq[Boolean] = {
+    var evaluated = false
```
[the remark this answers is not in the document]
johnynek: I’m nervous that if you pass a function that ignores the Boolean this is true trivially.

kailuowang:
@johnynek I added a commit to move `Function0` instance and added `Function1` instance. would you review?

core/src/main/scala/cats/data/IorT.scala
johnynek: looks like we are still missing the T.

kailuowang:
@johnynek mima should've been fixed by the latest commit. It was me being dumb and totally forgot about it.

kailuowang:
@LukaJCB do you want to take another quick look at the commits I made?

LukaJCB:
LGTM :+1:

core/src/main/scala/cats/data/IorT.scala
andyscott: I got you. https://github.com/typelevel/cats/pull/2688/files

PR #2284, opened 2018-06-07
Add Representable Functor

eli-jordan:
I have been working on adding representable functors to cats.

This PR adds;

- The `Representable` typeclass.
- The `RepresentableStore` data type, which is a generalisation of the `Store` comonad.


core/src/main/scala/cats/data/Kleisli.scala
```diff
@@ -7,7 +7,8 @@ import cats.arrow._
 /**
  * Represents a function `A => F[B]`.
  */
```
johnynek: can we revert the style changes in this file?

core/src/main/scala/cats/Representable.scala
```diff
@@ -0,0 +1,133 @@
...
+      }
+
+      def tabulate[A](f: Representation => A): Kleisli[M, E, A] = {
+        def curry[X, Y, Z](f: (X, Y) => Z): X => Y => Z = x => y => f(x, y)
+
+        Kleisli[M, E, A](curry(Function.untupled(f)) andThen R.tabulate)
+      }
+    }
+
+  def representableIsBimonad[F[_], R](implicit R: Representable.Aux[F, R], M: Monoid[R]): Monad[F] = new Bimonad[F] with StackSafeMonad[F] {
```
johnynek: don't we only get Monad if there is no `Monoid[R]`? Should we make that a lower priority implicit?

core/src/main/scala/cats/Representable.scala
```diff
@@ -0,0 +1,133 @@
...
+    override def pure[A](x: A): F[A] = R.tabulate(_ => x)
+
+    override def flatMap[A, B](fa: F[A])(f: A => F[B]): F[B] =
+      R.tabulate(a => R.index(f(R.index(fa)(a)))(a))
+
+    // TODO: Can we make this monad stack safe?
+    // override def tailRecM[A, B](a: A)(f: A => F[Either[A, B]]): F[B] = ???
```
johnynek: I think something like this works:
```scala
R.tabulate { r: R =>
  @annotation.tailrec
  def loop(a: A): B = 
    R.index(f(a))(r) match {
      case Right(b) => b
      case Left(a) => loop(a)
    }

  loop(a)
}
```

tests/src/test/scala/cats/tests/MonadSuite.scala
```diff
@@ -23,7 +23,7 @@ class MonadSuite extends CatsSuite {
     forAll(smallPosInt) { (max: Int) =>
       val (result, aggregation) = incrementAndGet.whileM[Vector](StateT.inspect(i => !(i >= max))).run(0)
       result should ===(Math.max(0, max))
-      aggregation should === ( if(max > 0) (1 to max).toVector else Vector.empty )
+      aggregation should === ( if (max > 0) (1 to max).toVector else Vector.empty )
```
johnynek: can we revert style changes?

tests/src/test/scala/cats/tests/KleisliSuite.scala
```diff
@@ -39,13 +39,13 @@ class KleisliSuite extends CatsSuite {
   checkAll("Semigroupal[Kleisli[Option, Int, ?]]", SerializableTests.serializable(Semigroupal[Kleisli[Option, Int, ?]]))
```
johnynek: can we revert the style changes? I'd prefer a PR with style only changes so we can focus on logic/design.

tests/src/test/scala/cats/tests/RepresentableSuite.scala
```diff
@@ -0,0 +1,43 @@
...
+
+class RepresentableSuite extends CatsSuite {
+
+  checkAll("Id[Int] <-> Unit => String", RepresentableTests[Id, Unit].representable[String])
+
+  checkAll("String => Int <-> String => Int", RepresentableTests[String => ?, String].representable[Int])
+
+  type Pair[A] = (A, A)
+  checkAll("Pair[String, String] <-> Boolean => String", RepresentableTests[Pair, Boolean].representable[String])
+
```
johnynek: can we check the monad/comonad laws on a Representable here?

core/src/main/scala/cats/Representable.scala
eli-jordan: Thanks! I hadn't got around to thinking this through. That solution works.

tests/src/test/scala/cats/tests/RepresentableSuite.scala
eli-jordan: I added a check for Bimonad

core/src/main/scala/cats/Representable.scala
eli-jordan: Yes thats right. I wasn't sure how to prioritise the implicits at the point, but just read up on it. I added `RepresentableInstances1` and `RepresentableInstances0`

eli-jordan:
@johnynek @tpolecat @kailuowang I thinks this is ready for a final review / approval

core/src/main/scala/cats/data/RepresentableStore.scala
```diff
@@ -0,0 +1,52 @@
+package cats.data
+
+import cats.{Comonad, Functor, Representable}
+
+/**
+ * A generalisation of the Store comonad, for any `Representable` functor.
+ * `Store` is the dual of `State`
+ */
+final case class RepresentableStore[F[_], S, A](fa: F[A])(val index: S)(implicit R: Representable.Aux[F, S]) {
```
johnynek: I would rather not use the second parameter for index. I think scala does not use that for equality (only the first item is checked) and I doing think we want that. Also, I’d rather take R as an implicit parameter at the callsites that use it.

kailuowang:
@eli-jordan it looks like there are some duplicated commits created when you merge master to your branch. You might need to hard reset to master and cherry-pick `cd82d38` to produce a clean PR.

core/src/main/scala/cats/Representable.scala
```diff
@@ -0,0 +1,84 @@
...
+
+private[cats] sealed abstract class RepresentableInstances0 extends RepresentableInstances1 {
+  def catsRepresentableIsBimonad[F[_], R](implicit Rep: Representable.Aux[F, R], Mon: Monoid[R]): Bimonad[F] = new RepresentableBimonad[F, R] {
+    override def R: Representable.Aux[F, R] = Rep
+    override def M: Monoid[R] = Mon
+  }
+}
+
+private[cats] sealed abstract class RepresentableInstances1 {
+  def catsRepresentableIsMonad[F[_], R](implicit Rep: Representable.Aux[F, R]): Monad[F] = new RepresentableMonad[F, R] {
```
kailuowang: This and the one above are not implicit and not much point to be imported and thus no need to follow the implicits naming convention. If we want them to be available for auto derivation of `Bimonad` and `Monad` we need to put them in a trait extendable by `cats.implicits` and better have some tests for potential conflicts. I vote for just rename them to better searchable if there aren't many real world use cases for them.

eli-jordan:
@kailuowang I have made the suggested change, and cleaned up the commit history.

tests/src/test/scala/cats/tests/RepresentableSuite.scala
```diff
@@ -0,0 +1,75 @@
...
+import cats.laws.discipline.{BimonadTests, RepresentableTests, SerializableTests}
+import cats.{Bimonad, Eq, Id, Representable}
+import org.scalacheck.Arbitrary
+import cats.data.Kleisli
+
+class RepresentableSuite extends CatsSuite {
+
+  type Pair[A] = (A, A)
+
+  checkAll("Id[Int] <-> Unit => String", RepresentableTests[Id, Unit].representable[String])
```
kailuowang: another nit: `Id[String]`

tests/src/test/scala/cats/tests/RepresentableSuite.scala
```diff
@@ -0,0 +1,75 @@
...
+    implicit val andMonoid = new cats.Monoid[Boolean] {
+      def empty: Boolean = true
+      override def combine(x: Boolean, y: Boolean): Boolean = x && y
+    }
+
+    implicit val isoPair = Isomorphisms.invariant[Pair]
+    implicit val bimonadInstance = Representable.bimonad[Pair, Boolean]
+    checkAll("Pair[Int]", BimonadTests[Pair].bimonad[Int, Int, Int])
+    checkAll("Bimonad[Pair]", SerializableTests.serializable(Bimonad[Pair]))
+  }
```
kailuowang: would you test the `monad` as well?

core/src/main/scala/cats/data/RepresentableStore.scala
```diff
@@ -0,0 +1,51 @@
...
+
+  def map[B](f: A => B): RepresentableStore[F, S, B] = {
+    RepresentableStore(R.F.map(fa)(f), index)
+  }
+
+  /**
+   * Given a functorial computation on the index `S` peek at the value in that functor.
+   */
+  def experiment[G[_]](fn: S => G[S])(implicit G: Functor[G]): G[A] = {
+    G.map(fn(index))(peek)
```
kailuowang: mind to add a test for this method? I think a doctest would probably suffice.

tests/src/test/scala/cats/tests/RepresentableSuite.scala
eli-jordan: I didn't because `BimonadTests` runs the laws for both `Monad` and `Comonad` plus some others and the implementation is shared. Happy to add one of monad directly though.

tests/src/test/scala/cats/tests/RepresentableSuite.scala
eli-jordan: Nice catch! :-)

eli-jordan:
@kailuowang Thanks for the review. I have updated based on your comments.

LukaJCB:
I'm super late to this PR and I apoligize for that. I've never used representable functors before, but I took a look at the Haskell package you mentioned at the top and I see that it implies a `Distributive` instead of just `Functor`, is there a reason why you didn't choose to use our `Distributive` instead? :)

eli-jordan:
@LukaJCB Because you can actually derive `distribute` for any `Representable`

```haskell
distributeRep :: (Representable f, Functor w) => w (f a) -> f (w a)
distributeRep wf = tabulate (\k -> fmap (`index` k) wf)
```

I added the derived `Monad` and `Bimonad` instance, but did not add the `Distributive` because I figured it could be added at a later date if desired.

LukaJCB:
Sounds good :+1: We can add it later :)

tests/src/test/scala/cats/tests/RepresentableSuite.scala
kailuowang: Thanks for adding it.  It wasn't a big deal, but just one more red spot on the coverage report that can always cost some attention in the future. Also it would be nice to have it covered in case  
someone for whatever reason changes the monad implementation.

kailuowang:
Looking good to me. I am not very familiar with the practical use of it but the API looks clean and straightforward. Will be great to have two other review approvals.

kailuowang:
@johnynek any other issues?

core/src/main/scala/cats/syntax/representable.scala
```diff
@@ -0,0 +1,18 @@
...
+
+  implicit final def catsSyntaxIndex[F[_], A, R](fa: F[A])(implicit R: Representable.Aux[F, R]): IndexOps[F, A, R] =
+    new IndexOps[F, A, R](fa)
+}
+
+final class IndexOps[F[_], A, R](fa: F[A]) {
+  def index(implicit R: Representable.Aux[F, R]): R => A = R.index(fa)
+}
+
+final class TabulateOps[A, R](f: R => A) {
```
johnynek: Why not extend AnyVal here and above?

johnynek: If there is a reason can we comment?

eli-jordan: When I first wrote it, I had the ops class taking implicit params and AnyVal doesn't allow multiple param lists. I missed adding it back when I changed that, will fix that up now.

core/src/main/scala/cats/syntax/representable.scala
```diff
@@ -0,0 +1,18 @@
...
+
+trait RepresentableSyntax {
+  implicit final def catsSyntaxTabulate[A, R](f: R => A): TabulateOps[A, R] =
+    new TabulateOps[A, R](f)
+
+  implicit final def catsSyntaxIndex[F[_], A, R](fa: F[A])(implicit R: Representable.Aux[F, R]): IndexOps[F, A, R] =
+    new IndexOps[F, A, R](fa)
+}
+
+final class IndexOps[F[_], A, R](val fa: F[A]) extends AnyVal {
```
johnynek: Should this have R on the class? Maybe it doesn’t matter but it seems like it should be on index.

eli-jordan: In my tests it didn't seem to make a difference, but I think you're right, it makes more sense to put it on index. I'll make that change now.

core/src/main/scala/cats/package.scala
```diff
@@ -77,6 +77,18 @@ package object cats {
       override def isEmpty[A](fa: Id[A]): Boolean = false
   }
 
+  /**
+   * Witness for: Id[A] <-> Unit => A
```
johnynek: we could add something similar for `Eval[A]` potentially.

kailuowang: +1 , Since this has been blocking 1.2.0 release, I propose we tackle that in a separate PR
I added an issue https://github.com/typelevel/cats/issues/2320

Issue #2282, opened 2018-06-05
Add `groupByNes`


kubukoz:
Sometimes it'd be nice to group a Set to non-empty Sets, like:

```scala
Set("a", "aa", "aaa", "bbb").groupByNes(_.length) <-> SortedMap(
  1 -> NonEmptySet.one("a"),
  2 -> NonEmptySet.one("aa"),
  3 -> NonEmptySet.of("aaa", "bbb")
)
```

I suggest that a `SetSyntax` trait is added with syntax for this, similar to the existing `ListSyntax`.
The way `ListOps#groupByNel` is implemented (`list.toNel.groupBy(...)`)suggests that this could also involve having `groupBy` in `NonEmptySet`.

I'll gladly contribute that, given some pointers :)

Obarros:
@kubukoz I can take a look on this one.

kubukoz:
Sure, go for it :)

kubukoz:
Closed in #2285

PR #2285, opened 2018-06-09
Add groupBy for NonEmptySet

Obarros:
Should fixe #2282.

@LukaJCB PTAL.

### DIFF
--- a/core/src/main/scala/cats/data/NonEmptySet.scala
+++ b/core/src/main/scala/cats/data/NonEmptySet.scala
@@ -346,6 +346,21 @@ sealed class NonEmptySetOps[A](val value: NonEmptySet[A]) {
     */
   def zipWithIndex: NonEmptySet[(A, Int)] =
     NonEmptySetImpl.create(toSortedSet.zipWithIndex)
+
+  /**
+    * Groups elements inside this `NonEmptySet` according to the `Order`
+    * of the keys produced by the given mapping function.
+    */
+  def groupBy[B](f: A => B)(implicit B: Order[B]): NonEmptyMap[B, NonEmptySet[A]] = {
+     reduceLeftTo(a => NonEmptyMap.one(f(a), NonEmptySet.one(a))) { (acc, a) =>
+       val key = f(a)
+       val result = acc.lookup(key) match {
+         case Some(nes) => nes.add(a)
+         case _ => NonEmptySet.one(a)
+       }
+       acc.add((key, result))
+     }
+  }
 }
 
 private[data] sealed abstract class NonEmptySetInstances {
--- a/tests/src/test/scala/cats/tests/SyntaxSuite.scala
+++ b/tests/src/test/scala/cats/tests/SyntaxSuite.scala
@@ -1,8 +1,10 @@
 package cats
 package tests
 
+import scala.collection.immutable.SortedSet
+import scala.collection.immutable.SortedMap
 import cats.arrow.Compose
-import cats.data.{Binested, Nested}
+import cats.data.{Binested, Nested, NonEmptyList, NonEmptySet}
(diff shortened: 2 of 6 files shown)


core/src/main/scala/cats/data/NonEmptySet.scala
```diff
@@ -346,6 +346,21 @@ sealed class NonEmptySetOps[A](val value: NonEmptySet[A]) {
     */
   def zipWithIndex: NonEmptySet[(A, Int)] =
     NonEmptySetImpl.create(toSortedSet.zipWithIndex)
+
+  /**
+    * Groups elements inside this `NonEmptySet` according to the `Order`
+    * of the keys produced by the given mapping function.
+    */
+  def groupBy[B](f: A => B)(implicit B: Order[B]): NonEmptyMap[B, NonEmptySet[A]] = {
```
kubukoz: If we're returning `NonEmptyMap` here, does it make sense to make `NonEmptyList#groupBy` also return a non-empty map? @LukaJCB

LukaJCB: Yeah that's probably true, but we can't change it without breaking compatibility

kubukoz:
Right. I'm sooo excited for the 2.x branch ;)

kailuowang:
to make this available with `import cats.implicits._` , we also need to add this to this
[binary compatible trait](https://github.com/typelevel/cats/blob/d4bc3493d51c1a42c29e5eaed7922729f24286ad/core/src/main/scala/cats/syntax/all.scala#L62)

LukaJCB:
@kailuowang that's already there, no? https://github.com/typelevel/cats/pull/2285/files#diff-97bffe9b5ca6c2e744d09d9ce308043dR67

kailuowang:
Sorry I missed that.

LukaJCB:
What we did forget is to add a specific import for these syntax extensions, we should add an `object set extends SetSyntax` to here: https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/syntax/package.scala#L46

@Obarros

Obarros:
@LukaJCB any test that we can have for this?

kubukoz:
@Obarros I think there is a spec that checks if syntax compiles. Look for `SyntaxSuite` :)

Obarros:
Ping @kubukoz @LukaJCB!
Updated the `SyntaxSuite` with list and set syntax.

kubukoz:
looks good, :shipit:

tests/src/test/scala/cats/tests/SyntaxSuite.scala
```diff
@@ -358,5 +360,22 @@ object SyntaxSuite extends AllSyntaxBinCompat with AllInstances with AllSyntax {
...
+
+  def testNonEmptySet[A: Order, B: Order] : Unit = {
+    val f = mock[A => B]
+    val set = mock[SortedSet[A]]
+
+    val nes: Option[NonEmptySet[A]] = set.toNes
+    val grouped: SortedMap[B, NonEmptySet[A]] = set.groupByNes(f)
+  }
+
+  def testNonEmptyList[A: Order, B: Order] : Unit = {
```
LukaJCB: I think we don't need this `Order` constraint on `A`, leaving it on `B` should be enough. :)

Obarros: Thanks, @LukaJCB.
Let me remove it, I will also rebase on top of the latest master before squashing the commits and push.

kailuowang:
@obarros, it's no big deal but it's not necessary to squash before merge. We usually do squash merge (meaning all commits on pr will be squashed on merge by GitHub) . Having separate commits helps reviewing the latest changes.

PR #2291, opened 2018-06-10
Update Bitraverse docs 

Zelenya:
- update bisequence doc (not sure if it got any better)
- add bitraverse doctest example


Zelenya:
@ceedubs yeah, I've initially did `parseInt `and `parseBoolean`, but those preparations took more space than example itself.

PR #2292, opened 2018-06-10
Add foldMapK, reduceMapK

### LINKED ISSUE #2275: Add foldMapK and reduceMapK

kubukoz:
Closes #2275.

Not sure if we need to forward the target effect parameter of these functions in all the test classes, or whether we can just use `Id` or some other hardcoded type having a `MonoidK`, but for now I added appropriate implicits etc. everywhere.

### DIFF
--- a/core/src/main/scala/cats/syntax/reducible.scala
+++ b/core/src/main/scala/cats/syntax/reducible.scala
@@ -21,11 +21,13 @@ final class ReducibleOps0[F[_], A](val fa: F[A]) extends AnyVal {
    * Apply `f` to each element of `fa` and combine them using the
    * given `SemigroupK[G]`.
    *
-   * scala>import cats._, cats.implicits._
-   * scala>val f: Int => Endo[String] = i => (s => s + i)
-   * scala>val x: Endo[String] = NonEmptyList(1, 2, 3).reduceMapK(f)
-   * scala>val a = x("foo")
+   * {{{
+   * scala> import cats._, cats.data._, cats.implicits._
+   * scala> val f: Int => Endo[String] = i => (s => s + i)
+   * scala> val x: Endo[String] = NonEmptyList.of(1, 2, 3).reduceMapK(f)
+   * scala> val a = x("foo")
    * a: String = "foo321"
+   * }}}
    * */
   def reduceMapK[G[_], B](f: A => G[B])(implicit F: Reducible[F], G: SemigroupK[G]): G[B] =
     F.reduceLeftTo(fa)(f)((b, a) => G.combineK(b, f(a)))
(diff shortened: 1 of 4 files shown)


kubukoz:
BTW. we could also leave it as it is and use the `G` type instead of `Id` in the laws for `foldM`.

kailuowang:
I am not sure about this one. It seems that a more generic approach would be provide an importable conversion from MonoidK to Monoid, something like 
```scala
implicit def monoidKAlgebra[G[_], A])(implicit ev: MonoidK[G]): Monoid[G[A]] = ev.algebra
```
and then wherever a `Monoid[G[A]]` is needed you can just import this method. I am not sure how I feel about having a bunch of pairs of methods that one takes a `Monoid` and the other takes a `MonoidK`.

kubukoz:
I don't know about others, but I personally don't like implicit functions that aren't converting to extension classes (which is a workaround for no value classes in traits/classes)

LukaJCB:
I think that particular method doesn't work when a type has different instances for Monoid and MonoidK. Maybe @tpolecat can chime in as he was the one with a use case for it :)

kubukoz:
I just found another usecase for myself, WDYT about moving it to syntax for now (for bincompat) and continuing with it?

LukaJCB:
Sounds good to me! :)

ceedubs:
```scala
implicit def monoidKAlgebra[G[_], A])(implicit ev: MonoidK[G]): Monoid[G[A]] = ev.algebra
```

@kailuowang @kubukoz I don't think that we want to go down this path, as `MonoidK` and `Monoid` instances aren't always equivalent. For example in Cats the `Option` `Monoid` instance is based on combining with the inner `Semigroup` while the `MonoidK` instance is based on `orElse`. I think that there's a similar story for `Function1/Endo` but with function composition instead of `orElse`.

kailuowang:
It's a fair point that MonoidK and Monoid instance could be different. Let's proceed with adding this to bincompat syntax traits.

LukaJCB:
@kubukoz Are you still working on this? :)

kubukoz:
tbh I completely forgot about it. I think I'll start over (especially with all these conflicts) and get something out there soon :)

kubukoz:
@LukaJCB @kailuowang this one should be ready when the build is green :) sorry for the long wait

kubukoz:
Rebased to resolve conflicts after #2670

core/src/main/scala/cats/syntax/reducible.scala
```diff
@@ -9,3 +9,24 @@ trait ReducibleSyntax extends Reducible.ToReducibleOps {
...
+  implicit final def catsSyntaxReducibleOps0[F[_], A](fa: F[A]): ReducibleOps0[F, A] =
+    new ReducibleOps0[F, A](fa)
+}
+
+final class ReducibleOps0[F[_], A](val fa: F[A]) extends AnyVal {
+
+  /**
+   * Apply `f` to each element of `fa` and combine them using the
+   * given `SemigroupK[G]`.
+   *
```
LukaJCB: I think in order for this to be valid you need to open it with `{{` and close after, that's why it's not actually running the doctests.

kubukoz: True, plus there must be a space after each `scala>`. Fixed now :)

kubukoz:
Looks about green :) @lukajcb

kubukoz:
ping @kailuowang

PR #2293, opened 2018-06-12
added `liftTo` syntax to `Validated`

kailuowang:
so that you can lift a `Validated[E, A]` to any `F[A]` that has an `Applicative[F, E]`. 

Note that I placed the syntax in a `ValidatedOps` value class mainly work around the variance requirement.  (If the method is inside `Validated` you wouldn't be able to use `E` and `A` in invariant/contravariant position, so you would have to introduce an EE >: E and AA >: A but that requires extra type parameters to be passed in, which beat the purpose of this thing which is mostly ergonomic.

### DIFF
--- a/core/src/main/scala/cats/ApplicativeError.scala
+++ b/core/src/main/scala/cats/ApplicativeError.scala
@@ -175,7 +175,12 @@ trait ApplicativeError[F[_], E] extends Applicative[F] {
    * }}}
    */
   def fromEither[A](x: E Either A): F[A] =
-    x.fold(raiseError, pure)
+    x match {
+      case Right(a) => pure(a)
+      case Left(e)  => raiseError(e)
+    }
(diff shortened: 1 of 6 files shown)


core/src/main/scala/cats/ApplicativeError.scala
```diff
@@ -175,7 +175,12 @@ trait ApplicativeError[F[_], E] extends Applicative[F] {
    * }}}
    */
   def fromEither[A](x: E Either A): F[A] =
-    x.fold(raiseError, pure)
+    x match {
+      case Right(a) => pure(a)
+      case Left(e)  => raiseError(e)
+    }
```
kailuowang: replace `fold` to optimize performance ( this way there is no need to create anonymous functions, which take a toll. )

PR #2294, opened 2018-06-12
Add empty to State and StateT objects

sderosiaux:
Hi,

I had some `State`s where I needed to deal with empty values, such as:

```scala
def orEmpty[F[_]: MonadError[?[_], Unit]: SemigroupK, A: Monoid](s: Parser[F, A]) =
    s <+> StateT.pure(Monoid[A].empty)
```
In order to be explicit, why not add a `empty` constructor using a `Monoid[A]` that calls `pure` behind the scene? Note that `IndexedStateT` has `runEmpty` variants with the same `Monoid[A]` logic.

If you think it's an API overhead, feel free to discard and close the PR, no worry.

### DIFF
--- a/core/src/main/scala/cats/data/IndexedStateT.scala
+++ b/core/src/main/scala/cats/data/IndexedStateT.scala
@@ -173,6 +173,9 @@ private[data] trait CommonStateTConstructors {
   def pure[F[_], S, A](a: A)(implicit F: Applicative[F]): IndexedStateT[F, S, S, A] =
     IndexedStateT(s => F.pure((s, a)))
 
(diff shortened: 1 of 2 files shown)


kailuowang:
Could this be written as `s <+> A.emtpy.pure`? Finger crossed that ScalaC can infer the type.

sderosiaux:
@kailuowang No, I get diverging and ambiguous implicits (`cats.implicits._`). It works with `s <+> Monoid[A].empty.pure[Parser[F, ?]]` (which is the inverse notation of my initial example).

kailuowang:
yeah, unfortunately, scalac fails on inferring higher kinded types.

larsrh:
Rebased in #2533.

PR #2297, opened 2018-06-16
Make cats.data.AndThen public

alexandru:
I'm increasingly finding use-cases for `AndThen` and it is frustrating that I have to copy it around.

Now I need it in Monix, to work with `Iterant`'s new encoding. And I remember another use-case for it in cats-effect. As explained in the ScalaDoc, it's an incredibly useful data type for whenever you have to work with `A => F[B]`.

In such cases you can piggyback on `F[_]` if it implements a stack-safe `flatMap`, however `F[_]` can be more expensive than usage of `AndThen`.

I'd like `AndThen` to be in `cats-core`, because we need it for `IndexedStateT` anyway and because it is reusable.

N.B. this doesn't break binary compatibility and so it could be released in the 1.x series.

### DIFF
--- a/core/src/main/scala/cats/data/AndThen.scala
+++ b/core/src/main/scala/cats/data/AndThen.scala
@@ -1,6 +1,9 @@
-package cats.data
+package cats
+package data
 
 import java.io.Serializable
+import cats.arrow.{ArrowChoice, CommutativeArrow}
+
 
 /**
  * A function type of a single input that can do function composition
@@ -16,8 +19,48 @@ import java.io.Serializable
  *   // This should not trigger stack overflow ;-)
  *   f(0)
  * }}}
+ *
+ * This can be used to build stack safe data structures that make
+ * use of lambdas. The perfect candidates for usage with `AndThen`
+ * are the data structures using a signature like this (where
+ * `F[_]` is a monadic type):
+ *
+ * {{{
+ *   A => F[B]
+ * }}}
+ *
+ * As an example, if we described this data structure, the native
+ * solution for that `map` is stack unsafe:
+ *
+ * {{{
+ *   case class Resource[F[_], A, B](
+ *     acquire: F[A],
+ *     use: A => F[B],
+ *     release: A => F[Unit]) {
+ *
+ *     def flatMap[C](f: B => C)(implicit F: Functor[F]): Resource[F, A, C] = {
+ *       Resource(
+ *         ra.acquire,
+ *         // Stack Unsafe!
+ *         a => ra.use(a).map(f),
+ *         ra.release)
+ *     }
+ *   }
+ * }}}
+ *
+ * To describe a `flatMap` operation for this data type, `AndThen`
+ * can save the day:
+ *
+ * {{{
+ *   def flatMap[C](f: B => C)(implicit F: Functor[F]): Resource[F, A, C] = {
+ *     Resource(
+ *       ra.acquire,
+ *       AndThen(ra.use).andThen(_.map(f)),
+ *       ra.release)
+ *   }
+ * }}}
  */
-private[cats] sealed abstract class AndThen[-T, +R]
+sealed abstract class AndThen[-T, +R]
   extends (T => R) with Product with Serializable {
 
   import AndThen._
@@ -97,7 +140,7 @@ private[cats] sealed abstract class AndThen[-T, +R]
     "AndThen$" + System.identityHashCode(this)
 }
 
-private[cats] object AndThen {
+object AndThen extends AndThenInstances1 {
   /** Builds an [[AndThen]] reference by wrapping a plain function. */
   def apply[A, B](f: A => B): AndThen[A, B] =
     f match {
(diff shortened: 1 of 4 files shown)


core/src/main/scala/cats/data/AndThen.scala
```diff
@@ -124,3 +167,81 @@ private[cats] object AndThen {
...
+      def compose[A, B, C](f: AndThen[B, C], g: AndThen[A, B]): AndThen[A, C] =
+        f.compose(g)
+    }
+}
+
+private[data] abstract class AndThenInstances0 {
+  /**
+   * [[cats.Contravariant]] instance for [[AndThen]].
+   */
+  implicit def catsStdContravariantForAndThen[R]: Contravariant[AndThen[?, R]] =
```
LukaJCB: Minor nitpick, but the standard convention is using the package name, so `catsDataContravariantForAndThen` :)

alexandru: Ah, I had no idea from where `Std` comes from 🙂

core/src/main/scala/cats/data/AndThen.scala
```diff
@@ -124,3 +167,81 @@ private[cats] object AndThen {
    */
   private final val fusionMaxStackDepth = 127
 }
+
+private[data] abstract class AndThenInstances1 extends AndThenInstances0 {
+  /**
+   * [[cats.Monad]] instance for [[AndThen]].
+   */
+  implicit def catsStdMonadForAndThen[T]: Monad[AndThen[T, ?]] =
```
LukaJCB: Same thing here.

core/src/main/scala/cats/data/AndThen.scala
```diff
@@ -124,3 +167,81 @@ private[cats] object AndThen {
...
+        AndThen(f).compose(fa)
+
+      def tailRecM[A, B](a: A)(f: A => AndThen[T, Either[A, B]]): AndThen[T, B] =
+        AndThen(fn1.tailRecM(a)(f))
+    }
+
+  /**
+   * [[cats.ContravariantMonoidal]] instance for [[AndThen]].
+   */
+  implicit def catsStdContravariantMonoidalForAndThen[R : Monoid]: ContravariantMonoidal[AndThen[?, R]] =
```
LukaJCB: Here.

core/src/main/scala/cats/data/AndThen.scala
```diff
@@ -124,3 +167,81 @@ private[cats] object AndThen {
...
+      def product[A, B](fa: AndThen[A, R], fb: AndThen[B, R]): AndThen[(A, B), R] =
+        AndThen(fn1.product(fa, fb))
+    }
+
+  /**
+   * [[cats.arrow.ArrowChoice ArrowChoice]] and
+   * [[cats.arrow.CommutativeArrow CommutativeArrow]] instances
+   * for [[AndThen]].
+   */
+  implicit val catsStdArrowInstancesForAndThen: ArrowChoice[AndThen] with CommutativeArrow[AndThen] =
```
LukaJCB: And here.

alexandru:
I renamed the instances.

core/src/main/scala/cats/data/AndThen.scala
```diff
@@ -124,3 +167,81 @@ private[cats] object AndThen {
    */
   private final val fusionMaxStackDepth = 127
 }
+
+private[data] abstract class AndThenInstances1 extends AndThenInstances0 {
```
kailuowang: sorry to nitpick here, but the Cats conversion is to use 0 to represent the highest priority in implicit search. So we might want to switch the 0 and 1 here.

tests/src/test/scala/cats/tests/AndThenSuite.scala
```diff
@@ -1,9 +1,42 @@
...
+    checkAll("Semigroupal[AndThen[Int, ?]]", SerializableTests.serializable(Semigroupal[AndThen[Int, ?]]))
+  }
+
+  {
+    implicit val iso = SemigroupalTests.Isomorphisms.invariant[AndThen[?, Int]]
+    checkAll("AndThen[Int, Int]", ContravariantMonoidalTests[AndThen[?, Int]].contravariantMonoidal[Int, Int, Int])
+    checkAll("ContravariantMonoidal[AndThen[?, Int]]", SerializableTests.serializable(ContravariantMonoidal[AndThen[?, Int]]))
+  }
+
+  checkAll("AndThen[Int, Int]", MonadTests[Int => ?].monad[Int, Int, Int])
```
kailuowang: This is not testing `AndThen` but `Int => ?`

tests/src/test/scala/cats/tests/AndThenSuite.scala
```diff
@@ -1,9 +1,42 @@
...
+  checkAll("AndThen[Int, Int]", CommutativeArrowTests[AndThen].commutativeArrow[Int, Int, Int, Int, Int, Int])
+  checkAll("Arrow[AndThen]", SerializableTests.serializable(CommutativeArrow[AndThen]))
+
+  checkAll("AndThen[Int, Int]", ChoiceTests[AndThen].choice[Int, Int, Int, Int])
+  checkAll("Choice[AndThen]", SerializableTests.serializable(Choice[AndThen]))
+
+  checkAll("AndThen[Int, Int]", ArrowChoiceTests[AndThen].arrowChoice[Int, Int, Int, Int, Int, Int])
+  checkAll("ArrowChoice[AndThen]", SerializableTests.serializable(ArrowChoice[AndThen]))
+
+  checkAll("AndThen[Int, Int]", ContravariantTests[? => Int].contravariant[Int, Int, Int])
```
kailuowang: Same here. Not testing `AndThen`

core/src/main/scala/cats/data/AndThen.scala
```diff
@@ -124,3 +167,81 @@ private[cats] object AndThen {
...
+      def compose[A, B, C](f: AndThen[B, C], g: AndThen[A, B]): AndThen[A, C] =
+        f.compose(g)
+    }
+}
+
+private[data] abstract class AndThenInstances0 {
+  /**
+   * [[cats.Contravariant]] instance for [[AndThen]].
+   */
+  implicit def catsContravariantForAndThen[R]: Contravariant[AndThen[?, R]] =
```
kailuowang: sorry another nitpick: the Cats [naming convention for implicits](https://typelevel.org/cats/guidelines.html#a-idimplicit-naming-hrefimplicit-naminga-implicit-naming) requires the package name to be included as well. So these should be `catsDataContravariantForAndThen`, but since they are in the companion object it doesn't matter much, just to be consistent.

alexandru:
@kailuowang I fixed your comments.

Issue #2228, opened 2018-04-13
Add MonadError instance for OptionT where F is a Monad


LukaJCB:
Currently we have `(Monad f) => Monad (OptionT f)` as well as `(MonadError f e) => MonadError (OptionT f) e`, but we could just as well change that first definition to have a `MonadError[OptionT[F, ?], Unit]`. :)

LukaJCB:
I've even got a free implementation to give away:
```scala
private[effect] trait OptionTMonadError[F[_]] extends MonadError[OptionT[F, ?], Unit] {
  implicit def F: Monad[F]

  def pure[A](a: A): OptionT[F, A] = OptionT.pure(a)

  def flatMap[A, B](fa: OptionT[F, A])(f: A => OptionT[F, B]): OptionT[F, B] = fa.flatMap(f)

  override def map[A, B](fa: OptionT[F, A])(f: A => B): OptionT[F, B] = fa.map(f)

  def tailRecM[A, B](a: A)(f: A => OptionT[F, Either[A, B]]): OptionT[F, B] =
    OptionT(F.tailRecM(a)(a0 => F.map(f(a0).value)(
      _.fold(Either.right[A, Option[B]](None))(_.map(b => Some(b): Option[B]))
    )))

  def raiseError[A](e: Unit): OptionT[F, A] = OptionT.none

  def handleErrorWith[A](fa: OptionT[F, A])(f: Unit => OptionT[F, A]): OptionT[F, A] =
    OptionT(F.flatMap(fa.value) {
      case s @ Some(_) => F.pure(s)
      case None => f(()).value
    })
}
```

oleg-py:
Btw, is `MonadError[Option, Unit]` actually useful?

LukaJCB:
As useful as it ever as, I wager ;)

abhishek7:
@LukaJCB I'm new to Cats and wanted to help contribute. While I'm not familiar with the code base yet, I'm happy to give this a try if we think it's appropriate for a newcomer?

LukaJCB:
Sure, go right ahead, I'll guide you through it if you hit any hurdles! :)

abhishek7:
@LukaJCB, appears that this already had a PR merged. I guess we can go ahead and close this open issue? I'll try to pick up another open one.

PR #2299, opened 2018-06-18
2228-add MonadError instance for optionT where F is a Monad

heyrutvik:
related to #2228


core/src/main/scala/cats/data/OptionT.scala
```diff
@@ -287,6 +290,29 @@ private[data] trait OptionTMonad[F[_]] extends Monad[OptionT[F, ?]] {
     )))
 }
 
+private[data] trait OptionTMonadErrorMonad[F[_]] extends MonadError[OptionT[F, ?], Unit] {
```
kubukoz: Wondering if it'd be possible to extend `OptionTMonad[F]` here to reduce duplication (like the trait below does).

heyrutvik: I see.. You are right. I should have mixin `OptionTMonad[F]`. Changing it..

PR #2306, opened 2018-06-24
Drop 2.10 support / update to 2.13-M4

larsrh:
There was some discussion in the [cats-dev](https://gitter.im/typelevel/cats-dev?at=5b2df1ab6ceffe4eba2a6e96) channel about dropping 2.10 support. The major reason is that it becomes increasingly cumbersome to support 2.10–2.13 in discipline.

I've browsed around the issue tracker here and didn't see any major discussion/consensus around 2.10 support, so I thought I'd give it a try and see what people say.


rossabaker:
I'd like to see a concrete example where 2.10 support causes friction, but once that appears, I'm in favor of dropping it.  cats-effect already did, and that means a huge swath of the ecosystem followed.

larsrh:
@rossabaker typelevel/discipline#54: There's no uniform ScalaCheck version for 2.10–2.13. Scalatest and Specs2 differ in the supported ScalaCheck versions, too.

travisbrown:
I'm :+1: on dropping 2.10 support, but I think it's worth having someone with `org.typelevel` rights look at the Sonatype stats, if only to have an argument ready when people complain (circe-core downloads for example have been ~3% 2.10 for a while, and most of those are for non-current circe versions, anyway).

larsrh:
## 2.10

stats1

## 2.11

stats2

## 2.12

stats3

kailuowang:
Only 147 Cats_core_2.10 1.1.0 downloads vs 46K+ 2.11 and 2.12 downloads (3 months after the 1.1.0 release), I think it's safe to drop 2.10 support post 1.2.0. I vote that we let 1.2.0 be the last release supporting scala 2.10

larsrh:
Alright. I'm going to release a 2.11+ version of discipline then, which means that Cats has to wait until 1.2.0 for proper 2.13 support.

kailuowang:
Sounds good to me. Thanks @larsrh

kailuowang:
thanks @larsrh I merged the conflict. will merge it after build green.

Issue #2309, opened 2018-06-28
Add `flatTransform` to `OptionT`


amilkov3:
```scala
def flatTransform[F[_]: FlatMap, A, B](
  t: OptionT[F, A], 
  f: Option[A] => F[Option[B]]
): OptionT[F, B] = OptionT(t.value.flatMap(f))
```

abhishek7:
@amilkov3 was giving this a try, but a bit confused with `OptionT(t.value.flatMap(f))` on the last line. Isn't `flatMap` not a member of `F[Option[A]]`?

amilkov3:
@abhishek7 it is if you give `F` a `FlatMap` constraint

kubukoz:
Since only `flatMap` is needed, I'd leave a constraint just for `FlatMap`.

abhishek7:
@amilkov3 I like the alternative you provided! If it's alright with you, I'll open up a quick PR with that implementation along with a test? Let me know if you, @kubukoz, or anyone else has any other suggestions.

kubukoz:
Closed by #2314

PR #2314, opened 2018-07-07
Adding on flatTransform to OptionT as suggested. (Issue #2309)

abhishek7:
#2309

### DIFF
--- a/tests/src/test/scala/cats/tests/OptionTSuite.scala
+++ b/tests/src/test/scala/cats/tests/OptionTSuite.scala
@@ -290,7 +290,7 @@ class OptionTSuite extends CatsSuite {
     }
   }
 
-  test("flatTransform consistent with value.map") {
+  test("flatTransform consistent with value.flatMap") {
     forAll { (o: OptionT[List, Int], f: Option[Int] => List[Option[String]]) =>
       o.flatTransform(f) should === (OptionT(o.value.flatMap(f)))
     }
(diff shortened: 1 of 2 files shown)


tests/src/test/scala/cats/tests/OptionTSuite.scala
```diff
@@ -290,6 +290,12 @@ class OptionTSuite extends CatsSuite {
     }
   }
 
+  test("flatTransform consistent with value.map") {
```
kubukoz: should be `with value.flatMap`

kubukoz:
Do you think we should add this for `EitherT` and `IorT` in some way too? Once this is merged we could prepare additional PRs for them.

I'm still not sure about the name, so I'd wait for one of the maintainers to take a look and decide.

tests/src/test/scala/cats/tests/OptionTSuite.scala
abhishek7: Good catch, will update.

abhishek7:
@kubukoz I think we should, and I do agree with you that it would make sense to have one of the maintainers decide. I also agree that separate PRs for each would be the way to go if the decision is to go ahead with it.

abhishek7:
@kubukoz, any particular maintainers we should request to take a look at this?

LukaJCB:
I think `flatTransform` is probably fine in terms of naming :)

abhishek7:
@kubukoz @LukaJCB friendly ping on this.

kubukoz:
I'm not a maintainer so I'll just delegate to Luka ;)

abhishek7:
Thanks @LukaJCB! It appears that I don't have the ability to merge. @kailuowang, @LukaJCB, not sure if you're able to?

PR #2315, opened 2018-07-07
Add choice syntax

catostrophe:
[Choice ](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/arrow/Choice.scala) doesn't have a separate syntax object in `cats.syntax` package and its syntax (`choice`, `|||`) is not accessible via `import cats.implicits._`


kailuowang:
Thanks @catostrophe

PR #2318, opened 2018-07-08
Remove useless Order constraint in NonEmptyMap constructors

Kambius:
Constructors which take `SortedMap` as a parameter don't need to take extra `Order` implicit as it is already there.

### DIFF
--- a/core/src/main/scala/cats/data/NonEmptyMapImpl.scala
+++ b/core/src/main/scala/cats/data/NonEmptyMapImpl.scala
@@ -31,15 +31,15 @@ private[data] object NonEmptyMapImpl extends NonEmptyMapInstances with Newtype2 
   private[cats] def unwrap[K, A](m: Type[K, A]): SortedMap[K, A] =
     m.asInstanceOf[SortedMap[K, A]]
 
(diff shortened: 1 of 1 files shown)


Kambius:
@johnynek yes, but it's not a big deal to fix client code with this changes. `NonEmptyMap` isn't used widely so maybe we could break binary compatibility? What other options we have if it's not the case?

LukaJCB:
The other option is to wait until 2.0 unfortunately :/

larsrh:
Couldn't the same hack apply here as elsewhere? Duplicate the methods and mark them as `private[cats]`?

diesalbla:
@LukaJCB Was this Pull Request applied on a different form? The `cats 2.0` already came.

diesalbla:
Would it work to introduce a binary-compatible yet deprecated method, which used no implicits for the `Order` paramter, but which would generate the same bytecode-level, uncurrified method? 

```Scala
  def fromMap[K, A](as: SortedMap[K, A], k: Order[K]): Option[NonEmptyMap[K, A]] = 
    this.fromMap(as)
```

joroKr21:
Superseded by #3473

Issue #2152, opened 2018-01-12
IntelliJIDEA's support for simulacrum disabled in cats


kailuowang:
IntelliJIDEA's support for simulacrum is disabled by this [trick](https://github.com/typelevel/cats/blob/master/build.sbt#L40-L44) to enable compile-time-only dependency. 

See discussion here. 
https://youtrack.jetbrains.com/issue/SCL-13221

IntelliJ is a great tool to navigate the code base, without simulacrum support, it would be less useful in the Cats codebase.
 
3 Options I can see:
1. keep as is. wait for https://youtrack.jetbrains.com/issue/SCL-13221 to fix 
2. document the issue so that user can temporarily comment out the `% CompileTime` when using IntelliJ IDEA with cats. 
3. switch back to use `% Provided` and leave an issue to switch back when jetbrains adds support for it.

jgogstad:
Is this related to other incorrect red squiggles issues with cats and IntelliJ?

"Good code highlighted in red using cats mapN": https://youtrack.jetbrains.com/issue/SCL-12892
"False error (good code red): sequence from cats": https://youtrack.jetbrains.com/issue/SCL-10259

Issue #1549, opened 2017-03-08
Add support for Scala Native


djneades:
It is early days for Scala Native (https://github.com/scala-native/scala-native), but support would be be very helpful.

jcranky:
Perhaps it is a naive question, but would supporting scala native means in practice?

dneades:
That’s a good question. I have only just begun to look at Scala Native myself, but immediately ran into problems because none of my project’s dependencies had Scala Native support (hence this issue). I see that the scalaz folks have recently started to support Scala Native, so their work may give a few useful pointers as to what could be required: https://github.com/scalaz/scalaz/wiki/7.3.0-M10

I suspect that some of the Cats dependencies (e.g. Simulacrum) will need themselves to acquire Scala Native support before Cats itself can do so.

mscharley:
So, compiler plugins should continue to work and not need any changes. (I believe)

Any JVM dependencies however will need to be

aoiroaoino:
FYI
scalaz - https://github.com/scalaz/scalaz/pull/1322
Monocle - https://github.com/julien-truffaut/Monocle/pull/508

mkotsbak:
@mscharley 2. JVM dependencies should not be a problem as Cats is already supporting Scala.Js which also does not support any JVM dependencies.

4lex1v:
I'd to try Cats with some experimental Scala Native code, any chance that this issue will get some attention? How hard to add support for Scala Native?

johnynek:
I don't know of anyone working on it. Someone interested should just try. It is probably just some sbt surgery to get it to work. I very much doubt dependencies or jdk usage will be a barrier.

ChetanBhasin:
I would love to try out my hand at this, but I think a bigger bottleneck is Simulacrum at this point.

denisrosset:
Adding support for scala native to simulacrum was not a big deal; see https://github.com/mpilquist/simulacrum/pull/92

However, it looks like scalacheck/scalatest will not be supported any time soon!

schrepfler:
Just to highlight, if we get cats-kernel for SN, we can probably pull spire fore SN which would be really interesting to benchmark.

cb372:
I managed to cross-build cats-kernel for scala-native. Here's the [commit](https://github.com/cb372/cats/commit/f44b13fe0917710502c8a84fbf2b434b8f455809) with my changes. Happy to submit a PR if it's deemed useful.

[...]

Issue #2267, opened 2018-05-23
Publish for 2.13.0-M4


BennyHill:
Here we go again...

djspiewak:
I took a few minutes to look at this. In no particular order:

- [x] Updated `.travis.yml`
- [x] Updated sbt-scalajs
- [x] Updated scalatest
- [ ] Add sbt-crossbuild (since that mechanism is deprecated in the scalajs plugin)
- [ ] Patch the partial-unification plugin to no longer add `-Ypartial-unification` flag (it was removed in 2.13.0-M4) fiadliel/sbt-partial-unification#4
- [ ] Update catalysts, discipline, and machinist to 2.13.0-M4 (https://github.com/typelevel/machinist/pull/29, plus some issues)

kailuowang:
- [ ] update sbt-partial-unification dependency https://github.com/typelevel/cats/pull/2302
- [ ] update scalacheck https://github.com/rickynils/scalacheck/pull/411
- [ ] update scalatest https://github.com/scalatest/scalatest/pull/1383

SethTisue:
machinist 0.6.5 is out now for M4, so is ScalaTest 3.0.6-SNAP1

kailuowang:
Awesome, if nobody beats me, I'll create a PR enable 2.13M4 and drop 2.10

djspiewak:
1.3.x is kind of a long time to wait for 2.13 support. I'm a little vague on the timelines, but I get the impression that 2.13 will actually be *released* by that point. Given how foundational cats is to such a large swath of the ecosystem, waiting that long is probably not a good idea. I know the issues with scalacheck and what not, but it's probably worth fiddling around with annoying matrix builds, given the cross-build hell this kind of delay is about to unleash on the ecosystem.

johnynek:
This issue is another reminder of the non-zero costs of dependencies. It would be nice to see about minimizing dependencies (would some of our dependencies be open to being submodules, e.g. machinist, discipline, so more of the cats ecosystem is publishable atomically).

djspiewak:
[...]

PR #2323, opened 2018-07-11
Use portable-scala crossProject

BennyHill:
Also added native to the kernel project, just to get the ball rolling there - ref https://github.com/typelevel/cats/issues/1549

There was an issue with issue with the workaround used by simulacrum. However, and with reference to https://github.com/typelevel/cats/issues/2152, I switched it back to `Provided` _and_ then added a post process step to remove from the POM: so hopefully, that is now OK, too.


build.sbt
```diff
@@ -39,16 +38,23 @@ lazy val commonSettings = Seq(
...
     "org.typelevel" %%% "machinist" % "0.6.4",
     compilerPlugin("org.scalamacros" %% "paradise" % "2.1.0" cross CrossVersion.patch),
     compilerPlugin("org.spire-math" %% "kind-projector" % "0.9.6")
   ),
+  pomPostProcess := { (node: xml.Node) =>
+    new RuleTransformer(new RewriteRule {
+      override def transform(node: xml.Node): Seq[xml.Node] = node match {
+        case e: xml.Elem
+          if e.label == "dependency" && e.child.exists(child => child.label == "groupId" && child.text == "com.github.mpilquist") => Nil
+        case _ => Seq(node)
```
kailuowang: Is this blanket removing every dependencies under "com.github.mpilquist" for the sake of simulacrum. Is this worthwhile than just using `Provided`?

BennyHill: it just removes the simulacrum dependency the POM, a known concern, as done in the past elsewhere- eg https://github.com/typelevel/cats-effect/pull/77#issuecomment-340068924 .

But if you would rather a more fine controlled selection, by artifact name, sure-np

kailuowang: yeah, I think artifact name would be better, but if it's trickier than group name I'd vote for just leave it as "Provdided", not ideal, but didn't cause any problems in the past.

kailuowang:
Thanks @BennyHill !

BennyHill:
Any updates on this, please? As others are right now, just trying to cross things off my open list.

BennyHill:
Blimey....fast response!! Many thanks!

kailuowang:
Merged into 1.2.x branch which will be merged into master soon (supposedly every commits to 1.2.x should be automatically merge back to master, I just want to avoid creating too many PRs, we are going to deal with this 2 branch situation until we get a 2.13-M4 build released)

BennyHill:
Just fyi, this was originally done for the 2.13.0-m4 work, see https://github.com/typelevel/cats/issues/2267#issuecomment-395489518 . Looks like it is not actually required for that, so can be merged in as you require

PR #2325, opened 2018-07-16
Add Cofree.ana

andyscott:
Adds `Cofree.ana`, which generalizes `Cofree.unfold` with a fused `map`. The map fused version is a bit more useful than the basic unfold, especially when you're unfolding from some other messy structure.

I stuck with the name from the Haskell category extras package but I suppose `unfoldMap` (or `mapUnfold`?) might also be a good name for this.


LukaJCB:
Looks good to me :+1:

About naming, I think `ana` is a good name personally, but `anaE` is a bit strange, we have a bunch of functions that use `Eval` as a postfix, like `map2Eval` or `productREval` so maybe it's appropriate here too? :)

andyscott:
Seems reasonable to me. I'll unfold `E` to `Eval` 😄.

andyscott:
The build seems to have freaked out-- possibly related to GitHub being down briefly. I don't have permissions to restart it so maybe someone here could?

tpolecat:
done

LukaJCB:
Ready to merge @andyscott? :)

andyscott:
Yes, definitely 🤘

PR #2337, opened 2018-07-26
Update the outdated FAQ on IO/Task

### LINKED ISSUE #2326: FAQ on IO is out of date
https://typelevel.org/cats/faq.html#task is pretty outdated. Specifically, Cats-effect has support for parallelization. There may be other items that should be addressed, but I don't have much experience with Cats-effect or Monix. I'm hoping that someone with more experience with these projects would be willing to get us up to date.

cc @alexandru

ceedubs:
Resolves #2326.

This probably isn't the best answer, but at least it is more current
than what's in there now, and should work until someone more familiar
with these projects steps in and improves the doc.


kailuowang:
Thanks!

PR #2350, opened 2018-07-31
Remove unnecessary context bound in NonEmptyMap

NicolaeNMV:
Hello,

I was trying to construct a NonEmptyMap with fromMap, but it is context bound on Order[K]. This looks unnecessary unless I'm missing something. It is also inconsistent with https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/NonEmptySet.scala#L35 for example.

Thanks.

### DIFF
--- a/core/src/main/scala/cats/data/NonEmptyMapImpl.scala
+++ b/core/src/main/scala/cats/data/NonEmptyMapImpl.scala
@@ -31,10 +31,10 @@ private[data] object NonEmptyMapImpl extends NonEmptyMapInstances with Newtype2 
   private[cats] def unwrap[K, A](m: Type[K, A]): SortedMap[K, A] =
     m.asInstanceOf[SortedMap[K, A]]
 
-  def fromMap[K: Order, A](as: SortedMap[K, A]): Option[NonEmptyMap[K, A]] =
+  def fromMap[K, A](as: SortedMap[K, A]): Option[NonEmptyMap[K, A]] =
     if (as.nonEmpty) Option(create(as)) else None
 
-  def fromMapUnsafe[K: Order, A](m: SortedMap[K, A]): NonEmptyMap[K, A] =
+  def fromMapUnsafe[K, A](m: SortedMap[K, A]): NonEmptyMap[K, A] =
     if (m.nonEmpty) create(m)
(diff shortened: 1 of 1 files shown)


johnynek:
This would be fine, but it is not binary compatible so we can’t do it now.

oskoi:
Duplicates this PR #2318

NicolaeNMV:
Ok, we can close this PR. 
I've just realized that for constructing `SortedMap` that's needed for NonEmptyMap, you need an `implicit ord: scala.math.Ordering[A]` anyway. 
And `cats.kernel.Order` can be easily created from `scala.math.Ordering`.

I'm just not sure why such a restrictive datatype that needs `Ordering` was used as an underlying type for `NonEmptyMap` ? Was wondering maybe @LukaJCB can shed some light on that as I am genuinely curious.

LukaJCB:
The reason is that you either need an `Order` or a `Hash` to create a good `Map` implementation. Both should be parametrized using type classes. In that sense I don't think `Order` is really restrictive at all, you can define it for almost all data types :)

NicolaeNMV:
@LukaJCB I see your point. Thanks for the answer. But aren't there more datatypes implementing `Hash` than `Order` ?

LukaJCB:
Sorry, I meant the `Hash` type class. Universal `hashCode` and `equals` are very problematic and the type class based approach for these types is superior IMO. So if were to create a `Map` based on hashing instead of sorting, it'd be using the `Hash` type class, which you'd have to implement yourself again.

That said, have a look at kittens to get automatic derivation so that you no longer have to write these instances yourself :)

NicolaeNMV:
>it'd be using the Hash type class, which you'd have to implement yourself again.
Looks like there is a easy way to derive a `Hash` type class with `Hash.fromUniversalHashCode[A]` compared to `Order`, where there is no such universal method ?

It's good to know that there is automatic derivation of `Order` using kittens. But again, if we can remove its need altogether, why not do it and use the simpler `Hash` type class ?

At the moment I think it's quite counterintuitive that `NonEmptyMap` requires an implicit on `Order` when one would expect it to be closer to the default scala Map implementation HashMap. I think `NonEmptyList` was really nice in this sense.

NicolaeNMV:
Duplicates https://github.com/typelevel/cats/pull/2318

PR #2351, opened 2018-07-31
Add parTraverse_ and parSequence_ syntax #2341

### LINKED ISSUE #2341: Syntax for `parTraverse_` and `parSequence_`
Seems these are missing, should be fairly easy to add :)

denisrosca:
Closes #2341

### DIFF
--- a/core/src/main/scala/cats/syntax/parallel.scala
+++ b/core/src/main/scala/cats/syntax/parallel.scala
@@ -1,6 +1,6 @@
 package cats.syntax
 
-import cats.{Monad, Parallel, Traverse, FlatMap}
+import cats.{Monad, Parallel, Traverse, FlatMap, Foldable}
 
 trait ParallelSyntax extends TupleParallelSyntax {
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/syntax/parallel.scala
```diff
@@ -40,6 +54,12 @@ final class ParallelSequenceOps[T[_], M[_], A](val tma: T[M[A]]) extends AnyVal
     Parallel.parSequence(tma)
 }
 
+final class ParallelSequence_Ops[T[_], M[_], A](val tma: T[M[A]]) extends AnyVal {
```
kailuowang: Totally nitpicking, the `_` in the class name is a bit unconventional in this code base. Normally it wouldn't be a big deal, we could always change it after, but we can't really in cats....

denisrosca: I kind of agree, but couldn't come up with a better name. Any suggestions?

PR #2353, opened 2018-07-31
fixing the scalac option in unidoc

kailuowang:
add publishLocal back to build

### DIFF
--- a/scripts/travis-publish.sh
+++ b/scripts/travis-publish.sh
@@ -20,10 +20,10 @@
 
 sbt_cmd="sbt ++$TRAVIS_SCALA_VERSION"
 
-export publish_cmd=""
+export publish_cmd="publishLocal"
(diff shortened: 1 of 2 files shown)


scripts/travis-publish.sh
```diff
@@ -20,10 +20,10 @@
 
 sbt_cmd="sbt ++$TRAVIS_SCALA_VERSION"
 
-export publish_cmd=""
+export publish_cmd="publishLocal"
```
ChristopherDavenport: Is this right?

kailuowang: I believe so. It tests the ability to publish artifacts locally. It was originally there but I mistakenly removed it as being redundant in my previous PR.

kailuowang:
didn't work, closing for now.

PR #2361, opened 2018-08-03
Make kittens officially a Cats module (still separate repo)

kailuowang:
I had a discussion with @milessabin see (https://github.com/typelevel/kittens/issues/96). We think that making kittens an official cats module (still in its own separate repo) would be a good idea. @milessabin moved it to under github typelevel.  I'm proposing this to the Cats community through this PR.

Update: I updated the title to further clarify the intention.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ functionality, you can pick-and-choose from amongst these modules
  * [`cats-effect`](https://github.com/typelevel/cats-effect): standard `IO` type together with `Sync`, `Async` and `Effect` type classes 
  * [`cats-mtl`](https://github.com/typelevel/cats-mtl): transformer typeclasses for Cats' Monads, Applicatives and Functors.
  * [`mouse`](https://github.com/typelevel/mouse): a small companion to Cats that provides convenient syntax (aka extension methods) 
-
+ * [`kittens`](https://github.com/typelevel/kittens): automatic type class instance derivation for Cats and generic utility functions
 
 Release notes for Cats are available in [CHANGES.md](https://github.com/typelevel/cats/blob/master/CHANGES.md).
 


ceedubs:
I'm a bit hesitant on this. Kittens depends on Shapeless, right? Here are some concerns that I would have:

- The Cats builds are already cross-building against several Scala versions and Scala.js. I think that them potentially needing to cross-compile against multiple Shapeless versions someday would be rough (in a combinatorial way).
- Because of its more complex usage of the type system and heavier usage of macros, I could see Shapeless being quicker to drop older versions of Scala. This could get tricky to try to not cross-build just the kittens module for some scala versions.

I'm happy to see kittens in the `typelevel` org, but I question the cost/benefit ratio of making it a module within the cats repo.

kailuowang:
@ceedubs not a module inside cats repo, it's still in its own repo with its own release cycle just like the other 3 "outside" modules cats-mtl, cats-effect and mouse.

tpolecat:
Does "official cats module" mean anything other than appearing on this bullet list?

kailuowang:
@tpolecat that's mostly it. Although I'd say that cats modules get a commitment from Cats maintainers to help to maintain them.

ceedubs:
@kailuowang ah okay. I understand. I'm all for this then.

I do have a concern about kittens derivation that has prevented me from using it and I've been meaning to look for a discussion of, but I'll pursue that separately.

LukaJCB:
Regarding commitment to maintain, I'd happily help out, but I fear my shapeless skills really aren't up to par. I do think `kittens` is an important project that we (heh) should give more focus, so I really like this move :)

kailuowang:
seems that we have enough approval. It's official!

PR #2366, opened 2018-08-05
Add Foldable extension collectFirstSomeM

catostrophe:
Monadic version of `collectFirstSome`. Implemented as a syntax extension for binary compatibility.


core/src/main/scala/cats/syntax/foldable.scala
```diff
@@ -92,4 +94,29 @@ final class FoldableOps[F[_], A](val fa: F[A]) extends AnyVal {
...
+    * scala> val keys2 = List("1", "x", "4", "5")
+    * scala> keys2.collectFirstSomeM(parseInt(_) map map1.get)
+    * res3: scala.util.Either[String,Option[String]] = Left(For input string: "x")
+    * scala> val keys3 = List("1", "2", "4", "x")
+    * scala> keys3.collectFirstSomeM(parseInt(_) map map1.get)
+    * res4: scala.util.Either[String,Option[String]] = Right(Some(Four))
+    * }}}
+    */
+  def collectFirstSomeM[G[_], B](f: A => G[Option[B]])(implicit F: Foldable[F], G: Monad[G]): G[Option[B]] =
+    F.foldRight(fa, Eval.now(OptionT.none[G, B]))((a, lb) =>
```
kubukoz: Would it be possible to implement this without `OptionT`? I think there's some overhead involved when wrapping values in monad transformers, and this is library code so it shouldn't be that problematic to use the wrapped values directly

catostrophe:
What's the preferred way to restart a failed Travis build if no file changes needed? push force?

kubukoz:
Add an empty commit, I guess. `git commit --allow-empty`

PR #2368, opened 2018-08-07
fix typo

letusfly85:
modified `partial-unification` because it didn't work.


letusfly85:
@kailuowang 

Thank you for seeing!

PR #2369, opened 2018-08-07
Add a documentation page for Nested

cb372:
If anyone has more motivating examples, I'm happy to add them.

### DIFF
--- a/docs/src/main/tut/datatypes/nested.md
+++ b/docs/src/main/tut/datatypes/nested.md
@@ -1,0 +1,119 @@
+---
+layout: docs
+title:  "Nested"
+section: "data"
+source: "core/src/main/scala/cats/data/Nested.scala"
+scaladoc: "#cats.data.Nested"
+---
+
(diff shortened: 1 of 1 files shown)


cb372:
Yeah actually the example we discussed on Twitter originally used `IO` but I "downgraded" it to Future  :)

I guess porting the examples to `IO` would mean adding a library dependency on `cats-effect` for the docs module. That feels a little weird. Or move the docs module into its own repo? That has its downsides as well: it means you can't add a new feature and write the docs for it at the same time.

LukaJCB:
Yeah we've had the discussion before about Free and what we'd interpret into. I think we could definitely take a dependcy just for the docs module once cats-effect is stable :)

kailuowang:
@LukaJCB having cats-effect as a dependency in docs might make things hairy when we start to have breaking changes (say going 3.0), but I guess in that case we need to coordinate with Cats-effect anyway. So 👍

PR #2376, opened 2018-08-10
Updates 47 Degrees homepage URL

juanpedromoreno:
In the _Adopters_ section given that https://47deg.com is not working.


LukaJCB:
Thanks! :)

Issue #2364, opened 2018-08-05
Use composite projects in build


ghost:
Ref https://github.com/portable-scala/sbt-crossproject/pull/99

Alistair-Johnson:
I'll grab this one too :+1:

PR #2382, opened 2018-08-11
Removed not used sbt-scala-native plugin and fixed importing the project into intellij

LukaszMarchewka:
I have had a problem during importing the project into intellij (version 2018.2) . Intellij tried to download `scala-native` for scala 2.12. I have solved the problem by:
- removed not used `scala-native-plugin`
- set fixed scala in version 2.11 for a native kernel


johnynek:
Why do you say it is a dead line?

Looks to me like support for scala native in the kernel module.

BennyHill:
Wow....this work in progress for native, only recently added. More work is in the pipeline with new release of of sbt-crossproject like 2 days ago

BennyHill:
I have a PR for https://github.com/typelevel/cats/issues/2364 waiting to go. We could expedite this by merging as-is, and my new PR can tweak afterwards.

LukaszMarchewka:
@BennyHill sorry for delay, I wasn't able to update this faster. I have applied all your suggestions.

BennyHill:
@LukaszMarchewka  Thanks! 

You'll see in next PR for composite projects that the lines 

```scala
lazy val kernelJVM = kernel.jvm
lazy val kernelJS = kernel.js
lazy val kernelNative = kernel.native
```

and similar will all disappear...so I'm not just being picky here :smile:

LukaJCB:
@kailuowang Should we merge this for now, as it seems to resolve a common blocker for people wanting to contribute to cats?

Avasil:
@LukaJCB @kailuowang any progress? I actually know about some potential contributors that are blocked by this :P

PR #2394, opened 2018-08-13
removed kittens and mainecoon from list

kailuowang:
kittens is an external cats module now. 
mainecoon became cats-tagless which is yet to be announced, but likely to become an external cats module as well.


kailuowang:
merged with 1 maintainer approval as a simple website update.

PR #2402, opened 2018-08-14
Add Compose instance for Map

### LINKED ISSUE #2399: Add a `Compose` instance for `Map`
Should be relatively easy to define and lawful as well :)

denisrosca:
Closes #2399

### DIFF
--- a/core/src/main/scala/cats/instances/map.scala
+++ b/core/src/main/scala/cats/instances/map.scala
@@ -85,8 +85,19 @@ trait MapInstances extends cats.kernel.instances.MapInstances {
 
 trait MapInstancesBinCompat0 {
 
-  implicit def catsStdComposeForMap: Compose[Map] = new Compose[Map] {
+  implicit val catsStdComposeForMap: Compose[Map] = new Compose[Map] {
 
+    /**
+      * Compose two maps `g` and `f` by using the values in `f` as keys for `g`.
+      * {{{
+      * scala> import cats.arrow.Compose
+      * scala> import cats.implicits._
+      * scala> val first = Map(1 -> "a", 2 -> "b", 3 -> "c", 4 -> "a")
+      * scala> val second = Map("a" -> true, "b" -> false, "d" -> true)
+      * scala> Compose[Map].compose(second, first)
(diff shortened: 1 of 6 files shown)


tests/src/test/scala/cats/tests/MapSuite.scala
```diff
@@ -15,6 +15,8 @@ class MapSuite extends CatsSuite {
   checkAll("Map[Int, Int] with Option", UnorderedTraverseTests[Map[Int, ?]].unorderedTraverse[Int, Int, Int, Option, Option])
   checkAll("UnorderedTraverse[Map[Int, ?]]", SerializableTests.serializable(UnorderedTraverse[Map[Int, ?]]))
 
+  checkAll("Compose[Map]", ComposeTests[Map].compose[Int, Long, String, Double])
```
kailuowang: nit, can we add the `SerializableTests` as the other instances here?

core/src/main/scala/cats/instances/map.scala
```diff
@@ -79,4 +80,23 @@ trait MapInstances extends cats.kernel.instances.MapInstances {
 
     }
   // scalastyle:on method.length
+
+}
+
+trait MapInstancesBinCompat0 {
+
+  implicit def catsStdComposeForMap: Compose[Map] = new Compose[Map] {
```
ceedubs: Should this be a `val` instead of a `def` so the same instance can be reused repeatedly?

denisrosca:
@ceedubs Would a doctest like this be enough?

    /**
      * Compose two maps `g` and `f` by using the values in `f` as keys for `g`.
      * {{{
      * scala> import cats.arrow.Compose
      * scala> import cats.implicits._
      * scala> val first = Map(1 -> "a", 2 -> "b", 3 -> "c", 4 -> "a")
      * scala> val second = Map("a" -> true, "b" -> false, "d" -> true)
      * scala> Compose[Map].compose(second, first)
      * res0: Map[Int, Boolean] = Map(1 -> true, 2 -> false, 4 -> true)
      * }}}
      */
      def compose[A, B, C](f: Map[B, C], g: Map[A, B]): Map[A, C]

ceedubs:
@denisrosca that example looks great!

PR #2405, opened 2018-08-15
Add FunctorFilter and TraverseFilter

### LINKED ISSUE #2348: Move FunctorEmpty and TraverseEmpty back into cats-core?
`FunctorEmpty` and `TraverseEmpty` (used to be `FunctorFilter` and `TraverseFilter` were moved to `cats-mtl` when that was extracted from cats-core, as it was one of multiple leaves for `Functor`.
However, when I look at these two type classes, they seem to be awfully unfit for `cats-mtl`, as they don't have anything at all to do with monad transformers and can't even be lifted through monad stacks. I suspect the only reason they were moved at all was to use the new composition-based type class encoding (which I agree is a good reason). 

I propose moving them back along with their instances as they are really useful and we shouldn't have any trouble using the composition-based type class encoding (we already use it for things like `Parallel`). Interested in hearing thoughts :)

LukaJCB:
Should fix #2348. Needs to be coordinated with their removal from `cats-mtl` :)


core/src/main/scala/cats/FunctorEmpty.scala
```diff
@@ -0,0 +1,28 @@
...
+  def flattenOption[A](fa: F[Option[A]]): F[A] =
+    mapFilter(fa)(identity)
+
+  def filter[A](fa: F[A])(f: A => Boolean): F[A] =
+    mapFilter(fa)(a => if (f(a)) Some(a) else None)
+}
+
+object FunctorEmpty {
+  implicit def catsFunctorForFunctorEmpty[F[_]](fe: FunctorEmpty[F]): Functor[F] =
+    fe.functor
```
LukaJCB: This function here allows us to call things like `FunctorEmpty[F].map`, do you think this is worthwhile? I feel like this is an easy change that doesn't cause any ambiguities and let's us atleast have some of the niceties that we lose when doing composition instead of inheritance :)

tests/src/test/scala/cats/tests/ChainSuite.scala
```diff
@@ -25,6 +25,9 @@ class ChainSuite extends CatsSuite {
   checkAll("Chain[Int]", OrderTests[Chain[Int]].order)
   checkAll("Order[Chain]", SerializableTests.serializable(Order[Chain[Int]]))
 
+  checkAll("Chain[Int]", TraverseEmptyTests[Chain].traverseEmpty[Int, Int, Int])
+  checkAll("TraverseEmpty[Chain]", SerializableTests.serializable(TraverseEmpty[Chain]))
```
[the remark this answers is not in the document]
ceedubs: I think that I got it -- https://github.com/LukaJCB/cats/pull/4

core/src/main/scala/cats/FunctorEmpty.scala
kailuowang: I am not sure if it's needed either. 
What would be the use case of it? 
If user `import FunctorEmpty._` and all of a sudden whenever there is a `FunctorEmpty` in scope there is also a `Functor` in scope, then will there be ambiguities with other possible `Functor` instances?
Should user be just writing the following instead?  
```def foo[F[_]: FunctorEmpty : Functor]: ....```

Comment (LukaJCB) [reply core/src/main/scala/cats/FunctorEmpty.scala]: There won't be ambiguities, because only an explicit instance will be converted, notice it is `(fe: FunctorEmpty[F]): Functor[F]` and not `(implicit fe: FunctorEmpty[F]): Functor[F]`.

The only real benefit is that when you have a method like this:
```scala
def foo[F[_]](fa: F[Int])(implicit F: FunctorEmpty[F]): F[String] =
  F.map(fa)(_.toString) 
```
You'll be able to call `map` et al. directly on it instead of doing `F.functor.map`. So the benefit is rather small, but at the same time it doesn't cause ambiguities. Also note that `import FunctorEmpty._` won't actually do anything as this instance will always be in scope whenever `FunctorEmpty` is in scope :)

kailuowang: ah, sorry I missed that `fe` isn't implicit. As you said, the benefit is a bit small, especially since you usually don't need to write `implicit F: FunctorEmpty[F]` at the first place, you could just use context bound, so not much typing to be saved. 
I'd prefer encouraging users to the more conventional type class syntax enabled style. 
```scala
def foo[F[_]: FunctorEmpty : Functor](fa: F[Int]): F[String] = {
   fa.map(_.toString).filter(_.length > 2)
}
```

LukaJCB: Sounds good to me :)

core/src/main/scala/cats/FunctorEmpty.scala
```diff
@@ -0,0 +1,28 @@
...
+
+import simulacrum.typeclass
+
+/**
+ * `FunctorEmpty[F]` allows you to `map` and filter out elements simultaneously.
+ */
+@typeclass
+trait FunctorEmpty[F[_]] extends Serializable {
+  def functor: Functor[F]
+
```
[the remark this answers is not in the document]
LukaJCB: Can we define any extra methods if we require implementors to implement an `empty` method?  Or just for the laws? The Haskell precedent doesn't seem to define a method like this and I'm not sure if we should either, need to think about what the actual benefits are 🤔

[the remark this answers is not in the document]
johnynek: well, I guess if you have an empty, since you can't combine it, you are always stuck with an empty. But still, that's interesting for laws `empty.map(fn) == empty` etc...

I hate to have a hacky way to make an empty, but not provide it (namely give me any instance, and I will filter it).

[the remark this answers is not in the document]
LukaJCB: Here's some discussion on it https://github.com/typelevel/cats/issues/1365 and here's the haskell counter part http://hackage.haskell.org/package/witherable-0.2/docs/Data-Witherable.html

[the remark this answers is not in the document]
LukaJCB: The original PR also has some discussion on why it wasn't included in the first place https://github.com/typelevel/cats/pull/1225/files#discussion_r71987653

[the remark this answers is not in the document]
ceedubs: More specifically, this comment: https://github.com/typelevel/cats/pull/1225#discussion_r71981348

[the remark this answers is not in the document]
LukaJCB: Yeah after thinking about it more a bit, I think I'd vote to keep it as is without the empty method.

core/src/main/scala/cats/TraverseEmpty.scala
```diff
@@ -0,0 +1,26 @@
...
+ * combined operation (`traverseFilter`).
+ *
+ * Based on Haskell's [[https://hackage.haskell.org/package/witherable-0.1.3.3/docs/Data-Witherable.html Data.Witherable]]
+ */
+
+@typeclass
+trait TraverseEmpty[F[_]] extends FunctorEmpty[F] {
+  def traverse: Traverse[F]
+
+  override def functor: Functor[F] = traverse
```
rossabaker: I haven't spent a lot of time on the this type of encoding.  Is there a reason that `functor` should ever _not_ be the `traverse`?  In other words, do we lose anything by making this final?

non:
I understand the arguments against providing `empty` -- but given that what we have is a functor that can be filtered, I'd prefer a name like `FunctorFilter` or `FilterableFunctor` or something similar. I think mentioning empty in the name is not a good idea if we don't want to require that method.

core/src/main/scala/cats/TraverseEmpty.scala
LukaJCB: Nope, I agree fully, I'll change it :)

LukaJCB:
@non I'm fine with renaming them back to `TraverseFilter` and `FunctorFilter` as they were originally called. :)

ceedubs:
When I originally added these in #1225, I had some doctest examples. It would be nice to see those come back. Also is any of the compose work in that PR still relevant?

LukaJCB:
Hi Cody, I'll add those doctests back :)
W.r.t to compose, I just added instances for `Nested`, not sure if we also need the explicit `compose` 🤔

core/src/main/scala/cats/data/OptionT.scala
```diff
@@ -234,6 +234,20 @@ private[data] sealed abstract class OptionTInstances extends OptionTInstances0 {
       def defer[A](fa: => OptionT[F, A]): OptionT[F, A] =
         OptionT(F.defer(fa.value))
     }
+
+  implicit def optionTFunctorFilter[F[_]: Functor]: FunctorFilter[OptionT[F, ?]] = {
```
ceedubs: There was a `TraverseFilter` instance for `OptionT` in #1225. Is there an issue with it?

core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -514,6 +515,28 @@ private[data] abstract class EitherTInstances extends EitherTInstances1 {
...
+          F0.traverseFilter[G, Either[L, A], Either[L, B]](fa.value) {
+            case l@Left(_) => G.pure(Option(l.rightCast[B]))
+            case Right(a) => G.map(f(a))(_.map(Either.right))
+          })(EitherT(_))
+
+      override def filterA[G[_], A]
+      (fa: EitherT[F, L, A])
+      (f: A => G[Boolean])
+      (implicit G: Applicative[G]): G[EitherT[F, L, A]] =
+        G.map(F0.filterA(fa.value)(_.fold(_ => G.pure(true), f)))(EitherT[F, L, A])
```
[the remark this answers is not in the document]
LukaJCB: This is just the implementation we had in cats-mtl, so I'm not sure what the motivation was. I agree that it seems weird though, maybe we should delete it. Did you provide one in your initial PR?

core/src/main/scala/cats/data/OptionT.scala
LukaJCB: No idea, just took what was there from `cats-mtl`

core/src/main/scala/cats/data/EitherT.scala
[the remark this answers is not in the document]
ceedubs: No, I didn't. I'd be inclined to leave it out until someone makes a good argument for it.

core/src/main/scala/cats/data/OptionT.scala
ceedubs: I'm having trouble tracking down the history in cats-mtl, but I think that we should have the `OptionT` instance. It should be have identically to the corresponding `Nested` instance.

core/src/main/scala/cats/data/EitherT.scala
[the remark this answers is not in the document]
LukaJCB: Yeah agree, can always add it back later :+1:

LukaJCB:
@ceedubs I pushed a new commit, addressing your points :)

core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -514,6 +514,8 @@ private[data] abstract class EitherTInstances extends EitherTInstances1 {
       def defer[A](fa: => EitherT[F, L, A]): EitherT[F, L, A] =
         EitherT(F.defer(fa.value))
     }
+
+
```
ceedubs: Minor: extra newline here

core/src/main/scala/cats/data/Nested.scala
```diff
@@ -54,13 +53,27 @@ private[data] sealed abstract class NestedInstances extends NestedInstances0 {
...
         Nested(F.defer(fa.value))
     }
+
+  implicit def catsDataTraverseFilterForNested[F[_], G[_]](implicit F0: Traverse[F], G0: TraverseFilter[G]): TraverseFilter[Nested[F, G, ?]] =
+    new NestedTraverseFilter[F, G] {
+      implicit val F: Traverse[F] = F0
+      implicit val G: TraverseFilter[G] = G0
+    }
+
+
```
ceedubs: Minor: extra newline here.

PR #2371, opened 2018-08-08
Add Chain

LukaJCB:
Adds Catenable from fs2 and fixes #2358 

Still needs:

- [x] tests 
- [x] law tests
- [x] license info
- [x] benchmarks
- [x] Eq, PartialOrder, Order, Coflatmap instances? (could add those in a later PR)
- [ ] Docs (couls also be added in a later PR)

PR #2406, opened 2018-08-15
Add NonEmptyChain

LukaJCB:
Based on the work in #2371, though we might still need to change a couple of things first, including the `Arbitrary` Generator.

### DIFF
--- a/core/src/main/scala/cats/data/Chain.scala
+++ b/core/src/main/scala/cats/data/Chain.scala
@@ -320,9 +320,9 @@ sealed abstract class Chain[+A] {
   /**
    * Returns the number of elements in this structure
    */
-  final def length: Int = {
+  final def length: Long = {
     val iter = iterator
-    var i: Int = 0
+    var i: Long = 0
     while(iter.hasNext) { i += 1; iter.next; }
     i
   }
@@ -330,7 +330,7 @@ sealed abstract class Chain[+A] {
   /**
    * Alias for length
    */
-  final def size: Int = length
+  final def size: Long = length
 
 
   /**
@@ -555,7 +555,7 @@ private[data] sealed abstract class ChainInstances extends ChainInstances1 {
       def coflatMap[A, B](fa: Chain[A])(f: Chain[A] => B): Chain[B] = {
         @tailrec def go(as: Chain[A], res: ListBuffer[B]): Chain[B] =
           as.uncons match {
-            case Some((h, t)) => go(t, res += f(t))
+            case Some((_, t)) => go(t, res += f(as))
             case None => Chain.fromSeq(res.result())
           }
 
--- a/core/src/main/scala/cats/data/Chain.scala
+++ b/core/src/main/scala/cats/data/Chain.scala
@@ -1,11 +1,11 @@
 package cats
 package data
 
-import cats.implicits._
 import Chain._
 
 import scala.annotation.tailrec
 import scala.collection.immutable.SortedMap
+import scala.collection.mutable.ListBuffer
 
 /**
  * Trivial catenable sequence. Supports O(1) append, and (amortized)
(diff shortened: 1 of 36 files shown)


core/src/main/scala/cats/data/NonEmptyChain.scala
```diff
@@ -0,0 +1,462 @@
...
+
+  def one[A](a: A): NonEmptyChain[A] = create(Chain.one(a))
+
+  implicit def catsNonEmptyChainOps[A](value: NonEmptyChain[A]): NonEmptyChainOps[A] =
+    new NonEmptyChainOps(value)
+}
+
+
+sealed class NonEmptyChainOps[A](val value: NonEmptyChain[A]) {
+
```
johnynek: Can this extend AnyVal?

core/src/main/scala/cats/data/NonEmptyChain.scala
```diff
@@ -0,0 +1,462 @@
...
+  }
+
+  /**
+   * Reduce using the Semigroup of A
+   */
+  final def reduce[AA >: A](implicit S: Semigroup[AA]): AA = {
+    val iter: Iterator[AA] = toChain.iterator
+    var result = iter.next
+    while (iter.hasNext) { result = S.combine(result, iter.next) }
+    result
```
johnynek: We should use combineAllOption.get so we don’t deoptimize combineAll

LukaJCB:
Urgh, I'm apparently really bad at using git 🤦🏻‍♂️

laws/src/main/scala/cats/laws/discipline/Arbitrary.scala
```diff
@@ -76,6 +76,12 @@ object arbitrary extends ArbitraryInstances0 {
   implicit def catsLawsCogenForNonEmptyList[A](implicit A: Cogen[A]): Cogen[NonEmptyList[A]] =
     Cogen[List[A]].contramap(_.toList)
 
+  implicit def catsLawsArbitraryForNonEmptyChain[A](implicit A: Arbitrary[A]): Arbitrary[NonEmptyChain[A]] =
+    Arbitrary(implicitly[Arbitrary[List[A]]].arbitrary.flatMap(fa => A.arbitrary.map(a => NonEmptyChain(a, fa: _*))))
+
```
johnynek: Can we use the Arbitrary for Chain and add one more to the front or back also as a variant?

kailuowang:
@LukaJCB it's not you are bad at using git, it's just the other PR is merged squashed. Since the changes look good, I think if we squash this one as well it will be okay.

core/src/main/scala/cats/data/NonEmptyChain.scala
```diff
@@ -0,0 +1,456 @@
...
+      override def exists[A](fa: NonEmptyChain[A])(p: A => Boolean): Boolean =
+        fa.exists(p)
+
+      override def toList[A](fa: NonEmptyChain[A]): List[A] = fa.toChain.toList
+
+      override def toNonEmptyList[A](fa: NonEmptyChain[A]): NonEmptyList[A] =
+        fa.toNonEmptyList
+    }
+
+  implicit def catsDataEqForNonEmptyChain[A: Eq]: Eq[NonEmptyChain[A]] =
```
johnynek: we only have `Eq` here, no Order, PartialOrder, right? Can we add those so we get the priority right there?

core/src/main/scala/cats/data/Chain.scala
```diff
@@ -320,17 +320,17 @@ sealed abstract class Chain[+A] {
   /**
    * Returns the number of elements in this structure
    */
-  final def length: Int = {
+  final def length: Long = {
     val iter = iterator
-    var i: Int = 0
+    var i: Long = 0
```
kailuowang: just curious, did we actually tried one instance with more than 2 billion items?

LukaJCB: No I don't think so, I just wanted to be consistent with NonEmptyList and NonEmptyVector :)

LukaJCB:
Merging with two signoffs :)

PR #2408, opened 2018-08-15
Add Wegtam GmbH to adopters.

jan0sch:


johnynek:
it would be interesting to put some files that definitely have no CI impact into a system that allows the tests to be skipped.

LukaJCB:
Indeed, especially as our CI is already slow as is 😬

jan0sch:
It seems this is possible: https://docs.travis-ci.com/user/customizing-the-build/#skipping-a-build

Should have looked this up beforehand. :-|
Maybe we should add it to the adopters section of the readme?

PR #2416, opened 2018-08-16
add append method to NEL

julien-truffaut:


core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -100,6 +100,15 @@ final case class NonEmptyList[+A](head: A, tail: List[A]) {
   def prepend[AA >: A](a: AA): NonEmptyList[AA] =
     NonEmptyList(a, head :: tail)
 
+  /**
+    * Alias for append
+    */
+  def :+[AA >: A](a: AA): NonEmptyList[AA] =
```
martin-g: this method should be `final`

julien-truffaut: sure but I would prefer we do that in a separate PR as there is no final method in NEL yet

julien-truffaut:
@LukaJCB good idea, done

PR #2420, opened 2018-08-17
Better Chain Arbitrary

LukaJCB:
This should fix the concerns @johnynek had in https://github.com/typelevel/cats/pull/2371

Also optimizes the `Chain.apply` builder :)


LukaJCB:
The small bug in `uncons` has been fixed, thanks a lot to @johnynek for pushing for the better arbitrary instance :)

LukaJCB:
Merging with two sign-offs :)

ceedubs:
Sorry I didn't get to the review @LukaJCB; I was on a plane. I'll submit some PRs that I crafted up while I was on it though :)

PR #2365, opened 2018-08-05
Add findM to Foldable

KentShikama:

PR #2421, opened 2018-08-18
Add/fix Foldable extensions: findM and collectFirstSomeM

catostrophe:
#2365 seems abandoned, so this is a rework of `findM` implemented as a syntax extension for binary compatibility. Another but equivalent implementation chosen, because `Foldable.Source` is private.

### DIFF
--- a/core/src/main/scala/cats/syntax/foldable.scala
+++ b/core/src/main/scala/cats/syntax/foldable.scala
@@ -95,31 +95,40 @@ final class FoldableOps[F[_], A](val fa: F[A]) extends AnyVal {
 
   /**
     * Monadic version of `collectFirstSome`.
+    *
+    * If there are no elements, the result is `None`. `collectFirstSomeM` short-circuits,
+    * i.e. once a Some element is found, no further effects are produced.
+    *
+    * For example:
     * {{{
     * scala> import cats.implicits._
     * scala> def parseInt(s: String): Either[String, Int] = Either.catchOnly[NumberFormatException](s.toInt).leftMap(_.getMessage)
     * scala> val keys1 = List("1", "2", "4", "5")
     * scala> val map1 = Map(4 -> "Four", 5 -> "Five")
     * scala> keys1.collectFirstSomeM(parseInt(_) map map1.get)
-    * res1: scala.util.Either[String,Option[String]] = Right(Some(Four))
+    * res0: scala.util.Either[String,Option[String]] = Right(Some(Four))
+    *
     * scala> val map2 = Map(6 -> "Six", 7 -> "Seven")
     * scala> keys1.collectFirstSomeM(parseInt(_) map map2.get)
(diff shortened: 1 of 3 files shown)


core/src/main/scala/cats/syntax/foldable.scala
```diff
@@ -120,4 +120,32 @@ final class FoldableOps[F[_], A](val fa: F[A]) extends AnyVal {
...
+    * res2: Either[String,Option[Int]] = Right(Some(2))
+    *
+    * scala> list.findM(n => Either.cond(n < 3, false, "error"))
+    * res3: Either[String,Option[Int]] = Left(error)
+    * }}}
+    */
+  def findM[G[_]](p: A => G[Boolean])(implicit F: Foldable[F], G: Monad[G]): G[Option[A]] =
+    F.foldRight(fa, Eval.now(G.pure(Option.empty[A])))((a, lb) =>
+      Eval.now(G.flatMap(p(a))(if (_) G.pure(Some(a)) else lb.value))
+    ).value
```
johnynek: I’m worried about stack safety here. Note findM, foldM and existsM use tailRecM to be stack safe.

Can we make Source package private so that we can use it here with tailRecM?

catostrophe: Good point. I assumed stack-safety is provided by safe flatMap implementation. Will rework via tailRecM.

catostrophe: Also added stack-safety tests for other monadic folds.

catostrophe:
Can anything else be improved here to give it the green light?

core/src/main/scala/cats/syntax/foldable.scala
```diff
@@ -95,29 +95,67 @@ final class FoldableOps[F[_], A](val fa: F[A]) extends AnyVal {
...
+    *
+    * scala> list.findM(n => Either.cond(n < 3, n >= 2, "error"))
+    * res2: Either[String,Option[Int]] = Right(Some(2))
+    *
+    * scala> list.findM(n => Either.cond(n < 3, false, "error"))
+    * res3: Either[String,Option[Int]] = Left(error)
+    * }}}
+    */
+  def findM[G[_]](p: A => G[Boolean])(implicit F: Foldable[F], G: Monad[G]): G[Option[A]] =
+    G.tailRecM(Foldable.Source.fromFoldable(fa))(_.uncons match {
```
johnynek: couldn't this be `collectFirstSome { a => p(a).map(if (_) Some(a) else None) }`?

catostrophe: Sure it could but would require an extra `map` for each iteration. I just followed the recommendation of other contributors to prefer efficiency over conciseness in code.

LukaJCB:
Shall we merge this, then? :)

KentShikama:
@catostrophe Thanks for cleaning up after me

PR #2423, opened 2018-08-19
Fix NestedIdOps.value leaking to types of shapes `F[G[A]]`

zarthross:
Made 'value'  private in `NestedIdOps`
This prevents 'value' from leaking into types with shape `F[G[A]]`.  

I encountered this issue while using Scalatests `OptionValues` on things with a `G[A]` shape, so I had a `Option[G[A]]` and tried to use `.value`.  It caused an ambiguous implicit match with `NestedIdOps`.

I didn't add a test, because I didn't know where it should go.  Would `SyntaxSuite.testNested` be an appropriate place?

### DIFF
--- a/core/src/main/scala/cats/syntax/nested.scala
+++ b/core/src/main/scala/cats/syntax/nested.scala
@@ -8,7 +8,7 @@ trait NestedSyntax {
     new NestedIdOps[F, G, A](value)
 }
 
-final class NestedIdOps[F[_], G[_], A](val value: F[G[A]]) extends AnyVal {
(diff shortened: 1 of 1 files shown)


zarthross:
@kailuowang Should I have changed that to 
`private[syntax] val value: F[G[A]]`
Instead?  

If I'm reading [this](https://github.com/jatcwang/binary-compatibility-guide#do-mark-methods-as-package-private-when-making-breaking-api-changes) right, it avoids binary incompatibility?

zarthross:
@kailuowang I've fixed the binary compatibility issue, mind removing the label?

PR #2425, opened 2018-08-19
Override size in Chain instance

ceedubs:
This should be a slight optimization.


core/src/main/scala/cats/data/Chain.scala
```diff
@@ -538,6 +538,7 @@ private[data] sealed abstract class ChainInstances extends ChainInstances1 {
       override def exists[A](fa: Chain[A])(p: A => Boolean): Boolean = fa.exists(p)
       override def forall[A](fa: Chain[A])(p: A => Boolean): Boolean = fa.forall(p)
       override def find[A](fa: Chain[A])(f: A => Boolean): Option[A] = fa.find(f)
+      override def size[A](fa: Chain[A]): Long = fa.length.toLong
```
LukaJCB: `fa.length` already returns `Long`, so no need for the call to `toLong` :)

ceedubs:
I'm pretty sure that it returned a long when I looked. Either i was looking
at outdated code or I'm seeing things :)

It might be a while before I'm at a computer to fix this.

PR #2426, opened 2018-08-19
Enrich return type for Representable.apply and add doctests

ceedubs:
The `Representable.apply` method was discarding information about its
return type in a way that made it not very helpful for some use-cases. I
believe that the change that I made to it makes it strictly more useful
in a compatible way, but let me know if that doesn't seem to be the
case.

While creating this I noticed that our Representable instance for Tuple2
uses true for the first index and false for the second index. I'm not
familiar with any precedents in this area, but I would have expected
false = 0 index and true = 1 index.  I assume that this is intended?

I also added several scaladoc examples, because why not?!

### DIFF
--- a/core/src/main/scala/cats/Representable.scala
+++ b/core/src/main/scala/cats/Representable.scala
@@ -21,11 +21,41 @@ trait Representable[F[_]] extends Serializable {
 
   /**
    * Create a function that "indexes" into the `F` structure using `Representation`
+   *
+   * Example:
+   * {{{
+   * scala> import cats.implicits._
+   *
+   * scala> type Pair[A] = (A, A)
+   *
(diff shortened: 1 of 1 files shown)


eli-jordan:
@ceedubs The change looks good to me. Thanks for adding the doc tests, probably should have done that originally :-/

Regarding the instance for `Tuple2` using `true` for the left element and `false` for the right element. This was an arbitrary decision - `Representable` just mandates there is a "slot" for every inhabitant of the type. I don't have a strong opinion on whether it should be `(true, false)` of `(false, true)`

ceedubs:
Thanks for the prompt response @eli-jordan. As you said, ultimately the slot assignments for `true`/`false` are arbitrary; and I think that changing them at this point would be unnecessarily breaking. Probably best to leave it as it is.

PR #2430, opened 2018-08-20
Make Chain Arbitraries recursively build concatenations

johnynek:
forgive me @LukaJCB I wanted a direct Arbitrary that constructs in all the ways we can (empty, single, concat or fromList).

I think this current generator should create any possible chain with some probability.

### DIFF
--- a/core/src/main/scala/cats/data/Chain.scala
+++ b/core/src/main/scala/cats/data/Chain.scala
@@ -421,10 +421,10 @@ object Chain extends ChainInstances {
     fromSeq(as)
 
   // scalastyle:off null
-  class ChainIterator[A](self: Chain[A]) extends Iterator[A] {
-    var c: Chain[A] = if (self.isEmpty) null else self
-    val rights = new collection.mutable.ArrayBuffer[Chain[A]]
-    var currentIterator: Iterator[A] = null
+  private class ChainIterator[A](self: Chain[A]) extends Iterator[A] {
+    private[this] var c: Chain[A] = if (self.isEmpty) null else self
+    private[this] val rights = new collection.mutable.ArrayBuffer[Chain[A]]
+    private[this] var currentIterator: Iterator[A] = null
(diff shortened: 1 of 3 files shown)


core/src/main/scala/cats/data/Chain.scala
```diff
@@ -37,12 +37,9 @@ sealed abstract class Chain[+A] {
...
           result = Some((seq.head, next))
         case Empty =>
-          if (rights.isEmpty) {
-            result = None
-          } else {
-            c = rights.last
-            rights.trimEnd(1)
-          }
+          // Empty is only top level, it is never internal to an Append
+          if (rights.nonEmpty) throw new IllegalStateException(s"found internal Empty in $this")
```
johnynek: should we just remove this since 1. we believe this to be true, 2. we have laws to catch this. By removing entirely we improve code coverage and we possibly slightly improve performance.

core/src/main/scala/cats/data/Chain.scala
```diff
@@ -461,8 +455,8 @@ object Chain extends ChainInstances {
               rights.clear()
               currentIterator = seq.iterator
               currentIterator.next
-            case Empty =>
-              go // This shouldn't happen
+            case null | Empty =>
```
johnynek: without the null here we had a match error previously on empty.

core/src/main/scala/cats/data/Chain.scala
```diff
@@ -505,8 +499,8 @@ object Chain extends ChainInstances {
               lefts.clear()
               currentIterator = seq.reverseIterator
               currentIterator.next
-            case Empty =>
-              go // This shouldn't happen
+            case null | Empty =>
```
johnynek: without the null here we had a match error previously on empty.

LukaJCB:
Ready to merge this @johnynek?

PR #2431, opened 2018-08-20
Give NonEmptyChain more presence

LukaJCB:
This adds the `ValidatedNec` type alias as well as several api enhancements that were previously only available to `NonEmptyList`

### DIFF
--- a/core/src/main/scala/cats/data/EitherT.scala
+++ b/core/src/main/scala/cats/data/EitherT.scala
@@ -205,6 +205,8 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
   def toValidatedNel(implicit F: Functor[F]): F[ValidatedNel[A, B]] =
     F.map(value)(_.toValidatedNel)
 
+  def toValidatedNec(implicit F: Functor[F]): F[ValidatedNec[A, B]] = ???
+
   /** Run this value as a `[[Validated]]` against the function and convert it back to an `[[EitherT]]`.
    *
    * The [[Applicative]] instance for `EitherT` "fails fast" - it is often useful to "momentarily" have
@@ -274,6 +276,12 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
    */
   def toNestedValidatedNel(implicit F: Functor[F]): Nested[F, ValidatedNel[A, ?], B] =
     Nested[F, ValidatedNel[A, ?], B](F.map(value)(_.toValidatedNel))
+
+  /**
+   * Transform this `EitherT[F, A, B]` into a `[[Nested]][F, ValidatedNec[A, ?], B]`.
+   */
+  def toNestedValidatedNec(implicit F: Functor[F]): Nested[F, ValidatedNec[A, ?], B] =
+    Nested[F, ValidatedNec[A, ?], B](F.map(value)(_.toValidatedNec))
 }
 
 object EitherT extends EitherTInstances {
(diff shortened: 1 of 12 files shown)


core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -205,6 +205,8 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
   def toValidatedNel(implicit F: Functor[F]): F[ValidatedNel[A, B]] =
     F.map(value)(_.toValidatedNel)
 
+  def toValidatedNec(implicit F: Functor[F]): F[ValidatedNec[A, B]] = ???
```
johnynek: implementation coming, I assume....

core/src/main/scala/cats/data/Validated.scala
```diff
@@ -602,3 +602,38 @@ private[data] trait ValidatedFunctions {
...
+  /**
+   * Converts a `B` to a `ValidatedNec[A, B]`.
+   *
+   * For example:
+   * {{{
+   * scala> Validated.validNec[IllegalArgumentException, String]("Hello world")
+   * res0: ValidatedNec[IllegalArgumentException, String] = Valid(Hello world)
+   * }}}
+   */
+  def validNec[A, B](b: B): ValidatedNec[A, B] = Validated.Valid(b)
```
johnynek: validated is already covariant. What about `ValidatedNec[Nothing, B]` here?

core/src/main/scala/cats/data/Validated.scala
```diff
@@ -602,3 +602,38 @@ private[data] trait ValidatedFunctions {
...
+  /**
+   * Converts an `A` to a `ValidatedNec[A, B]`.
+   *
+   * For example:
+   * {{{
+   * scala> Validated.invalidNec[IllegalArgumentException, String](new IllegalArgumentException("Argument is nonzero"))
+   * res0: ValidatedNec[IllegalArgumentException, String] = Invalid(Chain(java.lang.IllegalArgumentException: Argument is nonzero))
+   * }}}
+   */
+  def invalidNec[A, B](a: A): ValidatedNec[A, B] = Validated.Invalid(NonEmptyChain.one(a))
```
johnynek: same variance question, `invalidNec[A](a: A): ValidatedNec[A, Nothing]`?

core/src/main/scala/cats/syntax/either.scala
```diff
@@ -374,6 +375,49 @@ final class EitherIdOps[A](val obj: A) extends AnyVal {
...
+   * res0: Either[NonEmptyChain[String], Int] = Right(1)
+   * }}}
+   */
+  def rightNec[B]: Either[NonEmptyChain[B], A] = Right(value)
+}
+
+final class EitherOpsBinCompat0[A, B](val value: Either[A, B]) extends AnyVal {
+  /** Returns a [[cats.data.ValidatedNec]] representation of this disjunction with the `Left` value
+   * as a single element on the `Invalid` side of the [[cats.data.NonEmptyList]]. */
+  def toValidatedNec[AA >: A]: ValidatedNec[AA, B] = value match {
```
johnynek: since Validated is covariant, do we need this type parameted?

LukaJCB:
These are good questions, I just adopted the type signatures from the `NonEmptyList` counterparts. I'm not sure why those were chosen the way they were chosen in the first place, maybe someone can chime in :)

core/src/main/scala/cats/data/Validated.scala
johnynek: So, why do we need it? In cases where type inference falls down? e.g. in folds the initial value?

core/src/main/scala/cats/data/EitherT.scala
johnynek: this is still a `???`

We should consider a CI check that we can't merge with that present.

johnynek: Also, can we add a test to exercise this, since we must not have it now.

ceedubs: @johnynek I think that you are looking at outdated code. For me there's no longer `???` in the PR.

PR #2437, opened 2018-08-22
Add FP for Mortals entry to Microsite

emilypi:
Hi Everyone, 

As you know, Sam has finally finished his book, and a fair bit of it deals with Cats, as well as Scalaz. As the most modern resource on the subject of FP in Scala (chronologically), I feel it has a place here, considering it _is_ free and very well-written. Thanks.

### DIFF
--- a/docs/src/main/tut/resources_for_learners.md
+++ b/docs/src/main/tut/resources_for_learners.md
@@ -15,6 +15,12 @@ position: 30
    to introduce monads, functors, and other functional programming patterns
    as a way to structure program design, and to explain how these 
    concepts are implemented in Cats.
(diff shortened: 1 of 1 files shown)


emilypi:
@LukaJCB can i get a hallelujah?

stew:
Sam has been openly hostile the typelevel project, to the point of inventing and spreading lies about it and its members. 

I also don't think that  the book has an accurate description of cats.

fommil:
I actually didn't feel cats, or typelevel, were relevant enough to mention.

emilypi:
@stew which part wasn't accurate? The parent part? I can change that.

Issue #2440, opened 2018-08-23
Add missing semi-/sub-/-F/-Tap versions of flatMap to monad transformers


catostrophe:
Monad transformers lack some `semi-`/`sub-`/`-F`/`-Tap` versions of `flatMap` that free a user from manual lifting or mapping.

I find the following items reasonable for implementation as to make structures and transformers more consistent:
- OptionT:
  - [x] add member methods `flatTap`, `flatTapF`, `semiflatTap`, `subflatTap`
- Either:
  - [x] add extensions `flatTap`, `leftFlatTap` to `EitherOps` syntax class
- EitherT:
  - [x] add member methods `leftFlatMapF`, `leftSubflatMap`
  - [x] add member methods `flatTap`, `flatTapF`, `semiflatTap`, `subflatTap`
  - [x] add member methods `leftFlatTap`, `leftFlatTapF`, `leftSemiflatTap`, `leftSubflatTap`
  - [x] add member methods `biSemiflatTap`
- Ior:
  - [x] add member methods `leftFlatMap`
  - [x] add member methods `flatTap`, `leftFlatTap`
- IorT:
  - [x] add member methods `leftFlatMapF`, `leftSubflatMap`
  - [x] add member methods `flatTap`, `flatTapF`, `semiflatTap`, `subflatTap`
  - [x] add member methods `leftFlatTap`, `leftFlatTapF`, `leftSemiflatTap`, `leftSubflatTap`
- IdT:
  - [x] add member methods `flatTap`, `flatTapF`
- WriterT:
  - [x] add member methods `flatMapF`, `semiflatMap`, `subflatMap`
  - [x] add member methods `flatTap`, `flatTapF`, `semiflatTap`, `subflatTap`
- Kleisli:
  - [x] add member methods `flatTap`, `flatTapF`
- IndexedStateT:
  - [x] add member methods `flatTap`, `flatTapF`
- IndexedReaderWriterStateT:
  - [x] add member methods `flatTap`, `flatTapF`

Open questions:
- [x] `flatTap` automatically comes from `FlatMap` syntax. For some transformers it has a special implementation (e.g. for EitherT), for others it is standard (e.g. for OptionT). Should we add it to the latter for symmetry and consistency? (**Added**)
- [ ] EitherT has `biSemiflatMap`. Do `Either`/`EitherT`/`Ior`/`IorT` need `biFlatMap` with all the `semi-`/`sub-`/`-F`/`-Tap` versions?

[...]

PR #2442, opened 2018-08-26
Add left-/semi-/sub-/-F/-Tap flatMaps to monad transformers

catostrophe:
This is to address #2440.

### DIFF
--- a/core/src/main/scala/cats/data/EitherT.scala
+++ b/core/src/main/scala/cats/data/EitherT.scala
@@ -538,7 +538,7 @@ private[data] abstract class EitherTInstances extends EitherTInstances1 {
     }
 
   implicit def catsDataTraverseForEitherT[F[_], L](implicit FF: Traverse[F]): Traverse[EitherT[F, L, ?]] =
-    new EitherTTraverse[F, L] with EitherTFunctor[F, L] {
+    new EitherTTraverse[F, L] with EitherTFunctor[F, L] with EitherTFunctorExtBinCompat0[F, L] {
       val F0: Traverse[F] = FF
       val F: Functor[F] = FF
     }
@@ -574,7 +574,7 @@ private[data] abstract class EitherTInstances1 extends EitherTInstances2 {
     }
 
   implicit def catsDataMonadErrorForEitherT[F[_], L](implicit F0: Monad[F]): MonadError[EitherT[F, L, ?], L] =
-    new EitherTMonadError[F, L] {
+    new EitherTMonadError[F, L] with EitherTFunctorExtBinCompat0[F, L] {
       implicit val F = F0
       override def ensure[A](fa: EitherT[F, L, A])(error: => L)(predicate: (A) => Boolean): EitherT[F, L, A] =
         fa.ensure(error)(predicate)(F)
@@ -600,7 +600,7 @@ private[data] abstract class EitherTInstances2 extends EitherTInstances3 {
    * }}}
    */
   implicit def catsDataMonadErrorFForEitherT[F[_], E, L](implicit FE0: MonadError[F, E]): MonadError[EitherT[F, L, ?], E] =
-    new EitherTMonadErrorF[F, E, L] { implicit val F = FE0 }
+    new EitherTMonadErrorF[F, E, L] with EitherTFunctorExtBinCompat0[F, L] { implicit val F = FE0 }
 
 
   implicit def catsDataSemigroupKForEitherT[F[_], L](implicit F0: Monad[F]): SemigroupK[EitherT[F, L, ?]] =
@@ -614,7 +614,7 @@ private[data] abstract class EitherTInstances2 extends EitherTInstances3 {
 
 private[data] abstract class EitherTInstances3 {
   implicit def catsDataFunctorForEitherT[F[_], L](implicit F0: Functor[F]): Functor[EitherT[F, L, ?]] =
-    new EitherTFunctor[F, L] { implicit val F = F0 }
+    new EitherTFunctor[F, L] with EitherTFunctorExtBinCompat0[F, L] { implicit val F = F0 }
 }
 
 private[data] trait EitherTSemigroup[F[_], L, A] extends Semigroup[EitherT[F, L, A]] {
@@ -640,6 +640,9 @@ private[data] trait EitherTSemigroupK[F[_], L] extends SemigroupK[EitherT[F, L, 
 private[data] trait EitherTFunctor[F[_], L] extends Functor[EitherT[F, L, ?]] {
   implicit val F: Functor[F]
   override def map[A, B](fa: EitherT[F, L, A])(f: A => B): EitherT[F, L, B] = fa map f
+}
+
+private[data] trait EitherTFunctorExtBinCompat0[F[_], L] { self: EitherTFunctor[F, L] =>
   override def as[A, B](fa: EitherT[F, L, A], b: B): EitherT[F, L, B] = fa as b
 }
 
--- a/core/src/main/scala/cats/data/EitherT.scala
+++ b/core/src/main/scala/cats/data/EitherT.scala
@@ -100,18 +100,20 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
     EitherT(F.map(value)(f))
 
   def flatTap[AA >: A, D](f: B => EitherT[F, AA, D])(implicit F: Monad[F]): EitherT[F, AA, B] =
-    flatMap(b => f(b).map(_ => b))
+    flatMap(b => f(b).as(b))
 
   def flatTapF[AA >: A, D](f: B => F[Either[AA, D]])(implicit F: Monad[F]): EitherT[F, AA, B] =
     flatTap(f andThen EitherT.apply)
 
   def subflatTap[AA >: A, D](f: B => Either[AA, D])(implicit F: Functor[F]): EitherT[F, AA, B] =
-    subflatMap(b => f(b).map(_ => b))
+    subflatMap(b => f(b).as(b))
 
   def semiflatTap[D](f: B => F[D])(implicit F: Monad[F]): EitherT[F, A, B] =
     semiflatMap(b => F.as(f(b), b))
 
   def map[D](f: B => D)(implicit F: Functor[F]): EitherT[F, A, D] = bimap(identity, f)
+
+  def as[D](d: D)(implicit F: Functor[F]): EitherT[F, A, D] = EitherT(F.map(value)(_.as(d)))
 
   /**
    * Modify the context `F` using transformation `f`.
@@ -119,6 +121,8 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
   def mapK[G[_]](f: F ~> G): EitherT[G, A, B] = EitherT[G, A, B](f(value))
 
   def leftMap[C](f: A => C)(implicit F: Functor[F]): EitherT[F, C, B] = bimap(f, identity)
+
+  def leftAs[C](c: C)(implicit F: Functor[F]): EitherT[F, C, B] = EitherT(F.map(value)(_.leftAs(c)))
 
   def leftFlatMap[BB >: B, C](f: A => EitherT[F, C, BB])(implicit F: Monad[F]): EitherT[F, C, BB] =
     EitherT(F.flatMap(value) {
@@ -139,13 +143,13 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
     })
 
   def leftFlatTap[BB >: B, C](f: A => EitherT[F, C, BB])(implicit F: Monad[F]): EitherT[F, A, BB] =
-    leftFlatMap(a => f(a).leftMap(_ => a))
+    leftFlatMap(a => f(a).leftAs(a))
 
   def leftFlatTapF[BB >: B, C](f: A => F[Either[C, BB]])(implicit F: Monad[F]): EitherT[F, A, BB] =
     leftFlatTap(f andThen EitherT.apply)
 
   def leftSubflatTap[BB >: B, C](f: A => Either[C, BB])(implicit F: Functor[F]): EitherT[F, A, BB] =
-    leftSubflatMap(a => f(a).leftMap(_ => a))
+    leftSubflatMap(a => f(a).leftAs(a))
 
   def leftSemiflatTap[C](f: A => F[C])(implicit F: Monad[F]): EitherT[F, A, B] =
     leftSemiflatMap(a => F.as(f(a), a))
@@ -636,6 +640,7 @@ private[data] trait EitherTSemigroupK[F[_], L] extends SemigroupK[EitherT[F, L, 
 private[data] trait EitherTFunctor[F[_], L] extends Functor[EitherT[F, L, ?]] {
   implicit val F: Functor[F]
   override def map[A, B](fa: EitherT[F, L, A])(f: A => B): EitherT[F, L, B] = fa map f
+  override def as[A, B](fa: EitherT[F, L, A], b: B): EitherT[F, L, B] = fa as b
 }
 
 private[data] trait EitherTMonad[F[_], L] extends Monad[EitherT[F, L, ?]] with EitherTFunctor[F, L] {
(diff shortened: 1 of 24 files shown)


core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -90,11 +90,26 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
...
+
+  def semiflatMap[D](f: B => F[D])(implicit F: Monad[F]): EitherT[F, A, D] =
+    flatMap(b => EitherT.right(f(b)))
+
   def transform[C, D](f: Either[A, B] => Either[C, D])(implicit F: Functor[F]): EitherT[F, C, D] =
     EitherT(F.map(value)(f))
 
-  def subflatMap[AA >: A, D](f: B => Either[AA, D])(implicit F: Functor[F]): EitherT[F, AA, D] =
-    transform(_.flatMap(f))
+  def flatTap[AA >: A, D](f: B => EitherT[F, AA, D])(implicit F: Monad[F]): EitherT[F, AA, B] =
```
LukaJCB: I think this one is already provided by `Monad` :)

core/src/main/scala/cats/data/IdT.scala
```diff
@@ -42,6 +42,12 @@ final case class IdT[F[_], A](value: F[A]) {
   def ap[B](f: IdT[F, A => B])(implicit F: Apply[F]): IdT[F, B] =
     IdT(F.ap(f.value)(value))
 
+  def flatTap[B](f: A => IdT[F, B])(implicit F: FlatMap[F]): IdT[F, A] =
```
LukaJCB: Same here :)

core/src/main/scala/cats/data/IndexedReaderWriterStateT.scala
```diff
@@ -109,6 +109,13 @@ final class IndexedReaderWriterStateT[F[_], E, L, SA, SB, A](val runF: F[(E, SA)
       }
     }
 
+  def flatTap[SC, B](f: A => IndexedReaderWriterStateT[F, E, L, SB, SC, B])(
```
LukaJCB: This one as well.

core/src/main/scala/cats/data/IndexedStateT.scala
```diff
@@ -167,6 +167,13 @@ final class IndexedStateT[F[_], SA, SB, A](val runF: F[SA => F[(SB, A)]]) extend
    */
   def get(implicit F: Functor[F]): IndexedStateT[F, SA, SB, SB] =
     inspect(identity)
+
+  def flatTap[B, SC](fas: A => IndexedStateT[F, SB, SC, B])(implicit F: FlatMap[F]): IndexedStateT[F, SA, SC, A] =
```
LukaJCB: Here too.

core/src/main/scala/cats/data/Ior.scala
```diff
@@ -81,6 +81,22 @@ sealed abstract class Ior[+A, +B] extends Product with Serializable {
...
+    case Ior.Left(a) => f(a)
+    case r @ Ior.Right(_) => r
+    case Ior.Both(a, b1) => f(a) match {
+      case Ior.Left(c) => Ior.Both(c, b1)
+      case Ior.Right(b2) => Ior.Right(BB.combine(b1, b2))
+      case Ior.Both(c, b2) => Ior.Both(c, BB.combine(b1, b2))
+    }
+  }
+
+  final def flatTap[AA >: A, D](f: B => AA Ior D)(implicit AA: Semigroup[AA]): AA Ior B =
```
LukaJCB: Here too.

core/src/main/scala/cats/data/IorT.scala
```diff
@@ -120,6 +126,31 @@ final case class IorT[F[_], A, B](value: F[Ior[A, B]]) {
 
   def combine(that: IorT[F, A, B])(implicit F: Apply[F], A: Semigroup[A], B: Semigroup[B]): IorT[F, A, B] =
     IorT(F.map2(this.value, that.value)(_ combine _))
+
+  def flatTap[AA >: A, D](f: B => IorT[F, AA, D])(implicit F: Monad[F], AA: Semigroup[AA]): IorT[F, AA, B] =
```
LukaJCB: Here too.

core/src/main/scala/cats/data/Kleisli.scala
```diff
@@ -33,6 +33,12 @@ final case class Kleisli[F[_], A, B](run: A => F[B]) { self =>
   def flatMapF[C](f: B => F[C])(implicit F: FlatMap[F]): Kleisli[F, A, C] =
     Kleisli.shift(a => F.flatMap(run(a))(f))
 
+  def flatTap[C](f: B => Kleisli[F, A, C])(implicit F: FlatMap[F]): Kleisli[F, A, B] =
```
LukaJCB: Here too.

core/src/main/scala/cats/data/OptionT.scala
```diff
@@ -54,6 +54,18 @@ final case class OptionT[F[_], A](value: F[Option[A]]) {
   def subflatMap[B](f: A => Option[B])(implicit F: Functor[F]): OptionT[F, B] =
     transform(_.flatMap(f))
 
+  def flatTap[B](f: A => OptionT[F, B])(implicit F: Monad[F]): OptionT[F, A] =
```
LukaJCB: Here too.

core/src/main/scala/cats/data/WriterT.scala
```diff
@@ -47,6 +47,32 @@ final case class WriterT[F[_], L, V](run: F[(L, V)]) {
...
+      val (l2, u) = f(v)
+      (semigroupL.combine(l1, l2), u)
+    })
+
+  def semiflatMap[U](f: V => F[U])(implicit flatMapF: FlatMap[F]): WriterT[F, L, U] =
+    WriterT(flatMapF.flatMap(run) { case (l, v) =>
+      flatMapF.map(f(v))((l, _))
+    })
+
+  def flatTap[U](f: V => WriterT[F, L, U])(implicit flatMapF: FlatMap[F], semigroupL: Semigroup[L]): WriterT[F, L, V] =
```
LukaJCB: Here too.

core/src/main/scala/cats/syntax/either.scala
```diff
@@ -153,6 +153,12 @@ final class EitherOps[A, B](val eab: Either[A, B]) extends AnyVal {
     case r @ Right(_) => EitherUtil.leftCast(r)
   }
 
+  def flatTap[AA >: A, D](f: B => Either[AA, D]): Either[AA, B] =
```
LukaJCB: Here too.

core/src/main/scala/cats/data/EitherT.scala
catostrophe: Please look at the open questions in issue #2440. `flatTap` is provided by Monad but for EitherT it doesn't support widening of the second type parameter.

catostrophe: This version allows widening of A. `Monad`'s one doesn't.

core/src/main/scala/cats/data/IndexedReaderWriterStateT.scala
catostrophe: This version allows changing the state type. `Monad` instance exists for SA = SB only.

core/src/main/scala/cats/data/IndexedStateT.scala
catostrophe: This version allows changing the state type. `Monad` instance exists for SA = SB only.

core/src/main/scala/cats/data/Ior.scala
catostrophe: This version allows widening of A. `Monad`'s one doesn't.

core/src/main/scala/cats/data/IorT.scala
catostrophe: This version allows widening of A. `Monad`'s one doesn't.

catostrophe:
@LukaJCB just to confirm. Should we keep only those `flatTap`s adding some value (see comments) or should we keep them all for symmetry?

core/src/main/scala/cats/syntax/either.scala
catostrophe: This version allows widening of A. `Monad`'s one doesn't.

core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -90,11 +90,26 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
...
+    flatMap(b => f(b).map(_ => b))
+
+  def flatTapF[AA >: A, D](f: B => F[Either[AA, D]])(implicit F: Monad[F]): EitherT[F, AA, B] =
+    flatTap(f andThen EitherT.apply)
+
+  def subflatTap[AA >: A, D](f: B => Either[AA, D])(implicit F: Functor[F]): EitherT[F, AA, B] =
+    subflatMap(b => f(b).map(_ => b))
```
johnynek: `.as(b)` can be optimized and by law should be the same as `.map(_ => b)`.

For instance, as for `EitherMonad` can be a cast of the type if we have a `Left` or allocate `Right(b)`, avoiding allocating the lambda and calling a function (which is hard to inline for the JIT due to Function1 being megamorphic).

core/src/main/scala/cats/data/IorT.scala
```diff
@@ -120,6 +126,31 @@ final case class IorT[F[_], A, B](value: F[Ior[A, B]]) {
...
+    flatMap(b => f(b).map(_ => b))
```
johnynek: `.as` here too, right?

core/src/main/scala/cats/data/EitherT.scala
catostrophe: @johnynek There are many `.map(_ => a)` here and there in my change. To do the optimization you suggest I need to implement `as` method for each class changed, override `as` in their `Functor` instances, express `flatTap`/`subflatTap` via `as` instead of `map`. For `Either`/`Ior` `leftAs` should be added for the same purpose. Right?

catostrophe:
@johnynek `MonadTransformer` typeclass would be really useful but arrow-like transformers such as ReaderT or StateT don't fit.

core/src/main/scala/cats/data/EitherT.scala
kailuowang: I would say for this PR express `flatTap/subflatTap` via `as` instead of `map` would suffice. We can create a separate PR or maybe just an issue for optimizing `as`.

catostrophe: Ok, I will rework this PR.

catostrophe:
I stuck a bit.
Adding the optimized versions of `as` to the `Functor`-implementing traits led to bincompat issues. This could be fixed by moving `as` to a separate trait and mixing it directly in the instances extending `Functor`. But there are lots of them for each transformer. This workaround looks ugly for me.

Any other ideas on fixing bincompat issues here?

kailuowang:
You can't add any method to any trait without breaking BC on 2.11. 
The alternative is to create a new trait with the methods you want to add and have you class extending that. Or add the methods directly to the class.

catostrophe:
If the whole `as` optimization is worth it, `as` method should be overridden in all instances extending `Functor`. Due to BC reasons, it can be done only in a separate trait. E.g. for EitherT it could be done as follows:
```scala
private[data] trait EitherTFunctorExtBinCompat0[F[_], L] { self: EitherTFunctor[F, L] =>
  override def as[A, B](fa: EitherT[F, L, A], b: B): EitherT[F, L, B] = fa as b
}
```

This trait shoud be mixed in every instance extending `Functor`, e.g.:
```scala
implicit def catsDataTraverseForEitherT[F[_], L](implicit FF: Traverse[F]): Traverse[EitherT[F, L, ?]] =
  new EitherTTraverse[F, L] with EitherTFunctor[F, L] with EitherTFunctorExtBinCompat0[F, L] { ... }

implicit def catsDataMonadErrorForEitherT[F[_], L](implicit F0: Monad[F]): MonadError[EitherT[F, L, ?], L] =
  new EitherTMonadError[F, L] with EitherTFunctorExtBinCompat0[F, L] { ... }

implicit def catsDataMonadErrorFForEitherT[F[_], E, L](implicit FE0: MonadError[F, E]): MonadError[EitherT[F, L, ?], E] =
  new EitherTMonadErrorF[F, E, L] with EitherTFunctorExtBinCompat0[F, L] { ... }

implicit def catsDataFunctorForEitherT[F[_], L](implicit F0: Functor[F]): Functor[EitherT[F, L, ?]] =
  new EitherTFunctor[F, L] with EitherTFunctorExtBinCompat0[F, L] { ... }
```

There are lots of such instances on different priority levels: e.g. there are 12 for WriterT. Is it a reasonable change?

catostrophe:
Btw do we really need to keep BC for private traits that cannot leak out of cats code?

kailuowang:
@catostrophe, sorry to drag you into this messy BC business. I suggested that we don't do the `as` optimization in this PR. merely, we just use `as` in the newly added methods in lieu of `map(_ => x)`. We can optimize `as` in a later effort possibly at Cats 2.0 with which we drop 2.11 and with scala 2.12 we can add methods to traits without breaking BC.

catostrophe:
Ok, I suggest choosing from 2 options:

1. reverting `as` optimization only in `Functor` and related instances where it breaks BC. We'll still have the optimized `as`-es (and `flatTap`s) in transformer data classes themselves. But those `as`-es invoked via typeclass instances will be non-optimized, while semantically identical to the optimized ones due to laws confirmed by tests.

2. adding optimized `as`-es to `Functor` and related instances via BC mixins such as in the comment above. This will be more consistent but will make the codebase more fragmented and harder to maintain.

catostrophe:
I pushed option 2. If it's not ok, I will revert to option 1.

kailuowang:
@catostrophe I vote for 1.

core/src/main/scala/cats/data/EitherT.scala
kailuowang: Will this hide the `flatTap` syntax provided by `Monad`? The type parameters are different. It might break source compatibility, I haven't tested it though so I could be wrong.

catostrophe: Thanks, will check this

catostrophe: Tried to compose some examples. Can't imagine what can break source compatibility here. Any ideas on what to check? Do you mean type inference might break?

Hiding may cause `unused import` but it's irrelevant to widening.

kailuowang: The `flatTap` syntax provided by `Monad` take one type parameters, while this one takes two. Both could be inferred, in which case they are source compatible. But there might be use sites where the type parameter of `flatTap` is provided, with this new version, it would no longer compile, right?

catostrophe: Yes, you're right. Does source breaking require a major version upgrade?

kailuowang: sorry I dropped the ball on this. It's a rather minor source breaking change, I wouldn't say we have to bump the major version. I am okay to have this merge as is, but I'd like to hear from other maintainers. @LukaJCB?

catostrophe: @LukaJCB @kailuowang Ok, source compatibility is important. But instead of removing I would rather keep `flatTap` in data types with the same signature as in `FlatMap.Ops` (`def flatTap[B](f: A => F[B]): F[A]`):
- pros: it can use the optimized `as`, while the syntax extension cannot, since we didn't override `as` in `Functor` instances due to bincompat issues
- cons: its signature will be different from other similar functions in the `flat` family for some data types (no widening of the left type param). and this probably needs a TODO comment

blast-hardcheese:
Any updates on this? Just ran into this.

valdo404:
Any news on that issue ? :) It seems everything was written and ready to be merged.

larsrh:
There's a ton of conflicts in this PR. If anyone wants to pick it up again, feel free to open a new one.

PR #2444, opened 2018-08-28
Add Various Projects to Ecosystem List

ChristopherDavenport:
The syntax of capitalization versus lowercase was not exceptionally clear. Would Like some guidance on proper formatting.

Fixes https://github.com/ChristopherDavenport/cats-time/issues/11

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,10 +76,14 @@ By sharing the same set of type classes, instances and data types provided by Ca
 
 #### General purpose libraries to support pure functional programming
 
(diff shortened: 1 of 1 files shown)


LukaJCB:
To @johnynek's point, they don't add any dependencies, but it's a bit complicated due to them not being available for Scala.js AFAIK

ChristopherDavenport:
Yes, I've been assured the scalajs support lib is stable by the author, despite being a milestone. That being said I imagine cats would have to more concerns with the status of that dependency than a micro-library.

PR #2448, opened 2018-08-31
Fixes #2434 Rename NonEmptySet to NonEmptySortedSet

### LINKED ISSUE #2434: Rename NonEmptySet to NonEmptySortedSet
Current implementation of  the `cats.data.NonEmptySet` confuses, because it's not `NonEmptySet`, it's `NonEmptySortedSet`. 

When it was introduced in cats 1.1.0 I was very excited about it, because there is a lot of use cases in my codebase for the `NonEmptySet`. But when I tried to use it I noticed that it uses `Order` typeclass and in general this new `NonEmptySet` was based on `SortedSet`, not simple `Set`. I understand, that there is a lot of use cases for `NonEmptySortedSet`, but `NonEmptySortedSet` is not `NonEmptySet`. For a lot of use cases, I don't need any ordering in my set, I just want to have a set with 1 or more elements. 

Сurrent naming is misleading, especially for newcomers. I see no reason why `NonEmptySortedSet` should be named as `NonEmptySet`. So, I suggest to rename `NonEmptySet` to `NonEmptySortedSet`.

Because this change will break compatibility, I think the best way for renaming should look like this:
1. Rename `NonEmptySortedSet` to `NonEmptySet`.
2. Create an alias from `NonEmptySet` to `NonEmptySortedSet`.
3. Mark this alias as deprecated.
4. Remove this alias in the next major release.

LMnet:
Fixes #2434

I renamed `NonEmptySet` to the `NonEmptySortedSet`. To keep backward compatibility I created an alias from `NonEmptySet` to `NonEmptySortedSet`. This alias marked as deprecated.

To save binary compatibility I had to preserve names for:
  * `NonEmptySetImpl`
  * `NonEmptySetImpl.catsNonEmptySetOps`
  * `NonEmptySetOps`
  * `NonEmptySetInstances`
  * `NonEmptySetInstances.catsDataInstancesForNonEmptySet`
  * `NonEmptySetInstances.catsDataEqForNonEmptySet`
  * `NonEmptySetInstances.catsDataShowForNonEmptySet`
  * `NonEmptySetInstances.catsDataSemilatticeForNonEmptySet`

I'm not happy with this, but I don't know another way to save binary compatibility.

### DIFF
--- a/core/src/main/scala/cats/data/NonEmptyMapImpl.scala
+++ b/core/src/main/scala/cats/data/NonEmptyMapImpl.scala
@@ -97,7 +97,7 @@ sealed class NonEmptyMapOps[K, A](val value: NonEmptyMap[K, A]) {
   /**
     * Returns a `SortedSet` containing all the keys of this map.
     */
-  def keys: NonEmptySet[K] = NonEmptySet.fromSetUnsafe(toSortedMap.keySet)
+  def keys: NonEmptySortedSet[K] = NonEmptySortedSet.fromSetUnsafe(toSortedMap.keySet)
 
   /**
     * Returns the first key-value pair of this map.
@@ -238,7 +238,7 @@ sealed class NonEmptyMapOps[K, A](val value: NonEmptyMap[K, A]) {
     * Typesafe equality operator.
     *
     * This method is similar to == except that it only allows two
-    * NonEmptySet[A] values to be compared to each other, and uses
+    * NonEmptySortedSet[A] values to be compared to each other, and uses
     * equality provided by Eq[_] instances, rather than using the
     * universal equality provided by .equals.
     */
--- a/core/src/main/scala/cats/data/NonEmptySortedSet.scala
+++ b/core/src/main/scala/cats/data/NonEmptySortedSet.scala
@@ -32,26 +32,26 @@ private[data] object NonEmptySetImpl extends NonEmptySetInstances with Newtype {
     s.asInstanceOf[SortedSet[A]]
 
 
-  def fromSet[A](as: SortedSet[A]): Option[NonEmptySet[A]] =
+  def fromSet[A](as: SortedSet[A]): Option[NonEmptySortedSet[A]] =
     if (as.nonEmpty) Option(create(as)) else None
 
-  def fromSetUnsafe[A](set: SortedSet[A]): NonEmptySet[A] =
+  def fromSetUnsafe[A](set: SortedSet[A]): NonEmptySortedSet[A] =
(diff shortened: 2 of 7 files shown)


jozic:
there's also NonEmptyMap

LMnet:
@jozic in this issue/pull request I aimed only on `NonEmptySet` renaming. I think renaming `NonEmptyMap` should be discussed separately.

jozic:
it can be (and probably should be) a separate PR, but imo NonEmptyMap should also be renamed if this PR is merged, at least for consistency

LukaJCB:
I'm fairly neutral w.r.t this change, not sure this is worth giving current users deprecation warnings, what do others think?

kailuowang:
I am -1 on deprecation.

LMnet:
@kailuowang could you elaborate on this? Why do you think the current name is better?

kailuowang:
@LMnet sorry, my comment was made in a hurry. I am not suggesting that the current name is better or worse. I don't have a strong view over that. On the other hand, we want to avoid any breaking change before 2.0. In some sense, deprecations can be deemed a sort of breaking change, since it would probably force projects to modify their code to avoid deprecation warnings.

LMnet:
@kailuowang in my opinion deprecation -> changing cycle is always better than just changing in the major version. You still can use functionality, but deprecation mark gives you an understanding, that at some point in the future you will need to rewrite your code. I think it's more gradual than a simple change.

kailuowang:
@LMnet I agree with you that deprecation is more preferable than changing the major version. In this particular case, I think the question here is whether to deprecate the old name or leave it alone.  I understand that the original name might have some issues, but I'd propose that we do not burden our users with any breaking changes for it.

Leammas:
I think more verbose name would be better. Because it makes you think more of `Order` of your elements and that NonEmptySet actually use it to calculate if values are equal.

easel:
Is there an intention to implement a `NonEmptyUnorderedSet`? If not, I'm -1 on deprecating the old name at this time as well. I do think the new name is better and more clear, however. If you had asked me prior I would have assumed NonEmptySet was unordered.

v-gerasimov:
`NonEmptySet` also makes me think that it based on universal equality but it does not. So I think `NonEmptySortedSet` would be better.

armanbilge:
Probably, it's a bit late to change this now, esp. since the current policy is to be conservative with deprecations within Cats. The good news is we will soon have a `NonEmptyHashMap` and `NonEmptyHashSet` :)

PR #2450, opened 2018-08-31
Add Kleisli#mapK equivalent returning FunctionK

kubukoz:
Related: https://github.com/typelevel/cats-effect/pull/334

I don't like the name either, so feel free to suggest a better one 😅

### DIFF
--- a/core/src/main/scala/cats/data/Kleisli.scala
+++ b/core/src/main/scala/cats/data/Kleisli.scala
@@ -125,12 +125,12 @@ sealed private[data] trait KleisliFunctions {
    * {{{
    * scala> import cats._, data._, implicits._
    * scala> val a: OptionT[Eval, Int] = 1.pure[OptionT[Eval, ?]]
-   * scala> val b: OptionT[Kleisli[Eval, String, ?], Int] = a.mapK(Kleisli.liftK)
+   * scala> val b: OptionT[Kleisli[Eval, String, ?], Int] = a.mapK(Kleisli.liftFunctionK)
    * scala> b.value.run("").value
    * res0: Option[Int] = Some(1)
    * }}}
    */
-  def liftK[F[_], A]: F ~> Kleisli[F, A, ?] =
+  def liftFunctionK[F[_], A]: F ~> Kleisli[F, A, ?] =
     λ[F ~> Kleisli[F, A, ?]](Kleisli.liftF(_))
 
   @deprecated("Use liftF instead", "1.0.0-RC2")
--- a/core/src/main/scala/cats/data/Kleisli.scala
+++ b/core/src/main/scala/cats/data/Kleisli.scala
@@ -83,7 +83,12 @@ final case class Kleisli[F[_], A, B](run: A => F[B]) { self =>
   def apply(a: A): F[B] = run(a)
 }
 
-object Kleisli extends KleisliInstances with KleisliFunctions with KleisliFunctionsBinCompat with KleisliExplicitInstances {
+object Kleisli
+    extends KleisliInstances
+    with KleisliFunctions
+    with KleisliFunctionsBinCompat
+    with KleisliExplicitInstances {
+
   /**
    * Internal API — shifts the execution of `run` in the `F` context.
    *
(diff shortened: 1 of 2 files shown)


kubukoz:
Sure, I wanted to get some initial feedback before taking the time to write (a) test(s) :) as mentioned in the issue, I'd gladly rename it too, but `mapK` is taken (even though it's somewhere else, it clearly has different types)

kailuowang:
This `mapK` type of operation keeps coming back, maybe we should reconsider this PR #1815. The `liftNT` (or we could name it `liftNaturalTransformation` ) provides this `mapKK` feature for all those data types through type class.

kubukoz:
`cats-tagless`'s FunctorK would be a generalization of `mapK`, but it has
`def mapK[F[_], G[_]](af: A[F])(fk: F ~> G): A[G]`
and doesn't provide anything of the shape `A[F] ~> A[G]`.

I would start by defining this for Kleisli for now, and if we see this used by people enough to make an abstraction, maybe then we'll add a typeclass? (`FunctorK`)

kailuowang:
I am ok with adding it to Kleisli first under the name liftK or liftNT

kubukoz:
There's already a `liftK`. `liftNT` is free, but I'm afraid the `lift` prefix doesn't match what that function would do. After all, this operates on an already existing Kleisli, and the `lift`ing functions create Kleislis out of values like `F[B]` or `B`. Here's a potentially stupid idea: `fmapK`, just to disambiguate it from `mapK`?

kailuowang:
The `lift` method in `Functor` is the inspiration of this name. 
```Scala
def lift[A, B](f: A => B): F[A] => F[B] =
    fa => map(fa)(f) 
```
I think `liftNT` or `liftFunctionK` is fine since it's about lifting a FunctionK to another FunctionK to the Kleisli space. `lift` is an general action, to be honest, I don't see much confusion if you have `lift` that  lifts a `F[B]` or `B` to a `Kleisli`, and `liftNT` that lifts a `FunctionK` of `F`, `G` to another FunctionK of `Kleisli[F]` and `Kleisli[G]`

kubukoz:
That makes sense, yeah. I think `liftNT` will be good enough then :)

kubukoz:
Renamed to `liftNT`, added a test for consistency with `mapK` and some scaladoc - let's proceed :)

kubukoz:
Apparently the addition isn't binary compatible, but only on 2.11. Any ideas, @kailuowang?

LukaJCB:
Make a new trait `KleisliFunctionsBinCompat`, add the function there and then mix it into `KleisliFunctions` :)

kubukoz:
hmm, now I see that you mentioned `object Kleisli`, won't it work if I mix it into `KleisliFunctions`?

kubukoz:
@LukaJCB hi, is there anything else I need to do with this one?

core/src/main/scala/cats/data/Kleisli.scala
```diff
@@ -111,21 +110,21 @@ object Kleisli extends KleisliInstances with KleisliFunctions with KleisliExplic
     }
 }
 
-sealed private[data] trait KleisliFunctions {
+private[data] sealed trait KleisliFunctions {
```
LukaJCB: Was this done for a particular reason? :)

core/src/main/scala/cats/data/Kleisli.scala
```diff
@@ -111,21 +110,21 @@ object Kleisli extends KleisliInstances with KleisliFunctions with KleisliExplic
...
-   */
+    * Same as [[liftF]], but expressed as a FunctionK for use with mapK
+    * {{{
+    * scala> import cats._, data._, implicits._
+    * scala> val a: OptionT[Eval, Int] = 1.pure[OptionT[Eval, ?]]
+    * scala> val b: OptionT[Kleisli[Eval, String, ?], Int] = a.mapK(Kleisli.liftK)
+    * scala> b.value.run("").value
+    * res0: Option[Int] = Some(1)
+    * }}}
+    */
```
LukaJCB: This is IntelliJ autoformating, I assume?

kubukoz: ugh. For some reason it didn't enable scalafmt in cats, it seems...

LukaJCB:
Sorry @kubukoz, it looks good, just a couple of nitpicks :)

core/src/main/scala/cats/data/Kleisli.scala
kubukoz: no, just a rebase issue - I'll revert

kailuowang:
@kubukoz build failed due to scalafmt check, would you run `sbt prePR` and commit the changes?

kubukoz:
Done :)

kubukoz:
@kailuowang WDYT about merging this?

kailuowang:
@kubukoz sorry I dropped the ball on this one. And after another look, I am not sure about the `liftNT` name I suggested. We renamed `NaturalTransformation` to `FunctionK` a while ago so `NT` might be confusing.  Would it be possible to rename it to `liftFK` or `liftFunctionK`? Sorry about the back and forth.

kubukoz:
It's okay, I know how important it is to get the API right when you can't rename it for a while :)

We can definitely rename, and although I like shorter names I'll say `liftFK`'s `FK` part reminds me too much of "foreign key" and I'd go with `liftFunctionK` to be more explicit.

LukaJCB:
I think you renamed the wrong function :P

kubukoz:
Yeah, just noticed. Time does things to people...

kubukoz:
(Rerunning due to the neverending Travis build on 2.13)

Issue #2322, opened 2018-07-10
Add partial folding methods to Foldable


LukaJCB:
We can define functions that only tear down a subset of the structure by using something like a `PartialFunction` or `Option` to exclude individual elements from the aggregation.
E.g.:

```scala
def filterFold[A, M: Monoid](fa: F[A])(f: A => Option[M]): M =
  foldLeft(fa, Monoid[M].empty)((acc, cur) => f(cur) match {
    case Some(m) => acc |+| m
    case _ => acc
  })

def collectFold[A, M: Monoid](fa: F[A])(f: PartialFunction[A, M]): M =
   foldLeft(fa, Monoid[M].empty)((acc, cur) => acc |+| f.applyOrElse(cur, _ => Monoid[M].empty))

```

These should be equivalent to `collect(f).fold` and `filter(f).fold` respectively, so we could add some laws for that.

satansk:
Let me give it a try :)

LukaJCB:
Go ahead :)

peterneyens:
I wonder in what way this relates to `FunctorFilter` and `TraverseFilter` that were moved to cats-mtl ( `FunctorEmpty` and `TraverseEmpty`).

denisrosca:
@LukaJCB afaict `Foldable` doesn't have `collect` and `filter` defined so I'm not sure how the laws you propose would be implemented?

LukaJCB:
@denisrosca they'd have to be implemented on `FunctorEmpty` in cats-mtl, though I guess we could formulate some different laws, i.e. by using `toList` or using some other properties:

For example:
```scala
forAll  { (fa: F[A], f: A => Boolean) =>
  fa.forall(f) <-> (fa.filterFold(a => if (f(a)) Some(a) else None) === fa.combineAll)
}
```

LukaJCB:
@satansk Have you had a shot at this? Are you blocked by something? Is it something I could help with? :)

satansk:
@LukaJCB thanks, and yes, I did encounter a problem, I can't import the project into intellij idea, and here is the error message: 

```Scala
[error] coursier.ResolutionException: Encountered 5 error(s) in dependency resolution:
[error]     org.scala-native:nativelib_native0.3_2.12:0.3.7:
[error]         not found:
[error]             /Users/satansk/.ivy2/local/org.scala-native/nativelib_native0.3_2.12/0.3.7/ivys/ivy.xml
[...]

PR #2452, opened 2018-09-01
Add filterFold and collectFold to Foldable

satansk:
fix https://github.com/typelevel/cats/issues/2322

### DIFF
--- a/core/src/main/scala/cats/Foldable.scala
+++ b/core/src/main/scala/cats/Foldable.scala
@@ -262,7 +262,7 @@ import Foldable.sentinel
   def filterFold[A, M: Monoid](fa: F[A])(f: A ⇒ Option[M]): M = {
     val m = Monoid[M]
     foldLeft(fa, m.empty)((acc, a) ⇒ f(a) match {
-      case Some(x) ⇒ m.combine(x, acc)
+      case Some(x) ⇒ m.combine(acc, x)
       case None    ⇒ acc
     })
   }
--- a/core/src/main/scala/cats/Foldable.scala
+++ b/core/src/main/scala/cats/Foldable.scala
@@ -633,11 +633,11 @@ object Foldable {
    * It could be made a value class after
    * https://github.com/scala/bug/issues/9600 is resolved.
    */
-  private sealed abstract class Source[+A] {
+  private[cats] sealed abstract class Source[+A] {
     def uncons: Option[(A, Eval[Source[A]])]
   }
 
-  private object Source {
+  private[cats] object Source {
     val Empty: Source[Nothing] = new Source[Nothing] {
       def uncons = None
     }
--- a/core/src/main/scala/cats/Foldable.scala
+++ b/core/src/main/scala/cats/Foldable.scala
@@ -272,9 +272,7 @@ import Foldable.sentinel
    *{{{
    * scala> import cats.implicits._
    * scala> val xs = List(1, 2, 3, 4)
-   * scala> xs.collectFold {
-   *          case n if n % 2 == 0 => n
-   *        }
+   * scala> xs.collectFold { case n if n % 2 == 0 => n }
    * res0: Int = 6
    *}}}
    */
--- a/core/src/main/scala/cats/Foldable.scala
+++ b/core/src/main/scala/cats/Foldable.scala
@@ -250,6 +250,40 @@ import Foldable.sentinel
   def combineAll[A: Monoid](fa: F[A]): A = fold(fa)
 
   /**
+   * Tear down a subset of this structure using a `A => Option[M]`.
+   *{{{
+   * scala> import cats.implicits._
+   * scala> val xs = List(1, 2, 3, 4)
+   * scala> def f(n: Int): Option[Int] = if (n % 2 == 0) Some(n) else None
+   * scala> xs.filterFold(f)
+   * res0: Int = 6
+   *}}}
+   */
+  def filterFold[A, M: Monoid](fa: F[A])(f: A ⇒ Option[M]): M = {
+    val m = Monoid[M]
+    foldLeft(fa, m.empty)((acc, a) ⇒ f(a) match {
+      case Some(x) ⇒ m.combine(x, acc)
+      case None    ⇒ acc
+    })
+  }
+
+  /**
+   * Tear down a subset of this structure using a `PartialFunction`.
+   *{{{
+   * scala> import cats.implicits._
+   * scala> val xs = List(1, 2, 3, 4)
+   * scala> xs.collectFold {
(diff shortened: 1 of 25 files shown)


core/src/main/scala/cats/Foldable.scala
```diff
@@ -249,6 +249,38 @@ import Foldable.sentinel
...
+   * scala> val xs = List(1, 2, 3, 4)
+   * scala> def f(n: Int): Option[Int] = if (n % 2 == 0) Some(n) else None
+   * scala> xs.filterFold(f)
+   * res0: Int = 6
+   *}}}
+   */
+  def filterFold[A, M: Monoid](fa: F[A])(f: A ⇒ Option[M]): M = {
+    val m = Monoid[M]
+    foldLeft(fa, m.empty)((acc, a) ⇒ f(a) match {
+      case Some(x) ⇒ m.combine(x, acc)
```
johnynek: I think the order on combine is wrong. I think this won’t match collectFold for a non commutative Monoid.

tests/src/test/scala/cats/tests/FoldableSuite.scala
```diff
@@ -85,6 +85,20 @@ abstract class FoldableSuite[F[_]: Foldable](name: String)(
     }
   }
 
+  test(s"Foldable[$name] partial summation") {
+    forAll { (fa: F[Int], f: Int ⇒ Boolean) ⇒
+      val m: Monoid[Int] = Monoid[Int]
+
```
johnynek: Let’s test with a non commutative Monoid like String.

core/src/main/scala/cats/Foldable.scala
satansk: Yes, that's right, thanks, I will fix it soon :)

satansk:
The CI complains an binary compatibility issue:

```Scala
[error]  * method filterFold(java.lang.Object,scala.Function1,cats.kernel.Monoid)java.lang.Object in trait cats.Foldable is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("cats.Foldable.filterFold")
[error]  * method collectFold(java.lang.Object,scala.PartialFunction,cats.kernel.Monoid)java.lang.Object in trait cats.Foldable is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("cats.Foldable.collectFold")
[error]  * method filterFold(scala.Function1,cats.kernel.Monoid)java.lang.Object in trait cats.Foldable#Ops is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("cats.Foldable#Ops.filterFold")
[error]  * method collectFold(scala.PartialFunction,cats.kernel.Monoid)java.lang.Object in trait cats.Foldable#Ops is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("cats.Foldable#Ops.collectFold")
```

I will fix it latter.

core/src/main/scala/cats/syntax/foldable.scala
```diff
@@ -158,4 +158,37 @@ final class FoldableOps[F[_], A](val fa: F[A]) extends AnyVal {
...
+    * Tear down a subset of this structure using a `A => Option[M]`.
+    *{{{
+    * scala> import cats.implicits._
+    * scala> val xs = List(1, 2, 3, 4)
+    * scala> def f(n: Int): Option[Int] = if (n % 2 == 0) Some(n) else None
+    * scala> xs.filterFold(f)
+    * res0: Int = 6
+    *}}}
+    */
+  def filterFold[M: Monoid](f: A ⇒ Option[M])(implicit F: Foldable[F]): M = {
```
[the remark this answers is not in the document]
LukaJCB: We could indeed define these in terms of each other. I'm not sure if they'd be more performant than just using `foldLeft`, but if we do this, then it should be `filterFold` in terms of `collectFold`, since `A => Option[M]` has to allocate an extra `Option` :) 

Up to you, @satansk :)

core/src/main/scala/cats/syntax/foldable.scala
```diff
@@ -158,4 +158,37 @@ final class FoldableOps[F[_], A](val fa: F[A]) extends AnyVal {
...
+    * Tear down a subset of this structure using a `PartialFunction`.
+    *{{{
+    * scala> import cats.implicits._
+    * scala> val xs = List(1, 2, 3, 4)
+    * scala> xs.collectFold { case n if n % 2 == 0 => n }
+    * res0: Int = 6
+    *}}}
+    */
+  def collectFold[M: Monoid](f: PartialFunction[A, M])(implicit F: Foldable[F]): M = {
+    val m = Monoid[M]
```
kailuowang: totally nitpick - there are two styles of implicit type class instance summoning in this one method (and the `filterFold` method below). I think it would help readability a little bit if we just use `(implicit F: Foldable[F], M: Monoid[M])`

satansk: This is a nice advice, it will be more readable, and I'll fix it :)

Thanks!

core/src/main/scala/cats/syntax/foldable.scala
```diff
@@ -158,4 +158,33 @@ final class FoldableOps[F[_], A](val fa: F[A]) extends AnyVal {
...
+    * Tear down a subset of this structure using a `A => Option[M]`.
+    *{{{
+    * scala> import cats.implicits._
+    * scala> val xs = List(1, 2, 3, 4)
+    * scala> def f(n: Int): Option[Int] = if (n % 2 == 0) Some(n) else None
+    * scala> xs.filterFold(f)
+    * res0: Int = 6
+    *}}}
+    */
+  def filterFold[M](f: A ⇒ Option[M])(implicit F: Foldable[F], M: Monoid[M]): M =
```
kailuowang: The word `filter`, to me, kind of implies using a boolean value. 
We already have a `collectFirstSome` https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Foldable.scala#L233
how about naming this one `collectSomeFold`?

LukaJCB: We do have `mapFilter` which does use `Option` though. 
Maybe this should be `mapFilterFold`? I don't know

kailuowang: Right, I am fine with `mapFilterFold` too. Sorry about the bikeshedding @satansk , your call.

LukaJCB:
:+1:

hepin1989:
@satansk +100 ：）

satansk:
@hepin1989 :)

PR #2454, opened 2018-09-01
Add maxByOption and minByOption for Foldable

### LINKED ISSUE #2385: Add Foldable#[max|min]ByOption, Reducible#[maxBy|minBy]

leusgalvan:
Fixes #2385 

It's not ready for merging, though. I didn't implement methods for Reducible because I'm not quite sure where to add them. I didn't find a class similar to FoldableOps (something like ReducibleOps) in the syntax for Reducible. 

What's the correct approach? Should I just add ReducibleOps with this new method?


leusgalvan:
Yes, no problem.

leusgalvan:
@johnynek It's ready now. Please review it when you like. Thanks!

leusgalvan:
@LukaJCB Yes, I left it because I think the implementation with reduceLeftOption (in Foldable) might be a little bit faster than the one with unorderedFoldMap. I haven't actually tested it, though.

leusgalvan:
@kailuowang Anything I can do to help push this forward? Perhaps merging with the latest master and making Travis pass? :)

core/src/main/scala/cats/syntax/reducible.scala
```diff
@@ -11,3 +10,48 @@ final class NestedReducibleOps[F[_], G[_], A](val fga: F[G[A]]) extends AnyVal {
...
+    * scala> val list1 = List[String]("Two", "Three", "Four")
+    * scala> list1.maxBy(_.length)
+    * res0: String = Three
+    *
+    * scala> val list2 = List[Int](41, 32, 23)
+    * scala> list2.maxBy(_ % 10)
+    * res1: Int = 23
+    * }}}
+    */
+  def maxBy[B](f: A => B)(implicit F: Reducible[F], B: Order[B]): A =
```
johnynek: same foldMap question

core/src/main/scala/cats/syntax/foldable.scala
```diff
@@ -159,6 +159,57 @@ final class FoldableOps[F[_], A](val fa: F[A]) extends AnyVal {
...
+    * scala> list2.maxByOption(_.length)
+    * res1: Option[String] = Some(Three)
+    *
+    * scala> val list3 = List[Int](41, 32, 23)
+    * scala> list3.maxByOption(_ % 10)
+    * res2: Option[Int] = Some(23)
+    * }}}
+    */
+  def maxByOption[B](f: A => B)(implicit F: Foldable[F], B: Order[B]): Option[A] =
+    F.reduceLeftOption[A](fa)((a1, a2) => if (B.gteqv(f(a1), f(a2))) a1 else a2)
```
johnynek: why not use foldMap here which could potentially leverage associativity?

core/src/main/scala/cats/syntax/foldable.scala
```diff
@@ -159,6 +159,57 @@ final class FoldableOps[F[_], A](val fa: F[A]) extends AnyVal {
...
+    * scala> list2.minByOption(_.length)
+    * res1: Option[String] = Some(Two)
+    *
+    * scala> val list3 = List[Int](41, 32, 23)
+    * scala> list3.minByOption(_ % 10)
+    * res2: Option[Int] = Some(41)
+    * }}}
+    */
+  def minByOption[B](f: A => B)(implicit F: Foldable[F], B: Order[B]): Option[A] =
+    F.reduceLeftOption[A](fa)((a1, a2) => if (B.lteqv(f(a1), f(a2))) a1 else a2)
```
johnynek: why not use foldMap here?

johnynek:
I don't see why the implementation using unorderedFoldMap would be slower (indeed, perhaps faster since you are promising less requirements for in order iteration). I'd rather just do the Unordered versions unless we have evidence we can get a win.

leusgalvan:
@johnynek Yes, you are right. I got carried away by the bias that subclasses, being more concrete, tend to have faster implementations that superclasses. In this case I forgot that unorderedFoldMap can actually provide much faster iteration. Thanks for your clarification.

ceedubs:
Thanks @leusgalvan, this is a great addition! Sorry, but you have some merge conflicts; would you mind resolving those?

The only addition that I can think to make (though I'm not sure whether or not it is worth it) is that potentially we could mane the `CommutativeMonoid` instances that you've written package-private instead of `private` so that we could test the `CommutativeMonoid` laws for them.

johnynek:
@ceedubs will that have any different binary compatibility impacts? I don’t see the needs to test the monoids AND the functions and the functions seem well tested, which I think is all we care about.

core/src/main/scala/cats/syntax/unorderedFoldable.scala
```diff
@@ -0,0 +1,79 @@
...
+    * scala> set2.maxByOption(_.length)
+    * res1: Option[String] = Some(Three)
+    *
+    * scala> val set3 = Set(41, 32, 23)
+    * scala> set3.maxByOption(_ % 10)
+    * res2: Option[Int] = Some(23)
+    * }}}
+    */
+  def maxByOption[B](f: A => B)(implicit F: UnorderedFoldable[F], B: Order[B]): Option[A] =
+    F.unorderedFoldMap(fa)(Option.apply)(orderMaxCommutativeMonoidBy(f))
```
ceedubs: I think that we should use `Some.apply` instead of `Option.apply` here. We probably don't want to silently remove any `null` values that they have (though hopefully they don't have any).

core/src/main/scala/cats/syntax/unorderedFoldable.scala
```diff
@@ -0,0 +1,79 @@
...
+    * scala> set2.minByOption(_.length)
+    * res1: Option[String] = Some(Two)
+    *
+    * scala> val set3 = Set[Int](41, 32, 23)
+    * scala> set3.minByOption(_ % 10)
+    * res2: Option[Int] = Some(41)
+    * }}}
+    */
+  def minByOption[B](f: A => B)(implicit F: UnorderedFoldable[F], B: Order[B]): Option[A] =
+    F.unorderedFoldMap(fa)(Option.apply)(orderMinCommutativeMonoidBy(f))
```
ceedubs: Same as other comment about `Option.apply`

ceedubs:
In the current implementation, `f` is repeatedly called on the current "winner" as it is compared to each item in the structure. I think that we could prevent this by doing something like the following:

```scala
  private def orderMaxCommutativeMonoidBy[B](implicit ev: Order[B]): CommutativeMonoid[Option[(A, B)]] =
    new CommutativeMonoid[Option[(A, B)]] {
      override def empty = None
      override def combine(optX: Option[(A, B)], optY: Option[(A, B)]): Option[(A, B)] = (optX, optY) match {
        case (None, y) => y
        case (x, None) => x
        case (l @ Some(x), r @ Some(y)) => if (ev.gteqv(x._2, y._2)) l else r
      }
    }


  def maxByOption[B](f: A => B)(implicit F: UnorderedFoldable[F], B: Order[B]): Option[A] =
    F.unorderedFoldMap[A, Option[(A, B)]](fa)(a => Some((a, f(a))))(orderMaxCommutativeMonoidBy)
    .map(_._1)
```

The tradeoff is that we repeatedly allocate a `Tuple2` as opposed to repeatedly calling `f`, and it's quite possible that calling `f` is cheaper (perhaps it's just referencing an `Int` in the `A` value). However, the tuple approach should have a better worse case (expensive `f` operation), and I would imagine would be less of a potential gotcha to the user. I'd be particularly interested in what @johnynek thinks about this, as he is good about being performance-conscious :)

core/src/main/scala/cats/syntax/unorderedFoldable.scala
```diff
@@ -0,0 +1,79 @@
...
+}
+
+final class UnorderedFoldableOps[F[_], A](val fa: F[A]) extends AnyVal {
+  private def orderMaxCommutativeMonoidBy[B](f: A => B)(implicit ev: Order[B]): CommutativeMonoid[Option[A]] =
+    new CommutativeMonoid[Option[A]] {
+      override def empty = None
+      override def combine(optX: Option[A], optY: Option[A]): Option[A] = (optX, optY) match {
+        case (None, y) => y
+        case (x, None) => x
+        case (Some(x), Some(y)) => if (ev.gteqv(f(x), f(y))) Some(x) else Some(y)
```
ceedubs: We should be able to prevent some `Some` allocations here by changing this to `case (l @ Some(x), r @ Some(y)) => if (ev.gteqv(f(x), f(y))) l else r`. Similarly for the min monoid below.

leusgalvan: Yes, you're right.

leusgalvan:
@ceedubs I post a PR with the original solution with the changes you proposed and the conflicts fixed. If we have @johnynek opinion and conclude that the best idea is to switch to your proposed solution I'll be more than happy to update it!

One further question if I may. Up until now I've been adding empty commits to force travis to build again for the cases where it fails for reasons not related to my code. Is there any other preferred way to do it?

johnynek:
Thinking on this more, if we are uptight @ceedubs is right.

The commutative monoid here is not commutative under a general `Eq[A]`. It is commutative under `Order[B].contramap(fn)` which is to say the ordering the user claims to care about (because it is exactly the max operation on that ordering.

It does make me a little nervous that two different equivalent `f: F[_]: UnorderedFoldable` could be `eqv(f1, f2)` but `neqv(f1.maxByOption(g), f2.maxByOption(g))`.

The differences are kind of exactly what the user claims not to care about, never the less, the non-deterministic nature is annoying to me.

A probably terrible solution would be to also require `Hash[A]` and return a `Set[A]` of values that all take the maximum value for `B`.

In Foldable we don't have this issue since we have a single linearizable structure, so we should give the same result in both orders using `foldMap`....

It may seem like we are being uptight, but usually when we do something less principled we eventually get burned.

johnynek:
of course, we don't actually have a `Set[A]` that uses a typeclass `Hash[A]` so, that's terrible...

ceedubs:
This is a tricky one.

I think that in https://github.com/typelevel/cats/pull/2454#issuecomment-426872070 @johnynek provides a pretty good example of a potentially confusing result of this.

In this implementation we are creating a `CommutativeMonoid` instance that we know could fail the commutativity laws because `f: A => B` can be throwing away `A` information, even if the `Order[B]` is a completely lawful instance that doesn't throw away any `B` information. I think that this is a little different than if we were to do this with `maximumOption` and `minimumOption`, since those take an `Order[A]` instance, so your `CommutativeMonoid` instance would in fact be commutative if you use this `Order[A]` instance as your definition of equality. Somebody could still provide an `Order[A]` instance by doing something like `Order.by(_.someField)`, but their definition of equality for `A` is then explicitly defined as being by only `someField`. This is a bit different than the scenario in which we have an `Order[A]` and a separate `(A => B, Order[B])` and the two don't necessarily provide consistent results.

Ultimately I'm inclined to think the following:

- Compatibility permitting, we could reasonably move `maximumOption` and `minimumOption` from `Foldable` to `UnorderedFoldable`.
- `maxByOption` and `minByOption`, if they exist, should probably live on `Foldable` and not `UnorderedFoldable`.
- If `maximumOption` and `minimumOption` were to exist on `UnorderedFoldable`, then they could pretty easily be used instead of `maxByOption` and `minByOption` by providing an `Order` instance that uses `Order.by` or something similar. But `maxByOption` and `minByOption` might be nice helper methods, and they could help satisfy people who are adamant about type class coherence who wouldn't want to create a local `Order` instance.

LukaJCB:
Hey folks, I still think this would be a good thing to have! 
@leusgalvan do you want to continue working on this? Otherwise I could take a shot at getting this to a mergeable state :)

leusgalvan:
@LukaJCB I won't be able to work on it so feel free to take it!

PR #2459, opened 2018-09-03
Add documentation for FlatMap and Bifunctor

asoltysik:


docs/src/main/tut/typeclasses/bifunctor.md
```diff
@@ -0,0 +1,52 @@
+---
+layout: docs
+title:  "Functor"
```
LukaJCB: This should be Bifunctor I assume? :)

kailuowang:
`Tuple2` is a good example to add that as well. I propose we merge and add that in a separate PR.

PR #2460, opened 2018-09-03
#2429 Replace all usage of the list Monoid in the docs with the chain Monoid

### LINKED ISSUE #2429: Replace all usage of the list Monoid in the docs with the chain Monoid
Using `Chain`s Monoid instead of `List`s can give us an enormous performance boost. We should encourage our users to use it instead of list in many cases, such as, `Validated`, `Ior`, `Writer` or `Const`.

jan0sch:
* replace `NonEmptyList` with `NonEmptyChain` in `Ior` and `IorT` docs

* fixes #2429

### DIFF
--- a/docs/src/main/tut/datatypes/ior.md
+++ b/docs/src/main/tut/datatypes/ior.md
@@ -49,9 +49,9 @@ Here's an example of how we might be able to do that:
 
 ```tut:silent
 import cats.implicits._
-import cats.data.{ NonEmptyList => Nel }
+import cats.data.{ NonEmptyChain => Nec }
 
-type Failures = Nel[String]
+type Failures = Nec[String]
 
 case class Username(value: String) extends AnyVal
 case class Password(value: String) extends AnyVal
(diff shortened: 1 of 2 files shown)


jan0sch:
Please comment if further changes should be included in the pull request.

docs/src/main/tut/datatypes/ior.md
```diff
@@ -98,22 +98,23 @@ To extract the values, we can use the `fold` method, which expects a function fo
...
-type IorNel[B, A] = Ior[NonEmptyList[B], A]
+type IorNec[B, A] = Ior[NonEmptyChain[B], A]
 ```
 
 
 ```tut
+import cats.implicits._, cats.data.NonEmptyChain
 
-val left: IorNel[String, Int] = Ior.leftNel("Error")
+val left: IorNec[String, Int] = Ior.fromEither("Error".leftNec[Int])
```
LukaJCB: Seems I missed the convenience function here, we should add an Issue for that too. :)

PR #2467, opened 2018-09-04
Post 1.3.0 Release commits

kailuowang:


kailuowang:
@johnynek mima pointer is automatic it checks against all prior 1.x releases

PR #2471, opened 2018-09-05
add back the implicit method as explicit to maintain BC

kailuowang:


kailuowang:
@rossabaker I just removed the exception

rossabaker:
We could have automated test on exceptions.  Compile this in an unpublished cats module that builds against cats-core in Provided scope:

```scala
object MimaExceptions {
    def isBinaryCompatible = {
      cats.data.OptionT.catsDataTraverseForOptionT[List]
      cats.data.Kleisli.catsDataCommutativeArrowForKleisliId
      cats.data.OptionT.catsDataMonoidKForOptionT[List]
      cats.data.OptionT.catsDataMonoidForOptionT[List, Int]
      cats.data.Kleisli.catsDataMonadForKleisliId[Int]
      cats.data.Kleisli.catsDataCommutativeArrowForKleisli[Option]
      cats.data.Kleisli.catsDataCommutativeFlatMapForKleisli[Option, Int]
      cats.data.IRWST.catsDataStrongForIRWST[List, Int, Int, Int]
      cats.data.OptionT.catsDataMonadErrorMonadForOptionT[List]
      true
  }
}
```

And then the test, which depends on the unpublished module:

```scala
class MimaExceptionsTest {
  test("is binary compatible") { 
    MimaExceptions.isBinaryCompatible
  }
}
```

It's more or less what I did in a side repo to verify the others.  And it would have caught this one.

rossabaker:
Note: I don't want the above test idea to slow the bugfix.  If others like the idea, we can always add it after 1.3.1.

PR #2472, opened 2018-09-05
Replace List with Chain in Validated's docs

satansk:
* `ValidatedNel` to `ValidatedNec`
* `NonEmptyList` to `NonEmptyChain`
* `List` to `Chain`


LukaJCB:
This is great, thanks!
It still seems to run into some errors with `tut` though, it'd be great if you could fix those :)

satansk:
Hi @LukaJCB , the reason for the `tut` errors is the missing `toValidatedNec` method of `Validated`, there's a `toValidatedNel` in it.

So should I add `toValidatedNec` to `Validated`, or what else I can do to avoid the errors?

Thanks :)

LukaJCB:
Can you merge this with master now that #2487 is merged? :)

kailuowang:
Thank you!

PR #2475, opened 2018-09-06
Add Either#toEitherNel extension and friends

kubukoz:
Also brings `Ior#toIorNel`.

I'm also pondering submitting an `EitherTNel` type alias in the `cats.data` package object, together with conversions from `EitherT`, but then we'd need to do the same for `IorT` I guess.


kubukoz:
I imagine it's going to get tedious if we have all these helpers for `NonEmptySet` and `NonEmptyChain`...

ceedubs:
To add to the permutations: should we be making these for `NonEmptyChain` now? That seems to be the direction that we are encouraging people to go with #2472. I'm guessing that supporting both would be most helpful (but starts to feel like potential bloat).

kubukoz:
Maybe we could generalize with something like `.leftPure[NonEmptyChain]`
for now

catostrophe:
Agree with @ceedubs - the whole codebase should be reviewed and each `*Nel` function should get its `*Nec` pair. I tried to switch to `Nec` in my project via "replace all" but couldn't due to lack of some functions.

kubukoz:
Created #2539 to track NEC equivalents. I suggest we proceed with this one as it is :)

kubukoz:
Can we merge? It'll unblock #2543 which will conflict with this

kubukoz:
Thanks!

PR #2480, opened 2018-09-07
improve interaction with subtyping in OptionOps#liftTo

mberndt123:
Currently OptionOps#liftTo interacts badly with subtyping. Something like this won't compile:
```scala
42.some.liftTo[IO](new Exception("foo"))
```
One needs to write this instead: 
```scala
42.some.liftTo[IO](new Exception("foo"): Throwable)
```
This patch fixes this.

### DIFF
--- a/core/src/main/scala/cats/ApplicativeError.scala
+++ b/core/src/main/scala/cats/ApplicativeError.scala
@@ -187,7 +187,7 @@ object ApplicativeError {
   def apply[F[_], E](implicit F: ApplicativeError[F, E]): ApplicativeError[F, E] = F
 
   private[cats] final class LiftFromOptionPartially[F[_]](val dummy: Boolean = true) extends AnyVal {
-    def apply[E, A](oa: Option[A], ifEmpty: => E)(implicit F: ApplicativeError[F, E]): F[A] =
+    def apply[E, A](oa: Option[A], ifEmpty: => E)(implicit F: ApplicativeError[F, _ >: E]): F[A] =
       oa match {
         case Some(a) => F.pure(a)
         case None => F.raiseError(ifEmpty)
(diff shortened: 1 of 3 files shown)


kailuowang:
Also the liftTo on either can be improved the same way right?

kailuowang:
Thanks! BTW, in case you haven't used doctest before, [here is an example of PR adding doctests](https://github.com/typelevel/cats/pull/898/files), it would nice to have test make sure that this new interaction works as intended.

mberndt123:
I added the same change to EitherOps#liftTo.

I also added doctests. The latter turned out to be a PITA. I added the doctests, and they compiled and ran. In order to see whether they actually exercise my new code, I changed the `_ >: E` in `OptionOps.LiftToPartiallyApplied` part back to `E` and expected them to fail. But they still ran just fine, and I couldn't figure out what the hell was going on… Until, after quite a bit of head scratching, I ran `clean` in sbt before `coreJVM/test` and, lo and behold, the tests started failing as expected!

Oh well, one more scala war story to tell.

mberndt123:
I've simplified the tests a bit. The reason I used `EitherT[Option, CharSequence, ?]` instead of just `Either[CharSequence, ?]` in the test is that my test was passing when I expected it to fail. I suspected that `Either`'s covariance might have something to do with it, so I changed the test to use `EitherT`, which is invariant. But the actual cause turned out to be the above-mentioned incremental compilation bug,

mberndt123:
@kailuowang @LukaJCB could you approve again?

mberndt123:
This PR was approved and all review comments were addressed, so to me it looks like it's ready to be merged. Is there anything else to do to get this merged?

LukaJCB:
Sorry for the delay, merging now :)

PR #2482, opened 2018-09-08
Update applicative.md

amarrella:
Added observation wrt `Semigroup[A].combine` versus `Applicative[F[_]].product`.

I'm not sure whether you want to add this information (which might be obvious) and if this is the right place to add it but to me discovering that combining via applicative and combining via semigroup was different was quite surprising.

### DIFF
--- a/docs/src/main/tut/typeclasses/applicative.md
+++ b/docs/src/main/tut/typeclasses/applicative.md
@@ -66,6 +66,12 @@ Such an `Applicative` must obey three laws:
     * `fa.product(pure(())) ~ fa`
     * As an equality: `fa.product(pure(())).map(_._1) = fa`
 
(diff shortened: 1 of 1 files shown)


kailuowang:
The types of `Semigroup.combine` and `Applicative.product` is quite different. It actually never occured to me that people associate them together. The example you gave acutally hide the type difference - the `None` in the second line is of type `Option[(Int, Nothing)]`. So maybe that was where the confusion came from?

amarrella:
Thanks for the clarification! I thought they were more similar (as i thought that `Applicative` - or more precisely Semigroupal) was using a Semigroup instance to combine the values but turns out in Scala it's just doing putting them into a tuple "manually" as in the snippet below (from cats.Apply).
```scala
    ap(map(fa)(a => (b: B) => (a, b)))(fb)
```

kailuowang:
I am glad that it helped. shall we close this one?

amarrella:
Definitely, thanks again :) Closing it

PR #2485, opened 2018-09-10
Proposal - add cats-tagless as an external module

kailuowang:
Not a module inside Cats repo, it's still in its own repo with its own release cycle just like the other 4 "outside" modules cats-mtl, cats-effect, mouse and kittens. 
Nothing really special about being an official cats external module other than some sort of commitment that Cats maintainers are going to help its maintenance. As of now, the majority of the code of cats-tagless came from two projects from two Cats maintainer @LukaJCB and myself. I plan to remain a maintainer of the project. (how about you @LukaJCB ?)


LukaJCB:
I'm on board! :)

non:
👍

ChristopherDavenport:
👍

PR #2487, opened 2018-09-12
Add toValidatedNec to Validated

satansk:
fix the problems in https://github.com/typelevel/cats/pull/2472


kailuowang:
shall we add some tests?

satansk:
@kailuowang I'll add tests tonight :)

PR #2488, opened 2018-09-12
Kill scala-collection-compat

travisbrown:
Are we sure we even need this? I just ran `buildJVM` on my laptop for 2.13.0-M4 and it worked just fine.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -36,8 +36,6 @@ lazy val commonSettings = Seq(
     Resolver.sonatypeRepo("releases"),
     Resolver.sonatypeRepo("snapshots")),
   fork in test := true,
-  libraryDependencies ++= Seq(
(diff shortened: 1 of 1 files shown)


kailuowang:
Just went through the original PR adding M4 support, it is not needed indeed.
What happened was that there were a lot of deprecations in collections whose migration can't be cross compiled without `scala-collection-compat`. it turned out that even with `scala-collections-compat` it was still too much work, so we ended up disabling the deprecation warnings in 2.13-M4 build and roll back those changes. The dependency, however, was left alone. 

Once build green I am going to merge this one and cherry-pick it back to 1.3.x branch and release a 1.3.2. 
Thanks @travisbrown!

travisbrown:
Thanks much @kailuowang, @LukaJCB!

PR #2490, opened 2018-09-13
Invariant For Applicative Error

ChristopherDavenport:
Baseline is just an implementation of Invariant for ApplicativeErrors error type. Not sure if this is wanted so wanted to open discussion before I built out the full tests.

### DIFF
--- a/core/src/main/scala/cats/Invariant.scala
+++ b/core/src/main/scala/cats/Invariant.scala
@@ -121,4 +121,22 @@ object Invariant {
     }
 
   }
+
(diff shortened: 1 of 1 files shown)


ceedubs:
I can't say that I've encountered a need for this personally, but it looks reasonable to me, and admittedly I don't use many of the MTL classes much (probably to a fault).

This does make me mentally go down a rabbit hole of how much more Cats could be doing with instances based on isomorphisms, but I probably shouldn't open up that can of worms here :)

kailuowang:
@LukaJCB it's not ready for review yet, just would like to hear your opinion on this one.

LukaJCB:
I haven't encountered the need for this yet either personally, but we already have typeclass instances for a bunch of other type classes, so I think this is a good addition :)

larsrh:
Closing this for lack of activity.

PR #2505, opened 2018-09-16
Add Chain docs

LukaJCB:
Some documentation for `Chain`. Partially adopted from the blog post I wrote recently, but with more detail and a short mention of `NonEmptyChain` as well.


kailuowang:
thank you @easel and @Daenyth for the review, really helpful.

PR #2506, opened 2018-09-16
Add ContT monad

johnynek:
ContT is a challenging Monad to implement if you need to implement tailRecM.

The addition of the Defer typeclass solves this problem. This PR replaces the unmerged #1400.

cc @non

### DIFF
--- a/core/src/main/scala/cats/data/ContT.scala
+++ b/core/src/main/scala/cats/data/ContT.scala
@@ -9,15 +9,16 @@ package data
  * a tailRecM implementation we leverage the Defer type class to
  * obtain stack-safety.
  */
-sealed trait ContT[M[_], A, B] extends Serializable { self =>
+sealed abstract class ContT[M[_], A, B] extends Serializable { self =>
   final def run: (B => M[A]) => M[A] = runAndThen
   protected def runAndThen: AndThen[B => M[A], M[A]]
 
-  def map[C](fn: B => C): ContT[M, A, C] = {
+  def map[C](fn: B => C)(implicit M: Defer[M]): ContT[M, A, C] = {
     // allocate/pattern match once
     val fnAndThen = AndThen(fn)
     ContT { fn2 =>
-      run(fnAndThen.andThen(fn2))
+      val cb = fnAndThen.andThen(fn2)
+      M.defer(run(cb))
     }
   }
 
@@ -35,13 +36,13 @@ sealed trait ContT[M[_], A, B] extends Serializable { self =>
     // lazy to avoid forcing run
     ContT.later(AndThen(fn).andThen(runAndThen))
 
-  def flatMap[C](fn: B => ContT[M, A, C]): ContT[M, A, C] = {
+  def flatMap[C](fn: B => ContT[M, A, C])(implicit M: Defer[M]): ContT[M, A, C] = {
     // allocate/pattern match once
     val fnAndThen = AndThen(fn)
     ContT[M, A, C] { fn2 =>
       val contRun: ContT[M, A, C] => M[A] = (_.run(fn2))
       val fn3: B => M[A] = fnAndThen.andThen(contRun)
-      run(fn3)
+      M.defer(run(fn3))
     }
   }
 }
@@ -79,7 +80,7 @@ object ContT {
       def go(a: A): M[C] =
         fn(a).run {
           case Left(a) => M.defer(go(a))
-          case Right(b) => cb(b)
+          case Right(b) => M.defer(cb(b))
         }
 
       go(a)
--- a/core/src/main/scala/cats/data/ContT.scala
+++ b/core/src/main/scala/cats/data/ContT.scala
@@ -6,14 +6,14 @@ package data
  * the Haskell package Control.Monad.Cont
  *
  * This is reasonably straight-forward except that to make
- * a tailRecM implementation we leverage the Defer typeclass to
+ * a tailRecM implementation we leverage the Defer type class to
  * obtain stack-safety.
  */
 sealed trait ContT[M[_], A, B] extends Serializable { self =>
-  def run(cb: B => M[A]): M[A]
+  def run: (B => M[A]) => M[A]
 
   def map[C](fn: B => C): ContT[M, A, C] =
-    ContT[M, A, C] { fn2 =>
+    ContT { fn2 =>
       run(AndThen(fn).andThen(fn2))
     }
 
@@ -21,13 +21,14 @@ sealed trait ContT[M[_], A, B] extends Serializable { self =>
    * c.mapCont(f).run(g) == f(c.run(g))
    */
   def mapCont(fn: M[A] => M[A]): ContT[M, A, B] =
-    ContT.MapCont(this, fn)
+    // Use later here to avoid forcing run
+    ContT.later(AndThen(run).andThen(fn))
 
   /**
    * cont.withCont(f).run(cb) == cont.run(f(cb))
    */
   def withCont[C](fn: (C => M[A]) => B => M[A]): ContT[M, A, C] =
-    ContT.WithCont(this, fn)
+    ContT { cb => run(fn(cb)) }
 
   def flatMap[C](fn: B => ContT[M, A, C]): ContT[M, A, C] =
     ContT[M, A, C] { fn2 =>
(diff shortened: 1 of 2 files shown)


djspiewak:
Not to be Debbie Downer, but this (like all Cont implementations) still doesn’t have a stack safe tailRecM given a truly arbitrary value of Cont. Specifically, arbitrary left-associated flattens sequenced through continue should blow the stack. This is encodable via tailRecM just as it is via flatMap. There are older examples of this in the cats-effect history (IO is basically Defer + Cont).

johnynek:
It’s an interesting question how many current Monads may fail this test with better generators.

It may be that this isn’t stack safe with more aggressive generators but the use of AndThen doesn’t make it totally obvious to me that is the case. The WithCont/MapCont isn’t safe I guess in recursions.

In any case, without Defer this implementation fails, which I think shows a nice example of using Defer.

kailuowang:
to save you some time digging travis log, scalastyle was complaining
> [info]scalastyle using config /home/travis/build/typelevel/cats/scalastyle-config.xml
[error]/home/travis/build/typelevel/cats/core/src/main/scala/cats/data/ContT.scala:50:8: Public method must have explicit type
[error]/home/travis/build/typelevel/cats/core/src/main/scala/cats/data/ContT.scala:54:8: Public method must have explicit type
[error]/home/travis/build/typelevel/cats/core/src/main/scala/cats/data/ContT.scala:58:8: Public method must have explicit type
[error]/home/travis/build/typelevel/cats/core/src/main/scala/cats/data/ContT.scala:9:51: Use 'type class' instead of typeclass (See #441)

danslapman:
@johnynek FYI: https://github.com/danslapman/cats-conts

core/src/main/scala/cats/data/ContT.scala
```diff
@@ -0,0 +1,108 @@
...
+    // lazy to avoid forcing run
+    ContT.later(AndThen(fn).andThen(runAndThen))
+
+  def flatMap[C](fn: B => ContT[M, A, C]): ContT[M, A, C] = {
+    // allocate/pattern match once
+    val fnAndThen = AndThen(fn)
+    ContT[M, A, C] { fn2 =>
+      val contRun: ContT[M, A, C] => M[A] = (_.run(fn2))
+      val fn3: B => M[A] = fnAndThen.andThen(contRun)
+      run(fn3)
```
johnynek: I wonder if we should take `(implicit M: Defer[M])` and do `M.defer(run(fn3))` here. If we do this, I think it is possibly that this will be a stack safe Monad without a specialized tailRecM. I'll experiment with that.

johnynek:
okay, interestingly, using Defer on each call in map and flatMap makes this a StackSafeMonad, but I went ahead and kept the custom implementation which I think should be more efficient.

What do you think now @djspiewak ? I think this may actually be fully stack safe (at the expense of requiring to defer inside each map/flatMap).

djspiewak:
Reading and pondering… :-D

core/src/main/scala/cats/data/ContT.scala
```diff
@@ -0,0 +1,109 @@
...
+    // lazy to avoid forcing run
+    ContT.later(AndThen(fn).andThen(runAndThen))
+
+  def flatMap[C](fn: B => ContT[M, A, C])(implicit M: Defer[M]): ContT[M, A, C] = {
+    // allocate/pattern match once
+    val fnAndThen = AndThen(fn)
+    ContT[M, A, C] { fn2 =>
+      val contRun: ContT[M, A, C] => M[A] = (_.run(fn2))
+      val fn3: B => M[A] = fnAndThen.andThen(contRun)
+      M.defer(run(fn3))
```
djspiewak: I stand very much corrected here. **I'm convinced that this *is* stack-safe.** I owe you an apology, @johnynek. Dear internets, please link here if you're looking for a handy example of me being wrong.

When I first reviewed this PR on my phone, it looked like a slightly different version of the trick you used in #1400, but this is considerably better. You're basically inheriting the stack-safety of the underlying monad by explicitly hitting a stack-safe join at this point and returning to its trampoline (which it must have due to `Defer`). I'm not sure you even need the `AndThen`, since the stack is cut by the `defer`.

core/src/main/scala/cats/data/ContT.scala
```diff
@@ -0,0 +1,109 @@
...
+  def pure[M[_], A, B](b: B): ContT[M, A, B] =
+    apply { cb => cb(b) }
+
+  def apply[M[_], A, B](fn: (B => M[A]) => M[A]): ContT[M, A, B] =
+    FromFn(AndThen(fn))
+
+  def later[M[_], A, B](fn: => (B => M[A]) => M[A]): ContT[M, A, B] =
+    DeferCont(() => FromFn(AndThen(fn)))
+
+  def tailRecM[M[_], A, B, C](a: A)(fn: A => ContT[M, C, Either[A, B]])(implicit M: Defer[M]): ContT[M, C, B] =
```
djspiewak: Can just delegate to `flatMap` now. Probably slightly slower than this though.

core/src/main/scala/cats/data/ContT.scala
```diff
@@ -0,0 +1,109 @@
...
+      go(a)
+    }
+
+  implicit def catsDataContTDefer[M[_], B]: Defer[ContT[M, B, ?]] =
+    new Defer[ContT[M, B, ?]] {
+      def defer[A](c: => ContT[M, B, A]): ContT[M, B, A] =
+        DeferCont(() => c)
+    }
+
+  implicit def catsDataContTMonad[M[_]: Defer, A]: Monad[ContT[M, A, ?]] =
```
djspiewak: Random thought…

What if we implicitly prioritize a bit here? `ContT` is still useful with very short bind chains even when the underlying monad is not `Defer`, it just wouldn't be stack-safe. So make the higher priority `Monad` instance a `Monad with Defer` and require the `Defer[M]` constraint. Provide a lower priority instance which doesn't require or provide the `Defer[M]` (and thus also has a stack-unsafe `tailRecM`). Obviously this means removing the `flatMap` and `map` implementations from the `ContT` class itself and only making them visible via implicit enrichment (if the imports are a problem for usability, we can `with MonadSyntax` on the companion object).

The advantage would be a more broadly applicable `ContT` (admittedly, I'm not sure how much people care about continuations on the JVM outside of async effects, so… maybe not a big advantage?) which transparently preserves stack-safety whenever possible, and gracefully degrades whenever not. Maybe this is taking the "`tailRecM` doesn't really need to be stack-safe to be lawful" idea too far though.

core/src/main/scala/cats/data/ContT.scala
```diff
@@ -0,0 +1,109 @@
...
+      val fn3: B => M[A] = fnAndThen.andThen(contRun)
+      M.defer(run(fn3))
+    }
+  }
+}
+
+object ContT {
+
+  // Note, we only have two instances of ContT in order to be gentle on the JVM JIT
+  // which treats classes with more than two subclasses differently
```
djspiewak: You can just make `ContT` take `runAndThen` as a parameter to get the same benefit without the subtype limit. :-) A more impactful micro-optimization would be to make the implementations within `ContT` marked as `final`, since then it won't matter how many subtypes you have.

core/src/main/scala/cats/data/ContT.scala
```diff
@@ -0,0 +1,109 @@
...
+      }
+
+    lazy val runAndThen: AndThen[B => M[A], M[A]] = loop(next).runAndThen
+  }
+
+  def pure[M[_], A, B](b: B): ContT[M, A, B] =
+    apply { cb => cb(b) }
+
+  def apply[M[_], A, B](fn: (B => M[A]) => M[A]): ContT[M, A, B] =
+    FromFn(AndThen(fn))
```
djspiewak: I'd also like to see an `applyPure` or something like that defined as:

```scala
def apply[M[_]: Applicative, A, B](fn: (B => A) => A): ContT[M, A, B] =
  apply[M, A, B](_(fn.andThen(_.pure)).pure)
```

Or thereabouts…

johnynek: actually, I don't think we can do that I think we need Applicative and Comonad (we need to go `B => M[A]` to `B => A` to call `fn`.

core/src/main/scala/cats/data/ContT.scala
johnynek: We all make mistakes here and there. I very much admire folks who feel no shame about it. Kudos to you.

core/src/main/scala/cats/data/ContT.scala
djspiewak: Oh I see what you mean. That's definitely unfortunate. I'm not sure the function would be useful at all requiring `Comonad`

ceedubs:
I'd also be interested in the motivation for this. I've read about Cont and ContT a few times, but I've never seen them used in the wild. As someone who doesn't have much intuition for them, in the use cases that people describe it seems to me like a pull- based stream API like fs2 would be a cleaner solution. But I'm definitely open to learning more about this approach. Do you have a current use-case for this?

johnynek:
I mostly did it to show we can get a safe tailRecM.

Whatever. I’ll let y’all close it if you don’t want it. Not here to try to change minds about it.

rossabaker:
@iravid has an interesting [proof of concept](https://gist.github.com/iravid/7c4b3d0bbd5a9de058bd7a5534073b4d) that shows an akka-http-like DSL expressed with `ContT`.

kailuowang:
@djspiewak is there anything else you would like to be addressed in this PR?

lemastero:
> I'd also be interested in the motivation for this. I've read about Cont and ContT a few times, but I've never seen them used in the wild. As someone who doesn't have much intuition for them, in the use cases that people describe it seems to me like a pull- based stream API like fs2 would be a cleaner solution. But I'm definitely open to learning more about this approach. Do you have a current use-case for this?

TLDR; there are many :)

People do mind blowing things with ContT. Usually they write interpreters, games, web servers, streaming libraries: 
* [snoyberg/conduit](https://github.com/snoyberg/conduit/blob/master/resourcet/Control/Monad/Trans/Resource/Internal.hs#L104-L105)
* byorgey/MonadRandom [Strict](https://github.com/byorgey/MonadRandom/blob/master/Control/Monad/Trans/Random/Strict.hs#L197-L198), [Lazy](https://github.com/byorgey/MonadRandom/blob/master/Control/Monad/Trans/Random/Lazy.hs#L196-L197)
* [mrkkrp/megaparsec](https://github.com/mrkkrp/megaparsec/blob/master/Text/Megaparsec/Internal.hs#L237-L241)
* [gitpan/Perl6-Pugs](https://github.com/gitpan/Perl6-Pugs/blob/master/src/Pugs/AST/Eval.hs#L27)
* [snapframework/heist](https://github.com/snapframework/heist/blob/master/src/Heist/Internal/Types/HeistState.hs#L543-L544)
* [simonmar/monad-par](https://github.com/simonmar/monad-par/blob/master/monad-par/Control/Monad/Par/Scheds/Direct.hs)
* [mvoidex/hsdev](https://github.com/mvoidex/hsdev/blob/master/src/System/Win32/FileMapping/Memory.hs)
* [paolino/reactivegas Server](https://github.com/paolino/reactivegas/blob/master/Lib/Server/Server.hs#L54), [Passo (1)](https://github.com/paolino/reactivegas/blob/master/Lib/Passo.hs#L67) [(2)](https://github.com/paolino/reactivegas/blob/master/Lib/Passo.hs#L100), [Interazione](https://github.com/paolino/reactivegas/blob/master/Lib/Interazione.hs#L31)
* [motemen/jusk](https://github.com/motemen/jusk/blob/master/src/Main.hs#L27)
* [aleino/thesis](https://bitbucket.org/aleino/thesis/src/2a6657136dc2525f8c2afcbf0e1fceb98b94892c/poga/Graphics/GraphicalGame.hs?at=master&fileviewer=file-view-default)
* [orbitz/web_typed](https://github.com/orbitz/web_typed/blob/master/libs/pa_monad/cc.ml)
* [exFalso/Sim](https://github.com/exFalso/Sim/blob/master/SimNode.hs#L189)
* [chris-taylor/Haskeme](https://github.com/chris-taylor/Haskeme/blob/master/src/Language/Types.hs#L64)
* [vpetro/heopl](https://bitbucket.org/vpetro/heopl/src/d717615d311e8d7d45594e1e6b1b20a5680a521d/continuation-examples.hs?at=default&fileviewer=file-view-default)
* [Rabbit: A Compiler for Scheme/Chapter 8 D. Conversion to Continuation-Passing Style](https://en.wikisource.org/wiki/Rabbit:_A_Compiler_for_Scheme/Chapter_8)

Last one is special :heart:  to me - That is how I got hooked by FP. Long before I learned Scala. When ContT lands in Cats, I will revisit my plan to write optimizing Lisp compiler using CPS in Scala :)

There are more sophisticated applications as well: [Is there a real-world applicability for the continuation monad outside of academic use?](https://stackoverflow.com/questions/41202721/is-there-a-real-world-applicability-for-the-continuation-monad-outside-of-academ#answer-41214580)

And it is good to have this implemented in library, with PR that passed review :eyes:  like this one. Otherwise everyone will need to write it by himself.

djspiewak:
@kailuowang Sorry missed this. Nothing more. I'm 👍

kailuowang:
Thanks @djspiewak! I am merging.

fommil:
a stack safe `ContT` in Scala is an achievement, well done @johnynek !

SemanticBeeng:
> There are more sophisticated applications as well: [Is there a real-world applicability for the continuation monad outside of academic use?]

Sorry if being lazy but how to best compare `ConT` monad and Scala `delimited continuations` ?
Asking in context of `Lantern` https://github.com/feiwang3311/Lantern.

PR #2509, opened 2018-09-17
Adding tests for binary compat that can't be tested by Mima 

kailuowang:
I am trying to add automated tests to test binary compatibility at places that we had to add mima exceptions. The idea is based off @rossabaker 's feedback [here](https://github.com/typelevel/cats/pull/2471#issuecomment-418810148). 

At the moment, I couldn't get the tests to work properly. It passes even when I introduced BC breaking changes. 

@rossabaker since you created a similar test before, would you help spot where I got things wrong here? 
Thanks so much!


build.sbt
```diff
@@ -216,10 +216,10 @@ lazy val docSettings = Seq(
   includeFilter in Jekyll := (includeFilter in makeSite).value
 )
 
-def mimaSettings(moduleName: String) = {
+def mimaPrevious(moduleName: String, scalaVer: String, ver: String): List[ModuleID] = {
```
kailuowang: Note for @rossabaker , no need to read these changes regarding change to mimaPreviousArtifacts change. For now I'm not using them and hard coded the two cats versions to test binary compat (1.2.0 and 1.3.0)

build.sbt
```diff
@@ -494,6 +502,25 @@ lazy val bench = project.dependsOn(macrosJVM, coreJVM, freeJVM, lawsJVM)
   ))
   .enablePlugins(JmhPlugin)
 
+
+lazy val binCompactTest = project
```
larsrh: Should this be `binCompatTest` (and also in the source folder)?

kailuowang: oops

PR #2518, opened 2018-09-19
improved support for subtyping in applicativeError syntax

mberndt123:
This is a case that I missed in my last pull request.
`new Exception("foo").raiseError[Either[Throwable, ?], Nothing]` now compiles; without this patch, you need to write `(new Exception("foo"): Throwable).raiseError[Either[Throwable, ?], Nothing]`

@LukaJCB , @kailuowang since you merged my last PR, you might want to take a look at this too.


kailuowang:
This change was already introduced by #2511 which is merged 9 hours ago. It's already in [master](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/syntax/applicativeError.scala#L70) so I am totally confused why I am seeing this diff (meaning the change is not on master) in this PR.

mberndt123:
I based my PR on an older master, I guess that must be why.

PR #2519, opened 2018-09-21
Use the Foldable, Luka!

kubukoz:
This PR makes sure that in instances for List we don't use the standard library's `foldRight` (which reverses the list and then folds left) but the faster `Foldable[List].foldRight` implementation.


kubukoz:
Renamed because I found yet another pun opportunity!

ceedubs:
@kubukoz I like the idea of this. Do you have benchmarks that show that this does indeed have the desired effect?

kubukoz:
I don't, it's based solely on my manual one-time measurements and the fact that stdlib's foldRight traverses the list twice (reverse and foldLeft)

kubukoz:
Totally agreed - I'll get some numbers then :)

kailuowang:
@kubukoz are you still interested in adding the benchmark? (maybe also solve the conflict? )

kubukoz:
I am... I'm sure I'll get back to it this week or so 😅

PR #2520, opened 2018-09-22
Add count as syntax to UnorderedFoldable

### LINKED ISSUE #2447: Add `count` method to Foldable
I'm thinking of this one here:

```scala
def count[A](p: A => Boolean): Long
```

We could also provide these specialized for the various collections we have :)

leusgalvan:
Fixes #2447 .

### DIFF
--- a/core/src/main/scala/cats/syntax/unorderedFoldable.scala
+++ b/core/src/main/scala/cats/syntax/unorderedFoldable.scala
@@ -15,19 +15,17 @@ final class UnorderedFoldableOps[F[_], A](val fa: F[A]) extends AnyVal {
     * For example:
     * {{{
     * scala> import cats.implicits._
-    * scala> val set1 = Set[String]()
-    * scala> set1.count_(_.length > 0)
+    * scala> val map1 = Map[Int, String]()
+    * scala> val p1: String => Boolean = _.length > 0
+    * scala> map1.count(p1)
     * res0: Long = 0
     *
-    * scala> val set2 = Set("hello", "world", "!")
-    * scala> set2.count_(_.length > 1)
+    * scala> val map2 = Map(1 -> "hello", 2 -> "world", 3 -> "!")
+    * scala> val p2: String => Boolean = _.length > 1
+    * scala> map2.count(p2)
     * res1: Long = 2
-    *
-    * scala> val set3 = Set(41, 32, 23)
-    * scala> set3.count_(_ % 2 == 0)
-    * res2: Long = 1
     * }}}
     */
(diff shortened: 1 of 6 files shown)


leusgalvan:
@LukaJCB only because when I tried to use it in a simple way, the `count` method from the standard library (which returns an `Int`) gets called instead of the cats one (which returns `Long`):

```
import cats._
import cats.implicits._
Set("hi", "world").count(_.length > 3)
res0: Int = 1
```

core/src/main/scala/cats/syntax/unorderedFoldable.scala
```diff
@@ -0,0 +1,33 @@
...
+    * scala> val set2 = Set("hello", "world", "!")
+    * scala> set2.count_(_.length > 1)
+    * res1: Long = 2
+    *
+    * scala> val set3 = Set(41, 32, 23)
+    * scala> set3.count_(_ % 2 == 0)
+    * res2: Long = 1
+    * }}}
+    */
+  def count_(p: A => Boolean)(implicit F: UnorderedFoldable[F]): Long =
```
kubukoz: Maybe it'd be better to have `count` defined in the `UnorderedFoldable` trait itself, and only call `count` here? Otherwise you can only use it with the syntax implicits, and not just `F.count`

LukaJCB: You can't do that if you want to maintain binary compat.

LukaJCB:
@leusgalvan I think I'd prefer to just call it `count`, we have a convention where an underscore after a name usually means it's some kind of operation that doesn't quite do what the same function without the underscore would do. 

See `filter_`, `takeWhile_`, `traverse_`, etc.

core/src/main/scala/cats/syntax/unorderedFoldable.scala
kubukoz: Aww damn :( the encoding of traits strikes again

kubukoz: we should be able to define extensions for typeclass instances, like:

```scala
implicit class UFOps[F[_]](val instance: UnorderedFoldable[F]) extends AnyVal {
  def count(...)
}
```

kubukoz:
@LukaJCB if it's called `count`, using it with a concrete type that has `count` defined on it will make it impossible to use - similar to `fold` and `combineAll`. 

Although it may be alright to keep it that way here, because there's a perfectly sane `count` on Lists and similar types that works the same as the one here...

leusgalvan:
@kubukoz @LukaJCB Ok let's go with `count`.

Issue #2479, opened 2018-09-07
Add doctests (excellent first-time issue)


kailuowang:
Doctests are a great way to provide compilable and testable example code for scaladoc. We should cover as many methods with doctests as possible. 
[Here is an example of PR adding doctests](https://github.com/typelevel/cats/pull/898/files).
Note that to test the doctests run `sbt coreJVM/test`
This issue encompasses the effort of adding more doctest coverage to our type classes and data types. 
It would be a great first-time contribution.

 If you are interested, please comment here on what type classes or data types you'd like to work on.

DennisVDB:
I'll be writing them for `Semigroup`, `Group`, and `Monoid`.

nasadorian:
@kailuowang So since Kleisli implements quite a few typeclasses, how much doctesting is too much? Should every method be covered?

kailuowang:
@nasadorian hmm, good question. I think I would focus on methods that are not part of type classes. There a couple of them in `KleisliFunctions`. Thanks!

hamednourhani:
Hi @kailuowang can i start for `Applicatives` ?

vasiliybondarenko:
@kailuowang can I start with `OptionT`?

LukaJCB:
@vasiliybondarenko go right ahead!

hamednourhani:
@kailuowang i checked Applicatives. That was done previously. I will work on EitherT

hamednourhani:
@kailuowang i will start `Validated`.

hamednourhani:
@kailuowang i checked `Ior` data type. that was interesting and good opportunity to learn. I will start `Ior`.

hamednourhani:
@kailuowang i checked the `Chain` data type as it used widely in cats and like to add more doctest to it to learn more.

muskan3006:
I'll pick Show

Gemini-3-Pro-AV:
e is the patch:
```diff
diff --git a/core/src/main/scala/cats/Foldable.scala b/core/src/main/scala/cats/Foldable.scala
index 70db2e7cd..5840ba85c 100644
--- a/core/src/main/scala/cats/Foldable.scala
+++ b/core/src/main/scala/cats/Foldable.scala
@@ -204,6 +204,16 @@ trait Foldable[F[_]] extends UnorderedFoldable[F] with FoldableNFunctions[F] { s
   /**
    * Find the minimum `A` item in this structure according to the `Order[A]`.
    *     *
[...]

PR #2523, opened 2018-09-22
Add doctests for Semigroup, Group, and Monoid

DennisVDB:
PR for part of #2479.

Adds doctests for the `Semigroup`, `Group`, and `Monoid` typeclasses.

### DIFF
--- a/kernel/src/main/scala/cats/kernel/Group.scala
+++ b/kernel/src/main/scala/cats/kernel/Group.scala
@@ -14,7 +14,7 @@ trait Group[@sp(Int, Long, Float, Double) A] extends Any with Monoid[A] {
    *
    * Example:
    * {{{
-   * scala> import cats.implicits._
+   * scala> import cats.kernel.instances.int._
    *
    * scala> Group[Int].inverse(5)
    * res0: Int = -5
@@ -29,7 +29,7 @@ trait Group[@sp(Int, Long, Float, Double) A] extends Any with Monoid[A] {
    *
    * Example:
    * {{{
-   * scala> import cats.implicits._
+   * scala> import cats.kernel.instances.int._
    *
    * scala> Group[Int].remove(5, 2)
    * res0: Int = 3
--- a/kernel/src/main/scala/cats/kernel/Group.scala
+++ b/kernel/src/main/scala/cats/kernel/Group.scala
@@ -11,6 +11,14 @@ trait Group[@sp(Int, Long, Float, Double) A] extends Any with Monoid[A] {
    * Find the inverse of `a`.
    *
    * `combine(a, inverse(a))` = `combine(inverse(a), a)` = `empty`.
+   *
+   * Example:
+   * {{{
+   * scala> import cats.implicits._
+   *
+   * scala> Group[Int].inverse(5)
+   * res0: Int = -5
+   * }}}
    */
   def inverse(a: A): A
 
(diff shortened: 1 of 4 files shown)


DennisVDB:
`sbt validate` is failing with:

```
error] /Users/dennis/Documents/Programming/scala/cats/kernel/.jvm/target/scala-2.12/src_managed/test/cats/kernel/GroupDoctest.scala:4:19: object scalacheck is not a member of package org
[error] import _root_.org.scalacheck.Prop._
[error]                   ^
[error] /Users/dennis/Documents/Programming/scala/cats/kernel/.jvm/target/scala-2.12/src_managed/test/cats/kernel/GroupDoctest.scala:7:24: object scalacheck is not a member of package org
[error]     extends _root_.org.scalacheck.Properties("Group.scala") {
[error]                        ^
[error] /Users/dennis/Documents/Programming/scala/cats/kernel/.jvm/target/scala-2.12/src_managed/test/cats/kernel/GroupDoctest.scala:7:46: no arguments allowed for nullary constructor Object: ()Object
[error]     extends _root_.org.scalacheck.Properties("Group.scala") {
[error]                                              ^
[error] /Users/dennis/Documents/Programming/scala/cats/kernel/.jvm/target/scala-2.12/src_managed/test/cats/kernel/GroupDoctest.scala:15:3: not found: value include
[error]   include(new _root_.org.scalacheck.Properties("Group.scala:10: inverse") {
...
```

kailuowang:
probably need to add this setting to the kernel settings
```
.settings(libraryDependencies += "org.scalacheck" %%% "scalacheck" % scalaCheckVersion(scalaVersion.value) % "test")

```

kernel/src/main/scala/cats/kernel/Monoid.scala
```diff
@@ -31,6 +64,17 @@ trait Monoid[@sp(Int, Long, Float, Double) A] extends Any with Semigroup[A] {
 
   /**
    * Given a sequence of `as`, sum them using the monoid and return the total.
+   *
+   * Example:
+   * {{{
+   * scala> import cats.implicits._
+   *
+   * scala> Monoid[String].combineAllOption(List("One ", "Two ", "Three"))
```
LukaJCB: Don't you mean `combineAll` here?

kailuowang:
I might missed something but I don't see build.sbt updated

DennisVDB:
You didn't miss it. 

I now pushed the updated `build.sbt` with the fix you suggested but it still fails the same way as before. It seems that the code generated for the doctest  doesn't compile.

The generated code: `/cats/kernel/.jvm/target/scala-2.12/src_managed/test/cats/kernel/*Doctest.scala`

Travis failure: https://travis-ci.org/typelevel/cats/jobs/432625136
Local failure: https://gist.github.com/DennisVDB/1fa934962f2f192af80731b1336adb1d

kernel/src/main/scala/cats/kernel/Group.scala
```diff
@@ -11,13 +11,29 @@ trait Group[@sp(Int, Long, Float, Double) A] extends Any with Monoid[A] {
    * Find the inverse of `a`.
    *
    * `combine(a, inverse(a))` = `combine(inverse(a), a)` = `empty`.
+   *
+   * Example:
+   * {{{
+   * scala> import cats.implicits._
```
larsrh: The problem is that `cats.implicits` is only defined in `cats-core`, not in `cats-kernel`, so importing it here won't work.

Issue #2486, opened 2018-09-12
Cats logo


kailuowang:
The current log of cats is cute but a little bit out of place in the typelevel project family. I don't know its history and our commitment to it. Will us be open to change it, maybe to a more modern design with geometry shapes, more conforming to the typelevel and scala theme? 

@thoefer designed one in #1807, see below: the one in the hexagon on the left.
![cats_friendly_big](https://user-images.githubusercontent.com/202410/44628798-b341b480-a947-11e8-966f-ce1e121d78ad.png)

 Several of us like it. What do you think?

kailuowang:
page @non @ceedubs @tpolecat @adelbertc @rossabaker @djspiewak @johnynek @milessabin @LukaJCB @mpilquist @fthomas and all cats contributors.

longcao:
I like the previous pixelated cat a bit more personally (in my head it's closely associated with the brand, so to speak), but I think this version could be made to work.

Maybe the original red from https://github.com/typelevel/cats/blob/f2850487732a185572b8a1752a24f13b875176d3/docs/src/main/resources/microsite/img/cats-logo.png? Keeping in mind that the original logo has a very slight gradient on it.

alexandru:
Just one more thought for why the color of a "Cats friendly" logo should be less strident ... because it's going to be shown near other logos that may not share a similar color palette. 

See for example [Monix](https://github.com/monix/monix#monix) which is OK near the "Typelevel red", because it's a cooler red, but would probably not be OK near this strident red:

![image](https://user-images.githubusercontent.com/11753/45432423-8c5ae080-b6b2-11e8-935f-ee1345908736.png)

rossabaker:
I like the Cats friendly graphic but I prefer the pixelated cat.  Even after realizing just now that the cat is asymmetrical.

Agree that the abstract cat could make a fine secondary logo.  favicon could be one  ideal context for it.

adelbertc:
I like the original pixelated cats, probably for most things except, like @non mentioned, in maybe like really small sizes or something where due to the pixellation style it won't work out.

kailuowang:
Looks like that we reached a consensus that we'd like to keep the original pixelated cat for logo. If we can create a smaller, i.e. more pixelated, version for smaller images it would be great too. @thoefer what do you think?

thoefer:
@non Do you have your pixelated cat as vector image somewhere?

PeterPerhac:
[...]

Issue #1807, opened 2017-08-09
a "works with cats" badge


kailuowang:
So that projects that integrate with cats can use.

DavidGregory084:
You can create custom badges on shields.io:

![custom badge](https://img.shields.io/badge/works%20with%20cats-1.0.0--MF-green.svg)

thoefer:
![versions](https://user-images.githubusercontent.com/202410/43990285-4b8301da-9d5a-11e8-82d5-6abb858b2dd6.png)

thoefer:
proposals, hints, suggestions?

milessabin:
:+1: for "Cats friendly".

johnynek:
👍 cats friendly, but I like using the hexagonal cat vs lambda, which I think is more distinctive (lambda being very generically used across FP).

BennyHill:
Cats friendly + hexagonal cats => :+1:

davegurnell:
+1 for hexagons and "cats friendly" :)

If people are interested I could speak to Jenny, the artist for the Scala with Cats book (http://underscore.io/books/scala-with-cats). She's more of a comic book artist and IMO what's needed here is pixel art, but I can have the conversation and see what she says.

thoefer:
@davegurnell Polishing an icon at the pixel level is what I usually do last. Right now the interpolation is somehow arbitrary and not optimized for being show at this small size. The end result is a more crisp icon, what you maybe expect. Nevertheless: I'd not consider that final icon to be "real" pixel art.

thoefer:
![versions2](https://user-images.githubusercontent.com/202410/44302850-8f6fe480-a331-11e8-8142-f227e2a7c5ed.png)

The design process flows (as usual; at least for me) from top left to right bottom. In the first half of everything I tried a pictographic cat, which I abandoned at some point. I simply tried a view concepts. In the final pieces on the right, especially bottom-right I tried to visually resolve the dualism between cat and hexagon by integrating the former into the last. Actually I like how it turned out.

thoefer:
![cats_friendly_big](https://user-images.githubusercontent.com/202410/44628798-b341b480-a947-11e8-966f-ce1e121d78ad.png)
![cats_friendly_normal](https://user-images.githubusercontent.com/202410/44628799-b341b480-a947-11e8-97a2-267b60707b6c.png)
[...]

PR #2524, opened 2018-09-23
cats logo as svg and new "cats friendly" badge

thoefer:
* add original, pixelated cats logo as svg image (#2486)

* add new "cats friendly" badge as svg image (#1807)

* add exported .png images of both also


thoefer:
TravisCI gave me a "No output has been received in the last 10m0s" on Scala. 2.13... not sure why that happens. I triggered a rebuild again.

kailuowang:
Thank so much! Once this is merge, we could update the README.md to provide it on the homepage.

rossabaker:
What's the recommended way for projects to embed these?  Copy them into our repos?  Solemnly swear that we'll never change the URL?

kailuowang:
I think copy them to their own repos might be easier. do we need to specify the license of this and the Cats logo? or MIT is going to cover this?

thoefer:
Embedding them with a link to our repo has the advantage that we can change it later and other have an always up-to-date version without any work on their side.

kailuowang:
Let's merge this and add links to the README

PR #2532, opened 2018-09-25
Add `allOf` to Eq

danslapman:
Motivation:
Composing of multiple Eq's with `and` quickly becomes awkward, Eq.allOf makes it more convenient.

### DIFF
--- a/kernel/src/main/scala/cats/kernel/Eq.scala
+++ b/kernel/src/main/scala/cats/kernel/Eq.scala
@@ -76,6 +76,14 @@ object Eq extends EqFunctions[Eq] with EqToEquivConversion {
       def eqv(x: A, y: A) = eq1.eqv(x, y) || eq2.eqv(x, y)
     }
 
+  /**
+    * Return an Eq that gives the result of conjunction of all arguments
+    * Note this is idempotent
+    */
+  def compose[@sp A](eqs: Eq[A]*): Eq[A] =
+    new Eq[A] {
+      def eqv(x: A, y: A) = eqs.forall(_.eqv(x, y))
+    }
 
   /**
    * Create an `Eq` instance from an `eqv` implementation.
--- a/kernel/src/main/scala/cats/kernel/Eq.scala
+++ b/kernel/src/main/scala/cats/kernel/Eq.scala
@@ -80,7 +80,7 @@ object Eq extends EqFunctions[Eq] with EqToEquivConversion {
     * Return an Eq that gives the result of conjunction of all arguments
     * Note this is idempotent
     */
-  def compose[@sp A](eqs: Eq[A]*): Eq[A] =
+  def allOf[@sp A](eqs: Eq[A]*): Eq[A] =
     new Eq[A] {
       def eqv(x: A, y: A) = eqs.forall(_.eqv(x, y))
     }
--- a/tests/src/test/scala/cats/tests/EqSuite.scala
+++ b/tests/src/test/scala/cats/tests/EqSuite.scala
@@ -17,4 +17,27 @@ class EqSuite extends CatsSuite {
     checkAll("Eq[Int]", ContravariantMonoidalTests[Eq].contravariantMonoidal[Int, Int, Int])
     checkAll("ContravariantMonoidal[Eq]", SerializableTests.serializable(ContravariantMonoidal[Eq]))
 
+    case class Something(a: Int, b: Int, c: Int)
+
+    test("Eq returned by Eq.compose yields true on all underlying Eq's true") {
+        val somethingEq = Eq.compose[Something](
+            Eq.by(_.a), Eq.by(_.b), Eq.by(_.c)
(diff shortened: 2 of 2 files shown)


danslapman:
Surely! Here they are

danslapman:
Hmm, can't figure out why JS build with 2.13.0-M4 failed...
`No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.`

kernel/src/main/scala/cats/kernel/Eq.scala
```diff
@@ -76,6 +76,14 @@ object Eq extends EqFunctions[Eq] with EqToEquivConversion {
       def eqv(x: A, y: A) = eq1.eqv(x, y) || eq2.eqv(x, y)
     }
 
+  /**
+    * Return an Eq that gives the result of conjunction of all arguments
+    * Note this is idempotent
+    */
+  def compose[@sp A](eqs: Eq[A]*): Eq[A] =
```
ceedubs: I think that it would be good to have a more descriptive name for this. Maybe `allOf`?

tests/src/test/scala/cats/tests/EqSuite.scala
```diff
@@ -17,4 +17,27 @@ class EqSuite extends CatsSuite {
     checkAll("Eq[Int]", ContravariantMonoidalTests[Eq].contravariantMonoidal[Int, Int, Int])
     checkAll("ContravariantMonoidal[Eq]", SerializableTests.serializable(ContravariantMonoidal[Eq]))
 
+    case class Something(a: Int, b: Int, c: Int)
+
+    test("Eq returned by Eq.compose yields true on all underlying Eq's true") {
```
ceedubs: Can we test the created instances against the `Eq` laws?

kernel/src/main/scala/cats/kernel/Eq.scala
danslapman: Done

tests/src/test/scala/cats/tests/EqSuite.scala
danslapman: Done

kernel/src/main/scala/cats/kernel/Eq.scala
```diff
@@ -76,6 +76,14 @@ object Eq extends EqFunctions[Eq] with EqToEquivConversion {
       def eqv(x: A, y: A) = eq1.eqv(x, y) || eq2.eqv(x, y)
     }
 
+  /**
+    * Return an Eq that gives the result of conjunction of all arguments
+    * Note this is idempotent
+    */
+  def allOf[@sp A](eqs: Eq[A]*): Eq[A] =
```
ceedubs: I think that you are just following what's done in similar methods here, but I question whether specialization is likely to help us here. It seems unlikely to me that someone is going to define multiple definitions of equivalence for a primitive type and what to check that all of them are satisfied. I suspect that this would only really serve to increase the cats jar size. WDYT @johnynek?

tests/src/test/scala/cats/tests/EqSuite.scala
```diff
@@ -17,4 +18,34 @@ class EqSuite extends CatsSuite {
...
+    )
+
+    implicit def arbitrarySomething(implicit i1: Arbitrary[Int], i2: Arbitrary[Int], i3: Arbitrary[Int]): Arbitrary[Something] =
+        Arbitrary(for {
+            a <- i1.arbitrary
+            b <- i2.arbitrary
+            c <- i3.arbitrary
+        } yield Something(a, b, c))
+
+    implicit val codenSomething: Cogen[Something] = Cogen(_.hashCode().toLong)
```
ceedubs: Typo coden

danslapman: Fixed

danslapman:
```
# There is insufficient memory for the Java Runtime Environment to continue.
# Native memory allocation (mmap) failed to map 415596544 bytes for committing reserved memory.
# An error report file with more information is saved as:
# /home/travis/build/typelevel/cats/hs_err_pid5557.log
```

kailuowang:
I restarted the build job

danslapman:
Same trouble

kailuowang:
I thought having multiple Eq instances of the same type is a rare case. What would be concrete use case?

danslapman:
Concrete use case is when You have about 20 fields in you case class, but want only few of then are significant for Eq.

kailuowang:
thanks @danslapman , in that case, would `Eq` by `Tuple` be even more convenient? 
```scala
implicit val fooEq: Eq[Foo] = Eq.by[Foo](f => (f.a, f.b, f.c))

```

danslapman:
Hmmm... May be

kailuowang:
closing since an alternative solution exists

PR #2533, opened 2018-09-25
Add empty to State and StateT objects

larsrh:
Work by @sderosiaux in #2294. Rebased to current master. For some reason, the bincompat error also went away (at least locally).

### DIFF
--- a/core/src/main/scala/cats/data/IndexedStateT.scala
+++ b/core/src/main/scala/cats/data/IndexedStateT.scala
@@ -203,7 +203,7 @@ private[data] trait CommonStateTConstructors {
     IndexedStateT(s => F.pure((s, s)))
 }
 
-object IndexedStateT extends IndexedStateTInstances with CommonStateTConstructors {
+object IndexedStateT extends IndexedStateTInstances with CommonStateTConstructors0 {
   def apply[F[_], SA, SB, A](f: SA => F[(SB, A)])(implicit F: Applicative[F]): IndexedStateT[F, SA, SB, A] =
     new IndexedStateT(F.pure(f))
 
@@ -221,6 +221,11 @@ object IndexedStateT extends IndexedStateTInstances with CommonStateTConstructor
 
   def setF[F[_], SA, SB](fsb: F[SB])(implicit F: Applicative[F]): IndexedStateT[F, SA, SB, Unit] =
     IndexedStateT(_ => F.map(fsb)(s => (s, ())))
+}
+
+private[data] trait CommonStateTConstructors0 extends CommonStateTConstructors {
+  def empty[F[_], S, A](implicit A: Monoid[A], F: Applicative[F]): IndexedStateT[F, S, S, A] =
+    pure(A.empty)
 }
 
 private[data] abstract class StateTFunctions extends CommonStateTConstructors {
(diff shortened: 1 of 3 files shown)


LukaJCB:
Seems like there's still a bincompat issue for 2.11

larsrh:
Ah, right. 2.11 doesn't like methods added to traits. But this should be fine with an exception, no?

LukaJCB:
We can add new methods by creating a new trait and mixing it in, similar to what we've been doing with the syntax additions :)

larsrh:
Smart. I've updated the commit accordingly.

larsrh:
And we're green!

core/src/main/scala/cats/data/IndexedStateT.scala
```diff
@@ -223,6 +223,11 @@ object IndexedStateT extends IndexedStateTInstances with CommonStateTConstructor
     IndexedStateT(_ => F.map(fsb)(s => (s, ())))
 }
 
+private[data] trait CommonStateTConstructors0 extends CommonStateTConstructors {
+  def empty[F[_], S, A: Monoid](implicit F: Applicative[F]): IndexedStateT[F, S, S, A] =
+    pure(Monoid[A].empty)
```
ceedubs: Is there a reason to use `: Monoid` for `A` but not use `: Applicative` for `F`? I'd be inclined to not use the syntactic sugar for either so you can avoid do something like `F.pure(A.empty)` and avoid an extra `Monoid.apply[A]` call.

larsrh:
@ceedubs Is this fine with you now?

core/src/main/scala/cats/data/IndexedStateT.scala
```diff
@@ -223,6 +223,11 @@ object IndexedStateT extends IndexedStateTInstances with CommonStateTConstructor
     IndexedStateT(_ => F.map(fsb)(s => (s, ())))
 }
 
+private[data] trait CommonStateTConstructors0 extends CommonStateTConstructors {
```
ceedubs: Is there any particular reason to introduce this new trait as opposed to just putting this method directly into the `object`?

core/src/main/scala/cats/data/IndexedStateT.scala
```diff
@@ -299,6 +304,11 @@ private[data] abstract class StateFunctions {
    */
   def pure[S, A](a: A): State[S, A] = State(s => (s, a))
 
+  /**
+    * Return `A`'s empty monoid value and maintain the input state.
+    */
+  def empty[S, A: Monoid]: State[S, A] = pure(Monoid[A].empty)
```
ceedubs: Should this use the same approach used above for getting the `empty` of the `Monoid`?

core/src/main/scala/cats/data/IndexedStateT.scala
larsrh: Binary compatibility: https://github.com/typelevel/cats/pull/2533#issuecomment-424601003

larsrh: To be precise, the function is not only provided on the `IndexedStateT` object, but also on `StateT`. This is the reason why it has to go into a trait in the first place.

larsrh:
@ceedubs Updated accordingly.

core/src/main/scala/cats/data/IndexedStateT.scala
ceedubs: Oh okay that makes sense. Thanks for the explanation.

PR #2534, opened 2018-09-25
avoid deprecated macro APIs post 2.10

larsrh:


core/src/main/scala/cats/arrow/FunctionK.scala
```diff
@@ -139,7 +139,7 @@ private[arrow] object FunctionKMacros extends MacroCompat {
 
     private[this] def punchHole(tpe: Type): Tree = tpe match {
```
larsrh: I would like to take a moment to appreciate the name of this method.

larsrh:
The removed trait is `private[cats]`. Is it allowed to add a bincompat exception there?

kailuowang:
It looks to me that this change should be bin compat (although I am not really sure since macro is involved). Regardless, I think it would be a good practice to add a test to verify that the intended end usage is still bin compat when we add a Mima exception. The PR to add such tests and related infrastructure is https://github.com/typelevel/cats/pull/2509 @larsrh would you take a look? Once that is merged, we can add one for `FunctionK.lift`. 

Another thing, maybe I should mention it first, is that @milessabin suggested that we don't really need macro here.  we can just do asInstanceOf thanks to errasure. In the following code I also made the example a doc test. 
```scala
  /**
    * Lifts function `f` of `F[A] => G[A]` into a `FunctionK[F, G]`.
    *
    * {{{
    * scala> import cats.arrow.FunctionK
    * scala> def headOption[A](list: List[A]): Option[A] = list.headOption
    * scala> val lifted: FunctionK[List, Option] = FunctionK.lift(headOption)
    * scala> lifted(List(1,2,3))
    * res0: Option[Int] = Some(1)
    * }}}
    *
    * Note: This method has a macro implementation that returns a new
    * `FunctionK` instance as follows:
    *
    * {{{
    *   new FunctionK[F, G] {
    *     def apply[A](fa: F[A]): G[A] = f(fa)
    *   }
    * }}}
    *
    * Additionally, the type parameters on `f` must not be specified.
    */
  def lift[F[_], G[_]](f: (F[α] ⇒ G[α]) forSome { type α }): FunctionK[F, G] =
    new FunctionK[F, G] {
      def apply[A](fa: F[A]): G[A] = f.asInstanceOf[F[A] => G[A]](fa)
    }
```

kailuowang:
cc @andyscott who added the macro to see if he has over concerns over the `asInstanceOf` implementation.

larsrh:
Here's a problem with this approach:

```scala
scala> import cats.arrow._
import cats.arrow._

scala> def sample[A](option: Option[A]): List[A] = option.toList
sample: [A](option: Option[A])List[A]

scala> FunctionK.lift(sample[String])
res0: cats.arrow.FunctionK[Option,List] = cats.arrow.FunctionK$$anon$5@53cd7d41
```

This should supposedly be ruled out.

kailuowang:
How did the macro version respond, if you have tried?

larsrh:
That's checked in the test suite with `assertTypeError`.

core/src/main/scala/cats/arrow/FunctionK.scala
```diff
@@ -124,7 +124,7 @@ private[arrow] object FunctionKMacros extends MacroCompat {
         val G = punchHole(evG.tpe)
 
         q"""
-        new FunctionK[$F, $G] {
+        new _root_.cats.arrow.FunctionK[$F, $G] {
```
larsrh: This is a hygiene issue that I fixed along the way.

build.sbt
```diff
@@ -516,10 +522,10 @@ lazy val binCompatTest = project
         else //We are not testing BC on Scala 2.13 yet.
           "org.typelevel" %% "cats-core" % version.value % Provided
       },
-      "org.typelevel" %% "cats-core" % version.value % Test,
       "org.scalatest" %%% "scalatest" % scalatestVersion(scalaVersion.value) % Test
     )
   )
+  .dependsOn(coreJVM % Test)
```
larsrh: I didn't realize that the approach in #2509 would require a `publishLocal` first. This here may be simpler.

larsrh:
@kailuowang This is still the macro solution, with bincompat tests & filters on top.

build.sbt
kailuowang: right. I was testing against existing jars.

core/src/main/scala/cats/arrow/FunctionK.scala
andyscott: 👍

andyscott:
I definitely am in favor of removing the macro implementation!

larsrh:
> I definitely am in favor of removing the macro implementation!

@andyscott Just to be clear, this PR doesn't do that, because the non-macro solution is unsafe.

LukaJCB:
I'm pretty clueless wrt macros, so I'm not sure I'm a good authority to review this PR, but I trust you two to have done the correct thing, so I think we can probably merge this :)

larsrh:
@LukaJCB I left the macro in place with minor changes, really nothing fancy happening here (@kailuowang did all the work on BC) :smile:

larsrh:
For posterity: https://github.com/typelevel/cats/issues/2553#issuecomment-427446985

Issue #2346, opened 2018-07-28
Add NEM[K, V] => NEL[(K, V)] and vice versa, NEL#toNes


kubukoz:

denisrosca:
Wouldn't `NEL[(K, V)]` have to change to `NEM[K, NEL[V]]`? Or should the transformation require `V: Semigroup`?

kubukoz:
I think that would resemble `groupBy` or something like that.

What I meant in the issue was equivalents to stdlib's `Map#toList` and `List#toMap` :)

Also, adding `NEL#toNes` to the issue as I've had a few usecases for it recently too.

CucumisSativus:
Guys, I'd like to work on this if its still relevant :)

LukaJCB:
@CucumisSativus Go right ahead :)

calvinbrown085:
@kubukoz @CucumisSativus Can we close this? Looks like the PR has been merged

CucumisSativus:
@calvinbrown085 I split it into 3 PR to make reviews easier. One is merged. The second is awaiting second confirmation https://github.com/typelevel/cats/pull/2541 and my third is waiting for the second to be merged :)

calvinbrown085:
Ah fair enough :)

CucumisSativus:
Okay, issue can be closed. All pull requests are merged

kubukoz:
Closing ;) thanks!

PR #2535, opened 2018-09-25
Add NonEmptyMap#toNonEmptyList (issue #2346)

CucumisSativus:
Hello, this is my first pull request there so please go easy with me :)

As for #2346 started with small steps - implementing just first method. Frankly i have no idea if I placed this in proper place, no idea if my test is sufficient and no idea if the thing I implemented is performant and if my git message is sufficient. Would love any feedback from you. Then after initial feedback I will do my best to implement rest of the methods from that issue

### DIFF
--- a/tests/src/test/scala/cats/tests/NonEmptyMapSuite.scala
+++ b/tests/src/test/scala/cats/tests/NonEmptyMapSuite.scala
@@ -201,4 +201,10 @@ class NonEmptyMapSuite extends CatsSuite {
       nem.length should ===(nem.toSortedMap.size)
     }
   }
+
+  test("NonEmptyMap#toNonEmptyList is consistent with Map#toList and creating NonEmptyList from it"){
+    forAll{ nem: NonEmptyMap[String, Int] =>
+      nem.toNel should ===(NonEmptyList.fromListUnsafe(nem.toSortedMap.toList))
+    }
+  }
 }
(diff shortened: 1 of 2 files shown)


tests/src/test/scala/cats/tests/NonEmptyMapSuite.scala
```diff
@@ -201,4 +201,10 @@ class NonEmptyMapSuite extends CatsSuite {
       nem.length should ===(nem.toSortedMap.size)
     }
   }
+
+  test("NonEmptyMap#toNonEmptyList is consistent with Map#toList and creating NonEmptyList from it"){
+    forAll{ nem: NonEmptyMap[String, Int] =>
+      nem.toNonEmptyList should ===(NonEmptyList.fromListUnsafe(nem.toSortedMap.toList))
```
LukaJCB: Unfortunately the `toNonEmptyList` of `NonEmptyMap` collides with `toNonEmptyList` of `Reducible`, so the `nem.toNonEmptyList` here seems to return a `NonEmptyList[Int]` instead of the `NonEmptyList[(String, Int)]` that we want. 
Maybe we should rename the method to `toNel` instead? :)

CucumisSativus: Worked, thanks a lot!

PR #2536, opened 2018-09-25
Introduce `OrElse` type class

danslapman:
This PR introduces `OrElse` typeclass into alleycats. `OrElse` is a generalization of the idea of getting of "alternative" value instead of inappropriate contents of a container. The original problem was abstracting from throwing `NoSuchElementException` by the `filter` method of `Future` when I worked on my https://github.com/danslapman/asyncstreams library

### DIFF
--- a/alleycats-core/src/main/scala/alleycats/OrElse.scala
+++ b/alleycats-core/src/main/scala/alleycats/OrElse.scala
@@ -1,0 +1,8 @@
+package alleycats
+
+import simulacrum.op
+import simulacrum.typeclass
+
+@typeclass trait OrElse[F[_]] {
+  @op("orElse") def orElse[A](fa: F[A], alternative: => F[A]): F[A]
+}
--- a/alleycats-core/src/main/scala/alleycats/std/future.scala
+++ b/alleycats-core/src/main/scala/alleycats/std/future.scala
@@ -3,7 +3,7 @@ package std
 
 import export._
 
-import scala.concurrent.Future
+import scala.concurrent.{ExecutionContext, Future}
 
 @reexports(FutureInstances)
 object future
@@ -13,5 +13,12 @@ object FutureInstances {
   implicit val exportFuturePure: Pure[Future] =
     new Pure[Future] {
       override def pure[A](a: A): Future[A] = Future.successful(a)
+    }
+
+  @export(Orphan)
+  implicit def exportFutureOrElse(implicit ec: ExecutionContext): OrElse[Future] =
+    new OrElse[Future] {
+      override def orElse[A](fa: Future[A], alternative: => Future[A]): Future[A] =
+        fa.recoverWith { case _: NoSuchElementException => alternative }
     }
 }
--- a/alleycats-core/src/main/scala/alleycats/std/option.scala
+++ b/alleycats-core/src/main/scala/alleycats/std/option.scala
@@ -13,4 +13,11 @@ object OptionInstances {
     new EmptyK[Option] {
       def empty[A]: Option[A] = None
     }
+
+  @export(Orphan)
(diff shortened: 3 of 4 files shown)


danslapman:
Future's empty is `NoSuchElementException` by design in scala library (https://github.com/scala/scala/blob/2.13.x/src/library/scala/concurrent/Future.scala#L286).
Example from asyncstreams' readme clearly shows the problem:

```scala
val stream = genS(0) {
      for {
        s <- getS[Int]
        if s < 3
        _ <- putS(s + 1)
      } yield s
    }

stream.to[List].map(_ shouldBe (0 :: 1 :: 2 :: Nil))
```
`if` desugars to `filter` which throws NoSuchElementException

As for Try I followed the behavior defined in standart library.

I'm not sure that any laws can be defined for OrElse (that's why I propose it for alleycats).

alleycats-core/src/main/scala/alleycats/OrElse.scala
```diff
@@ -0,0 +1,8 @@
+package alleycats
+
+import simulacrum.op
+import simulacrum.typeclass
+
+@typeclass trait OrElse[F[_]] {
+  @op("orElse") def orElse[A](fa: F[A], alternative: => F[A]): F[A]
```
kubukoz: I think `@op def ...` will work the same :)

danslapman: Truly!

kubukoz: hmm, interesting. I was pretty sure it had at least an overload that doesn't take a name. Sorry :)

danslapman:
Have no idea, how to rebuild...

kubukoz:
If you can't control the travis restarts, maybe make an empty commit with
--allow-empty?

danslapman:
It seems that close/re-open triggered re-build

kailuowang:
What would be the practical use of this over `ApplicativeError`?

danslapman:
Again, the abstraction `OrElse` provides is different: `OrElse` says, that some states of `F` are considered incorrect and then alternative value is used. We can't raise an arbitrary error on Option, but we have orElse defined at Option naturally. `OrElse` comes for convenience when dealing with existing code, maybe lawless, but working, we can't always replace `s.c.Future` or twitter's `Future` with something lawful.

kubukoz:
Wouldn't MonadError[Option, Unit] suffice for that?

danslapman:
The probles is when You want to abstract over a "incorrect", You can't use MonadError as a context bound:
You need
- MonadError[Future, NoSuchElementException] for scala Future
- MonadError[Future, PredicateDoesNotObtain] for twitter Future
- MonadError[Try, Throwable] for Try
- etc
The type of error needed to be catched varies from type to type, and `OrElse` incapsulates that type. 
For a real-world practical applications please take a look at https://github.com/danslapman/asyncstreams/blob/master/core/src/main/scala/asyncstreams/AsyncStream.scala#L21-L26 or https://github.com/danslapman/asyncstreams/blob/master/core/src/main/scala/asyncstreams/AsyncStream.scala#L87-L92
Alse, here is the test, from which my previous example came from: https://github.com/danslapman/asyncstreams/blob/master/core/src/test/scala/asyncstreams/AsyncStreamMonadSyntaxTests.scala#L69-L81

kag0:
I have no experience writing laws for cats, but it seems we can create at least a few rules for `OrElse`
abstractly some might be 
if orElse(a, b) = b 
  then orElse(a, c) = c
    and if orElse(b, c) = c
      then orElse(b, a) = a
      else orElse(b, a) = b

Issue #2517, opened 2018-09-19
Add FiniteDuration instances


LukaJCB:
We currently have instances for `Duration` only, which means that you can use `FiniteDuration` with it, but must upcast it, which can be annoying:

```scala
3.minutes |+| 3.minutes // Fails to compile because `|+|` is not a member of `FiniteDuration`.
```

satansk:
I will do it :)

adelbertc:
I use `FiniteDuration`s all the time, `Duration`s not so much 💯

calvinbrown085:
@satansk If you're still working on this feel free to keep going :) Can we just reuse https://github.com/typelevel/cats/blob/master/kernel/src/main/scala/cats/kernel/instances/duration.scala

With `FiniteDuration` instead of `Duration?

Where do these instances go? @codecentric

Thanks!

LukaJCB:
Yes the instances should be exactly the same :) and you should be able to place them analogously to where the duration instances are placed

calvinbrown085:
I opened a PR here, @satansk if you wanted to work on this i'm more than happy to close it!

satansk:
@calvinbrown085 go ahead :)

I'm busy for a release in company, and I was planning to work on this on my 7 days vocation one day later. Since you already have made a PR, just keep on working on it.

calvinbrown085:
@satansk Sounds good! Just making sure I don't trample on anyone :)

PR #2544, opened 2018-09-29
Adding FiniteDuration instances

calvinbrown085:
closes https://github.com/typelevel/cats/issues/2517


calvinbrown085:
@LukaJCB Yes! I'll get those added

LukaJCB:
I'm sorry I think worded myself incorrectly, what I meant was, that we should re-export the instances in `cats.instances` and `cats.implicits` not duplicate them. :)

You can look at other instances in kernel to see how it's done.

calvinbrown085:
@LukaJCB There we go, this should look a little better.

calvinbrown085:
@LukaJCB Thank you, I got those added

calvinbrown085:
Anyone more familiar with this repo than me, why is the CodeCov pipeline failing?

Am I missing some tests?

LukaJCB:
@calvinbrown085 You're not testing any of the instances for their laws. Check the `LawTests` class to see how it's done for other types.

kernel/src/main/scala/cats/kernel/instances/all.scala
```diff
@@ -14,6 +14,7 @@ trait AllInstances
     with EqInstances
     with EitherInstances
     with DurationInstances
+    with FiniteDurationInstances
```
LukaJCB: This is binary incompatible, you should create another `AllInstancesBinCompat0` trait here, and then mix that into `all`. :)

calvinbrown085: Ah, that might be what I am missing :) Thank you

LukaJCB: Of course :)

PR #2546, opened 2018-09-30
Implement scalafmt

### LINKED ISSUE #1206: Consider adding scalafmt

easel:
Resolves #1206 and replaces #1731


easel:
I don't think the codecov/patch failure is relevant. There's perhaps some additions to the CONTRIBUTING.md or something that should go along with this. Any thoughts?

.scalafmt.conf
```diff
@@ -0,0 +1 @@
+// default settings
```
LukaJCB: Okay, someone has to start the bikeshedding, so here goes:

I'd like to see
```
align=none
maxColumn = 120
```

The rest of the default settings I think are good :)

easel:
I've pushed another commit applying @LukaJCB's bikeshed suggestions. Both seem reasonable. I like alignment, but for a larger project it will create diff noise that may be less than desirable. 120 columns is a reasonable max in the modern world, although I'd say the "style guide" should recommend 80.

ceedubs:
Awesome, @easel! Thanks a bunch.

This looks good to me, but... how is it such a small changeset? I would expect it to touch a ton of lines. Is it actually running on all of the modules?

ceedubs:
The only changes are in the `jvm`/`js` projects which makes me think that maybe our JVM/JS cross-building setup is throwing off scalafmt.

easel:
@ceedubs Thats an interesting possibility, the cats cross build stuff is definitely beyond the basics. From my perspective if we can bikeshed out the settings it would be best to just get this merged and then work on applying it to any missing submodules as we can. That way the settings are documented and people can use them on their own.

.scalafmt.conf
```diff
@@ -0,0 +1,3 @@
...
```
joan38: Why this comment?

kailuowang:
shall we close this since #2562 is merged.

PR #2550, opened 2018-10-03
Proposal: add `cats-collections` as an external cats module

kailuowang:
@stew generously [donated dogs to become typelevel/cats-collections](https://github.com/typelevel/cats-collections/issues/104). @larsrh completed the migration and started actively maintaining it. @LukaJCB, @ChristopherDavenport and I also signed up for help maintaining it. 
We'd like to propose to the Cats community to make cats-collections an external Cats module (separate repo and release cycle). 
The main implication for being a Cats external module is that at least some Cats maintainers will help maintaining it, which is the case now.


kailuowang:
Looks like we got enough approvals. It's official then.

PR #2552, opened 2018-10-04
Optimize the foldM implementation for Vector

ceedubs:
I've also added a benchmark to compare the new implementation with the
old. Here are the results that I got on my computer when running
`bench/jmh:run -i 10 -wi 10 -f1 -t1 .*FoldBench\.vector.*`:

```
Benchmark                     Mode  Cnt        Score       Error  Units
FoldBench.vectorIndexFoldM   thrpt   10  1185138.829 ± 45633.259  ops/s
FoldBench.vectorToListFoldM  thrpt   10   817330.438 ± 21998.851  ops/s
```


LukaJCB:
Nice work, thanks @ceedubs! :)

PR #2557, opened 2018-10-14
Add NonEmptyList#toNes (issue #2346)

CucumisSativus:
The last missing pull request for #2346


hepin1989:
Why.not toSet or toNonEmptySet

satansk:
@hepin1989 Beacuse there are already some codes where `toNes` refers to `toNonEmptySet`. In fact, this is confusing sometimes, beacuse it can be interperted as `toNonEmptyStream` too.

CucumisSativus:
@hepin1989 Everything that @satansk + the fact I wanted to be consistent with the rest of naming from this issue where I implemented NonEmptyMap#toNel and NonEmptyList#toNem. I think that `.toSet` might because it does not mention that this set should be non empty. 

If you guys think that this naming is misleading, i'm not really attached to it.

CucumisSativus:
Sorry for the mess in commits, screwed a bit with merging and then everything escalated quickly.

CucumisSativus:
The build failed because of `Java HotSpot(TM) 64-Bit Server VM warning: INFO: os::commit_memory(0x000000078ce4e000, 857415680, 0) failed; error='Cannot allocate memory' (errno=12)`. Could someone restart it?

PR #2560, opened 2018-10-15
Add ApplicativeError.redeem and MonadError.redeemWith

joan38:


core/src/main/scala/cats/ApplicativeError.scala
```diff
@@ -72,6 +72,26 @@ trait ApplicativeError[F[_], E] extends Applicative[F] {
...
+
+  /**
+    * Returns a new value that transforms the result of the source,
+    * given the `recover` or `bind` functions, which get executed depending
+    * on whether the result is successful or if it ends in error.
+    *
+    * This is an optimization on usage of [[attempt]] and [[flatMap]].
+    */
+  def redeemWith[A, B](fa: F[A])(recover: E => F[B], bind: A => F[B]): F[B] =
+    flatMap(attempt(fa))(_.fold(recover, bind)))
```
joan38: `flatMap` doesn't exist here so what do you think I should do?

kailuowang:
I might've missed something. Is this a shortcut for `fa.map(f).handleError(ef)`?

joan38:
@kailuowang Well it's supposed to be in the type class but because of 2.11 bin incompat it's juts in the ops.
This is to match the `IO.redeem`.

kailuowang:
Thanks, that clarified the intention for me, I think. The issue is that if we can't add it to type class then we can't have instances to override it, then there isn't much point of adding it before 2.0 with which we shall be able to add it to type class without breaking BC.

joan38:
@kailuowang It probably make sense.
How do we handle stuff that will go in 2.0? Do we keep breaking PRs like this open?

kailuowang:
yeah for cases like this , we can add it to the type class, let the mima test fail but schedule it to milesstone 2.0.

diesalbla:
This PR seems to be doing a very similar thing to @alexandru PR here https://github.com/typelevel/cats/pull/2237.

joan38:
@diesalbla Thanks I'm closing this one

PR #2562, opened 2018-10-17
Introduce Scalafmt

joan38:
This is with mostly the default settings.

I removed some overlapping scalastyle settings.
I raised the scalastyle method line limit from 50 to 55, because I don't see how to decompose those functions like in `NonEmptyTraverseTests.nonEmptyTraverse`.

The important files are:
- `project/plugins.sbt`
- `build.sbt`
- `.scalafmt`
- `scalastyle-config.xml`

The rest is applied formatting.

Let me know your thoughts.


alleycats-laws/src/main/scala/alleycats/laws/discipline/FlatMapRecTests.scala
```diff
@@ -7,20 +7,16 @@ import org.scalacheck.Arbitrary
...
-                            ): RuleSet = {
-    new DefaultRuleSet(
-      name = "flatMapTailRec",
-      parent = None,
-      "tailRecM consistent flatMap" -> forAll(laws.tailRecMConsistentFlatMap[A] _))
-  }
+                             EqFA: Eq[F[A]]): RuleSet =
+    new DefaultRuleSet(name = "flatMapTailRec",
+                       parent = None,
+                       "tailRecM consistent flatMap" -> forAll(laws.tailRecMConsistentFlatMap[A] _))
```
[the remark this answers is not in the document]
joan38: It lets you do this if you'd like, you just need to add the returns. I'm not going to go though all of them tho.

kailuowang:
so is scalafmt applied on every compile?

joan38:
@kailuowang No, you need to run `sbt fmt`

ikhoon:
@joan38 I saw some people use vscode, vim and other editors. [neo-sbt-scalafmt](https://github.com/lucidsoftware/neo-sbt-scalafmt) automatically reformat code with [some configuration](https://github.com/lucidsoftware/neo-sbt-scalafmt/blob/master/README.md#task-configuration] on every compile.

kailuowang:
Sorry I missed the previous discussion on this, what did we decide to do IRT enforcing this? add to commit hook? or is there a way to run scalafmt only against changed file (using git)?

build.sbt
```diff
@@ -207,14 +206,19 @@ lazy val docSettings = Seq(
...
-    "-Ymacro-annotations"
-  )),
+  ) ++ (if (priorTo2_13(scalaVersion.value))
+          Seq(
+            "-Yno-adapted-args",
+          )
+        else
+          Seq(
+            "-Ymacro-annotations"
+          )),
```
ceedubs: This seems a bit excessive with respect to whitespace to me, but I don't know if this is something that's easily configurable?

build.sbt
```diff
@@ -636,31 +684,23 @@ lazy val publishSettings = Seq(
...
-
+addCommandAlias("validateJVM", ";buildJVM;bench/test;validateBC;makeMicrosite")
 addCommandAlias("validateJS", ";catsJS/compile;testsJS/test;js/test")
-
 addCommandAlias("validateKernelJS", "kernelLawsJS/test")
-
 addCommandAlias("validateFreeJS", "freeJS/test") //separated due to memory constraint on travis
-
-addCommandAlias("validate", ";clean;validateJS;validateKernelJS;validateFreeJS;validateJVM")
+addCommandAlias("validate", ";clean;scalastyle;fmtCheck;validateJS;validateKernelJS;validateFreeJS;validateJVM")
```
ceedubs: I think that it might be better to move the scalastyle and format check back to `validateJVM`. While it's not JVM-specific, our build runs `validateJVM` or `validateJS` depending on whether it's a JVM or JS build (combining both in a single build leads to more memory issues). So to make sure that this is run, I think that we need to choose one for it to run in, and `validateJVM` is what we were using.

Our builds probably need a bit of an overhaul, but I don't think that we should try to tackle that in a PR that has this big of a changeset, or we'll never get either merged :)

core/src/main/scala/cats/Alternative.scala
```diff
@@ -24,35 +25,35 @@ import simulacrum.typeclass
...
-   * {{{
-   * scala> import cats.implicits._
-   * scala> def even(i: Int): Option[String] = Alternative[Option].guard(i % 2 == 0).as("even")
-   * scala> even(2)
-   * res0: Option[String] = Some(even)
-   * scala> even(3)
-   * res1: Option[String] = None
-   * }}}
-   */
+    * Return ().pure[F] if `condition` is true, `empty` otherwise
```
ceedubs: While we probably haven't been entirely consistent, I think that we've been using javadoc-style comments, so I think that `docstrings = JavaDoc` might be a good setting for us. Unless I'm wrong and that actually generates a bigger diff...

build.sbt
joan38: 👍

core/src/main/scala/cats/Alternative.scala
joan38: 👍

joan38:
@ikhoon not only neo-sbt-scalafmt can do it.
Lets decide what you guys want as @kailuowang questionned. Currently you need to run `fmt`.

build.sbt
joan38: It just wants the closing bracket to be on the same indent as the opening one or on a single line.
We just need to put it on a single line the thing. Tho I'm not going to do it for all of them.

core/src/main/scala/cats/Alternative.scala
joan38: The diff is much smaller now :)

build.sbt
joan38: I reverted this

ceedubs:
@kailuowang my vote would be to run the scalafmt _check_ as part of `validateJVM` and have people explicitly call `fmt` (or whatever) to apply formatting. Doing it on every compile could potentially slow things down or get in the way. But I'm not set in stone on that.

This looks good to me. My only concern is that some builds have been timing out, but I don't know how much that's exacerbated by this PR. I'm checking out a few things locally, but I think that I'd be good with this being merged in its current form.

ceedubs:
@kailuowang yeah, I can see that concern. It's possible that running format on compile would be just fine; I'm not sure. And maybe we could switch things around a bit so `validate` is something that we would expect people to run before pushing and we'd have a new `validateAll` that did both JVM and JS (like the current `validate` does).

BUT since this PR has a huge changeset that's likely to cause merge conflicts, I'm inclined to get this merged and figure that out afterwards.

joan38:
Thanks guys for your comments. Let me if I can do anything.

kailuowang:
follow up PR https://github.com/typelevel/cats/pull/2567

PR #2563, opened 2018-10-18
Add an `applyK` function to `Kleisli`.

diesalbla:
We add an `applyK` function to the `Kleisli` companion object. This takes a parameter of type `A`, and generates a natural transformation (`FunctionK`) from `Kleisli[F,A,?]` to `F`, i.e., it maps over the Kleisli object by applying it to the given parameter.

An alternative name for this function could be `dipK`, or something that conveys the idea of it being contrary to the `liftK` function.

### DIFF
--- a/core/src/main/scala/cats/data/Kleisli.scala
+++ b/core/src/main/scala/cats/data/Kleisli.scala
@@ -113,6 +113,33 @@ object Kleisli
       case _ =>
         Kleisli(run)
     }
+
+  /**
+   * Creates a `FunctionK` that transforms a `Kleisli[F, A, B]` into an `F[B]` by applying the value of type `a:A`.
+   * {{{
+   * scala> import cats.{~>}, cats.data.{Kleisli, EitherT}
+   *
+   * scala> def f(i: Int): Option[Either[Char, Char]] = if (i > 0) Some(Right('n')) else if (i < 0) Some(Left('z')) else None
+   *
+   * scala> type KOI[A] = Kleisli[Option, Int, A]
+   * scala> val b: KOI[Either[Char, Char]] = Kleisli[Option, Int, Either[Char, Char]](f _)
+   * scala> val nt: Kleisli[Option, Int, ?] ~> Option = Kleisli.applyK[Option, Int](1)
+   * scala> nt(b)
+   * res0: Option[Either[Char, Char]] = Some(Right('n'))
+   *
+   * scala> type EKOIC[A] = EitherT[KOI, Char, A]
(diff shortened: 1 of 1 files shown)


core/src/main/scala/cats/data/Kleisli.scala
```diff
@@ -131,6 +131,20 @@ private[data] sealed trait KleisliFunctions {
   def liftK[F[_], A]: F ~> Kleisli[F, A, ?] =
     λ[F ~> Kleisli[F, A, ?]](Kleisli.liftF(_))
 
+  /**
+    * Same as [[apply]], but expressed as a FunctionK for use with mapK
+    * {{{
+    * scala> import cats._, data._, implicits._
+    * scala> val a: Int = 1
+    * scala> def f(a: Int): Option[Int] = if (a > 0) Some('y') else None
```
kailuowang: Option[Int] -> Option[String]

core/src/main/scala/cats/data/Kleisli.scala
```diff
@@ -131,6 +131,19 @@ private[data] sealed trait KleisliFunctions {
   def liftK[F[_], A]: F ~> Kleisli[F, A, ?] =
     λ[F ~> Kleisli[F, A, ?]](Kleisli.liftF(_))
 
+  /**
+    * Same as [[apply]], but expressed as a FunctionK for use with mapK
+    * {{{
+    * scala> import cats._, data._, implicits._
+    * scala> def f(a: Int): Option[Char] = if (a > 0) Some('y') else None
+    * scala> val b: Kleisli[Option, Int, Char] = Kleisli[Option, Int, Char](f _)
+    * scala> b.mapK( Kleisli.applyK(1) )
```
peterneyens: The `mapK` on `Kleisli` takes an `F ~> G` to transform an `Kleisli[F, A, B]` into `Kleisli[G, A, B]`.

It seems you are trying to use the `mapK` originating from cats-tagless `FunctorK`.

kailuowang:
@diesalbla are you still interested in fixing the doctest?

diesalbla:
@kailuowang I'll take a look later today.

diesalbla:
@kailuowang Fixed the doctests, and used `mapK` within the `EitherT` transformer.

kailuowang:
Oh we can't add it the trait without breaking Binary compatibility, make just add it to the companion

diesalbla:
Did you mean the `Kleisli` companion object?

kailuowang:
Yup

Issue #2530, opened 2018-09-25
Add `Tuple2` as example of `Bifunctor`


kailuowang:
See comment in #2459 https://github.com/typelevel/cats/pull/2459#pullrequestreview-151860750

PR #2566, opened 2018-10-18
Add Tuple2 Example for Bifunctor

stephen-lazaro:
The example is certainly a bit contrived, but I do think it gets the general point across. Very open to feedback.
Intended to resolve https://github.com/typelevel/cats/issues/2530


stephen-lazaro:
Sure! I'll simplify and try to get something less complicated.

stephen-lazaro:
@ceedubs How's that?

PR #2567, opened 2018-10-18
Make scalafmt reformat changed files on compile. 

kailuowang:
Added a temporary solution to reformat all changed files on compile. 
It's needed because because sbt-scalafmt's formatOnCompile option reformats all files on every compile. 
So my solution is to run a `scalafmt --diff-branch HEAD` on every compile which is a roughly 5 second overhead on `coreJVM` alone. 
It's not ideal because it reformats all files that are changed from git HEAD on compile, but it's mostly manageable 1-2 seconds overhead.


build.sbt
```diff
@@ -3,15 +3,31 @@ import ReleaseTransformations._
 import scala.xml.transform.{RewriteRule, RuleTransformer}
 import sbtcrossproject.CrossProject
 import sbtcrossproject.CrossPlugin.autoImport.{crossProject, CrossType}
-
```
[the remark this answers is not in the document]
kailuowang: mistake

.scalafmt.conf
```diff
@@ -4,3 +4,5 @@ assumeStandardLibraryStripMargin = true
 danglingParentheses = true
 rewrite.rules = [AvoidInfix, SortImports, RedundantBraces, RedundantParens, SortModifiers]
 docstrings = JavaDoc
+align.openParenCallSite = true
+align.openParenDefnSite = true
```
ceedubs: Can we split the formatting changes into a separate PR? There are two separate changes that both could be controversial in this PR and it might be simpler to not have the reviews/comments of the two intertwined.

ceedubs:
@kailuowang you mention that this is temporary -- could you please elaborate on that? Is there a PR/issue tracking this on the sbt-scalafmt side?

Personally I'm a bit hesitant with auto-formatting on compile. My usual development flow is vim in one pane editing code and `~compile` in another, so I think that this will prompt vim to keep complaining that the file has changed outside of my edits. But I'm not sure it's been a while since I've been on a project that was set up to do this.

.scalafmt.conf
kailuowang: Sorry I should've noted that these two settings were default to be `true` in scalafmt 1.5, which is the version used before this PR. In scalafmt 1.6 the defaults were changed to `false`. So I added them here to avoid more changes than necessary to this PR. If we want to change to sue 1.6's default, we can open a new PR to remove these two settings.

joan38: Damn my bad I tried to anticipate 1.6 by adding `danglingParentheses = true` but I missed those ones.
So you can remove `danglingParentheses = true` then since in 1.6 it's the default.

kailuowang:
@ceedubs it's temporary because there is a setting in sbt-scalafmt called formatOnCompile, which means it can reformat on compile as well, but it checks all files so it adds more overhead. I just created an issue about it. https://github.com/scalameta/sbt-scalafmt/issues/9

The root problem is that any manual step hard to communicate and enforce. So IMO we can't afford to have a manual reformat step. People would submit create a PR and couple of hours later found out format  check failed (it's almost bound to happen now if you forget to run scalafmt that scalafmt is a lot more strict) 

There are two potential ways for automating this. 
1. reformat on compile. 
2. reformat on commit. 

I thought reformat on compile is okay but now you pointed out, breaking VIM workflow is probably a deal-breaker
Timing wise, reformat on commit with a git pre-commit hook makes a lot more sense. The downside is that we can't automate the addition of the git pre-commit hook, contributors will have to do a one-time manual step to add that pre-commit hook (we can certainly provide script to do that. ). Now that's probably an acceptable cost for contributors, but I struggled to think of a way to communicate that one-time manual step to new contributors. We can add it to contribution.md, but I suspect many people read that through, maybe we just add it to README. 

What do you think?

.scalafmt.conf
ceedubs: @kailuowang ah okay I see that makes sense. 👍

ceedubs:
@kailuowang I'm inclined to think that a pre-commit hook isn't a great solution, since you can't really automate it. I'll pull your changes and play around a bit and see how it goes.

ceedubs:
Okay I tried this out a bit locally and I do get the Vim warnings about a file changing from underneath me, but it's hot too bad, and I can probably configure Vim to automatically load new changes or something if I want to. I'm okay with this being merged, but I'll let someone else weigh in.

Does anyone have insight into how this would work for IntelliJ users? My understanding is that IntelliJ has its own presentation compiler and that people often don't use SBT to run compile if they use IntelliJ. Would the formatting still get triggered?

Also it's possible that this won't catch some cases, such as if someone adds a new file and doesn't add it to `git` until after they've done all of their coding and testing. But that's fine; it will be caught by Travis and there's probably not much else that we could do for that case.

joan38:
For me Intellij and formatting outside of it is a mess since it doesn't pick up the change underneath and then as soon as I change the file and try to save it would prompt me which version I want.

Intellij finds the `.scalafmt` file and prompt you to use scalafmt instead, and then you can check the box `Format` in the commit dialogue. It's perfect!

kailuowang:
@joan38 On my team's machine, intellJ picks up external change automatically without a blink. Maybe you and I are using different settings?

joan38:
screenshot 2018-10-22 at 18 18 21

I'm using default settings everywhere.

joan38:
@kailuowang Yes, it's on by default

milessabin:
I'm also not at all enthusiastic about this.

Could we add a "prepr" sbt task which reformats and runs all relevant pre-PR tests and document that prominently that the task must be run locally before submitting a PR?

milessabin:
FWIW, format-on-compile was easily my least favourite part of working on SBT.

olafurpg:
I don't recommend format on compile in the build tool since it messes up with undo buffers in the editor and it slows down compilation. This functionality was contributed externally and I was hesitant to merge it in the sbt-scalafmt plugin (it should probably go out when I think of it 🤔).

I recommend running scalafmt from the editor in addition to keeping a slim 12kb bootstrap script in the repo (https://scalameta.org/scalafmt/docs/installation.html#coursier) that you can run as `./bin/scalafmt --diff` before git push.  Scalafmt is included out-of-the-box with the latest IntelliJ Scala plugin and for people using other editors like vim/emacs you can build a graalvm native-image thanks to https://github.com/scalameta/scalafmt/pull/1266

project/plugins.sbt
```diff
@@ -17,4 +17,4 @@ addSbtPlugin("org.portable-scala" % "sbt-scalajs-crossproject" % "0.6.0")
 addSbtPlugin("org.portable-scala" % "sbt-scala-native-crossproject" % "0.6.0")
 addSbtPlugin("org.scala-js" % "sbt-scalajs" % "0.6.24")
 addSbtPlugin("org.scala-native" % "sbt-scala-native" % "0.3.8")
-addSbtPlugin("com.geirsson" % "sbt-scalafmt" % "1.5.1")
+addSbtPlugin("com.geirsson" % "sbt-scalafmt" % "1.6.0-RC4")
```
olafurpg: Beware that v1.6 is incompatible with the version that comes bundled with IntelliJ. It's currently not possible to override the scalafmt version that comes installed by default with IntelliJ https://gitter.im/JetBrains/intellij-scala?at=5bcf1a52ae7be94016a22b6c

kailuowang:
looks like format on compile will cause some friction, I just thought of another idea to mitigate the problem of adding a manual step - PR template. In the PR template we could add a reminder for people run a certain sbt task before submitting the PR. What do you guys think?

project/plugins.sbt
joan38: Yes I have project with 1.6 and it's annoying because it invalidates the use of the Intellij plugin.

kailuowang:
replaced with a different solution #2583

PR #2568, opened 2018-10-19
give test more memory

kailuowang:
I noticed that the test is running in fork and was using the default java options. 
This setting change shall boost it's max memory to 6GB, according to travis, we have 7.5 G on the machine.


ceedubs:
I see this warning when running tests locally: `[warn] javaOptions will be ignored, fork is set to false`

kailuowang:
The Travis machine we use has 7.5GB. So we have some room left. 
The warning you got worth investigating. The fork in test is set to true right above...I am on my phone, will check later.

kailuowang:
@ceedubs finally got some time to correct the setting in #2580

PR #2573, opened 2018-10-20
Add biFlatMap to EitherT

### LINKED ISSUE #2438: Add biFlatMap to EitherT
`EitherT` has a `bimap` but no `biFlatMap`. This would be useful when chaining computations and needing to process both sides of the `Either` with functions that return an `EitherT`. For example:

```scala
def findVertex(id: String): EitherT[Future, APIError, Vertex]
def addVertex(...): EitherT[Future, APIError, String]
def doSomethingWithVertex(...): EitherT[Future, APIError, String]

findVertex("id").biFlatMap( {
  case _: APIError.NotFound => addNewVertex(...)
  case error => EitherT.left(error)
},
vertex => doSomethingWithVertex(vertex)
)
```

agajek:
Fixes #2438

### DIFF
--- a/core/src/main/scala/cats/data/EitherT.scala
+++ b/core/src/main/scala/cats/data/EitherT.scala
@@ -77,6 +77,13 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
                                                          applicativeG: Applicative[G]): G[EitherT[F, C, D]] =
     applicativeG.map(traverseF.traverse(value)(axb => Bitraverse[Either].bitraverse(axb)(f, g)))(EitherT.apply)
 
+  def biFlatMap[AA >: A, BB >: B](fa: A => EitherT[F, AA, BB],
+                                  fb: B => EitherT[F, AA, BB])(implicit F: Monad[F]): EitherT[F, AA, BB] =
+    EitherT(F.flatMap(value) {
+      case Left(a)  => fa(a).value
+      case Right(a) => fb(a).value
+    })
+
   def applyAlt[D](ff: EitherT[F, A, B => D])(implicit F: Apply[F]): EitherT[F, A, D] =
     EitherT[F, A, D](F.map2(this.value, ff.value)((xb, xbd) => Apply[Either[A, ?]].ap(xbd)(xb)))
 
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -77,6 +77,13 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
                                                          applicativeG: Applicative[G]): G[EitherT[F, C, D]] =
     applicativeG.map(traverseF.traverse(value)(axb => Bitraverse[Either].bitraverse(axb)(f, g)))(EitherT.apply)
 
+  def biFlatMap[AA >: A, BB >: B](fa: A => EitherT[F, AA, BB],
+                                  fb: B => EitherT[F, AA, BB])(implicit F: Monad[F]): EitherT[F, AA, BB] =
```
ceedubs:
```suggestion
                                  fb: B => EitherT[F, AA, BB])(implicit F: FlatMap[F]): EitherT[F, AA, BB] =
```

agajek:
@ceedubs is it possible to rerun travis' build? It indicates some memory problem of host rather than issues with my code.

ceedubs:
We have a 2-approver rule for merging non-documentation PRs. @kailuowang or @LukaJCB would one of you want to take a look?

ochrons:
A question on naming, should this actually be `biflatMap` instead of `biFlatMap` as there are `bimap` and `bitraverse` with no capitalization in the middle?

core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -77,6 +77,13 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
                                                          applicativeG: Applicative[G]): G[EitherT[F, C, D]] =
     applicativeG.map(traverseF.traverse(value)(axb => Bitraverse[Either].bitraverse(axb)(f, g)))(EitherT.apply)
 
+  def biFlatMap[AA >: A, BB >: B](fa: A => EitherT[F, AA, BB],
+                                  fb: B => EitherT[F, AA, BB])(implicit F: FlatMap[F]): EitherT[F, AA, BB] =
+    EitherT(F.flatMap(value) {
+      case Left(a)  => fa(a).value
+      case Right(a) => fb(a).value
```
phderome: would not it be more intuitive to say `case Right(b) => fb(b).value` at L.84 to match with the type `B` fb expects?

PR #2578, opened 2018-10-22
Add awk script to parse test durations

ceedubs:
This is something that I have recreated a couple of times when I've been
looking into builds timing out, so I wanted to save it somewhere in case
I want to use it again. I figured I might as well submit a PR in case
other people might find this to be useful. But I can understand if we
don't want to litter the Cats code base with one-off scripts.


LukaJCB:
I've never used `awk` in my life, but from what I can tell it looks good :+1:

PR #2579, opened 2018-10-22
Added `Chain#distinct` and `NonEmptyChain#distinct` to match `List#distinct` and `NonEmptyList#distinct`

fmsbeekmans:
Implemented and tested `Chain#distinct` and `NonEmptyChain#distinct`.

Still needs examples.

### DIFF
--- a/core/src/main/scala/cats/data/Chain.scala
+++ b/core/src/main/scala/cats/data/Chain.scala
@@ -360,8 +360,8 @@ sealed abstract class Chain[+A] {
     }
 
   /**
-    * Remove duplicates. Duplicates are checked using `Order[_]` instance.
-    */
+   * Remove duplicates. Duplicates are checked using `Order[_]` instance.
+   */
   def distinct[AA >: A](implicit O: Order[AA]): Chain[AA] = {
     implicit val ord = O.toOrdering
 
--- a/core/src/main/scala/cats/data/Chain.scala
+++ b/core/src/main/scala/cats/data/Chain.scala
@@ -5,6 +5,7 @@ import Chain._
 
 import scala.annotation.tailrec
 import scala.collection.immutable.SortedMap
+import scala.collection.immutable.TreeSet
 import scala.collection.mutable.ListBuffer
 
 /**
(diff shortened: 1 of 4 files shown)


core/src/main/scala/cats/data/Chain.scala
```diff
@@ -358,6 +359,24 @@ sealed abstract class Chain[+A] {
...
 
+  /**
+   * Remove duplicates. Duplicates are checked using `Order[_]` instance.
+   */
+  def distinct[AA >: A](implicit O: Order[AA]): Chain[AA] = {
+    implicit val ord = O.toOrdering
+
+    var alreadyIn = TreeSet.empty[AA]
+
+    foldLeft(Chain.empty[AA]) { (elementsSoFar, b) =>
```
CucumisSativus: Just a question of curiosity. Why are you using TreeSet instead of pure scala's Set?

core/src/main/scala/cats/data/NonEmptyChain.scala
```diff
@@ -372,6 +372,24 @@ class NonEmptyChainOps[A](val value: NonEmptyChain[A]) extends AnyVal {
   /** Reverses this `NonEmptyChain` */
   final def reverse: NonEmptyChain[A] =
     create(toChain.reverse)
+
+  /**
+   * Remove duplicates. Duplicates are checked using `Order[_]` instance.
+   */
+  final def distinct[AA >: A](implicit O: Order[AA]): NonEmptyChain[AA] = {
```
CucumisSativus: I wonder. Dont you think that it would be better to to call `toChain` and then `distinct` on chain and then rebuild NonEmptyChain? It would remove duplication but I'm not sure if it will still be ok with performance.

core/src/main/scala/cats/data/Chain.scala
fmsbeekmans: This was done for symmetry with `NonEmptyList#distinct` but it's a good question. `Set` uses a `HashSet` with sets bigger than 4 elements.  `HashSet#contains` has O(1) and `TreeSet#contains` has O(log(n)).

Changing `Set` would also remove the artificial dependency on `Order`. Was there a reason for this dependency when `NonEmptyList#distinct` was introduced?

core/src/main/scala/cats/data/NonEmptyChain.scala
fmsbeekmans: The other way around would be cheaper, something along the lines of `NonEmptyChain.fromChain(this).fold(empty[A])(_.distinct.toChain)`

core/src/main/scala/cats/data/Chain.scala
nasadorian: I wrote up a functional set implementation that should speed up `distinct` in the future and let it depend on `Eq` instead of `Order`. Closed issue here, was suggested to put up a PR on cats-collection instead. https://github.com/typelevel/cats/issues/2593

nasadorian: Actually, it seems a structure like this already exists in cats-collections called ISet. I'd be curious to see the performance difference between TreeSet and ISet for this use case.

PR #2580, opened 2018-10-23
fix fork test

kailuowang:
follow up for #2568 
This seems to be the correct setting.


ceedubs:
Oh haha we submitted basically the same PR at the same time 😆 .

We have a couple of slight differences. Does the `Test` / `test` difference matter? Also does it make sense to have the argument in Java settings for JS?

kailuowang:
yeah, I think yours makes more sense.

PR #2588, opened 2018-10-30
Add flap function to Apply

ssanj:
Add the `flap` function to `Apply` to facilitate combining a function in an
`Apply` context with a value which is not.

This is similar to `ap`:

```
```

with the minor change of the value not being in a context:

```
def flap[A, B](ff: F[A => B])(a: A): F[B]
                              ^^^^
```

This is a kind reminder to run `sbt prePR` and commit the changed files before submitting.

### DIFF
--- a/core/src/main/scala/cats/syntax/all.scala
+++ b/core/src/main/scala/cats/syntax/all.scala
@@ -31,6 +31,7 @@ trait AllSyntax
     with EqSyntax
     with FlatMapSyntax
     with FoldableSyntax
+    with Function1Syntax
     with FunctorSyntax
     with GroupSyntax
     with HashSyntax
--- a/core/src/main/scala/cats/syntax/all.scala
+++ b/core/src/main/scala/cats/syntax/all.scala
@@ -31,7 +31,6 @@ trait AllSyntax
     with EqSyntax
     with FlatMapSyntax
     with FoldableSyntax
-    with Function1Syntax
     with FunctorSyntax
     with GroupSyntax
     with HashSyntax
@@ -77,4 +76,4 @@ trait AllSyntaxBinCompat2
     with ListSyntaxBinCompat0
     with ValidatedSyntaxBincompat0
 
-trait AllSyntaxBinCompat3 extends UnorderedFoldableSyntax
+trait AllSyntaxBinCompat3 extends UnorderedFoldableSyntax with Function1Syntax
--- a/core/src/main/scala/cats/Apply.scala
+++ b/core/src/main/scala/cats/Apply.scala
@@ -40,6 +40,28 @@ trait Apply[F[_]] extends Functor[F] with InvariantSemigroupal[F] with ApplyArit
   def ap[A, B](ff: F[A => B])(fa: F[A]): F[B]
 
   /**
+   * Given a function in the Apply context and a plain value, supplies the
+   * value to the function.
+   *
+   * Example:
+   * {{{
+   * scala> import cats.implicits._
+   *
+   * scala> val someF: Option[Int => Long] = Some(_.toLong + 1L)
+   * scala> val noneF: Option[Int => Long] = None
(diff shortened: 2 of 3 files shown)


LukaJCB:
Hi @ssanj, thanks for contributing! I don't think `Apply` is the correct type class here as it could simply be implemented with `Functor` instead. Futhermore I'm not sure `flap` is the best name for this, but I'm also not coming up with a better name right now unfortunately, might think about this some more :)

ssanj:
You're right. This could be moved into `Functor`. The reason I moved it into `Apply` was that it had a function in a context (`F[A => B]`) and was verify similar to `ap`. Happy to move it.

Yeah, naming this is hard. I hard a look at some of the Haskell equivalents and the variants I saw were:
- [flip](http://hackage.haskell.org/package/functors-0.1/docs/Data-Functor-Syntax.html#v:flip)
- [flap](https://hackage.haskell.org/package/relude-0.3.0/docs/Relude-Functor-Fmap.html#v:flap)
- [other options](http://hayoo.fh-wedel.de/?query=f+%28a+-%3E+b%29+-%3E+a+-%3E+f+b)

Another question I had was should this obey any specific Laws?

ssanj:
When I move the `flap` function into `Functor` I get the following compiler error:

```
[error] /Volumes/Work/projects/code/scala/opensource/cats/core/src/main/scala/cats/Functor.scala:56:21: not found: type A
[error]   def widen[A, B >: A](fa: F[A]): F[B] = fa.asInstanceOf[F[B]]
[error]                     ^
[error] /Volumes/Work/projects/code/scala/opensource/cats/core/src/main/scala/cats/Functor.scala:12:2: type arguments [C,B] do not conform to method widen's type parameter bounds [A,B >: A]
[error] @typeclass trait Functor[F[_]] extends Invariant[F] { self =>
[error]  ^
[error] two errors found
[error] /Volumes/Work/projects/code/scala/opensource/cats/core/src/main/scala/cats/Functor.scala:56:21: not found: type A
[error]   def widen[A, B >: A](fa: F[A]): F[B] = fa.asInstanceOf[F[B]]
[error]                     ^
[error] /Volumes/Work/projects/code/scala/opensource/cats/core/src/main/scala/cats/Functor.scala:12:2: type arguments [C,B] do not conform to method widen's type parameter bounds [A,B >: A]
[error] @typeclass trait Functor[F[_]] extends Invariant[F] { self =>
[error]  ^
[error] two errors found
```

I've enquired whether this is a [Simulcrum issue](https://github.com/mpilquist/simulacrum/issues/115). I can get this to work if I rewrite `widen` not to have a upper bound:

```
def widen[A, B](fa: F[A])(implicit ev: A <:< B): F[B] = fa.asInstanceOf[F[B]]
```

Has anyone seen this issue before?

kailuowang:
How about we call this `mapApply` with a `F[_] :Functor`  bound add it to a Function1 syntax extension trait? We can't add any method to any trait without breaking binary compatibility in Scala 2.11. And we promised to not break it until 2.0

core/src/main/scala/cats/syntax/all.scala
```diff
@@ -31,6 +31,7 @@ trait AllSyntax
     with EqSyntax
     with FlatMapSyntax
     with FoldableSyntax
+    with Function1Syntax
```
kailuowang: due to binary compatibility restrictions, you can't add it to this trait either. However, you can add it to `AllSyntaxBinCompat3 ` which is a new trait that hasn't been released yet.

ssanj: Thanks. 👍

ssanj:
@kailuowang @ssanj does this look ok now?

kailuowang:
I am going to schedule this to 1.5 so that you don't have to have change your BinaryCompat trait.

PR #2589, opened 2018-10-31
updated to 2.13-M5

kailuowang:
Hopefully we can merge this one quickly and publish 1.5


erikerlandson:
+1, this is on the path for `algebra` and `spire`

ChristopherDavenport:
Looks like some sort of scalatest or scalactic bug.

kailuowang:
The previous error was fixed, now  I have a randomly occurring error (roughly 50% of the time) on scalajs
>[info] - Try with Throwable.monadError.applicativeError recover consistent with recoverWith *** FAILED ***
[info]   UndefinedBehaviorError was thrown during property evaluation.
[info]     Message: An undefined behavior was detected: scala.runtime.Statics$PFMarker$@26 is not an instance of scala.util.Try
[info]       arg0 = Failure(java.lang.Error),
[info]       arg1 = <function1>
[info]     )

erikerlandson:
log spew might be a problem unto itself here?
`The job exceeded the maximum log length, and has been terminated.`

kailuowang:
turns out to be a `scala.util.Failure` regression, reported by @sjrd https://github.com/scala/bug/issues/11242

SethTisue:
can it be worked around for now?

kailuowang:
@SethTisue the scala.util.Failure regression can probably be circumvented by disabling the test on 2.13 temporarily. I am not sure about the scalatest output one. I'll poke around scalatest settings.

erikerlandson:
just one test failing for (2.13, js_build false) in `OpSuite`
```
[info] OpSuite:
[info] - Op[Function1, Char, Int].eq.antisymmetry
[info] - Op[Function1, Char, Int].eq.reflexitivity
[info] - Op[Function1, Char, Int].eq.symmetry
[info] - Op[Function1, Char, Int].eq.transitivity
[info] - Eq[Op[Function1, Char, Int]].serializable.can serialize and deserialize *** FAILED ***
[info]   ClassCastException was thrown during property evaluation.
[info]     Message: cannot assign instance of java.lang.invoke.SerializedLambda to field org.scalacheck.Gen$$anon$5.f$5 of type scala.Function2 in instance of org.scalacheck.Gen$$anon$5
[info]   
[info]     )
[info] - Op[Function1, Char, Int].category.category left identity
[info] - Op[Function1, Char, Int].category.category right identity
[info] - Op[Function1, Char, Int].category.compose associativity
[info] - Category[Op[Function1, ?, ?]].serializable.can serialize and deserialize
```

SethTisue:
looks like https://github.com/rickynils/scalacheck/pull/427#issuecomment-424330310 / https://github.com/scala/scala-dev/issues/562

kailuowang:
the fix to this one is to turn `fork in test` back to true, it was turned to false as a workaround for the scalatest issue. So, if you allow me,
(╯°□°）╯︵ ┻━┻

erikerlandson:
can you just disable this single test using `BuildInfo.scalaVersion != "2.13.0-M5"` like you did with the other test in TrySuite?

kailuowang:
I'd like to turn `fork in test` back to `true`, since that's better memory wise. Cheeseng is looking at https://github.com/scalatest/scalatest/issues/1423 so I am hopeful.

SethTisue:
> I'd like to turn fork in test back to true, since that's better memory wise

I'm mostly just curious, but how come it isn't better to just crank up `-Xmx` some more? I don't recall seeing other projects setting `fork` to `true` for this reason (unless they were doing weird things with threads and classloaders, something like that, which I wouldn't expect in this repo)

kailuowang:
@SethTisue good question! Due to the need to generate coverage report, we need to run test from root module and congregate coverage reports (although that may be hackable). Since each test run classloader is not garbage collected, it's probably better memory wise that each module's test is run in a fresh forked jvm.

erikerlandson:
🎉

tests/src/test/scala/cats/tests/TrySuite.scala
```diff
@@ -17,7 +17,11 @@ class TrySuite extends CatsSuite {
   checkAll("Try[Int]", CoflatMapTests[Try].coflatMap[Int, Int, Int])
   checkAll("CoflatMap[Try]", SerializableTests.serializable(CoflatMap[Try]))
 
-  checkAll("Try with Throwable", MonadErrorTests[Try, Throwable].monadError[Int, Int, Int])
+  //temporarily disable this test due to scala.util.Failure regression https://github.com/scala/bug/issues/11242
+  if (BuildInfo.scalaVersion != "2.13.0-M5") {
```
ceedubs: We might want to open an issue in this repo to remind ourselves to remove this hack.

erikerlandson: done: #2601

kailuowang: thanks @erikerlandson

Issue #2288, opened 2018-06-09
Add an update method to NonEmptyMap


LukaJCB:
Something along these lines:

```scala
def update[K, A](nem: NonEmptyMap[K, A], key: K, f: A => A): NonEmptyMap[K, A]
```

Unfortunately `SortedMap` doesn't expose such a method.

johnynek:
Why is this different from mapValues or the Functor map?

johnynek:
Oh, I see. Only update one key. Duh.

kubukoz:
You surely meant `updated`, right? :P

kubukoz:
Nobody said they were working on it, so I guess you're good to go :P

kubukoz:
Btw. wouldn't it be better to have `modify` in the name? `update` reminds me of the standard library's `updated`, which just takes a key and a new value, but this is more like `modifyAt` if you look at how `modify` is named in `Ref`: https://github.com/typelevel/cats-effect/blob/master/core/shared/src/main/scala/cats/effect/concurrent/Ref.scala#L115

Obarros:
Yeah, good point @kubukoz. 
When I look at `updated` usually I think of something that will create a new map obtained by updating `this` map with a given key/value pair as well ...
Here we are after Modifying the initial state using `f`, it would make sense to make it part of the name somehow.

nasadorian:
Hi @Obarros @LukaJCB was this issue already addressed in a PR? Happy to help out as it seems like somewhat low hanging fruit.

LukaJCB:
Hi @nasadorian, as far as I know there's been no PR yet

nasadorian:
@LukaJCB https://github.com/typelevel/cats/pull/2592

PR #2592, opened 2018-11-02
Add `updateWith` method to NonEmptyMap

nasadorian:
- Named `updateWith` since `updated` would imply a simple setting of k -> v
- Added a couple small tests

Closes https://github.com/typelevel/cats/issues/2288

### DIFF
--- a/core/src/main/scala/cats/data/NonEmptyMapImpl.scala
+++ b/core/src/main/scala/cats/data/NonEmptyMapImpl.scala
@@ -94,6 +94,11 @@ sealed class NonEmptyMapOps[K, A](val value: NonEmptyMap[K, A]) {
   def lookup(k: K): Option[A] = toSortedMap.get(k)
 
   /**
+   * Applies f to the value stored at k. If lookup misses, does nothing.
+   */
+  def updateWith(k: K, f: A => A): NonEmptyMap[K, A] = lookup(k).fold(value)(v => add((k, f(v))))
+
+  /**
    * Returns a `SortedSet` containing all the keys of this map.
    */
   def keys: NonEmptySet[K] = NonEmptySet.fromSetUnsafe(toSortedMap.keySet)
--- a/core/src/main/scala/cats/data/NonEmptyMapImpl.scala
+++ b/core/src/main/scala/cats/data/NonEmptyMapImpl.scala
@@ -96,7 +96,11 @@ sealed class NonEmptyMapOps[K, A](val value: NonEmptyMap[K, A]) {
   /**
    * Applies f to the value stored at k. If lookup misses, does nothing.
    */
-  def updateWith(k: K)(f: A => A): NonEmptyMap[K, A] = lookup(k).fold(value)(v => add((k, f(v))))
+  def updateWith(k: K)(f: A => A): NonEmptyMap[K, A] =
+    lookup(k) match {
+      case Some(v) => add((k, f(v)))
+      case None    => value
+    }
 
   /**
    * Returns a `SortedSet` containing all the keys of this map.
--- a/core/src/main/scala/cats/data/NonEmptyMapImpl.scala
+++ b/core/src/main/scala/cats/data/NonEmptyMapImpl.scala
@@ -96,7 +96,7 @@ sealed class NonEmptyMapOps[K, A](val value: NonEmptyMap[K, A]) {
   /**
    * Applies f to the value stored at k. If lookup misses, does nothing.
    */
-  def updateWith(k: K, f: A => A): NonEmptyMap[K, A] = lookup(k).fold(value)(v => add((k, f(v))))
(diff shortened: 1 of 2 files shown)


nasadorian:
Build is flaky?

core/src/main/scala/cats/data/NonEmptyMapImpl.scala
```diff
@@ -93,6 +93,11 @@ sealed class NonEmptyMapOps[K, A](val value: NonEmptyMap[K, A]) {
    */
   def lookup(k: K): Option[A] = toSortedMap.get(k)
 
+  /**
+   * Applies f to the value stored at k. If lookup misses, does nothing.
+   */
+  def updateWith(k: K, f: A => A): NonEmptyMap[K, A] = lookup(k).fold(value)(v => add((k, f(v))))
```
kubukoz: I'm wondering if `updateWith[B >: A](k: K, f: A => B): NonEmptyMap[K, B]` wouldn't be better

kubukoz:
Can you add `Closes #2288` to the description? It'll close the issue automatically once this is merged.

ceedubs:
Thanks @nasadorian!

This looks good to me, but I don't think that the tests are hitting the case where the key isn't in the map. Could you explicitly test that case? It could probably be a unit test and/or an sbt-doctest example.

core/src/main/scala/cats/data/NonEmptyMapImpl.scala
kailuowang: I like the ability to widen, I also think it's more cats conventional to curry the `f` param.

nasadorian: So the underlying collection is covariant in the value type `SortedSet[A, +B]`. I guess it would line up correctly to allow the widening?

nasadorian: @kailuowang so just `updateWith(k: K)(f: A => A)`?

nasadorian: IDK if this will work because `updateWith` depends on `add` which is in A only...

```
[error]  found   : B
[error]  required: A
[error]   def updateWith[B >: A](k: K)(f: A => B): NonEmptyMap[K, B] = lookup(k).fold(value)(v => add((k, f(v))))
[error]                                                                                                    ^
[error] one error found
```

kailuowang: yeah now I realized that  `add` doesn't widen, it doesn't make much sense for updateWith to be able to widen .

tests/src/test/scala/cats/tests/NonEmptyMapSuite.scala
```diff
@@ -209,4 +209,12 @@ class NonEmptyMapSuite extends CatsSuite {
       nem.toNel should ===(NonEmptyList.fromListUnsafe(nem.toSortedMap.toList))
     }
   }
+
+  test("NonEmptyMap#updateWith consistent with equivalent operations") {
+    forAll { (nem: NonEmptyMap[String, Int], i: (String, Int)) =>
+      nem.add(i) should ===(nem.add(i).updateWith(i._1)(identity))
```
kubukoz: I think it would be better to have these three checks as separate tests :)

nasadorian: Good point

nasadorian: Separated the tests

core/src/main/scala/cats/data/NonEmptyMapImpl.scala
```diff
@@ -93,6 +93,11 @@ sealed class NonEmptyMapOps[K, A](val value: NonEmptyMap[K, A]) {
    */
   def lookup(k: K): Option[A] = toSortedMap.get(k)
 
+  /**
+   * Applies f to the value stored at k. If lookup misses, does nothing.
+   */
+  def updateWith(k: K)(f: A => A): NonEmptyMap[K, A] = lookup(k).fold(value)(v => add((k, f(v))))
```
johnynek: Pattern matching on the Option allows you to avoid the allocation in the empty case of fold. The code isn’t as pretty but we’ve benchmarked before and is measurable so we generally use pattern matching over functions taking lambdas inside cats methods.

nasadorian: I think I've seen a similar comment on SO before. What's the allocation for, the lambda?

johnynek: Both the lazy parameter to the empty case of fold and the lambda will allocate and neither will be inlined (in all probability) by the JVM.

By contrast neither problem exists with matching.

nasadorian: In this case, the lazy parameter to fold is just a reference to the current data structure. Is it copied or something? I totally agree with you I'm just curious.

PR #2597, opened 2018-11-05
Optimize `Apply` methods in `FlatMap`

kailuowang:
I noticed that these implementations can be slightly improved. Given their common usages I think it would worth it. 
related issue https://github.com/typelevel/cats-effect/issues/401

### DIFF
--- a/core/src/main/scala/cats/FlatMap.scala
+++ b/core/src/main/scala/cats/FlatMap.scala
@@ -22,12 +22,6 @@ import simulacrum.noop
   def flatMap[A, B](fa: F[A])(f: A => F[B]): F[B]
 
 
-  override def productR[A, B](fa: F[A])(fb: F[B]): F[B] =
-    flatMap(fa)(_ => fb)
-
-  override def productL[A, B](fa: F[A])(fb: F[B]): F[A] =
-    flatMap(fa)(a => map(fb)(_ => a))
-
   /**
    * "flatten" a nested `F` of `F` structure into a single-layer `F` structure.
    *
@@ -102,6 +96,11 @@ import simulacrum.noop
   override def map2[A, B, Z](fa: F[A], fb: F[B])(f: (A, B) => Z): F[Z] =
     flatMap(fa)(a => map(fb)(b => f(a, b)))
 
+  override def productR[A, B](fa: F[A])(fb: F[B]): F[B] =
+    flatMap(fa)(_ => fb)
+
+  override def productL[A, B](fa: F[A])(fb: F[B]): F[A] =
+    map2(fa, fb)((a, _) => a)
 
   /**
    * Pair `A` with the result of function application.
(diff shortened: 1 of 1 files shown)


ceedubs:
I think that this is probably a good change to make, but I'm wondering if it should go into a new minor version as opposed to a patch. My concern is that someone might have an `IO`/`Future`-like type where `ap` runs effects in parallel but a `FlatMap` instance is also defined. While we have laws that say that `ap` and `flatMap` should be consistent, our equality is only based on result values and not rumtime execution characteristics, so it's debatable whether or not their instance broke laws. So I wouldn't want to tank someone's performance without sufficient warning.

In fact I wonder whether this might apply to our instances for standard library `Future`?

kailuowang:
@ceedubs this is probably not a breaking change because currently `FlatMap` already overrides `product` with implementation using `flatMap`. And `product` is called by these methods. So if a `FlatMap` instance is used, it's already using `flatMap` v.s. `ap`, e.g. the current implementation of `productL` and `productR` in `FlatMap` is essentially, which you can see is wasteful in doing `map`s 
```
 map(flatMap(fa)(a => map(fb)(b => (a, b))))((a, _) => a)
 map(flatMap(fa)(a => map(fb)(b => (a, b))))((_, b) => b)
```
So I don't blieve this change will change the effect. 
I am not 100% sure if `product` being overriden in `FlatMap` is more correct, but my thinking is that user should avoid getting tied to the concrete effect type `F[_]`, and if they keep effect type abstract and use a context bound `F[_]: FlatMap` then they should expect sequential execution anyway.

core/src/main/scala/cats/FlatMap.scala
```diff
@@ -89,6 +89,18 @@ import simulacrum.noop
...
     flatMap(fa)(a => map(fb)(b => (a, b)))
 
+  override def ap2[A, B, Z](ff: F[(A, B) => Z])(fa: F[A], fb: F[B]): F[Z] =
+    flatMap(fa)(a => flatMap(fb)(b => map(ff)(_(a, b))))
+
+  override def map2[A, B, Z](fa: F[A], fb: F[B])(f: (A, B) => Z): F[Z] =
+    flatMap(fa)(a => map(fb)(b => f(a, b)))
+
+  override def productR[A, B](fa: F[A])(fb: F[B]): F[B] =
+    flatMap(fa)(_ => fb)
```
ceedubs: Coming back to this, the lack of symmetry between `productR` and `productL` feels a little off to me. I think that the override of `map2` makes sense. But maybe `productL` and `productR` should both just delegate to `map2`?

kailuowang: Unlike `productL`, `productR` using `map2` will incur one extra `map`. That's where the asymmetry comes from.

ceedubs: Oh right. Sorry that I've required so much explanation on this one 😬

LGTM. @mpilquist does this PR look good to you?

mpilquist: Yep looks good to me

PR #2600, opened 2018-11-07
Add `sequenceFilter` syntax

kubukoz:


core/src/main/scala/cats/syntax/all.scala
```diff
@@ -77,3 +78,5 @@ trait AllSyntaxBinCompat2
     with ValidatedSyntaxBincompat0
 
 trait AllSyntaxBinCompat3 extends UnorderedFoldableSyntax
+
+trait AllSyntaxBinCompat4 extends TraverseFilterSyntaxBinCompat0
```
kailuowang: `AllSyntaxBinCompat3` was introduced since last release, you should be able to just use this.

kubukoz: Ah, wonderful. I was actually thinking about that but hadn't checked before making this PR. I'll move the extension there, then.

I also plan on providing some test(s).

kubukoz:
@kailuowang is there a chance to get this in 1.5.0? I'll merge the bincompat syntax traits now.

kubukoz:
Looks like I'm missing the new instances (`TraverseFilter[List]`) in syntax tests, this shouldn't take long

core/src/main/scala/cats/syntax/traverseFilter.scala
```diff
@@ -2,3 +2,12 @@ package cats
 package syntax
 
 trait TraverseFilterSyntax extends TraverseFilter.ToTraverseFilterOps
+
+trait TraverseFilterSyntaxBinCompat0 {
```
kailuowang: We also want this to go to `cats.syntax.all` and `cats.implicits` right?

kailuowang:
ideally we want to test all non-bug fix changes in RC before rolling out. So if it's not urgent let's wait for 1.6. I plan to shorten the release cadence to 1-2 months.

core/src/main/scala/cats/syntax/traverseFilter.scala
kailuowang: also can we add at least a doctest for it?

kubukoz:
Sure, that fits me. I'll update this PR soon :)

core/src/main/scala/cats/syntax/traverseFilter.scala
kubukoz: It was in `cats.syntax.all` :) I added it to cats.implicits now, also made a doctest. Because there's been a release in the meantime, I added back the new bincompat trait. I'll push the changes once sbt gives me a green light.

kubukoz:
@kailuowang please take a look :)

PR #2606, opened 2018-11-12
Add syntax for Bitraverse to allow traversal and sequencing of the left side

andimiller:
This adds Bitraverse-based syntax for:
* leftTraverse
* leftSequence


tests/src/test/scala/cats/tests/SyntaxSuite.scala
```diff
@@ -299,9 +299,17 @@ object SyntaxSuite extends AllSyntaxBinCompat with AllInstances with AllSyntax {
...
+    val gfcb = fab.leftTraverse(f)
 
     val fgagb = mock[F[G[A], G[B]]]
     val gfab = fgagb.bisequence
+
+    val fgab = mock[F[G[A], B]]
+    val gfab = fgab.leftSequence
+
+    val fagb = mock[F[A, G[B]]]
+    val gfab = fagb.sequence
```
LukaJCB: `gfab` is defined twice

core/src/main/scala/cats/syntax/bitraverse.scala
```diff
@@ -11,14 +11,38 @@ private[syntax] trait BitraverseSyntax1 {
...
+  implicit final def catsSyntaxRightnestedBitraverse[F[_, _]: Bitraverse, G[_], A, B](
+    fagb: F[A, G[B]]
+  ): RightNestedBitraverseOps[F, G, A, B] =
+    new RightNestedBitraverseOps[F, G, A, B](fagb)
 }
 
 final class BitraverseOps[F[_, _], A, B](val fab: F[A, B]) extends AnyVal {
   def bitraverse[G[_]: Applicative, C, D](f: A => G[C], g: B => G[D])(implicit F: Bitraverse[F]): G[F[C, D]] =
     F.bitraverse(fab)(f, g)
+  def traverse[G[_], C](f: B => G[C])(implicit F: Bitraverse[F], G: Applicative[G]): G[F[A, C]] =
```
LukaJCB: I'm not sure about this one and `sequence`, we already have them defined on `Traverse` and I think that can cause complications. Maybe we should remove these or rename them to `rightTraverse` and `rightSequence`

tests/src/test/scala/cats/tests/SyntaxSuite.scala
andimiller: good catch, I was concentrating too much on getting the names right to notice I'd repeated some

core/src/main/scala/cats/syntax/bitraverse.scala
andimiller: I wasn't sure about this, but I'll change it to `rightTraverse` and `rightSequence` for now. You don't normally have syntax in scope to give you `traverse` and `sequence`, so I do think they're still valuable to have.

kubukoz:
Looks like this isn't binary compatible, @andimiller would you mind adding these implicit conversions to a new trait like https://github.com/typelevel/cats/pull/2600/files ?

andimiller:
> Looks like this isn't binary compatible, @andimiller would you mind adding these implicit conversions to a new trait like https://github.com/typelevel/cats/pull/2600/files ?

I have no idea how binary compatibility works, but I can give it a go copying how that PR does it

andimiller:
I think I've managed to make it binary compatible, I've left `{Left,Right}NestedBitraverseOps` without the `BinCompat0` suffix because they're new syntax classes, not sure if they should have the suffix to match the others, or if it's okay to leave them as is, because there are no "original" versions of those classes?

andimiller:
Should I have made a new `SyntaxBinCompat4` or put it into an existing one? No idea how these are meant to work

kubukoz:
I believe it'll be fine if you add it to bincompat3.

andimiller:
I've changed this to only provide `leftTraverse` and `leftSequence`, since adding the right-based versions is out of scope for this PR

ikempf:
@andimiller Hi, are you still on this ? Now that 1.5.0 is out this will have to go into a bincompat4.

andimiller:
@ikempf yeah I'll be updating this PR later today to put it in bincompat4

LukaJCB:
Seems like there's a bug in the travis reporting, I'm going to close/reopen :)

kailuowang:
Thanks! @andimiller  Adding some doctests, [an example](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Traverse.scala#L26-L35), would be helpful. Although I admit that a doctest here would be hard to find. I'll leave that to you.

andimiller:
I've added doctests to these syntax methods, I did look at adding these methods directly onto `Bitraverse` as derived methods, but that trips the binary compatibility checker, so it'd probably need to wait til 2.0.

I also can't figure out how to run doctests, so I've pushed it to see if travis manages it better than I can locally.

kailuowang:
@andimiller thanks for adding them.  simple `test` will run the doctest. In this case `sbt coreJVM/test`

Issue #2604, opened 2018-11-09
Contravariant instances for `Op`


stephen-lazaro:
I noticed there are a number of possible contravariant instances for `Op` but currently none are there. Do we feel like any of these would be beneficial? For example...

PR #2607, opened 2018-11-12
Add `Decidable`

### LINKED ISSUE #1935: Add Divisible and Decideable (contravariant Applicative and co.)

stephen-lazaro:
Resolves https://github.com/typelevel/cats/issues/1935 and https://github.com/typelevel/cats/issues/2604.

Seeing an issue with binary compatibility I'm not sure how to resolve correctly. 
Maybe someone can help me out? I'm probably misunderstanding the intended function of the BinCompat traits.

Included a very barebones doc page.

Since the laws are not formalized in ekmett's original, I've done a bit of guesswork / intiuitioneering on what the right laws look like. Open to discussion.


build.sbt
```diff
@@ -732,7 +732,7 @@ def commonScalacOptions(scalaVersion: String) =
     "-language:implicitConversions",
     "-language:experimental.macros",
     "-unchecked",
-    "-Ywarn-dead-code",
+    //"-Ywarn-dead-code",
```
stephen-lazaro: This probably needs to be discussed, it's required to define `empty`. We don't, in this expression of the type class, end up using `empty` much, so it might make more sense to drop `empty` than this, but I'd like to hear opinions before I go for it.

core/src/main/scala/cats/data/Const.scala
```diff
@@ -35,7 +35,7 @@ final case class Const[A, B](getConst: A) {
     s"Const(${A.show(getConst)})"
 }
 
-object Const extends ConstInstances {
+object Const extends ConstInstances with ConstInstancesBinCompat0 {
```
stephen-lazaro: This is causing a Mima issue... Had it as a method before but of course that makes it a ReverseMissingMethod issue. What's our practice on binary incompatible instances for data types?

LukaJCB: I think this will work if you just create a new ` ConstInstances2` and let ` ConstInstances1` extend from it. IIRC MiMa is fine with it when using sealed abstract classes. :)

build.sbt
LukaJCB: When you say `empty` you mean `zero`? :)

core/src/main/scala/cats/Decideable.scala
```diff
@@ -0,0 +1,27 @@
...
+@typeclass trait Decideable[F[_]] extends ContravariantMonoidal[F] {
+  def sum[A, B](fa: F[A], fb: F[B]): F[Either[A, B]]
+
+  def decide[A, B, C](fa: F[A], fb: F[B])(f: C => Either[A, B]): F[C] =
+    contramap(sum(fa, fb))(f)
+
+  def chosen[B, C](fb: F[B], fc: F[C]): F[Either[B, C]] =
+    decide(fb, fc)(identity[Either[B, C]])
+
+  def zero: F[Nothing] = trivial[Nothing]
```
LukaJCB: Are these two always the same? I feel like it's a bit weird that both the additive and multiplicative identities are the same, but it might very well be that my intuition for these sorts of things are off 🤔

docs/src/main/tut/typeclasses/decideable.md
```diff
@@ -0,0 +1,61 @@
...
+  def unit: Unit => Boolean = Function.const(true)
+  def contramap[A, B](fa: A => Boolean)(f: B => A): B => Boolean =
+    fa.compose(f)
+  def product[A, B](fa: A => Boolean, fb: B => Boolean): ((A, B)) => Boolean =
+    (ab: (A, B)) =>
+      ab match {
+        case (a, b) => fa(a) && fb(b)
+    }
+  def sum[A, B](fa: A => Boolean, fb: B => Boolean): Either[A, B] => Boolean =
+    either => either.fold(fa, fb)
```
LukaJCB: This is great, but maybe for posterity can we write it just like `product` except using `^` instead of `&&`? :)

docs/src/main/tut/typeclasses/decideable.md
```diff
@@ -0,0 +1,61 @@
...
+
+In other words, if we have two contexts `F[A]` and `F[B]` and we have a procedure `C => Either[A, B]` that could produce either an A or a B depending on the input, we have a way to produce a context `F[C]`.
+
+The classic example of a `Decideable` is that of predicates `A => Boolean`, leaning on the fact that both `&&` and `^` are valid monoids on `Boolean`.
+
+We can write the instance as:
+
+```tut:silent
+implicit val decideableForPredicates = new Decideable[? => Boolean] {
+  def empty[A]: A => Boolean = Function.const(false)
```
LukaJCB: Shouldn't this say `zero`?

core/src/main/scala/cats/instances/function.scala
```diff
@@ -58,6 +58,21 @@ trait FunctionInstancesBinCompat0 {
         Deferred(() => cachedFa)
       }
     }
+
+  implicit def catsStdDecideableForPredicate: Decideable[? => Boolean] =
+    new Decideable[? => Boolean] {
+      def empty[A]: A => Boolean = Function.const(false)
```
LukaJCB: This one should say `zero` as well, right?

docs/src/main/tut/typeclasses/decideable.md
```diff
@@ -0,0 +1,61 @@
...
+    (ab: (A, B)) =>
+      ab match {
+        case (a, b) => fa(a) && fb(b)
+    }
+  def sum[A, B](fa: A => Boolean, fb: B => Boolean): Either[A, B] => Boolean =
+    either => either.fold(fa, fb)
+}
+```
+
+This can be used to combine predicates over both coproducts and products.
```
LukaJCB: Maybe we should say *both sum types and product types*. I feel that these terms are a bit more common, but maybe that's just me

laws/src/main/scala/cats/laws/discipline/DecideableTests.scala
```diff
@@ -0,0 +1,84 @@
...
+          forAll(laws.decideableRightDistributivity[A, B, C] _)
+      )
+    }
+}
+
+object DecideableTests {
+  def apply[F[_]: Decideable](implicit ev: Isomorphisms[F]): DecideableTests[F] =
+    new DecideableTests[F] { def laws: DecideableLaws[F] = DecideableLaws[F] }
+
+  trait Isomorphisms[F[_]] {
```
LukaJCB: Do we really need these `Isomorphisms`? 
`Decidable` extends `Contravariant` so we can just that instance, no?

For `Semigroupal` which doesn't have access to an `Invariant` instance this sort of thing is needed to test the laws, but here it shouldn't be needed IMO :)

laws/src/main/scala/cats/laws/DecideableLaws.scala
```diff
@@ -0,0 +1,28 @@
...
+trait DecideableLaws[F[_]] extends ContravariantMonoidalLaws[F] {
+  implicit override def F: Decideable[F]
+
+  def decideableDecideRightAbsorption[A](fa: F[A]): IsEq[F[A]] =
+    (F.decide(F.trivial[A], fa))(Right.apply[A, A]) <-> fa
+
+  def decideableSumAssociativity[A, B, C](fa: F[A],
+                                          fb: F[B],
+                                          fc: F[C]): (F[Either[A, Either[B, C]]], F[Either[Either[A, B], C]]) =
+    (F.sum(fa, F.sum(fb, fc)), F.sum(F.sum(fa, fb), fc))
```
LukaJCB: Here we should just be able to contramap one to the other :)

laws/src/main/scala/cats/laws/DecideableLaws.scala
```diff
@@ -0,0 +1,28 @@
...
+
+  def decideableRightDistributivity[A, B, C](fa: F[A], fb: F[B], f: C => A, g: C => B): IsEq[F[Either[C, C]]] =
```
LukaJCB: This law tests for distributivity of `sum` and `contramap`, but I think we should also have one for `sum` and `product` :)

laws/src/main/scala/cats/laws/DecideableLaws.scala
```diff
@@ -0,0 +1,28 @@
+package cats
+package laws
+
+trait DecideableLaws[F[_]] extends ContravariantMonoidalLaws[F] {
+  implicit override def F: Decideable[F]
+
+  def decideableDecideRightAbsorption[A](fa: F[A]): IsEq[F[A]] =
```
LukaJCB: This one tests right identity, not absorption, right?
Absorption would be `product(fa, zero) == zero`

LukaJCB (commented):
Massive thanks @stephen-lazaro! Really great work on this PR. I've added a bunch of comments, but none are really major I think :)

laws/src/main/scala/cats/laws/DecideableLaws.scala
stephen-lazaro: Oooo good idea. I'll add that.

docs/src/main/tut/typeclasses/decideable.md
stephen-lazaro: Ah, just so.

build.sbt
LukaJCB: I think we should go with `def zero[A]: F[A]` then to make it easier. `Nothing` can be weird in a lot of ways :smile:

core/src/main/scala/cats/data/Const.scala
```diff
@@ -179,6 +178,20 @@ sealed abstract private[data] class ConstInstances4 {
     new ConstContravariant[C] {}
 }
 
+trait ConstInstancesBinCompat0 {
+  implicit def catsDataDecideableForConst[D: Monoid]: Decideable[Const[D, ?]] =
```
LukaJCB: I'm not super sure about this instance, here `product` and `sum` do the same thing, but IMO if we had `Semiring`s, it should be `*` and `+` respectively. Maybe we should remove this instance until we do?

laws/src/main/scala/cats/laws/DecideableLaws.scala
LukaJCB: For `ContravariantMonoidal` we already have left and right multiplicative identity as well as associativity.
So for `Decidable` I think we need :

1. `sum` and `zero` Left Identity
2. `sum` and `zero` Right Identity
3. `sum` Associativity
4. `product` and `zero` Right absorption
5. `sum` and `contramap` Right Distributivity
6. `sum` and `product` Right Distributivity

And then we should be good IMO :)

docs/src/main/tut/typeclasses/decideable.md
```diff
@@ -0,0 +1,61 @@
...
+  def decide[A, B, C](fa: F[A], fb: F[B])(f: C => Either[A, B]): F[C] =
+    contramap(sum(fa, fb))(f)
+}
+```
+
+`sum` lets us implement the `decide` function. Note how `decide` says, no matter what the outcome of a possibly multi-outcome procedure, I can consume either outcome.
+
+In other words, if we have two contexts `F[A]` and `F[B]` and we have a procedure `C => Either[A, B]` that could produce either an A or a B depending on the input, we have a way to produce a context `F[C]`.
+
+The classic example of a `Decideable` is that of predicates `A => Boolean`, leaning on the fact that both `&&` and `^` are valid monoids on `Boolean`.
```
LukaJCB: Another note on this one, the default `Rig` for `Boolean` uses `||` instead of `^`, should we use `||` here as well? https://github.com/typelevel/algebra/blob/master/core/src/main/scala/algebra/instances/boolean.scala

stephen-lazaro:
Thanks for the feedback @LukaJCB ! It may take me a day or two to incorporate, but good points all.

core/src/main/scala/cats/Decideable.scala
stephen-lazaro: Yeah, that was sort of my intuition as well. I'm a bit puzzled at the original's choice to use `Void` but, there's probably some reason. Regardless, will change.

core/src/main/scala/cats/data/Const.scala
stephen-lazaro: Yeah, you're right. Absolutely true, and changing `zero` as discussed above makes this issue more accute.

laws/src/main/scala/cats/laws/DecideableLaws.scala
stephen-lazaro: Ah I see how you'd write absorption I think now, yeah, sounds good.

core/src/main/scala/cats/Decideable.scala
```diff
@@ -0,0 +1,27 @@
...
+ * https://hackage.haskell.org/package/contravariant-1.4/docs/Data-Functor-Contravariant-Divisible.html#g:2
+ */
+@typeclass trait Decideable[F[_]] extends ContravariantMonoidal[F] {
+  def sum[A, B](fa: F[A], fb: F[B]): F[Either[A, B]]
+
+  def decide[A, B, C](fa: F[A], fb: F[B])(f: C => Either[A, B]): F[C] =
+    contramap(sum(fa, fb))(f)
+
+  def chosen[B, C](fb: F[B], fc: F[C]): F[Either[B, C]] =
+    decide(fb, fc)(identity[Either[B, C]])
```
LukaJCB: Why not just use `sum` here if it's just an alias?

core/src/main/scala/cats/Decideable.scala
catostrophe: See my comment in https://github.com/typelevel/cats/issues/2620#issuecomment-440459517

stephen-lazaro:
Just an update, due to work I haven't had much time to work on this, but I hope in the coming weekend to hammer out the remaining issues.

stephen-lazaro:
Finally had time to return to this, I'll be merging and resolving conflicts and incorporating any feedback I missed.

julienrf:
Why would `Decideable` be useful only on contravariant structures? I think it is orthogonal to variance. I personally use it with covariant and invariant structures. The minimal interface is:

~~~ scala
trait Decideable[F[_]] {
  def choose[A, B](fa: F[A], fb: F[B]): F[Either[A, B]]
}
~~~

[`Semigroupal`](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Semigroupal.scala) is another example of type class which is orthogonal to variance. Incidentally, `Decideable` is dual to `Semigroupal`.

kubukoz:
Hey @stephen-lazaro, do you think you'll have a chance to get back to this anytime soon? If not, I would like to take over, this looks really nice.

matil019:
Should this be named `Decidable` instead of `Decideable` (note the `e`)? I think it should have the same name as [Haskell's equivalent](https://hackage.haskell.org/package/contravariant-1.5.2/docs/Data-Functor-Contravariant-Divisible.html#t:Decidable) to avoid confusion.

stephen-lazaro:
@kubukoz I can definitely come back to this this weekend, I mostly just got distracted and was under the impression it might not be binary compatible.

build.sbt
```diff
@@ -815,9 +815,6 @@ def commonScalacOptions(scalaVersion: String, isDotty: Boolean) =
     "UTF-8",
     "-feature",
     "-unchecked",
-    //"-Ywarn-dead-code",
-    "-Ywarn-numeric-widen",
-    "-Ywarn-value-discard",
```
stephen-lazaro: Note to self, this looks a bit weird...

core/src/main/scala/cats/data/Const.scala
stephen-lazaro: By the way, do we think we'll have `Semiring` constraints available in the near future? Else I'll file an issue for it if we merge this.

laws/src/main/scala/cats/laws/DecideableLaws.scala
stephen-lazaro: Interestingly, having both left and right identities seems to rule out `OptionT` because you have to make _a_ choice for `None`, and similarly the sum and product distribution rules out `Order` and `Ordering` instances as you end up reversing orderings in the course of that, though I'm working to see if I can fix those instances to respect it.

stephen-lazaro:
Tests seem to pass for me locally, but I have to give quite a lot of memory and it still hangs, which is odd. Everything else except dotty compilation is a go, OTOH.

armanbilge:
I haven't looked in depth, but from the comments it seems like this PR is in nice shape and just needs to be revived :) let's see if we can land it in 2.9.0.

stephen-lazaro:
Cool, I merged and updated this a bit - I imagine there are a few things I'll have to fix trailing off of that, so I'll try and keep this polished and in approvable state in the near future.

armanbilge:
@stephen-lazaro nice, thank you so much for refreshing this! Yes, MiMa exceptions are now moved to `mima.sbt`, sorry about that. Let me know if I can help with anything!

build.sbt
```diff
@@ -1,3 +1,4 @@
+
```
armanbilge, line 1: Spurious newline.

core/src/main/scala/cats/Composed.scala
```diff
@@ -21,6 +21,8 @@
 
 package cats
 
+import cats.data.INothing
```
armanbilge, line 24: Is this still a thing? Seems in modern compilers `Nothing` may be working ok.

core/src/main/scala/cats/Decidable.scala
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * This is comparable to [[Alternative]] in the
+ * covariant case.
+ *
+ * Must obey laws in cats.laws.DecidableLaws.
+ *
+ * Based on ekmett's contravariant library:
+ * https://hackage.haskell.org/package/contravariant-1.4/docs/Data-Functor-Contravariant-Divisible.html#g:2
+ */
+@implicitNotFound("Could not find an instance of Decidable for ${F}")
```
armanbilge, line 19: We stopped using these, they were causing misleading compiler errors.

docs/datatypes/oneand.md
```diff
@@ -19,4 +19,4 @@ to represent other "non-empty" data structures e.g.
 
 ```scala mdoc:silent
 type NonEmptyStream[A] = OneAnd[Stream, A]
-```
\ No newline at end of file
+```
```
armanbilge, line 22: Huh, what's the diff here.

docs/typeclasses/decideable.md
```diff
@@ -0,0 +1,68 @@
+---
+layout: docs
+title: "Decidable"
+section: "typeclasses"
+source: "core/src/main/scala/cats/Decidable.scala"
+scaladoc: "#cats.Decidable"
+---
```
armanbilge, line 7: Yay docs! You can nix the header and also make sure to add the entry to the `directory.conf`.

docs/typeclasses/decideable.md
```diff
@@ -0,0 +1,68 @@
...
+# Decidable
+
+The `Decidable` type class is an extension of [`ContravariantMonoidal`](contravariantmonoidal.html) that allows the
+definition of a `sum` function, allowing us to make choices in consuming our input.
+
+It is the contravariant analogue of the [`Alternative`](alternative.html) type class.
+
+```tut:silent
```
armanbilge, line 15: `tut` -> `scala mdoc`. Hopefuly the modifiers are basically the same!

laws/src/main/scala/cats/laws/DecidableLaws.scala
```diff
@@ -0,0 +1,45 @@
+package cats
```
armanbilge, line 1: Run `headerCreateAll` to add the headers.

armanbilge (commented):
Did a drive-by with some pointers, hope it's helpful.

docs/typeclasses/decideable.md
stephen-lazaro: Cool, thanks, I'll update this, there's probably a lot of stuff like this where convention has changed.

laws/src/main/scala/cats/laws/DecidableLaws.scala
stephen-lazaro: Nice, thank you!

core/src/main/scala/cats/Decidable.scala
stephen-lazaro: Interesting! I guess I could see how that would happen with enough transitive chains of implicit resolution

core/src/main/scala/cats/Composed.scala
stephen-lazaro: I'm not honestly sure - originally I think this diff used normal `Nothing` and we switched to `INothing` for inference reasons. It may be simpler now than it was at that time.

build.sbt
stephen-lazaro: Ah yeah, residue of merge, will resolve, thanks.

core/src/main/scala/cats/Composed.scala
armanbilge: Context:
- https://github.com/typelevel/fs2/pull/2870

core/src/main/scala/cats/Decidable.scala
armanbilge: Context:
- https://github.com/typelevel/cats/issues/4059

core/src/main/scala/cats/Composed.scala
stephen-lazaro: Hmm I'm giving this a shot but I'm not sure it's possible for this changeset since a few of the instances then return `F[Nothing]` into invariant sites like `Tuple2K`. I'll keep playing with it but it might be a bit finicky

CONTRIBUTING.md
```diff
@@ -105,7 +105,7 @@ To build Cats you should have
  * `console`: launch a REPL
  * `test`: run the tests
  * `unidoc`: generate the documentation
- * `fmt`: run formatting of the code
+ * `scalafmt`: run formatting of the code
```
stephen-lazaro, line 108: Noticed that `fmt` does not work for me locally, but `scalafmt` does, so I went ahead and tossed an update here, can remove if unnecessary.

core/src/main/scala/cats/data/package.scala
```diff
@@ -142,4 +142,9 @@ package object data extends ScalaVersionSpecificPackage {
       RepresentableStoreT.pure[W, Function1[S, *], S, A](x)
 
   }
+
+  type INothing <: Nothing
```
johnynek, line 146: I think we need to remove this.

CONTRIBUTING.md
stephen-lazaro: Can do!

core/src/main/scala/cats/instances/eq.scala
```diff
@@ -22,8 +22,50 @@
...
+      def contramap[A, B](fa: Eq[A])(f: B => A): Eq[B] =
+        Eq.by(f)(fa)
+
+      def product[A, B](fa: Eq[A], fb: Eq[B]): Eq[(A, B)] =
+        Eq.instance { (left, right) =>
+          fa.eqv(left._1, right._1) && fb.eqv(left._2, right._2)
+        }
+
+      def sum[A, B](fa: Eq[A], fb: Eq[B]): Eq[Either[A, B]] =
+        Eq.instance { (left, right) =>
```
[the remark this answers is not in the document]
stephen-lazaro: Sure, I can switch to that.

core/src/main/scala/cats/Composed.scala
stephen-lazaro: Ah nevermind, just need quite a bit more type annotations. Will proceed

core/src/main/scala/cats/data/package.scala
stephen-lazaro: Removed.

core/src/main/scala/cats/Decidable.scala
```diff
@@ -0,0 +1,88 @@
...
+ * Based on ekmett's contravariant library:
+ * https://hackage.haskell.org/package/contravariant-1.4/docs/Data-Functor-Contravariant-Divisible.html#g:2
+ */
+trait Decidable[F[_]] extends ContravariantMonoidal[F] {
+  def sum[A, B](fa: F[A], fb: F[B]): F[Either[A, B]]
+
+  def decide[A, B, C](fa: F[A], fb: F[B])(f: C => Either[A, B]): F[C] =
+    contramap(sum(fa, fb))(f)
+
+  def chosen[B, C](fb: F[B], fc: F[C]): F[Either[B, C]] = sum(fb, fc)
```
johnynek, line 42: this is just a pure alias right? Do we need it? If we really need it can it be `final` so we don't have any issue where someone overrides them incorrectly?

core/src/main/scala/cats/Decidable.scala
```diff
@@ -0,0 +1,88 @@
...
+  def lose[A](f: A => Nothing): F[A] = contramap[Nothing, A](zero)(f)
+  def zero: F[Nothing]
```
johnynek, line 44: can we move this up to be under `sum` since `sum` and `zero` are the only two things we need to override (keep the abstract ones at the top).

core/src/main/scala/cats/Invariant.scala
```diff
@@ -176,10 +176,6 @@ object Invariant extends ScalaVersionSpecificInvariantInstances with InvariantIn
     cats.instances.ordering.catsContravariantMonoidalForOrdering
   implicit def catsContravariantMonoidalForPartialOrdering: ContravariantMonoidal[PartialOrdering] =
     cats.instances.partialOrdering.catsContravariantMonoidalForPartialOrdering
-  implicit def catsContravariantMonoidalForEq: ContravariantMonoidal[Eq] =
```
johnynek, line 179: I think this binary incompatible. Can we add the methods back, but remove the `implicit`? I think that is binary compatible.

core/src/main/scala/cats/data/package.scala
```diff
@@ -142,4 +142,6 @@ package object data extends ScalaVersionSpecificPackage {
       RepresentableStoreT.pure[W, Function1[S, *], S, A](x)
 
   }
+
+  def absurd[A](n: Nothing): A = throw new Exception("Somehow instantiated a value of INothing, this is a bug.")
```
johnynek, line 146: you don't need to `throw`. since `Nothing <: A` you can just return `n`:

```scala
def absurd[A](n: Nothing): A = n
```

also, the name `INothing` has persisted.

core/src/main/scala/cats/instances/equiv.scala
```diff
@@ -22,8 +22,51 @@
...
+      def sum[A, B](fa: Equiv[A], fb: Equiv[B]): Equiv[Either[A, B]] =
+        new Equiv[Either[A, B]] {
+          def equiv(x: Either[A, B], y: Either[A, B]): Boolean = (x, y) match {
+            case (Left(a1), Left(a2))   => fa.equiv(a1, a2)
+            case (Right(b1), Right(b2)) => fb.equiv(b1, b2)
+            case _                      => false
+          }
+        }
+
+      override def zero: Equiv[Nothing] = Equiv.by[Nothing, Unit](_ => ())
```
johnynek, line 64: can make this a `val` to avoid reallocating...

core/src/main/scala/cats/instances/function.scala
```diff
@@ -80,6 +79,19 @@ private[instances] trait FunctionInstancesBinCompat0 {
         Deferred(() => cachedFa)
       }
     }
+
+  implicit def catsStdDecidableForPredicate: Decidable[* => Boolean] =
+    new Decidable[* => Boolean] {
+      def unit: Unit => Boolean = Function.const(false)
```
johnynek, line 85: can be ` val`.

core/src/main/scala/cats/instances/function.scala
```diff
@@ -80,6 +79,19 @@ private[instances] trait FunctionInstancesBinCompat0 {
...
+      def contramap[A, B](fa: A => Boolean)(f: B => A): B => Boolean =
+        fa.compose(f)
+      def product[A, B](fa: A => Boolean, fb: B => Boolean): ((A, B)) => Boolean = { case (a, b) =>
+        fa(a) || fb(b)
+      }
+      def sum[A, B](fa: A => Boolean, fb: B => Boolean): Either[A, B] => Boolean =
+        either => either.fold(fa, fb)
+      def zero: Nothing => Boolean = _ => true
```
johnynek, line 93: can be a `val`.

core/src/main/scala/cats/ContravariantMonoidal.scala
```diff
@@ -38,7 +38,7 @@ trait ContravariantMonoidal[F[_]] extends ContravariantSemigroupal[F] with Invar
    * the diagonal
    */
   def trivial[A]: F[A] = contramap(unit)(_ => ())
-
+  def conquer[A]: F[A] = trivial[A]
```
johnynek, line 41: why do we need this alias? Can we remove it?

core/src/main/scala/cats/instances/order.scala
```diff
@@ -25,7 +25,6 @@ package instances
 import cats.kernel.instances.unit._
 
 trait OrderInstances extends kernel.instances.OrderInstances {
-
   implicit val catsContravariantMonoidalForOrder: ContravariantMonoidal[Order] =
```
johnynek, line 28: can we not do `Decidable[Order]`, `Decideable[Hash]` and `Decidable[Ordering]`? Seems doable (well, maybe Hash isn't totally clear, not sure).

Would be interesting to grep for `ContravariantMonoidal` and see how many instances are actually Decidable.

stephen-lazaro: We actually had `Order` and `Ordering` at one point, I believe I removed them based on feedback from a prior iteration, but those concerns may be gone now. I am happy to re-add them. (In practice almost all `ContravariantMonoidal` that are not derived from a base `ContravariantMonoidal` are `Decidable` it appears to me, but I'm sure in the wild there are more cases where these come apart than in the base).

core/src/main/scala/cats/ContravariantMonoidal.scala
stephen-lazaro: This is here primarily in deference to those Haskell users used to the original text - in that text `conquer` figures as a somewhat more primary combinator than in ours because they lack the `Semiringal` hierarchy that we have in this library.

core/src/main/scala/cats/instances/equiv.scala
stephen-lazaro: Good call, I'll give the `def`s a pass and make `val` where appropriate.

core/src/main/scala/cats/Decidable.scala
stephen-lazaro: Certainly - I think this like `conquer` exists in deference to the original text and the conventions users of that will expect.

core/src/main/scala/cats/Invariant.scala
stephen-lazaro: Yeah, let me take a look. That is likely possible.

core/src/main/scala/cats/data/package.scala
stephen-lazaro: I'll likely simply delete this function - I believe I only use something like this in one place so it is probably no longer necessary here.

docs/typeclasses/decidable.md
```diff
@@ -0,0 +1,61 @@
...
+The `Decidable` type class is an extension of [`ContravariantMonoidal`](contravariantmonoidal.html) that allows the
+definition of a `sum` function, allowing us to make choices in consuming our input.
+
+It is the contravariant analogue of the [`Alternative`](alternative.html) type class.
+
+```scala mdoc:silent
+import cats.ContravariantMonoidal
+import cats.data.INothing
+
+trait Decidable[F[_]] extends ContravariantMonoidal[F] {
```
johnynek, line 12: I think a good observation might be if you have a `Decidable[F]` then in scala 3 you can derive instances for `F[_]` for enums.

It might even be cool to have some scala 3 specific code that does this.

core/src/main/scala/cats/Composed.scala
```diff
@@ -176,6 +176,18 @@ private[cats] trait ComposedContravariantCovariant[F[_], G[_]] extends Contravar
...
 }
 
+private[cats] trait ComposedApplicativeDecidable[F[_], G[_]]
+    extends Decidable[λ[α => F[G[α]]]]
+    with ComposedApplicativeContravariantMonoidal[F, G] { outer =>
+  def F: Applicative[F]
+  def G: Decidable[G]
+
+  def sum[A, B](fa: F[G[A]], fb: F[G[B]]): F[G[Either[A, B]]] =
+    F.map(F.product(fa, fb))(Function.tupled(G.sum))
```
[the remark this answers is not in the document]
johnynek: Also I think it is causing a compilation error:

```
 [info] Fast optimizing /home/runner/work/cats/cats/kernel/.js/target/scala-3.1.3/cats-kernel-test-fastopt
[error] -- [E007] Type Mismatch Error: /home/runner/work/cats/cats/core/src/main/scala/cats/Composed.scala:186:45 
[error] 186 |    F.map(F.product(fa, fb))(Function.tupled(G.sum))
[error]     |                                             ^^^^^
[error]     |Found:    (G[Any], G[Any]) => G[Either[Any, Any]]
[error]     |Required: (G[A], G[B]) => G[Either[A, B]]
[error]     |
[error]     |where:    G is a type in traitComposedApplicativeDecidable with bounds <: [_] =>> Any
[error]     |
[error]     | longer explanation available when compiling with `-explain`
```

docs/typeclasses/decidable.md
stephen-lazaro: That's a good point, I can look into adding some. If not in this diff then maybe in a follow up, at least.

core/src/main/scala/cats/data/Op.scala
```diff
@@ -67,6 +87,39 @@ private[data] trait OpCompose[Arr[_, _]] extends Compose[Op[Arr, *, *]] {
...
+private[data] trait OpDecidable[Arr[_, _], R] extends Decidable[Op[Arr, R, *]] with OpContravariantMonoidal[Arr, R] {
+  implicit def Arr: ArrowChoice[Arr]
+  implicit def M: Monoid[R]
+
+  def sum[A, B](fa: Op[Arr, R, A], fb: Op[Arr, R, B]): Op[Arr, R, Either[A, B]] =
+    Op(Arr.choice(fa.run, fb.run))
+
+  override def zero: Op[Arr, R, Nothing] =
+    Op[Arr, R, Nothing](
+      Arr.lift[Nothing, R](_ => throw new Exception("Materialized a value of type `Nothing`, this is a bug!"))
```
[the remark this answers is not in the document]
stephen-lazaro: Interesting, I'm not sure why but it surprises me that that works. `Nothing` is a bit odd.
I think I had `absurd` and the `throw` generally originally to have something to put that would step around having to produce something from nothing.

core/src/main/scala/cats/ContravariantMonoidal.scala
stephen-lazaro: Got it, yeah I think there was a fair amount of discussion about whether to keep these names in earlier versions of the diff as well, and some folks explicitly wanted the aliases so they had familiar names. I can see the perspective you're putting forward here though as well, so I'll go ahead and nix these aliases.

docs/typeclasses/decidable.md
kubukoz: Let's do it separately, this PR has been open for way too long anyway ;)

core/src/main/scala/cats/Composed.scala
```diff
@@ -176,6 +176,18 @@ private[cats] trait ComposedContravariantCovariant[F[_], G[_]] extends Contravar
...
+private[cats] trait ComposedApplicativeDecidable[F[_], G[_]]
+    extends Decidable[λ[α => F[G[α]]]]
+    with ComposedApplicativeContravariantMonoidal[F, G] { outer =>
+  def F: Applicative[F]
+  def G: Decidable[G]
+
+  def sum[A, B](fa: F[G[A]], fb: F[G[B]]): F[G[Either[A, B]]] =
+    F.map2(fa, fb)(G.sum)
+
+  override val zero: F[G[Nothing]] = F.pure(G.zero)
```
johnynek, line 188: I think I may have steered you wrong and this may be triggering the val init order bug. Try changing to `lazy val` or `def` and see if it fixes:

```
   ==> X cats.tests.NestedSuite.initializationError0.001s java.lang.NullPointerException: Cannot invoke "cats.Decidable.zero()" because the return value of "cats.data.NestedDecidable.FG()" is null
      at cats.data.NestedDecidable.$init$(Nested.scala:411)
      at cats.data.NestedInstances1$$anon$10.<init>(Nested.scala:135)
      at cats.data.NestedInstances1.catsDataDecidableForApplicativeForNested(Nested.scala:135)
      at cats.tests.NestedSuite.<init>(NestedSuite.scala:144)
```

core/src/main/scala/cats/instances/order.scala
stephen-lazaro: Ah I remember now what the deal with `Order` and `Ordering` does - the natural lexicographic approach to ordering the eithers causes the instance to fail the `right sum distributivity` law we've defined in this changeset. I'll probably leave them out since I don't have a better idea on how that would be defined, but maybe `Hash` is more fortunate.

core/src/main/scala/cats/Composed.scala
stephen-lazaro: Nice, thank you.

stephen-lazaro:
I believe all checks pass and I've managed to get existing feedback - please let me know if I've overlooked anything. This is my last day of vacation ;)

armanbilge:
Thanks for all your work on this!!

mima.sbt
```diff
@@ -145,6 +145,10 @@ ThisBuild / mimaBinaryIssueFilters ++= {
       exclude[ReversedAbstractMethodProblem]("cats.free.ContravariantCoyoneda.k"),
       exclude[DirectAbstractMethodProblem]("cats.free.Coyoneda.k"),
       exclude[ReversedAbstractMethodProblem]("cats.free.Coyoneda.k")
+    ) ++ // https://typelevel/cats/pull/2607
+    Seq(
+      exclude[DirectMissingMethodProblem]("cats.instances.EqInstances.$init$"),
+      exclude[DirectMissingMethodProblem]("cats.instances.EquivInstances.$init$")
     )
```
armanbilge, line 152: So these are actually legit bincompatibility breaks, that we ideally should not exclude. Although Scala 3 only. I am thinking about how we might best workaround these. The issue is that we removed the `val` from `EqInstances`.

tests/shared/src/test/scala/cats/tests/OpSuite.scala
```diff
@@ -44,13 +44,14 @@ class OpSuite extends CatsSuite {
...
-    implicit val catsDataDecidableForOp = Op.catsDataDecidableForOp[Function1, Int]
+    implicit val catsDataDecidableForOp: Decidable[Op[Function1, Int, *]] = Op.catsDataDecidableForOp[Function1, Int]
     checkAll("Op[Function1, Monoid, ?]", DecidableTests[Op[Function1, Int, *]].decidable[MiniInt, MiniInt, MiniInt])
     checkAll("Decidable[Op[Function1, Monoid, ?]]", SerializableTests.serializable(Decidable[Op[Function1, Int, *]]))
   }
 
   {
-    implicit val catsDataContravariantMonoidalForOp = Op.catsDataContravariantMonoidalForOp[Function1, Int]
+    implicit val catsDataContravariantMonoidalForOp: ContravariantMonoidal[Op[Function1, Int, *]] =
+      Op.catsDataContravariantMonoidalForOp[Function1, Int]
```
armanbilge, line 54: If you'll forgive me, just taking a moment to inform about cross-platform Scala for any wider audience :) nothing personal!

The commit comment says:

> Fix JS linking in tests 

Actually, you are fixing vanilla Scala 3 compiler errors here (the need for explicit types on implicits). In CI sometimes the JS job happens to fail first, but that doesn't mean that the problem is specfiic to JS :)

In general, the compiler frontends for Scala JVM, JS, and Native are the same—it's only the backends that produce different output (bytecode vs javascript vs native binaries). So compile-time quirks like these related to Scala features (types, implicits, etc.) are usually general Scala things and not platform specific.

core/src/main/scala/cats/instances/eq.scala
```diff
@@ -22,8 +22,42 @@
 package cats
 package instances
 
-trait EqInstances extends kernel.instances.EqInstances {
-  implicit val catsContravariantMonoidalForEq: ContravariantMonoidal[Eq] =
+trait EqInstances extends kernel.instances.EqInstances with EqInstances0 {
```
armanbilge, line 25:
```suggestion
trait EqInstances extends kernel.instances.EqInstances with EqInstances0 {

  val catsContravariantMonoidalForEq: ContravariantMonoidal[Eq] =
    catsContravariantMonoidalForEqBinCompat

```

core/src/main/scala/cats/instances/eq.scala
```diff
@@ -22,8 +22,42 @@
...
+          case (Right(b1), Right(b2)) => fb.eqv(b1, b2)
+          case _                      => false
+        }
+
+      override val zero: Eq[Nothing] = Eq.allEqual[Nothing]
+    }
+}
+
+private[instances] trait EqInstances0 {
+  implicit def catsContravariantMonoidalForEq: ContravariantMonoidal[Eq] =
```
armanbilge, line 60:
```suggestion
  implicit def catsContravariantMonoidalForEqBinCompat: ContravariantMonoidal[Eq] =
```

armanbilge (commented):
Ok, this should fix the bincompat issue I think, so you can remove the mima filters. Same for `EquivInstances`.

mima.sbt
armanbilge: Proposed a solution in https://github.com/typelevel/cats/pull/2607#pullrequestreview-1026674706.

stephen-lazaro: Yeah, I'm afraid I rely on the guidance of the maintainers on these - I simply don't work with bincompat enough to have any intuition.

tests/shared/src/test/scala/cats/tests/OpSuite.scala
stephen-lazaro: I appreciate this a lot actually! I don't have opportunities to use Scala 3 or Scala JS so I did not know the actual ground of this.

stephen-lazaro:
I do find myself missing the `validate` command. ;)

armanbilge:
Ah, it's been replaced by `prePR` :)

mima.sbt
armanbilge: No worries! Bincompat can be very counter-intuitive even for experienced maintainers. I'm just glad we could fix this one, thanks!

core/src/main/scala/cats/instances/eq.scala
```diff
@@ -22,8 +22,45 @@
...
+
+      override val zero: Eq[Nothing] = Eq.allEqual[Nothing]
+    }
+
+  val catsContravariantMonoidalForEq: ContravariantMonoidal[Eq] =
+    catsContravariantMonoidalForEqForBinCompat
+}
+
+private[instances] trait EqInstances0 {
+  implicit def catsContravariantMonoidalForEqForBinCompat: ContravariantMonoidal[Eq] =
```
armanbilge, line 63: Actually ... I don't think we need this at all 😅 this is completely subsumed by the new `Decidable` instance :)

core/src/main/scala/cats/instances/eq.scala
```diff
@@ -22,8 +22,45 @@
...
+          case (Left(a1), Left(a2))   => fa.eqv(a1, a2)
+          case (Right(b1), Right(b2)) => fb.eqv(b1, b2)
+          case _                      => false
+        }
+
+      override val zero: Eq[Nothing] = Eq.allEqual[Nothing]
+    }
+
+  val catsContravariantMonoidalForEq: ContravariantMonoidal[Eq] =
+    catsContravariantMonoidalForEqForBinCompat
```
armanbilge, line 59:
```suggestion
  val catsContravariantMonoidalForEq: ContravariantMonoidal[Eq] =
    catsDecidableForEq
```

core/src/main/scala/cats/instances/equiv.scala
```diff
@@ -51,5 +51,51 @@ trait EquivInstances {
...
+
+      override val zero: Equiv[Nothing] = Equiv.by[Nothing, Unit](_ => ())
+    }
+
+  val catsContravariantMonoidalForEquiv: ContravariantMonoidal[Equiv] =
+    catsContravariantMonoidalForEquivForBinCompat
+}
+
+private[instances] trait EquivInstances0 {
+  implicit def catsContravariantMonoidalForEquivForBinCompat: ContravariantMonoidal[Equiv] =
```
armanbilge, line 72: Same comments here :)

tests/shared/src/test/scala/cats/tests/Predicate.scala
```diff
@@ -0,0 +1,49 @@
...
+
+import cats.Decidable
+import cats.kernel.Eq
+import cats.laws.discipline.eq._
+import cats.laws.discipline.ExhaustiveCheck
+import org.scalacheck.{Arbitrary, Cogen}
+
+case class Predicate[A](run: A => Boolean)
+object Predicate {
+  implicit val decidablePredicate: Decidable[Predicate] =
```
armanbilge, line 32: Can we implement this by delegating to the instance in Cats itself?
https://github.com/typelevel/cats/blob/7730debb3209aa4a28db92f01dd7e8ff943d33c8/core/src/main/scala/cats/instances/function.scala#L83

Or are we testing that instance elsewhere?

tests/shared/src/test/scala/cats/tests/KleisliSuite.scala
```diff
@@ -127,12 +127,29 @@ class KleisliSuite extends CatsSuite {
...
+      "Decidable[Kleisli[Option, Int, *]]",
+      SerializableTests.serializable[Decidable[Kleisli[Predicate, MiniInt, *]]](
+        Decidable[Kleisli[Predicate, MiniInt, *]]
+      )
+    )
+  }
+
+  {
+    implicit val catsDataContravariantMonoidalForKleisliLocal
+      : ContravariantMonoidal[Kleisli[Const[String, *], Int, *]] =
```
armanbilge, line 144: Why does this `implicit` need to be scoped like this, is there some sort of ambiguity issue?

tests/shared/src/test/scala/cats/tests/OpSuite.scala
```diff
@@ -42,6 +43,36 @@ class OpSuite extends CatsSuite {
     checkAll("Category[Op[Function1, *, *]]", SerializableTests.serializable(Category[Op[Function1, *, *]]))
   }
 
+  {
+    implicit val catsDataDecidableForOp: Decidable[Op[Function1, Int, *]] = Op.catsDataDecidableForOp[Function1, Int]
```
armanbilge, line 47: Same question here about ambiguities. It would be good if the tests can verify that it can be disambiguated in practice without needing help.

core/src/main/scala/cats/Composed.scala
```diff
@@ -176,6 +176,18 @@ private[cats] trait ComposedContravariantCovariant[F[_], G[_]] extends Contravar
     F.contramap(fga)(gb => G.map(gb)(f))
 }
 
+private[cats] trait ComposedApplicativeDecidable[F[_], G[_]]
```
armanbilge, line 179: Sorry if I missed it, is this being tested anywhere?

armanbilge: Ah I see, it's used to implement `Nested` and is tested via that. Cool 👍

armanbilge (commented):
Wow, this is a big PR. Thank you so much for sticking with it!

Did my best to do a detailed pass-through, apologies for any stupid comments 😅 overall it looks really good.

I didn't make an attempt to study the laws yet, since that requires real brainpower ... I need to dig into the old reviews, but I'm hopeful they've already been thought through by smarter folks!

core/src/main/scala/cats/Decidable.scala
```diff
@@ -0,0 +1,86 @@
...
+
+object Decidable {
+
+  /**
+   * Summon an instance of [[Decidable]] for `F`.
+   */
+  @inline def apply[F[_]](implicit instance: Decidable[F]): Decidable[F] = instance
+
+  @deprecated("Use cats.syntax object imports", "2.2.0")
+  object ops {
```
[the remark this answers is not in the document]
armanbilge: Oh also, we should add some tests somewhere to check the syntax is working. I think there's a suite for syntax tests.

core/src/main/scala/cats/Invariant.scala
stephen-lazaro: I opted for version `since 2.9.0` based on the tag I saw, let me know if another would be better.

core/src/main/scala/cats/data/OptionT.scala
```diff
@@ -553,11 +552,6 @@ sealed abstract private[data] class OptionTInstances0 extends OptionTInstances1
       implicit val F = F0
     }
 
-  implicit def catsDataContravariantMonoidalForOptionT[F[_]](implicit
-    F0: ContravariantMonoidal[F]
-  ): ContravariantMonoidal[OptionT[F, *]] =
-    new OptionTContravariantMonoidal[F] { implicit val F = F0 }
```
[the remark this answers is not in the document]
stephen-lazaro: Ah what happened here is I actually _added_ a `Decidable` but it turns out to be laws violating for similar reasons to `Order` and `Ordering`. Given our law of `right sum distributivity` you end up having to make a choice that then can't be consistent across distributions.
I'll move the instance back just for clarity of this diff.

tests/shared/src/test/scala/cats/tests/KleisliSuite.scala
stephen-lazaro: Doesn't look like it needs to be - probably just cargo culting when I wrote it.

tests/shared/src/test/scala/cats/tests/Predicate.scala
stephen-lazaro: Good call, I'll definite it via delegation!

stephen-lazaro: Really that instance in `instances.function` should have a `Ring` context bound if not `Rig` but I don't think `core` depends on `algebra` does it?

armanbilge: Hmm, I'm afraid not, `core` only depends on `kernel`. Why does it need to be `Ring`/`Rig`, how should it actually look like? If we're not doing it right maybe we shouldn't do it at all 😕

stephen-lazaro: I mean it's a fine instance and it's useful, but here's a way more generic one we could have, if I remember right:
```scala
implicit def catsStdDecidableForRigValuedFunctions[C: Rig] = 
  new Decidable[* => C] {
    def unit: * => C = Function.const(Rig[C].one)
    def zero: Nothing => A = Function.const(Rig[C].zero)
    def sum(fa: A => C, fb: B => C): Either[A, B]] => C = 
       _.fold(fa, fb)
     def product(fa: A => C, fb: B => C): (A, B) => C =
       t => Rig[C].combine(fa(t._1), fb(t._2))
  }
```
Which ought to generalize this one based on the `Rig` for `Boolean`, but it's not huge, it would provide a few more instances "for free".
So this instance is totally legitimate but we it's kind of the shadow of this other more fundamental instance.

armanbilge: Oh I see, that is quite nice. Maybe one day ring-like stuff can find its way into Cats proper.

armanbilge (commented):
Thanks for getting through my previous review, now I'll try and turn my attention to the laws 😅

armanbilge:
I see there's some discussion on laws in:
- https://github.com/ekmett/contravariant/issues/53
- https://github.com/scalaz/scalaz/pull/1881

laws/src/main/scala/cats/laws/DecidableLaws.scala
```diff
@@ -0,0 +1,64 @@
...
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package cats
+package laws
+
+trait DecidableLaws[F[_]] extends ContravariantMonoidalLaws[F] {
```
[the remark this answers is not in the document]
stephen-lazaro: I can add some on those certainly, I think keeping the identity laws is a good idea though because they nicely capture that `decide` should be sort of monoid-like with respect to the trivial value.

stephen-lazaro:
Yeah, I'd read that Kmett thread back when opening this - it's true that it's a bit difficult to say what exactly the right choices are. I believe we vaguely our choices here largely on the principle of lifting `Rig` laws when we were doing this a few years ago.
Luka also suggested some options here IIRC: https://github.com/typelevel/cats/pull/2607#discussion_r232940785

stephen-lazaro:
It gets pretty difficult to actually run some of the laws even once they are expressed, because generating things with `Nothing` in their type gets pretty hard, and much worse when the law can be run but there is no view to `Prop` from `IsEq[F[Nothing]]`.

laws/src/main/scala/cats/laws/discipline/DecidableTests.scala
```diff
@@ -46,12 +46,18 @@ trait DecidableTests[F[_]] extends ContravariantMonoidalTests[F] {
     iso: SemigroupalTests.Isomorphisms[F]
   ): RuleSet =
     new RuleSet {
+      implicit val ArbNothing: Arbitrary[Nothing => Nothing] = Arbitrary(Gen.const((n: Nothing) => n))
+      implicit val eqNothing: Eq[F[Nothing]] = Eq.by[F[Nothing], Unit]((_: F[Nothing]) => ())
```
armanbilge, line 50: Ha these are rigged 😂 actually I think `Eq[F[Nothing]]` needs be an implicit constraint of the tests, because we don't know how equality of `F[_]` is supposed to work. Then, maybe we can add `Eq[Nothing]` to Cats?

stephen-lazaro: Sounds good hah. I went this route mainly because scoping these sort of bizarre instances tightly seems least risky, but yeah I see your point on the constraint having to be user accessible for laws testing new instances.

stephen-lazaro: I do want to call out - just looking at this finally today, this is going to add some pretty weird things if we want to run these tests against the cats instances, e.g. `Eq[Eq[Nothing]]` is becomes a requirement... These feel like they should be test scoped, but I don't know if we have a standard way of doing that? Maybe you can advice @armanbilge

armanbilge: `Eq[Eq[Nothing]]` seems fine (should always return `true` I reckon), although I agree it should be scoped. I think it's fine to put it directly in the test suite for now. I can't remember where these things live, I'm sure we have some instances for `Invariant` tests or something.

tests/shared/src/test/scala/cats/tests/NestedSuite.scala
```diff
@@ -127,11 +127,25 @@ class NestedSuite extends CatsSuite {
...
+    )
+    checkAll(
+      "ContravariantMonoidal[Nested[Option, Const[String, *], *]",
+      SerializableTests.serializable(ContravariantMonoidal[λ[α => Nested[Option, Const[String, *], α]]])
+    )
+  }
+
+  {
+    // Applicative + Decidable functor composition
+    implicit val eqForOptionPredicateNothing: Eq[Nested[Option, Predicate, Nothing]] = Eq.allEqual
```
stephen-lazaro, line 143: Experimenting with this for now to see if it passes remotely (not all tests run for me locally [maybe it's my M1?]).
Anyway, it seems not quite right to me, the right instance should be `Eq.by(_.value.isDefined)` or something similar but that does not seem to pass.

armanbilge: It's weird, but I guess you can't have `Some[Nothing]`, so the only valid instance of `Option[Nothing]` is `None`, which is a singleton?

stephen-lazaro: Yeeeahhhhhh it's kind of "extensionally cool" basically since `Predicate[Nothing]` is essentially `Nothing`. Well, I'll leave it for now, anything stronger is just incompatible with the instance, and the instance makes more sense than the law does really.

armanbilge: Oh, right, `Eq.by(_.value.isDefined)` does seem right. CI seems to be passing, does that not pass?

stephen-lazaro: I've removed the `right absorption` law. I can't see any way to make it consistent with instances that are reasonable. It's also not a law in any other library implementing this I know of, and the set we have seems comparatively strong.
If it doesn't meet our standards that's alright - maybe someone else will be able to find a reasonable extension to this.

stephen-lazaro:
@armanbilge This is probably as complete as it's going to get at this stage - let me know if there are thoughts on it.

armanbilge:
Ah, great, thanks for the bump and all your work! I will try to take a look this weekend.

core/src/main/scala/cats/Decidable.scala
```diff
@@ -0,0 +1,73 @@
...
+ */
+trait Decidable[F[_]] extends ContravariantMonoidal[F] {
+  def sum[A, B](fa: F[A], fb: F[B]): F[Either[A, B]]
+  def zero: F[Nothing]
+
+  def decide[A, B, C](fa: F[A], fb: F[B])(f: C => Either[A, B]): F[C] =
+    contramap(sum(fa, fb))(f)
+
+  def lose[A](f: A => Nothing): F[A] = contramap[Nothing, A](zero)(f)
+}
```
armanbilge, line 44: Ok I realize this is a bit late in the day, but wouldn't it make more sense to define it like this? Basically trade two abstract defs (`contramap`+`sum`) for one (`decide`).

```scala
// abstract
def decide[A, B, C](fa: F[A], fb: F[B])(f: C => Either[A, B]): F[C]

// concrete
def contramap[A, B](fa: F[A])(f: B => A): F[B] =
  decide(fa, zero)(b => Left(f(b)))
def sum[A, B](fa: F[A], fb: F[B]): F[Either[A, B]] =
  decide(fa, fb)(identity)
```

armanbilge: Based on principle of power, I suspect that this similar adjustment should also be made.
```
// abstract
def lose[A](f: A => Nothing): F[A]

// concrete
def zero: F[Nothing] = lose(identity)
```

armanbilge (commented):
Sorry that it took me _forever_ to cycle back, had to work up the courage to face this PR again 😅 

Just a couple comments for now, but will slowly be combing through this. Really want to land this in the next Cats release.

~~Also can we add a `Decidable[Show]` instance?~~ Nopeee 😂 ignore my inane comment.

armanbilge:
Does anyone else have thoughts on my latest comments? Specifically https://github.com/typelevel/cats/pull/2607#discussion_r993029098. I am happy to try and push this across the finish line if Stephen is busy atm.

This is phenomenal work, and I'd really like to land it before another 4 years pass 😄

stephen-lazaro:
Apologies! I have been pre-occupied with my wedding. Seeing that there are comments here, I will endeavor to resolve them this weekend. If there is other feedback, I would appreciate having that up front so I can do them all in a few iterations, that said if it takes a few I understand.

core/src/main/scala/cats/Decidable.scala
stephen-lazaro: I'm fine with this! I think this actually better fits our current modeling with a monoidal / semigroupal hierarchy. This encoding I think predates some of the rest of that work and was originally part of that push.

core/src/main/scala/cats/instances/eq.scala
```diff
@@ -44,5 +44,17 @@ trait EqInstances extends kernel.instances.EqInstances {
         Eq.instance { (left, right) =>
           fa.eqv(left._1, right._1) && fb.eqv(left._2, right._2)
         }
+
+      def sum[A, B](fa: Eq[A], fb: Eq[B]): Eq[Either[A, B]] =
+        Eq.instance {
+          case (Left(a1), Left(a2))   => fa.eqv(a1, a2)
+          case (Right(b1), Right(b2)) => fb.eqv(b1, b2)
+          case _                      => false
+        }
```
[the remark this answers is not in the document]
stephen-lazaro: Good call! Thank you.

MavenRain:
@stephen-lazaro @armanbilge: half of #1935 landed long ago as
   `ContravariantMonoidal` (#2034), so this PR is really the Decidable half.
  
   If neither of you is actively driving it now, would you mind if I picked it
   up?  I would rebase #2607 onto current main, work through the open review
   threads, and either continue here or open a fresh PR preserving your commit
   history.

stephen-lazaro:
I personally have no issues with you taking this over if you want - my work and school commitments will likely keep me from having time for this in the foreseeable future. If you want to talk about the state of the feature or my memory of it in the course of that feel empowered to reach out on any of my public advertised communication channels.

PR #2609, opened 2018-11-13
Add an If-Then-Else operation to the Apply.

diesalbla:
The `FlatMap` type-class includes a function `ifM`, which allows evaluating a Boolean computation and, depending on the result, choosing between one effectful computation or other.

We add to the `Apply` type-class a similar function, `ifA`, which 1) takes as inputs an `F[Boolean]` and two `F[A]`, and 2) depending on the value of the boolean result, returns either the first or the second `F[A]` on the result. 

Although the profile is similar to `ifM`, they need to be separate functions due to the different way they handle the effects: 
- The `ifM` performs the effects of the condition `F[Boolean]`, and the effects of _one_ of the branches. Which branch, depends on the boolean result. 
- The `ifA` will always perform the effect of the `F[Boolean]`, those of the first `F[A]`, and those of the second `F[A]`, in that order. It will apply the `if-then-else` selection to the values.

In other words, `ifA` only maps the values, but does not allow skipping effects.


core/src/main/scala/cats/syntax/apply.scala
```diff
@@ -14,6 +14,34 @@ trait ApplySyntax extends TupleSemigroupalSyntax {
...
+   * scala> val b2: Option[Boolean] = Some(false)
+   * scala> val asInt2: Option[Int] = b2.ifA(Some(1), Some(0))
+   * scala> asInt2.get
+   * res1: Int = 0
+   *
+   * scala> val b3: Option[Boolean] = Some(true)
+   * scala> val asInt3: Option[Int] = b3.ifA(Some(1), None)
+   * asInt2: Option[Int] = None
+   *
+   * }}}
```
kailuowang: Unfortunately we can't add any methods to type classses without breaking binary compatibility. So for now, we have to settle adding to a new syntax trait. Sorry I am on my phone and couldn't point you to an example.

diesalbla: @kailuowang If I understand you correctly, we can not add the method to the `trait Apply[F[_]]` (the type-class) but can only leave it as part of a syntactic extension `implicit class`?

kailuowang: sorry for the delayed response. Yes, that's the idea. You already updated the PR, we also need to create a new `ApplySyntaxBinCompat` trait to include an implicit conversion from `F[Boolean]` to `IfAOps`, then there are a couple of more steps, 
1.  [here](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/syntax/all.scala) we need to add a new trait `AllSyntaxBinCompat4` to extends the new `ApplySyntaxBinCompat`,
 let the `AllSyntaxBinCompat` at the top extend this `AllSyntaxBinCompat4`
2. [here] https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/implicits.scala we also need to let the `implicits` object extends this new `AllSyntaxBinCompat4`
3. then we can probably add a doctest here that `import cats.implicits._` to validate that it's all hooked up. 

Sorry about the gymnastics you have to go through to overcome the Scala 2.11 binary compatibility restrictions.

LukaJCB:
Thanks @diesalbla!

PR #2610, opened 2018-11-13
Expose headOption for Chain

ChristopherDavenport:

This is a kind reminder to run `sbt prePR` and commit the changed files before submitting.

### DIFF
--- a/tests/src/test/scala/cats/tests/ChainSuite.scala
+++ b/tests/src/test/scala/cats/tests/ChainSuite.scala
@@ -57,7 +57,7 @@ class ChainSuite extends CatsSuite {
   }
 
   test("headOption") {
-    forAll { (s: Seq[Int])  =>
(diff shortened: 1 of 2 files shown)


tests/src/test/scala/cats/tests/ChainSuite.scala
```diff
@@ -56,6 +56,12 @@ class ChainSuite extends CatsSuite {
     }
   }
 
+  test("headOption") {
```
kailuowang: nitpick: this doesn't test the None scenario. How about 

```scala
forAll { (s: Seq[Int]) =>
      Chain.fromSeq(s).headOption should ===(s.headOption)
    }
```

ChristopherDavenport: Addressed. Thanks.

Issue #2613, opened 2018-11-15
Breaking change: NestedSyntax collides with Optionable in Scalatest causing codebases to no longer compile tests.


coltfred:
If you upgrade from cats 1.1.0 to 1.2+ and you use `OptionValues` from Scalatest you'll end up with errors like the following.

```
src/test/scala/com/ironcorelabs/admin/CookiesTest.scala:39: type mismatch;
 found   : gottenCookie.type (with underlying type Option[scodec.bits.ByteVector])
 required: ?{def value: ?}
Note that implicit conversions are not applicable because they are ambiguous:
 both method catsSyntaxNestedId in trait NestedSyntax of type [F[_], G[_], A](value: F[G[A]])cats.syntax.NestedIdOps[F,G,A]
 and method convertOptionToValuable in trait OptionValues of type [T](opt: Option[T])(implicit pos: org.scalactic.source.Position)CookiesTest.this.Valuable[T]
 are possible conversion functions from gottenCookie.type to ?{def value: ?}
      gottenCookie.value shouldBe Password(hexBytes.value)
```

This is caused by the fact that `.value` is syntax added to the `Option` type as well as the NestedSyntax conversion are at the same implicit level. I hadn't upgraded my cats version in a while so I didn't notice this until now, but it's a *huge* PITA. It means people might end up with cats all of a sudden not working for their project.

kailuowang:
related issue https://github.com/typelevel/cats/issues/2514

coltfred:
@kailuowang Should we close these on merge or when the release happens?

kailuowang:
yeah we can close this one.

PR #2614, opened 2018-11-15
Make the value of the class private

### LINKED ISSUE #2514: Value classes' public `val` field introduce unnecessary collisions
In Scala 2.10 the underlying value field in a value class has to be public, 
For example 
```scala
final class EitherIdOpsBinCompat0[A](val value: A) extends AnyVal
```
In this particular case, it introduced a `value` val to all types that simply returns itself. There is no real use case for this it's only due to the limitation of value classes in 2.10. It collides with any implicit extension that adds a `value` too. See https://github.com/typelevel/cats/issues/2491 for detail in this case. 
This limitation is removed since Scala 2.11. And Cats dropped support for 2.10 since 1.3. 

I thought of two approaches that can fix this. 

Option 1: make the existing [implicit conversion](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/syntax/either.scala#L382-L383) and [value class `EitherIdOpsBinCompat0`](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/syntax/either.scala#L386) package private and deprecated; copy the value class `EitherIdOpsBinCompat0` and paste it as a new value class with its `value` private and new implicit conversion for it. We also need a new `EitherSyntaxBinCompat1` trait for maintaining BC on Scala 2.11. 
I believe that this would be binary compatible and hide the `value` field going forward, but I haven't tested it yet. The downside is, of course, the duplicated code and more boilerplate. Also if we want to use the syntax inside the package we will run into collision.  

Option 2:  make the `value` field private. It seems to me that unless user calls the `value` field which has no practical usage, user won't run into binary compatibility issue. This can be verified by adding a `"232". leftNec[Int]` line to the [BC test](https://github.com/typelevel/cats/blob/3bb271406619434ba29fd31219ee8b98ef425f06/binCompactTest/src/main/scala/catsBC/MimaExceptions.scala#L8) added by #2509 ~(not merged yet)~ . We would also need to add a MiMa exception for this field. 

What do you guys think?

coltfred:
This makes the `.value` of two of the `Ops` classes private, since they shouldn't ever be referenced and they collide with Scalatest.

Fixes #2514 and #2613

### DIFF
--- a/binCompatTest/src/main/scala/catsBC/MimaExceptions.scala
+++ b/binCompatTest/src/main/scala/catsBC/MimaExceptions.scala
@@ -18,7 +18,9 @@ object MimaExceptions {
       cats.data.Kleisli.catsDataCommutativeFlatMapForKleisli[Option, Int],
       cats.data.IRWST.catsDataStrongForIRWST[List, Int, Int, Int],
       cats.data.OptionT.catsDataMonadErrorMonadForOptionT[List],
-      FunctionK.lift(headOption)
+      FunctionK.lift(headOption),
+      "blah".leftNec[Int],
+      List(Some(4), None).nested
   )
 
 }
--- a/build.sbt
+++ b/build.sbt
@@ -280,7 +280,10 @@ def mimaSettings(moduleName: String) =
         exclude[DirectMissingMethodProblem]("cats.data.KleisliInstances1.catsDataCommutativeArrowForKleisli"),
         exclude[DirectMissingMethodProblem]("cats.data.KleisliInstances4.catsDataCommutativeFlatMapForKleisli"),
         exclude[DirectMissingMethodProblem]("cats.data.IRWSTInstances1.catsDataStrongForIRWST"),
-        exclude[DirectMissingMethodProblem]("cats.data.OptionTInstances1.catsDataMonadErrorMonadForOptionT")
+        exclude[DirectMissingMethodProblem]("cats.data.OptionTInstances1.catsDataMonadErrorMonadForOptionT"),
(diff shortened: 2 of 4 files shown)


kailuowang:
Thanks! Would you also add usages of the syntax to https://github.com/typelevel/cats/blob/master/binCompatTest/src/main/scala/catsBC/MimaExceptions.scala
something like 
`"blah".leftNec[Int]`

This will make sure that such usages are still binary compatible.

kailuowang:
once this approach is approved, I'd like to update *all* value classes fields to private in Cats

coltfred:
@kailuowang Added.

coltfred:
I think I've proven to myself that things still work as expected, so this should be ready for actual review.

kailuowang:
Thanks @coltfred

kailuowang:
@rossabaker @ceedubs @LukaJCB what do you think of this? 
This is mostly to fix to the public field of value classes causing conflicts with other extensions. We had to make them public due to restrictions on scala 2.10 but we already dropped 2.10 support. If you all agree with this approach, I am going to create an issue to make **all*** value class fields private.

LukaJCB:
This looks very reasonable, If everything still works as is then I'm happy to change this for all the extensions. Are we going to do an 1.5.0-RC1? We could make sure everything still works with that release :)

kailuowang:
@LukaJCB yes I think we should do a RC1

coltfred:
I'm more than happy to give a 1.5.0-RC1 a try today or this weekend.

PR #2626, opened 2018-11-20
Fix typo

shokohara:
I'm sorry if these are not typo.


kailuowang:
thanks for contributing, `iff` is a notation for `if and only if`, maybe we can change this PR to add a note for that somewhere?

shokohara:
Thanks I understand. Please :+1:

kailuowang:
@shokohara sorry for the late response. Yes, let's close this one then.

PR #2629, opened 2018-11-21
WIP: Add Dotty build for kernel only

### LINKED ISSUE #2356: Add dotty build for kernel
/cc @smarter

Alistair-Johnson:
resolves #2356


kailuowang:
We probably need to change sbt release steps to accommodate this change (different crossScalaVersions for different modules). Basically, we need to do this 
```scala
    releaseCrossBuild := false,
      checkSnapshotDependencies,
      inquireVersions,
      releaseStepCommandAndRemaining("+clean"),
      releaseStepCommandAndRemaining("+test"),
      setReleaseVersion,
      commitReleaseVersion,
      tagRelease,
      releaseStepCommandAndRemaining("+publishSigned"),
      setNextVersion,
      commitNextVersion,
      releaseStepCommand("sonatypeReleaseAll"),
      pushChanges)
    
```

also set `crossScalaVersions` at root to `Nil`

Alistair-Johnson:
re the release, it's currently worse than that. 

In sbt shell, `++0.10.0-RC1` and then `compile` will start compiling all the projects under `cats`, even though `0.10.0-RC1` is only set in `kernelJVM`. So naturally, `+compile` won't work either.

Additionally, there are no tests in `kernelJVM`.

The `publish_cmd` in `travis-publish.sh` will go wrong for the same reasons.

So...I think next step is to fix the build in interactive sbt so it works cleanly, otherwise I'm not sure we will have much fun moving forward.

Alistair-Johnson:
@smarter 

The build issues I can/will look at,  but I was expecting  the `compat` files of  [TraversableOnce](https://github.com/typelevel/cats/blob/b2c7f1c7c37df457fd137a54f35687c419a89f02/kernel/src/main/scala-0.10/cats/kernel/compat/TraversableOnce.scala) and [WrappedMutableMapBase](https://github.com/typelevel/cats/blob/b2c7f1c7c37df457fd137a54f35687c419a89f02/kernel/src/main/scala-0.10/cats/kernel/compat/WrappedMutableMapBase.scala) to be the same as for the 2.13 build. In fact, it looks like the dotty version is a mix of 2.12 and 2.13

It's not a big deal for cats, but I would just to point these out to you in case there is indeed an issue with dotty here.

smarter:
>  In fact, it looks like the dotty version is a mix of 2.12 and 2.13

No. Dotty right now only uses the Scala 2.12 standard library (so that you can depend on Scala 2.12 projects in your Dotty build). We'll switch to 2.13 once it's released.

Alistair-Johnson:
Thanks...I'll change that back to 2,12 :+1:

build.sbt
```diff
@@ -483,9 +490,14 @@ lazy val kernel = crossProject(JSPlatform, JVMPlatform)
   .settings(scoverageSettings)
   .settings(sourceGenerators in Compile += (sourceManaged in Compile).map(KernelBoiler.gen).taskValue)
   .settings(includeGeneratedSrc)
-  .jsSettings(commonJsSettings)
+  .jsSettings(commonJsSettings ++ noDottySettings)
   .jvmSettings(commonJvmSettings ++ mimaSettings("cats-kernel"))
-  .settings(libraryDependencies += "org.scalacheck" %%% "scalacheck" % scalaCheckVersion(scalaVersion.value) % "test")
```
smarter: If you add `.withDottyCompat(scalaVersion.value)` you should be able to keep this dependency when compiling with Dotty too (https://github.com/lampepfl/dotty-example-project#getting-your-project-to-compile-with-dotty)

build.sbt
```diff
@@ -878,7 +907,8 @@ lazy val warnUnusedImport = Seq(
         Seq("-Ywarn-unused-import")
       case Some((2, n)) if n >= 12 =>
         Seq("-Ywarn-unused:imports")
-
+      case Some((0, 10))  =>
+        Seq("-Ywarn-unused:imports")
```
smarter: That setting doesn't exist in Dotty currently

build.sbt
```diff
@@ -906,6 +936,7 @@ lazy val xlint = Seq(
   scalacOptions += {
     CrossVersion.partialVersion(scalaVersion.value) match {
       case Some((2, scalaMajor)) if scalaMajor >= 12 => "-Xlint:-unused,_"
+      case Some((0, 10))                             => "-Xlint:-unused,_"
```
smarter: There's no -Xlint in Dotty currently (and I hope scalafix can completely replace -Xlint)

build.sbt
Alistair-Johnson: Oh...I missed that!! Cool

project/plugins.sbt
```diff
@@ -19,3 +19,4 @@ addSbtPlugin("org.scala-js" % "sbt-scalajs" % "0.6.25")
 addSbtPlugin("org.scala-native" % "sbt-scala-native" % "0.3.8")
 addSbtPlugin("com.geirsson" % "sbt-scalafmt" % "1.5.1")
 addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.9.0")
+addSbtPlugin("ch.epfl.lamp" % "sbt-dotty" % "0.2.4")
```
smarter: Latest is 0.2.6 (looks like we forgot to update the example projects)

Alistair-Johnson: Indeed... my proof that I _did_ Read The Fantastic Manual, but missed the important bit.

diesalbla:
Since @travisbrown is already working on a Scala 3 / Dotty migration branch, perhaps this one is no longer needed.

larsrh:
Thanks for the notice, @diesalbla. Closing this one in favour of #3269.

PR #2635, opened 2018-11-21
Replace List with Chain in Iort and EitherT docs

kun-song:


ceedubs:
Is there any particular motivation for this? To me `Chain` makes sense as the log accumulator for `WriterT` and the error accumulator for `Either`/`Validated` because of the efficiency of concatenation. But is there a particular reason to prefer it in these examples?

larsrh:

PR #2641, opened 2018-11-26
Adding partitionEitherM

blast-hardcheese:
- [x] Used `partitionEitherM` in a project to ensure there weren't any odd usability concerns
- [x] Added `Foldable[F].partitionEitherM[G, A, B, C]`
- [x] Duplicated all `partitionEither` tests and modified them to test `partitionEitherM`
- [x] Ran `sbt test`
- [x] Ran `sbt prePR`


blast-hardcheese:
Getting an unrelated error due to decimal precision in JS:
```
[info] Tests:
[info] - CommutativeGroup[BigDecimal].commutativeGroup.associative *** FAILED ***
[info]     Falsified after 96 successful property evaluations.
[info]       arg0 = 5.182509090085532E-184,
[info]       arg1 = -5.036270751383713E-218,
[info]       arg2 = 2.0611576410519167E-208
[info]     )
[info]       Expected: 5.182509090085532000000002061157641E-184
[info]   Received: 5.182509090085532000000002061157640E-184
```

ceedubs:
Thanks, @blast-hardcheese! This looks good to me, but introduces a binary incompatibility, so we'll have to figure out how to handle that. We'll either need to temporarily add it on as a syntax extension or wait until Cats 2.0 to merge it. cc @kailuowang @LukaJCB

blast-hardcheese:
@LukaJCB I wouldn't be opposed to that, but that would introduce the same binary incompatibility, just for the syntax class, wouldn't it?

LukaJCB:
Not quite, you can get it working by creating a new trait that gets mixed into the various imports :)
You can find a lot of examples if you search for e.g. `BinCompat1` in the codebase

blast-hardcheese:
Resolved conflict with another BinCompat that showed up in the interim

core/src/main/scala/cats/implicits.scala
```diff
@@ -7,6 +7,7 @@ object implicits
     with syntax.AllSyntaxBinCompat2
     with syntax.AllSyntaxBinCompat3
     with syntax.AllSyntaxBinCompat4
+    with syntax.AllSyntaxBinCompat5
```
kailuowang: `AllSyntaxBinCompat4` was introduced after the last release.  You can simply let it extends your `FoldableSyntaxBinCompat1`

blast-hardcheese:
@kailuowang Understood. I was on the fence about how this method should be consumed, as I try to strike a good balance between convenience and discoverability for my teammates. Are there objections to having both?

blast-hardcheese:
For the time being, I've made the changes you suggested

core/src/main/scala/cats/syntax/foldable.scala
```diff
@@ -214,4 +218,72 @@ final class FoldableOps0[F[_], A](val fa: F[A]) extends AnyVal {
...
+   * scala> val partitioned1 = Foldable[List].partitionEitherM(list)(a => if (a % 2 == 0) Eval.now(Either.left[String, Int](a.toString)) else Eval.now(Either.right[String, Int](a)))
+   * Since `Eval.now` yields a lazy computation, we need to force it to inspect the result:
+   * scala> partitioned1.value
+   * res0: (List[String], List[Int]) = (List(2, 4),List(1, 3))
+   * scala> val partitioned2 = Foldable[List].partitionEitherM(list)(a => Eval.later(Either.right(a * 4)))
+   * scala> partitioned2.value
+   * res1: (List[Nothing], List[Int]) = (List(),List(4, 8, 12, 16))
+   * }}}
+   */
+  def partitionEitherM[G[_], A, B, C](fa: F[A])(f: A => G[Either[B, C]])(implicit A: Alternative[F],
```
ceedubs: This code is duplicated in two places. What if instead we added a `partitionEitherM` method to the `Foldable` companion object and had the `FoldableOps0` method delegate to it? At that point, I'm not sure if it makes sense to have the `FoldableOps1` extension to the type class itself; people could just call the method on the companion object if they wanted it. WDYT @kailuowang and @blast-hardcheese?

Sorry I'm sure that this is annoying. I think that Cats maintainers need to put some effort into our approach to binary compatibility and making it easier on other contributors.

blast-hardcheese: Consolidating definitely felt like the right approach, but doing the same syntax lookup again while we're already working around bin compat seemed worse. I can make the change.

core/src/main/scala/cats/syntax/foldable.scala
```diff
@@ -214,4 +218,62 @@ final class FoldableOps0[F[_], A](val fa: F[A]) extends AnyVal {
...
```
ceedubs: Would it make sense to generalize this to any `Bifoldable` instead of just `Either`?

```scala
  def partitionBiffyM[G[_], H[_,_], A, B, C](fa: F[A])(f: A => G[H[B, C]])(implicit A: Alternative[F],
                                                                         M: Monad[G], H: Bifoldable[H]): G[(F[B], F[C])] = {
    import cats.instances.tuple._

    implicit val mb: Monoid[F[B]] = A.algebra[B]
    implicit val mc: Monoid[F[C]] = A.algebra[C]

    F.foldMapM[G, A, (F[B], F[C])](fa)(
      a =>
        M.map(f(a)){
          H.bifoldMap[B, C, (F[B], F[C])](_)(
            b => (A.pure(b), A.empty[C]),
            c => (A.empty[B], A.pure(c)))
      }
    )
  }
```

Then you could use it for the `partitionEitherM` use-case, but you could also use it for other `Bifoldable`s like `Tuple2`:

```
scala> Foldable[List].partitionBiffyM(list)(a => Eval.now((a - 1, a + 1)))
res5: cats.Eval[(List[Int], List[Int])] = cats.Eval$$anon$6@67bef13e

scala> res5.value
res6: (List[Int], List[Int]) = (List(0, 1, 2, 3),List(2, 3, 4, 5))
```

I'm hoping that someone else has a better name than mine 😛

blast-hardcheese: This seems to suggest a broader refactor that includes [`partitionEither`](https://github.com/typelevel/cats/blob/774fb51/core/src/main/scala/cats/Foldable.scala#L494-L507). I can generalize to `partitionBiffy` as well as `partitionBiffyM`, while providing the concretized `partitionEitherM` that just defers to `partitionBiffyM`, if that is desirable

blast-hardcheese: I gave it a shot, though the examples are getting increasingly convoluted I think.

kailuowang:
looks like this one can use another `sbt fmt`

blast-hardcheese:
@kailuowang just pushed a fix -- I seem to remember somewhere seeing that it's not desired to force push to branches in cats, is this a hard rule? I'd really like to squash some of the noise, only keeping the relevant commits

blast-hardcheese:
Understood.

LukaJCB:
I just want to preface this, that I really appreciate the time you take to make this PR perfect, so thanks a ton @blast-hardcheese!
I'm really sorry to continue the bikeshedding here, especially after you've already made a ton of changes to your PR, but can we name the functions `partitionBifold` instead? 
Sounds nicer to me than bifoldable and also reflects the implementation, we partition using some structure which supports a `bifold`, so it seems nicer and more accurate IMO. 

Again thanks for your work so far! :)

ceedubs:
Agreed. Thanks a lot, @blast-hardcheese!

LukaJCB:
Can you run scalafmt once more? :)

blast-hardcheese:
Done

core/src/main/scala/cats/syntax/foldable.scala
```diff
@@ -214,4 +218,145 @@ final class FoldableOps0[F[_], A](val fa: F[A]) extends AnyVal {
...
+  /**
+   * Separate this Foldable into a Tuple by an effectful separating function `A => H[B, C]` for some `Bifoldable[H]`
+   * Equivalent to `Functor#map` over `Alternative#separate`
+   *
+   * {{{
+   * scala> import cats.implicits._, cats.data.Const
+   * scala> val list = List(1,2,3,4)
+   * scala> list.partitionBifold(a => (a, "value " + a.toString))
+   * res0: (List[Int], List[String]) = (List(1, 2, 3, 4),List(value 1, value 2, value 3, value 4))
+   * `Const`'s second parameter is never instantiated, so we can use an impossible type:
```
ceedubs: Should line 231 have a `//` at the beginning of the comment? I guess that I don't know whether or not that's necessary in a doctest example; just wanted to make sure that this code is running when tests are run.

tests/src/test/scala/cats/tests/FoldableSuite.scala
```diff
@@ -78,6 +78,63 @@ abstract class FoldableSuite[F[_]: Foldable](name: String)(implicit ArbFInt: Arb
...
+      val vector = Foldable[F].toList(fi).toVector
+      val result = Foldable[Vector].partitionEitherM(vector)(f.andThen(Option.apply)).map {
+        case (lefts, rights) =>
+          (lefts <+> rights).size
+      }
+      result should ===(Option(vector.size))
+    }
+  }
+
+  test("Foldable#partitionEitherM consistent with List#partition") {
```
ceedubs: Ooh this is a good idea for a test 👍

core/src/main/scala/cats/syntax/foldable.scala
blast-hardcheese: I should say, I'm not opposed to changing it, just that it currently isn't causing the tool to stop compiling the other lines.

Issue #2474, opened 2018-09-06
Additional methods for Bitraverse


asoltysik:
What do you think about adding any/all of the following functions to `Bitraverse`? IMHO `leftTraverse` and `leftSequence` would be most useful. 

```scala
def leftTraverse[G[_]: Applicative, A, B, C](fab: F[A, B])(f: A => G[C]): G[F[C, B]] =
  bitraverse(fab)(f, Applicative[G].pure _)

def leftSequence[G[_]: Applicative, A, B](fab: F[G[A], B]): G[F[A, B]] = 
  leftTraverse(fab)(identity)

def biflatTraverse[G[_]: Applicative, A, B, C, D](fab: F[A, B])(f: A => G[F[C]], g: B => G[F[D]])(implicit F: FlatMap[F]): G[F[C, D]] = 
  Applicative[G].bimap(traverse(fab)(f, g))(F.flatten, F.flatten)

def leftFlatTraverse[G[_]: Applicative, A, B, C](fab: F[A, B])(f: A => G[F[C]])(implicit F: FlatMap[F]): G[F[C, B]] = 
  biflatTraverse(fab)(f, identity)
```

The code above is only a draft, might not compile.

sideeffffect:
I've looked into it briefly and `biflatTraverse` doesn't seem to be trivially implementable. Just as `flatTraverse` needs `FlatMap[F]`
```scala
def flatTraverse[G[_], A, B](fa: F[A])(f: A => G[F[B]])(implicit G: Applicative[G], F: FlatMap[F]): G[F[B]]
```
we would need a totally new `BiFlatMap[F]` for `biflatTraverse` to flatten `F[F[C, D], F[C, D]]` to mere `F[C, D]`.
```
def biflatTraverse[G[_], A, B, C, D](
    fa: F[A, B]
  )(f: A => G[F[C, D]], g: B => G[F[C, D]])(implicit G: Applicative[G], F: BiflatMap[F]): G[F[C, D]] =
    G.map(bitraverse(fa)(f, g))(F.biflatten)
...

trait BiflatMap[F[_, _]] extends Bifunctor[F] {
  def biflatten[A, B](ffa: F[F[A, B], F[A, B]]): F[A, B]
}
```

Would you accept a PR with `BiflatMap`? Seems a bit ad hoc to me, I'm not sure how many usages it would have...

PR #2646, opened 2018-11-28
Add leftTraverse and leftSequence to Bitraverse typeclass

ikempf:
Bitraverse should offer "left" traversing functionality.
Right-traversal is already offered through Traverse typeclass, bi-traversal and left-traversal should both be offered by Bitraverse. This keeps it consistent with Bifunctor (which offers leftMap).

This resolves issue #2474

### DIFF
--- a/core/src/main/scala/cats/syntax/all.scala
+++ b/core/src/main/scala/cats/syntax/all.scala
@@ -76,4 +76,4 @@ trait AllSyntaxBinCompat2
     with ListSyntaxBinCompat0
     with ValidatedSyntaxBincompat0
 
-trait AllSyntaxBinCompat3 extends UnorderedFoldableSyntax with Function1Syntax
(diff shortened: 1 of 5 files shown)


kailuowang:
Unfortunately we can't add methods to traits without breaking binary compatibility on Scala 2.11.
You can add this to a new syntax extension trait like in #2588.

ikempf:
I removed methods from existing traits (Bitraverse and BitraverseSyntax) and moved everything into a new trait BitraverseSyntaxBinCompat0 that is extended by AllSyntaxBinCompat3 and bitraverse.

LukaJCB:
Seems like this is a duplicate of https://github.com/typelevel/cats/pull/2606

ikempf:
@LukaJCB You are right it is a duplicate

Issue #2319, opened 2018-07-09
Evaluate moving to Circle CI 


kailuowang:
Travis' memory issue is a bit too much and our build there now takes more than 3 hours.

tpolecat:
Might also look at [BuildKite](https://buildkite.com) if we can get people to donate hardware. It's easy enough for me to set up, which means it's easy. An agent running on @larsrh's 3874629847653-core machine would be 🔥

larsrh:
@tpolecat That won't work, unfortunately. That machine is university property.

djspiewak:
@kailuowang I'm a bit out of the loop, I think. The build takes 3 hours? How long does it take locally? What is it doing? Thanks to working at SlamData, I have a remarkably vast swath of experience debugging slow Travis builds. I'd be happy to take a look if you want.

BennyHill:
the 2-3 hours is the combined total of the builds, each job is typically 20-30 minutes - https://travis-ci.org/typelevel/cats/builds/415542808

And a lot of that time is coverage testing, tut testing, doc testing, site building and so on.

djspiewak:
Ok taking a quick look at things, literally the first things that occur to me:

- Oh god, you're using a separated build script… I hate that convention.
- Why is `sudo: required`? I'm relatively certain those VMs are slower. Is it just for `codecov`? See below.
- The `travis-publish.sh` script goes to great lengths to push things all into a single SBT instance. In my experience, this is exactly the opposite of what you want to do when you have a slow build. Separate SBT processes, sequentially invoked, gives you better memory characteristics and is better understood by Travis (especially if you *don't* split the build script out of `.travis.yml`).
- `.jvmopts` uses `-Xmx6g`. This is problematic because Travis doesn't have that much memory! You should strongly consider dropping that option altogether and allowing it to be the default (ditto with `-Xms`), which will be scaled off of the reported system memory.
- We should have a discussion about whether or not code coverage is actually worth anything. Frankly, I've never seen it provide any value whatsoever, and it doubles the duration of the JVM build.
- Why is the Ivy cache not being sanitized prior to publication? This is resulting in re-caching quite often.
- Random best-practice: consider commenting on each of the `secure` variables so we know which one is which.

I didn't look at SBT itself. Looks like a lot of the logic is in tasks, so that may also contribute.

BennyHill:
[...]

Issue #2570, opened 2018-10-19
Get rid of `publish-travis.sh`


kailuowang:
The bash script is getting unmanageable. 
Instead we can define travis jobs explicitly in a clearer jobs spec, here is a simplified example. 

```yml

scala: 
    - 2.11.12
    - 2.12.6
    - 2.13.0-M4
jobs:
  include:
    - env:
        - NAME="Validate JVM"
      script:
        - sbt ++$TRAVIS_SCALA_VERSION validateJVM

    - if: env(TRAVIS_SCALA_VERSION) =~ /^2.12/ 
      env:
        - NAME="Validate docs"
      script:
        - sbt docs/tut

    - if: (branch = master) AND (type = pull_request)
      env:
        - NAME="Publish SNAPSHOT"
      script:
        - sbt ++$TRAVIS_SCALA_VERSION publish gitSnapshots publish

```

djspiewak:
FWIW, there is no filesystem caching between job stages, and even if you emulate it by hand (e.g. by using S3 to pass files between stages), you still run into the fact that Zinc is irrationally picky about file state and will often refuse to incrementally compile.

ceedubs:
In case it saves anyone else some time, I went down a bit of a rabbit hole trying to use Travis CI's new "stages" feature, and ran into trouble. I don't entirely understand it, but it seems that the "stages" are part of the build matrix, as is the scala version, and there's some sort of interaction where it tries to create a default stage that just runs `test` no matter what you do. Or at least that was my experience. Someone who knows more about Travis may be able to do better.

ceedubs:
I'm working on this. I'll submit a PR after https://github.com/typelevel/cats/pull/2732 is merged.

ceedubs:
On that note: does anyone know why our publish command is `publish gitSnapshots publish`? Why does `publish` show up twice?

PR #2648, opened 2018-11-29
Throttle list size in ListWrapper

Alistair-Johnson:
Running the tests locally, and in particular `ApplicativeSuite`, I managed to occasionally, but nonetheless repeatedly, experience the same test timeouts we see on travis. Additionally, I observed some tests taking much longer than average, always coupled with the test JVM memory jumping up from about 1.2G to the max of 3G.

A few `println`'s later...the `List[A]`'s generated by `Arbitrary` in the custom `ListWrapper` are generally in the sub 100 element size, but I have seen them balloon to over 8 million in the slow tests; this would, I believe, explain the OOM's.

The real fix is to make sure scalacheck creates reasonable size lists for `List[A]`'s in the first place, but unfortunately, I could not _quickly_ achieve this. So the current fix is suboptimal, but works. 

So if you are happy with the investigation, I might suggest that the current "fix" is better than what we currently have - a new issue can be raised for a more elegant solution, that might also take into account other slow tests that might suffer from a similar issue.

### DIFF
--- a/tests/src/test/scala/cats/tests/ListWrapper.scala
+++ b/tests/src/test/scala/cats/tests/ListWrapper.scala
@@ -32,7 +32,9 @@ import org.scalacheck.Arbitrary.arbitrary
  * }
  * }}}
  */
-final case class ListWrapper[A](list: List[A]) extends AnyVal
+final case class ListWrapper[A](listA: List[A]) extends AnyVal {
+  def list: List[A] = listA.take(100)
+}
 
 object ListWrapper {
   def order[A: Order]: Order[ListWrapper[A]] = Order.by(_.list)


Alistair-Johnson:
Also, perhaps resolves https://github.com/typelevel/cats/issues/2319 ?

kailuowang:
hmmm, this might be a little surprising to users. Imagine they intentionally created a 10k item ListWrapper to test stack safety of some type class. I am curious how this solution compared to changing the `Arbitrary` instance of `LIstWrapper`?  namely change [line 142](https://github.com/typelevel/cats/blob/b64c76bf5d63df86fa1d2d8a7171229667106bb1/tests/src/test/scala/cats/tests/ListWrapper.scala#L141-L142)
from
```scala
implicit def listWrapperArbitrary[A: Arbitrary]: Arbitrary[ListWrapper[A]] =
    Arbitrary(arbitrary[List[A]].map(ListWrapper.apply))
```
to 

```scala
implicit def listWrapperArbitrary[A: Arbitrary]: Arbitrary[ListWrapper[A]] =
    Arbitrary(arbitrary[List[A]].map(l => ListWrapper(l.take(100))
```

Alistair-Johnson:
It actually needs :

```scala
Arbitrary(arbitrary[List[A]].map((l: List[A]) => ListWrapper(l.take(100))))
```
But unfortunately, it does not work. Whilst re-testing, I actually got a new PB of 29 million list size...

Test code, for those interested, :

```scala
new Alternative[ListWrapper] {
      def pure[A](x: A): ListWrapper[A] = ListWrapper(M.pure(x))

      def ap[A, B](f: ListWrapper[A => B])(fa: ListWrapper[A]): ListWrapper[B] ={
        if (f.list.length > 1000 ||  fa.list.length >1000)
          println( f.list.length +" "+  fa.list.length )
        ListWrapper(M.ap(f.list)(fa.list))
      }
...
```

I'm totally with you that the current "fix" is not ideal and _could_ give surprising results. The problem is, the current implemention _does_ give surprising results, and does so regularly; it make my laptop sound like an air raid siren. "oooooOOOOOOoooooooOOOOOOOOoooooo"

Alistair-Johnson:
Just thinking out loud... I wonder if the issue has anything to do with the fact that we aren't directly using a `ListWrapper[A]`, but in fact an `Alternative[ListWrapper]` ?

kailuowang:
It's interesting that the list size was not caused by `Arbitrary` so there must be other code ballooning that size to millions. If we throw an exception in ListWrapper constructor when the given list has more than a large size, we maybe able to find out where it happened?

Alistair-Johnson:
Yeah...this is worth nailing once and for all...I'll have a look later :+1:

ceedubs:
@Alistair-Johnson when you saw the ballooned sizes, was it for passing tests or was it when Scalacheck was trying to shrink a failing test case?

kailuowang:
FYI just encountered another one of these on build (2.13)
```
[info] - Applicative[ListWrapper].monoid.monoid.combine all *** FAILED ***
[info]   OutOfMemoryError was thrown during property evaluation.
[info]     Message: GC overhead limit exceeded
[info]       arg0 = Vector(ListWrapper(List(1074835304, 0, -777179633, 1455578008, -25683529, 155715011)), ListWrapper(List(-1840890490, -2147483648, 1007200674, -1, -1, -2147483648, -1, 0, -107186259)), ListWrapper(List(0, -911786825, 1123360752, 0, -821246721, 0, -2147483648, 1495617553, -1, 2036308788)), ListWrapper(List(-916165509, 1950063819, -1961633005, 1, 1856854552, -2147483648, -1130776)), ListWrapper(List(-1, 711237612, 1424350193)), ListWrapper(List(0, -58409028, 1, -1897919184, -2134889206)), ListWrapper(List(0, -2147483648, -72482326, -1, -1641350473, -1717342098, 1, -1)), ListWrapper(List(0, -2147483648, -1996303227, -459503821, 0, -2147483648, 2147483647, 1, -1321329796, -1)), ListWrapper(List(0, -1, 930579752, 770218430, 1, 1469285130, 0)))
[info]     )
```

DavidGregory084:
@kailuowang I can reproduce the same behaviour just by running `sbt catsJS/test:fastOptJS`: it maxes out heap usage and a memory sampler shows a similar amount of `scala.collection.immutable.$colon$colon` allocations.

I was looking at some heap dumps earlier and they are dominated by scala.js ClearableLinkers.

Is it possible to increase the heap size to the ~3.5-4gb we need for this to run? Alternatively, can we split the build up so that fewer tasks are running in parallel? 

SBT will happily run four fastOptJS tasks in parallel which is very heavy in terms of memory usage. Perhaps we could use the tag system in SBT to cut down the number of fastOptJS tasks that can run at once.

kailuowang:
@DavidGregory084 the VM running our build, according to travis documentation, has 6G of memory, so we should be able to increase. I'd like to further break up the build task as well (as part of this proposal https://github.com/typelevel/cats/issues/2570), but that might be a heavier task.

DavidGregory084:
> I'm not convinced this is caused by ListWrapper, as it doesn't happen if you do `catsJVM/test`. Is it possible this is due to the Scala.js compiler somehow?

I think now that I was totally wrong about this as I have indeed seen the same thing on JVM tests while investigating #2319

larsrh:
Is this still an issue? If yes, feel free to re-open.

PR #2654, opened 2018-12-02
Add a shell.nix and instructions for using it

ceedubs:
I have mixed feelings about this. On the one hand, it feels like which
tools a developer uses in their environment shouldn't bleed into the
source repository. On the other hand, getting a proper Cats development
environment includes wrangling some finicky bits that can be tricky for
a Scala developer (such as managing Jekyll and Node.js versions). I'm
generally in favor of lowering this barrier and using a deterministic
build environment.

I've been using something like this for a while when working on Cats and
thought that I'd put together this PR in case others are interested in
doing the same. If people aren't crazy about this, it's no big deal;
it's easy enough to have it ignored by git.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -97,7 +97,9 @@ git clone git@github.com:typelevel/cats.git
 
 To build Cats you should have
 [sbt](http://www.scala-sbt.org/0.13/tutorial/Setup.html) and [Node.js](https://nodejs.org/)
- installed. Run `sbt`, and then use any of the following commands:
+ installed. If you'd like, you can use the [Nix Cats development environment](#nix-cats-development-environment).
+
+ Run `sbt`, and then use any of the following commands:
 
  * `compile`: compile the code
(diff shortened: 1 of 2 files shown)


ceedubs:
@tpolecat I've seen you comment on the mixed reviews that you've seen of Nix. For what it's worth, if you are on a Mac, I don't think that Nix is quite there as a Homebrew replacement, but I think that it works well for isolated development environments such as the one described in this PR. (I've also been using it as a Homebrew replacement, but it has required a lot of tinkering and some looking at source code, and not everything that I would want is there -- I think that the story is probably a bit better if you are running NixOS).

ceedubs:
This code coverage report makes no sense to me...

rossabaker:
I think this is harmless for those who don't know Nix it and useful for those who do, or are willing to try.

LukaJCB:
I haven't used Nix much yet (though I definitely want to), but I think this is small enough to be perfectly fine as part of this Repo :)

PR #2655, opened 2018-12-03
Add `rethrowT` method to EitherT

bplommer:
- Add `rethrowT` method to EitherT: inverse of `ApplicativeError#attemptT`, analogous to `MonadError#rethrow`
- Add/fix some tests
- Document `MonadErrorOps.reject` and refactor to one line

### DIFF
--- a/core/src/main/scala/cats/syntax/monadError.scala
+++ b/core/src/main/scala/cats/syntax/monadError.scala
@@ -18,10 +18,12 @@ final class MonadErrorOps[F[_], E, A](private val fa: F[A]) extends AnyVal {
   def ensureOr(error: A => E)(predicate: A => Boolean)(implicit F: MonadError[F, E]): F[A] =
     F.ensureOr(fa)(error)(predicate)
 
+  /**
+   * Turns a successful value into the error returned by a given partial function if it is
+   * in the partial function's domain.
+   */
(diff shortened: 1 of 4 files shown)


core/src/main/scala/cats/syntax/monadError.scala
```diff
@@ -18,10 +18,12 @@ final class MonadErrorOps[F[_], E, A](private val fa: F[A]) extends AnyVal {
...
 
+  /**
+   * Turns a successful value into the error returned by a given partial function if it is
+   * in the partial function's domain.
+   */
   def reject(pf: PartialFunction[A, E])(implicit F: MonadError[F, E]): F[A] =
-    F.flatMap(fa) { a =>
-      pf.andThen(F.raiseError[A]).applyOrElse(a, (_: A) => fa)
-    }
+    F.flatMap(fa)(pf.andThen(F.raiseError[A]).applyOrElse(_, (_: A) => fa))
```
kailuowang: Maybe I am biased, but I think the original version is more readable.

bplommer: I'm not attached to this change - I made it to match the style of other utility methods, but I'm happy to remove it if the original version is preferred.

PR #2659, opened 2018-12-03
change prePR to run BC test on Scala 2.11

kailuowang:
Just realized that currently prePR is testing BC  on 2.12 which probably won't catch the majority of the BC breaking changes.


LukaJCB:
Going to merge, since this is such a tiny change that doesn't affect the actual code :)

Issue #2651, opened 2018-11-30
Remove fine-grained imports from docs


tpolecat:
Some of the doc pages use fine-grained imports, which we should discourage as it's very tricky (so tricky in fact that most advanced users don't even try) and is a source of discouraging problems and many questions on the Gitter channel. Can we go through the doc and remove them?

calvinbrown085:
Hi @tpolecat do you have an example here? I'm guessing it's somewhere we can import `._` instead?

LukaJCB:
@calvinbrown085 
In general when there's docs where anything apart from `import cats.implicits._` is used, change it to `cats.implicits._` instead. So e.g. in https://github.com/typelevel/cats/blob/master/docs/src/main/tut/typeclasses/applicative.md

you'll see this:

```scala 
import cats.instances.list._
import cats.syntax.traverse._
```

That should be changed to just 
```scala
import cats.implicits._
```

calvinbrown085:
@LukaJCB Ah, that makes sense! I can try to go through and find some of these to fix

ceedubs:
I think that this can be closed out now. Thanks, @calvinbrown085!

PR #2664, opened 2018-12-05
Cleaning up docs to no longer use fine grained imports

calvinbrown085:
https://github.com/typelevel/cats/issues/2651
@LukaJCB @tpolecat 

How does this look to you guys? Hopefully I got them all, happy to take another sweep through if you know of any I missed.

### DIFF
--- a/docs/src/main/tut/typeclasses/applicative.md
+++ b/docs/src/main/tut/typeclasses/applicative.md
@@ -87,7 +87,7 @@ def product3[F[_]: Applicative, A, B, C](fa: F[A], fb: F[B], fc: F[C]): F[(A, B,
 Let's see what happens if we try to compose two effectful values with just `map`.
 
 ```tut:book:silent
-import cats.instances.option._
+import cats.implicits._
 
 val f: (Int, Char) => Double = (i, c) => (i + c).toDouble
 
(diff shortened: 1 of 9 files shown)


docs/src/main/tut/typeclasses/applicative.md
```diff
@@ -212,8 +212,7 @@ This function is provided by Cats via the `Traverse[List]` instance and syntax,
 tutorial.
 
 ```tut:book:silent
-import cats.instances.list._
-import cats.syntax.traverse._
+import cats.implicits
```
ChristopherDavenport: Did this miss the `._`?

kailuowang: oops, missed that, I will fix that in a separate PR.

fthomas: Why didn't this broke the build?

kailuowang: the import isn't really needed for the code since it's already imported at previous blocks. I guess unused import isn't checked in tut.

fthomas: If it is not needed I'd remove it or require that each paragraph needs to specify its imports again with maybe tut's `:reset` modifier.

kailuowang: I read the page again and I can see it become confusing. It looks like it would be cumbersome to reset every paragraph, I suggested that we remove `import cats.implicits._` from them and add it to the top of the page, with something like:
First, bring cats instances into the scope by
```scala
import cats.implicits._
```

PR #2665, opened 2018-12-06
minor fix

kailuowang:


LukaJCB:
How come this compiled? 🤔

kailuowang:
it's legitimate `import` right?

kailuowang:
sorry I wasn't clear. this `import cats.implicits._` is redundant because it's already imported in previous blocks. However for doc purpose, I think we might still want to keep it.

LukaJCB:
Sounds good to me :)

PR #2671, opened 2018-12-17
add syntax <&> for parallel

jcouyang:
in #2038 introduced `&>` and `<&` syntax for parallel which are very helpful

why not have a `<&>` too, so  I can
```scala
mab <&> ma
```
instead of
```scala
Parallel.parAp(mab)(ma)
```

### DIFF
--- a/core/src/main/scala/cats/syntax/parallel.scala
+++ b/core/src/main/scala/cats/syntax/parallel.scala
@@ -3,8 +3,6 @@ package cats.syntax
 import cats.{FlatMap, Foldable, Monad, Parallel, Traverse}
 
 trait ParallelSyntax extends TupleParallelSyntax {
-  implicit final def catsSyntaxParallelApply[F[_], A, B](fa: F[A => B]): ParallelApplyOps[F, A, B] =
-    new ParallelApplyOps[F, A, B](fa)
 
   implicit final def catsSyntaxParallelTraverse[T[_]: Traverse, A](ta: T[A]): ParallelTraversableOps[T, A] =
     new ParallelTraversableOps[T, A](ta)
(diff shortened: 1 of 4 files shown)


core/src/main/scala/cats/syntax/parallel.scala
```diff
@@ -3,6 +3,9 @@ package cats.syntax
 import cats.{FlatMap, Foldable, Monad, Parallel, Traverse}
 
 trait ParallelSyntax extends TupleParallelSyntax {
+  implicit final def catsSyntaxParallelApply[F[_], A, B](fa: F[A => B]): ParallelApplyOps[F, A, B] =
```
kailuowang: Unfortunately, you can't add a method to a trait without breaking binary compatibility in Scala 2.11. 
You need to add a new trait to add this method and have relevant objects extending it. And then you can have `AllSyntaxBinCompat4` (a new trait introduced post 1.5.0 release) extend this.

kubukoz:
I think you can add your syntax to the trait introduced here (in a similar way): https://github.com/typelevel/cats/pull/2600 (it was after 1.5.0)

jcouyang:
thanks @kailuowang @kubukoz I added it to `AllSyntaxBinCompat4` also the syntax object `parallel`
but I think the coverage drop because `SyntaxSuite.scala` is only compile time check, coverage report was not able picking that up
should I add a run time test case in `ParallelSuite.scala`?

kubukoz:
I don't think anyone looks at the coverage reports, but if you want you can add a doctest to ensure this works as expected (I'm not sure syntax is being tested in ParallelSuite)

jcouyang:
Good, I personally think the syntaxsuite is good enough for covering the syntax

PR #2672, opened 2018-12-17
Alter from Reflexitivity to Reflexivity

mattkohl:
Perhaps "reflexitivity" here should be "reflexivity"?


kailuowang:
another BC breaking law change.

PR #2674, opened 2018-12-18
minor fix to build

kailuowang:
removed redundant `dependsOn`


kailuowang:
thanks @ceedubs. Since this is such a simple fix I am going to merge with one approval.

PR #2677, opened 2018-12-29
Write documentation for ContT

cb372:
Had a crack at writing some documentation for ContT.

Corrections/suggestions welcome.

### DIFF
--- a/docs/src/main/tut/datatypes/contt.md
+++ b/docs/src/main/tut/datatypes/contt.md
@@ -1,0 +1,186 @@
+---
+layout: docs
+title:  "ContT"
+section: "data"
+source: "core/src/main/scala/cats/data/ContT.scala"
+scaladoc: "#cats.data.ContT"
+---
+# ContT
+
+A pattern that appears sometimes in functional programming is that of a function
+first computing some kind of intermediate result and then passing that result to
+another function which was passed in as an argument, in order to delegate the
+computation of the final result.
+
+For example:
+
+```tut:silent
+case class User(id: Int, name: String, age: Int)
+sealed abstract class UserUpdateResult
+case class Succeeded(updatedUserId: Int) extends UserUpdateResult
+case object Failed extends UserUpdateResult
+```
+
+```tut:book
+import cats.Eval
+
+def updateUser(persistToDatabase: User => Eval[UserUpdateResult])
+              (existingUser: User, newName: String, newAge: Int): Eval[UserUpdateResult] = {
+  val trimmedName = newName.trim
+  val cappedAge = newAge max 150
+  val updatedUser = existingUser.copy(name = trimmedName, age = cappedAge)
+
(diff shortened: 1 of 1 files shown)


kailuowang:
Thanks so much, @cb372. Our users can certainly use some documentation on this one. I think this draft explained well how to use it. There are a couple of points that not obvious to me and possibly other readers with no experience with `ContT`.
- for chaining computations, can we document the benefits of `flatMap`ing a chain of `A => ContT` over `andThen`ing a chain of `Kleisli`?
- less importantly, it seems a bit redundant that `ContT.apply` ask for a `(B => M[A]) => M[A])` while user just computes  a `B` and call the `B => M[A]` at the end. Why can't there be a simplified version that takes a `=> B` and construct the `(B => M[A]) => M[A])` for user? something like 
```scala
def defer[M[_], A, B](b: => B): ContT[M, A, B] =  apply(_(b))
```
I suspect that there must be reason but it's just not obvious to me, and probably worth documenting. 
also cc @johnynek @djspiewak @Leammas @rossabaker @ceedubs

djspiewak:
I can’t think of a reason not to have that function. It should work just fine and it sounds quite useful.

johnynek:
@kailuowang that fuction is hiding there. It is `defer(pure(b))`. We can certainly add a helper method for it.

It didn’t occur to me. Sorry.

cb372:
OK, I'll submit a separate PR to add that helper function. It does sounds useful. Once that's merged and released, we can update this documentation to use it.

It's worth pointing out that the `apply` function is still necessary to cater for slightly more exotic use cases of `ContT` in which you use the continuation to fill in a "hole" somewhere in the middle of your function, rather than at the end. e.g.

```scala
ContT.apply { hole =>
  for {
    a <- doFirstThing()
    b <- doSecondThing(a)
    c <- hole(b)
    d <- doFourthThing(c)
  } yield d
}
```

(Gabriel Gonzalez goes into more detail in [this blogpost](http://www.haskellforall.com/2012/12/the-continuation-monad.html).)

As for your first question,

> the benefits of flatMaping a chain of A => ContT over andThening a chain of Kleisli?

I don't really have a good answer to that. To be honest, I find the chain of Kleislis easier to follow and I struggled to come up with a motivating example for `ContT`.

One (rather weak) argument is that `ContT` acts as a form of documentation, making it explicit that "I'm doing CPS here".

johnynek:
One thing to note: foreach is a continuation, forall is a continuation, foldMap is a continuation, etc....

Many of our familiar functions take a `(A => B)` and return a `B`. This is a continuation. They are holes exactly like the example @cb372 made.

kailuowang:
Thanks @cb372. Yeah, the original `apply` is still needed for the `hole` use case.

kailuowang:
I propose we get this merged first and we can address the method addition and document update in a separate PR.

PR #2683, opened 2019-01-03
make the announcement section more prominent.

kailuowang:


kailuowang:
Merged with 1 approval for simple doc rearrangement

PR #2688, opened 2019-01-07
Add a missing T

andyscott:
Adds a missing T.

### DIFF
--- a/core/src/main/scala/cats/data/IorT.scala
+++ b/core/src/main/scala/cats/data/IorT.scala
@@ -443,7 +443,7 @@ abstract private[data] class IorTInstances extends IorTInstances1 {
     }
   }
 
-  implicit def catsDataDeferForIor[F[_], E](implicit F: Defer[F]): Defer[IorT[F, E, ?]] =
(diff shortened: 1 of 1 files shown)


LukaJCB:
Wait, this is potentially binary breaking, is it not? 🤔

LukaJCB:
Yeah seems, like it. I'm gonna file this one for 2.0 :)

mcanlas:
Just driving by. @LukaJCB Is there a resource/document that outlines which changes are binary breaking and which are not? And isn't there a tool that helps with this also?

kailuowang:
Yeah it's BC breaking.

andyscott:
Indeed!

LukaJCB:
@mcanlas Check out mima, it's the tool we use and it also describes what's binary breaking :)

kailuowang:
@mcanlas rule of thumbs is that 
1. you can't change any existing public members. 
2. you can't add a new method to a trait even if you give it a default implementation. 

@andyscott if we really care about this name we could simply change the existing to explicit and add a new implicit one with the correct name, although I am not sure if that's better than a miss-spell.

@LukaJCB this would be BC breaking on scala 2.12 as well. For 2.0 we still want 2.12 BC right?

Issue #2685, opened 2019-01-07
MonadError#adaptError could be on ApplicativeError instead


bplommer:
`MonadError#adaptError` is defined as 
```scala
def adaptError[A](fa: F[A])(pf: PartialFunction[E, E]): F[A] =
  flatMap(attempt(fa))(_.fold(e => raiseError(pf.applyOrElse[E, E](e, _ => e)), pure))
```
But it could be defined on `ApplicativeError` simply as
```scala
def adaptError[A](fa: F[A])(pf: PartialFunction[E, E]): F[A] =
  recoverWith(fa)(pf.andThen(raiseError))
```
I haven't opened a PR because this would be a breaking change.

bplommer:
@kailuowang I'm happy to pick this up. I it will need a different name in the `ApplicativeError` syntax to prevent ambiguity - maybe `applicativeAdaptError`?

kailuowang:
@bplommer thanks, yeah, that's a good catch. `applicativeAdaptError` sounds too long. how about `adaptErr` or `modifyError` with a comment on why it can't be the same as the one in MonadError for now? we shall be able to change this in 2.0.

kubukoz:
Can this be closed? :)

kailuowang:
I still like to move this method in the type class in 2.1

kubukoz:
On the >2.11 branch, right?

kailuowang:
Right, which is master.

travisbrown:
For what it's worth it should have been possible to fix the syntax in 2.0 by making the `MonadErrorOps` `adaptError` package private, but I just tried it for 2.1 and ran into a compiler bug: https://github.com/scala/bug/issues/11787

(`private` or `private[MonadErrorOps]` avoid the bug but break bincompat.)

PR #2689, opened 2019-01-07
Add `orElseRaise` syntax for `ApplicativeError`

kubukoz:


ceedubs:
Thanks, @kubukoz. Would you mind adding an sbt-doctest example for this? That would help motivate it and would also probably help with the code coverage drop.

kailuowang:
This is a nit: 
This method seems closer to `adaptError` to me, except that it ignores the original error. 
`orElse` is usually associated with recovering. We don't have `adaptError` in `ApplicativeError` but we probably should (at least in the syntax) related issue is https://github.com/typelevel/cats/issues/2685
I have a proposal there. maybe we call this one `adaptErrorTo` or `orRaise`? 
I also agree that we can use at least a doctest.

kubukoz:
I'm back on it. `orRaise` sounds good to me!

kubukoz:
@kailuowang can we get this to the finish line? :)

PR #2690, opened 2019-01-08
Override equals and hashCode on Chain

### LINKED ISSUE #2540: Add Chain equals & hashCode methods (via Chain.iterator?)
Currently `Chain[A]` only defines `===` operator and therefore different Chain instances with "same" contents are not `.equals` to each other:
```scala
import cats.data.Chain
import cats.implicits._

(Chain.one(1) |+| Chain.one(2) |+| Chain.one(3)) == Chain.fromSeq(List(1, 2, 3)) // false
(Chain.one(1) |+| Chain.one(2) |+| Chain.one(3)) === Chain.fromSeq(List(1, 2, 3)) // true
```
Not only this leads to confusion, but also could lead to bugs if `Chain[A]` used in `Set` or as `Map` keys, or in `Eq.fromUniversalEquals`.

ceedubs:
Resolves #2540.

### DIFF
--- a/core/src/main/scala/cats/data/Chain.scala
+++ b/core/src/main/scala/cats/data/Chain.scala
@@ -2,6 +2,7 @@ package cats
 package data
 
 import Chain._
+import cats.kernel.instances.StaticMethods
 
 import scala.annotation.tailrec
 import scala.collection.immutable.SortedMap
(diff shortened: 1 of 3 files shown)


simonyun:
thanks all!!! 🙏

Issue #2686, opened 2019-01-07
Provide an IndexedState data type


valydia:
It would be an alias for:
``` scala
type IndexedState[SA, SB, A] = IndexedStateT[Eval, SA, SB, A]
```

PR #2693, opened 2019-01-10
Add IndexedState alias (#2686)

valydia:

### DIFF
--- a/core/src/main/scala/cats/data/package.scala
+++ b/core/src/main/scala/cats/data/package.scala
@@ -43,6 +43,9 @@ package object data {
     def tell[L](l: L): Writer[L, Unit] = WriterT.tell(l)
   }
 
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/data/package.scala
```diff
@@ -43,6 +43,9 @@ package object data {
     def tell[L](l: L): Writer[L, Unit] = WriterT.tell(l)
   }
```
kailuowang: Would you mind adding a doctest for the `IndexedStateFunctions.apply` here?

valydia: Sure! Please see https://github.com/typelevel/cats/pull/2693/commits/b858c14a1b9e6f9a0fb95662a225928cb5a371e5

PR #2694, opened 2019-01-11
added Scala Pets store as an example project

kailuowang:

### DIFF
--- a/docs/src/main/tut/resources_for_learners.md
+++ b/docs/src/main/tut/resources_for_learners.md
@@ -24,8 +24,7 @@ position: 30
 
 # Example Projects
  * [Scala Pet Store](https://github.com/pauljamescleary/scala-pet-store) 
-   An implementation of the java pet store using FP techniques in scala. Stack includes Http4s, Circe,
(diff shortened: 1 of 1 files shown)


docs/src/main/tut/resources_for_learners.md
```diff
@@ -21,6 +21,12 @@ position: 30
  * [Scala Exercises](https://www.scala-exercises.org/cats) by 47 Degrees 
    offers multiple online exercises to learn about the type classes and 
    datatypes included in Cats.
+
+# Examples
+ * [Scala Pets Store](https://github.com/pauljamescleary/scala-pet-store) 
+   An implementation of the java pet store using FP techniques in scala. Stack includes Http4s, Circe,
+   Doobie, Cats, ScalaCheck, and PureConfig.
+
```
pauljamescleary: `Demonstrates how to apply FP techniques using Typelevel projects, while maintaining approachability for people with an OO background.  Uses tagless final for the core domain, and uses popular Typelevel libraries including Cats, Http4s, Circe and Doobie.`

pauljamescleary: 👍

PR #2696, opened 2019-01-12
Add a `defer` factory method to `ContT`

cb372:
As suggested in #2677.

Similar to the existing `pure` function, but evaluation of the argument is deferred.

This is useful for building a computation which calls its continuation as the final step. Instead of writing:

```scala
ContT.apply { next =>
  val x = foo()
  val y = bar(x)
  val z = baz(y)
  next(z)
}
```

you can write:

```scala
ContT.defer {
  val x = foo()
  val y = bar(x)
  baz(y)
}
```


ceedubs:
Thanks @cb372! Would you be willing to add a ScalaDoc description of the method and a unit test or two? One unit test that I think would be good is one that checks that the argument isn't eagerly evaluated.

PR #2698, opened 2019-01-13
Add `adaptErr` to `ApplicativeErrorOps`

bplommer:
Partially addresses https://github.com/typelevel/cats/issues/2685.

I've changed `MonadError#adaptError` to be an alias for `ApplicativeErrorOps#adaptErr`, but I wasn't sure whether I should do this or provide a duplicate implementation.

### DIFF
--- a/core/src/main/scala/cats/MonadError.scala
+++ b/core/src/main/scala/cats/MonadError.scala
@@ -1,7 +1,5 @@
 package cats
 
-import cats.syntax.ApplicativeErrorOps
-
 /**
  * A monad that also allows you to raise and or handle an error value.
  *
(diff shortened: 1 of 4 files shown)


kailuowang:
thanks so much for taking on this! Here is some feedback 
1. I would duplicate the code with the same comment you have on syntax. 
2. I would leave the laws unchanged and leave that to the future PR that moves the method to the `ApplicativeError`. The change breaks binary compatibility in cats-laws  (which we don't check at the moment), and conventionally everything in `XXXXlaws` should be in `XXXXTests`, which we can't do since it's in the syntax ops.  Instead, we can add a doctest to the syntax method in `ApplicativeErrorOps`

kailuowang:
Thanks @bplommer

LukaJCB:
Looks great, thanks @bplommer! 😊

Issue #2687, opened 2019-01-07
Provide an IndexedStore


valydia:
Analogue to https://github.com/scalaz/scalaz/blob/7e64f43872fc944b54a63e4262acaea2248f50c8/core/src/main/scala/scalaz/StoreT.scala

valydia:
I am happy to provide an implementation if that's useful

PR #2704, opened 2019-01-21
Implement IndexedStoreT (#2687)

valydia:


core/src/main/scala/cats/data/package.scala
```diff
@@ -58,6 +58,12 @@ package object data {
   type State[S, A] = StateT[Eval, S, A]
   object State extends StateFunctions
 
+  type StoreT[F[_], A, B] = IndexedStoreT[F, A, A, B]
+  object StoreT extends StoreTFunctions
+
+  type StoreId[A, B] = StoreT[Eval, A, B]
```
valydia: Ideally this should have been called `Store` but it clashes with [this](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/package.scala#L80)

tests/src/test/scala/cats/tests/IndexedStoreTSuite.scala
```diff
@@ -0,0 +1,118 @@
...
+
+sealed trait IndexedStoreTSuiteInstances {
+
+  implicit def indexedStoreTEq[F[_], I, A, B](implicit I: Arbitrary[I],
+                                              EqI: Eq[I],
+                                                FAB: Eq[F[A => B]]): Eq[IndexedStoreT[F, I, A, B]] =
+    Eq.by[IndexedStoreT[F, I, A, B], (F[A => B], I)](_.run)
+}
+
+/** This data type exists purely for testing.
```
valydia: Does this class need to go in its own file?

core/src/main/scala/cats/data/package.scala
LukaJCB: I'm really ignorant to this, but are these two `Store` implementations equivalent to each other?

tests/src/test/scala/cats/tests/IndexedStoreTSuite.scala
LukaJCB: I think we have precedent here with `ListWrapper` being its own file, so I think it could go into its own file, yeah :)

kailuowang:
Apologies, I didn't think too much when I thumbed up your original issue #2687, I assumed that you are going to build it on top of `RepresentableStore`. Maybe that's tricky or impossible?  cc @eli-jordan who added `RepresentableStore`.

eli-jordan:
Thanks for the ping @kailuowang 
@valydia I *think* you can derive `StoreT` in a way that is more general than what is in this PR, by replacing `F[A => B]` with `F[G[A]]` where `G` is `Representable` 

Since the identity representable is `Function1` https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/instances/function.scala#L16 this appears to make sense. However, I haven't attempted an implementation so it might not work out.

tests/src/test/scala/cats/tests/IndexedStoreTSuite.scala
valydia: Cool 👍

core/src/main/scala/cats/data/package.scala
valydia: It might be, see the discussion below!

valydia:
Hey! Thanks guys for your input, I can see I can make the `F[G[A]]` work... The question is  should we link those 2 implementations together? It seems like in Haskell, they are keeping it separate see [here](http://hackage.haskell.org/package/representable-functors-3.2.0.2/docs/Control-Comonad-Representable-Store.html) and [here](http://hackage.haskell.org/package/comonad-5.0.4/docs/Control-Comonad-Store.html)...

LukaJCB:
Do you have an idea what the downsides or upsides of having them separate vs unified would be? I have a lot of reading to do before I fully understand them.

valydia:
Not off the top of my head, I am afraid... I'll do a bit of investigation on both the justification and the implementation side of it!

valydia:
I've made an attempt in term if implementation [here](https://github.com/valydia/cats/pull/1/files)... Is that what you meant @eli-jordan ? I could implement some of the methods, but I got blocked at the definition of `coflatMap`, if we want a meaningful implementation, `coflatMap` should be defined in the [base class](https://github.com/valydia/cats/blob/indexedstore_representable/core/src/main/scala/cats/data/IndexedStoreT.scala#L4) (otherwise, you could argue if it should be [there](https://github.com/valydia/cats/blob/indexedstore_representable/core/src/main/scala/cats/data/IndexedStoreT.scala#L52), we may as well provide a complete different implementation (i.e. this PR)...). I couldn't define a correct signature of `coflatMap`, it looked like `G` requires 2 types parameters (analogous to `Function1`...)

valydia:
Following up on my previous comment, I don't think it is possible to build the `IndexedStoreT` on top of the `RepresentableStore` - or I am missing the point (in which case I could use some help 😄 ). Also from my researches, other libraries (Haskell, Scalaz) adopt the same approach as in this PR, as a matter of fact ( this pr is based on Scalaz ). So I would made the case for this current approach... What do you think?

eli-jordan:
@valydia I have been traveling, so just saw this. I'll take a sometime look over the weekend and see if I can figure out how coflatMap should be defined. I think it should be possible.

valydia:
Awesome thanks @eli-jordan ! I hope you had a nice time 👍

eli-jordan:
```scala
case class StoreT[G[_], W[_], A](wga: W[G[A]], a: A)(implicit W: Comonad[W], R: Representable.Aux[G, A]) {

  def extract: A = a

  def map[B](f: A => B)(implicit R0: Representable.Aux[G, B]): StoreT[G, W, B] =
    StoreT(W.map(wga)(ga => R.F.map(ga)(f)), f(a))

  def coflatten(implicit R0: Representable.Aux[G, StoreT[G, W, A]]): StoreT[G, W, StoreT[G, W, A]] =
    StoreT(W.coflatMap(wga) { x: W[G[A]] =>
      R.tabulate(aa => StoreT(x, aa))
    }, this)

  def coflatMap[B](f: StoreT[G, W, A] => B)
            (implicit R0: Representable.Aux[G, StoreT[G, W, A]], R1: Representable.Aux[G, B]): StoreT[G, W, B] =
    coflatten.map(f)
}
```

This works. I have simplified it a bit, but I think it shows the core of the implementation. There is probably a better way to encode the implicits so all the `Representable` instances don't need to be passed in, but I haven't spent the time to figure that out.

valydia:
Thanks for the help, in the light of this solution, I'll try to implement the other method

valydia:
I've been trying to extend this approach to `IndexedStoreT` without any success. @eli-jordan Do you see a way to do so?

larsrh:

Issue #2663, opened 2018-12-05
Add mkString_ that just takes separator string 


matthughes:
Cats provides a Show-compatible version of mkString but it requires all three parameters: prefix, separator, suffix.  The std library also provides a mkString that just takes a separator string; this is what I use 90% of the time. 

Can we add this?  I'd be happy to PR.

calvinbrown085:
Hi @matthughes did you end up working on this?

matthughes:
@calvinbrown085 I have not.  I wanted to see if there was interest before doing the PR as it's pretty simple.  Really just adding a new method that delegates to the form:

```scala
def _mkString(separator: String): String = 
  _mkString("", separator, "")
```

kailuowang:
@calvinbrown085 I think this would be a good addition. And it's rather simple, the code you wrote and a doctest should do.

dwijnand:
Fixed by https://github.com/typelevel/cats/pull/2709

PR #2709, opened 2019-01-25
Adds version of mkString_ with no prefix/suffix, matching the std lib.

matthughes:

### DIFF
--- a/core/src/main/scala/cats/syntax/foldable.scala
+++ b/core/src/main/scala/cats/syntax/foldable.scala
@@ -77,7 +77,9 @@ final class FoldableOps[F[_], A](private val fa: F[A]) extends AnyVal {
     A.combine(prefix, A.combine(F.intercalate(fa, delim), suffix))
 
   /**
-   * Make a string using `Show`, named as `mkString_` to avoid conflict
+   * Make a string using `Show`, prefix, delimiter, and suffix.
+   *
+   * Named as `mkString_` to avoid conflict.
    *
    * Example:
    * {{{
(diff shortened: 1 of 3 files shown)


matthughes:
Fixes #2663 but as it stands, this is going to fail binary compatibility rules.  Not sure how best to get around?  I could change the name but I got no suggestions.

kailuowang:
I am not sure if this is going to break binary compatibility. Right now the build break due to format issue, a `sbt prePR` will format the code and also check if it breaks binary compatibility.

kailuowang:
@matthughes thanks!
The short answer is: yes, please move it `FoldableOps0` and I'd perfer call `new FoldableOps(f).mkString_("", delim, "")`

The longer answer: 
ah, it looks like adding an override to a method also breaks BC. Good to learn something new.  
`FoldableOps0` and ``FoldableOps1` were created recently to get around binary incompatibility. Mima is happy when you move the method to `FoldableOps0` because that one is not officially released yet, it's released on 1.6.0-RC1, and mima doesn't check BC against RC releases. If we can make this one in 1.6.0 we should be good. Just noticed that your PR is against 1.6.x branch rather than master, so we are good on that as well.

mpilquist:
Note: adding an overload to a regular class doesn't break binary compat but doing so to a value class does. The existence of an overload changes the scheme used for generation of the synthetic method names in the companion class.

Consider these two classes, representing 2 different versions of the same class:

```scala
class Foo1(val x: Int) extends AnyVal {
  def foo(y: Int): Int = x + y
}

class Foo2(val x: Int) extends AnyVal {
  def foo(y: Int): Int = x + y
  def foo(y: Int, z: Int): Int = x + y + z
}
```

`Foo1.class` and `Foo2.class` are binary compatible (modulo the class name difference of Foo1 vs Foo2). However, `Foo1$.class` and `Foo2$.class` are not.

`Foo1$.class` has this method:
```
  public final int foo$extension(int, int);
    descriptor: (II)I
    flags: ACC_PUBLIC, ACC_FINAL
```

`Foo2$.class` has these methods:
```
  public final int foo$extension0(int, int);
    descriptor: (II)I
    flags: ACC_PUBLIC, ACC_FINAL
<snip>

  public final int foo$extension1(int, int, int);
    descriptor: (III)I
    flags: ACC_PUBLIC, ACC_FINAL
```

kailuowang:
@mpilquist ah, that makes sense, thanks for the detail! In hindsight, maybe the first one should've been made `foo$extension0` rather than `foo$extension`. I'll see if I can find an existing scalac issue for this one.

kailuowang:
@matthughes I updated the CatsSuite which is missing the newly added syntax trait and caused the broken build. Hopefully this one will be ready to merge soon.

Issue #2708, opened 2019-01-23
Provide a Semigroupal instance for SortedSet


LukaJCB:
Should be something like this:

```scala
def product[A, B](fa: SortedSet[A], fb: SortedSet[B]): SortedSet[(A,B)] = {
  implicit val orderingA = fa.ordering
  implicit val orderingB = fb.ordering
    
  fa.flatMap(a => fb.map(b => a -> b))
}
```

calvinbrown085:
Hi @LukaJCB I'd be happy to give this a shot, would that go in here? https://github.com/typelevel/cats/blob/774fb51d1365e1adef7fa71f09b1410941264f60/core/src/main/scala/cats/instances/sortedSet.scala#L9

LukaJCB:
Unfortunately I think it'd need to go in another trait for binary compatability.

calvinbrown085:
@LukaJCB if have a second, do you mind pointing me in the right direction? Or by `another trait` do you mean a new file? That we can create for bin compat issues

LukaJCB:
Something like this one: 
https://github.com/typelevel/cats/blob/774fb51d1365e1adef7fa71f09b1410941264f60/core/src/main/scala/cats/instances/option.scala#L130

You can't extend the trait directly so we mix it into the objects that need it, i.e. `cats.implicits` and `cats.instances.sortedSet`. :)

calvinbrown085:
Perfect, that's what I figured :)

calvinbrown085:
So I have that PR, I'm a bit confused on why it's failing the bin compat check

vladislavGutov:
Hi @LukaJCB 
Seems like this one was forgotten.
I created PR https://github.com/typelevel/cats/pull/2840. When you have time, could look at it, please?

calvinbrown085:
@vladislavGutov Thank you! I haven't had time to look at this unfortunately. I'll close this PR.

kailuowang:
closed by #2840

PR #2710, opened 2019-01-25
Adding product function for sorted set instances

calvinbrown085:
https://github.com/typelevel/cats/issues/2708
I ran `sbt prePR`, i'm kind of confused why it's failing the bin compatibility check if anyone has a second to look :)

### DIFF
--- a/core/src/main/scala/cats/instances/all.scala
+++ b/core/src/main/scala/cats/instances/all.scala
@@ -42,6 +42,7 @@ trait AllInstancesBinCompat1
     with StreamInstancesBinCompat0
     with MapInstancesBinCompat0
     with SortedMapInstancesBinCompat0
+    with SortedSetInstancesBinCompat0
 
 trait AllInstancesBinCompat2 extends DurationInstances with FiniteDurationInstances
 
--- a/core/src/main/scala/cats/instances/all.scala
+++ b/core/src/main/scala/cats/instances/all.scala
@@ -42,8 +42,9 @@ trait AllInstancesBinCompat1
     with StreamInstancesBinCompat0
     with MapInstancesBinCompat0
     with SortedMapInstancesBinCompat0
-    with SortedSetInstancesBinCompat0
 
 trait AllInstancesBinCompat2 extends DurationInstances with FiniteDurationInstances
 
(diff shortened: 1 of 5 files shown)


core/src/main/scala/cats/instances/all.scala
```diff
@@ -42,6 +42,7 @@ trait AllInstancesBinCompat1
     with StreamInstancesBinCompat0
     with MapInstancesBinCompat0
     with SortedMapInstancesBinCompat0
+    with SortedSetInstancesBinCompat0
```
LukaJCB: This is the problem right here, you can't add anything to existing traits, only final classes and objects :)

calvinbrown085: Ah, Do I just need to add `AllInstancesBinCompat4` that extends `SortedSetInstancesBinCompat0` ?

LukaJCB: Yup,I believe so

LukaJCB:
Can you also mix it into `cats.instances.all` and `cats.implicits`?

calvinbrown085:
@LukaJCB Yep!

tests/src/test/scala/cats/tests/SortedSetSuite.scala
```diff
@@ -2,13 +2,17 @@ package cats
...
+import cats.laws.discipline.{FoldableTests, SemigroupKTests, SemigroupalTests, SerializableTests}
 import cats.kernel.laws.discipline.{BoundedSemilatticeTests, HashTests, PartialOrderTests}
 import cats.laws.discipline.arbitrary._
 
 import scala.collection.immutable.SortedSet
 
 class SortedSetSuite extends CatsSuite {
+
+  checkAll("SortedSet[Int]", SemigroupalTests[SortedSet].semigroupal[Int, Int, Int])
+  checkAll("Semigroupal[SortedSet]", SerializableTests.serializable(Semigroupal[SortedSet]))
```
LukaJCB: `Semigroupal[SortedSet]` isn't in scope here :)

LukaJCB: You can mix it into `CatsSuite`

PR #2713, opened 2019-01-29
Switch to Scala CoC

LukaJCB:
The Scala Code of Conduct is an improvement to the original CoC, [endorsed by Typelevel](https://typelevel.org/blog/2016/12/17/scala-coc.html) and other downstream projects have already made the switch (fs2, cats-effect, http4s).


LukaJCB:
Looks like we have overwhelming support :+1:

PR #2722, opened 2019-02-02
Add more doctests to Kleisli

nasadorian:
Re: Issue #2479 -- Appreciate any feedback to improve or further clarify examples.

Adds  comments and some doctests for *Kleisli*:
* dimap
* map
* andThen
* local
* liftF
* pure
* ask

### DIFF
--- a/core/src/main/scala/cats/data/Kleisli.scala
+++ b/core/src/main/scala/cats/data/Kleisli.scala
@@ -12,9 +12,21 @@ final case class Kleisli[F[_], A, B](run: A => F[B]) { self =>
   def ap[C](f: Kleisli[F, A, B => C])(implicit F: Apply[F]): Kleisli[F, A, C] =
     Kleisli(a => F.ap(f.run(a))(run(a)))
 
+  /**
+   * Performs [[local]] and [[map]] simultaneously.
+   */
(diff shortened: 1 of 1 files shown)


core/src/main/scala/cats/data/Kleisli.scala
```diff
@@ -164,12 +210,33 @@ sealed private[data] trait KleisliFunctions {
...
+   * scala> pureOpt.run(())
+   * res0: Option[String] = Some(beam me up!)
+   * }}}
+   */
   def pure[F[_], A, B](x: B)(implicit F: Applicative[F]): Kleisli[F, A, B] =
     Kleisli(_ => F.pure(x))
 
+  /**
+   * Similar to [[pure]] except the input type is the same as the `A` in output type `F[A]`.
+   */
```
ceedubs: This comment is true, but I think that it could be more illustrative. In my mind, `ask` is to `Kleisli` as `identity` is to `Function1`. I think that it would be helpful to point out that in the resulting `Kleisli`, the output `A` is the same as the input `A` (the actual value not just the type). A code example might help get this point across.

nasadorian: Just updated and added an example to `ask`

nasadorian:
Okay @kailuowang I got the build to pass, looks like my commits overrode your approval 😅

PR #2724, opened 2019-02-04
adding helper method `toFreeT` to `Free`

jcouyang:
so any `Free[S, A]` can be converted to `FreeT[S, G, A]` if there is an Applicative instance of G

### DIFF
--- a/free/src/main/scala/cats/free/Free.scala
+++ b/free/src/main/scala/cats/free/Free.scala
@@ -183,12 +183,12 @@ sealed abstract class Free[S[_], A] extends Product with Serializable {
   final def inject[G[_]](implicit ev: InjectK[S, G]): Free[G, A] =
     mapK(λ[S ~> G](ev.inj(_)))
 
-  final def toFreeT: FreeT[S, Id, A] =
-    foldMap[FreeT[S, Id, ?]] { // this is safe because Free is stack safe
-      new (S ~> FreeT[S, Id, ?]) {
-        def apply[B](fa: S[B]): FreeT[S, Id, B] = FreeT.liftF((fa))
+  final def toFreeT[G[_]: Applicative]: FreeT[S, G, A] =
+    foldMap[FreeT[S, G, ?]] {
+      new (S ~> FreeT[S, G, ?]) {
+        def apply[B](fa: S[B]): FreeT[S, G, B] = FreeT.liftF(fa)
       }
-    }(FreeT.catsFreeMonadForFreeT[S,Id])
+    }(FreeT.catsFreeMonadForFreeT[S, G])
 
 
   override def toString: String =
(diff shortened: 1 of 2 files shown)


ceedubs:
Thanks @jcouyang! It looks like scalafmt is complaining about some whitespace. If you run `sbt prePR` it should resolve the issue.

jcouyang:
@ceedubs I ran `sbt prePR` and everything was ok, weird, just ran `sbt scalafmt` see if it will fix the format

free/src/main/scala/cats/free/Free.scala
```diff
@@ -183,6 +183,13 @@ sealed abstract class Free[S[_], A] extends Product with Serializable {
   final def inject[G[_]](implicit ev: InjectK[S, G]): Free[G, A] =
     mapK(λ[S ~> G](ev.inj(_)))
 
+  final def toFreeT[G[_]: Applicative]: FreeT[S, G, A] =
+    foldMap[FreeT[S, G, ?]] {
+      new (S ~> FreeT[S, G, ?]) {
+        def apply[B](fa: S[B]): FreeT[S, G, B] = FreeT.liftF(fa)
+      }
+    }(FreeT.catsFreeMonadForFreeT[S, G])
```
kailuowang: This can probably be simplified as 
```scala
foldMap( λ[S ~> FreeT[S, G, ?]](FreeT.lift(_))
```

free/src/test/scala/cats/free/FreeSuite.scala
```diff
@@ -45,6 +45,10 @@ class FreeSuite extends CatsSuite {
     rr.toString.length should be > 0
   }
 
+  test("toFreeT is stack-safe") {
```
kailuowang: Can we move this one down to be grouped with the other stack safety test?

jcouyang:
thanks @kailuowang, comment addressed, please have a look :)

kailuowang:
so just `foldMap(λ[S ~> FreeT[S, G, ?]](FreeT.lift(_))` (without the explicit type and instance param) wouldn't compile for you?

jcouyang:
🤔 not sure why it works now...looks like λ magic
the compile can't find the correct monad instance when it was `new (S ~> FreeT[S, Id, ?])` before so I explicitly specify that a FreeT Monad for S and G

anyway, removing the explicit paramter in mapK too

jcouyang:
hi @kailuowang can you please help review the latest change?

LukaJCB:
Hey @jcouyang, sorry for not getting to this earlier. Kai had his phone stolen on vacation and I've been away as well. Anyway, this looks good to me :) Will merge this soon!

jcouyang:
🙀 oh no, sorry to hear that

kailuowang:
@jcouyang thanks a lot! as Luka mentioned, sorry about the delay.

PR #2726, opened 2019-02-05
Add `TraverseFilter` instance for `Set` to alleycats

dcastro:


dcastro:
@LukaJCB Thanks! :)
I just realized this should probably also be done for `Map`s, what do you think? If so, should I add it here or in a different PR?

LukaJCB:
Good call, I think it's fine to add it to this PR if you like :)

dcastro:
@LukaJCB Apologies for the delay, it's done now :) I just reused the code from the `TraverseFilter` instance for `SortedMap`.

PR #2727, opened 2019-02-08
Add grouped method to foldable

frosforever:
This PR adds `grouped` on `F : Foldable : Alternative`. It's the deviates from the standard library's [IterableLike.grouped](https://www.scala-lang.org/api/2.12.3/scala/collection/IterableLike.html#grouped(size:Int):Iterator[Repr]) by returning an empty list if `size <= 0` instead of throwing.

This is a first pass. If there's interest, I'd be happy to flush this out further. 

TODO: 
- [ ] This may introduce binary compatibility issues and may need a similar treatment as https://github.com/typelevel/cats/pull/2641/commits/f36ce95e133b3e93e6afe4a27f0e798a43ce41e0
- [ ] Tests:
  - Match std lib's `grouped` save for invalid `size`
  - `F.grouped.flatten == F` or `F.grouped.unite == F`

Thanks in advance for your time reviewing!

### DIFF
--- a/core/src/main/scala/cats/Foldable.scala
+++ b/core/src/main/scala/cats/Foldable.scala
@@ -582,6 +582,32 @@ import Foldable.sentinel
 
   override def unorderedFoldMap[A, B: CommutativeMonoid](fa: F[A])(f: (A) => B): B =
     foldMap(fa)(f)
+
+  /**
+   * Partitions elements in fixed size.
+   * Returns empty `F` if `size <= 0`.
+   *
+   * Example:
+   * {{{
+   * scala> import cats.implicits._
+   * scala> val x: List[Int] = List(1, 2, 3, 4, 5)
+   * scala> Foldable[List].grouped(x)(2)
+   * res0: List[List[Int]] = List(List(1, 2), List(3, 4), List(5))
+   * }}}
+   */
+  def grouped[A](fa: F[A])(size: Int)(implicit F: Alternative[F]): F[F[A]] =
+    // Need to check on empty as otherwise the final combine would return F[Empty] rather than just Empty
+    if (size <= 0 || isEmpty(fa)) {
+      F.empty
+    } else {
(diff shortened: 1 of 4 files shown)


core/src/main/scala/cats/Foldable.scala
```diff
@@ -582,6 +582,32 @@ import Foldable.sentinel
...
+   * scala> val x: List[Int] = List(1, 2, 3, 4, 5)
+   * scala> Foldable[List].grouped(x)(2)
+   * res0: List[List[Int]] = List(List(1, 2), List(3, 4), List(5))
+   * }}}
+   */
+  def grouped[A](fa: F[A])(size: Int)(implicit F: Alternative[F]): F[F[A]] =
+    // Need to check on empty as otherwise the final combine would return F[Empty] rather than just Empty
+    if (size <= 0 || isEmpty(fa)) {
+      F.empty
+    } else {
```
LukaJCB: Potentially we could also check for `size = 1` and just map using `pure`, though not sure why someone would do that :smile:

LukaJCB:
Another property we could test is `F.fold(F.grouped(fa, n)) === fa

kailuowang:
Right after 2.0 release at the end of Q1 we will have the master branch dropping 2.11 support, at which point we can merge this to master - then we'll need to backport it to a new scala2_11 branch as a syntax trait. Later we are going to release Cats 2.1-RC1 from those two branches.

frosforever:
I think one would really want the optimization for instances as without it performance will suffer. Does it make sense to get this in now, without the optimizations and then have a new version on 2.0 with the optimizations or better off to just wait?

kailuowang:
@frosforever sorry about the delayed response. I think are better off to wait.

frosforever:
@kailuowang no worries! Do we have an issue label to track PRs that will have to updated for version 2? I don't yet see a version 2 branch?
Is it best to keep this open in the meantime or close until a v2 branch is available?

frosforever:
@travisbrown Thanks for the review and all good points! I'm all for closing this.

PR #2728, opened 2019-02-11
Optimize productR in Apply

maphi:
Further improvement similar to #2597 . Resolves problems like this: https://github.com/typelevel/cats-effect/issues/401 .

Reference to `a` was kept unnecessarily by prev implementation with map2:
```
//Expanding productR: map2 => product => ap => flatMap

fa *> fb == 
  fa
    .map(a => (b: B) => (a,b))
    .flatMap(f => fb.map(f)).map((_, b) => b)
```


kailuowang:
sorry about the delay, I missed this one while on vacation.

PR #2729, opened 2019-02-13
Add semiflatMap to kleisli

frosforever:
Here's another helpful method on Kleisli I've been reaching for recently. Very open to name changes!

Thanks for your time reviewing!

### DIFF
--- a/core/src/main/scala/cats/data/Kleisli.scala
+++ b/core/src/main/scala/cats/data/Kleisli.scala
@@ -45,6 +45,9 @@ final case class Kleisli[F[_], A, B](run: A => F[B]) { self =>
   def flatMapF[C](f: B => F[C])(implicit F: FlatMap[F]): Kleisli[F, A, C] =
     Kleisli.shift(a => F.flatMap(run(a))(f))
 
+  def semiflatMap[N[_], C](f: F[B] => Kleisli[N, A, C]): Kleisli[N, A, C] =
+    Kleisli(a => f(run(a)).run(a))
+
   /**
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/data/Kleisli.scala
```diff
@@ -45,6 +45,9 @@ final case class Kleisli[F[_], A, B](run: A => F[B]) { self =>
   def flatMapF[C](f: B => F[C])(implicit F: FlatMap[F]): Kleisli[F, A, C] =
     Kleisli.shift(a => F.flatMap(run(a))(f))
 
+  def semiflatMap[N[_], C](f: F[B] => Kleisli[N, A, C]): Kleisli[N, A, C] =
```
LukaJCB: Thanks a ton for opening this, it looks like a great addition! 
Unfortunately we already have a few methods named `semiflatMap` that have a different signature (in this case it'd be `Kleisli[F, A, B] => (B => F[C]) => Kleisli[F, A, C]`, but we already have that as `flatMapF`). 

This one looks a bit like `StateT#transformF`, but also not quite. 
I'm not quite sure what the best name is here, I'll try to think about it more :)

frosforever: I know, I added those https://github.com/typelevel/cats/pull/999 😄 and I'm highly unoriginal at naming. At first I thought this would be `flatMapF` because it's nearly `flatMap` but takes an `F[B]` and `flatMapF` would be `semiFlatMap` but changing that is more trouble than it's worth.

larsrh:

PR #2730, opened 2019-02-13
Do not include parent dir of root in build

frosforever:
At the moment, the parent of the root directory is included in the build due to `CrossType.Pure.sharedSrcDir(baseDirectory, "main")` being called on the root (cats) project whose base directory is `"."`. See the output of 
```
sbt:cats> show unmanagedSourceDirectories
...
[info] Compile / unmanagedSourceDirectories
[info] 	List(/Github/cats/src/main/scala-2.12, /Github/cats/src/main/scala, /Github/cats/src/main/java, /Github/src/main/scala-2.12-)
```
Notice that `/Github/src/main` is included in the build!

This PR limits the adding of cross type confs to the `root` build. It would be nice to reference `(cats/modueName).value` instead of the string `"root"` but that results in a circular dependency.

After this build the output is
```
...
[info] Compile / unmanagedSourceDirectories
[info] 	List(/Github/cats/src/main/scala-2.12, /Github/cats/src/main/scala, /Github/cats/src/main/java)
```
with nothing under `Github/` referenced besides `Github/cats`.

I'm not an SBT expert and this is a complicated build so I'm very open to better ways of doing this!

Thanks for reviewing.

See https://youtrack.jetbrains.com/issue/SCL-14951#focus=streamItem-27-3290672-0-0 for context

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -18,12 +18,16 @@ lazy val commonSettings = Seq(
     val bd = baseDirectory.value
     def extraDirs(suffix: String) =
       CrossType.Pure.sharedSrcDir(bd, "main").toList.map(f => file(f.getPath + suffix))
-    CrossVersion.partialVersion(scalaVersion.value) match {
-      case Some((2, y)) if y <= 12 =>
-        extraDirs("-2.12-")
-      case Some((2, y)) if y >= 13 =>
-        extraDirs("-2.13+")
-      case _ => Nil
+    if (moduleName.value == "root") {
+      Seq.empty
+    } else {
+      CrossVersion.partialVersion(scalaVersion.value) match {
+        case Some((2, y)) if y <= 12 =>
+          extraDirs("-2.12-")
+        case Some((2, y)) if y >= 13 =>
+          extraDirs("-2.13+")
+        case _ => Nil
+      }
     }
   },
   coverageEnabled := {


frosforever:
Hmm, I'm unsure what this build failure means. Is this just a timeout or did I break something here? https://travis-ci.org/typelevel/cats/jobs/492826785
```
...
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

kailuowang:
Thanks for fixing this! On my commuter cats is in a dedicated folder so I didn't notice this. The build is just some random issue (likely on the Travis side) I just restarted it.

frosforever:
Hmm, now that I'm thinking about it, perhaps the fix here is to not include `catsSettings` in the root `cats` module (https://github.com/typelevel/cats/blob/663f654707cc64497e780688dd1648f99be4a9f5/build.sbt#L393). 
My understanding is `cats` is only an aggregation module in which case the `catsSettings` applied to each individual module should be enough. Is that correct or are there settings from `catsSettings` that would be needed in the root module?

kailuowang:
It's needed for cross releasing with sbt-release (mostly just for the `crossScalaVersions` settings I believe). However, I think we should switch to [these release steps](https://github.com/typelevel/sbt-catalysts/blob/master/src/main/scala/sbtcatalysts/CatalystsPlugin.scala#L445-L459) and then set the `crossScalaVersions` to Nil at the root module. The only thing is that this change can only be tested with a real release. If you want, you can make this change and I will merge and then test with a release (probably to an obscure version like `0.0.1-TESTRELEASE`)

frosforever:
Sorry, @kailuowang I'm not sure I'm following.
> If you want, you can make this change and I will merge and then test with a release

Is that referring to what I've submitted here or to the release steps from sbt-catalysts you linked?
If the latter, I'll need some time to understand what's going on there. 

Thanks!

kailuowang:
@frosforever I meant the changed I was suggesting. Sorry I didn't tell you the whole story about cross releasing with sbt-release. I'll make the changes in a separate PR. We can keep this one open and see if that one fixes this issue.

frosforever:
No worries! Happy to help any way I can. Feel free to reach out if there's something I can contribute.

PR #2731, opened 2019-02-13
Add productR/L to semi/subflatMap in EitherT

ikempf:
I am using cats-effect and regularly find myself writing effectful code (involving F[Unit]) such as :

```scala
EitherT
  .right[String](logger.debug(show"Logging $foo"))
  .subflatMap(_ => parseValue("value"))
```

I think it would be quite useful to have `Functor.as` and `Apply.productR/L` like functionality on some of the monad transformer functions to explicitly drop the unused `Unit` return values. This would allow the better express the above code in the following way :

```scala
EitherT
  .right[String](logger.debug(show"Logging $foo"))
  .subProductR(parseValue("value"))
```  

If this seems interesting to anyone I would also consider reviewing other EitherT functions such as`flatMapF`, `leftSemiflatMap`, `biSemiflatMap`

Of course `OptionT` is also a candidate.

### DIFF
--- a/core/src/main/scala/cats/data/EitherT.scala
+++ b/core/src/main/scala/cats/data/EitherT.scala
@@ -1,7 +1,6 @@
 package cats
 package data
 
-import cats.Bifunctor
 import cats.instances.either._
 import cats.syntax.either._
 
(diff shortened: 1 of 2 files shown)


larsrh:

PR #2732, opened 2019-02-15
new better release settings

kailuowang:
The old release steps are from the sbt 0.13 era. Now that modular crossScalaVersions issue is improved in sbt 1.x, we need to switch to this new release steps - i.e. the root module should have a `Nil` `crossScalaVersions` setting, then each module can have their own `crossScalaVersions` settings and release accordingly.
This change also reduced significantly the number of unnecessary settings on the root module - which fixes #2730  

To test this build code, I released a `0.0.1-DONOTUSE`, and tested with cats-tagless

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -391,7 +391,7 @@ lazy val cats = project
   .in(file("."))
   .settings(moduleName := "root", crossScalaVersions := Nil)
   .settings(publishSettings) // these settings are needed to release all aggregated modules under this root module
-  .settings(noPublishSettings) // this is to exclue the root module itself from being published. 
+  .settings(noPublishSettings) // this is to exclue the root module itself from being published.
   .aggregate(catsJVM, catsJS)
   .dependsOn(catsJVM, catsJS, tests.jvm % "test-internal -> test")
 
--- a/build.sbt
+++ b/build.sbt
@@ -390,8 +390,8 @@ lazy val docs = project
 lazy val cats = project
   .in(file("."))
   .settings(moduleName := "root", crossScalaVersions := Nil)
-  .settings(publishSettings)
-  .settings(noPublishSettings)
+  .settings(publishSettings) // these settings are needed to release all aggregated modules under this root module
+  .settings(noPublishSettings) // this is to exclue the root module itself from being published. 
   .aggregate(catsJVM, catsJS)
   .dependsOn(catsJVM, catsJS, tests.jvm % "test-internal -> test")
 
--- a/build.sbt
+++ b/build.sbt
@@ -804,7 +804,6 @@ def priorTo2_13(scalaVersion: String): Boolean =
   }
 
 lazy val sharedPublishSettings = Seq(
-  releaseCrossBuild := true,
   releaseTagName := tagName.value,
   releaseVcsSign := true,
   useGpg := true, // bouncycastle has bugs with subkeys, so we use gpg instead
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -389,8 +389,8 @@ lazy val docs = project
 
 lazy val cats = project
   .in(file("."))
-  .settings(moduleName := "root")
-  .settings(catsSettings)
+  .settings(moduleName := "root", crossScalaVersions := Nil)
+  .settings(publishSettings)
```
ceedubs: It looks a little weird to have `publishSettings` followed by `noPublishSettings`. This is intentional, right? If so, it might be helpful to have a brief comment about why this is being done.

build.sbt
```diff
@@ -825,12 +824,11 @@ lazy val sharedReleaseProcess = Seq(
   releaseProcess := Seq[ReleaseStep](
     checkSnapshotDependencies,
     inquireVersions,
-    runClean,
-    releaseStepCommand("validate"),
+    releaseStepCommandAndRemaining("+validate"),
```
ceedubs: Is this a switch from running `validate` from the current version to all versions? If so, does this lead to OOMs or anything?

frosforever:
Verifying this does indeed fix #2730. Thanks @kailuowang!

build.sbt
kailuowang: Good catch. I added some comment here. The basic thing is that `sbt release` command is run through aggregation from this root module. Thus it requires some release settings, then the `noPublishSettings` makes sure the module itself is not published.

build.sbt
kailuowang: Not really. Before the validate is still run against all scala versions because the `crossScalaVersions` settings on the root module and the `releaseCrossBuild := true`. sbt release automatically runs certain steps on each modules against all crossScalaVersions set on the root module. Now, with the new sbt version , the `+validate` will performe on each aggregated module correctly on their own `crossScalaVersions`. The root module's `crossScalaVersions` should set as `Nil`, otherwise it will conflict with the aggregrated module's `crossScalaVersion`.

ceedubs:
Thanks, @kailuowang! LGTM but it looks like scalafmt is complaining about formatting in build.sbt.

build.sbt
```diff
@@ -389,9 +389,9 @@ lazy val docs = project
 
 lazy val cats = project
   .in(file("."))
-  .settings(moduleName := "root")
-  .settings(catsSettings)
-  .settings(noPublishSettings)
+  .settings(moduleName := "root", crossScalaVersions := Nil)
+  .settings(publishSettings) // these settings are needed to release all aggregated modules under this root module
+  .settings(noPublishSettings) // this is to exclue the root module itself from being published.
```
ceedubs:
```suggestion
  .settings(noPublishSettings) // this is to exclue the root module itself from being published.
```

There is a dangling space at the end of this line. I've removed it in this suggestion.

ceedubs:
@kailuowang once you fix the build.sbt formatting issue, you may run into an issue where now SBT complains about the build's `sbt ++TRAVIS_SCALA_VERSION` commands for projects that don't have that version number listed in their `crossScalaVersion`s. If so, I think that #2737 fixes this issue. Though I'm not sure that adding the `!` to the command is the best solution. It's possible that we should be removing `2.13` from the `crossScalaVersion` and that we should explicitly add a `!` to the `2.13` tests. Not sure what the best approach is.

kailuowang:
Thanks @ceedubs. I will be traveling without laptop for a week, if you want to, feel free to merge this PR into #2737

kailuowang:
@LukaJCB any feedback?

PR #2734, opened 2019-02-16
upgrade to scalacheck 1.14

kailuowang:

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -130,27 +130,23 @@ lazy val includeGeneratedSrc: Setting[_] = {
   }
 }
 
-// 2.13.0-M4 workarounds
-def catalystsVersion(scalaVersion: String): String =
-  if (priorTo2_13(scalaVersion)) "0.6" else "0.8"
+val catalystsVersion = "0.8"
 
 def scalatestVersion(scalaVersion: String): String =
   if (priorTo2_13(scalaVersion)) "3.0.5" else "3.0.6-SNAP5"
 
-def scalaCheckVersion(scalaVersion: String): String =
-  if (priorTo2_13(scalaVersion)) "1.13.5" else "1.14.0"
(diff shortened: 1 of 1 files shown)


ceedubs:
It looks like there are some differences in the format of strings in some of the SBT doctests. TBH I'm not sure why this PR would have changed these though? Maybe it changed some Scalacheck `Pretty` instances?

```
'Some((List(1, 2, 3, 4),List()))' is not equal to 'Some((List(1, 2, 3, 4), List()))'
```

```
'Some(Right(n))' is not equal to 'Some(Right('n'))'
```

ceedubs:
Hmm actually upon closer inspection it looks like we _already_ have this problem, and somehow this change just makes the build detect the failure. I'll submit a PR for master.

kailuowang:
close reopen to fix travis glitch

travisbrown:
Shouldn't this happen in sync with [the upcoming ScalaTest 3.1.0 release](https://twitter.com/travisbrown/status/1101956555136733187), which is the first (legitimate) release that depends on ScalaCheck 1.14? There will also need to be a new Discipline release.

Even if all tests pass here just fine, I'm concerned that mixing ScalaTest 3.0 and 3.1 and ScalaCheck 1.13 and 1.14 is going to cause problems and impose a lot of unnecessary fuss and worry on users of downstream libraries like circe-testing.

kailuowang:
@travisbrown that's valid concerns. We can wait for scalatest 3.1.0

kailuowang:
@travisbrown we need to merge this to fix  #2758 to fix build. In the mean time I created an issue https://github.com/typelevel/cats/issues/2761 to make sure that we upgrade to coming Scalatest 3.10 before our Cats next release

SethTisue:
🎉

Issue #2741, opened 2019-03-02
More efficient Alternative#separate via Foldable


tkroman:
I was looking at the current implementation of `separate` and looks like at least for the list-like `F`s we could benefit from a single-pass implementation (well, not for lists since list's combine is `++`, but for some structures which have more efficient `combine` implementations :)).

Compare current implementation:
```scala
def separate[G[_, _], A, B](fgab: F[G[A, B]])(implicit FM: Monad[F], G: Bifoldable[G]): (F[A], F[B]) = {
  val as = FM.flatMap(fgab)(gab => G.bifoldMap(gab)(pure, _ => empty[A])(algebra[A]))
  val bs = FM.flatMap(fgab)(gab => G.bifoldMap(gab)(_ => empty[B], pure)(algebra[B]))
  (as, bs)
}

```

to the one proposed:

```scala
def separateFoldable[G[_, _], A, B](fgab: F[G[A, B]])(implicit
                                                      G: Bifoldable[G],
                                                      FF: Foldable[F]): (F[A], F[B]) = {
  FF.foldLeft(fgab, (empty[A], empty[B])) { case (mamb, gab) =>
    G.bifoldLeft(gab, mamb)(
      (t, a) => (combineK(t._1, pure(a)), t._2),
      (t, b) => (t._1, combineK(t._2, pure(b)))
    )
  }
}
```

In any case, for foldables this would at least reduce the number of `empty` allocations etc etc.

Note: I'm obviously not suggesting replacing the current one which is strictly more generic, but adding an alternative (pun not intended) implementation alongside existing one.

PR #2742, opened 2019-03-04
Added separateFoldable

tkroman:
#2741 

Since this is a new method, mima is breaking but I haven't used it before, so not sure what to do here.


kailuowang:
Thanks for the contribution. There is ways to get around breaking binary compatibility for this type of addition, but there are quite some hoops to jump through. The good news is in Q2 we are going to allow this type of changes on master with a new branching scheme. So if you can wait, it might be easier to just schedule this to cats 2.1. Otherwise I can guide you on how to avoid breaking BC on Scala 2.11. Let me know.

tkroman:
@kailuowang I can definitely wait, but I'm OK with jumping through the hoops as well, so I defer to your judgement.

kailuowang:
If you can wait, let's wait to avoid the extra boilerplate code on master.

travisbrown:
@tkroman Do you mind merging master to fix the build? Otherwise looks good to me.

core/src/main/scala/cats/Alternative.scala
```diff
@@ -41,6 +41,30 @@ import simulacrum.typeclass
...
+   *
+   * Example:
+   * {{{
+   * scala> import cats.implicits._
+   * scala> val l: List[Either[String, Int]] = List(Right(1), Left("error"))
+   * scala> Alternative[List].separate(l)
+   * res0: (List[String], List[Int]) = (List(error),List(1))
+   * }}}
+   */
+  def separateFoldable[G[_, _], A, B](fgab: F[G[A, B]])(implicit G: Bifoldable[G], FF: Foldable[F]): (F[A], F[B]) =
```
travisbrown: I'd prefer for the `Foldable` instance to come first in the list to match the order it appears in the argument, but I guess matching `separate`'s order for consistency is probably more important.

core/src/main/scala/cats/syntax/alternative.scala
```diff
@@ -51,6 +56,25 @@ final class SeparateOps[F[_], G[_, _], A, B](private val fgab: F[G[A, B]]) exten
...
+   *
+   * Example:
+   * {{{
+   * scala> import cats.implicits._
+   * scala> val l: List[Either[String, Int]] = List(Right(1), Left("error"))
+   * scala> l.separateFoldable
+   * res0: (List[String], List[Int]) = (List(error),List(1))
+   * }}}
+   */
+  def separateFoldable(implicit
```
travisbrown: This enrichment method could be added to `SeparateOps` since the target has the same type.

travisbrown:
@tkroman Thanks! I think it'll still need merging with master to avoid the 2.11 bincompat check.

travisbrown:
CI failure was Travis nonsense, just restarted that job.

tkroman:
woohoo, thanks!

Issue #2745, opened 2019-03-08
add Selective


johnynek:
https://www.staff.ncl.ac.uk/andrey.mokhov/selective-functors.pdf

```scala
trait Selective[F[_]]  {
  def applicative: Applicative[F] // we could also extend, but that may create ambiguities
  // the law here is if you give `F[Right[B]` we never evaluate `fn`
  def select[A, B](fab: F[Either[A, B]])(fn: F[A => B]): F[B]
}

object Selective {
  def fromMonad[F[_]: Monad]: Selective[F] =
    new Selective[F] {
      def applicative = Monad[F]
      def select[A, B](fa: F[Either[A, B]])(fn: F[A => B]): F[B] =
        fa.flatMap {
          case Right(b) => Monad[F].pure(b)
          case Left(a) => fn.map(_(a))
        }
    }
}
```

Note the paper gives some interesting examples of things that don't have Monad but do have Selective, e.g. Validated.

LukaJCB:
I think Selective is super interesting and I can't wait to dive into it more fully. 
I also think we could benefit from a more "Scala-like" encoding, as the one in paper is based on `ap` which is hardly used in Scala, due to no auto-currying. I envisioned something like 
```scala
trait Selective[F[_]] extends Apply[F] {
  def selective[A, B, C](feab: F[Either[A, B]])(fc: F[C]): F[Either[(A, C), B]]
  def select[A, B](feab: F[Either[A, B]])(fc: F[A => B]): F[B] =
    map(selective(feab)(fc))(_.fold({ case (a, f) => f(a) }, identity))
}
```
modulo the inheritance bit. 
I think this signature is equal in power, but doesn't involve any function application and is therefore a bit more verbose.
It'd also be really interesting if we could extract something like `Semigroupal` that is just the least powered version of this abstraction :)

cb372:
I'm planning to take a stab at this today based on @johnynek's formulation. Can't guarantee I will get anywhere with it though.

LukaJCB:
Not to discourage you @cb372, but I think it's a good idea to take our time and be patient with this. It's a very new and experimental abstraction, that could probably use some time discussing how to implement it, especially with how it fits in the cats hierarchy :) 
Cats is supposed to be super stable, so anything that goes has to be "future-proofed" to some degree. With that said, I really appreciate you taking the time for this and I'm very much looking forward to see some of your conclusions :) 

[...]

PR #2746, opened 2019-03-08
WIP Selective applicative functors

cb372:
This is a collaboration with @hamishdickson. Made a first stab at implementing [Selective Applicative Functors](https://www.staff.ncl.ac.uk/andrey.mokhov/selective-functors.pdf) as suggested in #2745.

* Add Selective type class
* Add instance for Validated
* Start writing laws
* Start writing combinators

Here is the "shapes" example from the paper using Validated:

```
scala> import cats._, cats.data._, cats.data.Validated._
import cats._
import cats.data._
import cats.data.Validated._

scala> type Radius = Int
defined type alias Radius

scala> type Width = Int
defined type alias Width

scala> type Height = Int
defined type alias Height

scala> :paste
// Entering paste mode (ctrl-D to finish)

sealed trait Shape
case class Circle(r: Radius) extends Shape
case class Rectangle(w: Width, h: Height) extends Shape

// Exiting paste mode, now interpreting.

defined trait Shape
defined class Circle
defined class Rectangle

scala> def shape[F[_]](x: F[Boolean], r: F[Radius], w: F[Width], h: F[Height])(implicit F: Selective[F]): F[Shape] =
  F.ifS(x)(F.map(r)(radius => Circle(radius): Shape))(F.applicative.map2(w, h){ case (width, height) => Rectangle(width, height): Shape })
shape: [F[_]](x: F[Boolean], r: F[Radius], w: F[Width], h: F[Height])(implicit F: cats.Selective[F])F[Shape]

scala> type V[A] = Validated[String, A]
defined type alias V

scala> import cats.instances.string._
import cats.instances.string._

scala> shape[V](valid(true), valid(1), invalid("width?"), invalid("height?"))
res1: V[Shape] = Valid(Circle(1))

scala> shape[V](valid(false), invalid("radius?"), valid(2), valid(3))
res2: V[Shape] = Valid(Rectangle(2,3))

scala> shape[V](valid(false), valid(1), invalid("width?"), invalid("height?"))
res3: V[Shape] = Invalid(width?height?)

scala> shape[V](invalid("choice?"), invalid("radius?"), valid(2), invalid("height?"))
res4: V[Shape] = Invalid(choice?)
```

This is a work in progress (see the TODO comments) but happy to receive feedback at this stage.

After getting this far, we're still not sure about whether it's better for `Selective` to extend `Applicative` or to have it as a member. We went for the latter, but it's a bit clumsy to work with.

### DIFF
--- a/core/src/main/scala/cats/Selective.scala
+++ b/core/src/main/scala/cats/Selective.scala
@@ -1,0 +1,44 @@
+package cats
+
+// TODO simulacrum bug? @typeclass doesn't work
+trait Selective[F[_]]  {
+
+  def applicative: Applicative[F]
+
+  def select[A, B](fab: F[Either[A, B]])(fn: F[A => B]): F[B]
+
+  def pure[A](a: A): F[A] = applicative.pure(a)
+
+  def map[A, B](fa: F[A])(f: A => B): F[B] = applicative.map(fa)(f)
+
+  def branch[A, B, C](x: F[Either[A, B]])(l: F[A => C])(r: F[B => C]): F[C] = {
+    val lhs = {
+      val innerLhs: F[Either[A, Either[B, C]]] = map(x)(_.map(Left(_)))
+      val innerRhs: F[A => Either[B, C]] = map(l)(_.andThen(Right(_)))
+      select(innerLhs)(innerRhs)
+    }
+    select(lhs)(r)
+  }
+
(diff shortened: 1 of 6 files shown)


dwijnand:
Close in favour of implementation ideas and experimentation in https://github.com/cb372/cats-selective and theoretical discussions in #2745?

cb372:
Sounds good to me. I really need to get back to work on that repo I started...

PR #2752, opened 2019-03-19
added a styling stage to check the styles first

kailuowang:
before committing to run the full build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ before_install:
  - export PATH=${PATH}:./vendor/bundle
 
 stages:
+  - name: styling
   - name: test
   - name: publish snapshot
     if: (branch = master AND type = push)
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -68,19 +70,13 @@ jobs:
     - <<: *bincompat_check
       scala: *scala_version_212
 
-    - &publish_snapshot
-      stage: publish snapshot
+    - stage: publish snapshot
       script: |
         if [[ $(cat version.sbt) =~ "-SNAPSHOT" ]]; then
-          sbt ++$TRAVIS_SCALA_VERSION! publish gitSnapshots publish
+          sbt +publish gitSnapshots +publish
```
kailuowang: It makes it possible for each module to define which scala version to release on (based on their own `crossScalaVersions`), also simplifies the build here.

kailuowang:
@ceedubs , do you happen to have time to take a look? You mentioned something about Travis stage not working in some way. This run seems fine to me. Maybe they fixed it?

kailuowang:
Yes. That occurred to me as well. Since all cats PRs/branch build share the same concurrent VMs, I think it's worth the little delay to free up VMs for other PRs.

PR #2754, opened 2019-03-23
Add `replicatePar` syntax

kubukoz:
This PR introduces a function that runs the given effect N times in parallel and collects the result in a list.


kubukoz:
Honestly, I'm not sure this is even useful anymore. I'll close it for now.

Issue #2718, opened 2019-02-01
Chain.fromSeq vs NonEmptyChain.fromSeq


dcastro:
There seems to be an inconsistency between these two methods.

`Chain.fromSeq` takes a `scala.collection.Seq`, whereas `NonEmptyChain.fromSeq` takes a `scala.collection.immutable.Seq`.

I assume the correct fix is to change `Chain.fromSeq` to take an immutable sequence, but that sounds like a breaking change...

Cats version: 1.6.0.

diesalbla:
This seems to have been solved, or at least worked around, in #3133 .

PR #2756, opened 2019-03-25
NonEmptyChain.fromSeq should take base Seq

gheine:
... (ie. `scala.collection.Seq`) instead of `scala.collection.immutable.Seq`, similar to `Chain.fromSeq`


gheine:
Fixes #2718

kailuowang:
Thanks! Unfortunately this break backward binary compatibility. 
A remedy is probably adding an overload , like the following (I also included a doc test to make sure everything works)
```scala

  /**
   * {{{
   * scala> cats.data.NonEmptyChain.fromSeq(Seq(1,2,3))
   * res1> Option[NonEmptyChain[Int]] = Some(List(1, 2, 3))
   * }}}
   */
  def fromSeq[A](as: Seq[A]): Option[NonEmptyChain[A]] =
    if (as.nonEmpty) Option(create(Chain.fromSeq(as))) else None

  /**
   * For Binary Compatibility
   * {{{
   * scala> cats.data.NonEmptyChain.fromSeq(collection.immutable.Seq(1,2,3))
   * res1> Option[NonEmptyChain[Int]] = Some(List(1, 2, 3))
   * }}}
   */
  def fromSeq[A](as: collection.immutable.Seq[A]): Option[NonEmptyChain[A]] =
    if (as.nonEmpty) Option(create(Chain.fromSeq(as))) else None

```

mpilquist:
Also note the behavior when compiling for 2.13 is different (unqualified `Seq` is `s.c.immutable.Seq` under 2.13).

travisbrown:
I'd be 👎 on this except that `Chain` uses `scala.collection.Seq`, and I think we should be consistent. If there's a way to make the override work on 2.13 without version-specific code I'd be 👍.

travisbrown:
I've just tried a few different ways to make this bincompat and I think the best would be to have this for 2.12 (and the empty equivalent for 2.13):

```
package cats.data

private[data] trait ScalaVersionSpecificNonEmptyChainImpl {
  def fromSeq[A](as: scala.collection.Seq[A]): Option[NonEmptyChain[A]] =
    if (as.nonEmpty) Option(NonEmptyChainImpl.create(Chain.fromSeq(as))) else None
}
```

To be honest I don't think it's worth it, though, especially because on 2.13 bare `Seq` _is_ `immutable.Seq`.

PR #2759, opened 2019-03-26
Category[Is] and related tests

### LINKED ISSUE #2720: Category instance for Is
Is there any reason why **cats.evidence.Is** isn't a **Category**?
It's already have **compose** and **id** is **Is.refl**

ctongfei:
This should fix #2720 .

### DIFF
--- a/core/src/main/scala/cats/evidence/Is.scala
+++ b/core/src/main/scala/cats/evidence/Is.scala
@@ -1,6 +1,7 @@
 package cats.evidence
 
 import cats.Id
+import cats.arrow._
 
 /**
  * A value of `A Is B` is proof that the types `A` and `B` are the same. More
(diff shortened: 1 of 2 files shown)


kailuowang:
Thanks! There is just Scalafmt failure. Would you run sbt prePR and commit the changes ?

ctongfei:
@kailuowang Thanks! I'll figure it out soon.

ctongfei:
@kailuowang This should be OK now.

core/src/main/scala/cats/evidence/Is.scala
```diff
@@ -65,7 +66,19 @@ abstract class Is[A, B] extends Serializable {
...
 }
 
-object Is {
+sealed abstract class IsInstances {
+  import Is._
+
+  /**
+   * The category instance on Leibniz categories.
+   */
+  implicit val leibniz: Category[Is] = new Category[Is] {
```
LukaJCB: Can we give thise one a more verbose name? Something like `catsEvidenceCategoryForLeibniz`? :)

ctongfei: The `Category` instance for `As` is named as `liskov` -- so I copied that :-)

LukaJCB: Good call :)

PR #2766, opened 2019-03-28
Removed Ops and replaced them with forwarders.

drdozer:


kailuowang:
is it possible to add some benchmark [here](https://github.com/typelevel/cats/tree/master/bench/src/main/scala/cats/bench) to verify the performance

smarter:
There's a remaining use of Ops in https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/syntax/hash.scala. Afterwards I think you should be able to delete https://github.com/typelevel/cats/blob/master/macros/src/main/scala/cats/macros/Ops.scala and remove machinist from build.sbt.

core/src/main/scala/cats/arrow/FunctionK.scala
```diff
@@ -89,74 +89,14 @@ object FunctionK {
    *
    * Additionally, the type parameters on `f` must not be specified.
    */
-  def lift[F[_], G[_]](f: (F[α] ⇒ G[α]) forSome { type α }): FunctionK[F, G] =
-    macro FunctionKMacros.lift[F, G]
+  def lift[F[_]]: LiftPartiallyApplied[F] = new LiftPartiallyApplied
 
-}
-
-private[arrow] object FunctionKMacros {
```
kailuowang: We can't simply remove this code. we need to create two versions of it, one for scala 2.x with the current implementation and another with the new scala3 implementation. Like in this example
https://github.com/typelevel/cats/tree/master/core/src/main/scala-2.13%2B/cats/compat
we created two versions of `cats.compat.SortedSet`

core/src/main/scala/cats/arrow/FunctionK.scala
```diff
@@ -3,8 +3,11 @@ package arrow
 
 import scala.reflect.macros.blackbox.Context
 
+<<<<<<< HEAD
+=======
 import cats.data.{EitherK, Tuple2K}
 
+>>>>>>> 60c2fe6336f1e4dff1cd7d7eb839c19f83778b63
```
LukaJCB: Merge error?

smarter:
I've made updated PRs for the machinist part: https://github.com/typelevel/cats/pull/2925 and the FunctionK part (using a different solution proposed on the issue): https://github.com/typelevel/cats/pull/2926, so this PR can be closed.

PR #2768, opened 2019-03-29
Added Coflatmap instance to Kleisli

### LINKED ISSUE #2767: Add CoflatMap instance for Kleisli
Can't seem to find any that already exist in the codebase and the implementation passes all of the laws.

wunderk1nd-e:
Can't seem to find any that already exist in the codebase and the implementation passes all of the laws.

Resolves #2767

### DIFF
--- a/core/src/main/scala/cats/data/Kleisli.scala
+++ b/core/src/main/scala/cats/data/Kleisli.scala
@@ -450,10 +450,10 @@ sealed abstract private[data] class KleisliInstances8 extends KleisliInstances9 
   implicit def catsDataDistributiveForKleisli[F[_], R](implicit F0: Distributive[F]): Distributive[Kleisli[F, R, ?]] =
     new KleisliDistributive[F, R] with KleisliFunctor[F, R] { implicit def F: Distributive[F] = F0 }
 
-  implicit def catsDataCoflatMapForKleisli[F[_], In](implicit F0: Applicative[F]): CoflatMap[Kleisli[F, In, ?]] =
+  implicit def catsDataCoflatMapForKleisli[F[_], In](implicit F: Functor[F]): CoflatMap[Kleisli[F, In, ?]] =
     new CoflatMap[Kleisli[F, In, ?]] {
       override def coflatMap[A, B](fa: Kleisli[F, In, A])(f: Kleisli[F, In, A] => B): Kleisli[F, In, B] =
-        Kleisli(_ => F0.pure(f(fa)))
+        Kleisli(i => F.as(fa.run(i), f(fa)))
 
       override def map[A, B](fa: Kleisli[F, In, A])(f: A => B): Kleisli[F, In, B] = fa.map(f)
     }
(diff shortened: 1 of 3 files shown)


core/src/main/scala/cats/data/Kleisli.scala
```diff
@@ -455,6 +454,14 @@ sealed abstract private[data] class KleisliInstances8 extends KleisliInstances9
 sealed abstract private[data] class KleisliInstances9 {
   implicit def catsDataFunctorForKleisli[F[_], A](implicit F0: Functor[F]): Functor[Kleisli[F, A, ?]] =
     new KleisliFunctor[F, A] { def F: Functor[F] = F0 }
+
+  implicit def catsDataCoflatMapForKleisli[F[_], In](implicit F0: Applicative[F]): CoflatMap[Kleisli[F, In, ?]] =
```
kailuowang: This one has to go one level above `Functor` since `Coflatmap` inherits `Functor` and thus is more "specific" type wise. Let's move it to ` KleisliInstances8` .

tests/src/test/scala/cats/tests/KleisliSuite.scala
```diff
@@ -195,6 +196,12 @@ class KleisliSuite extends CatsSuite {
     checkAll("MonoidK[Kleisli[Option, String, ?]]", SerializableTests.serializable(monoidk))
   }
 
+  {
+    implicit val coflatmap = Kleisli.catsDataCoflatMapForKleisli[Option, String]
```
kailuowang: you shouldn't need this explicit instance. did it fail to compile without it?

tests/src/test/scala/cats/tests/KleisliSuite.scala
```diff
@@ -26,6 +25,8 @@ class KleisliSuite extends CatsSuite {
   implicit val iso = SemigroupalTests.Isomorphisms.invariant[Kleisli[Option, Int, ?]]
   implicit val iso2 = SemigroupalTests.Isomorphisms.invariant[Reader[Int, ?]]
 
+  implicit def cogenKlei[F[_], A, T](implicit cogen: Cogen[A => F[T]]): Cogen[Kleisli[F, A, T]] = cogen.contramap(_.run)
```
kailuowang: a better place for this might be laws.discipline.arbitrary
https://github.com/typelevel/cats/blob/master/laws/src/main/scala/cats/laws/discipline/Arbitrary.scala

core/src/main/scala/cats/data/Kleisli.scala
wunderk1nd-e: Cheers - that's been done

tests/src/test/scala/cats/tests/KleisliSuite.scala
wunderk1nd-e: You're right - I've removed it.

tests/src/test/scala/cats/tests/KleisliSuite.scala
wunderk1nd-e: Done 👍

LukaJCB:
I'm not quite sure if this instance makes sense, using `Applicative#pure` for `coflatmap` works for every single data type out there and using `pure` without also using `product` seems to me just like using `Monoid#empty` without using `Monoid#combine`. E.g. we dropped a PR that did something similar in cats-effect: https://github.com/typelevel/cats-effect/pull/285

kailuowang:
@LukaJCB good catch. 
how about using a `Functor` instead?
```scala
  implicit def catsDataCoflatMapForKleisli[F[_], In](implicit F: Functor[F]): CoflatMap[Kleisli[F, In, ?]] =
    new CoflatMap[Kleisli[F, In, ?]] {
      override def coflatMap[A, B](fa: Kleisli[F, In, A])(f: Kleisli[F, In, A] => B): Kleisli[F, In, B] =
        Kleisli(i =>  F.as(fa.run(i), f(fa)))

      override def map[A, B](fa: Kleisli[F, In, A])(f: A => B): Kleisli[F, In, B] = fa.map(f)
    }
```

PR #2770, opened 2019-04-02
Add foldRightDefer to Foldable

### LINKED ISSUE #2678: Foldable could have foldRightDefer
currently, we have:
```scala
def foldRight[A, B](fa: F[A], b: Eval[B])(fn: (A, Eval[B]) => Eval[B]): Eval[B]
```
which predates the `Defer[F[_]]` typeclass. In fact, I think in general, we could have:

```scala
def foldRightDefer[G[_]: Defer, A, B](fa: F[A], gb: G[B])(fn: (A, G[B]) => G[B]): G[B]
```
e.g.
```scala
      def foldRightDefer[G[_]: Defer, A, B](fa: List[A], gb: G[B])(f: (A, G[B]) => G[B]): G[B] = {
        def loop(as: List[A]): G[B] =
          as match {
            case Nil    => gb
            case h :: t => f(h, Defer[G].defer(loop(t)))
          }
        Defer[G].defer(loop(fa))
      }
```

If we had this, all effects extend Defer, so we could do foldRight with `IO` without going through `Eval`.

denisrosca:
This PR adds a new `foldRightDefer` function to `Foldable[F]` with the following signature:
```
def foldRightDefer[G[_]: Defer, A, B](fa: F[A], gb: G[B])(fn: (A, G[B]) => G[B]): G[B]
```

Additionally, I added the following law to check the `foldRight <-> foldRightDefer` equivalence where `type G[A] = Eval[A]`.

I didn't spend too much time on making this binary compatible because #2678 is marked as targeting the 2.1 milestone which is a major release. 

Closes #2678

### DIFF
--- a/core/src/main/scala/cats/Foldable.scala
+++ b/core/src/main/scala/cats/Foldable.scala
@@ -95,6 +95,13 @@ import Foldable.sentinel
    */
   def foldRight[A, B](fa: F[A], lb: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B]
 
+  def foldRightDefer[G[_]: Defer, A, B](fa: F[A], gb: G[B])(fn: (A, G[B]) => G[B]): G[B] =
+    Defer[G].defer(
+      this.foldLeft(fa, (z: G[B]) => z) { (acc, elem) => z =>
+        Defer[G].defer(fn(elem, acc(z)))
+      }(gb)
+    )
+
   def reduceLeftToOption[A, B](fa: F[A])(f: A => B)(g: (B, A) => B): Option[B] =
     foldLeft(fa, Option.empty[B]) {
       case (Some(b), a) => Some(g(b, a))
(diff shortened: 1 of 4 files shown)


kailuowang:
Not sure why you closed this one, but I think we probably can't make it binary compatible on scala 2.11. 
We can wait until we cut the branch for Cats 2.1, which should be soon (mostly pending on Scala test 3.1.0 release). Once the branch is cut, we will no longer support scala 2.11 on the master branch, so we should be able to merge PR like this in which is BC on 2.12+. For detail about this branching plan, you can see our roadmap https://github.com/typelevel/cats/blob/master/ROADMAP_2019.md

denisrosca:
I accidentally pushed an incomplete PR. 

The implementation that I thought is bullet proof turned out to fail the `foldRightDefer <-> foldRight` law that I introduced :confused: 

I'll keep working on it and create a new PR once I have something actually ready for review.

PR #2771, opened 2019-04-02
Fix our crossScalaVersions since the last change to travis.yml

kailuowang:
The sbt-travisci plugin was no longer able to parse travis.yml since we move away from build matrix. This, combined with other changes made to the build, broke `sbt prePR` as well snapshot release


build.sbt
```diff
@@ -743,7 +765,7 @@ addCommandAlias("validateKernelJS", "kernelLawsJS/test")
 addCommandAlias("validateFreeJS", "freeJS/test") //separated due to memory constraint on travis
 addCommandAlias("validate", ";clean;validateJS;validateKernelJS;validateFreeJS;validateJVM")
 
-addCommandAlias("prePR", ";fmt;++2.11.12;validateBC")
+addCommandAlias("prePR", ";fmt;++2.11.12 mimaReportBinaryIssues")
```
kailuowang: unfortunately, sbt doesn't seem to work when trying to fix to a scala version when running a command alias.
I filed an issue https://github.com/sbt/sbt/issues/4599 
by this change we no longer run `binCompatTest/test`, but I think it's rare that it's accidentally broken before PR.

PR #2772, opened 2019-04-04
Add foldRightDefer to Foldable

### LINKED ISSUE #2678: Foldable could have foldRightDefer

denisrosca:
This PR adds `foldRightDefer` to `Foldable[F]` with the following signature:
```
def foldRightDefer[G[_]: Defer, A, B](fa: F[A], gb: G[B])(fn: (A, G[B]) => G[B]): G[B]
```

A new law to check the `foldRight <-> foldRightDefer` equivalence where `type G[A] = Eval[A]` was also added to the test suite .

To avoid binary incompatibilities, I added the function as a syntax enrichment for `Foldable[F]` and `F[A] given a Foldable[F]`.

Closes #2678


kailuowang:
Thanks very much! 
It's a bit confusing, though, to have a law testing a method added through extension. 
I see two options
1. add the method to the type class directly and wait for the branch post 2.0 release
2. comment out the law and test with a note that it's pending the method being moved into the type class post cats 2.0

I lean towards option 1 unless having this method is urgent to you.

denisrosca:
I don't really mind either way. I just thought you would prefer something that could be merged immediately.

I'll move the method to the typeclass.

kailuowang:
Thanks very much!

core/src/main/scala/cats/Foldable.scala
```diff
@@ -587,11 +594,14 @@ import Foldable.sentinel
 object Foldable {
   private val sentinel: Function1[Any, Any] = new scala.runtime.AbstractFunction1[Any, Any] { def apply(a: Any) = this }
 
-  def iterateRight[A, B](iterable: Iterable[A], lb: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B] = {
-    def loop(it: Iterator[A]): Eval[B] =
-      Eval.defer(if (it.hasNext) f(it.next, loop(it)) else lb)
+  def iterateRight[A, B](iterable: Iterable[A], lb: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B] =
+    iterateRightDefer(iterable, lb)(f)
```
johnynek: I would imagine inlining `Eval.defer` will give a slight performance improvement since the jit can at best do this inlining later.

denisrosca: @johnynek not sure I understand. Are you saying you would prefer the previous implementation explicitly calling `Eval.defer`?

travisbrown:
@denisrosca Do you mind merging master and changing the `iterateRight` implementation back to use `Eval.defer` directly as @johnynek suggested? Otherwise this looks good to me and it'd be nice to get it into 2.1.0. Thanks!

denisrosca:
@travisbrown Don't mind. Hopefully I'll get to it this week.

travisbrown:
@denisrosca Would you mind if I do it now and push to this PR branch? We're aiming to publish 2.1.0-RC1 on Friday and it'd be great to get this in.

denisrosca:
I don't mind if you prefer to do it now. Thanks.

johnynek:
Note, the addition of Iterable to Foldable in #2380 would allow us to use the iterateRightDefer method rather than the foldLeft implementation which I think will be more efficient (and I think allow short circuiting).

johnynek:
Thanks for adding this BTW!

PR #2782, opened 2019-04-10
add FunctionK.apply to summon instances from the implicit scope

bpholt:
I'm assuming `FunctionK` would be annotated with `@typeclass`, if simulacrum supported type constructors with more than one argument. This PR adds the `apply` method that I believe simulacrum would have added if it could.

### DIFF
--- a/core/src/main/scala/cats/arrow/FunctionK.scala
+++ b/core/src/main/scala/cats/arrow/FunctionK.scala
@@ -66,6 +66,12 @@ trait FunctionK[F[_], G[_]] extends Serializable { self =>
 object FunctionK {
 
   /**
+   * Summon an instance from the implicit scope if such an instance
+   * is available.
+   */
+  def apply[F[_], G[_]](implicit instance: F ~> G): F ~> G = instance
+
+  /**
    * The identity transformation of `F` to `F`
    */
   def id[F[_]]: FunctionK[F, F] = λ[FunctionK[F, F]](fa => fa)
--- a/tests/src/test/scala/cats/tests/FunctionKSuite.scala
+++ b/tests/src/test/scala/cats/tests/FunctionKSuite.scala
@@ -104,4 +104,12 @@ class FunctionKSuite extends CatsSuite {
     assertTypeError("FunctionK.lift(sample[Nothing])")
   }
 
+  test("apply ops") {
+    implicit val lToO: List ~> Option = listToOption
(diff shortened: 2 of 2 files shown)


LukaJCB:
Hey @bpholt, I think there might be some confusion. I really don't think `FunctionK` is a type class, as there's really no way at all to have unique instances for any two types, just like something like `Function1` isn't a type class. Of course you can use it in a way similar to a type class using implicits, but I don't think that's the intention. I think for these purposes it might be better to write your own type class separate from `FunctionK`. WDYT?

kailuowang:
@lukaJCB I agree that FunctionK isn't a type class. I was about to raise the same concern. But then I thought what is the downside of using it as implicit evidence? In which case, an implicit instance summon method would be handy.

rossabaker:
Summoners are sometimes used for implementations of tagless algebras.  I don't feel strongly either way on adding this, but would not mention simulacrum in the scaladoc, as it's not a type class.

bpholt:
I misunderstood whether this was a type class or not, so I have no problem removing the simulacrum reference from the scaladoc. That change is reflected in fa754ec.

(I was thinking there would typically only be one or two ways to meaningfully implement `FunctionK` for a given pair of types, so it felt similar to `Monoid`, where there can be multiple ways to combine things and one has to pick the desired method. But I'm fairly new to this and don't understand all the subtleties yet 🙂)

tpolecat:
I think this is probably not a great practice so I'm not super enthusiastic about endorsing it. It also doesn't save much typing over `implicitly[F ~> G]`.

diesalbla:
Apart of the two reasons mentioned above, there is yet another reason, about ergonomics: 
since `FunctionK` is like a function, some would like its API to be like that of other function-like types within `cats`-related libraries, such as `circe` decoders/encoders, or `http4s` routes. Within these, an `apply` method in the companion object is used not to summon an implicit instance, but to build an instance from a function.

bpholt:
I think this PR was a mistake. IIRC when I originally put it together I was thinking that one could do something like 

```scala
val f: F[T] = ???
FunctionK[F, G](f): G[T]
```

(and specifically also that replacing `FunctionF.apply` there with `implicitly[F ~> G]` would not work)

But AFAICT, that syntax does not work, or at least, I can't recreate whatever it was that I was trying to do. I'm going to close this PR. Thanks to everyone who provided feedback! 🙂

PR #2784, opened 2019-04-10
Make WrappedMutableMapBase extend Serializable

ceedubs:
We are running into Spark `Task not serializable` issues when a closure
that executes on a Spark executor node involves a `Map` that is created
via running `foldMap` on a `List`. This commit makes the
`WrappedMutableMap` hierarchy extend `Serializable` and chex that the
cerealization works (this test failed before extending `Serializable`).


rossabaker:
Are we all just going to pretend we didn't see those puns?

PR #2785, opened 2019-04-10
Fix unconventional package objects in kernel

### LINKED ISSUE #2781: unconventional package objects
we have several package objects that are not really package object since there is no corresponding packages and folders. Also their name is not conventional `package.scala` which throws warning in compiler. 
e.g. `kernel/instances/list`, `kernel/instances/tuple` etc. 
We can't make them normal objects now since that will break BC. But we should at least address the compiler warning and move those to `package.scala` and into corresponding folders. 
e.g. 
`kernel/instances/list.scala` -> `kernel/instances/list/package.scala`

denisrosca:
Closes #2781


kernel/src/main/scala/cats/kernel/instances/eq/package.scala
```diff
@@ -1,6 +1,6 @@
 package cats.kernel
 package instances
 
-trait EqInstances extends EqToEquivConversion
+package object eq extends EqInstances
 
-object eq extends EqInstances
+trait EqInstances extends EqToEquivConversion
```
kailuowang: not sure if this change is necessary

kernel/src/main/scala/cats/kernel/instances/hash/package.scala
```diff
@@ -1,6 +1,6 @@
 package cats.kernel
 package instances
 
-trait HashInstances extends HashToHashingConversion
+package object hash extends HashInstances
 
-object hash extends HashInstances
+trait HashInstances extends HashToHashingConversion
```
kailuowang: same here. the original is fine IMO

kailuowang:
thanks so much! I think we should leave those existing non-package object alone.

denisrosca:
I'll revert the changes breaking binary compatibility and keep the rest.

denisrosca:
@kailuowang any idea what happened with the 2.12.7 build? It looks like it never actually completed.

kailuowang:
Sometimes it hangs. I just restarted it, should be green soon

Issue #2779, opened 2019-04-09
Order instance for NonEmptySet


jatcwang:
Since SortedSet has an ordered instance and `NonEmptySet ` is implemented based on that, should we implement an Order instance for `NonEmptySet`?

EDIT: Will put in a PR

LukaJCB:
Closed by https://github.com/typelevel/cats/pull/2787

PR #2783, opened 2019-04-10
[WIP] move to Scala 2.13.0-RC1 (drop -M5)

SethTisue:
this is just a sketch to get the ball rolling, I don't intend
to finish it myself

not even sure if it's to the right branch, not sure what the plan is around ScalaCheck 1.14.0 being the only ScalaCheck available for 2.13 (for now? not sure if anyone has tried to push for 1.13.x release on 2.13)

PR #2787, opened 2019-04-11
Add Order instance for NonEmptySet (#2779)

jatcwang:
- Add Serializable tests for NonEmptySet instances

### DIFF
--- a/binCompatTest/src/test/scala/cats/data/CatsDataMimaException.scala
+++ b/binCompatTest/src/test/scala/cats/data/CatsDataMimaException.scala
@@ -1,0 +1,8 @@
+package cats.data
+import cats.implicits._
+
+object CatsDataPackagePrivateMimaExceptions {
+  def isBinaryCompatible = (
+    cats.data.NonEmptySetImpl.catsDataEqForNonEmptySet[Int]
+  )
+}
--- a/binCompatTest/src/test/scala/cats/data/CatsDataMimaException.scala
+++ b/binCompatTest/src/test/scala/cats/data/CatsDataMimaException.scala
@@ -1,8 +1,0 @@
-package cats.data
-import cats.implicits._
-
-object CatsDataPackagePrivateMimaExceptions {
-  def isBinaryCompatible = (
-    cats.data.NonEmptySetImpl.catsDataEqForNonEmptySet[Int]
-  )
-}
--- a/core/src/main/scala/cats/data/NonEmptySet.scala
+++ b/core/src/main/scala/cats/data/NonEmptySet.scala
@@ -407,12 +407,6 @@ sealed abstract private[data] class NonEmptySetInstances extends NonEmptySetInst
         fa.toNonEmptyList
     }
 
-  // Left to keep binary compatibility
-  private[data] def catsDataEqForNonEmptySet[A: Order]: Eq[NonEmptySet[A]] =
-    new Eq[NonEmptySet[A]] {
-      def eqv(x: NonEmptySet[A], y: NonEmptySet[A]): Boolean = x === y
-    }
-
   implicit def catsDataShowForNonEmptySet[A](implicit A: Show[A]): Show[NonEmptySet[A]] =
     Show.show[NonEmptySet[A]](_.show)
 
@@ -422,7 +416,7 @@ sealed abstract private[data] class NonEmptySetInstances extends NonEmptySetInst
 }
 
 sealed abstract private[data] class NonEmptySetInstances0 extends NonEmptySetInstances1 {
-  implicit def catsDataOrderForNonEmptySetReal[A](implicit A: Order[A]): Order[NonEmptySet[A]] =
+  implicit def catsDataOrderForNonEmptySet[A](implicit A: Order[A]): Order[NonEmptySet[A]] =
     new NonEmptySetOrder[A] {
       implicit override def A0: Order[A] = A
     }
--- a/core/src/main/scala/cats/data/NonEmptySet.scala
+++ b/core/src/main/scala/cats/data/NonEmptySet.scala
@@ -20,7 +20,6 @@ package data
 import cats.instances.sortedSet._
 import cats.kernel._
 import cats.syntax.order._
-import cats.instances.sortedSet._
 
 import scala.collection.immutable._
 
@@ -407,6 +406,11 @@ sealed abstract private[data] class NonEmptySetInstances extends NonEmptySetInst
         fa.toNonEmptyList
     }
 
+  implicit def catsDataOrderForNonEmptySet[A](implicit A: Order[A]): Order[NonEmptySet[A]] =
+    new NonEmptySetOrder[A] {
+      implicit override def A0: Order[A] = A
+    }
+
   implicit def catsDataShowForNonEmptySet[A](implicit A: Show[A]): Show[NonEmptySet[A]] =
     Show.show[NonEmptySet[A]](_.show)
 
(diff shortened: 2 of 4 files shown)


30sergey:
Дубликат #

30sergey:
#2783

core/src/main/scala/cats/data/NonEmptySet.scala
```diff
@@ -417,3 +420,29 @@ sealed abstract private[data] class NonEmptySetInstances {
     def combine(x: NonEmptySet[A], y: NonEmptySet[A]): NonEmptySet[A] = x | y
   }
 }
+
+sealed abstract private[data] class NonEmptySetInstances0 extends NonEmptySetInstances1 {
+  implicit def catsDataOrderForNonEmptySetReal[A](implicit A: Order[A]): Order[NonEmptySet[A]] =
```
kailuowang: nit, is that `Real` suffix still needed?

core/src/main/scala/cats/data/NonEmptySet.scala
```diff
@@ -19,6 +19,8 @@ package data
 
 import cats.instances.sortedSet._
 import cats.kernel._
+import cats.syntax.order._
+import cats.instances.sortedSet._
```
kailuowang: duplicated from 2 lines above

core/src/main/scala/cats/data/NonEmptySet.scala
jatcwang: Not sure. catsDataOrderForNonEmptySet is the name of the existing one (which I removed the `implicit` keyword from). I can play around with it and see if it breaks bin compat

kailuowang: did you mean `private[data] def catsDataEqForNonEmptySet[A: Order]:` ?

jatcwang: yep since `NonEmptySetInstances` extends `NonEmptySetInstances0`

kailuowang: :) `catsDataOrderForNonEmptySet` is different from `catsDataEqForNonEmptySet`.

jatcwang: 🤦‍♂️ well if MiMa is happy then I'm happy :)
I'm gonna try remove the old `catsDataEqForNonEmptySet` and see what happens

kailuowang: sorry I wasn't clear. I meant that you didn't need to add a `Real` suffix, because it didn't conflict with anything else. It's the right approach to remove the `implict` keyword for `catsDataEqForNonEmptySet` , but you didn't need to add private[data] since the class is already `private[data]`

jatcwang: yes I got you. What I was also going to do is to remove the original `catsDataEqForNonEmptySet`, since there's already a method named the same in `NonEmptySetInstances1`. (I think JVM's use of invokevirtual call means it won't break bin compat - so it'll be both source and binary compatible)

kailuowang:
I would like to propose :
We don't need two new classes, just one `NonEmptySetInstances0` is enough, you can place the `eq` instance there and the `order` instance in `NonEmptySetInstances`
I think this change is BC but Mima will raise a false alarm. We need to add a mima exception
here https://github.com/typelevel/cats/blob/master/build.sbt#L288
```
ProblemFilters.exclude[IncompatibleMethTypeProblem]("cats.data.NonEmptySetInstances.catsDataEqForNonEmptySet")
```

To make sure we didn't actually break BC, we need to add a line calling the instance def here https://github.com/typelevel/cats/blob/master/binCompatTest/src/main/scala/catsBC/MimaExceptions.scala similar to the other tests there.

core/src/main/scala/cats/data/NonEmptySet.scala
jatcwang: done

core/src/main/scala/cats/data/NonEmptySet.scala
jatcwang: Done

jatcwang:
Done (Sorry I was AFK for a week). Simplified the instances hierachy as suggested and added a different test class for testing binary package private instances.

jatcwang:
CI is passing again @kailuowang :smile:

tests/src/test/scala/cats/tests/NonEmptySetSuite.scala
```diff
@@ -20,17 +20,42 @@ package tests
...
+             SerializableTests.serializable(Order[NonEmptySet[ListWrapper[Int]]]))
+
+    // NOTE: Order[NonEmptySet[A]] instance depends on Order[SortedSet[A]]. SortedSet only has Order instance which depends on Order[A] instance
+    //       and thus the only way we can get PartialOrder[NonEmptySet[A]] is via construction of Order[NonEmptySet[A]]
+    checkAll("NonEmptySet[ListWrapper[Int]]", PartialOrderTests[NonEmptySet[ListWrapper[Int]]].partialOrder)
+    checkAll("PartialOrder[NonEmptySet[ListWrapper[Int]]]",
+             SerializableTests.serializable(PartialOrder[NonEmptySet[ListWrapper[Int]]]))
+
+    Eq[NonEmptySet[ListWrapper[Int]]]
+    PartialOrder[NonEmptySet[ListWrapper[Int]]]
```
kailuowang: These two lines are unnecessary, no?

tests/src/test/scala/cats/tests/NonEmptySetSuite.scala
```diff
@@ -20,17 +20,42 @@ package tests
...
+    checkAll("Eq[NonEmptySet[ListWrapper[Int]]]", SerializableTests.serializable(Eq[NonEmptySet[ListWrapper[Int]]]))
+
+    checkAll("NonEmptySet[ListWrapper[Int]]", OrderTests[NonEmptySet[ListWrapper[Int]]].order)
+    checkAll("Order[NonEmptySet[ListWrapper[Int]]]",
+             SerializableTests.serializable(Order[NonEmptySet[ListWrapper[Int]]]))
+
+    // NOTE: Order[NonEmptySet[A]] instance depends on Order[SortedSet[A]]. SortedSet only has Order instance which depends on Order[A] instance
+    //       and thus the only way we can get PartialOrder[NonEmptySet[A]] is via construction of Order[NonEmptySet[A]]
+    checkAll("NonEmptySet[ListWrapper[Int]]", PartialOrderTests[NonEmptySet[ListWrapper[Int]]].partialOrder)
+    checkAll("PartialOrder[NonEmptySet[ListWrapper[Int]]]",
```
kailuowang: Maybe I am missing something here, but this is covered by the `Order` test above, since we don't have a different way to construct a `PartialOrder`, not sure if we need this test?

binCompatTest/src/test/scala/cats/data/CatsDataMimaException.scala
```diff
@@ -0,0 +1,8 @@
+package cats.data
+import cats.implicits._
+
+object CatsDataPackagePrivateMimaExceptions {
```
kailuowang: My bad. I should've checked that mima no longer falsely claims that this type of change breaks BC. So this is no longer needed? Would you mine removing it?

tests/src/test/scala/cats/tests/NonEmptySetSuite.scala
jatcwang: These instance tests are written based on https://github.com/typelevel/cats/blob/4f92efedd2c30534c07b5f75a72b22e3e1ba9b13/tests/src/test/scala/cats/tests/NonEmptyListSuite.scala and from what I can gather is that it's to test that an `Eq` and `PartialOrder` instance can be resolved when an `Order` instance is in scope. 

Do you think they are redundant?

tests/src/test/scala/cats/tests/NonEmptySetSuite.scala
jatcwang: you're right. Removed thx

binCompatTest/src/test/scala/cats/data/CatsDataMimaException.scala
jatcwang: done

tests/src/test/scala/cats/tests/NonEmptySetSuite.scala
kailuowang: The resolution is already tested in the law and serialization tests

jatcwang: ok

PR #2789, opened 2019-04-13
Remove unused type parameter for Parallel instances

Billzabob:


tpolecat:
Looks like changing `def` to `val` and/or removing the unused type parameter breaks binary compatibility.

Billzabob:
@tpolecat How does cats handle binary compatibility breaking changes?

kailuowang:
@Billzabob we made a promise to no breaking binary compatibility on core until Cats 3.0. 
Did it cause any issues for users? 
We could fix it in a BC way - we remove the implicit keyword of the original `def`, make it package private and deprecated, add your `implicit val` with a `BinCompat` suffix.

Billzabob:
@kailuowang No I don't think it causes any issues so it's probably not worth it to do all that. It was just something I noticed and mentioned on Gitter and someone suggested I make PR to fix it. Does that mean this PR just stays up until 3.0?

kailuowang:
Okay, I will schedule it for 3.0

johnynek:
isn't this superceded by @travisbrown 's great observation here: #3012

travisbrown:
@johnynek I hadn't seen this PR, but the problem isn't removing the type parameter, it's the change from `def` to `val` and the fact that this is a trait—you end up with the kind of silly-seeming synthetic public setter.

We could go ahead and make this change but without the `val`s.

johnynek:
actually, I didn't even look at the diff (like a fool). Sorry, I just read the title and assumed.

Billzabob:
@travisbrown Can you help me understand what you mean by "synthetic public setter"? I thought zero parameter `def`s and `val`s were basically equivalent if you're not side-effecting and aren't using type parameters.

travisbrown:
@Billzabob They're equivalent in the [universal access principle](https://en.wikipedia.org/wiki/Uniform_access_principle) sense—they look the same to callers—but they're implemented differently on the underlying platform, especially in traits, where the initialisation of vals is tricky. The compiler is mapping traits onto JVM interfaces, which don't have instance fields and up until Java 8 didn't even have instance method implementations, and the details of that mapping mean making a `def` a `val` breaks binary compatibility (definitely on 2.11 but IIRC also on 2.12 and 2.13).

travisbrown:
@Billzabob This is a good catch and I'd personally really like to see this change in 2.0.0. Would you mind opening a new PR that only removes the type parameters? Unfortunately it'll have to happen in two places, since these instances are now part of version-specific code, but if you search for the names you'll find them.

travisbrown:
@Billzabob Hope you don't mind but I took your commit, updated it, and opened #3042, with the hope that we can get this into today's 2.0.0.

Billzabob:
@travisbrown I don't mind at all! Also, thanks for the explanation.

travisbrown:
Included in #3042, which is now merged.

Issue #2776, opened 2019-04-08
publish for Scala 2.13.0-RC1, please


SethTisue:

SethTisue:
blocked by simulacrum at the moment, but there's probably other things too (catalysts?)

SethTisue:
note that kind-projector moved from org.spire-math and org.typelevel in their version 0.10.0 release

SethTisue:
simulacrum has now published. not sure what the next blocker is

SethTisue:
I submitted some WIP at https://github.com/typelevel/cats/pull/2783 , not even sure if it's to the right branch, not sure what the plan is around ScalaCheck 1.14.0 being the only ScalaCheck available for 2.13 (for now? not sure if anyone has tried to push for 1.13.x release on 2.13)

SethTisue:
catalysts, I think

SethTisue:
#2791 and #2792 are both merged. is there more work to do before publishing levers can be pulled...?

kailuowang:
@SethTisue I am waiting on a discipline maintainer releasing a milestone with this PR merged in https://github.com/typelevel/discipline/pull/89

SethTisue:
@kailuowang I see, would that be @larsrh ?

larsrh:
0.11.2-M1 is on its way to Maven Central

kailuowang:
Last PR (i hope)  #2806 before 2.0-M1 release on 2.13-RC1

djspiewak:
@kailuowang Are you going to do a cross-build in the 1.x line on 2.13?

kailuowang:
@djspiewak we don't plan to release any non-bug-fix releases in the 1.x line. Cats 2.0 is binary compatible with 1.x on `cats-core` and `cats-free`.  
The only major changes Cats 2.0 introduces is upgrading to Scalatest 3.1.0 and Scalacheck 1.14.0. (There are a couple of minor breaking changes in cats-laws and alleycats, but they are unlikely to impact use site)
[...]

PR #2791, opened 2019-04-13
Remove catalysts dependency

travisbrown:
The catalysts dependency has proven to be a drag on every update (right now it seems to be [blocking a release for 2.13.0-RC1](https://github.com/typelevel/cats/issues/2776)) and at this point it's almost completely unused, so I'm proposing that we just scrap it.

There's some dead code in `LawTests` that uses catalysts's `TypeTagM`, but it is unused and hasn't been touched since 2016, so I just scrapped it. Otherwise the only thing catalysts is used for is to check whether the current build is Scala.js in order to set appropriate test sizes.

In my view this really should be done with separate code trees, but to keep this PR simple I just introduced a new (package private) `cats.platform.Platform` that duplicates the ~dozen lines of code Cats needs from catalysts.

MiMa says this change is fine, all tests pass as usual with no changes (apart from the `Platform` import), etc.

r? @SethTisue @kailuowang


kailuowang:
Thanks so much for taking this on. I believe this is the right thing to do.

SethTisue:
one less thing to worry about :+1:

travisbrown:
I ran `validateJVM` and `prePR` and thought that would be enough to confirm that the build works, but apparently it's not. Fixing the alleycats issue now (and trying not to get extremely frustrated that `alleycats` is used as a top-level package and shares no path with `cats`…).

kailuowang:
I am on my phone, the code changes look good to me, but GitHub didn't really show the full path of the rename, so it appears that there is no actual path change. I am curious what they are.

kailuowang:
replaced by #2792

kailuowang:
@travisbrown I spoke too soon in #2792, there is a conflict and can't be resolved throught github ui. Would it be too much to ask you to resolve it?

travisbrown:
@kailuowang Merged, thanks.

Issue #2790, opened 2019-04-13
Make instance packages more consistent


travisbrown:
There's currently a lot of inconsistency in the way `cats.instance` and `cats.kernel.instances` are implemented, which can be extremely frustrating, especially for [people trying to work with Cats from Java](https://stackoverflow.com/a/55023418/334519)), or for contributors who want to know where to add something (e.g. me, in the past), or for anyone thinking of pointing to Cats as an example of a thoughtfully-structured Scala library (e.g. me, now).

For example, all `cats.instances.xyz` are defined as objects inside a `cats.instances` package object (which seems completely unnecessary and messy for Java users), except for `cats.instances.symbol`, which for some reason is defined as a package object in its own file. The cats-kernel instance packages follow the same approach, with each `cats.kernel.instances.xyz` being a package object.

Whether or not `BinCompat` and prioritizations traits are package private are also inconsistent across the `instances` packages.

In my view Cats should follow the cats-kernel approach consistently across both kernel and core, and all compat and prioritization traits should be made as private as possible without breaking the build. I'm happy to do this myself if people agree it should happen, but since these changes will break binary compatibility I guess they have to wait for 3.0? It's a shame to have to tell users who need Java compatibility that they have to roll a bunch of their own awful boilerplate because some bad decisions that made it into 1.0 will necessarily live 2+ years or whatever.

kailuowang:
:+1: We recently just merged a PR to fix package objects in kernel
https://github.com/typelevel/cats/pull/2785
I didn't know objects within package objects are pain for Java users. Under that light, I agree we should move them all to package objects.
We might be about to make the change without breaking BC in the following way:
Make current objects package private and deprecated. Create new replacements package objects. This passed the mima and compilation but I have tested it from user site yet.  Better to test it in binCompatTest module. I can investigate more when I get back to computer later today. 
 I also agree that all traits that aren't intended to be inherited externally should be made package private. That change should be BC. We should also add it to guidelines..

dwijnand:
The best way I've seen to ensure good ergonomics in Java is by having Java tests (see akka/akka for examples of some duo-testing, maybe even some tidbits).

travisbrown:
@dwijnand I'm not advocating putting a lot of effort into supporting usage from Java—this is more just avoiding gratuitously making life for Java users more difficult than it already is by doing stuff in Scala that isn't idiomatic anyway.

dwijnand:
[...]

PR #2792, opened 2019-04-14
2.13.0-RC1

travisbrown:
Includes #2783 and #2791. Running the full build locally is apparently a mystery beyond my skill level so I'm just creating my own version of #2783 to run it here.


travisbrown:
These two changes in 2.13.0-RC1 make a huge mess of some of the `Hash` instances in cats-kernel:

* https://github.com/scala/scala/pull/7693
* https://github.com/scala/scala/pull/7212

I really wish `cats.kernel.instances.StaticMethods` hadn't been made public—there's absolutely no good reason for it to be public, and the API needs to be different in a couple of places for 2.13. I've done that in a binary compatible way by just adding a few new methods (which I've made package private), but it's more of a mess than it needs to be.

In any case it seems to work now, and I've tried to minimize the version-specific code.

travisbrown:
Ugh, there are some stray `Hash` instances for `SortedSet` and `SortedMap` in cats-core. Is there a reason these aren't in cats-kernel? The `SortedSet` instances even have `catsKernel` in the name. I've had to widen the access of `product2HashWithPrefix` and `updateUnorderedHashC` from `private[kernel]` to `private[cats]` because now we need them in core as well.

These aren't the only instances that seem out of place. At a glance:

```
travis@sidmouth cats(update/2.13.0-RC1)$ ag Kernel core/
core/src/main/scala/cats/data/Op.scala
23:  implicit def catsKernelEqForOp[Arr[_, _], A, B](implicit ArrEq: Eq[Arr[B, A]]): Eq[Op[Arr, A, B]] =

core/src/main/scala/cats/data/ZipList.scala
4:import cats.instances.list.catsKernelStdEqForList

core/src/main/scala/cats/instances/partialOrdering.scala
29:      def unit: PartialOrdering[Unit] = cats.instances.unit.catsKernelStdOrderForUnit.toOrdering

core/src/main/scala/cats/instances/sortedSet.scala
69:  implicit def catsKernelStdOrderForSortedSet[A: Order]: Order[SortedSet[A]] =
74:  implicit def catsKernelStdHashForSortedSet[A: Order: Hash]: Hash[SortedSet[A]] =
77:  implicit def catsKernelStdSemilatticeForSortedSet[A: Order]: BoundedSemilattice[SortedSet[A]] =
```
We're stuck with these in the wrong place until 3.0?

travisbrown:
This is probably off-topic for this PR, but I have to say my experience contributing to Cats this weekend (for the first time in a year or so, at least with non-trivial changes) has been pretty miserable.

There are so many little broken things—these misnamed / misplaced instances, inconsistencies everywhere, #2790, stuff like `StaticMethods` that definitely should not be in the public API, the fact that the build is so complex (right at 1k lines for `build.sbt` and the Travis config) that I can't figure out how to run it locally in a useful way, the whole ScalaCheck / ScalaTest version mess, etc.—and I don't feel like I can do anything about any of it because any non-terrible-hack fixes break binary compatibility.

The worst part is that this is entirely self-inflicted: Cats 1.0 is almost a year and a half old—it would be perfectly reasonable to release a 99% source-compatible 2.0 that would break binary compatibility but also make the build hugely simpler and the code more consistent.

Okay, that's my rant—as I've been saying over and over for months I think [the 2019 roadmap](https://github.com/typelevel/cats/blob/master/ROADMAP_2019.md) is a big mistake. Hopefully the build will pass this time and you can publish for 2.13.0-RC1 and I can go back to being just a reasonably happy library user instead of an extremely frustrated contributor.

travisbrown:
Oh, cool, now I'm getting export-hook related errors locally in alleycats-tests, but only on JDK 8—the tests run just fine on Java 10. Looking now…

travisbrown:
Okay, the only failure is now the 2.13.0-RC1 JVM tests, and that's just an OOM. It might need attention but I just restarted to see if it was a one-off thing.

travisbrown:
Now seeing the alleycats export-hook errors on 2.13.0-RC1 in CI:

```
[info] alleycats.tests.SetSuite *** ABORTED ***
[info]   java.lang.IncompatibleClassChangeError: Expected non-static field alleycats.std.ListInstances$exports$.alleycats$std$ListInstances$exports$$alleycats$std$ListInstances$anonimplicit$1
[info]   at alleycats.std.all$.<clinit>(all.scala:6)
[info]   at alleycats.tests.SetSuite.<init>(SetSuite.scala:12)
[info]   at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
[info]   at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
[info]   at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
[info]   at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
[info]   at java.lang.Class.newInstance(Class.java:442)
[info]   at org.scalatest.tools.Framework$ScalaTestTask.execute(Framework.scala:438)
[info]   at sbt.ForkMain$Run.lambda$runTest$1(ForkMain.java:304)
[info]   at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[info]   ...
```

kailuowang:
I plan to remove export hook. Will take on it tomorrow now that it is a blocker.

milessabin:
+1 to removing the export-hook dependency, but even so ... why an `IncompatibleClassChangeError`? export-hook has been rebuilt for 2.13.0-RC1 so we really shouldn't be seeing that sort of thing.

travisbrown:
@milessabin Yeah, it doesn't make sense to me. I've tried a full clean, etc., but it still happens maybe a quarter of the time both locally for me and in Travis CI.

kailuowang:
@travisbrown, export hook is removed on master, if you merge master back in. 
Note that currently, master branch is targeted to release 2.0, but we are waiting on Scala test 3.1.0 for that.

travisbrown:
@kailuowang Great, thanks! Trying it out locally now.

core/src/main/scala/cats/data/WriterT.scala
```diff
@@ -357,6 +357,9 @@ sealed private[data] trait WriterTFlatMap1[F[_], L] extends WriterTApply[F, L] w
   implicit override def F0: FlatMap[F]
   implicit def L0: Monoid[L]
 
+  override def ap[A, B](f: WriterT[F, L, A => B])(fa: WriterT[F, L, A]): WriterT[F, L, B] =
```
kailuowang: Just curious, was this an optimization opportunity you happen to come across or was it broken on RC1?
Also would it be simpler to just call `fa.ap(f)`

kailuowang:
Thanks so much! Shall we remove the `[WIP]`? The changes look good to me. Just left a minor question.

core/src/main/scala/cats/data/WriterT.scala
travisbrown: This is from @SethTisue's commit, but the issue is that 2.13 sees the `ap` definitions in `FlatMap` and `WriterTApply` as conflicting. I personally think picking out the appropriate implementation via `super` is the right thing to do in this case, but I also don't think it matters much, and am happy to change it to `fa.ap(f)` if you'd prefer.

travisbrown:
@kailuowang I'd been thinking we'd merge #2791 separately and then I'd merge master back here, since the catalysts removal is logically separate—it just makes this update easier. We could also just merge this directly, though, so I've removed the `WIP`. Thanks for reviewing!

kailuowang:
@travisbrown sorry I misunderstood your intention in regards to #2791. I agree that it's more logical to merge that as a separate PR. I am going to squash merge that one, and one you merge master back in this PR, please don't worry about sort of duplicated commits (the ones originally in #2791), we are going to squash merge this one too.

tpolecat:
```
[error] /home/travis/build/typelevel/cats/build.sbt isn't formatted properly!
```
Well that's helpful.

travisbrown:
@tpolecat Somehow in the merge build.sbt picked up a stray newline. But yeah, it's funny how irrationally annoying scalafmt's contentless error messages are, given that they're specifically not intended to be acted on (beyond indicating that you need to run scalafmt).

PR #2794, opened 2019-04-16
removed export hook

kailuowang:
export hook seems to be blocking updating to 2.13.0-RC1 #2792


LukaJCB:
Thanks so much for taking care of this @kailuowang! Any objections to merge this ASAP? :)

kailuowang:
Just noticed some glitch with kittens. Please give me some time to investigate. There is probably nothing we can do in this PR to address those. Let me verify.

PR #2797, opened 2019-04-17
Tuning JVM options for throughput

DavidGregory084:
Changes the JVM options used by SBT somewhat.

Here is the reasoning for each change:

* `-Xmx4G` - the Travis infrastructure that we're using [has 7.5GB of memory](https://docs.travis-ci.com/user/reference/overview/), this could probably be increased further if forking tests was also disabled
* removal of `XX:-UseGCOverheadLimit` - I think that this is just creating zombie builds that are stuck thrashing the GC and will timeout anyway due to lack of progress, better to fail early IMO
* removal of `-XX:+CMSClassUnloadingEnabled` - this is relevant only for the CMS collector
* switch from `-XX:+UseConcMarkSweepGC` to `-XX:+UseParallelGC` - from what I understand the concurrent CMS collector is a response time collector, whereas the parallel GC is a throughput collector (more on this in the [Oracle docs](https://docs.oracle.com/javase/10/gctuning/toc.htm)), which is much more relevant for compiling code IMO

Here is some VisualVM output from running `sbt clean test/clean catsJS/test`, which normally stalls for me locally with the default options.

Before:

Screenshot 2019-04-17 at 12 43 29

Note that I had to kill this build as it got stuck.

After:

Screenshot 2019-04-17 at 12 43 37

Hopefully this will translate to improvements on Travis too! Who knows!

### DIFF
--- a/.jvmopts
+++ b/.jvmopts
@@ -1,8 +1,6 @@
 -Dfile.encoding=UTF8
 -Xms1G
--Xmx3G
+-Xmx4G
 -XX:ReservedCodeCacheSize=250M
 -XX:+TieredCompilation
--XX:-UseGCOverheadLimit
--XX:+CMSClassUnloadingEnabled
--XX:+UseConcMarkSweepGC
+-XX:+UseParallelGC


DavidGregory084:
One more comment about the choice of GC; the Scala compiler team noted [performance regressions](https://github.com/sbt/sbt/issues/3559) when moving to Java 9 due to the default GC changing from the ParallelGC (a throughput collector) to the GC1C (a response-time collector).

DavidGregory084:
Build time from ~2hrs down to 30min, pretty happy with that 😄 

Going to close and reopen to give it another go

DavidGregory084:
Second build took ~31min

On that basis I'm not sure this is going to yield any improvement in build times, but it may be less likely that builds will freeze

kailuowang:
Thanks so much! @DavidGregory084, I am not an expert on concurrent CMS collector v.s. parallel GC, but I vote that we give it a try.

PR #2799, opened 2019-04-18
Add attemptTap to monad error

frosforever:
Thanks in advance for your time reviewing this and maintaining cats!

This adds `attemptTap` to `MonadError`. This is similar to `flatTap` but gives you access to the error that `attempt` would give you.

### DIFF
--- a/core/src/main/scala/cats/syntax/monadError.scala
+++ b/core/src/main/scala/cats/syntax/monadError.scala
@@ -11,6 +11,13 @@ trait MonadErrorSyntax {
     new MonadErrorRethrowOps(fea)
 }
 
+trait MonadErrorSyntaxBinCompat0 {
+  implicit final def catsSyntaxMonadErrorBinCompat0[F[_], E, A](
+    fa: F[A]
+  )(implicit F: MonadError[F, E]): MonadErrorOps0[F, E, A] =
+    new MonadErrorOps0[F, E, A](fa)
+}
+
 final class MonadErrorOps[F[_], E, A](private val fa: F[A]) extends AnyVal {
   def ensure(error: => E)(predicate: A => Boolean)(implicit F: MonadError[F, E]): F[A] =
     F.ensure(fa)(error)(predicate)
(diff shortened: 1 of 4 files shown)


core/src/main/scala/cats/syntax/monadError.scala
```diff
@@ -31,6 +38,11 @@ final class MonadErrorOps[F[_], E, A](private val fa: F[A]) extends AnyVal {
     F.adaptError(fa)(pf)
 }
 
+final class MonadErrorOps0[F[_], E, A](private val fa: F[A]) extends AnyVal {
+  def attemptTap[B](f: Either[E, A] => F[B])(implicit F: MonadError[F, E]): F[A] =
+    F.flatMap(F.attempt(fa))(ea => F.flatMap(f(ea))(_ => fa))
+}
```
LukaJCB: This will run `fa` twice, I'm not sure that that's what I'd expect 🤔

frosforever: Doh! Good catch. thanks and fixed

core/src/main/scala/cats/syntax/monadError.scala
```diff
@@ -31,6 +38,11 @@ final class MonadErrorOps[F[_], E, A](private val fa: F[A]) extends AnyVal {
     F.adaptError(fa)(pf)
 }
 
+final class MonadErrorOps0[F[_], E, A](private val fa: F[A]) extends AnyVal {
```
kailuowang: I don't think any of the bin compat traits are needed. we can just add this method to the existing `MonadErrorOps`.
Also let's have a doctest. 
But more fundamentally would you help me understand the incentive for this? Because I think there might be something else missing. 
What if we have the following syntax on `MonadError`
`def foldF[B]((e: E) => F[B])((a: A) => F[B]) : F[B]`
and then you can do 
`me.foldF(onError)(onSuccess) *> me`
and if that's too much typing 
we can add a `foldTapE`
Would that work better for you than involving Either?

larsrh:

PR #2801, opened 2019-04-21
Introduce Clay: light-weight Kleisli-like trait

diesalbla:
This PR introduces the `Clay` class, a light-weight implementation of a `Kleisli`-like abstraction. 

The Kleisli class is one of the well-known data types of Cats. An Kleisli wraps a function (A => F[B]) into a different case class. By exposing the higher-kind parameter `F[_]` in the type system,
it makes it possible to define type-class instances and operations that we could not define for functions. This is an example of a type rotation: we re-arrange the types in the
nested type `=>[A, F[B]]` to make a plain `Kleisli[F, A, B]`.

However, one key problem of Kleisli is its performance. Because a Kleisli is an outer object `Kleisli` that wraps an inner object (a function), almost every method of the Kleisli class, and every
combinator of Kleisli objects, needs to allocate two objects: first the new inner function, and then the wrapper. This means that Kleisli-based libraries and applications are paying a performance penalty to use this abstraction.

To remedy this problem, we introduce a lightweight version of Kleisli, the trait `Clay[F[_], A, B]`, which is a trait with a single abstract method `def apply(a: A): F[B]`. This has some advantages:

- This makes it almost similar to a function of A to an F[B], but   without the methods of `Function1` that may get in the way.
- In Scala 2.12 and later we can use the Single-Abstract-Method syntax to define new Clay instances.

For the sake of feature parity, we implement in Clay all the methods of Kleisli, as well as same type-class instances. The code is pretty much copied from one to the other.

### DIFF
--- a/core/src/main/scala/cats/data/Clay.scala
+++ b/core/src/main/scala/cats/data/Clay.scala
@@ -1,0 +1,634 @@
+package cats
+package data
+
+import cats.{Contravariant, Id}
+import cats.arrow._
+
+/**
+ * A Clay represents a function `A => F[B]`, but provides an interface whose type symbol exposes the `F[_]` parameter.
+ */
+trait Clay[F[_], A, B] { self =>
+
+  def apply(a: A): F[B]
+
+  def ap[C](f: Clay[F, A, B => C])(implicit F: Apply[F]): Clay[F, A, C] =
+    new Clay[F, A, C] { def apply(a: A): F[C] = F.ap(f(a))(self(a)) }
+
+  /**
+   * Performs [[local]] and [[map]] simultaneously.
+   */
+  def dimap[C, D](f: C => A)(g: B => D)(implicit F: Functor[F]): Clay[F, C, D] =
+    new Clay[F, C, D] { def apply(c: C): F[D] = F.map(self(f(c)))(g) }
+
+  /**
+   * Modify the output of the Clay function with `f`.
+   * {{{
+   * scala> import cats.data.Clay, cats.implicits._
+   * scala> val takeHead = Clay[Option, List[Int], Int](_.headOption)
+   * scala> takeHead.map(_.toDouble).apply(List(1))
+   * res0: Option[Double] = Some(1.0)
+   * }}}
+   */
+  def map[C](f: B => C)(implicit F: Functor[F]): Clay[F, A, C] =
+    new Clay[F, A, C] { def apply(a: A): F[C] = F.map(self(a))(f) }
+
+  def mapF[N[_], C](f: F[B] => N[C]): Clay[N, A, C] =
+    new Clay[N, A, C] { def apply(a: A): N[C] = f(self(a)) }
+
+  /**
+   * Modify the context `F` using transformation `f`.
+   */
+  def mapK[G[_]](f: F ~> G): Clay[G, A, B] =
+    new Clay[G, A, B] { def apply(a: A): G[B] = f(self(a)) }
+
+  /**
+   * Tip to tail Clay arrow composition.
+   * Creates a function `A => F[C]` from [[apply]] (`A => F[B]`) and the given Clay of `B => F[C]`.
+   * {{{
+   * scala> import cats.data.Clay, cats.implicits._
+   * scala> val takeHead = Clay[Option, List[Int], Int](_.headOption)
+   * scala> val plusOne  = Clay[Option, Int, Int](i => Some(i + 1))
+   * scala> (takeHead andThen plusOne).apply(List(1))
+   * res0: Option[Int] = Some(2)
+   * }}}
+   */
+  def andThen[C](k: Clay[F, B, C])(implicit F: FlatMap[F]): Clay[F, A, C] =
+    new Clay[F, A, C] { def apply(a: A) = F.flatMap(self(a))(k.apply) }
+
+  def flatMap[C](f: B => Clay[F, A, C])(implicit F: FlatMap[F]): Clay[F, A, C] =
+    Clay.shift { a =>
+      F.flatMap[B, C](self(a))((b: B) => f(b)(a))
+    }
+
+  def flatMapF[C](f: B => F[C])(implicit F: FlatMap[F]): Clay[F, A, C] =
+    Clay.shift { a =>
+      F.flatMap(self(a))(f)
+    }
+
+  /**
+   * Contramap the input using `f`, where `f` may modify the input type of the Clay arrow.
+   * {{{
+   * scala> import cats.data.Clay, cats.implicits._
+   * scala> type ParseResult[A] = Either[Throwable, A]
+   * scala> val parseInt: Clay[ParseResult, String, Int] = Clay(s => Either.catchNonFatal(s.toInt))
+   * scala> parseInt.local[List[String]](_.combineAll).apply(List("1", "2"))
+   * res0: ParseResult[Int] = Right(12)
+   * }}}
+   */
+  def local[AA](f: AA => A): Clay[F, AA, B] = // SAM: aa => self(f(aa))
+    new Clay[F, AA, B] { def apply(aa: AA): F[B] = self(f(aa)) }
+
+  def lower(implicit F: Applicative[F]): Clay[F, A, F[B]] =
+    new Clay[F, A, F[B]] { def apply(a: A): F[F[B]] = F.pure(self(a)) }
+
+  def first[C](implicit F: Functor[F]): Clay[F, (A, C), (B, C)] =
+    new Clay[F, (A, C), (B, C)] {
+      def apply(ac: (A, C)): F[(B, C)] = F.fproduct(self(ac._1))(_ => ac._2)
+    }
+
+  def second[C](implicit F: Functor[F]): Clay[F, (C, A), (C, B)] =
+    new Clay[F, (C, A), (C, B)] {
+      def apply(ca: (C, A)): F[(C, B)] = F.map(self(ca._2))(ca._1 -> _)
+    }
+
+  def merge[C](g: Clay[F, A, C])(implicit F: Apply[F]): Clay[F, A, (B, C)] =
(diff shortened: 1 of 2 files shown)


kailuowang:
Nice idea. I am curious if you have benchmarked this v.s. kleisli?  Also have you considered instead is trait, use abstract class, which is lighter and more binary compatibility friendly?

travisbrown:
I'd be curious to see benchmarks as well. I kind of doubt this makes much of a difference in realistic use cases, and if it does, I'd personally much rather see `Kleisli` "fixed" to minimize the extra allocations (which could probably be done without breaking source compatibility, if not bincompat) instead of introducing an otherwise equivalent alternative (or just publish it as a separate micro-library?).

Also, why "clay"?

diesalbla:
@travisbrown I have no trouble trying to produce some benchmarks, to be sure. Evidence is always useful.

My basic interpretation is that, since `Kleisli` has an extra wrapper around the function, removing it is likely to halve object allocations. I don't know if the Scala compiler is clever enough to apply deforestation with a final case class like `Kleisli`, but I do not think this is the case.

I doubt you can _fix_ the Kleisli class without breaking apart from it altogether. The key point here is to make the Kleisli-like abstraction without wrapping, and wrapping is in the source and binary code. The idea of adding the light abstraction is to allow gradually deprecating the wrapped variant.

djspiewak:
I'm curious as to the motivation behind this. The allocation penalty is real, obviously, but we're talking about something which should never be in the hot path *anyway*. That's not to say that optimizations are meaningless, I'm just skeptical about whether the benefits are worth the added API load.

To that point as well, shouldn't it be possible to do this in a fashion which is API-compatible with the existing `Kleisli`? In order to do this in such a way that things are both binary and source compatible, it will require an overloaded `apply` on the companion object where one of them is a macro, but that doesn't seem like too much of a hurdle.

rossabaker:
Thanks for submitting this.  It is an interesting idea and discussion.

I agree with the calls above for _realistic_ benchmarks, which is admittedly a subjective qualifier.  There is a cottage industry of monad transformer benchmarks that show ghastly multipliers when flatmapping a trivial function to a primitive value in a trivial monad.  But in a non-trivial function over a larger payload in the IO monad, it doesn't matter.  My work is dominated by the latter.  Monad transformers are almost never the bottleneck.

Opaque types and Cats 3 both seem far enough away that I hate to keep waiting for them, but I can imagine a solution in that world that comes with all these benefits without the conceptual tax of a brand new data type.

diesalbla:
> I'm curious as to the motivation behind this. The allocation penalty is real, obviously, but we're talking about something which should never be in the hot path anyway.

There is another motivation: `Kleisli` is not just a redundant object and allocation: it is also a redundant concept in Scala. In the original [Haskell library](http://hackage.haskell.org/package/base-4.12.0.0/docs/Control-Arrow.html), it is defined as follows. 
```Haskell 
newtype Kleisli m a b = Kleisli { runKleisli :: a -> m b }
```
But in Haskell, this is not intended as a new concept: it is just a hack, special alias, used to give the type-class system the convenience of a different root type symbol (`Kleisli`), which it can use as hook for different type-class instances, separate of those it would have for `->`. It also helpe with other things, like partial type applications. 

In Scala, the `Kleisli` case class, as currently written in `cats` (and `scalaz`) follows this Haskell code to closely: the `newtype` as the `case class`, and the `runKleisli` (or `run`) as a `Function1`. But this fails to exploit some of the strengths of Scala a function can be anything with an `apply` method. And anything with an abstract method can be written as a lambda literal (since 2.12 SAM). 
The point of the `Clay` is to do both things, _swapping_ type symbols and exposing a function-like thing, in a single structure. This is not yet the optimal solution, though: the optimal solution would be to just have `type Kleisli[F[_], A, B] = A => F[B]`, and be done with it. 

> ... in a non-trivial function over a larger payload in the IO monad, it doesn't matter. My work is dominated by the latter. Monad transformers are almost never the bottleneck.

The point of type-classes is to give generic functions and combinators, to compose coarse-grained  computations out of fine-grained ones. The point of monad transformers, like `Kleisli`, is to build type-class instances for some nested data-structures. If you are already working with non-trivial functions, why would you need to use the type-classes and the transformers at all? 

To make it a more concrete example, @rossabaker, for how much of the code in `http4s` do you need the `Kleisli` data type and all the methods it brings? How difficult would it be to have those functions in an alternative way?

djspiewak:
> There is another motivation: Kleisli is not just a redundant object and allocation: it is also a redundant concept in Scala.

Oh I totally agree! Specifically, I very much agree that the current `Kleisli` encoding is not what it should be, and I'd certainly be in favor of fixing it, especially since I think we probably can fix it without breaking binary or source compatibility. I think that the encoding you're suggesting here with `Clay` is probably optimal in Scala, since we don't want to just use an alias on `Function1` due to the magical status it has within Scala's type system paired with the extreme fragility of aliases and their companions. But `Clay` as encoded is functionally equivalent to `Function1` without these issues, so it seems optimal to me.

My point was simply arguing against having `Clay` as a *separate* thing from `Kleisli`. `Clay` is motivated in the OP as a higher-performance `Kleisli` that people can choose to use, while really I think that's a strawman that no one is asking for (and I say this as someone who spends a lot of time using cats in software that has exceptionally tight performance requirements). I think the better way to do something like this would be to say that `Clay` is a *better* `Kleisli` and just replace it straight up. I would at least like to be really convinced that it *can't* be done without breaking compatibility before giving up on that avenue, since adding API surface area is a pretty serious thing that shouldn't be undertaken lightly.

kailuowang:
Just a quick note that https://github.com/estatico/scala-newtype support writing such monad transformers with zero runtime overhead. It uses the same encoding we used in cats like here
https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/Newtype2.scala
We didn't yet have the encoding for types like Kleisli, but potentially we can port it if we want to avoid dependency on scala-newtype.
 If we can't wait for opaque type, maybe we should start experimenting with scala-newtype encoding, have some benchmark. I know that the macro annotation might get frowned upon, but I think it might provide a solution for libraries to cross compile new type code between scala versions before and after opaque type.

diesalbla:
> I think the better way to do something like this would be to say that Clay is a better Kleisli and just replace it straight up. I would at least like to be really convinced that it can't be done without breaking compatibility before giving up on that avenue, since adding API surface area is a pretty serious thing that shouldn't be undertaken lightly.

Given how many hoops the `cats` maintainers go and the care they take not to break binary and source compatibility, I did not think such a proposal would fly. If it did, it would delay availability until a big release, at which point it would create a ripple though replacements in all libraries. I have no desire for that. I would prefer coexistence and a (short) deprecation cycle over a full replacement.

mpilquist:
This looks like a great technique for encoding `Kleisli`, though I'm very much against introducing another name for this concept. The cost of migrating existing code bases, training material, books, etc. isn't worth the gain in performance. If instead, this was implemented as a fix to `Kleisli`, I'd be in full support. Alternatively, this could wait until Cats 3, where binary compatibility breakages are planned.

djspiewak:
> Given that most of the Kleisli-using Byte-Code could be coupled attached to the accessor method .run, which gets an object of type Function1, I really can not see how. I am afraid that is not a nut I would bother to crack, but if you can see a way then I would certainly would see the advantages.

Actually most compiled `Kleisli`-using code almost certainly just uses `map` and `flatMap`. The majority of the pipeline would remain within there. At the end, they definitely call `run`, which is going to require allocating the `Function1`, but it can be deprecated to encourage migration.

Overall I honestly think that both source and binary compatibility can be preserved here. It may be necessary to do a bit of tricky overloading (and/or a macro) to preserve binary compatibility on the `Kleisli` companion object without foregoing the syntax that we would want, but it at least seems worth the attempt.

travisbrown:
More generally my impression is that this kind of fine-grained optimization to avoid allocations has just never been a goal in Cats—for example three years ago we were thinking of building some stuff in Finch on `StateT`, but the Cats encoding (wrapper for a `run` function) had too much of an impact on some benchmarks we cared about, so we proposed exactly the same change as here (from a `run` function member to `apply`), and then ended up publishing our own `StateT` as a separate module when nobody seemed interested in changing the Cats one.

I don't personally think this is the best approach (i.e. for Cats to provide simple, idiomatic implementations of these abstractions, and if the performance cost is too high for specific applications, you have to roll your own), but it's probably better than having either having inconsistency in the API because some abstractions are optimized and others aren't, and it's definitely better than having multiple implementations of these abstractions in Cats itself.

SystemFw:
> There is another motivation: Kleisli is not just a redundant object and allocation: it is also a redundant concept in Scala. In the original Haskell library, it is defined as follows.
```
newtype Kleisli m a b = Kleisli { runKleisli :: a -> m b }
```
But in Haskell, this is not intended as a new concept: it is just a hack, a special alias, used to give the type-class system the convenience of a different root type symbol (Kleisli), which it can use to hook different type-class instances, separate from those for ->. It also helps with other things like partial type applications, for which Scala offers type projections, and for which we can also use the kind-projector plugin.

For the record, this is not the case. The equivalent of `cats.Kleisli` in Haskell is `ReaderT`, and the point is that it needs to act as the `->` transformer, and therefore is not redundant is the same way as `OptionT` is not redundant. It just so happens that scala picked the same name as the newtype used in Haskell to give Arrow instances to `->` instead of using `ReaderT` more broadly (although it still exists as an alias).

Apart from that, I'd like to echo the previous suggestions of just fixing `Kleisli` instead, ideally with a benchmark to back up the effort

diesalbla:
>  the point is that it needs to act as the -> transformer, and therefore is not redundant in the same way as OptionT is not redundant. 

Actually, I completely agree with that.  :slightly_smiling_face:

> Apart from that, I'd like to echo the previous suggestions of just fixing Kleisli instead, ideally with a benchmark to back up the effort.

Again, I cannot see how one can modify Kleisli without breaking source and binary compatibility, and therefore delaying release for a little while. My intention was to let it out as soon as possible, even if it had to _coexist_ with the old implementation for a while. However, it seems that that is not a preferred way.

PR #2806, opened 2019-04-24
update scalatest and discipline.

kailuowang:
This shall enable a 2.0-M1 release


kailuowang:
the property based test API in scalatest 3.1.0 has more breaking change than I thought. Closing this PR for now. will reopen once I fix them.

LukaJCB:
This looks good, though it worries me a bit that the ScalaTest Api has breaking changes in a minor version. I imagine this might affect people, using `CatsSuite`, no?

kailuowang:
@LukaJCB I think in most use cases, it shouldn't affect users using CatsSuite. But part of the point of having this M1 release is so that we can test the impact. I can try it with Cats-tagless and see what migration is needed.

PR #2810, opened 2019-04-25
Change MonadErrorOps#reject so it no longer runs effects twice

### LINKED ISSUE #2809: MonadErrorOps#reject runs non-rejected effects twice
Example ammonite script:
```scala
import $ivy.`org.typelevel::cats-effect:1.2.0`
import cats.syntax.monadError._
import cats.effect.IO

IO(println("this should only be printed once")) // printed twice
  .reject { case _ if false => new Throwable}
  .unsafeRunSync()
```

Fixed in https://github.com/typelevel/cats/pull/2810

bplommer:
Fixes #2809


LukaJCB:
Thanks for bringing this to our attention, this indeed sounds like a bug. Technically this is a breaking change, but I think we should merge this as a bug fix 👍

PR #2817, opened 2019-04-28
added a new foldRight lazy law, move forallLazy and existLazy laws

kailuowang:
I noticed that `foldRight`'s laziness was tested through `forallLazy` and `existLazy` laws. Since `foldRight` is a type class method I think we should have the laziness law on it. In kittens we were caught by the lack of such a law. 

Also I noticed that although `forallLazy` and `existLazy` are defined in `UnorderedFoldableLaws`, they are actually only used in `FoldableTests`. Moreover, implementations in `UnorderedFoldable` breaks these two laws. So I think they should probably be `FoldableLaws`. 

Update: This actually caught a bug in NonEmptyChain whose `Foldable.foldRight` implementation is simply delegating to the native (scala std lib) `foldRight`.

### DIFF
--- a/laws/src/main/scala/cats/laws/FoldableLaws.scala
+++ b/laws/src/main/scala/cats/laws/FoldableLaws.scala
@@ -17,16 +17,15 @@ trait FoldableLaws[F[_]] extends UnorderedFoldableLaws[F] {
       b |+| f(a)
     }
 
-
   def foldRightLazy[A](fa: F[A]): Boolean = {
     var i = 0
     F.foldRight(fa, Eval.now("empty")) { (_, _) =>
-      i += 1
-      Eval.now("not empty")
-    }.value
+        i += 1
+        Eval.now("not empty")
+      }
+      .value
     i == (if (F.isEmpty(fa)) 0 else 1)
   }
-
 
   def rightFoldConsistentWithFoldMap[A, B](
     fa: F[A],
@@ -122,7 +121,6 @@ trait FoldableLaws[F[_]] extends UnorderedFoldableLaws[F] {
   def orderedConsistency[A: Eq](x: F[A], y: F[A])(implicit ev: Eq[F[A]]): IsEq[List[A]] =
     if (x === y) (F.toList(x) <-> F.toList(y))
     else List.empty[A] <-> List.empty[A]
-
 
   def existsLazy[A](fa: F[A]): Boolean = {
     var i = 0
--- a/laws/src/main/scala/cats/laws/FoldableLaws.scala
+++ b/laws/src/main/scala/cats/laws/FoldableLaws.scala
@@ -11,9 +11,9 @@ trait FoldableLaws[F[_]] extends UnorderedFoldableLaws[F] {
   def foldRightLazy[A](fa: F[A]): Boolean = {
     var i = 0
     F.foldRight(fa, Eval.now("empty")) { (_, _) =>
-      i += 1
-      Eval.now("not empty")
-    }
+        i += 1
+        Eval.now("not empty")
+      }
       .value
     i == (if (F.isEmpty(fa)) 0 else 1)
   }
--- a/laws/src/main/scala/cats/laws/FoldableLaws.scala
+++ b/laws/src/main/scala/cats/laws/FoldableLaws.scala
@@ -8,6 +8,16 @@ import scala.collection.mutable
 trait FoldableLaws[F[_]] extends UnorderedFoldableLaws[F] {
   implicit def F: Foldable[F]
 
+  def foldRightLazy[A](fa: F[A]): Boolean = {
+    var i = 0
+    F.foldRight(fa, Eval.now("empty")) { (_, _) =>
+      i += 1
+      Eval.now("not empty")
+    }
+      .value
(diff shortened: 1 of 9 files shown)


kailuowang:
BTW, strictly speaking this is breaking change for `UnorderedFoldableLaws` but it's unlikely to affect anyone and we are breaking BC in cats-laws in our next release.

joroKr21:
I guess it makes sense because someone might override `forall` and `exists` before testing the laws and then not realize that the `foldRight` implementation is not lazy.

joroKr21:
This mistake is actually incredibly easy to make. Here is a silly example:
```scala
    implicit val foldableSeq: Foldable[Seq] = new Foldable[Seq] {
      def foldLeft[A, B](fa: Seq[A], b: B)(f: (B, A) => B) = fa.foldLeft(b)(f)
      def foldRight[A, B](fa: Seq[A], lb: Eval[B])(f: (A, Eval[B]) => Eval[B]) = fa.foldRight(lb)(f)
      override def forall[A](fa: Seq[A])(p: A => Boolean) = fa.forall(p)
      override def exists[A](fa: Seq[A])(p: A => Boolean) = fa.exists(p)
    }
```

kailuowang:
@joroKr21 funny that's exactly what I just discovered on `NonEmptyChain`

kailuowang:
cc @non, @ceedubs and @johnynek  who may have opinion over this.

laws/src/main/scala/cats/laws/FoldableLaws.scala
```diff
@@ -8,6 +8,16 @@ import scala.collection.mutable
 trait FoldableLaws[F[_]] extends UnorderedFoldableLaws[F] {
   implicit def F: Foldable[F]
 
+  def foldRightLazy[A](fa: F[A]): Boolean = {
+    var i = 0
+    F.foldRight(fa, Eval.now("empty")) { (_, _) =>
+        i += 1
```
johnynek: We use += here but not below. Any reason for the difference?

laws/src/main/scala/cats/laws/UnorderedFoldableLaws.scala
```diff
@@ -21,24 +21,6 @@ trait UnorderedFoldableLaws[F[_]] {
       (F.isEmpty(fa) || F.exists(fa)(p))
     } else true // can't test much in this case
 
-  def existsLazy[A](fa: F[A]): Boolean = {
```
johnynek: Foldable extends UnorderedFoldable. To me, the bug here is that we weren’t calling these in UnorderedFoldable tests and that FoldableLaws were not calling through to them as we should do any time we extend a typeclass.

That seems like the better fix to me than moving the law (since we should still want these laws on UnorderedFoldable).

kailuowang: It does look that way but actually the default implementations of `exist` and `forall` in `UnorderedFoldable` do not obey these laws. These laws can only be supported in `Foldable` through `foldRight`, hence they should be in `Foldable`

laws/src/main/scala/cats/laws/FoldableLaws.scala
kailuowang: no real reason except this one is written by me and the other ones are from before. Just changed the other ones to be more consistent.

laws/src/main/scala/cats/laws/UnorderedFoldableLaws.scala
rossabaker: We can keep the laws in `UnorderedFoldableTests` if we implement the `CommutativeMonoid[Eval[Boolean]]`s directly in `UnorderedFoldable`.  See https://github.com/rossabaker/cats/commit/4af246d0f4e2377e9e94defe3eedc3ab64cd9dbf.

johnynek:
If we are going to keep a laziness law, my vote is to keep it on UnorderedFoldable, if we aren’t going to do that, I’d next prefer to remove the laziness requirement entirely.

joroKr21:
> That said, the worst case is already linear complexity so maybe that law is too restrictive anyway. Maybe the real mistake is forcing laziness on Foldable.

I think one might expect the laziness from the type signature.

rossabaker:
The commit I linked above keeps the lazy laws on `UnorderedFoldable`, passes the tests for `Map[K, ?]` and `Set` with the new default implementation, and preserves bincompat in core.  Are there any desiderata that it misses?

johnynek:
@rossbaker I think the approach is good but I have a few nits:

1. We should override exists and forall for our concrete instances.
2. In the Eval monoids I’d rather not reevaluate the Eval which could be expensive in principle.

kailuowang:
Thanks for all the feedbacks. I will incorporate @rossbaker's code with @johnynek's suggestions

rossabaker:
I agree with @johnynek's suggestions, but if we override for `Map` and `Set`, that will leave the default implementations untested.  I'm unsure whether we've developed an idiom for testing unused defaults.

kailuowang:
@rossabaker https://github.com/typelevel/cats/blob/master/tests/src/test/scala/cats/tests/UnorderedFoldableSuite.scala  is already setup to test some other default implemetations. This should be easy to add.

Issue #2818, opened 2019-04-28
Additional example for FreeT in docs


mat646:
Hi, 
I have recently used FreeT and found that docs provide only pretty specific example with State monad. 

Would you mind accepting docs PR with additional basic `Future` `Option` example?

After consulting on gitter I ended up with additional transformation for `Option ~> OptionT`, which I am not sure.

```scala

import cats.free._
import cats._
import cats.data._
import cats.implicits._
import scala.concurrent.ExecutionContext.Implicits.global
import scala.concurrent.Future

sealed trait Ctx[A]

case class Action(value: Int) extends Ctx[Int]

def op1: FreeT[Ctx, Option, Int] =
  FreeT.liftF[Ctx, Option, Int](Action(7))

def op2: FreeT[Ctx, Option, Int] =
  FreeT.liftT[Ctx, Option, Int](Some(4))

def op3: FreeT[Ctx, Option, Int] =
  FreeT.pure[Ctx, Option, Int](1)

val opComplete: FreeT[Ctx, Option, Int] =
  for {
    a <- op1
    b <- op2
    c <- op3
  } yield a + b + c


// ----interpreters----

type OptFut[A] = OptionT[Future, A]

def futureInterpreter: Ctx ~> OptFut = new (Ctx ~> OptFut) {
  def apply[A](fa: Ctx[A]): OptFut[A] = {
    fa match {
      case Action(value) => OptionT.liftF(Future(value))
    }
  }
}

def optFutLift: Option ~> OptFut = new (Option ~> OptFut) {
  def apply[A](fa: Option[A]): OptFut[A] = {
    fa match {
      case Some(value) =>
        OptionT(Future(Option(value)))
      case None =>
        OptionT.none
    }
  }
}

val hoisted = opComplete.hoist(optFutLift)
val evaluated = hoisted.foldMap(futureInterpreter)

val future = evaluated.value
// Some(12)

```

kailuowang:
👍 for Option. Mix feelings for `Future`. `Future` is widely used but we should be encouraging people to move away from it. Maybe we should add a FAQ item on this topic?

mat646:
#2819

PR #2819, opened 2019-04-28
added FreeT basic example

mtsokol:
#2818

### DIFF
--- a/docs/src/main/tut/datatypes/freemonad.md
+++ b/docs/src/main/tut/datatypes/freemonad.md
@@ -504,7 +504,7 @@ In the following example a basic console application is shown.
 When the user inputs some text we use a separate `State` monad to track what the user
 typed.
 
-As we can observe in this case `FreeT` offers us a the alternative to delegate denotations to `State`
+As we can observe in this case `FreeT` offers us the alternative to delegate denotations to `State`
 monad with stronger equational guarantees than if we were emulating the `State` ops in our own ADT.
 
 ```tut:book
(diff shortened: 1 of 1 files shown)


docs/src/main/tut/datatypes/freemonad.md
```diff
@@ -560,6 +560,67 @@ val initialState = Nil
...
+
+val opComplete: FreeT[Ctx, Option, Int] =
+  for {
+    a <- op1
+    b <- op2
+    c <- op3
+  } yield a + b + c
+
+
+/* Our interpreters e.g. for Future, DBIO, Id  */
```
kailuowang: I'd say let's not use `Future` in our documentation. Just don't want to give people the impression that it's not a bad idea to use `Future`.  For the purpose of your example, we could use `Either` or something

mat646: I ended up with `Try` instead of `Future`.

You have mentioned about adding FAQ item about moving away from `Future`.
So e.g. short paragraph about crucial reasons providing some sources for further reading?
Something like [cats effect docs](https://typelevel.org/cats-effect/datatypes/io.html#on-referential-transparency-and-lazy-evaluation) comparison?

mat646: One of the jobs has failed somehow with "GC overhead limit exceeded" and I'm not able to rerun it. Could you rerun it? Or should I trigger it (e.g. close and open pull)?

kailuowang: Sorry about the last response. Yeah, that cats effect docs is great as a further reading!

PR #2823, opened 2019-04-30
[WIP] Semaphore CI

### LINKED ISSUE #2319: Evaluate moving to Circle CI

DavidGregory084:
This PR is one possible option to fix #2319. It adds configuration for [Semaphore CI](https://semaphoreci.com). This looks like a good option but it's no so clear how or when they plan to support open source, which is strange because I believe Semaphore version 1 did so. The free plan is limited to $20 of build minutes per month at $0.00050/sec, which is ridiculously low.

If you guys think it's worth it I will contact them to try and clarify their policy on open source.

Compared with Travis it is very fast (about 15-16 min per build).

The configuration is well documented [here](https://docs.semaphoreci.com/article/50-pipeline-yaml) and it was reasonably straightforward to set up.

The execution model is easy to understand; each block runs sequentially and jobs within a block run in parallel.

There is also a matrix feature which can be used to declare a build matrix using multiple environment variables if needed.

The only niggle I would mention is that it is not possible to add unspecified keys at the top level to use to hold the scala versions so I had to use `env_vars` blocks for that purpose.

From what I have seen in evaluating different services the flakiest test step which often struggles to run within memory limits is the Applicative.monoid.combineAll test, which frequently hits the GC overhead limit. I will raise that as a separate issue.

Screenshot 2019-04-30 at 16 55 41

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -5,10 +5,6 @@ agent:
     type: e1-standard-4
     os_image: ubuntu1804
 
-scala_version_211: &scala_version_211 2.11.12
-scala_version_212: &scala_version_212 2.12.8
-scala_version_213: &scala_version_213 2.13.0-RC1
-
 blocks:
   - name: Linting
     task:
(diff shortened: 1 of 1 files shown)


fbrdaric:
@DavidGregory084 I'm Filip from Semaphore Customer Success team. I'm glad to see you're trying out Semaphore! The full OS support for Semaphore 2.0 is planned to be shipped in the upcoming weeks. It'll include clear guidelines on supporting OS projects in terms of consumption. I'd be happy to share more details around this via email. If you're interested, please reach out to customersuccess@semaphoreci.com. I might be able to help you out right away :).

larsrh:

PR #2824, opened 2019-04-30
[WIP] Drone.io

### LINKED ISSUE #2319: Evaluate moving to Circle CI

DavidGregory084:
This PR is one possible option to fix #2319. It adds configuration for [Drone CI](https://drone.io), which is free for open source.

It uses containers so I needed to create a [Docker image](https://github.com/DavidGregory084/cats-builder) with the right tools to build cats.

Compared with Travis it is very fast (about 15-16 min per build).

The configuration is reasonably well documented [here](https://readme.drone.io/user-guide/pipeline/) and it was straightforward to set up.

Unlike Semaphore and Travis there is no build matrix option, so there is a bit of annoying repetition in the config. I plan to tidy this up a bit with YAML anchors.

So far I have had no issue with builds running out of memory, instead I'm currently trying to debug a weird issue where the 2.12 build consistently fails to compile cats-kernel for reasons that I don't understand.

image

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -28,7 +28,7 @@ steps:
     - pwd
     - ls
     - env
-    - sbt ++$TRAVIS_SCALA_VERSION! validateBC
+    - sbt ++$TRAVIS_SCALA_VERSION validateBC
     environment:
       TRAVIS_SCALA_VERSION: *scala_version_211
     
@@ -38,7 +38,7 @@ steps:
     - pwd
     - ls
     - env
-    - sbt ++$TRAVIS_SCALA_VERSION! validateBC
+    - sbt ++$TRAVIS_SCALA_VERSION validateBC
     environment:
       TRAVIS_SCALA_VERSION: *scala_version_212
 
@@ -61,7 +61,7 @@ steps:
     - pwd
     - ls
     - env
-    - sbt ++$TRAVIS_SCALA_VERSION! buildJVM bench/test
+    - sbt ++$TRAVIS_SCALA_VERSION buildJVM bench/test
     environment:
       TRAVIS_SCALA_VERSION: *scala_version_211
     depends_on:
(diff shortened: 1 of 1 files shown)


kailuowang:
That's doctest. Not obvious to me why it couldn't find jars of cats-kernel...

DavidGregory084:
Yeah, it's very weird, I think my best bet is to try running it locally with the [Drone CLI](https://docs.drone.io/cli/) to see if I can reproduce it.
It looks like SBT doesn't even try to compile the main sources before it tries to compile the test sources?

kailuowang:
Drone build link if anyone is interested
https://cloud.drone.io/DavidGregory084/cats

DavidGregory084:
Haven't made a lot of progress on that lone build error - running locally with the Drone CLI it passes if I run that single step

If I don't get anywhere I'll ping some people on the SBT Gitter channel to see if they have any ideas

kailuowang:
In our travis we are forcing scala version , which I think we don't need any more. 
i.e. all `sbt ++$TRAVIS_SCALA_VERSION!` should be `sbt ++$TRAVIS_SCALA_VERSION`
I am not sure if it's related to this weird issue on drone, but it might worth a try.

DavidGregory084:
Hmm, seems like a weird result - now I have dependency resolution problems on 2.11. I'm afraid I don't have a lot of time to investigate this at the moment as I'm on holiday for a couple weeks.

Is master / 2.0.0-SNAPSHOT 2.12 and above @kailuowang?

kailuowang:
That 2.11 error is related to sbt cross build. I will take a look. 

 Right now our master is still cross building 2.11-2.13, but as soon as 2.0 release we'll remove 2.11 on master, Then we will separate out 2.11 to its own branch which will might reduce cross build complexity. 

Enjoy your holiday.

larsrh:

PR #2825, opened 2019-04-30
[WIP] Buildkite CI

### LINKED ISSUE #2319: Evaluate moving to Circle CI

DavidGregory084:
This PR is one possible option to fix #2319. It adds configuration for [Buildkite CI](https://drone.io), which is free for open source.

Buildkite is a hosted service but it provides only the coordination of builds, not the hardware. You need to install the Buildkite agent on your own hardware in order to run builds.

In order to test this I used a server I provisioned on [Packet](https://packet.com), which was really easy to set up. It's possible to use [cloud-init](https://cloudinit.readthedocs.io/en/latest/) to preconfigure your server so that it's ready to go immediately once it's provisioned. The configuration that I used can be viewed [here](https://gist.github.com/DavidGregory084/b3166d113ab81d13e49cbc296aadfffa). 

Packet do have an [open source policy](https://www.packet.com/developers/open-source/) so it's possible that they would donate build machines if we contact them.

Compared with Travis it is faster but once all of the cross builds were added, with the hardware that I used it's much slower to run a whole build than Drone or Semaphore (about 25-26min). This may just be due to the hardware that I used (4 cores @ 3.5GHz with 32GB RAM).

The configuration is well documented [here](https://buildkite.com/docs/pipelines) and it was straightforward to set up. There is also a visual editor for config on the web interface.

Like Drone there is no build matrix option, so there is a bit of annoying repetition in the config and I plan to tidy this up a bit with YAML anchors.

Our old friend ApplicativeTest.monoid.combineAll makes an appearance here too:

image

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,0 +1,53 @@
+env:
+  TRAVIS: "true"
+  
+scala_version_211: &scala_version_211 2.11.12
+scala_version_212: &scala_version_212 2.12.8
+scala_version_213: &scala_version_213 2.13.0-RC1
+
+steps:
+  - command: "sbt scalastyle fmtCheck"
+    label: ":scala: Linting"
+
+  - command: "cd scalafix && sbt tests/test"
+    label: ":scala: Scalafix"
+
+  - command: "sbt ++\\$TRAVIS_SCALA_VERSION! validateBC"
+    label: ":scala: Bincompat 2.11"
+    env:
+      TRAVIS_SCALA_VERSION: *scala_version_211
+
+  - command: "sbt ++\\$TRAVIS_SCALA_VERSION! validateBC"
+    label: ":scala: Bincompat 2.12"
+    env:
+      TRAVIS_SCALA_VERSION: *scala_version_212
+
+  - wait
+
+  - command: "sbt coverage buildJVM bench/test coverageReport"
+    label: ":codecov: Coverage"
+
+  - command: "sbt ++\\$TRAVIS_SCALA_VERSION! buildJVM bench/test"
+    label: ":java: JVM Tests 2.11"
+    env:
+      TRAVIS_SCALA_VERSION: *scala_version_211
+
+  - command: "sbt ++\\$TRAVIS_SCALA_VERSION! buildJVM bench/test"
+    label: ":java: JVM Tests 2.12"
+    env:
+      TRAVIS_SCALA_VERSION: *scala_version_212
+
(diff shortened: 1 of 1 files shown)


DavidGregory084:
One downside that I have seen with Buildkite is that the agents all run with the same user ID if they are on the same machine, so they share a coursier cache and numerous SBT lockfiles.

This can lead to weird errors like this, which I think might possibly be caused by multiple Coursier processes downloading the same dependency at once?

image

As a result, there is most likely a bit of investigation needed into the optimal SBT configuration for running multiple processes at once on the same box.

Alternatively, Buildkite [can be configured](https://buildkite.com/docs/tutorials/docker-containerized-builds) to use Docker to run build steps which would sidestep this problem at the cost of downloading all dependencies on each build.

tpolecat:
Unless they changed something, BuildKite only builds PRs from branches on the same repo by default, because a PR can change the build so it's arbitrary code execution. Is there a solution to this limitation?

DavidGregory084:
I think you can change that on the pipeline settings page @tpolecat, there is a checkbox you can enable to "Build pull requests from third-party forked repositories"

tpolecat:
> there is a checkbox you can enable to "Build pull requests from third-party forked repositories"

Right, but if you do this a malicious PR can destroy the build infrastructure since it can introduce arbitrary code that will be run as part of the (automatic) build.

DavidGregory084:
Ah, I see what you mean.

It looks like there are some discussion about how to make this more secure on buildkite/feedback#360.

There's also a [third-party tool](https://github.com/mvines/ci-gate) which allows you to supress CI for unknown contributors until a maintainer adds a tag to the PR.

That doesn't seem like a soluble problem in general without some kind of sandboxing tech though.

DavidGregory084:
One option is that the pipeline definition is not stored in the repo and is configured only through the BuildKite UI - that way it could only be edited by members of that organisation on BuildKite.
That may be an acceptable compromise as I suspect it's not too common that anyone other than cats maintainers edits the CI config.

kailuowang:
Supress CI for unkown contributor probably won't work for us, it's tricky to determine which new contributors are "known". Also it adds overhead for new contributors. 
I am not sure pipeline definition being secure alone would be enough to guard against malicious PR. afterall the PR can run any code as part of the test. 
Sounds to me that this BuildKite is more catered toward as an internal CI solution.

DavidGregory084:
Yeah that may be the case @kailuowang, and the burden of managing the hardware for the build agents would need to be taken into consideration too

larsrh:

PR #2828, opened 2019-05-03
make stage more explicit

kailuowang:
This also results in a very short lint stage and a longer test stage, which was my original intention.


rossabaker:
What runs `scalastyle` now?

kailuowang:
@rossabaker, my bad. Forgot to mention that I removed it since scalafmt is much more strict, scalastyle keeps failing on parsing from time to time. Not sure if we need to maintain two sets of styling rules.

rossabaker:
In that case, `validateJVM` still calls it.  If we're not going to enforce it on the build, we should also remove it from there.  It's also still mentioned in `CONTRIBUTING.md`.

We could even go a step further and remove the config and the plugin, unless we think it's still providing some marginal value when run manually.

kailuowang:
@rossabaker good point. I will udpate this PR with complete scalastyle removal.

kailuowang:
@rossabaker scalastyle was removed in another PR and merged. That should be it right?

PR #2829, opened 2019-05-03
remove deprecated methods from eq

kailuowang:
since cats-laws already breaks bin compat, we no longer need these bin compat and deprecated methods any more. Also they are causing deprecation warning in the downstream.


kailuowang:
replaced by #2830

PR #2832, opened 2019-05-06
#resolves 2715 add minimumoption and maximumoption to unorderedfoldable

mschuwalow:

### DIFF
--- a/core/src/main/scala/cats/Functor.scala
+++ b/core/src/main/scala/cats/Functor.scala
@@ -1,6 +1,6 @@
 package cats
 
-import simulacrum.typeclass
+import simulacrum.{noop, typeclass}
 
 /**
  * Functor.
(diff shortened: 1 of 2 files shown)


kailuowang:
Is this urgent? if not can we wait on adding them directly to the type class once 2.0 is released?

mschuwalow:
> Is this urgent? if not, can we wait a bit so that we can add them directly to the type class once 2.0 is released?

Sure we can do that. Just comment here once this becomes relevant and I'll update the pr

travisbrown:
@mschuwalow Would you want to revisit this now that 2.0 is out and the `BinCompat` trait isn't necessary?

mschuwalow:
Sure, I'll try to update the pr this week

LukaJCB:
Hey @mschuwalow, did you close this on purpose? :) 
If you no longer want to update the PR do you mind if someone else goes for it?

PR #2837, opened 2019-05-10
Delete unnecessary restriction on biFlatMap

AdrianRaFo:
This PR delete an unnecessary type restriction on the `EitherT#biFlatMap` implementation

### DIFF
--- a/core/src/main/scala/cats/data/EitherT.scala
+++ b/core/src/main/scala/cats/data/EitherT.scala
@@ -85,7 +85,7 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
                                                          applicativeG: Applicative[G]): G[EitherT[F, C, D]] =
     applicativeG.map(traverseF.traverse(value)(axb => Bitraverse[Either].bitraverse(axb)(f, g)))(EitherT.apply)
 
-  def biflatMap[AA >: A, BB >: B](fa: A => EitherT[F, AA, BB],
+  def biflatMap[AA, BB](fa: A => EitherT[F, AA, BB],
                                   fb: B => EitherT[F, AA, BB])(implicit F: FlatMap[F]): EitherT[F, AA, BB] =
     EitherT(F.flatMap(value) {
       case Left(a)  => fa(a).value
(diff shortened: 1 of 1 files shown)


core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -85,8 +85,8 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
                                                          applicativeG: Applicative[G]): G[EitherT[F, C, D]] =
     applicativeG.map(traverseF.traverse(value)(axb => Bitraverse[Either].bitraverse(axb)(f, g)))(EitherT.apply)
 
-  def biflatMap[AA >: A, BB >: B](fa: A => EitherT[F, AA, BB],
-                                  fb: B => EitherT[F, AA, BB])(implicit F: FlatMap[F]): EitherT[F, AA, BB] =
+  def biflatMap[AA, BB](fa: A => EitherT[F, AA, BB],
```
kailuowang: Thanks! Good catch. Would you change these type param names to `C` and `D` to be more consistent? 
Actually there is one more here
https://github.com/typelevel/cats/blob/5ca98e71a44cc0e0bbc060392c677a64f545a7b8/core/src/main/scala/cats/data/EitherT.scala#L34, change that `AA`to `C` will make it more consistent as well.

kailuowang:
sorry I didn't make it clear. I didn't mean to change all `AA` and `BB` to `C` and `D`, only those that do not have a `AA :> A` qualifier ones. I think it makes sense to differentiate between transforming a type to a new type and widening it.

AdrianRaFo:
Oh ok, that's make sense

AdrianRaFo:
Could you merge it? I have no permissions to do that

kailuowang:
@AdrianRaFo we have a policy of two maintainer sign off for a merge. Calling @LukaJCB ...

kailuowang:
thanks  @djspiewak

avdv:
:heart: Nice! I just noticed the same restriction and almost gave in to the reflex of creating a PR... ;-)

PR #2839, opened 2019-05-10
add some type class instances for PartialFunction

mberndt123:
Hi there,

I recently asked on gitter why there are no type class instances for `PartialFunction`. The answer I got was that it was discussed a while ago and no definitive conclusion was reached. So I thought I'll just implement some of them and see if anybody is interested. Let me know what you think!


mberndt123:
Apparently this code results in ambiguities because `PartialFunction` is a subtype of `Function1`. Could it be that this is the reason this wasn't done before?

mberndt123:
nobody seems to care, so whatever.

Issue #2749, opened 2019-03-10
should Kleisli contravariant on A?


jcouyang:
Kleisli[F[_], A, B] is supposed to be abstraction of `A => F[B]`, but `=>` has type of Function1[-A, +B], I can't figure out why Kleisli is invariant on both A and B, is there any benefit that it's design like this?

from what I observed, it will be much easier to let scala compiler find out what A should be

for instance, if I would like to choose from two Kleisli
### contravariant on -A
```scala
case class Kleisli[F[_], -A, B](run: A => F[B])

def first[A, B, C, F[_]](a: Kleisli[F, A, B], b: Kleisli[F, A, C]) = a

trait A1
trait A2
trait A12 extends A1 with A2

first(Kleisli((a: A1) => Some(a)), Kleisli((b:A2)=>Some(b)))
res7: Kleisli[Some, A2 with A1, A1] =  ...
```

compiler can infer the correct type of the input of Kleisli  should be `A2 with A1` and the output will be `A1`

to achieve the same thing without `-A` will be much difficult

### by contramap
```scala
def first[A1, A2, A12, B, C, F[_]](a: Kleisli[F, A1, B], b: Kleisli[F, A2, C])
  (implicit ev: A12<:<A1):Kleisli[F, A12, B] = 
  Contravariant[Kleisli[F, ?,B]].contramap(a)(ev)
```

```scala
@ first(Kleisli((a: A1) => Some(a)), Kleisli((b:A2)=>Some(b)))
cmd14.sc:1: type mismatch;
 found   : Nothing <:< Nothing
 required: A12 <:< ammonite.$sess.cmd9.A1
val res14 = first(Kleisli((a: A1) => Some(a)), Kleisli((b:A2)=>Some(b)))
                 ^
```
the compiler can't infer A12 for contramap unless I explicitly tell the compiler that I need `Kleisli[Some, A12, A1]`
```scala
@ first(Kleisli((a: A1) => Some(a)), Kleisli((b:A2)=>Some(b))) : Kleisli[Some, A12, A1]
res14: Kleisli[Some, A12, A1] = Kleisli(scala.Function1$$Lambda$35/695682681@1e3c4c12)
```

Maatary:
Hi @jcouyang I wonder if you could help me with some pointers on the issue you raised. 

Although I understand the explanation above, and experienced it when composing Kleisli, I am looking for the formal definition of things. 

[...]

PR #2843, opened 2019-05-14
Kleisli contravariant on input type A

jcouyang:
Rationale: #2749
also as described in the test, it's more composable
```scala
    val program = for {
      k1 <- Kleisli((a: A1) => List(1))
      k2 <- Kleisli((a: A2) => List("2"))
      k3 <- Kleisli((a: A3) => List(true))
    } yield (k1, k2, k3)
```
scala can infer type to be `Kleisli[List, A1 with A2 with A3, (Int, String, Boolean)]`
if not using scala subtyping it would looks like
```scala
    val program = for {
      k1 <- Kleisli((a: A1) => List(1)).local(identity[A1 with A2 with A3])
      k2 <- Kleisli((a: A2) => List("2")).local(identity[A1 with A2 with A3])
      k3 <- Kleisli((a: A3) => List(true)).local(identity[A1 with A2 with A3])
    } yield (k1, k2, k3)
```

### DIFF
--- a/core/src/main/scala/cats/data/Kleisli.scala
+++ b/core/src/main/scala/cats/data/Kleisli.scala
@@ -7,9 +7,9 @@ import cats.arrow._
 /**
  * Represents a function `A => F[B]`.
  */
-final case class Kleisli[F[_], A, B](run: A => F[B]) { self =>
+final case class Kleisli[F[_], -A, B](run: A => F[B]) { self =>
 
-  def ap[C](f: Kleisli[F, A, B => C])(implicit F: Apply[F]): Kleisli[F, A, C] =
+  def ap[C, AA <: A](f: Kleisli[F, AA, B => C])(implicit F: Apply[F]): Kleisli[F, AA, C] =
     Kleisli(a => F.ap(f.run(a))(run(a)))
 
   /**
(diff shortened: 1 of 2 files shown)


kailuowang:
I am a little concerned with the source breaking widening, although the chance of them actually breaking on usesite is probably not very high.
To avoid source breaking change, an alternative might be adding a `narrow` method to `Kleisli`.  Unfortunately the `narrow` syntax comes with `Contravariant` doesn't work due to fact that the contravariant type `A` is not the most right, and as a result, partial unification doesn't help.
We can still add a `def narrow[AA <: A]: Kleisli[F, AA, B] = this.asInstanceOf`, then user can manually call `kleisli.narrow` when needed. That would probably make life easier for a lot of use cases. 

I would like to invite feedbacks from @rossabaker and @ChristopherDavenport on this one, given that `Kleisli` is the primitive in `Http4s`.

rossabaker:
I don't have a great intuition for inference problems.  I'll try to publish a local version of cats and build http4s against that and look for trouble this week.

kailuowang:
Thanks a lot @rossabaker

jdegoes:
@jcouyang 👍 

Not only should Kleisli be contravariant in `A`, but it should also be covariant on `B`.

It will most likely be source breaking, because in some cases users will have to annotate methods with the newly-added type parameter.

LukaJCB:
@jdegoes I disagree, `B` can only be covariant if `F` has a `Functor` instance and vice versa for covariant. This means that by default `Kleisli` can only be invariant, unless we want to specialize it to only covariant functors.

jdegoes:
@LukaJCB Yes, I believe it's the correct decision to specialize for covariant functors.

If anyone uses Kleisli with contravariant functors — and I haven't seen that in the wild — then this use case should be specialized as `ContraKleisli`, for example, with contravariant annotation on `B`.

This is one of the many cases where embracing specialization results in unquestionably higher ergonomics at the loss of abstraction of questionable value that most users don't care about.

kaishh:
What about revisiting the stance on covariance in monad transformers if we're at it?
I don't find the arguments in [FAQ](https://github.com/typelevel/cats/blob/master/docs/src/main/tut/faq.md#monad-transformer-variance) particularly convincing, as ALL monads must be covariant by definition, if we desire to apply a monad-transformer like structure to a non-monad, maybe we ought to just make a different type? e.g. `OptionNested`, `KleisliLike`, etc.

jdegoes:
@kaishh 👍 Major surgery required.

jcouyang:
Thanks @jdegoes, yeah I think as well it should be -A and +B
while for now contravariant has more practical use case(it's hard to predict wisely what should a composition of various Kleisli take as input type beforehand, but it's relatively easy to reason and adapt output type) and less likely widely breaking source (I published local and one of my project that using http4s still compiled,  like to see what outcome @rossabaker get though )

PR #2844, opened 2019-05-14
added sponsors

kailuowang:


ChristopherDavenport:
All of the image links appear borken when I navigate to the README on your page. Is it due to it being your page?

kailuowang:
@ChristopherDavenport yes, that's because those are hard linked to the website (to support display on both github readme and website index.html). it will show up after we publish the website.

PR #2845, opened 2019-05-14
Use Arrow-like parameter syntax for Arrow type-classes

diesalbla:
This is a syntactic sugaring of the type-class definitions in the `cats.arrow` package. The idea is that, instead of using the `F` symbol, and since Scala has a flexible syntax when it comes to identifiers, let us use a parametric arrow-like symbol. 

Using an arrow-like symbol helps reinforce the idea that an arrow is like a `Function` (or `=>`( with some effects attached. Also, using the arrow symbol makes it easier to put it in infix type notation, which is handy when it comes to operations involving tuples or `Either`.

### DIFF
--- a/core/src/main/scala/cats/arrow/Arrow.scala
+++ b/core/src/main/scala/cats/arrow/Arrow.scala
@@ -6,7 +6,7 @@ import simulacrum.typeclass
 /**
  * Must obey the laws defined in cats.laws.ArrowLaws.
  */
-@typeclass trait Arrow[F[_, _]] extends Category[F] with Strong[F] { self =>
+@typeclass trait Arrow[ |==>[_, _]] extends Category[|==>] with Strong[|==>] { self =>
 
   /**
    *  Lift a function into the context of an Arrow.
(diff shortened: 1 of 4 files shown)


travisbrown:
I'm :-1: on this, for (at least) three reasons: `F` is immediately recognizable as a parameter, while anything symbolic is likely to get confused with an alias defined elsewhere; `|==>` specifically may collide with aliases defined in other libraries or user code; and forcing users to think about operator precedence, etc. in Scala is unnecessarily cruel.

diesalbla:
@travisbrown Thank you for your comments. Allow me to address some of your concerns. 

In general, I know non-alphanumeric characters can be tricky, and I do agree that funny syntax should be used with care. Nevertheless, given that `Arrow` is a rather advanced abstraction, I do not think syntax be the major difficulty to understand them; and this is offset by the major benefit, which is to reinforce the intuition of Arrows as "functions with effects". 

> `F` is immediately recognizable as a parameter

The `F` is used across cats as the symbol for a type parameter of high kind. However, IMHO we should not use the same symbol for type-parameters of kind `* -> *` (such as `Functor[F]`, `Monad[F]`, or `Kleisli[F, In, Out]`, and those of kind `(*,*) -> *` like `Arrow` or `Strong`. 

> anything symbolic is likely to get confused with an alias defined elsewhere; |==> specifically may collide with aliases defined in other libraries or user code

Although this can be a general problem, I cannot see how it would be here: as a type parameter, each `|==>` is completely private to the trait / type-class in which it is defined. Not even sub-classes see it. 

> forcing users to think about operator precedence, etc. in Scala is unnecessarily cruel.

I neither recommend nor was intending to rely on precedence rules, and I prefer to use parentheses to mix several infix type operators.

PR #2848, opened 2019-05-17
Enables Hydra on cats build 

kailuowang:
Thanks to @dotta 's generous help and donation, we can enable hydra on Cats build to potentially speed up build as well as to obtain compilation metrics. The metrics can be viewed here https://dashboard.triplequote.com
The code change here is simply moving the traits defined in a package object file into their own file. This is due to a hydra issue that triplequote is working on.


dotta:
One thing I noticed while moving bits around is that there is no `package object finiteDuration`​ inside `cats/kernel/instances/finiteDuration/package.scala`. Was it on purpose? I'm just raising this in case it was an oversight and you'd like to add it.

kailuowang:
Yeah, that's an oversight. I will add it.

dotta:
A quick note to inform you that in case you are considering adopting sbt 1.3.0-rc1, we have found an incompatibility with our sbt-hydra plugin, preventing the two to work together. The good news is that we have a fix ready, and a new release will likely be announced on Monday. I'll comment again here when the release is out so that you can upgrade the sbt-hydra version.

dotta:
As promised, here is a quick update about sbt 1.3.0-RC1. There is a regression in sbt affecting the sbt-hydra plugin (the hydra-bridge sources JAR aren't downloaded because a resolver is missing). We opened a ticket to track this problem (https://github.com/sbt/sbt/issues/4712).

Unless you wanted to upgrade to sbt 1.3.0-RC1 right away, this PR should not be blocked the linked issue.

kailuowang:
turns out that finiteDuration as a name of a package is not scalastyle compliant. I am going to resolve that in a different PR.

travisbrown:
I wish this change had been put forward for more public discussion before being merged. I personally didn't hear anything about this before seeing all the tweets about it after it was done.

I think there are many good reasons to avoid requiring a closed-source sbt plugin to run the Cats build, and that's even if it had been tested thoroughly and we had established that it wasn't likely to cause problems. As it is right now running `sbt prePR` from a clean checkout crashes with [this error](https://github.com/typelevel/cats/issues/2870) on both my Mac laptop and Linux desktop, and it's completely impossible to investigate because sbt-hydra is just a binary from a commercial repo.

I appreciate Triplequote's generosity, but the Cats build already has plenty of sharp edges, especially for new contributors, and I really don't think throwing a WIP closed-source sbt plugin into the mix was a good idea.

Issue #2751, opened 2019-03-18
Add FlatMapRequested on Validated to alleycats like scalaz's Validation for convenience


psilospore:
Although scalaz's Validation and cat's Validated are not monads. Scalaz provides a convenient import for providing flatmap on Validation to use on for comprehensions. 

```
  /** Import this if you wish to use `flatMap` */
  object FlatMap {
    @inline implicit def ValidationFlatMapRequested[E, A](d: Validation[E, A]): ValidationFlatMap[E, A] =
      new ValidationFlatMap(d)
  }
```

In our code base we simply import this right before we want to sequence our Validations. Example

```
def logNumOfPosts() = {
  import scalaz.Validation.FlatMap._
  for {
    user <- fetchUser(userId)
    posts <- fetchPostsByUser(user) //This could potentially go on much longer
  } {
    log.info(s"${user.name} has posted ${posts.size} times")
  }
}
```

Although I could use andThen to sequence I miss out on the nice for-comprehension syntax.

psilospore:
I would be interested in working on this if it gets a 👍

chuwy:
I think if this is going to be implemented it should be in alleycats. I personally prefer `Either` + `Parallel` in most cases to solve this problem.

PR #2851, opened 2019-05-18
Add FlatMapRequested on Validated to alleycats like scalaz's Validation for convenience

psilospore:
Related issue https://github.com/typelevel/cats/issues/2751

The contribution guide said I should put tests in cats.tests but I couldn't access alleycats from there. So I'm not sure what to do about tests.

### DIFF
--- a/alleycats-core/src/main/scala/alleycats/data/validated.scala
+++ b/alleycats-core/src/main/scala/alleycats/data/validated.scala
@@ -1,0 +1,23 @@
+package alleycats.data
+
+import cats.data.Validated
+import cats.data.Validated.Invalid
+
+object validated {
+
+  /*
+   * The contents of FlatMap is inspired by and derived from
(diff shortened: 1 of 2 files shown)


rossabaker:
Thanks for this.  I'm not entirely clear on the motivation for doing this vs. just using an `Either` in the simple case.  As suggested in #2751, there is a [more complex example that uses `Parallel`](https://typelevel.org/cats/typeclasses/parallel.html#parallel).  Is there a fuller example of why these existing techniques aren't sufficient?

psilospore:
Sorry for the late reply. I've been meaning to test it out a bit and see if I could see any disadvantages. Based on what I've read It seems that in my case I don't see a reason to use `Validated` in cats when `Either` + `Parallel` would able to allow me to accumulate and sequence. I'll close the PR for now then.

PR #2854, opened 2019-05-21
reduce constraint of foldMapM from Monad to Applicative

### LINKED ISSUE #2852: Reduce constraint for foldMapM
Currently 

```scala
def foldMapM[G[_]: Monad, A, B: Monoid](fa: F[A])(f: A => G[B]): G[B]
```

but as far as I can see, we can drop the Monad constraint to Applicative

```scala
def foldMapM[G[_]: Applicative, A, B: Monoid](fa: F[A])(f: A => G[B]): G[B]
```

For `Traverse`, it is easy to see why it is possible:
```scala
def foldMapM[G[_]: Applicative, A, B: Monoid](fa: F[A])(f: A => G[B]): G[B] =
  fa.traverse(f).map(fold)
```

for `Foldable`, I still think it is stack safe because it is a `foldLeft` with `map2`

julien-truffaut:
fix #2852

I am not sure if it is binary compatible since Applicative extends Monad and it is in a Contravariant position.

### DIFF
--- a/core/src/main/scala/cats/Foldable.scala
+++ b/core/src/main/scala/cats/Foldable.scala
@@ -302,7 +302,7 @@ import Foldable.sentinel
    * res1: Option[Int] = None
    * }}}
    */
-  def foldMapM[G[_], A, B](fa: F[A])(f: A => G[B])(implicit G: Monad[G], B: Monoid[B]): G[B] =
+  def foldMapM[G[_], A, B](fa: F[A])(f: A => G[B])(implicit G: Applicative[G], B: Monoid[B]): G[B] =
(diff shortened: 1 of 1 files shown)


kailuowang:
This is not BC, once we drop Scala 2.11 on master, we can add a foldMapA for this if it helps.

julien-truffaut:
@kailuowang yeah that's what I expected but I wasn't sure. Should I revert my last commit and add `foldMapA`? We can merge this whenever 2.11 is dropped

kailuowang:
Yeah, let's do that. It will be a bit more work though since you need to add tests.

LukaJCB:
Shall we add `foldA` at that point as well? :)

julien-truffaut:
I don't have the bandwidth to finish the work, closing for now

PR #2857, opened 2019-05-24
Updates for 2.13.0-RC2

travisbrown:
Depends on a new Discipline release with [this change](https://github.com/typelevel/discipline/pull/98) and Machinist but otherwise I think this is ready to go.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ jdk:
 
 scala_version_211: &scala_version_211 2.11.12
 scala_version_212: &scala_version_212 2.12.7
-scala_version_213: &scala_version_213 2.13.0-RC1
+scala_version_213: &scala_version_213 2.13.0-RC2
 
 before_install:
  - export PATH=${PATH}:./vendor/bundle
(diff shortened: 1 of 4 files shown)


kailuowang:
thanks to #2858 by @dotta, merging in master should probably fix the build here

travisbrown:
We’ll still need Machinist and Discipline releases but thanks!

travisbrown:
Just confirmed that `+buildJVM` is fine here with the current Discipline master locally published as 0.12.0 (thanks for the Machinist release, @larsrh).

larsrh:
I've fixed the Discipline build number. Now, waiting for tut.

larsrh:
tpolecat/tut#250

travisbrown:
Thanks @larsrh! I’m going to be away from the computer until next Wednesday so anyone should feel free to grab these commits or whatever.

larsrh:
Turns out we don't necessarily need `tut` because CI runs it only on 2.12.

kailuowang:
I am out of town this weekend. If no one beats me, I can cut a cats 2.0m2 release Tuesday

travisbrown:
Excellent, thanks all!

rossabaker:
I'm around tonight.  Is [releasing](https://github.com/typelevel/cats/blob/master/PROCESS.md#releasing) still the state of the art?

PR #2858, opened 2019-05-24
Don't create the hydra.license if $HydraLicense is empty

dotta:
The sbt-hydra checks if a hydra.license file exists to decide whether it should enable
itself. Therefore, if the $HydraLicense is empty, it's important not to
create the hydra.license file.

The reason why the $HydraLicense can be empty is because its content is a
secret, hence it is available only to PRs created from within the repository.
Otherwise, the secret would be easily leaked.

This commit is a followup fix for 8069e09

---

I'm afraid the PR #2857 needs to include this fix to be validated. Sorry!

Please, don't hesitate to pull me and @dragos in if you hit any hurdle with Hydra.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,9 @@ scala_version_213: &scala_version_213 2.13.0-RC1
 
 before_install:
  - export PATH=${PATH}:./vendor/bundle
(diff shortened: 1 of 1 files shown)


kailuowang:
@dotta thanks for finding the fix. 
I am going to merge with one sign off since it's simple build change but also urgent.

PR #2859, opened 2019-05-25
v2.0.0-M2 updates

rossabaker:

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -16,6 +16,7 @@ possible:
  * Aaron Levin
  * Adam Fisher
  * Adelbert Chang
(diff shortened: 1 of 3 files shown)


kailuowang:
I noticed that some authors are off. I am going to use [this script](https://github.com/typelevel/cats/blob/master/scripts/releaseNotes.scala) to generate a new release note to update this one when I get a chance

rossabaker:
Oh, I didn't know we had a script.  We should copy over the result to clean up the GitHub release as well.  It's all duplicated.

rossabaker:
Obsoleted by #2873.

PR #2796, opened 2019-04-17
Added collectFirst to Chain and NonEmptyChain

LMnet:
Fixed #2795

PR #2780, opened 2019-04-09
added `foldF` to `EitherT`

kailuowang:
very simple short cuts, kind of a counterpart for `getOrElseF`

PR #2760, opened 2019-03-26
Make MonoidK[Endo] stack safe on combineK

denisrosca:
Closes #2738

PR #2861, opened 2019-05-26
Backports for a cats-1.6.x patch

rossabaker:
Some of the features in v2.0.0-M1 and v2.0.0-M2 are suitable for backporting:

For v1.6.x:
* #2810
* #2842
* #2784
* #2728
* #2837 (arguably?)

Alternatively or additionally, we could do a 1.7.x, and get several more while we await 2.0.0:
* #2780 
* #2775 
* #2760 
* #2759 
* #2750 
* #2747 
* #2744 
* #2726 
* #2724 
* #2709 
* #2698 
* #2696 
* #2796 
* #2840 
* #2837, if it's not already on a 1.6.1

### DIFF
--- a/core/src/main/scala/cats/Apply.scala
+++ b/core/src/main/scala/cats/Apply.scala
@@ -72,7 +72,7 @@ trait Apply[F[_]] extends Functor[F] with InvariantSemigroupal[F] with ApplyArit
    *
    */
   def productR[A, B](fa: F[A])(fb: F[B]): F[B] =
-    map2(fa, fb)((_, b) => b)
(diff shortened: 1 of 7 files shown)


rossabaker:
Any thoughts on going all the way to a 1.7 vs this more modest approach?

LukaJCB:
I think the more modest approach makes sense, I'd rather focus on getting 2.0 out the door, e.g. by pulling out `cats-testkit`

rossabaker:
Okay.  I should be able to release this tonight.

PR #2863, opened 2019-05-30
add attemptNarrow to ApplicativeErrorOps

marcodippy:
Hello, I already copy-pasted this function in a few places in my codebases, so I thought it could be useful to have it in cats.

Example use case (assuming `F : ApplicativeError[F, Throwable]`):
```scala
def authenticate(r: Request[F]): F[Unit] = ???

val authentication: Kleisli[F, Request[F], Either[AuthenticationException, Unit]] =
  Kleisli { request => authenticate(request).attemptCase[AuthenticationException] }
```

### DIFF
--- a/core/src/main/scala/cats/syntax/applicativeError.scala
+++ b/core/src/main/scala/cats/syntax/applicativeError.scala
@@ -86,7 +86,7 @@ final class ApplicativeErrorOps[F[_], E, A](private val fa: F[A]) extends AnyVal
     F.attempt(fa)
 
   def attemptNarrow[EE](implicit F: ApplicativeError[F, E], tag: ClassTag[EE], ev: EE <:< E): F[Either[EE, A]] =
-    F.recover(F.map(fa)(Right[EE, A](_): Either[EE, A])) { case e: EE => Left[EE, A](e) }
+    F.attemptNarrow[EE, A](fa)
 
   def attemptT(implicit F: ApplicativeError[F, E]): EitherT[F, E, A] =
     F.attemptT(fa)
--- a/core/src/main/scala/cats/syntax/applicativeError.scala
+++ b/core/src/main/scala/cats/syntax/applicativeError.scala
@@ -85,7 +85,7 @@ final class ApplicativeErrorOps[F[_], E, A](private val fa: F[A]) extends AnyVal
   def attempt(implicit F: ApplicativeError[F, E]): F[Either[E, A]] =
     F.attempt(fa)
 
-  def attemptCase[EE](implicit F: ApplicativeError[F, E], tag: ClassTag[EE], ev: EE <:< E): F[Either[EE, A]] =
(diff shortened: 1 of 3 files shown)


core/src/main/scala/cats/syntax/applicativeError.scala
```diff
@@ -83,6 +86,9 @@ final class ApplicativeErrorOps[F[_], E, A](private val fa: F[A]) extends AnyVal
   def attempt(implicit F: ApplicativeError[F, E]): F[Either[E, A]] =
     F.attempt(fa)
 
+  def attemptCase[EE](implicit F: ApplicativeError[F, E], tag: ClassTag[EE], ev: EE <:< E): F[Either[EE, A]] =
+    F.recover(F.map(fa)(_.asRight[EE])) { case e: EE => e.asLeft[A] }
```
diesalbla: Given that extra inter-dependencies could lead to circular dependencies, perhaps we can by-pass the `either` syntax and use this not much longer alternative: 
  ```suggestion
      F.recover(F.map(fa)(x => Right[EE, A](x))) { case e: EE => Left[EE, A](e) }
```

**Comment by kailuowang** [comment]
I think this could be useful although I am not quite sure about the name. I am thinking `attemptNarrow` or `narrowAttempt`?

**Comment by marcodippy** [comment]
@kailuowang renamed to `attemptNarrow` 👍

File: core/src/main/scala/cats/syntax/applicativeError.scala [hunk 2]
```diff
@@ -83,6 +85,9 @@ final class ApplicativeErrorOps[F[_], E, A](private val fa: F[A]) extends AnyVal
   def attempt(implicit F: ApplicativeError[F, E]): F[Either[E, A]] =
     F.attempt(fa)
 
+  def attemptNarrow[EE](implicit F: ApplicativeError[F, E], tag: ClassTag[EE], ev: EE <:< E): F[Either[EE, A]] =
```
tkroman: why do we need a classtag here?

marcodippy: `EE` would be erased and we wouldn't be able to pattern match on it otherwise @tkroman

tpolecat: It's unsound without `ev` due to erasure, but `ev` is unused in the body so we might want to add a comment to this effect.

marcodippy:
@tpolecat I added the test, let me know what you think

marcodippy:
hey @kailuowang, is there anything I can do to get this merged?

kailuowang:
Hi @marcodippy sorry about the delay here, but can I ask one more thing? Now that we dropped 2.11 from master branch, we can add this method to the typeclass itself and have the syntax delegate to it. Would you  make that change?

marcodippy:
no problem @kailuowang, I just pushed the change :)

Issue #2870, opened 2019-06-01
Some Hydra sbt plugin error


travisbrown:
I have no idea what this is about but it's annoying. It's happening on current master on my machine.
```
$ sbt prePR
[info] Loading settings for project global-plugins-build from sbt-updates.sbt ...
[info] Loading global plugins from /home/travis/.sbt/1.0/plugins/project
[info] Loading settings for project global-plugins from sbt-updates.sbt ...
[info] Loading global plugins from /home/travis/.sbt/1.0/plugins
Waiting for lock on /home/travis/.ivy2/.sbt.ivy.lock to be available...
[info] Loading settings for project cats-build-build from plugins.sbt ...
[info] Loading project definition from /home/travis/code/projects/cats/project/project
[info] Loading settings for project cats-build from build.sbt,plugins.sbt ...
[info] Loading project definition from /home/travis/code/projects/cats/project
[info] Loading settings for project cats from build.sbt,version.sbt ...
[info] Resolving key references (23201 settings) ...
[info] Set current project to cats (in build file:/home/travis/code/projects/cats/)
[info] sbt started with 3818 MB memory heap
[info] Checking for Hydra latest version at https://repo.triplequote.com/artifactory/libs-release/...
[info] You are running the latest Hydra release.
[info] The Triplequote Hydra Scala compiler won't be used because the /home/travis/.triplequote/hydra.license is missing. Instead, compilation is carried out as usual with the vanilla Scala compiler.
[success] Total time: 17 s, completed Jun 1, 2019 7:56:45 AM
[success] Total time: 7 s, completed Jun 1, 2019 7:56:52 AM
[success] Total time: 1 s, completed Jun 1, 2019 7:56:53 AM
[info] Setting Scala version to 2.11.12 on 29 projects.
[info] Excluded 1 projects, run ++ 2.11.12 -v for more details.
[info] Reapplying settings...
[info] Set current project to cats (in build file:/home/travis/code/projects/cats/)
[info] sbt started with 3818 MB memory heap
[info] Checking for Hydra latest version at https://repo.triplequote.com/artifactory/libs-release/...
[info] You are running the latest Hydra release.
[info] The Triplequote Hydra Scala compiler won't be used because the /home/travis/.triplequote/hydra.license is missing. Instead, compilation is carried out as usual with the vanilla Scala compiler.
[error] java.util.NoSuchElementException
[error]         at scala.collection.LinearSeqOptimized.last(LinearSeqOptimized.scala:146)
[...]

PR #2871, opened 2019-06-01
Remove sbt-hydra for now

travisbrown:
See #2870 and my comment on #2848:
> I wish this change had been put forward for more public discussion before being merged. I personally didn't hear anything about this before seeing all the tweets about it after it was done.

> I think there are many good reasons to avoid requiring a closed-source sbt plugin to run the Cats build, and that's even if it had been tested thoroughly and we had established that it wasn't likely to cause problems. As it is right now running `sbt prePR` from a clean checkout crashes with [this error](https://github.com/typelevel/cats/issues/2870) on both my Mac laptop and Linux desktop, and it's completely impossible to investigate because sbt-hydra is just a binary from a commercial repo.

> I appreciate Triplequote's generosity, but the Cats build already has plenty of sharp edges, especially for new contributors, and I really don't think throwing a WIP closed-source sbt plugin into the mix was a good idea.

If we're going to use a closed-source sbt plugin in the Cats build, I think there's a pretty clear case for it being gated behind an environmental variable (Guillaume Martres pointed out [this example](https://github.com/scala/scala/blob/3274effb944f58cc6ac6cd99bcfd51abd6eab7a1/project/plugins.sbt#L39-L42) for sbt-dotty). This PR doesn't do that—it just disables the plugin for now so that when we tell people to run `sbt prePR` they don't get an opaque error message.

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -19,9 +19,11 @@ addSbtPlugin("org.scala-native" % "sbt-scala-native" % "0.3.9")
 addSbtPlugin("com.geirsson" % "sbt-scalafmt" % "1.5.1")
 addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.9.0")
 
+/* Temporarily disabling sbt-hydra, see #2870.
 resolvers += Resolver.url(
   "Triplequote Plugins Releases",
   url("https://repo.triplequote.com/artifactory/sbt-plugins-release/")
 )(Resolver.ivyStylePatterns)
 
 addSbtPlugin("com.triplequote" % "sbt-hydra" % "2.1.5")
+ */


kailuowang:
I totally agree that we shouldn't force contributors to use hydra. 
Hydra is only used on cats build on main repo branches (i.e, it is not used in PR builds or when run locally)
sbt-hydra 2.1.4 was doing just that. I think all we need to do is to revert back to sbt-hydra 2.1.4

travisbrown:
@kailuowang The objection isn't just to Hydra being required, it's sbt-hydra, which is also not open source. If it's behind a gate and not the default, I don't personally have a problem with it being in the build, but I don't think it should be required.

Issue #2561, opened 2018-10-16
Somewhat annoying to get at rethrow syntax due to invariance


djspiewak:
`MonadError` is invariant in `E`, which is problematic in turn because it means that `catsSyntaxMonadErrorRethrow` will not infer on `Either[E, A]` where `E <: Throwable`, which is *vastly* far and away the most common case. For example, if you have an expression that produces `F[Either[RuntimeException, A]]`, you might reasonably expect that `rethrow` would be available syntax on such a value (at least, I did). It won't work though, and it's quite difficult to tease out why this is the case.

We can either make `MonadError` contravariant in `E` (which I believe would cause a cascade of problematic changes, so I don't recommend it), or we can loosen the constraints on `catsSyntaxMonadErrorRethrow`. Something like:

```scala
implicit final def catsSyntaxMonadErrorRethrow[F[_], E2, E1 <: E2, A](fea: F[Either[E1, A]])(implicit F: MonadError[F, E2]): MonadErrorRethrowOps[F, E2, A]
```

Or the latter, which is likely to infer slightly better:

```scala
implicit final def catsSyntaxMonadErrorRethrow[F[_], E1, E2, A](fea: F[Either[E1, A]])(implicit F: MonadError[F, E2], ev: E1 <~< E2): MonadErrorRethrowOps[F, E2, A]
```

LukaJCB:
Closed by #2880

PR #2880, opened 2019-06-05
Improve MonadError rethrow syntax to be more flexible

bpholt:
This allows calling `MonadError`'s `rethrow` directly on an `F[Either[E, A]]`, where `E <: Throwable`. Currently I don't think this is possible: one has to explicitly `leftWiden` to `Throwable`.

### DIFF
--- a/core/src/main/scala/cats/MonadError.scala
+++ b/core/src/main/scala/cats/MonadError.scala
@@ -63,7 +63,7 @@ trait MonadError[F[_], E] extends ApplicativeError[F, E] with Monad[F] {
    * res1: scala.util.Try[Int] = Success(1)
    * }}}
    */
-  def rethrow[A](fa: F[Either[E, A]]): F[A] =
+  def rethrow[A, EE <: E](fa: F[Either[EE, A]]): F[A] =
     flatMap(fa)(_.fold(raiseError, pure))
 }
 
--- a/core/src/main/scala/cats/syntax/monadError.scala
+++ b/core/src/main/scala/cats/syntax/monadError.scala
@@ -7,7 +7,7 @@ trait MonadErrorSyntax {
 
   implicit final def catsSyntaxMonadErrorRethrow[F[_], E, A](
     fea: F[Either[E, A]]
-  )(implicit F: MonadError[F, E]): MonadErrorRethrowOps[F, E, A] =
+  )(implicit F: MonadError[F, _ >: E]): MonadErrorRethrowOps[F, E, A] =
     new MonadErrorRethrowOps(fea)
 }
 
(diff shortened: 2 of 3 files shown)


LukaJCB:
Hi @bpholt thanks for creating this PR, is there a reason you're specializing it to `Throwable`? I think we should be able to provide this for any type that's a supertype of the `E` . in question :)

kailuowang:
agree with @LukaJCB 
I haven't tried myself, but maybe changing the existing rethrow syntax in  https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/syntax/monadError.scala#L34-L36
```scala
final class MonadErrorRethrowOps[F[_], E, A](private val fea: F[Either[E, A]]) extends AnyVal {
  def rethrow(implicit F: MonadError[F, E]): F[A] = F.rethrow(fea)
}
```
to
```scala
  def rethrow(implicit F: MonadError[F, _ :> E]): F[A] = F.rethrow(fea)
```
will achieve what you want? 
and that would be binary compatible change.

bpholt:
@LukaJCB Widening beyond `Throwable` didn't occur to me. I don't have any principled objection to it if we can get it working.

@kailuowang That doesn't look like it compiles (I assume you meant `_ >: E`, not `_ :> E`):

```
[error] ~/cats/core/src/main/scala/cats/syntax/monadError.scala:35:68: type mismatch;
[error]  found   : MonadErrorRethrowOps.this.fea.type (with underlying type F[Either[E,A]])
[error]  required: F[Either[_$1,A]]
[error] Note: Either[E,A] <: Either[_$1,A], but type F is invariant in type _.
[error] You may wish to define _ as +_ instead. (SLS 4.5)
[error]   def rethrow(implicit F: MonadError[F, _ >: E]): F[A] = F.rethrow(fea)
[error]                                                                    ^
[error] one error found
```

I tried implementing this as a low-priority implicit (to avoid ambiguity with the existing one, ignoring bincompat for now) but I can't seem to get the type inference to work. 

```scala
implicit final def catsSyntaxMonadErrorWidenRethrow[F[_], E, EE >: E, A](
  fea: F[Either[E, A]]
)(implicit F: MonadError[F, EE]): MonadErrorWidenRethrowOps[F, E, EE, A] =
  new MonadErrorWidenRethrowOps(fea)
```

If I specify at the call site the super type `EE` for which the `MonadError` exists, it compiles. If anyone has ideas on how to help the inference, I'd appreciate them.

kailuowang:
@bpholt thanks for trying this out. 
It looks like we encountered a scala compiler bug - I filed it [here](https://github.com/scala/bug/issues/11562)

As a work around I suggested that we make two changes (both Binary compatible), I tested them it should work
Change 1,  in `MonadError`
We change `rethrow` to allow narrower `E`
```Scala
  def rethrow[A, EE <: E](fa: F[Either[EE, A]]): F[A] =
    flatMap(fa)(_.fold(raiseError, pure))
```
Change 2, in MonadError syntax we change the exising `rethrow` to
```scala
def rethrow(implicit F: MonadError[F, _ >: E]): F[A] =
    F.flatMap(fea)(_.fold(F.raiseError, F.pure))  //it's a dup from the type class implemenation, due to https://github.com/scala/bug/issues/11562. Once fixed should be able to replace with `F.rethrow(fea)`
```

kailuowang:
fixes #2561 cc @djspiewak

PR #2886, opened 2019-06-12
Add extension syntax for instantiating EitherT/OptionT from F[Either]/F[Option]

bplommer:


jhnsmth:
Are there any chances to get this merged ? I'm happy to help with updating this PR and resolving conflicts

rossabaker:
Sorry that this one escaped notice.

What is the advantage to this over the apply methods that already exist?  Does it improve type inference in some way?

bplommer:
> What is the advantage to this over the apply methods that already exist? Does it improve type inference in some way?

No, it's purely a convenience method to support method chaining.

bplommer:
I didn't do a great job of selling this back in... blimey, the end of the before times.

Finding myself once again working in some `EitherT`-heavy code bases though, I still think it's a good idea. Any thoughts?

bplommer:
I've messed something up with updating the bincompat mixins - will fix if the idea is approved.

rossabaker:
I wondered whether mouse had something like this, and as of 10 days ago, [it does](https://github.com/typelevel/mouse/pull/260).  I think that's the right place for it.

armanbilge:
Yep, this seems to have found a good home in mouse.

PR #2889, opened 2019-06-14
Add missing factories similar to existing factories

### LINKED ISSUE #2887: Should factories like Semigroup.instance be provided for other typeclasses?
While `Semigroup` has [`cats.kernel.Semigroup#instance`](https://github.com/typelevel/cats/blob/774fb51d1365e1adef7fa71f09b1410941264f60/kernel/src/main/scala/cats/kernel/Semigroup.scala#L120), `Monoid` has no equivalent, which could be implemented as:
```
object Monoid {
  /**
   * Create a `Monoid` instance from the given function and constant
   */
  @inline def instance[A](emptyVal: A, cmb: (A, A) => A): Monoid[A] = new Monoid[A] {
    override val empty: A = emptyVal
    override def combine(x: A, y: A): A = cmb(x,y)
}
``` 

Looking at the various typeclasses in `cats.kernel`, there doesn't seem to be a pattern for which ones have factories, and which don't:

| Typeclass              | Has Factory | Commentary                                                           |
|------------------------|-------------|----------------------------------------------------------------------|
| `Band`                 | No          | Should be identical to `Semigroup.instance`                            |
| `BoundedSemilattice`   | No          | Should be identical to `Monoid.instance`                               |
| `CommutativeGroup`     | No          | See note for parent type `Group`                                       |
| `CommutativeMonoid`    | No          | Should be identical to `Monoid.instance`                               |
| `CommutativeSemigroup` | No          | Should be identical to `Semigroup.instance`                            |
| `Eq`                   | Yes         |                                                                      |
| `Group`                | No          | `Group.instance(A, A => A, (A, A) => A)` might be a bit much           |
| `Hash`                 | Yes         |                                                                      |
| `Monoid`               | No          | `Monoid.instance(A, (A, A) => A)` should be easy to understand |
| `Order`                | Yes         |                                                                      |
| `PartialOrder`         | Yes         |                                                                      |
| `Semigroup`            | Yes         |                                                                      |
| `Semilattice`          | No          | Should be identical to `Semigroup.instance`                            |

Would a PR adding these be useful?

It doesn't look like factories for the typeclasses in `cats.core` would be possible, as they'd require accepting higher-order lambdas, which I don't think can be encoded as parameters in Scala.

morgen-peschke:
Fixes #2887

### DIFF
--- a/kernel/src/main/scala/cats/kernel/Band.scala
+++ b/kernel/src/main/scala/cats/kernel/Band.scala
@@ -14,4 +14,11 @@ object Band extends SemigroupFunctions[Band] {
    * Access an implicit `Band[A]`.
    */
   @inline final def apply[@sp(Int, Long, Float, Double) A](implicit ev: Band[A]): Band[A] = ev
+
+  /**
+   * Create a `Band` instance from the given function.
+   */
+  @inline def instance[A](cmb: (A, A) => A): Band[A] = new Band[A] {
+    override def combine(x: A, y: A): A = cmb(x, y)
+  }
 }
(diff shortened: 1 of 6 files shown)


morgen-peschke:
@kailuowang, is there anything I can do to help get this merged?

morgen-peschke:
Thanks @kailuowang :)

Issue #2893, opened 2019-06-18
Fix all compiler warnings on Scala 2.13


kailuowang:
Fatal-warning was temporarily disabled on Scala 2.13. Removing them might require breaking changes, hence we'd better do it before 2.0-RC1

poslegm:
Should cats 2.0 for Scala 2.13 contains instances for `Stream`? It's deprecated and replaced by `LazyList`.

kailuowang:
@poslegm thanks for looking at this. I would say no, cats 2.0 code base should not have any deprecation warnings. Thus we need to move all the `Stream` instances to the scala version specific source folders `/core/src/main/scala-2.12-/cats/`. (and probably need a blank StreamInstances trait) in the 2.13+ folder For the first PR I don't think we have to add the LazyList instance yet, that can be a different PR. 

`Stream` is also used in laws/ExhausiveCheck, in which case maintaining cross build might be hard, I propose we replace that with a `List` which was discussed during the PR introducing `ExhausiveCheck`. We ended didn't switch to `List` at that time due to lack of benefit but now we do. That could be a separate PR if you want to control the scope of the one you are working on. 

Thanks!!!

kailuowang:
@poslegm I realized that we probably still want to keep Stream instances in 2.13 but deprecate them only on 2.13, this will require duplication the code in two scala version specific folders. I think I am going to take stab myself.

kailuowang:
Update `Stream` deprecations are resolved by https://github.com/typelevel/cats/pull/2904

PR #2895, opened 2019-06-18
Replace ⇒ with => to avoid Scala 2.13 warnings

travisbrown:
Related to #2893. I thought I'd done this for Cats but it must have been other projects.


travisbrown:
Oh, messed up the `sed` command and missed a few. Fixed in the second commit.

PR #2898, opened 2019-06-20
fix LazyList tests

johnynek:
Here are some suggestions to a branch related to LazyList in scala 2.13.

cc @kailuowang


core/src/main/scala/cats/instances/lazyList.scala
```diff
@@ -124,10 +137,13 @@ trait LazyListInstances extends cats.kernel.instances.LazyListInstances {
           val (s, b) = in
           if (s.isEmpty)
             G.pure(Right(b))
-          else
-            G.map(f(b, s.head)) { bnext =>
-              Left((s.tail, bnext))
+          else {
+            val h = s.head
```
johnynek: this isn't needed... sorry I left it around.

kailuowang:
Thanks so much! I am going to close this PR and simply continue work on your branch.

Issue #2903, opened 2019-06-20
On Scala 2.13 replace `NonEmptyStream` with `NonEmptyLazyList`


kailuowang:
`OneAnd`'s `tailRecM` implementation is no longer stack safe on `LazyList`. 
It's time to introduce a `NonEmptyLazyList` that is a new type to `LazyList` similar to other `NonEmptyXXXX` (e.g. ~NonEmptyVector~ NonEmptyChain) data types. 

Then maybe (definitely in a separate PR) consider deprecate the `OneAnd` data type on 2.13 (need to duplicate the code to both 2.12- folder and 2.13+ folder and mark the one in 2.13 as deprecated, this probably have chain implications (e.g. the type alias `NonEmptyStream` in cats.data need to be somehow only deprecated on 2.12-)

I completed the 1st step in #2905. If anyone interested in continue on it please feel free to continue on it.

Update: I was wrong `NonEmptyVector` didn't not use new type impl, changed to NonEmptyChain

kailuowang:
I completed the 1st step in #2905. If anyone interested in continue on it please feel free to continue on it.

nathaniel-may:
I've pulled your changes and I'm going to give it a try. This will be my first attempt at a cats issue 😬.

kailuowang:
Thanks @nathaniel-may, please don't hesitate to ask questions here or on cats-dev gitter chatroom. Also please merge in master, which has resolved some Scala 2.13 issues.

kailuowang:
closed by #2941

PR #2905, opened 2019-06-21
[WIP] new NonEmptyLazyList

kailuowang:
This PR only includes an initial step for #2903. which makes the existing tests for `NonEmptyStream` Scala 2.12- specific. 

The next step will be creating a new type based `NonEmptyLazyList` following the example of `NonEmptySet` and `NonEmptyChain` and then create a test suite for it.


kailuowang:
continued on https://github.com/typelevel/cats/pull/2930

PR #2907, opened 2019-06-22
Update sbt-scalafix to 0.9.5

fthomas:
This updates Scalafix to 0.6.0-M9 by applying [this diff](https://github.com/scalacenter/scalafix.g8/blob/c22ae75cabb629aa4a83a95d5fdd8f99691cd92b/scalafix/readme.md). I've checked locally that the Scalafix tests still pass.

My ultimate goal is to update Scalafix to 0.9.x so that this rule (and potential future rules) can be used by Scala Steward (https://github.com/fthomas/scala-steward/pull/611) but it is currently unclear to me if updating to 0.9.x means that this rule needs to rewritten with Scalafix' `v1` API. @olafurpg do `v0` rules also work with 0.9.x?

### DIFF
--- a/scalafix/build.sbt
+++ b/scalafix/build.sbt
@@ -1,4 +1,4 @@
-val V = _root_.scalafix.Versions
+val V = _root_.scalafix.sbt.BuildInfo
 
 inThisBuild(
   List(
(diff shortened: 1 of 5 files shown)


olafurpg:
In Scalafix v0.9 there is a `import scalafix.v0._` that keeps the old API and should at least make old rules compile with the latest Scalafix. However, all symbol matchers like
```
_root_.cats.syntax.CartesianOps.`*>`.
```
will need to be written against the new symbol format
```
cats/syntax/CartesianOps#`*>`(). (assuming *> is a method and CartesianOps is a class)
```

olafurpg:
Scalafix 0.6.x only supports outdated Scala versions so it's not much value today. The 0.9.x API will probably stay stable for a long time, it should probably be released as v1.0.

fthomas:
Rewriting the symbol matchers did the trick. Thanks, @olafurpg!

sbt-scalafix is now on 0.9.5.

fthomas:
And it works for Scala Steward: https://github.com/fthomas/scalafix-test/pull/9/files :-)

Issue #1536, opened 2017-02-12
Add Bounded typeclass


julien-truffaut:
```
trait Bounded[A] extends Order[A]{
  def minValue: A
  def maxValue: A
}
```
such as `min(a, minValue) == minValue` and `max(a, maxValue) == maxValue`

kevinmeredith:
Why must `Bounded` extend `Order`? 

I don't believe that Haskell's `Bounded` imposes such a constraint:

```haskell
iPrelude> :i Bounded
class Bounded a where
  minBound :: a
  maxBound :: a
```

julien-truffaut:
It is correct but in this case `Bounded` will be lawless isn't it?

kevinmeredith:
What laws must a `Bounded` instance obey?

reference - http://stackoverflow.com/questions/43029076/type-class-w-min/43029124?noredirect=1#comment73326210_43029124

julien-truffaut:
@kevinmeredith see my comment (such as ...)

edmundnoble:
This seems to not be `Bounded` any more when you add those laws, but `BoundedLattice` (https://github.com/typelevel/algebra/blob/master/core/src/main/scala/algebra/lattice/BoundedLattice.scala) seeing as `<` and `>` can be implemented using `meet`/`join` and `Eq`.

johnynek:
I'm not crazy about Bounded. Would prefer LowerBounded and UpperBounded. String, for instance is LowerBounded. NonNegative numbers are LowerBounded. There are some examples of only UpperBounded as well.

julien-truffaut:
I agree with @johnynek it would be more useful to one typeclass for upper and another for lower bound. Here is an example https://github.com/julien-truffaut/newts/blob/master/core/shared/src/main/scala/newts/Min.scala#L35

kailuowang:
For lawless typeclasses, we might also want to consider adding them to alleycats instead.

johnynek:
I don't see how his is lawless when it has an associated Ordering. So, they could extend Order and everything is fine, no?

kailuowang:
@johnynek I didn't make it clear. I was referring to LowerBounded and UpperBounded.

johnynek:
```scala 
trait LowerBounded[A] extends Order[A] {
  // law: compare(minValue, a) <= 0 for all a
  def minValue: A
}
```
I'm probably missing it, but how is that lawless? Maybe I'm missing context?

kailuowang:
[...]

PR #2910, opened 2019-06-24
Add LowerBounded and UpperBounded type classes

izeigerman:
Implements https://github.com/typelevel/cats/issues/1536


LukaJCB:
This is a great PR, thank you so much! I think we could add a bunch more instances, if you're up to it :) 
We could also do it in a separate PR if you'd prefer? Or someone else could take them.
I'm thinking of 
- Duration
- FiniteDuration
- various collections are lower bounded, e.g. `List`, `Set`, etc.
- `Either`, `Validated`, `Ior`, `Option`
- `Tuple2` 
- `Const`

LukaJCB:
I can also see another problem. What happens if someone summons both `Order` and `Bounded`, it seems we'd get a clash, probably, no?

```scala
def foo[A: Order: Bound](x: A, y: A) = x === y
```
The above will probably result in ambiguity.
We could go around this by using composition instead of inheritance:

```scala
trait LowerBound[A] {
  def lowerBound: A
  def partialOrder: PartialOrder[A]
}
```

izeigerman:
@LukaJCB Thanks for pointing out the ambiguity issue!

> I think we could add a bunch more instances, if you're up to it

Yes, I'm definitely up to it! I just wasn't quite sure about boundaries for some types that you enumerated. Can you please help me determine whether my understanding is correct:

- `Duration` - I assume only Lower Bound is applicable here and the value should be a duration with a 0 length. Upper Bound can't be specified since the duration can be infinite. 
- `FiniteDuration` - Lower Bound - 0-length duration, Upper Bound - `Long.MaxValue`.
- `List`, `Set`, etc. - Lower Bound - empty collection. Upper Bound is not applicable. 
- `Either`, `Validated`, `Ior`, `Option` - not sure about the semantics here. For example the Lower Bound for `Option` might be either `None` or the Lower Bound of the underlying type. Similar question about `Either`, `Validated` and `Ior` - what is the expected behavior here?
- `Tuple2` - Lower Bound - is the Lower Bound for both types that constitute a sum. Same for the Upper Bound.
- `Const` - Lower and Upper bounds of the underlying constant type.

LukaJCB:
That seems inline with what I had in mind, except for:

> Duration - I assume only Lower Bound is applicable here and the value should be a duration with a 0 length. Upper Bound can't be specified since the duration can be infinite.

There is `Duration.Inf` for the upper bound :)

Either, Validated, Ior, Option - not sure about the semantics here. For example the Lower Bound for Option might be either None or the Lower Bound of the underlying type. Similar question about Either, Validated and Ior - what is the expected behavior here?

We have to go with what the existing `Order` implementations assume, so for example for `Either`,
`Left(x).compare(Right(y))` will always be `-1`. So `Left` is considered to be less than `Right`.

Does that make it more clear? :)

izeigerman:
@LukaJCB Yes, makes perfect sense to me. Thanks!

LukaJCB:
Btw, I don't see instances in `kernel` for `Ior` and `Validated`, so you can probably leave those out for now :)

rossabaker:
> Duration - I assume only Lower Bound is applicable here and the value should be a duration with a 0 length. Upper Bound can't be specified since the duration can be infinite.

Durations can be negative.  See `Duration.MinusInf`.

izeigerman:
Thanks @rossabaker !
@LukaJCB I have two more clarifications to make:
1. I noticed that `List`, `Set`, etc may actually have an upper bound according to their `Order` implementations. Basically if the underlying type `A` has an `UpperBound` instance then for example for `List` the `UpperBound` will look like:
```
def maxBound: List[A] = List(UpperBound[A].maxBound)
```
This still satisfies the law, so unless there are no objections I'm going to introduce those as well.
2. Making a `partialOrder` a member of the `*Bounded` seems to have a purely indicative/enforcing purpose. It's not being used in any other way. And of course it makes creating instances for this type class less convenient. At the same time I realize that the direct hierarchy is not an option. I just figured this worth pointing out.

LukaJCB:
> Durations can be negative. See `Duration.MinusInf`.

TIL, that's kinda weird, though, what kind of duration is -4 minutes? :smile:

ritschwumm:
> > Durations can be negative. See `Duration.MinusInf`.
> 
> TIL, that's kinda weird, though, what kind of duration is -4 minutes? smile

imho a duration is just the difference between two points in time, which of course can be negative.

djspiewak:
> TIL, that's kinda weird, though, what kind of duration is -4 minutes? 😄

I mean, time is a directional vector, not a magnitude. I think `-4.minutes` is a pretty sane thing to talk about. :-) At the very very least, `6.minutes - 2.minutes` is most definitely a sane thing to talk about, but that expression doesn't really make algebraic sense without inverses, which implies negative durations and thus a directionality to time.

izeigerman:
@LukaJCB I think this PR is ready to be revisited once you get a chance. Thanks

kailuowang:
Thanks so much for this contribution. I think we should avoid adding so many new bincompat traits in the master branch now we are very close to drop support for 2.11 on master. I.e. I am proposing separating this into two PRs, 
1. a type class (with `Const` instance which can be added without bincompat trait) PR which can be merged right now, 
2. an instances PR that simply added methods to existing traits, we can merge it after the master branch drops 2.11 support

izeigerman:
@kailuowang I think the suggested approach sounds reasonable except one thing. I'm not sure I can just add an instance for `Const` without adding an instance for at least a single other type (eg `Int`), since I won't be able to test it properly. Do you think we should introduce instances for types that don't require bincompat traits (like `Int`, `Short`, etc.) as part of step #1?

kailuowang:
@izeigerman ah I didn't realize that there are instances you can add without bincompat traits. Yes please make them as part of step 1. Thanks!!!

izeigerman:
The PR that covers step 1 is https://github.com/typelevel/cats/pull/2913

PR #2911, opened 2019-06-25
fixed more warnings on 2.13

kailuowang:
Also made default Scala version 2.12, which helps for IDE integration. 

Update: I removed ScalaStyle which is conflicting with scalafmt, I see no reason to maintain both.


tests/src/test/scala/cats/tests/BinCodecInvariantMonoidalSuite.scala
```diff
@@ -8,7 +8,7 @@ import cats.implicits._
 import cats.Eq
 import cats.kernel.laws.discipline.{MonoidTests, SemigroupTests}
 import org.scalacheck.{Arbitrary, Gen}
-import kernel.compat.lazyList._
+import kernel.compat.scalaVersionSpecific._
```
milessabin: Is it really necessary to have an import like this in user code? Can't we hide the compat layer somehow?

kailuowang: The compat layer is hidden (package private to `cats`).  It's used here because the test referenced `Stream`, and is cross-compiled against 2.12 and 2.13, so the tests itself need the same cross scala compat machinery provided by the `scalaVersionSpecific` 
Users who are not cross compiling do not use any of these compat to use cats. But users who are might need their own compat layer if they are using a deprecated type from 2.12. E.g. if a cross-compiling user is using `Stream`, there is simply no `Stream` instances in Cats for 2.13, only `Lazy List` instances. The user will need the same compat layer (could be copied from Cats) to use `Stream` in 2.12 and `LazyList` in 2.13.

kernel/src/main/scala-2.13+/cats/kernel/compat/scalaVersionSpecific.scala
```diff
@@ -0,0 +1,13 @@
+package cats
+package kernel
+
+package compat
+
+private[cats] object scalaVersionSpecific {
+  type LazyList[+A] = scala.LazyList[A] //this is needed only to avoid unused import warnings on Scala 2.13
+  type IterableOnce[+A] = scala.IterableOnce[A] //this is needed only to avoid unused import warnings on Scala 2.13
```
djspiewak: Burn it all down.

PR #2914, opened 2019-06-26
Fixed all 2.13 warnings, reenabled fatalWarning on 2.13

### LINKED ISSUE #2893: Fix all compiler warnings on Scala 2.13

kailuowang:
fix #2893

### DIFF
--- a/.jvmopts
+++ b/.jvmopts
@@ -1,6 +1,6 @@
 -Dfile.encoding=UTF8
 -Xms1G
--Xmx4G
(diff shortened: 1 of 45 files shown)


kailuowang:
We might want more reviews. Merged for now to unblock folks at Scala matsuri contribution event.

PR #2915, opened 2019-06-27
Add OrderTests for SortedSet

jooohn:
I found we have PartialOrderTests for SortedSet, but not OrderTests.

### DIFF
--- a/tests/src/test/scala/cats/tests/SortedSetSuite.scala
+++ b/tests/src/test/scala/cats/tests/SortedSetSuite.scala
@@ -1,7 +1,7 @@
 package cats
 package tests
 
-import cats.kernel.laws.discipline.{BoundedSemilatticeTests, HashTests, PartialOrderTests}
+import cats.kernel.laws.discipline.{BoundedSemilatticeTests, HashTests, OrderTests, PartialOrderTests}
 import cats.kernel.{BoundedSemilattice, Semilattice}
 import cats.laws._
 import cats.laws.discipline.SemigroupalTests.Isomorphisms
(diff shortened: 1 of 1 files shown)


tests/src/test/scala/cats/tests/SortedSetSuite.scala
```diff
@@ -19,6 +19,8 @@ class SortedSetSuite extends CatsSuite {
   checkAll("Semigroupal[SortedSet]", SerializableTests.serializable(Semigroupal[SortedSet]))
 
   checkAll("SortedSet[Int]", FoldableTests[SortedSet].foldable[Int, Int])
+  checkAll("Order[SortedSet[Int]]", OrderTests[SortedSet[Int]].order)
+  checkAll("Order.reverse(Order[SortedSet[Int]])", OrderTests(Order.reverse(Order[SortedSet[Int]])).order)
   checkAll("PartialOrder[SortedSet[Int]]", PartialOrderTests[SortedSet[Int]].partialOrder)
```
larsrh: @kailuowang Do we still need these?

kailuowang: It is redundant, but we are not very consistent when it comes testing laws already covered by the type class laws lower in the hierarchy. That's why I didn't raise an objection here. There is benefit in redundant tests, also sometimes it's hard to track which tests are already covered and remove them. Reducing redundancy to minimum (i.e. always test only the laws lowest in the hierarchy) might have its own problem as well. So right now we are doing it rather arbitrarily. We should probably come up with a principle and add it to the guideline. I created an issue for it https://github.com/typelevel/cats/issues/2919

Issue #2553, opened 2018-10-05
Thinking Longterm, Dotty support means removal of all macros


LukaJCB:
We should start discussing the best way to handle migration to Dotty.
Fortunately we don't have a lot of macros in cats, the few that I can think of are Simulacrum/Machinist and the `FunctionK.lift` macro. We could manually expand the former (as suggested by @smarter), but I'm not sure how to handle the latter. Maybe someone else has other good ideas :)

smarter:
> FunctionK.lift macro

Seems like this only exists to work around the lack of polymorphic function types, so if it's really a blocker that could force me to prioritize getting them in sooner :). In fact, I'd like to add polymorphic SAM auto-adaptation too, so you could just write `val lifted: FunctionK[List, Option] = headOption`.

More generally, I'd welcome removing all macros from cats given its central place in the Scala ecosystem. A macro-less cats could probably easily become part of the dotty community build, which would help us make sure we don't break it, and would make it much easier to experiment with new features related to typeclasses or type-level programming in general in a real codebase.

LukaJCB:
I played around with it a bit, and the fact that you can insert a type doesn't seem to cause any breakage, all of this compiles fine:

```scala
import cats._
import cats.arrow.FunctionK

sealed trait Example[A]

def lift[F[_], G[_]]
  (f: (F[α] ⇒ G[α]) forSome { type α }): FunctionK[F, G] =
    new FunctionK[F, G] {
      def apply[A](fa: F[A]): G[A] = 
        f.asInstanceOf[F[A] => G[A]](fa)
    }
  
def sample[A](option: Option[A]): List[A] = option.toList
  
val lifted1: FunctionK[Option, List] = lift(sample[Int])
val lifted2: FunctionK[Option, List] = lift(sample[Any])
val lifted3: FunctionK[Option, List] = lift(sample[Example[String]])
val lifted4: FunctionK[Option, List] = lift(sample)
  
def foo[F[_], A](fa: F[A])(f: F ~> List): List[A] =
  f(fa)
  
val res1: List[String] = foo(Option("2"))(lifted1)
val res2: List[String] = foo(Option("2"))(lifted2)
val res3: List[String] = foo(Option("2"))(lifted3)
val res4: List[String] = foo(Option("2"))(lifted4)

val res5: List[Int] = foo(Option(2))(lifted1)
val res6: List[Int] = foo(Option(2))(lifted2)
val res7: List[Int] = foo(Option(2))(lifted3)
[...]

PR #2925, opened 2019-06-30
Replace machinist macros by plain forwarders

smarter:
Partial fix for #2553.

Co-authored-by: drdozer <turingatemyhamster@gmail.com>


smarter:
Forgot to actually remove it from the build.

PR #2926, opened 2019-06-30
Remove macro in FunctionK.lift implementation

smarter:
Instead, use the trick proposed by Alex in
https://github.com/typelevel/cats/issues/2553#issuecomment-493633214,
this is perhaps not ideal but it seems simpler than maintaining two
separate source files for Scala 2 and Dotty, just for this one function.
Additionally, since Dotty is able to call into non-macro methods in
Scala 2 dependencies, this makes FunctionK.lift usable from
Dotty without cats havin to publish Dotty artifacts.

Co-authored-by: Alexander Konovalov <alex.knvl@gmail.com>

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -398,7 +398,10 @@ def mimaSettings(moduleName: String) =
           exclude[IncompatibleMethTypeProblem]("cats.arrow.FunctionKMacros.lift"),
           exclude[MissingTypesProblem]("cats.arrow.FunctionKMacros$"),
           exclude[IncompatibleMethTypeProblem]("cats.arrow.FunctionKMacros#Lifter.this"),
-          exclude[IncompatibleResultTypeProblem]("cats.arrow.FunctionKMacros#Lifter.c")
+          exclude[IncompatibleResultTypeProblem]("cats.arrow.FunctionKMacros#Lifter.c"),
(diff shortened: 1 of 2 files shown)


smarter:
Looks like there's no way to avoid at least regressing on type inference with any of the proposed approach, so I think we'll have to go with split sources.

kailuowang:
Just curious is the type inference regression also happens on 2.13?

smarter:
Yup, same thing:
```scala
error] /home/smarter/opt/cats/tests/src/test/scala/cats/tests/FunctionKSuite.scala:88:46: no type parameters for method lift: (f: F[cats.arrow.FunctionK.τ[F,G]] => G[cats.arrow.FunctionK.τ[F,G]])cats.arrow.FunctionK[F,G] exist so that it can be applied to arguments (Option[Nothing] => List[Nothing])
[error]  --- because ---
[error] argument expression's type is not compatible with formal parameter type;
[error]  found   : Option[Nothing] => List[Nothing]
[error]  required: ?F[cats.arrow.FunctionK.τ[?F,?G]] => ?G[cats.arrow.FunctionK.τ[?F,?G]]
[error]     val fOptionToList = cats.arrow.FunctionK.lift(optionToList _)
[error]                                              ^
```

PR #2931, opened 2019-07-05
Add par functions for UnorderedTraverse.

takayahilton:
Added `parUnordered`.


mberndt123:
@LukaJCB,

Hey, I just implemented the same thing. Unfortunately I only noticed this PR after submitting my PR #2946. However my PR also includes parUnorderedFlatTraverse and parUnorderedFlatSequence, so if you think we need those, you can merge my PR instead. If you feel they're not needed, you can just close my PR.

kailuowang:
close since #2946 is merged.

PR #2941, opened 2019-07-12
Added  `NonEmptyLazyList` to replace `NonEmptyStream`

kailuowang:
The `NonEmptyStream` based on `OneAnd` cannot be simply modified to wrap LazyList since the `OneAnd` no longer provides stack safety for `LazyList`. Thus I switched to use the newtype based approach.
I also refactor `NonEmptyChain` (also newtype based) to reduce code duplication. 

This extracted from #2930 hence the long history, will squash merge as usual


tests/src/test/scala/cats/tests/OneAndSuite.scala
```diff
@@ -1,49 +1,22 @@
 package cats
 package tests
 
-import cats.kernel.laws.discipline.{EqTests, SemigroupTests}
```
kailuowang: `NonEmptyStream` related tests in this suite were moved to `NonEmptyStreamSuite` which is now 2.12- specific

core/src/main/scala-2.13+/cats/data/NonEmptyLazyList.scala
```diff
@@ -0,0 +1,378 @@
...
+import instances.lazyList._
+
+import scala.collection.immutable.TreeSet
+
+object NonEmptyLazyList extends NonEmptyLazyListInstances {
+  private[data] type Base
+  private[data] trait Tag extends Any
+  type Type[+A] <: Base with Tag
+
+  private[cats] def create[A](s: LazyList[A]): Type[A] =
```
travisbrown: Do you happen to know if these aren't `private[data]` just because of the tests? It seems like e.g. `NonEmptyStreamSuite` should be in `cats.data`, not `cats.tests`.

kailuowang: Right now all tests suites are under `cats.tests`, do we want to make it more conventional (i.e. parallel package tree as in src)?

travisbrown: I'd say yes, especially if it's the only reason these aren't more restricted. That's pretty far from the topic of this PR, though. I can take a look this weekend.

kailuowang: just found out that the tests don't use them. So it's probably a simple artifact of copy paste code. I just changed all of such methods to be private[data]

travisbrown: Nice, thanks!

core/src/main/scala-2.13+/cats/data/NonEmptyLazyList.scala
```diff
@@ -0,0 +1,378 @@
...
+import NonEmptyLazyList.create
+import kernel.PartialOrder
+import instances.lazyList._
+
+import scala.collection.immutable.TreeSet
+
+object NonEmptyLazyList extends NonEmptyLazyListInstances {
+  private[data] type Base
+  private[data] trait Tag extends Any
+  type Type[+A] <: Base with Tag
```
johnynek: I don’t understand how these are abstract on a concrete `object`. What is Type?

tests/src/test/scala-2.13+/cats/tests/NonEmptyLazyListSuite.scala
```diff
@@ -0,0 +1,141 @@
...
+    }
+  }
+}
+
+class ReducibleNonEmptyLazyListSuite extends ReducibleSuite[NonEmptyLazyList]("NonEmptyLazyList") {
+  def iterator[T](nel: NonEmptyLazyList[T]): Iterator[T] = nel.toLazyList.iterator
+
+  def range(start: Long, endInclusive: Long): NonEmptyLazyList[Long] = {
+    // if we inline this we get a bewildering implicit numeric widening
+    // error message in Scala 2.10
```
johnynek: Since we don’t use 2.10 anymore can we remove?

tests/src/test/scala/cats/tests/NonEmptyChainSuite.scala
```diff
@@ -140,3 +140,14 @@ class NonEmptyChainSuite extends CatsSuite {
     }
   }
 }
+
+class ReducibleNonEmptyChainSuite extends ReducibleSuite[NonEmptyChain]("NonEmptyChain") {
+  def iterator[T](nel: NonEmptyChain[T]): Iterator[T] = nel.toChain.iterator
+
+  def range(start: Long, endInclusive: Long): NonEmptyChain[Long] = {
+    // if we inline this we get a bewildering implicit numeric widening
+    // error message in Scala 2.10
```
johnynek: Can we remove?

core/src/main/scala-2.13+/cats/data/NonEmptyLazyList.scala
kailuowang: This is the newtype encoding (kind of a hack of the Scala type system) that eliminates all runtime overhead.

johnynek: I know the trick, I just didn't know objects could have totally abstract types. Is that intended, or some quirk of the compiler?

Seems to me, anything we an abstract type should have to be abstract.

Do we have this documented somewhere that we can link to? I think it is not at all clear to the reader what the various parts are here and why they are all required:

1. why do we need Base?
2. why do we need Tag?
3. why should the above be `private[data]` vs `private` or `public`?
4. how is the aliasing of this type to `NonEmptyLazyList` accomplished? Is that a package alias?

kailuowang: good point. Link added.

kailuowang: also cc @LukaJCB to see if he has anything to add to the simple description and the link

kailuowang:
@johnynek did my changes address the issues?

kailuowang:
@LukaJCB, do you want to review the new changes?

PR #2946, opened 2019-07-14
add unordered parallel variants of (flat)traverse and (flat)sequence

mberndt123:
Hi,

There currently doesn't seem to be a convenient way to traverse a `Set` with `IO`. `traverse` doesn't work because there's no `Traverse` instance for `Set`, and `unorderedTraverse` doesn't work because there is no `CommutativeApplicative` instance for `IO`. So I've added `parUnorderedTraverse` and `parUnorderedSequence` for that purpose.

I also added `parUnorderedFlatTraverse` and `parUnorderedFlatSequence` for symmetry, but given that the non-par versions of these don't exist I'm not really sure we need these.

Another logical set of operations would be `parUnorderedTraverse_`  and `parUnorderedSequence_`. But again, there is no `unorderedTraverse_` or `unorderedSequence_` in `UnorderedFoldable`, so it's not going to be any more efficient than the non-_ variants, and given that it's easy to just discard the return value I haven't added them.

So overall I think that `parUnorderedTraverse` and `parUnorderedSequence` should be pretty uncontroversial. Which of the other ones do you think we want? I think they all make sense for symmetry, but then I'm not sure how useful they are in practice.

Oh, and one more question: is `FlatMap` actually the right typeclass to use for the `flat` variants, or should it be `CommutativeFlatMap`?

### DIFF
--- a/core/src/main/scala/cats/Parallel.scala
+++ b/core/src/main/scala/cats/Parallel.scala
@@ -169,6 +169,26 @@ object Parallel extends ParallelArityFunctions2 {
     P.sequential(gtb)
   }
 
+  def parUnorderedTraverse[T[_]: UnorderedTraverse, M[_], F[_]: CommutativeApplicative, A, B](
+    ta: T[A]
+  )(f: A => M[B])(implicit P: Parallel[M, F]): M[T[B]] =
+    P.sequential(UnorderedTraverse[T].unorderedTraverse(ta)(a => P.parallel(f(a))))
+
+  def parUnorderedSequence[T[_]: UnorderedTraverse, M[_], F[_]: CommutativeApplicative, A](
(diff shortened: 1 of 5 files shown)


mberndt123:
OK, somebody beat me to it: https://github.com/typelevel/cats/pull/2931

But hey, you probably really want those `flat` variants, don't you? :laughing:

mberndt123:
Hey @LukaJCB,

I've added some tests now. Thanks for the idea, it actually helped me dicover some silly mistakes in the implementation.

mberndt123:
I've also removed `ParallelUnorderedFlatTraverseOps` and moved its only method to `ParallelUnorderedTraverseOps`. I don't think there's a reason to have separate classes for those two methods.

PR #2953, opened 2019-07-19
Add  method init and last to NonEmptyChain like NonEmptyList.

takayahilton:


LukaJCB:
This looks great, thanks a bunch!

PR #2962, opened 2019-07-25
Add bifold to Bifoldable close #2947.

### LINKED ISSUE #2947: Add bifold to Bifoldable
Something like this:

```scala
def bifold[F[_, _]: Bifoldable, A: Monoid, B: Monoid](fab: F[A, B]): (A, B) =
  bifoldMap(fab)(a => (a, Monoid[B].empty))(b => (Monoid[A].empty, b))
```

takayahilton:


kailuowang:
We can't break BC yet. also it would be nice to have some tests.

travisbrown:
I just noticed this PR, which was also done in #3088, which was just merged. I'm going to close this one, but thanks, @takayahilton, and I'm sorry we didn't see this one.

I'm not really sure the test here is needed, since it's just checking the default implementation, and I wouldn't really expect `Bifoldable` implementers to override this method.

PR #2963, opened 2019-07-28
Alternative implementation for 2.13 LazyList `tailRecM`

andyscott:
I took at stab at this and wound up with a solution that more heavily leverages existing functionality.


andyscott:
Closing, since I ported the changes over to master in #2964.

PR #2964, opened 2019-07-28
Implement LazyList tailRecM using Iterator.unfold

andyscott:
Implements `LazyList`'s `tailRecM` using `Iterator.unfold`.

### DIFF
--- a/core/src/main/scala-2.13+/cats/instances/lazyList.scala
+++ b/core/src/main/scala-2.13+/cats/instances/lazyList.scala
@@ -69,56 +69,14 @@ trait LazyListInstances extends cats.kernel.instances.StreamInstances {
         fa.zipWithIndex
 
       def tailRecM[A, B](a: A)(fn: A => LazyList[Either[A, B]]): LazyList[B] = {
-        val it: Iterator[B] = new Iterator[B] {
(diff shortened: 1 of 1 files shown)


core/src/main/scala-2.13+/cats/instances/lazyList.scala
```diff
@@ -69,56 +69,14 @@ trait LazyListInstances extends cats.kernel.instances.StreamInstances {
...
-            case Left(a) =>
-              // this is the first run
-              stack = fn(a).iterator :: Nil
-              advance()
-              next()
-            case Right(o) =>
-              val b = o.get
-              advance()
-              b
+        val kernel = Iterator.unfold[Option[B], Iterator[Either[A, B]]](Iterator(Left(a))) { it =>
```
johnynek: two small notes, I bet this would be faster if we did:
```scala
unfold[Option[B], List[Iterator[Either[A, B]]])(Iterator.single(Left(a)) :: Nil) {
  case Nil => None
  case nonempty@(h :: t) =>
    if (!h.hasNext) Some((None, t))
    else h.next match {
      case Left(a) => Some((None, fn(a) :: nonempty))
      case Right(b) => Some((Some(b), nonempty))
    } 
}
```

PR #2965, opened 2019-07-30
fix typo for `Data` in traverseFilter for OptionT

gagandeepkalra:
stumbled across this minor typo when I was looking for specific `traverseFilter` instance for `OptionT`

### DIFF
--- a/core/src/main/scala/cats/data/OptionT.scala
+++ b/core/src/main/scala/cats/data/OptionT.scala
@@ -248,7 +248,7 @@ sealed abstract private[data] class OptionTInstances extends OptionTInstances0 {
         OptionT(F.defer(fa.value))
     }
 
-  implicit def catsDateTraverseFilterForOptionT[F[_]](implicit F0: Traverse[F]): TraverseFilter[OptionT[F, *]] =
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/data/OptionT.scala
```diff
@@ -248,7 +248,28 @@ sealed abstract private[data] class OptionTInstances extends OptionTInstances0 {
...
+
+      override def filterA[G[_], A](
+        fa: OptionT[F, A]
+      )(f: A => G[Boolean])(implicit G: Applicative[G]): G[OptionT[F, A]] =
+        G.map(Traverse[F].traverse(fa.value)(TraverseFilter[Option].filterA[G, A](_)(f)))(OptionT[F, A])
+
+    }
+
+  @deprecated("renamed to catsDataTraverseFilterForOptionT", "2.0.0")
+  def catsDateTraverseFilterForOptionT[F[_]](implicit F0: Traverse[F]): TraverseFilter[OptionT[F, *]] =
```
rossabaker: I think this could call the corrected one instead of duplicating the implementation. The key is that it's deprecated and no longer the implicit.

PR #2967, opened 2019-07-31
Fix FunctorFilter mapFilter doc example

regadas:


LukaJCB:
Thank you @regadas :)

PR #2970, opened 2019-08-02
Decouple published modules from scalatest

rossabaker:
This isolates our ScalaTest dependency to tests and docs so we can publish cats-2.0.0 before scalatest-3.1.0 is released.

* `CatsSuite` and `CatsEquality`, which depend on ScalaTest, are moved from cats-testkit to cats-tests' tests.  This means we can continue to use them internally, but they are no longer available to external projects.

* The docs depend on discipline-scalatest for a full example. These examples assume a ScalaTest 3.1.0, which we hope will follow quickly after cats-2.0.0 is published.  We can update the docs if there is a breaking change before scalatest-3.1.0.

If there is external demand for `CatsSuite`, we could consider a separate `cats-scalatest` repo with its own release cycle.

cats-testkit is no longer directly mentioned in the docs, calling into question its value as a module.  That debate is for another PR, but is worth considering before we commit to cats-2.0.0.


docs/src/main/tut/typeclasses/lawtesting.md
```diff
@@ -74,72 +70,73 @@ but you won't need it if you import `org.scalacheck.ScalacheckShapeless._`).
...
 
-### Testing cats.kernel instances
+### Testing `cats.kernel` instances
 
 For most of the type classes included in Cats, the above will work great.
 However, the law tests for the type classes inside `cats.kernel` are located in `cats.kernel.laws.discipline.*` instead.
 So we have to import from there to test type classes like `Semigroup`, `Monoid`, `Group` or `Semilattice`.
 
-Let's test it out, by defining a `Semigroup` instance for our `Tree` type.
+Let's test it out by defining a `Semigroup` instance for our `Tree` type.
```
djspiewak: Random grammar fixes in the middle of an otherwise pure refactoring will never cease to bemuse and satisfy me.

ceedubs:
I have no strong attachment to `CatsSuite` (the `even` method is evidence that it was never really intended for external usage). But `CatsEquality` is pretty handy for anyone who is using cats and Scalatest. It's small but finicky enough that it seems like it shouldn't be copied into every repo who wants its functionality. It might be useful to expose in a scalatest support module?

Now that I'm typing this out I'm realizing that I should probably just finally make the effort to convert some of my Scalatest tests to use a more lightweight testing library. But lots of people are still going to use Scalatest, and it's not easy for big code bases to transition all at once.

rossabaker:
I wouldn't object to that being a microlibrary in a separate repo -- and we even could bring back a `CatsSuite` cleaned up for external use.  But publishing that here would again put us and everything downstream at the mercy of ScalaTest's release cycle.

ceedubs:
Ah sorry about that I didn't take the time to fully understand the context. I agree that it doesn't belong in this repo. And it is a small enough amount of code that I would probably copy and paste it as opposed to getting a new repo and build set up 😵

kailuowang:
:+1: for creating Cats-testkit-scalatest repo. Copy pasting code from two files + finding out what extra dependencies are needed is a bit too much for someone trying to figure out how to law testing instances for cats type classes.

ronanM:
It would be fine to have this `cats-testkit-scalatest` when Cats release version `2.0.0`.

PR #2980, opened 2019-08-08
Remove redundant type annotation.

takayahilton:
NonEmptyTraverse already inherits Reduceable.


kailuowang:
I agree that if you already have `NonEmptyTraversable` then you don't need `Reducible`. The rest of the changes are probably misguided.  Unfortunately there is quite some inconsistency and legacy complexty in the `NonEmptyXXX` datatypes, which is causing quite some confusion. I tried to unify them in a binary compatible way in this PR https://github.com/typelevel/cats/pull/2930, but the community cannot reach a consensus on what should be the best approach. To be fair, I reckon the risk of #2930 brings even more confusion since it only unifies the code in 2.13 code branch, thus created more versions...

takayahilton:
Thanks, I'll just remove Reducible.

PR #2981, opened 2019-08-09
Add Hash instance for NonEmptySet.

takayahilton:


core/src/main/scala/cats/data/NonEmptySet.scala
```diff
@@ -422,12 +422,17 @@ sealed abstract private[data] class NonEmptySetInstances extends NonEmptySetInst
...
-sealed abstract private[data] class NonEmptySetInstances0 {
+sealed abstract private[data] class NonEmptySetInstances0 extends NonEmptySetInstances1 {
   implicit def catsDataEqForNonEmptySet[A](implicit A: Order[A]): Eq[NonEmptySet[A]] = new NonEmptySetEq[A] {
     implicit override def A0: Eq[A] = A
   }
 }
 
+sealed abstract private[data] class NonEmptySetInstances1 {
+  implicit def catsDataHashForNonEmptySet[A: Order: Hash]: Hash[NonEmptySet[A]] =
+    Hash[SortedSet[A]].asInstanceOf[Hash[NonEmptySet[A]]]
```
kailuowang: Hash inherits Eq and thus should swap positions with it.

Issue #2982, opened 2019-08-09
We need to talk about Stream


travisbrown:
## Problem

The current Cats 2.0.0 release candidate (RC1) assumes that Scala 2.13 users will be able to alias the new `LazyList` to `Stream` (which is deprecated in 2.13.0). For example, the `cats.instances.stream` object doesn't actually provide instances for `Stream` on 2.13:

```scala
scala> import cats.Traverse, cats.instances.stream._
import cats.Traverse
import cats.instances.stream._

scala> Traverse[Stream]
               ^
       error: could not find implicit value for parameter instance: cats.Traverse[Stream]
```
But it does provide `LazyList` instances:
```scala
scala> Traverse[LazyList]
res1: cats.Traverse[LazyList] = cats.instances.LazyListInstances$$anon$1@43b7181d
```
There's also a `lazyList` object that provides exactly the same instances:
```scala
scala> import cats.Traverse, cats.instances.lazyList._
import cats.Traverse
import cats.instances.lazyList._

scala> Traverse[LazyList]
res0: cats.Traverse[LazyList] = cats.instances.LazyListInstances$$anon$1@75decfcd
```
So the `stream` object is still around just to facilitate cross-compilation, but only for users who can alias `LazyList`. For everyone else it's just confusing—when you import the contents of something called `instances.stream` you're probably going to assume that you'll get some `Stream` instances.

We're not even really consistent about this, in that cats-kernel takes a slightly different approach: like cats-core, it provides a `stream` object in `cats.kernel.instances` that also contains `LazyList` instances but not `Stream` ones, but unlike cats-core it doesn't have a `lazyList` object.

None of this would be that big of a problem if everyone using 2.13 could just alias `LazyList` (there would still be the inconsistencies, but we have lots of inconsistencies), but that's not realistic, since there are lots of libraries that are published for 2.13 that still use `Stream`.

## Solutions

There are a few things we could do to address these problems in Cats 2.0:

### Provide stream instances

[...]

PR #2983, opened 2019-08-09
Treat Stream and LazyList as different types

travisbrown:
~~A quick implementation of the second proposal in #2982.~~

This became a much deeper fix for #2982 after conversation and a vote in #2991. In the initial version it only removed the aliasing in instances, but now it removes the aliasing altogether.


core/src/main/scala-2.13+/cats/instances/lazyList.scala
```diff
@@ -71,10 +69,11 @@ trait LazyListInstances extends cats.kernel.instances.StreamInstances {
       def tailRecM[A, B](a: A)(fn: A => LazyList[Either[A, B]]): LazyList[B] = {
         val kernel = Iterator.unfold[Option[B], Iterator[Either[A, B]]](Iterator(Left(a))) { it =>
           if (!it.hasNext) None
-          else it.next match {
-            case Left(a) => Some((None, fn(a).iterator ++ it))
-            case Right(b) => Some((Some(b), it))
-          }
+          else
```
travisbrown: This is Scalafmt, which doesn't seem to be running for 2.13-specific code?

core/src/main/scala-2.13+/cats/instances/package.scala
```diff
@@ -0,0 +1,62 @@
+package cats
+
+package object instances {
```
kailuowang: Can we use an abstract class for objects from both scala specific versions to inherit from and remove the duplication?

core/src/main/scala/cats/data/OneAnd.scala
```diff
@@ -209,7 +209,7 @@ sealed abstract private[data] class OneAndLowPriority4 {
         fa.head
 
       def map[A, B](fa: OneAnd[LazyList, A])(f: A => B): OneAnd[LazyList, B] =
-        fa.map(f)
+        fa.map(f)(crossVersionInstancesForLazyList)
```
kailuowang: so for a user who needs to cross compile, they either have to create this `crossVersionInstancesForLazyList` themselves or have to write version specific code for every `import cats.instances.stream._` they have in the past?

kailuowang:
I am bit concerned with the ergonomics for cross compiling users who migrated away from Stream on 2.13. 
Now that we have version specific `cats.instances` and `cats.kernel.instances` why not provide a 
`lazyListXStream` (or a better name) that simply points to `StreamInstances` on 2.12- and `LazyInstances` on 2.13? That way the users mentioned above can simply replace `import cats.intances.stream` with `import cats.instance.lazyListXStream._` in their codebase?

core/src/main/scala/cats/data/OneAnd.scala
kailuowang: see my comment below.

travisbrown:
@kailuowang Personally I think the alias is a bad idea in general. `Stream` and `LazyList` are different types with different semantics, and pretending they're the same is likely to cause subtle and annoying problems. I'm not going to use the alias in any code I maintain, and I think it's a mistake to use it here.

I've only included this `crossVersionInstancesForLazyList` stuff because it doesn't seem like there's any will to make a more complete separation of the types in the implementation of Cats itself. That's fine, I guess—I don't really care that much. I don't think something like this should be included in the public API, but if other people really want a `cats.instance.lazyListXStream` I can add it.

kailuowang:
If you can make **all** `Stream` dependent code version specific, then we don't need the type alias. The approach before this PR was that alias was used as much as possible to support cross compiling except where it can't, such as LazyList instances in `cats.core` or `NonEmptyStream`, in which case they become version specific. 
This PR made more `Stream` dependent code (namely stream instance in kernel) version specific, but still not all of them, hence we still need the alias and the inconsistency.

I'd argue that we either leave as is, or we just go ahead and really make **ALL** `Stream` dependent code version specific and remove the need for type alias all together. 

Either way, I think the `crossVersionInstancesForLazyList`  and `crossVersionEqForLazyList` is awkward even as an internal API in Cats. 
`cats.instance.lazyListXStream` would be better. I don't strongly believe it should be public either, but at the very least we should document to let users how to create their own `cats.instance.lazyListXStream`.

kailuowang:
One more thing, if we make all Stream dependent code version specific we then add back the stream instances on 2.13 and have them deprecated.

core/src/main/scala-2.13+/cats/instances/package.scala
travisbrown: Not without dropping in some `lazy val`s. Here's a minimization:
```scala
package foo

package object bar extends BarPackage
package bar {
  abstract class BarPackage {
    object baz extends Baz
  }
  trait Baz {
    val x = foo.bar.baz
  }
}
```
Referring to `baz` will stack overflow.

travisbrown:
@kailuowang 
> This PR made more `Stream` dependent code (namely stream instance in kernel) version specific, but still not all of them, hence we still need the alias and the inconsistency.

My goal was to remove any _public_ conflation of the two types. My position is that I think conflating the two is a mistake, but as long as Cats doesn't force that decision on me I don't really care about the implementation.

> I'd argue that we either leave as is, or we just go ahead and really make ALL Stream dependent code version specific and remove the need for type alias all together.

How should we decide on this?

kailuowang:
But there is so public conflation (e.g. ZipStream and parallel instance for stream. ), There doesn't seem to be a lot left with the changes from this PR, completely removing the conflation with version specific code seems a reasonable effort. 
Adding stream depending API deprecated back to 2.13 also becomes possible.
The trade off here is code duplication, hard to discover code duplication.


We can hold a vote among the maintainers, because if we chose duplication it will be maintainers burden.

travisbrown:
Okay! How do we call for a vote? I'm happy to make the remaining changes if we decide to go that route.

kailuowang:
Issue created for vote
https://github.com/typelevel/cats/issues/2991

build.sbt
```diff
@@ -523,6 +523,7 @@ lazy val kernelLaws = crossProject(JSPlatform, JVMPlatform)
   .settings(scoverageSettings)
   .settings(disciplineDependencies)
   .settings(testingDependencies)
+  .settings(scalacOptions in Test := (scalacOptions in Test).value.filter(_ != "-Xfatal-warnings"))
```
travisbrown: This is necessary because we now have deprecation warnings for `Stream` on 2.13.

core/src/main/scala-2.13+/cats/data/ScalaVersionSpecificPackage.scala
```diff
@@ -1,7 +1,19 @@
...
 
+import kernel.compat.scalaVersionSpecific._
+
 abstract private[data] class ScalaVersionSpecificPackage {
   type NonEmptyLazyList[+A] = NonEmptyLazyList.Type[A]
+  @deprecated("2.0.0-RC2", "Use NonEmptyLazyList")
+  type NonEmptyStream[A] = OneAnd[Stream, A]
+
+  @deprecated("2.0.0-RC2", "Use NonEmptyLazyList")
+  def NonEmptyStream[A](head: A, tail: Stream[A]): NonEmptyStream[A] =
```
travisbrown: Note that we can't have the default argument here that we have on 2.11 and 2.12, since it results in a deprecation warning (which is clearly a compiler bug).

travisbrown:
This PR is becoming a nightmare, but I think it's ready for review.

Some notes:

* All conflation of `Stream` and `LazyList` are gone.
* All `Stream` instances are still available on 2.13, but are deprecated.
* The `cats.instances.stream` object itself is also deprecated, but `cats.kernel.instances.stream` is not, because it's a package object, and package objects can't be deprecated (this is a particularly fun interaction of our pervasive inconsistency and general Scala brokenness).
* All `Stream`-related tests are still around on 2.13.
* Everything that exists for `Stream` (instances, tests, `NonEmptyStream`, etc.) has a `LazyList` equivalent on 2.13.
* In some cases 2.11 bincompat requires a horrifying amount of code duplication (e.g. if we want to deprecate `catsStdParallelForZipStream` I believe we have to duplicate most of parallel.scala).
* Comparison and equality for `SortedSet` was previously implemented via conversions to `Stream` / `LazyList`. I've reimplemented them in terms of `StaticMethods.iteratorCompare` / `iteratorEq`, which is equivalent and in my view a little clearer, anyway (and definitely more efficient).

There is one difference between `Stream`-related APIs on 2.13 and earlier versions: the `NonEmptyStream` constructor no longer includes a default argument for `tail`:

```scala
@deprecated("2.0.0-RC2", "Use NonEmptyLazyList")
   def NonEmptyStream[A](head: A, tail: Stream[A]): NonEmptyStream[A] = ...
```
While on 2.11-12 we have:

```scala
   def NonEmptyStream[A](head: A, tail: Stream[A] = Stream.empty): NonEmptyStream[A] = ...
```
I removed the default argument on 2.13 because it results in a deprecation warning, which is a compiler bug—it clearly shouldn't warn. We could alternatively leave the default argument and use the silencer scalac plugin to ignore this one (spurious) warning, but that seemed like overkill for a deprecated method.

core/src/main/scala-2.12-/cats/instances/stream.scala
```diff
@@ -6,11 +6,9 @@ import cats.syntax.show._
 import scala.annotation.tailrec
 
 /**
- * For cross compile with backward compatibility
+ * Needed only to avoid some version-specific code in `cats.instances.all`.
  */
-trait LazyListInstances extends StreamInstances with StreamInstancesBinCompat0 {
-  val catsStdInstancesForLazyList = catsStdInstancesForStream
-}
+private[instances] trait LazyListInstances
```
kailuowang: Is this still needed?

core/src/main/scala-2.13+/cats/instances/ScalaVersionSpecificParallelInstances.scala
```diff
@@ -0,0 +1,124 @@
+package cats
+package instances
+
+import cats.data._
+import cats.kernel.Semigroup
+import cats.syntax.either._
+import cats.{~>, Applicative, Apply, FlatMap, Functor, Monad, NonEmptyParallel, Parallel}
+import kernel.compat.scalaVersionSpecific._
```
kailuowang: In all version specific code (i.e. not being cross compiled), you shouldn't need to import a `compat.scalaVersionSpecific._` right? which also means that you shouldn't need the `@suppressUnusedImportWarningForScalaVersionSpecific` below as well.

core/src/main/scala-2.12-/cats/instances/stream.scala
travisbrown: If we don't want two copies of `cats/instances/all.scala` we do need it.

(I guess you could argue that we have two copies of everything else at this point so we might as well do it here.)

core/src/main/scala-2.13+/cats/instances/ScalaVersionSpecificParallelInstances.scala
travisbrown: I think you're right. Let me try removing all of these in the version-specific source trees.

travisbrown: In some cases we might still want to use some of this stuff to minimize the diff between version-specific files? Like `IterableOnce`? I'm not sure—let me try it.

travisbrown: Thanks for catching that, @kailuowang—there were only a couple of place we were using `IterableOnce` or `lazyZip` in version-specific code. Just pushed a commit removing  `scalaVersionSpecific` in all version-specific code (except where it's defined).

kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
```diff
@@ -132,7 +132,7 @@ class Tests extends AnyFunSuiteLike with Discipline {
   checkAll("Eq[List[HasEq[Int]]]", EqTests[List[HasEq[Int]]].eqv)
```
kailuowang: I couldn't find a version specific version of this test (to test the `LazyList` instances)?

core/src/main/scala-2.12-/cats/instances/stream.scala
travisbrown: Yes, but the definition of `AllInstances` in `all.scala` also needs to bring in the `LazyList` instances on 2.13.

kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
travisbrown: Hmm, I know I grepped `tests` but I may have missed these. If so will add them now.

travisbrown: Thanks, just pushed a commit adding these.

core/src/main/scala-2.12-/cats/instances/stream.scala
kailuowang: Ah i see. I didn't see you were able to deprecate all the methods in `StreamInstance` to avoid deprecating the trait. 
Since it confused me a bit, I think we might just as well create two copies for it so that hopefully we can get rid off all `Stream/LazyList` related `ScalaVersionSpecific` stuff in non test code?

kailuowang:
Thanks so much for this epic effort (I tried twice and it was too painful for me to push through). Let's release an RC2 after this gets merged.

travisbrown:
Thanks, @kailuowang! (And thanks to @permutive—most of this was done on their time yesterday :smile:).

travisbrown:
Oh, one other thing I wanted to flag. This compiles on 2.13:

```scala
import cats.{NonEmptyParallel, Parallel}, cats.data._, cats.implicits._

type ZipNonEmptyStream[A] = OneAnd[ZipStream, A]

Parallel[NonEmptyStream, ZipNonEmptyStream]
NonEmptyParallel[NonEmptyLazyList, NonEmptyLazyList.ZipNonEmptyLazyList]
```
But this doesn't:
```scala
Parallel[NonEmptyLazyList, NonEmptyLazyList.ZipNonEmptyLazyList]
```
This isn't just `NonEmptyLazyList`, though—the following also fails, on all Scala versions:
```scala
Parallel[NonEmptyVector, NonEmptyVector.ZipNonEmptyVector]
```
In this PR I just followed `NonEmptyVector` in the implementation of `NonEmptyLazyList`. Anyone know why `NonEmptyVector` doesn't have a `Parallel` instance, while `NonEmptyStream` does? (If not I can look into it in a follow-up.)

travisbrown:
Okay, @LukaJCB pointed out that we should have a `Parallel` for `NonEmptyLazyList` since it's non-strict. I should be able to get to that in the next hour or two.

core/src/main/scala-2.12-/cats/instances/stream.scala
travisbrown: Sure, I can go ahead and do this.

travisbrown:
@LukaJCB @kailuowang Would appreciate another look at that last commit. The issue is that I followed you in using the new newtype pattern for `NonEmptyLazyList`, @kailuowang, but that was based on `NonEmptyVector`, which has different semantics, and doesn't have a `Parallel` instance.

In this version I've provided a `Parallel` instance for `NonEmptyLazyList`, but have followed `NonEmptyStream` in making the parallel side `OneAnd[ZipLazyList, *]`. This puts `NonEmptyLazyList` in a kind of weird position in-between `NonEmptyVector` and `NonEmptyStream` in terms of implementation, but I think it's the best thing to do, given that we can't standardize on the newtype approach everywhere because of bincompat guarantees. Sound reasonable?

travisbrown:
@kailuowang Just pushed one last commit duplicating `all.scala`, so there are now zero references to `LazyList` in `scala-2.12-` source for kernel and core.

travisbrown:
The CI failure is `validateBC` on 2.11, and the error doesn't make any sense. I've just restarted the job after verifying that it's fine locally.

kailuowang:
Very bizarre indeed. `kernelJVM/compile` on Scala 2.11 failed, which passed on other jobs...

core/src/main/scala-2.13+/cats/data/NonEmptyLazyList.scala
```diff
@@ -362,8 +362,23 @@ sealed abstract private[data] class NonEmptyLazyListInstances extends NonEmptyLa
...
 
   implicit def catsDataShowForNonEmptyLazyList[A](implicit A: Show[A]): Show[NonEmptyLazyList[A]] =
-    Show.show[NonEmptyLazyList[A]](nec => s"NonEmpty${Show[LazyList[A]].show(nec.toLazyList)}")
+    Show.show[NonEmptyLazyList[A]](nell => s"NonEmpty${Show[LazyList[A]].show(nell.toLazyList)}")
 
+  implicit def catsDataParallelForNonEmptyLazyList: Parallel[NonEmptyLazyList, OneAnd[ZipLazyList, *]] =
+    new Parallel[NonEmptyLazyList, OneAnd[ZipLazyList, *]] {
+
+      def applicative: Applicative[OneAnd[ZipLazyList, *]] =
+        OneAnd.catsDataApplicativeForOneAnd(ZipLazyList.catsDataAlternativeForZipLazyList)
```
LukaJCB: I think this is the wrong `Applicative`, no? It's `pure` should be `LazyList.continually`.

travisbrown: @LukaJCB It's the same as the one used in the `Parallel` instance for `NonEmptyStream`, and works the same way:

```scala
scala> import cats.data._
import cats.data._

scala> val F = OneAnd.catsDataApplicativeForOneAnd(ZipLazyList.catsDataAlternativeForZipLazyList)
F: cats.Applicative[[β$22$]cats.data.OneAnd[[A]cats.data.ZipLazyList[A],β$22$]] = cats.data.OneAndLowPriority2$$anon$8@3e84b276

scala> F.pure(1).unwrap.value.take(10).toList
res0: List[Int] = List(1, 1, 1, 1, 1, 1, 1, 1, 1, 1)
```

travisbrown: (Which is just to say I'm pretty sure it's the right one, and its `pure` does return an infinite `OneAnd`.)

LukaJCB: Ah great, sorry about that. I guess I'm a bit confused as `catsDataApplicativeForOneAnd ` looks like this: https://github.com/typelevel/cats/blob/48c8b344aaed1f796aebbfdfa610432bee641d84/core/src/main/scala/cats/data/OneAnd.scala#L234

travisbrown: @LukaJCB Right, the `continually`-ing happens in `unwrap`.

travisbrown:
I don't understand this CI failure—I cleared the Travis caches for both the PR and the branch and it's still happening.

travisbrown:
Okay, after a bunch of slow experiments today, here's what I've figured out about the CI failure:

* Changing the Travis CI JDK to OpenJDK fixes the issue (see #2998).
* Restarting, clearing the cache, rebasing commits, etc. doesn't fix the issue.
* I can't reproduce the issue locally (on any JDK or on either Linux or OS X).
* I can't make any sense of how the changes since 67dc634c37938f87b9c7f97434cb416f76731b89 (which was green) could cause this.

kailuowang:
Thanks for the trouble shooting and update. 
I vote we just change the jdk in this PR and call it a day.

travisbrown:
@kailuowang Sounds reasonable. I've just pushed the change. I've also dropped the explicit `dist` config—I'm assuming it was only there because `oraclejdk8` isn't available on the default and more recent options.

LukaJCB:
Thanks for this monster of an effort 👍

tests/src/test/scala/cats/tests/ParallelSuite.scala
```diff
@@ -303,22 +303,6 @@ class ParallelSuite extends CatsSuite with ApplicativeErrorForEitherTest {
     }
   }
 
-  test("ParMap over Stream should be consistent with zip") {
```
kailuowang: I couldn't find the Stream version of this test any more?

tests/src/test/scala/cats/tests/ParallelSuite.scala
```diff
@@ -343,12 +327,6 @@ class ParallelSuite extends CatsSuite with ApplicativeErrorForEitherTest {
     }
   }
 
-  test("ParTupled of Stream should be consistent with ParMap of Tuple.apply") {
```
kailuowang: I couldn't find the Stream version of this test any more?

tests/src/test/scala/cats/tests/ParallelSuite.scala
```diff
@@ -361,12 +339,6 @@ class ParallelSuite extends CatsSuite with ApplicativeErrorForEitherTest {
     }
   }
 
-  test("ParTupled of Stream should be consistent with zip") {
```
kailuowang: I couldn't find the Stream version of this test any more?

tests/src/test/scala/cats/tests/ParallelSuite.scala
```diff
@@ -445,8 +417,6 @@ class ParallelSuite extends CatsSuite with ApplicativeErrorForEitherTest {
   checkAll("NonEmptyParallel[Vector, ZipVector]",
            NonEmptyParallelTests[Vector, ZipVector].nonEmptyParallel[Int, String])
   checkAll("NonEmptyParallel[List, ZipList]", NonEmptyParallelTests[List, ZipList].nonEmptyParallel[Int, String])
-  // Can't test Parallel here, as Applicative[ZipStream].pure doesn't terminate
-  checkAll("Parallel[Stream, ZipStream]", NonEmptyParallelTests[LazyList, ZipStream].nonEmptyParallel[Int, String])
```
kailuowang: I couldn't find the Stream version of this test any more?

tests/src/test/scala/cats/tests/RegressionSuite.scala
```diff
@@ -115,7 +115,7 @@ class RegressionSuite extends CatsSuite {
     // shouldn't have ever evaluted validate(8)
     checkAndResetCount(3)
 
-    LazyList(1, 2, 6, 8).traverse(validate) should ===(Either.left("6 is greater than 5"))
+    Stream(1, 2, 6, 8).traverse(validate) should ===(Either.left("6 is greater than 5"))
```
kailuowang: I don't know much of these set of regression tests, but are we sure that `LazyList` doesn't need them (couldn't find the `LazyList` version)?

tests/src/test/scala/cats/tests/ParallelSuite.scala
travisbrown: Oh, ouch, that's my fault. Fixed now.

travisbrown: Well, not quite, one second…

tests/src/test/scala/cats/tests/RegressionSuite.scala
travisbrown: I omitted these intentionally, but looking more closely I'm surprised the property being tested isn't covered by the `Traverse` laws, and we probably do want it for `LazyList`. Will add now.

travisbrown:
Thanks for the thorough review!

kubukoz:
it's happening

PR #2984, opened 2019-08-12
improve liftTo/rethrowT/raiseOrPure syntax to be more flexible

bpholt:
This takes the changes in #2880 and extends them to a few more methods with similar usage.

### DIFF
--- a/core/src/main/scala/cats/syntax/either.scala
+++ b/core/src/main/scala/cats/syntax/either.scala
@@ -276,6 +276,7 @@ final class EitherOps[A, B](private val eab: Either[A, B]) extends AnyVal {
 
   def toEitherNel[AA >: A]: EitherNel[AA, B] = leftMap(NonEmptyList.one)
 
+  @deprecated("use liftTo instead", "2.0.0")
   def raiseOrPure[F[_]](implicit ev: ApplicativeError[F, A]): F[B] =
     ev.fromEither(eab)
 
--- a/core/src/main/scala/cats/data/EitherT.scala
+++ b/core/src/main/scala/cats/data/EitherT.scala
@@ -49,7 +49,7 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
   /**
    * Inverse of `MonadError#attemptT`
    */
-  def rethrowT(implicit F: MonadError[F, A]): F[B] =
+  def rethrowT(implicit F: MonadError[F, _ >: A]): F[B] =
     F.rethrow(value)
 
   def valueOr[BB >: B](f: A => BB)(implicit F: Functor[F]): F[BB] = fold(f, identity)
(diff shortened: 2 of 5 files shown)


kailuowang:
Thanks!

core/src/main/scala/cats/syntax/either.scala
```diff
@@ -276,8 +276,9 @@ final class EitherOps[A, B](private val eab: Either[A, B]) extends AnyVal {
 
   def toEitherNel[AA >: A]: EitherNel[AA, B] = leftMap(NonEmptyList.one)
 
-  def raiseOrPure[F[_]](implicit ev: ApplicativeError[F, A]): F[B] =
-    ev.fromEither(eab)
+  @deprecated("use liftTo instead", "2.0.0")
+  def raiseOrPure[F[_]](implicit F: ApplicativeError[F, _ >: A]): F[B] =
```
LukaJCB: If we're deprecating this, why change the signature? Should we just leave this as is?

bpholt: That's a good question… as far as I could tell, there was no downside to changing it, so I thought I might as well, in case we end up not deprecating it or it comes in handy for someone. 🤷‍♂

As long as the deprecated version is there too, it might also improve discoverability? I stumbled across `liftTo` for this purpose because I asked about `raiseOrPure` in Gitter and @kubukoz said it sounded like `liftTo` with worse types. Maybe better types for `raiseOrPure` makes it easier to find, and then the deprecation warning points you to the canonically-named method? Idk.

kailuowang: The downside is that there is a small chance it breaks source for some users. If we deprecate it , there is really no reason to improve as it only encourage users to use it right?

bpholt: It makes sense to be conservative. I reverted the signature change and left the deprecation in place.

I'm a little curious when this change could break source, though… do you happen to know any examples?

kailuowang: @bpholt it's probably going to be some extreme edge cases but imagine where there are two `ApplicativeError` instances for `F` in scope,  one of them is `ApplicativeError[F, A]`

kailuowang: Again it's not really expected to break some users, that's why we are totally okay with the other changes.

tests/src/test/scala/cats/tests/EitherSuite.scala
```diff
@@ -350,6 +350,12 @@ class EitherSuite extends CatsSuite {
     }
   }
 
+  test("raiseOrPure syntax works with specialized errors") {
```
kailuowang: no longer valid

bpholt: Oops, good catch

PR #2989, opened 2019-08-14
Add Hash instance for NonEmptyMap.

takayahilton:


core/src/main/scala/cats/data/NonEmptyMapImpl.scala
```diff
@@ -318,10 +318,8 @@ sealed abstract private[data] class NonEmptyMapInstances {
         NonEmptyList(fa.head._2, fa.tail.toList.map(_._2))
     }
 
-  implicit def catsDataEqForNonEmptyMap[K: Order, A: Eq]: Eq[NonEmptyMap[K, A]] =
-    new Eq[NonEmptyMap[K, A]] {
-      def eqv(x: NonEmptyMap[K, A], y: NonEmptyMap[K, A]): Boolean = x === y
-    }
+  implicit def catsDataHashForNonEmptyMap[K: Hash: Order, A: Hash]: Hash[NonEmptyMap[K, A]] =
+    Hash[SortedMap[K, A]].asInstanceOf[Hash[NonEmptyMap[K, A]]]
```
LukaJCB: We should consider adding a `Contravariant[Hash]`, so we don't need to `asInstanceOf` these, WDYT @kailuowang ?

kailuowang: Contravariant[Hash] sounds reasonable, but that would add overhead to this particular case right. So maybe we add that in a separate PR?

LukaJCB: Yes, just in general :)

kailuowang: so shall we merge this one?

LukaJCB: indeed

PR #2992, opened 2019-08-17
Add guarded syntax for Alternative

ybasket:
This add a `.guarded` syntax for any `A`, allowing to lift this value into a `F[_]` with an `Alternative` instance based on whether some test on the value holds or not: `a.guarded[F](f)`. 

Does the same as `f(a).guard[F].as(a)`, but is very helpful when a is not a simple value, but something more expensive to compute, as the `guard` version would compute `a` twice unless it's extracted in to a variable which is ugly in nested calls, example:

```scala
(0 to 100).toList.flatMap(factorial(_).guarded[List](_ % 10 === 7))
```
Calculates all factorials which end with 7, avoiding duplicate calls to the (expensive) `factorial` function as well as the need to assign it.

### DIFF
--- a/core/src/main/scala/cats/implicits.scala
+++ b/core/src/main/scala/cats/implicits.scala
@@ -7,6 +7,9 @@ object implicits
     with syntax.AllSyntaxBinCompat2
     with syntax.AllSyntaxBinCompat3
     with syntax.AllSyntaxBinCompat4
+    with syntax.AllSyntaxBinCompat5
+    with syntax.AllSyntaxBinCompat6
+    with syntax.AllSyntaxBinCompat7
     with instances.AllInstances
     with instances.AllInstancesBinCompat0
     with instances.AllInstancesBinCompat1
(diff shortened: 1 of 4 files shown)


kailuowang:
I am a bit hesitated to add an extension method to all types. On the other hand I can't think of a better API. I am curious what is your real use case? is `F` a `List` and you need to filter?

ybasket:
@kailuowang I checked my projects the last days, often `F` would in the end be `Option`, sometimes `List` or others. The cases with `Option` are often checking whether some additional thing needs to be applied in case this additional thing is severe enough. Example from one of my projects is car routing, if the calculated route is ending too far away from the actual destination, add some additional walking to compensate:
```scala
val walkingDistance: Option[Distance] = distance(routeEnd, destination).guarded[Option](_ > Kilometers(3))
```

I absolutely see your point about not adding "unrestricted" implicit syntax and the use case isn't too strong as there are reasonable other ways to do it (both using `Alternative` or without), so I'd be OK if this PR gets rejected. Just found myself repeatedly having the use case, so I thought it might be worth adding.

PS: Sorry for the late answer.

kailuowang:
@ybasket no worries. For your use case, will `Option(distance(routeEnd, destination)).filter(_ > Kilometers(3))` be as concise and readable? 
To abstract out `Option` I think you can use `Applicative` + `FunctorFilter`

ybasket:
@kailuowang `pure` + `filter`  is one of the "workarounds" I'm using so far, so yes. I find it slightly odd to lift and then filter, but it solves the use case as well. Thank you for your time and comments!

I will close this PR now.

PR #2996, opened 2019-08-20
Add :== unapply to Chain.

takayahilton:


tests/src/test/scala/cats/tests/ChainSuite.scala
```diff
@@ -84,6 +85,10 @@ class ChainSuite extends CatsSuite {
       case h ==: t => (h, t) should ===(1 -> Chain(2, 3))
     }
 
+    Chain(1, 2, 3) match {
+      case h :== t => (h, t) should ===(Chain(1, 2) -> 3)
+    }
```
kailuowang: Nit can we use 'init' and 'last' instead of h and t?

PR #2998, opened 2019-08-21
Separate Stream and LazyList

### LINKED ISSUE #2982: We need to talk about Stream

travisbrown:
Trying this one more time, after rebasing all commits.

Fixes #2982. This PR is an attempt to resolve a really weird Travis CI issue in #2983 that's persisted through a couple of restarts and cache clearings.

All feedback from #2983 has been addressed and as far as I'm concerned this is ready to go.


travisbrown:
Moved this fix back to #2983.

PR #2999, opened 2019-08-21
optimize chain's get.

takayahilton:


kailuowang:
I know `get` is tested by `Foldable` laws but since now we added a `get` to the class, shall add a test here as well? something like `forAll(c: Chain, i: Int) => c.get(i) shouldBe c.toList.lift(i)`?

This is the same CI error seen in #2983 , you can either change the jdk as done there or wait for that PR to be merged and merge master back to this one.

takayahilton:
Thanks!
Fixed.

PR #3000, opened 2019-08-22
More consistency

travisbrown:
## Update

Because some of these fixes are more controversial than I expected, I've split the easy ones out into separate PRs: 

* Move appropriate SortedSet and SortedMap instances to kernel (#3001)
* Fix easy typos (#3002)
* Make bincompat and prioritization traits package-private (#3003)

I'm hoping we can get at least these merged before 2.0, even if the rest has to wait for 2.1.

## Original description

There's currently a lot of inconsistency in Cats, both in the implementation and in the public API. I know this isn't something most people care about, but I hate it, and even for normal people I imagine it has to make contributing and reading the code harder than it should be. This PR aims to fix a few things that can be fixed without breaking binary compatibility in Cats 2, and to put us in a better position to make things even less horrible in Cats 3!

It does break source compatibility, it breaks it all over the place, but in ways that shouldn't affect normal usage. This is part of the reason I'd really like to get these changes into 2.0.0.

### Misnamed instances

Most of these changes are pretty mechanical. I see an `Eq` instance named `zipNelEq`, I get extremely angry, I deprecate it, remove its `implicit`, make it package private, copy the implementation, rename it `catsDataEqForZipNonEmptyList` or whatever it is it actually should be, make sure `validateBC` still passes on all Scala versions, and then repeat dozens of times for the rest of the misnamed stuff. In some cases I've also had to add new `BinCompat` traits for misnamed instances defined in traits (thanks to 2.11).

### SortedSet and SortedMap instances

This is one of the bigger changes. Previously we only had instances for cats-kernel type classes (`Eq`, `Monoid`, etc.) for `SortedSet` and `SortedMap` in cats-core. I don't know why, but I don't think there's a good reason for it. To make things even more confusing, some of the instances for these types in cats-core followed the cats-kernel instance naming convention. I've moved these instances into cats-kernel and done all the necessary juggling to keep from breaking bincompat in cats-core.

### Cleaner public API

I've also made all (or approximately all) `BinCompat` and prioritization traits private to the smallest possible enclosing package (I'm pretty confident the `BinCompat` ones are all done, but the prioritization ones are easier to miss). Some of these were already package-private, but apparently at random. There's no good reason for them to be in the public interface, so they're gone now.

### DIFF
--- a/core/src/main/scala/cats/instances/function.scala
+++ b/core/src/main/scala/cats/instances/function.scala
@@ -22,24 +22,8 @@ trait FunctionInstancesBinCompat0 {
       override def index[A](f: E => A): E => A = f
     }
 
-  implicit val catsSddDeferForFunction0: Defer[Function0] =
-    new Defer[Function0] {
-      case class Deferred[A](fa: () => Function0[A]) extends Function0[A] {
-        def apply() = {
-          @annotation.tailrec
-          def loop(f: () => Function0[A]): A =
-            f() match {
-              case Deferred(f) => loop(f)
-              case next        => next()
-            }
-          loop(fa)
-        }
-      }
-      def defer[A](fa: => Function0[A]): Function0[A] = {
-        lazy val cachedFa = fa
-        Deferred(() => cachedFa)
-      }
-    }
+  @deprecated("2.0.0-RC2", "Use cats.instances.function.catsStdDeferForFunction0")
+  private[instances] def catsSddDeferForFunction0: Defer[Function0] = cats.instances.function.catsStdDeferForFunction0
 
   implicit def catsStdDeferForFunction1[A]: Defer[A => *] =
     new Defer[A => *] {
@@ -55,6 +39,27 @@ trait FunctionInstancesBinCompat0 {
         }
       }
       def defer[B](fa: => A => B): A => B = {
+        lazy val cachedFa = fa
+        Deferred(() => cachedFa)
+      }
+    }
+}
+
+private[instances] trait FunctionInstancesBinCompat1 {
+  implicit val catsStdDeferForFunction0: Defer[Function0] =
+    new Defer[Function0] {
+      case class Deferred[A](fa: () => Function0[A]) extends Function0[A] {
+        def apply() = {
+          @annotation.tailrec
+          def loop(f: () => Function0[A]): A =
+            f() match {
+              case Deferred(f) => loop(f)
+              case next        => next()
+            }
+          loop(fa)
+        }
+      }
+      def defer[A](fa: => Function0[A]): Function0[A] = {
         lazy val cachedFa = fa
         Deferred(() => cachedFa)
       }
--- a/core/src/main/scala/cats/instances/function.scala
+++ b/core/src/main/scala/cats/instances/function.scala
@@ -9,7 +9,7 @@ import annotation.tailrec
 
 trait FunctionInstances extends cats.kernel.instances.FunctionInstances with Function0Instances with Function1Instances
 
-trait FunctionInstancesBinCompat0 {
+private[instances] trait FunctionInstancesBinCompat0 {
 
   /**
    * Witness for: E => A <-> E => A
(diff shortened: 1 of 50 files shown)


core/src/main/scala/cats/instances/function.scala
```diff
@@ -9,7 +9,7 @@ import annotation.tailrec
 
 trait FunctionInstances extends cats.kernel.instances.FunctionInstances with Function0Instances with Function1Instances
 
-trait FunctionInstancesBinCompat0 {
+private[instances] trait FunctionInstancesBinCompat0 {
```
kailuowang: I am not sure about introducing a BinCompat trait to fix a typo. The plan is to branch out scala 2.11 code right after 2.0 release.  Which means we shall be able to leave the typo in the 2.11 code and fix it in master branch without an extra bincompat trait.

core/src/main/scala/cats/instances/sortedSet.scala
```diff
@@ -90,43 +91,27 @@ trait SortedSetInstancesBinCompat0 {
...
-class SortedSetOrder[A: Order] extends Order[SortedSet[A]] {
-  def compare(a1: SortedSet[A], a2: SortedSet[A]): Int =
-    Order[Int].compare(a1.size, a2.size) match {
-      case 0 => StaticMethods.iteratorCompare(a1.iterator, a2.iterator)
-      case x => x
-    }
-
-  override def eqv(s1: SortedSet[A], s2: SortedSet[A]): Boolean =
-    StaticMethods.iteratorEq(s1.iterator, s2.iterator)
+private[instances] trait SortedSetInstancesBinCompat1 extends LowPrioritySortedSetInstancesBinCompat1 {
```
kailuowang: same concern here, introducing a bincomp trait to fix an inconsistent placement while we should be able to fix this right after 2.0 release without it

core/src/main/scala/cats/instances/function.scala
travisbrown: I can pull these parts out, but in my view it'd be better to get all the instance name-related source compat breakage out of the way at once, instead of spreading it across 2.0 and 2.1.

docs/src/main/tut/guidelines.md
```diff
@@ -80,7 +80,7 @@ therefore name our implicits according to the following rules:
 - If the instance is for multiple type classes, use `InstancesFor` instead of a type class name.
 - If the instance is for a standard library type add `Std` after the package. i.e. `catsStdShowForVector` and `catsKernelStdGroupForTuple`.
 
-As an example, an implicit instance of `Monoid` for `List` defined in the package `Kernel` should be named `catsKernelMonoidForList`.
+As an example, an implicit instance of `Monoid` for `List` defined in the package `Kernel` should be named `catsKernelStdMonoidForList`.
```
kailuowang: I am not sure if the renaming of these `catskernelXXXX` instances are worth the extra code due to BC 
The `std` in implicit names in `cats-core` was due to the fact that `cats.instances` package was used to be `cats.std`. And instances in `cats.kernel.instances`  don't have the `instance` package name in their names because having instances in an instance name is kind of redundant.  We have this long name rule in place simply so that contributors can easily come up with long unique name for these implicit defs to avoid potential conflict with other libraries. I agree that ideally we should keep it consistent but I think we are paying too much a cost for this kind of rename fix.

core/src/main/scala/cats/instances/function.scala
kailuowang: The source breakage is highly unlikely to affect anyone (and if anyone it would be a super quick fix on compilation time), but the extra trait will be left there for a very long time.

travisbrown: So you're saying that breaking source compat in 2.x releases is acceptable?

travisbrown:
Closing this since the important parts got split out and merged in separate PRs. Maybe we can revisit the rest after the 2.11 code split.

PR #3003, opened 2019-08-23
Make bincompat and prioritization traits package-private

travisbrown:
One last piece of #3000 that shouldn't be controversial. This change makes all bincompat and prioritization traits (or classes, in a few cases) private to the smallest possible enclosing package.

This is a source-incompatible change, but it's verified to be binary compatible, and there's no good reason for anyone to use these traits explicitly.


travisbrown:
@kailuowang Does this look okay to you?

PR #3006, opened 2019-08-24
[issue-1222] Typeclass for Seq

VivekRagunathan:


larsrh:
Implemented by #3620.

PR #3012, opened 2019-08-26
Make Parallel's F a type member

### LINKED ISSUE #2233: Use an abstract type member for the parallel type of the Parallel type class
Right now we're using two type parameters which essentially means bidirectional functional dependencies. However all of our functions making use of parallel are uniquely determined by `F`, meaning we can make `G` an abstract type parameter which would get rid of the some of boilerplate when writing abstract code over `Parallel`. This is a breaking change, so would need to go to cats 2.0. :)

travisbrown:
Fixes #2233. We'd been holding off on this change because we thought it would necessarily require both source and binary compatibility breakage, but it turns out we can fix it without breaking binary compatibility with Cats 1.x.

Note that this change does break source compatibility pretty dramatically, but the migration path will generally be pretty clear (add `.Aux`, maybe an explicit type parameter here or there). Anyway, as Rob Norris says [here](https://gitter.im/typelevel/cats?at=5d63e222e403470ab6dd1976), the current API is so awkward to use that this is unlikely to break much code.

With this change you can use parallel syntax without having to thread an extra type parameter everywhere through your code. Taking an example from the README of [cats-par](https://github.com/ChristopherDavenport/cats-par) (a library for working around the limitations of the current API), this PR allows us to write the following:

```scala
import cats._
import cats.implicits._
import cats.data._

def withNewParallel[F[_]: Monad: Parallel, A, C, D](
  as: List[A],
  f: A => Kleisli[F, C, D]
): Kleisli[F, C, List[D]] =
  as.parTraverse(f)
```
With the current API we have to include an extra parameter `G[_]`, both here and all the way up our call stack:
```scala
def withOldParallel[F[_]: Monad, G[_], A, C, D](
  as: List[A],
  f: A => Kleisli[F, C, D]
)(implicit P: Parallel[F, G]): Kleisli[F, C, List[D]] =
  as.parTraverse(f)
```
Note that the only changes to the tests are a few `Aux`s (and one fewer type parameter in one `applicativeError` call); otherwise the usage exercised in the tests didn't require changes.

### DIFF
--- a/core/src/main/scala-2.12-/cats/instances/ScalaVersionSpecificParallelInstances.scala
+++ b/core/src/main/scala-2.12-/cats/instances/ScalaVersionSpecificParallelInstances.scala
@@ -21,24 +21,23 @@ trait ParallelInstances extends ParallelInstances1 {
         λ[Either[E, *] ~> Validated[E, *]](_.toValidated)
     }
 
-  implicit def catsParallelForOptionTNestedOption[F0[_], M[_]](
-    implicit P: Parallel.Aux[M, F0]
-  ): Parallel.Aux[OptionT[M, *], Nested[F0, Option, *]] = new Parallel[OptionT[M, *]] {
-    type F[x] = Nested[F0, Option, x]
+  implicit def catsParallelForOptionTNestedOption[M[_]](
+    implicit P: Parallel[M]
+  ): Parallel.Aux[OptionT[M, *], Nested[P.F, Option, *]] = new Parallel[OptionT[M, *]] {
+    type F[x] = Nested[P.F, Option, x]
 
-    implicit val appF: Applicative[F0] = P.applicative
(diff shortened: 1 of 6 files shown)


kubukoz:
y'all reading in my mind, I thought about this just a couple days ago... a very welcome change!

SethTisue:
recent discussion at https://github.com/scala/scala/pull/9303 has us wondering if this change was in fact safe. (which might affect decisions to make other such changes in the future)

kubukoz:
Wow, took us long enough!

I can't remember any case in the ecosystem where this broke anything, but I would wager Parallel isn't that popular an abstraction in libraries, which could be why - everyone just updated cats and their sources and they were fine. (I think)

SethTisue:
Perhaps y'all should `mimaReportSignatureProblems := true` and see whether there are too many false positives.

PR #3024, opened 2019-09-03
Adds option object ops with catchNonFatal method

stremlenye:
Introduces the `catchNonFatal` helper for Option type:

`Option.catchNonFatal { "foo".toInt }`

It breaks backwards compatibility and I don't really know where to proceed from here.


ybasket:
For the bin compat problem, check how `EitherSyntaxBinCompat0` is defined and used, same trick should work here.

Maybe it's also worth to add `catchOnly` while being on it?

stremlenye:
Thanks for the advice.
Regarding `catchOnly` it's surely doable. I am just not confident enough if its needed in general (even as the member of `EitherObjectOps`): seems odd to me to wrap the computation into effect and expect it to fail in certain cases.

That's as well leads to cases like this:

```scala 
type A = …
type B = …
type T <: Throwable = …
type T2 <: Throwable =!= T


// method signature suggests it already takes care of the side effects like exceptions
def foo(a : => A): Either[Throwable, B] =
  Either.catchOnly[T](a).map(…: B)

def a = throw new T2

foo(a) // fails in runtime even though it looked safe from the call site
```

ybasket:
I agree that your example is something we don't want to see in real code.

But I see some use cases where it can indeed be useful to expect certain exceptions, especially when dealing with Java APIs who declare them to be thrown, for example in number parsing as your initial example:

```scala
// using Integer.parseInt to make underlying Java call more visible, 
// .toInt is just syntax for it
Option.catchOnly[NumberFormatException](Integer.parseInt(...))
// vs idiomatic plain Scala - catches everything
Try(Integer.parseInt(...)).toOption
// or cats via Either
Either.catchOnly[NumberFormatException](Integer.parseInt(...)).toOption
```

The last is how mouse implements its `parseDoubleOption`, see https://github.com/typelevel/mouse/blob/master/shared/src/main/scala/mouse/string.scala. In these cases, you don't want to catch any exception except `NumberFormatException`, so a `Option.catchOnly` would be handy. But this just my 2 cents, just `catchNonFatal` alone is already something I'd like to use :)

stremlenye:
> But I see some use cases where it can indeed be useful to expect certain exceptions, especially when dealing with Java APIs who declare them to be thrown

Fair enough. I'll add `catchOnly`.

johnynek:
I'm not crazy about adding 1-off utility methods personally.

An alternative formulation might be adding:
```scala
trait ApplicativeError[F[_], E]] {
  def catchNonFatalOr[A](a: => A)(fn: Throwable => F[A]): F[A] =
    try pure(a)
    catch {
      case NonFatal(e) => fn(e)
    }
}
```
Then this could be `ApplicativeError[Option, Unit].catchNonFatalOr(a)(_ => None)`

Then `catchNonFatal` on ApplicativeError could just be: `catchNonFatalOr(a)(raiseError(_))`

more verbose, but also more general.

johnynek:
or, of course `Try(a).toOption` does the same thing as the `Option.catchNonFatal`.

stremlenye:
@johnynek  `ApplicativeError` already has `catchNonFatal` method, so the first solution you proposed is already available.

Regarding the second one: that one would do as well but requires instantiation of `Try` just to convert it into option immediately. Not a huge overhead but completely avoidable.

On the other hand the `Option.catchNonFatal` provides the uniformal with related data types API and being simply the syntactic sugar aims to rather not to generalise but to provide readable syntax with as less overhead as possible.

johnynek:
@stremlenye the first option isn’t available because Option doesn’t implement ApplicativeError for an error type that is a superclass of Throwable. The error type is Unit. So you can’t use the catchNonFatal method with Option currently. That is why I suggested adding a method the converts the Throwable to the Applicative value. Although now that I say it again I realize that could also be in Applicative not Applicative error because it doesn’t use raiseError.

stremlenye:
@johnynek I am sorry I am not entirely sure I understood your proposal. Could you please elaborate a bit more on it?

stremlenye:
I finally got some time to get back to this PR.
On the suggestion of `ApplicativeError[Option, Unit].catchNonFatalOr(a)(_ => None)`: `ApplicativeError.handleErrorWith` does pretty much that.

Regarding `a method the converts the Throwable to the Applicative value` I couldn't grasp the meaning of the proposal.

travisbrown:
> On the other hand the Option.catchNonFatal provides the uniformal with related data types API 

I'd put this in the cons column, since it makes it look a lot like it's doing the same thing as the `ApplicativeError` syntax but is actually throwing the error away.

I think this is a reasonable candidate for [Mouse](https://github.com/typelevel/mouse) (if it's not there already?), but isn't a good fit for Cats.

stremlenye:
@travisbrown makes sense. I guess this PR could be closed and as for me – I'll check if the `mouse` does not have it already.

travisbrown:
Thanks @stremlenye! Feel free to ping me if you'd like review on a Mouse PR.

stremlenye:
@stremlenye no worries 👍 Thanks for review!

PR #3041, opened 2019-09-08
Add scala.util.control.TailCalls.TailRec instances

johnynek:
Adds Defer and Monad instances for TailRec. Now that we don't support 2.10 anymore, we can also re-enable the benchmarks. For my (somewhat slow) machine, TailRec is about 17% faster than Trampoline (but much slower than Eval).
```
[info] Benchmark                    Mode  Cnt      Score       Error  Units    
[info] TrampolineBench.eval        thrpt    3  16603.447 ± 28185.656  ops/s    
[info] TrampolineBench.stdlib      thrpt    3   4829.133 ±  1911.742  ops/s    
[info] TrampolineBench.trampoline  thrpt    3   4152.917 ±  1948.304  ops/s
```

The motivation to add this are for users interacting with code that already uses TailRec and for instance interpreting Free into TailRec, or someone who wants to use TailRec with generic code that uses Defer (e.g. ContT).

We could add Semigroup, Monoid and Group to kernel. Since this is further from the main use of TailRec and independent of this current PR, I will not include it here.


core/src/main/scala/cats/Eval.scala
```diff
@@ -378,6 +378,7 @@ sealed abstract private[cats] class EvalInstances extends EvalInstances0 {
       def flatMap[A, B](fa: Eval[A])(f: A => Eval[B]): Eval[B] = fa.flatMap(f)
       def extract[A](la: Eval[A]): A = la.value
       def coflatMap[A, B](fa: Eval[A])(f: Eval[A] => B): Eval[B] = Later(f(fa))
+      override def unit: Eval[Unit] = Eval.Unit
```
travisbrown: This is unrelated?

johnynek: unrelated, but I noticed we could add this and avoid allocations for some folks that use `Applicative[F].unit` with `Eval`.

bench/src/main/scala/cats/bench/TrampolineBench.scala
```diff
@@ -30,14 +31,12 @@ class TrampolineBench {
       y <- Trampoline.defer(trampolineFib(n - 2))
     } yield x + y
 
-  // TailRec[A] only has .flatMap in 2.11.
+   @Benchmark
```
travisbrown: Good catch. Doing a search for "2.10" shows four or five other things like this we should do, but none of them look urgent.

core/src/main/scala-2.12-/cats/instances/all.scala
```diff
@@ -61,4 +61,7 @@ trait AllInstancesBinCompat4 extends SortedMapInstancesBinCompat1 with MapInstan
 
 trait AllInstancesBinCompat5 extends SortedSetInstancesBinCompat0
 
-trait AllInstancesBinCompat6 extends SortedSetInstancesBinCompat1 with SortedMapInstancesBinCompat2
+trait AllInstancesBinCompat6
```
travisbrown: Doesn't this break bincompat on 2.11? Only with the recent release candidates, I guess, so maybe it's fine?

core/src/main/scala/cats/Eval.scala
johnynek: yes. I added a comment, I just noticed it was missing while looking at Eval.

core/src/main/scala-2.12-/cats/instances/all.scala
johnynek: I would assume so. I think we should only bump mima on actual releases (not including any pre-release).

travisbrown: @johnynek Right, but it's still useful to know, so we can decide whether we need to publish new pre-releases of cats-effect or Circe after a Cats pre-release, etc.

kailuowang: the plan is to drop 2.11 on master immediately after 2.0.0 release (this evening). So we can add this to the root trait.

johnynek:
@kailuowang any concerns with merging?

kailuowang:
@johnynek sorry it was too last minute to be carried on the 2.0.0 train.
Now 2.0.0 is released this PR as is breaks BC on 2.11. 
The plan is to drop Scala 2.11 on master (I'll find time to do that this week if no one beats me to it), after that we can change this PR to simply extend the `AllInstances` instead of a BinCompat trait.

johnynek:
@kailuowang I'm confused why we can't merge now, and then when you make the change to merge in the trait and drop 2.11 everything is fine?

kailuowang:
@johnynek we can't merge the existing bincompat traits because of BC. If we merge this now then we need to 
a) remember to move it to a bincompat trait on the new 2.11 branch or remove it. 
and 
b) on master branch we can either leave it on the bin compat branch or move it to the AllInstance trait. 

It's just easier if we do it the other order. I am already working on dropping 2.11 on master so this shouldn't be held for long. 
https://github.com/typelevel/cats/pull/3051

travisbrown:
Just formatted, which should fix the build.

travisbrown:
Now there's a bincompat error, and it looks legit. I don't know how this ever passed, given that adding a `val` to a trait should always have been a breaking change?

johnynek:
I guess the idea that we can just add new traits to AllInstances in 2.0.0 is wrong in some way.

travisbrown:
@johnynek We can, just not as `val`s. I've got a bincompat version here (or if you don't mind I could just push to this branch instead): https://github.com/typelevel/cats/pull/3127

johnynek:
I would welcome your push, as my vanity has a slight preference for this PR be the one merged.

johnynek:
Thank you for doing this work.

travisbrown:
@johnynek Pushed, thanks! (They'd be your commits either way. :smile:)

johnynek:
👍 

Thanks @kailuowang and @travisbrown

PR #3042, opened 2019-09-09
Remove unused type parameter for Parallel instances

travisbrown:
This PR includes @Billzabob's commit from #2789 but rebased against current master, where the change has to happen in two places, and slightly adjusted to avoid breaking binary compatibility.

I missed @Billzabob's PR the first time around but should have caught this in #3012.

The change does technically break source compatibility, but (while it fixes a real problem) it's also pretty minor, so I'm not worried about it causing anyone problems. I personally think we can / should try to get it into 2.0.0 today.

### DIFF
--- a/core/src/main/scala-2.12-/cats/instances/ScalaVersionSpecificParallelInstances.scala
+++ b/core/src/main/scala-2.12-/cats/instances/ScalaVersionSpecificParallelInstances.scala
@@ -26,7 +26,7 @@ trait ParallelInstances extends ParallelInstances1 {
     implicit P: Parallel[M]
   ): Parallel.Aux[OptionT[M, *], Nested[P.F, Option, *]] = OptionT.catsDataParallelForOptionT[M]
 
-  implicit def catsStdNonEmptyParallelForZipList[A]: NonEmptyParallel.Aux[List, ZipList] =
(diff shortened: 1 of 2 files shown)


travisbrown:
The failing 2.13 test is unrelated, so I restarted the job. I'm not sure whether it needs attention:
```
[info] - CommutativeGroup[BigDecimal].commutativeGroup.associative *** FAILED ***
[info]    (Discipline.scala:12)
[info]     Falsified after 8 successful property evaluations.
[info]     Location: (Discipline.scala:12)
[info]       arg0 = 1.4502366979631793E-19,
[info]       arg1 = 3.2509612437180712E-9,
[info]       arg2 = -81805312727198.39
[info]     )
[info]       Expected: -81805312727198.38999999674903875613
[info]   Received: -81805312727198.38999999674903875614
```

kailuowang:
I am going to wait for Oscar until this afternoon. but I think it's such a trivial one that we can safely merge with 2 maintainer approvals.

johnynek:
Personally I’m far more worried about binary incompatibility because they show up at runtime when linking large sets of dependencies together.

Source incompatibility isn’t great, but can be detected with a compiler and doesn’t hit you at runtime. Also, for almost all users, they won’t see this because 1. Parallel isn’t that commonly used, 2. I imagine they are resolving this method implicitly.

PR #3046, opened 2019-09-09
Infer dependent parameter in NonEmpty/ParallelTests/Laws

djspiewak:
It's a little wonky. Also I left one `.Aux` usage in the tests which came from the fact that the compiler just wasn't able to infer the dependent version. I didn't look into why, but I suspect it's wonkiness with `OneAnd` and a scalac bug. Putting up the @milessabin signal!

Also this introduces an interesting pattern, which is adding `def Aux` alongside `type Aux` when you have `def apply`. I would propose we should do this for `Parallel` itself as well, but I didn't make that change. I think this is better than the current `DummyImplicit` approach, but I'm okay with going with `DummyImplicit` instead.

### DIFF
--- a/laws/src/main/scala/cats/laws/NonEmptyParallelLaws.scala
+++ b/laws/src/main/scala/cats/laws/NonEmptyParallelLaws.scala
@@ -22,8 +22,8 @@ object NonEmptyParallelLaws {
   type Aux[M[_], F0[_]] = NonEmptyParallelLaws[M] { type F[A] = F0[A]; val P: NonEmptyParallel.Aux[M, F0] }
 
   def apply[M[_]](implicit ev: NonEmptyParallel[M]): NonEmptyParallelLaws.Aux[M, ev.F] =
-    Aux[M, ev.F](ev)
+    apply[M, ev.F](ev, implicitly)
 
-  def Aux[M[_], F[_]](implicit ev: NonEmptyParallel.Aux[M, F]): NonEmptyParallelLaws.Aux[M, F] =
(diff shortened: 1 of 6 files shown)


djspiewak:
Random clarification: this is only source-breaking if someone wrote their own laws derived from `ParallelLaws`. Anything else should be source-compatible.

laws/src/main/scala/cats/laws/NonEmptyParallelLaws.scala
```diff
@@ -18,6 +19,11 @@ trait NonEmptyParallelLaws[M[_], F[_]] {
...
 
 object NonEmptyParallelLaws {
-  def apply[M[_], F[_]](implicit ev: NonEmptyParallel.Aux[M, F]): NonEmptyParallelLaws[M, F] =
-    new NonEmptyParallelLaws[M, F] { def P: NonEmptyParallel.Aux[M, F] = ev }
+  type Aux[M[_], F0[_]] = NonEmptyParallelLaws[M] { type F[A] = F0[A]; val P: NonEmptyParallel.Aux[M, F0] }
+
+  def apply[M[_]](implicit ev: NonEmptyParallel[M]): NonEmptyParallelLaws.Aux[M, ev.F] =
+    apply[M, ev.F](ev, implicitly)
+
+  def apply[M[_], F[_]](implicit ev: NonEmptyParallel.Aux[M, F], D: DummyImplicit): NonEmptyParallelLaws.Aux[M, F] =
```
travisbrown: It doesn't really matter but I did this in the other direction in #3031, with the `DummyImplicit` on the new one.

Issue #3040, opened 2019-09-08
Add `ifF` on Functor


kubukoz:
Signature:

```scala
//trait Functor[F[_]]
def ifF[A](fb: F[Boolean])(ifTrue: A, ifFalse: A): F[A]
```

Similar to `ifM` from Monad, but doesn't flatten.

Will be closed by #3058

stremlenye:
Shouldn't it go to the `IfFunctorOps` like it was done for `Apply` and `FlatMap`?

kubukoz:
I think it could be either of these places. Both in the type class with a default implementation and in the syntax. Probably only in the syntax until we drop 2.11 support though.

joroKr21:
That reads like "if and only if" :smile: 
Should the branches be by-name though?

kubukoz:
For laziness, they probably should, yeah.

Good point about the name, but it matches ifA and ifM.

gagandeepkalra:
shouldn't ifTrue and ifFalse be by name parameters ? 

`def ifF[A](fb: F[Boolean])(ifTrue: => A, ifFalse: => A): F[A] = map(fb)(x => if (x) ifTrue else ifFalse)`

vasiliybondarenko:
Could anybody review my solution?

kubukoz:
Merged to master in #3058!

PR #3058, opened 2019-09-12
Add `ifF` on Functor #3040

vasiliybondarenko:
Added `ifF` function to Functor:
```scala
def ifF[A](fb: F[Boolean])(ifTrue: A, ifFalse: A): F[A] = map(fb)(x => if (x) ifTrue else ifFalse)
```

### DIFF
--- a/tests/src/test/scala/cats/tests/FunctorSuite.scala
+++ b/tests/src/test/scala/cats/tests/FunctorSuite.scala
@@ -37,4 +37,12 @@ class FunctorSuite extends CatsSuite {
       assert(widened eq list)
     }
   }
+
+  test("ifF equals map(if(_) ifTrue else ifFalse)") {
+    forAll { (l: List[Boolean], o: Option[Boolean], m: Map[String, Boolean]) =>
+      Functor[List].ifF(l)(1, 0) should ===(l.map(if (_) 1 else 0))
+      Functor[Option].ifF(o)(1, 0) should ===(o.map(if (_) 1 else 0))
+    }
+  }
+
 }
--- a/tests/src/test/scala/cats/tests/FunctorSuite.scala
+++ b/tests/src/test/scala/cats/tests/FunctorSuite.scala
@@ -45,4 +45,13 @@ class FunctorSuite extends CatsSuite {
     }
   }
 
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -397,7 +397,8 @@ def mimaSettings(moduleName: String) =
           exclude[IncompatibleMethTypeProblem]("cats.arrow.FunctionKMacros.lift"),
           exclude[MissingTypesProblem]("cats.arrow.FunctionKMacros$"),
           exclude[IncompatibleMethTypeProblem]("cats.arrow.FunctionKMacros#Lifter.this"),
-          exclude[IncompatibleResultTypeProblem]("cats.arrow.FunctionKMacros#Lifter.c")
+          exclude[IncompatibleResultTypeProblem]("cats.arrow.FunctionKMacros#Lifter.c"),
+          exclude[ReversedMissingMethodProblem]("cats.Functor.ifF")
```
kubukoz: This will break binary compatibility on 2.11. you should add this method as an extension method instead (see the bincompat traits that already exist for other type classes)

tests/src/test/scala/cats/tests/FunctorSuite.scala
```diff
@@ -37,4 +37,12 @@ class FunctorSuite extends CatsSuite {
       assert(widened eq list)
     }
   }
+
+  test("ifF equals map(if(_) ifTrue else ifFalse)") {
```
kubukoz: That tests ensures the implementation is equivalent to the default one, but doesn't really check whether the default one works. I'd suggest adding some tests with concrete lists and concrete expected values.

build.sbt
kubukoz: I just realized 2.11 has been dropped on master so it doesn't matter. You should be able to rebase onto master and remove this line.

vasiliybondarenko: I wasn't able to rebase to recent master so I force pushed my changes.

tests/src/test/scala/cats/tests/FunctorSuite.scala
vasiliybondarenko: Tests were added

PR #3051, opened 2019-09-10
Drops Scala 2.11 support on master

kailuowang:
~Another PR to add the scala 2.11 branch coming soon~

PR to add the scala_2.11 branch  #3052

PR #3060, opened 2019-09-13
Hash[Map[K, V]] has unnecessary constraint #3039

### LINKED ISSUE #3039: Hash[Map[K, V]] has unnecessary constraint
Similar to #3038. Now we have:

```scala
  implicit def catsKernelStdHashForMap[K: Hash, V: Hash]: Hash[Map[K, V]] =
    new MapHash[K, V]
```
But the `K: Hash` isn't used in the implementation.

vasiliybondarenko:
Removed `Order` constraint from `Hash[Map[K, V]]`but bincomp is broken. Please advise how it can be overcome

Closes #3039

### DIFF
--- a/core/src/main/scala/cats/instances/sortedSet.scala
+++ b/core/src/main/scala/cats/instances/sortedSet.scala
@@ -73,7 +73,7 @@ trait SortedSetInstances extends SortedSetInstances1 {
 private[instances] trait SortedSetInstances1 {
   @deprecated("2.0.0-RC2", "Use cats.kernel.instances.sortedSet.catsKernelStdHashForSortedSet")
   private[instances] def catsKernelStdHashForSortedSet[A: Order: Hash]: Hash[SortedSet[A]] =
-    cats.kernel.instances.sortedSet.catsKernelStdHashForSortedSet[A]
+    cats.kernel.instances.sortedSet.catsKernelStdHashForSortedSet1[A]
 
   @deprecated("2.0.0-RC2", "Use cats.kernel.instances.sortedSet.catsKernelStdSemilatticeForSortedSet")
   def catsKernelStdSemilatticeForSortedSet[A: Order]: BoundedSemilattice[SortedSet[A]] =
@@ -94,7 +94,7 @@ private[instances] trait SortedSetInstancesBinCompat0 {
 private[instances] trait SortedSetInstancesBinCompat1 extends LowPrioritySortedSetInstancesBinCompat1 {
   // TODO: Remove when this is no longer necessary for binary compatibility.
   implicit override def catsKernelStdHashForSortedSet[A: Order: Hash]: Hash[SortedSet[A]] =
-    cats.kernel.instances.sortedSet.catsKernelStdHashForSortedSet[A]
+    cats.kernel.instances.sortedSet.catsKernelStdHashForSortedSet1[A]
 }
 
 private[instances] trait LowPrioritySortedSetInstancesBinCompat1
@@ -104,7 +104,7 @@ private[instances] trait LowPrioritySortedSetInstancesBinCompat1
     cats.kernel.instances.sortedSet.catsKernelStdOrderForSortedSet[A]
 
   override def catsKernelStdHashForSortedSet[A: Order: Hash]: Hash[SortedSet[A]] =
-    cats.kernel.instances.sortedSet.catsKernelStdHashForSortedSet[A]
+    cats.kernel.instances.sortedSet.catsKernelStdHashForSortedSet1[A]
 }
 
 @deprecated("2.0.0-RC2", "Use cats.kernel.instances.SortedSetHash")
--- a/kernel/src/main/scala/cats/kernel/instances/SortedSetInstances.scala
+++ b/kernel/src/main/scala/cats/kernel/instances/SortedSetInstances.scala
@@ -1,11 +1,14 @@
 package cats.kernel
 package instances
 
-import cats.kernel.{BoundedSemilattice, Hash, Order}
 import scala.collection.immutable.SortedSet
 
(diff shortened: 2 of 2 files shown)


travisbrown:
@vasiliybondarenko Adding a method to a trait breaks bincompat on Scala 2.11. Once #3051 is merged you'll be able to merge master and the bincompat check should be fine.

travisbrown:
> But these functions are ambiguous I using the same name

Personally I'd prefer to make the old over-constrained version of the method package-private if necessary in 2.1.0 rather than add a suffix to the unbroken one, but I think the overload should be fine as long as the broken one isn't implicit, right?

vasiliybondarenko:
> > But these functions are ambiguous I using the same name
> 
> Personally I'd prefer to make the old over-constrained version of the method package-private if necessary in 2.1.0 rather than add a suffix to the unbroken one, but I think the overload should be fine as long as the broken one isn't implicit, right?

Overloading for `catsKernelStdHashForSortedSet` works unless this function has no explicit usage but it has in the following traits:
`cats.instances.SortedSetInstances1`
`cats.instances.SortedSetInstancesBinCompat1`
`cats.instances.LowPrioritySortedSetInstancesBinCompat1`

So I assume the only solution is to add suffix to unbroken function with a proper constraint.

larsrh:
I took the liberty of rebasing this to current master. Let's see if this works.,

larsrh:
MiMa still complains:

```
[error]  * method this(cats.kernel.Order,cats.kernel.Hash)Unit in class cats.kernel.instances.SortedSetHash does not have a correspondent in current version
```

I think we'll have to add a deprecated extra constructor in there. I'll take care of it.

vasiliybondarenko:
> MiMa still complains:
> 
> ```
> [error]  * method this(cats.kernel.Order,cats.kernel.Hash)Unit in class cats.kernel.instances.SortedSetHash does not have a correspondent in current version
> ```
> 
> I think we'll have to add a deprecated extra constructor in there. I'll take care of it.

Cool! Thanks

core/src/main/scala/cats/instances/sortedSet.scala
```diff
@@ -94,7 +94,7 @@ private[instances] trait SortedSetInstancesBinCompat0 {
 private[instances] trait SortedSetInstancesBinCompat1 extends LowPrioritySortedSetInstancesBinCompat1 {
   // TODO: Remove when this is no longer necessary for binary compatibility.
   implicit override def catsKernelStdHashForSortedSet[A: Order: Hash]: Hash[SortedSet[A]] =
-    cats.kernel.instances.sortedSet.catsKernelStdHashForSortedSet[A]
+    cats.kernel.instances.sortedSet.catsKernelStdHashForSortedSet1[A]
```
kailuowang: @travisbrown 
Just noticed this, should this method be deprecated and un-implicit like the one above?  The `Todo` comment should probably be removed too?

core/src/main/scala/cats/instances/sortedSet.scala
```diff
@@ -104,7 +104,7 @@ private[instances] trait LowPrioritySortedSetInstancesBinCompat1
     cats.kernel.instances.sortedSet.catsKernelStdOrderForSortedSet[A]
 
   override def catsKernelStdHashForSortedSet[A: Order: Hash]: Hash[SortedSet[A]] =
-    cats.kernel.instances.sortedSet.catsKernelStdHashForSortedSet[A]
+    cats.kernel.instances.sortedSet.catsKernelStdHashForSortedSet1[A]
```
kailuowang: @travisbrown, not sure what this override is for?  Can we add a comment?

kernel/src/main/scala/cats/kernel/instances/SortedSetInstances.scala
```diff
@@ -28,9 +31,13 @@ class SortedSetOrder[A: Order] extends Order[SortedSet[A]] {
     StaticMethods.iteratorEq(s1.iterator, s2.iterator)
 }
 
-class SortedSetHash[A: Order: Hash] extends Hash[SortedSet[A]] {
+// FIXME use context bound in 3.x
+class SortedSetHash[A](implicit hashA: Hash[A]) extends Hash[SortedSet[A]] {
   import scala.util.hashing.MurmurHash3._
 
+  @deprecated("Use the constructor _without_ Order instead, since Order is not required", "2.0.1")
```
kailuowang: Nit: more likely that the derrecate version is `2.1.0`

kernel/src/main/scala/cats/kernel/instances/SortedSetInstances.scala
```diff
@@ -28,9 +31,13 @@ class SortedSetOrder[A: Order] extends Order[SortedSet[A]] {
     StaticMethods.iteratorEq(s1.iterator, s2.iterator)
 }
 
-class SortedSetHash[A: Order: Hash] extends Hash[SortedSet[A]] {
+// FIXME use context bound in 3.x
```
kailuowang: Just curious, context bound breaks BC here (even with the deprecated constructor)?

larsrh: Yes, because it affects the constructor.

core/src/main/scala/cats/instances/sortedSet.scala
travisbrown: @kailuowang What's happening here is tricky: these instances were in the wrong place (`cats.instances.SortedSetInstances1` instead of `cats.kernel.instances`), and now that they're in the right place these bincompat traits get the methods from both, which results in an `inherits conflicting members` compiler error, which means we have to bubble overrides all the way to the top.

Issue #3023, opened 2019-09-03
add unzip to Functor


julien-truffaut:
what do you think about adding

```scala
trait Functor[F[_]]{
  def unzip[A, B](fab: F[(A, B)]): (F[A], F[B]) = (fab.map(_._1), fab.map(_._2))
}
```

potentially we could have optimised version per type

reference https://hackage.haskell.org/package/base-4.12.0.0/docs/Data-List-NonEmpty.html#v:unzip

gagandeepkalra:
Is it okay if I pick this up?

stremlenye:
Wouldn't the implementation @julien-truffaut proposed invoke computation twice for the effect `F[_]`?

kubukoz:
Also concerned about this. This would probably be fine in Foldable, but for effects it'll require evaluating twice or break referential transparency.

LukaJCB:
We could call it `unzipDuplicate` or something to make this clear from the naming. WDYT? :)

kubukoz:
I'd probably like it more with a comment than that name tbh 😆 let's see what others think

kailuowang:
The effect duplication seems pretty obvious in type (2 `F[A]`s out) to me. I agree that if we want to add a note a comment will suffice.

joroKr21:
I don't think there is a way to unzip an effect without running it twice unless we do some memoization, but that's cats-effect territory.

kubukoz:
@joroKr21 even in cats-effect memoization without an additional effect breaks RT.

joroKr21:
Yes you would have to return `F(F[A], F[B])` which defeats the purpose.

gagandeepkalra:
PR merged, closing this now, thanks everyone :)

PR #3062, opened 2019-09-14
add unzip to Functor

gagandeepkalra:
addresses https://github.com/typelevel/cats/issues/3023


gagandeepkalra:
I need some help with this weird compiler situation

> [error] /home/travis/build/typelevel/cats/core/src/main/scala/cats/Functor.scala:56:21: not found: type A
[error]   def widen[A, B >: A](fa: F[A]): F[B] = fa.asInstanceOf[F[B]]
[error]                     ^
[error] /home/travis/build/typelevel/cats/core/src/main/scala/cats/Functor.scala:12:2: type arguments [C,B] do not conform to method widen's type parameter bounds [A,B >: A]
[error] @typeclass trait Functor[F[_]] extends Invariant[F] { self =>
[error]  ^
[error] two errors found

the current version won't compile, but if I just introduce a dummy parameter to unzip method, it works just fine. 

> def unzip[A, B](dummy: Int, fab: F[(A, B)]): (F[A], F[B]) = (map(fab)(_._1), map(fab)(_._2))

oleg-py:
@gagandeepkalra looks like an issue with Simulacrum-generated extension method (from `@typeclass` annotation.

Annotating `unzip` with `@noop` helps too, and I think we have to do it anyway for bincompat reasons.

gagandeepkalra:
> Annotating `unzip` with `@noop` helps too, and I think we have to do it anyway for bincompat reasons.

@oleg-py yes thank you, this fixes the compilation. I'm not so sure of the bincompat, I still get this error

> [error]  * method unzip(java.lang.Object)scala.Tuple2 in trait cats.Functor is present only in current version
[error]    filter with: ProblemFilters.exclude[[ReversedMissingMethodProblem]]("cats.Functor.unzip")
[error] java.lang.RuntimeException: Cats core: Binary compatibility check failed!

compiler suggested I add the above line to build.sbt, so I did, but I'm not sure this is the right way.

gagandeepkalra:
@LukaJCB @kailuowang some eyes on this one please

kailuowang:
Sorry I just realized the review comments I left earlier was in pending status and not actually posted. I suggested that it's odd that the mima bin compat check failed. I am curious, have you merged in the master?

gagandeepkalra:
@kailuowang @oleg-py I updated the method signature.

`def unzip[A, A1, A2](fab: F[A])(implicit asPair: A => (A1, A2)): (F[A1], F[A2])` 

This fixed the Simulacrum error, we won't need `@noop` now.

kailuowang:
> `def unzip[A, A1, A2](fab: F[A])(implicit asPair: A => (A1, A2)): (F[A1], F[A2])`

If I am not mistaken this relies on the implicit `$comply` in `Predef`. I am not familiar with this approach. But at the very least, it obscures the type signature.  Can you confirm again that the original version proposed in the issue breaks bincompat after you merged in master?

gagandeepkalra:
> But at the very least, it obscures the type signature.

yes, agreed. I was trying to avoid using `@noop` here, since we no longer need it for bincompat reasons, which got fixed with the recent master as you suggested.

travisbrown:
This looks fine to me.

gagandeepkalra:
Thank you everyone

yangzai:
(accidentally closed and reopened pr with an incomplete comment...)


@travisbrown As I was saying, the original attempt was a direct fix in the Functor trait that work with Simulacrum so it would be slightly beneficial if the intention is to move as much back to a single typeclass/trait when bincompat isn't an issue. However, having `FunctorTupleOps` per complex shape like `F[(A, B)]` seems like a defacto convention you would want to keep to. Let me know your thoughts.

On a side note regarding https://github.com/typelevel/cats/pull/3062#issuecomment-535743466 , having the type signature of `widen` changed to something like
```scala
  def widen[A, B: A <:< *](fa: F[A]): F[B] = fa.asInstanceOf[F[B]]
```
would fix(i.e. not require `@noop` for `unzip`) but that would break bincompat. Just an interesting note.

Issue #2400, opened 2018-08-14
Consider using a composition based encoding for type classes to avoid implicit collisions


LukaJCB:
We've already adopted this pattern for a few of our type classes to avoid the well known problem that inheritance based type classes bring. All of `cats-mtl`, as well as `Representable`, `Parallel` and `NonEmptyParallel` already make use of this encoding and can therefore be safely used with the other more canon type classes.

If we look closely at the [cats typeclass diagram](https://github.com/tpolecat/cats-infographic), we can see which of our type classes could potentially benefit and the important ones seem to be `Traverse`, `CoflatMap` and `Distributive`, which all extend `Functor`. Another candidate could be `Alternative`, which one might want to use together with `Monad`, though I think it's less of a problem, since you can just require `MonoidK` instead. 
And even less urgent, `Monad` currently has 3 subclasses, `Bimonad`, `MonadError` and `CommutativeMonad`. There could potentially be an issue if you wanted two of those together, but I think it's highly unlikely. 

I think we could provide this enhancement without breaking source compatibility and should think about it for our next major version :)

LukaJCB:
A fairly old PR that does part of the work can be found here: https://github.com/typelevel/cats/pull/1379

PR #3076, opened 2019-09-20
Add Align typeclass

### LINKED ISSUE #1253: Draft 0.7.0 release notes

### LINKED ISSUE #1263: Align type class

LukaJCB:
Should fix #1263 while replacing #1755 by @julianmichael who did most of the work.


build.sbt
```diff
@@ -401,6 +401,9 @@ def mimaSettings(moduleName: String) =
           exclude[MissingClassProblem](
             "cats.kernel.compat.scalaVersionMoreSpecific$suppressUnusedImportWarningForScalaVersionMoreSpecific"
           )
+        ) ++ //abstract package private classes
+        Seq(
+          exclude[DirectMissingMethodProblem]("cats.data.AbstractNonEmptyInstances.this")
```
kailuowang: Are we sure this doesn't break BC? If we are sure maybe add a test in `binCompatTest`?  If not can we just add a separate method to provide the instance for `Align`

kailuowang: I think that shall do. thanks! 
On a slightly related note, since we have been updating the build, I am a bit nervous that if something breaks `binCompatTest`, we wouldn't know. I'll create an issue.

core/src/main/scala-2.13+/cats/instances/lazyList.scala
```diff
@@ -122,6 +127,27 @@ trait LazyListInstances extends cats.kernel.instances.LazyListInstances {
...
+      def align[A, B](fa: LazyList[A], fb: LazyList[B]): LazyList[Ior[A, B]] =
+        alignWith(fa, fb)(identity)
+
+      override def alignWith[A, B, C](fa: LazyList[A], fb: LazyList[B])(f: Ior[A, B] => C): LazyList[C] = {
+        val iterA = fa.iterator
+        val iterB = fb.iterator
+
+        var result: LazyList[C] = LazyList.empty
+
+        while (iterA.hasNext || iterB.hasNext) {
```
kailuowang: This will trigger evaluation of all items right? I imagine we should be able to write a lazy implementation right?

LukaJCB: Oh does it? I haven't actually looked at how `LazyList` works at all. Help welcome :)

kailuowang: me neither. but I think you can create a new iterator something like 
```
val alighIterator = new Iterator[C] {
    def hasNext: Boolean = iterA.hasNext || iterB.hasNext
    def next(): C   = 
            val ior =
               if (iterA.hasNext && iterB.hasNext) Ior.both(iterA.next(), iterB.next())
               else if (iterA.hasNext) Ior.left(iterA.next())
               else Ior.right(iterB.next())
           f(ior)
}

LazyList.from(alignIterator)
```

LukaJCB: Thanks this seems to work just fine :)

core/src/main/scala/cats/Align.scala
```diff
@@ -0,0 +1,53 @@
...
+   * Same as `align`, but forgets from the type that one of the two elements must be present.
+   */
+  def padZip[A, B](fa: F[A], fb: F[B]): F[(Option[A], Option[B])] =
+    alignWith(fa, fb)(_.pad)
+
+  /**
+   * Same as `alignWith`, but forgets from the type that one of the two elements must be present.
+   */
+  def padZipWith[A, B, C](fa: F[A], fb: F[B])(f: (Option[A], Option[B]) => C): F[C] =
+    alignWith(fa, fb)(ior => Function.tupled(f)(ior.pad))
```
kailuowang: This and the `padZip` above seem not tested. I think a doctest will suffice.

core/src/main/scala/cats/instances/either.scala
```diff
@@ -142,6 +144,25 @@ trait EitherInstances extends cats.kernel.instances.EitherInstances {
 
       override def isEmpty[B](fab: Either[A, B]): Boolean =
         fab.isLeft
+
+      def functor: Functor[Either[A, *]] = this
+
+      def align[B, C](fa: Either[A, B], fb: Either[A, C]): Either[A, Ior[B, C]] =
```
kailuowang: This `Align` instance for `Either` seems not tested.

kailuowang:
just added some more additional comments. Sorry I couldn't find the chunk of time to review it in one go.

core/src/main/scala/cats/Align.scala
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * `Align` supports zipping together structures with different shapes,
+ * holding the results from either or both structures in an `Ior`.
+ *
+ * Must obey the laws in cats.laws.AlignLaws
+ */
+@typeclass trait Align[F[_]] {
+
+  def functor: Functor[F]
```
djspiewak: Maybe this was already discussed, but is there a particular reason to make this compositional rather than inherited? Is it because `Align` is a leaf typeclass (a la `Traverse`)?

LukaJCB: That's exactly the reason :)

djspiewak: Shouldn't we try to keep the encodings uniform though? After all, `Traverse` has this same problem, but it still uses subtyping.

LukaJCB: I think we should move away from it. We're already doing this for `FunctorFilter`, `TraverseFilter`, `Representable` and probably others :)

LukaJCB: https://github.com/typelevel/cats/issues/2400

CremboC:
Only partially related but I recently implemented very similar functionality (esp. `padZipWith`) for `fs2.Stream`:
https://gist.github.com/CremboC/e0515b5f15f17ee40a92659db78cbfd3

kailuowang:
@djspiewak in my day-to-day use, I find it more convenient having leaf type classes encoded this way than inheritance. IMO the benefit outweight the benefit of having a uniform encoding for all type classes.

build.sbt
travisbrown: This just makes me vaguely uncomfortable. It's only used twice and it seems like it'd be almost equivalently  noisy just to write out the `new AbstractNonEmptyInstances[F, NEF] with Align[NEF]` in those two places.

I know we don't promise bincompat for people using Java, etc., but breaking it for the sake of saving a few lines doesn't feel ideal.

LukaJCB:
@travisbrown I addressed your points :)

PR #3078, opened 2019-09-21
Add ScalaRL to Cats README

sritchie:
Hey all,

I haven't officially announced ScalaRL yet, but the project makes heavy use of Cats's typeclasses, so I wanted to make sure to include it here.

Thanks for everything!


LukaJCB:
Thanks!

PR #2954, opened 2019-07-19
Update sbt-mima to 0.5.0

LukaJCB:
Should supersede #2945 and #2934

PR #3079, opened 2019-09-22
update to sbt 1.3.x, uses coursier by default

larsrh:
Includes & supersedes #2954, #3075, #3077.

Open problems:
* [ ] for some reason, the BC tests won't work when Coursier is used (we can choose to ignore this)
* [ ] see #2934 for some details on the generic signature problem, but since then a bunch of additional errors surfaced (see below, blocker)

```
[error]  * static method catsParallelForEitherTNestedValidated(cats.Monad,cats.kernel.Semigroup)cats.Parallel in class cats.implicits does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("cats.implicits.catsParallelForEitherTNestedValidated")
[error]  * static method catsParallelForEitherTNestedParallelValidated(cats.kernel.Semigroup,cats.Parallel)cats.Parallel in class cats.implicits does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("cats.implicits.catsParallelForEitherTNestedParallelValidated")
[error]  * static method catsParallelForOptionTNestedOption(cats.Parallel)cats.Parallel in class cats.implicits does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("cats.implicits.catsParallelForOptionTNestedOption")
[error]  * static method catsStdEqForSortedMap(cats.kernel.Order,cats.kernel.Eq)cats.kernel.Eq in class cats.implicits does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("cats.implicits.catsStdEqForSortedMap")
[error]  * static method catsStdMonoidForSortedMap(cats.kernel.Order,cats.kernel.Semigroup)cats.kernel.Monoid in class cats.implicits does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("cats.implicits.catsStdMonoidForSortedMap")
[error]  * static method catsStdCommutativeMonoidForSortedMap(cats.kernel.Order,cats.kernel.CommutativeSemigroup)cats.kernel.CommutativeMonoid in class cats.implicits does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("cats.implicits.catsStdCommutativeMonoidForSortedMap")
[error]  * static method catsStdHashForSortedMap(cats.kernel.Hash,cats.kernel.Order,cats.kernel.Hash)cats.kernel.Hash in class cats.implicits does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("cats.implicits.catsStdHashForSortedMap")
[error]  * static method catsKernelStdSemilatticeForSortedSet(cats.kernel.Order)cats.kernel.BoundedSemilattice in class cats.implicits does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("cats.implicits.catsKernelStdSemilatticeForSortedSet")
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -63,6 +63,8 @@ lazy val commonScalaVersionSettings = Seq(
 
 commonScalaVersionSettings
 
+ThisBuild / mimaFailOnNoPrevious := false
+
 lazy val commonSettings = commonScalaVersionSettings ++ Seq(
   scalacOptions ++= commonScalacOptions(scalaVersion.value),
   Compile / unmanagedSourceDirectories ++= scalaVersionSpecificFolders("main", baseDirectory.value, scalaVersion.value),
@@ -390,7 +392,8 @@ def mimaSettings(moduleName: String) =
           exclude[IncompatibleMethTypeProblem]("cats.arrow.FunctionKMacros.lift"),
           exclude[MissingTypesProblem]("cats.arrow.FunctionKMacros$"),
           exclude[IncompatibleMethTypeProblem]("cats.arrow.FunctionKMacros#Lifter.this"),
-          exclude[IncompatibleResultTypeProblem]("cats.arrow.FunctionKMacros#Lifter.c")
+          exclude[IncompatibleResultTypeProblem]("cats.arrow.FunctionKMacros#Lifter.c"),
+          exclude[DirectMissingMethodProblem]("cats.arrow.FunctionKMacros.compatNewTypeName")
         ) ++ //package private classes no longer needed
         Seq(
           exclude[MissingClassProblem]("cats.kernel.compat.scalaVersionMoreSpecific$"),
@@ -398,8 +401,10 @@ def mimaSettings(moduleName: String) =
           exclude[MissingClassProblem](
             "cats.kernel.compat.scalaVersionMoreSpecific$suppressUnusedImportWarningForScalaVersionMoreSpecific"
           )
+        ) ++ // Only narrowing of types allowed here
+        Seq(
+          exclude[IncompatibleSignatureProblem]("*")
         )
-
     }
   )
 
@@ -652,9 +657,12 @@ lazy val bench = project
   .enablePlugins(JmhPlugin)
 
 lazy val binCompatTest = project
-  .disablePlugins(CoursierPlugin)
   .settings(noPublishSettings)
   .settings(
+    // workaround because coursier doesn't understand dependsOn(core.jvm % Test)
+    // see https://github.com/typelevel/cats/pull/3079#discussion_r327181584
+    // see https://github.com/typelevel/cats/pull/3026#discussion_r321984342
+    useCoursier := false,
     commonScalaVersionSettings,
     addCompilerPlugin("org.typelevel" %% "kind-projector" % kindProjectorVersion),
     libraryDependencies ++= List(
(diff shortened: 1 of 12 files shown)


kailuowang:
I am on my phone. I wonder what happened to those methods. Will take a look them Monday

build.sbt
```diff
@@ -652,9 +656,9 @@ lazy val bench = project
   .enablePlugins(JmhPlugin)
 
 lazy val binCompatTest = project
-  .disablePlugins(CoursierPlugin)
   .settings(noPublishSettings)
   .settings(
+    useCoursier := false,
```
fthomas: @larsrh Is this a workaround for the problem described in [#3026](https://github.com/typelevel/cats/pull/3026/commits/4ee38b22a2455c0a812b9bc86ef70e835744fcaa)? If yes, we could just bump sbt and deal with the MiMa update separately.

larsrh: Ah, I somehow didn't see that PR. But I can try to incorporate it here.

fthomas: That isn't necessary. The only change I did there that is not in this PR, is a change we don't want to merge. :-) I just wanted to note that we could bump sbt separately with the `useCoursier := false` workaround.

larsrh: Wait, now I'm confused. Is this PR good to merge or should I change anything?

larsrh: Oh, btw, we can't bump sbt without also bumping MiMa (there's some binary breakage there).

fthomas: Interesting, I didn't notice this in #3026.

larsrh: Try bumping to 1.3.2. At least I thought that's what happened in another PR, but maybe I was mistaken :thinking:

fthomas: You're right. Bumping from 1.3.0 to 1.3.2 produced errors like this in the "binary compatibility" job:
```
(coreJVM / mimaReportBinaryIssues) java.lang.NoSuchMethodError: com.typesafe.tools.mima.core.Settings.allSettings()Lscala/collection/mutable/HashSet;
```

kailuowang:
Okay there are indeed BC breakage that were caught be previous Mima versions. 
Discussion can be found [here](https://github.com/lightbend/mima/pull/375)
TLDR; making a public method package private in an object (in our case `cats.implicts` breaks BC for Java usage. So these were not caught in the wide because only java usage exposes them. 
Looks like we need to remove the package private on these methods. Shall we do them in this PR? Then we probably want to cherry pick to 2.0.x branch and release a patch.

larsrh:
Are those supposed to be deprecated? I can do that in this PR.

kailuowang:
Yes. Those methods were moved to somewhere else. They are kept at the original place deprecated to maintain BC. Making them package private was an extra measure (now proven invalid) for the deprecation.

build.sbt
kailuowang: Coursier was disabled because it seemed to me that it doesn't understand this setting:
`dependsOn(core.jvm % Test)`

I should've added a comment. Can we add one now?

larsrh:
If the build is green, I'll rebase & reorder for easier cherry-picking.

larsrh:
Ready for review.

fthomas:
Could you also bump the sbt version in `scalafix/project/build.properties`?

PR #3080, opened 2019-09-22
Improve performance of Kleisli: make it a value class

svalaskevicius:
Make Kleisli a value class, to avoid the runtime cost of new object instantiation.

`sbt 'bench/jmh:run "cats.bench.KleisliBench.*" -f1 -t1 -wi 2 -i2' `

before:
```
[info] Benchmark                      Mode  Cnt       Score   Error  Units
[info] KleisliBench.baselineFlatMap  thrpt    2  299375.289          ops/s
[info] KleisliBench.flatMap          thrpt    2  127897.572          ops/s
```

after:
```
[info] Benchmark                      Mode  Cnt       Score   Error  Units
[info] KleisliBench.baselineFlatMap  thrpt    2  294876.092          ops/s
[info] KleisliBench.flatMap          thrpt    2  155593.301          ops/s
```

### DIFF
--- a/bench/src/main/scala/cats/bench/KleisliBench.scala
+++ b/bench/src/main/scala/cats/bench/KleisliBench.scala
@@ -1,0 +1,30 @@
+package cats.bench
+
+import cats.Eval
+import cats.data.Kleisli
+import org.openjdk.jmh.annotations.{Benchmark, Scope, State}
+
+@State(Scope.Benchmark)
+class KleisliBench {
+
+  @Benchmark
+  def baselineFlatMap(): Long = {
+    def pure(x: Long) = Eval.now(x)
+    val count = 100
+    val result = (0 until count).foldLeft(pure(0)) { (acc, x) =>
+      acc.flatMap(y => pure(x+y))
+    }
+    result.value
+  }
+
+  @Benchmark
+  def flatMap(): Long = {
+    def pure(x: Long) = Kleisli.pure[Eval, Unit, Long](x)
+    val count = 100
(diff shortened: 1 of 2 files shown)


svalaskevicius:
Hmm binary compatiblity fails - expectedly tbh as some functions are renamed.. (otherwise the compiler is not happy)

Could this be part of a major version change then? Or does anyone have any better ideas?

djspiewak:
Hmm, could you leave the old functions as `private`? The benchmarks are impressive. It is worth noting that in a few cases, this will actually make performance *worse* (for example, `List[Kleisli[F, R, A]]`, which will result in twice as much boxing as without the value class), but I think those cases are relatively rare.

kailuowang:
I think it's not just a few renames. It seems to me that changing this to a value class itself breaks BC. We still don't have a plan on how we do a binary breaking version on cats-core yet but we are working on it.

svalaskevicius:
Thanks both! Yeah I don't think private will help - the error I got from compiler is that when it bacame a value class the functions where too similar for it to resolve. We can try though if you think it might help :) plus the reason for bin compat failure is likely to be the change to value class as @kailuowang wrote..

Out of curiosity - @djspiewak - why would a list box it twice? I've checked https://docs.scala-lang.org/overviews/core/value-classes.html and one case is actually a bit worrying - apparently even a call to identity will instantiate it (what about tagless final usage?)

armanbilge:
We can't make this change bin-compatibly in Cats 2, and hopefully by Cats 3 can use Scala 3's `opaque type` :)

Issue #2347, opened 2018-07-28
Reenable full build on 2.13


kailuowang:
Right now several steps in build on 2.13-M4 were disabled temporarily to help release 2.13-M4 artifacts quickly 

- [ ]  code coverage disabled, due to lack of scoverage on 2.13-M4  [blocking issue](https://github.com/scoverage/scalac-scoverage-plugin/issues/225#issuecomment-405698919)
- [x] deprecation were ignored. need to find a way to cross-compile deprecated collection API, for now, it's just too much work that I don't have the time to do. https://github.com/typelevel/cats/issues/2893
- [ ] tut docs are disable also due to the deprecated collection APIs. 
- [ ] doctests are disabled, due to deprecated collection API and different toString implementations in collections. 
- [ ] scaladoc is disabled due to simulacrum stopped working on scaladoc compile on 2.13, possibly due to a scala macro bug reported to https://github.com/scala/bug/issues/11045

We should probably achieve these with multiple PRs.
If anyone is interested in picking one up, please announce it here.

SethTisue:
should this stay open? is any of this still pending?

kailuowang:
Only one of the 5 was resolved. I will take a look of them later.

DieBauer:
Scaladoc and scoverage can be enabled. tut and doctest still have issues. Tut, as mentioned with the deprecation warnings and -Werror flag failing the build. 
doctest, for what I can see, unused imports in some snippets.

I'll create a PR for the first two.

DieBauer:
I see two things, one is enable the 'full' build in sbt. e.g., all tasks can execute for both supported scala versions. The other is that the CI (travis) should be enabled for both versions. 

For travis I can see that cross-compiling scalaJs, and scalaFmt should be enabled. I can create a PR for that.

PR #3096, opened 2019-10-02
Enable coverage and docs

DieBauer:
Addresses two points of https://github.com/typelevel/cats/issues/2347

Enable scoverage for 2.13 and enable scaladoc generation for 2.13. 

One caveat with scoverage is that, for scala 2.13.1 (which Cats currently doesn't use), there are issues as described here https://github.com/scoverage/sbt-scoverage/issues/295 and blocking an upgrade. 

This might be decisive in dropping that commit from this PR, let me know.


kailuowang:
As of now on travis we only run coverage on Scala 2.12 using this [job](https://github.com/typelevel/cats/blob/master/.travis.yml#L32-L35)

With this change we shall be able to delete this coverage dedicated job and run coverage for both 2.12 and 2.13 by changing the jvm tests job to do coverage as well.  I.e. change the following 
https://github.com/typelevel/cats/blob/master/.travis.yml#L47 
to 
```
      install: pip install --user codecov
      script: sbt ++$TRAVIS_SCALA_VERSION!  coverage buildJVM bench/test coverageReport && codecov
```

DieBauer:
@kailuowang thanks for the remark, I forgot about travis. 

I'm reading the PR (https://github.com/typelevel/cats/pull/2737) that added the 2.12 coverage job, and there it states that running coverage reports for multiple scala versions is 'undefined'. 

So I'm not entirely sure if this is what we want to enable here. What are your thoughts?

Also, regarding publishing API docs for 2.12 and 2.13 does anything need to change? I'm not familiar with microsite. In my projects I'm doing `+unidoc` and move the /target/docs/scala-x to a server.

kailuowang:
My thought was enabling coverage on both Scala versions makes some sense. True we will only get one codecov report, but I believe we have coverage threshold check, which fails the build if the delta coverage falls below (on the other hand I haven't seen that in action for a while, so maybe we need to check the settings, not necessarily in this PR though). For those coverage checks it kind of makes sense to have them on both Scala versions. 

Unfortunately our microsite is tied to only one Scala version. I propose we change it to Scala 2.13, although I am not 100% sure if sbt-microsite supports it yet.

DieBauer:
Ok that makes sense. Not sure what the correct place would be in the travis job though

It seems that `bench/test` works on 2.13.0 as well. So the comment (https://github.com/typelevel/cats/blob/master/.travis.yml#L51) is not valid anymore and we can merge the jobs
with something like this?
```yml
    - &jvm_tests
      stage: test
      env: TEST="JVM tests"
      script: sbt ++$TRAVIS_SCALA_VERSION! buildJVM bench/test
      scala:
      - *scala_version_212
      - *scala_version_213
```

Where would coverage reporting then fit in `&& codecov`? in this job, or still in the separate job that you mentioned?

kailuowang:
that's good news. 
yeah, let's put the coverage in the `jvm_tests` jobs

build.sbt
```diff
@@ -6,12 +6,7 @@ import scala.xml.transform.{RewriteRule, RuleTransformer}
 import sbtcrossproject.CrossPlugin.autoImport.{crossProject, CrossType}
 
 lazy val scoverageSettings = Seq(
-  coverageEnabled := {
-    if (priorTo2_13(scalaVersion.value))
-      coverageEnabled.value
-    else
-      false
-  },
+  coverageEnabled := coverageEnabled.value,
```
kailuowang: Nit, we don't need this line right?

build.sbt
```diff
@@ -77,11 +72,7 @@ lazy val commonSettings = commonScalaVersionSettings ++ Seq(
     val unchanged = doctestGenTests.value
     if (priorTo2_13(scalaVersion.value)) unchanged else Nil
   },
-  // TODO: re-enable disable scaladoc on 2.13 due to https://github.com/scala/bug/issues/11045
-  sources in (Compile, doc) := {
-    val docSource = (sources in (Compile, doc)).value
-    if (priorTo2_13(scalaVersion.value)) docSource else Nil
-  }
+  sources in (Compile, doc) := (sources in (Compile, doc)).value
```
kailuowang: This line is not doing anything.

build.sbt
DieBauer: Yes correct :)

build.sbt
DieBauer: Oops, fixed

DieBauer:
Fixed merge conflict with master.

PR #3103, opened 2019-10-09
Add TraverseFilter instance for Queue.

takayahilton:

### DIFF
--- a/core/src/main/scala/cats/instances/queue.scala
+++ b/core/src/main/scala/cats/instances/queue.scala
@@ -152,4 +152,33 @@ trait QueueInstances extends cats.kernel.instances.QueueInstances {
       def show(fa: Queue[A]): String =
         fa.iterator.map(_.show).mkString("Queue(", ", ", ")")
     }
+
+  implicit def catsStdTraverseFilterForQueue: TraverseFilter[Queue] = QueueInstances.catsStdTraverseFilterForQueue
+}
+
+private object QueueInstances {
+  private val catsStdTraverseFilterForQueue: TraverseFilter[Queue] = new TraverseFilter[Queue] {
+    val traverse: Traverse[Queue] = cats.instances.queue.catsStdInstancesForQueue
+
+    override def mapFilter[A, B](fa: Queue[A])(f: (A) => Option[B]): Queue[B] =
+      fa.collect(Function.unlift(f))
+
+    override def filter[A](fa: Queue[A])(f: (A) => Boolean): Queue[A] = fa.filter(f)
+
+    override def collect[A, B](fa: Queue[A])(f: PartialFunction[A, B]): Queue[B] = fa.collect(f)
+
+    override def flattenOption[A](fa: Queue[Option[A]]): Queue[A] = fa.flatten
+
+    def traverseFilter[G[_], A, B](fa: Queue[A])(f: (A) => G[Option[B]])(implicit G: Applicative[G]): G[Queue[B]] =
+      fa.foldRight(Eval.now(G.pure(Queue.empty[B])))(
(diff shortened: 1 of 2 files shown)


takayahilton:
I tried to add `catsStdTraverseFilterForQueue` to `QueueInstances`, but still Mima fails with `ReversedMissingMethodProblem`.
Could you tell me where should I add it?

core/src/main/scala/cats/instances/queue.scala
```diff
@@ -152,4 +152,33 @@ trait QueueInstances extends cats.kernel.instances.QueueInstances {
       def show(fa: Queue[A]): String =
         fa.iterator.map(_.show).mkString("Queue(", ", ", ")")
     }
+
+  implicit def catsStdTraverseFilterForQueue: TraverseFilter[Queue] = QueueInstances.catsStdTraverseFilterForQueue
```
takayahilton: I don't know why , but `implicit val catsStdTraverseFilterForQueue` breaks binary compatibility.

```
[error]  * abstract synthetic method cats$instances$QueueInstances$_setter_$catsStdTraverseFilterForQueue_=(cats.TraverseFilter)Unit in interface cats.instances.QueueInstances is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("cats.instances.QueueInstances.cats$instances$QueueInstances$_setter_$catsStdTraverseFilterForQueue_=")
[error]  * abstract method catsStdTraverseFilterForQueue()cats.TraverseFilter in interface cats.instances.QueueInstances is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("cats.instances.QueueInstances.catsStdTraverseFilterForQueue")
```

core/src/main/scala/cats/instances/queue.scala
```diff
@@ -152,4 +152,33 @@ trait QueueInstances extends cats.kernel.instances.QueueInstances {
...
+}
+
+private object QueueInstances {
+  private val catsStdTraverseFilterForQueue: TraverseFilter[Queue] = new TraverseFilter[Queue] {
```
takayahilton: This is just an instance cache.

kailuowang: I don't see this pattern as problematic but it's not very consistent with rest of the code base, can we just make the one in the trait a `val`?

takayahilton: I get an error that 
```
[error]  * abstract synthetic method cats$instances$QueueInstances$_setter_$cats$instances$QueueInstances$$_catsStdTraverseFilterForQueue_=(cats.TraverseFilter)Unit in interface cats.instances.QueueInstances is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("cats.instances.QueueInstances.cats$instances$QueueInstances$_setter_$cats$instances$QueueInstances$$_catsStdTraverseFilterForQueue_=")
[error]  * abstract synthetic method cats$instances$QueueInstances$$_catsStdTraverseFilterForQueue()cats.TraverseFilter in interface cats.instances.QueueInstances is present only in current version
```

kailuowang: how did you run the check? did you get that error from `sbt validateBC`?

takayahilton: Yes, I ran `sbt validateBC`.

travisbrown: Adding a `val` to a trait will break bincompat on all Scala versions so far because of the synthetic setter that's necessary because of the way trait initialization works. The workaround here looks reasonable to me.

PR #3107, opened 2019-10-11
Kleisli tapWith default mapping functions

bpholt:
My goal for this PR is to be able to write something like this (without providing a lot of explicit type information or identity-like mapping functions):

```scala
x.tapWith().mapF(_.use((runProgram[F] _).tupled))
```

where `x: Kleisli[Resource[F, *], CaseClass, EitherK[…] ~> F]` and `runProgram[F]: (CaseClass, EitherK[…] ~> F) => F[Unit]`

Before this PR, I think the above line would look like

```scala
x.tapWith[(CaseClass, EitherK[…] ~> F), CaseClass]((a, b) => (a, b))
  .mapF(_.use((runProgram[F] _).tupled))
```

This PR provides two implementations; the expectation is that one would be dropped before accepting the PR. 

One implementation adds a new `tapWith` method added directly on `Kleisli`. It has some type inference problems, because scalac isn't able to infer the `AA <: A` type. This means callers have to provide the `A` type, which seems redundant since it's already known to the `Kleisli`:

```scala
x.tapWith[CaseClass]().mapF(_.use((runProgram[F] _).tupled))
```

The other implementation is an enhancement method to `Kleisli`. It has the advantage of improved type inference. I couldn't find any other enhancement methods on `Kleisli`, though, so it is stylistically different. It also lacks the ability for the caller to narrow the type, but if they really need to do that they can still fall back to the existing `tapWith` method.

I picked the name `tapWithIdentity` for the enhancement method because I think it had to be something that didn't already exist as a method on `Kleisli`. I'm definitely open to suggestions for a better name.

@kailuowang and I [discussed this briefly on gitter](https://gitter.im/typelevel/cats-dev?at=5d9e9d2e874aeb2d231c31bb) too.

My initial preference is that the enhancement methods be added and we drop 
c867dd6, but I'm looking for feedback from maintainers.

### DIFF
--- a/core/src/main/scala/cats/data/Kleisli.scala
+++ b/core/src/main/scala/cats/data/Kleisli.scala
@@ -1,7 +1,6 @@
 package cats
 package data
 
-import cats.{Contravariant, Id}
 import cats.arrow._
 
 /**
@@ -107,10 +106,27 @@ final case class Kleisli[F[_], -A, B](run: A => F[B]) { self =>
   def tap[AA <: A](implicit F: Functor[F]): Kleisli[F, AA, AA] =
     Kleisli(a => F.as(run(a), a))
 
-  /** Yield computed B combined with input value. */
+  /**
+   * Yield computed B combined with input value.
+   *
+   * {{{
+   * scala> import cats.data.Kleisli, cats.implicits._
+   * scala> val base = Kleisli((i: Int) => (i + 1).pure[Option])
+   * scala> base.tapWith[String, Int]((a, b) => (a, b).toString).run(42)
+   * res0: Option[String] = Some((42,43))
+   * }}}
+   */
   def tapWith[C, AA <: A](f: (AA, B) => C)(implicit F: Functor[F]): Kleisli[F, AA, C] =
     Kleisli(a => F.map(run(a))(b => f(a, b)))
 
+  /**
+   * {{{
+   * scala> import cats.data.Kleisli, cats.implicits._
+   * scala> val base = Kleisli((i: Int) => (i + 1).pure[Option])
+   * scala> base.tapWithF[String, Int]((a, b) => Option((a, b).toString)).run(42)
+   * res0: Option[String] = Some((42,43))
+   * }}}
+   */
   def tapWithF[C, AA <: A](f: (AA, B) => F[C])(implicit F: FlatMap[F]): Kleisli[F, AA, C] =
     Kleisli(a => F.flatMap(run(a))(b => f(a, b)))
 
@@ -252,7 +268,7 @@ sealed private[data] trait KleisliFunctionsBinCompat {
    * Lifts a natural transformation of effects within a Kleisli
    * to a transformation of Kleislis.
    *
-   * Equivalent to running `mapK(f) on a Kleisli.
+   * Equivalent to running `mapK(f)` on a Kleisli.
    *
    * {{{
    * scala> import cats._, data._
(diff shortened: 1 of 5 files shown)


core/src/main/scala/cats/data/Kleisli.scala
```diff
@@ -107,13 +106,43 @@ final case class Kleisli[F[_], -A, B](run: A => F[B]) { self =>
...
+   * Yield computed B combined with input value.
+   *
+   * {{{
+   * scala> import cats.data.Kleisli, cats.implicits._
+   * scala> val base = Kleisli((_: Int).toString.pure[Either[Throwable, *]])
+   * scala> base.tapWith[Int]().run(42)
+   * res0: scala.util.Either[Throwable,(Int, String)] = Right((42,42))
+   * }}}
+   */
+  def tapWith[AA <: A]()(implicit F: Functor[F]): Kleisli[F, AA, (AA, B)] =
```
LukaJCB: I'm not a fan of overloading functions in this way. Often the type inference will struggle more when doing so. Maybe we can find a different name? `tapProduct`? I don't know :smile:

core/src/main/scala/cats/data/Kleisli.scala
```diff
@@ -281,6 +310,36 @@ sealed private[data] trait KleisliExplicitInstances {
 }
 
 sealed abstract private[data] class KleisliInstances extends KleisliInstances0 {
+  implicit class KleisliTapWith[F[_], A, B](private val kl: Kleisli[F, A, B]) {
```
LukaJCB: I'm not sure, why you placed this here? This trait is meant for type class instances not syntax extensions :)

bpholt: Fair enough. I couldn't find any syntax extensions for `Kleisli` so I just shoehorned it in. I'm happy to add a new trait if that's the right way to do it.

core/src/main/scala/cats/data/Kleisli.scala
bpholt: Honestly I'd prefer to drop these and go with the enhancement method implementation anyway, where I named the functions `tapWithIdentity`. I'm open to using `tapProduct` there though.

core/src/main/scala/cats/data/Kleisli.scala
LukaJCB: I think you should be able to just add them to Kleisli directly, since we dropped 2.11 :)

bpholt: Do you have a suggestion for working around the type inference problems described above? AFAICT the method can't use `A` directly, it has to use `AA <: A` due to the variance at the class level. That's the reason I'd prefer to use enhancement methods for these functions.

bpholt: I ended up moving this to the `cats.syntax` package.

bpholt:
I dropped the implementation directly on `Kleisli` due to the type inference concerns, and moved the enhancement methods to a new object in `cats.syntax`. I think this PR is ready for a real review now.

travisbrown:
I'm :-1:  on this approach, primarily because because we already have `tapWith` and `tapWithF` with `AA <: A` that expect users either to provide the type parameters or the types for their function arguments there, and having similarly named enrichment methods that take a different approach is confusing, hurts discoverabilty, etc. More generally, I'm not sure there's _any_ precedent in Cats for using enrichment methods to improve inference?

I'm also reluctant to fix the inconsistency for going in the other direction and doing this for other methods, because I think we should be aiming to reduce the amount of implicit stuff ordinary usage drags into scope, not the opposite.

djspiewak:
> More generally, I'm not sure there's any precedent in Cats for using enrichment methods to improve inference?

Not in cats, but fs2 does this all the time. It does hurt discoverability a bit, but the inference is very noticeably better than it would be in other cases.

larsrh:
I'm going to close this because there was no consensus on this change, and there hasn't been activity in over a year. If there's renewed interest, feel free to reopen.

PR #3084, opened 2019-09-24
Add minimumBy/maximumBy/Option to Foldable

joroKr21:
* `mimimumByOption` / `maximumByOption` on `Foldable`
  * `minumumBy` / `maximumBy` on `Reducible`

Thank you for contributing to Cats!

Fixes #2385

PR #3122, opened 2019-10-28
Follow-up #3084

travisbrown:
This is a follow-up for #3084, with three changes, two of which should be uncontroversial:

* These are methods on the `Foldable` and `Reducible` type classes, and shouldn't have `Foldable` or `Reducible` constraints, so I've removed them.
* I fixed the references in the first `@see` clauses in the API docs for the `Foldable` methods (they should point to the `*By` methods on `Reducible`).

I've also changed the method names from `minimumOptionBy` and `maximumOptionBy` to `minimumByOption` and `maximumByOption`. This is consistent with the standard library's `maxByOption`, and also with e.g. `reduceRightOption` (from both the standard library and Cats). 

Following the standard library here seems like the right thing to do, and these new methods haven't been in a release yet, so we don't need to worry about compatibility. If people would prefer to keep the `OptionBy` versions I can remove that commit, though.

/cc @joroKr21

### DIFF
--- a/core/src/main/scala/cats/Foldable.scala
+++ b/core/src/main/scala/cats/Foldable.scala
@@ -204,8 +204,8 @@ import Foldable.sentinel
    *
    * @see [[maximumOptionBy]] for maximum instead of minimum.
    */
-  def minimumOptionBy[A, B: Order](fa: F[A])(f: A => B)(implicit F: Foldable[F]): Option[A] =
-    F.minimumOption(fa)(Order.by(f))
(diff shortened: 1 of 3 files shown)


joroKr21:
Ohh, I think I see what happened here. Initially added them as bincompat syntax I think, then realized we can have them in the typeclasses after dropping 2.11, but I completely forgot to remove the constraints.

I wonder if `-Ywarn-unused:params` would've helped here.

Regardless, thank you @travisbrown!

I'm ok with the name change as well.

travisbrown:
@joroKr21 Ah, that makes sense!

PR #3124, opened 2019-11-04
SemigroupK sum: F[A], F[B] => F[A Either B]

### LINKED ISSUE #3112: Add sum: (F[A], F[B]) => F[A Either B]
Hey, I have

```
type X[T] = Kleisli[OptionT[F, *], Foo, T]
```
and would like to go from `(X[A], X[B])` to `X[Either[A, B]]`. 

So similar to what `<+>` does (e.g. in http4s) but on two different types joined by `Either`. Right now I `map` and use `<+>` afterwards.

Signature proposed by @LukaJCB :

```
def sum[F[_]: SemigroupK, Functor, A, B](fa: F[A], fb: F[B]): F[A Either B]
```

pk044:
Closes #3112

### DIFF
--- a/core/src/main/scala/cats/SemigroupK.scala
+++ b/core/src/main/scala/cats/SemigroupK.scala
@@ -69,15 +69,15 @@ import simulacrum.typeclass
     }
 
   /**
-    * Combines `F[A]` and `F[B]` into a `F[Either[A,B]]]`.
-    *
-    * Example:
-    * {{{
-    * scala> import cats.SemigroupK
-    * scala> SemigroupK[NonEmptyList].sum(NonEmptyList.one(1), NonEmptyList.one(2))
-    * res0: cats.data.NonEmptyList[Either[Int,Int]] = NonEmptyList(Left(1), Right(2))
-    * }}}
-    */
-  def sum[A, B](fa: F[A], fb: F[B])(implicit F: Functor[F]): F[Either[A,B]] =
+   * Combines `F[A]` and `F[B]` into a `F[Either[A,B]]]`.
+   *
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/SemigroupK.scala
```diff
@@ -67,4 +67,18 @@ import simulacrum.typeclass
...
     }
+
+  /**
+   * Combines `F[A]` and `F[B]` into a `F[Either[A,B]]]`.
+   *
+   * Example:
+   * {{{
+   * scala> import cats.SemigroupK
+   * scala> import cats.data.NonEmptyList
+   * scala> SemigroupK[NonEmptyList].sum(NonEmptyList.one(1), NonEmptyList.one(2))
```
travisbrown: This is nitpicky, but what do you think about using different types for this example? Otherwise looks good to me.

pk044: I was hesitant about using NonEmptyList here, but I wasn't sure about the alternatives either, so I stayed with NEL.

do you perhaps have any suggestions which type could be used here instead? :)

travisbrown: @pk044 Oh, sorry, that wasn't very clear. I think `NonEmptyList` is a good choice, just that using different values for `A` and `B` might make it a little clearer—e.g. something like:
```scala
SemigroupK[NonEmptyList].sum(NonEmptyList.one("abc"), NonEmptyList.one(2))
```

pk044: ok, i changed the value :+1:

PR #3127, opened 2019-11-05
Add instances for the standard library's TailRec

travisbrown:
This takes @johnynek's commits from #3041 and adds one commit changing the `val` in `TailRecInstances` to a method pointing to a cached instance, to avoid breaking bincompat.


travisbrown:
Updated branch in #3041.

PR #3128, opened 2019-11-05
Deprecate / override syntax methods with redundant Monad constraints

travisbrown:
This does the same thing as @ceedub's #2180 but in a binary-compatible way.

I was afraid we might run into the same bug as [here](https://github.com/typelevel/cats/issues/2685#issuecomment-547862937) but the syntax tests are still passing.


travisbrown:
Oh, wait, I could have sworn I ran `validateBC` on this. Taking a look now.

travisbrown:
I only checked bincompat on 2.13, where the synthetic methods for overrides in value classes aren't suffixed with an index, and this change doesn't break bincompat. They are indexed on 2.12, which means there's no way to do this there, as far as I can tell.

PR #3129, opened 2019-11-06
Check bincompat on stuff we should be checking bincompat for

### LINKED ISSUE #3109: Enable Binary compatibility tests for scala 2.13
This includes `binCompatTest` as well as mima tests

travisbrown:
Fixes #3109. I've also added bincompat checking (currently against 2.0.0 only) for cats-laws, cats-kernel-laws, cats-testkit, alleycats-core, and alleycats-laws.

One piece of breakage slipped into the 2.13 build in #3099. I've verified that Travis CI fails with just the first two commits.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,10 +60,13 @@ jobs:
       env: TEST="scalafix"
       script: cd scalafix && sbt tests/test
 
-    - stage: test
+    - &bincompat
+      stage: test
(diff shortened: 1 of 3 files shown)


travisbrown:
(As a side note, you can check this with `show mimaPreviousArtifacts`, and the results now look reasonable to me, on both 2.12 and 2.13.)

PR #3130, opened 2019-11-06
change constraint for foldMapM from Monad to Applicative

mberndt123:
Applicative is enough for this method, so it shouldn't require Monad.

### DIFF
--- a/core/src/main/scala/cats/Foldable.scala
+++ b/core/src/main/scala/cats/Foldable.scala
@@ -334,6 +334,14 @@ import Foldable.sentinel
     foldM(fa, B.empty)((b, a) => G.map(f(a))(B.combine(b, _)))
 
   /**
+   * Equivalent to foldMapM.
+   * The difference is that foldMapA only requires G to be an Applicative
+   * rather than a Monad. It is also slower due to use of Eval.
+   */
+  def foldMapA[G[_], A, B](fa: F[A])(f: A => G[B])(implicit G: Applicative[G], B: Monoid[B]): G[B] =
+    foldRight(fa, Eval.now(G.pure(B.empty)))((a, egb) => G.map2Eval(f(a), egb)(B.combine)).value
+
+  /**
    * Traverse `F[A]` using `Applicative[G]`.
    *
    * `A` values will be mapped into `G[B]` and combined using
--- a/tests/src/test/scala/cats/tests/FoldableSuite.scala
+++ b/tests/src/test/scala/cats/tests/FoldableSuite.scala
@@ -280,6 +280,18 @@ class FoldableSuiteAdditional extends CatsSuite with ScalaVersionSpecificFoldabl
       (Some(x): Option[String])
     }
     assert(sumMapM == Some("AaronBettyCalvinDeirdra"))
+
+    // foldMapM should short-circuit and not call the function when not necessary
+    val f = (_: String) match {
+      case "Calvin" => None
+      case "Deirdra" =>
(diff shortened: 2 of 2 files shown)


travisbrown:
This could probably be done in a binary-compatible way by making the current implementation package private and adding the new one as an overload. I agree with @LukaJCB that it probably makes more sense to add a new `foldMapA` with the appropriate constraint instead.

It seems like the bigger issue with this implementation is short circuiting. For example, this will hang forever with the `foldMap` implementation:

```scala
import cats.implicits._

Stream.iterate(0)(_ + 1).foldMapM(b => if (b > 10) Left(b) else Right(b))
```
I think an implementation like this should be fine both in terms of stack safety and short-circuiting:

```scala
def foldMapA[G[_], A, B](fa: F[A])(f: A => G[B])(implicit G: Applicative[G], B: Monoid[B]): G[B] =
  foldRight(fa, Eval.now(G.pure(B.empty)))((a, egb) =>
    G.map2Eval(f(a), egb)(B.combine)
  ).value
```

For example:

```scala
import cats.implicits._

scala> Stream.iterate(0)(_ + 1).foldMapA(b => if (b > 10) Left(b) else Right(b))
res0: scala.util.Either[Int,Int] = Left(11)

scala> Stream.iterate(0)(_ + 1).foldMapA(b => if (b > 1000000) Left(b) else Right(b))
res1: scala.util.Either[Int,Int] = Left(1000001)
```

mberndt123:
Hey there,

Thanks for the review! I didn't know about map2Eval, that's very interesting.

I've now added `foldMapA` as a new method and also a test to make sure the short-circuiting behaviour is preserved.

core/src/main/scala/cats/Foldable.scala
```diff
@@ -333,6 +333,14 @@ import Foldable.sentinel
   def foldMapM[G[_], A, B](fa: F[A])(f: A => G[B])(implicit G: Monad[G], B: Monoid[B]): G[B] =
     foldM(fa, B.empty)((b, a) => G.map(f(a))(B.combine(b, _)))
 
+  /**
+   * Equivalent to foldMapM.
+   * The difference is that foldMapA only requires G to be an Applicative
+   * rather than a Monad. It is also slower due to use of Eval.
```
travisbrown: Do we actually know that this is true? My intuition is that it generally will be, but there's also some `Either` overhead involved in the `tailRecM` implementation.

travisbrown:
I put together a little benchmark with a vector with 10k elements with examples that both short-circuit in `Either` and complete successfully, and yeah, the `foldMapM` implementation is a lot faster:

```
[info] Benchmark                            Mode  Cnt     Score     Error  Units
[info] FoldMapABench.foldMapA              thrpt   20  2113.567 ± 137.868  ops/s
[info] FoldMapABench.foldMapM              thrpt   20  3445.172 ± 131.904  ops/s
[info] FoldMapABench.foldMapAShortCircuit  thrpt   20  5476.648 ± 289.969  ops/s
[info] FoldMapABench.foldMapMShortCircuit  thrpt   20  6699.230 ± 259.521  ops/s
```

LukaJCB:
Let's update the scaladoc and get this merged :)

mberndt123:
@LukaJCB I'm not quite sure what more I'm supposed to add for scaladoc. Isn't it enough to say that this does the same thing as `foldMapM`?

LukaJCB:
Sorry had a brainfart there :smile:

PR #3131, opened 2019-11-07
Backporting ifF for scala 2.11

sdrafahl:
Backporting ifF for scala 2.11.

### DIFF
--- a/core/src/main/scala/cats/Functor.scala
+++ b/core/src/main/scala/cats/Functor.scala
@@ -144,20 +144,6 @@ import simulacrum.typeclass
    */
   def tupleRight[A, B](fa: F[A], b: B): F[(A, B)] = map(fa)(a => (a, b))
 
-  /**
-   * Lifts `if` to Functor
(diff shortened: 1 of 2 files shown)


travisbrown:
This unfortunately can't be added directly to the `Functor` trait on 2.11 because it breaks binary compatibility. You can add it to functor syntax via the `BinCompat` approach, and if you really wanted you could add it to `Functor` via an enrichment method.

sdrafahl:
@travisbrown Could you show me an example of the BinCompat approach? I have an idea of how to do the implicit class approach but would like to consider both?

travisbrown:
@sdrafahl [This PR](https://github.com/typelevel/cats/pull/2609/files) is a good example of what's necessary (it's quite painful, unfortunately, which is why we've dropped 2.11 on master :smile:).

larsrh:
I'm closing all 2.11-related tickets since there is not enough community interest in further releases.

PR #3132, opened 2019-11-08
(wip) move maximum/minimumOption to UnorderedFoldable

### LINKED ISSUE #2715: minimumOption and maximumOption could be on UnorderdFoldable
Order.max/min is commutative, so it could be used with UnorderedFoldable/UnorderedReducible.

They should be moved down there in cats 2.0, which I think would be binary compatible in scala 2.12 since UnorderedFoldable is a subtype of Foldable.

mschuwalow:
resolves #2715 and is a followup to #2832 

This currently breaks binary compatibility, but I am unsure on how to best fix this. It would be greatly
appreciated if someone could point me into the right direction.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -169,6 +169,7 @@ possible:
  * Matt Martin
  * Matthias Lüneberg
  * Max Worgan
+ * Maxim Schuwalow
  * Merlin Göttlinger
  * Michał Gutowski
  * Michael Ledin
--- a/core/src/main/scala/cats/Foldable.scala
+++ b/core/src/main/scala/cats/Foldable.scala
@@ -166,62 +166,6 @@ import Foldable.sentinel
     reduceRightToOption(fa)(identity)(f)
 
   /**
-   * Find the minimum `A` item in this structure according to the `Order[A]`.
-   *
-   * @return `None` if the structure is empty, otherwise the minimum element
-   * wrapped in a `Some`.
-   *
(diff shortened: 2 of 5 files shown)


travisbrown:
I'm afraid the only way to "fix" this would be to put the `maximumOption` and `minimumOption` implementations back in `Foldable` and to add a `dummy: DummyImplicit = DummyImplicit.dummyImplicit` argument to the versions in `UnorderedFoldable`.

It's a terrible workaround and I hate it, but if we care about having these in the right place I think it's the only way. Unless maybe @mpilquist knows of a way to have Simulacrum generate `Ops` methods for overrides?

(The `*ByOption` versions are fine as you have them because they're new since 2.0.0.)

mschuwalow:
> I'm afraid the only way to "fix" this would be to put the `maximumOption` and `minimumOption` implementations back in `Foldable` and to add a `dummy: DummyImplicit = DummyImplicit.dummyImplicit` argument to the versions in `UnorderedFoldable`.
> 
> It's a terrible workaround and I hate it, but if we care about having these in the right place I think it's the only way. Unless maybe @mpilquist knows of a way to have Simulacrum generate `Ops` methods for overrides?
> 
> (The `*ByOption` versions are fine as you have them because they're new since 2.0.0.)

Would you recommend proceeding with this workaround? In that case I'll update the pr

larsrh:
This PR is still marked as WIP and contains some conflicts. Please open a new PR if you're interested in pursuing this feature again.

PR #3134, opened 2019-11-09
Backports #3133 to scala_2.11

jackcviers:
* Adds scala version specific src directories
* Backports PR #3133 to add scala 2.11 support


travisbrown:
This looks reasonable to me, but I think we should wait for #3133 to be merged. It would also probably be easier to backport features in the order they were added to post-2.0.0 master, but I don't think that should necessarily block this PR. (Also, relatedly, I think it'd be better to separate the Scalafmt update commit from the other part.)

travisbrown:
Wait, this doesn't need Scala-version-specific code because the `scala_2.11` branch will only ever be used to publish for 2.11.

jackcviers:
Sure, I can remove the 2.11 branch and seperate out the build parts. It wasn't passing the scalafmt build step on travis, so I added it in.

jackcviers:
Also, I put in the 2.11, 2.12, and 2.13+ source directories in so that if you ever did want to merge the 2.11 back into master, that would be possible without a great deal of trouble.

I have to wonder, why not simply have a 2.11 source directory on master where all code changes go for 2.11, rather than a full branch? The way it is, you will have to maintain tags for two branches going forward.

travisbrown:
@jackcviers That would involve a lot of code duplication, and the plan is to drop 2.11 entirely very soon.

larsrh:
I'm closing all 2.11-related tickets since there is not enough community interest in further releases.

PR #3136, opened 2019-11-11
Add benchmark for foldMapA

travisbrown:
This is a follow-up to #3130. We're making claims about performance in the API docs there, so I figured we should have a micro-benchmark to back them up.

I also ran Scalafmt on the `bench` project in a separate commit, since it doesn't look like it's been formatted in a while.


travisbrown:
Merging this with one additional sign-off since it only touches `bench`.

PR #3137, opened 2019-11-12
Upgrades sbt version on the 2.11 branch to 1.3.3 to match master

jackcviers:
Thank you for contributing to Cats!

This is a kind reminder to run `sbt +prePR` and commit the changed files, if any, before submitting.

PR #3139, opened 2019-11-12
Update 2.11 branch build

travisbrown:
Backports a couple of @larsrh's changes from #3079 to the `scala_2.11` branch.

The CI failure in #3137 isn't what I'd expect (I got a different one with the same changes locally), but I think this should fix it.


travisbrown:
@jackcviers Mind taking a look / signing off on this?

PR #3144, opened 2019-11-13
Add Align for Stream

travisbrown:
Follow-up to #3076 adding `Align[Stream]` with the same implementation as `Align[LazyList]`. I don't think there's any reason not to include these, @LukaJCB?

I've also added a package-private `Align.alignWithIterator` helper method to clean up some repetition in `Align` definitions.


travisbrown:
Great, thanks @LukaJCB!

PR #3147, opened 2019-11-13
Add Parallel.parFoldMapA

nigredo-tori:
Aka `foldMapA` hoisted through a `Parallel`.

### DIFF
--- a/core/src/main/scala/cats/Parallel.scala
+++ b/core/src/main/scala/cats/Parallel.scala
@@ -314,13 +314,13 @@ object Parallel extends ParallelArityFunctions2 {
 
   /**
    * Like `Foldable[A].foldMapA`, but uses the applicative instance
-   * corresponding to the Parallel instance instead
+   * corresponding to the Parallel instance instead.
    */
-  def parFoldMapA[T[_]: Foldable, M[_], A, B: Monoid](
+  def parFoldMapA[T[_], M[_], A, B](
     ta: T[A]
-  )(f: A => M[B])(implicit P: Parallel[M]): M[B] = {
+  )(f: A => M[B])(implicit T: Foldable[T], P: Parallel[M], B: Monoid[B]): M[B] = {
     val fb: P.F[B] =
-      Foldable[T].foldMapA(ta)(a => P.parallel(f(a)))(P.applicative, implicitly)
+      Foldable[T].foldMapA(ta)(a => P.parallel(f(a)))(P.applicative, B)
     P.sequential(fb)
   }
 
--- a/core/src/main/scala/cats/Parallel.scala
+++ b/core/src/main/scala/cats/Parallel.scala
@@ -313,6 +313,18 @@ object Parallel extends ParallelArityFunctions2 {
   }
 
   /**
+   * Like `Foldable[A].foldMapA`, but uses the applicative instance
+   * corresponding to the Parallel instance instead
+   */
+  def parFoldMapA[T[_]: Foldable, M[_], A, B: Monoid](
+    ta: T[A]
+  )(f: A => M[B])(implicit P: Parallel[M]): M[B] = {
+    val fb: P.F[B] =
+      Foldable[T].foldMapA(ta)(a => P.parallel(f(a)))(P.applicative, implicitly)
+    P.sequential(fb)
+  }
+
+  /**
    * Like `Applicative[F].ap`, but uses the applicative instance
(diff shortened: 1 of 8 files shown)


LukaJCB:
This is a great feature!

I wonder though if it would be easier to just place this method straight on the `Foldable` trait?

nigredo-tori:
It would probably be easier (no extra syntax). I'm a bit out of the loop, though - is this the preferred way to do things at the moment? I put this on `Parallel` to keep it consistent with all the other `par*` functions, but I can put it on `Foldable` if you want.

core/src/main/scala/cats/Parallel.scala
```diff
@@ -312,6 +312,18 @@ object Parallel extends ParallelArityFunctions2 {
     P.sequential(ftab)
   }
 
+  /**
+   * Like `Foldable[A].foldMapA`, but uses the applicative instance
+   * corresponding to the Parallel instance instead
+   */
+  def parFoldMapA[T[_]: Foldable, M[_], A, B: Monoid](
+    ta: T[A]
+  )(f: A => M[B])(implicit P: Parallel[M]): M[B] = {
```
travisbrown: I think we've made a mistake in allowing context bounds to be mixed with explicit implicit parameter sections. In a case like this with multiple context bounds it's especially confusing.

In general it tends to be best for inference for implicit parameters should be in the same order as their types appear in the arguments, which in this case would put the `Parallel` before the `Monoid`, but as it's currently written they're switched. I think the following would be better:

```scala
  def parFoldMapA[T[_], M[_], A, B](
    ta: T[A]
  )(f: A => M[B])(implicit T: Foldable[T], P: Parallel[M], B: Monoid): M[B] = {
```

core/src/main/scala/cats/syntax/all.scala
```diff
@@ -94,3 +95,5 @@ trait AllSyntaxBinCompat4
 trait AllSyntaxBinCompat5 extends ParallelBitraverseSyntax
 
 trait AllSyntaxBinCompat6 extends ParallelUnorderedTraverseSyntax
+
+trait AllSyntaxBinCompat7 extends ParallelFoldMapASyntax
```
travisbrown: We shouldn't need new `BinCompat` traits now that we've dropped 2.11 on master. You should be able to add `extends ParallelFoldMapASyntax` directly to the end of the `trait AllSyntax` stack.

core/src/main/scala/cats/Parallel.scala
```diff
@@ -312,6 +312,18 @@ object Parallel extends ParallelArityFunctions2 {
...
+    val fb: P.F[B] =
+      Foldable[T].foldMapA(ta)(a => P.parallel(f(a)))(P.applicative, implicitly)
```
travisbrown: Would you mind using either `Monoid[B]` (or just `B` if you make the suggested change above) instead of `implicitly` here?

core/src/main/scala/cats/syntax/parallel.scala
```diff
@@ -176,3 +191,8 @@ final class ParallelLeftSequenceOps[T[_, _], M[_], A, B](private val tmab: T[M[A
   def parLeftSequence(implicit T: Bitraverse[T], P: Parallel[M]): M[T[A, B]] =
     Parallel.parLeftSequence(tmab)
 }
+
+final class ParallelFoldMapAOps[T[_], A](private val ma: T[A]) extends AnyVal {
```
travisbrown: It's really unfortunate what we've done here (not you—but us, the maintainers). `ParallelTraversableOps`, `ParallelTraversable_Ops`, `ParallelFlatTraversableOps`, `ParallelFoldMapAOps`, etc. could all be a single `ParallelFoldableOps` trait. As it is I guess we should go with what you've got here for consistency.

core/src/main/scala/cats/Parallel.scala
nigredo-tori: Done (and I did the same in the syntax wrapper).

core/src/main/scala/cats/syntax/all.scala
nigredo-tori: Done.

core/src/main/scala/cats/Parallel.scala
nigredo-tori: Done.

PR #3148, opened 2019-11-13
Move adaptError

### LINKED ISSUE #2685: MonadError#adaptError could be on ApplicativeError instead

travisbrown:
Fixes #2685.

Unfortunately I don't think we can fix the syntax without breaking bincompat due to a compiler bug: https://github.com/typelevel/cats/issues/2685#issuecomment-547862937

### DIFF
--- a/core/src/main/scala/cats/syntax/applicativeError.scala
+++ b/core/src/main/scala/cats/syntax/applicativeError.scala
@@ -124,8 +124,7 @@ final class ApplicativeErrorOps[F[_], E, A](private val fa: F[A]) extends AnyVal
    * }}}
    *
    * This is the same as `MonadErrorOps#adaptError`. It cannot have the same name because
-   * this would result in ambiguous implicits. `adaptError` will be moved from `MonadError`
-   * to `ApplicativeError` in Cats 2.0: see [[https://github.com/typelevel/cats/issues/2685]]
+   * this would result in ambiguous implicits.
    */
   def adaptErr(pf: PartialFunction[E, E])(implicit F: ApplicativeError[F, E]): F[A] =
     F.recoverWith(fa)(pf.andThen(F.raiseError[A] _))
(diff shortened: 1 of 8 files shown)


core/src/main/scala/cats/syntax/applicativeError.scala
```diff
@@ -124,8 +124,7 @@ final class ApplicativeErrorOps[F[_], E, A](private val fa: F[A]) extends AnyVal
    * }}}
    *
    * This is the same as `MonadErrorOps#adaptError`. It cannot have the same name because
-   * this would result in ambiguous implicits. `adaptError` will be moved from `MonadError`
-   * to `ApplicativeError` in Cats 2.0: see [[https://github.com/typelevel/cats/issues/2685]]
+   * this would result in ambiguous implicits.
    */
   def adaptErr(pf: PartialFunction[E, E])(implicit F: ApplicativeError[F, E]): F[A] =
     F.recoverWith(fa)(pf.andThen(F.raiseError[A] _))
```
rossabaker: Nit: is there any reason to keep a duplicate implementation, now that it's on the constraint?

travisbrown: Agreed, thanks for catching this!

travisbrown:
Okay, conflict is resolved and implementation is de-duplicated. I've also added some syntax tests.

PR #3151, opened 2019-11-14
fix link to moniodK and Applicative

stsatlantis:


travisbrown:
CI failure was unrelated / spurious so I've restarted.

PR #3152, opened 2019-11-14
"Move" type class syntax methods onto type classes

travisbrown:
Between 1.0.0 and 2.0.0 we added a lot of new type class operations as syntax methods in order to avoid breaking binary compatibility on 2.11. We don't have to worry about that any more, which means these methods can go on the type classes where they belong.

## Syntax methods

There are two categories of syntax methods that can be moved. The first is something like `findM` for `Foldable`, which should live next to `find` on the type class itself, but right now only exists as syntax for `F[A]` values:

```scala
final class FoldableOps[F[_], A](private val fa: F[A]) extends AnyVal {
  def findM[G[_]](p: A => G[Boolean])(implicit F: Foldable[F], G: Monad[G]): G[Option[A]] = ???
}
```
For methods like this I've added an implementation in the type class (`trait Foldable` in this case):

```scala
  @noop
  def findM[G[_], A](fa: F[A])(p: A => G[Boolean])(implicit G: Monad[G]): G[Option[A]] =
```
I've then changed the syntax implementation to use the new type class method. Note that the `@noop` isn't strictly necessary (except in the case of `leftSequence`, which I think hits a Simulacrum bug)—it's just an optimization to avoid Simulacrum generating redundant syntax methods (since we can't remove the current ones because of bincompat).

The big advantage here is discoverability. Someone looking at the `Foldable` API docs for `find` might wonder whether there's an effectful version, and there is, but they won't currently know that without also looking in `cats.syntax.foldable`.

Another advantage is ease of use. In some cases, like `count` for `UnorderedFoldable`, which currently exists only as syntax, it's very difficult to actually use the method in many cases, because the syntax version collides with the `count` on standard library collections. After this PR you can just `F.count(pred)` on your `UnorderedFoldable` instance.

It's a smaller thing, but these changes also mean less logic in the syntax package, and fewer changes to make when we eventually do break binary compatibility in Cats 3.0 and (fingers crossed) can let Simulacrum or some Simulacrum successor do even more of the syntax boilerplate work.

## Type class instance enrichment methods

The other category of methods I've moved here are things like `fromValidated` for `ApplicativeError`. These methods aren't enriched onto `F[A]` values or whatever, but onto the type class instance itself. In this case we can add the method to the type class and actually "remove" the syntax method, in the sense that we can make the enrichment conversion non-implicit and can deprecated everything involved (we can't _remove_ remove it thanks to bincompat). This means less unnecessary garbage cluttering up implicit search when you import form `cats.syntax`.

## Other issues

* We currently have `collectFirst` and `collectFirstSome` but `collectFold` and `collectSomeFold` (as syntax methods). I've used `collectFoldSome` for consistency in the new proper type class version of `collectFoldSome`, and have deprecated the `collectFoldSome` syntax method.

* There are a few syntax methods that could have been added to type classes that I've omitted. Two examples are `contains_` and `mkString_` for `Foldable`. In all of these cases (and there are only a couple others, I think) the implementations are trivial (like for `contains_`) or the operation is very specific (like `mkString_`) and doesn't feel to me like it really needs to go on the type class.

### DIFF
--- a/core/src/main/scala/cats/ApplicativeError.scala
+++ b/core/src/main/scala/cats/ApplicativeError.scala
@@ -1,6 +1,7 @@
 package cats
 
-import cats.data.EitherT
+import cats.data.{EitherT, Validated}
+import cats.data.Validated.{Invalid, Valid}
 
 import scala.reflect.ClassTag
 import scala.util.{Failure, Success, Try}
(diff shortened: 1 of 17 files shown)


travisbrown:
Ǹote that there are no changes to tests and the only change to a doctest is to avoid the `collectSomeFold` deprecation. This is a breaking change for people who for some strange reason are using e.g. `ApplicativeErrorExtensionOps` explicitly, but shouldn't affect any normal usage.

rossabaker:
I don't think I can review this in detail until tomorrow, but :+1: in principle.

PR #3154, opened 2019-11-15
Make 1.x deprecation targets package-private

travisbrown:
This PR makes everything that was deprecated in 1.x package-private (in the cases where that's possible without breaking bincompat).

This means less noise in the API docs, and that users will be in a better position to be ready once we can finally clean this stuff up for real.

I don't feel terribly strongly about getting this into 2.1.0, but I think it'd be nice to have.


travisbrown:
@rossabaker Out-of-character :+1:s are the best :+1:s. :smile:

Issue #2891, opened 2019-06-14
Ambiguous Vector instances


durban:
The implicits `catsKernelStdPartialOrderForVector` and `catsKernelStdHashForVector` ([here](https://github.com/typelevel/cats/blob/master/kernel/src/main/scala/cats/kernel/instances/VectorInstances.scala#L11)) are ambiguous if `A` has both `PartialOrder` and `Hash` instances, and we're looking for an `Eq[Vector[A]]`. (`List` seems to have the same problem.)

kailuowang:
I am not sure if we can fix it for Scala 2.11 without breaking BC. We might have to leave it as is on 2.11 and do a scala version specific fix for 2.12+. Or we simply drop 2.11 which is more likely now than before.

travisbrown:
The fix in #3100 ended up breaking bincompat (MiMa missed it), so we had to revert it for now.

UlisesTorrella:
Hi! I tried to replicate this error in Scala212, Scala213 and Scala3 with the following script:
```
import cats._

type A = String // tried also with Int, List[Int]

// checked that A has a both instances
implicitly(PartialOrder[A])
implicitly(Hash[A])

// replicate implicit ambiguity error
implicitly(Eq[Vector[A]])
implicitly(Eq[List[A]])
``` 
And got successful executions every time. Let me know if there is another situation that breaks, but otherwise the issue could be closed.

armanbilge:
Thanks!

Issue #2701, opened 2019-01-17
Ambiguous Eq instances for tuples


ceedubs:
I think that the situations in which this arises are somewhat uncommon, but I don't know of any good way to work around it when it does happen. For example here's what happens when you try to summon the `Eq[(Set[Boolean], Set[Boolean])]` instance:

```scala
scala> Eq[(Set[Boolean], Set[Boolean])]
<console>:18: error: ambiguous implicit values:
 both method catsKernelStdHashForTuple2 in trait TupleInstances1 of type [A0, A1](implicit A0: cats.kernel.Hash[A0], implicit A1: cats.kernel.Hash[A1])cats.kernel.Hash[(A0, A1)]                          
 and method catsKernelStdPartialOrderForTuple2 in trait TupleInstances1 of type [A0, A1](implicit A0: cats.kernel.PartialOrder[A0], implicit A1: cats.kernel.PartialOrder[A1])cats.kernel.PartialOrder[(A0,
A1)]
 match expected type cats.kernel.Eq[(Set[Boolean], Set[Boolean])]
       Eq[(Set[Boolean], Set[Boolean])]
         ^
```

Summoning `Eq[Set[Boolean]]` by itself works fine:

```scala
scala> Eq[Set[Boolean]]
res0: cats.kernel.Eq[Set[Boolean]] = cats.kernel.instances.SetHash@4493c47
```

And summoning `Eq[(List[Boolean], List[Boolean])]` works:

```scala
scala> Eq[(List[Boolean], List[Boolean])]
res2: cats.kernel.Eq[(List[Boolean], List[Boolean])] = cats.kernel.instances.TupleInstances$$anon$90@155e7e 
```

I think that the difference is due to implicit resolution happening differently for `List` since its `Eq` and `Hash` instances have constraints on the `A` type, while `Set`'s instances do not. I think that it comes down to [this code](https://github.com/typelevel/cats/blob/637fbe7a4b74eb82842e1131f14335c6c8fd8bad/project/KernelBoiler.scala#L198-L209), where the `Hash` and `PartialOrder` instances should be in different layers of the hierarchy since they share a common `Eq` superclass. However, I think that changing that to not be the case would be binary incompatible (but maybe only prior to 2.12?).

kailuowang:
[...]

PR #3163, opened 2019-11-19
Fix broken binary compatibility

travisbrown:
Reverts #3100, #3105, and undoes part of #2834. This unfixes #2891 and #2701, which I think will have to wait for 3.0.

In the case of #2834 I don't think the move was necessary, since the `CommutativeMonoid` one is more specific. In any case moving the other back doesn't break any tests and seems fine:

```scala
scala> import cats.kernel.instances.all._
import cats.kernel.instances.all._

scala> cats.kernel.CommutativeMonoid[Int]
res0: cats.kernel.CommutativeMonoid[Int] = cats.kernel.instances.IntGroup@5c7654d1

scala> cats.kernel.Monoid[Int]
res1: cats.kernel.Monoid[Int] = cats.kernel.instances.IntGroup@5c7654d1

scala> cats.kernel.Monoid[String]
res2: cats.kernel.Monoid[String] = cats.kernel.instances.StringMonoid@77c600b4
```

I've confirmed that my quick-fix MiMa branch [here](https://github.com/lightbend/mima/issues/426) doesn't find any problems after this change and #3162.

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -1,5 +1,5 @@
 Thank you for contributing to Cats!
 
-This is a kind reminder to run `sbt prePR` and commit the changed files, if any, before submitting. 
+This is a kind reminder to run `sbt +prePR` and commit the changed files, if any, before submitting.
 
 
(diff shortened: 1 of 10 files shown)


travisbrown:
@rossabaker I was thinking we'd reopen the issues that were closed once this is merged, which I think should cover it.

PR #3165, opened 2019-11-20
Add catchOnly to ApplicativeError.

takayahilton:


travisbrown:
Looks reasonable to me.

tests/src/test/scala/cats/tests/TrySuite.scala
```diff
@@ -61,6 +61,17 @@ class TrySuite extends CatsSuite {
...
+
+  test("catchOnly works") {
+    forAll { e: Either[String, Int] =>
+      val str = e.fold(identity, _.toString)
+      val res = MonadError[Try, Throwable].catchOnly[NumberFormatException](str.toInt)
+      // the above should just never cause an uncaught exception
+      // this is a somewhat bogus test:
+      res should not be (null)
+    }
+  }
```
rossabaker: Might it be useful to also test that `catchOnly` on an unrelated exception does not catch?

core/src/main/scala/cats/ApplicativeError.scala
```diff
@@ -216,6 +215,12 @@ trait ApplicativeError[F[_], E] extends Applicative[F] {
       case NonFatal(e) => raiseError(e)
     }
 
+  /**
+   * Evaluates the specified block, catching exceptions of the specified type. Uncaught exceptions are propagated.
+   */
+  def catchOnly[T >: Null <: Throwable]: CatchOnlyPartiallyApplied[T, F, E] =
+    new CatchOnlyPartiallyApplied[T, F, E](this)
```
satorg, line 222: Out of curiosity... I'm struggling to understand this.

What was the `T >: Null` constraint added for in there? To me it seems that due to the latter constraint `T <: Throwable` the former one will be always satisfied automatically, won't it? This is because `Throwable` is `AnyRef` and `Null <: T <: AnyRef` for any `T`.

Or am I missing something?
@takayahilton could you clarify please?

PR #3168, opened 2019-11-20
Use `getOrElse` in `NonEmptyList#last`

PeterPerhac:
seems like someone re-implemented Option's `getOrElse` inside NonEmptyList. Delegate to Option#getOrElse instead

### DIFF
--- a/core/src/main/scala/cats/data/NonEmptyList.scala
+++ b/core/src/main/scala/cats/data/NonEmptyList.scala
@@ -36,10 +36,7 @@ final case class NonEmptyList[+A](head: A, tail: List[A]) {
    * res0: Int = 5
    * }}}
    */
-  def last: A = tail.lastOption match {
-    case None    => head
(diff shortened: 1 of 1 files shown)


joroKr21:
It could be an optimization.

travisbrown:
I'm not sure this is the reason for the original implementation, but `getOrElse` takes a `Function0` argument, and before Scala 2.12 that meant that the pattern matching version was one allocation lighter. In this case now it's just an extra synthetic method and my guess is the two are probably close to the same.

Neither of those versions are going to be anywhere close to `if (tail.isEmpty) head else tail.last`, which actually does still avoid an allocation on 2.12+, but I don't think it matters much here.

djspiewak:
This is somewhat defensible as an optimization, but as @travisbrown pointed out, both of them are quite slow compared to the `if`/`else` equivalent. I don't see any problem with the simplification.

PR #3178, opened 2019-11-27
Add ApplicativeError for EitherT documentation

### LINKED ISSUE #3177: EitherT.fromMonadError
I think #1644 is related.

Basicly something like:
```
import cats.implicits._
import cats.data.EitherT
import cats.MonadError

def fromMonadError[F[_], A, B](me: F[B])(implicit ev: MonadError[F, A]) = {
  EitherT(me.map(Either.right[A, B]).recover{case ex => Either.left[A, B](ex)})
}
```

Or if we have this functionality already maybe it would worth a mention in the EitherT/MonadError documentation.

If others think this is useful, I try to PR it :)

tg44:
fixes #3177

### DIFF
--- a/docs/src/main/tut/datatypes/eithert.md
+++ b/docs/src/main/tut/datatypes/eithert.md
@@ -145,6 +145,19 @@ val myOptionET = EitherT.fromOption[Future](myOption, "option not defined")
 val myOptionListET = EitherT.fromOptionF(myOptionList, "option not defined")
 ```
 
+## From `ApplicativeError[F, E]` to `EitherT[F, E, A]`
+
+An `ApplicativeError[F, E]` or a `MonadError[F, E]`, can be converted into `EitherT[F, E, A]` 
+using the `attemptT` method on them.
+
+```tut:silent
+val myTry: Try[Int] = Try(2)
+val myFuture: Future[String] = Future.failed(new Exception())
+
+val myTryET: EitherT[Try, Throwable, Int] = myTry.attemptT
+val myFutureET: EitherT[Future, Throwable, String] = myFuture.attemptT
+```
+
 ## Extracting an `F[Either[A, B]]` from an `EitherT[F, A, B]`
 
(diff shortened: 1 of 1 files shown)


docs/src/main/tut/datatypes/eithert.md
```diff
@@ -145,6 +145,19 @@ val myOptionET = EitherT.fromOption[Future](myOption, "option not defined")
...
+
+An `ApplicativeError[F, E]` or a `MonadError[F, E]`, can be converted into `EitherT[F, E, A]` 
+using the `attemptT` method on them.
+
+```tut:book
+val myTry: Try[Int] = Try(2)
+val myFuture: Future[String] = Future.failed(new Exception())
+
+val myTryET = myTry.attemptT
+val myFutureET = myFuture.attemptT
```
LukaJCB: Should we add type annotations to these as well to make it more obvious what the function does? :)

tg44: I had a dilemma with this, but the tut:book creates a kinda great output so I let it go...
Probably using tut:silent and type annotations would be an another good option.

tg44: So what do you think? Should I add the return types, or let the `tut:book` to do its magic?

djspiewak: IMO I think it's better to add the type annotations, just to be fully clear even in the original form. Also at some point I'm assuming we'll be switching from tut to mdoc, so it's worth thinking a bit ahead.

tg44:
fixed it

solarmosaic-kflorence:
@LukaJCB This is very helpful documentation, but it does not appear on the cats website: https://typelevel.org/cats/datatypes/eithert.html

See related stackoverflow question: https://stackoverflow.com/questions/54931204/recovering-underlying-future-into-cats-eitherts-left

PR #3181, opened 2019-12-01
Fix typo

msinton:
nextBoolean correctly implemented initially, then second occurrence would be false biased (ever so slightly!)


msinton:
> Thanks for catching this!

You're welcome. Thanks for Cats! I hope I can make more significant contributions in future!

PR #3190, opened 2019-12-04
Mark match that is unchecked by Scala 2 as unchecked

travisbrown:
The Dotty exhaustivity checker is both smarter and more demanding than Scala 2's, and this is a case where it warns while Scala 2 just silently ignores the fact that it can't prove exhaustivity (it's the only case like this in cats-core).

It would be possible to make the match exhaustive in a way that both compilers would accept by writing something like this:

```scala
  as match {
    case h ==: t   => f(h, Eval.defer(loop(t)))
    case _ => lb
  }
```
…but that results in an extra `ArrayBuffer` instantiation in the empty case. Adding `@unchecked` seems to me like a reasonable fix, since it just flags what the Scala 2 compiler is already doing.


dwijnand:
> Adding `@unchecked` seems to me like a reasonable fix, since it just flags what the Scala 2 compiler is already doing.

True, but it's also big and loses the value Dotty is (... finally! IMO) adding, there.

> but that results in an extra `ArrayBuffer` instantiation in the empty case

It's a touch of feature creep, but maybe `uncons` and `initLast` should fast-track on `isEmpty`?  It would drop that cost.

Issue #3192, opened 2019-12-04
Rethinking type class boilerplate generation


travisbrown:
This issue is a follow-up to #2553 that focuses on the Simulacrum question and gives a concrete proposal for a replacement.

The problem is that [Simulacrum 1](https://github.com/typelevel/simulacrum) is based on [Macro Paradise](https://docs.scala-lang.org/overviews/macros/paradise.html)'s macro annotations, which means it has a couple of disadvantages:

* It doesn't work on Dotty, which doesn't have macros that can do this kind of thing (and probably never will, as far as I can tell).
* It can be difficult to troubleshoot when it doesn't do what you expect (you have to turn on tracing and dig through some pretty verbose output).

I've been experimenting with reworking Simulacrum's code generation as a [set of Scalafix rules](https://github.com/travisbrown/simulacrum-scalafix). This [Scalafix](https://github.com/scalacenter/scalafix)-based approach resolves both of the macro-related issues:

* It works exactly the same on Dotty as on Scala 2.
* The generated code is right there in front of you.

This second part is also a disadvantage: in my current proof-of-concept you end up with a lot ([like 2.2k lines](https://github.com/travisbrown/cats/commit/a6a0eb39808fd545ecd0ad8d0ab2a769145ffb38)) of boilerplate being added to the Cats source. It "reserves" the last part of the companion object of every type class for rule-managed code, and if contributors accidentally edit this part of these files, their changes will be overwritten.

I've tried a few ways of generating at least some of the syntax method boilerplate as managed source, in order to minimize the additions to checked-in code, but I don't think it's possible to do this in a way that makes sense without breaking binary compatibility. This might be a viable solution for Cats 3, though, and it would be pretty straightforward to change my proof-of-concept to make this configurable (i.e. your build indicates whether the generated code should go behind a banner in `src/main/scala` or in `src_managed`).

The alternative seems to be to implement the current Simulacrum functionality either as a Dotty compiler plugin or [directly in the compiler](https://github.com/typelevel/cats/issues/2553#issuecomment-507043866).

If we do decide to go the Scalafix route, there's also a question of when. In theory it could happen at any time, since whether the code generation is done by a macro or a Scalafix rule shouldn't make any difference to the published artifacts.

[...]

PR #3193, opened 2019-12-04
Expand kind-projector's syntax for polymorphic function values

travisbrown:
I've always hated this specific piece of kind-projector syntax. It feels very different from everything else in Scala in that it's not even possible to annotate the function argument with its type, because the type parameter is entirely synthetic. I find it confusing whenever I notice it, and I'm familiar with it, so I'm assuming it confuses other people too. Maybe it has a place in application code but I don't personally think it belongs in a library like Cats.

Now that we're thinking about Dotty compatibility (see #3192) I have a good excuse to propose getting rid of it entirely. For this PR I ran [this Scalafix rule](https://gist.github.com/travisbrown/663ea73d277707e1aef484b56b50fada) (which is a slightly modified version of [the one here](https://github.com/travisbrown/simulacrum-scalafix)) and then manually fixed a few instances where `A` collided with another type parameter.


travisbrown:
Thanks @LukaJCB, hopefully it will work this time. :smile:

dwijnand:
> it's not even possible to annotate the function argument with its type, because the type parameter is entirely synthetic

I think you should be able to, see https://github.com/sbt/sbt/blob/e5e462f2c2c659a4700c6fb757ea83d3d8ac0329/internal/util-collection/src/main/scala/sbt/internal/util/INode.scala#L36 where I'm using one of the synthetic type parameter names.

djspiewak:
I'm definitely :-1: on this; I find this syntax to be absolutely indispensable. I get that some people find it confusing though, and Dotty compatibility is a real concern, so I won't stand in the way of its removal, I'm just definitely not in favor.

travisbrown:
@dwijnand

 > I think you should be able to

Idk, in my view the fact that you can surface and rely on an undocumented implementation detail like that is much worse than not being able to refer to the type at all.

travisbrown:
@djspiewak My goal for this set of PRs is to make Dotty-friendly changes that also improve or at least don't hurt readability on Scala 2 now. If there's disagreement on this one I'm happy to put it on hold.

djspiewak:
> My goal for this set of PRs is to make Dotty-friendly changes that also improve or at least don't hurt readability on Scala 2 now. If there's disagreement on this one I'm happy to put it on hold.

At least in my opinion, this hurts readability a *lot*. Like I said, I present that as a subjective view, and I don't think there's anything wrong with your points that, for those who aren't familiar with it, it may be a bit magical. That's technically true about everything in kind-projector though.

Technically speaking, if Dotty is getting polymorphic SAMs (as @smarter mentioned), then the kind-projector lambda syntax is still compatible, since we can just define a `def λ[F[_], G[_]](f: FunctionK[F, G]): FunctionK[F, G] = f` helper which allows code to cross-compile. It wouldn't even need to be compiled specifically for Dotty and not for Scala 2, since kind-projector would shadow it.

travisbrown:
Just refreshed this branch and made the same changes for cats-free and the tests as I had for the other modules.

@djspiewak I don't exactly see how polymorphic SAMs in Dotty will help us have a shared version of the clean syntax here, because of this part:

> not exactly this since it would need to be partially-applied to keep the correct parameter shape, but you get the idea

Even supposing we will have polymorphic SAMs on Dotty, I don't see how we could write a `λ` there such that `λ[K]` where `K` is a `FunctionK[F, G]` would work. It would be different if kind-projector supported `λ[FunctionK, F, G](fa => ???)` as a way to construct `FunctionK[F, G]` values, but that would require changes in kind-projector that I'm not sure anyone will be particularly interested in making (I definitely am not :smile:).

That would also mean that we're forced to wait on significant changes with no specific timeline in both Dotty and kind-projector before we can have Cats cross-building on Dotty in master. I'm inclined not to do that, and to merge this sooner rather than later. We can always undo these expansions if someday we get a nicer solution that works on both Scala 2 and Dotty.

djspiewak:
Fun fact: you can instantly crash the Dotty repl if you enter the following text:

```scala
def λ[F[a[_], b[_]]] = new {
  def apply[A[_], B[_]](f: F[A, B]): F[A, B] = f
}
```

Mostly it appears the `F[a[_], b[_]]` is the culprit, since that shape messes things up in other types as well. Also that shape isn't exactly what I want *anyway*, but I thought it was a fun discovery.

Anyway, a bit of fiddling convinced me that you're right, barring any witchcraft in Scala 3 of which I am not currently aware. We really would need something pretty closely approximating polykinds in order to do this. In fact, by using the skolems polykinding trick, I was able to get this to work… in Scala 2.13:

```scala
trait PolyParent {
  type F0[A]
  type G0[A]
}

trait FunctionK[F[_], G[_]] {
  type F0[A] = F[A]
  type G0[A] = G[A]

  def apply[A](fa: F[A]): G[A]
}

object Skolem {
  type τ
}

def λ[K <: PolyParent](f: K#F0[Skolem.τ] => K#G0[Skolem.τ]): FunctionK[K#F0, K#G0] =
  new FunctionK[K#F0, K#G0] {
    def apply[A](fa: K#F0[A]): K#G0[A] = f(fa.asInstanceOf).asInstanceOf
  }
```

So that's fun. It actually does the right things. But trying to do this in Dotty gives me a `NullPointerException` (also it *shouldn't* work in Dotty due to the unrestricted type projection).

At any rate, I am (sadly) convinced that you're probably right and we need to remove the expansions. I withdraw my objection.

travisbrown:
@djspiewak For what it's worth I think Dotty should actually be fine with your first definition—you're probably seeing a [bug that affects some synthetic names](https://github.com/lampepfl/dotty/issues/7934) in the first line you type in the REPL. It's supposed to have been fixed since 0.22.0-RC1, but I've gotten in the habit of typing `val x = 1` every time I open a Dotty REPL and it's going to be hard to shake :smile:.

PR #3199, opened 2019-12-10
Fix foldA, reduceA, and reduceMapA short-circuiting

travisbrown:
As discussed in #3150, the only real reason I can see for `foldA` and `reduceA` to exist is because they can short-circuit, while `fold(fga)(Applicative.monoid)` etc. can't. This change fixes them so they actually do short-circuit, and adds some tests to verify this.


travisbrown:
I went ahead and added the API docs edits in this PR, hope you don't mind, @LukaJCB.

PR #3203, opened 2019-12-10
Fix more bincompat breakage

### LINKED ISSUE #3202: More binary compatibility breakage
Compile this with 2.0.0:

```scala
object Test {
  val instances = new cats.instances.AllInstancesBinCompat6 {}
  def main(args: Array[String]): Unit = println(
    instances.catsKernelStdHashForSortedSet(
      cats.kernel.instances.int.catsKernelStdOrderForInt,
      cats.kernel.instances.int.catsKernelStdOrderForInt
    )
  )
}
```

Then run it with 2.1.0-RC1 or RC2, and you get this:

```scala
Exception in thread "main" java.lang.NoSuchMethodError: cats.instances.SortedSetInstancesBinCompat1.catsKernelStdHashForSortedSet$(Lcats/instances/SortedSetInstancesBinCompat1;Lcats/kernel/Order;Lcats/kernel/Hash;)Lcats/kernel/Hash;
        at Test$$anon$1.catsKernelStdHashForSortedSet(Test.scala:2)
        at Test$.main(Test.scala:6)
        at Test.main(Test.scala)
        ...
```

This is the same basic issue as #3163 but my original MiMa fix didn't catch it. Working on a PR now.

travisbrown:
Fixes #3202 and all related issues I was able to turn up.

### DIFF
--- a/core/src/main/scala/cats/instances/sortedSet.scala
+++ b/core/src/main/scala/cats/instances/sortedSet.scala
@@ -95,14 +95,6 @@ private[instances] trait SortedSetInstancesBinCompat1 extends LowPrioritySortedS
   implicit def catsKernelStdHashForSortedSet1[A: Hash]: Hash[SortedSet[A]] =
     cats.kernel.instances.sortedSet.catsKernelStdHashForSortedSet[A](Hash[A])
 
-  // TODO: Remove when this is no longer necessary for binary compatibility.
-  // Note that the overrides here and below are only necessary because the
-  // definitions in `SortedSetInstances1` conflict with the ones in
-  // `cats.kernel.instances.SortedSetInstances`. Both are inherited here, so
-  // we have to "bubble" the "correct" ones up to the appropriate place.
-  override def catsKernelStdHashForSortedSet[A: Hash]: Hash[SortedSet[A]] =
-    cats.kernel.instances.sortedSet.catsKernelStdHashForSortedSet[A](Hash[A])
-
   @deprecated("Use cats.kernel.instances.sortedSet.catsKernelStdHashForSortedSet", "2.0.0-RC3")
   override def catsKernelStdHashForSortedSet[A: Order: Hash]: Hash[SortedSet[A]] =
     cats.kernel.instances.sortedSet.catsKernelStdHashForSortedSet[A](Hash[A])
--- a/core/src/main/scala/cats/instances/sortedSet.scala
+++ b/core/src/main/scala/cats/instances/sortedSet.scala
@@ -92,12 +92,19 @@ private[instances] trait SortedSetInstancesBinCompat0 {
 }
 
 private[instances] trait SortedSetInstancesBinCompat1 extends LowPrioritySortedSetInstancesBinCompat1 {
+  implicit def catsKernelStdHashForSortedSet1[A: Hash]: Hash[SortedSet[A]] =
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/instances/sortedSet.scala
```diff
@@ -92,12 +92,11 @@ private[instances] trait SortedSetInstancesBinCompat0 {
...
-  // definitions in `SortedSetInstances1` conflict with the ones in
-  // `cats.kernel.instances.SortedSetInstances`. Both are inherited here, so
-  // we have to "bubble" the "correct" ones up to the appropriate place.
-  implicit override def catsKernelStdHashForSortedSet[A: Hash]: Hash[SortedSet[A]] =
+  implicit def catsKernelStdHashForSortedSet1[A: Hash]: Hash[SortedSet[A]] =
+    cats.kernel.instances.sortedSet.catsKernelStdHashForSortedSet[A](Hash[A])
+
+  @deprecated("Use cats.kernel.instances.sortedSet.catsKernelStdHashForSortedSet", "2.0.0-RC3")
+  override def catsKernelStdHashForSortedSet[A: Order: Hash]: Hash[SortedSet[A]] =
     cats.kernel.instances.sortedSet.catsKernelStdHashForSortedSet[A](Hash[A])
```
kailuowang: It should be easy to add a test to binCompatTest for the original breakage right?

laws/src/main/scala/cats/laws/MonadErrorLaws.scala
```diff
@@ -19,6 +19,12 @@ trait MonadErrorLaws[F[_], E] extends ApplicativeErrorLaws[F, E] with MonadLaws[
 
   def redeemWithDerivedFromAttemptFlatMap[A, B](fa: F[A], fe: E => F[B], fs: A => F[B]): IsEq[F[B]] =
     F.redeemWith(fa)(fe, fs) <-> F.flatMap(F.attempt(fa))(_.fold(fe, fs))
+
+  override def adaptErrorPure[A](a: A, f: E => E): IsEq[F[A]] =
+    F.adaptError(F.pure(a)) { case x => f(x) } <-> F.pure(a)
+
+  override def adaptErrorRaise[A](e: E, f: E => E): IsEq[F[A]] =
+    F.adaptError(F.raiseError[A](e)) { case x => f(x) } <-> F.raiseError(f(e))
```
kailuowang: hmmm, do you mind to comment here on how this fixes BC? not quite obvious to me.

travisbrown: It's the same general problem and fix as in #3162—when you move a method from a trait to a super-trait, you no longer get the synthetic static method (the bridge method or forwarder or whatever) on the lower trait, which can cause problems in real code. Including an override in the lower trait fixes the issue.

I can add a comment in the code when I'm back at my desktop.

core/src/main/scala/cats/instances/sortedSet.scala
travisbrown: I don't think so. I don't know of a way to make it show up without referring to `AllInstancesBinCompat6`, which was introduced after 1.6.1, which `binCompatTest` uses.

travisbrown:
@kailuowang I added a comment about the overrides. What do you think?

PR #3205, opened 2019-12-11
2.1.0-RC3 follow-up

travisbrown:
This PR includes the release commit tagged by sbt-release—if you're merging it please be sure to choose the merge commit option.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -21,6 +21,7 @@ possible:
  * Alejandro Gómez
  * Alessandro Lacava
  * Alejandro Marín E.
+ * Alex Simkin
  * Alexandru Nedelcu
  * Alexander Semenov
  * Alexey Levan
@@ -108,6 +109,7 @@ possible:
  * Gagandeep Kalra
  * Gavin Bisesi
  * Georgi Krastev
+ * Gergő Törcsvári
  * Giovanni Ruggiero
  * Giulio De Luise
  * Giuseppe Cannella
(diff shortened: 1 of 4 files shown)


travisbrown:
@kailuowang Thanks for merging! But it really would be better not to squash these post-release PRs (see the comment in the description above), since now the `v2.1.0-RC3` release tag doesn't point to a commit on master.

We could delete the tag and push a new one, but that could cause other problems. I guess it doesn't matter all that much in this case since this is a pre-release, but we really don't want to be in this situation for actual releases.

travisbrown:
This also means I'll have to update and force-push the `2.1.0` tag, because otherwise it'll end up with noisy merge commits going both ways (i.e. it can't be fast-forwarded to master now, and if we merge-commit master onto it that will require another merge commit at the next release). I'll do that when I'm back at my desktop later today.

travisbrown:
Okay, just force-pushed `2.1.0` branch.

I also deleted the broken `v2.1.0-RC3` and pushed a new one. I had to re-set the tag in the release but otherwise it seems fine. The old tag was just a few hours old so hopefully this won't be too annoying for people (I think having accurate release tags is worth it).

kailuowang:
I apologize. It's my bad. I merged from my phone and my brain short circuited. Sorry about that, thanks for cleaning up the mess.

travisbrown:
@kailuowang No worries, the clean-up was actually really easy!

PR #3207, opened 2019-12-13
Avoid kind-projector syntax with variance annotations

travisbrown:
Dotty currently reserves `*` for use as a type lambda placeholder for compatibility with kind-projector's syntax on Scala 2, via a currently-unimplemented `-Ykind-projector` compiler option. It's not clear whether `-Ykind-projector` should or will support kind-projector's ``λ[`-α` => Whatever[α]]`` syntax, or even `-*`. (I'm working on implementing `-Ykind-projector` at the moment, and personally I don't see compatibility with the goofier corners of kind-projector's syntax as a priority.)

Cats only uses kind-projector's variance annotations in this one file, and they're pretty easy to replace, and in my view the vanilla Scala syntax is clearer, anyway.

/cc @smarter

### DIFF
--- a/core/src/main/scala/cats/evidence/As.scala
+++ b/core/src/main/scala/cats/evidence/As.scala
@@ -68,14 +68,18 @@ object As extends AsInstances {
   /**
    * We can witness the relationship by using it to make a substitution *
    */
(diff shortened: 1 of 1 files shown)


travisbrown:
Also /cc @djspiewak.

PR #3208, opened 2019-12-13
Add Defer.fix

johnynek:
A pattern I often use in working with scalacheck.Gen or with Parsers is:
```
lazy val x: Gen[X] = {
  val recurse = Gen.lzy(x)
  //....
}
```
It occurred to me all the types I use this pattern with are generally Defer instances, and more over, we can express `fix` nicely on `Defer` instances since we know we can always return a value (there is no risk of infinite loop at fix application, although it is up to the user if evaluating `F[A]` is safe.

I think this is a useful function to have on Defer.

### DIFF
--- a/core/src/main/scala/cats/Defer.scala
+++ b/core/src/main/scala/cats/Defer.scala
@@ -28,6 +28,27 @@ package cats
  */
 trait Defer[F[_]] extends Serializable {
   def defer[A](fa: => F[A]): F[A]
+
+  /**
+   * Defer instances, like functions, parsers, generators, IO, etc...
+   * often are used in recursive settings where this function is useful
+   *
+   * fix(fn) == fn(fix(fn))
+   *
+   * example:
+   *
+   * val parser: P[Int] =
+   *   Defer[P].fix[Int] { rec =>
+   *     CharsIn("0123456789") | P("(") ~ rec ~ P(")")
+   *   }
+   *
+   * Note, fn may not yield a terminating value in which case both
+   * of the above F[A] run forever.
+   */
+  def fix[A](fn: F[A] => F[A]): F[A] = {
+    lazy val res: F[A] = defer(fn(res))
+    res
+  }
 }
 
 object Defer {
--- a/tests/src/test/scala/cats/tests/FunctionSuite.scala
+++ b/tests/src/test/scala/cats/tests/FunctionSuite.scala
@@ -46,6 +46,18 @@ class FunctionSuite extends CatsSuite {
   // TODO: make an binary compatible way to do this
   // checkAll("Function1[Int => *]", DeferTests[Function1[Int, *]].defer[Int])
 
+  test("Defer[Function1[Int, *]].fix computing sum") {
+    val sum2 = Defer[Function1[Int, *]].fix[Int] { rec =>
+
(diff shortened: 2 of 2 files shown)


tests/src/test/scala/cats/tests/FunctionSuite.scala
```diff
@@ -46,6 +46,22 @@ class FunctionSuite extends CatsSuite {
...
+  test("Defer[Function1[Int, *]].fix computing sum") {
+    val sum2 = Defer[Function1[Int, *]].fix[Int] {
+      rec =>
+        { n: Int =>
+          if (n <= 0) 0 else n * n + rec(n - 1)
+        }
+    }
+
+    forAll { n0: Int =>
+      val n = n0 & 0x3FF // don't let it get too big
```
travisbrown: This is nitpicky but could you just write this out for clarity? I can't remember off the top of my head what this does for negative `n0` for example.

djspiewak: It's taking the 18 least significant bits. Perhaps a slightly clearer way to write it for those who aren't doing bit manipulation on a regular basis would be `math.abs(n0 % 262144)`.

core/src/main/scala/cats/Defer.scala
```diff
@@ -28,6 +28,27 @@ package cats
...
+   *     CharsIn("0123456789") | P("(") ~ rec ~ P(")")
+   *   }
+   *
+   * Note, fn may not yield a terminating value in which case both
+   * of the above F[A] run forever.
+   */
+  def fix[A](fn: F[A] => F[A]): F[A] = {
+    lazy val res: F[A] = defer(fn(res))
+    res
+  }
```
djspiewak: This is rather nice! I never knew I wanted this.

tests/src/test/scala/cats/tests/FunctionSuite.scala
johnynek: I've changed it to be more direct `Gen.choose(0, 1000)` which I think is nicer anyway. Note `0x3ff = 1024`, it is actually 8 + 2 bits.

johnynek:
One confusing aspect of cats contributions is that it is not clear when and why things are merged. e.g. two people have approved this PR without comment, but not merged. Is it waiting on a third or is there some other criteria that determines when a PR is merged?

LukaJCB:
@johnynek sorry about that, I must have forgotten to press the button.

travisbrown:
@johnynek Sorry, probably partly my fault—I should have responded with a follow-up 👍. (My understanding is if there are two maintainer sign-offs and no outstanding requested changes you should feel free to merge yourself or ask for a merge.)

johnynek:
Okay, thanks for the merge. I wasn't sure if there was some reason for delay due to staging for releases.

djspiewak:
> Okay, thanks for the merge. I wasn't sure if there was some reason for delay due to staging for releases.

In my experience, some of the reviewers just drive-by review things and don't loop back to check the merging status. This has made cats reviews a bit of a two phase process, where the second pass is just going through and pressing **merge** on a bunch of already-reviewed PRs.

Honestly I don't think this is *entirely* problematic so long as the maintainers are doing these phases frequently enough, but it is certainly confusing for contributors.

PR #3209, opened 2019-12-13
Avoid infix, tuple, and function syntax when using kind-projector

travisbrown:
This is more a question than a proposal, I guess. I'm working on `-Ykind-projector` for Dotty, and I have a WIP implementation that I've confirmed can compile cats-core after #3207, #3193, and the change here (and one big [Simulacrum rewrite](https://github.com/travisbrown/simulacrum-scalafix)).

I think it makes sense to try to minimize the scope of Dotty's kind-projector compatibility, and right now my implementation only handles the type-level `λ[x => ...]` syntax and `*` in ordinary type application (`F[A, *, B, *, C]`). This is a really tiny change—around 50 lines in the parser behind a single `if (ctx.settings.YkindProjector.value)`. Adding support for `*` in tuple types, function types, and infix types would be a lot more invasive.

Maybe we should do it anyway, but for Cats at least only supporting `*` in type application isn't a very painful limitation—it only requires the changes here, and I think some of these changes (like `A Ior *` to `Ior[A, *]`) are readability improvements, anyway.

### DIFF
--- a/core/src/main/scala/cats/data/Ior.scala
+++ b/core/src/main/scala/cats/data/Ior.scala
@@ -811,7 +811,7 @@ sealed abstract private[data] class IorInstances extends IorInstances0 {
 
 sealed abstract private[data] class IorInstances0 {
 
-  implicit def catsDataTraverseFunctorForIor[A]: Traverse[A Ior *] = new Traverse[A Ior *] {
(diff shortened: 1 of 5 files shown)


travisbrown:
@djspiewak Okay, thanks! I think "yes, we need `*` in function types and tuples" is a reasonable response, and I'm happy to implement it and advocate for it if people want it (I just don't really care, myself).

travisbrown:
See #3215 for a subset of these changes that I think are reasonable.

PR #3217, opened 2019-12-18
Remove sbt-partial-unification

travisbrown:
Seems reasonable to get rid of this plugin dependency now that we can replace it with one line of build config.


travisbrown:
Oh, I didn't run `validateBC` locally, and need `-Ypartial-unification` there, too. Fixing now…

travisbrown:
Should go green now.

PR #3223, opened 2019-12-21
Non empty map order

diesalbla:
Overrides and closes https://github.com/typelevel/cats/pull/2318

The methods `fromMap` and `apply` in the NonEmptyMap companion object included an "Order" restriction on the Key type parameter. Although Keys should be ordered, this is not needed for these methods.

- So as to not break binary compatibility, we add some variants that still take the `Order[Key]` parameter, but
- In some tests or locations, where before they had the implicit Order, we now use a `null` value .

### DIFF
--- a/core/src/main/scala/cats/data/NonEmptyMapImpl.scala
+++ b/core/src/main/scala/cats/data/NonEmptyMapImpl.scala
@@ -14,15 +14,27 @@ private[data] object NonEmptyMapImpl extends NonEmptyMapInstances with Newtype2 
   private[data] def unwrap[K, A](m: Type[K, A]): SortedMap[K, A] =
     m.asInstanceOf[SortedMap[K, A]]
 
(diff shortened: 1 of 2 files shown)


joroKr21:
Wow, this goes really deep (all the way to `SortedMapInstances` in kernel).
I think we should should remove `Order` as constraint from most instances.
E.g. we can have `Semigroup[SortedMap[K, V]]` without `Order` but we can't have `Monoid`.

Also note that we get the incoherent ordering problem if we combine two sorted maps created with different orderings (even with the current implementation). But I guess we inherit this from Scala's collections so users shouldn't be too surprised by that.

joroKr21:
I mean something like this starting from `Hash` and `Eq` all the way to `NonEmptyMap` - #3397
is this something that makes sense?

joroKr21:
Superseded by #3473

PR #3232, opened 2019-12-30
Update Scalafmt to 2.3.2

travisbrown:
This is the update in #3228 but with the updated formatter actually run. Personally I like these changes, but if anyone has objections we could try to figure out whether it's possible to change the config to avoid them.


djspiewak:
Agreed these changes are preferable. I do prefer a style where multi-line lambdas are always written with curlies instead of parens, but baby steps. This is still better than it was.

PR #3233, opened 2019-12-31
Add when and unless to OptionT

strong-zero:
In scala 2.13, we have Option.{when, unless}. So should cats, I guess 🤔

### DIFF
--- a/core/src/main/scala/cats/data/OptionT.scala
+++ b/core/src/main/scala/cats/data/OptionT.scala
@@ -225,14 +225,14 @@ object OptionT extends OptionTInstances {
     λ[F ~> OptionT[F, *]](OptionT.liftF(_))
 
   /**
-   * Creates a nonempty `OptionT[F, A]` from an `A` value if the given condition is `true`.
+   * Creates a non-empty `OptionT[F, A]` from an `A` value if the given condition is `true`.
    * Otherwise, `none[F, A]` is returned. Analogous to `Option.when`.
    */
   def when[F[_], A](cond: Boolean)(a: => A)(implicit F: Applicative[F]): OptionT[F, A] =
     if (cond) OptionT.some[F](a) else OptionT.none[F, A]
 
   /**
-   * Creates a nonempty `OptionT[F, A]` from an `F[A]` value if the given condition is `true`.
+   * Creates a non-empty `OptionT[F, A]` from an `F[A]` value if the given condition is `true`.
    * Otherwise, the empty `OptionT[F, A]` is returned. Analogous to `Option.when`.
    */
   def whenF[F[_], A](cond: Boolean)(fa: => F[A])(implicit F: Applicative[F]): OptionT[F, A] =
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/data/OptionT.scala
```diff
@@ -223,6 +223,46 @@ object OptionT extends OptionTInstances {
...
+   */
+  def when[F[_], A](cond: Boolean)(a: => A)(implicit F: Applicative[F]): OptionT[F, A] =
+    if (cond) OptionT.some[F](a) else OptionT.none[F, A]
+
+  /**
+   * Creates a non-empty `OptionT[F, A]` from an `F[A]` value if the given condition is `true`.
+   * Otherwise, the empty `OptionT[F, A]` is returned. Analogous to `Option.when`.
+   */
+  def whenF[F[_], A](cond: Boolean)(fa: => F[A])(implicit F: Applicative[F]): OptionT[F, A] =
+    if (cond) OptionT.liftF(fa) else OptionT(F.map(fa)(_ => Option.empty))
```
fthomas: I find the implementation of the `else` branch surprising. I would have thought that it is just `OptionT.none[F, A]` if `cond` is `false` (analogous to `OptionT.when`) but the current implementation runs the `F`-effect and then replaces the `A` with `None: Option[A]`.

strong-zero: It seems odd and it works when F = List and fa = List.empty because we want to get `OptionT(List.empty)` not `OptionT(List(None))`. But it breaks the case `OptionT.whenF[List, Int](false)(List(1,2))` because we have `OptionT(List(None, None))` not `OptionT(List.empty)`😱

I will create a follow-up pull request to fix it.

I should have implemented the case when F assumes the empty value or F is a Monoid separately. Any comments and suggestions are welcome!

fthomas: If these properties imply that `val launchMissiles: IO[Unit] = ...; OptionT.whenF(cond)(launchMissiles)` always launches the missiles, no matter what `cond` is, then I think we should get rid of either these properties or `whenF`. I would prefer the former because `whenF` seems like a useful utility to me.

rossabaker: I agree with @fthomas.  Good catch.

How about this for `whenF`?  Many of the `F` methods on transformers are specified in terms of `Id`, and passes with Frank's proposed change:

```
  test("OptionT.whenF[Id, A] consistent with Option.when") {
    def when[A] = (c: Boolean, a: A) => if (c) Some(a) else None
    forAll { (i: Int, b: Boolean) =>
      OptionT.whenF[Id, Int](b)(i).value should ===(when(b, i))
    }
  }
```

We might even relate it to `.sequence` for less trivial effects:

```scala
forAll { (i: List[Int], b: Boolean) =>
  OptionT.whenF[List, Int](b)(i).value should ===(when(b, i).sequence)
}
```

strong-zero: @fthomas @rossabaker Thank you both for your helpful examples. I will make a follow-up pull request.

Issue #3143, opened 2019-11-13
Backport 2.1.0 changes for Scala 2.11


travisbrown:
As explained [here](https://www.reddit.com/r/scala/comments/dsf1xr/preparing_for_cats_210_help_wanted/) and in the [Cats 2019 roadmap](https://github.com/typelevel/cats/blob/master/ROADMAP_2019.md), we dropped support for Scala 2.11 on the Cats master branch after the 2.0.0 release, and we're relying on Scala 2.11 users to backport features from master to the `scala_2.11` branch for all future Cats releases.

We're planning to publish Cats 2.1.0-RC1 this Friday (15 November) and 2.1.0 by the end of this month. I've backported many of the build config changes and dependency updates to the 2.11 branch, but there are still many features that need to be backported in order for us to publish Cats 2.1.0 for Scala 2.11 (we're not necessarily aiming for 100% source compatibility between the 2.11 and 2.12 releases, but want to be close).

Note that many of the changes listed below are substantially more difficult to make on Scala 2.11 than on 2.12+ because adding methods to traits on 2.11 breaks binary compatibility (which is why we dropped decided to drop 2.11 support on master in the first place :smile:). You can see an example of what's necessary in [this pre-2.0 PR](https://github.com/typelevel/cats/pull/2609/files) adding an `ifA` method for `Apply`.

Pull requests currently merged on master that still need to be backported:

- [x]  Add `unzip` to `Functor`: #3062
- [ ]  Add `ifF` to `Functor`: #3040
- [x]  Add `Align`: #3076 
- [x]  Add `bifold` to `Bifoldable`: #3088
- [x]  Properly arrange `Parallel` instances: #3099
- [ ]  Fix `Vector` instances: #3100
- [x]  Add `attemptNarrow` to `ApplicativeErrorOps`: #2863
- [ ]  Add `TraverseFilter` instance for `Queue`: #3103
- [ ]  Fix `Eq` instances for tuples: #3105
- [x]  Add `minimumByOption`, etc.: #3084, #3122
- [x]  Add `sum` for `SemigroupK`: #3124
- [x]  Add `foldMapA`: #3135
- [x]  Add instances for `TailCalls.TailRec`: #3041
- [x]  Use `Foldable`'s `foldRight` for `Traverse`, etc.: #3015
- [x]  `CommutativeMonoid` for `Option` from `CommutativeSemigroup`: #2834
- [x]  Add `foldRightDefer` to `Foldable`: #2772
[...]

Issue #3008, opened 2019-08-26
Decide how public-facing instance traits should be


travisbrown:
Right now we have `AllInstances` and `AllSyntax` traits that are in the public API and are sometimes used outside of Cats itself—for example in Circe we have a `CirceSuite` helper class that extends these traits so that we don't have to import `cats.implicits._` in every test file.

The problem with this kind of usage is that `AllInstances` and `AllSyntax` don't actually include all of anything, ever since we started stacking on `BinCompatN` layers to maintain binary compatibility on Scala 2.11. For example, want a `Defer` for `Function0`? You won't get it from `AllInstances`—you also need `AllInstancesBinCompat0`.

This means anyone using `AllInstances` or `AllSyntax` has to be careful to add the latest `AllSyntaxBinCompat37` or whatever traits included in each new Cats release if they actually want everything.

This is something we haven't managed to stay on track on even within Cats itself—e.g. see #3007, which is due to the fact that `cats.implicits` is lagging two `AllSyntaxBinCompat`s behind, or `SyntaxSuite`, which is four `AllInstanceBinCompat`s out of date (but happens to still be working), etc.

To make things even more confusing, there's a `AllSyntaxBinCompat` (note the missing number) abstract class that bundles up all the syntax bin-compat traits, but no equivalent `AllInstancesBinCompat`.

I've also recently proposed (and merged) making the bin-compat traits package-private (#3003), which on second thoughts I think we'll have to undo, at least until 2.1.

So there's a lot of miscellaneous little brokenness and inconsistency and mildly bad UX here, some of which I'm about to open a PR to address, but there's also a bigger question of whether we actually intend the `AllInstances` and `AllSyntax` traits to be used outside of the implementation of Cats itself.

Daenyth:
Just as a data point, I've used those in third-party code so that I could make one combined import for cats implicits + my library implicits + another library implicits, to help reduce brain effort on imports

djspiewak:
As another random data point… I originally structured djspiewak/shims to hide the "All" traits, so the only way to actually get at anything was via the `shims._` import. The problem with this is it made it impossible for any third parties to participate in the implicit prioritization, which created unavoidable conflicts. Given Cats' nature, I would suspect similar problems if we privatized the traits.

joroKr21:
But this is only relevant for syntax now, isn't it?

djspiewak:
> But this is only relevant for syntax now, isn't it?

[...]

PR #3234, opened 2019-12-31
Backport unzip method in functor typeclass for scala_2.11

gagandeepkalra:
addresses #3143


gagandeepkalra:
If this looks good I can start working on other subtask for #3143

gagandeepkalra:
I need some help with this error on running `docs/makeMicrosite`

> [error] (docs / Jekyll / mappings) java.io.IOException: Cannot run program "jekyll" (in directory "/Users/gagandeepkalra/IdeaProjects/cats/docs/target/scala-2.11/resource_managed/main/jekyll"): error=2, No such file or directory`

Seems like it's failing for scala_2.11 branch altogether, I'm not very sure how to solve this.

rossabaker:
Have you installed jekyll?  It's a Ruby utility, and the right way to [install it](https://jekyllrb.com/docs/installation/#guides) depends on your OS.

gagandeepkalra:
@rossabaker thank you, now I have, and `docs/makeMicrosite` runs successfully on both my branch and `scala_2.11`. Now I wonder why is the build failing at all. I tried restarting with a dummy commit (amend) thinking it might just be flaky, but no.

build error,
>[error] /home/travis/.rvm/gems/ruby-2.5.3/gems/octokit-4.14.0/lib/octokit/middleware/follow_redirects.rb:14:in `<module:Middleware>': uninitialized constant Faraday::Error::ClientError (NameError)
[error] Did you mean?  Faraday::ClientError

ends up with.
>[error] (docs / Jekyll / mappings) Could not run jekyll, error: 1

rossabaker:
It suspect it's related to https://github.com/typelevel/cats/pull/3218.  There was some recent bitrot in jekyll that broke several projects using sbt-microsites, and it doesn't look like any fix was backported to this branch.

core/src/main/scala/cats/syntax/all.scala
```diff
@@ -93,3 +94,5 @@ trait AllSyntaxBinCompat4
 trait AllSyntaxBinCompat5 extends ParallelBitraverseSyntax
 
 trait AllSyntaxBinCompat6 extends ParallelUnorderedTraverseSyntax
+
+trait AllSyntaxBinCompat7 extends FunctorSyntaxBinCompat0
```
rossabaker: I can't remember why some syntax traits are package private and some aren't.  I can't think of a reason this couldn't and shouldn't be, but I also was not very active in developing this idiom.

travisbrown: @rossabaker As far as I know there's not a good answer to this question—like a lot of stuff in Cats it's just the result of a lot of uncoordinated decisions. I opened a related issue [here](https://github.com/typelevel/cats/issues/3008) a few months ago.

rossabaker: @gagandeepkalra has been through a lot of incidental distractions already. I'm good with punting on it for the purposes of this ticket.

gagandeepkalra:
Thank you 😊

core/src/main/scala/cats/syntax/all.scala
gagandeepkalra: thank you

PR #3235, opened 2019-12-31
Added flatMapK to FreeT

djspiewak:
The analogous function on `Free` is simply `foldMap`, but `FreeT` has an extra (monadic) suspension functor which cannot be manipulated through the same mechanism. `mapK` already exists; this simply provides a way of doing a `mapK`-like transformation where the resulting `N` is *itself* a `FreeT`. This can be useful when you want to manipulate the structure of the interpretation in a composable fashion.

### DIFF
--- a/free/src/main/scala/cats/free/FreeT.scala
+++ b/free/src/main/scala/cats/free/FreeT.scala
@@ -32,6 +32,19 @@ sealed abstract class FreeT[S[_], M[_], A] extends Product with Serializable {
         Suspend(mn(m))
     }
 
+  /**
+   * Modify the context `M` using the transformation `mn`, flattening the free
+   * suspensions into the outer.
+   */
+  def flatMapK[N[_]: Monad](mn: M ~> FreeT[S, N, *])(implicit S: Functor[S]): FreeT[S, N, A] = {
+    def loop(ftft: FreeT[S, FreeT[S, N, *], A]): FreeT[S, N, A] =
+      ftft.resume.flatMap { e =>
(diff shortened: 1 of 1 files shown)


free/src/main/scala/cats/free/FreeT.scala
```diff
@@ -32,6 +32,19 @@ sealed abstract class FreeT[S[_], M[_], A] extends Product with Serializable {
         Suspend(mn(m))
     }
 
+  /**
+   * Modify the context `M` using the transformation `mn`, flattening the free
+   * suspensions into the outer.
+   */
+  def flatMapK[N[_]: Monad](mn: M ~> FreeT[S, N, *])(implicit S: Functor[S]): FreeT[S, N, A] = {
```
travisbrown: I think we should try to avoid mixing context bounds and implicit parameter sections, in part because it's confusing (there's a [whole subgenre](https://twitter.com/travisbrown/status/1097407766082347008) of Stack Overflow questions related to this), but also because in Dotty context bounds will desugar to `given`s, which will have slightly different semantics.

Also it seems like this should follow the rule of thumb that the order of constraints should follow the order the types appear in—so:

```scala
   def flatMapK[N[_]](mn: M ~> FreeT[S, N, *])(implicit S: Functor[S], N: Monad[N]): FreeT[S, N, A]
```
Which is the opposite order of what the current code desugars to.

djspiewak: I don't mind changing it. I mix context bounds all the time, but obviously if Dotty is changing the semantics then I should probably break that habbit.

travisbrown:
@LukaJCB I think you :+1:'d this in January but I can't tell from this GitHub interface—do you mind taking a quick look?

PR #3236, opened 2020-01-01
Remove redundant `IfApplyOps`

gagandeepkalra:
remove now redundant `IfApplyOps`, earlier used to ensure binary compatibility


travisbrown:
I think for now the best we can do here is what I did in #3152 (add "real" methods but leave the syntax). If you can come up with a way to get this to pass `validateBC`, though, I'm all for it!

gagandeepkalra:
yes, you're right. I'll close this one then. Thank you.

PR #3237, opened 2020-01-04
Follow up to #3233

strong-zero:
A follow up to #3233 based on the helpful comments by @fthomas and @rossabaker:  `OptionT.{whenF, unlessF}` should return `OptionT.none` when the given condition is `false`.

### DIFF
--- a/tests/src/test/scala/cats/tests/OptionTSuite.scala
+++ b/tests/src/test/scala/cats/tests/OptionTSuite.scala
@@ -278,14 +278,16 @@ class OptionTSuite extends CatsSuite {
     }
   }
 
-  test("OptionT.when[Id, A] consistent with the same implementation of Option.when") {
-    val when = (c: Boolean, j: Int) => if (c) Some(j) else None
+  test("OptionT.when[Id, A] consistent with Option.when") {
+    // Option.when is inlined here because it is not available before Scala 2.13
+    def when[A]: (Boolean, A) => Option[A] = (c: Boolean, a: A) => if (c) Some(a) else None
     forAll { (i: Int, b: Boolean) =>
       OptionT.when[Id, Int](b)(i).value should ===(when(b, i))
     }
   }
 
   test("OptionT.whenF[F, A] consistent with Option.when") {
+    // Option.when is inlined here because it is not available before Scala 2.13
     def when[A]: (Boolean, A) => Option[A] = (c: Boolean, a: A) => if (c) Some(a) else None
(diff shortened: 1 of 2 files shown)


tests/src/test/scala/cats/tests/OptionTSuite.scala
```diff
@@ -285,10 +285,14 @@ class OptionTSuite extends CatsSuite {
...
-    forAll { (li: List[Int], b: Boolean) =>
-      OptionT.whenF(b)(li).value should ===(li.map(when(b, _)))
+  test("OptionT.whenF[F, A] consistent with Option.when") {
+    def when[A]: (Boolean, A) => Option[A] = (c: Boolean, a: A) => if (c) Some(a) else None
+    forAll { (i: Int, b: Boolean) =>
+      OptionT.whenF[Id, Int](b)(i).value should ===(when(b, i))
+    }
+
+    forAll { (i: List[Int], b: Boolean) =>
+      OptionT.whenF[List, Int](b)(i).value should ===(when(b, i).sequence)
```
rossabaker: I'm unsure whether I'd ever reason about code by this property, but I noticed it when I thought about the types and it seems to pass.  I'm curious whether others think this one is actually useful.

tests/src/test/scala/cats/tests/OptionTSuite.scala
```diff
@@ -285,10 +285,14 @@ class OptionTSuite extends CatsSuite {
     }
   }
 
-  test("OptionT.whenF[F, A] consistent with the same implementation of Option.when") {
-    val when = (c: Boolean, j: Int) => if (c) Some(j) else None
-    forAll { (li: List[Int], b: Boolean) =>
-      OptionT.whenF(b)(li).value should ===(li.map(when(b, _)))
+  test("OptionT.whenF[F, A] consistent with Option.when") {
+    def when[A]: (Boolean, A) => Option[A] = (c: Boolean, a: A) => if (c) Some(a) else None
```
rossabaker: Not a blocker, but since the property refers to `Option.when`, it might be helpful to future readers to comment why it's inlined here (e.g., that it's not available before 2.13).

tests/src/test/scala/cats/tests/OptionTSuite.scala
strong-zero: The test case is very interesting to me, but certainly less obvious that it universally holds (not only in the case when F = List). 

Added abd798b to remove it for now.
:+1: to restore it if someone knows a good explanation about it.

PR #3238, opened 2020-01-04
backport #3218 to scala_2.11, fixes error with `sbt-microsites`

gagandeepkalra:


rossabaker:
I'm not sure why the build is failing.  I noticed [this tweet about 403s](https://twitter.com/dwijnand/status/1213406224450048001), but I would think caching `$HOME/.cache` already covers the recommendation there.

rossabaker:
I should have thought of this earlier, but we could probably disable the site generation altogether on this branch.  I don't think we intend to publish the site from it.  It increases the divergence from the main branch, which is bad, but it also cuts off something that tends to bitrot, which is good.

Generating the scaladoc is still a requirement for Maven Central, so it probably makes sense to run `sbt doc` (but not unidoc) in Travis, and fix the errors from that last build.

gagandeepkalra:
@rossabaker I think the `publishMicrosite` is only enabled from [master](https://github.com/typelevel/cats/blob/master/scripts/travis-publish.sh#L25) already. 

The build fixed itself somehow, should I still go ahead and update `.travis.yml` to include just `docs`.

PR #3239, opened 2020-01-06
add List#scanLeftNel and List#scanRightNel

enzief:
`.scanLeft` result is always non-empty.


LukaJCB:
Hey @enzief, thanks for the contribution, this seems pretty useful! Do you mind adding a test or two? Maybe a doctest? :) Much appreciated!

rossabaker:
I think a `.scanRightNel` would also be possible, if you're feeling ambitious. :smile:

enzief:
@LukaJCB @rossabaker another look please.

PR #3241, opened 2020-01-06
Reimplemented MonadError[FreeT[...]] to be correct

### LINKED ISSUE #3240: FreeT violates the monad laws
```scala
import cats.{Id, MonadError}
import cats.data.Kleisli
import cats.implicits._
import cats.free.FreeT

type F[A] = Kleisli[FreeT[Id, Either[Int, ?], ?], Unit, A]
val F = MonadError[F, Int]

def run[A](fa: F[A]): Either[Int, A] =
  fa.run(()).runM(ft => Right(ft))

def map1[A, B](fa: F[A])(f: A => B): F[B] =
  fa.map(f)

def map2[A, B](fa: F[A])(f: A => B): F[B] =
  fa.flatMap(a => f(a).pure[F])

val eff1 = F.handleErrorWith(map1(F.fromEither(42.asLeft[String]))(_.asRight[Int]))(_.asLeft[String].pure[F])
val eff2 = F.handleErrorWith(map2(F.fromEither(42.asLeft[String]))(_.asRight[Int]))(_.asLeft[String].pure[F])

run(eff1)   // => Right(Left(42))
run(eff2)   // => Left(42)
```

The only reason we don't see this on a regular basis is the *order* in which the traits are composed together within `Kleisli`. For example:

```scala
private[data] trait KleisliMonad[F[_], A]
    extends Monad[Kleisli[F, A, *]]
    with KleisliFlatMap[F, A]
    with KleisliApplicative[F, A] {
  implicit def F: Monad[F]
}
```

This means we get `Applicative`'s `map` rather than `Monad`'s, which avoids the bug.

I *believe* this is being caused by the fact that `flatMap` is `shift`ed but `map` is not:

```scala
  def map[C](f: B => C)(implicit F: Functor[F]): Kleisli[F, A, C] =
    Kleisli(a => F.map(run(a))(f))

// ...

  def flatMap[C, AA <: A](f: B => Kleisli[F, AA, C])(implicit F: FlatMap[F]): Kleisli[F, AA, C] =
    Kleisli.shift(a => F.flatMap[B, C](run(a))((b: B) => f(b).run(a)))
```

djspiewak:
There's an explanation of this in the comments in the code. I retained the old implementation for binary compatibility, but it is somewhat unfortunate since any code which is hitting that is outright-broken and people may not know it. Then again, I'm not sure how many people (other than me) are using `FreeT`, so maybe it's okay either way.

Fixes #3240

### DIFF
--- a/free/src/main/scala/cats/free/FreeT.scala
+++ b/free/src/main/scala/cats/free/FreeT.scala
@@ -240,7 +240,7 @@ sealed abstract private[free] class FreeTInstances extends FreeTInstances0 {
       override def M = E
 
       /*
-       * Quick explanation... The previous version of this function (retained about for
+       * Quick explanation... The previous version of this function (retained above for
        * bincompat) was only able to look at the *top* level M[_] suspension in a Free
        * program. Any suspensions below that in the compute tree were invisible. Thus,
        * if there were errors in that top level suspension, then they would be handled
--- a/free/src/main/scala/cats/free/FreeT.scala
+++ b/free/src/main/scala/cats/free/FreeT.scala
@@ -220,11 +220,61 @@ object FreeT extends FreeTInstances {
 }
 
 sealed abstract private[free] class FreeTInstances extends FreeTInstances0 {
-  implicit def catsFreeMonadErrorForFreeT[S[_], M[_], E](implicit E: MonadError[M, E]): MonadError[FreeT[S, M, *], E] =
+
+  // retained for binary compatibility. its results are incorrect though and it would fail the laws if we generated things of the form pure(()).flatMap(_ => fa)
+  private[this] def catsFreeMonadErrorForFreeT[S[_], M[_], E](implicit E: MonadError[M, E]): MonadError[FreeT[S, M, *], E] =
     new MonadError[FreeT[S, M, *], E] with FreeTMonad[S, M] {
       override def M = E
       override def handleErrorWith[A](fa: FreeT[S, M, A])(f: E => FreeT[S, M, A]) =
         FreeT.liftT[S, M, FreeT[S, M, A]](E.handleErrorWith(fa.toM)(f.andThen(_.toM)))(M).flatMap(identity)
+      override def raiseError[A](e: E) =
+        FreeT.liftT(E.raiseError[A](e))(M)
+    }
+
+  implicit def catsFreeMonadErrorForFreeT2[S[_], M[_], E](implicit E: MonadError[M, E], S: Functor[S]): MonadError[FreeT[S, M, *], E] =
+    new MonadError[FreeT[S, M, *], E] with FreeTMonad[S, M] {
+      override def M = E
+
+      /*
+       * Quick explanation... The previous version of this function (retained about for
+       * bincompat) was only able to look at the *top* level M[_] suspension in a Free
+       * program. Any suspensions below that in the compute tree were invisible. Thus,
+       * if there were errors in that top level suspension, then they would be handled
+       * by the delegate. Errors buried further in the tree were unhandled. This is most
+       * easily visualized by the following two expressions:
+       *
+       * - fa
+       * - pure(()).flatMap(_ => fa)
+       *
+       * By the monad laws, these *should* be identical in effect, but they do have
+       * different structural representations within FreeT. More specifically, the latter
+       * has a meaningless M[_] suspension which sits in front of the rest of the
+       * computation. The previous iteration of this function would be blind to fa in
+       * the latter encoding, while it would handle it correctly in the former.
+       *
+       * Historical sidebar: this became visible because of the "shift" mechanism in
+       * Kleisli.
+       */
+      override def handleErrorWith[A](fa: FreeT[S, M, A])(f: E => FreeT[S, M, A]) = {
+        val ft = FreeT liftT[S, M, FreeT[S, M, A]] {
+          val resultsM = E.map(fa.resume) {
+            case Left(se) =>
+              FreeT.liftF[S, M, FreeT[S, M, A]](S.map(se)(handleErrorWith(_)(f))).flatMap(identity)
+
+            case Right(a) =>
+              pure(a)
+          }
+
+          E.handleErrorWith(resultsM) { e =>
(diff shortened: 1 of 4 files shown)


free/src/main/scala/cats/free/FreeT.scala
```diff
@@ -220,14 +220,67 @@ object FreeT extends FreeTInstances {
...
         FreeT.liftT(E.raiseError[A](e))(M)
     }
+
+  implicit def catsFreeMonadErrorForFreeT2[S[_], M[_], E](implicit E: MonadError[M, E],
+                                                          S: Functor[S]): MonadError[FreeT[S, M, *], E] =
+    new MonadError[FreeT[S, M, *], E] with FreeTMonad[S, M] {
+      override def M = E
+
+      /*
+       * Quick explanation... The previous version of this function (retained about for
```
LukaJCB: typo: `...retained **above** for...`

kubukoz:
Is it possible to adjust the generators / equalities used in tests to verify that the old implementation does indeed break a law?

kailuowang:
Thanks for spotting and fixing this bug. 
A couple questions if you don't mind: 
It seems that making it `private[this]` still breaks BC, probably need to stick to package private. 
Why keeping the old faulty implementation (v.s. letting the old method delegate to the new correct one)? 
It kind of itches that a faulty `handleErrorWith` implementation doesn't break the `ApplicativeErrorLaws`. There is no way to add a new law to capture this right?

djspiewak:
@kubukoz I've been wondering the same thing myself. I tried fiddling around with some sort of distributivity law for `handleErrorWith` but it's surprisingly subtle, because you still need to ensure ordered short-circuiting. I would definitely like to codify this better.

@kailuowang See above. :-) I want to add a law. Thoughts welcome. My attempt was something like this:

```scala
@kubukoz I've been wondering the same thing myself. Something like this perhaps in `MonadErrorLaws`? We can't work with `ApplicativeErrorLaws` unfortunately because we don't want to accidentally force `handleErrorWith` to somehow handle errors raised by its handler:

```scala
def handleErrorWithDistributesOverAp[A, E](fa1: F[A], fa2: F[A], h: E => F[A]): F[A] =
  F.handleErrorWith(fa1 *> fa2)(h) <-> (F.handleErrorWith(fa1)(h) *> F.handleErrorWith(fa2)(h))
```

That's not actually right, though, because `fa1` should short-circuit `fa2` if it errors, whereas it will not in the right side.

> Why keeping the old faulty implementation (v.s. letting the old method delegate to the new correct one)?

The constraints are different. The new requires a `Functor[S]`. I *am* tempted to just add a `sys.error` though, because any code using the old implementation is broken in a very dangerous way.

> It seems that making it private[this] still breaks BC, probably need to stick to package private.

Actually it has to be public. The problem is the static forwarder. I've fixed it now.

kailuowang:
Thanks @djspiewak I missed the new constraint. This is source breaking then, I am adding the label just so that it doesn't surprise users.

djspiewak:
How about this as a law?

```scala
def attemptDistributivity[A, B](fa: F[A], f: A => F[B]): F[Either[E, B]] =
  F.attempt(F.flatMap(fa)(f)) <-> F.flatMap(F.attempt(fa))(_.traverse(f.andThen(F.attempt)))
```

djspiewak:
(btw, please don't merge this just yet; even aside from the discussion about a new law, which is important, I have some ancillary evidence that this isn't appropriately stack-safe.

djspiewak:
@kailuowang @kubukoz Okay, I've added a pair of laws in ApplicativeError which capture this case. I've verified that they do indeed fail on `FreeT` in master. They aren't *quite* as strong as what I would like (explained in a comment), but I think they should appropriately constrain any non-pathological implementations.

I've also added a `Defer` for `FreeT`, which was previously absent, and used it to ensure the stack-safety of `handleErrorWith` even when the suspension monad does not provide a stack-safe `flatMap`. An obvious objection here is that, if the suspension monad is not stack-safe, then presumably stack-safety in `handleErrorWith` is just delaying the inevitable. However, it's entirely possible to run `handleErrorWith` on `FreeT` when the suspension monad is not stack-safe, and then subsequently `mapK` into a stack-safe context prior to interpretation. That use-case was impossible with my initial implementation.

laws/src/main/scala/cats/laws/ApplicativeErrorLaws.scala
```diff
@@ -54,6 +54,21 @@ trait ApplicativeErrorLaws[F[_], E] extends ApplicativeLaws[F] {
 
   def redeemDerivedFromAttemptMap[A, B](fa: F[A], fe: E => B, fs: A => B): IsEq[F[B]] =
     F.redeem(fa)(fe, fs) <-> F.map(F.attempt(fa))(_.fold(fe, fs))
+
+  /*
+   * These laws, taken together with applicativeErrorHandle, show that errors dominate in
+   * ap, *and* show that handle has lexical semantics over ap. F.unit is used in both laws
+   * because we don't have another way of expressing "an F[_] which does *not* contain any
```
kubukoz: Another way might be `Arbitrary[F[_]].map(_.attempt)`... But that's probably problematic in other ways (depending on the behavior of attempt)

djspiewak: Well, the problem is then you need a function `E => A`. That's not doable without breaking binary compatibility.

kubukoz: Why binary compatibility? I fail to see that here...

djspiewak: In order to have an `Arbitrary[E => A]`, we need a `Cogen[E]` and an `Arbitrary[A]`. We have the former, but not the latter (in `ApplicativeErrorTests`). In order to *get* the latter, we would need to add it as a parameter to the `applicativeError` function, which breaks binary compatibility.

kubukoz: Ohhh. Yeah, you're absolutely right.

PR #3242, opened 2020-01-09
Backport `bifold` in `Bifoldable` typeclass to scala_2.11

gagandeepkalra:
addresses #3143

### DIFF
--- a/core/src/main/scala/cats/syntax/bifoldable.scala
+++ b/core/src/main/scala/cats/syntax/bifoldable.scala
@@ -6,11 +6,11 @@ import scala.language.implicitConversions
 trait BifoldableSyntax extends Bifoldable.ToBifoldableOps
 
 private[syntax] trait BiFoldableSyntaxBinCompat0 {
-  implicit final def catsSyntaxBiFoldOps[F[_, _], A, B](fa: F[A, B]): BifoldBifoldableOps[F, A, B] =
-    new BifoldBifoldableOps[F, A, B](fa)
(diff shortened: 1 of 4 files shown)


gagandeepkalra:
I was hoping to augment this PR with another `Bifoldable` law, but couldn't get around binary compatibility. I settled with a `doctest` for now.

core/src/main/scala/cats/syntax/bifoldable.scala
```diff
@@ -1,4 +1,29 @@
 package cats
 package syntax
 
+import scala.language.implicitConversions
+
 trait BifoldableSyntax extends Bifoldable.ToBifoldableOps
+
+private[syntax] trait BiFoldableSyntaxBinCompat0 {
+  implicit final def catsSyntaxBiFoldOps[F[_, _], A, B](fa: F[A, B]): BifoldBifoldableOps[F, A, B] =
```
rossabaker: It's all implicit or private, so it doesn't matter that much, but everything else is `bifold`, not `biFold`.  Would b egood to remain consistent.

core/src/main/scala/cats/syntax/bifoldable.scala
```diff
@@ -1,4 +1,29 @@
...
+    new BifoldBifoldableOps[F, A, B](fa)
+}
+
+final class BifoldBifoldableOps[F[_, _], A, B](private val fab: F[A, B]) extends AnyVal {
```
rossabaker: I think it's more consistent to just call this `BifoldableOps`.

PR #3243, opened 2020-01-09
Remove unused Scala Native sbt plugin

travisbrown:
We don't support Scala Native or even Scala 2.11 and don't need this plugin sitting around.


travisbrown:
Updated to get both of them.

PR #3246, opened 2020-01-09
Minor changes for Scala.js 1.0 compatibility

travisbrown:
These changes make it possible to build Cats on Scala.js 1.0 (currently 1.0.0-RC2) via an environmental variable, without actually changing anything significant about the current build, which is still on Scala.js 0.6.

This approach is currently used by Simulacrum, Discipline (although Discipline does test Scala.js 1.0.0-RC2 in CI, and this change doesn't), etc.

I've confirmed that the following succeeds:

```
SCALAJS_VERSION=1.0.0-RC2 sbt +validateJS
```

### DIFF
--- a/js/src/test/scala/cats/tests/FutureTests.scala
+++ b/js/src/test/scala/cats/tests/FutureTests.scala
@@ -8,22 +8,28 @@ import cats.js.instances.Await
 import cats.js.instances.future.futureComonad
 import cats.tests.{CatsSuite, ListWrapper}
 
-import scala.concurrent.Future
+import scala.concurrent.{ExecutionContextExecutor, Future}
 import scala.concurrent.duration._
 
 import org.scalacheck.{Arbitrary, Cogen}
 import org.scalacheck.Arbitrary.arbitrary
 import cats.laws.discipline.arbitrary._
 
-// https://issues.scala-lang.org/browse/SI-7934
-@deprecated("", "")
-class DeprecatedForwarder {
-  implicit def runNow = scala.scalajs.concurrent.JSExecutionContext.Implicits.runNow
-}
-object DeprecatedForwarder extends DeprecatedForwarder
-import DeprecatedForwarder.runNow
-
 class FutureTests extends CatsSuite {
+  // Replaces Scala.js's `JSExecutionContext.runNow`, which is removed in 1.0.
+  // TODO: We shouldn't do this! See: https://github.com/scala-js/scala-js/issues/2102
+  implicit private object RunNowExecutionContext extends ExecutionContextExecutor {
+    def execute(runnable: Runnable): Unit =
+      try {
+        runnable.run()
+      } catch {
+        case t: Throwable => reportFailure(t)
+      }
+
+    def reportFailure(t: Throwable): Unit =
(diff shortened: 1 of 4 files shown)


js/src/test/scala/cats/tests/FutureTests.scala
```diff
@@ -8,22 +8,28 @@ import cats.js.instances.Await
...
-@deprecated("", "")
-class DeprecatedForwarder {
-  implicit def runNow = scala.scalajs.concurrent.JSExecutionContext.Implicits.runNow
-}
-object DeprecatedForwarder extends DeprecatedForwarder
-import DeprecatedForwarder.runNow
-
 class FutureTests extends CatsSuite {
+  // Replaces Scala.js's `JSExecutionContext.runNow`, which is removed in 1.0.
+  // TODO: We shouldn't do this! See: https://github.com/scala-js/scala-js/issues/2102
```
Daenyth: It's probably worth creating and linking to a github issue about fixing this item

travisbrown: Done: #3247

djspiewak:
This effectively means that `sbt release` will not work anymore, since the release needs to be split across two separate `sbt` runs. I don't know another way to do this, but it also means that, in order to keep the release documented and reproducible, we should toss a shell script in here somewhere which replaces the functionality we're currently getting from sbt-release while juggling the environment variables.

travisbrown:
@djspiewak Sure, I'll add a script. My preference would be to keep this fairly low-key, though—I don't want to commit to cross-building in the long term, or to make CI builds even longer by testing 1.0.0-RC2 on every PR now.

dwijnand:
Perhaps you could switch to a system property, and do something like this:

```
  ...
  publishArtifacts,
  releaseStepCommand("""eval sys.prop("scalajs.version") = "1.0.0-RC2"; reload"),
  publishArtifacts,
  ...
)
```

travisbrown:
@dwijnand Noooooo, please, no. Every time I've ever done anything like that in any build I've regretted it. All of this stuff is already a fragile pile of crap on the edge of disaster. I want to unblock people who want 1.0.0 primarily because I want us to be in a position to drop 0.6 the minute 1.0.0 final is out, and that's unlikely to be more than one or two Cats releases away. I think we should stick to the absolute minimum amount of extra process.

dwijnand:
WFM 😄

travisbrown:
@dwijnand In any case that would publish the JVM artifacts twice.

dwijnand:
Yeah, but I think that's fine if you're using the new sonatype bundle thing.

travisbrown:
And here's the 2.1.0 backport branch, which also passes `SCALAJS_VERSION=1.0.0-RC2 sbt +validateJS`: https://github.com/typelevel/cats/tree/backport/scala.js-1.0/2.1.0

travisbrown:
@djspiewak Any objection to merging now?

cquiroz:
@travisbrown Thanks for this. it would help unlock a big part of the ecosystem that can't be ported to RC2 just yet

PR #3248, opened 2020-01-11
Remove 2.12 and 2.13 tests from 2.11 branch

gagandeepkalra:


gagandeepkalra:
@rossabaker yes, but 2.12 and 2.13 modules have already been removed from the `scala_2.11` branch, given this we maybe can remove the tests now.

rossabaker:
I'm not sure I made my concern clear.

1. We delete these tests.
2. A feature on master updates code and touches these tests.
3. We cherry-pick that back to this branch.
4. We get a modification-delete merge conflict.

If we just leave these tests alone as dead code, we would have less exposure to these type of conflicts, and future backporting will be easier.  Then again, maybe backporting a PR that touched version-specific code should force us to think.  I guess I could argue this one from both sides.

If this doesn't dissuade you, I don't object to it.

gagandeepkalra:
sure, make sense. Do we want to keep this PR open for now and merge after all promised has been back-ported? or before last release is decided.

rossabaker:
I think if we're going to merge it, it makes more sense to do it now than leave it open.

PR #3259, opened 2020-01-17
Update discipline-scalatest

travisbrown:
Okay, trying this again, with [1.0.0-RC4](https://github.com/typelevel/discipline-scalatest/releases/tag/v1.0.0-RC4).

-----

Update: the NPEs were because of an initialization issue in `cats.kernel.laws.Test`, which was happening because the implicit `PropertyCheckConfiguration` is now being resolved where `checkAll` is called instead of in discipline-scalatest (where previously it always used `generatorDrivenConfig`).

Unfortunately ScalaTest's `org.scalatest.prop.Configuration` defines `generatorDrivenConfig` as a `val`, so we can't just slap a `lazy` on the override. Instead I've added a configuration class that makes the initialization work out right.

This is the only place this issue showed up either here or in [Circe](https://github.com/circe/circe/pull/1369), which I've also updated to discipline-scalatest 1.0.0-RC4, and I think it's likely to be rare—it's only showing up here because of the kind of unusual arrangement of Scala-version-specific tests.

I also think the new behavior is the correct behavior. If the user was writing tests in terms of `check` directly, the implicit configuration would be resolved at the point where they call `check`—it wouldn't just use `generatorDrivenConfig` automatically. In my view our `checkAll` should do the same.

(I'll note the issue in the final 1.0.0 release notes.)

### DIFF
--- a/alleycats-tests/shared/src/test/scala/alleycats/tests/AlleycatsSuite.scala
+++ b/alleycats-tests/shared/src/test/scala/alleycats/tests/AlleycatsSuite.scala
@@ -8,7 +8,7 @@ import cats.syntax.{AllSyntax, EqOps}
 import cats.tests.StrictCatsEquality
 import org.scalatest.funsuite.AnyFunSuiteLike
 import org.scalatestplus.scalacheck.ScalaCheckDrivenPropertyChecks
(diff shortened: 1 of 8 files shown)


travisbrown:
Will look at the 2.13 failures in the morning.

PR #3263, opened 2020-01-20
Avoid syntax methods in implementations in cats-core

travisbrown:
This weekend I ran into a [Dotty bug](https://github.com/lampepfl/dotty/issues/8035) that turned up because of our use of `cats.syntax` methods in `cats.data`. The issues are easy enough to work around, but it's a good reminder that value classes are a mess and we might want to avoid extension methods in our own implementations.

This PR replaces all use of `cats.syntax` extension methods in `cats/*.scala` and `cats/data/*.scala`. Most of the replacements have a fairly neutral effect on readability, like using `Ior.fromEither` instead of `_.toIor`, or `algebra.remove(x, y)` instead of `x |-| y`, etc. In these cases we get to avoid some indirection and exposure to value class nonsense without bloating the code.

In a few cases the non-syntax version is a significantly more verbose—in particular we use `bimap` on `Either` a lot in `cats.data`. This is one case where the syntax method does have a non-trivial runtime cost. I've added a benchmark for two implementations of `map` for `EitherK`: the current one:

```scala
EitherK(e.run.bimap(F.lift(f), G.lift(f)))
```

…and one that doesn't use the `bimap` extension method:

```scala
EitherK(
  e.run match {
    case Right(ga) => Right(G.map(ga)(f))
    case Left(fa)  => Left(F.map(fa)(f))
  }
)
```
The non-syntax one gets 30% more throughput in this benchmark:

```scala
EitherKMapBench.incMap1  thrpt   20  19348653.553 ±  75679.278  ops/s
EitherKMapBench.incMap2  thrpt   20  25152245.244 ± 161475.884  ops/s
```
Given that the faster one is still arguably _more_ readable, even if it's a little more verbose, I think avoiding `bimap` in our own implementations is reasonable.

### DIFF
--- a/bench/src/main/scala/cats/bench/EitherKMapBench.scala
+++ b/bench/src/main/scala/cats/bench/EitherKMapBench.scala
@@ -1,0 +1,33 @@
+package cats.bench
+
+import cats.Functor
+import cats.data.EitherK
+import cats.implicits._
+import org.openjdk.jmh.annotations.{Benchmark, Scope, State}
+
+@State(Scope.Benchmark)
+class EitherKMapBench {
+  def map1[F[_], G[_], A, B](e: EitherK[F, G, A])(f: A => B)(implicit F: Functor[F], G: Functor[G]): EitherK[F, G, B] =
+    EitherK(e.run.bimap(F.lift(f), G.lift(f)))
+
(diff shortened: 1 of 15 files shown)


travisbrown:
For the record this isn't something I feel particularly strongly about—I just thought it was worth seeing how much noise this would add, and it turns out it isn't much. I'm happy to change the PR to only include the easy replacements if people would prefer that.

PR #186, opened 2015-02-12
move the Id type alias to the Id object

stew:
this fixes #185
also adds docs to the website for Id, so if you merge, please consider running ghpagesPushSite.

Issue #2276, opened 2018-06-02
Id instances cannot be removed from the scope


djspiewak:
This is the best thing ever:

```scala
cats.Monad[cats.Id]
```

No imports. That compiles. Credit to @edmundnoble for finding the original issue here: https://github.com/djspiewak/shims/issues/20

The problem here is that `catsInstancesForId` is defined in the `cats` package object. This is a problem because the `cats` package is a *component* of the `cats.Id` type, and the `cats` package object is its own companion, meaning that the `cats` package object part of the implicit scope (without any imports!) used when resolving *any* type in the `cats` package (or presumably, any subpackage). This would also be a problem if other implicits were defined in the `cats` package object, but the `Id` instances are the only ones.

As far as I can tell, this scoping issue is pretty innocuous (if a bit leaky) if you're *only* using cats. Unfortunately, the moment you have scalaz on the classpath (you don't even need imports!) and are using shims, you end up in a world of hurt. See the issue I linked for an example.

The solution here is to unfortunately break compatibility. Implicits should be banned from the `cats` package object, and the `Id` instances should be moved to something like `cats.implicits.id`.

kailuowang:
Thanks for reporting this. Do you mind provide a bit more specifics on how this hurts people with scalaz in class path? I mean all data types defined in cats have instances defined in companion objects and thus in scope without imports. I understand Id is special but curious how it affects differently in this scenario. Thanks!

johnynek:
Actually removing implicit annotation is Mima compatible. So, we could keep the vals/defs but remove implicit.

djspiewak:
@johnynek Oh good call! Not source compatible, but binary.

@kailouwang It affects people in the scalaz namespace (as scalaz affects people in the cats namespace) as described in the linked shims ticket. It appears that the compiler considers both Id types to be equivalent by some notion of compatibility when viewed *through* the shims implicit mechanisms. This may be a bug in scalac, I’m not sure. But the net result is that scalaz and cats having this problem results in ambiguities for the *other* when shims is in scope, since shims makes instances for each equivalent to the other.

kailuowang:
Thanks for detail @djspiewak

armanbilge:
> As far as I can tell, this scoping issue is pretty innocuous (if a bit leaky) if you're _only_ using cats.

I'd say that's a safe assumption nowadays.

PR #3264, opened 2020-01-21
Make Id covariant

travisbrown:
Making `Id` covariant was [discussed briefly in 2015](https://github.com/typelevel/cats/pull/186#issuecomment-74051456):

> Should Id be covariant? I ask mainly because you say,
>>    We can freely treat values of type A as values of type Id[A], and vice-versa.
>
> which is only strictly true if it is.

As far as I know this was never followed up on, but I just ran into [an issue while trying to cross-build the Cats tests on Dotty](https://github.com/lampepfl/dotty/issues/8049#issue-552831428) where implicit conversions for `Id` weren't being found because it's invariant. Making it covariant would fix this problem (and also just generally seems like the right thing to do, even if the Dotty team decides that the issue is a bug and fixes it on the Dotty side).

Unfortunately if we make `Id` covariant we hit a weird Scala 2 bug when defining instances for it, which minimized looks like this:

```scala
scala> trait Tc[F[_]]
defined trait Tc

scala> object X {
     |   type Id[+A] = A
     |   val x: Tc[Id] = new Tc[Id] {}
     | }
<console>:15: error: covariant type Id occurs in invariant position in type => Tc[X.Id] of value x
         val x: Tc[Id] = new Tc[Id] {}
             ^
```
And it gets even weirder:
```scala
scala> val x: Tc[Id] = new Tc[Id] {}
x: Tc[Id] = $anon$1@673c4f6e

scala> object X { type Id[+A] = A; new Tc[Id] {} }
defined object X
```

There's some explanation by @smarter in a [conversation from this afternoon on Gitter](https://gitter.im/typelevel/cats-dev?at=5e271431075a19397ce055b1).

Update: @smarter also has [a fix](https://github.com/scala/scala/pull/8651) for the issue.

We can work around this by changing the `Id` instances to `λ[A => A]`. ~~We could also use `({ type L[+A] => A })#L` or the weird ``λ[`+A` => A]`` thing but `λ[A => A]` seems to work fine.~~ Update: Scala 2 doesn't care if the variance doesn't match, but Dotty does, so I've gone with `({ type L[+A] => A })#L` for now.

No other changes are necessary, and we can verify that these instances are available without imports:

```scala
scala> cats.Monad[cats.Id]
res0: cats.Monad[cats.Id] = cats.package$$anon$1@9a00a50

scala> cats.Parallel[cats.Id, cats.Id]
res1: cats.Parallel.Aux[cats.Id,cats.Id] = cats.Parallel$$anon$2@12376480

scala> cats.Parallel[cats.Id]
res2: cats.Parallel.Aux[cats.Id,cats.package.catsParallelForId.F] = cats.Parallel$$anon$2@12376480
```

In the longer run, we're going to come up against another change in Dotty (this one definitely [not a bug](https://github.com/lampepfl/dotty/issues/8048)) that means that these type class instances for `Id` won't be available in implicit scope—they'll have to be imported explicitly.

There are a few ways we could address this. The first would be to do nothing and expect users to import `cats._` if they need these instances. We could also create a new `cats.instances.id` that they could be imported from more idiomatically (also making them available in `cats.instances.all` and `cats.implicits`).

In my view the best approach would be to provide these instances in the type class companion objects, which would ensure that they're in implicit scope on both Scala 2 and Dotty. This is something I've already done in #3043, but we could do it separately if necessary.

### DIFF
--- a/core/src/main/scala/cats/package.scala
+++ b/core/src/main/scala/cats/package.scala
@@ -33,16 +33,16 @@ package object cats {
    */
   type Id[+A] = A
   type Endo[A] = A => A
-  implicit val catsInstancesForId: Bimonad[λ[A => A]]
-    with CommutativeMonad[λ[A => A]]
-    with Comonad[λ[A => A]]
-    with NonEmptyTraverse[λ[A => A]]
-    with Distributive[λ[A => A]] =
-    new Bimonad[λ[A => A]]
-      with CommutativeMonad[λ[A => A]]
-      with Comonad[λ[A => A]]
-      with NonEmptyTraverse[λ[A => A]]
-      with Distributive[λ[A => A]] {
+  implicit val catsInstancesForId: Bimonad[({ type L[+x] = x })#L]
+    with CommutativeMonad[({ type L[+x] = x })#L]
+    with Comonad[({ type L[+x] = x })#L]
+    with NonEmptyTraverse[({ type L[+x] = x })#L]
+    with Distributive[({ type L[+x] = x })#L] =
+    new Bimonad[({ type L[+x] = x })#L]
+      with CommutativeMonad[({ type L[+x] = x })#L]
+      with Comonad[({ type L[+x] = x })#L]
+      with NonEmptyTraverse[({ type L[+x] = x })#L]
+      with Distributive[({ type L[+x] = x })#L] {
       def pure[A](a: A): A = a
       def extract[A](a: A): A = a
       def flatMap[A, B](a: A)(f: A => B): B = f(a)
@@ -88,8 +88,8 @@ package object cats {
   /**
    * Witness for: Id[A] <-> Unit => A
    */
-  implicit val catsRepresentableForId: Representable.Aux[λ[A => A], Unit] =
-    new Representable[λ[A => A]] {
+  implicit val catsRepresentableForId: Representable.Aux[({ type L[+x] = x })#L, Unit] =
+    new Representable[({ type L[+x] = x })#L] {
       override type Representation = Unit
       override val F: Functor[Id] = Functor[Id]
 
@@ -98,7 +98,8 @@ package object cats {
       override def index[A](f: Id[A]): Unit => A = (_: Unit) => f
     }
 
-  implicit val catsParallelForId: Parallel.Aux[λ[A => A], λ[A => A]] = Parallel.identity[λ[A => A]]
+  implicit val catsParallelForId: Parallel.Aux[({ type L[+x] = x })#L, ({ type L[+x] = x })#L] =
+    Parallel.identity[({ type L[+x] = x })#L]
 
   type Eq[A] = cats.kernel.Eq[A]
   type PartialOrder[A] = cats.kernel.PartialOrder[A]
--- a/core/src/main/scala/cats/package.scala
+++ b/core/src/main/scala/cats/package.scala
@@ -34,6 +34,7 @@ package object cats {
   type Id[+A] = A
 
   // Workaround for a compiler bug that should be fixed soon.
+  // See https://github.com/scala/scala/pull/8651 for details.
   private type IdWrapper = { type L[+A] = A }
 
   type Endo[A] = A => A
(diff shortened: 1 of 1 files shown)


smarter:
> Unfortunately if we make Id covariant we hit a weird Scala 2 bug when defining instances for it

PR: https://github.com/scala/scala/pull/8651

dwijnand:
Does it keep working if you type alias `{ type L[+x] = x }`?  Just to kill the repetition.

```scala
Welcome to Scala 2.13.1 (OpenJDK 64-Bit Server VM, Java 11.0.1).
Type in expressions for evaluation. Or try :help.

scala> trait Monad[M[_]]
defined trait Monad

scala> type Foo = { type L[+x] = x }
defined type alias Foo

scala> new Monad[Foo#L] {}
res2: Monad[[+x]x] = $anon$1@589d831e
```

smarter:
Yes, but I didn't want to add some other form of type aliasing to the testcase.

travisbrown:
@dwijnand Yes, thanks, I guess that's better even if this is just a temporary workaround. I've pushed an update with a private type alias for `{ type L[+A] = A }`.

core/src/main/scala/cats/package.scala
```diff
@@ -31,11 +31,22 @@ package object cats {
    * type `A` to get a pure value of type `B`. That is, the instance
    * encodes pure unary function application.
    */
-  type Id[A] = A
+  type Id[+A] = A
+
+  // Workaround for a compiler bug that should be fixed soon.
```
kailuowang: now that the PR on scala is in, can we refer to it here? 
https://github.com/scala/scala/pull/8651

travisbrown: @kailuowang Yes, good idea (I linked it in the PR description but the comment would be more helpful).

travisbrown:
@LukaJCB To follow up on your question on Gitter yesterday (about whether users will also have to use the type lambda workaround), there's a [type class instance for `Id`](https://github.com/typelevel/cats/blob/832abe6ae8e5b0b3d6483a50c843250408dd8f34/tests/src/test/scala/cats/tests/IdSuite.scala#L7) in `IdSuite` that shows that you don't need the workaround there.

smarter:
There's an open PR (https://github.com/lampepfl/dotty/pull/8082) against Dotty which will allow it to infer the variance of the parameters of a type alias / type lambda based on its structure (see https://github.com/lampepfl/dotty/pull/8082/files#diff-9d6330a6809ef256c9b6536358305ab3R3591-R3597 for examples). If that gets in, this PR shouldn't be necessary anymore.

travisbrown:
@smarter Is your sense that we can backport this story to Scala 2 in this case, since it's the way things happen to work anyway?

smarter:
Well if the current way of doing things works well enough with scalac in practice, it's probably not worth changing

neko-kai:
@travisbrown 

> In my view the best approach would be to provide these instances in the type class companion objects, which would ensure that they're in implicit scope on both Scala 2 and Dotty. This is something I've already done in #3043, but we could do it separately if necessary.

Another way would be to alias `Id` to an `Id` in an object. Package prefixes aren't part of implicit scope, but objects prefixes are:

```scala
package object cats {
  type Id[+A] = data.id.Id[A]
}
package data {
  object id {
    type Id[+A] = A
    implicit lazy val instances = ...
  }
}
```

Although I guess there's no harm in putting the instances into type class companions, if the other stdlib types are there already and work fine

djspiewak:
> Package prefixes aren't part of implicit scope, but object prefixes are

I'm assuming you mean on Dotty? Package prefixes are most definitely part of the scope *now*. #2276 demonstrates how insidious this can be, and also seems quite related to this effort.

travisbrown:
Putting this on hold since it's no longer necessary on either Dotty or Scala 2. If someone wants to argue that it's more correct, it might be better to take that up once [the Scala 2 fix](https://github.com/scala/scala/pull/8651) is available in a release.

PR #3265, opened 2020-01-22
Copyright and license

travisbrown:
This change fixes a few small things that have been bothering me.

The first is that a few files have Apache license headers, even though the project itself is MIT licensed. These also say "2018 Luka Jacobowitz", so I'm assuming this was an accidental addition from an IDE template or global sbt plugin. Can you confirm this, @LukaJCB?

The other two changes are trivial—I've just updated the copyright dates in `build.sbt` and used the `Test` configuration there instead of `"test"`.


travisbrown:
Great, thanks all!

PR #3273, opened 2020-01-28
Adding instances for ArraySeq

### LINKED ISSUE #2994: Functor / Monad instance for ArraySeq
There doesn't seem to be a Functor / Monad instance for ArraySeq, the new immutable covariant wrapper for Arrays in 2.13.

BalmungSan:
This is my attempt at adding common instances for **ArraySeq**.

I am not sure if we could implement more typeclasses and if the tests are ok. Also, many implementations of the methods were borrowed from the **Vector** and **LazySList** ones.
So, any feedback is welcome.

Closes #2994 

------

_BTW, I have a surgery this Wednesday. So, if reviewers leave feedback and I do not answer them, and some have the motivation and time to finish this PR, I would really appreciate it. Since, probably I won't touch a computer again until the next week._

### DIFF
--- a/core/src/main/scala-2.13+/cats/instances/arraySeq.scala
+++ b/core/src/main/scala-2.13+/cats/instances/arraySeq.scala
@@ -5,7 +5,7 @@ import scala.annotation.tailrec
 import scala.collection.immutable.ArraySeq
 
 trait ArraySeqInstances extends cats.kernel.instances.ArraySeqInstances {
-  implicit def catsStdInstancesForArraySeq: Traverse[ArraySeq] with Monad[ArraySeq] with MonoidK[ArraySeq] =
+  implicit def catsStdInstancesForArraySeq: Monad[ArraySeq] with MonoidK[ArraySeq] with Traverse[ArraySeq] =
     ArraySeqInstances.stdInstances
 
   implicit def catsStdTraverseFilterForArraySeq: TraverseFilter[ArraySeq] =
@@ -19,7 +19,7 @@ trait ArraySeqInstances extends cats.kernel.instances.ArraySeqInstances {
 
 object ArraySeqInstances {
   final private val stdInstances =
-    new Traverse[ArraySeq] with Monad[ArraySeq] with MonoidK[ArraySeq] {
+    new Monad[ArraySeq] with MonoidK[ArraySeq] with Traverse[ArraySeq] {
       def empty[A]: ArraySeq[A] =
         ArraySeq.untagged.empty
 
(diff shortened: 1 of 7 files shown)


core/src/main/scala-2.13+/cats/instances/arraySeq.scala
```diff
@@ -0,0 +1,146 @@
...
+      override def find[A](fa: ArraySeq[A])(f: A => Boolean): Option[A] =
+        fa.find(f)
+
+      override def collectFirst[A, B](fa: ArraySeq[A])(pf: PartialFunction[A, B]): Option[B] =
+        fa.collectFirst(pf)
+
+      override def collectFirstSome[A, B](fa: ArraySeq[A])(f: A => Option[B]): Option[B] =
+        fa.collectFirst(Function.unlift(f))
+
+      def functor: Functor[ArraySeq] = this
```
LukaJCB: What is this needed for?

BalmungSan: Oh, as I said, I start by copying some exiting instances and adapting what I understood.
I just checked and this is not needed anymore, it probably was required for another typeclass that I am not longer implementing, I will remove it?

kernel/src/main/scala-2.13+/cats/kernel/instances/ArraySeqInstances.scala
```diff
@@ -0,0 +1,93 @@
...
+  implicit def catsKernelStdHashForArraySeq[A: Hash]: Hash[ArraySeq[A]] =
+    new ArraySeqHash[A]
+}
+
+private[cats] trait ArraySeqInstances2 {
+  implicit def catsKernelStdEqForArraySeq[A: Eq]: Eq[ArraySeq[A]] =
+    new ArraySeqEq[A]
+}
+
+final private[cats] class ArraySeqOrder[A](implicit ev: Order[A]) extends Order[ArraySeq[A]] {
```
travisbrown: Can we make these proper private?

BalmungSan: @travisbrown  The **Monoid** one has to be like this, since it is used in the core module.

But, I guess the other ones can be made more private, maybe enclose them in the companion object?

-----

_Anyways, as far as I check other files, those classes were public_.

travisbrown: Ah, makes sense. And yeah, I think it was a mistake to make those public to start with, and it's caused actual problems IIRC. I think `private` at the top-level is best.

BalmungSan: @travisbrown the last commit attempts to make those kernel instances as private as possible.
Let me know what do you think of that change.

BalmungSan:
Uhm, I do not understand why this PR decreases the coverture.
Did I missed something in the tests?

LukaJCB:
@BalmungSan you can check right here https://codecov.io/gh/typelevel/cats/pull/3273/diff which methods currently aren't covered :)

Seems like for most of those you can just add more/different law tests

PR #3274, opened 2020-01-28
Add ContT.liftF and ContT.liftK

### LINKED ISSUE #3271: Add ContT.liftF
Implementation:

```scala
  def liftF[M[_]: FlatMap, A, B](mb: M[B]): ContT[M, A, B] =
    apply(FlatMap[M].flatMap(mb)(_))
```

lavrov:
Closes #3271

### DIFF
--- a/core/src/main/scala/cats/data/ContT.scala
+++ b/core/src/main/scala/cats/data/ContT.scala
@@ -88,6 +88,15 @@ object ContT {
 
   /**
    * Same as [[liftF]], but expressed as a FunctionK for use with mapK
+   * {{{
+   * scala> import cats._, data._
+   * scala> trait Foo[F[_]] { def bar: F[Int] }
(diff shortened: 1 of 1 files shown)


core/src/main/scala/cats/data/ContT.scala
```diff
@@ -71,6 +71,27 @@ object ContT {
...
+   * scala> c.run(_ => EitherT.leftT("a")).value.value
+   * res1: Either[String, Int] = Left(a)
+   * }}}
+   */
+  def liftF[M[_], A, B](mb: M[B])(implicit M: FlatMap[M]): ContT[M, A, B] =
+    apply(M.flatMap(mb)(_))
+
+  /**
+   * Same as [[liftF]], but expressed as a FunctionK for use with mapK
+   */
```
kailuowang: How about we add a doc test here as well?

PR #3279, opened 2020-01-31
Add Semigroup and Monoid combinators reverse and intercalate

johnynek:
This adds a few things:

1. reverse method on Semigroup which returns a new Semigroup with the order reversed
2. intercalate method on Semigroup which returns a new Semigroup that puts an element between each combine operation
3. some optimizations of default implementations using laws.
4. reduce the width of the duration Arbitraries so we don't overflow (which causes exceptions).

### DIFF
--- a/kernel/src/main/scala/cats/kernel/CommutativeMonoid.scala
+++ b/kernel/src/main/scala/cats/kernel/CommutativeMonoid.scala
@@ -7,7 +7,9 @@ import scala.{specialized => sp}
  *
  * A monoid is commutative if for all x and y, x |+| y === y |+| x.
  */
-trait CommutativeMonoid[@sp(Int, Long, Float, Double) A] extends Any with Monoid[A] with CommutativeSemigroup[A]
+trait CommutativeMonoid[@sp(Int, Long, Float, Double) A] extends Any with Monoid[A] with CommutativeSemigroup[A] { self =>
+  override def reverse: CommutativeMonoid[A] = self
+}
 
 object CommutativeMonoid extends MonoidFunctions[CommutativeMonoid] {
 
--- a/kernel/src/main/scala/cats/kernel/CommutativeMonoid.scala
+++ b/kernel/src/main/scala/cats/kernel/CommutativeMonoid.scala
@@ -7,7 +7,8 @@ import scala.{specialized => sp}
  *
  * A monoid is commutative if for all x and y, x |+| y === y |+| x.
  */
-trait CommutativeMonoid[@sp(Int, Long, Float, Double) A] extends Any with Monoid[A] with CommutativeSemigroup[A] { self =>
+trait CommutativeMonoid[@sp(Int, Long, Float, Double) A] extends Any with Monoid[A] with CommutativeSemigroup[A] {
+  self =>
   override def reverse: CommutativeMonoid[A] = self
 }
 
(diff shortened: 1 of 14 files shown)


non:
Looks like some annoying 2.13 deprecation warnings. But once those are cleaned up I'm 👍 on this!

johnynek:
someone suggested renaming intercalate to intersperse, which I think it actually a better name.

I'll update to that. @LukaJCB @non what do you think?

johnynek:
actually, `Foldable` already uses `intercalate` for this, so I guess we should stick with that.

johnynek:
also @kubukoz correctly pointed out if we test associativity of combine, we don't need to test associativity of reverse.combine or intercalate.combine after we check that they are actually reversing or intercalating because the fact they are associative is true for all associative combines:
```
val rev = semi.reverse
rev.combine(a, rev.combine(b, c)) =
  combine(combine(c, b), a) =
  combine(c, combine(b, a)) =
  rev.combine(rev.combine(a, b), c)

val intr = semi.intercalate(m)
// using <+> for intr and + for semi:
a <+> (b <+> c) =
  a + m + (b + m + c) =
  (a + m + b) + m + c =
  (a <+> b) <+> c
```
so I removed those associativity checks after we check that reverse and intercalate are actually doing reverse and intercalate.

johnynek:
looks like BigDecimal isn't really commutative in 2.13
```
[info] - CommutativeGroup[BigDecimal].commutativeGroup.intercalateRepeat2 *** FAILED ***

[info]    (LawTests.scala:321)

[info]     Falsified after 82 successful property evaluations.

[info]     Location: (LawTests.scala:321)

[info]       arg0 = 3.1935566048388094E+154,

[info]       arg1 = 1.1175491601257495E+136

[info]     )

[info]       Expected: 3.193556604838809402235098320251500E+154

[info]   Received: 3.193556604838809402235098320251499E+154
```

kernel/src/main/scala/cats/kernel/CommutativeMonoid.scala
```diff
@@ -7,7 +7,10 @@ import scala.{specialized => sp}
  *
  * A monoid is commutative if for all x and y, x |+| y === y |+| x.
  */
-trait CommutativeMonoid[@sp(Int, Long, Float, Double) A] extends Any with Monoid[A] with CommutativeSemigroup[A]
+trait CommutativeMonoid[@sp(Int, Long, Float, Double) A] extends Any with Monoid[A] with CommutativeSemigroup[A] {
+  self =>
+  override def reverse: CommutativeMonoid[A] = self
```
djspiewak: Minor nit, but why not just use `this`?

travisbrown:
@johnynek What do you think about reverting 86a5088 now that #3303 is in? The optimization is valid and the `BigDecimal` tests won't fail now.

johnynek:
I never know who has authority to merge things here. It seems people approve but don’t merge. I think it creates a bad experience for contributors. I think technically I have the merge bit so I could have clicked merge but culturally I think active maintainers should be the ones making the calls wrt release schedules etc...

travisbrown:
@johnynek I was planning to merge but wanted to give you a day or two to follow up on the question about the `intercalate` change, if you wanted to. I'll go ahead and merge now and we can do that later.

travisbrown:
@johnynek (For the record though I don't think anyone would have looked sideways if you'd merged it yourself after the second checkmark—I definitely wouldn't have.)

PR #3285, opened 2020-02-04
Remove machinist from colophon microsite page

MarkusA380:
Machinist is an archived project and has been removed from the cats project with [PR 2925](https://github.com/typelevel/cats/pull/2925) on 1 Jul 2019.

To avoid confusion, it should be removed from the colophon page in my opinion as well.


MarkusA380:
@larsrh Anything I need to do to merge it? Or will someone else do that?

PR #3300, opened 2020-02-16
Backport `sum` from SemigroupK to scala_2.11 

gagandeepkalra:
addresses #3143


gagandeepkalra:
I think it's because of the `mima` previous artefact version set; adding it to AllSyntaxBinCompat6 instead breaks bc.

rossabaker:
Oh, right, it probably works because `AllSyntaxBinCompat7` is still unreleased.

travisbrown:
Thanks @gagandeepkalra!

Issue #149, opened 2015-02-09
Scaladoc for the cats/package.scala object should document the package, rather than the package object


retronym:
Currently:

```
/**
 * Symbolic aliases for various types are defined here.
 */
```

Scaladoc uses the package object documentation as docs for the entire package.

arosien:
👍 yes please!

PR #3301, opened 2020-02-16
Cats package: add comprehensive scaladoc.

diesalbla:
Documentation only PR. No code change. 

State what is in the package and what is not. Also, serve as an entry point to the Cats library and its implementation and design.

Resolves https://github.com/typelevel/cats/issues/149


larsrh:
Cool stuff!

PR #3302, opened 2020-02-18
ArraySeq instance follow-up

travisbrown:
This builds on #3273, addressing #2994. It adds a few instances that were missing and makes the cats-kernel prioritization traits package-private.


LukaJCB:
Doesn't have to be this PR, but we could add a `TraverseFilter` instance as well. :)
Happy to merge this as is though.

travisbrown:
@LukaJCB @BalmungSan already got `TraverseFilter` in #3273! :smile:

Issue #3262, opened 2020-01-20
Use imports for instances and syntax in tests


travisbrown:
Over the weekend I hit a [Dotty bug](https://github.com/lampepfl/dotty/issues/8033) while trying to run the Cats tests on Dotty, and while the workaround is pretty trivial, it highlighted the fact that our tests aren't really exercising normal usage, since the instances and syntax come from traits that are mixed into `CatsSuite` (rather than imported from the usual objects).

This feels like overengineering. It lets us avoid one import in each of our test files, but at the expense of testing instances in a way that doesn't represent what our users are actually doing.

rossabaker:
I think I'm responsible for this in the early days, and I agree with you now.

travisbrown:
Done in #3304.

PR #3304, opened 2020-02-19
Replace instance trait inheritance with imports in tests

travisbrown:
This PR does several things to clean up the Cats tests (the first addresses #3262):

1. Uses imports instead of inheritance to bring instances and syntax into scope.
2. Uses imports instead of nested packaging to bring definitions in `cats` / `alleycats` into scope.
3. Sorts imports consistently.
4. Fixes a few minor packaging inconsistencies (like some of the cats-free tests being in `cats.tests` and others in `cats.free`).

The issue is that the tests are currently set up in a way that saves a few imports, but at the expense of not exercising normal library usage (see #3262 for an example of this causing problems).

After this change the tests do a better job of reflecting what users actually do (e.g. they don't nest stuff inside `package cats` and they don't extend a giant stack of 14 `AllInstancesBinCompatX` traits).

## Compile times

Speeding up test compilation was a non-goal for this change, but I was curious and it seems a little faster (consistently 76s vs. 82s for `testsJVM/test:compile` after `testsJVM/test:clean` on 2.12 on my desktop).

## ~The WIP part~

Update: the `catsSyntaxEq` "override" did work just fine, even after switching to imports, but I went ahead and changed to more granular imports for `cats.syntax` anyway, since trying it out turned up two existing bugs (fixed separately in #3305 and #3306). I don't see any point in switching to granular imports for `cats.instances` since these will probably be gone entirely very soon.

~I'm opening this so I don't lose track of it, but I still want to make sure the change doesn't cause problems with respect to the `===` situation (previously we were overriding `catsSyntaxEq` to make it non-explicit, and now might need to change all of the `cats.syntax.all` imports to exclude it instead).~

There's also one test class (`AlgebraInvariantSuite`) where some function instances needed in the tests weren't resolving correctly when I switched to imports, so I'm still using the instance traits there. I'll try to take a closer look at it, but I think it's something specific to `cats.laws.discipline.eq` and I'm not too worried about it.

### DIFF
--- a/alleycats-tests/js/src/test/scala/alleycats/tests/TestSettings.scala
+++ b/alleycats-tests/js/src/test/scala/alleycats/tests/TestSettings.scala
@@ -1,5 +1,4 @@
-package alleycats
-package tests
+package alleycats.tests
 
 import org.scalactic.anyvals.{PosInt, PosZDouble, PosZInt}
 import org.scalatest.matchers.should.Matchers
--- a/alleycats-tests/jvm/src/test/scala/alleycats/tests/TestSettings.scala
+++ b/alleycats-tests/jvm/src/test/scala/alleycats/tests/TestSettings.scala
@@ -1,5 +1,4 @@
-package alleycats
-package tests
+package alleycats.tests
 
 import org.scalactic.anyvals.{PosInt, PosZDouble, PosZInt}
 import org.scalatest.matchers.should.Matchers
--- a/alleycats-tests/shared/src/test/scala/alleycats/tests/AlleycatsSuite.scala
+++ b/alleycats-tests/shared/src/test/scala/alleycats/tests/AlleycatsSuite.scala
@@ -3,7 +3,6 @@ package alleycats.tests
 import alleycats.std.MapInstances
 import cats._
 import cats.instances.all._
-import cats.syntax.EqOps
 import cats.tests.StrictCatsEquality
 import org.scalatest.funsuite.AnyFunSuiteLike
 import org.scalatestplus.scalacheck.ScalaCheckDrivenPropertyChecks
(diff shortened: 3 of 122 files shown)


travisbrown:
@rossabaker I agree that there are trade-offs here. In my view moving away from inheritance for these instances is a major improvement in this respect (tests reflecting usage), and à la carte vs. kitchen sink is a smaller matter. I'd also prefer to use kitchen sink here, if we had decent tests for the à la carte style, but unfortunately we don't.

There's also the question about what the recommended style should be after #3043. At least initially in 2.2, `cats.implicits._` won't change at all, because if we made the standard library instances non-implicit, that would change prioritization in some cases, and I think we want to stay 100% source-compatible during a transition period. But for the vast majority of use cases `cats.implicits._` will be replaceable with `cats.syntax.all._`.

travisbrown:
@rossabaker Switching to kitchen sink wouldn't be all that hard, though. We'd have to reinstate `catsSyntaxEq`, delete all the `cats.syntax.xyz._` imports, and add `cats.implicits._` everywhere—maybe 15 minutes to write a little script and make sure things work.

travisbrown:
@rossabaker Oh, I almost already did that—I've just rebased to make it a little easier to isolate the kitchen-sink-to-à-la-carte change. If you check out 8dd80e527eb00bbb5a24cd82d99153ae1d7c6fb3 ("Replace instance trait inheritance with imports in tests") you'll get more or less the kitchen sink version (although not `cats.implicits._` but `cats.instances.all._` plus `cats.syntax.all._`).

When it's green I'll merge the three rebased commits without squashing in case we want to revisit this in the future (the changes are identical to what was reviewed, just reordered slightly).

travisbrown:
Ugh, sorry, will need one last :+1: after the rebase. @LukaJCB, mind giving it one more? :smile:

PR #3309, opened 2020-02-21
Introduce reduceOption for UnorderedFoldable and minimum/maximum(By)Option 

jorgeadriano:
This PR first introduces the following methods for `UnorderedFoldable`,

*  `unorderedReduceOption`. By lifting a `CommutativeSemigroup[A]` into a `CommutativeMonoid[Option[A]]` in the standard way. 

* `minimumOption`, `maximumOption`, `minimumByOption`, `maximumByOption`. Via `unorderedReduceOption` by defining the appropriate semigroups for min and max. 

It also includes a law to test the consistency between `unorderedReduceOption` and `unorderedFold`.

**Observations:**

The second part of the PR regarding min/max operations is the object of issue https://github.com/typelevel/cats/issues/2715. I've since realised there is another (wip) PR to address it https://github.com/typelevel/cats/pull/3132. I've opted to leave my tentative solution in this PR anyway since it follows a different approach, via `unorderedReduceOption`. This is more in line with the corresponding implementations in `Foldable` via `reduceLeftOption`.

My tentative implementation for the min/max however appears to suffer from the same issues described in https://github.com/typelevel/cats/issues/2715. Namely that the min/max operations in `Foldable` require override, which if I understand correctly affects simulacrum generation of syntax ops. Indeed I could tell syntax wasn't working when I tried it, but don't know much about simulacrum. I could try to fix this but would need some guidance here.

### DIFF
--- a/core/src/main/scala/cats/UnorderedFoldable.scala
+++ b/core/src/main/scala/cats/UnorderedFoldable.scala
@@ -73,6 +73,50 @@ import simulacrum.{noop, typeclass}
   @noop
   def count[A](fa: F[A])(p: A => Boolean): Long =
     unorderedFoldMap(fa)(a => if (p(a)) 1L else 0L)
+
+  /**
+   * Find the minimum `A` item in this structure according to the `Order[A]`.
+   *
+   * @return `None` if the structure is empty, otherwise the minimum element
+   * wrapped in a `Some`.
+   *
+   * @see [[maximumOption]] for maximum instead of minimum.
+   */
+  def minimumOption[A: Order](fa: F[A]): Option[A] =
+    unorderedReduceOption(fa)(UnorderedFoldable.minCommutativeSemigroup)
+
+  /**
+   * Find the maximum `A` item in this structure according to the `Order[A]`.
+   *
+   * @return `None` if the structure is empty, otherwise the maximum element
+   * wrapped in a `Some`.
+   *
+   * @see [[minimumOption]] for minimum instead of maximum.
+   */
+  def maximumOption[A: Order](fa: F[A]): Option[A] =
+    unorderedReduceOption(fa)(UnorderedFoldable.maxCommutativeSemigroup)
+
+  /**
+   * Find the minimum `A` item in this structure according to an `Order.by(f)`.
+   *
+   * @return `None` if the structure is empty, otherwise the minimum element
+   * wrapped in a `Some`.
+   *
+   * @see [[maximumByOption]] for maximum instead of minimum.
+   */
+  def minimumByOption[A, B: Order](fa: F[A])(f: A => B): Option[A] =
+    minimumOption(fa)(Order.by(f))
+
+  /**
+   * Find the maximum `A` item in this structure according to an `Order.by(f)`.
+   *
+   * @return `None` if the structure is empty, otherwise the maximum element
+   * wrapped in a `Some`.
+   *
+   * @see [[minimumByOption]] for minimum instead of maximum.
+   */
+  def maximumByOption[A, B: Order](fa: F[A])(f: A => B): Option[A] =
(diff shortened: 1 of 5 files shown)


core/src/main/scala/cats/UnorderedFoldable.scala
```diff
@@ -14,6 +15,9 @@ import simulacrum.{noop, typeclass}
   def unorderedFold[A: CommutativeMonoid](fa: F[A]): A =
     unorderedFoldMap(fa)(identity)
 
+  def unorderedReduceOption[A](fa: F[A])(implicit A: CommutativeSemigroup[A]): Option[A] =
+    unorderedFoldMap(fa)(_.some)(cats.instances.option.catsKernelStdCommutativeMonoidForOption(A))
```
travisbrown: Do you mind changing this to `Some(_)` (adding a type parameter to `unorderedFoldMap` if needed)? We're trying to avoid using `syntax` in core, to keep things simple and minimize cyclic dependencies between packages.

jorgeadriano: Type inference fails with `Some(_)` which is why I resorted to `_.some`. Should I use `Some(_): Option[A]`?

jorgeadriano:
@travisbrown some further comments on possible improvements: 

**Consistency laws: unordered and ordered fold/reduce:**

We could test consistency between unordered folds/reduce and their "ordered" counterparts. For  `unorderedReduceOption` the law would be, 
```
def reduceLeftOptionConsistentWithUnorderedReduceOption[A](
  fa: F[A]
)(implicit A: CommutativeSemigroup[A]): IsEq[Option[A]] =
  F.reduceLeftOption(fa)(A.combine) <-> F.unorderedReduceOption(fa)
```
I've implemented this law in a separate branch: https://github.com/jorgeadriano/cats/pull/2. I did not yet adapt the tests, but could do if there is interest in it. 

**Specification laws: minimum/maximumOption:** 

I've found no laws for `Foldable#minimumOption` and friends. I don't know if this is intentional or a todo. In any case I've implemented full spec laws in this separate branch for `UnorderedFoldable`.  For `minimumOption` the laws would be, 

```
def minimumOptionIsMinimal[A](fa: F[A])(implicit A: Order[A]): Boolean = {
  val optMin = F.minimumOption(fa)
  F.forall(fa) { a =>
    optMin.exists { min =>
      A.lteqv(min, a)
    }
  }
}

def minimumOptionIsContained[A: Order](fa: F[A]): Boolean =
  F.minimumOption(fa).forall { min =>
    F.exists(fa) { a =>
      Order.eqv(min, a)
    }
  }
```
I've also adapted the tests to include the respective tests for methods `minimumOption` and `minimum/maximumOption`. I'd have to adapt them further to also include tests for  `minimumByOption` and `minimum/maximumByOption`. 

The extra required parameters and implicits seem to propagate throughout the tests and add up to plenty of changes in their "interface", so I suppose that even if we'd like to add them that may be better suited for some other PR. In any case I could finish for this PR or some other if there is interest in it. 

**contains_ in UnorderedFoldableOps:** 
 
Method `FoldableOps#contains_` could be defined for / moved to `UnorderedFoldableOps`

joroKr21:
AFAIR the reason to not add these laws was the required implicit Order and the binary compatibility of the test-kit.

core/src/main/scala/cats/UnorderedFoldable.scala
jorgeadriano: > I'll take a closer look at the bincompat issues ASAP (probably this weekend). I think you'll probably also hit bincompat breakage with these laws changes, but that's probably less annoying to work around.

Hi @travisbrown. Would it make sense to revisit this for Cats 2.2?

PR #3312, opened 2020-02-21
Make scala.js 1.0 the default

cquiroz:
I propose that scala.js becomes the default going forward but you can still publish for 0.6.x


larsrh:
I'd go as far as dropping sjs 0.6 altogether.

cquiroz:
There is still work pending to go to 1.0 only (e.g. scalajs-react which is blocked by several downstream dependencies), I'd like it to be cross built for a whilo
But I agree that soon it should be 1.0 only

cquiroz:
Closed in favor of #3357

PR #3316, opened 2020-02-24
Add semiflatTap and leftSemiflatTap functions to EitherT

matwojcik:
`semiflatTaps` functions on `EitherT` are highly useful when you have some e.g. `Logger[F]` abstractions and you don't want to lift it to `EitherT[F,A,B]`

### DIFF
--- a/core/src/main/scala/cats/data/EitherT.scala
+++ b/core/src/main/scala/cats/data/EitherT.scala
@@ -447,6 +447,12 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
         }
     })
 
+  def semiflatTap[C](f: B => F[C])(implicit F: Monad[F]): EitherT[F, A, B] =
+    semiflatMap(b => Monad[F].as(f(b), b))
+
+  def leftSemiflatTap[C](f: A => F[C])(implicit F: Monad[F]): EitherT[F, A, B] =
+    leftSemiflatMap(a => Monad[F].as(f(a), a))
+
   def compare(that: EitherT[F, A, B])(implicit o: Order[F[Either[A, B]]]): Int =
     o.compare(value, that.value)
 
--- a/core/src/main/scala/cats/data/EitherT.scala
+++ b/core/src/main/scala/cats/data/EitherT.scala
@@ -448,10 +448,10 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
     })
 
   def semiflatTap[C](f: B => F[C])(implicit F: Monad[F]): EitherT[F, A, B] =
-    semiflatMap(b => Monad[F].as(f(b), b))
+    semiflatMap(b => F.as(f(b), b))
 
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -447,6 +447,12 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
         }
     })
 
+  def semiflatTap[C](f: B => F[C])(implicit F: Monad[F]): EitherT[F, A, B] =
+    semiflatMap(b => Monad[F].as(f(b), b))
```
travisbrown: This is a nitpick, but for the sake of consistency and to keep things simple, what do you think about using `F.as` here instead of `Monad[F].as`?

matwojcik: sure, will do ;)

matwojcik:
@travisbrown  I'm not fully convinced ;) In the end it uses `flatMap` which resembles your suggestion however it does indeed one extra wrapping and unwrapping.

So it adds one wrap/unwrap for a price of more readable code and additionally existing code is reused. Plus it isn't a method like `flatMap`/`map` which would be used all over again in the same stack, so the penalty should be low.

However I'm not that attached to the original solution, so can change it if you insist ;)

travisbrown:
@matwojcik It's not a blocker for me. The signature is the important part, if someone wants to make an argument that we should change the implementations later they can.

travisbrown:
The test failure was `BigDecimal` commutativity on 2.13, which is a known issue, so I'm restarting that job.

I have an only moderately controversial fix for that problem that's waiting for another sign-off: https://github.com/typelevel/cats/pull/3303

core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -447,6 +447,12 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
         }
     })
 
+  def semiflatTap[C](f: B => F[C])(implicit F: Monad[F]): EitherT[F, A, B] =
+    semiflatMap(b => F.as(f(b), b))
+
+  def leftSemiflatTap[C](f: A => F[C])(implicit F: Monad[F]): EitherT[F, A, B] =
```
LukaJCB: What do we think about using `Unit` here to signify that this resulting type is going to be discarded? Might be a bit more cumbersome on the usage side, but it's just an extra `.void`

travisbrown: We already have a `flatTap` on `FlatMap` that doesn't require `Unit`, and `tap`, `tapWith`, etc. on `Kleisli` don't constraint the output type to be `Unit`. My personal preference would be consistency—I think that plus slightly easier use outweighs a little extra expressivity in the types.

matwojcik: @LukaJCB I first thought about it, but then saw that `FlatMap[F].flatTap` takes `A => F[B]` instead of `A => F[Unit]`, so I wanted to be consistent with existing signature.

PR #3318, opened 2020-02-24
Drop noop for Functor unzip with Liskov evidence

yangzai:
The main motivation of this PR is to reduce the verbosity of calling `unzip`.

As mentioned by @kailuowang in https://github.com/typelevel/cats/pull/3062#issuecomment-546747506 the `$conforms` related implicit conversion obscures type signature.

This PR proposes an alternative solution with an implicit Liskov evidence that requires explicit conversions instead.


travisbrown:
We can't change the signature of the `unzip` in `Functor` before 3.0, but we could probably just add an `unzip` to `FunctorSyntax` (possibly with a workaround like this if that's necessary)? We've done that in many other cases where Simulacrum didn't work for some reason.

yangzai:
@travisbrown thanks for the feedback. I'm looking at how it it was implemented here: https://github.com/typelevel/cats/pull/3234 . Hope I'm on the same page.

core/src/main/scala/cats/Functor.scala
```diff
@@ -159,10 +159,16 @@ import simulacrum.typeclass
...
+  def unzip[A, B](fab: F[(A, B)]): (F[A], F[B]) = (map(fab)(_._1), map(fab)(_._2))
+
+  //TODO: Drop `final` when the `noop` version deprecates
+  // `final` restricts functor instance overriding of `unzip` to the noop version
+  final def unzip[A, X, Y](fa: F[A])(implicit ev: A <~< (X, Y)): (F[X], F[Y]) = {
     val fxy = map(fa)(ev)
 
-    (map(fxy)(_._1), map(fxy)(_._2))
+    unzip(fxy)
   }
```
yangzai: @travisbrown Overloading seems to work but I'm not sure if there are any negative implication.

travisbrown: @yangzai I think it's a good idea to avoid overloading when it's not absolutely necessary—see e.g. [this SO answer](https://stackoverflow.com/a/2512001/334519) for some reasons. Not everything there is still applicable in the latest Scala versions, but some is. For example, this is a nicer error message:

```scala
scala> import cats.Functor, cats.instances.all._
import cats.Functor
import cats.instances.all._

scala> Functor[Option].unzip(1)
                             ^
       error: type mismatch;
        found   : Int(1)
        required: Option[(?, ?)]
```
…than this:
```scala
scala> Functor[Option].unzip(1)
<console>:16: error: overloaded method value unzip with alternatives:
  [A, X, Y](fa: Option[A])(implicit ev: A <~< (X, Y))(Option[X], Option[Y]) <and>
  [A, B](fab: Option[(A, B)])(Option[A], Option[B])
 cannot be applied to (Int)
       Functor[Option].unzip(1)
                       ^
```
If we _had_ to do this to get the syntax method, it'd be different, but I don't think overloading is worth it just to get Simulacrum to generate the syntax method. Implementing it explicitly in `FunctorOps` is also more consistent with the approach we use in other cases where Simulacrum hasn't done what we needed.

yangzai: @travisbrown Yeah it was a lazy hack. Sorry was kind of busy the past week. Will reference #3234 for the `FunctorOps` approach if the `<~<` based type signature is fine going forward.

core/src/main/scala/cats/syntax/functor.scala
```diff
@@ -1,4 +1,34 @@
...
+   *
+   * scala> (5: Id[Int]).map(i => (i, i)).unzip == ((5, 5))
+   * res0: Boolean = true
+   * }}}
+   *
+   */
+  def unzip[X, Y](implicit F: Functor[F], ev: A <~< (X, Y)): (F[X], F[Y]) = {
+    val fxy = F.map(fa)(ev)
+
+    F.unzip(fxy)
```
travisbrown: Is there a reason not to write this as `F.unzip(F.map(fa)(ev))`? It's straightforward enough that I'm not sure the intermediate `val` helps much.

core/src/main/scala/cats/syntax/functor.scala
```diff
@@ -1,4 +1,34 @@
 package cats
 package syntax
 
+import cats.evidence.<~<
+
 trait FunctorSyntax extends Functor.ToFunctorOps
+
+private[syntax] trait FunctorSyntaxBinCompat0 {
```
travisbrown: Is there a reason you can't just add this to `FunctorSyntax` and skip all the `BinCompat` stuff? It shouldn't be necessary now that we don't support Scala 2.11 on master.

core/src/main/scala/cats/syntax/functor.scala
yangzai: @travisbrown Yeah I was thinking of doing that but I thought fxy would be clearer somehow. Will make the change to a one liner tomorrow.

core/src/main/scala/cats/syntax/functor.scala
yangzai: @travisbrown I wasn't too sure so I left that as-is. Will move `catsSyntaxFunctorOps0` into `FunctorSyntax`.

travisbrown:
Is there any reason not to just write the following?

```scala
final class FunctorTupleOps[F[_], A, B](private val fab: F[(A, B)]) extends AnyVal {
  def unzip(implicit F: Functor[F]): (F[A], F[B]) = F.unzip(fab)
}
```
I just tried it with your tests and everything seems fine.

yangzai:
@travisbrown Originally the approach (before I saw what was done in the backport) was to fix it directly in the `Functor` typeclass/trait with something that would get lifted by Simulacrum. I guess with the current implementation would be beneficial if the intention is to move as much back to the typeclass when bincompat isn't an issue. However having something like `FunctorTupleOps` per complex shape like `F[(A, B)]` seems to have already been somewhat of a convention as well. Let me know your thoughts.

On a sidenote regarding the compilation error in, changing `widen` to something like
```

File: core/src/main/scala/cats/syntax/functor.scala [hunk 4]
```diff
@@ -1,4 +1,28 @@
...
+   * {{{
+   * scala> import cats.Id
+   * scala> import cats.syntax.functor._
+   *
+   * scala> (5: Id[Int]).map(i => (i, i)).unzip == ((5, 5))
+   * res0: Boolean = true
+   * }}}
+   *
+   */
+  def unzip[X, Y](implicit F: Functor[F], ev: A <~< (X, Y)): (F[X], F[Y]) = F.unzip(F.map(fa)(ev))
```
johnynek, line 27: I don't think we shoud use `map` with `ev`. Instead, we should use substitution to avoid any chance we do any action. If instead we took `ev: A Is (X, Y)`, then we can use `F.unzip(ev.substitute[F](fa))`

yangzai: @johnynek sorry, I was MIA for awhile. Let me have a look, will get back to you sometime this week.

yangzai: @johnynek I've done some testing with`Is`/`Leibinz`. I had to flip the type parameters `A` and `(X, Y)` around to make it work, i.e:
```scala
def unzip[X, Y](implicit F: Functor[F], ev: (X, Y) Is A): (F[X], F[Y]) = F.unzip(ev.flip.substitute(fa))
```

Also, because tuples are covariant, there is still a possibility of a `Tuple2` to be a subtype of another `Tuple2`, eg:
```scala
    sealed trait Animal
    final case class Cat() extends Animal
    final case class Dog() extends Animal

    val l: NonEmptyList[(Cat, Dog)] = NonEmptyList one Tuple2(Cat(), Dog())
    Functor[NonEmptyList].unzip[Animal, Animal](l) //compiles for both `Is` and `As`
    l.unzip[Animal, Animal] //does not compile for `Is`
```
This would result in slight inconsistency between the prefix and postfix versions of `unzip`.


As for `As` I've tried to replace `map` with `widen` like this:
```scala
def unzip[X, Y](implicit F: Functor[F], ev: A <~< (X, Y)): (F[X], F[Y]) = F.unzip(F.widen[A, (X, Y)](fa))
```
but it would not fit the type bound. Otherwise, calling `asInstanceOf` directly would work:
```scala
def unzip[X, Y](implicit F: Functor[F], ev: A <~< (X, Y)): (F[X], F[Y]) = F.unzip(fa.asInstanceOf[F[(X, Y)]])
```
or we could implement another `widen` that requires an evidence of `As` instead of a type bound constraint:
```scala
  def widenAs[A, B: A <~< *](fa: F[A]): F[B] = fa.asInstanceOf[F[B]]
  def widen[A, B >: A](fa: F[A]): F[B] = widenAs(fa)
```

johnynek: what about this:

```scala
final class FunctorTuple2Ops[F[_], A, B](fa: F[(A, B)]) extends AnyVal {
  def _1F(implicit F: Functor[F]): F[A] = F.map(fa)(_._1)
  def _2F(implicit F: Functor[F]): F[B] = F.map(fa)(_._2)
  def unzip(implicit F: Functor[F]): (F[A], F[B]) = F.unzip(fa)
}
```
or something... sidestep using the evidence at all and only put the enrichment on Tuple2 types?

johnynek: actually I see @travisbrown suggested this originally before I derailed things: https://github.com/typelevel/cats/pull/3318#issuecomment-597524402

yangzai: @johnynek @travisbrown I've made the change to `FunctorTuple2Ops`. I think some of my comments on that might have gone to the wrong PR when I accidentally closed and reopened this PR. The original intention was to fix `unzip` directly in the `Functor` typeclass but either should be fine now since a method extension would be required either way. We might still need type evidence for efficient`unzip` for specific types.

johnynek:
I think this looks good other than a comment about the Functor to use in the doc tests.

yangzai:
> I think this looks good other than a comment about the Functor to use in the doc tests.

@johnynek do you mean that the doc tests with "Lifts `XXX` to Functor" should to be more concise about the shape of the functor?

johnynek:
I mean doing doc tests using `Id[(Int, Int)]` might be less enlightening to new comers since Id seems useless to newcomers. Option would be just about as easy to write the doc test examples and wouldn't be as trivial.

johnynek (approved):

core/src/main/scala/cats/syntax/functor.scala
```diff
@@ -1,4 +1,64 @@
...
+  def _1F(implicit F: Functor[F]): F[A] = F.map(fab)(_._1)
+
+  /**
+   * Lifts `Tuple2#_2` to Functor
+   *
+   * {{{
+   * scala> import cats.Id
+   * scala> import cats.syntax.functor._
+   *
+   * scala> ((1, 2): Id[(Int, Int)])._2F == 2
```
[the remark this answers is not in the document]
yangzai: @johnynek the doc test for `unzip` was ported from the 2.11 branch so I decided to keep to the same type. Tried `Option` but it failed for 2.13 because `Option#unzip` is already defined in 2.13. Switched to `Chain` as it is Cats specific.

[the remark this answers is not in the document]
yangzai: @johnynek is `Chain` fine or should replace it with something else?

barambani (dismissed):

johnynek (approved):

johnynek:
@travisbrown can you double check before we merge? Any time we change API surface area I like to make sure there is good consensus.

Side note: it would be interesting if mima could tell you if a patch does or doesn't change surface area and we could label each PR. PRs that only change implementations for which there is coverage can be merged with less review IMO.

travisbrown:
@johnynek Sure, I'll take a look tomorrow.

tests/src/test/scala/cats/tests/FunctorSuite.scala
```diff
@@ -34,10 +36,33 @@ class FunctorSuite extends CatsSuite {
...
+    //empty test for completeness
+    val emptyL = List.empty[Int]
+    val emptyM = Map.empty[String, Int]
+
+    assert(Functor[List].unzip(List.empty[(Int, Int)]) === ((emptyL, emptyL)))
+    assert(Functor[Map[String, *]].unzip(Map.empty[String, (Int, Int)]) === ((emptyM, emptyM)))
+  }
+
+  test("_1F, _2F and swapF form correct lists for concrete list of tuples") {
+    forAll { l: List[(Int, Int)] =>
```
larsrh, line 61:
```suggestion
    forAll { (l: List[(Int, Int)]) =>
```

larsrh (approved):

Issue #3313, opened 2020-02-22
Possibly corrupted sjs 1 artifact


larsrh:
See typelevel/paiges#243.

My suspicion is that a new Scala.js version requires a clean build, which may not have happened.

Mitigation is probably a 2.1.1 release, because Sonatype won't let you change released artifacts.

larsrh:
Same issue here typelevel/cats-testkit-scalatest#48.

larsrh:
And another one typelevel/cats-effect#796

jenshalm:
I would omit the word "Possibly" in the description. :-)

I just ran into this myself and out of curiosity I browsed the sjsir files in the cats-core_sjs1 jar.
Somehow `/cats/syntax/package$hash$.sjsir` contains the version string 0.6.29 whereas most other files contain 1.0.

A 2.1.1 release would be great. I'm currently blocked on releasing Laika with Scala.js support, and I'm probably not the only one.

travisbrown:
Oh, ouch, I ran this release and this is probably my fault. The tests were definitely passing. Has anyone tried publishing Cats locally for Scala.js 1.0.0 and using in these projects to make sure it's just an issue with the release?

travisbrown:
I'd really like to get a Cats 2.2.0-M1 published this week and my inclination would be to wait for that, since then on the off-chance that there are ongoing issues we can iterate more quickly without forcing the 97% of Cats users who don't care about Scala.js to bump patch versions.

travisbrown:
Also a lot of the ecosystem is still blocked on Shapeless, so while it would be nice to have working artifacts out there for Scala.js 1.0, I personally don't think it's worth triggering thousands of updates for JVM-only users by publishing 2.1.1 yet.

larsrh:
You make a good point, although I don't think the cost of a patch release is high. Contrastingly, the cost of all the broken scala-steward PRs is quite high.

travisbrown:
Okay, let me make sure I can publish a working Scala.js 1.0.0 release and I'll get 2.1.1 lined up.

jenshalm:
I just tested a local publish with current master and I progressed from the error about invalid sjsir files to linkage errors (which are pretty certain on my side as I am just about to add Scala.js support for the first time, so I expected some issues anyway)

travisbrown:
[...]

PR #3319, opened 2020-02-25
Update changelog

travisbrown:
[2.1.1](https://github.com/typelevel/cats/releases/tag/v2.1.1) is on its way to Maven Central. See [this issue](https://github.com/typelevel/cats/issues/3313#issuecomment-590782403) for some additional discussion.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,11 @@
+## Version 2.1.1
+
+_2020 February 25_
+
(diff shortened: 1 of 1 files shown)


travisbrown:
(If anyone's wondering I didn't think this patch release needed a spot in the "Community Announcements" section of the README. Happy to add it in a follow-up if someone thinks it should be there.)

PR #3321, opened 2020-02-26
Reinstate @johnynek's intercalate optimization from #3279

travisbrown:
This is a minor thing but I'm opening this PR so we don't lose track of it.

@johnynek originally included this optimization in #3279, but removed it because prior to #3303 `combine` for `BigDecimal` wasn't actually commutative on Scala 2.13. That's fixed now, so we can put this back.


travisbrown:
This commit is merged but for some reason the PR didn't close itself. I'm leaving it open for now in the hopes that the issue will resolve itself (and keep our PR history accurate), but please don't merge again.

PR #3325, opened 2020-02-26
Specific commutativity and associativity tests for BigDecimal

travisbrown:
Before the fix in #3303 we were seeing occasional failed tests on Scala 2.13, and we haven't seen any since then, but I should have added some tests that specifically highlighted the fix, to make sure the instance don't accidentally get changed back, etc.

In this PR I've added two specific cases where commutativity and associativity didn't hold for the old `|+|` for `BigDecimal` on Scala 2.13, which simply called the standard library's `+`:

```scala
import java.math.MathContext

val one = BigDecimal("1", MathContext.DECIMAL32)
val small = BigDecimal("0.00001111111", MathContext.DECIMAL32)
val xs = one :: List.fill(10)(small)
```
And then:
```scala
scala> xs.reduceLeft(_ + _)
res0: scala.math.BigDecimal = 1.000110

scala> xs.reduceRight(_ + _)
res1: scala.math.BigDecimal = 1.000111
```
And for commutativity:
```scala
scala> import java.math.MathContext
import java.math.MathContext

scala> val one = BigDecimal("1")
one: scala.math.BigDecimal = 1

scala> val small = BigDecimal("1e-7", MathContext.DECIMAL32)
small: scala.math.BigDecimal = 1E-7

scala> one + small
res0: scala.math.BigDecimal = 1.0000001

scala> small + one
res1: scala.math.BigDecimal = 1.000000
```

### DIFF
--- a/kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
+++ b/kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
@@ -313,11 +313,9 @@ class Tests extends TestsConfig with AnyFunSuiteLike with FunSuiteDiscipline wit
   checkAll("Hash[SortedMap[Int, String]]", HashTests[SortedMap[Int, String]].hash)
   checkAll("Hash[Queue[Int]", HashTests[Queue[Int]].hash)
 
-  {
-    checkAll("Order[BigDecimal]", OrderTests[BigDecimal].order)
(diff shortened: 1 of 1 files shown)


travisbrown:
This commit is merged but for some reason the PR didn't close itself. I'm leaving it open for now in the hopes that the issue will resolve itself (and keep our PR history accurate), but please don't merge again.

PR #3335, opened 2020-02-29
Add foldF, cataF and emptyflatTap to OptionT

ybasket:
Adds several handy methods to `OptionT`:
* `foldF` allows to fold effectually, similar to `EitherT.foldF`
* `cataF` is a variant of `foldF` with one parameter list, similar as `cata` is to `fold`
* `emptyflatTap` (suggestions for better names welcome) allows to perform an effect if the `OptionT` carries a `None` value inside, leaving the result unchanged (except if raising an error, that would propagate). 

An example for a use case for `emptyflatTap` is looking up a value in a cache (returning `F[Option[V]]` which you wrap in `OptionT` for further handling, but logging (as `F[Unit]`) if the value was empty:

```scala
val cachedResult: F[Option[V]] = cache.get(...)
OptionT(cachedResult)
  .emptyflatTap(logger.info("Cache was empty"))
  .doMoreStuffOnOptionT
```


core/src/main/scala/cats/data/OptionT.scala
```diff
@@ -63,6 +86,13 @@ final case class OptionT[F[_], A](value: F[Option[A]]) {
   def subflatMap[B](f: A => Option[B])(implicit F: Functor[F]): OptionT[F, B] =
     transform(_.flatMap(f))
 
+  /**
+   * Perform an effect if the value inside the is a [[None]], leaving the result untouched. Equivalent to [[orElseF]]
+   * with an effect returning [[None]] as argument.
+   */
+  def emptyflatTap[B](f: => F[B])(implicit F: Monad[F]): OptionT[F, A] =
```
ybasket: We could require `F[Unit]` here, but that would be inconsistent with `FlatMap.flatTap` which has the type parameter as well.

The name with lowercase `flat` follows the example of `subflatMap` and `semiflatMap`.

rossabaker: Yes, I'd keep the type parameter for consistency. Having similar things grouped by prefix is nice for discovery through autocompletion or browsing the scaladoc, so maybe `flatTapNone`?

ybasket: I agree with the autocompletion argument, so renamed to `flatTapNone`, thanks for the name :)

ybasket:
@travisbrown thank you (also for the offer to resolve the conflict)! Just fixed the formatting, all checks green now.

core/src/main/scala/cats/data/OptionT.scala
```diff
@@ -63,6 +89,13 @@ final case class OptionT[F[_], A](value: F[Option[A]]) {
   def subflatMap[B](f: A => Option[B])(implicit F: Functor[F]): OptionT[F, B] =
     transform(_.flatMap(f))
 
+  /**
+   * Perform an effect if the value inside the is a `None`, leaving the value untouched. Equivalent to [[orElseF]]
+   * with an effect returning `None` as argument.
+   */
+  def flatTapNone[B](f: => F[B])(implicit F: Monad[F]): OptionT[F, A] =
```
diesalbla, line 96:
```suggestion
  def flatTapNone[B](ifNone: => F[B])(implicit F: Monad[F]): OptionT[F, A] =
```

core/src/main/scala/cats/data/OptionT.scala
```diff
@@ -16,6 +16,21 @@ final case class OptionT[F[_], A](value: F[Option[A]]) {
...
+   * {{{
+   * scala> import cats.implicits._
+   * scala> import cats.data.OptionT
+   *
+   * scala> val optionT: OptionT[List, Int] = OptionT[List, Int](List(Some(23), None))
+   * scala> optionT.foldF(Nil)(v => List(v, v * 2))
+   * res0: List[Int] = List(23, 46)
+   * }}}
+   */
+  def foldF[B](default: => F[B])(f: A => F[B])(implicit F: FlatMap[F]): F[B] =
```
diesalbla, line 31:
```suggestion
  def foldF[B](ifNone: => F[B])(withSome: A => F[B])(implicit F: FlatMap[F]): F[B] =
```

core/src/main/scala/cats/data/OptionT.scala
```diff
@@ -24,6 +39,14 @@ final case class OptionT[F[_], A](value: F[Option[A]]) {
   def cata[B](default: => B, f: A => B)(implicit F: Functor[F]): F[B] =
     fold(default)(f)
 
+  /**
+   * Effectful catamorphism on the Option. This is identical to [[foldF]], but it only has
+   * one parameter list, which can result in better type inference in some
+   * contexts.
+   */
+  def cataF[B](default: => F[B], f: A => F[B])(implicit F: FlatMap[F]): F[B] =
```
diesalbla, line 47: I am aware of the notions of _catamorphisms_, but is that also a unified term or notion across `cats`? I can only grep the words  `def cata` in `OptionT`and in `CoFree`. 

Unless we want to extend the use across the whole of `cats`, perhaps we should drop them, to avoid confusion.

core/src/main/scala/cats/data/OptionT.scala
```diff
@@ -45,6 +68,9 @@ final case class OptionT[F[_], A](value: F[Option[A]]) {
   def semiflatMap[B](f: A => F[B])(implicit F: Monad[F]): OptionT[F, B] =
     flatMap(a => OptionT.liftF(f(a)))
 
+  def semiflatTap[B](f: A => F[B])(implicit F: Monad[F]): OptionT[F, A] =
```
diesalbla, line 71: We have had this conversation before: given that the `B` has no relevance on the types of the outputs, should we just make it an existential?
```suggestion
  def semiflatTap(f: A => F[_])(implicit F: Monad[F]): OptionT[F, A] =
```

core/src/main/scala/cats/data/OptionT.scala
ybasket: While I agree with your intention (and have hence changed it on `flatTapNone` – thx), I would value consistency with `fold` more and keep the parameter names the same.

core/src/main/scala/cats/data/OptionT.scala
ybasket: There's a `cata` to `fold` already, so when seeing `foldF`, I'd expect `cataF` there as well. Whether `cata` should have been introduced with that name in the first place seems to be outside of the scope of this PR…but that's just my two cents.

core/src/main/scala/cats/data/OptionT.scala
ybasket: `FlatMap` has this type parameter for `flatTap`, `EitherT` has it for it's `semiFlatTap`, so again, I would vote for consistency to give cats users a reliable experience in terms of how signatures look and when type inference works. Or do you see any other advantage of the existential which would outweigh this aspect?

ybasket:
@travisbrown As you approved the code once already and I changed only an argument name since then, would you mind merging? 🙏

travisbrown:
Sure, I’ll merge when I’m back at a computer.

PR #3337, opened 2020-02-29
Update Scalafmt to 2.4.2

travisbrown:
This is a more conservative version of #3315 that still rewrites `{ }` to `( )` for single-line expressions, but doesn't remove newlines in cases like this, [which was my primary complaint about the default config in 2.4.2](https://github.com/typelevel/cats/pull/3315#issuecomment-590907872):

```scala
  forAll { (nes: NonEmptySet[Int]) =>
    nes.reduce should ===(nes.fold)
  }
```
In #3315 that got rewritten to the following, which obscures what's being tested, at least to my eyes:

```scala
  forAll((nes: NonEmptySet[Int]) => nes.reduce should ===(nes.fold))
```

Instead of `+495 −1,432` it's `+58, −50`, so quite a bit less dramatic, and the changes look more reasonable to me.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -7,3 +7,4 @@ assumeStandardLibraryStripMargin = true
 danglingParentheses = true
 rewrite.rules = [AvoidInfix, SortImports, RedundantBraces, RedundantParens, SortModifiers]
 docstrings = JavaDoc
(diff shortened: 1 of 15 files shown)


travisbrown:
Thanks all! We can always reconsider going back to defaults later.

PR #3338, opened 2020-03-01
Backport #2772 add `foldRightDefer` to Foldable

gagandeepkalra:
addresses #3143

### DIFF
--- a/core/src/main/scala/cats/Foldable.scala
+++ b/core/src/main/scala/cats/Foldable.scala
@@ -1,10 +1,11 @@
 package cats
 
-import scala.collection.mutable
+import cats.Foldable.sentinel
 import cats.instances.either._
 import cats.kernel.CommutativeMonoid
 import simulacrum.typeclass
-import Foldable.sentinel
+
+import scala.collection.mutable
 
 /**
  * Data structures that can be folded to a summary value.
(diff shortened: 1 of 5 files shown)


gagandeepkalra:
> java.lang.RuntimeException: Cats laws: Binary compatibility check failed!

I see other laws added without care for BC, did we add this constraint recently?

It's gonna be tricky to follow through with this.

travisbrown:
We've always guaranteed (and checked) bincompat within 2.x for laws on mainline, and I think we should on the 2.11 branch as well (note that this kind of addition only breaks bincompat on 2.11).

My vote would be just not to include the consistency law on this branch, possibly adding a comment to the method along the lines of "Implementers are responsible for ensuring the maintain consistency with `foldRight`; this is not checked by laws on Scala 2.11".

gagandeepkalra:
@travisbrown Think I'm a little confused, I see this added in May yet no BC problem.

Screenshot 2020-03-06 at 10 06 33 PM

yes, adding `foldRightDefer` would only break BC on 2.11, but how was the above law added given it is part of 2.11.

travisbrown:
@gagandeepkalra See @kailuowang's comment on #2817:

> BTW, strictly speaking this is breaking change for `UnorderedFoldableLaws` but it's unlikely to affect anyone and we are breaking BC in cats-laws in our next release.

And mine above:

> We've always guaranteed (and checked) bincompat within 2.x for laws on mainline

That addition was first included in 2.0.0, so while it was a breaking change since 1.x on Scala 2.11, we'd decided that that was okay for cats-laws (we only guaranteed 100% bincompat across 1.x and 2.x for cats-core, etc.).

gagandeepkalra:
I see, thank you. Apologies for not checking the PR first 😔

travisbrown:
No worries, glad that helped!

PR #3339, opened 2020-03-01
Backport #3126 Remove `Order` constraint from `Hash` instance for SortedSet

gagandeepkalra:
addresses #3143


gagandeepkalra:
learned a lot of Scala with this one 😌

PR #3340, opened 2020-03-02
Backport #2380 Add `combineAllOption` to Foldable

gagandeepkalra:
addresses #3143


core/src/main/scala/cats/instances/stream.scala
```diff
@@ -140,6 +140,8 @@ trait StreamInstances extends cats.kernel.instances.StreamInstances {
 
       override def toList[A](fa: Stream[A]): List[A] = fa.toList
 
+      def toIterable[A](fa: Stream[A]): Iterable[A] = fa
```
gagandeepkalra: I don't know how to override this here since `toIterable` is not part of the `Foldable` typeClass, any suggestions?

travisbrown:
There's no need for `FoldableCompat` when building only for 2.11 because `Stream` isn't deprecated until 2.13, so I think it'd be better to implement this directly in the 2.11 branch.

core/src/main/scala/cats/instances/stream.scala
gagandeepkalra: @travisbrown just to confirm again, are we good here?

travisbrown: Oh, hmm. I think we should just remove it entirely, since it won't be part of the public interface anyway.

gagandeepkalra: okay, thanks.

gagandeepkalra: just curious though, if `toIterable` didn't have a default implementation how would have we implemented it then (in a binary compatible way). 

It looks like we would have had to introduce another typeclass in that case?

travisbrown: It's pretty rare to add methods to type classes that don't have default implementations (and this would break bincompat even on 2.12+), so fortunately the question doesn't come up much. But yeah, I think that's what we'd have to do if it did come up somehow.

core/src/main/scala/cats/syntax/foldable.scala
```diff
@@ -331,6 +331,18 @@ final class FoldableOps0[F[_], A](private val fa: F[A]) extends AnyVal {
...
+    if (F.isEmpty(fa)) None else ev.combineAllOption(toIterable)
+
+  /**
+   * Convert F[A] to an Iterable[A].
+   *
+   * This method may be overridden for the sake of performance, but implementers should take care
+   * not to force a full materialization of the collection.
+   */
+  def toIterable(implicit F: Foldable[F]): Iterable[A] =
+    F.foldRight[A, Stream[A]](fa, Eval.now(Stream.empty))((a, eb) => eb.map(Stream.cons(a, _))).value
```
gagandeepkalra: @travisbrown  

Hello, sorry to bother you again on this PR.

I was looking at the `toIterable` implementation to understand why would we use Stream.

These are the possible arguments that I read from the original PR-
> Still aiming at mitigating the possible cost of building a list, it could be delayed up to when the iterator is traversing it

and 

>@travisbrown .toList.iterator forces a full materialization, and we don't control the API of List. So, the uses of a lazy structure (like Iterator, Stream, LazyList) are very different since we can interact with other tools that will not force a full materialization of the data. This is useful in things like spark or scalding.

> That said, I think in a library focused on FP, we should use an immutable lazy structure if possible, so I would lean towards Stream/LazyList. Users can then call .iterator on that, but we don't control that API. Alternatively, they can use foldLeft to tear down the Stream without materializing the whole thing in memory at once.

this is when I added a few `printLn` in between, confirming my suspicion that the Stream is created from the end to the first element, this means that the entire Stream is already materialised yet not visible.

Screenshot 2020-03-05 at 9 09 28 PM

when replacing Stream with List, no difference.


Screenshot 2020-03-05 at 9 16 09 PM

and given this evidence we can replace this entire implementation with-

`F.foldLeft(fa, List.empty[A])((acc, a) => a :: acc).reverse`

Stream is materialised yet not visible then re-materialised (going through the whole `synchronise` thing).

Is my analysis correct?

gagandeepkalra: @travisbrown if you get a chance to go through this one

PR #3341, opened 2020-03-03
Backport #2742 Add `separateFoldable` to Alternative

gagandeepkalra:
addresses #3143

### DIFF
--- a/core/src/main/scala/cats/syntax/alternative.scala
+++ b/core/src/main/scala/cats/syntax/alternative.scala
@@ -66,13 +66,13 @@ final class SeparateOps[F[_], G[_, _], A, B](private val fgab: F[G[A, B]]) exten
    * }}}
    */
   def separateFoldable(implicit AA: Alternative[F], G: Bifoldable[G], FF: Foldable[F]): (F[A], F[B]) =
-      FF.foldLeft(fgab, (AA.empty[A], AA.empty[B])) {
-        case (mamb, gab) =>
-          G.bifoldLeft(gab, mamb)(
-            (t, a) => (AA.combineK(t._1, AA.pure(a)), t._2),
-            (t, b) => (t._1, AA.combineK(t._2, AA.pure(b)))
-          )
-      }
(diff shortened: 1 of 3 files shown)


core/src/main/scala/cats/syntax/alternative.scala
```diff
@@ -49,6 +49,30 @@ final class SeparateOps[F[_], G[_, _], A, B](private val fgab: F[G[A, B]]) exten
...
+   *
+   * Example:
+   * {{{
+   * scala> import cats.implicits._
+   * scala> val l: List[Either[String, Int]] = List(Right(1), Left("error"))
+   * scala> l.separateFoldable
+   * res0: (List[String], List[Int]) = (List(error),List(1))
+   * }}}
+   */
+  def separateFoldable(implicit AA: Alternative[F], G: Bifoldable[G], FF: Foldable[F]): (F[A], F[B]) =
```
travisbrown: It really bothers me now that these instance parameters are in a different order from the ones for `separate`, and that they don't follow the standard practice of matching the order the type parameters appear in, but I think matching the mainline (which is also wrong) is more important than internal consistency.

gagandeepkalra: so you mean the order should have been like this?

`implicit A: Alternative[F], F: Foldable[F], G: Bifoldable[G]`

travisbrown: Oh, sorry that wasn't very clear. I meant

```scala
FF: Foldable[F],  A: Alternative[F], G: Bifoldable[G]
```
…which is the same as the syntax method on mainline (which is correct), but not the type class method (which is wrong, but unfixable for now). We only have the syntax method in this PR, so we should use `Foldable, Alternative, Bifoldable` both because it's consistent with mainline and it matches our general convention.

gagandeepkalra: I see, thank you.

PR #3348, opened 2020-03-08
Backport #3146 Add redeem and redeemWith

gagandeepkalra:
This one had few stuff to backport;

since we are not porting consistency law over, see [here](https://github.com/typelevel/cats/pull/3338#issuecomment-595196728)

and cannot override for specific instances now given we had to maintain BC while implementing, see [here](https://github.com/typelevel/cats/pull/3340#discussion_r388249553)

addresses #3143

### DIFF
--- a/core/src/main/scala/cats/instances/future.scala
+++ b/core/src/main/scala/cats/instances/future.scala
@@ -1,8 +1,8 @@
 package cats
 package instances
 
-import scala.util.control.NonFatal
 import scala.concurrent.{ExecutionContext, Future}
+import scala.util.control.NonFatal
 
 trait FutureInstances extends FutureInstances1 {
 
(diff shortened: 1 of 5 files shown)


core/src/main/scala/cats/instances/future.scala
```diff
@@ -1,37 +1,47 @@
...
+      def handleErrorWith[A](fea: Future[A])(f: Throwable => Future[A]): Future[A] =
+        fea.recoverWith { case t => f(t) }
 
-      def handleErrorWith[A](fea: Future[A])(f: Throwable => Future[A]): Future[A] = fea.recoverWith { case t => f(t) }
+      def raiseError[A](e: Throwable): Future[A] =
+        Future.failed(e)
 
-      def raiseError[A](e: Throwable): Future[A] = Future.failed(e)
-      override def handleError[A](fea: Future[A])(f: Throwable => A): Future[A] = fea.recover { case t => f(t) }
+      override def handleError[A](fea: Future[A])(f: Throwable => A): Future[A] =
```
travisbrown: There are no actual changes in this file, right?

gagandeepkalra: yes, only formatting

PR #3349, opened 2020-03-08
Backport #3076 Add Align typeclass

gagandeepkalra:
addresses #3143

### DIFF
--- a/core/src/main/scala/cats/instances/vector.scala
+++ b/core/src/main/scala/cats/instances/vector.scala
@@ -1,7 +1,7 @@
 package cats
 package instances
 
-import cats.data.ZipVector
+import cats.data.{Ior, ZipVector}
 import cats.syntax.show._
 
 import scala.annotation.tailrec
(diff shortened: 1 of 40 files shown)


gagandeepkalra:
This was was surprisingly easier to do, though I was saving it up for last given the size.

core/src/main/scala/cats/instances/vector.scala
```diff
@@ -110,6 +110,17 @@ trait VectorInstances extends cats.kernel.instances.VectorInstances {
 
       override def algebra[A]: Monoid[Vector[A]] = new kernel.instances.VectorMonoid[A]
 
+      def functor: Functor[Vector] = this
+
+      def align[A, B](fa: Vector[A], fb: Vector[B]): Vector[A Ior B] = {
+        val aLarger = fa.size >= fb.size
+        if (aLarger) {
+          (fa, fb).zipped.map(Ior.both) ++ fa.drop(fb.size).map(Ior.left)
```
gagandeepkalra: main change, inlined here

PR #3350, opened 2020-03-08
Add Align lawful tests for Option instance

gagandeepkalra:
while back-porting `Align` to Scala 2.11, noticed we are missing tests for `Option` instance.


travisbrown:
Can you rebase or merge master and add the import? Thanks for catching this!

travisbrown:
I think it still needs the `cats.Align` import (we recently changed the tests to use imports instead of package nesting in order to match usage more directly).

gagandeepkalra:
Sorry, I did that in a hurry.

gagandeepkalra:
Thank you, Sir 🙂

PR #3357, opened 2020-03-12
Update Scala.js to 1.0, disable Scala.js coverage reporting

travisbrown:
This is like #3312, except:

* It disables coverage reporting on Scala.js, which was causing the `buildJVM` job to fail for some reason.
* It's 1.0.1 instead of 1.0.0.
* It doesn't change the `scala.js-1.0-publish.sh` script to work for 0.6.

If we decide someday that we need to publish a future release for Scala.js 0.6.x, I think we should add a new `scala.js-0.6-publish.sh` at that point, rather than repurposing the 1.0 one.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -122,7 +122,8 @@ lazy val commonJsSettings = Seq(
   scalaJSLinkerConfig := scalaJSLinkerConfig.value.withBatchMode(isTravisBuild.value),
   // currently sbt-doctest doesn't work in JS builds
   // https://github.com/tkawachi/sbt-doctest/issues/52
-  doctestGenTests := Seq.empty
+  doctestGenTests := Seq.empty,
+  coverageEnabled := false
 )
 
 lazy val commonJvmSettings = Seq(
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -10,7 +10,7 @@ addSbtPlugin("org.xerial.sbt" % "sbt-sonatype" % "3.9.2")
 addSbtPlugin("com.47deg" % "sbt-microsites" % "1.2.1")
 addSbtPlugin("org.tpolecat" % "tut-plugin" % "0.6.13")
 addSbtPlugin("org.portable-scala" % "sbt-scalajs-crossproject" % "1.0.0")
-addSbtPlugin("org.scala-js" % "sbt-scalajs" % "1.0.1")
+addSbtPlugin("org.scala-js" % "sbt-scalajs" % "1.1.0")
 addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.4.0")
 addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.9.0")
 
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -1,6 +1,3 @@
-val scalaJSVersion =
-  Option(System.getenv("SCALAJS_VERSION")).getOrElse("0.6.33")
-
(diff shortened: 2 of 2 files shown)


cquiroz:
I guess this is the way to go. I'm a bit worried that not every lib is on scala.js 1.0 and that would limit reach, e.g. we depend on scalajs-react which isn't yet on 1.0.
I'll close my PR about this

travisbrown:
This is green but note that the `validateJS` jobs are now incredibly slow. I'll take a look at this locally today.

travisbrown:
@cquiroz I'm happy to publish for 0.6 if we decide that that's what's needed—I just don't really want to sign up for that in advance. Let's see when 2.2.0 comes out, etc. It's always easy to add a new cross-publishing script for 0.6.

larsrh:
It looks like coverage isn't reported at all anymore?

travisbrown:
@larsrh Oh, wow, what a mess. I might not get to this today, but I don't think there's any real hurry.

larsrh:
After looking at the build logs, it seems Coverage _is_ reported, it just doesn't make its way into a comment on this PR: https://codecov.io/gh/typelevel/cats/pull/3357

travisbrown:
Just refreshed this branch (I also went ahead and bumped the Scala.js version, so this supersedes #3418).

I confirmed that `sbt clean validateJS` doesn't run more slowly than master locally (3m17.515s vs. 3m51.507s, so actually a little faster).

I also confirmed that `validateJS` coverage is working locally for JVM modules (and there's a Codecov notification now), so I think this is ready to merge.  @LukaJCB, @larsrh, @fthomas, mind a quick review?

travisbrown:
The Scala.js builds are still slower in CI, and I'm not sure why, but I don't think that should block this PR.

travisbrown:
Sorry, @ChristopherDavenport, @larsrh, just had to resolve a plugins.sbt conflict, so I'll need at least one re-review?

travisbrown:
Ugh, now we're back to "builds take incredibly long or time out" instead of just "builds are slower". I asked about this [in the Scala.js Gitter channel](https://gitter.im/scala-js/scala-js?at=5ecb8927ff7a920a720d657c), but I think we'll have to hold off on this PR for now.

travisbrown:
Capturing these comments from Sébastien on Gitter for the record:

> Does `ThisBuild / scalaJSLinkerConfig ~= { _.withCheckIR(false) }` help?
> >    used to take ~12 minutes now take 40+
>
> hum that's probably not `checkIR`, now that I think of it.
> It would be worth knowing exactly what step is slower, especially between you local build and the CI build.

>    Can you decompose the timings between `theProject/test:fastOptJS` and `theProject/test`? So that we know whether it is linking or actually running the tests which is slower. From the `[info] Run completed in 12 minutes` log line, I actually suspect the latter, but who knows.
>    If my guess is wrong and it's the former, could you output `last` just after running `theProject/test:fastOptJS`, to obtain the timings of individual steps inside linking?
> If it's actually running the tests which is slower, one thing that's worth trying is whether disabling parallel testing helps (or makes it worse). That would be with Test / parallelExecution := false.


And:

> You've got useless stuff there, but nothing harmful.
> Your settings for `scalaJSStage` and `jsEnv` are the default ones, so you could remove those without changing anything.
> Otherwise this is reasonable.

travisbrown:
This is green again, and I'm going to go ahead and merge it. Just switching to `FullOptStage` cuts the test times way down ([around 24 minutes in this run](https://travis-ci.org/github/typelevel/cats/builds/691682531)), and I think we can improve some other stuff, but that can happen in a follow-up PR.

PR #3358, opened 2020-03-15
ApplicativeError: add raiseOption operation.

diesalbla:
Add a `raiseOption` method to the `ApplicativeError` type-class, to raise possible errors if they are in there.

### DIFF
--- a/core/src/main/scala/cats/syntax/option.scala
+++ b/core/src/main/scala/cats/syntax/option.scala
@@ -335,6 +335,27 @@ final class OptionOps[A](private val oa: Option[A]) extends AnyVal {
   def liftTo[F[_]]: LiftToPartiallyApplied[F, A] = new LiftToPartiallyApplied(oa)
 
   /**
+   * Raise to an F[Unit], as long as it has an ApplicativeError[F, A] instance
+   * If the option is empty, an empty unit effect is given.
+   * If the option contains an error, it is raised.
+   *
+   * Example:
+   * {{{
+   * scala> import cats.implicits._
+   *
+   * scala> type F[A] = Either[String, A]
+   *
+   * scala> Option.empty[String].raiseTo[F]
+   * res0: scala.Either[String, Unit] = Right(())
+   *
+   * scala> Option("Failed").raiseTo[F]
+   * res1: scala.Either[String, Unit] = Left(Failed)
+   * }}}
+   */
+  def raiseTo[F[_]](implicit F: ApplicativeError[F, A]): F[Unit] =
+    oa.fold(F.unit)(F.raiseError)
+
+  /**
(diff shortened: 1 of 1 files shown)


core/src/main/scala/cats/ApplicativeError.scala
```diff
@@ -273,6 +273,28 @@ trait ApplicativeError[F[_], E] extends Applicative[F] {
       case None    => raiseError(ifEmpty)
     }
 
+  /**
+   * Convert from scala.Option when the content is a value of the error type.
+   * If the option is empty, an empty unit effect is given.
+   * If the option contains an error, it is raised.
+   * or returns an empty action
```
travisbrown: This `or returns an empty action` seems to be dangling, and duplicates the second sentence?

core/src/main/scala/cats/ApplicativeError.scala
```diff
@@ -273,6 +273,28 @@ trait ApplicativeError[F[_], E] extends Applicative[F] {
...
+   *
+   * Example:
+   * {{{
+   * scala> import cats.implicits._
+   * scala> import cats.ApplicativeError
+   * scala> val F = ApplicativeError[Either[String, *], String]
+   *
+   * scala> F.raiseOption(Option.empty[String])
+   * res0: scala.Either[String, ()] = Right(())
```
travisbrown: Scala won't accept `()` as a type like this, will it?

core/src/main/scala/cats/syntax/option.scala
```diff
@@ -334,6 +334,27 @@ final class OptionOps[A](private val oa: Option[A]) extends AnyVal {
...
+   * scala> import cats.implicits._
+   * scala> import cats.ApplicativeError
+   * scala> implicit val F = ApplicativeError[Either[String, *], String]
+   *
+   * scala> Option.empty[String].raiseTo[F]
+   * res0: scala.Either[String, Unit] = Right(())
+   *
+   * scala> Option("Failed").raiseTo[F]
+   * res1: scala.Either[String, Unit] = Left("Failed")
+   * }}}
```
barambani, line 353: To fix  the ci error, the doctest  could be changed to something like
```scala
/**
 * Raise to an F[Unit], as long as it has an ApplicativeError[F, A] instance
 * If the option is empty, an empty unit effect is given.
 * If the option contains an error, it is raised.
 *
 * Example:
 * {{{
 * scala> import cats.implicits._
 *
 * scala> type F[A] = Either[String, A]
 *
 * scala> Option.empty[String].raiseTo[F]
 * res0: scala.Either[String, Unit] = Right(())
 *
 * scala> Option("Failed").raiseTo[F]
 * res1: scala.Either[String, Unit] = Left(Failed)
 * }}}
 */
```

LukaJCB:
Thanks @diesalbla!

PR #3359, opened 2020-03-17
add pos and seek to RepresentableStore

massimosiani:
This adds a couple of helper functions to `RepresentableStore`, ported from the `ComonadStore` Haskell typeclass.
Thanks for your work!


larsrh:

Issue #3360, opened 2020-03-18
Consider removing attemptNarrow


travisbrown:
Somehow I'd never paid attention to `attemptNarrow`, which we added in 2.1.0. It has some problems. For example it's trivially easy to make it crash at runtime:

```scala
scala> val e: Either[Any, Unit] = Left(List("abc"))
e: Either[Any,Unit] = Left(List(abc))

scala> import cats.implicits._
import cats.implicits._

scala> def x = e.attemptNarrow[List[Int]] match {
     |   case Right(Left(List(x))) => x
     |   case _ => 0
     | }
x: Int

scala> x
java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Integer
  at scala.runtime.BoxesRunTime.unboxToInt(BoxesRunTime.java:99)
  at .x(<console>:2)
  ... 36 elided
```
It can also be used to launder a runtime type check:

```scala
scala> import cats.implicits._, scala.reflect.ClassTag
import cats.implicits._
import scala.reflect.ClassTag

scala> def isIt[A: ClassTag](any: Any): Boolean =
     |   (Left(any): Either[Any, Unit]).attemptNarrow[A].isRight
isIt: [A](any: Any)(implicit evidence$1: scala.reflect.ClassTag[A])Boolean

scala> val x = if (true) 1 else "abc"
x: Any = 1

scala> isIt[String](x)
res0: Boolean = false

scala> isIt[Int](x)
res1: Boolean = true
```
Maybe that's fine, and you should know that if you've got a `ClassTag` around you've given up any hope of parametricity, but that seems pretty far from the spirit of Cats.

@tpolecat [requested tests](https://github.com/typelevel/cats/pull/2863#pullrequestreview-270341674) for the erasure issues in the original PR, but unfortunately the tests that were added don't actually check for the relevant problems.

I guess the reasoning is that `E` is usually an exception type, and generally won't be generic? In any case I think we should make both the type class and syntax methods package-private in future 2.x releases and remove them in 3.0, or at least put an explicit `<: Throwable` constraint on them (which doesn't eliminate the possibility of runtime crashes, but does make them much less likely).

travisbrown:
This is partially addressed by #3361. In the longer run we should consider what to do with `attemptNarrow` and `catchOnly` (probably we should just add documentation about the risks, but we could also consider deprecation).

lourkeur:
[...]

PR #3361, opened 2020-03-18
Restrict attemptNarrow to subtypes of Throwable

travisbrown:
This change makes it impossible to call `attemptNarrow` with a type parameter that isn't `Throwable` or one of its subtypes. This makes it slightly less easy to shoot yourself in the foot with it (see #3360), while still supporting its most common use cases.

The change doesn't break binary compatibility, but definitely does break source compatibility. It only breaks dangerous code, though, so I think that's fine for 2.2.0.

Personally I'd prefer to scrap it entirely (by making it package-private), since even in this restricted form it breaks parametricity in a way that feels un-Cats-like. I've chosen this route because of the precedent of `catchOnly`, which does a similar thing with runtime reflection but makes it relatively safe by restricting to `Throwable`.

### DIFF
--- a/core/src/main/scala/cats/ApplicativeError.scala
+++ b/core/src/main/scala/cats/ApplicativeError.scala
@@ -80,7 +80,7 @@ trait ApplicativeError[F[_], E] extends Applicative[F] {
   /**
    * Similar to [[attempt]], but it only handles errors of type `EE`.
    */
-  def attemptNarrow[EE, A](fa: F[A])(implicit tag: ClassTag[EE], ev: EE <:< E): F[Either[EE, A]] =
+  def attemptNarrow[EE <: Throwable, A](fa: F[A])(implicit tag: ClassTag[EE], ev: EE <:< E): F[Either[EE, A]] =
     recover(map(fa)(Right[EE, A](_): Either[EE, A])) { case e: EE => Left[EE, A](e) }
 
   /**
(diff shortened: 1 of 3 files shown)


travisbrown:
@tpolecat Out of curiosity, do you feel the same way about our `catchOnly` for `Either` and `Validated`, which does pretty much exactly the same thing as the version of `attemptNarrow` in this PR? In principle I definitely hate them both, but I also know `catchOnly` at least can be useful.

tpolecat:
@travisbrown yes, same. I dislike them both, but I can accept that utility in the common case probably outweighs unsoundness in the general case. But it does seem like we should mention it in the doc.

rossabaker:
If the eventual goal is to remove it, would it make more sense to just deprecate it?

travisbrown:
@rossabaker I see this PR is a first step, to get `attemptNarrow` to the same level of safety as `catchOnly` now, for inclusion in the 2.2.0 release.

That puts us in a position to communicate the risks of `attemptNarrow` and `catchOnly` together (since they'll represent the same kind of use of runtime reflection), and to decide what to do about them together in the future (deprecate them both, make them both package-private, leave them, etc.).

For what it's worth I was looking at `attemptNarrow` yesterday because a colleague said "I just found `attemptNarrow` and it's really useful". It might be reasonable to decide it has a place in Cats—for now I just think we should make our two uses of class tags look the same, instead of having one be mostly safe except for really weird situations and the other completely unsafe.

PR #3362, opened 2020-03-19
Prepend a Vector to a NonEmptyVector

ssanj:
I needed to implement this recently and was surprised it didn't already exist.

### DIFF
--- a/core/src/main/scala/cats/data/NonEmptyVector.scala
+++ b/core/src/main/scala/cats/data/NonEmptyVector.scala
@@ -118,8 +118,9 @@ final class NonEmptyVector[+A] private (val toVector: Vector[A])
   /**
    * Prepend a `Vector` to this, producing a new `NonEmptyVector`.
    */
-  def prependVec[AA >: A](vector: Vector[AA]): NonEmptyVector[AA] = 
+  def prependVec[AA >: A](vector: Vector[AA]): NonEmptyVector[AA] =
     NonEmptyVector.fromVector(vector).fold[NonEmptyVector[AA]](this)(_.concatNev(this))
+
   /**
    * Alias for [[prepend]]
    */
--- a/core/src/main/scala/cats/data/NonEmptyVector.scala
+++ b/core/src/main/scala/cats/data/NonEmptyVector.scala
@@ -118,8 +118,8 @@ final class NonEmptyVector[+A] private (val toVector: Vector[A])
   /**
    * Prepend a `Vector` to this, producing a new `NonEmptyVector`.
    */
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/data/NonEmptyVector.scala
```diff
@@ -115,6 +115,12 @@ final class NonEmptyVector[+A] private (val toVector: Vector[A])
    */
   def prepend[AA >: A](a: AA): NonEmptyVector[AA] = new NonEmptyVector(a +: toVector)
 
+  /**
+   * Prepend a `Vector` to this, producing a new `NonEmptyVector`.
+   */
+  def prependVec[AA >: A](vector: Vector[AA]): NonEmptyVector[AA] =
```
travisbrown: What do you think of naming this `prependVector` for consistency, on the model of the existing `prependLazyList` and `prependChain`?

ssanj: Sounds good. Happy to rename it so it's consistent with similar constructs. 👍

travisbrown: Thanks much!

ssanj:
Only a sightly related note do you know why the internal implementation of `NonEmptyVector` is unsafe:

```
final class NonEmptyVector[+A] private (val toVector: Vector[A])
``` 

and we have  implementations like this:

```
def head: A = toVector.head
```

which are also unsafe?

I presume clients of the API can only create an instance of this class through its typesafe companion functions. But this would still open up the possibility to create a `NonEmptyVector` incorrectly from within the class itself.

`NonEmptyList` does not seem to be implemented similarly and is typesafe at construction:

```
final case class NonEmptyList[+A](head: A, tail: List[A]) extends NonEmptyCollection[A, List, NonEmptyList] {
```

larsrh:
Closing and re-opening to trigger build

travisbrown:
@ssanj There's a lot of unmotivated inconsistency in the non-empty collections implementations, but in this case the difference is intentional. We want conversions between the standard library collection types and their non-empty `cats.data` partners to be safe first, but we also want them to be efficient.

It's both safe and efficient to check that a `List` is non-empty and convert it into a `(head, tail)` pair. This isn't the case for `Vector`, which is represented as a balanced tree, which means that taking the tail involves an amount of copying that's significant for our purposes. While it would be ideal to use a representation like `NonEmptyList`'s that makes it impossible to construct invalid values at all, either using public or private APIs, in the case of `NonEmptyVector` that would be prohibitively inefficient, so we use an internally unsafe representation and exclude the unsafe parts from the public API.

ssanj:
Thanks for the explanation @travisbrown 👍

PR #3363, opened 2020-03-19
implement the EnvT data type and its instances

massimosiani:
I tried to implement the Env comonad, and its transformer EnvT, as the next to simplest example of a Comonad.
Thanks!

### DIFF
--- a/core/src/main/scala/cats/data/EnvT.scala
+++ b/core/src/main/scala/cats/data/EnvT.scala
@@ -1,6 +1,6 @@
 package cats.data
 
-import cats.{Applicative, Comonad, Eval, Foldable, Functor, Traverse}
+import cats.{Applicative, Comonad, Eval, Foldable, Functor, Monoid, Traverse}
 
 /**
  * EnvT[W, E, A] is the Env Comonad transformer
@@ -10,7 +10,7 @@ final case class EnvT[W[_], E, A](wa: W[A], env: E) {
   /**
    * Return the environment
    */
-  lazy val ask: E = env
+  def ask: E = env
 
   /**
    * Return a transformed environment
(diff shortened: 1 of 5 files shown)


core/src/main/scala/cats/data/package.scala
```diff
@@ -77,6 +77,11 @@ package object data extends ScalaVersionSpecificPackage {
   type RWS[E, L, S, A] = ReaderWriterState[E, L, S, A]
   val RWS = ReaderWriterState
 
+  type Env[E, A] = EnvT[Id, E, A]
+  object Env {
+    def apply[E, A](e: E, a: Id[A]): Env[E, A] = EnvT(a, e)
```
rossabaker: Any reason not just `a: A` here?

The arguments are inconsistent with the order in `EnvT.apply`.

core/src/main/scala/cats/data/EnvT.scala
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * EnvT[W, E, A] is the Env Comonad transformer
+ */
+final case class EnvT[W[_], E, A](wa: W[A], env: E) {
+
+  /**
+   * Return the environment
+   */
+  lazy val ask: E = env
```
rossabaker: I don't think this needs to be lazy.  I would just make it a def.

core/src/main/scala/cats/data/EnvT.scala
```diff
@@ -0,0 +1,58 @@
...
+  def local[E1](f: E => E1): EnvT[W, E1, A] = EnvT(wa, f(env))
+
+  def map[B](f: A => B)(implicit W: Functor[W]): EnvT[W, E, B] = EnvT(W.map(wa)(f), env)
+
+  lazy val runEnvT: (E, W[A]) = (env, wa)
+}
+
+object EnvT {
+
+  implicit def catsDataEnvTComonad[W[_]: Comonad, E]: Comonad[EnvT[W, E, *]] = new Comonad[EnvT[W, E, *]] {
```
rossabaker: You can get a `Functor[EnvT[W, E, *]]` from `Functor[W]`, if you want to split this up like `Foldable` and `Traversable`.

core/src/main/scala/cats/data/EnvT.scala
```diff
@@ -0,0 +1,58 @@
...
+   * Transform the environment
+   */
+  def local[E1](f: E => E1): EnvT[W, E1, A] = EnvT(wa, f(env))
+
+  def map[B](f: A => B)(implicit W: Functor[W]): EnvT[W, E, B] = EnvT(W.map(wa)(f), env)
+
+  lazy val runEnvT: (E, W[A]) = (env, wa)
+}
+
+object EnvT {
```
rossabaker: There's an `Applicative[EnvT[W, E, *]]` from `Monoid[E]` and `Applicative[W]`.

core/src/main/scala/cats/data/EnvT.scala
```diff
@@ -0,0 +1,58 @@
...
+  def asks[B](f: E => B): B = f(env)
+
+  /**
+   * Transform the environment
+   */
+  def local[E1](f: E => E1): EnvT[W, E1, A] = EnvT(wa, f(env))
+
+  def map[B](f: A => B)(implicit W: Functor[W]): EnvT[W, E, B] = EnvT(W.map(wa)(f), env)
+
+  lazy val runEnvT: (E, W[A]) = (env, wa)
```
rossabaker: I'd also skip the laziness here, though you do save a tuple if it's called repeatedly.

My gut feeling is these should be returned in the same order as the constructor.  I don't have a strong feeling which order is correct.

massimosiani:
Hi @rossabaker and thanks for review!
Besides my interest in its own, comonads can be used for writing interpreters for (free) monads. For instance, I'd like to contribute by adding CofreeT, and the corresponding typeclasses in cats-mtl (ComonadStore and ComonadEnv for instance).

Sorry if was sloppy, probably the discussion doesn't fit in a PR comment :)

Let me know what you think about this!

core/src/main/scala/cats/data/package.scala
massimosiani: Well, the compiler complained, but a type annotation worked.

core/src/main/scala/cats/data/EnvT.scala
massimosiani: Makes sense to me.

larsrh:

PR #3368, opened 2020-03-24
Update Scala 2.12 to 2.12.11

jhnsmth:
Release notes https://github.com/scala/scala/releases/tag/v2.12.11

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ jdk:
   - openjdk8
 
 
-scala_version_212: &scala_version_212 2.12.10
+scala_version_212: &scala_version_212 2.12.11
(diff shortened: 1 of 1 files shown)


larsrh:
This is weird. `tut` fails because the compiler complains about an inexhaustive pattern match (e.g. here https://github.com/typelevel/cats/blob/424569f9c7dd7e99f6a6d1b72980db00b75627cb/docs/src/main/tut/typeclasses/lawtesting.md#L40-L43).

jhnsmth:
Thanks for looking into this @larsrh. It's indeed strange. I guess the problem is that tut is not full cross compiled.

barambani:
This was resolved here https://github.com/typelevel/cats/pull/3471 and can be closed.

PR #3374, opened 2020-03-29
Add EitherT.liftAttemptK

kubukoz:


core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -742,6 +742,25 @@ object EitherT extends EitherTInstances {
...
+   * res0: Option[Either[Unit, Int]] = Some(Left(()))
+   *
+   * scala> val a2: Option[Int] = Some(42)
+   * scala> val b2: EitherT[Option, Unit, Int] = EitherT.liftAttemptK[Option, Unit].apply(a2)
+   * scala> b2.value
+   * res1: Option[Either[Unit, Int]] = Some(Right(42))
+   * }}}
+   */
+  final def liftAttemptK[F[_], E](implicit F: ApplicativeError[F, E]): F ~> EitherT[F, E, *] =
+    λ[F ~> EitherT[F, E, *]](fa => EitherT(F.attempt(fa)))
```
travisbrown: We've been removing the value-level kind-projector syntax for polymorphic functions, since there's no real possibility of it ever cross-building on Dotty, and since writing this out explicitly as:

```scala
new (F ~> EitherT[F, E, *]) { def apply[A](fa: F[A]): EitherT[F, E, A] = EitherT(F.attempt(fa)) }
```
…is only marginally less readable. I don't think that needs to be changed in this PR, though, since we're not actually cross-building on Dotty yet.

(Note that the type-level `λ` is fine on Dotty with `-Ykind-projector`.)

kubukoz: oh, I was under the impression this was also part of the `-Ykind-projector` feature. But it makes sense, I'll keep it in mind.

travisbrown: @kubukoz No, I hate kind-projector's value-level lambda syntax and couldn't be bothered to implement it in `-Ykind-projector`. :smile: (Also it would have required much more disruptive changes than anything else that made it into `-Ykind-projector`.)

Issue #3327, opened 2020-02-28
TraverseFilter won't short circuit 


gagandeepkalra:
`TraverseFilter` won't short-circuit for `Set`, `LazyList`, `Vector` and `Queue`. For other instances, it's done correctly.

e.g.

```
import cats.implicits._
import scala.collection.immutable.Queue

scala> val f: Int => Either[String, Option[Int]] = i => {
  print(s"$i ")
  if (i > 0) Right(Some(i)) else Left("not positive")
}

scala> List(2, 1, 0, -1).traverseFilter(f)
2 1 0 res0: Either[String,List[Int]] = Left(not positive)

scala> Vector(2, 1, 0, -1).traverseFilter(f)
-1 0 1 2 res1: Either[String,scala.collection.immutable.Vector[Int]] = Left(not positive)

scala> Queue(2, 1, 0, -1).traverseFilter(f)
-1 0 1 2 res2: Either[String,scala.collection.immutable.Queue[Int]] = Left(not positive)

```

[Conversation](https://github.com/typelevel/cats/pull/3292#discussion_r377590934) for reference.

travisbrown:
Thanks for opening this, @gagandeepkalra! I'm planning to take a closer look at this in the next few days.

travisbrown:
One thing that's worth noting here: the bug doesn't extend to laziness of effect evaluation, in which respect (as far as I've checked) all of these instances are correct. The issue is that for strict types with a failed state (like `Either`) the function is applied in cases where it doesn't need to be, and where it isn't for e.g. the `List` instance.

PR #3375, opened 2020-03-31
Add laws to check Short-Circuiting behaviour  

### LINKED ISSUE #3355: Specify short-circuiting behavior and add tests or laws
There's currently some inconsistency between our collection instances with respect to where short-circuiting happens. To take a couple of examples:

```scala
import cats.implicits._

scala> import cats.Eval, cats.implicits._
import cats.Eval
import cats.implicits._

scala> List.empty[String].map2Eval(Eval.later[List[Int]](sys.error("explode")))((x, _) => x)
res0: cats.Eval[List[String]] = Now(List())

scala> res0.value
res1: List[String] = List()

scala> Vector.empty[String].map2Eval(Eval.later[Vector[Int]](sys.error("explode")))((x, _) => x)
res2: cats.Eval[scala.collection.immutable.Vector[String]] = cats.Eval$$anon$4@2030bbf1

scala> res2.value
java.lang.RuntimeException: explode
  at scala.sys.package$.error(package.scala:30)
  at .$anonfun$res2$1(<console>:19)
  at cats.Later.value$lzycompute(Eval.scala:148)
  at cats.Later.value(Eval.scala:147)
  at cats.Eval$.loop$1(Eval.scala:345)
  at cats.Eval$.cats$Eval$$evaluate(Eval.scala:366)
  at cats.Eval$FlatMap.value(Eval.scala:305)
  ... 36 elided
```
…or:
```scala
import cats.implicits._

scala> val f: Int => Either[String, Option[Int]] = i => {
     |   print(s"$i ")
     |   if (i > 0) Right(Some(i)) else Left("not positive")
     | }
f: Int => Either[String,Option[Int]] = $$Lambda$7369/108699853@1ccdbb90

scala> List(2, 1, 0, -1).traverse(f)
2 1 0 res0: Either[String,List[Option[Int]]] = Left(not positive)

scala> List(2, 1, 0, -1).traverseFilter(f)
2 1 0 res1: Either[String,List[Int]] = Left(not positive)

scala> Vector(2, 1, 0, -1).traverse(f)
2 1 0 res2: Either[String,scala.collection.immutable.Vector[Option[Int]]] = Left(not positive)

scala> Vector(2, 1, 0, -1).traverseFilter(f)
-1 0 1 2 res3: Either[String,scala.collection.immutable.Vector[Int]] = Left(not positive)
```
Some specific cases of this are tracked in #3327 and fixed in #3328, but even after those changes we don't have tests to make sure we don't introduce more inconsistency, and we don't do a very good job of communicating to users what they can expect.

gagandeepkalra:
resolves #3355, previously #3327

### DIFF
--- a/alleycats-tests/shared/src/test/scala/alleycats/tests/MapSuite.scala
+++ b/alleycats-tests/shared/src/test/scala/alleycats/tests/MapSuite.scala
@@ -2,11 +2,13 @@ package alleycats.tests
 
 import cats.Traverse
 import cats.instances.all._
-import cats.laws.discipline.{SerializableTests, TraverseFilterTests}
 import cats.laws.discipline.arbitrary._
+import cats.laws.discipline.{SerializableTests, ShortCircuitingTests, TraverseFilterTests}
 
 class MapSuite extends AlleycatsSuite {
   checkAll("Traverse[Map[Int, *]]", SerializableTests.serializable(Traverse[Map[Int, *]]))
 
   checkAll("TraverseFilter[Map[Int, *]]", TraverseFilterTests[Map[Int, *]].traverseFilter[Int, Int, Int])
+
+  checkAll("Map[Int, *]", ShortCircuitingTests[Map[Int, *]].traverseFilter[Int])
 }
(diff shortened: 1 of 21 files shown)


travisbrown:
This looks good to me!

PR #3382, opened 2020-04-02
Optimise NonEmptyTraverse implementation

### LINKED ISSUE #3381: Optimise NonEmptyTraverse implementation
The current `NonEmptyTraverse` for `List` goes something like,

    def nonEmptyTraverse[G[_], A, B](nel: NonEmptyList[A])(f: A => G[B])(implicit G: Apply[G]): G[NonEmptyList[B]] =
      Foldable[List]
        .reduceRightToOption[A, G[List[B]]](nel.tail)(a => G.map(f(a))(_ :: Nil)) { (a, lglb) =>
        }
        .map {
          case None        => G.map(f(nel.head))(NonEmptyList(_, Nil))
          case Some(gtail) => G.map2(f(nel.head), gtail)(NonEmptyList(_, _))
        }
        .value

The difference and the real challenge with the implementation is the absence of G.pure given the constraint is now `Apply` instead of `Applicative`. Other `NonEmptyInstances` follow the same pattern for implementation.

The following is the default not overridden `reduceRightToOption` implementation

    def reduceRightToOption[A, B](fa: F[A])(f: A => B)(g: (A, Eval[B]) => Eval[B]): Eval[Option[B]] =
      foldRight(fa, Now(Option.empty[B])) { (a, lb) =>
        lb.flatMap {
          case Some(b) => g(a, Now(b)).map(Some(_))
          case None    => Later(Some(f(a)))
        }
      }

Proposal, the above implementation can be rewritten as 

    def nonEmptyTraverse[G[_], A, B](nel: NonEmptyList[A])(f: A => G[B])(implicit G: Apply[G]): G[NonEmptyList[B]] = {
      def loop(head: A, tail: List[A]): Eval[G[NonEmptyList[B]]] =
        tail match {
          case Nil    => Eval.now(G.map(f(head))(NonEmptyList(_, Nil)))
          case h :: t => G.map2Eval(f(head), Eval.defer(loop(h, t)))((b, acc) => NonEmptyList(b, acc.toList))
        }

      loop(nel.head, nel.tail).value
    }

This is better for it has,

1. less memory overhead
2. actually short-circuits if G.map2Eval shorts, the original implementation doesn't
3. more readable?

I can add this up with #3355 to add `NonEmptyTraverse` short-circuiting law, ensuring all other implementations follow suit. I couldn't come up with any other implementation that short-circuits.

gagandeepkalra:
resolves #3381

depends on #3375

### DIFF
--- a/core/src/main/scala-2.13+/cats/data/NonEmptyLazyList.scala
+++ b/core/src/main/scala-2.13+/cats/data/NonEmptyLazyList.scala
@@ -473,16 +473,14 @@ sealed abstract private[data] class NonEmptyLazyListInstances extends NonEmptyLa
 
       def extract[A](fa: NonEmptyLazyList[A]): A = fa.head
 
-      def nonEmptyTraverse[G[_]: Apply, A, B](fa: NonEmptyLazyList[A])(f: A => G[B]): G[NonEmptyLazyList[B]] =
-        Foldable[LazyList]
-          .reduceRightToOption[A, G[LazyList[B]]](fa.tail)(a => Apply[G].map(f(a))(LazyList.apply(_))) { (a, lglb) =>
-            Apply[G].map2Eval(f(a), lglb)(_ +: _)
+      def nonEmptyTraverse[G[_]: Apply, A, B](fa: NonEmptyLazyList[A])(f: A => G[B]): G[NonEmptyLazyList[B]] = {
+        def loop(head: A, tail: LazyList[A]): Eval[G[NonEmptyLazyList[B]]] =
+          tail.headOption.fold(Eval.now(Apply[G].map(f(head))(NonEmptyLazyList(_)))) { h =>
+            Apply[G].map2Eval(f(head), Eval.defer(loop(h, tail.tail)))((b, acc) => b +: acc)
           }
-          .map {
-            case None        => Apply[G].map(f(fa.head))(h => create(LazyList(h)))
-            case Some(gtail) => Apply[G].map2(f(fa.head), gtail)((h, t) => create(LazyList(h) ++ t))
-          }
-          .value
+
+        loop(fa.head, fa.tail).value
+      }
 
       def reduceLeftTo[A, B](fa: NonEmptyLazyList[A])(f: A => B)(g: (B, A) => B): B = fa.reduceLeftTo(f)(g)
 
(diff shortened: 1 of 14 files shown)


chenguohui49:
[IntegratingAccessibilityIntoYourApp 3.zip](https://github.com/typelevel/cats/files/4441137/IntegratingAccessibilityIntoYourApp.3.zip)

core/src/main/scala-2.13+/cats/data/NonEmptyLazyList.scala
```diff
@@ -458,16 +458,14 @@ sealed abstract private[data] class NonEmptyLazyListInstances extends NonEmptyLa
...
 
-      def nonEmptyTraverse[G[_]: Apply, A, B](fa: NonEmptyLazyList[A])(f: A => G[B]): G[NonEmptyLazyList[B]] =
-        Foldable[LazyList]
-          .reduceRightToOption[A, G[LazyList[B]]](fa.tail)(a => Apply[G].map(f(a))(LazyList.apply(_))) { (a, lglb) =>
-            Apply[G].map2Eval(f(a), lglb)(_ +: _)
+      def nonEmptyTraverse[G[_]: Apply, A, B](fa: NonEmptyLazyList[A])(f: A => G[B]): G[NonEmptyLazyList[B]] = {
+        def loop(head: A, tail: LazyList[A]): Eval[G[NonEmptyLazyList[B]]] =
+          tail.headOption.fold(Eval.now(Apply[G].map(f(head))(NonEmptyLazyList(_)))) { h =>
+            Apply[G].map2Eval(f(head), Eval.defer(loop(h, tail.tail)))((b, acc) => b +: acc)
           }
```
diesalbla, line 465: There is a bit of LISP here. This code may be more readable as follows: 

```suggestion
          val fh = f(head)
          tail match {
            case _ if tail.isEmpty => 
              Eval.now(Apply[G].map(fh)(NonEmptyLazyList(_)))
            case h #:: ttail => 
              Apply[G].map2Eval(fh, Eval.defer(loop(h, ttail)))(_ +: _)
          }
```
The changes are: 

- Extract `f(head)`, which is eagerly evaluated in both branches, as a `val`.
- Replace the `Option.fold` by a pattern-match.
- Replace the cases of `headOption` by two cases on the `tail` lazy list itself. For the first case, there seems not to be any case-object, like `Nil`, for the empty list. [Discussion](https://users.scala-lang.org/t/pattern-matching-lazylist/5105).
- In the second case, abbreviate `(b, acc) => b +: acc` by underscores `_ +: _`.

gagandeepkalra: Hello, I see what you mean there. 

On my part, usually I try to avoid pattern match (it's slow for the newer object allocations)

gagandeepkalra:
Thank you @travisbrown @LukaJCB, this was fun. I'm happy knowing short-circuiting is already useful (I see foldable there).

PR #3383, opened 2020-04-02
Fix a link in the documentation for Traverse

justinhj:


justinhj:
BTW I had to move into the folder `cats/docs/target/scala-2.12/resource_managed/main/jekyll` to test the site. The contributor docs say `docs/target/site` but that generated errors for when running Jekyll. Not sure if I'm doing something wrong.

PR #3390, opened 2020-04-11
Add document about Arrow Choice

jcouyang:
`Choice` is actually very useful typeclass, we use that a lot in our Scala and PureScript project, but it seems lack of document in cats elaborate how to actually make use of it.

### DIFF
--- a/docs/src/main/tut/typeclasses/arrowchoice.md
+++ b/docs/src/main/tut/typeclasses/arrowchoice.md
@@ -11,15 +11,15 @@ scaladoc: "#cats.arrow.ArrowChoice"
 
 Usually we deal with function more often, we're so familiar with `A => B`.
 
-If we have two function `A => C` and `B => C`, how can we compose them into a single function that can take either A or B and produce a C?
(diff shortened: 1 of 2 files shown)


jcouyang:
thanks @julienrf 
didn’t notice it’s still open 🤣

jcouyang:
hmm..anyone keen to have a look the docs?

docs/src/main/tut/typeclasses/arrowchoice.md
```diff
@@ -0,0 +1,113 @@
...
+If we like to have an authentication middleware compose the route, we can simply define middleware as:
+
+```scala
+type Middleware[F[_]] = Kleisli[OptionT[F, ?], Request[F], Either[Response[F], Request[F]]]
+def auth[F[?]]: Middleware[F] = ???
+```
+
+Which means the `Request[F]` goes through the middleware, will become option of `Either[Response[F], Request[F]]`, where `Left` means the request is denied and return immediately, `Right` means the authentication is OK and request will get pass.
+
+Now we need to define what should we do when middleware returns `Left`:
```
barambani, line 49:
```suggestion
Now we need to define what we should do when middleware returns `Left`:
```

PR #3391, opened 2020-04-11
Migrate to mdoc

### LINKED ISSUE #2927: Migrate to mdoc

larsrh:
_Note: this PR is free to be picked up by someone who wants to continue working on it_

Closes #2927.

## General observations

- `sbt-microsites` configures the extra MD thingy of mdoc incorrectly
- mdoc pretends that it processes all MD files into HTML, so it prints tons of bogus warnings on link hygiene (here, Jekyll is used to process the MD files into HTML)

## redeclaration semantics

mdoc has some wildly different semantics from tut when re-defining classes or values. tut just lets later definitions override previous ones. In mdoc they need to be unique, except for when using reset, which resets everything in the document. The problem is that some `:fail` blocks are now succeeding for the wrong reason; they should fail because there's a type error in there, but now they fail because of this. I'm unsure what to do about this.

(NB I'm aware of `nest`, but unfortunately when used liberally it appears to crash the compiler)


larsrh:
This PR is severely outdated, so I'm going to close it. Anyone attempting to migrate from tut to mdoc should probably start afresh.

PR #3392, opened 2020-04-13
Add PartialFunction instance for Profunctor typeclass

### LINKED ISSUE #3386: Profunctor[PartialFunction]
Relatively certain this is lawful (credit: @wemrysi). Should be straightforward to add.

gagandeepkalra:
resolves #3386

### DIFF
--- a/core/src/main/scala-2.12/cats/instances/all.scala
+++ b/core/src/main/scala-2.12/cats/instances/all.scala
@@ -10,7 +10,6 @@ abstract class AllInstancesBinCompat
     with AllInstancesBinCompat4
     with AllInstancesBinCompat5
     with AllInstancesBinCompat6
-    with AllInstancesBinCompat7
 
 trait AllInstances
     extends AnyValInstances
@@ -44,6 +43,7 @@ trait AllInstances
     with TupleInstances
     with UUIDInstances
     with VectorInstances
+    with PartialFunctionInstances
 
 trait AllInstancesBinCompat0 extends FunctionInstancesBinCompat0 with Tuple2InstancesBinCompat0
 
@@ -64,5 +64,3 @@ trait AllInstancesBinCompat4 extends SortedMapInstancesBinCompat1 with MapInstan
 trait AllInstancesBinCompat5 extends SortedSetInstancesBinCompat0
 
 trait AllInstancesBinCompat6 extends SortedSetInstancesBinCompat1 with SortedMapInstancesBinCompat2
-
-trait AllInstancesBinCompat7 extends PartialFunctionInstances
--- a/core/src/main/scala-2.12/cats/instances/all.scala
+++ b/core/src/main/scala-2.12/cats/instances/all.scala
@@ -10,6 +10,7 @@ abstract class AllInstancesBinCompat
     with AllInstancesBinCompat4
     with AllInstancesBinCompat5
     with AllInstancesBinCompat6
+    with AllInstancesBinCompat7
 
 trait AllInstances
     extends AnyValInstances
@@ -63,3 +64,5 @@ trait AllInstancesBinCompat4 extends SortedMapInstancesBinCompat1 with MapInstan
 trait AllInstancesBinCompat5 extends SortedSetInstancesBinCompat0
 
 trait AllInstancesBinCompat6 extends SortedSetInstancesBinCompat1 with SortedMapInstancesBinCompat2
+
+trait AllInstancesBinCompat7 extends PartialFunctionInstances
--- a/core/src/main/scala/cats/Applicative.scala
+++ b/core/src/main/scala/cats/Applicative.scala
@@ -202,7 +202,7 @@ object Applicative {
    * scala> import cats.Applicative.catsApplicativeForArrow
    * scala> val toLong: Int => Long = _.toLong
    * scala> val double: Int => Int = 2*_
-   * scala> val f: Int => (Long, Int) = catsApplicativeForArrow.product(toLong, double)
+   * scala> val f: Int => (Long, Int) = catsApplicativeForArrow[Function1, Int].product(toLong, double)
    * scala> f(3)
    * res0: (Long, Int) = (3,6)
    * }}}
--- a/core/src/main/scala/cats/arrow/Arrow.scala
+++ b/core/src/main/scala/cats/arrow/Arrow.scala
@@ -74,9 +74,6 @@ import scala.annotation.implicitNotFound
 
 object Arrow {
 
-  implicit val partialFunction: Arrow[PartialFunction] =
-    cats.instances.partialFunction.catsStdInstancesForArrowHierarchy
-
   /****************************************************************************/
   /* THE FOLLOWING CODE IS MANAGED BY SIMULACRUM; PLEASE DO NOT EDIT!!!!      */
   /****************************************************************************/
--- a/core/src/main/scala/cats/arrow/Arrow.scala
+++ b/core/src/main/scala/cats/arrow/Arrow.scala
@@ -74,6 +74,9 @@ import scala.annotation.implicitNotFound
 
 object Arrow {
 
+  implicit val partialFunction: Arrow[PartialFunction] =
+    cats.instances.partialFunction.catsStdInstancesForArrowHierarchy
+
   /****************************************************************************/
   /* THE FOLLOWING CODE IS MANAGED BY SIMULACRUM; PLEASE DO NOT EDIT!!!!      */
   /****************************************************************************/
--- a/core/src/main/scala/cats/arrow/Compose.scala
+++ b/core/src/main/scala/cats/arrow/Compose.scala
@@ -41,11 +41,11 @@ import scala.annotation.implicitNotFound
 }
 
 object Compose {
-  implicit val catsInstancesForFunction1: ArrowChoice[Function1] with CommutativeArrow[Function1] =
+  implicit def catsInstancesForFunction1: ArrowChoice[Function1] with CommutativeArrow[Function1] =
     cats.instances.function.catsStdInstancesForFunction1
-  implicit val catsComposeForMap: Compose[Map] = cats.instances.map.catsStdComposeForMap
+  implicit def catsComposeForMap: Compose[Map] = cats.instances.map.catsStdComposeForMap
 
-  implicit val catsInstancesForPartialFunction: ArrowChoice[PartialFunction] with CommutativeArrow[PartialFunction] =
+  implicit def catsInstancesForPartialFunction: ArrowChoice[PartialFunction] with CommutativeArrow[PartialFunction] =
     cats.instances.partialFunction.catsStdInstancesForPartialFunction
 
   /****************************************************************************/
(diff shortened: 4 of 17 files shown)


core/src/main/scala/cats/instances/partialFunction.scala
```diff
@@ -0,0 +1,27 @@
...
+     * scala> val f: Int => Double = x => x.toDouble / 2
+     * scala> val g: Double => Double = x => x * 3
+     * scala> val h = fab.dimap(f)(g)
+     * scala> h(3)
+     * res0: Double = 5.4
+     * }}}
+     */
+    override def dimap[A, B, C, D](fab: PartialFunction[A, B])(f: C => A)(g: B => D): PartialFunction[C, D] =
+      new PartialFunction[C, D] {
+        override def isDefinedAt(c: C): Boolean = fab.isDefinedAt(f(c))
```
gagandeepkalra, line 23: I don't think if there's any way out of running `f` twice

LukaJCB:
I wonder if we could add instances for the other parts of the `arrow` hierarchy. Seems like we could implement pretty much everything, no?

core/src/main/scala/cats/Applicative.scala
```diff
@@ -200,7 +200,7 @@ object Applicative {
    * scala> import cats.Applicative.catsApplicativeForArrow
    * scala> val toLong: Int => Long = _.toLong
    * scala> val double: Int => Int = 2*_
-   * scala> val f: Int => (Long, Int) = catsApplicativeForArrow.product(toLong, double)
+   * scala> val f: Int => (Long, Int) = catsApplicativeForArrow[Function1, Int].product(toLong, double)
```
gagandeepkalra, line 203: since `PartialFunction` is subtype of `Function1` there was implicit ambiguity here

core/src/main/scala/cats/arrow/Arrow.scala
```diff
@@ -69,3 +69,8 @@ import simulacrum.typeclass
   def merge[A, B, C](f: F[A, B], g: F[A, C]): F[A, (B, C)] =
     andThen(lift((x: A) => (x, x)), split(f, g))
 }
+
+object Arrow {
+  implicit val partialFunction: Arrow[PartialFunction] =
+    cats.instances.partialFunction.catsStdInstancesForArrowHierarchy
```
gagandeepkalra, line 75: keeping up with this new practice of having StdLib's instances in the companion object itself

core/src/main/scala/cats/instances/partialFunction.scala
```diff
@@ -0,0 +1,60 @@
...
+
+      override def split[A, B, C, D](
+        f: PartialFunction[A, B],
+        g: PartialFunction[C, D]
+      ): PartialFunction[(A, C), (B, D)] = {
+        case (a, c) if f.isDefinedAt(a) && g.isDefinedAt(c) => (f(a), g(c))
+      }
+
+      override def compose[A, B, C](f: PartialFunction[B, C], g: PartialFunction[A, B]): PartialFunction[A, C] = {
+        case a if g.isDefinedAt(a) && f.isDefinedAt(g(a)) => f(g(a))
```
gagandeepkalra, line 57: guess there's no better way

gagandeepkalra:
@LukaJCB I updated the same PR instead, is it okay?

core/src/main/scala/cats/arrow/Arrow.scala
barambani: It would be probably enough to have the `ArrowChoice[PartialFunction]` and `CommutativeArrow[PartialFunction]` instances in the `Compose` companion object and the `Strong[PartialFunction]` instance in `Profunctor`'s. Something like
```scala
implicit val catsInstancesForPartialFunction: ArrowChoice[PartialFunction] with CommutativeArrow[PartialFunction] =
    cats.instances.partialFunction.catsStdInstancesForArrowHierarchy
```
and
```scala
implicit val catsStrongForPartialFunction: Strong[PartialFunction] =
    cats.instances.partialFunction.catsStdInstancesForArrowHierarchy
```

core/src/main/scala/cats/instances/partialFunction.scala
```diff
@@ -0,0 +1,60 @@
+package cats.instances
+import cats.arrow.{ArrowChoice, CommutativeArrow}
+
+trait PartialFunctionInstances {
+
+  implicit val catsStdInstancesForArrowHierarchy: ArrowChoice[PartialFunction] with CommutativeArrow[PartialFunction] =
```
barambani, line 6:
```suggestion
  implicit val catsStdInstancesForPartialFunction: ArrowChoice[PartialFunction] with CommutativeArrow[PartialFunction] =
```

core/src/main/scala/cats/instances/partialFunction.scala
```diff
@@ -0,0 +1,60 @@
...
+       * }}}
+       */
+      override def choose[A, B, C, D](f: PartialFunction[A, C])(
+        g: PartialFunction[B, D]
+      ): PartialFunction[Either[A, B], Either[C, D]] = {
+        case Left(a) if f.isDefinedAt(a)  => Left(f(a))
+        case Right(b) if g.isDefinedAt(b) => Right(g(b))
+      }
+
+      override def lift[A, B](f: A => B): PartialFunction[A, B] = { case a if a.isInstanceOf[A] => f(a) }
```
barambani, line 29: Could this be ?
```suggestion
      override def lift[A, B](f: A => B): PartialFunction[A, B] = { case a => f(a) }
```

core/src/main/scala/cats/arrow/Arrow.scala
gagandeepkalra: I see thank you @barambani, had to revisit implicit resolution rules again :)

core/src/main/scala/cats/instances/partialFunction.scala
gagandeepkalra: yes, this works. 

I originally tried `{ case a: A => f(a) }` which led to this type erasure error-

> [error] /Users/gagandeepkalra/IdeaProjects/cats/core/src/main/scala/cats/instances/partialFunction.scala:29:77: abstract type pattern A is unchecked since it is eliminated by erasure
[error]       override def lift[A, B](f: A => B): PartialFunction[A, B] = { case a: A => f(a) }
[error]    

As I understand adding `: A` introduces another type check (which is not possible because of erasure), but the function `{ case a => f(a) }` in itself type checks at compile team no runtime check needed?

gagandeepkalra:
Thank you @LukaJCB @barambani for the review, learnt something new today.

barambani:
Thank you.
> learnt something new today.

same here.

core/src/main/scala/cats/arrow/Arrow.scala
barambani: It might be that you already know this but, in case you don't, this approach is brilliantly explained [here](https://meta.plasm.us/posts/2019/09/30/implicit-scope-and-cats/)

barambani:
@LukaJCB I suggested it to align to the approach taken here https://github.com/typelevel/cats/pull/3043

LukaJCB:
Ah of course! Sorry I misremembered a detail of the whole thing

core/src/main/scala/cats/arrow/Compose.scala
```diff
@@ -41,9 +41,12 @@ import scala.annotation.implicitNotFound
 }
 
 object Compose {
-  implicit def catsInstancesForFunction1: ArrowChoice[Function1] with CommutativeArrow[Function1] =
+  implicit val catsInstancesForFunction1: ArrowChoice[Function1] with CommutativeArrow[Function1] =
```
travisbrown, line 44: It's not really a big deal, but there is a reason for these to be methods instead of `val`s: they're already instantiated as `val`s in the `instances` objects, so using `def` here won't result in new instantiations, but using a `val` does result in a little extra overhead and indirection, since it generates both an accessor method and a field.

In general this is unlikely to matter much, but because there are a lot of these, we've had a policy of `def` instead of `val`.

core/src/main/scala-2.12/cats/instances/all.scala
```diff
@@ -10,6 +10,7 @@ abstract class AllInstancesBinCompat
     with AllInstancesBinCompat4
     with AllInstancesBinCompat5
     with AllInstancesBinCompat6
+    with AllInstancesBinCompat7
```
travisbrown, line 13: We shouldn't need any new `BinCompat` traits now that we've dropped 2.11. You should be able to add `PartialFunctionInstances directly to `AllInstances`, I believe?

core/src/main/scala/cats/arrow/Compose.scala
gagandeepkalra: I see, that's a nice way of saving up on a reference; apologies for oversight

core/src/main/scala-2.12/cats/instances/all.scala
gagandeepkalra: I'm not sure, getting this error when adding to `AllInstances`; I don't see anything obvious, might need some help

> [error]  * abstract synthetic method cats$instances$PartialFunctionInstances$_setter_$catsStdInstancesForPartialFunction_=(cats.arrow.ArrowChoice)Unit in interface cats.instances.PartialFunctionInstances is inherited by class AllInstances in current version.
[error]    filter with: ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("cats.instances.AllInstances.cats$instances$PartialFunctionInstances$_setter_$catsStdInstancesForPartialFunction_=")
[error]  * abstract method catsStdInstancesForPartialFunction()cats.arrow.ArrowChoice in interface cats.instances.PartialFunctionInstances is inherited by class AllInstances in current version.
[error]    filter with: ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("cats.instances.AllInstances.catsStdInstancesForPartialFunction")

travisbrown: @gagandeepkalra Hmm, thanks, I'll take a look now.

travisbrown:
Oh, right. The issue is that adding a `val` to a trait still breaks bincompat even on 2.12+, because of synthetic methods. We've had to work around this in a couple of other places—let me find a link…

travisbrown:
Here's some explanation of the issue: https://github.com/typelevel/cats/pull/2789#issuecomment-529306146

I think that something like this should work:

```scala
trait PartialFunctionInstances {
  def catsStdInstancesForPartialFunction: ArrowChoice[PartialFunction] with CommutativeArrow[PartialFunction] =
    PartialFunctionInstances.instance
}

private object PartialFunctionInstances {
  private val instance: ArrowChoice[PartialFunction] with CommutativeArrow[PartialFunction] =
    new ArrowChoice[PartialFunction] with CommutativeArrow[PartialFunction] {
      // ...
```
I'd be happy to merge this PR as-is, and I can fix this particular part in a follow-up, if you'd prefer?

gagandeepkalra:
@travisbrown thank you, summarising-

I understand why adding a def to trait won't break bc given we have method defaults in java 8 interface, while though `val` would still require each class this trait is mixed in with, recompilation.

Meanwhile I had another doc-test import related improvement, so have pushed both together.

travisbrown:
Going ahead and merging this since it had an earlier second :+1:.

PR #3397, opened 2020-04-19
Remove Order constraints from SortedMap and NonEmptyMap instances

joroKr21:
`SortedMap` carries around its own instance of `Ordering` so there is no need for an implicit `Order` constraint unless we have to create a new `SortedMap` from scratch (e.g. in `Monoid.empty`).

Some of the removed constraints were simply not used (e.g. `Show`, `Hash`), others were used only in some methods but not others (e.g. `Traverse`) - note that the standard lib `++` does *not* require an `Ordering`.

These inconsistencies are accidental at best and can lead to surprising behaviour at worst (when there are incoherent `Order` instances in a project).

Closes #3223 
Closes #2318


joroKr21:
@travisbrown I think we should make a decision on this one before 2.2.0 because of bincompat - it modifies directly the new implicit scope instances.

core/src/main/scala/cats/data/NonEmptyMapImpl.scala
```diff
@@ -14,13 +13,21 @@ private[data] object NonEmptyMapImpl extends NonEmptyMapInstances with Newtype2
   private[data] def unwrap[K, A](m: Type[K, A]): SortedMap[K, A] =
     m.asInstanceOf[SortedMap[K, A]]
 
-  def fromMap[K: Order, A](as: SortedMap[K, A]): Option[NonEmptyMap[K, A]] =
+  def fromMap[K, A](as: SortedMap[K, A]): Option[NonEmptyMap[K, A]] =
     if (as.nonEmpty) Option(create(as)) else None
 
-  def fromMapUnsafe[K: Order, A](m: SortedMap[K, A]): NonEmptyMap[K, A] =
+  @deprecated("Use fromMap override without Order", "2.2.0-M2")
```
barambani, line 19: super nitpick  😄 , now that M2 is out
```suggestion
  @deprecated("Use fromMap override without Order", "2.2.0-M3")
```
or if it doesn't make it for the M3 RC1 may be. Here and the others

travisbrown:
The conflict seems pretty simple and I think doesn't require reformatting or anything, so I'll just resolve it here through the web interface if there are no objections (and we can update the `since` arguments to M3 in a quick follow-up).

travisbrown:
Oh, looks like I was wrong and it does need a format. @joroKr21, sorry about that—I can grab your commits and open a new PR with them, or you could pull from here and push the format commit?

joroKr21:
Fine by me 👍

joroKr21:
Superseded by #3473

PR #3401, opened 2020-04-24
Replace FunctionK.lift macro

joroKr21:


kailuowang:
Referencing https://github.com/typelevel/cats/issues/2553. 
lots of discussions there regarding this macro in case you are not aware of.

joroKr21:
We can have both

joroKr21:
Couldn't reopen this one so: #3402

PR #3402, opened 2020-04-25
Simplify FunctionK.lift macro

### LINKED ISSUE #3400: Macro Error in FunctionK.lift
Hi,
I was wrapping doobie's liftIO in FunctionK:

```scala
val lifter: FunctionK[IO, ConnectionIO] = FunctionK.lift(Async[ConnectionIO].liftIO)
```

and got

```scala
Error:(24, 80) Unexpected tree {
  <synthetic> val eta$0$1: cats.effect.Async[doobie.free.connection.ConnectionIO] = cats.effect.Async.apply[doobie.free.connection.ConnectionIO](free.this.connection.AsyncConnectionIO);
  ((ioa: cats.effect.IO[Nothing]) => eta$0$1.liftIO[Nothing](ioa))
} when lifting to FunctionK
  val lifter: FunctionK[IO, ConnectionIO] = FunctionK.lift(Async[ConnectionIO].liftIO)
```

This fixed it:

```scala
val acio = Async[ConnectionIO]
val lifter: FunctionK[IO, ConnectionIO] = FunctionK.lift(acio.liftIO)
```

@milessabin suggested the macro might have been confused about whether it was supposed to be looking at the Async.apply or the liftIO.

joroKr21:
Add `FunctionK.liftFunction` that uses an abstract type member
to emulate a polymorphic function type.

Because `liftFunction` has worse type inference,
keep the macro based `lift` that now delegates to `liftFunction`,
instead of creating a new `FunctionK` instance every time.
Also accept eta-expanded functions in the macro implementation.

Fixes #3400


core/src/main/scala/cats/arrow/FunctionK.scala
```diff
@@ -79,85 +80,70 @@ object FunctionK {
...
+   *
+   * {{{
+   *   def headOption[A](list: List[A]): Option[A] = list.headOption
+   *   val lifted = FunctionK.liftFunction[List, Option](headOption)
+   * }}}
+   *
+   * Note: The weird `τ[F, G]` parameter is there to compensate for
+   * the lack of polymorphic function types in Scala 2.
+   */
+  def liftFunction[F[_], G[_]](f: F[τ[F, G]] => G[τ[F, G]]): FunctionK[F, G] = new FunctionK[F, G] {
```
djspiewak, line 106: Are we entirely sure this is sound? I know that similar tricks involving a `*`-kinded tau are not, but I find it difficult to reason about scalac's behavior in these cases.

joroKr21: I took it from the discussion in #2553. We could also call it `unsafeLift` but I guess it would have to be public because it's called in the expanded macro.

joroKr21: AFAIU because `F` and `G` are type parameters of `τ` you cannot reuse it with other `FunctionK`s which would be a way to exploit unsoundness if we had a bare `τ` without type parameters.

tests/src/test/scala/cats/tests/FunctionKSuite.scala
```diff
@@ -92,12 +92,27 @@ class FunctionKSuite extends CatsSuite {
...
+  test("lift eta-expanded function") {
+    val fSomeNel = FunctionK.lift[NonEmptyList, OptionOfNel](Applicative[Option].pure)
+    forAll { (a: NonEmptyList[Int]) =>
+      fSomeNel(a) should ===(Some(a))
+    }
+  }
+
+  test("lift a function directly") {
+    def headOption[A](list: List[A]): Option[A] = list.headOption
+    val fHeadOption = FunctionK.liftFunction[List, Option](headOption)
```
barambani, line 110: A question for my understanding, could this be
 ```suggestion
     val fHeadOption = FunctionK.lift[List, Option](headOption)
```
Isn't it forwarding to `liftFunction` anyway ? Or it's testing the public `liftFunction` explicitly on purpose ?
Apologies if this doesn't make sense :).

joroKr21: It's testing `liftFunction` explicitly. Do you think we should also try to add a dotty version? It has native polymorphic functions, but we don't provide a way to convert them to `FunctionK` and I think type inference is still not great.

barambani: It makes sense, thanks for clarifying. About the dotty version, I guess it's related to https://github.com/typelevel/cats/issues/2553 that's still an open issue for what I understand and specifically there is work going on as note here https://github.com/typelevel/cats/issues/2553#issuecomment-561076000. We could sync a dotty specific version with that effort.
So what do you thing (you and others) about letting this PR move on as it improves the current situation described here https://github.com/typelevel/cats/issues/3400 and might make some user's life easier ? My approval goes in that direction. Of course I happy to do otherwise if the general preference is different.

joroKr21: Yes I agree this can be merged and I can try to add a dotty version in another PR 👍

barambani: Brilliant, thank 🙇

djspiewak:
> Thanks for this. Hopefully we can find a dotty solution soon as well

FYI I actually have a Dotty solution already, but it doesn't compile because higher-rank type inference isn't fully implemented. @smarter has mentioned that this is something he wants to do, but he hasn't had the time.

joroKr21:
> FYI I actually have a Dotty solution already, but it doesn't compile because higher-rank type inference isn't fully implemented.

Yes, that's what I meant, type inference doesn't work. I'm not sure if this can be solved with a macro as in Scala 2 but we can at least have `lift` for Dotty that takes a polymorphic function.

barambani:
@djspiewak , sorry for the ping, this PR has two approvals so if you're happy with the changes and the conflicts are resolved I would merge it in.

barambani (dismissed):
👍

larsrh (approved):

PR #3403, opened 2020-04-25
Add Cont alias to ContT

RaasAhsan:
Let me know if I need tests or documentation of any kind.

### DIFF
--- a/core/src/main/scala/cats/data/package.scala
+++ b/core/src/main/scala/cats/data/package.scala
@@ -82,4 +82,17 @@ package object data extends ScalaVersionSpecificPackage {
     def apply[S, A](f: S => A, s: S): Store[S, A] =
       RepresentableStore[S => *, S, A](f, s)
   }
+
+  type Cont[A, B] = ContT[Id, A, B]
+
+  object Cont {
+    def apply[A, B](f: (B => A) => A): Cont[A, B] =
+      ContT[Id, A, B](f)
+
+    def pure[A, B](b: B): Cont[A, B] =
+      ContT.pure[Id, A, B](b)
+
+    def defer[A, B](b: => B): Cont[A, B] =
+      ContT.defer[Id, A, B](b)
+  }
 }
--- a/core/src/main/scala/cats/data/package.scala
+++ b/core/src/main/scala/cats/data/package.scala
@@ -83,16 +83,22 @@ package object data extends ScalaVersionSpecificPackage {
       RepresentableStore[S => *, S, A](f, s)
   }
 
-  type Cont[A, B] = ContT[Id, A, B]
+  type Cont[A, B] = ContT[Eval, A, B]
 
   object Cont {
-    def apply[A, B](f: (B => A) => A): Cont[A, B] =
-      ContT[Id, A, B](f)
+    def apply[A, B](f: (B => Eval[A]) => Eval[A]): Cont[A, B] =
+      ContT[Eval, A, B](f)
 
     def pure[A, B](b: B): Cont[A, B] =
-      ContT.pure[Id, A, B](b)
+      ContT.pure[Eval, A, B](b)
 
     def defer[A, B](b: => B): Cont[A, B] =
-      ContT.defer[Id, A, B](b)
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/data/package.scala
```diff
@@ -82,4 +82,17 @@ package object data extends ScalaVersionSpecificPackage {
     def apply[S, A](f: S => A, s: S): Store[S, A] =
       RepresentableStore[S => *, S, A](f, s)
   }
+
+  type Cont[A, B] = ContT[Id, A, B]
```
djspiewak, line 86: I think this needs to be `Eval` instead of `Id` in order to preserve stack safety. But otherwise, big 👍 on this!

RaasAhsan: So I tried to verify the stack safety of `ContT[Eval, A, B]`:
```scala
def add[R](a: Int, b: Int): ContT[Eval, Int, Int] =
    ContT(f => f(a + b))
val seed = add[Int](3, 2)
val f = (0 until 10000).foldLeft(seed)((acc, _) => acc.flatMap(x => Cont.pure(x + 1)))
println(f.run(x => Eval.later(x)).value)
```
But I seem to be getting stack overflows for this as well. Excerpt of the stack trace:
```scala
[error] 	at cats.data.AndThen.runLoop(AndThen.scala:97)
[error] 	at cats.data.AndThen.apply(AndThen.scala:67)
[error] 	at cats.data.ContT$.$anonfun$defer$1(ContT.scala:133)
[error] 	at cats.data.AndThen.runLoop(AndThen.scala:97)
[error] 	at cats.data.AndThen.apply(AndThen.scala:67)
[error] 	at cats.data.ContT.$anonfun$flatMap$2(ContT.scala:43)
[error] 	at scala.Function1.$anonfun$andThen$1(Function1.scala:57)
[error] 	at cats.data.AndThen.runLoop(AndThen.scala:97)
[error] 	at cats.data.AndThen.apply(AndThen.scala:67)
[error] 	at cats.data.ContT$.$anonfun$defer$1(ContT.scala:133)
```
Do you think there might be a deficiency in the `ContT#flatMap` implementation here?

djspiewak: Finally got a chance to poke at this a bit. Have you tried the same test with `tailRecM`? I haven't looked into extreme detail about what's going on here, but I remember `tailRecM` being very meticulously verified as stack-safe, but I don't think the same work was done on `flatMap` (in fact, I remember specifically making arguments that you *can't* make a stack-safe `flatMap` for `ContT`).

RaasAhsan: Yeah, no more stack overflows with `tailRecM`. I'll need to convince myself why it won't work for `flatMap`, but otherwise I think this is all ready. :) I noticed we don't have any inductive class instances for `ContT` in `cats-effect`, might take a look and see if those can be added

djspiewak: I'd like to add some, but I think the problem is the only one we could add in CE2 is `Bracket`. All `Sync` instances are constrained to be stack-safe in their `flatMap`, which obviously `ContT` isn't.

As an aside, the main reason why `flatMap` on `ContT` isn't stack-safe is similar to the reason why `flatMap` on `Option` isn't stack-safe: there's nothing to trampoline through. With `tailRecM`, you're forced to pass back to the trampoline loop with each iteration, while `flatMap` has no such guarantee. With `Option`, you need to have a recursive `flatMap` in order to really see the damage. With `ContT`, a long chain of `flatMap`s *is* recursive because it's effectively just a very long chain of `andThen`s. Since `flatMap` on `ContT` corresponds directly to the CPS transformation, we can see it more directly in imperative form:

```scala
def direct(i: Int): String = i.toString

def cps[A](i: Int)(k: String => A): A = k(i.toString)
```

Now imagine chaining 10,000 `cps`es together. Naturally, the result will be a stack overflow.

It's worth noting that even `IO` would have this issue if someone worked entirely in terms of `async`.

PR #3404, opened 2020-04-25
Add NonEmptyReducible to type signatures

zainab-ali:
This PR exposes the NonEmptyReducible typeclass instances.

Currently, the instances aren't exposed and can't be summoned.  The following code does not compile:
```scala
import cats.implicits._
import cats.data._

object Test {
  implicitly[NonEmptyReducible[NonEmptyVector, Vector]]
}
```
I assume this isn't the behaviour we want.

I noticed that there are no `NonEmptyReducibleTests`.  Let me know if you would like these added.


zainab-ali:
As @LukaJCB  pointed out, this isn't actually a typeclass.  It doesn't really need to be exposed.

PR #3423, opened 2020-05-20
Add toNev to NonEmptyList variants

### LINKED ISSUE #3422: Add toNev combinator in NonEmptyList
I wanted to get an instance of NonEmptyVector from a NonEmptyList and I noticed it was lacking that combinator.

Something like:

```
def toNev[B >: A](implicit order: Order[B]): NonEmptyVector[B] = 
    NonEmptyVector.fromVectorUnsafe(toList.toVector)
```

There might be more combinators missing.

saraiva132:
Closes #3422

### DIFF
--- a/core/src/main/scala-2.13+/cats/data/NonEmptyLazyList.scala
+++ b/core/src/main/scala-2.13+/cats/data/NonEmptyLazyList.scala
@@ -458,7 +458,7 @@ class NonEmptyLazyListOps[A](private val value: NonEmptyLazyList[A])
    *}}}
    */
   final def toNev[B >: A]: NonEmptyVector[B] =
-    NonEmptyVector.fromVectorUnsafe(toLazyList.toList.toVector)
+    NonEmptyVector.fromVectorUnsafe(toLazyList.toVector)
 
   final def show[AA >: A](implicit AA: Show[AA]): String = s"NonEmpty${Show[LazyList[AA]].show(toLazyList)}"
 }
--- a/core/src/main/scala-2.13+/cats/data/NonEmptyLazyList.scala
+++ b/core/src/main/scala-2.13+/cats/data/NonEmptyLazyList.scala
@@ -445,6 +445,21 @@ class NonEmptyLazyListOps[A](private val value: NonEmptyLazyList[A])
   final def toNes[B >: A](implicit order: Order[B]): NonEmptySet[B] =
     NonEmptySet.of(head, tail: _*)
 
+  /**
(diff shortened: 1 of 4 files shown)


core/src/main/scala-2.13+/cats/data/NonEmptyLazyList.scala
```diff
@@ -445,6 +445,21 @@ class NonEmptyLazyListOps[A](private val value: NonEmptyLazyList[A])
...
+   * scala> import cats.instances.int._
+   * scala> val nel = NonEmptyLazyList.fromLazyListPrepend(1, LazyList(2,3,4))
+   * scala> val expectedResult = NonEmptyVector.fromVectorUnsafe(Vector(1,2,3,4))
+   * scala> val result = nel.toNev
+   * scala> result === expectedResult
+   * res0: Boolean = true
+   *}}}
+   */
+  final def toNev[B >: A]: NonEmptyVector[B] =
+    NonEmptyVector.fromVectorUnsafe(toLazyList.toList.toVector)
```
LukaJCB, line 461: does `LazyList` not have its own `toVector` method?

saraiva132: You are right. It does. Additionally, I was missing an import. I will push changes. Thank you @LukaJCB :)

LukaJCB: Of course :)

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -437,6 +437,21 @@ final case class NonEmptyList[+A](head: A, tail: List[A]) extends NonEmptyCollec
...
+   * scala> import cats.data._
+   * scala> import cats.instances.int._
+   * scala> val nel = NonEmptyList(1, List(2,3,4))
+   * scala> val expectedResult = NonEmptyVector.fromVectorUnsafe(Vector(1,2,3,4))
+   * scala> val result = nel.toNev
+   * scala> result === expectedResult
+   * res0: Boolean = true
+   *}}}
+   */
+  def toNev[B >: A]: NonEmptyVector[B] =
```
travisbrown, line 453: Is there a reason this one isn't final but the one above is?

saraiva132: Yes. NonEmptyList is a final class. LazyNonEmptyList is not.

PR #3269, opened 2020-01-24
Dotty cross-building [WIP]

travisbrown:
This is a work-in-progress pull request that should pass CI but is missing or waiting on several things:

- [x] Official Dotty release of ScalaTest.
- [ ] Official Dotty release of the ScalaTest ScalaCheck compatibility module.
- [x] An official decision on [what to do about Simulacrum](https://github.com/typelevel/cats/issues/3192).
- [x] An official decision on [what to do about use of kind-projector's polymorphic lambda value syntax](https://github.com/typelevel/cats/pull/3193).
- [x] Also I just haven't bothered to migrate cats-free (which uses existential types) yet.

Note that while this PR looks like a big change—currently `+2,662 -297`—over 2,400 of the added lines have been automatically generated by [Scalafix rules](https://github.com/travisbrown/simulacrum-scalafix), and another hundred or so are just moving the `FunctionK.lift` macro into version-specific code. There aren't much more than a hundred lines of substantive changes.

I've been opening Dotty bug reports for issues that I've had to work around. Some of these have already been fixed, and it's likely that more will be resolved by the time this PR is ready to merge:

- [x] https://github.com/lampepfl/dotty/issues/7876
- [x] https://github.com/lampepfl/dotty/issues/7926
- [x] https://github.com/lampepfl/dotty/issues/7993
- [ ] https://github.com/lampepfl/dotty/issues/7999
- [ ] https://github.com/lampepfl/dotty/issues/8001
- [x] https://github.com/lampepfl/dotty/issues/8033 (but still needs workaround since this was merged since 0.22.0-RC1)
- [x] https://github.com/lampepfl/dotty/issues/8035
- [x] https://github.com/lampepfl/dotty/issues/8049
- [ ] https://github.com/lampepfl/dotty/issues/8383 (not _necessary_ but it means we have to disable fatal warnings on Dotty)

I'm opening this for tracking and discussion [now that `-Ykind-projector` is in the Dotty nightly releases](https://github.com/lampepfl/dotty/pull/7775), but we probably won't be in a position to merge it for a while, and I'll be rebasing and force-pushing to keep it updated.

I have a similar [branch for cats-effect](https://github.com/travisbrown/cats-effect/tree/topic/dotty-cross-building), but haven't opened a PR there yet (update: [WIP cats-effect PR is open](https://github.com/typelevel/cats-effect/pull/762)).

PR #3424, opened 2020-05-22
Use Simulacrum Scalafix

travisbrown:
Addresses #3192, using a [new release](https://github.com/typelevel/simulacrum-scalafix/releases/tag/v0.2.0) of [Simulacrum Scalafix](https://github.com/typelevel/simulacrum-scalafix). An earlier version of this change was included in #3269, but I think it should be reviewed and merged separately from the Dotty cross-build work (we'll need to decide which Dotty version to support, since for example there's still no official 0.23 or 0.24 release of the scalatestplus dependency, but I don't think that should block this change).

The Scalafix commit was generated by running the following commands:

```
sbt:cats> scalafix AddSerializable
sbt:cats> scalafix AddImplicitNotFound
sbt:cats> scalafix TypeClassSupport
sbt:cats> scalafmtAll
```

These commands will need to be run manually after any API change that needs syntax support. I haven't added this process to the `validateJVM` task yet, but they're idempotent and only take a few seconds, so I guess we could run them on every build.

### DIFF
--- a/alleycats-core/src/main/scala/alleycats/ConsK.scala
+++ b/alleycats-core/src/main/scala/alleycats/ConsK.scala
@@ -2,8 +2,10 @@ package alleycats
 
 import cats.SemigroupK
 import simulacrum.typeclass
+import scala.annotation.implicitNotFound
 
-@typeclass trait ConsK[F[_]] {
+@implicitNotFound("Could not find an instance of ConsK for ${F}")
+@typeclass trait ConsK[F[_]] extends Serializable {
   def cons[A](hd: A, tl: F[A]): F[A]
 }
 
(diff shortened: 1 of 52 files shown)


travisbrown:
@kailuowang I think API changes that map over to codegen changes will be infrequent enough that we can run it manually for now. If it turns out that that's not the case and we're getting a lot of PRs with missing codegen changes, we could tie these tasks into `validateJVM` or `prePR`, so that they'd be run when people build locally.

I guess we could also have something more automated in CI, but that seems unnecessary for something that's likely to happen once every month or two.

travisbrown:
Thanks all! I'm merging this with a merge commit to keep the automated changes separate from the configuration that sets them up. I'll update the Dotty cross-building branch later today or tomorrow.

PR #3426, opened 2020-05-24
Add EitherT and IorT constructors from Option with monadic default

ivan-klass:
In practice the `OptionT[F, A]` => `EitherT[F, E, A]` transformation may also require the left to be calculated as `F[E]`. I didn't find anything handy neither in `OptionT` nor `EitherT` having to work-around with `.leftSemiflatMap(identity)` like
```scala
OptionT(...) 
  .toRight(fe) // EitherT[F, F[E], A]
  .leftSemiflatMap(identity)
```
or 
```scala
EitherT.fromOptionF(..., fe) // EitherT[F, F[E], A]
  .leftSemiflatMap(identity)
```

### DIFF
--- a/core/src/main/scala/cats/data/EitherT.scala
+++ b/core/src/main/scala/cats/data/EitherT.scala
@@ -833,6 +833,15 @@ object EitherT extends EitherTInstances {
       }
     )
 
+  /** Similar to `fromOptionF` but the left is carried from monadic `F[_]` context when the option is `None` */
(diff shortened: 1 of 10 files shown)


kailuowang:
`Option`, like several other data types, already has a `liftToF` that can "lift" it into any `F[_]` with an `ApplicativeError[F]`, which is the case for both `IorT` and `EitherT`.
For lifting to the left side, I believe we had a PR adding a `raiseToF` but somehow I couldn't find it anymore. 
I am thinking that instead of adding n - 1 methods to n datatypes to support transformation between

ivan-klass:
Sure, will re-submit with addressed suggestion

PR #3429, opened 2020-05-24
Make new method final

travisbrown:
This is a minor thing to follow up on #3423, which introduces two new `toNev` methods but only makes one of them final. There are a lot of places where we missed the boat on making things final, and can't now because of bincompat, but for new methods like this I think we should try to do it consistently.


saraiva132:
NonEmptyList is a final class so there is no need for its methods to be final

travisbrown:
@saraiva132 Ah, thanks, I missed that! Agreed, it's not needed here.

saraiva132:
Np @travisbrown

PR #3431, opened 2020-05-25
Merge 2.2.x after 2.2.0-M2 release

travisbrown:
All JVM and Scala.js 0.6 artifacts are on Maven Central now, and Scala.js 1.x is promoted and on its way. Release notes are [here](https://github.com/typelevel/cats/releases/tag/v2.2.0-M2).

Please note that this needs to be a merge commit! If we squash or rebase, then the 2.2.0-M2 tag won't be on master.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -152,6 +152,7 @@ possible:
  * Julien Truffaut
  * Jun Tomioka
  * jurisk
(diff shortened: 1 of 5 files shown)


travisbrown:
Going to merge this with one +1 since it’s somewhat sensitive and the changes are basically mechanical.

PR #3432, opened 2020-05-25
NonEmptyList.groupByNelA

### LINKED ISSUE #3420: Add effectful groupByNelA(A => F[B]): F[SortedMap[B, NonEmptyList[A]]]
Add to `cats.syntax.ListOps[A]`:
```
def groupByNelA[F[_] : Applicative, B](f: A => F[B])(implicit B: Order[B]): F[SortedMap[B, NonEmptyList[A]]]
```

An optimised version of:
```
as.traverse(a => f(a).tupleLeft(a)).map(_.groupByNel(_._2).mapValues(_.map(_._1)))
```

pk044:
Closes #3420

### DIFF
--- a/core/src/main/scala/cats/syntax/list.scala
+++ b/core/src/main/scala/cats/syntax/list.scala
@@ -72,9 +72,10 @@ final class ListOps[A](private val la: List[A]) extends AnyVal {
   def groupByNelA[F[_], B](f: A => F[B])(implicit B: Order[B], F: Applicative[F]): F[SortedMap[B, NonEmptyList[A]]] = {
     implicit val ordering: Ordering[B] = B.toOrdering
 
-    toNel.fold(F.pure(SortedMap.empty[B, NonEmptyList[A]]))(nel => F.map(nel.traverse(a => F.tupleLeft(f(a), a)))(_.groupBy(_._2).mapValues(_.map(_._1))))
+    toNel.fold(F.pure(SortedMap.empty[B, NonEmptyList[A]]))(nel =>
+      F.map(nel.traverse(a => F.tupleLeft(f(a), a)))(_.groupBy(_._2).mapValues(_.map(_._1)))
+    )
   }
-
 
   /** Produces a `NonEmptyList` containing cumulative results of applying the
    * operator going left to right.
--- a/core/src/main/scala/cats/syntax/list.scala
+++ b/core/src/main/scala/cats/syntax/list.scala
@@ -52,6 +52,13 @@ final class ListOps[A](private val la: List[A]) extends AnyVal {
     toNel.fold(SortedMap.empty[B, NonEmptyList[A]])(_.groupBy(f))
   }
 
+  def groupByNelA[F[_], B](f: A => F[B])(implicit B: Order[B], F: Applicative[F]): F[SortedMap[B, NonEmptyList[A]]] = {
+    implicit val ordering: Ordering[B] = B.toOrdering
+
+    toNel.fold(F.pure(SortedMap.empty[B, NonEmptyList[A]]))(nel => F.map(nel.traverse(a => F.tupleLeft(f(a), a)))(_.groupBy(_._2).mapValues(_.map(_._1))))
+  }
+
+
   /** Produces a `NonEmptyList` containing cumulative results of applying the
    * operator going left to right.
    *
--- a/core/src/main/scala/cats/syntax/list.scala
+++ b/core/src/main/scala/cats/syntax/list.scala
@@ -52,6 +52,23 @@ final class ListOps[A](private val la: List[A]) extends AnyVal {
     toNel.fold(SortedMap.empty[B, NonEmptyList[A]])(_.groupBy(f))
   }
 
+  /**
+   * Groups elements inside this `List` according to the `Order` of the keys
+   * produced by the given mapping monadic function.
+   *
+   * {{{
+   * scala> import cats.data.NonEmptyList
+   * scala> import scala.collection.immutable.SortedMap
+   * scala> import cats.implicits._
+   *
+   * scala> val list = List(12, -2, 3, -5)
+   *
+   * scala> val expectedResult = Option(SortedMap(false -> NonEmptyList.of(-2, -5), true -> NonEmptyList.of(12, 3)))
+   *
+   * scala> list.groupByNelA(num => Option(0).map(num >= _)) === expectedResult
+   * res0: Boolean = true
+   * }}}
+   */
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/syntax/list.scala
```diff
@@ -52,6 +52,31 @@ final class ListOps[A](private val la: List[A]) extends AnyVal {
...
+   *
+   * scala> list.groupByNelA(num => Option(0).map(num >= _)) === expectedResult
+   * res0: Boolean = true
+   * }}}
+   */
+  def groupByNelA[F[_], B](f: A => F[B])(implicit B: Order[B], F: Applicative[F]): F[SortedMap[B, NonEmptyList[A]]] = {
+    implicit val ordering: Ordering[B] = B.toOrdering
+
+    toNel.fold(F.pure(SortedMap.empty[B, NonEmptyList[A]]))(nel =>
+      F.map(nel.traverse(a => F.tupleLeft(f(a), a)))(_.groupBy(_._2).mapValues(_.map(_._1)))
```
LukaJCB, line 76: `mapValues` was changed in the 2.13 collection rewrite, we could use its `Functor` instance here instead :)

pk044: thank you for the suggestion! i'll push a fix soon :)

core/src/main/scala/cats/syntax/list.scala
```diff
@@ -52,6 +51,33 @@ final class ListOps[A](private val la: List[A]) extends AnyVal {
...
+   * {{{
+   * scala> import cats.data.NonEmptyList
+   * scala> import scala.collection.immutable.SortedMap
+   * scala> import cats.implicits._
+   *
+   * scala> val list = List(12, -2, 3, -5)
+   *
+   * scala> val expectedResult = Option(SortedMap(false -> NonEmptyList.of(-2, -5), true -> NonEmptyList.of(12, 3)))
+   *
+   * scala> list.groupByNelA(num => Option(0).map(num >= _)) === expectedResult
```
LukaJCB, line 67: btw you don't need to do any explicit assertions here, you could just do 
```scala
scala> list.groupByNelA(num => Option(0).map(num >= _))
res0: Option(SortedMap(false -> NonEmptyList(-2, -5), true -> NonEmptyList(12, 3)))
``` 

which will be a little less noise for the reader, while still asserting that the resulting expression is the same as what you typed. :)

pk044: hmm, this one is pretty interesting
```
[info] > Labels of failing property: 

[info] 'Some(TreeMap(false -> NonEmptyList(-2, -5), true -> NonEmptyList(12, 3)))' is not equal to 'Some(Map(false -> NonEmptyList(-2, -5), true -> NonEmptyList(12, 3)))'
```
this seems to fail only on Scala 2.13.1 🤔 works for me locally and on the rest of Scala versions https://travis-ci.org/github/typelevel/cats/builds/691476028

LukaJCB: It works for you locally on 2.13?

pk044: 2.12.10

core/src/main/scala/cats/syntax/list.scala
```diff
@@ -52,6 +51,33 @@ final class ListOps[A](private val la: List[A]) extends AnyVal {
...
+   * scala> list.groupByNelA(num => Option(0).map(num >= _))
+   * res0: Option[SortedMap[Boolean, NonEmptyList[Int]]] = Some(Map(false -> NonEmptyList(-2, -5), true -> NonEmptyList(12, 3)))
+   * }}}
+   */
+  def groupByNelA[F[_], B](f: A => F[B])(implicit B: Order[B], F: Applicative[F]): F[SortedMap[B, NonEmptyList[A]]] = {
+    implicit val ordering: Ordering[B] = B.toOrdering
+
+    toNel.fold(F.pure(SortedMap.empty[B, NonEmptyList[A]]))(nel =>
+      F.map(nel.traverse(a => F.tupleLeft(f(a), a)))(list =>
+        Functor[SortedMap[B, *]].map(list.groupBy(_._2))(_.map(_._1))
```
travisbrown, line 76: This won't generally be a big deal, I guess, but instantiating this instance repeatedly doesn't seem ideal, and it'd be easy to lift it out.

core/src/main/scala/cats/syntax/list.scala
```diff
@@ -52,6 +51,33 @@ final class ListOps[A](private val la: List[A]) extends AnyVal {
...
+   *
+   * scala> val list = List(12, -2, 3, -5)
+   *
+   * scala> val expectedResult = Option(SortedMap(false -> NonEmptyList.of(-2, -5), true -> NonEmptyList.of(12, 3)))
+   *
+   * scala> list.groupByNelA(num => Option(0).map(num >= _))
+   * res0: Option[SortedMap[Boolean, NonEmptyList[Int]]] = Some(Map(false -> NonEmptyList(-2, -5), true -> NonEmptyList(12, 3)))
+   * }}}
+   */
+  def groupByNelA[F[_], B](f: A => F[B])(implicit B: Order[B], F: Applicative[F]): F[SortedMap[B, NonEmptyList[A]]] = {
```
travisbrown, line 71: Is there a reason to order the implicit parameters this way? There are a couple of rules of thumb that would suggest reversing them (follow the declaration order of the type parameters and put instances that characterize type constructors first).

travisbrown: SIde note: I just took a closer look at this, and Cats is currently inconsistent in this respect—there are a few methods like `WriterT.liftF` that use the `Monoid[L], Applicative[F]` order for no obvious reason, but in general the codebase follows the other order, and I think we should here (unless of course there's a specific reason you did it this way).

pk044: good point, done!

core/src/main/scala/cats/syntax/list.scala
pk044: quite not sure how to approach this - is there anything I can do to work it out or should I just stay with the explicit assertion?

travisbrown: This is an issue with `toString` changing for some collection types between 2.12 and 2.13. Unfortunately I think you have to do something like this: #3155

pk044:
seems like I adjusted the PR to all the suggestions, is there anything I should do more to get this merged? :)

PR #3438, opened 2020-05-28
Change value-level kind-projector syntax

travisbrown:
This is a trivial syntax switch away from kind-projector lambdas that aren't supported in Dotty.

This got merged in #3374—I just started refreshing #3269 and noticed it again. It's the only one that's snuck in, and we might as well go ahead and change it.


travisbrown:
Sorry, @LukaJCB, @fthomas—meant to request reviews, not assign you.

PR #3439, opened 2020-05-28
Make boilerplate syntax classes extend Serializable

ceedubs:
Today I ran into a serializability issue when using `mapN` in a Spark
environment 😬.

### DIFF
--- a/core/src/main/scala/cats/syntax/semigroupal.scala
+++ b/core/src/main/scala/cats/syntax/semigroupal.scala
@@ -11,7 +11,7 @@ trait SemigroupalSyntax {
     }
 }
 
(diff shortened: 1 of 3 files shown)


travisbrown:
Definite :+1: to the boilerplate additions. For `SemigroupalOps`, though, wouldn't it be better to do this at the Simulacrum level, for all `Ops` traits? I've just done that in #3440 via [this Simulacrum Scalafix change](https://github.com/typelevel/simulacrum-scalafix/pull/29).

travisbrown:
By the way if anyone else is annoyed by scrolling past the huge Codecov comment in PRs like this, #3428 fixes that, and just needs one more review to merge.

travisbrown:
Merging this now. I'll include changing back the `SemigroupalOps` addition when I rebase #3440.

PR #3440, opened 2020-05-29
Mark Ops traits as Serializable

travisbrown:
Addresses part of the issue noted by @ceedubs in #3439 more generally. This is a draft PR only because I used a locally-published version of Scalafix Simulacrum to make it. It's ready for review, and can be merged if [this PR](https://github.com/typelevel/simulacrum-scalafix/pull/29) is merged and released as 0.3.0.

Issue #3330, opened 2020-02-28
Deprecate nonInheritedOps and ops objects?


travisbrown:
Cats currently has three ways of importing syntax methods for particular type classes. You can use the subpackage in `syntax` (this is the standard approach to à la carte imports):

```scala
scala> import cats.syntax.functor._
import cats.syntax.functor._

scala> cats.Eval.now(List(0)).widen[Seq[Int]]
res1: cats.Eval[Seq[Int]] = Now(List(0))
```
You could also use the `nonInheritedOps` object in the type class companion, which does exactly the same thing:

```scala
scala> import cats.Functor.nonInheritedOps._
import cats.Functor.nonInheritedOps._

scala> cats.Eval.now(List(0)).widen[Seq[Int]]
res0: cats.Eval[Seq[Int]] = Now(List(0))
```

Or you can use the `ops` object in the companion of any descendant type class:

```scala
scala> import cats.Monad.ops._
import cats.Monad.ops._

scala> cats.Eval.now(List(0)).widen[Seq[Int]]
res0: cats.Eval[Seq[Int]] = Now(List(0))
```
While this last is at least different from the standard `cats.syntax.abc._` approach, and maybe potentially useful, I've never actually seen anyone use it, or recommend it, and it has a serious downside—if you import `ops._` from multiple type class companions, you lose the syntax methods for the intersection of their ancestors:

```scala
scala> import cats.instances.all._, cats.Traverse.ops._, cats.Alternative.ops._
import cats.instances.all._
import cats.Traverse.ops._
import cats.Alternative.ops._

scala> List(1) <+> List(2) <+> List(3)
res0: List[Int] = List(1, 2, 3)

scala> List(Some(1), None, Some(2)).sequence
res1: Option[List[Int]] = None

scala> List(1, 2, 3).fmap(_ + 1)
           ^
       error: type mismatch;
        found   : List[Int]
        required: ?{def fmap: ?}
       Note that implicit conversions are not applicable because they are ambiguous:
        both method toAllTraverseOps in object ops of type [F[_], C](target: F[C])(implicit tc: cats.Traverse[F])cats.Traverse.AllOps[F,C]{type TypeClassType = cats.Traverse[F]}
        and method toAllAlternativeOps in object ops of type [F[_], C](target: F[C])(implicit tc: cats.Alternative[F])cats.Alternative.AllOps[F,C]{type TypeClassType = cats.Alternative[F]}
[...]

PR #3442, opened 2020-05-30
Deprecate Simulacrum ops objects

travisbrown:
Addresses #3330 and depends on #3440 and [this change in Simulacrum Scalafix](https://github.com/typelevel/simulacrum-scalafix/pull/32).

Once this is approved (which shouldn't be a big deal since there are several sign-offs on #3330, and this is a pretty mechanical implementation) I'll publish Simulacrum Scalafix 0.3.0 with the two changes we need for this and #3440, then merge #3440 and rebase and merge this one.

### DIFF
--- a/.scalafix.conf
+++ b/.scalafix.conf
@@ -1,0 +1,10 @@
+rules = [
+  AddSerializable
+  AddImplicitNotFound
+  TypeClassSupport
+]
+
+TypeClassSupport.opsObjectDeprecation = {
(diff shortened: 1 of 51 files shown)


travisbrown:
Okay, this is rebased and updated now that Simulacrum Scalafix 0.3.0 is out and #3340 is merged, so it's ready for review!

travisbrown:
@rossabaker or @mpilquist, mind a quick review since you okayed the idea in #3330?

PR #3446, opened 2020-05-31
Add combineKEval and add short-circuiting for foldMapK

### LINKED ISSUE #2707: Provide a `combineKEval` method
For `Apply` we currently have `map2Eval`, which individual instances can override to get short-circuiting. Would be cool to have the same for `SemigroupK`, as I believe quite a few instances could do some short-circuiting in that way.

LukaJCB:
Will fix #2707


travisbrown:
Squashing since the second commit is just formatting.

PR #3449, opened 2020-06-01
Switch tut-plugin to sbt-mdoc.

takayahilton:
tut is deprecated.
https://github.com/tpolecat/tut#tut-is-deprecated-please-switch-to-mdoc-see-the-migration-guide


travisbrown:
Wow, this actually works? I know several of us have tried this and all ran into problems with differences in resetting, etc. In any case this is really great, thanks!

LukaJCB:
I don't think this is enough unfortunately, you'll need to change this line as well: https://github.com/typelevel/cats/blob/master/build.sbt#L197

travisbrown:
Yes, on a closer look this runs into the issues discussed in #2927 and #3391.

takayahilton:
Thanks,
I will close this PR.
If I can solve the problem I will resend PR.

PR #3455, opened 2020-06-04
Update faq for ammonite

psilospore:
* Partial unification is not a setting for 2.13 so this gives an error
* Bump from cats 1 to 2

### DIFF
--- a/docs/src/main/tut/faq.md
+++ b/docs/src/main/tut/faq.md
@@ -268,7 +268,7 @@ Of course. [sbt-catalysts](https://github.com/typelevel/sbt-catalysts) is create
 
 The easiest way is probably using [Ammonite-REPL](http://ammonite.io/). Install it following the instructions there. Then in the amm console you can type in
 ```scala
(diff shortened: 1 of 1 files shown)


psilospore:
No problem 😃

docs/src/main/tut/faq.md
```diff
@@ -268,8 +268,8 @@ Of course. [sbt-catalysts](https://github.com/typelevel/sbt-catalysts) is create
 
 The easiest way is probably using [Ammonite-REPL](http://ammonite.io/). Install it following the instructions there. Then in the amm console you can type in
 ```scala
-interp.configureCompiler(_.settings.YpartialUnification.value = true)
-import $ivy.`org.typelevel::cats-core:1.0.1`, cats.implicits._
+# interp.configureCompiler(_.settings.YpartialUnification.value = true) # If using scala 2.11 or 2.12
```
djspiewak, line 271: Should this be `//`?

psilospore: Opps wrong language 😅

Issue #2342, opened 2018-07-27
Add `Enum` type class


kubukoz:

Michaelt293:
See https://github.com/typelevel/cats/issues/1536#issuecomment-416027544

johnynek:
I have taken a stab at this before with mixed success:

https://github.com/twitter/algebird/blob/develop/algebird-core/src/main/scala/com/twitter/algebird/Successible.scala

Some challenges:
1. what if you are countable, but not bounded? e.g. BigInteger
2. what if you are upper bounded but not lower bounded?

If you have something unbounded, you want `def next(a: A): A`, but if you are bounded, you want `def next(a: A): Option[A]`.

As for a law, I think it is something like: `a < next(a)` and there is nothing in between: `forAll { (a, b) => if (a < b) (b == next(a) || next(a) < b) else true`

sideeffffect:
this ticket can be closed
cc @LukaJCB @travisbrown

LukaJCB:
good call, thanks @sideeffffect!

PR #3458, opened 2020-06-06
Enum typeclass

yilinwei:
Add the `Enum` typeclass #2342 .

A few comments:

  - There are some instances in `UpperBounded` and `LowerBounded` companion objects; should these be deleted?
 - Should the `partialOrder` be the same field as the `order` just with the bound overridden?

### DIFF
--- a/kernel/src/main/scala/cats/kernel/Enum.scala
+++ b/kernel/src/main/scala/cats/kernel/Enum.scala
@@ -3,7 +3,7 @@ package kernel
 
 import scala.{specialized => sp}
 
-trait PartialNext[@sp A] extends UpperBounded[A] {
+trait PartialNext[@sp A] {
   def partialOrder: PartialOrder[A]
   def partialNext(a: A): Option[A]
 }
@@ -30,8 +30,10 @@ trait UnboundedEnum[@sp A] extends Next[A] with Previous[A] {
 }
 
 trait BoundedEnum[@sp A] extends PartialPrevious[A] with PartialNext[A] with UpperBounded[A] with LowerBounded[A] {
+
   def order: Order[A]
   override def partialOrder: PartialOrder[A] = order
+
 }
 
 trait LowerBoundedEnum[@sp A] extends Next[A] with PartialPrevious[A] with LowerBounded[A] {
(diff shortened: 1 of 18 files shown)


kernel/src/main/scala/cats/kernel/Enum.scala
```diff
@@ -0,0 +1,85 @@
...
+  override def partialOrder: PartialOrder[A] = order
+
+  def cycleNext(a: A): A =
+    partialNext(a).getOrElse(minBound)
+
+  def cyclePrevious(a: A): A =
+    partialPrevious(a).getOrElse(maxBound)
+
+  /** Enumerate the members in ascending order. */
+  def members: List[A] = {
```
travisbrown, line 65: Should this maybe be an `Iterable` or `LazyList` / `Stream`? Having a generic method that returns all members as a list in a context where instances will often be for types like `Int` seems not really ideal.

travisbrown:
@yilinwei Also, to answer your first question: the instances will need to go in all of the type class roots (see the discussion [here](https://meta.plasm.us/posts/2019/09/30/implicit-scope-and-cats/)), and since this is a rare change that adds _new_ type class roots, it will be a little more complicated than usual. I'm happy to take a closer look at that part once the bincompat issues are resolved, if that works for you.

kernel/src/main/scala/cats/kernel/Enum.scala
yilinwei: I agree! I toyed around with `Iterable`, but didn't like the idea because it's not pure iirc, and the idea of having an impure return type irked me. I don't have a strong preference on either lazy types but I think a `NonEmptyLazyList` sounds good? I can then add `membersAscending`/`membersDescending` on the `LowerBounded` and `UpperBounded` `Enum` types.

yilinwei:
@travisbrown 

Thanks for looking at this!

I thought I had ran mima but evidently not correctly. I'll fix the mima issues and ask if I have any particular questions.

Regarding the naming change, I think it makes sense; even more so because we've split it up into `Next`/`Previous`.

yilinwei:
@travisbrown I've fixed the mima issues and addressed the comments; I seem to be having some dotty issues in the CI.

PR #3459, opened 2020-06-07
attemptTap

RaasAhsan:
This is a common enough pattern that I think it deserves its own combinator. I defined it in `MonadErrorSyntax` for now, but could consider moving it to `MonadError`, unsure what the criteria for that is. You could possibly write a more efficient implementation depending on your `F` and I don't think it would break compatibility either.


core/src/main/scala/cats/syntax/monadError.scala
```diff
@@ -32,6 +32,9 @@ final class MonadErrorOps[F[_], E, A](private val fa: F[A]) extends AnyVal {
 
   def redeemWith[B](recover: E => F[B], bind: A => F[B])(implicit F: MonadError[F, E]): F[B] =
     F.redeemWith[A, B](fa)(recover, bind)
+
+  def attemptTap[B](f: Either[E, A] => F[B])(implicit F: MonadError[F, E]): F[A] =
+    F.rethrow(F.flatTap(F.attempt(fa))(f))
```
djspiewak, line 37: I'm not sure exactly what common pattern this is replacing? I guess exactly what the implementation is doing?

RaasAhsan: We already have an analogue for `Future#onSuccess`/`Future#foreach` (`Monad#flatTap`) and `Future#onError` (`ApplicativeError#onError`). This would be the pure analogue to `Future#onComplete`, which is useful for use cases like reporting metrics or logging. This function makes call chains a bit more compact.

```scala
fa
  .attempt
  .flatTap {
    case Right(a) => log.info(s"succeeded with $a")
    case Left(e) => log.error(s"failed with $e")
  }
  .rethrow

// Ironically this one has fewer LOC but harder to factor out the bodies
fa
  .flatTap(a => log.info(s"succeeded with $a"))
  .onError(e => log.error(s"failed with $e"))

fa.attemptTap {
  case Right(a) => log.info(s"succeeded with $a")
  case Left(e) => log.error(s"failed with $e")
}
```

aside: I could see an argument for accepting `f` as a `PartialFunction`

travisbrown:
Also it'd be good to add a usage example test (like in e.g. #3467).

RaasAhsan:
@travisbrown `MonadError` and `ApplicativeError` don't seem to be annotated with `@typeclass`, so no syntax is generated, which I'm guessing is why they're all manually defined in the `cats.syntax.monadError` package. Should they be?

Added in an example, not very happy about the `println` but it demonstrates that some effect happens

LukaJCB:
@RaasAhsan, simulacrum can't handle multi-param typeclasses, that's why we have them manually defined :)

core/src/main/scala/cats/MonadError.scala
```diff
@@ -77,6 +77,33 @@ trait MonadError[F[_], E] extends ApplicativeError[F, E] with Monad[F] {
...
+   * Note that if the effect returned by `f` fails, the resulting effect will fail too.
+   *
+   * Alias for `fa.attempt.flatTap(f).rethrow` for convenience.
+   *
+   * Example:
+   * {{{
+   * scala> import cats.implicits._
+   * scala> import scala.util.{Try, Success, Failure}
+   *
+   * scala> def checkError(result: Either[Throwable, Int]): Try[String] = result.fold(_ => Failure(new java.lang.Exception), _ => Success("success"))
```
barambani, line 93: just a note, but don't consider it a blocker for the PR. I personally like the example in your comment [here](https://github.com/typelevel/cats/pull/3459#discussion_r437766077)
 
```scala
fa.attemptTap {
  case Right(a) => log.info(s"succeeded with $a")
  case Left(e) => log.error(s"failed with $e")
}
``` 
It describes very well a real use case for `attemptTap`. A simple program could be simulated with `State`. Something like
```scala
/**
 * {{{
 * scala> import cats._, data._, implicits._
 * scala> type F[A] = EitherT[State[String, *], String, A]
 * scala> object log { def info(m: String): F[Unit] = EitherT.liftF(State.modify(_ => s"info: $m")); def error(m: String): F[Unit] = EitherT.liftF(State.modify(_ => s"error: $m")) }
 *
 * scala> val prog1: F[Int] = EitherT.liftF(State.pure(1))
 * scala> val prog2: F[Int] = "this one failed".raiseError[F, Int]
 *
 * scala> val run1 = prog1.attemptTap { case Right(a) => log.info(a.toString); case Left(e) => log.error(e) }
 * scala> val run2 = prog2.attemptTap { case Right(a) => log.info(a.toString); case Left(e) => log.error(e) }
 *
 * scala> run1.value.runS("init").value
 * res0: String = info: 1
 *
 * scala> run2.value.runS("init").value
 * res1: String = error: this one failed
 * }}}
 */
```

Issue #3450, opened 2020-06-01
Kleisli ap method in Applicative has arguments reversed


justinhj:
Most Applicative instances have an `ap` which implements function application where both the function and the parameter are in an effect type. Most instances take the function first and the parameter second, but the Kleisli (and therefore the Reader) instances have the order of arguments swapped. 

Concretely:

```
@ Applicative[IntReader].pure(10).ap(Applicative[IntReader].pure((a: Int) => a + 1)).run(100)
res16: Id[Int] = 11

@ Applicative[Option].pure((a: Int) => a + 1).ap(Applicative[Option].pure(10))
res17: Option[Int] = Some(11)
```

Having spoken to @djspiewak on gitter about this, it's recommended to fix it. The order of parameters should be changed in the syntax enrichment. This should preserve binary compatibility... 

"by adding an overload and setting the existing one to private[data]"

justinhj:
Will take a stab at this.

dmndpl:
I tried adding the following 
` def ap[B, C](fa: Kleisli[F, A, B])(f: Kleisli[F, A, B => C]): Kleisli[F, A, C] =
    fa.ap(f)` in `KleisliApply[F[_], A]` and setting the previous `ap` to `private[data]` however the compiler is not happy. 

`double definition:
override def ap[B, C](f: cats.data.Kleisli[F,A,B => C])(fa: cats.data.Kleisli[F,A,B]): cats.data.Kleisli[F,A,C] at line 620 and
def ap[B, C](fa: cats.data.Kleisli[F,A,B])(f: cats.data.Kleisli[F,A,B => C]): cats.data.Kleisli[F,A,C] at line 623
have same type after erasure: (f: cats.data.Kleisli, fa: cats.data.Kleisli)cats.data.Kleisli`

To my understanding, this was the proposed solution, is there an alternative maybe?

djspiewak:
> To my understanding, this was the proposed solution, is there an alternative maybe?

The definition in `KleisliApply` is actually correct. The definition in `Kleisli` *itself* is what's wrong. What we need to add is the following:

```scala
  def ap[C, D, AA <: A](f: Kleisli[F, AA, C])(implicit F: Apply[F], ev: B As (C => D)): Kleisli[F, AA, D] = ???
```

Or something like it.

justinhj:
[...]

PR #3462, opened 2020-06-09
Reverse argument order for Kleisli ap method

justinhj:


justinhj:
WIP fixes #3450

PR #3467, opened 2020-06-10
Add Parallel traverseFilter functions

### LINKED ISSUE #3465: Add parallel variant for TraverseFilter
Has it been considered to add parallel variants for `TraverseFilter` ? I suppose `ParTraverseFilter`

` def parTraverseFilter[G[_], B]...`

` def parSequenceFilter[G[_], B]...` 

Anything preventing us from adding these ?

I would be happy to add these :)

LukaJCB:
Should fix #3465

### DIFF
--- a/core/src/main/scala/cats/syntax/all.scala
+++ b/core/src/main/scala/cats/syntax/all.scala
@@ -94,4 +94,4 @@ trait AllSyntaxBinCompat4
 
 trait AllSyntaxBinCompat5 extends ParallelBitraverseSyntax
 
-trait AllSyntaxBinCompat6 extends ParallelUnorderedTraverseSyntax
+trait AllSyntaxBinCompat6 extends ParallelUnorderedTraverseSyntax with ParallelTraverseFilterSyntax
--- a/core/src/main/scala/cats/syntax/all.scala
+++ b/core/src/main/scala/cats/syntax/all.scala
@@ -60,6 +60,7 @@ trait AllSyntax
     with VectorSyntax
     with WriterSyntax
     with ParallelFoldMapASyntax
+    with ParallelTraverseFilterSyntax
 
 trait AllSyntaxBinCompat0 extends UnorderedTraverseSyntax with ApplicativeErrorExtension with TrySyntax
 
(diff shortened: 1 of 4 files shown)


core/src/main/scala/cats/syntax/parallel.scala
```diff
@@ -100,6 +113,19 @@ final class ParallelTraversableOps[T[_], A](private val ta: T[A]) extends AnyVal
 
 }
 
+final class ParallelTraverseFilterOps[T[_], A](private val ta: T[A]) extends AnyVal {
```
travisbrown, line 116: I think it'd be better to consolidate these, so that all enrichment methods for `T[A]` are on a single enrichment class, mostly in order to reduce the number of implicits in `cats.syntax.all`, but also just to simplify this code. In this case that would mean putting these methods on something named `ParallelTraversableOps` or `ParallelFoldMapAOps`, though, which also seems kind of bad.

I think for now this version is probably best, but we should definitely revisit this in Cats 3.

SethTisue:
this PR brought to you by https://www.youtube.com/watch?v=SMMgwfhPCzM&feature=youtu.be :-)

saraiva132:
You are fast @LukaJCB Thanks for the addition :)

core/src/main/scala/cats/syntax/all.scala
```diff
@@ -94,4 +94,4 @@ trait AllSyntaxBinCompat4
 
 trait AllSyntaxBinCompat5 extends ParallelBitraverseSyntax
 
-trait AllSyntaxBinCompat6 extends ParallelUnorderedTraverseSyntax
+trait AllSyntaxBinCompat6 extends ParallelUnorderedTraverseSyntax with ParallelTraverseFilterSyntax
```
kailuowang, line 97: Forgive me for being out for a while, but just curious why we still extend a binCompat trait instead of the `AllSyntax` itself?

travisbrown: Ah, good catch. I agree that this should go directly on `AllSyntax`, as in #3392.

LukaJCB: ah of course, thanks!

travisbrown:
:+1: on green, but I think we still need one more review.

LukaJCB:
Thanks @barambani 😊 merging this

PR #3469, opened 2020-06-12
Move nomenclature questions below introductory paragraph

### LINKED ISSUE #3460: Describe example questions at top of nomenclature page.
[Comment from reader](https://gitter.im/typelevel/general?at=5ede73972c49c45f5ace5711):

> Should these three questions be on the nomenclature page? They look like they’re supposed to be on FAQ. There is no answer and they just kind of float there:
>
> What is the difference between unit and void?
> To discard the first value and keep only the first effect, is it <* or *>?
> How do I make a computation F[A] fail by checking a condition on the value?

https://typelevel.org/cats/nomenclature.html

This page could preface those questions the paragraph that currently follows them:

> This is a catalogue of the major functions, type classes, and data types in Cats. It serves as a bird’s-eye view of each class capabilities. It is also intended as a go-to reference for Cats users, who may not recall the answer to questions like **these:**
>
> <insert questions here>

☝️ suggested copy

rmehri01:
Closes #3460

### DIFF
--- a/docs/src/main/tut/nomenclature.md
+++ b/docs/src/main/tut/nomenclature.md
@@ -5,12 +5,11 @@ section: "Glossary"
 position: 60
 ---
 
-> - What is the difference between `unit` and `void`?
-> - To discard the first value and keep only the first effect, is it `<*` or `*>`?
-> - How do I make a computation `F[A]` fail by checking a condition on the value?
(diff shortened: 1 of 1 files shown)


rmehri01:
Np! :)

PR #3475, opened 2020-06-17
Merge 2.2.x after 2.2.0-M3 release

travisbrown:
I went ahead and removed the Scala.js cross-publishing script, but we can always recover it from the history if we need to publish for 0.6 again.

Please note that this needs to be a merge commit! If we squash or rebase, then the 2.2.0-M3 tag won't be on master.


travisbrown:
@LukaJCB This definitely isn't a big deal, but we'll need the `2.2.x` branch for at least the next couple of releases. I've just been following PROCESS.md on this: merge master into `2.2.x`, push and publish from there, and then PR `2.2.x` back onto master. There's no reason I guess that `2.2.x` has to stay on GitHub in the meantime, but I think it makes things a little clearer.

LukaJCB:
oh sorry, I must've done that completely subconsciously, I don't even remember doing it.

Issue #3353, opened 2020-03-11
Investigate antisymmetry failure for UpperBounded for Duration


travisbrown:
This turned up in #3350 on Scala 2.12, and doesn't look familiar to me. It's not related to that change, so I'm just restarting the job there, but wanted to capture the failure.
```scala
[info] Compiling 2 Scala sources to /home/travis/build/typelevel/cats/kernel-laws/jvm/target/scala-2.12/test-classes ...
[info] Done compiling.
[info] Tests: failing seed for upperBounded.antisymmetry is 8MP8QBDQNKwLLH9fN7tBvEZqp74Gi9P01N1JpHidFrI=
[info] - UpperBounded[Duration].upperBounded.antisymmetry *** FAILED ***
[info]   GeneratorDrivenPropertyCheckFailedException was thrown during property evaluation.
[info]    (LawTests.scala:227)
[info]     Falsified after 29 successful property evaluations.
[info]     Location: (LawTests.scala:227)
[info]     Occurred when passed generated values (
[info]       arg0 = 36 days,
[info]       arg1 = 864 hours,
[info]       arg2 = org.scalacheck.GenArities$$Lambda$524/1880085276@32700c2e
[info]     )
[info]     Label of failing property:
[info]       Expected: true
[info]   Received: false
[info] ScalaTest

**Comment by travisbrown** [comment]
Similar failure on bc77b900801c70800405a2b374c4dc25b2123ea1 for 2.13:

```scala
failing seed for order.antisymmetry is BIuiwWI2xJdRhpszwX4UH3ULEdGDr49xSwwkgj-w6bB=
[info] - Order[Duration].order.antisymmetry *** FAILED ***
[info]   GeneratorDrivenPropertyCheckFailedException was thrown during property evaluation.
[info]    (LawTests.scala:192)
[info]     Falsified after 89 successful property evaluations.
[info]     Location: (LawTests.scala:192)
[info]     Occurred when passed generated values (
[info]       arg0 = -156 days,
[info]       arg1 = -3744 hours,
[info]       arg2 = org.scalacheck.GenArities$$Lambda$510/1150077346@5b5bbfa8
[info]     )
[info]     Label of failing property:
[info]       Expected: true
[info]   Received: false
[info] ScalaTest
```

barambani:
[...]

PR #3478, opened 2020-06-18
Add ior docs

kazchimo:
add some docs to Ior

### DIFF
--- a/core/src/main/scala/cats/data/Ior.scala
+++ b/core/src/main/scala/cats/data/Ior.scala
@@ -613,7 +613,7 @@ sealed abstract class Ior[+A, +B] extends Product with Serializable {
    * res1: List[Ior[String,Int]] = List(Right(123), Right(246))
    *
    * scala> Ior.both("abc", 123).traverse(i => List(i, i * 2))
-   * val res2: List[Ior[String,Int]] = List(Both(abc,123), Both(abc,246))
+   * res2: List[Ior[String,Int]] = List(Both(abc,123), Both(abc,246))
    * }}}
    */
   final def traverse[F[_], AA >: A, D](g: B => F[D])(implicit F: Applicative[F]): F[AA Ior D] =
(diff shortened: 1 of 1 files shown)


core/src/main/scala/cats/data/Ior.scala
```diff
@@ -576,11 +576,46 @@ sealed abstract class Ior[+A, +B] extends Product with Serializable {
...
+   * scala> import cats.implicits._
+   *
+   * scala> "abc".leftIor[Int].traverse(i => List(i, i * 2))
+   * res0: List[Ior[String,Int]] = List(Left(abc))
+   *
+   * scala> 123.rightIor[String].traverse(i => List(i, i * 2))
+   * res1: List[Ior[String,Int]] = List(Right(123), Right(246))
+   *
+   * scala> Ior.both("abc", 123).traverse(i => List(i, i * 2))
+   * val res2: List[Ior[String,Int]] = List(Both(abc,123), Both(abc,246))
```
barambani, line 616:
```suggestion
   * res2: List[Ior[String,Int]] = List(Both(abc,123), Both(abc,246))
```
otherwise it won't be checked (see [here](https://travis-ci.org/github/typelevel/cats/jobs/699667197#L664)).

kazchimo:
@barambani thanks

barambani:
thank you for the quick fix 👍 .

barambani:
It's running now, thanks.

kazchimo:
hmm, law tests seems to fail, but I changed only docs.
How this can be described?

LukaJCB:
Unfortunately, it's a flaky test we're aware of. Can you restart the job?

kazchimo:
@LukaJCB thanks for your info.
Is this problem already reported on any issue?
If not, this issue should be made.

LukaJCB:
Yeah it's this: https://github.com/typelevel/cats/issues/3353

LukaJCB:
Thanks a bunch!

PR #3480, opened 2020-06-18
Rename master to main

travisbrown:
This is a minor thing, but it's easy and will likely [soon be the default](https://twitter.com/natfriedman/status/1271253144442253312) on GitHub.

I've only changed links and references to the `master` branch in current instructions or documentation, not references in historical contexts like the changelog. I've chosen `main` since there seems to be consensus that that's the best replacement, but I'm happy to change it to one of the other alternatives.

There will be a few other steps that are necessary (see [here](https://twitter.com/mpilquist/status/1272694351181856769) or [here](https://www.hanselman.com/blog/EasilyRenameYourGitDefaultBranchFromMasterToMain.aspx)) in conjunction with this PR, which just ensures that we don't break links or CI.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -185,7 +185,7 @@ As a side note, the latter command uses [sbt-mima](https://github.com/lightbend/
 ## Contributing documentation
 
 ### source for the documentation
-The documentation for this website is stored alongside the source, in the [docs subproject](https://github.com/typelevel/cats/tree/master/docs).
+The documentation for this website is stored alongside the source, in the [docs subproject](https://github.com/typelevel/cats/tree/main/docs).
 
 * The source for the tut compiled pages is in `docs/src/main/tut`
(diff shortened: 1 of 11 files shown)


travisbrown:
(When I call this a "minor thing" I don't mean to downplay concerns about exclusionary language, just to note that e.g. GitHub's actions on this issue don't really deserve much celebration in the context of their continued relationship with ICE.)

mpilquist:
Thanks for doing this!

rossabaker:
So far, I've only made this move on repos with empty PR queues.  Do we just change the target branch on the open ones, or do we have to reopen them?

I'm :+1: on the move either way.

travisbrown:
@rossabaker I made this change [in Dhall for Java](https://github.com/travisbrown/dhallj), and had to change the target branch for open PRs, but that's all—I didn't have to reopen them.

djspiewak:
This appears to be overwhelmingly approved. Just staying open to track the other juggling required to make this happen.

larsrh:
Superseded by #3833.

PR #3482, opened 2020-06-18
Rerun Simulacrum after rearrangement for Scala.js

travisbrown:
See [this bug](https://github.com/scala/bug/issues/11997) and the Simulacrum Scalafix workaround by @joroKr21 [here](https://github.com/typelevel/simulacrum-scalafix/pull/38). In short the problem is that defining a lowercase `ops` after the capitalized `Ops` trait causes problems on Scala.js 1.x on some platforms. (This bug also affects the macro annotation implementation of Simulacrum.)

This change uses the new 0.5.0 release of Simulacrum Scalafix I just published with that and one other fix (which doesn't affect Cats's usage).

### DIFF
--- a/alleycats-core/src/main/scala/alleycats/ConsK.scala
+++ b/alleycats-core/src/main/scala/alleycats/ConsK.scala
@@ -24,6 +24,17 @@ object ConsK {
    */
   @inline def apply[F[_]](implicit instance: ConsK[F]): ConsK[F] = instance
 
+  @deprecated("Use cats.syntax object imports", "2.2.0")
+  object ops {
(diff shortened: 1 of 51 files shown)


barambani:
If that's the problem, isn't this change triggering it ?

travisbrown:
@barambani Sorry, yes, I wrote "before" where I meant "after". I've edited the description.

travisbrown:
Hmm, the Simulacrum Scalafix release is on Sonatype Releases but doesn't seem to have synced to Maven Central yet. I'll try restarting the build a little later.

travisbrown:
Okay, this should be ready for review now.

PR #3486, opened 2020-06-19
Another attempt at Dotty cross-building

travisbrown:
This PR partially includes and supersedes #3269. It also depends on #3484, which is ~trivial and should be merged soon.

There are a lot of issues with building and running our tests on Dotty, and I don't have the interest or will to deal with them right now, but I don't think that should block getting the core modules cross-building now. This PR enables Dotty cross-building (on the JVM only) for kernel, kernel-laws, core, laws, alleycats, and alleycats-laws, but does not run any tests.

Merging this now gets us a couple of things:

* We catch changes that break on Dotty sooner, at least for most of the code.
* It's easier for people other than me to work on the issues that are turning up in the tests.

None of the changes here should have any impact on the Scala 2 build, except for the introduction of a package-private `FunctionKMacroMethods` trait.

----

Note that I've disabled fatal warnings (for Dotty only) because of two issues:

* A [false positive from an exhaustivity check](https://github.com/lampepfl/dotty/issues/9067), which should be fixed in 0.25.
* Some weird type test error in `ContT` that I think I investigated once and maybe even reported, but can't find now.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -782,7 +782,8 @@ addCommandAlias("validateJVM", ";fmtCheck;buildJVM;bench/test;validateBC;makeMic
 addCommandAlias("validateJS", ";catsJS/compile;testsJS/test;js/test")
 addCommandAlias("validateKernelJS", "kernelLawsJS/test")
 addCommandAlias("validateFreeJS", "freeJS/test") //separated due to memory constraint on travis
-addCommandAlias("validate", ";clean;validateJS;validateKernelJS;validateFreeJS;validateJVM")
+addCommandAlias("validateDotty", ";++0.24.0!;alleycatsLawsJVM/compile")
+addCommandAlias("validate", ";clean;validateJS;validateKernelJS;validateFreeJS;validateJVM;validateDotty")
 
 addCommandAlias("prePR", "fmt")
 
--- a/build.sbt
+++ b/build.sbt
@@ -44,9 +44,9 @@ def scalaVersionSpecificFolders(srcName: String, srcBaseDir: java.io.File, scala
     List(CrossType.Pure, CrossType.Full)
       .flatMap(_.sharedSrcDir(srcBaseDir, srcName).toList.map(f => file(f.getPath + suffix)))
   CrossVersion.partialVersion(scalaVersion) match {
-    case Some((2, y)) if y >= 13 =>
-      extraDirs("-2.13+")
-    case _ => Nil
+    case Some((2, y)) => extraDirs("-2.x") ++ (if (y >= 13) extraDirs("-2.13+") else Nil)
+    case Some((0, _)) => extraDirs("-2.13+") ++ extraDirs("-3.x")
+    case _            => Nil
   }
 }
 
@@ -59,8 +59,11 @@ commonScalaVersionSettings
 
 ThisBuild / mimaFailOnNoPrevious := false
 
+def doctestGenTestsDottyCompat(isDotty: Boolean, genTests: Seq[File]): Seq[File] =
+  if (isDotty) Nil else genTests
+
 lazy val commonSettings = commonScalaVersionSettings ++ Seq(
-  scalacOptions ++= commonScalacOptions(scalaVersion.value),
+  scalacOptions ++= commonScalacOptions(scalaVersion.value, isDotty.value),
   Compile / unmanagedSourceDirectories ++= scalaVersionSpecificFolders("main", baseDirectory.value, scalaVersion.value),
   Test / unmanagedSourceDirectories ++= scalaVersionSpecificFolders("test", baseDirectory.value, scalaVersion.value),
   resolvers ++= Seq(Resolver.sonatypeRepo("releases"), Resolver.sonatypeRepo("snapshots")),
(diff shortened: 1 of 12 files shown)


laws/src/main/scala/cats/laws/NonEmptyTraverseLaws.scala
```diff
@@ -1,6 +1,8 @@
 package cats.laws
 
-import cats.{Apply, Id, NonEmptyTraverse, Semigroup}
+// The `catsInstancesForId` import is necessary to work around a Dotty
+// issue related to https://github.com/lampepfl/dotty/issues/9067.
```
barambani, line 4: The issue seems resolved in 0.25.0's release candidates. I guess we like to stay on stable versions ? I'm just curious.

travisbrown: @barambani I thought about using 0.25.0-RC2, and I don't think we should avoid the Dotty RCs on principle, but the 3.2.0 ScalaTest dependencies are currently only published for 0.24.0, and I think it's more important to make it easy for people to work on the test issues.

build.sbt
```diff
@@ -68,19 +72,26 @@ lazy val commonSettings = commonScalaVersionSettings ++ Seq(
...
 lazy val simulacrumSettings = Seq(
-  addCompilerPlugin(scalafixSemanticdb),
-  scalacOptions ++= Seq(s"-P:semanticdb:targetroot:${baseDirectory.value}/target/.semanticdb", "-Yrangepos"),
-  libraryDependencies += "org.typelevel" %% "simulacrum-scalafix-annotations" % "0.5.0",
+  libraryDependencies ++= (if (isDotty.value) Nil else Seq(compilerPlugin(scalafixSemanticdb))),
+  scalacOptions ++= (
+    if (isDotty.value) Nil else Seq(s"-P:semanticdb:targetroot:${baseDirectory.value}/target/.semanticdb", "-Yrangepos")
+  ),
+  libraryDependencies +=
+    ("org.typelevel" %% "simulacrum-scalafix-annotations" % "0.5.0").withDottyCompat(scalaVersion.value),
```
djspiewak, line 94: Does this have to be in the *runtime* classpath?

build.sbt
```diff
@@ -140,15 +151,25 @@ lazy val includeGeneratedSrc: Setting[_] = {
...
+    "org.scalatest" %%% "scalatest-funsuite" % scalatestVersion % Test,
     "org.scalatestplus" %%% "scalacheck-1-14" % scalatestplusScalaCheckVersion % Test
+  ),
+  libraryDependencies ++= Seq(
+    ("org.typelevel" %%% "discipline-scalatest" % disciplineScalatestVersion % Test)
+  ).map(
+    _.exclude("org.scalatestplus", "scalacheck-1-14_2.13")
+      .exclude("org.scalactic", "scalactic_2.13")
+      .exclude("org.scalatest", "scalatest_2.13")
+      .withDottyCompat(scalaVersion.value)
```
djspiewak, line 172: I'm *tentatively* okay with merging this as-is, but only if we follow up on discipline-scalatest asap to update its dependency. Exclusion rules in published dependencies are dangerous on several levels.

build.sbt
```diff
@@ -773,29 +810,33 @@ lazy val crossVersionSharedSources: Seq[Setting[_]] =
     }
   }
 
-def commonScalacOptions(scalaVersion: String) =
+def commonScalacOptions(scalaVersion: String, isDotty: Boolean) =
```
djspiewak, line 813: As an fyi, the `isDotty` setting is computed by string matching on `scalaVersion`. Though, it's not exposed as such, so there may be no way we can take advantage of it.

laws/src/main/scala/cats/laws/NonEmptyTraverseLaws.scala
djspiewak: In general, I think the *right* thing for us (and the whole ecosystem) to do would be to publish for the latest stable Dotty (in this case, 0.24.0) *as well as* the exact latest RC (in this case, 0.25.0-RC2). Obviously it's a pain keeping up with RCs, but it helps quite a bit. Dependencies which don't do this, particularly those which lack macro usage, effectively force their downstream dependents to fall back on the 2.13 cross-publish, which is annoying.

build.sbt
travisbrown: What do you think about changing the exclusion to only apply for Dotty (which we’re not currently publishing anyway)? I think it’s likely I’ll be the one publishing discipline-scalatest, and I don’t really want to start doing that for every Dotty release since it’s just a test dependency.

travisbrown: @djspiewak Oh, nevermind, I forgot that this is only a test dependency (and that's configured here, so there's no danger of them accidentally missing an `% Test` at the use site), so I don't see any problem with keeping the exclusions as they are now.

build.sbt
travisbrown: Right, I just thought this was clearer than trying to figure out and replicate whatever logic sbt-dotty uses for `isDotty`, but I have no objection to using `.startsWith("0")` or something similar instead.

build.sbt
travisbrown: Hmm, this isn't new to this PR, but it's a good point—I'll change this to `Provided` before 2.2.0, and make sure that we're doing the same thing with Simulacrum Scalafix as we were with Simulacrum (the macro annotation version) in this respect. But I think that should be a follow-up.

build.sbt
djspiewak: Ah if it's only a test dependency then I'm okay with it for now.

PR #3495, opened 2020-06-23
Add implicit variance support for Profunctor typeclass

### LINKED ISSUE #3452: Add implicit variance materialization for profunctor
Similar to #3436, we should have a simultaneous co- and contravariant materialization for any `F[_, _]` that forms a `Profunctor`.

gagandeepkalra:
resolves #3452

### DIFF
--- a/core/src/main/scala/cats/arrow/Profunctor.scala
+++ b/core/src/main/scala/cats/arrow/Profunctor.scala
@@ -41,6 +41,36 @@ import scala.annotation.implicitNotFound
    */
   def rmap[A, B, C](fab: F[A, B])(f: B => C): F[A, C] =
     dimap[A, B, A, C](fab)(identity)(f)
+
+  /**
+   * Narrows A into a subtype AA.
+   * Example:
+   * {{{
+   * scala> import cats.syntax.profunctor._
+   * scala> import cats.instances.function._
+   * scala>
+   * scala> sealed trait Foo
+   * scala> case object Bar extends Foo
+   * scala> val x1: Foo => Int = _ => 1
+   * scala> val x2: Bar.type => Int = x1.leftNarrow
+   * }}}
+   */
+  def leftNarrow[A, B, AA <: A](fab: F[A, B]): F[AA, B] = fab.asInstanceOf[F[AA, B]]
+
+  /**
+   * Widens B into a supertype BB.
+   * Example:
+   * {{{
+   * scala> import cats.syntax.profunctor._
+   * scala> import cats.instances.function._
+   * scala>
+   * scala> sealed trait Foo
+   * scala> case object Bar extends Foo
+   * scala> val x1: Int => Bar.type = _ => Bar
+   * scala> val x2: Int => Foo = x1.rightWiden
+   * }}}
+   */
+  def rightWiden[A, B, BB >: B](fab: F[A, B]): F[A, BB] = fab.asInstanceOf[F[A, BB]]
 }
 
 object Profunctor {
@@ -77,6 +107,8 @@ object Profunctor {
     def dimap[C, D](f: C => A)(g: B => D): F[C, D] = typeClassInstance.dimap[A, B, C, D](self)(f)(g)
     def lmap[C](f: C => A): F[C, B] = typeClassInstance.lmap[A, B, C](self)(f)
     def rmap[C](f: B => C): F[A, C] = typeClassInstance.rmap[A, B, C](self)(f)
+    def leftNarrow[C <: A]: F[C, B] = typeClassInstance.leftNarrow[A, B, C](self)
+    def rightWiden[C >: B]: F[A, C] = typeClassInstance.rightWiden[A, B, C](self)
   }
   trait AllOps[F[_, _], A, B] extends Ops[F, A, B]
   trait ToProfunctorOps extends Serializable {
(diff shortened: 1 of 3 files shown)


core/src/main/scala/cats/arrow/Profunctor.scala
```diff
@@ -41,6 +41,36 @@ import scala.annotation.implicitNotFound
...
+   * scala> import cats.syntax.profunctor._
+   * scala> import cats.instances.function._
+   * scala>
+   * scala> sealed trait Foo
+   * scala> case object Bar extends Foo
+   * scala> val x1: Int => Bar.type = _ => Bar
+   * scala> val x2: Int => Foo = x1.rightWiden
+   * }}}
+   */
+  def rightWiden[A, B, BB >: B](fab: F[A, B]): F[A, BB] = fab.asInstanceOf[F[A, BB]]
```
gagandeepkalra, line 73: added these both to avoid `lmap` and `rmap` with identity

tests/src/test/scala/cats/tests/VarianceSuite.scala
```diff
@@ -23,24 +25,45 @@ class VarianceSuite extends CatsSuite {
   }
 
   test("Auto-variance should widen a bifunctor automatically") {
-    def shouldInfer[F[_, _]: Bifunctor](fi: F[Int, Int]) =
+    def shouldInfer[F[_, _]: Bifunctor](fi: F[Int, Int]): F[Either[Bar, Baz.type], Either[Bar, Baz.type]] =
```
gagandeepkalra, line 28: think that the test would be more readable if explicitly typed

LukaJCB: The idea was to assert here that scalac can infer the correct result type :)

core/src/main/scala/cats/conversions/VarianceConversions.scala
```diff
@@ -1,10 +1,14 @@
 package cats
 package conversions
+import cats.arrow.Profunctor
 
 trait VarianceConversions extends VarianceConversionsLowPriority {
   implicit def autoWidenBifunctor[F[_, _]: Bifunctor, A, B >: A, C, D >: C](fac: F[A, C]): F[B, D] =
     Bifunctor[F].leftWiden(Bifunctor[F].rightFunctor.widen(fac))
 
+  implicit def autoWidenProfunctor[F[_, _]: Profunctor, A, B <: A, C, D >: C](fac: F[A, C]): F[B, D] =
```
ritschwumm, line 9: i think the name of this method should reflect the fact that it does not only widen, but narrow, too.

tests/src/test/scala/cats/tests/VarianceSuite.scala
```diff
@@ -23,24 +25,45 @@ class VarianceSuite extends CatsSuite {
...
   }
 
+  test("Auto-variance should widen a profunctor automatically") {
+    def shouldInfer[F[_, _]: Profunctor](fi: F[Int, Int]): F[Either[FooBar, FooBar], Either[Bar, Baz.type]] =
+      fi.dimap((_: Either[FooBar, FooBar]) => 1)(i => if (true) Left(Bar(i)) else Right(Baz))
+
+    def inferred[F[_, _]: Profunctor](fi: F[Int, Int]): F[Either[Foo, Foo], Either[Foo, Foo]] = shouldInfer[F](fi)
+  }
+
+  test("Auto-variance should left widen a profunctor automatically") {
```
ritschwumm, line 55: left-narrow, right?

tests/src/test/scala/cats/tests/VarianceSuite.scala
```diff
@@ -23,24 +25,45 @@ class VarianceSuite extends CatsSuite {
...
 
   test("Auto-variance should right widen a bifunctor automatically") {
-    def shouldInfer[F[_, _]: Bifunctor](fi: F[Int, Int]) =
+    def shouldInfer[F[_, _]: Bifunctor](fi: F[Int, Int]): F[Either[Bar, Baz.type], Int] =
       fi.bimap(i => if (true) Left(Bar(i)) else Right(Baz), identity)
 
     def inferred[F[_, _]: Bifunctor](fi: F[Int, Int]): F[Either[Foo, Foo], Int] = shouldInfer[F](fi)
   }
 
+  test("Auto-variance should widen a profunctor automatically") {
```
ritschwumm, line 48: right-widen, maybe?

core/src/main/scala/cats/conversions/VarianceConversions.scala
gagandeepkalra: The idea as I understood is that we get the behaviour `F[A, C] <: F[B, D]` given constraints `A >: B` and `C <: D`, the opposite however won't hold with these them.

The same is true for other implicits here, if the return type is `A` then we can return any instance of `A` or it's subtypes

tests/src/test/scala/cats/tests/VarianceSuite.scala
gagandeepkalra: looks like I botched up left and right

tests/src/test/scala/cats/tests/VarianceSuite.scala
barambani: I agree. If we don't want to be specific about the side we could also stay general like
```suggestion
  test("Auto-variance should convert a profunctor's variance automatically") {
```

tests/src/test/scala/cats/tests/VarianceSuite.scala
barambani: The test is still checking that the conversion happens at calls site
```scala
def inferred[F[_, _]: Bifunctor](fi: F[Int, Int]): F[Either[Foo, Foo], Either[Foo, Foo]] = shouldInfer[F](fi)
```
If we remove the `import cats.conversions.all._` the compilation fails also in this version, so the test is still valid. I don't have strong preferences about the type annotation.

gagandeepkalra: Now if I think some more, between the two not having types is still more useful, thanks @LukaJCB

PR #3498, opened 2020-06-25
Enable breakout in functions reduceRightToOption and reduceRightTo.

takayahilton:
now
```scala
import cats.data.NonEmptyList
import cats.implicits._

val notAllEven = NonEmptyList.of(2, 4, 6, 9, 10, 12, 14)
notAllEven.reduceMapA { a => println(a); if (a % 2 == 0) Some(a) else None }

14
12
10
9
6
4
2
res: Option[Int] = None
```

After applying this PR

```scala
import cats.data.NonEmptyList
import cats.implicits._

val notAllEven = NonEmptyList.of(2, 4, 6, 9, 10, 12, 14)
notAllEven.reduceMapA { a => println(a); if (a % 2 == 0) Some(a) else None }

2
4
6
9
res: Option[Int] = None
```

### DIFF
--- a/core/src/main/scala/cats/data/NonEmptyChain.scala
+++ b/core/src/main/scala/cats/data/NonEmptyChain.scala
@@ -2,7 +2,6 @@ package cats
 package data
 
 import NonEmptyChainImpl.create
-import cats.{Order, Semigroup}
 import cats.kernel._
 import scala.collection.immutable.SortedMap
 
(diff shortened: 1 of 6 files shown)


takayahilton:
ref: https://github.com/typelevel/cats/pull/3015

core/src/main/scala/cats/data/NonEmptyChain.scala
```diff
@@ -2,7 +2,6 @@ package cats
 package data
 
 import NonEmptyChainImpl.create
-import cats.{Order, Semigroup}
```
barambani, line 5: Nice cleanup here and below, thanks.

PR #3502, opened 2020-06-28
Add Tuple2K Semigroupal

### LINKED ISSUE #3410: Missing TupleK Semigroupal
I don't see an implementation (although there is ContravariantMonoidal, which is more constrained).

I'll plan to send a PR, but if anyone sees a reason why this can't exist, please feel free to let me know.

rmehri01:
Closes #3410.

I wasn't really sure where in the Tuple2kInstances this should be added so I just put it in the last one so please tell me if there is somewhere else it should go. Also, should tests be added because I had a bit of trouble doing so. Thanks!

### DIFF
--- a/core/src/main/scala/cats/data/Tuple2K.scala
+++ b/core/src/main/scala/cats/data/Tuple2K.scala
@@ -189,6 +189,22 @@ sealed abstract private[data] class Tuple2KInstances8 {
       def F: Functor[F] = FF
       def G: Functor[G] = GG
     }
+  implicit def catsDataSemigroupalForTuple2K[F[_], G[_]](implicit
+    FF: Semigroupal[F],
+    GG: Semigroupal[G]
+  ): Semigroupal[λ[α => Tuple2K[F, G, α]]] =
+    new Tuple2KSemigroupal[F, G] {
+      def F: Semigroupal[F] = FF
+      def G: Semigroupal[G] = GG
+    }
+}
+
+sealed private[data] trait Tuple2KSemigroupal[F[_], G[_]] extends Semigroupal[λ[α => Tuple2K[F, G, α]]] {
+  def F: Semigroupal[F]
+  def G: Semigroupal[G]
+
+  override def product[A, B](fa: Tuple2K[F, G, A], fb: Tuple2K[F, G, B]): Tuple2K[F, G, (A, B)] =
+    Tuple2K(F.product(fa.first, fb.first), G.product(fa.second, fb.second))
 }
 
 sealed private[data] trait Tuple2KFunctor[F[_], G[_]] extends Functor[λ[α => Tuple2K[F, G, α]]] {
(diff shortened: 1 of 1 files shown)


core/src/main/scala/cats/data/Tuple2K.scala
```diff
@@ -189,6 +189,22 @@ sealed abstract private[data] class Tuple2KInstances8 {
       def F: Functor[F] = FF
       def G: Functor[G] = GG
     }
+  implicit def catsDataSemigroupalForTuple2K[F[_], G[_]](implicit
+    FF: Semigroupal[F],
+    GG: Semigroupal[G]
+  ): Semigroupal[λ[α => Tuple2K[F, G, α]]] =
```
LukaJCB, line 195: any reason for not using the shorthand `*` here? :)

LukaJCB: Oh I see, it's not used in any of the others either, that's odd, but should be fine :)

rmehri01: Yeah I was just trying to follow the style of the others, unless you want me to change them?

LukaJCB: no, it's all good! thanks!

rmehri01:
Alright I can do that. Just another thing, does it matter which tuple2k instance class this goes in because there was 0-8 so I wasn't really sure and should this be tested?

core/src/main/scala/cats/data/Tuple2K.scala
rmehri01: Alright sounds good :)

rmehri01:
Also is that import at the top of the file `import cats.Contravariant` needed? Intellij says it isn't used.

joroKr21:
> Alright I can do that. Just another thing, does it matter which tuple2k instance abstract class this goes in because there was 0-8 so I wasn't really sure and also should this be tested?

This is the correct place. I'm not sure about testing, I guess you would need a type that has `Semigroupal` but not `Applicative` to test this.

rmehri01:
> > Alright I can do that. Just another thing, does it matter which tuple2k instance abstract class this goes in because there was 0-8 so I wasn't really sure and also should this be tested?
> 
> This is the correct place. I'm not sure about testing, I guess you would need a type that has `Semigroupal` but not `Applicative` to test this.

Ah okay thanks for the response, I meant more like does it matter which tuple2k instances class this goes in because I'm not really sure why they are labelled 0-8.

joroKr21:
> Also does it matter which tuple2k instances class this goes in because I'm not really sure why they are labelled 0-8, I thought maybe it was because they had some sort of structure.

It matters but this is the correct place to put it. The reason they exist is implicit priority. `Semigroupal` should have the lowest priority (same as `Functor`) because it has the least constraints. There is also no overlap between `Functor` and `Semigroupal` so they can be in one place without clashing.

rmehri01:
> It matters but this is the correct place to put it. The reason they exist is implicit priority. `Semigroupal` should have the lowest priority (same as `Functor`) because it has the least constraints. There is also no overlap between `Functor` and `Semigroupal` so they can be in one place without clashing.

Ah gotcha, thank you! :)

rmehri01:
Np, thanks for the help everyone! :)

PR #3508, opened 2020-07-05
Put implicit support for evidence from predef types

### LINKED ISSUE #2569: cats.evidence.Is and Scala 2.13
As far as I can tell, [`=:=` in Scala 2.13](https://github.com/scala/scala/blob/v2.13.0-M5/src/library/scala/Predef.scala#L662) is very similar (identical?) in functionality to `cats.evidence.Is`. Would it make sense to revisit the decision regarding `Is.unsafeFromPredef`?

- One thing is the `unsafe` prefix. Is it really unsafe? Especially now that in 2.13, it seems possible to write `(A =:= B) => (A Is B)` without any `asInstanceOf`.
- The other (probably more important) is implicitness: *if* they're equivalent, maybe having an `Is` implicitly available when `=:=` is, would not be so bad.

johnynek:
close #2569

There is no reason not to create Is and As instances from the predef ones. The changes in 2.13 make this perfectly safe, but nothing in 2.12 (or before) made it unsafe, the types just lacked a substitute method.

This PR improves matters a bit.

I did notice that `Is` is not sealed, but `As` is. In general these two types have a pretty different API considering how similar they are conceptually.

I only did a bit, I added `toPredef` to both of them to convert back to predef types.

### DIFF
--- a/core/src/main/scala/cats/evidence/As.scala
+++ b/core/src/main/scala/cats/evidence/As.scala
@@ -39,8 +39,10 @@ sealed abstract class As[-A, +B] extends Serializable {
    * A value `A As B` is always sufficient to produce a similar `Predef.<:<`
    * value.
    */
-  @inline final def toPredef: A <:< B =
-    substitute[<:<[*, B]](implicitly[B <:< B])
+  @inline final def toPredef: A <:< B = {
+    type F[-Z] = <:<[Z, B]
+    substitute[F](implicitly[B <:< B])
+  }
 }
 
 sealed abstract class AsInstances {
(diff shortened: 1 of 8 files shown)


core/src/main/scala/cats/evidence/As.scala
```diff
@@ -34,6 +34,13 @@ sealed abstract class As[-A, +B] extends Serializable {
   @inline final def compose[C](that: (C As A)): (C As B) = As.compose(this, that)
 
   @inline final def coerce(a: A): B = As.witness(this)(a)
+
+  /**
+   * A value `A As B` is always sufficient to produce a similar `Predef.<:<`
+   * value.
+   */
+  @inline final def toPredef: A <:< B =
+    substitute[<:<[*, B]](implicitly[B <:< B])
```
travisbrown, line 43: I think this needs the same treatment as above? Something like:
```scala
type F[-Z] = Z <:< B
substitute[F](implicitly[B <:< B])
```

johnynek: yes, you are right! I got impatient because on my slow machine the tests took too long. I've fixed the tests on 2.12 and 2.13 so I think it will be green now.

Thank you for the review and thanks for considering it for 2.2.0.

PR #3515, opened 2020-07-11
Add some more implementations to Function0 and Function1 Monads

johnynek:
This is just filling out some methods on these Monads that were left to default implementations. I think implementing them by hand slightly reduces the stack depth we are at when calling these

### DIFF
--- a/core/src/main/scala/cats/instances/function.scala
+++ b/core/src/main/scala/cats/instances/function.scala
@@ -81,10 +81,7 @@ sealed private[instances] trait Function0Instances extends Function0Instances0 {
         () => (fa(), fb())
 
       override def ap[A, B](f: () => A => B)(fa: () => A): () => B =
-        () => {
-          val fnAB = f()
-          fnAB(fa())
-        }
+        () => f()(fa())
 
       def flatMap[A, B](fa: () => A)(f: A => () => B): () => B =
         () => f(fa())()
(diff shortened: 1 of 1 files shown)


core/src/main/scala/cats/instances/function.scala
```diff
@@ -81,7 +81,10 @@ sealed private[instances] trait Function0Instances extends Function0Instances0 {
         () => (fa(), fb())
 
       override def ap[A, B](f: () => A => B)(fa: () => A): () => B =
-        () => f().apply(fa())
```
ritschwumm: why was that `.apply` here in the first place? does `() => f()(fa())` work on dotty, maybe?

core/src/main/scala/cats/instances/function.scala
```diff
@@ -71,6 +71,21 @@ sealed private[instances] trait Function0Instances extends Function0Instances0 {
...
+
+      override def map2[A, B, C](fa: () => A, fb: () => B)(fn: (A, B) => C): () => C =
+        () => fn(fa(), fb())
+
+      override def product[A, B](fa: () => A, fb: () => B): () => (A, B) =
+        () => (fa(), fb())
+
+      override def ap[A, B](f: () => A => B)(fa: () => A): () => B =
+        () => {
+          val fnAB = f()
```
travisbrown, line 85: You can write this as either `f.apply.apply(fa())` or `f()(fa())` on Dotty, both of which are I think a little nicer than the extra `val`.

core/src/main/scala/cats/instances/function.scala
johnynek: I think it starts to look like a soup of parens, so I didn't write it that way.

Issue #2587, opened 2018-10-30
Stack-overflow using Eval.defer


johnynek:
partial stack here:

https://gist.github.com/johnynek/2adb46aecde687e2ab993c3cf9c2deb0

Seems to possibly be memoization related.

johnynek:
this is in 1.4.0.

cc @non

PR #3519, opened 2020-07-12
remove casts from Eval, fix stack overflow in Eval

johnynek:
similar to #3518 we use the 2.12 ability to do tailrec when the generic parameter changes to remove the casts.

Instead of using `List[Any => Eval[Any]]` we have to introduce a new type-aligned stack of functions, but I make this private.

This simplification leads to one optimization: if we have a finished `Memoize` in FlatMap we don't need to `loop(Now...` we can directly apply the flatMap function.

Also, I noticed perhaps the source of stack overflows I've seen in the past: #2587

### DIFF
--- a/core/src/main/scala/cats/Eval.scala
+++ b/core/src/main/scala/cats/Eval.scala
@@ -263,21 +263,6 @@ object Eval extends EvalInstances {
   }
 
   /**
-   * Advance until we find a non-deferred Eval node.
-   *
-   * Often we may have deep chains of Defer nodes; the goal here is to
-   * advance through those to find the underlying "work" (in the case
-   * of FlatMap nodes) or "value" (in the case of Now, Later, or
-   * Always nodes).
-   */
-  @tailrec private def advance[A](fa: Defer[A]): Eval[A] =
-    fa.thunk() match {
-      case call: Eval.Defer[A] =>
-        advance(call)
-      case other => other
-    }
-
-  /**
    * FlatMap is a type of Eval[A] that is used to chain computations
    * involving .map and .flatMap. Along with Eval#flatMap it
    * implements the trampoline that guarantees stack-safety.
@@ -338,7 +323,7 @@ object Eval extends EvalInstances {
               // by removing the Defer, we can nest defers,
               // so defer(defer(x)).flatMap(f) could mean c.start()
               // returns a Defer. We have to handle it here
-              loop(advance(call), Many(c.run, fs))
+              loop(call.thunk(), Many(c.run, fs))
             case mm @ Memoize(eval) =>
               mm.result match {
                 case Some(a) =>
@@ -353,7 +338,7 @@ object Eval extends EvalInstances {
               loop(c.run(xx.value), fs)
           }
         case call: Defer[A1] =>
-          loop(advance(call), fs)
+          loop(call.thunk(), fs)
         case m: Memoize[a] =>
           // a <:< A1
           m.result match {
--- a/core/src/main/scala/cats/Eval.scala
+++ b/core/src/main/scala/cats/Eval.scala
@@ -116,7 +116,7 @@ sealed abstract class Eval[+A] extends Serializable { self =>
  * This type should be used when an A value is already in hand, or
  * when the computation to produce an A value is pure and very fast.
  */
-final case class Now[A](value: A) extends Eval[A] {
+final case class Now[A](value: A) extends Eval.Leaf[A] {
   def memoize: Eval[A] = this
 }
 
@@ -134,7 +134,7 @@ final case class Now[A](value: A) extends Eval[A] {
  * by the closure) will not be retained, and will be available for
  * garbage collection.
  */
-final class Later[A](f: () => A) extends Eval[A] {
+final class Later[A](f: () => A) extends Eval.Leaf[A] {
   private[this] var thunk: () => A = f
 
   // The idea here is that `f` may have captured very large
@@ -167,7 +167,7 @@ object Later {
  * required. It should be avoided except when laziness is required and
  * caching must be avoided. Generally, prefer Later.
  */
-final class Always[A](f: () => A) extends Eval[A] {
+final class Always[A](f: () => A) extends Eval.Leaf[A] {
   def value: A = f()
   def memoize: Eval[A] = new Later(f)
 }
@@ -177,6 +177,13 @@ object Always {
 }
 
 object Eval extends EvalInstances {
+
+  /**
+   * A Leaf does not depend on any other Eval
+   * so calling .value does not trigger
+   * any flatMaps or defers
+   */
+  sealed abstract class Leaf[A] extends Eval[A]
 
   /**
    * Construct an eager Eval[A] value (i.e. Now[A]).
@@ -340,7 +347,7 @@ object Eval extends EvalInstances {
                   val nextFs = Many(c.run, fs)
                   loop(eval, Many(addToMemo(mm), nextFs))
               }
-            case xx =>
+            case xx: Leaf[c.Start] =>
               // xx must be Now, Later, Always, all of those
               // have safe .value:
               loop(c.run(xx.value), fs)
@@ -358,7 +365,7 @@ object Eval extends EvalInstances {
             case None =>
               loop[a](m.eval, Many[a, A1, A](addToMemo[a](m), fs))
           }
-        case x =>
+        case x: Leaf[A1] =>
           // Now, Later or Always don't have recursions
           // so they have safe .value:
           val a1 = x.value
--- a/core/src/main/scala/cats/Eval.scala
+++ b/core/src/main/scala/cats/Eval.scala
@@ -270,7 +270,7 @@ object Eval extends EvalInstances {
       case compute: Eval.FlatMap[A] =>
         new Eval.FlatMap[A] {
           type Start = compute.Start
-          val start: () => Eval[Start] = () => compute.start()
+          val start: () => Eval[Start] = compute.start
           val run: Start => Eval[A] = s => advance1(compute.run(s))
         }
       case other => other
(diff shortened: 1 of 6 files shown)


core/src/main/scala/cats/Eval.scala
```diff
@@ -318,52 +318,61 @@ object Eval extends EvalInstances {
...
                 case Some(a) =>
-                  loop(Now(a), c.run.asInstanceOf[C] :: fs)
+                  loop(nextFs.first(a), nextFs.rest)
                 case None =>
-                  loop(eval, addToMemo(mm.asInstanceOf[M]) :: c.run.asInstanceOf[C] :: fs)
+                  loop(eval, Many(addToMemo(mm), nextFs))
               }
             case xx =>
+              // note: xx could be a Defer so calling .value
+              // could blow the stack?
```
johnynek, line 351: this looks unsafe to me. I don't see why we don't need to check for `Defer` here. I think this could be the cause of #2587

If this is safe, I think we need a comment as to why `xx.value` is okay.

core/src/main/scala/cats/Eval.scala
```diff
@@ -318,52 +318,61 @@ object Eval extends EvalInstances {
...
-              loop(eval, addToMemo(m.asInstanceOf[M]) :: fs)
+              loop[a](m.eval, Many[a, A1, A](addToMemo[a](m), fs))
           }
         case x =>
+          // Now, Later or Always don't have recursions
           fs match {
-            case f :: fs => loop(f(x.value), fs)
-            case Nil     => x.value
+            case Many(f, fs) => loop(f(x.value), fs)
+            case Ident(ev)   => ev(x.value)
```
johnynek, line 371: the calls to `x.value` here are safe because `Now`, `Later` and `Always` don't do any recursion on Eval.

core/src/main/scala/cats/Eval.scala
```diff
@@ -318,52 +318,61 @@ object Eval extends EvalInstances {
...
           }
       }
 
-    loop(e.asInstanceOf[L], Nil).asInstanceOf[A]
+    loop(e, Ident(implicitly[A <:< A]))
```
johnynek, line 375: here we have 1 allocation more than we did before: `Ident` where as before we did `Nil`. But since this is one allocation per call to `.value`, I think this is well worth the cost considering how many allocations we already make (1 per depth at least to build up the user level stack).

core/src/main/scala/cats/Eval.scala
```diff
@@ -318,52 +318,61 @@ object Eval extends EvalInstances {
...
           c.start() match {
-            case cc: FlatMap[_] =>
-              loop(cc.start().asInstanceOf[L], cc.run.asInstanceOf[C] :: c.run.asInstanceOf[C] :: fs)
+            case cc: FlatMap[c.Start] =>
+              loop(cc.start(), Many(cc.run, nextFs))
             case mm @ Memoize(eval) =>
               mm.result match {
                 case Some(a) =>
-                  loop(Now(a), c.run.asInstanceOf[C] :: fs)
+                  loop(nextFs.first(a), nextFs.rest)
```
johnynek, line 345: here is an optimization: since we know we already have a `Many`, we don't need to recurse.

core/src/main/scala/cats/Eval.scala
johnynek: @non do you know why we wouldn't do `loop(xx, Many(c.run, fs))` when we have a Defer?

johnynek:
Here are the benchmarks before and after:
```
benchmarks 2.12:                                                                  
                                                
master branch:                                                          
[info] Benchmark                    Mode  Cnt      Score      Error  Units    
[info] TrampolineBench.eval        thrpt    5  15266.439 ± 1106.948  ops/s    
                
this PR:                            
[info] Benchmark              Mode  Cnt      Score      Error  Units    
[info] TrampolineBench.eval  thrpt    5  14891.105 ± 4289.588  ops/s  
```

core/src/main/scala/cats/Eval.scala
johnynek: Okay, so `defer(e).flatMap(f)` is safe because `.flatMap` checks for `Defer` and uses the thunk as `start`, BUT `defer(defer(e)).flatMap(f)` isn't safe because flatMap only unwraps one layer of `Defer` and then we wind up calling `.value` on a defer which blows the stack.

I reproed this failure mode in a test.

Should I combine with this code for the fix or make a follow up PR?

johnynek:
I went ahead and added the fix for stack safety here.

BTW: this, IMO, shows the win of removing the casts. Without the casts, it was clearer to me to follow the logic and it was easier to see what was going on.

johnynek:
fixes #2587

core/src/main/scala/cats/Eval.scala
```diff
@@ -263,26 +263,13 @@ object Eval extends EvalInstances {
...
-    fa match {
+  @tailrec private def advance[A](fa: Defer[A]): Eval[A] =
+    fa.thunk() match {
       case call: Eval.Defer[A] =>
-        advance(call.thunk())
-      case compute: Eval.FlatMap[A] =>
-        new Eval.FlatMap[A] {
-          type Start = compute.Start
-          val start: () => Eval[Start] = () => compute.start()
-          val run: Start => Eval[A] = s => advance1(compute.run(s))
```
johnynek, line 274: @non can you see why we did this? Can you write a test that shows the need? Removing it is a bit more efficient since we don't add a level of stack to all FlatMaps we call, BUT note, we should only really need to call advance when we know it is a Defer (note I change the argument type).

core/src/main/scala/cats/Eval.scala
```diff
@@ -318,52 +305,70 @@ object Eval extends EvalInstances {
...
       curr match {
-        case c: FlatMap[_] =>
+        case c: FlatMap[A1] =>
           c.start() match {
-            case cc: FlatMap[_] =>
-              loop(cc.start().asInstanceOf[L], cc.run.asInstanceOf[C] :: c.run.asInstanceOf[C] :: fs)
+            case cc: FlatMap[c.Start] =>
+              val nextFs = Many(c.run, fs)
+              loop(cc.start(), Many(cc.run, nextFs))
+            case call: Defer[c.Start] =>
```
johnynek, line 329: this branch needs to be added. Note I added a test to the PR the exercises this. Without this change we blow the stack.

johnynek:
@non can you remind us here why we have `Defer` at all? Why not just:
```scala
def defer[A](ea: => Eval[A]): Eval[A] =
  Unit.flatMap(_ => ea)
```
all the tests seem to pass and it would, I assume, improve performance of code using Defer since we have fewer branches to test.

Also, if we did this, in `FlatMap` we could have `val start: Eval[Start]` rather than `val start: () => Eval[Start]` so we avoid having to invoke another `Function0` there.

We probably can't remove `Defer` as a class, but we can avoid using it and maybe the branch predictor would pick up on that fact. But our current benchmarks don't exercise `defer` so we'd have to spend effort on benchmarking to make this case.

core/src/main/scala/cats/Eval.scala
```diff
@@ -178,6 +178,13 @@ object Always {
 
 object Eval extends EvalInstances {
 
+  /**
+   * A Leaf does not depend on any other Eval
+   * so calling .value does not trigger
+   * any flatMaps or defers
+   */
+  sealed abstract class Leaf[A] extends Eval[A]
```
johnynek, line 186: with this, we can use `x: Leaf[A]` below to make it clear calling `.value` is safe. This helps the reader review the `def evaluate` method and see that it is safe (we only call .value on Leafs).

it doesn't impact the performance (hopefully scala removes the type check when it can statically prove that it would always pass, and since this was in the last position of the match that is the case currently).

johnynek:
@travisbrown would you have time to look at this one? I know you have a lot on your plate.

travisbrown:
@johnynek Sure, I'll take a look today.

core/src/main/scala/cats/Eval.scala
```diff
@@ -318,52 +312,70 @@ object Eval extends EvalInstances {
...
 
-    @tailrec def loop(curr: L, fs: List[C]): Any =
+    @tailrec def loop[A1](curr: Eval[A1], fs: FnStack[A1, A]): A =
       curr match {
-        case c: FlatMap[_] =>
+        case c: FlatMap[A1] =>
           c.start() match {
-            case cc: FlatMap[_] =>
-              loop(cc.start().asInstanceOf[L], cc.run.asInstanceOf[C] :: c.run.asInstanceOf[C] :: fs)
+            case cc: FlatMap[c.Start] =>
```
johnynek, line 333: I'm not entirely sure why we don't do:
```scala
case c: FlatMap[A1] =>
  loop(c.start(), Many(c.run, fs))
```
and let the rest of the code run.

That would be correct, but I assume @non inlined this because for a FlatMap we know that we have at least one item on the stack and if we do my suggestion above, we allocate a `Many` just to pattern match on it in many cases and discard it. So, by nesting the pattern match on the `c.start()` we often can often sometimes avoid the allocation and pattern match.

If @non has time to chime in, it would be appreciated, I could update a comment to make it clearer.

johnynek:
If I were designing Eval today I would make all subclasses private. It's a bit strange that the comments say that the details of Eval are opaque yet all the subclasses are public.

That said Leaf is somewhat useful to users: it is a case where calling .value is safe at any location.

For instance if you were doing a loop where you will call .value at the end you could conceivably match on Leaf and call .value at the beginning, e.g. some map2Eval implementations could possibly make use of it.

Note, they could have already matched on Now, Later or Always which are public but this unifies those three into a single subtype.

For me, I saw that there were some internal .value calls but I wasn't sure why that was safe: now you can see very clearly. I think that's a win. I even thought of giving Leaf a safeValue method to make it more clear that we only call that inside the evaluate loop. But I ultimately didn't.

Anyway, thanks for being flexible and accepting a design you didn't think was your first choice.

PR #3524, opened 2020-07-18
A method `StateT.fromState` turning `State[A, F[B]]` into `StateT[F,A, B]` is added.

akopich:


akopich:
@yilinwei , thank you for the discussion and encouragement.

core/src/main/scala/cats/data/IndexedStateT.scala
```diff
@@ -199,6 +199,17 @@ private[data] trait CommonStateTConstructors {
 
   def get[F[_], S](implicit F: Applicative[F]): IndexedStateT[F, S, S, S] =
     IndexedStateT(s => F.pure((s, s)))
+
+  /**
+   * Turn `State[A, F[B]]` into `StateT[F, A, B]`
+   */
+  def fromState[F[_]: Applicative, A, B](s: State[A, F[B]]): StateT[F, A, B] =
+    s.transformF { eval =>
+      import cats.implicits._
```
travisbrown, line 208: This is a minor thing, but we've been moving away from using Cats syntax in the core Cats implementation, in part because some cases had been causing issues on Dotty, but also because avoiding it makes runtime and compile-time performance a little more predictable. In this case I think the desugared version doesn't lose much readability:

```scala
def fromState[F[_], A, B](s: State[A, F[B]])(implicit F: Applicative[F]): StateT[F, A, B] =
  s.transformF { eval =>

    val (a, fb) = eval.value
    F.map(fb)((a, _))
  }
```

This definitely shouldn't block the PR, though, and I'd be happy to see it merged as-is.

akopich: Thanks for pointing that out. I'll desugar that. At least, for the sake of consistency with the rest of the file.

PR #3533, opened 2020-07-27
Improve invariants and performance in Chain

johnynek:
I realized that if we keep a suptype of Chain to note that items are non-empty, we can (at no cost, I think) simplify the code and make it safer.

There is a minor bit of ugliness in Append because to make it binary compatible, we need to put back some methods that were there before. I went ahead and did this to satisfy mima even though these classes are private to data.

Side note: unless we really need to expose classes like this, I'm a big fan of keeping them totally private.

also is a possible replacement for #3373


johnynek:
I also avoid writing `case Empty =>` since that triggers a call to `def equals` which triggers a call to `iterator`. I managed to hit an infinite loop in one version of the code, but no reason to hit that for a sealed private class, we know Empty by exclusion and it is the least common case I would expect.

johnynek:
While looking at this I found a performance improvement by not using an internal ArrayBuffer of 16% for large Chains. This is for iterator, which is used broadly. foldLeft exercises iterator.
```
This change:

[info] Benchmark                           Mode  Cnt         Score        Error  Units
[info] ChainBench.foldLeftLargeCatenable  thrpt    5        56.035 ±      0.661  ops/s
[info] ChainBench.foldLeftLargeChain      thrpt    5       199.145 ±      1.080  ops/s
[info] ChainBench.foldLeftLargeList       thrpt    5       158.537 ±      7.086  ops/s
[info] ChainBench.foldLeftLargeOldChain   thrpt    5       148.596 ±      1.560  ops/s
[info] ChainBench.foldLeftLargeVector     thrpt    5        77.262 ±      0.680  ops/s
[info] ChainBench.foldLeftSmallCatenable  thrpt    5  11846407.099 ±  22880.990  ops/s
[info] ChainBench.foldLeftSmallChain      thrpt    5  48157018.969 ± 246883.975  ops/s
[info] ChainBench.foldLeftSmallList       thrpt    5  55998179.055 ± 241976.102  ops/s
[info] ChainBench.foldLeftSmallOldChain   thrpt    5  55485668.350 ±  90607.034  ops/s
[info] ChainBench.foldLeftSmallVector     thrpt    5  14805536.562 ± 749006.508  ops/s


master:
[info] Benchmark                           Mode  Cnt         Score        Error  Units
[info] ChainBench.foldLeftLargeCatenable  thrpt    5        57.285 ±      1.891  ops/s
[info] ChainBench.foldLeftLargeChain      thrpt    5       170.942 ±      2.281  ops/s
[info] ChainBench.foldLeftLargeList       thrpt    5       159.709 ±      4.833  ops/s
[info] ChainBench.foldLeftLargeOldChain   thrpt    5       145.871 ±      7.981  ops/s
[info] ChainBench.foldLeftLargeVector     thrpt    5        76.742 ±      1.634  ops/s
[info] ChainBench.foldLeftSmallCatenable  thrpt    5  11798663.388 ± 606629.280  ops/s
[info] ChainBench.foldLeftSmallChain      thrpt    5  47750078.331 ± 473361.638  ops/s
[info] ChainBench.foldLeftSmallList       thrpt    5  56818234.798 ± 420088.798  ops/s
[info] ChainBench.foldLeftSmallOldChain   thrpt    5  63018070.265 ± 515461.221  ops/s
[info] ChainBench.foldLeftSmallVector     thrpt    5  14840988.935 ± 381582.742  ops/s
```

@LukaJCB can you look again?

johnynek:
@barambani could you take another look (merged with master, a few minor cleanups).

barambani:
LGTM on green. I think the failures are due to the update to scala 2.13.3. There are few missing parentheses on 'next' calls that now makes the build fail.

johnynek:
Thanks @barambani I think I fixed it. Sorry for so many revisions here. :/ one more :+1:

PR #3535, opened 2020-07-27
Improve traverseViaChain API

johnynek:
follow up to #3521 

In that PR, I added an API that consumes an Iterator. That is a bummer to have an API with a mutable arg, even if most people know that iterators must be considered with ownership semantics.

In fact, the first thing we do is materialize an IndexedSeq, so, why not just take an IndexedSeq and have callers materialize one if needed. This also removes duplication since Vector can share the same implementation now.

As an added bonus, you can use `Chain.traverseViaChain(0 until 100000)(f)` since Range is an IndexedSeq as well, which is nice.

It would be nice to get this in before 2.2.0 so we aren't stuck with the mutable API.

### DIFF
--- a/core/src/main/scala/cats/data/Chain.scala
+++ b/core/src/main/scala/cats/data/Chain.scala
@@ -629,13 +629,11 @@ object Chain extends ChainInstances {
   def apply[A](as: A*): Chain[A] =
     fromSeq(as)
 
-  def traverseViaChain[G[_], A, B](iter: Iterator[A])(f: A => G[B])(implicit G: Applicative[G]): G[Chain[B]] =
-    if (!iter.hasNext) G.pure(Chain.nil)
+  def traverseViaChain[G[_], A, B](as: IndexedSeq[A])(f: A => G[B])(implicit G: Applicative[G]): G[Chain[B]] =
+    if (as.isEmpty) G.pure(Chain.nil)
     else {
       // we branch out by this factor
       val width = 128
-      val as = collection.mutable.Buffer[A]()
-      as ++= iter
       // By making a tree here we don't blow the stack
       // even if the List is very long
       // by construction, this is never called with start == end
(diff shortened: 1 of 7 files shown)


johnynek:
@travisbrown I know I'm making your life a pain here with these PRs. I'm sorry. I do hope we can get this in and not have mima force us to keep an unneeded mutable API.

travisbrown:
No, this is great. Happy to review in the morning.

johnynek:
Okay, the 2.13 build brought up an interesting issue:

in 2.12, `IndexedSeq` is `collection.IndexedSeq` which may be mutable, which accepts `ArrayBuffer`. But that isn't safe for a lazy Applicative, because the computation may run later and the caller could mutate the input still. So, I added a wrapper which just makes an immutable.IndexedSeq from a mutable one when the caller promises not to mutate. Then, the public API can require an immutable IndexedSeq, which makes the contract clear.

This is probably a minor performance hit, but I imagine very very minimal (since the real meat of what traverse is doing is still the Eval and the applicative context I assume.

I think I'm happy with this now if you can take a look @travisbrown

johnynek:
@travisbrown what do you think about this one?

kernel/src/main/scala/cats/kernel/instances/StaticMethods.scala
```diff
@@ -17,6 +19,22 @@ object StaticMethods extends cats.kernel.compat.HashCompat {
...
+   * When you "own" this m, and will not mutate it again, this
+   * is safe to call. It is unsafe to call this, then mutate
+   * the original collection.
+   *
+   * You are giving up ownership when calling this method
+   */
+  def wrapMutableIndexedSeq[A](m: mutable.IndexedSeq[A]): ImIndexedSeq[A] =
+    new WrappedIndexedSeq(m)
+
+  private[kernel] class WrappedIndexedSeq[A](m: mutable.IndexedSeq[A]) extends ImIndexedSeq[A] {
```
travisbrown, line 32: I can never remember exactly how this works inside objects, but is there any reason to make this package-private (which is just public in the JVM API) instead of just private (which I think should be properly JVM-private)?

core/src/main/scala/cats/data/Chain.scala
```diff
@@ -508,11 +500,11 @@ sealed abstract class Chain[+A] {
     val builder = new StringBuilder("Chain(")
     var first = true
 
-    foreach { a =>
+    foreachUntil { a =>
```
travisbrown, line 503: This is just miscellaneous clean-up, right, since `foreach` isn't used anywhere else?

johnynek: yes. foreach is used exactly once and is private. Just use foreachUntil (or iterator).

kernel/src/main/scala/cats/kernel/instances/StaticMethods.scala
johnynek: not that I know of, I just copied the idiom below. Should I fix that up?

PR #3540, opened 2020-07-29
Remove traverseForEither in Traverse companion object

LukaJCB:
I'm not quite sure how this got here, but it's messing up things in dotty 0.25 so we should get rid of it asap.

### DIFF
--- a/core/src/main/scala/cats/Traverse.scala
+++ b/core/src/main/scala/cats/Traverse.scala
@@ -133,7 +133,6 @@ import scala.annotation.implicitNotFound
 }
 
 object Traverse {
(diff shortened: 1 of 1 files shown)


LukaJCB:
Seems like this was introduced here: https://github.com/typelevel/cats/commit/56d94da7120e778e7565af213fc680fc9dc214c7#diff-dfc66c84b778099ad28a81e9b0ece505

@travisbrown I'm assuming this was a mistake, but maybe there was some motivation I'm unaware of?

barambani:
I guess it's pretty safe to merge this in.

travisbrown:
Hmm, just saw this PR, but yes, looks like a mistake to me.

PR #3545, opened 2020-08-03
Enable breakout in functions nonEmptyTraverse_ and nonEmptySequence_

takayahilton:
ref: #3015 #3498

### DIFF
--- a/core/src/main/scala/cats/Reducible.scala
+++ b/core/src/main/scala/cats/Reducible.scala
@@ -182,7 +182,7 @@ import scala.annotation.implicitNotFound
    * the traversal.
    */
   def nonEmptyTraverse_[G[_], A, B](fa: F[A])(f: A => G[B])(implicit G: Apply[G]): G[Unit] =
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/Reducible.scala
```diff
@@ -182,7 +182,7 @@ import scala.annotation.implicitNotFound
    * the traversal.
    */
   def nonEmptyTraverse_[G[_], A, B](fa: F[A])(f: A => G[B])(implicit G: Apply[G]): G[Unit] =
-    G.void(reduceLeftTo(fa)(f)((x, y) => G.map2(x, f(y))((_, b) => b)))
+    G.void(reduceRightTo(fa)(f)((x, y) => G.map2Eval(f(x), y)((_, b) => b)).value)
```
johnynek, line 185: should we not void after each `f`? couldn't that be a win rather than voiding at the end?

`f1 = f.andThen(G.void(_))`

PR #3538, opened 2020-07-28
Preliminary Munit port

LukaJCB:
Should resolve #3534

PR #3546, opened 2020-08-03
Fix alleycats-tests on Scala.js

joroKr21:
* Add them to the PR build
  * Some minor build tweaks

I guess accidentally broken by #3538 and went unnoticed because it was not part of the build.


joroKr21:
@barambani sorry for the new commit but it was easier this way to avoid conflicts. JS tests are much faster now!

PR #3548, opened 2020-08-03
Test freeJS with FastOptStage to save some CO2

joroKr21:


joroKr21:
Hmm I didn't see any difference - but now we should be able to run `all` JS tests concurrently - I will try that next.

joroKr21:
It could also depend on the node version / OS - we use a really old one on Travis (4)

djspiewak:
Btw, doing a bit of digging… I think this setting is defaulted.

joroKr21:
> but now we should be able to run all JS tests concurrently - I will try that next.

Done in #3546 - it works

PR #3551, opened 2020-08-05
Add FlatMap.ifElseM

mtomko:
The actual implementation is from @djspiewak:

https://gist.github.com/djspiewak/1b92a6e338f4e1537692e748c54b9743

We discussed this function briefly on the cats-effect gitter channel: 

https://gitter.im/typelevel/cats-effect?at=5f297e4314c413356f56d230

As indicated in the discussion there, it wasn't totally clear where it belonged. There is no package object in `cats.syntax.flatMap` and I'm not sure adding a package object now makes sense, with Scala 3 on the horizon. This is my first code contribution to cats and I'm still learning the code layout so I'm happy to move things around to suit people's preferences.

Also, I did see some failures locally in the alleycats JS tests, but a clean checkout of the master branch also had those, and I don't see how I could have broken them or made the situation any worse with the changes I've made here. I'm curious to see what the CI build does - if it has problems I'll investigate further.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,9 +34,7 @@ jobs:
       stage: test
       name: Js tests 2.12
       env: TEST="JS tests"
-      # http://austinpray.com/ops/2015/09/20/change-travis-node-version.html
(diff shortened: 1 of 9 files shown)


mtomko:
Hmm, in my attempt to update my local fork, I seem to have picked up someone else's changes. I can make a new fork and re-apply my changes if this is a problem.

mtomko:
Closing in favor of https://github.com/typelevel/cats/pull/3553 which has a cleaner git log

PR #3552, opened 2020-08-06
Port tests to Dotty

LukaJCB:


.scalafmt.conf
```diff
@@ -10,3 +10,4 @@ docstrings = JavaDoc
 newlines.afterCurlyLambda = preserve
 docstrings.style = Asterisk
 docstrings.oneline = unfold
+project.excludeFilters = [ "core/src/main/scala-3.x" ]
```
LukaJCB, line 13: scalafmt can't deal with dotty syntax yet.

core/src/main/scala-3.x/src/main/scala/cats/syntax/MonadOps.scala
```diff
@@ -0,0 +1,12 @@
+package cats.syntax
+
+import cats.{Monad, Alternative}
+
+final class MonadOps[F[_], A](private val fa: F[A]) extends AnyVal {
```
LukaJCB, line 5: Had to split this out due to https://github.com/lampepfl/dotty/issues/9480

core/src/main/scala/cats/Invariant.scala
```diff
@@ -122,7 +122,6 @@ object Invariant extends ScalaVersionSpecificInvariantInstances with InvariantIn
   implicit def catsFlatMapForSortedMap[K]: FlatMap[SortedMap[K, *]] =
     cats.instances.sortedMap.catsStdInstancesForSortedMap[K]
   implicit def catsBimonadForFunction0: Bimonad[Function0] = cats.instances.function.catsStdBimonadForFunction0
-  implicit def catsMonadForFunction1[I]: Monad[I => *] = cats.instances.function.catsStdMonadForFunction1[I]
```
LukaJCB, line 125: Moved this down because dotty thinks `Monad[List]` and `Monad[Function1[Int, *]]` are ambiguous

core/src/main/scala/cats/Invariant.scala
```diff
@@ -280,29 +279,35 @@ object Invariant extends ScalaVersionSpecificInvariantInstances with InvariantIn
...
-private trait TupleInstances0 extends TupleInstances1 {
+private[cats] trait InvariantInstances1 extends InvariantInstances2 {
+  implicit def catsMonadForFunction1[I]: Monad[I => *] = cats.instances.function.catsStdMonadForFunction1[I]
+}
+
+private trait InvariantInstances2 extends TupleInstances0 {
   implicit def catsCommutativeFlatMapForTuple2[X](implicit X: CommutativeSemigroup[X]): CommutativeFlatMap[(X, *)] =
     cats.instances.tuple.catsStdCommutativeFlatMapForTuple2[X]
+
+  implicit def catsApplicativeForArrow[F[_, _], A](implicit F: Arrow[F]): Applicative[F[A, *]] =
```
LukaJCB, line 301: Same with this one

core/src/main/scala/cats/Traverse.scala
```diff
@@ -133,7 +133,7 @@ import scala.annotation.implicitNotFound
 }
 
 object Traverse {
-  implicit def catsTraverseForEither[A]: Traverse[Either[A, *]] = cats.instances.either.catsStdInstancesForEither[A]
+  //implicit def catsTraverseForEither[A]: Traverse[Either[A, *]] = cats.instances.either.catsStdInstancesForEither[A]
```
LukaJCB, line 136: See https://github.com/typelevel/cats/pull/3540

tests/src/test/scala-2.13+/cats/tests/ScalaVersionSpecific.scala
```diff
@@ -10,6 +10,7 @@ import cats.syntax.parallel._
 import cats.syntax.traverse._
 import cats.syntax.eq._
 import org.scalacheck.Prop._
+import cats.catsInstancesForId
```
LukaJCB, line 13: Id instances are no longer automatically in scope in dotty.

tests/src/test/scala/cats/tests/IndexedStateTSuite.scala
```diff
@@ -506,7 +509,7 @@ class IndexedStateTSuite extends CatsSuite {
     Alternative[IndexedStateT[ListWrapper, Int, Int, *]]
     Applicative[IndexedStateT[ListWrapper, Int, Int, *]]
     Apply[IndexedStateT[ListWrapper, Int, Int, *]]
-    Functor[IndexedStateT[ListWrapper, Int, Int, *]]
+    //Functor[IndexedStateT[ListWrapper, Int, Int, *]]
```
LukaJCB, line 512: Ambiguous here because both `G` and `F` are `Functor`s

tests/src/test/scala/cats/tests/OptionTSuite.scala
```diff
@@ -536,7 +536,7 @@ class OptionTSuite extends CatsSuite {
     Traverse[OptionT[List, *]]
 
     implicit val T: Traverse[ListWrapper] = ListWrapper.traverse
-    implicit val M: Monad[ListWrapper] = ListWrapper.monad
+    //implicit val M: Monad[ListWrapper] = ListWrapper.monad
```
LukaJCB, line 539: Same here

tests/src/test/scala/cats/tests/ParallelSuite.scala
```diff
@@ -88,34 +83,37 @@ class ParallelSuite extends CatsSuite with ApplicativeErrorForEitherTest with Sc
     }
   }
 
-  type ListTuple2[A, B] = List[(A, B)]
+  case class ListTuple2[A, B](value: List[(A, B)])
```
LukaJCB, line 86: Had to turn this into a case class for it to work

tests/src/test/scala/cats/tests/ReducibleSuite.scala
```diff
@@ -131,14 +131,14 @@ class ReducibleSuiteAdditional extends CatsSuite {
     val n = 100000
     val xs = NES(0, Stream.from(1))
 
-    assert(xs.reduceMapM(i => if (i < n) Right(i) else Left(i)) === Left(n))
+    assert(xs.reduceMapM(i => if (i < n) i.asRight[Int] else i.asLeft[Int]) === Left(n))
```
LukaJCB, line 134: See https://github.com/lampepfl/dotty/issues/9479

tests/src/test/scala/cats/tests/ApplicativeErrorSuite.scala
```diff
@@ -61,7 +61,7 @@ class ApplicativeErrorSuite extends CatsSuite {
     assert(compileErrors("e2.attemptNarrow[Num]").nonEmpty)
 
     val e3: Either[List[T[String]], Unit] = List(Str).asLeft[Unit]
-    assert(compileErrors("e3.attemptNarrow[List[Str.type]]").nonEmpty)
+    //assertEquals(compileErrors("e3.attemptNarrow[List[Str.type]]"), "")
```
LukaJCB, line 64: This is an odd one, because when I try it in the console it gives me an error but munit returns an empty string here

tests/src/test/scala/cats/tests/ExtraRegressionSuite.scala
```diff
@@ -1,25 +0,0 @@
...
-
-import cats.Show
-import ExtraRegressionSuite._
-
-class ExtraRegressionSuite extends CatsSuite {
-
-  /**
-   * Placed here to work around scala/bug#6260 on scala 2.10
-   */
-  test("#1802 duplicated method name") {
```
LukaJCB, line 11: Deleted this for now, since it was giving me an error and seems like this is pretty old

tests/src/test/scala/cats/tests/RegressionSuite.scala
```diff
@@ -103,7 +103,7 @@ class RegressionSuite extends CatsSuite with ScalaVersionSpecificRegressionSuite
 
   test("#500: foldMap - traverse consistency") {
     assert(
-      List(1, 2, 3).traverse(i => Const.of[List[Int]](List(i))).getConst == List(1, 2, 3).foldMap(List(_))
+      List(1, 2, 3).traverse(i => Const[List[Int], List[Int]](List(i))).getConst == List(1, 2, 3).foldMap(List(_))
```
LukaJCB, line 106: Dotty was inferring `Int => Int`, I'll try to reach out to the dotty team on how to get this working

core/src/main/scala-3.x/src/main/scala/cats/syntax/MonadOps.scala
LukaJCB: I actually don't know how dotty compiles this down. Is `explicitly` still a thing? Do you have a suggestion on how to fix this given the limitation above?

free/src/main/scala/cats/free/Free.scala
```diff
@@ -80,12 +80,12 @@ sealed abstract class Free[S[_], A] extends Product with Serializable {
   final def foldStep[B](
     onPure: A => B,
     onSuspend: S[A] => B,
-    onFlatMapped: ((S[X], X => Free[S, A]) forSome { type X }) => B
+    onFlatMapped: ((S[Any], Any => Free[S, A])) => B
```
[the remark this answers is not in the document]
LukaJCB: This is bincompat due to type erasure, but I believe having separate sources here is probably a good idea 👍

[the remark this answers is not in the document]
djspiewak: It's definitely bincompat, I was mostly just concerned about the loss of type safety in a public-facing API.

core/src/main/scala-3.x/src/main/scala/cats/syntax/MonadOps.scala
djspiewak: I would summon @smarter actually if I hit this. Probably worth looking at the bytecode quickly to see what the parameter order is. Can we simply put the `using` block second?

smarter: It's source-compatible (to explicitly pass parameters to a using block, you have to write `using` at use-site in front of the block too). I suggested putting the using block first because doing the implicit search requires instantiating type variables which in turns means that in `increment.whileM_(inspect(i => i > 4)).run(3)` we can figure out that `i` needs to be `Int`, see https://github.com/lampepfl/dotty/issues/9480#issuecomment-668545232 for the full explanation.

djspiewak: Ahhh, I see. Okay I'm in favor, though I think it might be best if we also do this for all of the other Ops classes.

smarter: That would increase the maintenance burden though. But if you did want to rewrite all the ops classes for dotty, then you might as well use extension syntax which would look more natural:
```scala
extension [F[_], A](fa: F[A])(using M: Monad[F]) {
  def whileM_(p: F[Boolean]): F[Unit] = ???
  def untilM(...)
}
```

LukaJCB:
Seeing a new failure now, opened a bug report at https://github.com/lampepfl/dotty/issues/9793

core/src/main/scala-3.x/src/main/scala/cats/syntax/MonadOps.scala
djspiewak: I'm just concerned that we're going to end up hitting something like this again in the future, but once we release on Dotty, we can't futz with the binary compatibility any further.

smarter: Well, you can break BC every time we break BC which is every six weeks until 3.0 is out probably, but afterwards it gets tricky yeah :) (though I guess you can always play trick with something like a package-protected overload to preserve BC?)

LukaJCB: Are we promising binary compatibility for dotty artifacts? Should we? (We certainly don't test it yet)

djspiewak: I don't think we're promising BC until Scala 3 goes final. Once it does though, we 100% should be promising it.

larsrh:
This appears to be fixed in lampepfl/dotty#9937, which is scheduled for 3.0.0-M1, so I guess we'll have to wait for that.

larsrh:
I just realized that the milestone date is end of year :grimacing: Maybe let's try this instead: 300a9cd02bb932dbb638aa1f873d6dfb2b5298b8

smarter:
M1 should be out next week, the milestone date hasn't been updated

larsrh:
I made a new PR that is up-to-date: #3636. Obviously we can continue the discussion there.

PR #3553, opened 2020-08-06
add ifElseM

mtomko:
The actual implementation is from @djspiewak:

https://gist.github.com/djspiewak/1b92a6e338f4e1537692e748c54b9743

We discussed this function briefly on the cats-effect gitter channel:

https://gitter.im/typelevel/cats-effect?at=5f297e4314c413356f56d230

As indicated in the discussion there, it wasn't totally clear where it belonged. There is no package object in cats.syntax.flatMap and I'm not sure adding a package object now makes sense, with Scala 3 on the horizon. This is my first code contribution to cats and I'm still learning the code layout so I'm happy to move things around to suit people's preferences.

Also, I did see some failures locally in the alleycats JS tests, but a clean checkout of the master branch also had those, and I don't see how I could have broken them or made the situation any worse with the changes I've made here. I'm curious to see what the CI build does - if it has problems I'll investigate further.


johnynek:
PS: I can see why the the ergonomics is nicer on the object (in fact, we explored this pattern many years ago, but cats never seemed to like it, it makes type inference better. See e.g.: https://github.com/typelevel/cats/blob/a26f1612f58d8f2ea091d0fbeb3136bd5e69b20f/kernel/src/main/scala/cats/kernel/Monoid.scala#L98 which we copied from algebird. Being able to do `Monoid.combine(a, b)` is great for type inference, but copying each method to the companion object is a bit of a pain, so I eventually buckled and felt that `Monoid[Int].combine` was a fine compromise in general. When the type is really ugly, that isn't so great of course).

mtomko:
Will do, running tests now...

core/src/main/scala/cats/FlatMap.scala
```diff
@@ -202,6 +202,33 @@ import scala.annotation.implicitNotFound
...
+   * scala> import cats._
+   * scala> FlatMap.ifElseM(Eval.later(false) -> Eval.later(1), Eval.later(true) -> Eval.later(2))(Eval.later(5)).value
+   * res0: Int = 2
+   * }}}
+   *
+   * Based on a [[https://gist.github.com/1b92a6e338f4e1537692e748c54b9743 gist]] by Daniel Spiewak.
+   * @see See [[https://gitter.im/typelevel/cats-effect?at=5f297e4314c413356f56d230]] for the discussion.
+   */
+  def ifElseM[F[_]: FlatMap, A](branches: (F[Boolean], F[A])*)(els: F[A]): F[A] = {
+    def loop(branches: List[(F[Boolean], F[A])]): F[A] =
```
johnynek, line 220: I thought I left this comment but I don't see it now.

If we have a long list of branches, this will stack overflow. We can use tailRecM to address that if we work on Monad:
```scala
type Branches = List[(F[Boolean], F[A])]
def step(branches: Branches): F[Either[Branches, A]] =
  branches match {
    case (cond, conseq) :: tail =>
      cond.flatMap { b => if (b) conseq.map(Right(_)) else pure(Left(tail)) }
    case Nil => els.map(Right(_))
}

tailRecM(branches)(step(_))
```

mtomko: Thanks for pointing that out. I have a couple of questions. I'm not sure how to spot when a `flatMap` isn't stack-safe - is it dependent on the underlying monad's own stack-safety or is there something I can see in the chain of calls to `flatMap` itself?

Also, if simply using `tailRecM` is the solution, there is a `tailRecM` for `FlatMap`. I'm not averse to moving this code over to `Monad` but if we don't need the full power of `Monad` for stack-safety, then I guess I'd like to save the effort.

Thanks again for your comments and suggestions!

mtomko: Ah, I just reread the cats FAQ on `tailRecM`. So it's nothing particular to this implementation, it depends on the underlying `Monad`, and using `tailRecM` avoids that problem. I'm going to reframe the implementation in terms of `tailRecM`.

mtomko: And, now I think I see why it has to be on `Monad`...

mtomko:
Just a wild thought, too, we could call this `condM` as a nod to our FP roots (Lisp's `cond` form)... It almost has that feel.

mtomko:
@johnynek thanks so much for your help. I'm digging into some more resources on stack safety now - I need to learn a few things before I can properly understand the Phil Freeman paper but I think it'll be good for me to build that background.

I think ultimately the shape of this method isn't quite the same as `cond` because the `els` has to be in a separate argument list. I'll just stick with `ifElseM` for that reason and for the reason you suggested (approachability).

PR #3554, opened 2020-08-07
Order for Ior

TimWSpence:

### DIFF
--- a/core/src/main/scala/cats/data/Ior.scala
+++ b/core/src/main/scala/cats/data/Ior.scala
@@ -763,7 +763,7 @@ sealed abstract private[data] class IorInstances extends IorInstances0 {
       }
   }
 
-  implicit def catsDataEqForIor[A: Order, B: Order]: Order[A Ior B] =
+  implicit def catsDataOrderForIor[A: Order, B: Order]: Order[A Ior B] =
     new Order[A Ior B] {
 
       def compare(x: Ior[A, B], y: Ior[A, B]): Int = x.compare(y)
--- a/core/src/main/scala/cats/data/Ior.scala
+++ b/core/src/main/scala/cats/data/Ior.scala
@@ -708,6 +708,17 @@ sealed abstract class Ior[+A, +B] extends Product with Serializable {
       (a, b) => that.fold(a2 => false, b2 => false, (a2, b2) => AA.eqv(a, a2) && BB.eqv(b, b2))
     )
 
+  final def compare[AA >: A, BB >: B](that: AA Ior BB)(implicit AA: Order[AA], BB: Order[BB]): Int =
+    (this, that) match {
+      case (Ior.Left(a1), Ior.Left(a2))         => AA.compare(a1, a2)
+      case (Ior.Left(_), _)                     => -1
+      case (Ior.Both(a1, b1), Ior.Both(a2, b2)) => Order[(AA, BB)].compare((a1, b1), (a2, b2))
+      case (Ior.Both(_, _), Ior.Left(_))        => 1
+      case (Ior.Both(_, _), Ior.Right(_))       => -1
+      case (Ior.Right(b1), Ior.Right(b2))       => BB.compare(b1, b2)
+      case (Ior.Right(_), _)                    => 1
+    }
+
   final def show[AA >: A, BB >: B](implicit AA: Show[AA], BB: Show[BB]): String =
     fold(
       a => s"Ior.Left(${AA.show(a)})",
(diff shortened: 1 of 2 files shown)


TimWSpence:
Apologies if there's a reason why this  doesn't already exist but I couldn't think of one!

TimWSpence:
Also I'll follow up with an instance for IorT if this is wanted :)

TimWSpence:
> Also I'll follow up with an instance for IorT if this is wanted :)

https://github.com/typelevel/cats/pull/3555

core/src/main/scala/cats/data/Ior.scala
```diff
@@ -708,6 +708,17 @@ sealed abstract class Ior[+A, +B] extends Product with Serializable {
       (a, b) => that.fold(a2 => false, b2 => false, (a2, b2) => AA.eqv(a, a2) && BB.eqv(b, b2))
     )
 
+  final def compare[AA >: A, BB >: B](that: AA Ior BB)(implicit AA: Order[AA], BB: Order[BB]): Int =
+    (this, that) match {
+      case (Ior.Left(a1), Ior.Left(a2))         => AA.compare(a1, a2)
+      case (Ior.Left(_), _)                     => -1
+      case (Ior.Both(a1, b1), Ior.Both(a2, b2)) => Order[(AA, BB)].compare((a1, b1), (a2, b2))
```
travisbrown, line 715: This is kind of nitpicky but it would be pretty easy to avoid the allocation of a new `Order` instance here.

core/src/main/scala/cats/data/Ior.scala
```diff
@@ -708,6 +708,17 @@ sealed abstract class Ior[+A, +B] extends Product with Serializable {
       (a, b) => that.fold(a2 => false, b2 => false, (a2, b2) => AA.eqv(a, a2) && BB.eqv(b, b2))
     )
 
+  final def compare[AA >: A, BB >: B](that: AA Ior BB)(implicit AA: Order[AA], BB: Order[BB]): Int =
```
travisbrown, line 711: I think it might be useful to have a comment with a high-level explanation of the semantics here. I think what you have makes sense as the natural way to do this, but what exactly its doing isn't necessarily clear at a glance.

TimWSpence: To be honest, I copied the semantics of `===` above and didn't think too much about it! I guess it's just standard practice for a covariant datatype?

core/src/main/scala/cats/data/Ior.scala
TimWSpence: Do you think it's worth it for the loss of clarity? (Genuine question, not passive-aggressive 😂 )

travisbrown: Looking at it again, it's actually at least three allocations (for the instance and then the tuples). Given that one of the places this could be used is for sorting collections, I'd be inclined to say three fewer object allocations per comparison is worth it.

core/src/main/scala/cats/data/Ior.scala
travisbrown: It's mostly the between-ness of `Both` that I think isn't necessarily obvious. I could imagine both `Left` and `Right` preceding `Both`, for example.

travisbrown: (Sorry, by "semantics here" I meant the details of how the comparison happens, not the `AA >: A` part.)

core/src/main/scala/cats/data/Ior.scala
travisbrown: (I wouldn't block the PR for this, though.)

core/src/main/scala/cats/data/Ior.scala
TimWSpence: Oh sorry!! Yeah I had a look in Haskell and found https://hackage.haskell.org/package/these-1.1.1.1/docs/Data-These.html. It puts `Both` after `Right` so I'm happy to change this. Although I don't know how widely used (and therefore canonical) it is. In my head, I feel like `Both` is kind-of between `Left` and `Right` so the ordering makes sense to me but I'm happy to be persuaded otherwise!

core/src/main/scala/cats/data/Ior.scala
TimWSpence: 👍

core/src/main/scala/cats/data/Ior.scala
travisbrown: Hmm, my vote would be to follow the Haskell instance, then, but I don't have a strong opinion.

travisbrown: I think there is an argument for the `These` choice in that if you think of this as an `(Option, Option)` pair with two `None`s prohibited, both `Left` and `Right` would precede `Both`.

rossabaker: The Haskell instance is derived and I think not a product of deep thought, but I also don't see a compelling reason to diverge from prior art.  I think of the type as a cleaner `Either[A, (Option[A], B)]`, which `Both` after `Right` is consistent with.

TimWSpence: Cool, I'll change it. Thanks for the feedback! :)

PR #3555, opened 2020-08-07
Order for IorT

TimWSpence:

PR #3556, opened 2020-08-07
Order for writert

TimWSpence:

### DIFF
--- a/core/src/main/scala/cats/data/WriterT.scala
+++ b/core/src/main/scala/cats/data/WriterT.scala
@@ -320,6 +320,9 @@ final case class WriterT[F[_], L, V](run: F[(L, V)]) {
     G.map(
       F.traverse(run)(lv => G.tupleLeft(f(lv._2), lv._1))
     )(WriterT.apply)
+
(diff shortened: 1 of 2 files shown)


TimWSpence:
Dependent on https://github.com/typelevel/cats/pull/3555

TimWSpence:
> Dependent on #3555

This is now merged so this PR should be ready for review :)

core/src/main/scala/cats/data/WriterT.scala
```diff
@@ -320,6 +320,9 @@ final case class WriterT[F[_], L, V](run: F[(L, V)]) {
     G.map(
       F.traverse(run)(lv => G.tupleLeft(f(lv._2), lv._1))
     )(WriterT.apply)
+
+  def compare(that: WriterT[F, L, V])(implicit Ord: Order[F[(L, V)]]): Int =
+    Ord.compare(run, that.run)
```
djspiewak, line 325: Is this a normal thing to add directly to the type? I actually don't know what our idiom is here.

PR #3559, opened 2020-08-10
Merge 2.2.x after 2.2.0-RC2 release

travisbrown:
I meant to open this after publishing [2.2.0-RC2](https://github.com/typelevel/cats/releases/tag/v2.2.0-RC2) but forgot. It'll need to be merged before the next release (which should probably be another RC).

Please note that this PR should be merged with a merge commit, since if it is squashed or rebased the `v2.2.0-RC2` tag won't point to a commit on the main branch.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -275,6 +275,7 @@ possible:
  * Trond Bjerkestrand
  * Tya
  * Valentin Willscher
+ * Valeriy Avanesov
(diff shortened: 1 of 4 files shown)


travisbrown:
(Also note that we shouldn't delete the `2.2.x` branch here, but that's less important than the merge commit part.)

PR #3560, opened 2020-08-10
Improve AndThen use of Single

johnynek:
follow up to #3527 

Here we still do O(1) work on andThen and compose, but we check improve in two ways:

1. we check to see if we have a `Concat(_, Single(_, _))` that we can extend. Otherwise the first Concat will break the andThen chaining up to fusionMaxStackDepth.
2. we check to see if the andThen/compose arguments are AndThen instances early. This is a safety fix, otherwise Concat(_, _).andThen(Concat(_, _)) puts the right concat inside a single apply call in the loop, which could blow the stack since inside there could also do the same.

Note, in the fallback we were already testing for `_: AndThen`, here we inline that test to always happen early so we can leverage that information.


PeterPerhac:
￼￼￼￼

PR #3562, opened 2020-08-11
Comments on NonEmptyReducible not being a typeclass

akopich:
`NonEmptyReducible` is not a typeclass and should not be used outside of Cats. Thus, the corresponding comments are added to the `NonEmptyReducible` along with `catsDataInstancesForNonEmptyVector` and `catsDataInstancesForNonEmptyList` implementing it.


travisbrown:
:+1: but I think the "should not be used outside of Cats" might not be a universal opinion.

PR #3565, opened 2020-08-12
Fix ReducibleLaws causing stack overflow by calling `Eval.now` early

bastewart:
The test `reduceRightConsistentWithReduceRightOption` currently calls
`value` on an `Eval` before the end of the computation, i.e. it calls `value`
inside the functions passed to `reduceRight` and `reduceRightOption`.
This can cause a stack overflow in otherwise lawful `Reducible` instances.

The test has been changed to instead use `map` on the `Eval` which is
the intended way of using `Eval` in methods like `reduceRight`.

### DIFF
--- a/laws/src/main/scala/cats/laws/ReducibleLaws.scala
+++ b/laws/src/main/scala/cats/laws/ReducibleLaws.scala
@@ -26,8 +26,8 @@ trait ReducibleLaws[F[_]] extends FoldableLaws[F] {
       fa.reduceRightTo(f)((a, eb) => eb.map(f(a) |+| _)).map(Option(_)).value
 
   def reduceRightConsistentWithReduceRightOption[A](fa: F[A], f: (A, A) => A): IsEq[Option[A]] =
-    fa.reduceRight((a1, e2) => Now(f(a1, e2.value))).map(Option(_)).value <->
-      fa.reduceRightOption((a1, e2) => Now(f(a1, e2.value))).value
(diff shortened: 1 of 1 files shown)


bastewart:
It would be hard for me to write a test for this behaviour, it only happens in a complex and recursive data-type* that I have defined a `Reducible` instance for. I have run `sbt validate` locally and do get a failure, but it is in the documentation generation (`tut`) so am pretty sure it's unrelated!

I did check if this test was intentionally on Gitter before opening this PR but got no response, I'm pretty sure it is a mistake in the way the laws test is written though. (https://gitter.im/typelevel/cats?at=5f3165af733c00338f036dbd).

I also checked the other laws suites for cases of `value` being called before the final `Eval` has been resolved but couldn't see any others to fix.

\* The type I see the failure for is a zipper for navigating a rose tree (multi-way tree), the `TreeLoc` below. The test I have changed in this PR fails with an SO when testing `Reducible[TreeLoc]`, with my change it doesn't though.

```scala
// Scala 2.13 LazyList
final case class Tree[A](rootLabel: A, subForest: LazyList[Tree[A]])

type TreeForest[A] = LazyList[Tree[A]]
type Parent[A]     = (TreeForest[A], A, TreeForest[A])
type Parents[A]    = LazyList[Parent[A]]

final case class TreeLoc[A](tree: Tree[A], lefts: TreeForest[A], rights: TreeForest[A], parents: Parents[A])
```

travisbrown:
This looks reasonable to me but it'd probably be good for @johnynek to take a quick look since he added the test in #1475.

PR #3566, opened 2020-08-13
Scalafix rule to remove instance imports when upgrading to 2.2.0

### LINKED ISSUE #3563: Scalafix rule to help migration to 2.2.0
I'm really looking forward to upgrading to 2.2.0 and deleting all my `cats.instances.*` imports!

Is it worth providing an official scalafix rule to help people do this?

It could be just a simple grep to delete all `import cats.instances.*` lines, or it could be a bit smarter and also delete `import cats.implicits._` lines that are only being used to import instances, not syntax.

cb372:
Fixes #3563 

A Scalafix rule for removing instance imports that become unnecessary in v2.2.0 thanks to #3043

Behaviour is as follows:

* `import cats.instances.foo._`
    * remove unconditionally
* `import cats.implicits._`
    * rewrite to `import cats.syntax.all._` if there is any use of a syntax extension method within the lexical range of the import statement
    * otherwise remove

What I mean by "lexical range" is the range where the import statement can affect lexical scope, i.e. from the end of the import statement to the end of its enclosing block, method body or class/object body. Not sure if there's an accepted name for this.

## Other changes

* Moved the existing v1.0.0 tests and test data into a subdirectory and created separate sbt projects for them. This gives us a template to follow if we want to add Scalafix rules for future migrations (v3.0.0?).
* ~Bumped sbt-scalafix to the latest version~ <- Gave up on this is as it caused one of the existing rules to break. Let's save that for another PR.
* Updated the readme for the Scalafix rules


travisbrown:
This looks good to me, with a couple of comments.

Can you give more details about this?

> Gave up on this is as it caused one of the existing rules to break. Let's save that for another PR.

Also the fact that there's so much special-casing needed here for the `Future` instances makes me think that maybe they should just go in implicit scope too? (Personally I think I'd prefer just to banish them to alleycats, but I also don't think it's the right time to do that.)

cb372:
>Can you give more details about this?

The failing test is the one for the `ContraMapToLMap` rule. In [this line](https://github.com/typelevel/cats/blob/330ac256abb0d654f990f525ae260a489af432a5/scalafix/rules/src/main/scala/fix/Cats_v1_0_0.scala#L87) the rule inspects the `denotation` of a term and checks whether it contains a certain value.

It looks like something in the behaviour of scalafix/scalameta changed between sbt-scalafix 0.9.16 and 0.9.17. The expected value no longer appears in the denotation, so the rule never applies any patches.

I had a quick look at the scalafix and scalameta changelogs but couldn't find a likely culprit for the change.

In the grand scheme of things I don't think this is very important - how many people out there are still running Cats v0.9 and looking to migrate to 1.0.0, and make use of `contramap` syntax? Probably the simplest solution is just to delete that rule.

>Also the fact that there's so much special-casing needed here for the Future instances makes me think that maybe they should just go in implicit scope too? (Personally I think I'd prefer just to banish them to alleycats, but I also don't think it's the right time to do that.)

I don't have a strong opinion on whether the Future instances should remain in core or move to alleycats, especially as I rarely use Future myself. I know this has been discussed at length and I don't really have any insight to add.

But I think keeping them in core but leaving them out of implicit scope, just for the sake of making them a second class citizen, is probably the worst of both worlds. It'll confuse people and lead to increased support burden on gitter.

I would suggest we add them to implicit scope for v2.2.0, and you lobby to get them moved to alleycats in v3.0.0.

travisbrown:
I was just looking at the `Future` instances issue again, and remembered that part of the original reason I didn't add them was because there's not currently any `Future` machinery at all in cats-kernel, and the `Monoid` and `Semigroup` instances would have to go there. IIRC when we discussed that last year the consensus was not to introduce the `scala.concurrent` stuff into cats-kernel just to get the `Future` instances into implicit scope, which still makes sense to me.

I'm still definitely open to discussion about these instances, but I think this is a real reason to exclude them from implicit scope that isn't just "nobody really likes them".

cb372:
I see. If there’s a valid technical reason not to add them, that’s fair enough.

In that case I guess there are no changes needed on this PR?

barambani:
>     * `import cats.instances.all._`
>       
>       * remove only if there is no use of a Future instance within the lexical range (see below) of the import statement
assuming (if there are no oversights) that Future's instances are the only one left out of scope, I was wondering if a possible strategy could be to replace the import with
```
import cats.instances.future._
```
instead of leaving there `all` in case Future's are used in the range.
I'm not super familiar with scalafix and I'm not sure how complex that is, so I hope that it makes sense. Thanks.

cb372:
Good idea. That should be pretty easy to do.

I was also wondering whether, in the case that `import cats.implicits._` is used for syntax extensions (but not for Future instances), we should replace it with `import cats.syntax.all._`.

mpilquist:
I'm in favor of the Future instances being implicitly available.

LukaJCB:
Yeah I agree with @mpilquist we should add them to cats-kernel

cb372:
I've updated the rule in light of Future instances now being available in implicit scope, so this is ready for another review.

Also updated the PR description to match the new behaviour.

fthomas:
I tested the migration in the Scala Steward repo here https://github.com/scala-steward-org/scala-steward/pull/1582 but it didn't add any `cats.syntax.all._` imports which I think is the reason why the build failed. Any idea what went wrong there, @cb372?

travisbrown:
Thanks for setting that up, @fthomas!

cb372:
@fthomas Thanks for testing. I’ll try to reproduce it. I should add a test for global imports.

One thought: the rule relies on the `-P:semanticdb:synthetics:on` option being set in order to find uses of syntax extensions. Did you set that option in your test? And could we set up scala-steward to set it when it runs scalafix?

fthomas:
> One thought: the rule relies on the -P:semanticdb:synthetics:on option being set in order to find uses of syntax extensions. Did you set that option in your test?

I added https://github.com/scala-steward-org/scala-steward/pull/1582/files#diff-0df4385bf6295c0060ce939a8299f27b before running sbt and scalafix. I also inspected `scalacOptions` and checked that this option is set.

> And could we set up scala-steward to set it when it runs scalafix?

That would be the first migration that requires an additional option. We currently have no way to set them in Scala Steward. The easiest way would be to set this option globally for all migrations if it is safe to do so.

fthomas:
@cb372 Turned out that I set `scalacOptions` only for the root project in the build. Sorry for not checking more carefully. I tried again with `scalacOptions` set for all projects and some `cats.implicits._` imports are now rewritten to `cats.syntax.all._` imports. Unfortunately the migration  failed halfway with a `MissingSymbolException`:
```
[error] scalafix.internal.v1.FileException: unexpected error processing file /home/frank/data/code/scala-steward/core/modules/core/src/main/scala/org/scalasteward/core/edit/package.scala
[error] Caused by: scalafix.v1.MissingSymbolException: Missing symbol local24
[error]         at scalafix.internal.v0.LegacyCodePrinter.$anonfun$pprint$2(LegacyCodePrinter.scala:79)
[error]         at scalafix.internal.v0.LegacyCodePrinter.$anonfun$pprint$2$adapted(LegacyCodePrinter.scala:71)
[error]         at scalafix.internal.v0.LegacyCodePrinter.$anonfun$mkString$1(LegacyCodePrinter.scala:46)
[error]         at scalafix.internal.v0.LegacyCodePrinter.$anonfun$mkString$1$adapted(LegacyCodePrinter.scala:40)
[error]         at scala.collection.immutable.List.foreach(List.scala:431)
[error]         at scalafix.internal.v0.LegacyCodePrinter.mkString(LegacyCodePrinter.scala:40)
[error]         at scalafix.internal.v0.LegacyCodePrinter.$anonfun$pprint$1(LegacyCodePrinter.scala:71)
[error]         at scalafix.internal.v0.LegacyCodePrinter.$anonfun$pprint$1$adapted(LegacyCodePrinter.scala:70)
[error]         at scala.collection.Iterator.foreach(Iterator.scala:943)
[error]         at scala.collection.Iterator.foreach$(Iterator.scala:943)
[error]         at scala.collection.AbstractIterator.foreach(Iterator.scala:1431)
[error]         at scala.collection.IterableLike.foreach(IterableLike.scala:74)
[error]         at scala.collection.IterableLike.foreach$(IterableLike.scala:73)
[error]         at scala.collection.AbstractIterable.foreach(Iterable.scala:56)
[error]         at scalafix.internal.v0.LegacyCodePrinter.pprint(LegacyCodePrinter.scala:70)
[error]         at scalafix.internal.v0.LegacyCodePrinter.$anonfun$pprint$4(LegacyCodePrinter.scala:98)
[error]         at scalafix.internal.v0.LegacyCodePrinter.$anonfun$pprint$4$adapted(LegacyCodePrinter.scala:98)
[error]         at scala.collection.Iterator.foreach(Iterator.scala:943)
[error]         at scala.collection.Iterator.foreach$(Iterator.scala:943)
[error]         at scala.collection.AbstractIterator.foreach(Iterator.scala:1431)
[error]         at scala.collection.IterableLike.foreach(IterableLike.scala:74)
[error]         at scala.collection.IterableLike.foreach$(IterableLike.scala:73)
[error]         at scala.collection.AbstractIterable.foreach(Iterable.scala:56)
[error]         at scalafix.internal.v0.LegacyCodePrinter.pprint(LegacyCodePrinter.scala:98)
[error]         at scalafix.internal.v0.LegacyCodePrinter.$anonfun$pprint$9(LegacyCodePrinter.scala:129)
[error]         at scalafix.internal.v0.LegacyCodePrinter.$anonfun$pprint$9$adapted(LegacyCodePrinter.scala:129)
[error]         at scala.Option.foreach(Option.scala:407)
[error]         at scalafix.internal.v0.LegacyCodePrinter.pprint(LegacyCodePrinter.scala:129)
[error]         at scalafix.internal.v0.LegacyCodePrinter.$anonfun$loop$4(LegacyCodePrinter.scala:201)
[error]         at scalafix.internal.v0.LegacyCodePrinter.$anonfun$loop$4$adapted(LegacyCodePrinter.scala:201)
[error]         at scalafix.internal.v0.LegacyCodePrinter.$anonfun$mkString$1(LegacyCodePrinter.scala:46)
[error]         at scalafix.internal.v0.LegacyCodePrinter.$anonfun$mkString$1$adapted(LegacyCodePrinter.scala:40)
[error]         at scala.collection.Iterator.foreach(Iterator.scala:943)
[error]         at scala.collection.Iterator.foreach$(Iterator.scala:943)
[error]         at scala.collection.AbstractIterator.foreach(Iterator.scala:1431)
[error]         at scala.collection.IterableLike.foreach(IterableLike.scala:74)
[error]         at scala.collection.IterableLike.foreach$(IterableLike.scala:73)
[error]         at scala.collection.AbstractIterable.foreach(Iterable.scala:56)
[error]         at scalafix.internal.v0.LegacyCodePrinter.mkString(LegacyCodePrinter.scala:40)
[error]         at scalafix.internal.v0.LegacyCodePrinter.loop(LegacyCodePrinter.scala:201)
[error]         at scalafix.internal.v0.LegacyCodePrinter.convertSynthetic(LegacyCodePrinter.scala:206)
[error]         at scalafix.internal.v0.LegacySemanticdbIndex$.syntheticToLegacy(LegacySemanticdbIndex.scala:169)
[error]         at scalafix.internal.v0.LegacySemanticdbIndex.$anonfun$synthetics$1(LegacySemanticdbIndex.scala:51)
[error]         at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:285)
[error]         at scala.collection.Iterator.foreach(Iterator.scala:943)
[error]         at scala.collection.Iterator.foreach$(Iterator.scala:943)
[error]         at scala.collection.AbstractIterator.foreach(Iterator.scala:1431)
[error]         at scala.collection.IterableLike.foreach(IterableLike.scala:74)
[error]         at scala.collection.IterableLike.foreach$(IterableLike.scala:73)
[error]         at scala.collection.AbstractIterable.foreach(Iterable.scala:56)
[error]         at scala.collection.TraversableLike.map(TraversableLike.scala:285)
[error]         at scala.collection.TraversableLike.map$(TraversableLike.scala:278)
[error]         at scala.collection.AbstractTraversable.map(Traversable.scala:108)
[error]         at scalafix.internal.v0.LegacySemanticdbIndex.synthetics(LegacySemanticdbIndex.scala:50)
[error]         at fix.v2_2_0.RemoveInstanceImports$$anonfun$fix$1.applyOrElse(/tmp/scalafix595746135189370008/Cats_v2_2_0.scala8000476978071258990.scala:27)
[error]         at fix.v2_2_0.RemoveInstanceImports$$anonfun$fix$1.applyOrElse(/tmp/scalafix595746135189370008/Cats_v2_2_0.scala8000476978071258990.scala:14)
[error]         at scala.PartialFunction$Lifted.apply(PartialFunction.scala:228)
[error]         at scala.PartialFunction$Lifted.apply(PartialFunction.scala:224)
[error]         at scala.meta.transversers.Api$XtensionCollectionLikeUI$traverser$3$.apply(Api.scala:32)
[error]         at scala.meta.transversers.SimpleTraverser.$anonfun$apply$1(SimpleTraverser.scala:5)
[error]         at scala.meta.transversers.SimpleTraverser.$anonfun$apply$1$adapted(SimpleTraverser.scala:5)
[error]         at scala.collection.immutable.List.foreach(List.scala:431)
[error]         at scala.meta.transversers.SimpleTraverser.apply(SimpleTraverser.scala:5)
[error]         at scala.meta.transversers.Api$XtensionCollectionLikeUI$traverser$3$.apply(Api.scala:33)
[error]         at scala.meta.transversers.SimpleTraverser.$anonfun$apply$1(SimpleTraverser.scala:5)
[error]         at scala.meta.transversers.SimpleTraverser.$anonfun$apply$1$adapted(SimpleTraverser.scala:5)
[error]         at scala.collection.immutable.List.foreach(List.scala:431)
[error]         at scala.meta.transversers.SimpleTraverser.apply(SimpleTraverser.scala:5)
[error]         at scala.meta.transversers.Api$XtensionCollectionLikeUI$traverser$3$.apply(Api.scala:33)
[error]         at scala.meta.transversers.Api$XtensionCollectionLikeUI.collect(Api.scala:36)
[error]         at fix.v2_2_0.RemoveInstanceImports.fix(/tmp/scalafix595746135189370008/Cats_v2_2_0.scala8000476978071258990.scala:14)
[error]         at scalafix.internal.v0.LegacySemanticRule.fix(LegacySemanticRule.scala:29)
[error]         at scalafix.internal.v1.Rules.$anonfun$semanticPatch$1(Rules.scala:75)
[error]         at scala.collection.immutable.List.map(List.scala:293)
[error]         at scalafix.internal.v1.Rules.semanticPatch(Rules.scala:73)
[error]         at scalafix.internal.v1.MainOps$.unsafeHandleFile(MainOps.scala:232)
[error] (core / Compile / scalafix) scalafix.sbt.ScalafixFailed: UnexpectedError
```

travisbrown:
@fthomas Just to confirm, that error happens after a `clean`?

Given these complications, I'm tempted to say we should leave this as a manual step, and not part of the Scala Steward update, since it's not strictly necessary for the 2.2.0 update. It's true that removing the `implicits` / `instances` imports can speed up compilation, but I haven't personally seen cases where 2.2.0 is significantly _worse_ than 2.1 even with the now-unnecessary imports.

travisbrown (approved):
(Either way—doing this in the Scala Steward update or not—I'm 👍 on merging this PR as-is and working out any possible remaining issues in a follow-up.)

fthomas:
Yes, it also happens with a `clean` before running scalafix.

I'm also OK with merging this PR now but would still like to run it as part of the Scala Steward update if we can fix the above error.

cb372:
@fthomas I think you ran into https://github.com/scalacenter/scalafix/issues/1123. I don't think there's much we can do about it, but as a workaround I added a try-catch to log it when it happens and stop it from crashing scalafix.

Also added another test.

fthomas:
Thanks for adding the workaround, @cb372! I'll try out the new version again ASAP.

fthomas:
@cb372 The scalafix now succeeds but the build still fails because some imports of `cats.implicits._` are not replaced with imports of `cats.syntax.all._`. One example is https://github.com/scala-steward-org/scala-steward/pull/1582/files#diff-b90dc7bb760a09e4a4e86ba83ab0715a where `flatMap` on `F[_]` is required for the `for` comprehension:
```
[error] /home/frank/data/code/scala-steward/core/modules/core/src/main/scala/org/scalasteward/core/application/SelfCheckAlg.scala:29:23: value flatMap is not a member of type parameter F[Unit]
[error]       _ <- logger.info("Run self checks")
[error]                       ^
[error] /home/frank/data/code/scala-steward/core/modules/core/src/main/scala/org/scalasteward/core/application/SelfCheckAlg.scala:30:12: value map is not a member of type parameter F[Unit]
[error]       _ <- checkHttpExistenceClient
[error]            ^
```
Adding `import cats.syntax.all._` to this file fixes these errors.

fthomas:
a1b8d09 fixed 100 errors in my test, only 9 errors are remaining. They are all similar to this ([source file](https://github.com/scala-steward-org/scala-steward/pull/1582/files#diff-005f86029026efb6de48cbcce6a58826)):
```
[error] /home/travis/build/scala-steward-org/scala-steward/modules/core/src/main/scala/org/scalasteward/core/buildtool/sbt/parser.scala:36:47: No implicit Ordering defined for org.scalasteward.core.data.Resolver.
[error]       else Some(Scope(dependencies, resolvers.sorted))
[error]                                               ^
```
In this file `cats.implicits._` was replaced with `cats.syntax.all._` and it seems that the `Order.catsKernelOrderingForOrder` is now missing in the implicit scope.

cb372:
@fthomas your build should pass this time 🤞

fthomas:
Yes it did! Many thanks for the quick fixes, @cb372! :heart:

fthomas:
This rule can now be tested really easily by anyone who is using Scala Steward: just bump Cats to 2.2.0-RC3 and Scala Steward will use this rule for the update to 2.2.0-RC4.

My tests with refined (https://github.com/fthomas/refined/pull/804) and scala-steward (https://github.com/scala-steward-org/scala-steward/pull/1590) were successful.

scalafix/README.md
```diff
@@ -1,22 +1,35 @@
 # Scalafix rules for cats
 
-## Try this!
+## How to use
 
-[Install the Scalafix sbt plugin](https://scalacenter.github.io/scalafix/docs/users/installation)
+1. [Install the Scalafix sbt plugin](https://scalacenter.github.io/scalafix/docs/users/installation)
 
-To run all rules that apply to version `1.0.0-RC1` run
+2. Configure the SemanticDB compiler plugin to enable synthetics:
```
fthomas, line 7: How about moving this to the "Migration to Cats v2.2.0" section since it is only required for this rule and not Cats_v1_0_0? Future rules also might require different `scalacOptions` or no additional options at all.

cb372: done

cb372:
@travisbrown Yep, I'd say this is ready to merge

Sorry I think I confused the GitHub check by putting `[skip ci]` in my final commit message. I'll just do a force-push to try and fix that.

fthomas (approved):

mzuehlke:
You can use this query to get a list of all migrations done by Scala Steward to cats `2.2.0-RC4`:

https://github.com/search?q=author%3Ascala-steward+is%3Apr+cats+2.2.0-RC4

LukaJCB (approved):
This is pretty cool, thanks!

PR #3567, opened 2020-08-14
Avoid all evaluation of LazyList#foldRightDefer

takayahilton:


travisbrown:
One of the CI failures is just a connectivity thing and the other seems definitely unrelated (see #3568), so I just restarted both jobs.

tests/src/test/scala-2.13+/cats/tests/LazyListSuite.scala
```diff
@@ -52,6 +55,16 @@ class LazyListSuite extends CatsSuite {
     assert(LazyList.empty[Int].show === (s"LazyList()"))
   }
 
+  test("Avoid all evaluation of LazyList#foldRightDefer") {
```
travisbrown, line 58: Just to confirm: this test hangs without the change above?

takayahilton: Yes,  The test never finishes.

travisbrown:
One Scala.js build is still failing even after a couple of restarts. It couldn't be related, so I think we can just wait and try again in a couple of hours or so.

barambani:
It seems it was a corrupted cache. I deleted the PR's cache and it went through.

PR #3569, opened 2020-08-14
Change AndThen to directly check isRightAssociated

### LINKED ISSUE #3568: toRightAssociated test failure in CI
Just noticed this test failure in [this job](https://travis-ci.org/github/typelevel/cats/jobs/717834344) for #3567, probably related to #3527:

```
==> X cats.tests.AndThenSuite.toRightAssociated works  0.202s munit.FailException: Failing seed: fSK8i_ADMdWh0P7wdLaXzD1SVThALHwqzii_jRMBtHE=
7051You can reproduce this failure by adding the following override to your suite:
7052
7053  override val scalaCheckInitialSeed = "H6sIq1EUUvxUiPa11mJ4PoKuyBKd3swtk-oTfbaaH6D="
7054
7055Exception raised on property evaluation.
7056> ARG_0: AndThen$201277246
7057> Exception: java.lang.StackOverflowError: null
7058    at munit.ScalaCheckSuite.propToTry(ScalaCheckSuite.scala:89)
7059Caused by: java.lang.StackOverflowError
7060    at cats.data.AndThen$Concat.equals(AndThen.scala:163)
7061    at cats.data.AndThen$Concat.equals(AndThen.scala:163)
7062    at cats.data.AndThen$Concat.equals(AndThen.scala:163)
7063    at cats.data.AndThen$Concat.equals(AndThen.scala:163)
```

I don't personally think this should block the next release candidate, but we should look into it. @johnynek, any idea?

johnynek:
fix #3568 

The change to address the above is to avoid `equal` which on a very deep structure can blow the stack. Instead, directly check the property we want: isRightAssociated.

I removed a microoptimization (using != instead of <) since I would be *SHOCKED* if that changed the perf of this code which has to do quite a bit of pattern matching already. The != is not very safe since if a refactor gets it wrong, it can easily blow past the limit.

Lastly, I couldn't resist a bit of refactoring to share more of the code and improve associations (still at O(1) cost per andThen/compose).

cc @travisbrown

### DIFF
--- a/core/src/main/scala/cats/data/AndThen.scala
+++ b/core/src/main/scala/cats/data/AndThen.scala
@@ -72,19 +72,12 @@ sealed abstract class AndThen[-T, +R] extends (T => R) with Product with Seriali
     // technique implemented for `cats.effect.IO#map`
     g match {
       case atg: AndThen[R, A] =>
-        (this, atg) match {
-          case (Single(f, indexf), Single(gs, indexg)) if indexf + indexg < fusionMaxStackDepth =>
-            Single(f.andThen(gs), indexf + indexg + 1)
-          case (Concat(left, Single(f, indexf)), Single(gs, indexg)) if indexf + indexg < fusionMaxStackDepth =>
-            Concat(left, Single(f.andThen(gs), indexf + indexg + 1))
-          case _ =>
-            Concat(this, atg)
-        }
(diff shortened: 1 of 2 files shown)


travisbrown:
By the way I'm marking this as an enhancement instead of a bugfix for the release notes, since the buggy part has never been released.

PR #3570, opened 2020-08-16
Converted build to GitHub Actions

djspiewak:
This is a bit of an experiment for now. I'm using sbt-github-actions, which isn't a *perfect* fit for what we were originally doing in the Travis build. Specifically, some things that were entirely separate jobs are now part of the build job because I'm being lazy. However, GitHub Actions is so much faster than Travis that it probably doesn't matter.

Snapshot publication has been removed for the moment, as has Hydra usage, because I don't have those credentials.


djspiewak:
NB: the build is failing because of code coverage. Some projects (like alleycats) are below the minimum, apparently. I wonder how long that's been a thing?

LukaJCB:
Pretty sure we haven't had coverage as a success condition for a while now, but I don't remember when exactly we stopped.

kailuowang:
Alleycats was moved into cats. I am not sure if it came with enough coverage.

djspiewak:
> Pretty sure we haven't had coverage as a success condition for a while now, but I don't remember when exactly we stopped

Well, bizarrely, if you run `coverageReport`, it will fail the build. I'm not sure why it *isn't* failing the build in Travis, but it certainly does here (and locally).

LukaJCB:
That is bizarre indeed

larsrh (approved):
SHIP IT ALL THE WAY TO THE MOON

Issue #3536, opened 2020-07-28
Future instances not in implicit scope


mrdziuban:
Importing `cats.instances.future._` is still required to summon typeclass instances for `scala.concurrent.Future`, e.g.

```scala
import cats.Applicative
import scala.concurrent.Future
import scala.concurrent.ExecutionContext.Implicits.global

Applicative[Future]
// Could not find an instance of Applicative for scala.concurrent.Future

import cats.instances.future._

Applicative[Future]
// cats.Applicative[concurrent.Future] = cats.instances.FutureInstances$$anon$1@479111ba
```

This was the only `cats.instances` import I needed to keep around when upgrading a large codebase that strictly uses a la carte imports from v2.1.1 to v2.2.0-RC2.

LukaJCB:
Thanks for bringing this up! Will have a look soon

travisbrown:
I believe there was some discussion about this somewhere, and that excluding them was a design decision, but I don't remember the details, and I'm not committed to either including or excluding them from implicit scope.

barambani:
This https://github.com/typelevel/cats/issues/2334 may or may not be what @travisbrown is referring to, but I'm adding it here for reference as that discussion is still open and marked for milestone 3.0

barambani:
Referencing relevant discussion here https://github.com/typelevel/cats/pull/3566#issuecomment-674367117

PR #3572, opened 2020-08-17
Add Future instances to implicit scope

travisbrown:
See discussion [here](https://github.com/typelevel/cats/pull/3566#issuecomment-674367117). I've not built or run tests locally yet but I think it should work.

### DIFF
--- a/core/src/main/scala/cats/Invariant.scala
+++ b/core/src/main/scala/cats/Invariant.scala
@@ -4,10 +4,11 @@ import cats.arrow.Arrow
 import cats.kernel._
 import simulacrum.typeclass
 import cats.kernel.compat.scalaVersionSpecific._
+import scala.annotation.implicitNotFound
 import scala.collection.immutable.{Queue, SortedMap}
+import scala.concurrent.{ExecutionContext, Future}
 import scala.util.Try
 import scala.util.control.TailCalls.TailRec
-import scala.annotation.implicitNotFound
 
 /**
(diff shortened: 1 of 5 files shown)


barambani:
This should also resolve this https://github.com/typelevel/cats/issues/3536. Adding it for reference.

barambani:
We could remove `FutureInstances` from here https://github.com/typelevel/cats/blob/a0752346dd0eb2aca54c5a2e153340694d33bada/js/src/test/scala/cats/tests/FutureTests.scala#L18 and the import from here https://github.com/typelevel/cats/blob/a0752346dd0eb2aca54c5a2e153340694d33bada/jvm/src/test/scala/cats/tests/FutureSuite.scala#L3 now. Or you would prefer keep testing the instances through them ? Happy to go on either way 👍 .

barambani (approved):
Silly me :). The test changes are already in.
Reminder to self: never look at prs from the phone.

cb372:
Once this is merged, it would be handy to have another RC release so I can bump the dependency in `scalafix/build.sbt` on #3566

travisbrown:
Yes, I'm traveling at the moment but could probably publish an RC tonight if this gets merged before then.

cb372:
@travisbrown Don't worry about it, I didn't realise we were publishing snapshots to Sonatype. I was able to use a snapshot. Enjoy your holiday!

PR #3603, opened 2020-09-11
Add method `distinct` to TraverseFilter

takayahilton:
It comes from Data.Witherable's hashNub and ordNub.

### DIFF
--- a/core/src/main/scala/cats/TraverseFilter.scala
+++ b/core/src/main/scala/cats/TraverseFilter.scala
@@ -1,7 +1,10 @@
 package cats
 
+import cats.data.State
 import simulacrum.{noop, typeclass}
+
 import scala.annotation.implicitNotFound
+import scala.collection.immutable.{HashSet, TreeSet}
 
 /**
  * `TraverseFilter`, also known as `Witherable`, represents list-like structures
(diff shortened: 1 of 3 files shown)


LukaJCB (approved):

22majda22 (commented):
Ggg

core/src/main/scala/cats/TraverseFilter.scala
```diff
@@ -85,6 +88,32 @@ trait TraverseFilter[F[_]] extends FunctorFilter[F] {
...
+   * Removes duplicate elements from a list, keeping only the first occurrence.
+   * This is usually faster than ordDistinct, especially for things that have a slow comparion (like String).
+   */
+  def hashDistinct[A](fa: F[A])(implicit H: Hash[A]): F[A] =
+    traverseFilter[State[HashSet[A], *], A, A](fa)(a =>
+      State(alreadyIn => if (alreadyIn(a)) (alreadyIn, None) else (alreadyIn + a, Some(a)))
+    )
+      .run(HashSet.empty)
+      .value
+      ._2
```
satorg, line 116: @takayahilton unfortunately it is not going to work the way it is supposed to.

The passed `Hash` typeclass is not used here since `HashSet` from Scala library relies on the Java's `hashCode` instead.

satorg: (linking to discussions in #4147 and #4185 just for reference)

armanbilge: @satorg note that the law I mentioned in https://github.com/typelevel/cats/issues/4147#issuecomment-1100377989 effectively makes it okay, I think. At least at the time the PR was merged.

satorg: Yeah, I see... Although I'm not getting it – if it is strictly required by the law for `Hash` to produce the same value as the universal hash does, then why would `Hash` be useful for at all?

armanbilge: You're right, it doesn't really make sense 😕 see discussion on the topic in https://github.com/typelevel/cats/pull/4118#issuecomment-1023473730.

The problem is, it's difficult to fix this backwards-compatibly. Because there may be code (such as this PR) that relies on this "law" ...

johnynek: I think we should fix this. This "law" precludes many `Hash` implementations on opaque types.

PR #3606, opened 2020-09-14
Adding TraverseFilter instance for Either

takayahilton:


LukaJCB:
Hi @takayahilton thanks a bunch for opening this PR, I'm not sure if this instance is a good idea, given it uses `E: Monoid` merely for `empty`, which is somewhat of a smell IMHO.

takayahilton:
@LukaJCB
Thanks for your review.
I implemented this with reference to [hear](https://hackage.haskell.org/package/witherable-0.1.3.3/docs/src/Data-Witherable.html#line-206)
But, as you say, I’ve certainly noticed that there are few situations where this really helps.

PR #3610, opened 2020-09-18
Added implicit syntax for Free monad

vincentdehaan:
This PR introduces a syntax with less boilerplate for Free monads. Given a case class `case class DoSomething(x: Int, y: String)`, one can now write:

```scala
DoSomething.free(7, "hey!")
```

without having to define a `def doSomething(x: Int, y: String)` first.

The PR also refactors the boilerplate generation code to reduce code duplication.


project/FreeBoiler.scala
```diff
@@ -0,0 +1,30 @@
...
+
+    override def content(tv: TemplateVals): String = {
+      import tv._
+
+      block"""
+      |package cats
+      |package free
+      |
+      |trait FreeSyntax {
+      |  implicit class WithFreeSyntax0[F[_], A](fa: F[A]) {
```
johnynek, line 19: can these not extend `AnyVal`? also below?

docs/src/main/mdoc/datatypes/freemonad.md
```diff
@@ -133,6 +133,8 @@ def update[T](key: String, f: T => T): KVStore[Unit] =
   } yield ()
 ```
 
+If the words in the grammar do not take type parameters, a shorthand is available after importing `cats.free.implicits._`: one can just write `Delete.free("key")` without defining `def delete(key: String)` first.
```
johnynek, line 136: can you extend the example here? I still don't totally follow the limitations and the benefits. It might be nice to be explicit with an example showing the verbose way, and the way using the syntax.

vincentdehaan:
Thanks @johnynek for your comments. I agree. I found a neater and quite different approach so I will close this PR and open a new one.

PR #3615, opened 2020-09-22
Add Group instance for Map to alleycats

tmccarthy:
This PR adds a `Group` for `Map` to alleycats. It can be constructed where the value type has a `Group` instance.

```scala
import cats.syntax.group._

val x = Map("a" -> 2, "b" -> 1)
val y = Map("a" -> 3)

assert((x |-| y) === Map("a" -> -1, "b" -> 1))
```

The resulting instance violates the left- and right-inverse laws, since an empty `Map` is not equal to one with a key bound to an empty value:

```scala
import cats.syntax.group._

val x = Map("a" -> 1)

assert((x |-| x) === Map.empty) // Fails, since x |-| x is Map("a" -> 0)
```

Despite this, I've found this `Group` instance useful in contexts where there is no semantic difference between the lack of a binding and a binding to an empty value.

I believe you can construct a `CommutativeGroup` in the same way, which I would be happy to add as a higher-priority implicit.

I realise I've raised this PR without first discussing the change in an issue or on Gitter. Please feel free to provide any feedback or reject this PR as appropriate.

### DIFF
--- a/alleycats-core/src/main/scala/alleycats/std/map.scala
+++ b/alleycats-core/src/main/scala/alleycats/std/map.scala
@@ -80,7 +80,7 @@ trait MapInstances {
           }) { chain => chain.foldLeft(Map.empty[K, B]) { case (m, (k, b)) => m.updated(k, b) } }
     }
 
-  implicit def alleycatsStdMapGroup[K, V : Group]: Group[Map[K, V]] = new MapMonoid[K, V] with Group[Map[K, V]] {
+  implicit def alleycatsStdMapGroup[K, V: Group]: Group[Map[K, V]] = new MapMonoid[K, V] with Group[Map[K, V]] {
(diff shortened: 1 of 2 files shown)


johnynek:
BTW: algebird dealt with this by removing zeros from the maps when it does +/-. That brought its own problems.

On the other hand, I think you can define another `Eq[Map[K, V]]` on which this Group is lawful, basically one where you each key matches and any items only on one side have to have `V === Monoid[V].zero`

To me, I think the better approach would not be to put it in alleycats and instead make a `newtype` like `SparseMap[K, V]` which is just a `Map[K, V]` where `Eq[SparseMap[K, V]]` works as described above. Then I believe Group and Ring are lawful.

armanbilge:
The `SparseMap` thing sounds good, if there's interest to add it to Cats I suppose it can be a new PR.

PR #3616, opened 2020-09-26
Added and edited new applicativeerror.md

dhinojosa:

### DIFF
--- a/docs/src/main/mdoc/typeclasses/applicativeerror.md
+++ b/docs/src/main/mdoc/typeclasses/applicativeerror.md
@@ -1,0 +1,180 @@
+---
+layout: docs
+title:  "Applicative Error"
+section: "typeclasses"
+source: "core/src/main/scala/cats/ApplicativeError.scala"
+scaladoc: "#cats.ApplicativeError"
+---
+# ApplicativeError
+
+## Description
+`ApplicativeError` extends `Applicative` to provide handling for types
+that represent the quality of an exception or an error, for example, `Either[E, A]`
+
+## TypeClass Definition
+`ApplicativeError` is defined by the following `trait`
+
+```
+trait ApplicativeError[F[_], E] extends Applicative[F] {
+  def raiseError[A](e: E): F[A]
+  def handleErrorWith[A](fa: F[A])(f: E => F[A]): F[A]
+  def handleError[A](fa: F[A])(f: E => A): F[A]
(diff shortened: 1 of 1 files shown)


docs/src/main/mdoc/typeclasses/applicativeerror.md
```diff
@@ -0,0 +1,180 @@
...
+form of `Applicative`.
+
+```scala mdoc:silent
+import cats.implicits._
+def attemptDivideApplicativeErrorWithMap2[F[_]](x: Int, y: Int)(implicit ae: ApplicativeError[F, String]): F[_] = {
+   if (y == 0) ae.raiseError("divisor is error")
+   else {
+     val fa = ae.pure(x)
+     val fb = ae.pure(y)
+     ae.map2(fa, fb)(_ + _)
```
Daenyth, line 129: Stylistic question: do people prefer using `map2`? Ubiquitously in codebases I've worked on, it's unused, and instead people use the tuple ops, `(fa, fb).mapN(_ + _)`, which works for any number of arguments (no need to use `map2` vs `map3` vs `map4`, so on)

Daenyth (approved):
Very helpful to have these docs improved.

What do you think about putting `MonadError` on the same page? When I'm teaching people the API I usually show them `attempt` and `rethrow` as the sort of "foundational" pair of functions, since basically all the rest can be built on top of those.

docs/src/main/mdoc/typeclasses/applicativeerror.md
dhinojosa: I was thinking of `MonadError` as a separate page just purely for the reason so that it fits on the left border of the website. Also, the original `Applicative` and `Monad` were introduced separately.  I can try to do `Applicative and Monad Error`.  I could go either way, TBH. ;)  As per `map2`, I just chose one of the `Applicative` methods. I can change it to be more realistic.

dhinojosa: I'll try to include both on the same page with the menu and title of "Applicative and Monad Error"

larsrh (approved):
LGTM :+1:

docs/src/main/mdoc/typeclasses/applicativeerror.md
larsrh: I'd be happy to have two separate pages and have them link to each other.

barambani (approved):

PR #3620, opened 2020-09-30
Add typeclasses for immutable.Seq

### LINKED ISSUE #1222: No typeclasses for Seq
I see there are typeclasses under cats.std for List and Vector (among others), but none for Seq. Is there a reason no typeclasses for Seq are provided? If it's simply that there's been no demand previously, then I'm happy to raise a PR to add this.

JosephMoniz:
Adds type class instances for `immutable.Seq`. This is based off of the implementations of the `List` and `Vector` typeclasses.

Closes #1222

### DIFF
--- a/core/src/main/scala-2.12/cats/compat/Seq.scala
+++ b/core/src/main/scala-2.12/cats/compat/Seq.scala
@@ -1,0 +1,8 @@
+package cats.compat
+
+import scala.collection.immutable.{Seq => ISeq}
+
+private[cats] object Seq {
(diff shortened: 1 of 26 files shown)


LukaJCB (commented):
Thanks a bunch for this PR, it's very thorough! :) I think we should future proof a little bit, based on my experiences in https://github.com/typelevel/cats/pull/3552

JosephMoniz:
Thanks for the review @LukaJCB!

I went ahead and lowered the priority of the `immutable.Seq` instances you called out. Hopefully they wont cause any problems for dotty going forward now.

LukaJCB (approved):
Excellent, thank you!

djspiewak (approved):
This is quite nice!

PR #3621, opened 2020-10-01
improve WriterT factory methods with PartiallyApplied technique

nikiforo:
The goal of PR is to change
```
WriterT.putT[F[_], L, V] -> WriterT.putT[F[_]]
WriterT.put[F[_], L, V] -> WriterT.put[F[_]]
WriterT.tell[F[_], L] -> WriterT.tell[F[_]]
WriterT.value[F[_], L, V] -> WriterT.value[F[_], L]
WriterT.valueT[F[_], L, V] -> WriterT.valueT[F[_], L]

Writer.value[L: Monoid, V] -> value[L]
```
So that you have to provide only those types, that can't be inferred automatically.

The biggest downside I see in this PR is that it backward incompatible, in a sense that it forces clients to change all the places, where mentioned methods are used. But the change is obvious - you just have to reduce the amount of information you provide to the compiler `putT[F, L, V](l)(v) -> putT[F](l)(v)`

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -414,7 +414,7 @@ def mimaSettings(moduleName: String, includeCats1: Boolean = true) =
           exclude[DirectMissingMethodProblem]("cats.data.WriterT.tell"),
           exclude[DirectMissingMethodProblem]("cats.data.WriterT.value"),
           exclude[DirectMissingMethodProblem]("cats.data.WriterT.valueT"),
-          exclude[DirectMissingMethodProblem]("cats.data.package#Writer.value"),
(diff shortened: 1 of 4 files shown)


nikiforo:
So yes, backward compatibility matters. I've changed mima exclusion rules to get a feedback.
 - It's better to have put[F] than put[F, L, V]
 - But it's wrong to break backward compatibility.
 - I can extend Writer with some other method names, but they wouldn't be so nice and semantically precise as original Writer ones.
The question is, what should be done next?

Issue #3622, opened 2020-10-01
Follow up to Seq Instances


djspiewak:
They don't appear to be working quite correctly? Follow up to #3620

```scala
scala> import cats._
import cats._

scala> import cats.syntax.all._
import cats.syntax.all._

scala> Seq(1, 2, 3) |+| Seq(3, 4, 5)
<console>:21: error: value |+| is not a member of Seq[Int]
       Seq(1, 2, 3) |+| Seq(3, 4, 5)
                    ^

scala> def foo[F[_]: Monad, A](fa: F[A]) = ()
foo: [F[_], A](fa: F[A])(implicit evidence$1: cats.Monad[F])Unit

scala> foo(Seq(1, 2, 3))
<console>:19: error: Could not find an instance of Monad for Seq
       foo(Seq(1, 2, 3))
          ^

scala> import cats.instances.seq._
import cats.instances.seq._

scala> foo(Seq(1, 2, 3))
<console>:22: error: Could not find an instance of Monad for Seq
       foo(Seq(1, 2, 3))
          ^
```

I'm guessing we need to add some tests here and see if we can introspect what's going on, but I would have expected this to work.

djspiewak:
Update to the above: these are Scala 2.12 problems. On Scala 2.13, more of it works:

```scala
scala> import cats._, cats.syntax.all._
import cats._
import cats.syntax.all._

scala> Seq(1, 2, 3) |+| Seq(3, 4, 5)
                    ^
       error: value |+| is not a member of Seq[Int]
       did you mean ++: or :++?

scala> List(1, 2, 3) |+| List(3, 4, 5)
val res1: List[Int] = List(1, 2, 3, 3, 4, 5)

scala> def foo[F[_]: Monad, A](fa: F[A]) = ()
def foo[F[_], A](fa: F[A])(implicit evidence$1: cats.Monad[F]): Unit

scala> foo(Seq(1, 2, 3))

scala> foo(List(1, 2, 3))

scala> Seq(1, 2, 3) *> List(3, 4, 5)
val res5: Seq[Int] = List(3, 4, 5, 3, 4, 5, 3, 4, 5)

```

So the semigroup syntax still doesn't quite work, but the monad resolution does. The final expression though is something we should probably talk about, since I think a lot of us would have expected that to fail to compile.

JosephMoniz:
[...]

PR #3624, opened 2020-10-01
immutable.Seq missing semigroup

JosephMoniz:
Looks like I missed adding a semigroup instances on `immutable.Seq`.

Before: 
```scala
scala> import cats._, cats.syntax.all._, scala.collection.immutable.Seq
import cats._
import cats.syntax.all._
import scala.collection.immutable.Seq

scala> Seq(1, 2, 3) |+| Seq(3, 4, 5)
<console>:19: error: value |+| is not a member of scala.collection.immutable.Seq[Int]
       Seq(1, 2, 3) |+| Seq(3, 4, 5)
```

After:
```scala
scala> import cats._, cats.syntax.all._, scala.collection.immutable.Seq
import cats._
import cats.syntax.all._
import scala.collection.immutable.Seq

scala> Seq(1, 2, 3) |+| Seq(3, 4, 5)
res0: scala.collection.immutable.Seq[Int] = List(1, 2, 3, 3, 4, 5)
```

See: #3622

### DIFF
--- a/kernel/src/main/scala/cats/kernel/Semigroup.scala
+++ b/kernel/src/main/scala/cats/kernel/Semigroup.scala
@@ -1,7 +1,7 @@
 package cats.kernel
 
 import scala.annotation.tailrec
-import scala.collection.immutable.{BitSet, Queue, SortedMap, SortedSet}
+import scala.collection.immutable.{BitSet, Queue, SortedMap, SortedSet, Seq}
 import scala.concurrent.{ExecutionContext, Future}
 import scala.concurrent.duration.{Duration, FiniteDuration}
 import scala.{specialized => sp}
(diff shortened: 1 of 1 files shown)


djspiewak (approved):
I do think we need to talk a bit about source compatibility though and 2.12 support. These instances basically don't resolve on 2.12 because `Seq` is a different type. Is that something which is worth addressing?

JosephMoniz:
All of these instances resolve on 2.12 just fine as long as you reference `scala.collection.immutable.Seq` and not  `scala.collection.mutable.Seq`. The downside here is that in order to use any of these typeclasses in 2.12 or earlier you need to more explicitly reference `immutable.Seq`.

Note: My example in the original PR description is running in 2.12

In terms of source compatibility though I think this is fine for 2.12 because unless you're going out of your way to convert `mutable.Seq` into `immutable.Seq` and your were previously avoiding `mutable.Seq` then these changes should have no impact.

JosephMoniz:
Added some additional thoughts on source parody here: https://github.com/typelevel/cats/issues/3622#issuecomment-702848584

johnynek (approved):

PR #3627, opened 2020-10-05
🔥 Universal Show 🔥

djspiewak:
Bring out your pitchforks!

Right now, `Show` is in a really weird spot. It's not as convenient as `toString` for ad-hoc debugging, so I often just don't use it. It doesn't really have any meaningful laws, since it isn't *enough* to characterize encoding, and we don't even *have* `Read`. So all in all, `Show` in its present form only makes sense for three use-cases:

- Overridability (since you can't provide a custom `toString` for a third-party type)
- Pretty-printing that is dependent on some polymorphic type (including inductive derivation)
- `Array` (which has a horrific `toString`)

Also, let's remember that we're on the JVM: *all* types have a `String` representation. And you can't even say it's "not a useful representation", since in many cases, just determining the pointer value is actually a very useful thing.

This PR embraces that. You can still provide custom `Show`s – and there are many reasons to do so! – but you are no longer *forced* to. This also makes it so that `.show` is just a strictly better `.toString` in that it handles all of the same use-cases in all the same scenarios, but with added advantages.

Downsides:

- You can no longer fail compilation when using `.show` on a type which just… doesn't have one
  + Notably: you could forget to add a `: Show` constraint and the compiler won't catch you
- Some types have absolutely incredibly evil `toString`s (e.g. `URL`), and now you could end up getting smacked with that in a polymorphic context. I find it unlikely anyone will be really bitten by this.

Motivating example: [Cats Effect 3's `IO.println` function](https://gitter.im/typelevel/cats-effect-dev?at=5f7b3f3765cb0320606f86f9).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .hydra
 .bloop
+.bsp
 .metals
 project/boot
 target
--- a/core/src/main/scala/cats/Show.scala
+++ b/core/src/main/scala/cats/Show.scala
@@ -102,7 +102,11 @@ object Show extends ScalaVersionSpecificShowInstances with ShowInstances {
     cats.instances.sortedMap.catsStdShowForSortedMap[K, V]
 }
 
-private[cats] trait ShowInstances {
+private[cats] trait ShowInstances extends LowPriorityShowInstances { this: Show.type =>
   implicit def catsShowForFiniteDuration: Show[FiniteDuration] =
     cats.instances.finiteDuration.catsStdShowForFiniteDurationUnambiguous
+}
+
+private[cats] trait LowPriorityShowInstances { this: Show.type =>
+  implicit def catsShowFromUniversalToString[A]: Show[A] = fromToString[A]
 }


Daenyth:
Personally I find `Show` useful as "just" a `toString` which sometimes doesn't exist - for types that it makes no sense on. When I use it, it's because I want to either expose an api where a caller is supposed to think about what the string version should be.

So I'm 👎 on this, for what it's worth

gabro:
Sorry for the flyby comment, but have you considered keeping Show as it is, and introducing a new typeclass with the “better toString” semantics (e.g ToString)?

kubukoz:
> Sorry for the flyby comment, but have you considered keeping Show as it is, and introducing a new typeclass with the “better toString” semantics (e.g ToString)?

This.

90% of the times I use `Show` it is specifically to avoid using the `toString` method in an interpolator, which is ridiculously hard to find the usages of, for one (not even Metals can do it).

I wouldn't mind `UShow` or `ToString`. Probably the latter, for outsider friendliness.

djspiewak:
> Sorry for the flyby comment, but have you considered keeping Show as it is, and introducing a new typeclass with the “better toString” semantics (e.g ToString)?

Thought about it! That's on the table. But… something I realized is I've now one-offed that three times in separate codebases, but I've literally never written code using `Show` directly which did what I wanted. Which is to say, I never want `Show` as it is, and I *always* want the "better `toString`" class.

kubukoz:
What about an import, like `import Show.auto._`? Relying on implicit priority is a bit too much IMO

djspiewak:
> What about an import, like import Show.auto._? Relying on implicit priority is a bit too much IMO

Then you may as well just call `.toString` explicitly. I'm not sure I understand the concern around implicit priority specifically. If I can paraphrase your concern from earlier, it sounds like you (and @Daenyth) are using `Show` in the exclusionary sense, which is sort of fundamentally at odds with this PR's change. (to be clear, I think that's a valid viewpoint, I'm just pointing out that there's no way to make this PR's change in any way compatible with that use-case, since the whole *point* of this PR is to eliminate that scenario)

tpolecat:
I suspect this is the kind of thing maintainers will like because we all hate `Show` anyway, but I also suspect there are a lot of less-weird people out there who are using `Show` as a way to construct strings to show to users, thinking of it as a safer `toString` since you have to opt in. I have coaxed otherwise upstanding functional programmers away from this interpretation from time to time, but it's not obviously incorrect.

In any case I'm morally :+1: but spiritually 👎 although I could be flipped to 👍 👍 if you also renamed it `Debug` and added deprecated aliases for `Show` and friends. Bonus points if you go back in time and change `toString` to `debugString`.

djspiewak:
@kubukoz 

> 90% of the times I use Show it is specifically to avoid using the toString method in an interpolator, which is ridiculously hard to find the usages of, for one (not even Metals can do it).

Can you elaborate on this a little bit? What's so problematic about `toString` *itself*? Obviously it's not parametric (in some sense), and some types (`URL`) use it to perform satanic rituals, but aside from that where is the issue? If the problem is *control* over the `toString`, then I would think (correct me if I'm mistaken) that this PR still preserves that property, since you can always just… define a custom `Show`.

clhodapp:
What if there were another extension method, `debugString`, and a sealed-off typeclass, `DebugString`, which tries to use `Show` but falls back to `toString` if an instance isn't available. That way:
  * Library authors don't need to rewrite all of their instances
  * Folks who want to restrict themselves to cases where a `Show` instance is available still can (without relying on something that is deprecated)
  * We can opportunistically get better strings for debugging

djspiewak:
> What if there were another extension method, debugString, and a sealed-off typeclass, DebugString, which tries to use Show but falls back to toString if an instance isn't available.

That's basically @gabro's suggestion as well, and there's merit to it! I've built that typeclass on three separate occasions embedded into various codebases, so I think it's 100% worth putting something like it into Cats. It's mostly a question of whether it should *replace* `Show` or just augment it.

Also note that a separate typeclass like this creates a layer of indirection which results in worse implicit resolution diagnoses, though since there's always a fallback, maybe it's not a big deal.

I guess my position is that I don't see a valid use of `Show` *other* than `debugString`, and I pretty regularly want to `debugString` completely arbitrary things, thus I believe that `Show` itself should provide this functionality. But that's really the crux of this debate.

clhodapp:
Whether it's compelling or not, I think the core use-case for a restrictive `.show` would be to allow writing code that pretends that `toString` doesn't exist, in much the same way as we do with `null`.

You also have to consider that a bunch of people have already written code against `show`, such that if you want to change the semantics like this, you really owe it to them to change the name. Otherwise, you are handing them a footgun against invariants that it may be key to their design to maintain.

djspiewak:
> Whether it's compelling or not, I think the core use-case for a restrictive .show would be to allow writing code that pretends that toString doesn't exist, in much the same way as we do with null.

I think part of what I'm getting at is this: what code *actually* benefits from that (pretending `toString` doesn't exist)? Really, what is the use-case? Because `toString` *always* exists, and it, like `Show`, is really only useful for debugging. If you want encoding, you need something much more powerful (and less `String`y). So why are you bothering to pretend that `toString` doesn't exist? What is it *really* gaining you?

> You also have to consider that a bunch of people have already written code against show, such that if you want to change the semantics like this, you really owe it to them to change the name. Otherwise, you are handing them a footgun against invariants that it may be key to their design to maintain.

This is very true, though I think part of my implicit argument here is that anyone who's code is *broken* by this change was probably relying on invariants that `Show` *already* doesn't provide. Again, debugging vs encoding.

JosephMoniz:
I guess the main user behavior change here would be that it could possibly encourage people to explicitly write more `toString` methods on their non case classes. Which is interesting, mostly because debuggers + test frameworks tend to use `toString` and not be aware of `Show`. I think that's actually a pretty big ergonomic win.

kubukoz:
> @kubukoz
> 
> > 90% of the times I use Show it is specifically to avoid using the toString method in an interpolator, which is ridiculously hard to find the usages of, for one (not even Metals can do it).
> 
> Can you elaborate on this a little bit? What's so problematic about `toString` _itself_? Obviously it's not parametric (in some sense), and some types (`URL`) use it to perform satanic rituals, but aside from that where is the issue? If the problem is _control_ over the `toString`, then I would think (correct me if I'm mistaken) that this PR still preserves that property, since you can always just… define a custom `Show`.

@djspiewak 

It's easier to see where the instance is used than where `toString` is used. This is likely a tooling problem, as usages of `toString` (maybe with the inclusion of `any2stringadd`) could be detected automatically for a concrete type.

I have mixed feelings about writing new code with `show` without writing an instance, but the typeclass isn't meant to be used for encoding so maybe I'm using it wrong anyway.

djspiewak:
> I don't have time right now to go through everything, but was it suggested to have Show and Debug, where Show stays as it is but Debug can be derived directly from Show or defaulted to toString?

That has been suggested in a couple different variants. :-) It's certainly on the table!

One objection to that plan, particularly if it's literally named `Debug`, is it raises an obvious question: if the `Debug` typeclass is for debugging, then what is `Show` for? My concern is people may answer that with any answer other than "also debugging". You see the problem.

Daenyth:
You know, on reflection, and reading the comments here, I think my previous 👎 was incorrect. Reading the comments, I was thinking "does `Show` even need to exist in the first place", which brought me around to exactly what your first proposal post is saying; it gives us a toString that's controlled by code in our own codebase, nothing more. If it's not that, it doesn't make sense to exist at all. This isn't haskell, so everything does have `toString` whether we like it or not.

Baccata:
I agree with @kubukoz and @clhodapp on this. I've used `Show` in the past to constrain coworkers to think of how things should appear in logs, and prevent mis-haps where the type of a value would change and the logs would accidentally become much less readable as a result (which is greatly annoying when you need to debug a production issue). 

A change in semantics to provide a universal implementation would annihilate this use-case, and I think it warrants a different type. 

Something something don't break user-space.

guizmaii:
I didn't read all your comments (sorry).

I 👎 this change because I use `Show` as non-universal `.toString`.
The usage of `toString` is forbidden in my codebases, only `.show` can be used. It makes refactoring safer because is a `Show` instance doesn't exist, the code doesn't compile anymore.

This PR removes this possibly and so we're back to the initial problem: everything as a `.toString`, even when it doesn't make sense.
What is the purpose of the `Show` typeclass in this case? (Rethorical question)

djspiewak:
> I've used Show in the past to constrain coworkers to think of how things should appear in logs, and prevent mis-haps where the type of a value would change and the logs would accidentally become much less readable as a result (which is greatly annoying when you need to debug a production issue).

That use-case remains intact though! I agree this is very important. If you have an explicit `Show` for something and then you change that type, that `Show` will cease compiling. The only difference is that you might be tempted to *not* write an explicit `Show` since you already have an implicit one. But the tradeoff is, now you can use `show` for debugging without having to worry about threading instances everywhere, and if one happens to be available, it'll automatically work.

> A change in semantics to provide a universal implementation would annihilate this use-case, and I think it warrants a different type.

So then we'll have some other type – say, `Debug` – which is useful for rendering debug strings from values. As distinct from `Show` which is… useful for rendering debug strings from *some* values (but not all). See the problem? What I'm *particularly* concerned about is having a separate typeclass which makes people think that `Show` should be used for user-visible rendering or encoding or something similarly problematic.

> What is the purpose of the Show typeclass in this case? (Rethorical question)

I'll answer your question with a question of my own: what is the purpose of `Show` otherwise? For what types does `toString` *not* make sense? (and before you say `Function1`, note that the `toString` for `Function1` prints the pointer identity, which is actually really useful for debugging in most cases)

I guess my premise here is that I actually *like* the fact that everything has a `toString`. It's the only way that `println` debugging works (and also, remember, every IDE debugger uses `toString` as well).

nafg:
There's a big advantage to statically ensuring I'm not stringifying things that shouldn't be by accident.

For instance, I've had bugs where I generated a CSV where things were stringified, and (later, I think) something was made optional, and the CSV would have cells like Some(John Smith)

If there's a universal Show how would that be prevented?

clhodapp:
@nafg That might not be such a great example:

https://github.com/typelevel/cats/blob/dca5b2af3de1087100971c7e2385516c303dcaaa/core/src/main/scala/cats/instances/option.scala#L197

nafg:
True. In fact I was bitten by something like that very recently but it was
while writing new code so I caught it right away. I think it was with
Either.

I would rather a typeclass, whatever it would be called (certainly not
Debug) that would not have instances like that automatically in implicit
scope. Something for user-facing textual descriptions of data.

I had been under the impression that parametric reasoning and Scala's
controllable implicit scoping meant that Show was for anything that
involved "showing" something by way of an A=>String and any further
specificity would be achieved by scoping. In a typeclass coherence world
I'd imagine you'd want either a name less ambiguous than Show (like Debug,
as in zio-prelude) or a type less ambiguous than String. Perhaps a new data
type DebugInfo, either a newtype or hold something richer than an old
String. If it's meaning is broad though then you need to have control over
different instances in different scopes.

nafg:
In any case, example aside, my original point was that refactoring
shouldn't silently change behavior. If I'm using Show user-facing then if I
change something's type I want the compiler to make sure I don't suddenly
start showing the use something I wasn't expecting.

kubukoz:
I'm not entirely convinced, but I'm not strongly opposing. Seeing this in a codebase would likely change my mind.

To be honest, in reality many instances my team defines (esp. in apps, not libraries) are just `Show.fromToString`. Also, if I can put an instance of the new thing next to my type, I can still see where its `Show` is being used - which solves my biggest problem.

djspiewak:
@nafg 

> If there's a universal Show how would that be prevented?

I think what you describe is an interesting case. For starters though, no one should be using `Show` for any sort of encoding *at all*, CSV or otherwise. However, even in the case you describe, it still would be fine because you presumably already have an explicit `Show` for the type which contained the thing which became optional. That `Show` would cease to compile (unless you're using some sort of auto-derivation) once the field was made optional.

Having a fallback doesn't remove all relevant compile-time checks. It just means you now have a default.

> I would rather a typeclass, whatever it would be called (certainly not
Debug) that would not have instances like that automatically in implicit
scope. Something for user-facing textual descriptions of data.

I've used the name `Render` for this concept in the past.

But more precisely, `String` itself is already kind of a terrible type and only useful for either low-level representations or... debugging. So any class with the shape `A => String` is probably in debug-land to begin with. Which... is what `Show` is for.

> In any case, example aside, my original point was that refactoring
shouldn't silently change behavior. 

I agree with this in principle, but at the same time, I would posit that anyone who's code is adversely affected by this change *already* has problems that perhaps they don't realize. It's a bit like the `ExecutionContext` change in 2.13: it only adversely affected code which was already using things in an unspecified or unsupported fashion.

This PR doesn't change the *contract* of `Show`. What it changes is the convenience within that contract, and it changes behavior *outside* of its contract.

joroKr21:
I see two distinct use cases of `Show` - debugging and logging.

Neither is user facing but logging is still code that runs in production and requires some care, such as not building too large strings (performance issues, overloading indices, etc.) and not accidentally showing sensitive information such as personal user data (in Europe, GDPR) or secret configuration.

The premise of this PR is that only debugging is a valid use case and I disagree. You could argue that structured logging is the proper solution to the problem, but that requires at least some (change in) infrastructure and is not under the control of every programmer out there. You could also argue that `Show` is not bulletproof and that is also true but some compile-time guarantee is better than none. The good thing about it is that at least it makes you think what kind of string representation you want. If you automatically add `Show.fromToString` instances, I would argue you shouldn't be using `Show` in the first place - string interpolation would be good enough for you.

Also I think adding a fallback instance will actually make `Show` redundant - why not delete it then? Yes you can use it to customise a type you don't control, but then I bet a lot of application code will mess up the threading of instances through polymorphic code and you will often end up using the fallback instance for type parameters. And then your customised `Show` will not actually work the way you intend it to.

Another unrelated problem I see is that this will mess up with typeclass derivation for `Show`. E.g. `kittens` can derive for you pretty `Show` instances that also print parameter names of case classes. But it takes care to give priority to existing instances. Which means that if we have an existing fallback instance, then derivation will never kick in. `magnolia` works the same way so if anyone is using it for `Show` derivation they would have the same issue.

Overall 👎 for changing `Show`. I have no strong opinion on a new `Debug` typeclass with a fallback.

alexandru:
In our $work project, we banned `.toString` via linting, because this risks logging sensitive data to standard output or via logging. We use `Show` when we want `.toString`, along without our own custom type-class (named `LogShow`), because it's opt-in.

That said `Show` is actually not that great for our use-case either, even when opt-in, because the default `Show[String]` instance is a problem. But we tolerated `Show`.

If this goes through, we'd have to ban `Show` via linting too. And this makes it a breaking change IMO.

As for whether we should do it or not, I don't know, it certainly would make it easier, but at that point one has to wonder what's the point of `Show`, other than to be able to work with newtype encodings, which people don't do anyway.

alexandru:
As for `IO.println`, what would be the difference between this, and simply doing:

```scala
object IO {
  def println[A](a: A)(implicit A: Show[A] = Show.fromToString[A]): IO[Unit] = ??
}
```

If this signature feels wrong, why wouldn't universal `Show` feel wrong too?

(here I'm probably wrong 🙂 but it's something I'm genuinely wondering)

Baccata:
> That use-case remains intact though! I agree this is very important. If you have an explicit Show for something and then you change that type, that Show will cease compiling. 

I think I failed to communicate my concern. Here's a dummy example explaining what I mean by "when the type of a value changes".  

#### At time t0 

```
def foo(x : Int, y : String) = for {
    ... 
    _ <- log(show"x: $x, y: $y")  // (or equivalent macro-expanded call) 
   ... 
} yield (...) 
```

#### At time t1

Change in requirements, string `y` now should be lazily computed for whatever reason 

```
def foo(x : Int, y : IO[String]) = for {
    ... 
    _ <- log(show"x: $x, y: $y") // still compiles and now logs something that is not quite as useful 
   ... 
    _ <- bar(y) 
} yield (...) 
```

With a universal `Show`, this `log` call now still compiles. Build is green so people overlook the line (sometimes it won't even appear in github because it's literally not changed).  I'd have rather the information had been communicated for people to make a decision on whether the log statement should have been moved or amended .

>  It's a bit like the ExecutionContext change in 2.13: it only adversely affected code which was already using things in an unspecified or unsupported fashion.

I get the point, but I disagree with the argument. The EC change was introduced in a significant language update which begs for a closer inspection of the changelog.

### A fallacious argument of mine :trollface: 

Playing devil's advocate, but similar [arguments](https://github.com/typelevel/cats-effect/issues/555#issuecomment-500096831) that were in favour of `Blocker` being introduced in CE could be used to reject this proposal. 

### How about ?  

``` 
object Show { 
    // sealed to ensure consistency with `Show`
    sealed trait Loose[A] {  // naming is hard
         def show(a: A) : String 
    } 
    object Loose extends LowPriorityLoose { 
       implicit def fromShow[A](show: Show[A]) : Loose[A] = (a : A) => show.show(a) 
    } 
    trait LowPriorityLoose {
       implicit def universal = ??? 
    }
} 
```

SystemFw:
Although I don't think `Show` should be used for anything rather than `println` debugging, I would actually go with @alexandru's solution and fight this battle another day

dimitarg:
I think there is a pedagogical aspect to this. I.e. should a pure functional programming library expose something which breaks parametricity? 
Arguably, `Show.fromToString` already does that. In my humble opinion it's questionable to continue further in that direction and make the escape hatch even easier.

arosien:
> I think there is a pedagogical aspect to this. I.e. should a pure functional programming library expose something which breaks parametricity?
> 
> I.e. you've now postulated
> 
> `forall a. a -> String`
> 
> which does not just return some constant, with all the consequences.

@dimitarg i'm not sure that's applicable, as every Scala type has a `toString` method, so `_.toString` could be a valid implementation of that signature.

tpolecat:
> @dimitarg i'm not sure that's applicable, as every Scala type has a `toString` method, so `_.toString` could be a valid implementation of that signature.

I think it's a reasonable statement. For parametricity to hold we must pretend that there are no operations on `Any`, so `forall A. A => String` must necessarily be a constant function.

arosien:
@dimitarg @tpolecat ah, i misread the part about "should".

acamillo:
> You can no longer fail compilation when using .show on a type which just… doesn't have one 

Honestly I always thought the major selling point of `Show` was to provide a kind of safety when dealing with hierarchy of classes. In that case figuring out what `toString` method would be used  is like a round at "roulette", really unpredictable.

From your comment I get to understand we loose that safety.

johnynek:
I'm -1 on this. Since show is lawless, a change to default implementations is even harder for people to reason about. Previously, about the only guarantee you got was that you didn't "just get toString". This breaks even that guarantee.

I'm +1 on the other approaches (another typeclass that can resolve from Show but fallback to toString, called Loose above), or enabling this by a specific import, e.g. an implicit def that is not in normal implicit scope.

kubukoz:
We should probably tackle this by the time CE3 releases... I think @alexandru's solution (default implicit parameter) was quite popular, should we just make `println` in IO / Console use that?

larsrh:
Since there is no consensus on this change, I'm going to close it (sorry Daniel!). If anyone feels very strongly about the opposite, feel free to reopen.

PR #3630, opened 2020-10-07
FunctionK: methods to narrow input / widen output types

bplommer:
This can also be done without `asInstanceOf` but that requires a wasteful allocation.


LukaJCB (approved):
Thanks!

Maybe a little test would be nice?

djspiewak (approved):
We should consider adding something similar to `Profunctor` if it isn't already there.

bplommer:
> Maybe a little test would be nice?

In principle yes, but I honestly can't think what to test given the lack of any runtime behaviour.

joroKr21:
> We should consider adding something similar to Profunctor if it isn't already there.

`ProfunctorK` ❤️

djspiewak:
> ProfunctorK ❤️

I really wish we had polykinds :-D

PR #3636, opened 2020-10-17
Port tests to Dotty (the 2nd)

larsrh:
Supersedes #3552.

Lots of the ground work done by @LukaJCB. I just added a few commits.

@smarter writes:

> M1 should be out next week, the milestone date hasn't been updated

I still think we should get this in sooner than later, because other development in `master` may easily break this again (see e.g. a449e16 which was caused by #3620).


LukaJCB (approved):

tests/src/test/scala-2.x/cats/tests/FunctionKLiftSuite.scala
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  test("hygiene") {
+    trait FunctionK
+    def optionToList[A](option: Option[A]): List[A] = option.toList
+    val fOptionToList = cats.arrow.FunctionK.lift(optionToList _)
+    forAll { (a: Option[Int]) =>
+      assert(fOptionToList(a) === (optionToList(a)))
+    }
+  }
```
satorg, line 36: @larsrh this trait FunctionK is not used here. I wonder is left here intentionally or just an oversight?

larsrh: That's on purpose, since it is a hygiene test. We're checking whether the macro uses the correct `FunctionK` trait.

PR #3639, opened 2020-10-18
Created documentation for monaderror, merged it with applicative erro…

dhinojosa:
This is the documentation for `MonadError` which is merged into one document. Also didn't realize there is menu.yaml file which needed to be updated so I turned on the page for Monad and Applicative Error.

### DIFF
--- a/docs/src/main/mdoc/typeclasses/applicativemonaderror.md
+++ b/docs/src/main/mdoc/typeclasses/applicativemonaderror.md
@@ -5,13 +5,15 @@ section: "typeclasses"
 source: "core/src/main/scala/cats/ApplicativeError.scala"
 scaladoc: "#cats.ApplicativeError"
 ---
-# ApplicativeError
+# ApplicativeError and Monad Error
 
-## Description
+## Applicative Error
+
+### Description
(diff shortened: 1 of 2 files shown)


docs/src/main/mdoc/typeclasses/applicativemonaderror.md
```diff
@@ -5,13 +5,15 @@ section: "typeclasses"
 source: "core/src/main/scala/cats/ApplicativeError.scala"
 scaladoc: "#cats.ApplicativeError"
 ---
-# ApplicativeError
+# ApplicativeError and Monad Error
```
Daenyth, line 8:
```suggestion
# ApplicativeError and MonadError
```

⛏️ Tiny detail but might affect search, so who knows

Daenyth (approved):
:shipit: 

I think the `liftTo` on Either and Option for MonadError deserve a mention: For example `someEither.liftTo[IO]` is a very common pattern, that's worth being in the docs. But not worth blocking a merge on that detail. Anything is better than what we have today

larsrh (approved):
Thanks a lot!

PR #3640, opened 2020-10-18
add .bsp/ directory to gitignore, I suspect it is because of mdoc

dhinojosa:

Directory .bsp is created,  it is usually empty when creating the microsite.  Not sure it is useful


LukaJCB (approved):

LukaJCB:
Thanks!

PR #3644, opened 2020-10-21
Add Kleisli.fromFunction, WriterT.liftFunctionK

kubukoz:


LukaJCB (approved):
LGTM!

core/src/main/scala/cats/data/Kleisli.scala
```diff
@@ -257,6 +257,18 @@ sealed private[data] trait KleisliFunctions {
...
+  /**
+   * Lifts a function to a Kleisli.
+   * {{{
+   * scala> import cats.data.Kleisli
+   * scala> val stringify = Kleisli.fromFunction[Option, Int, String](_.toString)
+   * scala> stringify.run(42)
+   * res0: Option[String] = Some(42)
+   * }}}
+   */
+  def fromFunction[M[_]: Applicative, A, R](f: R => A): Kleisli[M, R, A] =
```
rossabaker, line 270: Is this a candidate for some partial application tomfoolery so we might be able to say `Kleisli.fromFunction[Option]` and infer the `R` and `A`?

kubukoz: it's a good idea, I will give it a closer look

barambani, line 270:
```suggestion
  def fromFunction[M[_]: Applicative, R, A](f: R => A): Kleisli[M, R, A] =
```
should fix the doctest

kubukoz: Actually, I don't think it will (R is still unknown because there are no hints of the type on the LHS). I'll make R required in the partially applied magic.

kubukoz: I guess we can't infer R anyway unless there's a type annotation for the whole thing. Probably the best we can do is `Kleisli.fromFunction[Option, Int]` and inferring the result.

kubukoz:
@rossabaker when you have time, I'd appreciate another look :)

LukaJCB (approved):

rossabaker (approved):
I like it. :+1:

PR #3647, opened 2020-10-26
Implement traverseTap

majk-p:
I'd like to introduce `traverseTap`. I found it would be useful for me when attempting to perform debug log executing an effect, but dropping the result.

I'd like to add some tests, could someone please advice the best place to add them?

### DIFF
--- a/core/src/main/scala/cats/Traverse.scala
+++ b/core/src/main/scala/cats/Traverse.scala
@@ -38,6 +38,25 @@ import scala.annotation.implicitNotFound
   def traverse[G[_]: Applicative, A, B](fa: F[A])(f: A => G[B]): G[F[B]]
 
   /**
+   * Given a function which returns a G effect, thread this effect
+   * through the running of this function on all the values in F,
+   * returning an F[A] in a G context, ignoring the values
+   * returned by provided function.
+   *
+   * Example:
+   * {{{
+   * scala> import cats.implicits._
+   * scala> import java.io.IOException
+   * scala> type IO[A] = Either[IOException, A]
+   * scala> def debug(msg: String): IO[Unit] = Right(())
+   * scala> List("1", "2", "3").traverseTap(debug)
+   * res1: IO[List[String]] = Right(List(1, 2, 3))
+   * }}}
+   */
+  def traverseTap[G[_]: Applicative, A, B](fa: F[A])(f: A => G[B]): G[F[A]] =
+    traverse(fa)(a => Applicative[G].as(f(a), a))
+
+  /**
    * A traverse followed by flattening the inner result.
    *
    * Example:
@@ -160,6 +179,8 @@ object Traverse {
     val typeClassInstance: TypeClassType
     def traverse[G[_], B](f: A => G[B])(implicit ev$1: Applicative[G]): G[F[B]] =
       typeClassInstance.traverse[G, A, B](self)(f)
+    def traverseTap[G[_], B](f: A => G[B])(implicit ev$1: Applicative[G]): G[F[A]] =
+      typeClassInstance.traverseTap[G, A, B](self)(f)
     def flatTraverse[G[_], B](f: A => G[F[B]])(implicit G: Applicative[G], F: FlatMap[F]): G[F[B]] =
       typeClassInstance.flatTraverse[G, A, B](self)(f)(G, F)
     def sequence[G[_], B](implicit ev$1: A <:< G[B], ev$2: Applicative[G]): G[F[B]] =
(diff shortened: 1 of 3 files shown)


core/src/main/scala/cats/Traverse.scala
```diff
@@ -37,6 +37,26 @@ import scala.annotation.implicitNotFound
...
 
+  /**
+   * Given a function which returns a G effect, thread this effect
+   * through the running of this function on all the values in F,
+   * returning an F[A] in a G context, ignoring the values
+   * returned by provided function.
+   *
+   * Example:
+   * {{{
+   * scala> import cats.implicits._
```
kubukoz, line 48: Double import (counts as unused on 2.13, which is why I think the build failed on that version)

core/src/main/scala/cats/Traverse.scala
```diff
@@ -37,6 +37,26 @@ import scala.annotation.implicitNotFound
...
+   * scala> import java.io.IOException
+   * scala> type IO[A] = Either[IOException, A]
+   * scala> def debug(msg: String): IO[Unit] = Right(())
+   * scala> List("1", "2", "3").traverseTap(debug)
+   * res1: IO[List[String]] = Right(List(1, 2, 3))
+   * }}}
+   */
+  def traverseTap[G[_]: Applicative, A, B](fa: F[A])(f: A => G[B]): G[F[A]] =
+    traverse(fa)(a => Applicative[G].as(f(a), a))
```
kubukoz, line 58: Technically we could use `Functor[G].as`, but we have the applicative anyway, so it probably doesn't matter.

kubukoz (approved):

core/src/main/scala/cats/Traverse.scala
majk-p: Would that work even if `traverse` requires an `Applicative` for `G`?

kubukoz: it'll work because Functor <:< Applicative

core/src/main/scala/cats/Traverse.scala
majk-p: fixed now

LukaJCB (approved):
I like it 👍

johnynek (dismissed):

johnynek:
my personal opinion is that each function like this should be added to the laws, and the law is just a duplication of the implementation. That way, if someone overrides to make a more optimized version, we have a law that it behaves exactly as the direct implementation.

majk-p:
> my personal opinion is that each function like this should be added to the laws, and the law is just a duplication of the implementation. That way, if someone overrides to make a more optimized version, we have a law that it behaves exactly as the direct implementation.

That make sense. Implemented in https://github.com/typelevel/cats/pull/3647/files#diff-5ee4897e7202dfd0b4fd61f8afac00bbc7a9b28dc9309b3006f3ffc12b2cd36eR55. Couldn't find a better name for this law.

johnynek:
this is great! Unfortunately, we also need to add a line here:

https://github.com/typelevel/cats/blob/e0f0a831eb6c0590f8b506b2403cf03175acfcfd/laws/src/main/scala/cats/laws/discipline/TraverseTests.scala#L59

to actually call this code when the tests are run.

majk-p:
Thank you for pointing this out @johnynek. Sorry I've missed that, I'm not yet familiar enough with the project.
Just fixed that.

johnynek (dismissed):
Thanks! This looks useful for logging.

majk-p:
Had to make one more push to avoid breaking binary compatibility in laws

laws/src/main/scala/cats/laws/discipline/NonEmptyTraverseTests.scala
```diff
@@ -38,6 +38,7 @@ trait NonEmptyTraverseTests[F[_]] extends TraverseTests[F] with ReducibleTests[F
     EqA: Eq[A],
     EqB: Eq[B],
     EqXYFC: Eq[X[Y[F[C]]]],
+    EqXFA: Eq[X[F[A]]],
```
johnynek, line 41: I think this is still failing and binary incompatible.

kubukoz: looks unused to me, so probably just a leftover

majk-p: Thank you for pointing it out. It was a leftover, just removed that :+1:

johnynek (approved):

kubukoz (approved):

PR #3661, opened 2020-11-03
Fix typo in the OptionT#foldF scaladoc

fixxer:
According to the signature, the scaladoc should say "Transform this `OptionT[F, A]` into a `F[B]`"


LukaJCB (approved):
thanks!

PR #3666, opened 2020-11-05
Things weren't working, now I think they are

djspiewak:
This would replace 2.3.0-M1 entirely. We would need to release an M2. This also builds on 3.0.0-M1.


LukaJCB (dismissed):

djspiewak:
Formatted, but don't merge this yet. Simulacrum is still problematic

mpilquist (approved):

PR #3672, opened 2020-11-09
enable DottyJS

larsrh:
This is cool but I think we should wait until Scala 3.0.0-M2 for a proper release to get rid of all matrix exclusion shenanigans.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -24,10 +24,10 @@ scalafixDependencies in ThisBuild += "org.typelevel" %% "simulacrum-scalafix" % 
 
 val scalaCheckVersion = "1.15.1"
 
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -69,12 +69,6 @@ ThisBuild / githubWorkflowBuildMatrixExclusions ++=
     ) // 3.0.0-M1 doesn't work on JDK 14+
   }
 
-// exclude DottyJS for now
```
larsrh, line 72: @djspiewak this exclusion was actually redundant

djspiewak: I really really wanted excluded.

PR #3673, opened 2020-11-09
tweak Band impl so it can more easily be combined with Monoid in a single instance

bpholt:
This is helpful for noncommutative idempotent instances.

Without this change, the `monoid.repeat0` law fails for these instances unless the `combineN` method is overridden (which is why it worked with the original `Monoid[Order[A]] with Band[Order[A]]` instance).


LukaJCB (approved):
good catch!

PR #3677, opened 2020-11-11
Added EmptyK instance for Map

LMnet:
I added the missing `EmptyK` instance for a `Map`.

Also, I made the `Empty.fromEmptyK` function implicit. I don't know why it was not marked as implicit, and it looks like it is a pretty safe change. This change automatically added an `Empty` instance for a `Map` too.

### DIFF
--- a/alleycats-core/src/main/scala/alleycats/Empty.scala
+++ b/alleycats-core/src/main/scala/alleycats/Empty.scala
@@ -21,7 +21,7 @@ object Empty extends EmptyInstances0 {
   def apply[A](a: => A): Empty[A] =
     new Empty[A] { lazy val empty: A = a }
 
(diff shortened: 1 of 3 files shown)


larsrh (approved):

larsrh:
Thanks!

alleycats-core/src/main/scala/alleycats/Empty.scala
```diff
@@ -21,7 +21,7 @@ object Empty extends EmptyInstances0 {
   def apply[A](a: => A): Empty[A] =
     new Empty[A] { lazy val empty: A = a }
 
-  def fromEmptyK[F[_], T](implicit ekf: EmptyK[F]): Empty[F[T]] = ekf.synthesize[T]
+  implicit def fromEmptyK[F[_], T](implicit ekf: EmptyK[F]): Empty[F[T]] = ekf.synthesize[T]
```
joroKr21, line 24: Looks like kittens can't handle this: typelevel/kittens#299

Issue #3650, opened 2020-10-30
Add `grouped` to NonEmptyCollection


bbstilson:
I chatted with @johnynek on Gitter about adding this. For example, in Scala one can call `grouped` on any `Iterable`:

```scala
@ val xs = List(1,2,3,4) 
xs: List[Int] = List(1, 2, 3, 4)

@ xs.grouped(2).toList 
res1: List[List[Int]] = List(List(1, 2), List(3, 4))
```

My current plan is for grouped to return an `Iterable[NE[A]]`, but I'm also open to returning the same collection type, e.g.,`NE[NE[A]]`. As for handling invalid grouped size, I am open to ideas. I'm not a fan of throwing exceptions. One option is to make the name obvious, something like `unsafeGrouped`. Another is to return some error handler like `Either`, e.g., `Either[Iterator[NE[A]]]`.

A (likely bad) option would be to use a Refined type like:

```scala
def grouped[A](size: Int Refined Positive): Iterator[NE[A]]
```

This is a bad option as bringing in another library is likely not desirable.

kubukoz:
I think we'll be fine with a runtime check on `size`, we do that on `Monoid.combineN`, AFAIK. Users will be able to use a refined type on their own, plus `refined.auto._` for convenient auto-unwrapping.

PR #3679, opened 2020-11-12
add grouped to NonEmptyCollection

bbstilson:
Fixes https://github.com/typelevel/cats/issues/3650

This PR adds `grouped` to all implementers of `NonEmptyCollection`. It seemed common to jump out of a cats data type to the scala representation, do some transformation/calculation, then jump back to the cats type, so that's mainly what I did here.


tests/src/test/scala/cats/tests/NonEmptyCollectionSuite.scala
```diff
@@ -153,6 +153,12 @@ abstract class NonEmptyCollectionSuite[U[+_], NE[+_], NEC[x] <: NonEmptyCollecti
     }
   }
 
+  test("grouped is consistent with iterator.toList.grouped") {
+    forAll { (is: NE[Int], i: Int) =>
+      (i > 0) ==> assert(is.grouped(i).toList.map(toList) === is.iterator.toList.grouped(i).toList)
```
larsrh, line 158: The above snippet gives a type error, but this should fix it

```suggestion
      (i > 0) ==> is.grouped(i).toList.map(toList) === is.iterator.toList.grouped(i).toList
```

larsrh:
Got bitten by a missing set of parentheses, sorry for the noise.

core/src/main/scala/cats/data/NonEmptyChain.scala
```diff
@@ -384,6 +384,24 @@ class NonEmptyChainOps[A](private val value: NonEmptyChain[A])
   final def groupBy[B](f: A => B)(implicit B: Order[B]): NonEmptyMap[B, NonEmptyChain[A]] =
     toChain.groupBy(f).asInstanceOf[NonEmptyMap[B, NonEmptyChain[A]]]
 
+  /**
+   * Partitions elements in fixed size `NonEmptyChain`s.
+   *
+   * {{{
+   * scala> import cats.data.NonEmptyChain
+   * scala> import cats.implicits._
+   * scala> val nel = NonEmptyChain.of(12, -2, 3, -5)
```
larsrh, line 393: There's no `NonEmptyChain.of` method, this is why the build is failing.

larsrh: #3771

larsrh:
@LukaJCB Tagged you as a reviewer because of the MiMa exception.

larsrh (approved):

larsrh:
Thanks Brandon!

PR #3680, opened 2020-11-12
Add more group-like methods to NonEmptyList & Chain

BalmungSan:
Scope:

+ [x] `NonEmptyList#groupMap`
+ [x] `NonEmptyList#groupMapNem`
+ [x] `NonEmptyList#groupMapReduceWith`
+ [x] `NonEmptyList#groupMapReduceWithNem`
+ [x] `NonEmptyList#groupMapReduce` _(requires a **Semigroup**)_
+ [x] `NonEmptyList#groupMapReduceNem`  _(requires a **Semigroup**)_
+ [x] `Chain#groupMap`
+ [x] `NonEmptyChain#groupMapNem`
+ [x] `Chain#groupMapReduceWith` 
+ [x] `NonEmptyChain#groupMapReduceWithNem`
+ [x] `Chain#groupMapReduce` _(requires a **Semigroup**)_
+ [x] `NonEmptyChain#groupMapReduceNem` _(requires a **Semigroup**)_


core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -376,40 +376,79 @@ final case class NonEmptyList[+A](head: A, tail: List[A]) extends NonEmptyCollec
    * res0: Boolean = true
    * }}}
    */
-  def groupBy[B](f: A => B)(implicit B: Order[B]): SortedMap[B, NonEmptyList[A]] = {
-    implicit val ordering: Ordering[B] = B.toOrdering
-    var m = TreeMap.empty[B, mutable.Builder[A, List[A]]]
+  def groupBy[K](key: A => K)(implicit K: Order[K]): SortedMap[K, NonEmptyList[A]] =
```
BalmungSan, line 379: I am not sure if the change of the name of the generic and the function can affect binary compatibility.
And if the break of source compatibility is acceptable.

core/src/main/scala/cats/syntax/list.scala
```diff
@@ -31,24 +32,95 @@ final class ListOps[A](private val la: List[A]) extends AnyVal {
...
+   * {{{
+   * scala> import scala.collection.immutable.SortedMap
+   * scala> import cats.implicits._
+   * scala> val list = List("Hello", "World", "Goodbye", "World")
+   * scala> val expectedResult = SortedMap("goodbye" -> 1, "hello" -> 1, "world" -> 2)
+   * scala> list.groupMapReduce(_.strip.toLowerCase)(_ => 1) === expectedResult
+   * res0: Boolean = true
+   * }}}
+   */
+  def groupMapReduce[K, B](key: A => K)(f: A => B)(implicit K: Order[K], B: Semigroup[B]): SortedMap[K, B] =
```
BalmungSan, line 90: Not sure how to name this one so it doesn't clash with the one introduced by the stdlib in `2.13`

core/src/main/scala/cats/syntax/list.scala
```diff
@@ -31,24 +32,95 @@ final class ListOps[A](private val la: List[A]) extends AnyVal {
...
+   * scala> val list = List("Hello", "World", "Goodbye", "World")
+   * scala> val expectedResult = SortedMap("goodbye" -> 1, "hello" -> 1, "world" -> 2)
+   * scala> list.groupMapReduceWith(_.strip.toLowerCase)(_ => 1)(_ + _) === expectedResult
    * res0: Boolean = true
    * }}}
    */
-  def groupByNel[B](f: A => B)(implicit B: Order[B]): SortedMap[B, NonEmptyList[A]] = {
-    implicit val ordering: Ordering[B] = B.toOrdering
-    toNel.fold(SortedMap.empty[B, NonEmptyList[A]])(_.groupBy(f))
+  def groupMapReduceWith[K, B](key: A => K)(f: A => B)(combine: (B, B) => B)(implicit K: Order[K]): SortedMap[K, B] = {
```
BalmungSan, line 110: Not sure if it is worth it to add this method.

It is mostly the same as the one in the stdlib, just that this returns a **SortedMap**.
_(And that it is available in `2.12`)_

BalmungSan:
I should have targeted `master` instead of `2.2x` right?
I will replicate the work in another PR.

larsrh:
Sorry for the long wait, I will take a look at this PR.

core/src/main/scala/cats/data/Chain.scala
```diff
@@ -356,22 +356,111 @@ sealed abstract class Chain[+A] {
...
+   * And finally they are all reduced into a single value
+   * using their `Semigroup`
+   *
+   * {{{
+   * scala> import scala.collection.immutable.SortedMap
+   * scala> import cats.data.Chain
+   * scala> import cats.implicits._
+   * scala> val chain = Chain("Hello", "World", "Goodbye", "World")
+   * scala> val expectedResult = SortedMap("goodbye" -> 1, "hello" -> 1, "world" -> 2)
+   * scala> val result = chain.groupMapReduce(_.strip.toLowerCase)(_ => 1)
```
larsrh, line 422: The doctest appears to fail in multiple places:

```
[error] /home/runner/work/cats/cats/core/.jvm/target/scala-2.13/src_managed/test/cats/data/ChainDoctest.scala:70:41: value strip is not a member of String
[error] did you mean trim?
[error]     val result = chain.groupMapReduce(_.strip.toLowerCase)(_ => 1)
[error]                                         ^
```

BalmungSan: Ups, sorry.

I will take a look at that ASAP.

BalmungSan:
Hi, @larsrh thanks for looking into this!

I have fixed the documentation error.
AFAIK this now passes all checks in `2.12`, it doesn't in `2.13` due to the naming clash with the stdlib _(which I asked about above )_.

Let me know what I can do to fix that :)

larsrh:
My suggestion to move forward on this: could you remove all the extensions to the stdlib `List` from this PR and open another one? This one should only contain the ones for `NonEmptyList` and `Chain`. Then, we can discuss how to proceed for the smaller subset in the other PR and merge this one without conflicts.

BalmungSan:
@larsrh 

> could you remove all the extensions to the stdlib List from this PR

Done and the build is green now!

> and open another one? 

I will when we merge this to avoid some rebasing.

core/src/main/scala/cats/data/Chain.scala
```diff
@@ -356,22 +356,111 @@ sealed abstract class Chain[+A] {
...
+   * scala> val result = chain.groupBy(_ >= 0)
+   * scala> result === expectedResult
+   * res0: Boolean = true
+   * }}}
    */
-  final def groupBy[B](f: A => B)(implicit B: Order[B]): SortedMap[B, NonEmptyChain[A]] = {
-    implicit val ordering: Ordering[B] = B.toOrdering
-    var m = SortedMap.empty[B, NonEmptyChain[A]]
-    val iter = iterator
+  final def groupBy[K](key: A => K)(implicit K: Order[K]): SortedMap[K, NonEmptyChain[A]] =
```
larsrh, line 372: I'm really torn on the renaming.

As for the type parameter, it does not break source compatibility on Scala 2, but Dotty has named type parameters.

The method parameter however is a source-breaking change, even though it is unlikely that someone would give a name to the singleton parameter of a list.

Having said that, I still think we should err on the side of being conservative, so I'd rather avoid the renaming.

larsrh (changes requested):
I have only one point regarding the renamed parameters (it applies to all instances).

larsrh (approved):

larsrh:
Thanks a lot, merging this.

PR #3694, opened 2020-11-25
Revert "Issue 3353: Investigate antisymmetry failure for UpperBounded for Duration"

djspiewak:
Reverts typelevel/cats#3470

I think this is what broke the build


mpilquist (approved):
We'll need that PR to be updated for munit instead of scalacheck. That can come after 2.3.0 release.

LukaJCB (approved):

barambani (approved):

PR #3696, opened 2020-11-26
Add `ApplicativeThrow` and `MonadThrow` to cats package object (from cats-effect)

agustafson:
I noticed that the `ApplicativeThrow` and `MonadThrow` type aliases had been added to cats-effect:
https://github.com/typelevel/cats-effect/blob/series/2.x/core/shared/src/main/scala/cats/effect/package.scala#L41-L43

I felt that it made more sense for it to be cats as `ApplicativeError` and `MonadError` are in cats rather than cats-effect.

Also, it seems like `ApplicativeThrow` has erroneously had an unused `A` type param added, so I figured it was best to remove it. I'm not sure if that will clash or cause problems if both `cats._` and `cats.effect._` are imported. I'll open a separate PR against cats-effect 2.x branch to remove the `A`

> This is a kind reminder to run `sbt +prePR` and commit the changed files, if any, before submitting.

As an aside, I tried running this task but it failed on master as well as my PR, but the code all seems to be formatted correctly.

### DIFF
--- a/core/src/main/scala/cats/Invariant.scala
+++ b/core/src/main/scala/cats/Invariant.scala
@@ -127,11 +127,11 @@ object Invariant extends ScalaVersionSpecificInvariantInstances with InvariantIn
     cats.instances.function.catsStdContravariantMonoidalForFunction1[R]
   implicit def catsFunctorForPair: Functor[λ[P => (P, P)]] = cats.instances.tuple.catsDataFunctorForPair
 
-  implicit def catsInstancesForTry: MonadError[Try, Throwable] with CoflatMap[Try] =
+  implicit def catsInstancesForTry: MonadThrow[Try] with CoflatMap[Try] =
     cats.instances.try_.catsStdInstancesForTry
   implicit def catsInstancesForFuture(implicit
     ec: ExecutionContext
-  ): MonadError[Future, Throwable] with CoflatMap[Future] =
+  ): MonadThrow[Future] with CoflatMap[Future] =
     cats.instances.future.catsStdInstancesForFuture(ec)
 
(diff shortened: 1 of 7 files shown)


agustafson:
It'd also be kinda nice if we could add a
```scala
object MonadThrow {
  def apply[F[_]](implicit ev: MonadThrow[F]): MonadThrow[F] = ev
}
```
but wasn't sure where we might put that

bplommer:
If/when this is released I'd suggest we remove these from CE3 and change the ones in CE2 to be aliases for these.

bplommer:
> It'd also be kinda nice if we could add a
> 
> ```scala
> 
> object MonadThrow {
> 
>   def apply[F[_]](implicit ev: MonadThrow[F]): MonadThrow[F] = ev
> 
> }
> 
> ```
> 
> but wasn't sure where we might put that

Good idea - I think it could either go directly below the alias or in a `package cats {}` block in the same file directly below the package object

agustafson:
> > It'd also be kinda nice if we could add a
> > ```scala
> > object MonadThrow {
> > 
> >   def apply[F[_]](implicit ev: MonadThrow[F]): MonadThrow[F] = ev
> > 
> > }
> > ```
> > 
> > 
> > but wasn't sure where we might put that
> 
> Good idea - I think it could either go directly below the alias or in a `package cats {}` block in the same file directly below the package object

Yeah, I wondered if putting it in the same file as `ApplicativeError/MonadError` made sense as an alternative. Just so that the package object didn't get too crowded.

LukaJCB (approved):
I like it 👍

djspiewak (approved):
I like this a lot. So much better than having it in Cats Effect

PR #3698, opened 2020-11-28
Add addLeft/addRight for Ior

### LINKED ISSUE #3689: addLeft/addRight convenience methods for Iors?
Whilst contributing to a cats-based codebase that makes extensive use of Iors (https://github.com/gemini-hlsw/gsp-graphql/pull/96) I found myself needing a "addLeft" function, which given an Ior would either `combine` the new value with the existing LHS or in the case where it was a `Right`, turn it in to a `Both` with the new value on the left, ie

```scala
  def addLeft[A: Semigroup, B](ior: Ior[A,B], a: A): Ior[A, B] = ior match {
    case Ior.Right(r) => Both(a, r)
    case Both(l, r) => Both(l combine a, r)
    case Ior.Left(l) => Ior.left(l combine a)
  }
```

Would this be something that would make sense to include within Cats itself?

keirlawson:
Fixes #3689

### DIFF
--- a/core/src/main/scala/cats/data/Ior.scala
+++ b/core/src/main/scala/cats/data/Ior.scala
@@ -99,6 +99,58 @@ sealed abstract class Ior[+A, +B] extends Product with Serializable {
     fold(Ior.both(_, right), _ => Ior.right(right), (a, _) => Ior.both(a, right))
 
   /**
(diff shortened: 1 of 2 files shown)


LukaJCB (approved):
Looks good to me 👍

PR #3699, opened 2020-11-30
Allow finding a list of maxiumum/minimum items for Foldable/Reducible

Billzabob:
Whereas before you could only retrieve a single maximum or minimum item for a Foldable/Reducible, this allows getting all of the maximum or minimum items, since there can be several that are all equal.


larsrh (approved):
LGTM, thanks for your contribution!

larsrh (approved):

PR #3702, opened 2020-12-01
specialized EqOps

YannMoisan:
- `1L === 2L` triggers boxing of both longs and so costs 2 allocations
- `Eq.eqv(1L, 2L)` do not trigger boxing because `Eq` is specialized

This PR aims at fixing that by using specialization

As mentioned by @djspiewak , the code has been moved into source tree specific folder in order to _specialize_ only for Scala 2.x (and not Scala 3 that do not support it).

There are 3 smalls changes
- method `catsSyntaxEq` is specialized
- class `EqOps` is specialized
- class `EqOps` is now a case class (otherwise `@specialized` do not have the expected effect)


YannMoisan:
retest

djspiewak:
At present, Scala 3 doesn't support specialization, so this will need to be source tree specific. I do think this is a good idea though, particularly since Eq itself is already specialized!

core/src/main/scala-2.x/src/main/scala/cats/syntax/eq.scala
```diff
@@ -0,0 +1,19 @@
+package src.main.scala.cats.syntax
```
djspiewak, line 1: :-) I think intellij is drunk.

djspiewak (changes requested):
You'll also need to add a non-specialized version of `EqOps` within the scala-3.x source directory. Also worth noting that you only need to push `EqOps` into the scala version-specific directories; `EqSyntax` works across both.

YannMoisan:
@djspiewak I've fixed the PR

LukaJCB (approved):
thanks!

YannMoisan:
@djspiewak Just in case you miss the notification, the requested changes have been taken into account.

larsrh (approved):

larsrh:
Thanks, @YannMoisan!

larsrh:
This change is source-breaking, whenever someone overrides it:

```
[error] /home/runner/work/cats-testkit-scalatest/cats-testkit-scalatest/core/src/main/scala/cats/tests/CatsSuite.scala:61:29: Type parameter has to be specialized at least for the same types as in the overridden method. Missing types: Double, Float, Int, Long
[error]   override def catsSyntaxEq[A: Eq](a: A): EqOps[A] = new EqOps[A](a)
```

Unfortunately it also demands that anyone who overrides it has to do it depending on the Scala version.

PR #3708, opened 2020-12-08
Use as rather than map and discard

johnynek:
I noticed a few places where we do `map(f)(_ => a)`. This is the same as Functor.as, but it hides it from the typeclass implementation. Some types can optimize as by discarding any trailing map operators from their internal AST (such as parsers or other compute types that are not evaluated until later). By not calling typeclass methods we potentially slow those implementations down.


rossabaker (approved):
A scalafix linter for patterns like this, or `.pure(())`, would be neat.

PR #3709, opened 2020-12-09
WIP: Selective

rossabaker:
Experimental, incomplete integration of `Selective` into the hierarchy.

See #2745

### DIFF
--- a/core/src/main/scala/cats/Monad.scala
+++ b/core/src/main/scala/cats/Monad.scala
@@ -13,9 +13,15 @@ import scala.annotation.implicitNotFound
  * Must obey the laws defined in cats.laws.MonadLaws.
  */
 @implicitNotFound("Could not find an instance of Monad for ${F}")
-@typeclass trait Monad[F[_]] extends FlatMap[F] with Applicative[F] {
+@typeclass trait Monad[F[_]] extends FlatMap[F] with Selective[F] {
   override def map[A, B](fa: F[A])(f: A => B): F[B] =
     flatMap(fa)(a => pure(f(a)))
+
+  override def select[A, B](fab: F[Either[A, B]])(ff: F[A => B]): F[B] =
+    flatMap(fab) {
+      case Left(a)  => map(ff)(_(a))
+      case Right(b) => pure(b)
+    }
 
   /**
    * Execute an action repeatedly as long as the given `Boolean` expression
--- a/core/src/main/scala/cats/Select.scala
+++ b/core/src/main/scala/cats/Select.scala
@@ -1,0 +1,17 @@
+package cats
+
+import simulacrum.{typeclass}
+import scala.annotation.implicitNotFound
+
+@implicitNotFound("Could not find an instance of Select for ${F}")
+@typeclass trait Select[F[_]] extends Apply[F] {
+  def select[A, B](fab: F[Either[A, B]])(ff: F[A => B]): F[B]
+}
+
+object Select {
+  def fromApply[F[_], A, B](fab: F[Either[A, B]])(ff: F[A => B])(implicit F: Apply[F]): F[B] =
+    F.map2(fab, ff) {
+      case (Left(a), f)  => f(a)
+      case (Right(b), f) => b
+    }
+}
--- a/core/src/main/scala/cats/Selective.scala
+++ b/core/src/main/scala/cats/Selective.scala
@@ -1,0 +1,17 @@
+package cats
+
+import simulacrum.{typeclass}
+import scala.annotation.implicitNotFound
+
+@implicitNotFound("Could not find an instance of Selective for ${F}")
+@typeclass trait Selective[F[_]] extends Applicative[F] {
+  def select[A, B](fab: F[Either[A, B]])(ff: F[A => B]): F[B]
+}
+
+object Selective {
+  def fromApplicative[F[_], A, B](fab: F[Either[A, B]])(ff: F[A => B])(implicit F: Applicative[F]): F[B] =
+    F.map2(fab, ff) {
+      case (Left(a), f)  => f(a)
+      case (Right(b), f) => b
+    }
+}
--- a/core/src/main/scala/cats/Selective.scala
+++ b/core/src/main/scala/cats/Selective.scala
@@ -4,14 +4,4 @@ import simulacrum.{typeclass}
 import scala.annotation.implicitNotFound
 
 @implicitNotFound("Could not find an instance of Selective for ${F}")
-@typeclass trait Selective[F[_]] extends Applicative[F] {
-  def select[A, B](fab: F[Either[A, B]])(ff: F[A => B]): F[B]
-}
-
-object Selective {
-  def fromApplicative[F[_], A, B](fab: F[Either[A, B]])(ff: F[A => B])(implicit F: Applicative[F]): F[B] =
-    F.map2(fab, ff) {
-      case (Left(a), f)  => f(a)
-      case (Right(b), f) => b
-    }
-}
+@typeclass trait Selective[F[_]] extends Applicative[F] with Select[F]
--- a/core/src/main/scala/cats/SelectiveError.scala
+++ b/core/src/main/scala/cats/SelectiveError.scala
@@ -1,0 +1,15 @@
+package cats
+
+import cats.syntax.either._
+
+trait SelectiveError[F[_], E] extends ApplicativeError[F, E] with Selective[F] {
+
+  /**
+   * Turns a successful value into an error if it does not satisfy a given predicate.
+   */
+  def ensure[A](fa: F[A])(error: => E)(predicate: A => Boolean): F[A] =
+    select(map(fa) { a: A =>
+      if (predicate(a)) Right(a)
+      else Either.leftUnit
+    })(raiseError(error))
+}
--- a/core/src/main/scala/cats/data/Validated.scala
+++ b/core/src/main/scala/cats/data/Validated.scala
@@ -931,9 +931,8 @@ sealed abstract private[data] class ValidatedInstances extends ValidatedInstance
         fab.leftMap(f)
     }
 
-  implicit def catsDataApplicativeErrorForValidated[E](implicit E: Semigroup[E]): ApplicativeError[Validated[E, *], E] =
-    new ValidatedApplicative[E] with ApplicativeError[Validated[E, *], E] {
-
+  implicit def catsDataSelectiveErrorForValidated[E](implicit E: Semigroup[E]): Selective[Validated[E, *]] with ApplicativeError[Validated[E, *], E] =
+    new ValidatedSelective[E] with ApplicativeError[Validated[E, *], E] {
       def handleErrorWith[A](fa: Validated[E, A])(f: E => Validated[E, A]): Validated[E, A] =
         fa match {
           case Validated.Invalid(e)   => f(e)
@@ -941,6 +940,10 @@ sealed abstract private[data] class ValidatedInstances extends ValidatedInstance
         }
       def raiseError[A](e: E): Validated[E, A] = Validated.Invalid(e)
     }
+
+  @deprecated("Use catsDataSelectiveErrorForValidated", "2.4.0")
+  def catsDataApplicativeErrorForValidated[E](implicit E: Semigroup[E]): ApplicativeError[Validated[E, *], E] =
+    catsDataSelectiveErrorForValidated
 }
 
 sealed abstract private[data] class ValidatedInstances1 extends ValidatedInstances2 {
@@ -953,9 +956,13 @@ sealed abstract private[data] class ValidatedInstances1 extends ValidatedInstanc
       def combine(x: Validated[A, B], y: Validated[A, B]): Validated[A, B] = x.combine(y)
     }
 
-  implicit def catsDataCommutativeApplicativeForValidated[E: CommutativeSemigroup]
-    : CommutativeApplicative[Validated[E, *]] =
-    new ValidatedApplicative[E] with CommutativeApplicative[Validated[E, *]]
+  implicit def catsDataCommutativeSelectiveForValidated[E: CommutativeSemigroup]
+    : Selective[Validated[E, *]] with CommutativeApplicative[Validated[E, *]] =
+    new ValidatedSelective[E] with CommutativeApplicative[Validated[E, *]]
+
+  @deprecated("Use catsDataCommutativeSelectiveForValidated", "2.4.0")
+  def catsDataCommutativeApplicativeForValidated[E: CommutativeSemigroup]: CommutativeApplicative[Validated[E, *]] =
+    catsDataCommutativeApplicativeForValidated
 
   implicit def catsDataPartialOrderForValidated[A: PartialOrder, B: PartialOrder]: PartialOrder[Validated[A, B]] =
     new PartialOrder[Validated[A, B]] {
@@ -1033,6 +1040,15 @@ sealed abstract private[data] class ValidatedInstances2 {
       override def isEmpty[A](fa: Validated[E, A]): Boolean = fa.isInvalid
     }
   // scalastyle:off method.length
+}
+
+private[data] class ValidatedSelective[E: Semigroup] extends ValidatedApplicative[E] with Selective[Validated[E, *]] {
+  override def select[A, B](fab: Validated[E, Either[A, B]])(ff: Validated[E, A => B]): Validated[E, B] =
+    fab match {
+      case Valid(Left(a))  => ff.map(_(a))
+      case Valid(Right(b)) => Valid(b)
+      case i @ Invalid(_)  => i
+    }
 }
 
 private[data] class ValidatedApplicative[E: Semigroup] extends CommutativeApplicative[Validated[E, *]] {
--- a/core/src/main/scala/cats/data/Validated.scala
+++ b/core/src/main/scala/cats/data/Validated.scala
@@ -931,8 +931,8 @@ sealed abstract private[data] class ValidatedInstances extends ValidatedInstance
         fab.leftMap(f)
     }
 
-  implicit def catsDataSelectiveErrorForValidated[E](implicit E: Semigroup[E]): Selective[Validated[E, *]] with ApplicativeError[Validated[E, *], E] =
-    new ValidatedSelective[E] with ApplicativeError[Validated[E, *], E] {
+  implicit def catsDataSelectiveErrorForValidated[E](implicit E: Semigroup[E]): SelectiveError[Validated[E, *], E] =
+    new ValidatedSelective[E] with SelectiveError[Validated[E, *], E] {
       def handleErrorWith[A](fa: Validated[E, A])(f: E => Validated[E, A]): Validated[E, A] =
         fa match {
           case Validated.Invalid(e)   => f(e)
--- a/core/src/main/scala/cats/syntax/either.scala
+++ b/core/src/main/scala/cats/syntax/either.scala
@@ -395,6 +395,11 @@ final class EitherObjectOps(private val either: Either.type) extends AnyVal { //
    * Cached value of `Right(())` to avoid allocations for a common case.
    */
   def unit[A]: Either[A, Unit] = EitherUtil.unit
+
+  /**
+   * Cached value of `Left(())` to avoid allocations for a common case.
+   */
+  def leftUnit[B]: Either[Unit, B] = EitherUtil.leftUnit
 }
 
 final class LeftOps[A, B](private val left: Left[A, B]) extends AnyVal {
@@ -509,4 +514,5 @@ private[cats] object EitherUtil {
     left.asInstanceOf[Either[A, C]]
 
   private[cats] val unit = Right(())
+  private[cats] val leftUnit = Left(())
 }
--- a/core/src/main/scala/cats/FlatMap.scala
+++ b/core/src/main/scala/cats/FlatMap.scala
@@ -20,7 +20,7 @@ import scala.annotation.implicitNotFound
  * Must obey the laws defined in cats.laws.FlatMapLaws.
  */
 @implicitNotFound("Could not find an instance of FlatMap for ${F}")
-@typeclass trait FlatMap[F[_]] extends Apply[F] {
+@typeclass trait FlatMap[F[_]] extends Select[F] {
   def flatMap[A, B](fa: F[A])(f: A => F[B]): F[B]
 
   /**
--- a/core/src/main/scala/cats/MonadError.scala
+++ b/core/src/main/scala/cats/MonadError.scala
@@ -5,13 +5,7 @@ package cats
  *
  * This type class allows one to abstract over error-handling monads.
  */
-trait MonadError[F[_], E] extends ApplicativeError[F, E] with Monad[F] {
-
-  /**
-   * Turns a successful value into an error if it does not satisfy a given predicate.
-   */
-  def ensure[A](fa: F[A])(error: => E)(predicate: A => Boolean): F[A] =
-    flatMap(fa)(a => if (predicate(a)) pure(a) else raiseError(error))
+trait MonadError[F[_], E] extends SelectiveError[F, E] with Monad[F] {
 
   /**
    * Turns a successful value into an error specified by the `error` function if it does not satisfy a given predicate.
--- a/core/src/main/scala/cats/data/EitherT.scala
+++ b/core/src/main/scala/cats/data/EitherT.scala
@@ -922,7 +922,7 @@ abstract private[data] class EitherTInstances extends EitherTInstances1 {
       implicit val monadEither: Monad[Either[E, *]] = cats.instances.either.catsStdInstancesForEither
 
       def applicative: Applicative[Nested[P.F, Validated[E, *], *]] =
-        cats.data.Nested.catsDataApplicativeForNested(P.applicative, Validated.catsDataApplicativeErrorForValidated)
+        cats.data.Nested.catsDataApplicativeForNested(P.applicative, Validated.catsDataSelectiveErrorForValidated)
 
       def monad: Monad[EitherT[M, E, *]] = cats.data.EitherT.catsDataMonadErrorForEitherT
 
@@ -983,7 +983,7 @@ abstract private[data] class EitherTInstances1 extends EitherTInstances2 {
     new Parallel[EitherT[M, E, *]] {
       type F[x] = Nested[M, Validated[E, *], x]
 
-      implicit val appValidated: Applicative[Validated[E, *]] = Validated.catsDataApplicativeErrorForValidated
+      implicit val appValidated: Applicative[Validated[E, *]] = Validated.catsDataSelectiveErrorForValidated
       implicit val monadEither: Monad[Either[E, *]] = cats.instances.either.catsStdInstancesForEither
 
       def applicative: Applicative[Nested[M, Validated[E, *], *]] =
--- a/core/src/main/scala/cats/instances/either.scala
+++ b/core/src/main/scala/cats/instances/either.scala
@@ -207,7 +207,7 @@ trait EitherInstances extends cats.kernel.instances.EitherInstances {
     new Parallel[Either[E, *]] {
       type F[x] = Validated[E, x]
 
-      def applicative: Applicative[Validated[E, *]] = Validated.catsDataApplicativeErrorForValidated
+      def applicative: Applicative[Validated[E, *]] = Validated.catsDataSelectiveErrorForValidated
       def monad: Monad[Either[E, *]] = cats.instances.either.catsStdInstancesForEither
 
       def sequential: Validated[E, *] ~> Either[E, *] =


core/src/main/scala/cats/Monad.scala
```diff
@@ -13,10 +13,16 @@ import scala.annotation.implicitNotFound
  * Must obey the laws defined in cats.laws.MonadLaws.
  */
 @implicitNotFound("Could not find an instance of Monad for ${F}")
-@typeclass trait Monad[F[_]] extends FlatMap[F] with Applicative[F] {
+@typeclass trait Monad[F[_]] extends FlatMap[F] with Selective[F] {
   override def map[A, B](fa: F[A])(f: A => B): F[B] =
     flatMap(fa)(a => pure(f(a)))
 
+  override def select[A, B](fab: F[Either[A, B]])(ff: F[A => B]): F[B] =
```
rossabaker, line 20: An implementation equivalent to this is the law in Haskell for monads.  Left open to overrides for efficiency.

core/src/main/scala/cats/Selective.scala
```diff
@@ -0,0 +1,7 @@
+package cats
+
+import simulacrum.{typeclass}
+import scala.annotation.implicitNotFound
+
+@implicitNotFound("Could not find an instance of Selective for ${F}")
+@typeclass trait Selective[F[_]] extends Applicative[F] with Select[F]
```
rossabaker, line 7: I haven't implemented any derived methods yet.  Some will go here, some can go in `Select`.

core/src/main/scala/cats/SelectiveError.scala
```diff
@@ -0,0 +1,15 @@
+package cats
+
+import cats.syntax.either._
+
+trait SelectiveError[F[_], E] extends ApplicativeError[F, E] with Selective[F] {
```
rossabaker, line 5: I haven't seen this typeclass in Haskell, but, uh, seems legit?  More MonadError methods could be pushed down.

core/src/main/scala/cats/SelectiveError.scala
```diff
@@ -0,0 +1,15 @@
+package cats
+
+import cats.syntax.either._
```
rossabaker, line 3: This is for a handle on our cached `Left(())`.  I think syntax in the typeclasses is frowned on.

rossabaker (commented):
* Needs laws
* Needs tests
* Needs consideration of more efficient implementations for the monads
* Needs to be hammered in the community build
* Needs a review from someone with scars from our inheritance encoding

... but it passes MiMa so far!

rossabaker:
I thought this would get stopped cold by binary compatibility and only expected to spend a few minutes on this. Alas, it seems to work. I'll pause here and see if anyone thinks this approach is fruitful.

/cc @cb372 @hamishdickson whose prior art can be spliced in here if we proceed.

/cc @LukaJCB @johnynek who both showed interest in the original.

core/src/main/scala/cats/Monad.scala
rossabaker: `FlatMap` should extend `Select`, but once it does, we need to push this down.  And once we do, we no longer have `pure` and lose our default implementation.

I think we can salvage `Selective`, but not `Select`.

rossabaker: Well, we still can do it from `apply`.  Maybe it lives.

core/src/main/scala/cats/Select.scala
```diff
@@ -0,0 +1,13 @@
+package cats
+
+import simulacrum.{typeclass}
+import scala.annotation.implicitNotFound
+
+@implicitNotFound("Could not find an instance of Select for ${F}")
+@typeclass trait Select[F[_]] extends Apply[F] {
+  def select[A, B](fab: F[Either[A, B]])(ff: F[A => B]): F[B] =
+    map2(fab, ff) {
```
johnynek, line 9: I don't like this default since it is no more powerful than Apply.

Since FlatMap is so rarely used as a typeclass I would suggest removing Select and just make Monad extend Selective.

core/src/main/scala/cats/Selective.scala
johnynek: I'm -1 on Select.

core/src/main/scala/cats/SelectiveError.scala
```diff
@@ -0,0 +1,15 @@
...
+
+trait SelectiveError[F[_], E] extends ApplicativeError[F, E] with Selective[F] {
+
+  /**
+   * Turns a successful value into an error if it does not satisfy a given predicate.
+   */
+  def ensure[A](fa: F[A])(error: => E)(predicate: A => Boolean): F[A] =
```
johnynek, line 10: Maybe comment why error is lazy but we always evaluate it by calling raiseError

core/src/main/scala/cats/data/Validated.scala
```diff
@@ -1035,6 +1042,15 @@ sealed abstract private[data] class ValidatedInstances2 {
   // scalastyle:off method.length
 }
 
+private[data] class ValidatedSelective[E: Semigroup] extends ValidatedApplicative[E] with Selective[Validated[E, *]] {
+  override def select[A, B](fab: Validated[E, Either[A, B]])(ff: Validated[E, A => B]): Validated[E, B] =
```
johnynek, line 1046: Are there any laws that relate select to ap? I worry this won't pass any such law, but maybe there aren't.

johnynek (commented):
I added my comments.

I'd love to see this move forward but I would suggest we not add Select at this point. If we are adding a select method that can be implemented with Apply I don't see any reason not to add it to Apply.

core/src/main/scala/cats/Selective.scala
rossabaker: I agree.  Some more methods derived from `select` may end up on `Apply` then.  We'll see.

core/src/main/scala/cats/SelectiveError.scala
rossabaker: I overlooked the laziness.  And that laziness is useful if `error` is `Throwable`, so we don't pay for stack traces on the happy path.  Maybe this is just a new function without laziness?

I don't yet know whether `SelectiveError` is a useful abstraction.  This was a "because we can."

core/src/main/scala/cats/data/Validated.scala
rossabaker: In the Haskell implementation, if there is a monad, then `select == selectM`:

```haskell
selectM :: Monad f => f (Either a b) -> f (a -> b) -> f b
selectM mx mf = do
    x <- mx
    case x of
        Left  a -> fmap ($a) mf
        Right b -> pure b
```

I haven't done the algebra, but they say that then implies `apS == <$>`:

```haskell
apS :: Selective f => f (a -> b) -> f a -> f b
apS f x = select (Left <$> f) ((&) <$> x)
```

I don't see any such laws for non-monads.

We are of course free to explore our own.  I'm just using the Haskell implementation as a starting point to see if this is even feasible in Cats 2.

johnynek: I'm +1 for adding those to MonadLaws.

johnynek:
Here is an idea we can experiment with here: for every function on the typeclass we add a law that it behaves the same as the default. This allows us to override fearlessly and know that if the laws pass it must be correct.

core/src/main/scala/cats/Selective.scala
rossabaker: Started pulling on the string to prove this was so. `ifS` can be expressed with `apply`.  `whenS` needs `Applicative` so I put it on `Selective`, but why not `Applicative`?  And then what's left of `Selective`?

Maybe the `Selective` typeclass is really `RigidSelective`, which adds the `apS == <$>` law.  All monads are rigidly selective, but maybe not all rigid selectives are monads.

johnynek: I am +1 on this. I maybe wouldn't put select on Apply in that case. I think weak law typeclasses aren't that interesting. I also don't like it when you have something that may or may not be true.

If non rigid selective can be implemented with Apply, in my view, it isn't a new typeclass.

rossabaker: Relevant discussion from [the original paper](https://dl.acm.org/ft_gateway.cfm?id=3341694).

> Such loose requirements on select with respect to unnecessary effects might seem troublesome.
Below we list three reasons why we chose to keep the requirements loose.
• Requiring unnecessary effects to be skipped rules out any instances that could be used for
static analysis. Indeed, the only way to obey this law would be to look at actual values at
runtime, since it is impossible to statically know if f (Either a b) contains a Right value.
• One might suggest having two classes Selective and StrictSelective, the latter with
stricter requirements. But this second class would be useless: to statically analyse a computation you would have to express it via Selective, since it is inhabited by instances like Over.
As for the execution, a Monad is perfectly suitable, as we will see in ğ5.2.
• Finally, there is a good precedent: the Applicative type class has no requirements on the
order in which effects are executed: left-to-right, right-to-left, or in parallel. This loose
specification allows some instances to execute effects sequentially (typically from left to
right), and other instances to execute effects in parallel. Note that as soon as we also have a
Monad, we gain an additional requirement (<*>) = ap, which tells us that, at the semantic
level, the result should be as if the effects were executed in sequence from left to right. We
follow the same approach by requiring select = selectM if f is also a Monad (see below

* I think `select` does belong somewhere below `RigidSelective`, for the static analysis reasons.
* `RigidSelective` could still be a fine constraint if there are rigid selectives that aren't also monads, but it may be more obscure than `FlatMap`.
* If all `Apply` are `Selective`, the only value in a `Selective` constraint is providing extended operations that an `Apply` promises not to do.   `F: Applicative` means "I know my conditionals will evaluate both branches." and `F: Selective` says "eh, maybe not."  Is that parametricity worth it?

johnynek: I am thinking of parsing, or IO use cases where "maybe not" would totally break things.

I don't quite see why static analysis is impossible if some effects are skipped, you still know it is one of two cases. Yes it removes branches, but only by saying both run, how is that different that knowing that there are some nodes that can be labeled to possibly be unneeded.

I'm confused why Selective exists at all if all Apply are lawful Selectives. That seems very strange to me. I can't think of another typeclass for which that's true. It is true some things aren't totally locked down by laws (as suggested), but I don't recall a case where a subclass is enough to implement the superclass. That just seems wrong to me.

So, I guess it is okay to put select on Apply, and derived functions, but I would want the law on RigidSelective (which I would consider just calling Selective in cats) that in select you definitely don't evaluate the right arg when the Either is Right.

rossabaker: > I don't quite see why static analysis is impossible if some effects are skipped,

I don't think this is the claim.  I think the claim is that different static analyses can be performed by skipping or not.  `Over` records all possible effects, and `Under` records all guaranteed effects.  Establishing these bounds is possible because of this vagary in the law.

> I'm confused why Selective exists at all if all Apply are lawful Selectives. 

I'm still forming my thoughts on this.  The authors hang their hats on this:

> Some applicative functors are not monads, e.g. the Const
functor (ğ2.2), but every applicative functor is also a selective functor, as witnessed by the function
selectA. The subclass relationship Applicative f => Selective f is justified only by the extra
method select in Selective. While select = selectA is a valid implementation of select, it is
not the only useful implementation,

I can't think of another typeclass in Cats that can be lawfully derived from a subclass, but is there one where we would override a default implementation to change behavior, rather than optimize?

rossabaker: Taking a straight path up the hierarchy, almost every subclass introduces exactly one abstract method.  `Monad` and `Async` are two exceptions, but in each case, the relationship between the two abstract methods is legislated.  `Concurrent` does not have a snappy law relating its two abstract operations, but has laws that utilize both.

* `Invariant` -> `imap`
* `Functor` -> `map`
* `Apply` -> `ap`
* `Applicative` -> `pure`
* `Monad` -> `flatMap`, `tailRecM`
* `Bracket` -> `bracketCase`
* `Sync` -> `suspend`
* `Async` -> `async`, `asyncF`
* `Concurrent` -> `racePair`, `start`
* `ConcurrentEffect` -> `runCancelable`

If we put `select` on `Apply` as an abstract method, we'd violate the principle of all abstract methods on a typeclass being legally related.  If we put it with a default implementation equivalent to `selectA`, we'd be encouraging overrides that change behavior, which goes against your principle of laws checking consistency with default behavior.  We get around both of these with a `Selective` typeclass.  

I'm back to leaning toward this being a real typeclass.

rossabaker:
The Haskell implementation introduces an operator for `select`: `<*?`.  Should we?

laws/src/main/scala/cats/laws/discipline/MonadTests.scala
```diff
@@ -25,11 +26,18 @@ trait MonadTests[F[_]] extends ApplicativeTests[F] with FlatMapTests[F] {
...
     EqFInt: Eq[F[Int]],
     iso: Isomorphisms[F]
-  ): RuleSet =
+  ): RuleSet = {
+    implicit def ArbFAA: Arbitrary[F[Either[A, A]]] =
+      Arbitrary(
+        Gen.oneOf(ArbFA.arbitrary.map(fa => laws.F.map(fa)(_.asLeft[A])),
+                  ArbFA.arbitrary.map(fa => laws.F.map(fa)(_.asRight[A]))
+        )
+      )
```
rossabaker, line 35: We need `F[Either[A, A]]`, `F[Either[A, B]]`, `F[Boolean]`, `F[Unit]`, and maybe more to test all the selective laws, but for bincompat, we have what we have here.

Since we're a functor, we can make all those from what we have.  Is that the right approach?  And should we make them here and parameterize them on SelectiveTests, or should SelectiveLaws make their own from this set?

core/src/main/scala/cats/SelectiveError.scala
johnynek: I would be conservative and not do "because we can".

laws/src/main/scala/cats/laws/SelectiveLaws.scala
```diff
@@ -0,0 +1,48 @@
+package cats
+package laws
+
+import cats.syntax.all._
+
+/**
+ * Laws that must be obeyed by any `Selective`.
+ */
+trait SelectiveLaws[F[_]] extends ApplicativeLaws[F] {
```
johnynek, line 9: Why if we make RigidSelectiveLaws that has this plus rigidity and MonadLaws extend that?

laws/src/main/scala/cats/laws/discipline/MonadTests.scala
johnynek: I say we derive these from the ones passed in and not ask for more. I think we should always use a minimal set.

laws/src/main/scala/cats/laws/SelectiveLaws.scala
rossabaker: "Why if" == "what if"?  In other words, you're proposing `RigidSelectiveLaws` between `SelectiveLaws` and `ApplicativeLaws`? I'm good with that.  

Does that imply an empty `RigidSelective` typeclass as a marker?  I'm not sure where we'd use it as a constraint, but it's weird to have laws with no corresponding typeclass.

laws/src/main/scala/cats/laws/discipline/MonadTests.scala
rossabaker: I think that's a good idea.  It does make it harder to test against pure values.  For instance, if we derive `F[Boolean]` from `F[A]`, the `F[Boolean]` will be the result of a `map`.  If we request an `F[Boolean]`, we could get pure `F[Boolean]`s.  Per the functor laws, it shouldn't matter.  In more complex functors, like `IO` or `Parser`, a pure value and a mapped value may have different internal representations with different optimizations.  I think we'll leave it to the functor laws to detect violations there.

core/src/main/scala/cats/SelectiveError.scala
rossabaker: I prefer not to lean too hard on Haskell, but there's no precedent there to justify its usefulness.  I'll remove it for now.

laws/src/main/scala/cats/laws/SelectiveLaws.scala
rossabaker: We'd need to express the rigidity law without `FlatMap`, or else there isn't much point.

johnynek: I think the law for parsers you want is `select(pa.map(Right))(pb) == pa`, so you don't want any action on the second parser if the first one succeeds and returns Right.

To test this with for instance IO, if you have some Ref, and the second item increments, if the first IO returns right, the increment should not happen.

In any case, I think the laws I want are:
```scala
select(pa.map(Right))(pb) == pa
select(pa.map(Left))(pb) == (pa, pb).mapN { (b, fn) => fn(b) }
```

Is that overdetermined?

rossabaker: The Haskell implementation proposes two similar properties, true whenever `apS == <$>`:

```haskell
select (pure (Left  x)) y = ($x) <$> y -- Pure-Left
select (pure (Right x)) y = pure x     -- Pure-Right
```

But as long as `Eq[IO[_]]` is based on `unsafeRunSync()`, then I don't see how that law precludes evaluating `pb`.  I think we have to go to SyncLaws to generically test that the effect is not being evaluated.

The laws for rigid selectives in the paper are:

```haskell
-- Laws for rigid selective functors (in particular, for monads)
(<*>) = apS -- Selective apply
x *> (y <*? z) = (x *> y) <*? z -- Interchange
```

rossabaker: I suppose we could also use tricks like the defer laws to ensure the optional effect isn't evaluated, and not go all the way to `Sync`.

johnynek: the law I suggest implies your law, and it can detect evaluating pb. Is it too strong of a law? I think it may be exactly the rigidity law stated a different way.

rossabaker: Let's make these three laws "rigid".  One from the paper, and two from you:

```scala
  def selectiveApply[A, B](ff: F[A => B], fa: F[A]): IsEq[F[B]] =
    ff.apS(fa) <-> ff.ap(fa)

  def oscarsRight[A, B](fb: F[B], ff: F[A => B]): IsEq[F[B]] =
    fb.map(Right(_)).select(ff) <-> fb

  def oscarsLeft[A, B](fa: F[A], ff: F[A => B]): IsEq[F[B]] =
    fa.map(Left(_)).select(ff) <-> (fa, ff).mapN { (a, fn) => fn(a) }
```

Override `Select[Eval]`:

```scala
      override def select[A, B](fa: Eval[Either[A, B]])(ff: Eval[A => B]): Eval[B] =
        fa.flatMap {
          case Left(a) => ff.map(_(a))
          case Right(b) => ff.map(_ => b)
        }
```

It's lawful:

```
  + Eval[Int]: rigidSelective.rigid selective Oscar's left 0.001s
  + Eval[Int]: rigidSelective.rigid selective Oscar's right 0.0s
  + Eval[Int]: rigidSelective.rigid selective apply 0.001s
  + Eval[Int]: rigidSelective.selective associativity 0.001s
  + Eval[Int]: rigidSelective.selective branch consistency 0.001s
  + Eval[Int]: rigidSelective.selective distributivity 0.001s
  + Eval[Int]: rigidSelective.selective identity 0.001s
  + Eval[Int]: rigidSelective.selective ifS consistency 0.001s
  + Eval[Int]: rigidSelective.selective whenS consistency 0.001s
  + Eval[Int]: rigidSelective.semigroupal associativity 0.001s
```

But it evaluates `pb` even on Right:

```scala
scala> val pa = Eval.now("Hi")
val pa: cats.Eval[String] = Now(Hi)

scala> val pb = Eval.later(println("ope!")) *> Eval.now((i: Int) => i.toString)
val pb: cats.Eval[Int => String] = cats.Eval$$anon$4@370b3acf

scala> pa.map(Right(_)).select(pb).value === pa
ope!
val res14: Boolean = true
```

We can check anything in the purview of the the `Eq` instance, but I don't think we can legislate the evaluation of the second effect without either a `Defer` constraint (`Sync` unites `Defer` with `Select`), or by making the second argument lazy.

johnynek: yeah, it passes the law but it evaluates the right, but you had to make a side-effecting Eval to do that. I say you broke the rules at that point and you got what you asked for.

With IO, which allows you to put side effects in the IO, a proper equality includes all the state of the world that might change after the IO is run.

We could actually do that with Eval, and detect the bug here: make an Eval that reads from a mutable variable using Eval.always. With that, your law won't pass I believe.

btw: isn.t apS == ap just oscarsLeft in the above?

core/src/main/scala/cats/data/Validated.scala
```diff
@@ -1035,6 +1044,15 @@ sealed abstract private[data] class ValidatedInstances2 {
...
```
johnynek, line 1048: I'm a bit surprised this is lawful since if there is a failure in fab and ff you only see the fab failure.

Would it still pass the laws if you accumulate any errors from ff in the fab fail case?

Then again, if select is trying to be minimal here, I guess we don't want to do that since we want to avoid ff if at all possible.

core/src/main/scala/cats/syntax/either.scala
```diff
@@ -509,4 +514,5 @@ private[cats] object EitherUtil {
     left.asInstanceOf[Either[A, C]]
 
   private[cats] val unit = Right(())
+  private[cats] val leftUnit = Left(())
```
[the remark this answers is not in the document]
johnynek: aren't these both useful in branch? I made a comment earlier in this review about it.

laws/src/main/scala/cats/laws/SelectiveLaws.scala
```diff
@@ -0,0 +1,53 @@
...
+  implicit override def F: Selective[F]
+
+  def selectiveIdentity[A, B](faa: F[Either[A, A]]): IsEq[F[A]] =
+    faa.select[A, A](F.pure(identity)) <-> faa.map(_.merge)
```
johnynek, line 13: this almost has to be true due to parametricity unless you just left one side or the other into an error, which I guess you could do for some Fs

laws/src/main/scala/cats/laws/SelectiveLaws.scala
johnynek: If we don't have the Right law, I don't know how you really program against this typeclass: I may or may not evaluate both arguments even if I don't need to (judged by the types)... okay, I guess...

To me, without that law we are back to allowing Applicative as a valid Selective and I'm pretty firmly in the no-camp on superclasses being valid implementations of their subclasses. In fact: I would propose that as a principle in general: we should never admit a typeclass whose laws are so weak that superclasses can satifsy them.

tests/src/test/scala/cats/tests/ValidatedSuite.scala
```diff
@@ -45,6 +46,9 @@ class ValidatedSuite extends CatsSuite {
            SerializableTests.serializable(ApplicativeError[Validated[String, *], String])
   )
 
+  checkAll("Validated[String, Int]", SelectiveTests[Validated[String, *]].selective[Int, Int, Int])
+  checkAll("Selective[Validated[String, *]]", SerializableTests.serializable(Selective[Validated[String, *]]))
```
johnynek, line 50: is this the only pure selective in cats? Are there any other Applicatives that could possibly be raised up?

What about the transformers (Kleisli, StateT, WriterT), some of those can have applicatives when their inner type supports applicative, can we also do selective, I bet...

laws/src/main/scala/cats/laws/SelectiveLaws.scala
rossabaker: > With IO, which allows you to put side effects in the IO, a proper equality includes all the state of the world that might change after the IO is run.

Yes, `IO` makes a better example than `Eval`.  Like `Eq[Eval[A]]`, we don't have an `Eq[IO[A]]` that can detect the unwanted evaluation of `ff`.  (It should, but it's hard.)  Unlike `Eval`, side effects are welcome in `IO`.  I see two paths forward:

1. The law is that `ff` mustn't be evaluated for any effect type.  In that case, we should make `ff` by name.

2. The law is that `ff` mustn't perform unnecessary side effects.  In that case, we add a `SyncLaw`, which would cover `IO` and similar, and declare that my `Eval` example voids the warranty.

Since we're concerned about this property in `Parser`, which has no `Sync`, I'm leaning toward the former.

> btw: isn.t apS == ap just oscarsLeft in the above?

I haven't done the algebra yet, but the very similar `Pure-Left` and `Pure-Right` are proven in Coq. (I don't read Coq).

> If we don't have the Right law, I don't know how you really program against this typeclass: I may or may not evaluate both arguments even if I don't need to (judged by the types)... okay, I guess...

This flexibility is crucial to `Over` and `Under`, but that could be done with an `Applicative` constraint if `select` were pushed down.  

We might rearrange the laws like this:
* Selective associativity in `Apply`... I think.
* Distributivity and identity in `Applicative`.
* No laws, and no typeclass, for `Selective`.
* Rigidity by a new `RigidSelective` type class.  Few of these won't also be monads.
* selectM consistency as a Monad law.  I'm not sure it's interesting as long as all monads are rigid.
* Non-evaluation of second argument in RigidSelective if it's lazy, or Sync if it's not.

core/src/main/scala/cats/syntax/either.scala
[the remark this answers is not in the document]
rossabaker: They are, but I wasn't wild about referencing syntax from the type classes.  I'd like to find a home for this.

core/src/main/scala/cats/Selective.scala
```diff
@@ -0,0 +1,80 @@
...
+
+import simulacrum.{noop, typeclass}
+import scala.annotation.implicitNotFound
+
+@implicitNotFound("Could not find an instance of Selective for ${F}")
+@typeclass trait Selective[F[_]] extends Applicative[F] {
+  def select[A, B](fab: F[Either[A, B]])(ff: F[A => B]): F[B]
+
+  def branch[A, B, C](fab: F[Either[A, B]])(fl: F[A => C])(fr: F[B => C]): F[C] = {
+    val lhs = {
```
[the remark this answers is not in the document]
rossabaker: I'm good with that.  I bootstrapped this from cats-selective, which will need proper credit when I fill in the docs as we apply the final polish.

core/src/main/scala/cats/syntax/either.scala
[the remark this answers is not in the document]
rossabaker: I find no references to syntax in the root package.  Let's keep it and find a better home.

core/src/main/scala/cats/Selective.scala
```diff
@@ -0,0 +1,80 @@
...
+      val innerLhs: F[Either[A, Either[B, C]]] = map(fab)(_.map(Left(_)))
+      val innerRhs: F[A => Either[B, C]] = map(fl)(_.andThen(Right(_)))
+      select(innerLhs)(innerRhs)
+    }
+    select(lhs)(fr)
+  }
+
+  @noop
+  def ifS[A](x: F[Boolean])(t: F[A])(e: F[A]): F[A] = {
```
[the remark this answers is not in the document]
rossabaker: I think that's only a guarantee for rigid selectives.  In static analysis, you might gather all the possible effects.

core/src/main/scala/cats/data/Validated.scala
rossabaker: Yes, I think this could be a `RigidSelective`.  A rare one that isn't also a `Monad`.

tests/src/test/scala/cats/tests/ValidatedSuite.scala
rossabaker: `Validated` is the first one I thought of that can be rigid without being a `Monad`.  `Over` is another (and `Under` would be demoted to `Applicative`, if there is no `Selective`).  Anything else that's `Applicative` without being a `Monad` is a candidate for directly being a `RigidSelective`, but I don't have a list yet.  (I wish we had a means to query for instances!)

The inductive instances on transformers all sound like good candidates, too.

laws/src/main/scala/cats/laws/SelectiveLaws.scala
johnynek: I'm fairly unconvinced that Over and Under are the main use cases here.

I don't really love the name RigidSelective personally. I would just use Selective, personally, since there is no compelling other definition for Selective (disregarding the idea of making a new typeclass with methods that can be lawfully implemented by Apply/Applicative).

If you really want to put select on Apply, I would suggest we call it selectA, and add a law in (RigidSelective) that selectA == select.

I think the non-evaluation law I want is just the Right law (pa.map(Right).select(f) == pa). For a parser, you would definitely detect if select triggered f. I'm not really concerned with the Eval example since Eval is purely to model how a computation is done, so, presumably we wouldn't make a bad implementation, so it is fine.

I hear you that IO may or may not generate complex enough values in its Arbitraries (I think the work Erik showed which we developed at Stripe and I used in cats-parse could be used to make sure IO really exercises the possible space).

Lastly, I would reiterate my point about using a call-by-name for ff just to allow to implement ifM on (RigidSelective). I kind of hate introducing a new method that is identical to something on a subclass.

Finally, I think untilDefinedM can be defined with selective:

```
def untilDefinedM[A](fa: F[Option[A]]): F[A] = {
  val init: F[Either[Unit, A]] = fa.map {
    case Some(a) => Right(a)
    case None => Left(()) // once again an Either unit value
  }
  lazy val fn: F[Unit => A] = untilDefinedM(fa).map { a => (u: Unit) => a }
 
  // if the second param is call-by-name, we only loop as many times as needed
  select(init)(fn)
}
``` 
this an be an override in Monad which can use tailRecM which might be needed for very long chains.

core/src/main/scala/cats/EitherUtil.scala
```diff
@@ -0,0 +1,14 @@
...
+ * Convenience methods and values for Either.
+ */
+private[cats] object EitherUtil {
+  def leftCast[A, B, C](right: Right[A, B]): Either[C, B] =
+    right.asInstanceOf[Either[C, B]]
+  def rightCast[A, B, C](left: Left[A, B]): Either[A, C] =
+    left.asInstanceOf[Either[A, C]]
+
+  private[cats] val unit = Right(())
+  private[cats] val leftUnit = Left(())
```
johnynek, line 13: Why do we need private cats here and on the object?

core/src/main/scala/cats/Selective.scala
```diff
@@ -0,0 +1,86 @@
...
+  def select[A, B](fab: F[Either[A, B]])(ff: F[A => B]): F[B]
+
+  def branch[A, B, C](fab: F[Either[A, B]])(fl: F[A => C])(fr: F[B => C]): F[C] = {
+    val innerLhs: F[Either[A, Either[B, C]]] = map(fab)(_.map(Left(_)))
+    val innerRhs: F[A => Either[B, C]] = map(fl)(_.andThen(Right(_)))
+    val lhs = select(innerLhs)(innerRhs)
+    select(lhs)(fr)
+  }
+
+  def apS[A, B](ff: F[A => B])(fa: F[A]): F[B] = {
```
johnynek, line 17: Why do we want this?

core/src/main/scala/cats/EitherUtil.scala
rossabaker: Inertia. We don't.

core/src/main/scala/cats/Selective.scala
rossabaker: Parroting a law.  I think we agree the `ff` should be lazy on `select`, but I don't see an opportunity for laziness in either parameter here.

It avoids `flatMap` on `FlatMap`s, which may be faster for `Parser` and more amenable to static analysis.  But if that's the rationale, where does it end? `productS`? `map2S`? `productRS`?

laws/src/main/scala/cats/laws/SelectiveLaws.scala
rossabaker: > If you really want to put select on Apply, I would suggest we call it selectA, and add a law in (RigidSelective) that selectA == select.

`selectA` seems reasonable, but I'd expect it to be strict in both arguments, like other `A` methods?  And then we still need `select` with a lazy second argument somewhere below RigidSelective if we want the `Over`/`Under` that you don't seem to want. :laughing: 

`select == selectM` is the law in Haskell once you get to Monad.  I don't think `select == selectA` and `select == selectM` are compatible as laws.

core/src/main/scala/cats/Selective.scala
rossabaker: Wow. I didn't notice similar was being discussed in #3712 today!

core/src/main/scala/cats/Selective.scala
```diff
@@ -0,0 +1,78 @@
+package cats
+
+import simulacrum.{noop, typeclass}
+import scala.annotation.implicitNotFound
+
+@implicitNotFound("Could not find an instance of Selective for ${F}")
+@typeclass trait Selective[F[_]] extends Applicative[F] {
```
johnynek, line 7: Just to reiterate, I'm strongly -1 on adding a typeclass whose superclass is a valid implementation.

I believe that is what we have here unless I'm mistaken.

I think we should remove this and instead just move all these methods up to Applicative if you are really intent on keeping select weak.

core/src/main/scala/cats/Selective.scala
```diff
@@ -0,0 +1,78 @@
...
+
+  def branch[A, B, C](fab: F[Either[A, B]])(fl: => F[A => C])(fr: => F[B => C]): F[C] = {
+    val innerLhs: F[Either[A, Either[B, C]]] = map(fab)(_.map(Left(_)))
+    def innerRhs: F[A => Either[B, C]] = map(fl)(_.andThen(Right(_)))
+    val lhs = select(innerLhs)(innerRhs)
+    select(lhs)(fr)
+  }
+
+  @noop
+  def ifS[A](fCond: F[Boolean])(fTrue: => F[A])(fFalse: => F[A]): F[A] = {
```
johnynek, line 18: this is the exact same signature as ifM. Why can't we use the same name?

This is going to be really confusing. I would rather just name this method ifM.

core/src/main/scala/cats/data/Validated.scala
```diff
@@ -1035,6 +1044,15 @@ sealed abstract private[data] class ValidatedInstances2 {
   // scalastyle:off method.length
 }
 
+private[data] class ValidatedSelective[E: Semigroup] extends ValidatedApplicative[E] with Selective[Validated[E, *]] {
+  override def select[A, B](fab: Validated[E, Either[A, B]])(ff: => Validated[E, A => B]): Validated[E, B] =
```
johnynek, line 1048: is this not rigid? Looks like to me.

laws/src/main/scala/cats/laws/SelectiveLaws.scala
```diff
@@ -0,0 +1,53 @@
+package cats
+package laws
+
+import cats.syntax.all._
+
+/**
+ * Laws that must be obeyed by any `Selective`.
```
johnynek, line 7: if none of these laws require rigidity, why not put them on Applicative while testing selectA?

laws/src/main/scala/cats/laws/ApplyLaws.scala
```diff
@@ -26,6 +26,12 @@ trait ApplyLaws[F[_]] extends FunctorLaws[F] with SemigroupalLaws[F] {
 
   def productLConsistency[A, B](fa: F[A], fb: F[B]): IsEq[F[A]] =
     F.productL(fa)(fb) <-> F.map2(fa, fb)((a, _) => a)
+
+  def selectAConsistency[A, B](fab: F[Either[A, B]], ff: F[A => B]): IsEq[F[B]] =
```
johnynek, line 30: I don't know if we need this law thinking more on it. If we remove this law, and move all the other non-rigid laws down to Apply or Applicative, I think we can allow selectA == select in a RigidSelective.

This allows users who don't care about rigidity to just pull in Apply/Applicative and use selectA, and users that require rigidity to pull in Selective (or if we must RigidSelective).

To me that is the best way to allow weak and strong versions of select.

johnynek (changes requested):
I guess I am a broken record now, so I don't need to keep commenting.

To summarize:

1. I am strongly -1 on any typeclass which can lawfully be implemented by a superclass. In this case, it appears to me that Selective and Applicative do this. Just because Andrey did this in Haskell is not a sufficient argument, IMO. If we want select on Applicative, just put it there.
2. I don't like to see us duplicate methods. Some of FlatMap/Monad methods seem to only require a RigidSelective (which I would call Selective and make all cats Selectives rigid, and if you want non-rigid, just use Applicative). I think we should move as many methods down as we can. I think ifM can be moved down, but I think others as well. Duplication confuses newcomers.

core/src/main/scala/cats/Selective.scala
rossabaker: Got rid of this by caling it `ap` and making it a law on `RigidSelective`.

core/src/main/scala/cats/Selective.scala
rossabaker: I'm relying heavily Andrey's paper for inspiration, but my hesitation is not Because Haskell™.  The problem here is two principles we both agree on are in tension:

1. No typeclasses that can be implemented by a superclass.
2. Laws that overrides equate with their base implementations.  

* The status quo violates principle #1.
* We can default `select == selectA`, but that's incompatible with rigidity, violating principle #2.
* Moving an abstract `select` down to `Apply` breaks compatibility.
* Not introducing `select` before rigidity eschews important functionality.
* `Selective` could just not participate in this hierarchy, but then we can't enforce rigidity of monads by law.

I'm leaning toward `Apply.select == selectA`, but no law.  Principle #2 above is violated, but something has to yield here.

core/src/main/scala/cats/Selective.scala
rossabaker: If this were a clean slate, we'd pull `ifM` down and call it `ifS`.  `ifM` is equivalent to `ifS` where it presently exists, but if we pull it down, some implementations will behave like `ifA`, and we'd be unexpectedly changing the behavior of an established name.

I would like to call it `ifS` and deprecate `ifM`, but I'm not sure how bad the warnings would be.

laws/src/main/scala/cats/laws/SelectiveLaws.scala
rossabaker: Will address in [this thread](https://github.com/typelevel/cats/pull/3709#discussion_r544741355).

laws/src/main/scala/cats/laws/ApplyLaws.scala
rossabaker: Because we don't want one operation that's always rigid and another one that's always not.  We need one that is maybe rigid or maybe not.  The canonical example of this is analysis that overapproximates and underapproximates effects.

I think our failure to agree on this point is the root of every other open debate on this PR.

core/src/main/scala/cats/Selective.scala
```diff
@@ -0,0 +1,80 @@
+package cats
+
+import simulacrum.{noop, typeclass}
+import scala.annotation.implicitNotFound
+
+@implicitNotFound("Could not find an instance of Selective for ${F}")
+@typeclass trait Selective[F[_]] extends Applicative[F] {
+  def select[A, B](fab: F[Either[A, B]])(ff: F[A => B]): F[B]
```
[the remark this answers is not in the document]
rossabaker: `ff` is now call-by-name.  The rest is discussed in more detail [here](https://github.com/typelevel/cats/pull/3709#discussion_r544797954).

core/src/main/scala/cats/Selective.scala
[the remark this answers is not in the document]
rossabaker: Discussed in more detail [here](https://github.com/typelevel/cats/pull/3709#discussion_r544797954).

core/src/main/scala/cats/data/Validated.scala
rossabaker: Is now rigid.

rossabaker:
Nope, that last commit is bad.  My solution was dubious, and I failed to test the rigid laws.

core/src/main/scala/cats/data/Validated.scala
rossabaker: It skips the effect appropriately, but is inconsistent with `ap`.

This is consistent with `ap`, but fails the skip laws, which I think could be relaxed to work only on pure `fab`.  More importantly, it fails associativity, and I'm entering the stupid hours.

```scala
  override def select[A, B](fab: Validated[E, Either[A, B]])(ff: => Validated[E, A => B]): Validated[E, B] =
    fab match {
      case Valid(Right(b)) => Valid(b)
      case Valid(Left(a)) => ff.map(_(a))
      case e @ Invalid(e1) =>
        ff match {
          case Valid(_) => e
          case Invalid(e2) => Invalid(EE.combine(e1, e2))
        }
    }
```

core/src/main/scala/cats/Selective.scala
johnynek: Okay, thank you for clarifying this. I think we have never violated 1, but I think we may have violated 2 without knowing since we didn't write the laws previously. For instance ap2 uses a strange order in the default implementation. I think many people won't notice if they don't use it, or use it with a commutative monad.

I tend to agree with your final result: provide a default implementation for select == selectA, but no law about it. The law I would like is that Monad.select == selectM

core/src/main/scala/cats/data/Validated.scala
rossabaker: It's the `Invalid - Valid(Right)) - Invalid` combination that's not associative in the above definition.  If we select `fb <*? fc` first, we get a valid function that discards the invalid `fc`, so the result is equivalent to the invalid `fa`.  If we select `fa <*? fb` first, the result is `fa`, and then combined with the invalid `fc` in the final step.

We can regain associativity by not combining two `Invalid`s, but then `ap != apS`, where `ap == apS` is the paper's definition of rigidity:

```scala
val left: F[Either[A => B, B]] = ff.map(Left(_))
val right: F[(A => B) => B] = fa.map((a: A) => _(a))
left.select(right)
```

core/src/main/scala/cats/SelectiveError.scala
rossabaker: This is out of scope.

core/src/main/scala/cats/Selective.scala
rossabaker: `Selective` is out, but its spirit lives on.

core/src/main/scala/cats/Selective.scala
rossabaker: `select == selectM` is already added to `MonadLaws`, so everyone agrees here.  And `ap == apS` looks like it could apply to all `FlatMap`.  I can't think of any `FlatMap`s for which that's untrue.

So then what is `RigidSelective` and not `FlatMap`?  Well, `Validated` can avoid evaluating the `ff` effect, but isn't associative when `ap == apS`.  The only two unflatmappable rigids I can think of are `Over` and a free selective, both of which are interesting, neither of which I'm convinced belong in cats-core.  And `Over` often runs with `Under`: it's the canonical example of why `select == selectA` is a fine default but a bad law.

I propose that the rigidity laws move to `FlatMap`, and no new typeclass is introduced until a need for a non-flatmapping rigidity constraint is demonstrated.

core/src/main/scala/cats/data/Validated.scala
johnynek: If you ask me, associativity is a must. The law ap == apS is one I would at least consider relaxing.

That law feels like it may preclude any rigid selects which are not also Monads.

And if you are a monad you are back to the part where we aren't really adding something new.

The struggle here feels like:

1. Add something lawful
2. Has non-trivial implementations that can't be implemented with Applicative
3. Cannot be extended to Monad.

I'm not sure if I've seen an existence proof of something matching all three.

core/src/main/scala/cats/Apply.scala
```diff
@@ -257,6 +257,30 @@ trait Apply[F[_]] extends Functor[F] with InvariantSemigroupal[F] with ApplyArit
     def ite(b: Boolean)(ifTrue: A, ifFalse: A) = if (b) ifTrue else ifFalse
     ap2(map(fcond)(ite))(ifTrue, ifFalse)
   }
+
+  def select[A, B](fab: F[Either[A, B]])(ff: => F[A => B]): F[B] =
+    selectA(fab)(ff)
+
+  def selectA[A, B](fab: F[Either[A, B]])(ff: F[A => B]): F[B] =
```
johnynek, line 264: Remind me again why we need both select and selectA if setting them equal is lawful? Why not just add select which may or may not be override to be rigid?

rossabaker: * `selectA` is like `ifA`: requires 'Apply`, not rigid even for `FlatMap`s
* `selectM` (not added yet) is like `ifM`: requires `FlatMap`, rigid
* `select` is like `ifS`: requires 'Apply`, rigid for `FlatMap`s, otherwise effect's choice

I can live without `selectA` and `selectM`, but their existence parallels some other functions.  (I'll note that @SystemFw just dumped on `ifA` on Gitter today.)

johnynek: I suspect ifA without rigidity is a recipe for exponential blow ups in cost.

core/src/main/scala/cats/data/Validated.scala
rossabaker: Several are close, but none hit the mark:

* `Const` (aka `Over`) has no `Monad` and is rigid per the paper's definition: it passes the selective laws and `ap == apS`.  It fails the our rigid skip laws.

 * There's [FreeRigidSelective](https://github.com/cb372/cats-selective/blob/master/core/src/main/scala/cats/free/FreeRigidSelective.scala).  It's `ap == apS`, but I think would also fail our rigid skip laws.  

* `Validated` and `Under` follow our skip laws, but not `ap == apS`.

*  All `FlatMap`s appear to pass `ap == apS` and selective associativity, but at least `Map[K, *]` and `Tuple[X, *]` fail our skip laws.

We may have conflated two concepts with what the paper calls rigid and our rigid skip laws, but everything that satisfies both seems to have a `Monad`.

core/src/main/scala/cats/Apply.scala
SystemFw: >  (I'll note that @SystemFw just dumped on ifA on Gitter today.

I have barely skimmed the whole conversation, so I do apologise for the lack of context, but I do think `ifA` is one of the worse things added in cats.
An `if` that executes both branches is what you want 0.01% of the time, and users have been told to "use the least powerful constraint you can", which makes it a recipe for disaster. We can't even claim we didn't see this, because the `if` + macro interaction in sbt has been hated for the same reason for years.

Personally I see Selective as just pulling things down from Monad, to get you more static analysis, so in an ideal world `ifA` would be gone (and similarly for `select*` on Apply), `Selective` would called just that, and have `ifM` (`ifS`?), etc. So (barring misunderstanding of what I've skimmed), I think I'm largely in agreement with @johnynek

johnynek: So, I caused a bit of the wild goose chase here I think.

One thing I was and am concerned about is a lawful implementation of Selective using Applicative, why not add to Applicative? So, I pushed to lower things to Applicative or Apply. But that means the default implementation winds up executing both sides, but as you note, you would almost never want that. It seems weird to open the door for exactly one use case that we can name (over approximating dependencies in a modestly dynamic graph).

So, now my thinking has gotten to either:

1. Reintroduce Selective (basically back where we started, yes you could implement Selective with Applicative, but it is probably not what you want, so it isn't a default, this allows you to implement it when you are sure that is what you want).
2. I'm dubious RigidSelective is much of a thing. I tend to wonder if the only instances are things that could be monads but are being restricted in some way (e.g. CLI parsers being restricted so that they can generate a static help).
3. A conservative way out would be to add `def select` to Monad and implement it with flatMap, add the select laws there (and require it to be rigid). This allows Monad instances to optimize select if they can (like Parser or Gen could).

So, one idea would be:

```scala
// Comment that this is a rarely used type that is generally used to make static analysis of
// F[_] values that model a deferred computation.
trait Selective[F[_]] extends Applicative[F] {
  def select[A, B](e: F[Either[A, B]])(fn: F[A => B]): F[B]
}

trait RigidSelective[F[_]] extends Selective[F] {
  def ifM... // all the monad methods that can be implemented in terms of select
}

trait Monad[F[_]] extends RigidSelective[F] {
...
}
```

rossabaker: I hadn't seen Oscar's reply while doing this research, but dumping it while the kids are entertained.

The `select == selectM` relationship for monads has never been controversial.  And I think `ap == apS` follows for flatMaps.  Every `Apply` can be selective (associativity) with more laws (identity, distributivity) once `Applicative`.

So we should focus on instances that are `Apply` but not `FlatMap`.  There are 55 such implementations in cats and cats-effect.  Several are inductive instances.

What's left are worth thinking about, because they each:
1. May have an interesting `select != selectA`.
2. May have an offensive `select` that justifies a `Selective` typeclass.

## The data types

### Parallel versions of otherwise monadic types

We already identified `Validated` as an interesting `Selective`.  I bet all of these are interesting, too:

* cats.data.NonEmptyList.ZipNonEmptyList
* cats.data.NonEmptySeq.ZipNonEmptySeq
* cats.data.NonEmptyVector.ZipNonEmptyVector
* cats.data.Validated
* cats.data.ZipLazyList
* cats.data.ZipList
* cats.data.ZipSeq
* cats.data.ZipStream
* cats.data.ZipVector
* cats.effect.IO.Par

### Static analysis

* cats.data.Const: has two interesting selectives, and the only known interesting one where `select == selectA`.

### Data types we need to think harder about

* cats.data.AppFunc
* cats.data.Func
* cats.data.Nested 
* cats.effect.Fiber

## Type classes to think about

* cats.arrow.Arrow: we (non-implicitly) derive Applicative from Arrow. What does this imply about its default `Selective`?  Can we derive a more interesting one?

## Composed instances

Could there be a `ComposedSelective` to match? This is closely related to `Nested`.

* cats.ComposedApply
* cats.ComposedApplicative
* cats.ComposedAlternative

johnynek: Great summary Ross.

To simplify, I think ZipList, NonEmptyList.ZipNonEmptyList, Validated, IO.Par and Nested are the core things to think about. Clearly if Zip*List works, the other collections which are isomorphic should work too.

rossabaker: I need to cook now, but quick response to "So, one idea would be:..."

* Yes, let's set aside the difference between `Apply` and `Applicative` for now.  We can circle back on maps and tuples in a second PR, or not, after we're done learning from this.
* Agree with reviving `Selective[F]`.  Several of the laws (identity, associativity, distributivity) can stick there.
* Agree on monad having rigidity laws.  (That has remained constant through this PR.)

I have some qualms pulling the `*M` functions down to `RigidSelective`:
1. As you note, few things are without being monads.  Which things?  Depends whether we stick to our skip laws, or accept the `ap == apS` law, or both.
2. Depending how we resolve the above, some things may have a viable `ifM` without being a `RigidSelective`, but we'd have no typeclass for these.  We could pull those functions down to `Selective`, but the name would lose its current sense of rigidity.
3. Some of the`'*M` functions are on `FlatMap`.  Now we need to think about that relationship earlier than we'd like.

Haskell gets around all these problems with an `*S` set of functions that's usually like `*M`. Great for laws, confusing for new users autocompleting weird names.

rossabaker: Oof. `ZipList` is in the pile of things that are `Apply` but not `Applicative` that I'm trying to forget about for now.

`ZipStream`, `ZipLazyList`, and `IO.Par` are all `Applicative`.

rossabaker: The Applicative[ZipLazyList] laws don't terminate, so I don't know if our `Selective[ZipLazyList]` is lawful.  But it appears to be Haskell-rigid: the `Apply` laws pass when `ap` is implemented as `apS`.  It's not Cats-Issue-3709-rigid, because we have to evaluate `ff` even if we get a lazy stream of Rights.

j-mie6:
Wanted to weigh in on this (a bit late, I know!): I'd like to point out that `branch` might be a better abstract operation to support than `select`:

* `branch[A, B, C](mx: F[Either[A, B]])(left: F[A => C], right: F[B => C]): F[C]` (shuffle the currying around to taste) is nice because it can implement `select` using one operation (and it's as easy as `branch[A, B, B](mx)(my, identity[B].pure)`. The inverse direction requires _two_ `select`s
* `branch` very obviously corresponds to a lifted fold on `Either` (or in other words a lifted pattern match), whereas `select`s high-level idea is less immediately obvious. This is a nice symmetry with `<*>` being a basic lifted function application and `flatMap` being a lifted variable-binding. 
* The selective laws apply similarly to `branch` with a couple of intuitive modifications.

Also I believe that there are some nice interactions between `select` and failure for an `Alternative`. Having both `Alternative` and `Selective` gives you access to a `filter` combinator, as well.

core/src/main/scala/cats/Apply.scala
j-mie6: > 2. I'm dubious RigidSelective is much of a thing. I tend to wonder if the only instances are things that could be monads but are being restricted in some way (e.g. CLI parsers being restricted so that they can generate a static help).

One very good reason for having a specific `Selective` thing that _isn't_ also a monad is precisely to leverage static structure with dynamic behaviours. Take parsers, for example: `Applicative` isn't strong enough to do context-sensitive things, `Monad` is, and `Selective` gives you some (of the most common) context-sensitive behaviours but not all of them. This is important, because if you wanted to generate fast hand-written recursive descent for a parser combinator implementation, you need a fully static structure. That rules out `Monad`, but not `Selective` or `Applicative`: luckily these can almost always be enough for your practical parsing needs, but such a library wouldn't be able to make a `Monad` instance (see https://github.com/j-mie6/ParsleyHaskell as an example of such a library)

rossabaker:
I'm unsure at this point whether it's easier to scrape the bitrot off this PR or start over and hope to not reopen the same issues that we did manage to work through here, but I'm interested in trying or handing the baton to someone else.

johnynek:
somehow this PR popped into my head today, and it *seemed* to me like I found a solution.

I propose the following:

1. `Selective[F] extends Applicative[F]`
2. `Monad[F] extends Selective[F]`
3. `Selective[F]` has one abstract function: `def select[A, B](fab: F[Either[A, B]], fn: F[A => B]): F[B]`
4. the laws are: `forall { (fa: F[A], fn: F[A => B]) => equiv(select(fa.map(Right(b)), fn), fa) }` and `forall { (fb: F[B], fn: F[A => B]) => equiv(select(fa.map(Left(b)), fn), ap(fb, fn)) }`

I don't see that we proposed this law before, but I think it rules out an implementation using `Applicative`. For example, consider 
```scala
select(fab, fn) = map2(fab, fn) {
  case (Right(b), _) => b
  case (Left(a), f) => f(a)
}
```
now consider `Applicative[List]`, we know that `select(Right(a) :: Nil, Nil)` *should* be `a :: Nil` if we were using the monadic definition:

```
select(fab, fn) = flatMap(fab) {
  case Right(b) => pure(b)
  case Left(a) => fn.map(_(a))
}
```
but using the definition above (map2-based) it will return `Nil`. Similarly with `ApplicativeError[E, F]` then `select(fa.map(RIght(_)), raise(e)) == fa`, but the product based implementation can't give that.

I think this is what we want.

Then we rule out `Selective` being different from `Applicative` and yet, I think all the instances we want, retain implementations.

For instance, I think `Validated` can do:

```scala
def select[A, B](fab: Validated[E, Either[A, B]], fn: Validated[E, A => B]) =
  fab match {
    case Valid(Right(b)) =>
      // do to Validated being pure and parametricity, this is the only value that can type check
      Valid(b)
    case Valid(Left(a)) =>
      // do to Validated being pure and parametricity, this is the only value that can type check
      fn.map(_(a))
    case Invalid(e) =>
      // we have to combine fn on error to satisfy the left/apply law
      fn match {
        case Valid(_) => Invalid(e)
        case Invalid(e1) => Invalid(combine(e, e1))
     }
  }
```
I *think* this is the only function that can match the above two laws.

So, unless I'm mistaken, this solves the issues I had with Selective being the same as Applicative, and we don't need to add Selective and RigidSelective: we only add Selective (which is always rigid).

Now, we could always add a function like `selectA` but I wouldn't do it, since the the Applicative implementation will conflict with `select` when you can implement both (for instance in the Validated case one returns Invalid and one Valid when you have `Valid(Right(_))` with a fn that is `Invalid(_)`. Having two functions with very similar type signatures, but different results I think invites errors. If you only have Applicative, you just have to use `map2, ap, product, mapN` etc...

johnynek:
Looking back at the original paper I don't think I actually made much progress here except to state rigid laws a bit more clearly in scala.

I guess I still disagree with the paper and think Selective should be rigid and if you want to violate rigidity just make a Selective implemented with Applicative and don't check the Selective laws (just check the Applicative laws).

j-mie6:
> Looking back at the original paper I don't think I actually made much progress here except to state rigid laws a bit more clearly in scala.
> 
> I guess I still disagree with the paper and think Selective should be rigid and if you want to violate rigidity just make a Selective implemented with Applicative and don't check the Selective laws (just check the Applicative laws).

If you want to go the rigid only direction, then perhaps the discussion here might be of interest: https://github.com/snowleopard/selective/issues/74

I think one of the reasons this might not happen is because of supporting non-rigidity?

j-mie6:
To be fair, I think I'm on board with assuming rigidity, because otherwise, you can just implement the desired `selectA` with `Applicative` and ignore selective entirely. In which case, great, I'm pretty sure you can go ahead and do that immediately without introducing a new typeclass for it. You want selectiveness when you can't quite get monadic power but want decision making, in my experience. However, the other argument is for static analysis purposes.

johnynek:
I also still don't see why a single use case stops the law which I think for many effects you want: you would almost certainly want to know that an effect isn't run unless required not that "maybe it will maybe it won't".

If you want static analysis just make a new typeclass and instantiate it from Applicative or Selective.

johnynek:
Also your argument about branch makes sense to me.

johnynek:
One last point: if you want to do static analysis, can't you do it with a Free selective? The key issue is that there are no functions like bind/flatMap to evaluate in selective: nothing has the shape `A => F[B]` so with a free selective you could do static analysis.

PR #3722, opened 2020-12-24
Rename productM to mproduct in the glossary

BenFradet:
I believe `mproduct` is the available name and not `productM`. At least I couldn't find `productM` in the documentation nor in the code.


larsrh (approved):
Good catch, thanks!

PR #3730, opened 2021-01-11
Add NonEmptyList.zip

saeltz:
This adds `NonEmptyList.zip` as it was still missing and we could use it for our project. Including tests.

I don't know why the unrelated test is failing. Is it flaky?

### DIFF
--- a/core/src/main/scala/cats/data/NonEmptyList.scala
+++ b/core/src/main/scala/cats/data/NonEmptyList.scala
@@ -287,6 +287,22 @@ final case class NonEmptyList[+A](head: A, tail: List[A]) extends NonEmptyCollec
   }
 
   /**
+   * Zips this `NonEmptyList` with another `NonEmptyList` and returns the pairs of elements.
+   *
+   * {{{
+   * scala> import cats.data.NonEmptyList
+   * scala> val as = NonEmptyList.of(1, 2, 3)
+   * scala> val bs = NonEmptyList.of("A", "B", "C")
+   * scala> as.zip(bs)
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -286,6 +286,22 @@ final case class NonEmptyList[+A](head: A, tail: List[A]) extends NonEmptyCollec
...
+   * {{{
+   * scala> import cats.data.NonEmptyList
+   * scala> val as = NonEmptyList.of(1, 2, 3)
+   * scala> val bs = NonEmptyList.of("A", "B", "C")
+   * scala> as.zip(bs)
+   * res0: cats.data.NonEmptyList[(Int, String)] = NonEmptyList((1,A), (2,B), (3,C))
+   * }}}
+   */
+  def zip[B](nel: NonEmptyList[B]): NonEmptyList[(A, B)] = {
+    val (NonEmptyList(head1, tail1), NonEmptyList(head2, tail2)) = (this, nel)
```
johnynek, line 301: This relies on the compiler to avoid a needless allocation. Can we just explicitly write the code without the pattern matching or tuple?

```scala
NonEmptyList((head, nel.head), tail.zip(nel.tail))
```

johnynek:
Sorry for the review latency.

core/src/main/scala/cats/data/NonEmptyList.scala
saeltz: Good suggestions. Applied it in https://github.com/typelevel/cats/pull/3730/commits/d5605656277dcfa14c537a328d3f51e0e3d7a357!

johnynek (approved):

johnynek:
Thanks!

PR #3732, opened 2021-01-12
scaladoc fixes for PartialOrder

martijnhoekstra:
use Markdown tables and lists, so that it renders nicely in the output HTML


LukaJCB (approved):
thanks! :)

PR #3747, opened 2021-01-27
Add callCC to ContT

TimWSpence:
Add "call with current continuation" to ContT

### DIFF
--- a/core/src/main/scala/cats/data/ContT.scala
+++ b/core/src/main/scala/cats/data/ContT.scala
@@ -125,9 +125,12 @@ object ContT {
    *
    * }}}
    */
-  def callCC[M[_], R, A, B](f: (A => ContT[M, R, B]) => ContT[M, R, A]): ContT[M, R, A] =
+  def callCC[M[_], R, A, B](f: (A => ContT[M, R, B]) => ContT[M, R, A])(implicit M: Defer[M]): ContT[M, R, A] =
     apply { cb =>
-      f(a => apply(_ => cb(a))).run(cb)
+      val cont = f { a =>
+        apply(_ => cb(a))
+      }
+      M.defer(cont.run(cb))
     }
 
   /**
--- a/core/src/main/scala/cats/data/ContT.scala
+++ b/core/src/main/scala/cats/data/ContT.scala
@@ -40,7 +40,9 @@ sealed abstract class ContT[M[_], A, +B] extends Serializable {
     // allocate/pattern match once
     val fnAndThen = AndThen(fn)
     ContT[M, A, C] { fn2 =>
-      val contRun: ContT[M, A, C] => M[A] = _.run(fn2)
+      val contRun: ContT[M, A, C] => M[A] = { c =>
+        M.defer(c.run(fn2))
+      }
       val fn3: B => M[A] = fnAndThen.andThen(contRun)
       M.defer(run(fn3))
     }
--- a/core/src/main/scala/cats/data/ContT.scala
+++ b/core/src/main/scala/cats/data/ContT.scala
@@ -105,6 +105,29 @@ object ContT {
       def apply[A](ma: M[A]): ContT[M, B, A] = ContT.liftF(ma)
     }
 
+  /*
+   * Call with current continuation
+   *
+   * Passes the current continuation to f, meaning we can model short-circuit
+   * evaluation eg exception handling
+   *
+   * {{{
+   *   for {
+   *     _ <- ContT.callCC( (k: Unit => ContT[IO, Unit, Unit]) =>
+   *       ContT.liftF(IO.println("this will print first")) >>
+   *         k(()) >>
+   *         ContT.liftF(IO.println("this will NOT print as we short-circuit to the contination"))
+   *     )
+   *     _ <- ContT.liftF(IO.println("this will print second")])
+   *   } yield ()
+   *
+   * }}}
+   */
+  def callCC[M[_], R, A, B](f: (A => ContT[M, R, B]) => ContT[M, R, A]): ContT[M, R, A] =
+    apply { cb =>
+      f(a => apply(_ => cb(a))).run(cb)
+    }
+
   /**
    * Similar to [[pure]] but evaluation of the argument is deferred.
    *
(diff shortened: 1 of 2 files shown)


LukaJCB (dismissed):

core/src/main/scala/cats/data/ContT.scala
```diff
@@ -105,6 +105,29 @@ object ContT {
...
+   *         ContT.liftF(IO.println("this will NOT print as we short-circuit to the contination"))
+   *     )
+   *     _ <- ContT.liftF(IO.println("this will print second")])
+   *   } yield ()
+   *
+   * }}}
+   */
+  def callCC[M[_], R, A, B](f: (A => ContT[M, R, B]) => ContT[M, R, A]): ContT[M, R, A] =
+    apply { cb =>
+      f(a => apply(_ => cb(a))).run(cb)
```
johnynek, line 128: I'm trying to see:

1. Why we can't move cb(a) out
2. If we should use Defer[M] anywhere to make this lazier or safer (as is done in tailRecM)

core/src/main/scala/cats/data/ContT.scala
```diff
@@ -105,6 +105,29 @@ object ContT {
...
 
+  /*
+   * Call with current continuation
+   *
+   * Passes the current continuation to f, meaning we can model short-circuit
+   * evaluation eg exception handling
+   *
+   * {{{
+   *   for {
+   *     _ <- ContT.callCC( (k: Unit => ContT[IO, Unit, Unit]) =>
```
arosien, line 116: Use brackets instead of parentheses to make it cleaner looking?

core/src/main/scala/cats/data/ContT.scala
TimWSpence: Sorry, what did you mean by move `cb(a)` out?

The `Defer` thing is something I was worried about. I am not aware that we need it anywhere but was hoping that somebody with more experience would catch it in review if I'm wrong

johnynek: for instance:
```scala
apply {
  val cont = f { a =>
    val cba = cb(a)
    apply(_ => cba)
  }

  cont.run(cb)
}
```
has potentially different execution behavior, although, if we have purity, I think the same result. It isn't clear to me which is likely to be more efficient.

Secondly, if we have a `Defer[M]` then we could put the entire `cont.run(cb)` in a defer: `Defer[M].defer(cont.run(cb))`

Since we do that on map and flatMap, I assume there are recursive cases that would want that here as well, and if nothing else, for consistency, I guess I'd do it...

But it would be nice to find a recursive example and see if we can show we can do deep recursions.

TimWSpence: Hmm it would appear that it is not stack-safe. I pushed a failing test for it but no amount of poking it with `defer` and `AndThen` has fixed it yet 😂

johnynek: this seems to fix it:
```diff
oscar@Patricks-MacBook-Air cats % git diff
diff --git a/core/src/main/scala/cats/data/ContT.scala b/core/src/main/scala/cats/data/ContT.scala
index 4bef4e13c..d4864b80b 100644
--- a/core/src/main/scala/cats/data/ContT.scala
+++ b/core/src/main/scala/cats/data/ContT.scala
@@ -40,7 +40,10 @@ sealed abstract class ContT[M[_], A, +B] extends Serializable {
     // allocate/pattern match once
     val fnAndThen = AndThen(fn)
     ContT[M, A, C] { fn2 =>
-      val contRun: ContT[M, A, C] => M[A] = _.run(fn2)
+      val contRun: ContT[M, A, C] => M[A] =  { c =>
+        M.defer(c.run(fn2))
+      }
+
       val fn3: B => M[A] = fnAndThen.andThen(contRun)
       M.defer(run(fn3))
     }

```
in the flatMap function.

Can you try that (I pulled your PR and it worked for me).

TimWSpence: 😍 THANK YOU! Very good call. Just out of curiosity: what made you realize that flatMap was the issue?

I've added an explicit stack-safety test for `flatMap` as well (which I just assumed already existed hence not even thinking to suspect `flatMap`)

johnynek: Honestly, I just looked at all the places a Defer could go and I tried it.

Afterwords, it seemed more obvious that flatMap needed one there: it was internally doing a run which starts a new stack on whatever M you have. By using a Defer, you should push that evaluation onto the same final heap used to interpret M.

I think a good rule of thumb should be: any time we do a `run` internally, we should wrap that inside `Defer[M].defer`.

core/src/main/scala/cats/data/ContT.scala
```diff
@@ -105,6 +107,29 @@ object ContT {
...
+   *         ContT.liftF(IO.println("this will NOT print as we short-circuit to the contination"))
+   *     )
+   *     _ <- ContT.liftF(IO.println("this will print second")])
+   *   } yield ()
+   *
+   * }}}
+   */
+  def callCC[M[_], R, A, B](f: (A => ContT[M, R, B]) => ContT[M, R, A]): ContT[M, R, A] =
+    apply { cb =>
+      f(a => apply(_ => cb(a))).run(cb)
```
johnynek, line 130: according to the principal above, I think we should put the `run` call in callCC inside a defer.
```scala
apply { cb =>
  val cont = f { a =>
    apply(_ => cb(a))
  }

  Defer[M].defer(cont.run(cb))
}
```

johnynek (changes requested):
request one more change and then we can merge this I think.

core/src/main/scala/cats/data/ContT.scala
TimWSpence: 👍 done. If you could re-trigger the build that would be great! I created https://github.com/typelevel/cats/issues/3762 to track/capture the scalacheck seed for the `Order[FiniteDuration]` laws failure

core/src/main/scala/cats/data/ContT.scala
TimWSpence: Thanks, just trying to visualize the execution and the interaction between `AndThen` and `Defer`. Any insights much appreciated 😂

johnynek (approved):
This is great. Glad we have the flatMap stack safety checks too. Thanks!

PR #3748, opened 2021-01-27
Delimited continuations via shiftT/resetT for ContT

TimWSpence:
Delimited continuations for `ContT` via `shiftT`/`resetT`


core/src/main/scala/cats/data/ContT.scala
```diff
@@ -45,6 +45,8 @@ sealed abstract class ContT[M[_], A, +B] extends Serializable {
       M.defer(run(fn3))
     }
   }
+
+  final def eval(implicit M: Monad[M], ev: B <:< A): M[A] = run(b => M.pure(ev(b)))
```
johnynek, line 49: Why monad here? Seems you are only using pure?

tests/src/test/scala/cats/tests/ContTSuite.scala
```diff
@@ -86,4 +86,27 @@ class ContTSuite extends CatsSuite {
...
+  test("ContT.resetT and shiftT delimit continuations") {
+    forAll { (cb: Unit => Eval[Unit]) =>
+      var counter = 0
+
+      val contT: ContT[Eval, Unit, Unit] = ContT.resetT(
+        ContT.shiftT { (k: Unit => Eval[Unit]) =>
+          ContT.defer[Eval, Unit, Unit] {
+            counter = counter + 1
+          } >>
+            ContT.liftF(k(())) >>
```
johnynek, line 98: I have no intuition for why this is testing anything when it is true for all cbs.

core/src/main/scala/cats/data/ContT.scala
```diff
@@ -160,6 +162,38 @@ object ContT {
...
+   *                     }
+   *                   )
+   *              _ <- ContT.liftF(IO.println("3"))
+   *            } yield ()
+   *         )
+   *     _ <- ContT.liftF(IO.println("5"))
+   *   } yield ()
+   * }}}
+   */
+  def shiftT[M[_]: Monad, R, A](f: (A => M[R]) => ContT[M, R, R]): ContT[M, R, A] =
```
johnynek, line 194: Same question about monad here.

core/src/main/scala/cats/data/ContT.scala
TimWSpence: Thanks. Yeah, cargo-culted from Haskell wtithout thinking too much about it. Will relax to `Applicative`

tests/src/test/scala/cats/tests/ContTSuite.scala
TimWSpence: Yes, good point 😂 I'll try to come up with a better test. Really we want to assert on the evaluation order, because that is different from other monads but it's not so obvious to do that. I guess we could have multiple `var`s and assign them a monotonic timestamp as we evaluate?

TimWSpence: Actually just realized `AtomicInteger#incrementAndGet()` would be much better

core/src/main/scala/cats/data/ContT.scala
TimWSpence: Actually, that is required - it needs `Applicative` for `eval` and `FlatMap` for `liftF`

core/src/main/scala/cats/data/ContT.scala
```diff
@@ -45,6 +45,8 @@ sealed abstract class ContT[M[_], A, +B] extends Serializable {
       M.defer(run(fn3))
     }
   }
+
+  final def eval(implicit M: Applicative[M], ev: B <:< A): M[A] = run(b => M.pure(ev(b)))
```
johnynek, line 49: I think to be safe (following #3747 ) any call to run should probably be done inside a Defer, so this should probably be
```scala
def eval(implicit M: Applicative[M], defer: Defer[M], ev: B <:< A): M[A] = defer.defer(run(b => M.pure(ev(b))))
```

johnynek: alternatively, maybe we only need Defer when we are nesting, so, maybe below inside reset we should put the Defer there, since I guess this call can't stack overflow since AndThen calls shouldn't stack overflow.

johnynek: or maybe just ignore this since we have no evidence that it matters... I think the flatMap fix may be the only one we need.

core/src/main/scala/cats/data/ContT.scala
```diff
@@ -160,6 +162,38 @@ object ContT {
...
+   *                   }
+   *              _ <- ContT.liftF(IO.println("3"))
+   *            } yield ()
+   *         )
+   *     _ <- ContT.liftF(IO.println("5"))
+   *   } yield ()
+   * }}}
+   */
+  def shiftT[M[_]: Monad, R, A](f: (A => M[R]) => ContT[M, R, R]): ContT[M, R, A] =
+    apply(cb => f(cb).eval)
```
johnynek, line 195: putting this inside a `Defer` seems more likely to be needed...

Can we add a stack-safety test here in some way?

core/src/main/scala/cats/data/ContT.scala
TimWSpence: Thanks. Yeah, I've put a `defer ` in just to be safe but I wrote stack-safety tests for both `shiftT` and `resetT` and they passed without it. Definitely good tests to have though so thank you 👍

johnynek:
do you want to merge #3747 ?

TimWSpence:
> do you want to merge #3747 ?

Yep, done 👍 Still trying to get my head around the stack-usage behaviour of that. With `defer` and `AndThen` does it run using constant stack? (I guess so if it's stack-safe for arbitrary `flatMap` chains)

core/src/main/scala/cats/data/ContT.scala
```diff
@@ -188,6 +190,38 @@ object ContT {
...
+   *                       _ <- ContT.liftF(k(()))
+   *                       _ <- ContT.liftF(IO.println("4"))
+   *                     }
+   *                   }
+   *              _ <- ContT.liftF(IO.println("3"))
+   *            } yield ()
+   *         )
+   *     _ <- ContT.liftF(IO.println("5"))
+   *   } yield ()
+   * }}}
```
johnynek, line 220: can you add what this is supposed to do? It you don't understand shift and reset (and I am foggy on it), I don't think this comment helps you.

core/src/main/scala/cats/data/ContT.scala
```diff
@@ -47,6 +47,8 @@ sealed abstract class ContT[M[_], A, +B] extends Serializable {
       M.defer(run(fn3))
     }
   }
+
+  final def eval(implicit M: Applicative[M], D: Defer[M], ev: B <:< A): M[A] = run(b => D.defer(M.pure(ev(b))))
```
johnynek, line 51: I think we need instead:
```scala
D.defer(run(b => M.pure(ev(b))))
```
as you have it now, it is not needed, since `b` is already evaluated, so `D.defer(M.pure(ev(b)))` is not helpful.

TimWSpence: Yes, good point! 👍

core/src/main/scala/cats/data/ContT.scala
TimWSpence: Yup 👍

core/src/main/scala/cats/data/ContT.scala
johnynek: I don't see why `Monad` is needed here. It seems to me that only `Applicative` should be since I think `apply` requires nothing and `eval` just requires Applicative and Defer.

core/src/main/scala/cats/data/ContT.scala
```diff
@@ -188,6 +190,58 @@ object ContT {
   def later[M[_], A, B](fn: => (B => M[A]) => M[A]): ContT[M, A, B] =
     DeferCont(() => FromFn(AndThen(fn)))
 
+  /*
+   * Limits the continuation of any inner [[shiftT]]
+   */
+  def resetT[M[_]: Monad: Defer, A, B](contT: ContT[M, A, A]): ContT[M, B, A] =
```
johnynek, line 196: can we reverse `B` and `A` here so that we return `ContT[M, A, B]` as in other cases?

johnynek (changes requested):
Two minor issues.

TimWSpence:
> Two minor issues.

👍 thanks again!

johnynek (approved):
Thanks so much for doing this. I was so obsessed with finding a way to make tailRecM not blow the stack when I implemented this that once I solved that I didn't even add the classic functions. Thanks for taking it over the finish line.

TimWSpence:
> Thanks so much for doing this. I was so obsessed with finding a way to make tailRecM not blow the stack when I implemented this that once I solved that I didn't even add the classic functions. Thanks for taking it over the finish line.

No worries!! Thanks so much for helping me make it stack-safe 😂 I still find the `Defer`/`AndThen` interaction a bit brain-bending but I trust the tests that it works

PR #3750, opened 2021-01-29
Support scala-native 0.4

arashi01:
~~Requires typelevel/discipline-munit#73 and discipline-munit to be published for scala-native first for builds to go green. (Tested locally)~~

- Enables _all_ cross modules currently published for Scala.js for [Scala Native](https://www.scala-native.org) as well.
- Updates Scala 2.12 from 2.12.12 to 2.12.13
- Updates discipline-munit to 1.0.5
- ~~Reorganises build to split coverage job out of main build matrix.~~


larsrh:
discipline-munit 1.0.5 release was successful :tada:

arashi01:
Looks like the current build errors are specifically related to scoverage/sbt-scoverage#319 / sbt/sbt#6251

One solution I can think of (other than waiting for scoverage/sbt-scoverage#319 to be resolved) is splitting the coverage tasks into their own CI steps and temporarily disabling coverage / coverageReport for Scala 2.12.13.

@larsrh Any thoughts on this?

arashi01:
Solution, if acceptable, was to move coverageReport and associated prep tasks out of the matrix build and into separate job. 

This is arguably a cleaner solution than including coverage generation in validateJVM command, as was the case previously. Coverage was only being run on JVM platform projects, and only for Scala 2.12 / 2.13. This is still the case, with 2.12 not included due to scoverage/sbt-scoverage#319 mentioned above.

arashi01:
@larsrh I think we're ready to go on this. Let me know if anything needs adjusting.

build.sbt
```diff
@@ -52,44 +52,44 @@ ThisBuild / scalaVersion := Scala213
...
+  githubWorkflowJavaVersions.value.filterNot(Set(PrimaryJava)).flatMap { java =>
+    Seq(MatrixExclude(Map("platform" -> "js", "java" -> java)),
+        MatrixExclude(Map("platform" -> "native", "java" -> java))
+    )
+  }
+
+ThisBuild / githubWorkflowBuildMatrixExclusions ++=
+  Seq("jvm", "js", "native").map { platform =>
+    MatrixExclude(
+      Map("platform" -> platform, "java" -> LatestJava, "scala" -> DottyOld)
```
larsrh, line 67: You can rebase on latest master, this is not necessary anymore.

native/src/test/scala/cats/native/tests/FutureSuite.scala
```diff
@@ -0,0 +1,65 @@
...
+import org.scalacheck.rng.Seed
+import org.scalacheck.{Arbitrary, Cogen}
+
+import scala.concurrent.{Await, ExecutionContextExecutor, Future}
+import scala.concurrent.duration._
+
+class FutureSuite extends CatsSuite {
+  val timeout = 3.seconds
+
+  // TODO: We shouldn't do this! See: https://github.com/scala-js/scala-js/issues/2102
```
larsrh, line 19: Is this still accurate for Scala Native?

arashi01: It seems so. Using global ExecutionContext resulted in intermittent test failures.

The actual test suite is copied from the JVM platform project, modified with with synchronous ExecutionContext replacing the global import.

build.sbt
arashi01: Done.

larsrh:
Sorry to rain on your parade, but we might kill coverage entirely. PR incoming, which we'll discuss first.

arashi01:
> Sorry to rain on your parade, but we might kill coverage entirely. PR incoming, which we'll discuss first.

No problem at all. Will rebase again when necessary!

larsrh:
@arashi01 FYI, we've removed coverage from the build now.

arashi01:
@larsrh Done. Let me know if any changes are needed.

build.sbt
```diff
@@ -46,34 +46,40 @@ ThisBuild / scalaVersion := Scala213
...
 ThisBuild / githubWorkflowArtifactUpload := false
 
+ThisBuild / githubWorkflowBuildMatrixFailFast := Some(false)
+
 val JvmCond = s"matrix.platform == 'jvm'"
 val JsCond = s"matrix.platform == 'js'"
+val NativeCond = s"matrix.platform == 'native'"
 
 val Scala2Cond = s"(matrix.scala != '$DottyOld' && matrix.scala != '$DottyNew')"
 val Scala3Cond = s"(matrix.scala == '$DottyOld' || matrix.scala == '$DottyNew')"
```
larsrh, line 72: We don't need this anymore, do we? I wonder why this split was in here in the first place; maybe @djspiewak can tell.

build.sbt
```diff
@@ -46,34 +46,40 @@ ThisBuild / scalaVersion := Scala213
...
-  ),
-  WorkflowStep.Sbt(List("buildJVM", "bench/test"),
-                   name = Some("Validate JVM (scala 3)"),
-                   cond = Some(JvmCond + " && " + Scala3Cond)
+                   name = Some("Validate JVM (Scala ${{ matrix.scala }})"),
+                   cond = Some(JvmCond)
   ),
   WorkflowStep.Sbt(
-    List("clean", "validateBC"), // cleaning here to avoid issues with codecov
+    List("clean", "validateBC"),
```
larsrh, line 82: If the `clean` was just here for codecov, maybe we can also get rid of it.

build.sbt
```diff
@@ -134,6 +140,7 @@ lazy val commonSettings = Seq(
   resolvers ++= Seq(Resolver.sonatypeRepo("releases"), Resolver.sonatypeRepo("snapshots")),
   parallelExecution in Test := false,
   testFrameworks += new TestFramework("munit.Framework"),
+  doctestTestFramework := DoctestTestFramework.Munit,
```
larsrh, line 143: Why is this necessary?

build.sbt
arashi01: Looking at the build I can't see any reason not to. It will also likely save quite a bit on build time as it'll avoid a rebuild for binary compatibility checks on each build in the matrix.

build.sbt
arashi01: > We don't need this anymore, do we? I wonder why this split was in here in the first place; maybe @djspiewak can tell.

It isn't used anywhere anymore. Can remove.

build.sbt
arashi01: It isn't. Can remove.

Although isn't a bad idea to explicitly set `doctestTestFramework`, and ensure relevant test library dependencies are available. Since MUnit dependencies are explicitly provided in the build for most modules, set `doctestTestFramework` to that, but can be set to any supported framework.

build.sbt
arashi01: Actually just looking at the build again there are a couple of spots `Scala2Cond` is used as a condition. `Scala3Cond`, not anymore.

build.sbt
arashi01: Removed.

build.sbt
arashi01: Reverted to original setup. Can make a separate pull request for build specific changes.

arashi01:
@larsrh I've remove all build related changed not related to adding Scala Native to build. Can make a separate pull request for build cleanup if necessary. Probably the more correct thing to do anyway!

larsrh:
> Can make a separate pull request for build cleanup if necessary. Probably the more correct thing to do anyway!

Sounds like a good plan. I'd like to solicit some reviews from others (also, take a closer look myself) before asking you to do even more work though :wink: But regardless, thanks already for this effort. It's nice to see that it doesn't require that many code changes.

larsrh (approved):
I'm happy with this change. Thanks for pushing through! However I will let it sit for a few more days in case anyone complains.

djspiewak (approved):

larsrh:
... and we have lift-off :rocket: Thanks a ton @arashi01.

This will be included in the upcoming 2.4.0 release.

Independently of that, if you're so inclined, it'd be great if you could file a follow-up PR with the build changes.

arashi01:
> ... and we have lift-off rocket Thanks a ton @arashi01.
> 
> This will be included in the upcoming 2.4.0 release.
> 
> Independently of that, if you're so inclined, it'd be great if you could file a follow-up PR with the build changes.

Okay great! And will do!

PR #3761, opened 2021-01-31
remove coverage

larsrh:
Completely rips out coverage from the Cats build.

Scoverage continues to be painful to support, see for example #3750. Furthermore, upstream hasn't released anything in quite a while, which blocks even Scala 2.12 and 2.13 support.

tl;dr coverage is not pulling its weight


rossabaker (approved):

mpilquist (approved):

larsrh:
Let the record reflect that @rossabaker approved, but I had to force-push :wink:

PR #3767, opened 2021-02-03
Replace actions/setup-ruby with ruby/setup-ruby

sloshy:
The cats microsite build is silently failing or otherwise having some problems: https://github.com/typelevel/cats/runs/1825538057#step:7:356

This problem twofold: first is, the `actions/setup-ruby` action becoming deprecated in favor of `ruby/setup-ruby`. Second is, there is no version of ruby specified. This swaps out the old action for the new one, and specifies a stable version of ruby. A similar fix was applied to 47deg repositories recently, and this fix will also need to be applied to other Typelevel ecosystem repositories if they also use Jekyll-based microsites.


rossabaker (approved):
I had to do similar on case-insensitive recently.  :+1:

We also do tricks with Gemfile.lock and bundler, which I don't see in this repo.  From what I remember, even installing a specific version of jekyll is insufficient as default dependencies change. Is that no longer true?

sloshy:
@rossabaker I have no idea, I'm a bit new to the world of jekyll. I'll defer to your judgement on whether we need that here.

ChristopherDavenport (approved):

rossabaker:
I think a Gemfile.lock is still required for a properly repeatable build, but what you've done here is both necessary and immediately beneficial.  Let's merge, and we can talk about more and better later.

Issue #3766, opened 2021-02-03
Proposal: merge `cats-time` into `cats`


diesalbla:
[`cats-time`](https://github.com/typelevel/cats-time) is a microlibrary that implements instances of kernel typeclasses, such as `Show`, `Ord`, `Eq`, or `Hash`, for the classes of the `java.time` package. This package was introduced in version 8.0 of the JDK, in order to replace previous unsatisfactory classes (such as `java.Date`), and was derived from [`joda-time`](https://www.joda.org/joda-time/).

The proposal is that these files be merged into the `kernel` package, and the `cats-time` library archived. 

Some considerations on this matter are as follows: 
- **Precedent**: there is a trait of instances for `java.util.UUID` already in `kernel`.
- `java.time` is just as standard a library as `Option`, `Either`, so it is reasonable to also include its instances in the package. 
- Adding instances for a JDK-specific class would add a coupling between `cats` and the JVM platform specifically, but as mentioned that already happens with `java.util.UUID`.
- This would make the `cats` kernel incompatible with JDK pre-8 versions. The existing instance for `UUID` only adds incompatibility with pre-1.5 JDKs. Note that JDK 8 was released in March 2014, almost seven years ago, so I am not sure how far `cats` retrocompatibility should go.

DavidGregory084:
Since cats already requires Java 8+ as of version 2.1.0 because it's only published for Scala 2.12+, I think this proposal is very sensible

djspiewak:
The only major concern here would be around ScalaJS support.

diesalbla:
> The only major concern here would be around ScalaJS support.

In other projects there is a separation of sources that are JS-specific or JVM-specific. Perhaps we can do the same here? 
As noted, this already happens with `UUID`.

rossabaker:
Strong :+1:.  We shouldn't accept a lesser experience on our flagship platform when a feature is not supported by a secondary platform.  And for the Java 8 requirement, public support for Java 7 ended in 2015.

diesalbla:
Taking existing support into account, I have opened issue to integrate this into the `cats.kernel`...

PR #3768, opened 2021-02-03
Integrate cats-time into cats-kernel.

diesalbla:
The `cats-time` microlibrary publishes instances of the typeclasses of cats.kernel and cats for the classes in the java.time package. In [Issue 3766](https://github.com/typelevel/cats/issues/3766) it was proposed to integrate this library into cats itself, considering that for JVM users the `java.time` classes are just as standard as the UUID or the Scala Standard Library ones.

This is started as a draft Pull Request, because there are a couple preferences that people may want to recommend. The code from `cats-time` is copied mostly without changes, at opening the PR. The only change being the names of the packages.

- The current structure is a folder, with a lot of 
- Due notice has to be added in the `LICENSE` to the original authors. https://github.com/typelevel/cats-time/blob/master/LICENSE
-

### DIFF
--- a/kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
+++ b/kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
@@ -14,6 +14,7 @@ import scala.collection.immutable.{BitSet, Queue, SortedMap, SortedSet}
 import scala.util.Random
 import java.util.UUID
 import java.util.concurrent.TimeUnit.{DAYS, HOURS, MICROSECONDS, MILLISECONDS, MINUTES, NANOSECONDS, SECONDS}
+import java.time.{Instant, Duration, ZonedDateTime, ZoneId, Instant, LocalTime, Year, YearMonth}
 import compat.scalaVersionSpecific._
 import munit.ScalaCheckSuite
 import org.scalacheck.Test.Parameters
@@ -128,6 +129,8 @@ class Tests extends TestsConfig with DisciplineSuite {
   checkAll("Eq[Vector[HasEq[Int]]]", EqTests[Vector[HasEq[Int]]].eqv)
   checkAll("Eq[Stream[HasEq[Int]]]", EqTests[Stream[HasEq[Int]]].eqv)
   checkAll("Eq[Queue[HasEq[Int]]]", EqTests[Queue[HasEq[Int]]].eqv)
+  checkAll("Eq[java.time.ZoneId]", EqTests[ZoneId].eqv)
+
 
   checkAll("PartialOrder[Set[Int]]", PartialOrderTests[Set[Int]].partialOrder)
   checkAll("PartialOrder.reverse(PartialOrder[Set[Int]])",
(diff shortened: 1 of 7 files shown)


mpilquist:
Any reason this needs to be in kernel versus core?

diesalbla:
> Any reason this needs to be in kernel versus core?

What would the key criteria to separate between them be? The main distinction I knew was that kernel only have kind `*` typeclasses (Show, Eq, Ord, Hash, etc) whereas the core has typeclasses of kind `* -> *`, such as Functor, Monad, Traverse, etc.

kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
```diff
@@ -148,7 +149,7 @@ class TestsConfig extends ScalaCheckSuite {
 
 class Tests extends TestsConfig with DisciplineSuite {
 
-  import KernelCheck._
+nnnn  import KernelCheck._
```
larsrh, line 152:
```suggestion
  import KernelCheck._
```

kernel/src/main/scala/cats/kernel/instances/time/all.scala
```diff
@@ -0,0 +1,21 @@
...
+  with offsetdatetime
+  with offsettime
+  with period
+  with year
+  with yearmonth
+  with zoneddatetime
+  with zoneid
+  with zoneoffset
+
+trait all2 extends month
```
larsrh, line 19: Why are there two traits here?

larsrh:
> Any reason this needs to be in kernel versus core?

I think kernel is exactly the right spot for those instances. We already have precedent: https://github.com/typelevel/cats/blob/v2.3.1/kernel/src/main/scala/cats/kernel/instances/FiniteDurationInstances.scala

diesalbla:
@larsrh the only inconvenient is with the `Show` instances... is there any conversation about the `kernel` vs `core` separation, the criteria, or whether `Show` should be in one or the other?

larsrh:
I don't understand your comment. `Show` is defined in core, so that's where those instances should go. All others can stay in kernel.

johnynek:
+1 to keep anything that only depends on the jdk and kernel in kernel.

kernel/src/main/scala/cats/kernel/instances/time/month.scala
```diff
@@ -0,0 +1,18 @@
...
+import java.time.format.TextStyle
+import java.util.Locale
+
+trait month {
+  implicit final val monthInstances = new Show[Month] with Order[Month] with Hash[Month] {
+    def compare(x: Month, y: Month): Int = x.compareTo(y)
+
+    def hash(x: Month): Int = x.hashCode
+
+    def show(t: Month): String = t.getDisplayName(TextStyle.SHORT, Locale.ENGLISH)
```
johnynek, line 14: I wonder about baking this in.

diesalbla:
@johnynek So, after thinking for a while, I think I am going to leave just the Kernel instances for `Eq, Order, Hash`, and Monoid where appropiate, and avoid `Show`.

diesalbla:
Closing this PR to clear the cruft.

PR #3770, opened 2021-02-05
Adding group-like extension methods

BalmungSan:
Added:

+ `toNec`
+ `groupMapNel`
+ `groupMapNec`
+ `groupMapReduce` _(requires a **Semigroup**)_
+ `groupMapReduceWith`

Related to #3680

### DIFF
--- a/core/src/main/scala/cats/syntax/list.scala
+++ b/core/src/main/scala/cats/syntax/list.scala
@@ -2,7 +2,8 @@ package cats
 package syntax
 
 import cats.data.{NonEmptyChain, NonEmptyList}
-import scala.collection.immutable.SortedMap
+import cats.kernel.Semigroup
+import scala.collection.immutable.{SortedMap, TreeMap}
 
 trait ListSyntax {
   implicit final def catsSyntaxList[A](la: List[A]): ListOps[A] = new ListOps(la)
@@ -34,14 +35,11 @@ final class ListOps[A](private val la: List[A]) extends AnyVal {
    * produced by the given mapping function.
    *
    * {{{
-   * scala> import cats.data.NonEmptyList
    * scala> import scala.collection.immutable.SortedMap
+   * scala> import cats.data.NonEmptyList
    * scala> import cats.implicits._
-   *
    * scala> val list = List(12, -2, 3, -5)
-   *
    * scala> val expectedResult = SortedMap(false -> NonEmptyList.of(-2, -5), true -> NonEmptyList.of(12, 3))
-   *
    * scala> list.groupByNel(_ >= 0) === expectedResult
    * res0: Boolean = true
    * }}}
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/syntax/list.scala
```diff
@@ -51,6 +49,80 @@ final class ListOps[A](private val la: List[A]) extends AnyVal {
...
+   * {{{
+   * scala> import scala.collection.immutable.SortedMap
+   * scala> import cats.implicits._
+   * scala> val list = List("Hello", "World", "Goodbye", "World")
+   * scala> val expectedResult = SortedMap("goodbye" -> 1, "hello" -> 1, "world" -> 2)
+   * scala> list.groupMapReduce(_.trim.toLowerCase)(_ => 1) === expectedResult
+   * res0: Boolean = true
+   * }}}
+   */
+  def groupMapReduce[K, B](key: A => K)(f: A => B)(implicit K: Order[K], B: Semigroup[B]): SortedMap[K, B] =
```
BalmungSan, line 90: Not sure how to name this one so it doesn't clash with the one introduced by the stdlib in `2.13`

Also, not sure if it worth adding since in `2.13` one can easily pass the `combine` operation provided by the **Semigroup**.

core/src/main/scala/cats/syntax/list.scala
```diff
@@ -51,6 +49,80 @@ final class ListOps[A](private val la: List[A]) extends AnyVal {
...
+   * {{{
+   * scala> import scala.collection.immutable.SortedMap
+   * scala> import cats.implicits._
+   * scala> val list = List("Hello", "World", "Goodbye", "World")
+   * scala> val expectedResult = SortedMap("goodbye" -> 1, "hello" -> 1, "world" -> 2)
+   * scala> list.groupMapReduceWith(_.trim.toLowerCase)(_ => 1)(_ + _) === expectedResult
+   * res0: Boolean = true
+   * }}}
+   */
+  def groupMapReduceWith[K, B](key: A => K)(f: A => B)(combine: (B, B) => B)(implicit K: Order[K]): SortedMap[K, B] = {
```
BalmungSan, line 110: Not sure if it is worth it to add this method.

It is mostly the same as the one in the stdlib, just that this returns a **SortedMap**.
_(And that it is available in `2.12`)_

larsrh:
For the record, I asked @BalmungSan to remove this from #3680 and create a new PR for the extensions (this one).

I'm unsure how to proceed with this one. Any opinions?

mcanlas:
What happened to the `toNec` mentioned at the top of this PR?

BalmungSan:
@mcanlas My memory is fuzzy, but IIRC, we decided not to add extensions to `List` and only to type that **cats** control like `NEL`.
Not sure why tho, I really liked the `toNec`.

PR #3772, opened 2021-02-06
another attempt at Cogen for (Finite)Duration

larsrh:
This is #3470 redux, although with another approach.

Contains code by @barambani to test the issue.

I've confirmed that the newly-added tests fail _without_ the `Cogen` change.

cc @TimWSpence

### DIFF
--- a/kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
+++ b/kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
@@ -107,30 +107,11 @@ object KernelCheck {
       if (d == Duration.Inf) 3896691548866406746L
       else if (d == Duration.MinusInf) 1844151880988859955L
       else if (d == Duration.Undefined) -7917359255778781894L
-      else
-        d.length * (d.unit match {
-          case DAYS         => -6307593037248227856L
-          case HOURS        => -3527447467459552709L
-          case MINUTES      => 5955657079535371609L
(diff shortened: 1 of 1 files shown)


barambani (approved):

kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
```diff
@@ -107,30 +107,11 @@ object KernelCheck {
...
-        case DAYS         => -6307593037248227856L
-        case HOURS        => -3527447467459552709L
-        case MINUTES      => 5955657079535371609L
-        case SECONDS      => 5314272869665647192L
-        case MILLISECONDS => -2025740217814855607L
-        case MICROSECONDS => -2965853209268633779L
-        case NANOSECONDS  => 6128745701389500153L
-      })
-    }
+    Cogen[Long].contramap(_.toNanos)
```
LukaJCB, line 114: @larsrh this is the same as the one defined in ScalaCheck currently, so maybe we should just delete it?

larsrh: Excellent point. I just checked ans ScalaCheck has all combinations of {`Arbitrary`, `Cogen`} and {`Duration`, `FiniteDuration`}. Should I delete them too?

larsrh: #3775

PR #3222, opened 2019-12-21
Add `Invariant[Numeric]`

kubukoz:
Closes #3220.

I was surprised to see binary compatibility errors when I put the instance inside `InvariantMonoidalInstances`:

```
sbt:cats> coreJVM/mimaReportBinaryIssues
[info] Compiling 1 Scala source to /Users/kubukoz/IdeaProjects/cats/core/.jvm/target/scala-2.12/classes ...
[info] Compiling 4 Scala sources to /Users/kubukoz/IdeaProjects/cats/core/.jvm/target/scala-2.12/classes ...
[error]  * abstract synthetic method cats$instances$InvariantInstances$_setter_$catsInvariantNumeric_=(cats.Invariant)Unit in interface cats.instances.InvariantInstances is inherited by class AllInstances in current version.
[error]    filter with: ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("cats.instances.AllInstances.cats$instances$InvariantInstances$_setter_$catsInvariantNumeric_=")
[error]  * abstract method catsInvariantNumeric()cats.Invariant in interface cats.instances.InvariantInstances is inherited by class AllInstances in current version.
[error]    filter with: ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("cats.instances.AllInstances.catsInvariantNumeric")
[error]  * abstract synthetic method cats$instances$InvariantInstances$_setter_$catsInvariantNumeric_=(cats.Invariant)Unit in interface cats.instances.InvariantInstances is inherited by class InvariantMonoidalInstances in current version.
[error]    filter with: ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("cats.instances.InvariantMonoidalInstances.cats$instances$InvariantInstances$_setter_$catsInvariantNumeric_=")
[error]  * abstract method catsInvariantNumeric()cats.Invariant in interface cats.instances.InvariantInstances is inherited by class InvariantMonoidalInstances in current version.
[error]    filter with: ProblemFilters.exclude[InheritedNewAbstractMethodProblem]("cats.instances.InvariantMonoidalInstances.catsInvariantNumeric")
[error] stack trace is suppressed; run last coreJVM / mimaReportBinaryIssues for the full output
[error] (coreJVM / mimaReportBinaryIssues) Cats core: Binary compatibility check failed!
[error] Total time: 4 s, completed Dec 21, 2019, 2:39:27 PM
```

So I added a workaround (another trait). If anyone has a better idea I'm open for suggestions :)

PR #3773, opened 2021-02-07
Add Invariant instances for Numeric and Integral

### LINKED ISSUE #3220: Add `Invariant[Numeric]`
Here's a ready instance one can use:

```scala
implicit val invariantNumeric: Invariant[Numeric] = new Invariant[Numeric] {

  def imap[A, B](fa: Numeric[A])(f: A => B)(g: B => A): Numeric[B] = new Numeric[B] {
    def compare(x: B, y: B): Int = fa.compare(g(x), g(y))
    def plus(x: B, y: B): B = f(fa.plus(g(x), g(y)))
    def minus(x: B, y: B): B = f(fa.minus(g(x), g(y)))
    def times(x: B, y: B): B = f(fa.times(g(x), g(y)))
    def negate(x: B): B = f(fa.negate(g(x)))
    def fromInt(x: Int): B = f(fa.fromInt(x))
    def toInt(x: B): Int = fa.toInt(g(x))
    def toLong(x: B): Long = fa.toLong(g(x))
    def toFloat(x: B): Float = fa.toFloat(g(x))
    def toDouble(x: B): Double = fa.toDouble(g(x))
  }
}
```

Feel free to grab it if you have time before I do.

tmccarthy:
Fixes #3220.

This PR continues the work done in #3222 by @kubukoz. My additional work is:

* Add instances on the `Invariant` companion
* Fix binary compatibility issues with `AllInstances`
* Support `2.12` and `2.13+` by handling the `parseString` method that was added to `Numeric` in `2.13`
* Add `Invariant[Integral]`

Note that I haven't added `Invariant[Fractional]` because the laws testing looks to be a bit messy. I might think a bit more on how to test such an instance and add it in a separate PR.

### DIFF
--- a/core/src/main/scala-2.13+/cats/instances/ScalaVersionSpecificNumeric.scala
+++ b/core/src/main/scala-2.13+/cats/instances/ScalaVersionSpecificNumeric.scala
@@ -1,0 +1,16 @@
+package cats.instances
+
+abstract private[instances] class ScalaVersionSpecificNumeric[A, B](fa: Numeric[A])(f: A => B)(g: B => A)
+    extends Numeric[B] {
+  def compare(x: B, y: B): Int = fa.compare(g(x), g(y))
+  def plus(x: B, y: B): B = f(fa.plus(g(x), g(y)))
+  def minus(x: B, y: B): B = f(fa.minus(g(x), g(y)))
+  def times(x: B, y: B): B = f(fa.times(g(x), g(y)))
+  def negate(x: B): B = f(fa.negate(g(x)))
+  def fromInt(x: Int): B = f(fa.fromInt(x))
+  def toInt(x: B): Int = fa.toInt(g(x))
+  def toLong(x: B): Long = fa.toLong(g(x))
+  def toFloat(x: B): Float = fa.toFloat(g(x))
+  def toDouble(x: B): Double = fa.toDouble(g(x))
+  def parseString(str: String): Option[B] = fa.parseString(str).map(f)
+}
--- a/tests/src/test/scala-2.12/cats/tests/ScalaVersionSpecific.scala
+++ b/tests/src/test/scala-2.12/cats/tests/ScalaVersionSpecific.scala
@@ -12,7 +12,7 @@ trait ScalaVersionSpecificRegressionSuite
 trait ScalaVersionSpecificTraverseSuite
 
 trait ScalaVersionSpecificAlgebraInvariantSuite {
-  protected val numericForMiniInt: Integral[MiniInt] = new Integral[MiniInt] {
+  protected val integralForMiniInt: Integral[MiniInt] = new Integral[MiniInt] {
     def compare(x: MiniInt, y: MiniInt): Int = Order[MiniInt].compare(x, y)
     def plus(x: MiniInt, y: MiniInt): MiniInt = x + y
     def minus(x: MiniInt, y: MiniInt): MiniInt = x + (-y)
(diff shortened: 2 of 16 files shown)


core/src/main/scala-2.13+/cats/instances/ScalaVersionSpecificNumeric.scala
```diff
@@ -0,0 +1,16 @@
...
+  def plus(x: B, y: B): B = f(fa.plus(g(x), g(y)))
+  def minus(x: B, y: B): B = f(fa.minus(g(x), g(y)))
+  def times(x: B, y: B): B = f(fa.times(g(x), g(y)))
+  def negate(x: B): B = f(fa.negate(g(x)))
+  def fromInt(x: Int): B = f(fa.fromInt(x))
+  def toInt(x: B): Int = fa.toInt(g(x))
+  def toLong(x: B): Long = fa.toLong(g(x))
+  def toFloat(x: B): Float = fa.toFloat(g(x))
+  def toDouble(x: B): Double = fa.toDouble(g(x))
+  def parseString(str: String): Option[B] = fa.parseString(str).map(f)
```
tmccarthy, line 15: This differs from the `2.12` implementation because `parseString` was added in `2.13`.

tests/src/test/scala/cats/tests/AlgebraInvariantSuite.scala
```diff
@@ -237,11 +268,6 @@ class AlgebraInvariantSuite extends CatsSuite {
   checkAll("Invariant[Monoid]", InvariantTests[Monoid].invariant[Option[MiniInt], Boolean, Boolean])
   checkAll("Invariant[Monoid]", SerializableTests.serializable(Invariant[Monoid]))
 
-  Eq[Band[Set[Boolean]]]
-  cats.laws.discipline.ExhaustiveCheck[Set[Boolean]]
-  Eq[(Set[Boolean], Boolean)]
-  Eq[(Set[Boolean], Set[Boolean] => (Set[Boolean], Boolean))]
-  Eq[CommutativeSemigroup[Set[Boolean]]]
```
tmccarthy, line 244: There was a [question](https://github.com/typelevel/cats/pull/3222#discussion_r360649967) about these lines in the previous PR. I've removed them but happy to revert if they are necessary.

tests/src/test/scala-2.12/cats/tests/ScalaVersionSpecific.scala
```diff
@@ -1,6 +1,57 @@
...
+
+  implicit protected def eqNumeric[A: Eq: ExhaustiveCheck]: Eq[Numeric[A]] = {
+
+    // These allow us to capture the cases where operations on Numeric throw (eg when causing an overflow). These are
+    // represented by `None` and allow us to compare two Numeric instances, verifying that when one throws, the other
+    // also throws.
+    def makeUnaryFnSafe[X, R](f: X => R): X => Option[R] =
+      x => Either.catchOnly[IllegalArgumentException](f(x)).toOption
+    def makeBinaryFnSafe[X, Y, R](f: (X, Y) => R): (X, Y) => Option[R] =
+      (x, y) => Either.catchOnly[IllegalArgumentException](f(x, y)).toOption
```
tmccarthy, line 38: Note that here and in a couple of other places, I've had to find a way of reconciling the unsafe nature of the methods on `Numeric` with defining an `Eq`.

My approach has been to require that for two `Numeric` instances to be equal, when one throws the other throws. Lifting the values into `Option` and using `None` to represent cases that throw erases the difference between different types of thrown exceptions. I've decided that this is okay given the alternative would be a dirty `Eq` for `ArithmaticException` and `IllegalArgumentException`.

Happy to change this approach to something else.

tmccarthy:
The [test failure](https://github.com/typelevel/cats/pull/3773/checks?check_run_id=1848317821#step:7:14380) looks to be unrelated to my changes.

tests/src/test/scala-2.12/cats/tests/ScalaVersionSpecific.scala
tmccarthy: Actually I think this is unnecessary except in the `fromInt` case. I will rework it...

tests/src/test/scala/cats/tests/AlgebraInvariantSuite.scala
larsrh: They might be compilation tests, no?

larsrh:
> The test failure looks to be unrelated to my changes.

Probably a flaky test.

tests/src/test/scala/cats/tests/AlgebraInvariantSuite.scala
larsrh: On second thought, they probably don't belong there at all. So let's keep them removed.

larsrh (approved):

PR #3779, opened 2021-02-08
add Chain#splitAt

mtomko:

add Chain#splitAt

### DIFF
--- a/core/src/main/scala/cats/data/Chain.scala
+++ b/core/src/main/scala/cats/data/Chain.scala
@@ -181,6 +181,36 @@ sealed abstract class Chain[+A] {
   }
 
   /**
+   * Splits the chain into two, placing the initial `n` elements in the first element of the tuple
+   * and the remaining elements in the second element of the tuple. If `n` is greater than the length
+   * of this Chain, the first element of the tuple will contain the entire Chain and the second element
+   * will be an empty chain.
+   */
+  final def splitAt(n: Int): (Chain[A], Chain[A]) = {
+    @tailrec
+    def loop(togo: Int, acc: Chain[A], rest: Chain[A]): (Chain[A], Chain[A]) = {
+      if (togo == 0) (acc, rest)
+      else
+        rest match {
+          case Empty            => (acc, Empty)
+          case s @ Singleton(_) => (Chain.concat(acc, s), Empty)
+          case Wrap(seq) =>
+            val (l, r) = seq.splitAt(togo)
+            (Chain.concat(acc, Wrap(l)), Wrap(r))
+          case Append(l, r) =>
+            if (l.size > togo) {
+              val (a, b) = l.splitAt(togo)
+              (Chain.concat(acc, a), Chain.concat(b, r))
+            } else {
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/data/Chain.scala
```diff
@@ -180,6 +180,36 @@ sealed abstract class Chain[+A] {
...
+    @tailrec
+    def loop(togo: Int, acc: Chain[A], rest: Chain[A]): (Chain[A], Chain[A]) = {
+      if (togo == 0) (acc, rest)
+      else
+        rest match {
+          case Empty            => (acc, Empty)
+          case s @ Singleton(_) => (Chain.concat(acc, s), Empty)
+          case Wrap(seq) =>
+            val (l, r) = seq.splitAt(togo)
+            (Chain.concat(acc, Wrap(l)), Wrap(r))
```
johnynek, line 199: this Wrap is not safe... I think an invariant of Wrap is that it has length 2 or more. By manually constructing this, you could violate that. I think we should use fromSeq on both of these.

core/src/main/scala/cats/data/Chain.scala
```diff
@@ -180,6 +180,36 @@ sealed abstract class Chain[+A] {
...
+          case Append(l, r) =>
+            if (l.size > togo) {
```
johnynek, line 201: size is O(N) on chain, and you compute it twice in this call. Let's at least memoize it.

that said, I think it would be better to do something like:
```scala
val (a, b) = l.splitAt(togo)
if (b.isEmpty) {
  // then a.size <= togo, we have to compute the size
  val (a2, b2) = r.splitAt(togo - a.size)
  (Chain.concat(a, a2), b2)
}
else {
  // we know that a.size == togo
  (a, Chain.concat(b, r))
}
```
If we do this, I think we actually never recurse on loop at all, which does open the question: is this stack safe?

An alternate approach would be:

```scala
val left = Chain.fromSeq(toIterator.take(n).toVector)
val right = Chain.fromSeq(toIterator.drop(n).toVector)
(left, right)
```
The above is safer, but may be slower since it always does O(N) work.

I'd love to find a way to do O(n) (the parameter) work, and be stack safe.

core/src/main/scala/cats/data/Chain.scala
mtomko: Will fix.

core/src/main/scala/cats/data/Chain.scala
mtomko: Good points. I suspect it's possible to traverse the spine of the chain in a stack-safe manner but the O(n) size calculation may make the point moot. I will think about this some more.

mtomko: Looking at this further, I suspect that it's only possible to do the stack-safe traversal efficiently if we can compute the size efficiently. I'll give it a tiny bit more thought but if I don't have any great ideas, I'll probably just close this.

mtomko:
Closing. I don't think I have a good idea how to do this efficiently and preserve stack safety. Thanks for your input!

Issue #3783, opened 2021-02-10
Follow-up : IorT's Parallel's need amending as well 


Baccata:
### The actual issue, following Daniel's axiom of "the outer most effect should take precedence in parallel composition"

Re-re-re-reading IorT code, this line needs changing to avoid error aggregation on the unhappy-path : https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/IorT.scala#L452 

Follow-up to https://github.com/typelevel/cats/issues/3776

### Regarding cats-effect Concurrent => Parallel : 

Here's the original snippet that lead to the opening of the original issue, amended to present the issue with IorT instead of EitherT, proving that the Parallel instance derived from Concurrent is still incoherent 

```scala
import $ivy.`org.typelevel::cats-effect:3.0.0-M5`

import cats._
import cats.effect._
import cats.syntax.all._
import cats.data.IorT

import scala.concurrent.duration._

val foo: IorT[IO, Int, Unit] =
  IorT.liftF[IO, Int, Unit](IO.sleep(1.second)) *>
    IorT.fromEither[IO](1.asLeft[Unit])


val withStandardInstance: IO[Unit] = {
  (foo, foo).parTupled.value.flatMap(IO.println)
}

val withCEInstance: IO[Unit] = {
  import cats.effect.implicits._
  (foo, foo).parTupled.value.flatMap(IO.println)
}

import cats.effect.unsafe.implicits.global

(withStandardInstance &> withCEInstance).unsafeRunTimed(1.5.second)
```

This prints the following, showing incoherence
  
```
Left(2)
Left(1)
 ```

### Below is to be disregarded

~~So, people are gonna hate me, but I'm re-opening the debate because I think `Ior/IorT's` semantics fundamentally disprove the arguments that lead to the changes in EitherT's Parallel.~~

~~Now one of the arguments that Daniel used to justify the change to EitherT was that MonadTransformers in general ought to apply [a single layer](https://github.com/typelevel/cats/pull/3777#issuecomment-775322944) of effects.~~ 

~~Although I disagreed with that point of view reflecting something that is fundamental, I decided to agree on the account of it making it possible to achieve coherence when deriving `Parallel` from `Concurrent`, which is something that a lot of people want.~~ 

~~But when it comes to `Ior/IorT`, it's not that simple,  because `Ior` is kind of inherently parallel.~~

```scala
@ type Inner[A] = Ior[Int, A]
defined type Inner

[...]

PR #3791, opened 2021-02-12
Prevent additive effects of IorT when F is parallel. 

Baccata:
Fixes https://github.com/typelevel/cats/issues/3783

* Applies the same fix as https://github.com/typelevel/cats/pull/3777 
* Adds tests to verify non-additive behaviour of Left in  EitherT / IorT when F is Parallel 
* Adds law checks for additiveParallel on both IorT and EitherT

### DIFF
--- a/core/src/main/scala/cats/data/IorT.scala
+++ b/core/src/main/scala/cats/data/IorT.scala
@@ -488,10 +488,9 @@ abstract private[data] class IorTInstances extends IorTInstances1 {
   def accumulatingParallel[M[_], E](implicit
     P: Parallel[M],
     E: Semigroup[E]
-  ): Parallel.Aux[IorT[M, E, *], IorT[P.F, E, *]] { type Dummy } =
+  ): Parallel.Aux[IorT[M, E, *], IorT[P.F, E, *]] =
     new Parallel[IorT[M, E, *]] {
       type F[x] = IorT[P.F, E, x]
-      type Dummy // fix to make this one more specific than the catsDataParallelForIorTWithSequentialEffect, see https://github.com/typelevel/cats/pull/2335#issuecomment-408249775
 
       val parallel: IorT[M, E, *] ~> IorT[P.F, E, *] =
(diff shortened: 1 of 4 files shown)


core/src/main/scala/cats/data/IorT.scala
```diff
@@ -439,6 +439,60 @@ abstract private[data] class IorTInstances extends IorTInstances1 {
...
+   * (a, b).parTupled  // => IorT(IO(Chain(error1, error2).leftIor[Unit]))
+   * }}}
+   */
+  def accumulatingParallel[M[_], E](implicit
+    P: Parallel[M],
+    E: Semigroup[E]
+  ): Parallel.Aux[IorT[M, E, *], IorT[P.F, E, *]] { type Dummy } =
+    new Parallel[IorT[M, E, *]] {
+      type F[x] = IorT[P.F, E, x]
+      type Dummy // fix to make this one more specific than the catsDataParallelForIorTWithSequentialEffect, see https://github.com/typelevel/cats/pull/2335#issuecomment-408249775
```
djspiewak, line 494: I think we can remove the `Dummy` types here, since they're just an implicit search workaround.

djspiewak (approved):
Looks good aside from my one minor comment.

core/src/main/scala/cats/data/IorT.scala
Baccata: Addressed 49fafb0

larsrh:
I'm having trouble understanding the diff. It looks like there's just one additional non-implicit definition, no? In that case, why is this marked as source breaking?

Baccata:
@larsrh I swapped to methods to help the diff. [Here's](https://github.com/typelevel/cats/pull/3791/files#diff-2bdfce46cbcc6ea0c7911c010533de85fe0e87a8ae7d4aaa854529d67a8b8369R504-R505) what's making it source-breaking.

larsrh (approved):

PR #3792, opened 2021-02-13
Revert implicitness of Empty.fromEmptyK

joroKr21:
This breaks kittens: typelevel/kittens#299
PR that made it implicit: #3677

But more importantly, `MonoidK -> Monoid` and `SemigroupK -> Semigroup`
are not implicit either. I would err on the side of consistency here.


barambani (approved):

LukaJCB (approved):

larsrh:
Do we need another hotfix release? :sweat_smile:

joroKr21:
> Do we need another hotfix release?

I think not - nobody has complained yet and there is no rush to update cats in kittens.

PR #3795, opened 2021-02-15
Fix year in changelog

me-ydv-5:


larsrh:
Thanks, good catch!

larsrh (approved):

PR #3798, opened 2021-02-16
Adding NonEmptyList.unapply

### LINKED ISSUE #3663: unapply for NonEmptyList (and friends)
I wonder why there aren't any `unapply` functions on structures like the NEL? It would allow me to safely operate on lists in pattern matching without explicitly wrapping them in NEL, e.g. given:

```scala
object NonEmptyList {
  def unapply[A](list: List[A]): Option[NonEmptyList[A]] =
      NonEmptyList.fromList(list)
}
```

Then I could use it like:

```scala
val someList: List[Item] = ...
...
someList match {
  case NonEmptyList(items) => ...
  ...
}
```

Any reason not to include it?

hmemcpy:
Fixes #3663 

This PR adds 2 `unapply` functions: one is the irrefutable unapply which works on any NEL (to keep existing code compiling in the presence of a custom extractor) and a custom `unapply` that works with regular Lists, optionally turning them into a NEL, so it can be used in pattern matching:

```scala
val list = List(1,2,3)
...

list match {
  case NonEmptyList(nel) => ...
  ...
}
```

I don't have a better idea for the test, would love any suggestions!

### DIFF
--- a/core/src/main/scala/cats/data/NonEmptyList.scala
+++ b/core/src/main/scala/cats/data/NonEmptyList.scala
@@ -640,6 +640,10 @@ final case class NonEmptyList[+A](head: A, tail: List[A]) extends NonEmptyCollec
 }
 
 object NonEmptyList extends NonEmptyListInstances {
+  def unapply[A](list: List[A]): Option[NonEmptyList[A]] = fromList(list)
+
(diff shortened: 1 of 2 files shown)


hmemcpy:
I can add `.unapply` for the rest of the NonEmpty* family as well, if this is accepted.

hmemcpy:
Hmm, unfortunately doing this seems to break mima... any hints on how to possibly solve it?

PR #3805, opened 2021-02-19
Add Ior.mergeWith and AlignMergeWith

### LINKED ISSUE #3804: Add Ior#mergeWith and Align#alignMergeWith
These are versions of `Ior#merge` and `Align#alignCombine` that use explicit functions rather than `Semigroup` instances.

Motivation: `alignMergeWith` is useful in cases where the logic in case of a "conflict" doesn't match the default `Semigroup` instance. It is also useful in places where relying on a specific `Semigroup` makes the code brittle. For example, when combining nested `Map`s and `NonEmptyMap`s, changing `NonEmptyMap` to `Map` might introduce subtle errors due to different `Semigroup` semantics - see #3377.

nigredo-tori:
Closes #3804

### DIFF
--- a/core/src/main/scala/cats/Align.scala
+++ b/core/src/main/scala/cats/Align.scala
@@ -57,6 +57,19 @@ import scala.annotation.implicitNotFound
     alignWith(fa1, fa2)(_.merge)
 
   /**
+   * Align two structures with the same element, combining results according to the given function.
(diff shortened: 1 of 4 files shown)


larsrh (approved):

larsrh:
Thanks!

PR #3807, opened 2021-02-19
Replace Band instance for NonEmptyMap with a Semigroup instance consistent with SortedMap

### LINKED ISSUE #3377: Semigroup for NonEmptyMap is kind of weird
In any recent Cats version (back to before 2.0) `|+|` is `++` for non-empty maps:
```
scala> import cats.data.NonEmptyMap, cats.implicits._
import cats.data.NonEmptyMap
import cats.implicits._

scala> NonEmptyMap.of("a" -> List(1)) |+| NonEmptyMap.of("a" -> List(2))
res0: cats.data.NonEmptyMap[String,List[Int]] = Map(a -> List(2))
```

I don't really use `NonEmptyMap`, so maybe I'm missing something, but this seems like the wrong thing to do, since `|+|` for other maps combines at the element level:

```scala
scala> Map("a" -> List(1)) |+| Map("a" -> List(2))
res1: scala.collection.immutable.Map[String,List[Int]] = Map(a -> List(1, 2))
```
We can fix this but it'll take some fussing with, since we'll have to make the current `Band` instance non-implicit, add some new prioritized instances, etc.

nigredo-tori:
Fixes #3377 

:warning: This is source-breaking!

### DIFF
--- a/core/src/main/scala/cats/data/NonEmptyMapImpl.scala
+++ b/core/src/main/scala/cats/data/NonEmptyMapImpl.scala
@@ -317,9 +317,18 @@ sealed abstract private[data] class NonEmptyMapInstances extends NonEmptyMapInst
   implicit def catsDataShowForNonEmptyMap[K: Show, A: Show]: Show[NonEmptyMap[K, A]] =
     Show.show[NonEmptyMap[K, A]](_.show)
 
-  implicit def catsDataBandForNonEmptyMap[K, A]: Band[NonEmptyMap[K, A]] =
+  @deprecated("Use catsDataSemigroupForNonEmptyMap", "2.5.0")
+  def catsDataBandForNonEmptyMap[K, A]: Band[NonEmptyMap[K, A]] =
     new Band[NonEmptyMap[K, A]] {
       def combine(x: NonEmptyMap[K, A], y: NonEmptyMap[K, A]): NonEmptyMap[K, A] = x ++ y
+    }
+
+  implicit def catsDataSemigroupForNonEmptyMap[K, A: Semigroup]: Semigroup[NonEmptyMap[K, A]] =
+    new Semigroup[NonEmptyMap[K, A]] {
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/data/NonEmptyMapImpl.scala
```diff
@@ -317,10 +317,19 @@ sealed abstract private[data] class NonEmptyMapInstances extends NonEmptyMapInst
   implicit def catsDataShowForNonEmptyMap[K: Show, A: Show]: Show[NonEmptyMap[K, A]] =
     Show.show[NonEmptyMap[K, A]](_.show)
 
-  implicit def catsDataBandForNonEmptyMap[K, A]: Band[NonEmptyMap[K, A]] =
+  @deprecated("Use catsDataSemigroupForNonEmptyMap", "2.5.0")
+  def catsDataBandForNonEmptyMap[K, A]: Band[NonEmptyMap[K, A]] =
```
johnynek, line 321: can we not move this to a lower priority? I think we could.

nigredo-tori: I don't think we should. Having a surprising `Semigroup` instance (as we do right now) is error-prone. But having two instances (one of them surprising), and choosing between them implicitly based on whether or not `A` has a `Semigroup` instance in scope - that is downright evil, IMO.

joroKr21: Also it wouldn't work, because `Band <: Semigroup` so if we put a `Band` instance in a lower-priority scope, they will become ambiguous (lower priority gives you -1 points, but being a subtype gives you +1 points, making the score even).

johnynek: Yeah. This was a bad suggestion on my part.

larsrh (approved):

PR #3808, opened 2021-02-19
Chain add fromOption factory method

nikiforo:
```
bench/ jmh:run -i 5 -wi 3 -f1 -t1 cats.bench.ChainBench

[info] Benchmark                         Mode  Cnt          Score         Error  Units
[info] ChainBench.createChainOption     thrpt    5  190985469,005 ± 5173146,902  ops/s
[info] ChainBench.createChainSeqOption  thrpt    5   24305512,739 ±  594330,476  ops/s
```


larsrh (approved):

larsrh:
Thanks!

PR #3813, opened 2021-02-25
Invariant[Fractional] and MiniFloat

### LINKED ISSUE #3794: Add Invariant[Fractional]
`Invariant[Numeric]` and `Invariant[Integral]` were added in #3773, but I left out `Invariant[Fractional]` primarily because it can't be tested using `MiniInt`.

I'm raising this issue to track the lack of `Invariant[Fractional]`. I'm working on this in the background and hope to have a PR in the coming weeks.

tmccarthy:
Fixes #3794

This PR adds `Invariant[Fractional]`. In order to test this instance, I have provided a new class `MiniFloat`, which is the floating-point analogue to `MiniInt`. Its range is small enough that an `ExhaustiveCheck` instance can be defined.

I'm conscious that a class like `MiniFloat` is a more significant addition than might be strictly necessary to add an instance like `Invariant[Fractional]`, but I'm hoping that it's judged to be sufficiently useful to be included in the `discipline` library.

I'll make more explanatory comments inline on the PR.

Note that the build is failing due to a [bug in the Scala Native implementation of isFinite](https://github.com/scala-native/scala-native/issues/2178). I'll try to find a solution to ignore this test.

### DIFF
--- a/laws/src/main/scala/cats/laws/discipline/MiniFloat.scala
+++ b/laws/src/main/scala/cats/laws/discipline/MiniFloat.scala
@@ -77,7 +77,7 @@ object MiniFloat {
      * Returns `None` if the given float cannot fit in an instance of `Finite`.
      */
     def from(float: Float): Option[Finite] = {
-      val exponent: Int = Math.getExponent(float)
+      val exponent: Int = getExponent(float)
       val significand: Int = math.round(float / math.pow(Finite.base.toDouble, exponent.toDouble).toFloat)
 
       if (significand == 0 || exponent < minExponent) {
@@ -110,6 +110,16 @@ object MiniFloat {
     private def withinBounds(significand: Int, exponent: Int): Boolean =
       (minExponent <= exponent && exponent <= maxExponent) &&
         (minSignificand <= significand && significand <= maxSignificand)
+
+    private val floatExponentStartBit: Int = 23
+    private val floatExponentLength: Int = 8
+    private val floatExponentBias: Int = 127
+    private val floatExponentMask: Int = ((1 << floatExponentLength) - 1) << floatExponentStartBit
+
+    // This does the same thing as java.lang.Math.getExponent, but that method is not available in scalaJS so we have to
+    // do the same thing here.
+    private def getExponent(float: Float): Int =
+      ((floatExponentMask & java.lang.Float.floatToIntBits(float)) >> floatExponentStartBit) - floatExponentBias
 
   }
 
--- a/laws/src/main/scala/cats/laws/discipline/MiniFloat.scala
+++ b/laws/src/main/scala/cats/laws/discipline/MiniFloat.scala
@@ -123,7 +123,7 @@ object MiniFloat {
 
   }
 
-  val Zero: MiniFloat = MiniFloat.from(0f)
+  val Zero: MiniFloat = Finite.zero
   val NegativeOne: MiniFloat = MiniFloat.from(-1f)
   val One: MiniFloat = MiniFloat.from(1f)
 
--- a/laws/src/main/scala/cats/laws/discipline/MiniFloat.scala
+++ b/laws/src/main/scala/cats/laws/discipline/MiniFloat.scala
@@ -134,7 +134,8 @@ object MiniFloat {
   def from(long: Long): MiniFloat     = from(long.toFloat)
 
   /**
-   * Note that since `MiniFloat` is used primarily for tests, NaN is considered equal to itself in this context.
+   * Note that since `MiniFloat` is used primarily for tests, this `Eq` instance defines `NaN` as equal to itself. This
+   * differs from the `Order` defined for `Float`.
    */
   implicit val catsLawsEqInstancesForMiniFloat: Order[MiniFloat] with Hash[MiniFloat] =
     new Order[MiniFloat] with Hash[MiniFloat] {
--- a/laws/src/main/scala/cats/laws/discipline/MiniFloat.scala
+++ b/laws/src/main/scala/cats/laws/discipline/MiniFloat.scala
@@ -24,6 +24,7 @@ sealed abstract class MiniFloat private (val toFloat: Float) {
   def -(that: MiniFloat): MiniFloat = MiniFloat.from(this.toFloat - that.toFloat)
   def *(that: MiniFloat): MiniFloat = MiniFloat.from(this.toFloat * that.toFloat)
   def /(that: MiniFloat): MiniFloat = MiniFloat.from(this.toFloat / that.toFloat)
+  // TODO %
   def unary_- : MiniFloat = MiniFloat.from(-this.toFloat)
 
   def isNaN: Boolean = toFloat.isNaN
--- a/laws/src/main/scala/cats/laws/discipline/MiniFloat.scala
+++ b/laws/src/main/scala/cats/laws/discipline/MiniFloat.scala
@@ -3,6 +3,7 @@ package laws
 package discipline
 
 import cats.implicits.toTraverseFilterOps
+import java.lang.Math
 
 /**
  * Similar to `Float`, but with a much smaller domain. The exact range of [[MiniFloat]] may be tuned from time to time,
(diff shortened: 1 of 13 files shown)


laws/src/main/scala/cats/laws/discipline/MiniFloat.scala
```diff
@@ -0,0 +1,177 @@
...
+
+  private[MiniFloat] object Finite {
+
+    private[MiniFloat] val base = 2
+
+    private val minSignificand = -2
+    private val maxSignificand = 2
+
+    private val minExponent = -1
+    private val maxExponent = 2
```
tmccarthy, line 60: These define the range of `MiniFloat`. The finite values currently are:

  *  `-8.0`
  *  `-4.0`
  *  `-2.0`
  *  `-1.0`
  *  `-0.5`
  *  `0.0`
  *  `0.5`
  *  `1.0`
  *  `2.0`
  *  `4.0`
  *  `8.0`

laws/src/main/scala/cats/laws/discipline/MiniFloat.scala
```diff
@@ -0,0 +1,177 @@
...
+
+    private val floatExponentStartBit: Int = 23
+    private val floatExponentLength: Int = 8
+    private val floatExponentBias: Int = 127
+    private val floatExponentMask: Int = ((1 << floatExponentLength) - 1) << floatExponentStartBit
+
+    // This does the same thing as java.lang.Math.getExponent, but that method is not available in scalaJS so we have to
+    // do the same thing here.
+    private def getExponent(float: Float): Int =
+      ((floatExponentMask & java.lang.Float.floatToIntBits(float)) >> floatExponentStartBit) - floatExponentBias
```
tmccarthy, line 122: Note that I've had to write a dedicated function to extract the exponent from a `Float` since `java.lang.Math.getExponent` is not currently defined in ScalaJS. I'll probably look to add this over there soon.

laws/src/main/scala/cats/laws/discipline/MiniFloat.scala
```diff
@@ -0,0 +1,177 @@
...
+
+  /**
+   * Note that since `MiniFloat` is used primarily for tests, this `Eq` instance defines `NaN` as equal to itself. This
+   * differs from the `Order` defined for `Float`.
+   */
+  implicit val catsLawsEqInstancesForMiniFloat: Order[MiniFloat] with Hash[MiniFloat] =
+    new Order[MiniFloat] with Hash[MiniFloat] {
+      override def compare(x: MiniFloat, y: MiniFloat): Int = Order[Float].compare(x.toFloat, y.toFloat)
+      override def hash(x: MiniFloat): Int = Hash[Float].hash(x.toFloat)
+    }
```
tmccarthy, line 165: Because we're using this class in tests it's really only useful if we define `Eq` such that `NaN == NaN`. I'm aware that this is slightly unintuitive but otherwise it becomes useless.

An alternative approach would be to provide this `Eq` in such a way that it needs to be manually imported when used. I'd be happy to make that change as needed.

laws/src/main/scala/cats/laws/discipline/MiniFloat.scala
```diff
@@ -0,0 +1,177 @@
...
+
+  implicit val catsLawsExhaustiveCheckForMiniInt: ExhaustiveCheck[MiniFloat] = new ExhaustiveCheck[MiniFloat] {
+    override def allValues: List[MiniFloat] = MiniFloat.allValues
+  }
+
+  val miniFloatMax: CommutativeMonoid[MiniFloat] with BoundedSemilattice[MiniFloat] =
+    new CommutativeMonoid[MiniFloat] with BoundedSemilattice[MiniFloat] {
+      override def empty: MiniFloat = MiniFloat.NegativeInfinity
+      override def combine(x: MiniFloat, y: MiniFloat): MiniFloat = if (x > y) x else y
+    }
```
tmccarthy, line 175: Assuming we define `Eq[MiniFloat]` as above, I believe this is the main lawful Monoid you can define for `MiniFloat`. I'd be happy to provide instances for addition and multiplication (with appropriate tests and disclaimers) if requested, but I'm not sure they'd be super useful.

tests/src/test/scala-2.12/cats/tests/ScalaVersionSpecific.scala
```diff
@@ -26,27 +25,20 @@ trait ScalaVersionSpecificAlgebraInvariantSuite {
...
+    def negate(x: MiniFloat): MiniFloat = -x
+    def fromInt(x: Int): MiniFloat = MiniFloat.from(x)
+    def toInt(x: MiniFloat): Int = x.toInt
+    def toLong(x: MiniFloat): Long = x.toLong
+    def toFloat(x: MiniFloat): Float = x.toFloat
+    def toDouble(x: MiniFloat): Double = x.toDouble
+    def div(x: MiniFloat, y: MiniFloat): MiniFloat = x / y
   }
+
+  protected def versionSpecificNumericEq[A: Eq: ExhaustiveCheck]: Eq[Numeric[A]] = Eq.allEqual
```
tmccarthy, line 42: There are no version-specific components of `Numeric` in 2.12, as opposed to 2.13 which added `parseString`.

tests/src/test/scala-2.12/cats/tests/ScalaVersionSpecific.scala
```diff
@@ -26,27 +25,20 @@ trait ScalaVersionSpecificAlgebraInvariantSuite {
...
-      numeric.plus _,
-      numeric.minus _,
-      numeric.times _,
-      numeric.negate _,
-      fromMiniInt,
-      numeric.toInt _,
-      numeric.toLong _,
-      numeric.toFloat _,
-      numeric.toDouble _
-    )
```
tmccarthy, line 50: Most of this code for defining an `Eq[Numeric]` is common between 2.12 and 2.13, so I've refactored to move the common stuff into the common file. Only the version-specific behaviour sits in `ScalaVersionSpecific.scala` via the `versionSpecificNumericEq` below.

tests/src/test/scala/cats/tests/AlgebraInvariantSuite.scala
```diff
@@ -158,8 +159,46 @@ class AlgebraInvariantSuite extends CatsSuite with ScalaVersionSpecificAlgebraIn
...
+        numeric.negate _,
+        fromMiniInt,
+        numeric.toInt _,
+        numeric.toLong _,
+        numeric.toFloat _,
+        numeric.toDouble _
+      )
+    }
+
+    Eq.and(versionSpecificNumericEq, versionAgnosticNumericEq)
```
tmccarthy, line 200: Open to feedback about this approach of breaking the version-specific equality checks out into a dedicated `Eq` in `ScalaVersionSpecific.scala`.

tests/src/test/scala/cats/tests/MiniFloatSuite.scala
```diff
@@ -0,0 +1,393 @@
...
+import cats.laws.discipline.MiniFloat
+import cats.laws.discipline.MiniFloat._
+import cats.laws.discipline.arbitrary._
+import cats.syntax.eq._
+import cats.tests.MiniFloatSuite.FloatArithmeticOp
+import munit.Location
+import org.scalacheck.Prop._
+import org.scalacheck.{Arbitrary, Gen}
+
+class MiniFloatSuite extends CatsSuite {
```
tmccarthy, line 20: There are a _lot_ of tests in here, which I hope is OK. I wanted to be thorough when checking the corner-case floating point behaviours.

tests/src/test/scala/cats/tests/MiniFloatSuite.scala
```diff
@@ -0,0 +1,393 @@
...
+    test(s"$mf isNan $expectIsNaN") {
+      assert(mf.isNaN === expectIsNaN)
+    }
+
+    test(s"$mf isFinite $expectIsFinite") {
+      assert(mf.isFinite === expectIsFinite)
+    }
+  }
+
+  testSpecialNumberExpectations(MiniFloat.NaN, expectIsNaN = true, expectIsFinite = false)
```
tmccarthy, line 181: This test fails in Scala Native due to https://github.com/scala-native/scala-native/issues/2178.

johnynek:
I'd personally like to separate the PR. One for the `Invariant[Fractional]` and another for MiniFloat.

I'm more skeptical to MiniFloat. I think something like Float16:

https://github.com/stripe/agate/blob/master/core/src/main/scala/com/stripe/agate/tensor/Float16.scala

or BFloat16:

https://github.com/stripe/agate/blob/master/core/src/main/scala/com/stripe/agate/tensor/BFloat16.scala

would be more useful.

It is small enough to enumerate still, but also potentially useful for low precision applications.

tmccarthy:
>I'd personally like to separate the PR. One for the `Invariant[Fractional]` and another for `MiniFloat`.

I probably could have been clearer in the issue description, but these are bundled together in this PR because `MiniFloat` is needed for the tests.

In order to define an `Eq[Fractional[A]]`, we can either:

* Use the deprecated [`catsLawsEqForFn1`](https://github.com/typelevel/cats/blob/aa6374b/laws/src/main/scala/cats/laws/discipline/Eq.scala#L139) using `Arbitrary[Float]` or somesuch, or
* Use the recommended [`catsLawsEqForFn1Exhaustive`](https://github.com/typelevel/cats/blob/aa6374b/laws/src/main/scala/cats/laws/discipline/Eq.scala#L16), for which we need a type `A` that has both `Fractional[A]` and `ExhaustiveCheck[A]`.

No type currently exists for which you can define both `Fractional` and `ExhaustiveCheck`, hence the need for a type like `MiniFloat`.

If the scope of this PR is too big I'm happy to use the deprecated `catsLawsEqForFn1` using `Arbitrary[Float]` in order to add `Invariant[Fractional]`, but my preference was to avoid the use of the deprecated function.

> I'm more skeptical to MiniFloat. I think something like Float16 or BFloat16 would be more useful.
> It is small enough to enumerate still, but also potentially useful for low precision applications.

I might be missing something, but a Float16 type has a cardinality of roughly 2¹⁶ = 65,536 (a bit less if you equate all NaNs). This seems much too big for `ExhaustiveCheck`. Certainly it's orders of magnitude more than for the other types for which cats provides a definition (2 for `Boolean`, 16 for `MiniInt`, 14 for `MiniFloat` here).

Also, my understanding was that `MiniFloat` (and `MiniInt`, which inspired it) are intended solely for use in laws testing using `ExhaustiveCheck`. I'm not sure either of them needs to have any utility for "low precision applications". Perhaps cats (or spire) might like to define a `Float16` type separately, but that would presumably in the main library and not in `discipline` like `MiniFloat`.

tmccarthy:
I'll break this into two PRs.

PR #3814, opened 2021-02-26
Update applicativemonaderror.md

wibisono:
The attemptDivideApplicativeError should divide instead of add.


larsrh (approved):
Great catch, thanks!

PR #3817, opened 2021-02-28
remove unnecessary Duration scalacheck instances

### LINKED ISSUE #3775: Remove some unneeded `Arbitrary` and `Gen` instances
Follow-up to #3772.

ScalaCheck already provides instances for `Duration` and `FiniteDuration`. We should use them and delete ours. If it turns out that the ScalaCheck provided instances don't work for us, we should fix them upstream.

Requires deleting some lines in https://github.com/typelevel/cats/blob/master/kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala.

tnielens:
Closes  #3775.


larsrh:
Thanks for the PR! Could you please rebase since there is a conflict in `LawTests.scala`?

larsrh (commented):
LGTM! Thanks.

PR #3823, opened 2021-03-09
Add sumAll & productAll

### LINKED ISSUE #3643: Add `.sum` method on Foldables using `scala.Numeric`
It would be good to have to enable summing NonEmptyLists like standard collections, similarly with other structures.

phongngtuan:
Fix #3643


larsrh (approved):
Given that algebra hasn't materialized into this repo yet, I think this addition should be fine.

LukaJCB (approved):

PR #3825, opened 2021-03-10
Use -scalajs-mapSourceURI in Scala 3 builds

fthomas:
See https://github.com/lampepfl/dotty/issues/9783#issuecomment-731718011.


larsrh (approved):
Thanks!

PR #3831, opened 2021-03-17
Add composition to Representable

TimWSpence:


TimWSpence:
Realized I forgot to do `Nested` - will push that shortly

tests/src/test/scala/cats/tests/NestedSuite.scala
```diff
@@ -264,4 +266,36 @@ class NestedSuite extends CatsSuite {
              SerializableTests.serializable(Distributive[Nested[Function1[Int, *], Function0, *]])
     )
   }
+
+  {
+    type Pair[A] = (A, A)
+
+    //Scala 2.12 implicit resolution absolutely loses its mind here
```
johnynek, line 273: do we need `implicit` if you are passing them explicitly below?

johnynek (commented):
looks good to me other than a question about if we need some `implicit` modifiers.

tests/src/test/scala/cats/tests/NestedSuite.scala
TimWSpence: Nope, good point. If we remove the modifier then we  could probably summon some of them via `implicitly` as well. Thanks, I'll fix this in both PRs 👍

johnynek (approved):

PR #3832, opened 2021-03-19
Representable instance for Tuple2K

TimWSpence:


tests/src/test/scala/cats/tests/Tuple2KSuite.scala
```diff
@@ -41,6 +42,37 @@ class Tuple2KSuite extends CatsSuite {
 
   checkAll("Show[Tuple2K[Option, Option, Int]]", SerializableTests.serializable(Show[Tuple2K[Option, Option, Int]]))
 
+  {
+    type Pair[A] = (A, A)
+
+    //Scala 2.12 implicit resolution absolutely loses its mind here
```
johnynek, line 48: do these need to be implicit if you are passing them explicitly? Can we just make them normal vals?

TimWSpence: Yep, good call! 👍

johnynek (approved):

PR #3833, opened 2021-03-20
Rename master to main, take 2

larsrh:
Based on work by @travisbrown in #3480.

Updated to latest HEAD and removed some outdated content along the way.


larsrh:
Branch rename via GitHub UI successful.

PR #3834, opened 2021-03-23
Scala-3: Use Opaque Types.

diesalbla:
Scala-3 comes with its own zero-cost Newtype implementation, as Opaque Types. This is an improvement over existing case classes, or Newtype encodings we are using. 

In this commit, we bifurcate the `IdT` wrapper in two implementations: the existing one goes to the Scala-2.x branch, a new one with newtype is used on the new.

#3835


rossabaker:
The goal is compelling, but this is 3500 extra lines of code, and maintaining parity across Scala 2 and Scala 3 will be exceedingly difficult.  I wonder if there's a less duplicative encoding of this?

diesalbla:
> The goal is compelling, but this is 3500 extra lines of code, and maintaining parity across Scala 2 and Scala 3 will be exceedingly difficult. I wonder if there's a less duplicative encoding of this?

I doubt we can avoid duplication, since we are writing in two languages. The goal is not just compellingPlus, the sooner we start using features of Dotty, the sooner we can find the minefields and bugs on the language... as I just found some.

_Update_: I have converted the `OptionT` case class into an opaque type. Of course, with opaque types all methods are extension methods. One feature of extension methods is that we can group methods that use a same type-class constraint, such as "Functor[F]" or "Monad[F]". Currently, this is crashing the Scala compiler.

PR #3836, opened 2021-03-25
RepresentableStoreT

TimWSpence:
Add the `RepresentableStoreT` (co)monad-transformer. Unfortunately I don't _think_ it's possible to unify the existing `RepresentableStore` with this (by setting `W = Id`) due to bin-compat problems. If that is not the case then please do let me know!


core/src/main/scala/cats/data/RepresentableStore.scala
```diff
@@ -3,7 +3,7 @@ package cats.data
 import cats.{Comonad, Functor, Representable}
 
 /**
- * A specialization of the `Store` comonad, for any `Representable` functor.
+ * A generalization of `StoreT`, where the underlying functor `F` hasfor any `Representable` functor.
```
ritschwumm, line 6: "hasfor" ?

larsrh (approved):

TimWSpence:
@larsrh thanks for reviewing! Is this just waiting for someone else to look at it or do I need to do anything else?

larsrh:
I was hoping to get another :+1: but I think this addition is good regardless :smile:

TimWSpence:
> I was hoping to get another 👍 but I think this addition is good regardless 😄

😂  Thanks!!

PR #3837, opened 2021-03-28
Provide instance of `Alternative[Set]` in alleycats

rtyley:
This allows the use of methods from [`Alternative`](https://typelevel.org/cats/typeclasses/alternative.html) on a `Set` - for instance, I want to be able to use `separate` on a `Set`:

```scala
import cats.implicits._
import alleycats.std.set._

val stringsAndInts: Set[Either[String, Int]] = Set(Right(6),Left("Foo"))
val (strings: Set[String], ints: Set[Int]) = stringsAndInts.separate
```

The `Alternative` typeclass just requires `MonoidK` (already provided for `Set` in `cats-core`) & `Applicative` (already provided for `Set` by [`Monad`](https://typelevel.org/cats/typeclasses/monad.html) in [`alleycats`](https://github.com/typelevel/cats/tree/main/alleycats-core#set_-instances)), so adding it to `alleycats` is a small change.

Here's a diagram I drew mainly for myself (!), as a novice contributing to Cats, to understand the typeclass relationships involved:

CatsAlternativeForSet

> This is a kind reminder to run `sbt +prePR` and commit the changed files, if any, before submitting.

I did try running this - but it crashed on unrelated files like this:

```
[info] Forcing Scala version to 3.0.0-M3 on all projects.
[info] Reapplying settings...
[info] set current project to cats (in build file:/home/roberto/development/cats/)
[info] Formatting 2 Scala sources...
[info] Formatting 2 Scala sources...
[info] Formatting 2 Scala sources...
[info] Formatting 1 Scala sources...
[info] Formatting 1 Scala sources...
[info] Formatting 1 Scala sources...
[error] /home/roberto/development/cats/free/src/main/scala-3.x/cats/free/FreeFoldStep.scala:19: error: identifier expected but [ found
[error]     onFlatMapped: [X] => (S[X], X => Free[S, A]) => B
```

### DIFF
--- a/alleycats-core/src/main/scala/alleycats/std/set.scala
+++ b/alleycats-core/src/main/scala/alleycats/std/set.scala
@@ -2,7 +2,7 @@ package alleycats
 package std
 
 import alleycats.compat.scalaVersionSpecific._
-import cats.{Always, Applicative, Eval, Foldable, Monad, Monoid, Traverse, TraverseFilter}
+import cats.{Alternative, Always, Applicative, Eval, Foldable, Monad, Monoid, Traverse, TraverseFilter}
 
 import scala.annotation.tailrec
 
@@ -10,7 +10,7 @@ object set extends SetInstances
 
 @suppressUnusedImportWarningForScalaVersionSpecific
 trait SetInstances {
-  // This method advertises parametricity, but relies on using
+  // Monad advertises parametricity, but Set relies on using
   // universal hash codes and equality, which hurts our ability to
   // rely on free theorems.
   //
@@ -30,8 +30,12 @@ trait SetInstances {
   // contain three. Since `g` is not a function (speaking strictly)
   // this would not be considered a law violation, but it still makes
   // people uncomfortable.
-  implicit val alleyCatsStdSetMonad: Monad[Set] =
-    new Monad[Set] {
+  //
+  // If we accept Monad for Set, we can also have Alternative, as
+  // Alternative only requires MonoidK (already accepted by cats-core) and
+  // the Applicative that comes from Monad.
+  implicit val alleyCatsStdSetMonad: Monad[Set] with Alternative[Set] =
+    new Monad[Set] with Alternative[Set] {
       def pure[A](a: A): Set[A] = Set(a)
       override def map[A, B](fa: Set[A])(f: A => B): Set[B] = fa.map(f)
       def flatMap[A, B](fa: Set[A])(f: A => Set[B]): Set[B] = fa.flatMap(f)
@@ -65,6 +69,10 @@ trait SetInstances {
         go(f(a))
         bldr.result()
       }
+
+      override def empty[A]: Set[A] = Set.empty
+
+      override def combineK[A](x: Set[A], y: Set[A]): Set[A] = x | y
     }
 
   // Since iteration order is not guaranteed for sets, folds and other
--- a/alleycats-tests/shared/src/test/scala/alleycats/tests/SetSuite.scala
+++ b/alleycats-tests/shared/src/test/scala/alleycats/tests/SetSuite.scala
@@ -2,18 +2,24 @@ package alleycats.tests
 
 import alleycats.laws.discipline._
 import alleycats.std.all._
-import cats.Foldable
+import cats.{Alternative, Foldable}
 import cats.instances.all._
 import cats.kernel.laws.discipline.SerializableTests
+import cats.laws.discipline.SemigroupalTests.Isomorphisms
 import cats.laws.discipline.arbitrary._
-import cats.laws.discipline.{ShortCircuitingTests, TraverseFilterTests}
+import cats.laws.discipline.{AlternativeTests, ShortCircuitingTests, TraverseFilterTests}
 
 class SetSuite extends AlleycatsSuite {
+  implicit val iso: Isomorphisms[Set] = Isomorphisms.invariant[Set](alleyCatsStdSetMonad)
+
   checkAll("FlatMapRec[Set]", FlatMapRecTests[Set].tailRecM[Int])
 
   checkAll("Foldable[Set]", SerializableTests.serializable(Foldable[Set]))
 
   checkAll("TraverseFilter[Set]", TraverseFilterTests[Set].traverseFilter[Int, Int, Int])
+
+  checkAll("Set[Int]", AlternativeTests[Set].alternative[Int, Int, Int])
+  checkAll("Alternative[Set]", SerializableTests.serializable(Alternative[Set]))
 
   checkAll("Set[Int]", ShortCircuitingTests[Set].traverseFilter[Int])
 }


alleycats-core/src/main/scala/alleycats/std/set.scala
```diff
@@ -30,8 +30,12 @@ trait SetInstances {
   // contain three. Since `g` is not a function (speaking strictly)
   // this would not be considered a law violation, but it still makes
   // people uncomfortable.
-  implicit val alleyCatsStdSetMonad: Monad[Set] =
-    new Monad[Set] {
+  //
+  // If we accept Monad for Set, we can also have Alternative, as
+  // Alternative only requires MonoidK (already accepted by cats-core) and
+  // the Applicative that comes from Monad.
+  implicit val alleycatsStdInstancesForSet: Monad[Set] with Alternative[Set] =
```
rtyley, line 37: This name follows [the Cats implicit-naming guidelines](https://typelevel.org/cats/guidelines.html#-implicit-naming), I think.

alleycats-core/src/main/scala/alleycats/std/set.scala
```diff
@@ -65,6 +69,10 @@ trait SetInstances {
         go(f(a))
         bldr.result()
       }
+
+      override def empty[A]: Set[A] = Set.empty
+
+      override def combineK[A](x: Set[A], y: Set[A]): Set[A] = x | y
```
rtyley, line 75: These are effectively re-implemented from `catsStdInstancesForSet` in `cats-core`:

https://github.com/typelevel/cats/blob/3d900dedf121223debec27207a13cbd374381235/core/src/main/scala/cats/instances/set.scala#L23-L25

...I wasn't sure if delegating to the `cats-core` typeclass instance was the thing to do, or if it was clearer just to duplicate!

rtyley:
I think this PR is finally ready for review, but I'm not sure who to ask for review! - maybe @larsrh or @kailuowang ?

alleycats-core/src/main/scala/alleycats/std/set.scala
rossabaker: I'm okay with this.  The duplication is slight and the laws ought to catch problems.

rossabaker (commented):
Thanks for this!

What happens if we try to summon a `MonoidK[Set]` with this in scope?  There's only one `Alternative[Set]`, so your test passes, but there are now two `MonoidK[Set]` instances, and I don't see a prioritization.  I am not too familiar with alleycats, so perhaps there is precedent for a workaround.  Or maybe it works, but I'd throw in a test for that to be sure.  Also maybe try it from a REPL with the standard imports, just to be super sure.

rtyley:
> There's only one `Alternative[Set]`, so your test passes, but there are now two `MonoidK[Set]` instances, and I don't see a prioritization. I am not too familiar with alleycats, so perhaps there is precedent for a workaround. Or maybe it works, but I'd throw in a test for that to be sure.

Thanks - that all makes sense, I will take a look at this a bit later 👍

rtyley:
I tried summoning a `MonoidK[Set]`, and as you anticipated, ambiguous implicit values are a problem:

```scala
scala> import cats._
     | import cats.data._
     | import cats.implicits._
     | import alleycats.std.all._
import cats._
import cats.data._
import cats.implicits._
import alleycats.std.all._

scala> implicitly[MonoidK[Set]]
                 ^
       error: ambiguous implicit values:
        both value catsStdInstancesForSet in trait SetInstances of type cats.UnorderedTraverse[Set] with cats.MonoidK[Set]
        and value alleyCatsStdSetMonad in trait SetInstances of type cats.Monad[Set] with cats.Alternative[Set]
        match expected type cats.MonoidK[Set]
```

I don't know how to get round this - if both imports (`cats.implicits._` & `alleycats.std.all._`) make an implicit instance of `MonoidK[Set]` available, I don't know how to resolve that ambiguity for the compiler. Although I've read https://typelevel.org/cats/guidelines.html#-implicit-instance-priority I can't see how to apply it to this situation, where the concrete instance in `cats-core` can't refer to anything in `alleycats`. Is any kind of disambiguation possible?!

larsrh:
I don't think we can get around that. Possibly the instance could be made non-implicit?

rossabaker:
Clients could get around it by excluding the cats one in the import, or using a la carte imports.  But, yuck.

Is this the first time we've needed to duplicate a cats instance as part of the hierarchy in an alleycats instance?  I'm surprised this hasn't come up before.

rossabaker:
Alleycats' `Traverse[Map[K, *]]` vs. Cats' `UnorderedTraverse[Map[K, *]]`?  Both are implicit.  Seems analogous to what is being done here.

larsrh:
I think the problem here is that neither implicit is more specific, whereas `Traverse` is more specific than `UnorderedTraveres`.

rossabaker:
But if someone asked for a `UnorderedTraverse[Map[K, *]]`, they'd have the same problem as a `MonoidK[Set]`, no?  And `Traverse[Map[K, *]]` is fine, like `Alternative[Set]` is fine.

larsrh:
The more specific implicit beats the less specific one every time:

```
@ implicitly[Traverse[({ type l[a] = Map[String, a] })#l]] 
res3: Traverse[Map[String, a]] = alleycats.std.MapInstances$$anon$1@124dac75

@ implicitly[UnorderedTraverse[({ type l[a] = Map[String, a] })#l]] 
res4: UnorderedTraverse[Map[String, a]] = alleycats.std.MapInstances$$anon$1@750f64fe
```

rossabaker:
I must have weekend brain, because I don't see why that should be different.  What is "more specific"?  `Alternative` is a subtype of `MonoidK` like `Traverse` is a subtype of `UnorderedTraverse`.

larsrh:
You might be right ... maybe the actual problem here is that the `with` in the implicit definition in this PR breaks the specificity rule somehow?

rossabaker:
Is it your REPL that's allowing the `Map` instances?  Nested objects as an implementation of the REPL?

joroKr21:
> You might be right ... maybe the actual problem here is that the with in the implicit definition in this PR breaks the specificity rule somehow?

Yes, that is correct - neither is subtype of the other: `cats.UnorderedTraverse[Set] with cats.MonoidK[Set]` and `cats.Monad[Set] with cats.Alternative[Set]`. In this case bundling instances together coming to bite us. If we wish to avoid ambiguity we should separate them at the cost of duplication (or factor out common implementation into private traits).

rossabaker:
`cats.implicits._` isn't necessary for standard instances anymore, and I think that's the difference:

```scala
@ import $ivy.`org.typelevel::alleycats-core:2.5.0` 
import $ivy.$                                    

@ import cats._, cats.data._, alleycats.std.all._ 
import cats._, cats.data._, alleycats.std.all._

@ implicitly[Traverse[({ type l[a] = Map[String, a] })#l]]  
res2: Traverse[Map[String, a]] = alleycats.std.MapInstances$$anon$1@5f395ce1

@ implicitly[UnorderedTraverse[({ type l[a] = Map[String, a] })#l]]  
res3: UnorderedTraverse[Map[String, a]] = alleycats.std.MapInstances$$anon$1@16f7f59f

@ import cats._, cats.data._, cats.implicits._, alleycats.std.all._ 
import cats._, cats.data._, cats.implicits._, alleycats.std.all._

@ implicitly[Traverse[({ type l[a] = Map[String, a] })#l]]  
res5: Traverse[Map[String, a]] = alleycats.std.MapInstances$$anon$1@486dd616

@ implicitly[UnorderedTraverse[({ type l[a] = Map[String, a] })#l]]  
cmd6.sc:1: ambiguous implicit values:
 both method alleycatsStdInstancesForMap in trait MapInstances of type [K]cats.Traverse[[β$0$]scala.collection.immutable.Map[K,β$0$]]
 and method catsStdInstancesForMap in trait MapInstances of type [K]cats.UnorderedTraverse[[β$0$]scala.collection.immutable.Map[K,β$0$]] with cats.FlatMap[[β$1$]scala.collection.immutable.Map[K,β$1$]] with cats.Align[[β$2$]scala.collection.immutable.Map[K,β$2$]]
 match expected type cats.UnorderedTraverse[[a]scala.collection.immutable.Map[String,a]]
val res6 = implicitly[UnorderedTraverse[({ type l[a] = Map[String, a] })#l]] 
                     ^
Compilation Failed
```

I bet this works without that import, too.

larsrh:
@rossabaker is correct, so this PR is good to go.

larsrh (approved):

PR #3851, opened 2021-04-05
Update 'Jump Start Guide' page to modern cats

chaitanyawaikar:
Updated the Jump Start Guide.

### DIFF
--- a/docs/src/main/mdoc/jump_start_guide.md
+++ b/docs/src/main/mdoc/jump_start_guide.md
@@ -20,7 +20,7 @@ Let's go through the library package-wise, looking at the syntax available in ea
 
 # Helpers for `Option` and `Either`
 
-## `import cats.syntax.option._`
+## `import cats.syntax.all._`
 
 Importing this package enables `obj.some` syntax — equivalent to `Some(obj)`.
 The only real difference is that the value is already upcast to `Option[T]` from `Some[T]`.
@@ -39,7 +39,7 @@ implicit class ToFutureSuccessful[T](obj: T) {
 then you can use the chained syntax shown below:
 
 ```scala mdoc:silent
-import cats.syntax.option._
+import cats.syntax.all._
 
 class Account { /* ... */ }
 
@@ -64,14 +64,14 @@ Chaining `.some.asFuture` at the end rather than putting it at the front also he
 Providing a more specialized type sometimes helps the Scala compiler properly infer the type of expressions containing `None`.
 
 
-## `import cats.syntax.either._`
+## `import cats.syntax.all._`
 
 `obj.asRight` is `Right(obj)`, `obj.asLeft` is `Left(obj)`.
 In both cases the type of returned value is widened from `Right` or `Left` to `Either`.
 Just as was the case with `.some`, these helpers are handy to combine with `.asFuture` to improve readability:
 
 ```scala mdoc:silent
-import cats.syntax.either._
+import cats.syntax.all._
 
 case class User(accountId: Long) { /* ... */ }
 
@@ -94,7 +94,7 @@ If the provided `option` is `Some(x)`, it becomes `Right(x)`.
 Otherwise it becomes `Left` with the provided `ifNone` value inside.
 
 
-## `import cats.syntax.apply._`
+## `import cats.syntax.all._`
 
 The `apply` package provides `(..., ..., ...).mapN` syntax, which allows for an intuitive construct for applying a function that takes more than one parameter to multiple effectful values (like futures).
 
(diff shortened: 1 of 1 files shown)


docs/src/main/mdoc/jump_start_guide.md
```diff
@@ -94,27 +94,6 @@ If the provided `option` is `Some(x)`, it becomes `Right(x)`.
...
-Usually, it's just enough to import the appropriate `cats.instances` package.
-For example, when you're doing the transformations on futures, you'll need to `import cats.instances.future._`.
-The corresponding packages for options and lists are called `cats.instances.option._` and `cats.instances.list._`.
-They provide the implicit type class instances that Cats syntax needs to work properly.
-
-As a side note, if you have trouble finding the necessary `instances` or `syntax` package, the quick workaround is to just `import cats.implicits._`.
-This is not a preferred solution, though, as it can significantly increase compile times — especially if used in many files across the project.
-It is generally considered good practice to use narrow imports to take some of the implicit resolution burden off the compiler.
-
 ## `import cats.syntax.apply._`
```
Daenyth, line 97: 🤔 We may want to change the heading here, since this was styled after a-la-carte syntax imports

docs/src/main/mdoc/jump_start_guide.md
```diff
@@ -148,9 +126,8 @@ def processAsync: Future[ProcessingResult] = {
 }
 ```
 
-As pointed out before, to provide the implicit instances (namely, [`Functor[Future]`](http://typelevel.org/cats/api/cats/Functor.html) and
-[`Semigroupal[Future]`](https://typelevel.org/cats/api/cats/Semigroupal.html)) required for `mapN` to work properly,
-you should import `cats.instances.future._`.
+By default the implicit instances (namely, [`Functor[Future]`](http://typelevel.org/cats/api/cats/Functor.html) and
+[`Semigroupal[Future]`](https://typelevel.org/cats/api/cats/Semigroupal.html)) required for `mapN` to work properly are automatically imported. These are present in the respective companion objects of the instances and hence we do not need to import them explicitly.
```
Daenyth, line 130:
```suggestion
[`Semigroupal[Future]`](https://typelevel.org/cats/api/cats/Semigroupal.html)) required for `mapN` to work properly are always visible. They are present in the respective companion objects of the instances and hence we do not need to import them explicitly.  
```

docs/src/main/mdoc/jump_start_guide.md
```diff
@@ -192,8 +169,6 @@ If you call `traverse` instead of `map`, like `obj.traverse(fun)`, you'll get `G
 
 ```scala mdoc:silent
 import cats.syntax.traverse._
```
Daenyth, line 171: I think we probably want to update all these examples to use the `all` syntax import

docs/src/main/mdoc/jump_start_guide.md
```diff
@@ -453,7 +422,7 @@ If you're instead looking for validation that accumulates the errors (e.g. when
 
 # Common issues
 
-If anything doesn't compile as expected, first make sure all the required Cats implicits are in the scope — just try importing ```cats.implicits._``` and see if the problem persists.
+The Cats type class instances for standard library types are available in implicit scope and hence no longer have to be imported. If anything doesn't compile as expected, first make sure all the required Cats implicits are in the scope — just try importing ```cats.syntax.all._``` and see if the problem persists. The only exception is here is Cat's own ```Order``` and ```PartialOrder``` type classes which are available by importing ```cats.implicits._```.
```
Daenyth, line 425:
```suggestion
The Cats type class instances for standard library types are available in implicit scope and hence no longer have to be imported. If anything doesn't compile as expected, first make sure all the required Cats syntax implicits are in the scope — try importing ```cats.syntax.all._``` and see if the problem persists. The only exception is here is Cat's own ```Order``` and ```PartialOrder``` type classes which are available by importing ```cats.implicits._```. 
```

Daenyth (approved):
I think we can keep iterating on this, but the changes here are already better than what's on the website today, so it's worth just getting them out there

docs/src/main/mdoc/jump_start_guide.md
chaitanyawaikar: Changed to ```import cats.syntax.all._```

docs/src/main/mdoc/jump_start_guide.md
chaitanyawaikar: Incorporated these changes

larsrh (approved):
LGTM, thanks!

PR #3852, opened 2021-04-06
fix show interpolation ambiguity with Seq/List

### LINKED ISSUE #3720: Ambiguous implicit values for derived `Show` in 2.3.x
After instances for `Seq` was introduced in 2.3.0 code such
```scala
object test {

    import cats.Show
    import cats.syntax.all._

    final case class Foo(bar: String)

    implicit val showFoo = Show.fromToString[Foo]

    val foo = Foo("bar")

    show"$foo"

    val listFoo: List[Foo] = List(foo)

    show"$listFoo"
}
```
started to fail with, a rather unfriendly, error:
```
[error]  found   : List[test.Foo]
[error]  required: cats.Show.Shown
[error]     show"$listFoo"
```

The `Shown` derivation gives better error to understand what happened:
```scala
val shown  = cats.Show.Shown.mat(listFoo)
```
```
[error]  both method catsShowForList in object Show of type [A](implicit evidence$3: cats.Show[A]): cats.Show[List[A]]
[error]  and method catsShowForSeq in object Show of type [A](implicit evidence$4: cats.Show[A]): cats.Show[Seq[A]]
[error]  match expected type cats.Show.ContravariantShow[List[test.Foo]]
[error]     val shown: Show.Shown = cats.Show.Shown.mat(listFoo)
```

I suppose that this is because we need `ContravatiantShow[-T]` and you can easily fix the problem with importing the `List` instances. But it would be nice to either fix it somehow - maybe prioritise the `Seq` instances lower - or just drop a note in the release notes that you need to add `import cats.instances.list._`.

tnielens:
Fixes #3720 .

### DIFF
--- a/core/src/main/scala/cats/Show.scala
+++ b/core/src/main/scala/cats/Show.scala
@@ -89,7 +89,6 @@ object Show extends ScalaVersionSpecificShowInstances with ShowInstances {
   implicit def catsShowForOption[A: Show]: Show[Option[A]] = cats.instances.option.catsStdShowForOption[A]
   implicit def catsShowForTry[A: Show]: Show[Try[A]] = cats.instances.try_.catsStdShowForTry[A]
   implicit def catsShowForList[A: Show]: Show[List[A]] = cats.instances.list.catsStdShowForList[A]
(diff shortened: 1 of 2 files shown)


rossabaker (approved):

LukaJCB (approved):
awesome thanks!

PR #3856, opened 2021-04-08
Remove redundant Monad constraints from Parallel syntax

### LINKED ISSUE #2254: Syntax extensions for Parallel require Monad[M], but implementation doesn't
I discovered today that to use syntax extension for `parTraverse`/`parSequence`, a `Monad[M]` instance is required.

```
final class ParallelTraversableOps[T[_], A](val ta: T[A]) extends AnyVal {

  def parTraverse[M[_]: Monad, F[_], B]
  (f: A => M[B])(implicit T: Traverse[T], P: Parallel[M, F]): M[T[B]] =
    Parallel.parTraverse(ta)(f)

}
```
While `Parallel.parTraverse` is defined as follows (notice lack of `Monad[M]`)
```scala
  def parTraverse[T[_]: Traverse, M[_], F[_], A, B]
  (ta: T[A])(f: A => M[B])(implicit P: Parallel[M, F]): M[T[B]] = {
    val gtb: F[T[B]] = Traverse[T].traverse(ta)(f andThen P.parallel.apply)(P.applicative)
    P.sequential(gtb)
  }
```
Same is true for `parSequence`. Is that an oversight?

joroKr21:
Binary compatible version of #2180
Fixes #2254

### DIFF
--- a/core/src/main/scala/cats/syntax/parallel.scala
+++ b/core/src/main/scala/cats/syntax/parallel.scala
@@ -15,102 +15,124 @@ import cats.{
 
 trait ParallelSyntax extends TupleParallelSyntax {
 
-  implicit final def catsSyntaxParallelTraverse[T[_]: Traverse, A](ta: T[A]): ParallelTraversableOps[T, A] =
-    new ParallelTraversableOps[T, A](ta)
+  @deprecated("Kept for binary compatibility", "2.6.0")
+  final def catsSyntaxParallelTraverse[T[_]: Traverse, A](ta: T[A]): ParallelTraversableOps[T, A] =
+    new ParallelTraversableOps(ta)
 
-  implicit final def catsSyntaxParallelSequence[T[_]: Traverse, M[_]: Monad, A](
-    tma: T[M[A]]
-  ): ParallelSequenceOps[T, M, A] = new ParallelSequenceOps[T, M, A](tma)
+  implicit final def catsSyntaxParallelTraverse1[T[_]: Traverse, A](ta: T[A]): ParallelTraversableOps1[T, A] =
+    new ParallelTraversableOps1(ta)
+
+  @deprecated("Kept for binary compatibility", "2.6.0")
+  final def catsSyntaxParallelSequence[T[_]: Traverse, M[_]: Monad, A](tma: T[M[A]]): ParallelSequenceOps[T, M, A] =
+    new ParallelSequenceOps(tma)
+
+  implicit final def catsSyntaxParallelSequence1[T[_]: Traverse, M[_], A](tma: T[M[A]]): ParallelSequenceOps1[T, M, A] =
+    new ParallelSequenceOps1(tma)
 
   implicit final def catsSyntaxParallelAp[M[_]: FlatMap, A](ma: M[A]): ParallelApOps[M, A] =
-    new ParallelApOps[M, A](ma)
-
+    new ParallelApOps(ma)
 }
 
 trait ParallelApplySyntax {
   implicit final def catsSyntaxParallelApply[F[_], A, B](fa: F[A => B]): ParallelApplyOps[F, A, B] =
-    new ParallelApplyOps[F, A, B](fa)
+    new ParallelApplyOps(fa)
 }
 
 trait ParallelFlatSyntax {
-  implicit final def catsSyntaxParallelFlatTraverse[T[_]: Traverse: FlatMap, A](
-    ta: T[A]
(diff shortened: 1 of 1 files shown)


core/src/main/scala/cats/syntax/parallel.scala
```diff
@@ -18,9 +18,16 @@ trait ParallelSyntax extends TupleParallelSyntax {
   implicit final def catsSyntaxParallelTraverse[T[_]: Traverse, A](ta: T[A]): ParallelTraversableOps[T, A] =
     new ParallelTraversableOps[T, A](ta)
 
-  implicit final def catsSyntaxParallelSequence[T[_]: Traverse, M[_]: Monad, A](
-    tma: T[M[A]]
-  ): ParallelSequenceOps[T, M, A] = new ParallelSequenceOps[T, M, A](tma)
+  @deprecated("Kept for binary compatibility", "2.5.1")
```
joroKr21, line 21: Not sure if this should be 2.5.1 or 2.6.0

joroKr21:
Since this is not working, the other option is to create new `AnyVal`s - should I go for it?

tnielens:
Would something along this line be binary compatible? 
```scala 
final class ParallelTraversableOps[T[_], A](private val ta: T[A]) extends AnyVal {
  def parTraverse[M[_]: Monad, B](f: A => M[B])(implicit T: Traverse[T], P: Parallel[M]): M[T[B]] =
    Parallel.parTraverse(ta)(f)
}

final class ParallelTraversableOps2[T[_], A](private val ta: T[A]) extends AnyVal {
  def parTraverse[M[_], B](f: A => M[B])(implicit T: Traverse[T], P: Parallel[M]): M[T[B]] =
    Parallel.parTraverse(ta)(f)
}
```

joroKr21:
> Would something along this line be binary compatible?

Yes, that's what I meant. Unfortunately adding an overload in Scala 2.12 changes the name mangling of extension methods.
There is also an opportunity to consolidate extension methods (if we drop the constraints from the implicit conversions).

core/src/main/scala/cats/syntax/parallel.scala
```diff
@@ -108,9 +123,11 @@ trait ParallelFoldMapASyntax {
 }
 
 final class ParallelTraversableOps[T[_], A](private val ta: T[A]) extends AnyVal {
-  def parTraverse[M[_]: Monad, B](f: A => M[B])(implicit T: Traverse[T], P: Parallel[M]): M[T[B]] =
+  @deprecated("Kept for binary compatibility", "2.5.1")
+  def parTraverse[M[_], B](f: A => M[B], M: Monad[M], T: Traverse[T], P: Parallel[M]): M[T[B]] =
```
tnielens, line 127: shouldn't this have a second parameter list instead?

joroKr21: WDYM? 🤔

tnielens: You moved the monad constraint in the first position of the implicit parameter list. My understanding of context bounds is that a new parameter list is appended to the method definition in this way:
```scala
   def parTraverse[M[_]: Monad, B](f: A => M[B])(implicit T: Traverse[T], P: Parallel[M]): M[T[B]] = 
   // is equivalent to
    def parTraverse[M[_], B](f: A => M[B])(implicit T: Traverse[T], P: Parallel[M])(implicit M: Monad[M]): M[T[B]] = 
 ```

Comment (joroKr21) [reply core/src/main/scala/cats/syntax/parallel.scala]: Nope, in Scala 2 you can have only one implicit list and the context bounds come first.

Comment (tnielens) [reply core/src/main/scala/cats/syntax/parallel.scala]: Didn't know that, thanks for the explanation.

Comment (joroKr21) [reply core/src/main/scala/cats/syntax/parallel.scala]: But in the bytecode there is only one parameter list anyway (and no implicits) which is why both are binary compatible:
```scala
def parTraverse[M[_]: Monad, B](f: A => M[B])(implicit T: Traverse[T], P: Parallel[M]): M[T[B]]
def parTraverse[M[_], B](f: A => M[B], M: Monad[M], T: Traverse[T], P: Parallel[M]): M[T[B]]
```

That's a neat trick for library authors 😄

larsrh (dismissed):
LGTM, but I think this warrants a bump to 2.6.0

larsrh (approved):

diesalbla (approved):

PR #3858, opened 2021-04-12
Representable instance for Cofree

TimWSpence:

### DIFF
--- a/free/src/main/scala/cats/free/Cofree.scala
+++ b/free/src/main/scala/cats/free/Cofree.scala
@@ -1,6 +1,8 @@
 package cats
 package free
 
+import cats.Representable
+
 /**
  * A free comonad for some branching functor `S`. Branching is done lazily using [[Eval]].
  * A tree with data at the branches, as opposed to [[Free]] which is a tree with data at the leaves.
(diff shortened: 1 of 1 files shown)


larsrh:
Can we add a test for this?

TimWSpence:
Thanks @larsrh! Sorry, I've been meaning to leave a comment on this for ages. I tried to start a discussion on this  (https://gitter.im/typelevel/cats-dev?at=607434731f84d71853b1223f) but nobody took the bait 😭

armanbilge:
Copy-pasta from the linked gitter.
> I opened a PR for adding a Representable instance for Cofree - typelevel/cats#3858. I’m pretty sure the implementation is sound but laws testing it is an interesting challenge. The issue is that sum types are not representable (as far as I know and it certainly seems intuitively obvious) and neither are trivial product types like Const. But if we have a genuine product type F[A] with a hole of type A then Cofree[F, A] is necessarily infinite so we can’t construct an Eq instance for checking the laws. My only thought is that we construct an Eq instance that evaluates the tree to some fixed depth and asserts that the sub-trees are equal?
> 
> I think the instance is still useful because it seems reasonable for an application to generate this infinite tree structure but only index some finite subtree

armanbilge:
@TimWSpence if law-testing this is challenging, can we add some more vanilla tests or property checks instead?

PR #3863, opened 2021-04-21
Apply Scalafix rule to migrate to sbt slash syntax

joroKr21:
https://www.scala-sbt.org/1.x/docs/Migrating-from-sbt-013x.html#Migrating+to+slash+syntax

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,10 +1,7 @@
-import microsites._
-import ReleaseTransformations._
-import sbt.io.Using
 import com.jsuereth.sbtpgp.PgpKeys
-
-import scala.xml.transform.{RewriteRule, RuleTransformer}
+import microsites._
 import sbtcrossproject.CrossPlugin.autoImport.{crossProject, CrossType}
+import sbtrelease.ReleasePlugin.autoImport.ReleaseTransformations._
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -13,8 +10,8 @@ lazy val publishLocalSignedIfRelevant =
   taskKey[Unit]("Runs publishLocalSigned but only if scalaVersion in crossScalaVersions")
 Global / publishLocalSignedIfRelevant := PgpKeys.publishLocalSigned.value
 
-organization in ThisBuild := "org.typelevel"
-scalafixDependencies in ThisBuild += "org.typelevel" %% "simulacrum-scalafix" % "0.5.3"
+(ThisBuild / organization) := "org.typelevel"
```
larsrh, line 13: I think all those parens are unneeded, right?

joroKr21: Hmm good point, I just ran the scalafix rule - somehow I didn't notice the parens

rossabaker (approved):

larsrh (approved):

larsrh:
Thanks for doing the unglamorous work :wink:

PR #3864, opened 2021-04-22
Fix GroupFunctions inheritance

armanbilge:
Just a small typo, as far as I can tell.


rossabaker (approved):
Variance still confuses me after a decade, but this looks right and I got a second opinion on cats-dev and CI likes it, so, :+1:.

armanbilge:
If it helps support this change, an identical pattern is used for all the other `Functions` traits e.g. see [MonoidFunctions](https://github.com/typelevel/cats/blob/main/kernel/src/main/scala/cats/kernel/Monoid.scala#L98), [SemilatticeFunctions](https://github.com/typelevel/cats/blob/main/kernel/src/main/scala/cats/kernel/Semilattice.scala#L56), etc.

PR #3867, opened 2021-04-25
remove sbt-dotty plugin

xuwei-k:
https://github.com/sbt/sbt/releases/tag/v1.5.0

> sbt 1.5.0 adds built-in Scala 3 support


larsrh (approved):
Thanks, very timely!

Issue #3870, opened 2021-04-26
Use helper constructors to instantiate type classes 


joroKr21:
Since we don't support Scala 2.11 anymore this is an option.
See https://www.scala-lang.org/news/2.12.0/

> Scala and Java 8 interop is also improved for functional code, as methods that take functions can easily be called in both directions using lambda syntax. The FunctionN classes in Scala’s standard library are now Single Abstract Method (SAM) types, and all SAM types are treated uniformly – from type checking through code generation. No class file is generated for a lambda; invokedynamic is used instead.

"No class file is generated" sounds pretty good to me 🤓

joroKr21:
Unfortunately this is mostly useful only for Scala 3

joroKr21:
I looked for a scalafix rule but it doesn't exist yet.

rossabaker:
What makes this mostly useful only for Scala 3?

joroKr21:
> What makes this mostly useful only for Scala 3?

There is a problem in Scala 2 for SAM traits with method implementations (i.e. having concrete methods that derive from the single abstract one which all type classes in Cats do):

 * Methods implemented in traits that can be translated to interfaces become default methods in the byte code. But the compiler also uses a method called `isInterfaceMember` to determine if a trait needs an initializer. The initializer is normally used to initialize trait vals and vars and also to run any statements in the body of the trait. But `isInterfaceMember` has a but and it says that a non-abstract method is not a legal interface member definition (obviously false). In this case an empty initializer is generated.
 * Later in the "delambdafy" phase which translates anonymous functions to methods the compiler also takes care of eliminating generated classes for SAM expressions. But it must be careful to preserve semantics (i.e. if there are vals / vars / statements in the trait we cannot avoid instantiating a class). It does that by checking for the existence of an initializer (it must also work under separate compilation). Due to the bug in `isInterfaceMember` we now have an empty initializer that tricks us into believing this trait is not a pure interface.
 * This bug with generating unnecessary initializers is fixed in Scala 3 but can't be fixed in Scala 2 due to binary compatibility guarantees.

Anyway long story short, I found it's anyway good to use `instance` constructors because they can also work for type classes with multiple abstract methods. See the linked PR: #3871

diesalbla:
@joroKr21 From the last message, can we understand that you would like to withdraw this proposal? Is this something that should be reconsidered once we deprecate Scala 2.13?

joroKr21:
[...]

PR #3871, opened 2021-04-26
Optimise boilerplate generators, use instance constructors

joroKr21:
Make it a bit more readable by formatting constraints on a new line.

See #3870

This change results in around 40% jar size reduction on average of cats-kernel 😲

Artifact | This PR | main | Reduction
-- | -- | -- | --
cats-kernel_2.12.jar | 3068152 | 5034406 | 39.0%
cats-kernel_2.13.jar | 3299524 | 5262009 | 37.3%
cats-kernel_3.0.0-RC3.jar | 1418767 | 2491382 | 43.0%
cats-kernel_sjs1_2.12.jar | 6336434 | 10614940 | 40.3%
cats-kernel_sjs1_2.13.jar | 6883734 | 10614940 | 35.1%
cats-kernel_sjs1_3.0.0-RC3.jar | 2197991 | 4255929 | 48.3%
cats-kernel_native0.4_2.12.jar | 6330092 | 10796161 | 41.4%
cats-kernel_native0.4_2.13.jar | 6862749 | 11322497 | 39.4%

I've also give cats-core a similar treatment, but the savings will be much smaller there.


joroKr21:
Hmm I still see classes generated - is that expected?

joroKr21:
TLDR - it's good for Scala 3, Scala 2 is deceiving itself into generating a class file 😭

joroKr21:
As an alternative - use `instance` constructors that take functions. The externally provided functions will be delambdafied in both Scala 2 and Scala 3.

joroKr21:
This change results in around 40% jar size reduction on average 😲 

| Artifact                       | This PR | main | Reduction |
|--------------------------------|---------|----------|-------|
| cats-kernel_2.12.jar           | 3068152 | 5034406  | 39.0% |
| cats-kernel_2.13.jar           | 3299524 | 5262009  | 37.3% |
| cats-kernel_3.0.0-RC3.jar      | 1418767 | 2491382  | 43.0% |
| cats-kernel_sjs1_2.12.jar      | 6336434 | 10614940 | 40.3% |
| cats-kernel_sjs1_2.13.jar      | 6883734 | 10614940 | 35.1% |
| cats-kernel_sjs1_3.0.0-RC3.jar | 2197991 | 4255929  | 48.3% |
| cats-kernel_native0.4_2.12.jar | 6330092 | 10796161 | 41.4% |
| cats-kernel_native0.4_2.13.jar | 6862749 | 11322497 | 39.4% |

johnynek (commented):
I think this is worth doing but I do wonder if we can make it a bit more principled by using InvariantFunctor, ContravariantCartesian, or similar typeclasses on these kernel typeclasses (typeclasses can have typeclasses too!)

kernel/src/main/scala/cats/kernel/CommutativeGroup.scala
```diff
@@ -14,4 +14,14 @@ object CommutativeGroup extends GroupFunctions[CommutativeGroup] {
    * Access an implicit `CommutativeGroup[A]`.
    */
   @inline final def apply[A](implicit ev: CommutativeGroup[A]): CommutativeGroup[A] = ev
+
+  /**
+   * Create a `CommutativeGroup` instance from the given inverse and combine functions and empty value.
+   */
+  @inline def instance[A](emp: A, inv: A => A, cmb: (A, A) => A): CommutativeGroup[A] =
```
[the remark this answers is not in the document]
joroKr21: You probably mean `InvariantSemigroupal` so that we can use `tupledN` right? That's doable but it would more allocations and reduced performance at _runtime_ (going from `(a, b, c, d)` to `(a, (b, (c, d)))` and back) so I'm not sure that would be acceptable for `cats-kernel` which is also used by `albra`.

rossabaker (dismissed):
The build failure could be related to #3974.  I'll bring it up there.

johnynek (dismissed):

rossabaker (dismissed):

armanbilge:
Personally, I'm not convinced we need or even want this change. Apologies if I'm missing the point. 

- In https://github.com/typelevel/cats/pull/3877#discussion_r642021197 @joroKr21 mentioned that classes are loaded/unloaded dynamically.
- Scala.js and Native having a static linking step that removes all unused code from the final generated JS/binary.
- I understand the relative numbers are significant, but in absolute terms we're basically talking megabytes, right?

Additionally, isn't this de-optimizing the currently implementations? By replacing dedicated classes with lambdas and thus adding a level of indirection and a larger per-instance footprint. Like, these are micro-optimizations, but isn't saving a few megabytes? :)

joroKr21:
> In Move typelevel/algebra into cats repo #3877 (comment) @joroKr21 mentioned that classes are loaded/unloaded dynamically.

Yes, that's mostly true on the JVM these days unless you are using the CMS GC without class unloading enabled.

> Scala.js and Native involve a static linking step that removes all unused code from the final generated JS/binary.

Oh that's cool, I didn't think much about that. Do you know how fine-grained that is? Does it mean that the concerns for including `java.time` instances are unfounded (e.g. #3910)?

> I understand the relative numbers are significant, but in absolute terms we're basically talking megabytes, right?

Well yes, that's megabytes per download per jar. I don't know what's the multiplier, but there is some multiplier. I would prefer to fix that in the compiler to be honest but we can't because of binary compatibility constraints. (Aside: at least that's what I've been told. Now that I think about it - those classes should be private anyway so why not? 🤔 Maybe I will give it a try but let's not discuss that here).

> Additionally, isn't this de-optimizing the current implementations? By replacing dedicated classes with lambdas and thus adding a level of indirection and a larger per-instance footprint. Like, these are micro-optimizations, but isn't saving a few megabytes? :)

That's a good question - do we have any benchmarks on that? I can't say without trying.

joroKr21:
Some shower thoughts - the biggest cost of these instances are probably boxing of primitives (would be interesting to check in which cases it occurs) and tuple allocations. Lambda calls should not be that expensive because otherwise we would never get SAMs on the JVM. That leaves open the question about the cost of the additional indirection (which is necessary because of the compiler bug).

armanbilge:
Thanks for the detailed response! Much appreciated :)

> Do you know how fine-grained that is? Does it mean that the concerns for including java.time instances are unfounded (e.g. #3910)?

I would say very fine-grained, if done right. [This comment](https://github.com/scala-js/scala-js/blob/master/javalanglib/src/main/scala/java/lang/System.scala#L24-L35) in Scala.js is the best explanation I've found of how this works. I'll respond to your question about #3910 over there.

> Well yes, that's megabytes per download per jar.

Ok, thanks, that clarifies things. So the concern here is (also) about download size? In general, or are there specific situations where this is especially a problem? Without knowing anything, I would think anywhere cats is being downloaded probably some other (bigger) things are being downloaded too, that will make a few megabytes saved in cats all the more trivial.

> I would prefer to fix that in the compiler to be honest but we can't because of binary compatibility constraints.

> That leaves open the question about the cost of the additional indirection (which is necessary because of the compiler bug).

Sorry, I'm certain you've explained this before, but what exactly is this problem?

joroKr21:
> Sorry, I'm certain you've explained this before, but what exactly is this problem?

It's here:
https://github.com/scala/scala/blob/2.13.x/src/compiler/scala/tools/nsc/transform/Erasure.scala#L1418-L1421

Now I remember what is the problem - the bug is that traits with method implementations get an empty initializer (totally unnecessary) which triggers this check. We can't fix it in a binary compatible way because older versions of Scala would try to call this initializer and it wouldn't exist in the fixed version.

> Ok, thanks, that clarifies things. So the concern here is (also) about download size? In general, or are there specific situations where this is especially a problem?

I don't think there is any concern per se. It's just that if we can reduce the jar size by so much, then why not?

armanbilge:
Thanks for that reference. So IIUC, the traits are implemented as mix-ins rather than vanilla interfaces, causing an additional indirection?

> I don't think there is any concern per se. It's just that if we can reduce the jar size by so much, then why not?

👍 fair enough. Personally I'm of the opinion that few megabytes is not so much at all and I would rather avoid compromising runtime performance. OTOH, I don't really believe either of these micro-optimizations (saving a few mbs vs dedicated classes for these instances) will make much difference in practice. So, anything goes :)

joroKr21:
> Thanks for that reference. So IIUC, the traits are implemented as mix-ins rather than vanilla interfaces, causing an additional indirection?

Yes - traits with method implementations, which become `default` methods in the byte code - so there is actually no reason why they couldn't be vanilla interfaces if it weren't for that bug.

> Personally I'm of the opinion that few megabytes is not so much at all and I would rather avoid compromising runtime performance. OTOH, I don't really believe either of these micro-optimizations (saving a few mbs vs dedicated classes for these instances) will make much difference in practice. So, anything goes :)

I don't think any of this code was written with runtime performance in mind 😜 - but it's a valid concern to not introduce regressions. I see there are some benchmarks in the project but nothing related to the boilerplate. I might give it a try but it won't be soon.

satorg:
Just noticed that this PR has got all the approvals. But the PR looks a bit stale now (although supposed to be a nice improvement).
Are there plans to have it merged?
cc @joroKr21

joroKr21:
> Just curious – are there plans to have it merged?

I think it can be merged but people have asked about benchmarks and I have no time to work on that.

joroKr21:
There's been a lot of changes here, what happened? I have massive conflicts.

armanbilge:
> There's been a lot of changes here, what happened? I have massive conflicts.

Sorry 😬 I think it got caught up in the https://github.com/typelevel/cats/pull/4116 drama.

> I think it can be merged but people have asked about benchmarks and I have no time to work on that.

Well, by "people" I think it was only me ;) if the consensus is 👍 I say go ahead. I was really just pointing out that:
1. We are not reducing jar size "for free": there is a tradeoff, due to increased indirection and memory pressure at runtime (partially due to a scalac bug).
2. Even though the jar sizes are reduced by 40% which is substantial relatively-speaking, in practice we are only talking about at most 3 mb per jar. This seems most (only) important for download sizes, but I bet anyone repeatedly downloading Cats artifacts is also repeatedly downloading far bigger stuff, which probably makes a few megabyte savings less significant.

Of course my assumptions could be wrong about those, and at the end of the day I think that prioritizing performance vs jar size are both micro-optimizations anyway. So if folks are happy I say we can merge :)

joroKr21:
Yeah I agree about the indirection and ideally we would have some benchmarks. So if anyone wants to take over it's up for grabs. I don't think there's any more memory pressure though. Except for boxing of course which would incur the most significant cost anyway if we deal with primitives.

armanbilge:
> I don't think there's any more memory pressure though.

Maybe I'm confused about how it works, but IIUC we're replacing a specialized class with references to `N` delegates (for a tuple of size `N`) with a non-specialized class with `M` references to lambdas (implementing the `M` methods of the typeclass) and each of those lambdas contains `N` references to the delegates. So the footprint is on the order of `N * M`.

joroKr21:
Well I botched the merge 😞

joroKr21:
> Maybe I'm confused about how it works, but IIUC we're replacing a specialized class with references to N delegates (for a tuple of size N) with a non-specialized class with M references to lambdas (implementing the M methods of the typeclass) and each of those lambdas contains N references to the delegates. So the footprint is now N * M.

We are replacing classes with static methods and `LambdaMetafactory` - delegates should translate to local variables. Check the bytecode for `Monoid` of `Tuple2`:

```
  // access flags 0x1
  // signature <A0:Ljava/lang/Object;A1:Ljava/lang/Object;>(Lcats/kernel/Monoid<TA0;>;Lcats/kernel/Monoid<TA1;>;)Lcats/kernel/Monoid<Lscala/Tuple2<TA0;TA1;>;>;
  // declaration: cats.kernel.Monoid<scala.Tuple2<A0, A1>> catsKernelMonoidForTuple2<A0, A1>(cats.kernel.Monoid<A0>, cats.kernel.Monoid<A1>)
  public default catsKernelMonoidForTuple2(Lcats/kernel/Monoid;Lcats/kernel/Monoid;)Lcats/kernel/Monoid;
    // parameter final  A0
    // parameter final  A1
   L0
    LINENUMBER 2574 L0
    GETSTATIC cats/kernel/Monoid$.MODULE$ : Lcats/kernel/Monoid$;
   L1
    LINENUMBER 2575 L1
    NEW scala/Tuple2
    DUP
    ALOAD 1
    INVOKEINTERFACE cats/kernel/Monoid.empty ()Ljava/lang/Object; (itf)
    ALOAD 2
    INVOKEINTERFACE cats/kernel/Monoid.empty ()Ljava/lang/Object; (itf)
    INVOKESPECIAL scala/Tuple2.<init> (Ljava/lang/Object;Ljava/lang/Object;)V
   L2
    LINENUMBER 2576 L2
    ALOAD 1
    ALOAD 2
    INVOKEDYNAMIC apply(Lcats/kernel/Monoid;Lcats/kernel/Monoid;)Lscala/Function2; [
      // handle kind 0x6 : INVOKESTATIC
      java/lang/invoke/LambdaMetafactory.altMetafactory(Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/invoke/MethodType;[Ljava/lang/Object;)Ljava/lang/invoke/CallSite;
      // arguments:
      (Ljava/lang/Object;Ljava/lang/Object;)Ljava/lang/Object;, 
      // handle kind 0x6 : INVOKESTATIC
      cats/kernel/instances/TupleMonoidInstances.$anonfun$catsKernelMonoidForTuple2$1(Lcats/kernel/Monoid;Lcats/kernel/Monoid;Lscala/Tuple2;Lscala/Tuple2;)Lscala/Tuple2; itf, 
      (Lscala/Tuple2;Lscala/Tuple2;)Lscala/Tuple2;, 
      5, 
      1, 
      (Lscala/Tuple2;Lscala/Tuple2;)Lscala/Tuple2;
    ]
    INVOKEVIRTUAL cats/kernel/Monoid$.instance (Ljava/lang/Object;Lscala/Function2;)Lcats/kernel/Monoid;
    ARETURN
   L3
    LOCALVARIABLE this Lcats/kernel/instances/TupleMonoidInstances; L0 L3 0
    LOCALVARIABLE A0 Lcats/kernel/Monoid; L0 L3 1
    LOCALVARIABLE A1 Lcats/kernel/Monoid; L0 L3 2
    MAXSTACK = 5
    MAXLOCALS = 3
```

```
// access flags 0x1009
  public static synthetic $anonfun$catsKernelMonoidForTuple2$1(Lcats/kernel/Monoid;Lcats/kernel/Monoid;Lscala/Tuple2;Lscala/Tuple2;)Lscala/Tuple2;
    // parameter final  A0$178
    // parameter final  A1$169
    // parameter final  x
    // parameter final  y
   L0
    LINENUMBER 2576 L0
    NEW scala/Tuple2
    DUP
    ALOAD 0
    ALOAD 2
    INVOKEVIRTUAL scala/Tuple2._1 ()Ljava/lang/Object;
    ALOAD 3
    INVOKEVIRTUAL scala/Tuple2._1 ()Ljava/lang/Object;
    INVOKEINTERFACE cats/kernel/Monoid.combine (Ljava/lang/Object;Ljava/lang/Object;)Ljava/lang/Object; (itf)
    ALOAD 1
    ALOAD 2
    INVOKEVIRTUAL scala/Tuple2._2 ()Ljava/lang/Object;
    ALOAD 3
    INVOKEVIRTUAL scala/Tuple2._2 ()Ljava/lang/Object;
    INVOKEINTERFACE cats/kernel/Monoid.combine (Ljava/lang/Object;Ljava/lang/Object;)Ljava/lang/Object; (itf)
    INVOKESPECIAL scala/Tuple2.<init> (Ljava/lang/Object;Ljava/lang/Object;)V
    ARETURN
   L1
    LOCALVARIABLE A0$178 Lcats/kernel/Monoid; L0 L1 0
    LOCALVARIABLE A1$169 Lcats/kernel/Monoid; L0 L1 1
    LOCALVARIABLE x Lscala/Tuple2; L0 L1 2
    LOCALVARIABLE y Lscala/Tuple2; L0 L1 3
    MAXSTACK = 6
    MAXLOCALS = 4
```

```
 // access flags 0x1
  // signature <A:Ljava/lang/Object;>(TA;Lscala/Function2<TA;TA;TA;>;)Lcats/kernel/Monoid<TA;>;
  // declaration: cats.kernel.Monoid<A> instance<A>(A, scala.Function2<A, A, A>)
  public instance(Ljava/lang/Object;Lscala/Function2;)Lcats/kernel/Monoid;
    // parameter final  emptyValue
    // parameter final  cmb
   L0
    LINENUMBER 120 L0
    NEW cats/kernel/Monoid$$anon$2
    DUP
    ALOAD 1
    ALOAD 2
    INVOKESPECIAL cats/kernel/Monoid$$anon$2.<init> (Ljava/lang/Object;Lscala/Function2;)V
    ARETURN
   L1
    LOCALVARIABLE this Lcats/kernel/Monoid$; L0 L1 0
    LOCALVARIABLE emptyValue Ljava/lang/Object; L0 L1 1
    LOCALVARIABLE cmb Lscala/Function2; L0 L1 2
    MAXSTACK = 4
    MAXLOCALS = 3
```

armanbilge:
Sure, for `Monoid` it is effectively the same, because it only has `cmb`. But what about for `Group`? Are the `inv` and `cmb` methods really able to share the references to the delegates? I assumed they would get duplicated between the two lambdas.

joroKr21:
> Sure, for Monoid it is effectively the same, because it only has cmb (empty is a pre-computed constant). But what about for Group? Are the inv and cmb methods really able to share the references to the delegates? I assumed they would get duplicated between the two lambdas.

I'm not sure what you mean by being shared across lambdas. I don't know details about how `LambdaMetafactory` works under the hood. Are you saying that these closures are too expensive for the GC?

armanbilge:
I don't claim to have any knowledge of how these things work. But my intuition says that if I define:
```scala
val cmb = (x, y) => (A1.cmb(x._1, y._1), A2.cmb(x._2, y._2))
```
Then that `cmb` lambda must be storing references to `A1` and `A2` somewhere.

Then, if I define:
```scala
val inv = x => (A1.inv(x._1), A2.inv(x._2))
```
This `inv` lambda must also be storing references to `A1` and `A2`.

So now, we've stored these same references twice.

However, if I define:
```scala
class TupleGroup[A1: Group, A2: Group] extends Group[(A1, A2)] {
  def cmb(x: (A1, A2), y: (A1, A2)) = (A1.cmb(x._1, y._1), A2.cmb(x._2, y._2))
  def inv(x: (A1, A2)) = (A1.inv(x._1), A2.inv(x._2))
  // ...
}

Then the references to `A1` and `A2` only need to be stored once. So it seems like it would use less memory.

Apologies if I completely misunderstand how this works 😅

**Comment by joroKr21** [comment]
Since we've reached the limit of our JVM knowledge - what about applying this only to the SAM type classes and leaving the multiple method type classes as they were? SAM conversion is supposed to work automatically by the compiler but because of the bug it's not the case. Obviously that means our jar size savings will be much less, but there will be no danger of performance regressions.

**Comment by armanbilge** [comment]
👍 that sounds like a good way forward.

**Comment by joroKr21** [comment]
@armanbilge done - now only SAM type classes are optimised in this way. I converted multi method type class instances back to anonymous classes. There were not that many actually - only `Hash`, `Group`, `CommutativeGroup`, `Ring`, `Rig` and `Rng`.

File: core/src/main/scala/cats/Show.scala [hunk 2]
```diff
@@ -42,18 +42,12 @@ object Show extends ScalaVersionSpecificShowInstances with ShowInstances {
   /**
    * creates an instance of [[Show]] using the provided function
    */
-  def show[A](f: A => String): Show[A] =
-    new Show[A] {
-      def show(a: A): String = f(a)
-    }
+  def show[A](f: A => String): Show[A] = f(_)
```
armanbilge, line 45: Is there a practical difference between these? e.g. no class is emitted.

joroKr21: Yes, indeed 👍 - because `Show` has only one method

armanbilge: Right, thanks. Is this affected by the aforementioned Scala 2 bug?

Also, can `Semigroup` etc. get the same treatment?
https://github.com/typelevel/cats/blob/6ba7e4b782c0355c7fa2b5a787355a19b22d469a/kernel/src/main/scala/cats/kernel/Semigroup.scala#L151-L154

joroKr21: I think for `Semigroup` it doesn't matter because it has other (non-abstract) methods. So then the bug applies and there would be an anonymous classes generated even if we use the SAM syntax. So the only benefit would be aesthetics.

armanbilge: I see, so it's not just that `Show` has one abstract method, it's that it has no other methods.

I agree it's only aesthetics on Scala 2, but does the bug apply to Scala 3 as well?

joroKr21: > I agree it's only aesthetics on Scala 2, but does the bug apply to Scala 3 as well?

I don't remember - that's a good question.

armanbilge: The reason I ask is, I don't care so much about the handful of `Semigroup.instance` etc. that can be re-written.

But I'm wondering if in Scala 3 the boilerplate instances themselves could be written directly like this. So we get the win in terms of jar size without introducing any indirection at all.

joroKr21: Even if that's possible it would be quite hard to split like this 😄

joroKr21: FTR Scala 3 doesn't have this bug - but I again I'm sceptical about version-specific boilerplate generators 🤔

armanbilge: If there's interest it can be a followup PR. I think it shouldn't be _too_ hard (famous last words): currently all the instances are in the same file I think, when they could easily be split among a few files. And some of those files can go into the scala-2 and scala-3 srcs.

armanbilge (approved):
I didn't review the boilerplate code in detail, but I did spent some time studying the decompiled bytecode of the generated classes. The extra indirection is unfortunate but quite possibly not a big deal (and potentially avoidable for Scala 3 in follow-up work). Given the various constraints I think this is good 👍

DavidGregory084:
> This `inv` lambda must also be storing references to `A1` and `A2`.

Sorry for coming to this late but there is a great page about the lambda translation process [here](http://cr.openjdk.java.net/~briangoetz/lambda/lambda-translation.html) and a good blog post with some further explanation of what happens at runtime [here](https://blogs.oracle.com/javamagazine/post/behind-the-scenes-how-do-lambda-expressions-really-work-in-java).

Scalac uses the [LambdaMetafactory](https://docs.oracle.com/javase/8/docs/api/java/lang/invoke/LambdaMetafactory.html) provided in `java.lang.invoke` and it works in a similar way.

In the example that you provide:

```scala
val cmb = (x, y) => (A1.cmb(x._1, y._1), A2.cmb(x._2, y._2))
val inv = x => (A1.inv(x._1), A2.inv(x._2))
```

These lambdas are capturing lambdas, but they are not instance-capturing lambdas (they do not use `this` or `super` because `A1` and `A2` are constructor arguments of `TupleGroup`), so they are encoded as static methods. That does make me wonder if it behaves differently when you use an `implicit val` argument instead of using a context bound or just an `implicit` argument though!

The static methods will be declared with the lambda parameters with any captured variables prepended to the parameter list. You can actually see this in the bytecode for the `INVOKEDYNAMIC` call that @joroKr21 posted above:

```
// handle kind 0x6 : INVOKESTATIC
cats/kernel/instances/TupleMonoidInstances.$anonfun$catsKernelMonoidForTuple2$1(Lcats/kernel/Monoid;Lcats/kernel/Monoid;Lscala/Tuple2;Lscala/Tuple2;)Lscala/Tuple2; itf, 
```

The first two parameters are `cats.kernel.Monoid` instances. The `ALOAD 1` and `ALOAD 2` calls just before the `INVOKEDYNAMIC` bytecode stack those "static arguments" for the bootstrap method that creates the lambda object.

The [VM spec](https://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.27.4) says that:

> At run time, evaluation of a lambda expression is similar to evaluation of a class instance creation expression, insofar as normal completion produces a reference to an object. Evaluation of a lambda expression is distinct from execution of the lambda body.
>
> Either a new instance of a class with the properties below is allocated and initialized, or an existing instance of a class with the properties below is referenced. If a new instance is to be created, but there is insufficient space to allocate the object, evaluation of the lambda expression completes abruptly by throwing an OutOfMemoryError. 

So like @armanbilge suggests, these lambda expressions will capture `A1` and `A2` separately as the static arguments to those lambda expression objects. Each static lambda expression will *probably* result in the creation of one lambda object with a reference to each of those static arguments, although the behaviour is deliberately left completely up to the VM implementors so it's hard to be sure that different VMs will even behave the same:

> These rules are meant to offer flexibility to implementations of the Java programming language, in that:
>
>    A new object need not be allocated on every evaluation.
>
>    Objects produced by different lambda expressions need not belong to different classes (if the bodies are identical, for example).
>
>    Every object produced by evaluation need not belong to the same class (captured local variables might be inlined, for example).
>
>    If an "existing instance" is available, it need not have been created at a previous lambda evaluation (it might have been allocated during the enclosing class's initialization, for example).

As of relatively recent versions of the JDK it seems that:

> In the current implementation, the metafactory delegates to code that uses an internal, shaded copy of the [ASM bytecode libraries](https://asm.ow2.io/) to spin up an inner class that implements the target type.

joroKr21:
Thanks for detailed explanation @DavidGregory084 - I think that supports our current approach to apply this change to SAM type classes only

joroKr21:
Hey, is there any reason not to get this merged? 🙏

armanbilge:
I'm happy to move forward with this as of https://github.com/typelevel/cats/pull/3871#pullrequestreview-879681120. I feet like I remember someone (you?) saying we should still benchmark it or something before merging, but maybe I'm confused.

joroKr21:
> I feet like I remember someone (you?) saying we should still benchmark it or something before merging, but maybe I'm confused.

I thought that referred to the instances which have more than one abstract method which are now removed.
I personally won't have time to benchmark the rest 😄

danicheg (approved):
This is tremendous.

armanbilge:
We have previous approvals from Ross and Oscar as well, let's go ahead 🚀

armanbilge:
In unrelated research, I found out that SAMs still generate classes on JS (unsurprisingly). So this PR actually didn't help things there, although I don't think it made things worse.

Not sure the situation on Native, but if I had to guess it would be the same as JS.

joroKr21:
I assume both JS and Native have functions though - otherwise I can't explain the reduction of jar size I observed in both. Note that we use functions here, not SAM syntax directly to work around the Scala 2 bug.

armanbilge:
Ah, sorry, to clarify I'm not talking about the bytecode/SJSIR size. I'm talking about the size of the final generated JavaScript. Since JS is delivered into browsers this is a sensitive subject 😉

armanbilge:
> Note that we use functions here, not SAM syntax directly to work around the Scala 2 bug.

Right, sorry, I forgot this :)

Issue #2041, opened 2017-11-25
Move typelevel/algebra into cats repo.


LukaJCB:
This was discussed during the cats meeting at Scala By The Bay this year and I think most of us thought it'd be a good idea.

johnynek:
+1

kailuowang:
+1 under a different root package or cats.algebra?

johnynek:
maybe it should be cats algebra. It is a shame when we have two different artifact ids with the same class names inside.

ctongfei:
+1

johnynek:
on second thought... I don't know why I said that. The artifact can be the same: `org.typelevel:algebra` so I don't see why we would have the artifact ID collision.

So in that case, it is best to just keep it as it is.

denisrosset:
What is the timeframe for that move? I'd like to discuss in algebra some changes I am making in the Spire hierarchy, so that could impact what goes finally into cats.

ctongfei:
@johnynek I prefer `cats.algebra`. Sometimes you have your own package called `X` and you have a package `X.algebra` and this sometimes causes path collision -- to refer to Typelevel Algebra you must do `_root_.algebra`. See use cases like this:
https://github.com/ctongfei/nexus/blob/master/algebra/src/main/scala/nexus/algebra/IsReal.scala#L12:15

So maybe `cats.algebra._` is a better choice?

LukaJCB:
Would we want a separate module without the `cats-core` dependency, or a "real" merge with the dependency or possibly even same module?

benhutchison:
IIRC, Kernel was split out to enable the Algebra project to depend on Cats (and not the other way round). If that was the predominant reason (?), then I prefer condensing them into kernel than adding an intermediate node in the dependency tree.

LukaJCB:
What was the reason to split algebra into kernel? 🤔 Anyone of the old guard still have some info? @non maybe?

johnynek:
There is a cats issue where it was discussed. The reason was cats wanted fewer dependencies I think and didn’t care about Rings.

LukaJCB:
I've been trying to find an issue that accurately describes the actual problem with having the `Ring` and `Lattice` family in cats, and I can't really seem to find a good reason anywhere? Is it bloat? Jar size? Compatibility guarantees?

johnynek:
I think it is just OCD around minimization. This code is very stable and had binary compatibility guarantees much longer than cats did.

[...]

PR #3877, opened 2021-05-01
Move typelevel/algebra into cats repo

armanbilge:
This PR will migrate algebra into cats. As far as I can tell @denisrosset got the final word in at https://github.com/typelevel/cats/issues/2041#issuecomment-425500016.
> I wouldn't put algebra in cats-kernel, but as a separate module it's fine.

So far I've followed his proposal and moved algebra core and laws into the cats repo as their own modules, which I believe should already be compiling/passing tests for 2.12.

### Questions

1. Should algebra be its own module or merged into kernel?
2. What should be the package, `algebra` or `cats.algebra`? `cats.kernel`??
3. How to handle collections-related deprecations? Possible solutions:
    1. Add `@nowarn` annotations for 2.13; disable fatal warnings for 3.0
    2. Migrate source to 2.13 collections w/ [scala-collection-compat](https://github.com/scala/scala-collection-compat)
    3. Maintain separate version-dependent sources
 4. Where should the algebra docs go?
 5. Do we want to keep this [tiny benchmark](https://github.com/typelevel/algebra/tree/master/benchmark)?
 6. This effort is specifically mentioned as [paid maintainer's responsibility](https://opencollective.com/typelevel); might I kindly expect some compensation? :)


armanbilge:
Pending final answers to questions 1 and 2 above, I think this is in good shape for a basic code migration. Is there any follow-up work to do? e.g. https://github.com/typelevel/algebra/issues/199

denisrosset:
I'd vote for `cats-algebra`.

Note that I'd like to include one change from Spire if this gets baked into cats: right now, `typelevel/algebra` does not support Euclidean division nor GCD/LCM operations, whereas Spire does (with proper commutative ring types and laws).

So the algebra hierarchy goes `CRing -> Field`, while Spire's hierarchy goes `CRing -> GCDRing -> EuclideanRing -> Field`, and we have the ugly consequence that `algebra.ring.Field` is not `spire.algebra.Field`.

larsrh:
Thanks for that work! It'll take me a bit to go through everything.

To answer your questions:

> Should algebra be its own module or merged into kernel? What should we name the artifact?

Like @denisrosset I'd say `cats-algebra`.

> What should be the package, `algebra` or `cats.algebra`? `cats.kernel`??

`cats.algebra`

> Do we care about binary compatibility wrt the previous algebra artifacts?

No, I think that's too hard.

> This effort is specifically mentioned as a [paid maintainer's responsibility](https://opencollective.com/typelevel); might I kindly expect some compensation? :)

It would be best to ask first before sending the PR :slightly_smiling_face: But on a case by case basis we can discuss this, please email me.

denisrosset:
Note: I've moved the Spire stuff that should be in algebra in these two PRs: typelevel/algebra#246 typelevel/algebra#247 .

The second one can wait forever, but we should consider merging the first one -- as we're breaking binary compatibility, we can make that change now.

armanbilge:
@denisrosset Brilliant! That helps a lot, thanks.

denisrosset:
I think I understand the ring stuff in algebra pretty well, including the tradeoffs in naming stuff, duplication between standard/additive/multiplicative group-like structure. We also have pretty common data types (polynomials, integers, rational numbers) take exercise this stuff at different levels in the hierarchy. Well... except the noncommutative rings, or worse, structure with noncommutative addition; but I see that algebird depends on those, and they fit into a regular naming scheme already in cats-kernel.

However, I don't understand well the structure of the lattice subpackage. Why is there duplication between `Logic` and `GenBool`? See also typelevel/algebra#108

Should we merge all of that as it this into cats?

algebra-core/src/main/scala/cats/algebra/ring/EuclideanRing.scala
```diff
@@ -0,0 +1,60 @@
...
+
+  /**
+   * Simple implementation of Euclid's algorithm for gcd
+   */
+  @tailrec final def euclid[@sp(Int, Long, Float, Double) A: Eq: EuclideanRing](a: A, b: A): A = {
+    if (EuclideanRing[A].isZero(b)) a else euclid(b, EuclideanRing[A].emod(a, b))
+  }
+
+  /*  @tailrec final def extendedEuclid[@sp(Int, Long, Float, Double) A: Eq: EuclideanRing](a: A, b: A): (A, A, A) = {
+      if (EuclideanRing[A].isZero(b)) a else euclid(b, EuclideanRing[A].emod(a, b))*/
```
armanbilge, line 58: @denisrosset what to do about this?

denisrosset: I wanted to implement the extended Euclidean algorithm; this can be removed for now (it is not present in algebra anyway).

larsrh:
@denisrosset How do you think should we move forward on this? Personally I am a big fan of not letting best be the enemy of good :smile: So if we have some actionable changes then we should do them but otherwise we can let it be.

armanbilge:
https://github.com/typelevel/algebra/issues/108 looked easily actionable to me and also seemed to achieve a consensus in the original discussion (I assume it was never actioned due to binary compatibility constraints). At least for me, the biggest confusion working on this is not determining what is/isn't actionable but what there is/isn't consensus to *take* action on :) I think it's worth trawling algebra issues/pulls for other binary-incompatible changes since this is our chance to get them in and then ... put them to a vote??

algebra-laws/js/src/main/scala/cats/algebra/laws/platform/Platform.scala
```diff
@@ -0,0 +1,10 @@
+package cats.algebra.laws.platform
+
+private[laws] object Platform {
```
larsrh, line 3: Can we use the same definition from `kernel-laws`? There, it is defined as `private[cats]` so it should be usable here.

larsrh:
> I think it's worth trawling algebra issues/pulls for other binary-incompatible changes since this is our chance to get them in and then ... put them to a vote??

AFAICT most issues on Algebra are 4+ years old, so not sure if there's renewed interest in voting. Since I have no horse in the (algebra) race I'd suggest that @denisrosset and you just pick all changes that you feel are appropriate?

algebra-core/src/main/scala/cats/algebra/Priority.scala
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * This type class will attempt to provide an implicit instance of `P`
+ * (the preferred type). If that type is not available it will
+ * fallback to `F` (the fallback type). If neither type is available
+ * then a `Priority[P, F]` instance will not be available.
+ *
+ * This type can be useful for problems where multiple algorithms can
+ * be used, depending on the type classes available.
+ */
+sealed trait Priority[+P, +F] {
```
larsrh, line 15: It looks like this trait is never actually used, so I think we should scrap it.

algebra-laws/js/src/main/scala/cats/algebra/laws/platform/Platform.scala
armanbilge: Yup, I think so. Adding the `kernel-laws` dependency also lets us drop a bunch of other dupes, like serialization laws.

algebra-core/src/main/scala/cats/algebra/Priority.scala
armanbilge: Funny, I make good use of it in my own project and maybe it even got submitted to Scala? https://github.com/typelevel/algebra/issues/67

Anyway, scrapped.

algebra-laws/js/src/main/scala/cats/algebra/laws/platform/Platform.scala
armanbilge: Done. Now also using Serialization, Eq, and (Partial)Order tests directly from kernel-laws.

armanbilge:
@denisrosset @johnynek I just noticed your parallel discussions in https://github.com/typelevel/algebra/pull/246 and https://github.com/typelevel/algebra/pull/247, which makes my work here confusing (if algebra is not frozen, then we should put this migration effort on halt). May I kindly request that either we discuss here, or perhaps @denisrosset can make his PRs against [armanbilge:migrate-algebra](https://github.com/armanbilge/cats/tree/migrate-algebra)?

denisrosset:
@armanbilge I suggest we finish discussing the details on the separate PRs, and then I'll be happy to rewrite the changes against your PR. Works for you?

I'll dig a bit into the logic/lattice stuff, but I never used it in my own code. Right now, I don't get some of the design decisions, but it may click at some point (anyway, do you need the lattice/logic stuff for your code? at least the lattice stuff is textbook clear, we could merge that part first if needed).

armanbilge:
@denisrosset No problem, perhaps I was too gung-ho with this PR. I'm personally not in any rush for it so (assuming no other urgency) it can lay low until things stabilize over at algebra, at which point I can continue migration.

project/AlgebraBoilerplate.scala
```diff
@@ -0,0 +1,162 @@
...
+      val headerLines = header.split('\n')
+      val raw = range.map(n => content(new TemplateVals(n)).split('\n').filterNot(_.isEmpty))
+      val preBody = raw.head.takeWhile(_.startsWith("|")).map(_.tail)
+      val instances = raw.flatMap(_.filter(_.startsWith("-")).map(_.tail))
+      val postBody = raw.head.dropWhile(_.startsWith("|")).dropWhile(_.startsWith("-")).map(_.tail)
+      (headerLines ++ preBody ++ instances ++ postBody).mkString("\n")
+    }
+  }
+
+  object GenTupleInstances extends Template {
```
joroKr21, line 80: This can be optimised to reduce the number of generated classes - see #3871
Can also be done post merge ofc

armanbilge:
@denisrosset @johnynek I haven't seen any more activity over at algebra, so deferring to @larsrh's wise words about "letting best be the enemy of good" I'm going to try and wrap this up :)

project/AlgebraBoilerplate.scala
armanbilge: I will wait to do this after #3871 is reviewed/merged in case there are any concerns. Essentially, it's trading static memory (generated classes) for dynamic memory (pointers to lambdas)?

joroKr21: These days classes are also loaded in dynamic memory and can be unloaded by default.

johnynek:
I'm very strongly of the opinion that users do not benefit from us changing from `algebra` to `cats.algebra`.

There is a nice motivation, from which users benefit, from us having the code in this repo: we can more easily publish and we are more likely to leverage algebra in cats proper.

But, breaking every previous user of algebra just to make the package match cats, to me, is a loss. Algebra was an effort started many years ago to allow Spire and Algebird to have interoperable types. It took a long time to even achieve that goal. It was easy to make the algebra code, but to actually include it in algebird and spire took a few years. Every breaking change puts a cost downstream that we don't see here. It's a negative that generally slows, not speeds, adoption.

This PR proposes we will say to the two projects that came together to make algebra: you need to push a binary incompatible change onto your users so we can have a package name we prefer more.

In fact, algebra has changed VERY little over the years. This is not surprising: it was already a distillation of the common parts of algebird and spire, each of which already had some learnings in the design space.

I would MUCH prefer if we just keep the same artifact names `org.typelevel:algebra` and package names *and* keep binary compatibility with the previous package.

I don't want to see this as an opportunity to break things just because we feel it is a fine time to make some cosmetic changes.

larsrh:
Algebra has been released for Scala 2.12, 2.13 and 3, so users do not have any pressing need to migrate to `cats-algebra` (or whatever it ends up being called). No substantial development is happening there, so they are free to ignore whatever is going on there.

In fact, the two proposed changes (artifact name + package name) go hand in hand and make migration a ton easier, since both artifacts can coexist. OTOH, keeping both the same and shoe-horning Algebra into the Cats version scheme would be difficult from a maintenance PoV.

Since Cats is where most of the action happens nowadays, the ability to re-start development greatly outweighs the (negligible) downstream costs.

johnynek:
Okay so spire and algebird, the folks who made this project, should just ignore cats-algebra and then we have two typelevel algebra projects?

Or are you really saying there are negligible downstream users of those two projects (who are impacted by changing the names of these types of folks adopt them).

It's also weird how this decision is made. Being a founder of this project doesn't carry much weight? Is the idea we just find one typelevel committer to agree?

larsrh:
> It's also weird how this decision is made. Being a founder of this project doesn't carry much weight? Is the idea we just find one typelevel committer to agree?

Oscar, I'm not sure where you're getting the idea that your voice is not being heard. We are quite literally discussing right now about the decision, which is because you're commenting on an open not a closed PR. Obviously this contribution has significant impact so we're keeping this open for a bit to discuss. If you scroll up in this thread you see that a lot of discussion has already happened.

> Okay so spire and algebird, the folks who made this project, should just ignore cats-algebra and then we have two typelevel algebra projects?

That's not what I said at all. My point is that they can migrate at their own pace. Of course I can't speak for Algebird but I'm pretty sure we can commit to porting Spire.

> Or are you really saying there are negligible downstream users of those two projects (who are impacted by changing the names of these types of folks adopt them).

With "negligible downstream cost" I mean hypothetically merging this PR as it is incurs 0 cost to downstream users. It would be different if we'd – for the sake of argument – kept the artifact name but still incur breaking changes.

armanbilge:
> If you scroll up in this thread you see that a lot of discussion has already happened.

It's also worth pointing to the original thread in #2041. As far as I can tell this particular issue was never raised (and certainly not so strongly) and at the time @johnynek seemed open to artifact/package changes, although I recognize that's an older thread, and opinions/requirements certainly evolve. 

> OTOH, keeping both the same and shoe-horning Algebra into the Cats version scheme would be difficult from a maintenance PoV.

Difficult maybe but since Cats versioning is ahead of Algebra's, technically this should be possible right? Algebra would skip from 2.2 to 2.6 I think. However, personally I am inclined to agree with @larsrh's point that

> Algebra has been released for Scala 2.12, 2.13 and 3, so users do not have any pressing need to migrate to cats-algebra

Admittedly, I am also not a user of Algebird. And doesn't Spire potentially have some breaking changes ahead, as we figure out how to port it to Scala 3? I can certainly commit to migrating Spire to cats-algebra.

johnynek:
@armanbilge I think the comment I made here is very in line with what I'm saying now:
https://github.com/typelevel/cats/issues/2041#issuecomment-348034250

Lars, yes I realize that the issue is still open, but to the point: how will we come to a conclusion. This comment: https://github.com/typelevel/cats/pull/3877#issuecomment-841255675

speaks with a sound of certainty: e.g. what package? `cats.algebra`.

It does not discuss how we will decide. Obviously, it seems you and I are diametrically opposed on visions for moving forward.

Like @armanbilge I don't see a problem with advancing the version number of algebra up to match cats. That seems fine to me.

I'd like to see a list of the concrete things we really want to change.

If we don't really have those now, why not defer the package/artifact rename until we do. Some of what @denisrosset has suggested amounts to basically a rewrite. If we want to go down that avenue, we shouldn't merge first. Note, I would disagree with his argument (renaming things to CommutativeAdditiveMoniod to Plus0 would be like renaming cats.Functor to Mappable, but we can discuss that separately). If a rewrite is desired, just start as cats.algebra. But importing all the code, renaming the package, but then not actually making the changes incurs the cost without the benefit.

Lastly, I think there should be a strong prior on *NOT* breaking compatibility, rather than assuming a break is negligible costs. Look at the pain of the binary breaks in cats-effect 3. In that example, there is a strong reason for it, yet still the ecosystem has to resort to suffix versioning of downstream dependencies. I think we should put the burden of proof on those who argue for a break, not on those who argue against it.

johnynek:
Also in fairness to you all, if you want to reply to my subtweet of this event:

https://twitter.com/posco/status/1400138900362854401

larsrh:
> speaks with a sound of certainty

You are overinterpreting. I made my comment about _two weeks_ after the initial PR has been created with nobody else chiming in. Forgive me for not prefixing literally everything I write with "IMHO".

> Obviously, it seems you and I are diametrically opposed on visions for moving forward.

That much is clear. Then again it does not give you the right to complain that you're not being heard.

> But importing all the code, renaming the package, but then not actually making the changes incurs the cost without the benefit.

I don't get this, there are changes in this PR beyond renaming the package.

> Look at the pain of the binary breaks in cats-effect 3. In that example, there is a strong reason for it, yet still the ecosystem has to resort to suffix versioning of downstream dependencies.

CE3 is not comparable to this because they didn't change the artifact coordinates.

armanbilge:
> not actually making the changes incurs the cost without the benefit

In case you missed it, I did make the changes that you and @denisrosset discussed in typelevel/algebra#246 and typelevel/algebra#247. As well as some other changes (open to review/discussion of course) to provide a better cats experience: removing conflicting implicits (although it may be possible to do this in a binary-compat way) and using the laws/test distinction.

johnynek:
@larsrh you say

    That much is clear. Then again it does not give you the right to complain that you're not being heard.

I am not complaining about not being heard. I asked a question about how this decision is made:

    It's also weird how this decision is made. Being a founder of this project doesn't carry much weight? Is the idea we just find one typelevel committer to agree?

I think I do have the right to ask how a decision is made on this. If I don't who does? I am a committer in cats and a founder of algebra. Not enough contribution to get a seat at the table?

It sounds like you are saying "we have left the PR open, you can comment here" but then? How do we decide to merge? I mean, I believe in free and open source software, so you can do whatever you want, but taking someone else's project copying the code and then saying "yeah, we heard your objections" isn't a friendly way to work, IMO.

About the changes: the changes you two reference could have been included in algebra (and indeed were PRs before @denisrosset ran out of time). So, the motivation to make a new artifact and package name, to me, should be to correct things that *cannot be done* in a compatible way.

larsrh:
> I think I do have the right to ask how a decision is made on this. If I don't who does? I am a committer in cats and a founder of algebra. Not enough contribution to get a seat at the table?

Literally nobody is denying you the right to discuss this here. You made a point. I made a counter-point. You complained that your opinion does not carry weight, which is a non-sequitur.

> It sounds like you are saying "we have left the PR open, you can comment here" but then? How do we decide to merge? I mean, I believe in free and open source software, so you can do whatever you want, but taking someone else's project copying the code and then saying "yeah, we heard your objections" isn't a friendly way to work, IMO.

I don't know, in the absence of any formal voting process, wait for a few other people to chime in? We're not on a deadline here. What makes this PR so different from others before it? Do you claim a veto right?

And please don't tell me this isn't a friendly way to work when you're at the same time insinuating that I'm not listening to you.

armanbilge:
> the changes you two reference could have been included in algebra (and indeed were PRs before @denisrosset ran out of time)

My understanding was @denisrosset made those PRs in response to this one (see https://github.com/typelevel/cats/pull/3877#issuecomment-841300254) and they would be binary-breaking to algebra.

> I think I do have the right to ask how a decision is made on this.

I was definitely confused about this when I started as well https://github.com/typelevel/cats/pull/3877#issuecomment-841626881. I was open to go any direction. So far, @denisrosset @larsrh and @johnynek have chimed in, 2 for the artifact/package change, 1 against. Plenty in the original thread #2041 also seemed to support a change.

johnynek:
@lars I guess we can't discuss this. You keep putting words in my mouth:

I said:

> I think I do have the right to ask how a decision is made on this. If I don't who does? I am a committer in cats and a founder of algebra. Not enough contribution to get a seat at the table?

and you replied:

> Literally nobody is denying you the right to discuss this here. You made a point. I made a counter-point. You complained that your opinion does not carry weight, which is a non-sequitur.

1. Where did I say or imply that someone is denying me a right to discuss? Can you link to the comment? I don't feel that I am being prevented from sharing my voice. What I am concerned about is that there is no clear process, and it sounds like you are not interested in accommodating my views in the least. I haven't seen any notion of compromise. How is your position different than "Oscar, you can add all the comments you like, but I'm going to merge it the way I want it."? That's not rhetorical. I really am trying to communicate here and see what the difference is.
2. I don't think me asking about the relative weight of my voice on this decision is a non-sequitur. I also disagree with the characterization of my asking your view about my weight is equivalent to me complaining (unless we perhaps presume your answer to that question to be that weight to be very limited, or no weight).

> What makes this PR so different from others before it? Do you claim a veto right?

Are you really saying that it should be totally fine to take some project, import it into cats, make a package name change, and ignore the folks who made the project? That's what makes this different? In 2013 @non @tixxit @sritchie and I had a call to try to unify the typeclasses in Spire and Algebird. I think the fact that was the mission of the project and we worked on it for the past 8 years is what is different here. You have the right to copy any OSS code you want. But I don't think you should surprised if you copy someone else's project into cats, but then don't come to an agreement that they feel frustrated. This is not a new story.

mpilquist:
If we're going to break binary compatibility, then changing the package name is a good approach, which will avoid linkage errors and allow co-existence until downstream projects can publish for the new artifacts. It seems like the critical question here is whether or not we need to break binary compatibility. I suggest we enumerate the issues that cannot be done without breaking compatibility then figure out a path forward from there. Is that something @denisrosset and @armanbilge can provide?

armanbilge:
This PR was binary compatible up until https://github.com/typelevel/cats/pull/3877/commits/c51b72671ee9f0399c4f9bba4fe1c7cf37ddebb3.

I'm no expert on binary compatibility, so I can't be 100% sure without running MiMa, but to the best of my knowledge these are the binary-breaking changes made:

a6ec3b822de83d4eebd0a523c2061105994aafa4 Merge DeMorganLaws into LogicLaws per TODO
8ddf2a40a3dd5c7271db8ce89a42c7a9203c3b40 Merge in typelevel/algebra#246
513086720153726c3eedc8eba843040d3da07a70 Add kernel-laws dependency and refactor
7ff1809c065a020d269bc64e34a9a83f8e2c2417 Delete priority
048ce2ef1364873488057d7a6a53cb42d0768c37 Implement typelevel/algebra#108
b81e1fbfcbe1041ce26ac3f00a88b4d62f3f752d Make Field inherit from DivisionRing https://github.com/typelevel/algebra/pull/246#discussion_r633507731
844670d7de63899e65a09bec4d13885f90f75418 Rename laws -> tests, remove redundancies
44eb36c88eac73ccbe8979a7ab6570562b398618 Refactor instances to avoid conflicts/ambiguities

johnynek:
1. I'd like to also mention, compatibility is a continuum: if you are totally compatible, then there is not a problem, but every break MAY or MAY NOT break a user (whose code may or may not touch that thing).

We shouldn't think of it as "once we break one thing we might as well change anything we want".

With that in mind, binary compatibility for the `algebra` package vs the `algebra-laws` package is much more important to me (simply because the laws package I believe to have far fewer users, and almost no transitive users: you use it to compile the tests in a single repo, not to build another testing library to appear in diamonds).

2. Note: we can remove `implicit` without hurting binary compatibility. cats has done this several times.

I don't see anything in that list that looks like a huge win to users, and I think we can find a way to do what we want and remain compatible. As is probably clear: I am very -1 on us breaking compatibility as perfectionist library authors when that has no significant benefit to users. The users aren't here advocating for themselves. The library authors, who want the code to be as clean as possible and often don't have to deal with building apps composed of hundreds or even thousands of dependencies, are represented. I feel it is important to weight heavily the downstream effects of changes.

3. The fact that both algebird and spire depend on this means algebra has a good chance to be a diamond dependency (something you get from more than one path in your dependency graph).

armanbilge:
> It seems like the critical question here is whether or not we need to break binary compatibility.

Personally, I feel the binary-breaking changes could fall into roughly two categories:
1. Changes like https://github.com/typelevel/algebra/pull/246 that improve Algebra itself
2. Changes like https://github.com/typelevel/cats/commit/844670d7de63899e65a09bec4d13885f90f75418 that make Algebra more "Cats-friendly" (similar coding style/organization, etc.)

I just took a read through https://github.com/typelevel/cats/pull/762 which I believe is the origins of `cats-kernel`. Specifically, I found https://github.com/typelevel/cats/pull/762#issuecomment-168749770 by @non to be extremely insightful.

> Algebra was small, relatively unopinionated, and already existing. (Being unopinionated around implicits/syntax was necessary to even be in consideration as an Algebird dependency.)

Coupled with @johnynek's https://github.com/typelevel/cats/pull/3877#issuecomment-853227734 above:

> Algebra was an effort started many years ago to allow Spire and Algebird to have interoperable types

Now, I understand much better where Algebra's goals (to be agnostic and unopinionated) diverge with Cats' (to define an ecosystem). This leads to pain points such as https://github.com/typelevel/algebra/issues/199 and explains why `cats-kernel` lacks syntax: `cats-kernel` was accommodating Algebra which in turn was accommodating Algebird.

I am strongly of the opinion that, if one of the goals of moving Algebra into the Cats repo is to re-vitalize development (something that @larsrh and @johnynek seem to agree on), then we should use this opportunity for Algebra to more closely follow the patterns of a Cats-style project. I can certainly understand why 4-5 years ago when Cats was younger the preference for agnosticism, but now Cats seems to be a much more established part of the ecosystem. Any `cats-algebra` user already following Cats-style patterns (such as Spire) could definitely benefit from this.

On the other hand, if Algebra's mission first and foremost continues to be to be the bridge between Spire and Algebird, I'm not sure at all it belongs in the Cats repo.

I started writing this before @johnynek's comment above about `algebra` vs `algebra-laws` artifacts, in which case I add it should possible to accomplish at least some of the changes I mentioned above while still retaining binary-compatibility in `algebra`.

Finally, In case it needs to be said: all just my humble opinion, as a relatively new participant in the Typelevel community (although longtime user of Spire). I opened the PR open to keeping binary compatibility and would be happy to return to that path.

mpilquist:
Totally agree on `algebra-laws` vs `algebra` -- great point.

In addition to removing `implicit`, there are some other tricks like changing things that were once public to `private[algebra]`. Not sure if that's useful in this PR or not but worth considering.

>  I am very -1 on us breaking compatibility as perfectionist library authors when that has no significant benefit to users.

Me too (see my feedback every time we discuss cats changes for Scala 3, or the great lengths I've gone to in scodec-bits) and I appreciate your reminders about the importance of binary compatibility. I don't think it's fair to say this is a case of perfectionist library authors though -- folks are trying to implement changes that seemed generally agreeable. Let's continue to discuss those changes and figure out if we can do the ones we want without breaking compatibility.

We should also figure out how we handle suggestions or ideas which break compatibility and document that process. I want to avoid situations where folks put a bunch of work in to a PR and then we have to abandon it b/c the bincompat implications weren't made clear.

> The fact that both algebird and spire depend on this means algebra has a good chance to be a diamond dependency (something you get from more than one path in your dependency graph).

Yep! Hence, *if* we break compat, a new package name is warranted -- especially given the wildly different release cadences of those projects. But let's defer that discussion until we figure out whether we really need to break compatibility or not.

How about this path forward: let's create a new PR that starts at the last bincompat commit, add mima checking for `algebra` but not `algebra-laws`, and then see what of the additional commits can be made safely?

armanbilge:
> How about this path forward: let's create a new PR that starts at the last bincompat commit, add mima checking for algebra but not algebra-laws, and then see what of the additional commits can be made safely?

Very nearly all of them. See the full story over at #3918.

johnynek:
just wanted to add a data point:

https://mvnrepository.com/artifact/org.typelevel/algebra/usages?p=1

there are 35 direct dependencies on algebra. The top three of which have more than 200 dependencies total.

We are talking about hundreds of downstream projects impacted here. It's hard to guess the number of developers but I would guess it is likely more than 10k (perhaps much more). To me this is what merits being very biased to not breaking.

PR #3895, opened 2021-05-18
Support lifting polymorphic functions to FunctionK in Scala 3

bplommer:


bplommer:
The CI failure seems to be GH actions playing up.

tests/src/test/scala-3.x/cats/tests/FunctionKLiftSuite.scala
```diff
@@ -0,0 +1,16 @@
...
+
+import cats.arrow.FunctionK
+import cats.implicits._
+import org.scalacheck.Prop._
+import cats.laws.discipline.arbitrary._
+
+class FunctionKLiftSuite extends CatsSuite {
+
+  test("lift a polymorphic function directly") {
+    val fHeadOption = FunctionK.lift[List, Option]([X] => (_: List[X]).headOption)
```
smarter, line 11: Note that in the future the compiler should support polymorphic sam, so you'll be able to write `val fHeadOption: List ~> Option = [X] => (_: List[X]).headOption` (or maybe even `= _.headOption`)

core/src/main/scala-3/src/main/scala/cats/arrow/FunctionKMacros.scala
```diff
@@ -22,4 +22,14 @@
...
-private[arrow] class FunctionKMacroMethods
+private[arrow] class FunctionKMacroMethods:
+  /**
+   * Lifts function `f` of `[X] => F[X] => G[X]` into a `FunctionK[F, G]`.
+   *
+   * {{{
+   *   val headOptionK = FunctionK.lift[List, Option]([X] => (_: List[X]).headOption)
+   * }}}
+   */
+  def lift[F[_], G[_]](f: [X] => F[X] => G[X]): FunctionK[F, G] =
```
johnynek, line 33: This isn't a macro is it? Why the name?

I wonder if there could be a Conversion here so you could use a literal in scala 3 without having to call lift.

armanbilge: I think it's convenient to put it in this file/class, because it was already present (the Scala 2 side uses it for macros). Ideally it could be renamed something more appropriate.

FTR I personally haven't had a good experiencing using `Conversion` on Scala 3, as it still a feature import to enable implicit conversions. That makes me tend to stay away.

armanbilge:
@bplommer do you mind fixing this up? My efforts to revive it have fallen short.

armanbilge (approved):

core/src/main/scala-3/src/main/scala/cats/arrow/FunctionKMacros.scala
joroKr21: Maybe I'm wrong but it shouldn't need a feature import if the conversion is on the companion object.

armanbilge: Thanks, I tried slapping an `implicit` in front of this but actually it doesn't seem to work 🤔 

```suggestion
  implicit def lift[F[_], G[_]](f: [X] => F[X] => G[X]): FunctionK[F, G] =
```

```
[error] -- [E007] Type Mismatch Error: /workspace/cats/tests/shared/src/test/scala-3/cats/tests/FunctionKLiftSuite.scala:40:77 
[error] 40 |    val fHeadOption: FunctionK[List, Option] = [X] => (_: List[X]).headOption
[error]    |                                                                             ^
[error]    |    Found:    [X] => (List[X]) => Option[X]
[error]    |    Required: cats.arrow.FunctionK[List, Option]
```

I also tried using `given Conversion[...]` but the compiler didn't like that with our current `scalacOptions`.

armanbilge: Yeah, it seems this just doesn't work. Maybe worth opening a dotty issue?
```scala
//> using scala "3.1.2"
//> using lib "org.typelevel::cats-core::2.7.0"

import cats.arrow.FunctionK

def lift[F[_], G[_]](f: [X] => F[X] => G[X]): FunctionK[F, G] =
  new FunctionK[F, G] {
    def apply[A](fa: F[A]): G[A] = f(fa)
  }

given [F[_], G[_]]: Conversion[[X] => F[X] => G[X], FunctionK[F, G]] = lift(_)

@main def main =
  val fHeadOption: FunctionK[List, Option] = [X] => (_: List[X]).headOption
  val a = List(1, 2, 3)
  assert(fHeadOption(a) == a.headOption)
```

joroKr21: I think just type inference fails here:
```scala
scala> summon[Conversion[[X] => List[X] => Option[X], FunctionK[List, Option]]](using given_Conversion_X_FunctionK)
-- [E007] Type Mismatch Error: -------------------------------------------------
1 |summon[Conversion[[X] => List[X] => Option[X], FunctionK[List, Option]]](using given_Conversion_X_FunctionK)
  |                                                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |Found:    Conversion[[X] => (Nothing) => Any, FunctionK[F, G]]
  |Required: Conversion[[X] => (List[X]) => Option[X], FunctionK[List, Option]]
  |
  |where:    F is a type variable with constraint <: [_] =>> Any
  |          G is a type variable with constraint <: [_] =>> Any

longer explanation available when compiling with `-explain`
1 error found
```

johnynek: yeah, this seems like maybe it is a bug that could be filed with scala.

armanbilge: I opened https://github.com/lampepfl/dotty/issues/15276.

PR #3896, opened 2021-05-18
Make FunctionK extend PolyFunction in Scala 3

bplommer:
Depends on #3895

### DIFF
--- a/core/src/main/scala/cats/arrow/FunctionK.scala
+++ b/core/src/main/scala/cats/arrow/FunctionK.scala
@@ -13,7 +13,7 @@ import cats.data.{EitherK, Tuple2K}
  *   val listToOption = λ[FunctionK[List, Option]](_.headOption)
  * }}}
  */
-trait FunctionK[F[_], G[_]] extends Serializable { self =>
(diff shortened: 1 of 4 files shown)


core/src/main/scala/cats/arrow/FunctionK.scala
```diff
@@ -13,7 +13,7 @@ import cats.data.{EitherK, Tuple2K}
  *   val listToOption = λ[FunctionK[List, Option]](_.headOption)
  * }}}
  */
-trait FunctionK[F[_], G[_]] extends Serializable { self =>
+trait FunctionK[F[_], G[_]] extends PolyFunction with Serializable { self =>
```
smarter, line 16: While extending PolyFunction should work I wouldn't really recommend it given that it's likely under-tested and we're still figuring out how these things should work exactly, so if there's a problem you'll be stuck with it due to the binary-compatibility requirements of cats. This also won't be of much use once we have polymorphic SAMs. Also it might make more sense to redefine `type FunctionK[F[_], G[_]] = [A] => F[A] => G[A]` whenever you'll be able to break binary-compatibility.

PR #3899, opened 2021-05-21
Applicative whenA/unlessA syntax extensions should respect call-by-name parameter

### LINKED ISSUE #3687: `whenA` and `unlessA` by-name makes syntax inconsistent with typeclass
`Applicative.whenA` and `Applicative.unlessA` are defined using a by-name argument. Syntax extension is via a value-class though which naturally can't use a by-name in it's constructor. This means that calling `whenA` is either eager or lazy depending on whether it is called from the typeclass (lazy) or the syntax (eager).

Definition:
```scala
def whenA[A](cond: Boolean)(f: => F[A]): F[Unit]
```

Syntax:
```scala
final class ApplicativeOps[F[_], A](private val fa: F[A]) extends AnyVal {
  def whenA(cond: Boolean)(implicit F: Applicative[F]): F[Unit] = F.whenA(cond)(fa)
}
```

I had a brief discussion with @SystemFw (sorry for ping!) [on gitter](https://gitter.im/typelevel/cats?at=5fb65e4f06fa0513ddc15a85) and they suggested opening the issue as it's a messy problem.

catostrophe:
The current implementation of the `.whenA`/`.unlessA` syntax extensions doesn't respect `F[A]` being passed as call-by-name.

This leads to issues where it's important.
See: https://github.com/typelevel/fs2/pull/2412

Unfortunately we can't change whenA/unlessA directly due to bincompat guarantees.
But the new method names even better reflect that we throw `A` out (as in `traverse_`). I would also deprecate the old ones.

Closes #3687

### DIFF
--- a/core/src/main/scala/cats/syntax/applicative.scala
+++ b/core/src/main/scala/cats/syntax/applicative.scala
@@ -6,6 +6,8 @@ trait ApplicativeSyntax {
     new ApplicativeIdOps[A](a)
   implicit final def catsSyntaxApplicative[F[_], A](fa: F[A]): ApplicativeOps[F, A] =
     new ApplicativeOps[F, A](fa)
+  implicit final def catsSyntaxApplicativeByName[F[_], A](fa: => F[A]): ApplicativeByNameOps[F, A] =
+    new ApplicativeByNameOps[F, A](() => fa)
 }
 
 final class ApplicativeIdOps[A](private val a: A) extends AnyVal {
@@ -16,4 +18,9 @@ final class ApplicativeOps[F[_], A](private val fa: F[A]) extends AnyVal {
   def replicateA(n: Int)(implicit F: Applicative[F]): F[List[A]] = F.replicateA(n, fa)
   def unlessA(cond: Boolean)(implicit F: Applicative[F]): F[Unit] = F.unlessA(cond)(fa)
   def whenA(cond: Boolean)(implicit F: Applicative[F]): F[Unit] = F.whenA(cond)(fa)
+}
+
+final class ApplicativeByNameOps[F[_], A](private val fa: () => F[A]) extends AnyVal {
+  def unlessA_(cond: Boolean)(implicit F: Applicative[F]): F[Unit] = F.unlessA(cond)(fa())
+  def whenA_(cond: Boolean)(implicit F: Applicative[F]): F[Unit] = F.whenA(cond)(fa())
 }


core/src/main/scala/cats/syntax/applicative.scala
```diff
@@ -17,3 +19,8 @@ final class ApplicativeOps[F[_], A](private val fa: F[A]) extends AnyVal {
   def unlessA(cond: Boolean)(implicit F: Applicative[F]): F[Unit] = F.unlessA(cond)(fa)
```
catostrophe, line 19: I also suggest deprecating these two

joroKr21: You could just make the old syntax not implicit and substitute the new one with the same names.

catostrophe: @joroKr21 then `replicateA` will become a syntax extenstion on` => F[A]`, while the method itself accepts `F[A]`. Is it ok?

joroKr21: Hmm I'm not sure

joroKr21: @catostrophe how about `replicateA(0)`? Shouldn't that also **not** execute `fa`?

catostrophe: Yes, but replicateA in Applicative takes a call-by-value param

joroKr21: Yes, and exposing a by-name API for `replicateA` might mislead users to think of a more imperative style of API like `List.fill` where `=> F[A]` might have side-effects.

Another option would be to extract `ApplicativeStrictOps` for `replicateA`

core/src/main/scala/cats/syntax/applicative.scala
```diff
@@ -17,3 +19,8 @@ final class ApplicativeOps[F[_], A](private val fa: F[A]) extends AnyVal {
...
   def whenA(cond: Boolean)(implicit F: Applicative[F]): F[Unit] = F.whenA(cond)(fa)
 }
+
+final class ApplicativeByNameOps[F[_], A](private val fa: () => F[A]) extends AnyVal {
+  def unlessA_(cond: Boolean)(implicit F: Applicative[F]): F[Unit] = F.unlessA(cond)(fa())
```
johnynek, line 24: I think we should use .void on these two.

johnynek (changes requested):
Can we add some test coverage?

Also I think this should probably be added to the Applicative trait and then the syntax is calling it so it can potentially be specialized by different type constructors when some efficiency can be gained.

core/src/main/scala/cats/syntax/applicative.scala
catostrophe: Hmm, what for?

johnynek: sorry, I got confused. Often we have `foo_` as a variant of `foo` that discards the value. But I see that here we are doing something else (just using it as another version that uses call by name correctly.

I would like to see another name since, I think, that pattern is established and this would be a violation of it.

core/src/main/scala/cats/syntax/applicative.scala
```diff
@@ -17,3 +19,8 @@ final class ApplicativeOps[F[_], A](private val fa: F[A]) extends AnyVal {
   def unlessA(cond: Boolean)(implicit F: Applicative[F]): F[Unit] = F.unlessA(cond)(fa)
   def whenA(cond: Boolean)(implicit F: Applicative[F]): F[Unit] = F.whenA(cond)(fa)
 }
+
+final class ApplicativeByNameOps[F[_], A](private val fa: () => F[A]) extends AnyVal {
```
satorg, line 23: Perhaps a minor thing, but wouldn't it better to keep constructors of *-`Ops` classes private to the `syntax` package? I know, not all the existing classes do follow it, but perhaps it is a good idea for all new classes, wdyt?
```scala
final class ApplicativeByNameOps[F[_], A] private[syntax] (private val fa: () => F[A])
  extends AnyVal
```
I mean, it is unlikely that we want to allow users to create `Ops` files explicitly, isn't? Or is there some use case for that?

Issue #2089, opened 2017-12-10
Remove OneAnd?


ceedubs:
`OneAnd` doesn't seem very useful to me now that Cats has first-class `NonemptyList` and `NonemptyVector` types. It can be awkward to work with, especially if you don't have type aliases and "companion" object set up for the type that you are working with.

It keeps its head element separate like a cons-list, making it not very efficient to use with a vector-like structure, but it `combine`s by delegating to an `Alternative` instance, which isn't efficient for list-like structures.

People might reach for `OneAnd` when they want a non-empty version of a structure that doesn't have first-class non-empty support in Cats. Some common types that they might reach for are non-empty sets and non-empty maps. However, the `OneAnd` model doesn't actually work for those (see #1909 for an example).

I'm not sure that `OneAnd` cuts it as an abstraction, and I don't know anyone who is actually using it. I vote for removing it. Is there anyone who uses it and/or has reason for it to stay?

Here is a related [gitter discussion](https://github.com/typelevel/cats/issues/1909).

LukaJCB:
I'm generally +1 on this, but would like to see a replacement for things like `NonEmptyStream` first :)

PR #3902, opened 2021-05-23
Deprecate OneAnd class.

diesalbla:
The class `cats.data.OneAnd` represetns a pair of one element and a collection of items. The main uses for this class, lists and lazy lists (fka streams), have their own specialised class. In issue https://github.com/typelevel/cats/issues/2089 it was suggested that this class is not really useful and should be removed.

I have checked some of the main TypeLevel projects: `cats-effect`, `doobie`, `fs2`, and `http4s` (core). None of these seem to be using the `OneAnd` class.

This commit marks the class, and the major methods that create instances involving it, as deprecated.

Related to https://github.com/typelevel/cats/issues/2089

### DIFF
--- a/core/src/main/scala-2.12/cats/data/OneAndLowPriority4.scala
+++ b/core/src/main/scala-2.12/cats/data/OneAndLowPriority4.scala
@@ -5,6 +5,7 @@ import scala.annotation.tailrec
 import scala.collection.mutable.Builder
 
 abstract private[data] class OneAndLowPriority4 {
+  @deprecated("For collections, use special NonEmpty types instead of OneAnd (see #3089 for details)", "2.6.2")
   implicit val catsDataComonadForNonEmptyStream: Comonad[OneAnd[Stream, *]] =
     new Comonad[OneAnd[Stream, *]] {
       def coflatMap[A, B](fa: OneAnd[Stream, A])(f: OneAnd[Stream, A] => B): OneAnd[Stream, B] = {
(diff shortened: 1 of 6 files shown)


rossabaker:
The "2.6.1" isn't right -- it would be 2.6.2 or 2.7.0.  But :+1: to the rest.

larsrh:
I think a deprecation annotation is missing here: https://github.com/typelevel/cats/blob/df8a49bad1cea64985f04caae5f507c8daa44ad9/laws/src/main/scala/cats/laws/discipline/arbitrary.scala#L46

core/src/main/scala-2.12/cats/data/OneAndLowPriority4.scala
```diff
@@ -5,6 +5,7 @@ import scala.annotation.tailrec
 import scala.collection.mutable.Builder
 
 abstract private[data] class OneAndLowPriority4 {
+  @deprecated("Will remove OneAnd in cats v3", "2.6.1")
```
larsrh, line 8: I suggest to change the deprecation message to something like:

> For collections, use special NonEmpty types instead of OneAnd (see #3089 for details).

diesalbla: This has been added to the PR.

satorg:
Just a side note: I'm not aware of any usage of `OneAnd` in Prod, but I personally found it sometimes kind of handy in tests (not that often though) to represent a concept of "one-or-two" items: `OneAnd[Option, *]`.
I mean, I can make use of just `Option` for "one-or-zero" items, or `OneAnd[Option, *]` for "one-or-two".
I wouldn't say it is something I cannot survive without, but can come handy in some particular cases.

PR #3906, opened 2021-05-26
Move Id instances into implicit scope for Scala 3

joroKr21:

### DIFF
--- a/core/src/main/scala/cats/Invariant.scala
+++ b/core/src/main/scala/cats/Invariant.scala
@@ -104,7 +104,9 @@ import scala.util.control.TailCalls.TailRec
 
 @suppressUnusedImportWarningForScalaVersionSpecific
 object Invariant extends ScalaVersionSpecificInvariantInstances with InvariantInstances0 {
-  implicit def catsInstancesForId: Distributive[Id] with Comonad[Id] = cats.catsInstancesForId
+  implicit def catsInstancesForId
+    : Distributive[Id] with Bimonad[Id] with CommutativeMonad[Id] with NonEmptyTraverse[Id] =
+    cats.catsInstancesForId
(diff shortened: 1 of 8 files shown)


johnynek (approved):

djspiewak (approved):

joroKr21:
I think I missed `Semigroupal`

armanbilge:
Ah, I was having trouble with this as well.

armanbilge:
Btw, I think we also should consider putting `Id`'s implicits in its companion object.

joroKr21:
It doesn't have a companion object - it's a type alias

armanbilge:
Ok, well whatever you call the `object Id` that sits next to the type alias :) I'm pretty sure anything there is in implicit scope in Scala 3, I recently did this in https://github.com/typelevel/feral/pull/52/commits/744013e3ccd6123f0318f24f8fa3191b478aa919#diff-de6a1bd431e8326aa1f0db9116e7f8d27c15a9c8a6bb6344aec9984873eead98

smarter:
For abstract types, scala 3 will look for implicits in the same-named object, but not for type alias if I'm not mistaken.

armanbilge:
By abstract type, do you mean the difference between `type INothing <: Nothing` and `type INothing = Nothing`?

smarter:
Yes.

smarter:
Or well, types were the lower and upper bound actually differ, the `INothing <: Nothing` thing should in theory behave exactly like `INothing = Nothing` although the compiler implementation might not respect that

armanbilge:
Well, FTR I just changed `INothing <: Nothing` to `INothing = Nothing` in that code I linked and it continues to grab the implicit from the `INothing` companion object.

PR #3907, opened 2021-05-26
Save a few allocations in Applicative.replicateA

joroKr21:
Cheat by creating an IndexedSeq that always returns the same element

 * Before: List -> ArrayBuffer -> Chain -> List
 * After: Chain -> List

### DIFF
--- a/core/src/main/scala/cats/Applicative.scala
+++ b/core/src/main/scala/cats/Applicative.scala
@@ -1,8 +1,10 @@
 package cats
 
 import cats.arrow.Arrow
+import cats.data.Chain
 import simulacrum.typeclass
 import scala.annotation.implicitNotFound
+import scala.collection.immutable.IndexedSeq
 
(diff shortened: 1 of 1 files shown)


diesalbla (approved):

core/src/main/scala/cats/Applicative.scala
```diff
@@ -65,7 +67,13 @@ import scala.annotation.implicitNotFound
    * }}}
    */
   def replicateA[A](n: Int, fa: F[A]): F[List[A]] =
-    Traverse[List].sequence(List.fill(n)(fa))(this)
+    if (n <= 0) pure(Nil)
+    else {
+      map(Chain.traverseViaChain(new IndexedSeq[F[A]] {
```
johnynek, line 72: Awesome idea!

johnynek (approved):

PR #3910, opened 2021-05-28
Add Kernel instances for java-time datatypes

diesalbla:
[Issue 3766](https://github.com/typelevel/cats/issues/3766) proposed integrating `cats-time` into `cats`, and adding instances for the data types in `java.time`, which is after all just as standard as the Scala library or UUID. 

All the instances, as defined in `cats-time`, just use the natural equality, hash code, or comparable implementations defined in the Java library. To ease the definition, within the companion objects of the typeclasses `Eq`, `Order`, and `Hash`, we define new "mixin" traits that provide implementations of each type-class based on the Any-hash, Any-eq, or Comparable.

With that, each class in `java.time` has an instance of the kernel typeclasses written in two lines.

### DIFF
--- a/core/src/main/scala-2.12/cats/instances/all.scala
+++ b/core/src/main/scala-2.12/cats/instances/all.scala
@@ -25,6 +25,7 @@ trait AllInstances
     with FutureInstances
     with HashInstances
     with InvariantMonoidalInstances
+    with JavaTimeInstances
     with ListInstances
     with MapInstances
     with OptionInstances
--- a/core/src/main/scala-2.13+/cats/instances/all.scala
+++ b/core/src/main/scala-2.13+/cats/instances/all.scala
@@ -27,6 +27,7 @@ trait AllInstances
     with FutureInstances
     with HashInstances
     with InvariantMonoidalInstances
+    with JavaTimeInstances
     with LazyListInstances
     with ListInstances
     with MapInstances
--- a/kernel/src/main/scala/cats/kernel/instances/JavaTimeInstances.scala
+++ b/kernel/src/main/scala/cats/kernel/instances/JavaTimeInstances.scala
@@ -1,0 +1,55 @@
+package cats.kernel.instances
+
+import cats.kernel.{CommutativeGroup, Eq, Hash, Order}
+import java.time._
+
+trait JavaTimeInstances {
+
+  implicit final val catsKernelStdOrderForduration
+    : Hash[Duration] with Order[Duration] with CommutativeGroup[Duration] =
+    new Order.FromComparable[Duration] with Hash.FromUniversal[Duration] with CommutativeGroup[Duration] {
+      override def empty: Duration = Duration.ZERO
+      override def combine(x: Duration, y: Duration): Duration = x.plus(y)
+      override def inverse(x: Duration): Duration = x.negated()
+    }
+
+  implicit final val catsKernelStdOrderForInstant: Order[Instant] with Hash[Instant] =
+    new Order.FromComparable[Instant] with Hash.FromUniversal[Instant] {}
+
+  implicit final val catsKernelStdEqForLocalDate: Eq[LocalDate] with Hash[LocalDate] =
+    new Eq.FromUniversal[LocalDate] with Hash.FromUniversal[LocalDate] {}
+
+  implicit final val catsKernelStdEqForLocalDateTime: Eq[LocalDateTime] with Hash[LocalDateTime] =
+    new Eq.FromUniversal[LocalDateTime] with Hash.FromUniversal[LocalDateTime] {}
+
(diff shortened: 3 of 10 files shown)


ChristopherDavenport (commented):
This is just in kernel? Should this have forwarders so cats imports get this automatically or am I missing them?

kernel/src/main/scala/cats/kernel/instances/JavaTimeInstances.scala
```diff
@@ -0,0 +1,55 @@
+package cats.kernel.instances
+
+import cats.kernel.{CommutativeGroup, Eq, Hash, Order}
+import java.time._
+
+trait JavaTimeInstances {
+
+  implicit final val catsKernelStdOrderForduration
+    : Hash[Duration] with Order[Duration] with CommutativeGroup[Duration] =
```
johnynek, line 9: we aren't testing the commutative group are we? I don't see where those laws are checked.

djspiewak:
Once this is compiling on ScalaJS, we need to double-check that downstream projects are not *forced* to take on the Java Time shim dependency.

kernel/src/main/scala/cats/kernel/instances/JavaTimeInstances.scala
diesalbla: The laws seem to be failing. I am removing that instance until that is fixed.

diesalbla: So, the problem I am having with some of the tests is that, if it generates very large values, it is causing overflows. Not sure how we should address this problem?

diesalbla: I was able to fix it by defining a generator for the `Duration` class itself. It is just copied from the one for `FiniteDuration`.

core/src/main/scala-2.12/cats/instances/all.scala
```diff
@@ -25,6 +25,7 @@ trait AllInstances
     with FutureInstances
     with HashInstances
     with InvariantMonoidalInstances
+    with JavaTimeInstances
```
diesalbla, line 28: > This is just in kernel? Should this have forwarders so cats imports get this automatically or am I missing them?

@ChristopherDavenport IIUC this should add the instances to the core itself too

core/src/main/scala-2.13+/cats/instances/all.scala
```diff
@@ -27,6 +27,7 @@ trait AllInstances
     with FutureInstances
     with HashInstances
     with InvariantMonoidalInstances
+    with JavaTimeInstances
```
diesalbla, line 30: > This is just in kernel? Should this have forwarders so cats imports get this automatically or am I missing them?

@ChristopherDavenport  I think adding this mixin should add the instances to the `core` instances as well.

diesalbla:
> Once this is compiling on ScalaJS, we need to double-check that downstream projects are not forced to take on the Java Time shim dependency.

The model followed here is similar to that of `java.util.UUID`, which is also included in the `kernel`, in the `main` branch.

ChristopherDavenport:
> Once this is compiling on ScalaJS, we need to double-check that downstream projects are not _forced_ to take on the Java Time shim dependency.

This is super super important. Time is an enormous increase to JS file size where the prospect of having it in a JS file is generally referred to as crazy by the community.

armanbilge:
I was pointed here by @joroKr21 in https://github.com/typelevel/cats/pull/3871#issuecomment-979730831.

Actually, if @diesalbla or someone else wants to, I think we can move forward with this. Here's how:
1. Move all the `java.time` instances into shared code (as is done in this PR).
2. _Don't_ add a dependency to scala-java-time. It's only needed for linking (not compiling) and only if `java.time` classes are actually used.
3. _Don't_ test `java.time` instances in the JS test suite. If the test suite still runs, this demonstrates that the linker was able to remove the (unused) references to `java.time`.
4. (Optional) Create another test (sbt) project/module with scala-java-time as a test dependency, and use it to test the `java.time` instances on JS.

What this achieves:
- if downstreams don't care about `java.time`, this PR will have no impact on them. They won't have scala-java-time on the classpath, and as long as they don't use `java.time` instances in their code the linker will not complain, as demonstrated in (2).
- If a downstream _does_ want to use `java.time` stuff, it's all there! They'll just have to explicitly add a dependency to scala-java-time (or any alternative shims of this stuff!) to get their code to link.

Risks:
- Because code using `java.time` will always compile (regardless of the availability of implementations for it to link against), having these instances will make it easier for libraries to unwittingly use this code without realizing the impact on their downstreams. I mean, this is true right now, because anyone can use `java.time` anytime anywhere, not test their code, and only find out a few downstreams later whey they link what consequences that has. The difference is that currently, the lack of these instances for Scala.js in cats might create enough of a compiler stumbling block that they get properly informed on the situation :)

PR #3918, opened 2021-06-04
Binary-compatibly move typelevel/algebra into cats repo

armanbilge:
Following directly from https://github.com/typelevel/cats/pull/3877#issuecomment-854998836.

Re-played commits from #3877:
- [x] 2c51138cd Initial dump of algebra into cats
- [x] 71f397b30 Disable MiMa _subsequently re-enabled_
- [x] 080a53910 @nowarn deprecations
- [x] 700c0b7cb Migrate docs
- [x] c51b72671 Reference cats-algebra in README/docs
- [ ] 66b807480 Rename package algebra -> cats.algebra
- [x] a6ec3b822 Merge DeMorganLaws into LogicLaws per TODO
- [x] 8ddf2a40a Merge in algebra#246
- [x] 1ce0efdde Remove unimplemented stubs
- [x] 39716dc13 Formatting
- [x] 513086720 Add kernel-laws dependency and refactor
- [x] 7ff1809c0 Delete priority _deprecated instead_
- [x] 048ce2ef1 Implement algebra#108
- [x] b81e1fbfc Make Field inherit from DivisionRing
- [ ] 844670d7d Rename laws -> tests, remove redundancies
- [x] 7ab3d9052 Add Signed and TruncatedDivision typeclasses
- [x] b0c1ced8a More laws for Signed
- [ ] 44eb36c88 Refactor instances to avoid conflicts/ambiguities


armanbilge:
Ok everyone, good news: I was able to rework @denisrosset's proposals typelevel/algebra#246 and typelevel/algebra#247 as well as typelevel/algebra#108 so that they are binary compatible 😅 

Unfortunately I had issues with https://github.com/typelevel/cats/commit/44eb36c88eac73ccbe8979a7ab6570562b398618 which tries to reduce potential for conflicts when using `import algebra.instances.all._`. It seems that marking as `private[algebra]` a trait method that is extended by a (package) object causes the static-forwarder for that method to disappear completely.

Still haven't brought in any of the significant changes to laws yet, but since it seems breaking bin-compat there could be okay they are less of a concern.

armanbilge:
@denisrosset @larsrh @johnynek @mpilquist Ready for review. Hopefully we can wrap this up? :) 

I decided not to make any binary-breaking changes to laws, doing so doesn't feel right under semantic versioning. Furthermore, I decided to keep everything in both core/laws source-compatible and added `@deprecation` warnings instead.

In the end I was able to replay most of the changes from the original PR. The glaring omissions are:

1. Currently, both `cats` and `algebra` instances inherit from `cats-kernel` instances. So importing both could cause conflicts/problems, possibly? It doesn't seem possible to fix this in a binary-compatible way.
2. The organization/naming scheme of `algebra-laws` is not very Cats-like (separation of laws and tests, etc.). Since breaking bincompat for `laws` seemed a possibility, I could still do this, but this would also mean breaking semantic versioning rules.

johnynek:
I'll look soon. Note: removing "implicit" is actually binary (although not source) compatible.

Since source incompatibilities are always discovered by compilation but binary are often discovered via runtime crashes in transitive dependencies, I would consider removing implicits that conflict to be acceptable.

armanbilge:
> Note: removing "implicit" is actually binary (although not source) compatible.

Thanks, I am aware of this! See my comment above https://github.com/typelevel/cats/pull/3918#issuecomment-855170288 about the problems I had with this. I don't think removing `implicit` would help without also marking as `private[algebra]`, because implicits and non-implicits alike get imported with the usual wildcard imports.

algebra-core/src/main/scala/algebra/Priority.scala
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * This type class will attempt to provide an implicit instance of `P`
+ * (the preferred type). If that type is not available it will
+ * fallback to `F` (the fallback type). If neither type is available
+ * then a `Priority[P, F]` instance will not be available.
+ *
+ * This type can be useful for problems where multiple algorithms can
+ * be used, depending on the type classes available.
+ */
+@deprecated("No replacement", since = "2.7.0")
```
johnynek, line 16: why is this deprecated? It seems it is not deprecated in algebra?

The motivation is that some functions can be written that e.g. use either a Group or a Monoid. If you have a Group, it can be more efficient, or else you can fall back to Monoid (at a constant multiplicative cost).

algebra-core/src/main/scala/algebra/Priority.scala
```diff
@@ -0,0 +1,71 @@
...
+    fold(_ => false)(_ => true)
+
+  def getPreferred: Option[P] =
+    fold[Option[P]](p => Some(p))(_ => None)
+
+  def getFallback: Option[F] =
+    fold[Option[F]](_ => None)(f => Some(f))
+}
+
+@deprecated("No replacement", since = "2.7.0")
```
johnynek, line 52: can we remove this?

algebra-laws/jvm/src/main/scala/algebra/laws/platform/Platform.scala
```diff
@@ -0,0 +1,10 @@
+package algebra.laws.platform
+
+private[laws] object Platform {
```
johnynek, line 3: since this is private, can we remove this and use:

https://github.com/typelevel/cats/blob/main/kernel-laws/jvm/src/main/scala/cats/platform/Platform.scala

docs/src/main/mdoc/algebra.md
```diff
@@ -0,0 +1,113 @@
...
+|--------------------------|-----------|-----------|-----------|----------|------------|----------|----------------|
+|JoinSemilattice           |          ✓|           |           |          |            |          |                |
+|MeetSemilattice           |           |          ✓|           |          |            |          |                |
+|BoundedJoinSemilattice    |          ✓|           |          ✓|          |            |          |                |
+|BoundedMeetSemilattice    |           |          ✓|           |         ✓|            |          |                |
+|Lattice                   |          ✓|          ✓|           |          |            |          |                |
+|DistributiveLattice       |          ✓|          ✓|           |          |           ✓|          |                |
+|BoundedLattice            |          ✓|          ✓|          ✓|         ✓|            |          |                |
+|BoundedDistributiveLattice|          ✓|          ✓|          ✓|         ✓|           ✓|          |                |
+|Heyting                   |          ✓|          ✓|          ✓|         ✓|           ✓|         ✓|                |
```
johnynek, line 104: I think Logic is missing from this table, btw...

johnynek (commented):
I'm mostly good. I had a couple of minor comments.

algebra-core/src/main/scala/algebra/Priority.scala
johnynek: Note: we could imagine pushing this onto `Either[A, B]` in the scala standard library (prefer Right, fallback to Left). Similarly with `Option[A]` and tuples, but I think the standard library currently does not allow this.

armanbilge: I was just replaying 7ff1809c065a020d269bc64e34a9a83f8e2c2417 which was suggested on the original PR since it's not used in algebra itself. My bad, now I see its intrinsic value. I'll bring it back.

Btw I did see your plans to submit it to Scala standard library in https://github.com/typelevel/algebra/issues/67!

algebra-laws/jvm/src/main/scala/algebra/laws/platform/Platform.scala
armanbilge: I would have liked to, but that one is also private! Since we don't share the same package. Ok to open the `cats-kernel` one up?

docs/src/main/mdoc/algebra.md
armanbilge: If you tell me which boxes to check I'll add it in real quick :) I'm ok with the first 5 columns or so, but a bit confused about the last two.

armanbilge:
@johnynek Thanks for the review, much appreciated. I've restored `Priority` to its rightful place.

@larsrh also expressed his overall approval, hopefully he can confirm here? :)

johnynek (approved):

larsrh:
I remain unconvinced that this one is better than #3877, but will not stand in the way of getting this done.

armanbilge:
Just a polite bump :) does anyone else want to chime in? @denisrosset or @mpilquist?

PR #3919, opened 2021-06-05
Opaque transformers types

diesalbla:
This Pull Request adds a implementation of the monad transformers `OptionT`, `IdT`, and `Kleisli`, for Scala 3. This new implementation uses the new feature of Scala 3, [Opaque Types](https://docs.scala-lang.org/scala3/book/types-opaque-types.html), which may allows us to have them implemented without overhead on the runtime memory.

This may slightly improve performance for libraries that are based on these monad transformers, such as `http4s`.


ChristopherDavenport (commented):
I can really tell how well this source aligns. Are there source incompatibilities introduced with this?

rossabaker:
This is similar to #3834.  I still like the idea, but I remain concerned about the maintenance burden.of two duplicate implementations.  Could we mitigate that with some tests that assure compatibility across Scala 2 and Scala 3?  Or maybe our test coverage is already sufficient?

core/src/main/scala-3.x/src/main/scala/cats/data/Kleisli.scala
```diff
@@ -0,0 +1,675 @@
...
+      * Contramap the input using `f`, where `f` may modify the input type of the Kleisli arrow.
+      * {{{
+      * scala> import cats.data.Kleisli, cats.implicits._
+      * scala> type ParseResult[A] = Either[Throwable, A]
+      * scala> val parseInt = Kleisli[ParseResult, String, Int](s => Either.catchNonFatal(s.toInt))
+      * scala> parseInt.local[List[String]](_.combineAll).run(List("1", "2"))
+      * res0: ParseResult[Int] = Right(12)
+      * }}}
+      */
+    def local[AA](f: AA => A): Kleisli[F, AA, B] = aa => x(f(aa))
```
johnynek, line 35: what about `f.andThen(x)`

core/src/main/scala-3.x/src/main/scala/cats/data/Kleisli.scala
```diff
@@ -0,0 +1,675 @@
...
+      a => F.map(x(a))(f)
+
+    /**
+      * Performs [[local]] and [[map]] simultaneously.
+      */
+    def dimap[C, D](f: C => A)(g: B => D): Kleisli[F, C, D] =
+      c => F.map(x(f(c)))(g)
+
+    def mapF[N[_], C](f: F[B] => N[C]): Kleisli[N, A, C] =
+      a => f(x(a))
```
johnynek, line 71: `x.andThen(f)`

johnynek (commented):
I think this is worth pursuing personally.

I think we should relax our binary compat for scala 3 to be able to merge something like this (as long as we can avoid any special casing in the tests and we expect source compatibility).

smarter:
> I think we should relax our binary compat for scala 3

There's many libraries published for scala 3 already that depends on cats, are they all OK with throwing bincompat out of the window for hypothetical performance benefits? IMO this belongs in a cats 3 branch or in https://github.com/typelevel/spotted-leopards

mpilquist:
Totally agree with @smarter on bin compatibility. If we wanted to get this in to cats 2.x, we'd have needed to do it prior to Scala 3 final release. I'd really like to see this included in the future though (perhaps Cats 3) along with built-in support for derivation (not via an add-on project).

armanbilge:
Btw, for anyone who cares about preserving bin-compat on Scala 3, we still have to enable MiMa for it. My PR needs feedback at https://github.com/typelevel/cats/pull/4063#issuecomment-984096068.

armanbilge:
I don't think this one is happening until Cats 3.

PR #3920, opened 2021-06-05
Delete methods and classes deprecated in the 1.0 release candidates.

diesalbla:
Cats 1.0 was released at the end of 2017, three and a half years ago. That should be time enough for a deprecation cycle, so we can now delete these methods.


ChristopherDavenport:
This would still be a binary incompatible change though? While we can agree that folks shouldn't be using these, if this is a 1.x line this is not a compatible change.

When the library goes 2.x this will be an exceptional PR.

rossabaker:
This should absolutely be done for the eventual Cats 3, but yeah, bincompat forbids this on Cats 2.

PR #3925, opened 2021-06-12
Add `ReferentialEq[A]` to Alley Cats

armanbilge:
This PR implements a helper `Eq.fromReferentialEquals[A]` to provide an instance of `Eq[A]` based on referential equality `eq`.

It is similar in spirit to `Eq.fromUniversalEquals[A]` and, as the [finest](https://en.wikipedia.org/wiki/Equivalence_relation#Comparing_equivalence_relations) possible equivalence relation, I believe makes it the dual of the coarsest `Eq.allEqual[A]`.

My use case is for fast equivalence comparisons in certain situations with nested case classes (where using universal equals could involve traversing the entire object hierarchy).

### DIFF
--- a/alleycats-core/src/main/scala/alleycats/ReferentialEq.scala
+++ b/alleycats-core/src/main/scala/alleycats/ReferentialEq.scala
@@ -1,0 +1,13 @@
+package alleycats
+
+import cats.Eq
+
+/**
+ * An `Eq[A]` that delegates to referential equality (`eq`).
+ * Note that it is not referentially transparent!
+ */
+object ReferentialEq {
+  def apply[A <: AnyRef]: Eq[A] = new Eq[A] {
+    def eqv(x: A, y: A) = x eq y
+  }
+}
--- a/alleycats-tests/shared/src/test/scala/alleycats/tests/ReferentialEqSuite.scala
+++ b/alleycats-tests/shared/src/test/scala/alleycats/tests/ReferentialEqSuite.scala
@@ -1,0 +1,33 @@
+package alleycats.tests
+
+import alleycats.ReferentialEq
+import cats.kernel.Eq
+import cats.kernel.laws.discipline._
+import cats.kernel.laws.EqLaws
+import org.scalacheck.Arbitrary
+import org.scalacheck.Prop.forAll
+import org.typelevel.discipline.Laws
+
+class ReferentialEqSuite extends AlleycatsSuite {
+
+  class ReferentialEqTests[A](eq: Eq[A]) extends Laws {
+    def laws = EqLaws(eq)
+
+    def eqv(implicit arbA: Arbitrary[A]): RuleSet = {
+      implicit val eqA: Eq[A] = laws.E
+      new DefaultRuleSet(
+        "referentialEq",
+        None,
+        "reflexivity eq" -> forAll(laws.reflexivityEq _),
+        "symmetry eq" -> forAll(laws.symmetryEq _),
+        "transitivity eq" -> forAll(laws.transitivityEq _)
+      )
+    }
+  }
+
+  implicit val arbObject: Arbitrary[Object] = Arbitrary(Arbitrary.arbUnit.arbitrary.map(_ => new Object))
+  implicit val eqObject: Eq[Object] = ReferentialEq[Object]
+
+  checkAll("ReferentialEq[Object]", new ReferentialEqTests(ReferentialEq[Object]).eqv)
+
+}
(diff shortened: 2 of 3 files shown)


armanbilge:
Now that I've publicly embarrassed myself, obviously `eq` isn't referentially transparent so none of this is valid. Thanks to @ChristopherDavenport for politely pointing this out to me on Discord.

armanbilge:
... does Alley Cats want it? 😁

djspiewak:
IMO this makes sense for Alleycats. 😃 Though I will note that, in most cases, if you're using `eq` you probably care a lot about the performance you would sacrifice when using `Eq`, so the intersection of use cases might be small.

armanbilge:
Cool! Re-opening as PR adding `ReferentialEq` to Alley Cats :)

> if you're using `eq` you probably care a lot about the performance you would sacrifice when using `Eq`

Quite possibly! However, `Eq` is an entrypoint to various Cats-based APIs, so those would be off the table too. In my case the performance of `Eq` is negligible compared to traversing a graph of hundreds/thousands of objects 😅

alleycats-core/src/main/scala/alleycats/ReferentialEq.scala
```diff
@@ -0,0 +1,13 @@
...
+import cats.Eq
+
+/**
+ * An `Eq[A]` that delegates to referential equality (`eq`).
+ * Note that it is not referentially transparent!
+ */
+object ReferentialEq {
+  def apply[A <: AnyRef]: Eq[A] = new Eq[A] {
+    def eqv(x: A, y: A) = x eq y
+  }
```
johnynek, line 12: can we add some tests? I think this actually passes the laws associated with Eq except `antiSymmetryEq`.

Maybe we could test the other three in an ad-hoc fashion?

johnynek: see: https://github.com/typelevel/cats/blob/main/kernel-laws/shared/src/main/scala/cats/kernel/laws/EqLaws.scala

armanbilge: @johnynek good point, you are right! I just added it

alleycats-tests/shared/src/test/scala/alleycats/tests/ReferentialEqSuite.scala
```diff
@@ -0,0 +1,33 @@
...
+        "referentialEq",
+        None,
+        "reflexivity eq" -> forAll(laws.reflexivityEq _),
+        "symmetry eq" -> forAll(laws.symmetryEq _),
+        "transitivity eq" -> forAll(laws.transitivityEq _)
+      )
+    }
+  }
+
+  implicit val arbObject: Arbitrary[Object] = Arbitrary(Arbitrary.arbUnit.arbitrary.map(_ => new Object))
```
johnynek, line 28: nit: this will always make distinct objects, what about:

```scala
Arbitrary(
  Gen.oneOf(
    Gen.oneOf(List.fill(5)(new Object)),
    Arbitrary.arbUnit.arbitrary.map(_ => new Object)
  )
)
```
So with some probability we select from a small set of objects, otherwise make a totally new one.

armanbilge: Clever!

Hossein2525Hp121212 (approved):

alleycats-tests/shared/src/test/scala/alleycats/tests/ReferentialEqSuite.scala
armanbilge: Also not a nit, otherwise we aren't fully checking some of the laws right?

johnynek (approved):

SimY4:
Do we want to add `IdentityHash` to complement `ReferentialEq`? 🤔 
Something like:
```
object IdentityHash {
  def apply[A <: AnyRef]: Hash[A] = new Hash[A] { def hash(x: A): Int = java.lang.System.identityHashCode(x);  }
}
```

armanbilge:
@SimY4 Seems reasonable to me. Although not sure about the name, it's consistent but sounds confusing at least to me. I'd prefer `SystemIdentityHash`.

Now I regret not making my `ReferentialEq` and `ReferentialEqTests` more extensible. You should factor those out as so you can extend them. Also you'll probably want to reuse the arbitrary `Object` code as well. Sorry about that!

SimY4:
@armanbilge It's not yet released so we can move things around, just saying. 😀

armanbilge:
@SimY4 please PR! :)

SimY4:
@armanbilge Looked into it a bit more and found that referential eq already exist in cats through Hash.fromUniversalHashCode here: https://github.com/typelevel/cats/blob/main/kernel/src/main/scala/cats/kernel/Hash.scala#L51

armanbilge:
@SimY4 actually no! Easy mistake though. On Java `==` is referential equality (cannot override) and `x.equals(y)` is universal equality (objects can override this; default is referential equality). On Scala, `==` is universal equality and `eq` is referential equality. Similar for `hashCode` and `System.identityHashCode`.

PR #3935, opened 2021-07-07
Add NonEmptyList.min and NonEmptyList.max

JannikArndt:
This adds `.min` and `.max` to `NonEmptyList[A]` if there's an Ordering for `A`.

---------

Background: We ran into the issue that scapegoat complained that `List(a, b, c).min` isn't safe. Since we _know_ the list is not empty, we tried `NonEmptyList`, but `min` and `max` don't exist there. 

The implementation uses `reduceLeft`, same as the standard lib for `IterableOnce`: https://github.com/scala/scala/blob/262c97250a614bf19a0e9e8994ee59abea82a5db/src/library/scala/collection/IterableOnce.scala#L912-L916

Note: I did run `+prePR`, and it failed in `free/src/main/scala-3.x/cats/free/FreeFoldStep.scala`, but not in any file touched by this PR.

### DIFF
--- a/core/src/main/scala/cats/data/NonEmptyList.scala
+++ b/core/src/main/scala/cats/data/NonEmptyList.scala
@@ -388,6 +388,29 @@ final case class NonEmptyList[+A](head: A, tail: List[A]) extends NonEmptyCollec
     NonEmptyList.fromListUnsafe(toList.sorted(AA.toOrdering))
 
   /**
+   * Returns the minimum value of this `NonEmptyList` according to an `Ordering`
+   *
+   * {{{
+   * scala> import cats.data.NonEmptyList
+   * scala> val nel = NonEmptyList.of(12, 4, 3, 9)
+   * scala> nel.min
+   * res0: Int = 3
+   * }}}
+   */
+  def min[AA >: A](implicit ord: Ordering[AA]): A = reduceLeft(ord.min[A])
+  /**
+   * Returns the maximum value of this `NonEmptyList` according to an `Ordering`
+   *
+   * {{{
+   * scala> import cats.data.NonEmptyList
+   * scala> val nel = NonEmptyList.of(12, 4, 3, 9)
+   * scala> nel.max
+   * res0: Int = 12
+   * }}}
+   */
+  def max[AA >: A](implicit ord: Ordering[AA]): A = reduceLeft(ord.max[A])
+
+  /**
    * Groups elements inside this `NonEmptyList` according to the `Order`
    * of the keys produced by the given mapping function.
    *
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -387,6 +387,30 @@ final case class NonEmptyList[+A](head: A, tail: List[A]) extends NonEmptyCollec
...
+   * Returns the minimum value of this `NonEmptyList` according to an `Ordering`
+   *
+   * {{{
+   * scala> import cats.data.NonEmptyList
+   * scala> val nel = NonEmptyList.of(12, 4, 3, 9)
+   * scala> nel.min
+   * res0: Int = 3
+   * }}}
+   */
+  def min[AA >: A](implicit ord: Ordering[AA]): A = reduceLeft(ord.min[A])
```
johnynek, line 400: This should be a cats.Order.

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -387,6 +387,30 @@ final case class NonEmptyList[+A](head: A, tail: List[A]) extends NonEmptyCollec
...
+   * Returns the maximum value of this `NonEmptyList` according to an `Ordering`
+   *
+   * {{{
+   * scala> import cats.data.NonEmptyList
+   * scala> val nel = NonEmptyList.of(12, 4, 3, 9)
+   * scala> nel.max
+   * res0: Int = 12
+   * }}}
+   */
+  def max[AA >: A](implicit ord: Ordering[AA]): A = reduceLeft(ord.max[A])
```
johnynek, line 412: cats.Order

johnynek (changes requested):

tests/src/test/scala/cats/tests/NonEmptyListSuite.scala
```diff
@@ -321,6 +321,18 @@ class NonEmptyListSuite extends NonEmptyCollectionSuite[List, NonEmptyList, NonE
...
+      assert(nel.min === nel.toList.min)
+    }
+  }
+
+  test("NonEmptyList#max is consistent with List#max") {
+    forAll { (nel: NonEmptyList[Int]) =>
+      assert(nel.max === nel.toList.max)
+    }
+  }
+
```
rossabaker, line 335: After @posco's suggestions, which I agree with, these properties assume `Order` and `Ordering` are consistent.  That holds for `Int`, and I can't think of a counterexample.  But that assumption might be worth a comment.

Vaguely related discussion: #3936.

JannikArndt:
#3937 solves the problem in a nicer way

PR #3936, opened 2021-07-07
Add implicit OrderFromOrderingConversion

saeltz:
https://github.com/typelevel/cats/pull/3935 wants to add `min` and `max` to `NonEmptyList`.

I guess we already have these in `Reducible` if you import `cats.syntax.all._` and provide a `cats.Order[A]`.

This can be achieved using `cats.Order.fromOrdering` but it would be so much easier for users of the library to have an implicit conversion between `Order[A]` and `Ordering[A]`.

This PR adds this in a new `trait` that extends `cats.Order`.
One could debate whether this should always be available or a separate import.

### DIFF
--- a/kernel/src/main/scala/cats/kernel/Order.scala
+++ b/kernel/src/main/scala/cats/kernel/Order.scala
@@ -125,7 +125,16 @@ trait OrderToOrderingConversion {
 
 }
 
-object Order extends OrderFunctions[Order] with OrderToOrderingConversion {
+trait OrderFromOrderingConversion {
+
+  /**
+   * Implicitly derive a `Order[A]` from a `scala.math.Ordering[A]`
+   * instance.
+   */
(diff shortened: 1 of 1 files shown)


rossabaker:
Authors of Scala `Ordering`s aren't bound by Cats' `Order`'s laws.  I suspect nearly all of them are lawful, but given Cats' stronger laws, I definitely don't think they should be implicitly converted with the standard imports.

Do you have an example where implicits would be better than a handful `fromOrdering` calls?  I wonder whether this problem might be solved by adding Cats instances to some library giving you trouble.

saeltz:
Lawfulness is a good point. I opened this primarily to start the discussion.

We were missing instances for `Deadline`. I just opened another PR https://github.com/typelevel/cats/pull/3937 to add that.
I would appreciate your feedback over there and would close this one then.

rossabaker:
Closing in favor of #3937.

PR #3937, opened 2021-07-08
Add DeadlineInstances

saeltz:
We wanted to have `min` and `max` on a `NonEmptyList`. That requires a `cats.Order[A]`.
For the standard library's `Deadline` no instances were defined. This adds it.

(I'm sure I need to touch some other places to make it available everywhere. Please tell me where. Thanks!)

### DIFF
--- a/kernel/src/main/scala/cats/kernel/instances/DeadlineInstances.scala
+++ b/kernel/src/main/scala/cats/kernel/instances/DeadlineInstances.scala
@@ -1,0 +1,34 @@
+package cats.kernel
+package instances
+
+import java.util.concurrent.TimeUnit
+import scala.concurrent.duration.{Deadline, FiniteDuration}
+
+trait DeadlineInstances {
+  implicit val catsKernelStdOrderForDeadline
+    : Order[Deadline] with Hash[Deadline] with LowerBounded[Deadline] with UpperBounded[Deadline] = new DeadlineOrder
+}
+
+trait DeadlineBounded extends LowerBounded[Deadline] with UpperBounded[Deadline] {
+  override def minBound: Deadline = Deadline(FiniteDuration(-Long.MaxValue, TimeUnit.NANOSECONDS))
+  override def maxBound: Deadline = Deadline(FiniteDuration(Long.MaxValue, TimeUnit.NANOSECONDS))
+}
+
+class DeadlineOrder extends Order[Deadline] with Hash[Deadline] with DeadlineBounded { self =>
+
+  def hash(x: Deadline): Int = x.hashCode()
+  def compare(x: Deadline, y: Deadline): Int =
+    if (x == y) 0 else if (x > y) 1 else -1
+
+  override def eqv(x: Deadline, y: Deadline): Boolean = x == y
+  override def neqv(x: Deadline, y: Deadline): Boolean = x != y
+  override def gt(x: Deadline, y: Deadline): Boolean = x > y && x != y
+  override def lt(x: Deadline, y: Deadline): Boolean = x < y && x != y
+  override def gteqv(x: Deadline, y: Deadline): Boolean = x == y || x > y
+  override def lteqv(x: Deadline, y: Deadline): Boolean = x == y || y > x
+
+  override def min(x: Deadline, y: Deadline): Deadline = if (x < y) x else y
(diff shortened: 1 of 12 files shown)


rossabaker:
What's here looks like a good start!

Search for `FiniteDurationInstances`, and I would imagine it'll be in exactly the same places.  I think that will include an `all` package object and maybe a new `deadline` package object for a la carte imports.

See also `FiniteDurationSuite`.  We'll want a similar `DeadlineSuite` to check the laws.

saeltz:
@rossabaker, I added the `package object`s, all `trait`s and included this in `all`.

The `FiniteDurationSuite` only tests for the `Show` typeclass which is not available for `Deadline`.
The tests for `FiniteDuration` are in the `OrderSuite` where I added some tests which are passing on JVM but failing on Scala.js. Any hints on how to prevent that?

saeltz:
@rossabaker, I fixed all the tests and binary compatibility issues. Please review again. Thanks!

kernel/src/main/scala/cats/kernel/instances/DeadlineInstances.scala
```diff
@@ -0,0 +1,34 @@
...
+  implicit val catsKernelStdOrderForDeadline
+    : Order[Deadline] with Hash[Deadline] with LowerBounded[Deadline] with UpperBounded[Deadline] = new DeadlineOrder
+}
+
+trait DeadlineBounded extends LowerBounded[Deadline] with UpperBounded[Deadline] {
+  override def minBound: Deadline = Deadline(FiniteDuration(-Long.MaxValue, TimeUnit.NANOSECONDS))
+  override def maxBound: Deadline = Deadline(FiniteDuration(Long.MaxValue, TimeUnit.NANOSECONDS))
+}
+
+class DeadlineOrder extends Order[Deadline] with Hash[Deadline] with DeadlineBounded { self =>
```
johnynek, line 17: I think a BoundedSemilattice may also make sense here (on max?) or maybe it is unclear so we should let users make their own.

saeltz: It definitely is unclear to me how to implement that. So I'd appreciate either an approval or some help.

rossabaker: The `BoundedSemilattice` is a `Monoid`, so you'd need to define `empty` and `combine`.  You also inherit these typeclasses, and therefore laws:

* a `CommutativeSemigroup`: `combine(x, y) === combine(y, x)`
* a `Band`: `combine(x, x) == x`

If you combine two deadlines, would you want the result to be the earlier deadline (`min`) or the later (`max`)?  I can imagine both, so I'm a bit skeptical of the instance.

And what should `empty` be?  `Deadline.now` is tempting, but changes on each call.  I suppose we'd use `minBound` if we combined with `max`, and `maxBound` if we combined with `min`.  Or we could escape this question by being a `Semilattice` instead of a `BoundedSemilattice`.

I think if someone can make a persuasive case for min or max being the canonical instance, we could handle this in a separate PR.

rossabaker (approved):
:+1:

kernel/src/main/scala/cats/kernel/instances/DeadlineInstances.scala
johnynek: empty would be either the minimum or maximum value depending on which BoundedSemilattice you pick. `now` would not be lawful.

johnynek: Note: we already have a constructor function for picking a max or min semilattice if you give the corresponding empty value. I agree it isn't clear if you want min or max here.

Probably some applications would want one and others the other.

saeltz: If it is dependent on the application, I would refrain from implementing it here and merging this PR.
Later one could implement the `Semilattice` in a follow-up PR anyways.

rossabaker: I'm pretty well convinced the `empty` follows from `combine` and any eventual instance should be monoidal.  But there's no obvious canonical instance, and no reason to block this waiting for one.  Let's get a second approval and get this shipped.

johnynek (approved):

PR #3938, opened 2021-07-08
add methods for creating generic errors from string

yangzai:
Getting quick-and-dirty errors from string error messages is a pretty common use case. This PR is similar to `asLeft`/`asRight` and `some`/`none` but not widened for `Error`/`Exception` subtypes as they are not ADTs.

Instead of writing this:
```scala
val leftError = Left(new Error("error"))
val ioThrow = new Throwable("error").raiseError[IO, Unit]
```
this PR lets you do this:
```scala
val leftError = "error".asError.asLeft
val ioThrow = "error".asThrowable.raiseError[IO, Unit]
```
Saves you from writing the `new` keyword and parentheses.


rossabaker:
Thanks!  I think this might fit better in [Mouse](https://typelevel.com/typelevel/mouse).

yangzai:
> Thanks! I think this might fit better in [Mouse](https://typelevel.com/typelevel/mouse).

@rossabaker sure, I will move it over.

rossabaker:
Closed in favor of Mouse PR.

PR #3939, opened 2021-07-08
Add missing import in string |+| example

LannyRipple:
String |+| example discussion mentions "the second import" but
that import is missing.

### DIFF
--- a/docs/src/main/mdoc/typeclasses/imports.md
+++ b/docs/src/main/mdoc/typeclasses/imports.md
@@ -45,6 +45,7 @@ For example, if you'd like to import the `Monoid` instance for `String` and the 
 
 ```scala mdoc
 import cats.instances.string._
+import cats.syntax.semigroup._
 
 "Hello, " |+| "World!"
 ```


djspiewak:
Is this preferable or should we encourage people to `import cats.syntax.all._`?

ChristopherDavenport:
I think `import cats.syntax.all._` should always be preferred. Memorizing all the little syntax blocks is too much, and adds a mental barrier.

LannyRipple:
A counter argument would be that the description says, "while the second
import adds the |+| syntax",
and not "pulls in all syntax".

Oh.  I just noticed the next line of description in that area says

You can also import all syntax or all instances by importing
cats.syntax.all._ or cats.instances.all._ respectively.
so that usage seems to be already covered.

My $0.02 would be for being specific with syntax.semigroup.

YMMV,
  -ljr

LannyRipple:
This is what I get for not proof-reading.

  My $0.02 would be for being specific with syntax.semigroup since this is
the import guide.

-ljr

rossabaker:
Is the `import cats.instances.string._` even necessary anymore?

LannyRipple:
That's bringing in the semigroup instance on String isn't it?  For the
example to work and description to make sense I think we need both

   import cats.instances.string._
   import cats.syntax.semigroup._

-ljr

djspiewak:
> That's bringing in the semigroup instance on String isn't it?  For the
example to work and description to make sense I think we need both

`Semigroup[String]` is always in scope now. (since Cats 2.3, I believe) There's no need to import for anything other than syntax.

rossabaker (commented):
So I think the thrust of this PR is correct: it's showing the a la carte method described in the paragraph above it.  And mdoc didn't catch it because we did the full import earlier in the doc.

I think the `cats.instances.string._` example is poor now that it's automatic.  (Can we think of one that's not?)  We should replace that, fix up the paragraph in line 43, and keep @LannyRipple's suggestion on the a la carte import.

LannyRipple:
Agree that string isn't a good example now that Semigroup[String] is always
in scope.  The bug in the docs has been there for a few years.  Just
finally got around to reading it again and wanting to fix it.

Short of something better than the string example my suggestion would be to
show something like

   import cats.syntax.semigroup._

    "Hello" |+| ", World!"

With a description along the lines of

   The above example shows how to access a symbolic operator, in this case
for semigroup combine.
   The import brings in the operator definition which can then be used on
the strings.  (Semigroup[String]
   is now always in scope so there's no need to bring it into scope
directly with import cats.instances.strings._.)

armanbilge:
Ok, I did a botched merge trying to freshen this PR up ...

... but I also realized, it's already fixed on the published site? Not sure when that happened.
https://typelevel.org/cats/imports.html

Anyway, thanks for bringing this to our attention!

PR #3943, opened 2021-07-15
Use new sbt syntax for fmt commands

danicheg:
This is a minor thing but cleans up warnings in the console.


djspiewak (approved):

rossabaker (approved):
Workflow hadn't run, but :+1: assuming green.

PR #3944, opened 2021-07-17
add identity hash to alleycats

SimY4:
This is a follow-up PR based on the discussion in https://github.com/typelevel/cats/pull/3925

To compliment ReferentialEq I've added SystemIdentityHash to JVM alleycats.

### DIFF
--- a/alleycats-core/.jvm/src/main/scala/alleycats/SystemIdentityHash.scala
+++ b/alleycats-core/.jvm/src/main/scala/alleycats/SystemIdentityHash.scala
@@ -1,0 +1,13 @@
+package alleycats
+
+import cats.Hash
+
+class SystemIdentityHash[A <: AnyRef] extends ReferentialEq[A] with Hash[A] {
+  override def hash(a: A): Int = java.lang.System.identityHashCode(a)
+}
+
+object SystemIdentityHash {
+  private[this] val identityHash: Hash[AnyRef] = new SystemIdentityHash[AnyRef]
+
+  def apply[A <: AnyRef]: Hash[A] = identityHash.asInstanceOf[Hash[A]]
+}
--- a/alleycats-core/src/main/scala/alleycats/ReferentialEq.scala
+++ b/alleycats-core/src/main/scala/alleycats/ReferentialEq.scala
@@ -6,12 +6,12 @@ import cats.Eq
  * An `Eq[A]` that delegates to referential equality (`eq`).
  * Note that it is not referentially transparent!
  */
-class ReferentialEq[A <: AnyRef] extends Eq[A] {
-  def eqv(x: A, y: A) = x eq y
+trait ReferentialEq[A <: AnyRef] extends Eq[A] {
+  def eqv(x: A, y: A): Boolean = x eq y
 }
 
 object ReferentialEq {
-  private[this] val referentialEq: Eq[AnyRef] = new ReferentialEq[AnyRef]
+  private[this] val referentialEq: Eq[AnyRef] = new ReferentialEq[AnyRef] {}
 
   def apply[A <: AnyRef]: Eq[A] = referentialEq.asInstanceOf[Eq[A]]
 }
--- a/alleycats-core/src/main/scala/alleycats/ReferentialEq.scala
+++ b/alleycats-core/src/main/scala/alleycats/ReferentialEq.scala
@@ -6,8 +6,12 @@ import cats.Eq
  * An `Eq[A]` that delegates to referential equality (`eq`).
  * Note that it is not referentially transparent!
  */
+class ReferentialEq[A <: AnyRef] extends Eq[A] {
+  def eqv(x: A, y: A) = x eq y
+}
+
 object ReferentialEq {
-  def apply[A <: AnyRef]: Eq[A] = new Eq[A] {
-    def eqv(x: A, y: A) = x eq y
-  }
+  private[this] val referentialEq: Eq[AnyRef] = new ReferentialEq[AnyRef]
+
+  def apply[A <: AnyRef]: Eq[A] = referentialEq.asInstanceOf[Eq[A]]
 }
--- a/alleycats-core/src/main/scala/alleycats/SystemIdentityHash.scala
+++ b/alleycats-core/src/main/scala/alleycats/SystemIdentityHash.scala
@@ -2,12 +2,12 @@ package alleycats
 
 import cats.Hash
 
-class SystemIdentityHash[A <: AnyRef] extends ReferentialEq[A] with Hash[A] {
+trait SystemIdentityHash[A <: AnyRef] extends ReferentialEq[A] with Hash[A] {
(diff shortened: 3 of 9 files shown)


armanbilge (changes requested):
Just saw this, nice addition and clean-up on my PR! I don't think anything is supposed to go in the `.jvm` folder. The correct way to handle this is to change the build to a full cross with a proper `jvm` folder. Is `System.identityHash` not available on Scala.js and/or Native?

Also in your clean up efforts you removed `ReferentialEqTests`, I'm surprised the test suite passed (must be chance?) since it actually doesn't satisfy all the `Eq` laws.

alleycats-core/.jvm/src/main/scala/alleycats/SystemIdentityHash.scala
```diff
@@ -0,0 +1,13 @@
+package alleycats
+
+import cats.Hash
+
+class SystemIdentityHash[A <: AnyRef] extends ReferentialEq[A] with Hash[A] {
```
armanbilge, line 5: Maybe just me, but perhaps make it a trait so that it can be mixed with other instances?

alleycats-core/src/main/scala/alleycats/ReferentialEq.scala
```diff
@@ -6,8 +6,12 @@ import cats.Eq
  * An `Eq[A]` that delegates to referential equality (`eq`).
  * Note that it is not referentially transparent!
  */
+class ReferentialEq[A <: AnyRef] extends Eq[A] {
```
armanbilge, line 9: Same as above: trait maybe?

alleycats-tests/shared/src/test/scala/alleycats/tests/ReferentialEqSuite.scala
```diff
@@ -3,42 +3,9 @@ package alleycats.tests
...
 import cats.kernel.laws.discipline._
-import cats.kernel.laws.EqLaws
-import org.scalacheck.Arbitrary
-import org.scalacheck.Gen
-import org.scalacheck.Prop.forAll
-import org.typelevel.discipline.Laws
 
 class ReferentialEqSuite extends AlleycatsSuite {
-
-  class ReferentialEqTests[A](eq: Eq[A]) extends Laws {
```
armanbilge, line 14: Wait a minute: where did this go? Referential eq doesn't pass all the laws of `Eq`.

armanbilge (changes requested):

alleycats-tests/shared/src/test/scala/alleycats/tests/ReferentialEqSuite.scala
armanbilge: See https://github.com/typelevel/cats/pull/3925#discussion_r669111963

armanbilge:
I checked and `System.identityHashCode` is available on both Scala.js and Native.

SimY4:
@armanbilge I may have messed something up with my test suite but it seem to pass all the eq laws.

alleycats-core/.jvm/src/main/scala/alleycats/SystemIdentityHash.scala
SimY4: I don't mind either way. I'll modify the change

alleycats-tests/shared/src/test/scala/alleycats/tests/ReferentialEqSuite.scala
SimY4: @armanbilge I may have messed something up but it seem to me that all laws have passed now. I see them all being green on my branch.

armanbilge: @johnynek can probably explain it better than me, but IIUC:

Essentially, the anti-symmetry law states that `a === b <=> f(a) === f(b)` where `f` is a pure function. The definition of a pure function is that it is referentially transparent i.e. `f(a) === f(a)`. But, because we are implementing referential equality this may not hold. Example:
```scala
def prepend1(list: List[Int]) = 1 :: list

prepend1(Nil) == prepend1(Nil) // true
prepend1(Nil) eq prepend1(Nil) // false
```

Thus, we drop this law.

I'm not sure how/why the test suite is passing, but I suspect something is not quite right.

alleycats-tests/shared/src/test/scala/alleycats/tests/AlleycatsSuite.scala
```diff
@@ -20,7 +19,21 @@ trait AlleycatsSuite extends munit.DisciplineSuite with TestSettings with TestIn
...
+        Arbitrary.arbUnit.arbitrary.map(_ => new Object)
+      )
+    )
+
+  implicit val arbObjectF: Arbitrary[Object => Object] =
+    Arbitrary {
+      for {
+        obj <- arbitrary[Object]
+      } yield _ => obj
+    }
```
armanbilge, line 38: @SimY4 this is why the antisymmetry laws pass in your tests. This creates an arbitrary function `Object => Object` that is in fact a constant (i.e., does not depend on the input). Therefore `f(a) eq f(b)` for any choice of `a` or `b` and the law seems to pass.

If you instead use a function that depends on the input e.g. `(x: Object) => Some(x)` then the law will fail.

johnynek: yes, I agree we should not use this arbitrary to test these instances. They are not lawful for the reasons mentioned in https://github.com/typelevel/cats/pull/3944/files#r677021500 so, that's a counter example.

alleycats-core/src/main/scala/alleycats/SystemIdentityHash.scala
```diff
@@ -0,0 +1,13 @@
+package alleycats
+
+import cats.Hash
+
+trait SystemIdentityHash[A <: AnyRef] extends ReferentialEq[A] with Hash[A] {
+  override def hash(a: A): Int = java.lang.System.identityHashCode(a)
+}
```
armanbilge, line 7: Maybe add a brief scaladoc? :)

armanbilge (dismissed):
Thanks for addressing all the PR comments, nice work! ✅ but please add a quick scaladoc to `SystemIdentityHash`.

SimY4:
The thing that I haven't implemented but could be potential optimisation is: merge both ReferentialEq and SystemIdentityHash into one alleycats definition since they are very closely related.

armanbilge:
> merge both ReferentialEq and SystemIdentityHash into one alleycats definition since they are very closely related

I think it's fine to have them as separate traits, more modular :) As it was pointed out in the original PR https://github.com/typelevel/cats/pull/3925#issuecomment-860050951 using typeclasses isn't particularly performant anyway!

johnynek (approved):

PR #3949, opened 2021-07-23
Added implementation of Eq[Free]

djspiewak:
Look ma, no documentation! Okay that's not *entirely* true, but it's close to true. Documentation coming.

So this probably looks a little odd if you're unfamiliar with recursion schemes. `Free` *is* a fixed point (similar to `Fix` or `Mu`), but we don't normally think of it that way. IMO `Free` is a lot more useful in this role than `Fix` is, and this `Eq` instance is one of the first steps in being able to unlock this power on top of vanilla Cats (rather than requiring something like Droste for even trivial cases).

Anyway, examples to come. There are some nice ones of how this can be leveraged in practice, I just haven't written them down yet.

Poke @Baccata since I promised you this a while back.

### DIFF
--- a/free/src/test/scala-2.13+/cats/free/FreeStructuralSuite.scala
+++ b/free/src/test/scala-2.13+/cats/free/FreeStructuralSuite.scala
@@ -1,0 +1,25 @@
+package cats.free
+
+import cats.Functor
+import cats.kernel.laws.discipline.EqTests
+import cats.tests.CatsSuite
+
+import org.scalacheck.Cogen
+
+// this functionality doesn't exist on Scala 2.12
+class FreeStructuralSuite extends CatsSuite {
+  import FreeSuite.freeArbitrary
+
+  implicit def freeCogen[S[_]: Functor, A](implicit S: => Cogen[S[Free[S, A]]], A: Cogen[A]): Cogen[Free[S, A]] =
+    Cogen { (seed, f) =>
+      f.resume match {
+        case Left(sf) =>
+          S.perturb(seed, sf)
+
+        case Right(a) =>
+          A.perturb(seed, a)
+      }
+    }
+
+  checkAll("Free[Option, Int]", EqTests[Free[Option, Int]].eqv)
+}
--- a/free/src/test/scala-2.13+/cats/free/FreeStructuralSuite.scala
+++ b/free/src/test/scala-2.13+/cats/free/FreeStructuralSuite.scala
@@ -1,7 +1,7 @@
 package cats.free
 
 import cats.Functor
-import cats.kernel.laws.discipline.EqTests
+import cats.kernel.laws.discipline.{/*HashTests,*/ PartialOrderTests}
 import cats.tests.CatsSuite
 
 import org.scalacheck.Cogen
@@ -21,5 +21,7 @@ class FreeStructuralSuite extends CatsSuite {
       }
     }
 
-  checkAll("Free[Option, Int]", EqTests[Free[Option, Int]].eqv)
+  // TODO HashLaws#sameAsUniversalHash is really dodgy
+  // checkAll("Free[Option, Int]", HashTests[Free[Option, Int]].hash)
+  checkAll("Free[Option, Int]", PartialOrderTests[Free[Option, Int]].partialOrder)
 }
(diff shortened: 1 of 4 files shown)


LukaJCB (dismissed):
lgtm! Might want to test it with something other than option as well perhaps? not a blocker though

djspiewak:
@LukaJCB I'm going to add a couple other instances actually. Can also test with something other than `Option` if you like.

Hosseinjoon2020 (approved):

djspiewak:
@LukaJCB I think this is ready for a final look

free/src/test/scala-2.13+/cats/free/FreeStructuralSuite.scala
```diff
@@ -0,0 +1,133 @@
...
+        case Right(a) =>
+          A.perturb(seed, a)
+      }
+    }
+
+  Show[Free[Option, Int]]
+
+  // TODO HashLaws#sameAsUniversalHash is really dodgy
+  // checkAll("Free[Option, Int]", HashTests[Free[Option, Int]].hash)
+  checkAll("Free[Option, Int]", PartialOrderTests[Free[Option, Int]].partialOrder)
```
johnynek, line 30: is it the case that `f1: Free[Option, Int] == f2: Free[Option, Int] <=> f1.runTailRec == f2.runTailRec`?

That seems like a nice property to have, but I don't see that if the current code can get it.

If it can't get it, what does this notion of equality mean?

johnynek: I guess if I read the docs I might learn something...

So, I guess the property I mention is not true, but one way: if we are "free-equal" then we are option-equal.

That might be a nice test to add to strengthen the tests here (e.g. I think returning that all items are equivalent always passes the laws, but isn't what we really want, adding a law I suggest here rules out many bad implementations since happening to get equal `Option[Int]` is unlikely.

djspiewak: > So, I guess the property I mention is not true, but one way: if we are "free-equal" then we are option-equal.

I *believe* this is the case. Caffeine is still working its way into my brain, but at first glance, I think this property holds*at least* because `Either` is monotone with respect to `Eq` (meaning that `a1 === a2` implies `Right(a1) === Right(a2)` and `Left(a1) === Left(a2)` *and* the converse), which means that this would reduce to an inductive argument over that monotonicity. But, as I'll argue in a minute, this restriction on the pattern functor is probably not needed.

The stronger bidirectional implication definitely doesn't hold, since `Free.pure(42).runTailRec === Free.suspend(Some(42)).runTailRec`, but they would *not* be considered equal by the structural definition of `Eq`. You can build versions of this which don't rely on coyoneda as well. Interpreting a `Free` into a monad is a stronger process than simply inspecting its defining structure, which is what this `Eq` does. In a sense, this is the reification of the intuition that, for almost any value, there are many different programs which can produce that outcome. The structural `Eq` inspects the *program*, while `runTailRec` inspects the *outcome*.

What you're saying is that, for any *programs* that are structurally equal, the *outcomes* they produce when run must also be equal. I believe this holds for all suspension functors which are deterministic monads (e.g. `race` makes `IO` a nondeterministic monad, which is why it is generally ignored as a combinator).

free/src/test/scala-2.13+/cats/free/FreeStructuralSuite.scala
```diff
@@ -0,0 +1,133 @@
...
+  checkAll("Free[ExprF, String]", EqTests[Free[ExprF, String]].eqv)
```
LukaJCB, line 31: 👍

free/src/test/scala-2.13+/cats/free/FreeStructuralSuite.scala
```diff
@@ -0,0 +1,133 @@
...
+      f.resume match {
+        case Left(sf) =>
+          S.perturb(seed, sf)
+
+        case Right(a) =>
+          A.perturb(seed, a)
+      }
+    }
+
+  Show[Free[Option, Int]]
```
LukaJCB, line 26: Is this just for checking that you can summon the show instance?

LukaJCB (approved):

PR #3951, opened 2021-07-25
Override `map2Eval` in `FlatMap`

vasilmkd:
We noticed that if not explicitly overridden to support lazy evaluation of the right argument, `map2Eval` can cause StackOverflowErrors for effect types.

Relevant Cats Effect issue.
https://github.com/typelevel/cats-effect/issues/2095


djspiewak (approved):

LukaJCB (approved):

xuwei-k:
- 😢 https://github.com/circe/circe/pull/1891

johnynek:
@xuwei-k I'm confused how your instance can be lawful and impacted by this change.

That said, leaving map2/map2Eval to be the default in FlatMap is probably a deoptimization for many cases.

PR #3956, opened 2021-08-05
Configure microsite image directory

### LINKED ISSUE #3690: Sponsor logos are broken in README (and on website)
Screenshot_2020-11-22 Cats Home

rossabaker:
Fixes #3690.  I'm just trying to put out a fire, and circling back with a more thoughtful approach from someone who understands jekyll would be good.

Concerns:
* I don't know why the source is in `docs` while the output is in `cats-docs`.
* There are a bunch more directory settings and I don't know what else is subtly broken.  I'm just taking care of the sponsors.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -273,6 +273,7 @@ lazy val docSettings = Seq(
     )
   ),
   micrositeGithubRepo := "cats",
+  micrositeImgDirectory := (LocalRootProject / baseDirectory).value / "docs" / "src" / "main" / "resources" / "microsite" / "img",
   micrositeTheme := "pattern",
   micrositePalette := Map(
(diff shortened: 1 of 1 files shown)


arosien:
Thanks so much @rossabaker!!

FYI there are still some broken non-sponsor images with dubious looking URLs like https://placehold.it/15/f03c15/000000?text=+. Maybe I'll just submit a patch to remove them, as they don't look particularly necessary.

rossabaker:
That sounds good.  For whatever reason, I'm not seeing them in the source.  Perhaps they were there when the site was published with the last release?

djspiewak (approved):

Issue #3957, opened 2021-08-06
`Eq[NonEmptySet[A]]` requires `Order[A]`, but only needs `Eq[A]`


domaspoliakas:
Hello folks 👋 

[`NonEmptySetEq[A]`](https://github.com/typelevel/cats/blob/main/core/src/main/scala/cats/data/NonEmptySet.scala#L426) has an `Eq[A]` member, but the implict construction of `NonEmptySetEq[A]` [here](https://github.com/typelevel/cats/blob/main/core/src/main/scala/cats/data/NonEmptySet.scala#L413) declares an `implicit A: Order[A]` parameter, meaning that you can't have a `Eq[NonEmptySet[A]]` unless `A` has an instance of `Order`, despite the fact that an `Eq` should suffice.

I'd be more than happy to get this fixed, but I don't know how to approach it with regards to binary compatibility.

joroKr21:
Hey - the binary-compatible approach is to make the old instance `private[data]` and not implicit and just add a new instance with the relaxed constraints. In this case you can even simply forward the old one to the new because `Order[A] <:< Eq[A]`.

PR #3958, opened 2021-08-06
Eq[NonEmptySet[A]] now needs Eq[A], not Order[A]

domaspoliakas:
Addresses #3957 

`+prePR` failed on my local machine for some reason complaining about code I was nowhere near, so I'm opening it as is for now and I'll fix whatever issues occur on CI if necessary


rossabaker (approved):
:+1: on green

djspiewak (approved):

PR #3959, opened 2021-08-08
Add mongo4cats

Kirill5k:
[mongo4cats](https://github.com/Kirill5k/mongo4cats) is a MongoDB Scala client wrapper compatible with Cats Effect ans Fs2. Available for Scala 2.12, 2.13 and 3.0.


rossabaker (approved):
Thanks!

Issue #3947, opened 2021-07-20
Stack-safety issues in `Kleisli` instances


catostrophe:
Repro 1:
```scala
import cats._
import cats.data._
import cats.implicits._

val res1 = (1 to 10000).toList.traverse_[Id, Unit](_ => ())  // ok

val res2 = (1 to 1000).toList.traverse_(_ => Kleisli.liftF[Id, String, Unit](())).run("") // fails with SO
```

`traverse_` relies on `List`'s`foldRight` and `Kleisli`'s`map2Eval` under the hood, which in turn relies on `F.map2Eval`

Repro 2 (involves Cats Effect):
```scala
import cats._
import cats.data._
import cats.implicits._
import cats.effect._

val res3 = (1 to 10000).toList.traverse_[IO, Unit](_ => IO.unit) // ok
val res4 = (1 to 10000).toList.traverse_(_ => Kleisli.liftF[IO, String, Unit](IO.unit)).run("") // ok

val res5 = (1 to 1000).toList.parTraverse_(_ => Kleisli.liftF[IO, String, Unit](IO.unit)).run("") // fails with SO
```

Thus, using `Kleisli[IO, R, A]` may result in SO at runtime if operations involving `IO.Par` are used, e.g. `parTraverse_`.

johnynek:
I think an issue there is a deep, cats3-level issue that `Eval[A]` was used in early cats where we really wanted to use `Defer[F]` but we had yet to understand that. Calling `.value` on an Eval is code smell and doing so in a loop is reasonably likely to break stack safety.

I could fix this issue with this diff:
```
+trait KleisliDeferApplicative[F[_], A] extends KleisliApplicative[F, A] {
+  implicit def deferF: Defer[F]
+
+  override def map2Eval[B, C, Z](fa: Kleisli[F, A, B], fb: Eval[Kleisli[F, A, C]])(
+    f: (B, C) => Z
+  ): Eval[Kleisli[F, A, Z]] = {
+    
+    val kb = Defer[Kleisli[F, A, *]].defer(fb.value)
+
+    Eval.now(Kleisli { a =>
+      val ec: Eval[F[C]] = Eval.always(kb.run(a))
+      val efz: Eval[F[Z]] = F.map2Eval(fa.run(a), ec)(f)
+      deferF.defer(efz.value)
+    })
+  }
+}
+

[...]

Issue #3790, opened 2021-02-12
Optimize traverseWithChain for StackSafeMonad(s)


djspiewak:
The `traverseWithChain` implementation is really good and performs well in benchmarks on downstream datatypes, but it has a meaningful flaw: it forces added allocations through `Eval` at every step of the iteration. This is something that is definitely necessary when you're trying to implement it on an arbitrary applicative, but that really isn't the case most of the time.

The common scenario I'm thinking about is traversing `List` with `IO`, which happens both in the context of `parTraverse` and the more conventional `traverse`, and it happens *very* frequently. When you instantiate `G = IO` within `traverseWithChain`, the `map2Eval` and corresponding machinery is no longer necessary because `IO` *itself* is stack-safe, and thus we can entirely remove that wrapping allocation in each step.

My proposal would be to do this by checking to see if `G.isInstanceOf[StackSafeMonad[G]]`. The reasoning here is that *most* such implementations will take advantage of the `StackSafeMonad` utility mixin. For those that don't, the old `Eval` implementation will still *work*, but for those that do, we have the guarantee (based on their own `tailRecM` laws) that their `flatMap` is capable of suspending the laziness just as safely as `Eval`'s. This covers the common case of `IO` as well as several other similar cases, and achieves a significant speed-up in downstream performance without sacrificing anything other than ugliness of implementation.

johnynek:
Forgive the verbose reply (much known to @djspiewak but I am adding for additional context for other readers).

This is an important question. Arguably, `traverse` is the most important function in cats since it is the functional programming answer to effectful loops, which are ubiquitous in programming. Cats has always been trying to serve two goals: 1. pure functional programming on scala 2. but fast.

If we want fast, we generally want to avoid laziness since that generally adds boxing and wrapping. But also, we want to preserve "short circuiting" error semantics when we can. Since we want fast, cats defaults to strict arguments, thus `def map2[A, B](fa: F[A], fb: F[B])...` requires the left and right to have been evaluated to `F[_]` before we can merge them. Early in cats, `Eval[A]` was the answer when we wanted targeted laziness since it was explicit, it can be composed (unlike call-by-name), and is reasonably fast for an pure execution monad (it is not for side effects). Thus we have both map2 and map2Eval, the latter taking an `Eval[F[B]]` to perserve laziness.

The problem comes when `F[_]` is already a non-strict type constructor, and we have to wrap with laziness twice: once in `F[_]` and once in `Eval[_]`.

[...]

PR #3960, opened 2021-08-09
Fix traverse_ for List and Vector to be stack safe.

johnynek:
This fixes traverse_ for List and Vector to avoid stack overflow errors.

This is a partial fix for #3947

### DIFF
--- a/tests/src/test/scala/cats/tests/TraverseSuite.scala
+++ b/tests/src/test/scala/cats/tests/TraverseSuite.scala
@@ -32,7 +32,7 @@ abstract class TraverseSuite[F[_]: Traverse](name: String)(implicit ArbFInt: Arb
     }
   }
 
-  test(s"Traverse[$name].traverseMatches_ with Option") {
+  test(s"Traverse[$name].traverse matches traverse_ with Option") {
     forAll { (fa: F[Int], fn: Int => Option[Int]) =>
       assert(Applicative[Option].void(fa.traverse(fn)) == fa.traverse_(fn))
     }
--- a/tests/src/test/scala/cats/tests/TraverseSuite.scala
+++ b/tests/src/test/scala/cats/tests/TraverseSuite.scala
@@ -32,6 +32,12 @@ abstract class TraverseSuite[F[_]: Traverse](name: String)(implicit ArbFInt: Arb
     }
   }
 
+  test(s"Traverse[$name].traverseMatches_ with Option") {
+    forAll { (fa: F[Int], fn: Int => Option[Int]) =>
+      assert(Applicative[Option].void(fa.traverse(fn)) == fa.traverse_(fn))
+    }
+  }
+
 }
 
 object TraverseSuite {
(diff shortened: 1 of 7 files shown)


johnynek:
take a look @armanbilge

armanbilge:
Cool!! So, if I'm understanding correctly, this doesn't _actually_ make it stack safe (in the theoretical sense) but more so in the practical sense, that it's effectively impossible to blow the stack due since its depth is now the logarithm of the size of the input?

johnynek:
So, I think we generally say that going `log(N)` depth stack is safe, since we can go 10000 deep, that means we are dealing with things on O(2^10000) before we blow the stack, which is not a size we can every actually reach.

Note, the sun does not have enough energy to enumerate all numbers to 2^200, so 2^1000 even is an unimaginably large number.

tests/src/test/scala/cats/tests/TraverseSuite.scala
```diff
@@ -32,6 +32,12 @@ abstract class TraverseSuite[F[_]: Traverse](name: String)(implicit ArbFInt: Arb
     }
   }
 
+  test(s"Traverse[$name].traverseMatches_ with Option") {
+    forAll { (fa: F[Int], fn: Int => Option[Int]) =>
+      assert(Applicative[Option].void(fa.traverse(fn)) == fa.traverse_(fn))
```
johnynek, line 37: there is no law on this, and probably should be, but requires source and binary changes to laws.

joroKr21:
Maybe we need `traverseViaChain_`? 🤔

johnynek:
I don't think so. The point of traverseViaChain is to build back up a sequence from a Tree structure. But in the case of traverse_ we don't build a resulting sequence: we just return the Applicative `G[Unit]`.

joroKr21:
Ahh right it could be just `traverseIndexedSeq_` - I think it also serves to reduce code duplication so that all collections can use it.

LukaJCB (approved):

tests/src/test/scala/cats/tests/TraverseSuite.scala
```diff
@@ -32,6 +32,12 @@ abstract class TraverseSuite[F[_]: Traverse](name: String)(implicit ArbFInt: Arb
     }
   }
 
+  test(s"Traverse[$name].traverseMatches_ with Option") {
```
LukaJCB, line 35: did you mean to say `traverse_ matches traverse with Option`?

LukaJCB (dismissed):
LGTM

djspiewak (commented):
This looks really good, but while you're in here… I would like to interest you in #3790. What you're touching right here is *the* bottleneck for any and all `par` operators, and the `Eval` wrapping is the most substantial cost difference between this generic implementation and a hand-optimized and specialized implementation.

tests/src/test/scala/cats/tests/TraverseSuite.scala
johnynek: good catch, thanks. Will you take a look again?

johnynek:
@djspiewak I'll reply on that issue with some thoughts. I agree we are in a suboptimal situation at the moment.

tests/src/test/scala/cats/tests/TraverseSuite.scala
djspiewak: FYI laws can be introduced without breaking either source or binary compatibility

djspiewak (approved):
I think it would be interesting to add a law, but we can hit it in a follow-up if necessary

PR #3961, opened 2021-08-09
Optimize void in several cases

johnynek:
`void` in Functor can be a significant optimization in that it can discard work that does not need to be done in a pure setting (which we assume in cats).

Even when we are dealing with a strict data structure, `void` can often be implemented more efficiently than using `map(_ => ())` since we can often allocate a singular `unit` type and cheaply check if we can return that.

### DIFF
--- a/core/src/main/scala/cats/instances/tailrec.scala
+++ b/core/src/main/scala/cats/instances/tailrec.scala
@@ -22,5 +22,6 @@ private object TailRecInstances {
         fa.flatMap(f)
 
       override val unit: TailRec[Unit] = done(())
+      override def void[A](ta: TailRec[A]): TailRec[Unit] = unit
     }
 }
--- a/core/src/main/scala/cats/instances/try.scala
+++ b/core/src/main/scala/cats/instances/try.scala
@@ -104,10 +104,7 @@ trait TryInstances extends TryInstances1 {
         if (idx == 0L) fa.toOption else None
 
       override def size[A](fa: Try[A]): Long =
-        fa match {
-          case Failure(_) => 0L
-          case Success(_) => 1L
-        }
+        if (fa.isSuccess) 1L else 0L
 
       override def find[A](fa: Try[A])(f: A => Boolean): Option[A] =
         fa.toOption.filter(f)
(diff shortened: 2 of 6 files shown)


core/src/main/scala/cats/instances/list.scala
```diff
@@ -179,6 +179,19 @@ trait ListInstances extends cats.kernel.instances.ListInstances {
 
       override def collectFirstSome[A, B](fa: List[A])(f: A => Option[B]): Option[B] =
         fa.collectFirst(Function.unlift(f))
+
+      override val unit: List[Unit] = () :: Nil
```
djspiewak, line 183: Given that this is being done with an eye towards optimization, my recommendation would be to follow the pattern used in Cats Effect to avoid the extra volatile read (on the object bitmap):

```suggestion
      override def unit: List[Unit] = _unit
      private[this] val _unit: List[Unit] = () :: Nil
```

core/src/main/scala/cats/instances/tailrec.scala
```diff
@@ -22,5 +22,6 @@ private object TailRecInstances {
         fa.flatMap(f)
 
       override val unit: TailRec[Unit] = done(())
```
djspiewak, line 24:
```suggestion
      override def unit: TailRec[Unit] = _unit
      private[this] val unit: TailRec[Unit] = done(())
```

core/src/main/scala/cats/instances/try.scala
```diff
@@ -141,6 +138,14 @@ trait TryInstances extends TryInstances1 {
       override def catchNonFatal[A](a: => A)(implicit ev: Throwable <:< Throwable): Try[A] = Try(a)
 
       override def catchNonFatalEval[A](a: Eval[A])(implicit ev: Throwable <:< Throwable): Try[A] = Try(a.value)
+
+      private val successUnit: Try[Unit] = Success(())
```
djspiewak, line 142:
```suggestion
      private[this] val successUnit: Try[Unit] = Success(())
```

`private[this]` is much faster than `private`

djspiewak (changes requested):

johnynek:
@djspiewak thanks for taking a look. Accepted these suggestions.

LukaJCB (approved):

djspiewak (approved):

rossabaker (approved):

PR #3962, opened 2021-08-10
Make StackSafeMonad extend Defer

johnynek:
Make StackSafeMonad extend Defer.

See the discussion in #3790 for more context.

The argument for why this is safe is given in the comments on StackSafeMonad. If they don't seem to make sense, please add a comment there.

cc @djspiewak @non 

I think we can follow up with using an idea similar to `shift` from Kleisli to make an Applicative map2 abstraction. As a sketch:

```scala
trait Map2Strategy[F[_]] {
  type Lazy[_]
  def map2[A, B](fa: F[A], fb: Lazy[F[B]]): Lazy[F[B]]
}
```
and then we can either make `Lazy[A] = Eval[A]` or `Lazy[A] = Id[A]` depending on the `F[_]` which maybe we can reflectively select based on if the `F[_]` is a `StackSafeMonad` or not.

Or something...

### DIFF
--- a/core/src/main/scala/cats/StackSafeMonad.scala
+++ b/core/src/main/scala/cats/StackSafeMonad.scala
@@ -24,7 +24,7 @@ trait StackSafeMonad[F[_]] extends Monad[F] with Defer[F] {
       case Right(b) => pure(b)
     }
 
-  /**
+  /*
    * This is always safe for a StackSafeMonad.
    * proof: we know flatMap can't blow the stack
    * because if it could, tailRecM would not be safe:
--- a/core/src/main/scala/cats/StackSafeMonad.scala
+++ b/core/src/main/scala/cats/StackSafeMonad.scala
@@ -9,12 +9,38 @@ import scala.util.{Either, Left, Right}
  * will inherit will not be sound, and will result in unexpected stack overflows.  This
  * trait is only provided because a large number of monads ''do'' define a stack-safe
  * flatMap, and so this particular implementation was being repeated over and over again.
+ *
+ * Note, tailRecM being safe and pure implies that the function passed to flatMap
+ * is not called immediately on the current stack (since otherwise tailRecM would
+ * stack overflow for sufficiently deep recursions). This implies we can implement
+ *
+ * defer(fa) = unit.flatMap(_ => fa)
  */
-trait StackSafeMonad[F[_]] extends Monad[F] {
+trait StackSafeMonad[F[_]] extends Monad[F] with Defer[F] {
 
   override def tailRecM[A, B](a: A)(f: A => F[Either[A, B]]): F[B] =
     flatMap(f(a)) {
       case Left(a)  => tailRecM(a)(f)
       case Right(b) => pure(b)
+    }
+
+  /**
+   * This is always safe for a StackSafeMonad.
+   * proof: we know flatMap can't blow the stack
+   * because if it could, tailRecM would not be safe:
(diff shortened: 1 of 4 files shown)


core/src/main/scala/cats/StackSafeMonad.scala
```diff
@@ -9,12 +9,38 @@ import scala.util.{Either, Left, Right}
...
+  def defer[A](fa: => F[A]): F[A] =
+    flatMap(unit)(_ => fa)
+}
+
+object StackSafeMonad {
+  def shiftFunctor[F[_], A, B](fn: A => F[B])(implicit F: Functor[F]): A => F[B] =
+    F match {
+      case ssm: StackSafeMonad[F] @unchecked => { a => ssm.defer(fn(a)) }
+      case _                                 => fn
+    }
```
djspiewak, line 45: I wonder if there's a way to make this more general. `A => F[B]` works extremely well for `Kleisli` and extremely not-well for almost anything else, since it will force boxing into a `Function1` in cases where it just isn't needed.

core/src/main/scala/cats/instances/tailrec.scala
```diff
@@ -11,7 +11,7 @@ trait TailRecInstances {
 private object TailRecInstances {
   val catsInstancesForTailRec: StackSafeMonad[TailRec] with Defer[TailRec] =
     new StackSafeMonad[TailRec] with Defer[TailRec] {
```
djspiewak, line 13:
```suggestion
  val catsInstancesForTailRec: StackSafeMonad[TailRec] =
    new StackSafeMonad[TailRec] {
```

Minor cleanup…

core/src/main/scala/cats/StackSafeMonad.scala
```diff
@@ -9,12 +9,38 @@ import scala.util.{Either, Left, Right}
...
-trait StackSafeMonad[F[_]] extends Monad[F] {
+trait StackSafeMonad[F[_]] extends Monad[F] with Defer[F] {
 
   override def tailRecM[A, B](a: A)(f: A => F[Either[A, B]]): F[B] =
     flatMap(f(a)) {
       case Left(a)  => tailRecM(a)(f)
       case Right(b) => pure(b)
     }
+
+  /**
```
djspiewak, line 27:
```suggestion
  /*
```

Can we actually make this *not* scaladoc? The problem here is if we make it a scaladoc comment, it will override the inherited one from `Defer`, which in turn results in *this* being the documentation shown in downstream generated scaladoc.

djspiewak (changes requested):
I like where this is going. I had a similar thought around the `Lazy[_]` idea but still trying to piece through how we can make it not-horrible.

core/src/main/scala/cats/StackSafeMonad.scala
johnynek: Maybe it shouldn't be public at all. I just took what was in Kleisli and made it more particular. I think the current Kleisli hack is adding cost for any Monad that isn't lazy (e.g. Option, or Either), and this function will help there.

johnynek:
@djspiewak please take a look and see what you think with your changes merged.

core/src/main/scala/cats/StackSafeMonad.scala
rossabaker: When in doubt, leave it private.  Can always publicize it later if someone finds a second use.

PR #3963, opened 2021-08-14
NonEmptyMap: add new methods

satorg:
Inspired by [this](https://discord.com/channels/632277896739946517/632278570512678923/875434472735146024) discussion.

Introduces a few new methods which make `NonEmptyMap` more convenient and useful:
  - `leftMap`: maps over the keys rather than values (in contrast to `map` which maps over the values).
    The name `leftMap` looks `Bifunctor`-ish although there's a required `Order` constraint added to the method.
  - `fullMap`: maps over both keys and values. It behaves like a regular `Map.map` from Scala collections, but also includes the `Order` constraint. The name is controversial though, just the best one I managed to come up with. But I am totally open to change it to something better.
  - `transform`: looks and behaves like `Map.transform`. Does not require any additional constraint.

Nothing is "set in stone", of course – feel free to make any suggestions.

**UPD**: Some names are changed: `leftMap` -> `mapKeys`, `fullMap` -> `mapBoth`.

### DIFF
--- a/core/src/main/scala/cats/data/NonEmptyMapImpl.scala
+++ b/core/src/main/scala/cats/data/NonEmptyMapImpl.scala
@@ -79,6 +79,28 @@ sealed class NonEmptyMapOps[K, A](val value: NonEmptyMap[K, A]) {
     NonEmptyMapImpl.create(Functor[SortedMap[K, *]].map(toSortedMap)(f))
 
   /**
+   * Applies f to all the keys leaving elements unchanged.
+   */
+  def leftMap[L](f: K => L)(implicit orderL: Order[L]): NonEmptyMap[L, A] = {
+    implicit val orderingL: Ordering[L] = orderL.toOrdering
+    NonEmptyMapImpl.create(toSortedMap.map(Bifunctor[Tuple2].leftMap(_)(f)))
+  }
+
+  /**
+   * Applies f to both keys and elements simultaneously.
+   */
+  def fullMap[L, B](f: (K, A) => (L, B))(implicit orderL: Order[L]): NonEmptyMap[L, B] = {
+    implicit val orderingL: Ordering[L] = orderL.toOrdering
+    NonEmptyMapImpl.create(toSortedMap.map(Function.tupled(f)))
+  }
+
+  /**
+   * Transforms the elements only by applying f to both elements and associated keys.
+   */
+  def transform[B](f: (K, A) => B): NonEmptyMap[K, B] =
+    NonEmptyMapImpl.create(toSortedMap.transform(f))
+
+  /**
    * Optionally returns the value associated with the given key.
    */
   def lookup(k: K): Option[A] = toSortedMap.get(k)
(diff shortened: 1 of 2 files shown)


tests/src/test/scala/cats/tests/NonEmptyMapSuite.scala
```diff
@@ -5,12 +5,13 @@ import cats.data.{NonEmptyList, NonEmptyMap}
 import cats.kernel.laws.discipline.{SerializableTests => _, _}
 import cats.laws.discipline._
 import cats.laws.discipline.arbitrary._
+import cats.syntax.bifunctor._
+import cats.syntax.eq._
 import cats.syntax.foldable._
 import cats.syntax.functor._
-import cats.syntax.show._
 import cats.syntax.reducible._
+import cats.syntax.show._
```
satorg, line 13: Just sorted alphabetically.

core/src/main/scala/cats/data/NonEmptyMapImpl.scala
```diff
@@ -78,6 +78,28 @@ sealed class NonEmptyMapOps[K, A](val value: NonEmptyMap[K, A]) {
...
+   */
+  def leftMap[L](f: K => L)(implicit orderL: Order[L]): NonEmptyMap[L, A] = {
+    implicit val orderingL: Ordering[L] = orderL.toOrdering
+    NonEmptyMapImpl.create(toSortedMap.map(Bifunctor[Tuple2].leftMap(_)(f)))
+  }
+
+  /**
+   * Applies f to both keys and elements simultaneously.
+   */
+  def fullMap[L, B](f: (K, A) => (L, B))(implicit orderL: Order[L]): NonEmptyMap[L, B] = {
```
johnynek, line 92: I wonder if mapBoth might be a better name here (which is used in WriterT in related way).

core/src/main/scala/cats/data/NonEmptyMapImpl.scala
```diff
@@ -78,6 +78,28 @@ sealed class NonEmptyMapOps[K, A](val value: NonEmptyMap[K, A]) {
   def map[B](f: A => B): NonEmptyMap[K, B] =
     NonEmptyMapImpl.create(Functor[SortedMap[K, *]].map(toSortedMap)(f))
 
+  /**
+   * Applies f to all the keys leaving elements unchanged.
+   */
+  def leftMap[L](f: K => L)(implicit orderL: Order[L]): NonEmptyMap[L, A] = {
```
johnynek, line 84: I think `mapKeys` might be a better name here.

core/src/main/scala/cats/data/NonEmptyMapImpl.scala
satorg: done

core/src/main/scala/cats/data/NonEmptyMapImpl.scala
satorg: updated

satorg:
@johnynek, thank you for your suggestion. The names are changed to `mapBoth` and `mapKeys`.

johnynek (commented):
Thanks for the nice PR.

johnynek (approved):

PR #3972, opened 2021-08-24
Render sponsors from Open Collective REST API

rossabaker:
Makes sponsors appear automatically on the site, at the expense of appearing in the README.

1. Call the Open Collective REST API to retrieve all the active sponsors
2. Merge with sponsors who donated through other means
3. Remove obsolete language about previous sponsorship levels

Issues:
* This makes the README a bit unsightly.  Help wanted.
* I don't think it will render in local jekyll without doing unspeakable things to disable CORS.
* I didn't implement pagination, if we get more sponsors.  Help wanted.
* This is more JavaScript than I've written in at least a decade.

Local rendering below:

Screen Shot 2021-08-24 at 3 00 09 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,51 +36,39 @@ information about how you can get involved as a developer. If you are looking fo
 
 ### Financial Contributors
 
-[Become a financial contributor](https://opencollective.com/typelevel) and help us sustain our community. Donations directly support office hour for maintainers, better documentation and strategic initiatives. 
+[Become a financial contributor](https://opencollective.com/typelevel) and help us sustain our community. Donations directly support office hours for maintainers, better documentation and strategic initiatives.
 
+<h4>Platinum Sponsors</h4>
+<a href="https://opencollective.com/typelevel/contribute/platinum-sposor-12420/checkout">Platinum sponsorship</a> starts at $950 USD/month.
+<div id="platinum-sponsors">
+  <noscript>Platinum Sponsors appear here at <a href="https://typelevel.org/cats/">https://typelevel.org/cats</a></noscript>
+</div>
 
+<h4>Gold Sponsors</h4>
+<a href="https://opencollective.com/typelevel/contribute/gold-sponsor-12419/checkout">Gold Sponsorship</a> starts at $420 USD/month.
+<div id="gold-sponsors">
+  <noscript>Gold Sponsors appear here at <a href="https://typelevel.org/cats/">https://typelevel.org/cats</a></noscript>
+</div>
 
-### Gold Sponsors 
+<h4>Silver Sponsors</h4>
+<a href="https://opencollective.com/typelevel/contribute/silver-sponsor-11780/checkout">Silver Sponsorship</a> starts at $180 USD/month.
+<div id="silver-sponsors">
+  <noscript>Silver Sponsors appear here at <a href="https://typelevel.org/cats/">https://typelevel.org/cats</a></noscript>
+</div>
 
-Gold Sponsors are those who have pledged $5,000 to $10,000.
+<h4>Backers</h4>
+Become a <a href="https://opencollective.com/typelevel/contribute/backer-11779/checkout">Backer</a> with a recurring donation of just $5 USD/month.
+<div id="backers">
+  <noscript>Backs appear here at <a href="https://typelevel.org/cats/">https://typelevel.org/cats</a></noscript>
+</div>
 
-<a href="http://47deg.com"><img src="https://typelevel.org/img/media/sponsors/47_degrees.png" style="margin-left: 10px; margin-right: 10px" /></a>
-<a href="https://www.iteratorshq.com"><img src="https://typelevel.org/img/media/sponsors/iterators.png" style="margin-bottom:8px;margin-left: 10px; margin-right: 10px"/></a>
-<a href="https://triplequote.com/"><img src="https://typelevel.org/img/media/sponsors/triplequote.png" style="margin-bottom:9px; margin-right: 10px" /></a>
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -36,51 +36,39 @@ information about how you can get involved as a developer. If you are looking fo
...
+<h4>Silver Sponsors</h4>
+<a href="https://opencollective.com/typelevel/contribute/silver-sponsor-11780/checkout">Silver Sponsorship</a> starts at $180 USD/month.
+<div id="silver-sponsors">
+  <noscript>Silver Sponsors appear here at <a href="https://typelevel.org/cats/">https://typelevel.org/cats</a></noscript>
+</div>
+
+<h4>Backers</h4>
+Become a <a href="https://opencollective.com/typelevel/contribute/backer-11779/checkout">Backer</a> with a recurring donation of just $5 USD/month.
+<div id="backers">
+  <noscript>Backs appear here at <a href="https://typelevel.org/cats/">https://typelevel.org/cats</a></noscript>
```
armanbilge, line 62:
```suggestion
  <noscript>Backers appear here at <a href="https://typelevel.org/cats/">https://typelevel.org/cats</a></noscript>
```

docs/src/main/resources/microsite/js/sponsors.js
```diff
@@ -0,0 +1,96 @@
...
+                        if (member.totalAmountDonated > 0) {
+                            addSponsor('other-contributors', member, ContributorSize);
+                        }
+                    }
+                };
+            }
+        }
+    };
+    xhr.send();
+};
```
armanbilge, line 55: NBD, but [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) is the modern way. Something like this I think:

```suggestion
var sponsors = async function() {
    var response = await fetch('https://opencollective.com/typelevel/members/all.json');
    if (response.ok) {
        var members = await response.json();
        for(i = 0; i < members.length; i++) {
            var member = members[i];
            if (member.isActive) {
                switch (member.tier) {
                case 'Platinum Sponsor':
                    addSponsor('platinum-sponsors', member, PlatinumSize);
                case 'Gold Sponsor':
                    addSponsor('gold-sponsors', member, GoldSize);
                case 'Silver Sponsor':
                    addSponsor('silver-sponsors', member, SilverSize);
                case 'backer':
                    addSponsor('backers', member, BackerSize);
                    break;
                default:
                    if (member.totalAmountDonated > 0) {
                        addSponsor('other-contributors', member, ContributorSize);
                    }
                }
            }
        }
    }
};
```

rossabaker: Thanks. I don't know and I don't have time to retest it, but if a knowledgeable reviewer agrees, please feel free to commit it!

armanbilge: If it ain't broke don't fix it! 😆

ChristopherDavenport: Lets move forward with this, and then we can PR this in a later review. We want to fix whats here most importantly.

ChristopherDavenport (approved):

LukaJCB (approved):

docs/src/main/resources/microsite/js/sponsors.js
rossabaker: :+1:  My CSS is dated and my JS was never competent, and would be delighted to see either improved once the fire is out.

rossabaker:
I worry the missing avatar SVG isn't very gender neutral.  Or also that it represents companies.  It's what I found on Wikimedia.  Better ideas welcome there too, if people make a second pass.

rossabaker:
This didn't work and in fact made things worse because that endpoint doesn't add CORS headers.  Only the static ones rendered. :rage:

rossabaker:
Apparently it needs to be converted to GraphQL, which I haven't the slightest idea how to do.

armanbilge:
I'll give this a try!

rossabaker:
`https://rest.opencollective.com/v2/typelevel/orders/incoming` gets close, without mucking around with any GraphQL libraries.

This reveals three statuses: `ACTIVE`, `CANCELLED`, and `PAID`.  I'd keep the active ones in the tiers, but include them all in the "other contributors", as some of those were very generous one-time contributions, and they're all appreciated.

This one doesn't seem to sort by amount like the old endpoint did.  That was nice, but not critical.

armanbilge:
Just got this:
https://rest.opencollective.com/v1/typelevel/members.json

armanbilge:
Ok, but that one I posted is missing cors :(

armanbilge:
> https://rest.opencollective.com/v2/typelevel/orders/incoming gets close

I think this is the only way to go. Using the GraphQL API would require an API key which is not worth the trouble.

armanbilge:
Instead I created a PR at opencollective to add cors to the endpoint. They seem active on that repo and responsive on slack, so if you're willing to wait a day and see how this plays out? Otherwise we'll fallback to exactly what @rossabaker proposed in his comment above.

rossabaker:
The one I posted has CORS, but has the more complex GraphQL response, which I think is still not too hard to traverse as pure JSON.

rossabaker:
Sorry, when I responded, it hadn't shown the most recent comments.  I saw on another issue that the /v1 is no longer maintained, so I think we're going to need to go ahead with the /v2.

armanbilge:
The code for `rest.opencollective.com/v{1,2}` is in the same repo (mixed in the same files) and just forwards to the new GraphQL API. There is active development there as of 12 hours ago. If you're willing to wait overnight (they are European time zone), let's just see what they say. `orders/incoming` as you propose is okay but sub-optimal since it's not really a member list, but more a list of transactions IIUC. As you say we can make it work though.

Issue #3974, opened 2021-08-25
Restore Open Collective sponsors to site


rossabaker:
I tried with #3972 and made it worse, because I used a non-CORS endpoint after hacking CORS locally.  The conversation continued on that issue, but it is very important we find and publish a solution.  This is to hold us accountable for doing so quickly, or to revert #3972 and manually add a couple more sponsors if there is no quick solution.

@armanbilge has PR'ed a fix upstream, and also has a backup plan.

armanbilge:
I'll invoke the backup plan first thing tomorrow unless upstream moves fast tonight.

armanbilge:
FYI some movement on my upstream PR with the promise that "Maybe not today, but it could be landed sometime this week."

Therefore, backup plan it is.

rossabaker:
@djspiewak is the only admin right now, so I think he'd need to provide the key.

This means we're going to have to keep it as part of the build process, doesn't it?  And if we do that, periodically refresh the site?  I guess we could have a scheduled GitHub workflow just to fetch this JSON, and then our fetch could fetch that.

armanbilge:
> This means we're going to have to keep it as part of the build process, doesn't it?

Well, until I bug them enough to get my CORS fix live 😛 

Alternatives:
1. We can host a tiny proxy (e.g., a free cloudflare worker) that adds CORS to the request and point to that. Actually, I tried this first but ran into problems without the API key.
2. With an API key, we _may_ be able to use the GraphQL API directly in the site. Not sure, I'm confused about whether the key should be kept private, and if so, how it's even possible to use GraphQL in the frontend. I'll research this again.
3. Use your idea to extract the sponsors from the transactions API which does have CORS enabled. I'm sorry I was stubborn and tried to find "better" ways, but more and more that might be the best fix.

PR #3979, opened 2021-08-26
Add `Semifield` and `CommutativeSemifield` to algebra

armanbilge:
Introduces a `Semifield` and `CommutativeSemifield` to algebra. I'm going by the [second definition on Wikipedia](https://en.wikipedia.org/wiki/Semifield):
* a commutative monoid for addition (+)
* a group for multiplication (*)

Thanks in advance!

### DIFF
--- a/algebra-core/src/main/scala/algebra/ring/CommutativeSemifield.scala
+++ b/algebra-core/src/main/scala/algebra/ring/CommutativeSemifield.scala
@@ -1,0 +1,19 @@
+package algebra
+package ring
+
+import scala.{specialized => sp}
+
+/**
+ * CommutativeSemifield is a Semifield that is commutative under multiplication.
+ */
+trait CommutativeSemifield[@sp(Int, Long, Float, Double) A]
+    extends Any
+    with Semifield[A]
+    with CommutativeRig[A]
+    with MultiplicativeCommutativeGroup[A]
+
+object CommutativeSemifield
+    extends AdditiveMonoidFunctions[CommutativeSemifield]
+    with MultiplicativeGroupFunctions[CommutativeSemifield] {
+  @inline final def apply[A](implicit r: CommutativeSemifield[A]): CommutativeSemifield[A] = r
+}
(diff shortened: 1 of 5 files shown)


armanbilge:
@johnynek would you mind reviewing this? Thanks!

johnynek (approved):
Looks good to me.

Are there examples you have in mind that are not other structures?

Are there any instances we can create that we couldn't fully express before?

armanbilge:
Thanks for the review! I'm using semifields for non-negative reals in my probabilistic programming library, where probabilities are added, multiplied, and divided (but never negative). For example, I have a `LogDouble` type which will have this: https://github.com/armanbilge/schrodinger/blob/main/log-double/src/main/scala/schrodinger/LogDouble.scala#L67

Not sure if this is anything we can add to Algebra itself.

johnynek:
I can't restart the microsite build. Can you push again? I want to make sure it was a flake.

armanbilge:
Absolutely. But this was a known issue, see https://github.com/typelevel/cats/issues/3974#issuecomment-907176663

armanbilge:
@johnynek good to go.

benhutchison:
Thank you @armanbilge & @johnynek for adding `Semifield` & friends 🙏 

I'm guessing the algebra type-classes must sometimes feel obscure and even that no one notices/appreciates the work. So I thought I'd share my story.. 

Having defined a non-negative sub-species of `BigDecimal`, I went looking for the right algebra to implement +, * and /, without negation. And lo and behold, I discovered Semifield was exactly what I needed. At first, I assumed it was part of @non's original vision, but then noticed it is missing from the [algebra docs](https://typelevel.org/cats/algebra.html#ring-like-structures) and got curious how/when it came to be.

On docs, I do think an extra two rows are needed in the [Ring table](https://typelevel.org/cats/algebra.html#ring-like-structures). But then it further qualifies the statement that "every lower structure is also an instance of the structures above it"; I actually wonder if with another two exceptions, that over-generalization has gone past time?

armanbilge:
Thanks!

> I do think an extra two rows are needed in the [Ring table](https://typelevel.org/cats/algebra.html#ring-like-structures).

Indeed!

> I actually wonder if with another two exceptions, that over-generalization has gone past time?

Yes, agree, probably an inheritance diagram would be more useful at this point :) (in addition to the existing table.)

PR #3984, opened 2021-08-28
Use OpenCollective GraphQL API to render sponsors

### LINKED ISSUE #3974: Restore Open Collective sponsors to site

armanbilge:
This should close #3974 once and for all. It reverts my previous fix in #3977 and instead uses the latest and greatest OpenCollective GraphQL v2 API to get a live list of sponsors on page load. Turns out we don't need an API key to use it after all. Bonus points: fixes the missing avatar issue.


rossabaker (approved):
:+1: Publishing now to see how it looks

rossabaker:
It works great.  Thanks for persevering on this.

PR #3987, opened 2021-09-01
Implement `InvariantCoyoneda`

armanbilge:
This came up in https://github.com/typelevel/vault/pull/264, but @tpolecat's stack was about to overflow 😝

### DIFF
--- a/free/src/main/scala/cats/free/InvariantCoyoneda.scala
+++ b/free/src/main/scala/cats/free/InvariantCoyoneda.scala
@@ -20,22 +20,22 @@ sealed abstract class InvariantCoyoneda[F[_], A] extends Serializable { self =>
   val fi: F[Pivot]
 
   /**
-   * The list of transformer functions, to be composed and lifted into `F` by `run`.
+   * The list of the `Pivot => *` halves of the transformer functions, to be composed and lifted into `F` by `run`.
    */
   private[cats] val ks0: List[Any => Any]
 
   /**
-   * The list of transformer functions, to be composed and lifted into `F` by `run`.
+   * The list of the `* => Pivot` halves of the transformer functions, to be composed and lifted into `F` by `run`.
    */
   private[cats] val ks1: List[Any => Any]
 
   /**
-   * The list of transformer functions composed into a single function, to be lifted into `F` by `run`.
+   * Half of the composed transformer function, to be lifted into `F` by `run`.
    */
   final def k0: Pivot => A = Function.chain(ks0.reverse)(_).asInstanceOf[A]
 
   /**
-   * The composed transformer function, to be lifted into `F` by `run`.
+   * The other half of the composed transformer function, to be lifted into `F` by `run`.
    */
   final def k1: A => Pivot = Function.chain(ks1)(_).asInstanceOf[Pivot]
 
--- a/free/src/main/scala/cats/free/InvariantCoyoneda.scala
+++ b/free/src/main/scala/cats/free/InvariantCoyoneda.scala
@@ -67,7 +67,7 @@ object InvariantCoyoneda {
   /**
    * `F[A]` converts to `InvariantCoyoneda[F,A]` for any `F`
    */
-  def lift[F[_], A](fa: F[A]): InvariantCoyoneda[F, A] =
+  def lift[F[_], A](fa: F[A]): Aux[F, A, A] =
     apply(fa)(identity[A])(identity[A])
 
   /**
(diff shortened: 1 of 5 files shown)


free/src/main/scala/cats/free/InvariantCoyoneda.scala
```diff
@@ -0,0 +1,108 @@
...
+
+  /**
+   * The list of transformer functions composed into a single function, to be lifted into `F` by `run`.
+   */
+  final def k0: Pivot => A = Function.chain(ks0.reverse)(_).asInstanceOf[A]
+
+  /**
+   * The composed transformer function, to be lifted into `F` by `run`.
+   */
+  final def k1: A => Pivot = Function.chain(ks1)(_).asInstanceOf[Pivot]
```
armanbilge, line 40: Are there better names for these? Also, I see the comments are broken, will fix.

tpolecat: I think this is fine, just to keep it consistent with the co/contravariant versions.

LukaJCB (dismissed):

tpolecat (dismissed):

free/src/main/scala/cats/free/InvariantCoyoneda.scala
```diff
@@ -0,0 +1,108 @@
...
+  /**
+   * Lift the `Pivot` type member to a parameter. It is usually more convenient to use `Aux` than
+   * a refinment type.
+   */
+  type Aux[F[_], A, B] = InvariantCoyoneda[F, A] { type Pivot = B }
+
+  /**
+   * `F[A]` converts to `InvariantCoyoneda[F,A]` for any `F`
+   */
+  def lift[F[_], A](fa: F[A]): InvariantCoyoneda[F, A] =
```
johnynek, line 78: Why not preserve the Aux type here but do so in other places?

free/src/main/scala/cats/free/InvariantCoyoneda.scala
```diff
@@ -0,0 +1,108 @@
...
+
+  /**
+   * The underlying value.
+   */
+  val fi: F[Pivot]
+
+  /**
+   * The list of the `Pivot => *` halves of the transformer functions, to be composed and lifted into `F` by `run`.
+   */
+  private[cats] val ks0: List[Any => Any]
```
johnynek, line 25: Do we not want to use AndThen here so we know the function composition is safe (since we control it)?

free/src/main/scala/cats/free/InvariantCoyoneda.scala
johnynek: Actually I don't see the reason to ever preserve the Aux type. I can't see how anyone can use it. Isn't it better to use the type that hides the Pivot?

free/src/main/scala/cats/free/InvariantCoyoneda.scala
armanbilge: I think so too, I was just following the style of the other `Coyoneda`s. Should I make the change across the board?

free/src/main/scala/cats/free/InvariantCoyoneda.scala
armanbilge: I'm just following the style of the other `Coyoneda`s, happy to make this change.

johnynek: I think it is worth it if we have AndThen designed explicitly for this purpose but it is probably binary incompatible on the existing ones since you will change List to AndThen.

armanbilge:
@johnynek I reworked all the Coyonedas to use `AndThen`. Seems binary compatible to me, what do you think?

Re `Aux`, I'm a little less clear about this change?

free/src/main/scala/cats/free/InvariantCoyoneda.scala
johnynek: If we are going to retain the Aux in other places I think this should be `Aux[F, A, A]`.

johnynek:
I don't follow your comment "seems binary compatible" since you had to add mima exclusions. It seems binary *incompatible* to me, but I believe it is safe because: 1. these types are sealed 2. the changed values were `private[cats]`.

free/src/main/scala/cats/free/InvariantCoyoneda.scala
armanbilge: Makes sense. I made this change to all the Coyonedas.

armanbilge:
> is safe because: 1. these types are sealed 2. the changed values were private[cats].

Apologies, "safe" is what I meant, I'll be more precise next time. Or, binary-compatible from the perspective of users anyway :)

tpolecat (approved):

tpolecat:
This looks good, are we waiting on anything else?

johnynek (approved):

PR #3991, opened 2021-09-10
adding override modifier

gokyo:

### DIFF
--- a/core/src/main/scala/cats/Align.scala
+++ b/core/src/main/scala/cats/Align.scala
@@ -119,7 +119,7 @@ import scala.annotation.implicitNotFound
 object Align extends ScalaVersionSpecificAlignInstances {
   def semigroup[F[_], A](implicit F: Align[F], A: Semigroup[A]): Semigroup[F[A]] =
     new Semigroup[F[A]] {
-      def combine(x: F[A], y: F[A]): F[A] = Align[F].alignCombine(x, y)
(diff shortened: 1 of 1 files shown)


rossabaker (commented):
Thanks, @gokyo.  Is this just a style thing, or did it fix a specific problem?

gokyo:
@rossabaker 
It is not an issue. 
The code would behave in the same way, regardless if you use or not the `override` modifier. 
I find the code easier to read when the overridden methods have the `override` modifier. 
I wonder what your views are.

ritschwumm:
i wouldn't put an `override` here: it's abstract in Semigroup, so imho it's "implemented" rather than"overridden", as there _is_ nothing to override.

i prefer using the `override` modifier as a warning sign when someone actually did override a concrete, existing implementation, which i consider a code smell.

gokyo:
I understand your point. 
It is a shame that we don't have an `implemented` modifier in Scala. 

I personally find it useful to quickly see if a method is introduced in the class/object/trait you are reading, or somewhere in its upper hierarchy. 

This is just a suggestion based on my personal preference though. 
So, if you decide to ignore this PR, that's fine.

rossabaker (commented):
I generally use the modifier the same as @ritschwumm.

armanbilge:
I also try to use `override` only where its explicitly necessary (i.e. enforced by the compiler). If we want to change the code style for this in Cats I reckon it will be a much bigger PR :)

PR #3997, opened 2021-09-19
Relax a constraint for Alternative's unite and separate methods from Monad to FlatMap

satorg:
The initial discussion is [here](https://discord.com/channels/632277896739946517/632278570512678923/888660638073749514).

### Rationale

- The `unite` and `separate` methods don't need a `Monad` actually to do their job, because they only use `flatMap` in addition to methods from `Alternative`. Therefore the `FlatMap[F]` constraint is just enough there.
- Moreover, both `Alternative` and `Monad` implement `Applicative`. That means the `Applicative` methods can come by two different paths which creates sort of a loophole for ambiguity:
  ```scala
  implicit val fooMonad: Monad[Foo] = ??? // assume Monad[Foo] is different from Alternative[Foo]
  Alternative[Foo].unite[Bar, Int](Foo(Bar(123)) // which `pure` should it use – one from `Alternative` or `Monad` ? 
  ```

### Discussion

In fact, if it is okay to constraint the `unite` and `separate` methods with a `Monad`, then the `Alternative` typeclass is absolutely redundant in such a case – we can simply put these two methods directly into `MonoidK`.

Another approach (perhaps even better one) could be a new typeclass (let's name it `Monadditive` for the sake of this example) which would inherit to `Monad` and `MonoidK` (via `Alternative`) instead of just `Applicative` and `MonoidK` (which the current `Alternative` inherits to). In this case the definition of those two methods – `unite` and `separate` – could be simplified to the following:
```scala
trait Monadditive[F[_]] extends Alternative[F] with Monad[F] { // implies `MonoidK` as well
  def unite[G[_], A](fga: F[G[A]])(implicit G: Foldable[G]): F[A] = ??? // implementation remains the same
  def separate[G[_, _], A, B](fgab: F[G[A, B]])(implicit G: Bifoldable[G]): (F[A], F[B]) = ??? // implementation remains the same
}
```
In other words, an additional constraint for `F[_]` is not necessary anymore.

Therefore such a new typeclass would allow to reduce number of additional constraints passed to those methods.
And looks like it would make a lot of sense anyway – most of the containers which Cats supports implement corresponding `Monad` instances, not only `Applicative`.

### DIFF
--- a/core/src/main/scala/cats/Alternative.scala
+++ b/core/src/main/scala/cats/Alternative.scala
@@ -6,6 +6,15 @@ import scala.annotation.implicitNotFound
 @implicitNotFound("Could not find an instance of Alternative for ${F}")
 @typeclass trait Alternative[F[_]] extends NonEmptyAlternative[F] with MonoidK[F] { self =>
 
+  // Note: `protected` is only necessary to enforce binary compatibility
+  // since neither `private` nor `private[cats]` work properly here.
+  @deprecated("use a FlatMap-constrained version instead", "2.6.2")
+  protected def unite[G[_], A](fga: F[G[A]])(FM: Monad[F], G: Foldable[G]): F[A] = {
+    implicit def FM0: FlatMap[F] = FM
+    implicit def G0: Foldable[G] = G
+    unite(fga)
+  }
+
   /**
    * Fold over the inner structure to combine all of the values with
    * our combine method inherited from MonoidK. The result is for us
@@ -15,29 +24,34 @@ import scala.annotation.implicitNotFound
    *
    * Example:
    * {{{
-   * scala> import cats.implicits._
    * scala> val x: List[Vector[Int]] = List(Vector(1, 2), Vector(3, 4))
    * scala> Alternative[List].unite(x)
    * res0: List[Int] = List(1, 2, 3, 4)
    * }}}
    */
-  def unite[G[_], A](fga: F[G[A]])(implicit FM: Monad[F], G: Foldable[G]): F[A] =
-    FM.flatMap(fga) { ga =>
-      G.foldLeft(ga, empty[A])((acc, a) => appendK(acc, a))
-    }
+  def unite[G[_], A](fga: F[G[A]])(implicit FM: FlatMap[F], G: Foldable[G]): F[A] =
+    FM.flatMap(fga) { G.foldMapK(_)(pure)(self) }
+
+  // Note: `protected` is only necessary to enforce binary compatibility
+  // since neither `private` nor `private[cats]` work properly here.
+  @deprecated("use a FlatMap-constrained version instead", "2.6.2")
+  protected def separate[G[_, _], A, B](fgab: F[G[A, B]])(FM: Monad[F], G: Bifoldable[G]): (F[A], F[B]) = {
+    implicit def FM0: FlatMap[F] = FM
+    implicit def G0: Bifoldable[G] = G
+    separate(fgab)
+  }
 
   /**
-   * Separate the inner foldable values into the "lefts" and "rights"
+   * Separate the inner foldable values into the "lefts" and "rights".
    *
    * Example:
    * {{{
-   * scala> import cats.implicits._
    * scala> val l: List[Either[String, Int]] = List(Right(1), Left("error"))
    * scala> Alternative[List].separate(l)
    * res0: (List[String], List[Int]) = (List(error),List(1))
    * }}}
    */
-  def separate[G[_, _], A, B](fgab: F[G[A, B]])(implicit FM: Monad[F], G: Bifoldable[G]): (F[A], F[B]) = {
+  def separate[G[_, _], A, B](fgab: F[G[A, B]])(implicit FM: FlatMap[F], G: Bifoldable[G]): (F[A], F[B]) = {
     val as = FM.flatMap(fgab)(gab => G.bifoldMap(gab)(pure, _ => empty[A])(algebra[A]))
     val bs = FM.flatMap(fgab)(gab => G.bifoldMap(gab)(_ => empty[B], pure)(algebra[B]))
     (as, bs)
@@ -45,9 +59,9 @@ import scala.annotation.implicitNotFound
 
   /**
    * Separate the inner foldable values into the "lefts" and "rights".
-   * A variant of [[separate]] that is specialized
-   * for Fs that have Foldable instances
-   * which allows for a single-pass implementation
+   * 
+   * A variant of [[[separate[G[_,_],A,B](fgab:F[G[A,B]])(implicitFM:cats\.FlatMap[F]* separate]]]
+   * that is specialized for Fs that have Foldable instances which allows for a single-pass implementation
    * (as opposed to {{{separate}}} which is 2-pass).
    *
    * Example:
@@ -115,9 +129,11 @@ object Alternative {
     def self: F[A]
     val typeClassInstance: TypeClassType
     def unite[G[_], B](implicit ev$1: A <:< G[B], FM: Monad[F], G: Foldable[G]): F[B] =
-      typeClassInstance.unite[G, B](self.asInstanceOf[F[G[B]]])(FM, G)
+      // Note: edited manually since seems Simulacrum is not able to handle the bin-compat redirection properly.
+      typeClassInstance.unite[G, B](self.asInstanceOf[F[G[B]]])
     def separate[G[_, _], B, C](implicit ev$1: A <:< G[B, C], FM: Monad[F], G: Bifoldable[G]): (F[B], F[C]) =
-      typeClassInstance.separate[G, B, C](self.asInstanceOf[F[G[B, C]]])(FM, G)
+      // Note: edited manually since seems Simulacrum is not able to handle the bin-compat redirection properly.
+      typeClassInstance.separate[G, B, C](self.asInstanceOf[F[G[B, C]]])
     def separateFoldable[G[_, _], B, C](implicit ev$1: A <:< G[B, C], G: Bifoldable[G], FF: Foldable[F]): (F[B], F[C]) =
       typeClassInstance.separateFoldable[G, B, C](self.asInstanceOf[F[G[B, C]]])(G, FF)
   }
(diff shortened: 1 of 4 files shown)


satorg:
Converting the PR to Draft since Mime is not happy with this it:
```
[error] (coreJVM / mimaReportBinaryIssues) Failed binary compatibility check against org.typelevel:cats-core_2.13:2.6.1! Found 1 potential problems
```
Let me investigate the issue first. If it is not worth the effort to fix it, I'll close the PR.

rossabaker:
If you keep the old one and make it `private[cats]`, you'll fix that MiMa complaint.  Then you'll be fighting ambiguous overloads at call sites within cats, and cats-rooted projects like cats-effect.  You can fix the ambiguous overload by putting the old one in a bincompat parent trait, but then a different MiMa error pops up.  We've done this successfully on instances, but I don't remember a precedent on typeclasses.

There's also syntax to wrestle with.

satorg:
@rossabaker thank you for the advice. As I can see now it won't be an easy lift. Seems I managed to calm MiMa down at this point, although the solution I came up with looks quite ugly to me. As you can see in the last commit, I added a new `AlternativeBinCompat0` trait for that. Initially I tried to move the old `unite` method to this trait, but it didn't make MiMa happy unfortunately. It simply changed the way it complaints to
```
[error] Cats core: Failed binary compatibility check against org.typelevel:cats-core_2.13:2.6.1! Found 1 potential problems
[error]  * in current version, classes mixing cats.Alternative need be recompiled to wire to the new static mixin forwarder method all super calls to method unite(java.lang.Object,cats.Monad,cats.Foldable)java.lang.Object
[error]    filter with: ProblemFilters.exclude[NewMixinForwarderProblem]("cats.Alternative.unite")
```
So I decided to try the opposite: place the new `unite` into `AlternativeBinCompat0` while leaving the old one in place (but restricting its visibility via `private[cats]`. Currently all checks pass, but I haven't got started with the syntax yet.

tests/src/test/scala/cats/tests/AlternativeSuite.scala
```diff
@@ -9,7 +9,9 @@ class AlternativeSuite extends CatsSuite {
     forAll { (list: List[Option[String]]) =>
       val expected = list.collect { case Some(s) => s }
 
-      assert(Alternative[List].unite(list) === expected)
+      assert(
+        (Alternative[List]: AlternativeBinCompat0[List]).unite(list) === expected
```
satorg, line 13: To make sure the new version is used here.

core/src/main/scala/cats/Alternative.scala
```diff
@@ -89,6 +88,16 @@ import scala.annotation.implicitNotFound
...
 }
 
+private[cats] trait AlternativeBinCompat0[F[_]] { self: Alternative[F] =>
+
+  def unite[G[_], A](fga: F[G[A]])(implicit FM: FlatMap[F], G: Foldable[G]): F[A] =
+    unite0(fga)
+
+  // Allows to share implementation between the old and new `unite` method versions.
+  final protected def unite0[G[_], A](fga: F[G[A]])(implicit FM: FlatMap[F], G: Foldable[G]): F[A] =
+    FM.flatMap(fga) { G.foldMapK(_)(pure)(self) }
```
satorg, line 98: I tried to change implementation from
```
FM.flatMap { a =>
  G.foldLeft(ga, empty[A])((acc, a) => combineK(acc, pure(a)))
}
```
to just
```
FM.flatMap(fga) { G.foldMapK(_)(pure)(self) }
```
It should be equivalent to the previous one. But can be even more efficient since `foldMapK` has optimized implementations at least for some of supported collections.

satorg: ... but if there're any concerns or objections with this approach, I won't resist to switch back to the original implementation.

satorg:
Seems I managed to get through all the thorns with this task. The only caveat: Simulacrum Scalafix rules don't work correctly with all these binary compatibility tricks in `Alternative` I had to apply. But that Simulacrum generated syntax is not currently in use for `Alternative` anyway.

core/src/main/scala/cats/Alternative.scala
```diff
@@ -15,39 +24,44 @@ import scala.annotation.implicitNotFound
...
    * scala> Alternative[List].unite(x)
    * res0: List[Int] = List(1, 2, 3, 4)
    * }}}
    */
-  def unite[G[_], A](fga: F[G[A]])(implicit FM: Monad[F], G: Foldable[G]): F[A] =
-    FM.flatMap(fga) { ga =>
-      G.foldLeft(ga, empty[A])((acc, a) => combineK(acc, pure(a)))
-    }
+  def unite[G[_], A](fga: F[G[A]])(implicit FM: FlatMap[F], G: Foldable[G]): F[A] =
+    FM.flatMap(fga) { G.foldMapK(_)(pure)(self) }
```
satorg, line 33: The original implementation is replaced with this better optimized one which utilizes `Foldable.foldMapK` (instead of the former's `foldLeft`).

satorg:
cc @rossabaker, in case if you're curious in how I've fought it :)

satorg:
Hmm.. Seems the build failure is not related to the PR:
```
[error] org.scalajs.ir.IRVersionNotSupportedException: Failed to deserialize a file compiled with Scala.js 1.7 (supported up to: 1.5): /home/runner/.cache/coursier/v1/https/repo1.maven.org/maven2/org/scala-js/scalajs-library_2.13/1.7.1/scalajs-library_2.13-1.7.1.jar:/scala/volatile.sjsir
[error] You may need to upgrade the Scala.js sbt plugin to version 1.7 or later.
[error] 	at org.scalajs.sbtplugin.ScalaJSPluginInternal$.enhanceIRVersionNotSupportedException(ScalaJSPluginInternal.scala:83)
[error] 	at org.scalajs.sbtplugin.ScalaJSPluginInternal$.$anonfun$scalaJSStageSettings$19(ScalaJSPluginInternal.scala:227)
```
I wonder – is the main branch broken (how come?)

rossabaker:
Yes, that scala.js issue came up on another branch.  Pretty sure it's a result of #4018.

rossabaker:
#4048 should fix the build.

core/src/main/scala/cats/syntax/alternative.scala
```diff
@@ -13,11 +13,20 @@ trait AlternativeSyntax {
     new GuardOps(b)
 }
 
-final class UniteOps[F[_], G[_], A](private val fga: F[G[A]]) extends AnyVal {
+final class UniteOps[F[_], G[_], A] private[syntax] (protected val fga: F[G[A]])
+    extends AnyVal
+    with UniteOpsBinCompat0[F, G, A] {
+
+  @deprecated("use a FlatMap-constrained instead", "2.6.2")
```
rossabaker, line 20: For consistency with your other deprecation messages:

```suggestion
  @deprecated("use a FlatMap-constrained version instead", "2.6.2")
```

rossabaker (commented):
Nice work!  I've used package private to hide overloads, but can't recall ever using a protected.  It seems to work.  Have you tried calling it anywhere with something that has a `FlatMap` but no `Monad`?  I'm mostly sure it will still work, but want to be 100% sure.  (Maybe this encoding isn't as novel as I think.)

satorg:
> I've used package private to hide overloads, but can't recall ever using a protected. It seems to work.

Right, good catch. First I tried to make it `private[cats]` indeed. And it works in terms of binary compatibility too. But it makes testing the new methods a way more sophisticated since apparently all the tests are also located inside the `cats` package and the compiler tries to route all the calls to `unite` and `separate`  to the old methods instead of the new ones. Using `protected` instead of `private[cats]` basically solves the problem while preserving the binary compatibility.

Interesting observation: this trick with `protected` does not work for syntax – it breaks BC for Scala 2.12 (but still works for Scala 2.13 and Scala 3). I don't know exactly why.

rossabaker:
> Interesting observation: this trick with protected does not work for syntax – it breaks BC for Scala 2.12 

Oh, so this is still going to fail MiMa even after #4048 is merged? 🤔

satorg:
> Oh, so this is still going to fail MiMa even after #4048 is merged? 🤔

No, I don't think so. I mean, all the checks had been succeeding before #4018. But later I rebased this PR onto `main` to test it against the latest version – and the checks began failing. I'm going to make the changes you requested and we will see what happens anyway :)

core/src/main/scala/cats/syntax/alternative.scala
satorg: Fixed.

satorg:
> Have you tried calling it anywhere with something that has a FlatMap but no Monad?

Good catch, thanks. Seems I overlooked that it has to be explicitly tested. Should be good now.

rossabaker (approved):

core/src/main/scala/cats/syntax/alternative.scala
```diff
@@ -13,11 +13,20 @@ trait AlternativeSyntax {
     new GuardOps(b)
 }
 
-final class UniteOps[F[_], G[_], A](private val fga: F[G[A]]) extends AnyVal {
```
satorg, line 16: **Note**: the `fga`'s access is changed from `private` to `protected` because `fga` needs to be accessible from `UniteOpsBinCompat0`. The same change is made for the `SeparateOps` below.

**One more note**: the `UniteOps`'s constructor is made `private[syntax]`. That makes sense in general – I believe it is better to keep its access scope as narrow as possible. And apparently it does not affect the binary compatibility. But if someone has a concern regarding this change – speak up please, we can change this access back to `public`.

rossabaker: The `fga` seems fine.

Changing the constructor visibility is a source-breaking change that's highly unlikely to break anybody's source.  I'd probably leave it public to be conservative in Cats, but I'm not too fussed either way.

rossabaker (dismissed):

satorg:
@rossabaker thank you for keeping eye on this PR and helping me with polishing it, I appreciated it.

Yeah, as you might notice, I eventually decided to revert  access for `UniteOps` and `SeparateOps` constructors back to public. Although I still believe it would be better to make them private to `cats`, but the improvement would be so minutiae then it wouldn't be worth the risk it imposes.

satorg:
Ironically, this PR had got conflicts with another my PR that have been merged recently.

rossabaker (approved):

johnynek (approved):

PR #3931, opened 2021-06-25
Add missing methods to VectorOps

satorg:
Adds methods `groupByNev`, `groupByNevA`, `scanLeftNev` and `scanRightNev` to `VectorOps`,
which are similar to the methods found in `ListOps`.

The implementation is mostly copied from `ListOps` and adjusted accordingly.

Initial discussion is [here](https://discord.com/channels/632277896739946517/632278570512678923/857709871448588309).

PR #3998, opened 2021-09-23
Add missing methods to NE data stuctures

satorg:
The initial discussion is [here](https://discord.com/channels/632277896739946517/632278570512678923/890152626027626506).

The following methods are added:

* `NonEmptyList`:
  - `prependList` 

**Note:** this PR is based on #3931 therefore changes from the latter may show up here until the latter is merged.

### DIFF
--- a/core/src/main/scala/cats/syntax/vector.scala
+++ b/core/src/main/scala/cats/syntax/vector.scala
@@ -1,11 +1,127 @@
 package cats.syntax
 
 import cats.data.NonEmptyVector
+import cats.{Applicative, Functor, Order, Traverse}
+
+import scala.collection.immutable.SortedMap
 
 trait VectorSyntax {
   implicit final def catsSyntaxVectors[A](va: Vector[A]): VectorOps[A] = new VectorOps(va)
 }
 
 final class VectorOps[A](private val va: Vector[A]) extends AnyVal {
+
+  /**
+   * Returns an Option of NonEmptyVector from a Vector
+   *
+   * Example:
+   * {{{
+   * scala> import cats.data.NonEmptyVector
+   * scala> import cats.implicits._
+   *
+   * scala> val result1: Vector[Int] = Vector(1, 2)
+   * scala> result1.toNev
+   * res0: Option[NonEmptyVector[Int]] = Some(NonEmptyVector(1, 2))
+   *
+   * scala> val result2: Vector[Int] = Vector.empty[Int]
+   * scala> result2.toNev
+   * res1: Option[NonEmptyVector[Int]] = None
+   * }}}
+   */
   def toNev: Option[NonEmptyVector[A]] = NonEmptyVector.fromVector(va)
+
+  /**
+   * Groups elements inside this `Vector` according to the `Order` of the keys
+   * produced by the given mapping function.
+   *
+   * {{{
+   * scala> import cats.data.NonEmptyVector
+   * scala> import scala.collection.immutable.SortedMap
+   * scala> import cats.implicits._
+   *
+   * scala> val vector = Vector(12, -2, 3, -5)
+   *
+   * scala> val expectedResult = SortedMap(false -> NonEmptyVector.of(-2, -5), true -> NonEmptyVector.of(12, 3))
+   *
+   * scala> vector.groupByNev(_ >= 0) === expectedResult
+   * res0: Boolean = true
+   * }}}
+   */
+  def groupByNev[B](f: A => B)(implicit B: Order[B]): SortedMap[B, NonEmptyVector[A]] = {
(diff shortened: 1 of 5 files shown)


core/src/main/scala/cats/syntax/vector.scala
```diff
@@ -1,11 +1,141 @@
...
+   *
+   * scala> val vector = Vector(12, -2, 3, -5)
+   *
+   * scala> val expectedResult = Option(SortedMap(false -> NonEmptyVector.of(-2, -5), true -> NonEmptyVector.of(12, 3)))
+   *
+   * scala> vector.groupByNevA(num => Option(0).map(num >= _)) === expectedResult
+   * res0: Boolean = true
+   * }}}
+   */
+  def groupByNevA[F[_], B](
```
johnynek, line 86: I wonder if this should live here:
https://github.com/typelevel/cats/blob/main/core/src/main/scala/cats/NonEmptyTraverse.scala

```scala
def groupByA[G[_], B](fn: A => G[B])(implicit G: Apply[G], B: Order[B]): G[SortedMap[B, F[A]]]
```
and then you can call that here:
```scala
  toNev.fold(F.pure(SortedMap.empty[B, NonEmptyVector[A]])(_.groupByA(f))
```

johnynek (commented):
To me this highlights a weakness of the `NonEmpty` types...

I can imagine:
```scala
trait Reducible[F[_]] {
  type Empty[_]
  def emptyFoldable: Foldable[Empty]

  def toEmpty[A](f: F[A]): Empty[A]
  def fromEmpty[A](f: Empty[A]): Option[F[A]]
}

trait NonEmptyTraverse[F[_]] extends Reducible[F] {
  type Empty[_]
  def emptyTraverse: Traverse[Empty]
}
```
We could potentially add a typeclass:

```scala
trait HasNonEmpty[F[_], TC[_[_]]] {
  type NonEmpty[_]
  def emptyTC: TC[F]
  def nonEmptyTC: TC[NonEmpty]
  def toEmpty[A](f: F[A]): Empty[A]
  def fromEmpty[A](f: Empty[A]): Option[F[A]]
}
```
and then use the `HasNonEmpty[List, Traverse]`, `HasNonEmpty[Vector, Traverse]`, etc...

core/src/main/scala/cats/syntax/vector.scala
satorg: Well, these particular methods are just an attempt to maintain some degree of consistency and interoperability among all linear "non-empty" collections we have in `cats.data`. Because initially there was `ListOps#groupByNelA` only but no corresponding methods in `VectorOps` nor `SeqOps`. Which is a bit confusing.

satorg: But I like the idea of generalization. I also have a feeling, that keeping stuffing various `*Ops` classes with almost identical methods is not the best way to deal with it. I like your suggestion in particular, but why `NonEmptyTraverse` ?

Seems, it should work with just a regular `Traverse` – shouldn't it?

I mean, when we do grouping on some collection by some predicate, all the groups we will get eventually are going to be non-empty regardless of whether the source collection was empty or not. I.e. we will either get a non-empty group or don't get it at all.

satorg: We can use the way you suggested of course:
```
toNev.fold(F.pure(SortedMap.empty[B, NonEmptyVector[A]])(_.groupByA(f))
```
But to me it has at least two flaws:
1. It is too verbose and may require explicit types.
2. It mimics the idea of type safety for grouping: i.e. grouping over a regular collection ALWAYS results in zero or more groups of corresponding non-empty collections.

johnynek: to your point 2: putting `groupByA` on `NonEmptyTraverse` we could actually return `NonEmptyMap[B, F[A]]`, so we know both the map is non-empty and since `F[_]: NonEmptyTraverse` we also know `F[_]` is non-empty.

johnynek:
I appreciate the time you have taken to do this. I find it hard to review code that just adds ad-hoc methods to enrich standard types. It's difficult to know that we are being consistent or that we aren't missing a more general concept that should properly be added to a typeclass (which is the main focus of cats).

core/src/main/scala/cats/syntax/vector.scala
satorg: Yes, this is a valid case too. But I'd suppose that grouping over a regular `Traverse` is a more common case anyway. And it's guaranteed to produce a regular `Map` of `NonEmpty*` groups. While grouping over `NonEmptyTraverse` is guaranteed to produce a `NonEmptyMap` of `NonEmpty*` groups. I mean both cases looks legit to me and I'd like to think on how to incorporate both of them.

PR #4002, opened 2021-09-30
Migration to Scaladoc 3

pikinier20:
Try `sbt ++3.1.1-RC1 docs_scala3/unidoc` to generate only API documentation or  `sbt ++3.1.1-RC1 docs_scala3/makeMicrosite` to generate whole microsite.

Preview can be found there: https://scala3docstorage.blob.core.windows.net/cats/index.html

Current problem: snippets in Markdown files don't compile with Scala 3. ~~Usually, there are indent errors (Scala 3 doesn't like mixed spaces and tabs) but sometimes~~ there are type errors.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -316,6 +316,7 @@ lazy val docSettings = Seq(
 )
 
 lazy val docs_scala3Settings = docSettings ++ Seq(
+  Compile / scalacOptions ++= Seq("-no-indent"), // Fix for indent errors in mdoc
   ScalaUnidoc / unidoc / scalacOptions ++= Seq(
     "-external-mappings:" +
       ".*scala/.*::scaladoc3::https://dotty.epfl.ch/api/," +
--- a/build.sbt
+++ b/build.sbt
@@ -39,7 +39,7 @@ ThisBuild / githubWorkflowJavaVersions := Seq(PrimaryJava, LTSJava, GraalVM8)
 
 val Scala212 = "2.12.15"
 val Scala213 = "2.13.6"
-val Scala3 = "3.0.2"
+val Scala3 = "3.1.1-RC1"
 
 ThisBuild / crossScalaVersions := Seq(Scala212, Scala213, Scala3)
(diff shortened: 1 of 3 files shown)


docs/src/main/mdoc/datatypes/freemonad.md
```diff
@@ -525,31 +525,31 @@ type TeletypeState[A] = State[List[String], A]
...
   def apply[A](fa : Teletype[A]) : TeletypeState[A] = {
-	fa match {
-	  case ReadLine(prompt) =>
-		println(prompt)
-		val userInput = "hanging in here" //scala.io.StdIn.readLine()
-		StateT.pure[Eval, List[String], A](userInput)
-	  case WriteLine(line) =>
-		StateT.pure[Eval, List[String], A](println(line))
-	}
+  fa match {
```
johnynek, line 545: this looks weird. It looks like the indentation is constant. Is it due to mixed tabs/spaces? Can we have a way that this compiles but also doesn't look weird in text?

build.sbt
```diff
@@ -315,6 +315,23 @@ lazy val docSettings = Seq(
...
+      "github::https://github.com/typelevel/cats," +
+      "gitter::https://gitter.im/typelevel/cats," +
+      "twitter::https://twitter.com/typelevel," +
+      "discord::https://discord.gg/hWd4eS244g",
+    "-Ygenerate-inkuire",
+    "-project-logo",
+    "docs/src/main/resources/microsite/img/cats-logo.png"
+  )
+)
+
```
armanbilge, line 334: None of this is particularly specific to Cats actually :) we should explore upstreaming these into sbt-typelevel.

armanbilge:
Thanks for the PR! Scaladoc 3 seems pretty interesting. If you plan to keep working on this, would you mind re-opening it as a PR against sbt-typelevel? Then we can roll out these new features to all Typelevel projects :)

https://github.com/typelevel/sbt-typelevel/

PR #4007, opened 2021-10-03
Introduce parReplicateA

RafalSumislawski:
This PR introduces `parReplicateA` as a parallel version of `Applicative.replicateA`.

### DIFF
--- a/core/src/main/scala/cats/Parallel.scala
+++ b/core/src/main/scala/cats/Parallel.scala
@@ -106,6 +106,14 @@ trait Parallel[M[_]] extends NonEmptyParallel[M] {
 
       override def whenA[A](cond: Boolean)(f: => F[A]): F[Unit] = applicative.whenA(cond)(f)
     }
+
+  /**
+   * Like [[Applicative.replicateA]], but uses the apply instance
+   * corresponding to the Parallel instance instead.
+   */
+  def parReplicateA[A](n: Int, ma: M[A]): M[List[A]] =
+    if (n <= 0) monad.pure(List.empty[A])
+    else Parallel.parSequence(List.fill(n)(ma))(UnorderedFoldable.catsTraverseForList, this)
 }
 
 object NonEmptyParallel extends ScalaVersionSpecificParallelInstances {
--- a/core/src/main/scala/cats/syntax/parallel.scala
+++ b/core/src/main/scala/cats/syntax/parallel.scala
@@ -245,6 +245,10 @@ final class ParallelApOps[M[_], A](private val ma: M[A]) extends AnyVal {
 
   def parProduct[B](mb: M[B])(implicit P: Parallel[M]): M[(A, B)] =
     Parallel.parProduct(ma, mb)
+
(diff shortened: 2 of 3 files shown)


core/src/main/scala/cats/Parallel.scala
```diff
@@ -106,6 +106,14 @@ trait Parallel[M[_]] extends NonEmptyParallel[M] {
...
       override def whenA[A](cond: Boolean)(f: => F[A]): F[Unit] = applicative.whenA(cond)(f)
     }
+
+  /**
+   * Like [[Applicative.replicateA]], but uses the apply instance
+   * corresponding to the Parallel instance instead.
+   */
+  def parReplicateA[A](n: Int, ma: M[A]): M[List[A]] =
+    if (n <= 0) monad.pure(List.empty[A])
+    else Parallel.parSequence(List.fill(n)(ma))(UnorderedFoldable.catsTraverseForList, this)
```
satorg, line 116: To me it seems that `trait Parallel` is designed to just represent a relationship between a Monad and Applicative with parallel composition. But most of complimentary methods are placed into `object Parallel` (including `parAp` per se). I'm not absolutely confident, but shouldn't `parReplicateA` be also placed in the object rather than the trait?

core/src/main/scala/cats/syntax/parallel.scala
```diff
@@ -245,6 +245,10 @@ final class ParallelApOps[M[_], A](private val ma: M[A]) extends AnyVal {
 
   def parProduct[B](mb: M[B])(implicit P: Parallel[M]): M[(A, B)] =
     Parallel.parProduct(ma, mb)
+
+  def parReplicateA(n: Int)(implicit P: Parallel[M]): M[List[A]] =
+    if (n <= 0) P.monad.pure(List.empty[A])
+    else Parallel.parSequence(List.fill(n)(ma))
```
satorg, line 251: I guess these two lines can be replaced with just `P.parReplicateA(...)` if you decide to leave `Parallel#parReplicateA` in the `trait Parallel` or with `Parallel.parReplicateA(...)` if you move it into the object. Just to avoid code duplication.

tests/src/test/scala/cats/tests/SyntaxSuite.scala
```diff
@@ -184,6 +184,7 @@ object SyntaxSuite {
     val mb4: M[B] = ma.parProductR(mb)
     val mab2: M[(A, B)] = ma.parProduct(mb)
     val mb5: M[B] = mab.parAp(ma)
+    val mla: M[List[A]] = ma.parReplicateA(3)
```
satorg, line 187: `mock[Int]` instead of just `3`?

RafalSumislawski:
Hello @satorg I agree with you comments and so I've pushed corresponding changes.

satorg (approved):
Looks good to me, thanks.

Although I think it will be also appreciated if you add a corresponding test case to `ParallelSuite`.

satorg (approved):
Looks great, thx!

PR #4009, opened 2021-10-06
Added flatMapN

### LINKED ISSUE #4003: Introducing `.flatMapN(f)` as a shorthand for `.mapN(f).flatten`
Hello 👋 this follows from my question on discord and is basically the title. 

I find myself doing `.mapN(f).flatten` very frequently and I think it'd be nice to have a `.flatMapN` for this scenario. 

I am told this was rejected in the past, but I could not find a past issue for this and so I was hoping I could reopen a discussion.

Thanks!

domaspoliakas:
Hello 👋 

As promised here is `.flatMapN`. Took me a little while to get my head around `Boilerplate`. Let me know if this works for you, or if you need anything else.

resolves #4003


project/Boilerplate.scala
```diff
@@ -522,6 +522,12 @@ object Boilerplate {
         else
           s"def traverseN[G[_]: Applicative, Z](f: (${`A..N`}) => G[Z])(implicit traverse: Traverse[F], semigroupal: Semigroupal[F]): G[F[Z]] = Semigroupal.traverse$arity($tupleArgs)(f)"
 
+      val flatMap =
+        if (arity == 1)
+          s"def flatMap[Z](f: (${`A..N`}) => F[Z])(implicit flatMap: FlatMap[F]): F[Z] = flatMap.flatMap($tupleArgs)(f)"
+        else
+          s"def flatMapN[Z](f: (${`A..N`}) => F[Z])(implicit flatMap: FlatMap[F], semigroupal: Semigroupal[F]): F[Z] = flatMap.flatten(Semigroupal.map$arity($tupleArgs)(f))"
+
```
satorg, line 530: Looking into the Boilerplate code I noticed, that support for `mapN` is currently implemented in this way:
- `SemigroupalArityFunctions` contains a bunch of map***X*** (***2 ≤ X ≤ 22***) functions (along with other arity functions like imap***X*** and contramap***X***.
- These implementations are used then from all other arity syntax, including `TupleSemigroupalSyntax`.

Shouldn't we follow the same pattern for `flatMapN` – i.e. add implementations for all flatMap***X*** into `SemigroupalArityFunctions` in the first place and then make use of it from everywhere else?

I am not certain about it though, just an observation.

satorg: Also there's the `ApplyArityFunctions` trait that adds the map***X*** syntax directly to the `Apply` typeclass.

Curious, shouldn't we create a `FlatMapArityFunctions` trait (it does not exist yet) and add flatMap***X*** method directly into the `FlatMap` typeclass as well?

satorg: Agree, jar size is a concern. Did you try by chance to measure how much the resulting jars get increased in size comparing to the base version? May be it adds not too much...

domaspoliakas: I did not actually. I'll give it a shot today/tomorrow and see how it compares.

armanbilge:
Personally, I frequently find myself wanting `flatMapN`. Also, IMHO jar size is not _that_ big of a deal, see commentary in https://github.com/typelevel/cats/pull/3871#issuecomment-979604249.

In theory, we add methods to typeclasses (and syntax merely delegates to those) so that instances can override those with more efficient implementations (you can't override syntax extensions :)

In practice, not sure if it makes a meaningful difference here. But still I think that adding these methods to a `FlatMapArityFunctions` would probably be the way to do it.

domaspoliakas:
Rebased on latest main, added `FlatMapArityFunctions` and made tuple syntax delegate to those instead of the previous implementation

tests/src/test/scala/cats/tests/SyntaxSuite.scala
```diff
@@ -372,10 +372,14 @@ object SyntaxSuite {
...
 
-  def testFlatMap[F[_]: FlatMap, A, B, C, D]: Unit = {
+  def testFlatMap[F[_]: FlatMap, A, B, C, D, Z]: Unit = {
     val a = mock[A]
     val returnValue = mock[F[Either[A, B]]]
     val done = a.tailRecM[F, B](a => returnValue)
+    val tfabc = mock[(F[A], F[B], F[C])]
+    val ff = mock[(A, B, C) => F[Z]]
+
+    tfabc.flatMapN(ff)
```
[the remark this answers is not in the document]
armanbilge: Yup, I think this is easy to add.

project/Boilerplate.scala
```diff
@@ -516,6 +547,12 @@ object Boilerplate {
         else
           s"def traverseN[G[_]: Applicative, Z](f: (${`A..N`}) => G[Z])(implicit traverse: Traverse[F], semigroupal: Semigroupal[F]): G[F[Z]] = Semigroupal.traverse$arity($tupleArgs)(f)"
 
+      val flatMap =
+        if (arity == 1)
+          s"def flatMap[Z](f: (${`A..N`}) => F[Z])(implicit flatMap: FlatMap[F]): F[Z] = flatMap.flatMap($tupleArgs)(f)"
+        else
+          s"def flatMapN[Z](f: (${`A..N`}) => F[Z])(implicit flatMap: FlatMap[F]): F[Z] = flatMap.flatMap$arity($tupleArgs)(f)"
```
armanbilge, line 554: So I haven't been able to wrap my head around whether this technically belongs in `TupleSemigroupalSyntax`. It's certainly the easiest place to put it, and it works, I'm just confused by the "semigroupal" in the name :)

armanbilge: > [Semigroupal](https://typelevel.org/cats/api/cats/Semigroupal.html) captures the idea of composing independent effectful values.

Ah, ok that makes sense :)

armanbilge (approved):
Thank you very much, I think this looks good!

armanbilge (approved):

tests/src/test/scala/cats/tests/SyntaxSuite.scala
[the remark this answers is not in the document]
satorg: I would really like to have this addressed (since the syntax for Tuple1 seems to be slightly different), but I'm not going to hold the entire PR on just because of this concern only.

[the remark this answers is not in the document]
satorg: I would really like to have this addressed (since the syntax for Tuple1 seems to be slightly different), but I'm not going to hold the entire PR on just because of this concern only.

satorg (dismissed):
lgtm, thanks a lot. The test for the `arity = 1` case would be appreciated though.

domaspoliakas:
I'll add the test, but I'm honestly not sure why there even is a `Tuple1` syntax. Is that useful?

danicheg (approved):
I was dreaming of `flatMapN` for years. It's nice to see this happening!

armanbilge (approved):
> I'll add the test, but I'm honestly not sure why there even is a `Tuple1` syntax. Is that useful?

Agree, pretty dubious about this 😆 but it's good for completeness. Thanks!

satorg:
> I'll add the test, but I'm honestly not sure why there even is a `Tuple1` syntax. Is that useful?

Perhaps just for completeness and consistency with other syntax, I guess.
Thank you, looks really great!

satorg (approved):

PR #4014, opened 2021-10-09
[proposal] optimize Alternative (part 1): introduce NonEmptyAlternative with prependK and appendK methods

satorg:
### Rationale

There's a pretty common pattern can be found for `Alternative` even in the Cats sources:
`F.combineK(F.pure(a), fa)` and `F.combineK(fa, F.pure(a))`
which basically means prepending or appending an element `a` lifted into `F[_]`.
But when it is used for some standard collections like `List` or `Vector` it basically results in creating an interim singleton collection on each `F.pure` which gets discarded right after `F.combineK` is completed.

The proposed PR allows to avoid creating the interim singleton collections by various `Alternative` implementations in many cases.

### Description

This PR does the following:
1. Introduces a new type class `NonEmptyAlternative` as a combination of `Applicative` and `SemigroupK`. 
2. Adds `prependK` and `appendK` methods to `NonEmptyAlternative`. These methods have default implementations as shown above. But they can be overridden for a particular `Alternative` implementation.
3. Makes `Alternative` inheriting `NonEmptyAlternative` instead of just `Applicative`. The `MonoidK` mix-in type class remains in place.
4. Switches existing `F.pure + F.combineK` patterns found in core Cats (apart from `cats.data`) to `F.prependK` or `F.appendK` appropriately where it is safe and reasonable.
5. Introduces `NonEmptyAlternativeLaws` which becomes a base for `AlternativeLaws`.
6. Both the laws above get an ability to test composed (via `compose`) type-classes (see `composed` constructor in the corresponding companion objects).

Things to be done in a subsequent PR (see the discussion below):

7. Add `NonEmptyAlternative` instance to non-empty containers in `cats.data`.
8. Override `F.prependK` and `F.appendK` in the `Alternative` and `NonEmptyAlternative` implementations where it reasonable.

Therefore this PR along with the next one will allow to utilize more efficient methods in some collections (e.g. ` :: ` in `List`) while preserving binary and source compatibility with existing `Alternative` implementations. Moreover they should enable further optimizations for non-empty data types too.

### Caveats

- Simulacrum's ScalaFix rules do not recognize the `prependK` method because it is defined in this way:
  ```scala
  def prependK(a: A, fa: F[A]): F[A] // seems Simulacrum cannot understand `fa` as a second parameter
  ```
  I believe it is an issue with Simulacrum itself and not a big deal anyway since neither `Alternative` nor `NonEmptyAlternative` make use of extension methods generated by Simulacrum actually.

### DIFF
--- a/core/src/main/scala/cats/Alternative.scala
+++ b/core/src/main/scala/cats/Alternative.scala
@@ -4,7 +4,7 @@ import simulacrum.typeclass
 import scala.annotation.implicitNotFound
 
 @implicitNotFound("Could not find an instance of Alternative for ${F}")
-@typeclass trait Alternative[F[_]] extends Applicative[F] with MonoidK[F] { self =>
+@typeclass trait Alternative[F[_]] extends NonEmptyAlternative[F] with MonoidK[F] { self =>
 
   /**
    * Fold over the inner structure to combine all of the values with
@@ -23,7 +23,7 @@ import scala.annotation.implicitNotFound
    */
   def unite[G[_], A](fga: F[G[A]])(implicit FM: Monad[F], G: Foldable[G]): F[A] =
     FM.flatMap(fga) { ga =>
-      G.foldLeft(ga, empty[A])((acc, a) => combineK(acc, pure(a)))
+      G.foldLeft(ga, empty[A])((acc, a) => appendK(acc, a))
     }
 
   /**
@@ -61,8 +61,8 @@ import scala.annotation.implicitNotFound
   def separateFoldable[G[_, _], A, B](fgab: F[G[A, B]])(implicit G: Bifoldable[G], FF: Foldable[F]): (F[A], F[B]) =
     FF.foldLeft(fgab, (empty[A], empty[B])) { case (mamb, gab) =>
       G.bifoldLeft(gab, mamb)(
-        (t, a) => (combineK(t._1, pure(a)), t._2),
-        (t, b) => (t._1, combineK(t._2, pure(b)))
+        (t, a) => (appendK(t._1, a), t._2),
+        (t, b) => (t._1, appendK(t._2, b))
       )
     }
 
@@ -104,12 +104,11 @@ object Alternative {
   object ops {
     implicit def toAllAlternativeOps[F[_], A](target: F[A])(implicit tc: Alternative[F]): AllOps[F, A] {
       type TypeClassType = Alternative[F]
-    } =
-      new AllOps[F, A] {
-        type TypeClassType = Alternative[F]
-        val self: F[A] = target
-        val typeClassInstance: TypeClassType = tc
-      }
+    } = new AllOps[F, A] {
+      type TypeClassType = Alternative[F]
+      val self: F[A] = target
+      val typeClassInstance: TypeClassType = tc
+    }
   }
   trait Ops[F[_], A] extends Serializable {
     type TypeClassType <: Alternative[F]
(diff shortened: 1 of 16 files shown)


core/src/main/scala/cats/Alternative.scala
```diff
@@ -82,6 +82,28 @@ import scala.annotation.implicitNotFound
...
+  /**
+   * Lift `a` into `F[_]` and prepend it to `fa`.
+   *
+   * Example:
+   * {{{
+   * scala> Alternative[List].prependK(List(2, 3, 4), 1)
+   * res0: List[Int] = List(1, 2, 3, 4)
+   * }}}
+   */
+  def prependK[A](fa: F[A], a: A): F[A] = combineK(pure(a), fa)
```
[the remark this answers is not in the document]
satorg: I would like to, but not sure how to map it into the syntax. I.e.:
For `Alternative#prependK(a : A, fa: F[A])` should the syntax become `(A: a).prependK(fa: F[A])` or remain `(fa: F[A]).pependK(a: A)` ?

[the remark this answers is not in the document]
satorg: Ok, I'll give it a try. The only concern is the simulacrum scalafix rules. Although right now seems the syntax generated by simulacrum is not used, but as I understand it is supposed to be used in the future eventually. So I'll try to swap arguments and see what happens )

johnynek (dismissed):
one comment, take or leave.

Thanks for doing this!

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -712,23 +712,37 @@ object NonEmptyList extends NonEmptyListInstances {
 
 sealed abstract private[data] class NonEmptyListInstances extends NonEmptyListInstances0 {
 
+  // TODO: put correct version number.
+  @deprecated("Use catsDataInstancesForNonEmptyListBinCompat1", "2.6.2-?")
+  def catsDataInstancesForNonEmptyList
+    : SemigroupK[NonEmptyList] with Bimonad[NonEmptyList] with NonEmptyTraverse[NonEmptyList] with Align[NonEmptyList] =
+    catsDataInstancesForNonEmptyListBinCompat1
+
```
satorg, line 720: Not sure if it is the best way to address binary compatibility though.

johnynek (approved):

satorg:
Turned out it was not an easy lift and required a way more changes that I initially anticipated. But seems the PR is competed now. Finally)

There were several bin-compat issues which I had to tackle. The solutions don't look esthetically great but they're the best I managed to come up with. And seems MiMa is happy now.

But if someone has got an idea on how to make it better, I'd really appreciate any help/hint on that.

laws/src/main/scala/cats/laws/AlternativeLaws.scala
```diff
@@ -3,22 +3,28 @@ package laws
...
 }
 
 object AlternativeLaws {
   def apply[F[_]](implicit ev: Alternative[F]): AlternativeLaws[F] =
     new AlternativeLaws[F] { def F: Alternative[F] = ev }
+
+  def composed[M[_], N[_]](implicit M: Alternative[M], N: Applicative[N]): AlternativeLaws[λ[α => M[N[α]]]] =
+    new AlternativeLaws[λ[α => M[N[α]]]] {
+      def F: Alternative[λ[α => M[N[α]]]] = M.compose[N]
+    }
```
satorg, line 29: Just a call-out: this introduces a new ability to validate laws for composed instances. This feature is brand new one, it hasn't existed before. Currently only composed `Alternative` and `NonEmptyAlternative` instances (see below) are supported to keep the PR scope narrow.

johnynek:
This is really good work.

It did get very large.

I'm now wondering if we can break it into two:

1. introduce NonEmptyAlternative and the laws.
2. override prependK/appendK to be specialized for more instances.

It is much harder to review larger PRs and if we have to revert, we revert more code at once. I would be on the lookout for ways we can decompose into separately mergeable parts.

What do you think?

satorg:
> I'm now wondering if we can break it into two:
> 
> 1. introduce NonEmptyAlternative and the laws.
> 2. override prependK/appendK to be specialized for more instances.

Agree, that makes sense to me too. I'll take on this.

"introduce NonEmptyAlternative and the laws" – do you mean to add `NonEmptyAlternative` typeclass along with `prependK`/`appendK` methods all together (but without their specialized versions)? Because otherwise the new typeclass is going to be just empty...

johnynek:
yeah, I was thinking add the default implementation of `prependK/appendK` using `pure` + `combineK`. And then a separate pr with all the specializations.

Just as one example way to split the PR.

core/src/main/scala/cats/syntax/all.scala
```diff
@@ -14,6 +14,7 @@ abstract class AllSyntaxBinCompat
 
 trait AllSyntax
     extends AlternativeSyntax
+    with NonEmptyAlternativeSyntax
```
satorg, line 17: Not sure, where to insert this new trait. I'd prefer to maintain all mix-in traits in some particular order, but I cannot figure out any specific order here.

core/src/main/scala/cats/syntax/package.scala
```diff
@@ -71,4 +71,5 @@ package object syntax {
   object vector extends VectorSyntax
   object writer extends WriterSyntax
   object set extends SetSyntax
+  object nonEmptyAlternative extends NonEmptyAlternativeSyntax
```
satorg, line 74: Seems that the initial intent was to keep all inner objects of object `syntax` just in alphabetic order by their names. But later the order was broken (perhaps due to an oversight). So I'm putting the new object to the end of all existing objects for now.

satorg:
@johnynek FYI: I truncated this PR to make it containing basic `prependK`/`appendK` changes only.

LukaJCB (approved):
Thank you so much!

johnynek (approved):

satorg:
@johnynek @LukaJCB thank you guys for following up this PR.

PR #4015, opened 2021-10-10
Implement `OptionT#foreachF`

armanbilge:


core/src/main/scala/cats/data/OptionT.scala
```diff
@@ -30,6 +30,13 @@ final case class OptionT[F[_], A](value: F[Option[A]]) {
   def foldF[B](default: => F[B])(f: A => F[B])(implicit F: FlatMap[F]): F[B] =
     F.flatMap(value)(_.fold(default)(f))
 
+  /**
+   * Transform this `OptionT[F, A]` into a `F[Unit]`.
+   * This is identical to `foldF(F.unit)(f)`.
+   */
+  def foreachF[B](f: A => F[Unit])(implicit F: Monad[F]): F[Unit] =
```
johnynek, line 37: side note, this `foldMap` where `B =:= F[Unit]`. But I think we can't implicitly resolve `Monoid[F[A]]` given `F[_]: Applicative` (since I don't think that resolves without additional imports).

johnynek, line 37: `B` type parameter is unused. Can we remove it?

armanbilge: Oops, right. Thanks.
```suggestion
  def foreachF(f: A => F[Unit])(implicit F: Monad[F]): F[Unit] =
```

PR #4018, opened 2021-10-19
Bump back to Scala.js 1.5.1

armanbilge:
Opening this for consideration in parallel to the discussion in #4016. The [scala-js-dom](https://github.com/scala-js/scala-js-dom) project for example specifically stays on SJS 1.5.x (and only abandoned SJS 1.0.x for Scala 3 support).


rossabaker:
I don't know the ramifications of a downgrade of Scala.js for something already published.  Is it true that whatever is selected here is not transitive to Scala.js apps or downstream libraries?

I'm a [deferential :+1:](https://github.com/typelevel/cats/pull/4016#issuecomment-947046024) if this doesn't affect downstream and a stronger :-1: if it could.

armanbilge:
> I don't know the ramifications of a downgrade of Scala.js for something already published.

Oh sorry, I should have clarified. The last cats release was on SJS 1.5.1.

https://github.com/typelevel/cats/blob/v2.6.1/project/plugins.sbt#L14

rossabaker:
Oh, cool.  For my education: say cats-2.6.1 were already on 1.7.1.  Would this have any affect outside this repo, other than enhancing its reach?

This is a one-line change, with no apparent user-facing disadvantage.  But:
* Do we lose any warnings?  This is why on the Scala 3.x debate, a Scala 3.latest run is proposed even if not published.
* Do we lose any efficiency of compiled artifacts?  I would think no, because that will be up to the app version at linking time?

djspiewak (approved):

PR #4032, opened 2021-10-31
Invariant[Fractional]

tmccarthy:
Fixes #3794.

This PR adds `Invariant[Fractional]`, and follows on from #3773, where `Invariant[Numeric]` and `Invariant[Integral]` were added.

PR #4033, opened 2021-11-01
Add small-domain floating-point type MiniFloat to discipline package

tmccarthy:
This PR adds a new data class to the `cats.laws.discipline` package: `MiniFloat`. Similar to `MiniInt`, `MiniFloat` is intended to provide a floating-point type whose domain is small enough to have an [`ExhaustiveCheck`](https://github.com/typelevel/cats/blob/8787d59f0c53e9ba6cfa5dea63da9d0a5aaecb43/laws/src/main/scala/cats/laws/discipline/ExhaustiveCheck.scala) instance.

This PR is a re-attempt at #3813, but I have broken out the introduction of `MiniFloat` into a separate PR.

Note that there is a failing test in the scala-native build. This is due to [a bug in scala-native](https://github.com/scala-native/scala-native/issues/2178) that was fixed in `0.4.1`. Accordingly, this PR depends on #4022.

#### How big is the domain of `MiniFloat?`

As written, `MiniFloat` has 14 values: `NaN`, `PositiveInfinity`, `NegativeInfinity`, `-8.0`, `-4.0`, `-2.0`, `-1.0`, `-0.5`, `0.0`, `0.5`, `1.0`, `2.0`, `4.0`, and `8.0`. This compares to `16` for `MiniInt`, and `2` for `Boolean`, which are the other types for which `ExhaustiveCheck` is defined.

#### Why not just use a 8- or 16-bit float?

These types have much larger cardinalities (roughly 2⁸ and 2¹⁶ respectively) than `MiniFloat` that make them too big to be used for exhaustive laws checks.

#### Why would `MiniFloat` be useful in the discipline package?

Since #2577, testing function equivalence using `Arbitrary` has been deprecated with `ExhaustiveCheck` now the preferred way to test. `MiniInt` serves this purpose, but I found a use for something like `MiniFloat` when defining an `Eq` instance for `Fractional`. Without something like `MiniFloat`, we have to fall back on using `Arbitrary` to define `Eq[Fractional[A]]` (see [here](https://github.com/typelevel/cats/pull/4032#discussion_r739895694)). More broadly, It seems like it would be nice to have another option for tests requiring an `ExhaustiveCheck` instance. It's also a nice small type that can be used to test the quirky behaviours of very large, very small, and infinite floating-point values.


tests/src/test/scala/cats/tests/MiniIntSuite.scala
```diff
@@ -24,7 +24,7 @@ class MiniIntSuite extends CatsSuite {
 
   {
     implicit val m: CommutativeMonoid[MiniInt] = miniIntMultiplication
-    checkAll("MiniInt addition", CommutativeMonoidTests[MiniInt].commutativeMonoid)
+    checkAll("MiniInt multiplication", CommutativeMonoidTests[MiniInt].commutativeMonoid)
```
tmccarthy, line 27: Unrelated change that fixes a typo in the tests for `MiniInt`.

laws/src/main/scala/cats/laws/discipline/MiniFloat.scala
```diff
@@ -0,0 +1,177 @@
...
+
+  /**
+   * Note that since `MiniFloat` is used primarily for tests, this `Eq` instance defines `NaN` as equal to itself. This
+   * differs from the `Order` defined for `Float`.
+   */
+  implicit val catsLawsEqInstancesForMiniFloat: Order[MiniFloat] with Hash[MiniFloat] =
+    new Order[MiniFloat] with Hash[MiniFloat] {
+      override def compare(x: MiniFloat, y: MiniFloat): Int = Order[Float].compare(x.toFloat, y.toFloat)
+      override def hash(x: MiniFloat): Int = Hash[Float].hash(x.toFloat)
+    }
```
tmccarthy, line 165: Because we're using this class in tests it's really only useful if we define `Eq` such that `NaN == NaN`. I'm aware that this is slightly unintuitive but otherwise it becomes useless.

An alternative approach would be to provide this `Eq` in such a way that it needs to be manually imported when used. I'd be happy to make that change as needed.

laws/src/main/scala/cats/laws/discipline/MiniFloat.scala
```diff
@@ -0,0 +1,177 @@
...
+
+  implicit val catsLawsExhaustiveCheckForMiniInt: ExhaustiveCheck[MiniFloat] = new ExhaustiveCheck[MiniFloat] {
+    override def allValues: List[MiniFloat] = MiniFloat.allValues
+  }
+
+  val miniFloatMax: CommutativeMonoid[MiniFloat] with BoundedSemilattice[MiniFloat] =
+    new CommutativeMonoid[MiniFloat] with BoundedSemilattice[MiniFloat] {
+      override def empty: MiniFloat = MiniFloat.NegativeInfinity
+      override def combine(x: MiniFloat, y: MiniFloat): MiniFloat = if (x > y) x else y
+    }
```
tmccarthy, line 175: Assuming we define `Eq[MiniFloat]` as above, I believe this is the main lawful Monoid you can define for `MiniFloat`. I'd be happy to provide instances for addition and multiplication (with appropriate tests and disclaimers) if requested, but I'm not sure they'd be super useful.

armanbilge:
Hi, thanks for the PR! Apologies, just working our way through some of the backlog.

Sorry if I missed it, but I'm confused about one point: do we actually need/use `MiniFloat` anywhere within Cats itself?

tmccarthy:
Thanks @armanbilge. This PR is obviously a bit old but I think it's in a good place and can still add value.

> Sorry if I missed it, but I'm confused about one point: do we actually need/use MiniFloat anywhere within Cats itself?

I've added changes to this PR since you had a look that use `MiniFloat` to test `Invariant[Fractional]` (see #4032). This is the where I originally ran into the need for this type.

In order to test whether our `Invariant[Fractional]` instance is lawful [in `AlgebraInvariantSuite`](https://github.com/typelevel/cats/blob/2e92db57cf726cba1804196110bb0f67ea832f65/tests/shared/src/test/scala/cats/tests/AlgebraInvariantSuite.scala#L231), we need to be able to define an `Eq[Fractional[A]]` for some `A`. The [existing code](https://github.com/typelevel/cats/blob/4e211c59cf7aff9049b05e180dc4e0596d65c876/tests/shared/src/test/scala/cats/tests/AlgebraInvariantSuite.scala#L185) uses `Float` and derives a `Eq[Fractional[Float]]`. Unfortunately this is done using `Arbitrary[Float]` and the deprecated `DeprecatedEqInstances.catsLawsEqForFn1` [here](https://github.com/typelevel/cats/blob/4e211c59cf7aff9049b05e180dc4e0596d65c876/tests/shared/src/test/scala-2.13%2B/cats/tests/ScalaVersionSpecific.scala#L246). I _believe_ this is the only place in the core cats tests that this deprecated instance is still used.

Ideally we would instead use [`catsLawsEqForFn1Exhaustive`](https://github.com/typelevel/cats/blob/4e211c59cf7aff9049b05e180dc4e0596d65c876/laws/src/main/scala/cats/laws/discipline/Eq.scala#L37), which has been the recommended approach since #2577. However, the key problem is that there is **currently no type for which both `ExhaustiveCheck` and `Fractional` are defined**. Hence, we are in need of a small-domain, floating-point type. One that has the behaviour of `Fractional`, but is small enough that we can define an `ExhaustiveCheck` instance. That is what this PR provides.

It's possible there are other ways to an `Eq[Fractional[A]]` and test the lawfulness of `Invariant[Fractional]`. If the decision here is that `MiniFloat` is confusing and doesn't add enough value I'll probably go away and try to find another solution. Because I think we need to get rid of the dependency on `DeprecatedEqInstances.catsLawsEqForFn1`.

tmccarthy:
> unless there are some other possible uses outside of Cats

Personally I think this is a fascinating data type that makes some of the interesting corner-cases of floating-point numbers a lot more obvious. But I'm not sure that's a great reason to put it in the core discipline package, particularly given I've had to trade away some of that behaviour in order to make it useful for testing (I've made `NaN == NaN`, which isn't the case for `Float` and `Double`).

> Out of curiosity could we use some sort of rational extended with infinities/NaN here? At least personally I could reason about that more easily than about floating point arithmetic :)

This might be doable but I think it might just be easier to cheat a little bit and use `Fractional[MiniInt]` to test the laws. It's a little spurious but it never escapes the test suite and it's a lot easier to reason about than this full `MiniFloat` type.

I have raised #4216 taking this approach which may be a more fruitful way of removing the `DeprecatedEqInstances`. I'm happy for this PR to be closed in lieu of that one.

tmccarthy:
For posterity's sake, I should note that the [Hash consistency test is flakey in Scala native](https://github.com/typelevel/cats/pull/4033/commits/32f06d0f2123058a6bbea8ff34f747c7810f2097#). It seems to work sometimes and not other times, even on the CI? If we ever return to this code in the future that'll need to be fixed before this is merged.

PR #4036, opened 2021-11-04
Adding Id.apply

BalmungSan:
I hope there isn't a reason for not doing this.

### DIFF
--- a/core/src/main/scala/cats/package.scala
+++ b/core/src/main/scala/cats/package.scala
@@ -64,6 +64,10 @@ package object cats {
    * encodes pure unary function application.
    */
   type Id[A] = A
+  object Id {
+    def apply[A](a: A): Id[A] = a
+  }
+
   type Endo[A] = A => A
 
   val catsInstancesForId: Bimonad[Id] with CommutativeMonad[Id] with NonEmptyTraverse[Id] with Distributive[Id] =
--- a/tests/src/test/scala/cats/tests/IdSuite.scala
+++ b/tests/src/test/scala/cats/tests/IdSuite.scala
@@ -2,6 +2,9 @@ package cats.tests
 
 import cats.{Bimonad, CommutativeMonad, Id, Reducible, Traverse}
 import cats.laws.discipline._
+import cats.syntax.applicative._
+import cats.syntax.eq._
+import org.scalacheck.Prop._
 
 class IdSuite extends CatsSuite {
   implicit val iso: SemigroupalTests.Isomorphisms[Id] =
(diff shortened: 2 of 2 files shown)


joroKr21:
You could just do `a: Id[A]`

BalmungSan:
@joroKr21 sure, but a lot of times I have wanted to use `Id` to then use `map` as a _pipe_ of pure functions to pure values. But then if I have to do `(foo: Id[Foo]).map(f)` looks worse and usually breaks this chain idea I had in mind in the first place.
The idea of this PR is to allow us to do `Id(foo).map(f)`

satorg:
@BalmungSan another way with help of the existing Cats syntax could be `a.pure[Id]`:

```scala
import cats._
import cats.syntax.all._

val a = 123
println(a.pure[Id].map(_ + 456))
```

https://scastie.scala-lang.org/satorg/Qq3pz0kaT8CLJqfTye2VBA/17

joroKr21:
Since Scala 2.13 there is also a standard library `pipe` syntax in `scala.util.chaining` btw

BalmungSan:
@satorg oh nice, I kind of always forget about it.
@joroKr21 right, I personally don't like the names the used but yeah that is a valid point.

I guess the point of the PR is not really if this is the best way or not to access it. But, rather why not having the `Id` object?
Even more, I think this could serve as the initial step of a bigger change of moving `Id` into `cats.data` and moving the typeclass instances of `Id` into its companion object. The problem is that such changes are source and binary breaking so they must wait and I am not sure if everyone agree if those make sense.

joroKr21:
Yeah it's definitely a valid point to discuss on it's own merits whether `Id.apply` should exist or not.
I'm just giving you some alternatives you can apply today.

Regarding the type class instances, that wouldn't work in Scala 2 anyway because type aliases can't have a companion object. I.e. you can have an `object Id`, it's just not a companion. A completely separate question is if `Id` should be an `opaque type` in a hypothetical Scala 3 only version of Cats.

tpolecat (approved):
I like this change. It's trivial and it provides an ergonomic improvement that I would use from time to time.

satorg:
I personally don't mind this change. Just to note: there's `IdSuite` as well. I wonder would it make sense to put a small test for this change there? Something like
```scala
property("Id#apply") {
  forAll { (a: Int) =>
    assert(Id(a) == (a: Id[Int])) // or `== a.pure[Id]` or even just `== a`
  }
}
```

BalmungSan:
Uhm that error is not related to my change, is it?

tpolecat:
I restarted the build.

rossabaker (approved):
There are other ways, but I've looked for this way before.  It's intuitive to me.

I wish it lived in `cats.data`, but it's not easy to get there from here in Cats 2.

tpolecat (approved):

PR #3073, opened 2019-09-19
add more doctest to WriterT

hamednourhani:
Re: Issue #2479

Add more doctest to WriterT methods:

- tell
- written
- value
- listen
- ap
- map
- imap
- mapK
- flatMap
- mapBoth
- bimap
- mapWritten
- swap
- reset
- show
- foldLeft
- foldRight
- traverse

PR #4038, opened 2021-11-07
DocTests for OptionT

FelAl:
https://github.com/typelevel/cats/issues/2479


core/src/main/scala/cats/data/OptionT.scala
```diff
@@ -101,30 +111,109 @@ final case class OptionT[F[_], A](value: F[Option[A]]) {
...
   def exists(f: A => Boolean)(implicit F: Functor[F]): F[Boolean] =
     F.map(value)(_.exists(f))
 
+  /**
+   * Example:
+   * {{{
+   *  scala> import cats.data.OptionT
+   *
+   *  scala> val optionT: OptionT[List, Int] = OptionT(List(Some(2), None, Some(414), None, None))
+   *  scala> optionT.filter(el => (el % 2 ==0 ))
```
rossabaker, line 136: No big deal, but the whitespace on these is a bit weird:

```suggestion
   *  scala> optionT.filter(el => (el % 2 == 0))
```

rossabaker (commented):
Thanks.  I :heart: doctests.

Since most of these are overrides, are we going to lose summaries of what they do for just "Example"?  I like examples, but I wonder if there should be a first sentence explaining `map`, `filter`, etc.  Have we added doctests like this for other data types as precedent?

FelAl:
@rossabaker thank you for the review.
I've checked previous PRs(https://github.com/typelevel/cats/pull/3053/files, https://github.com/typelevel/cats/pull/3073/files) and most of them have only section `Example`

rossabaker (dismissed):
:+1: Thanks!

johnynek:
looks like the collect examples don't yet work.

FelAl:
`sbt ++2.12.15 buildJVM bench/test` works locally right now without errors.

core/src/main/scala/cats/data/OptionT.scala
```diff
@@ -108,30 +118,109 @@ final case class OptionT[F[_], A](value: F[Option[A]]) {
   def getOrElseF[B >: A](default: => F[B])(implicit F: Monad[F]): F[B] =
     F.flatMap(value)(_.fold(default)(F.pure))
 
+  /**
+   * Example:
+   * {{{
+   *  scala> import cats.data.OptionT
+   *
+   *  scala> val optionT: OptionT[List, Int] = OptionT(List(Some(2), None, Some(414), None, None))
+   *  scala> optionT.collect{ case i: Int => i == 2 }
```
johnynek, line 127: I think this example would be better if you did `case i if i == 2 => i `

Because the way you wrote it this is the same as a map function?

core/src/main/scala/cats/data/OptionT.scala
```diff
@@ -108,30 +118,109 @@ final case class OptionT[F[_], A](value: F[Option[A]]) {
...
+   *
+   *  scala> optionT.filter(el => (el % 3 == 0))
+   *  res1: OptionT[List, Int] = OptionT(List(None, None, Some(414), None, None))
+   * }}}
+   */
   def filter(p: A => Boolean)(implicit F: Functor[F]): OptionT[F, A] =
     OptionT(F.map(value)(_.filter(p)))
 
+  /**
+   * Example:
```
johnynek, line 154: I think we should mention this is for `for` comprehensions which desugar using withFilter. Otherwise it is identical to filter.

johnynek (changes requested):

core/src/main/scala/cats/data/OptionT.scala
rossabaker: Just like @FelAl's is a glorified `map`, @johnynek's is a glorified `filter`.  Doing an inexhaustive match _and_ doing something other than returning `i` on the right would best show both those characteristics of `collect`.

FelAl: I've provided both examples.

core/src/main/scala/cats/data/OptionT.scala
FelAl: I've added a brief description.

FelAl:
Sorry for the delay.

rossabaker (approved):
Thanks for the touch-ups!

johnynek (approved):

johnynek:
the job seemed to have a transient failure on future tests.

Issue #4045, opened 2021-11-17
ApplicativeError final example - raiseError ignored


steve-e:
The final example for ApplicativeError is wrong IMO. It calls raiseError but the result is ignored.
```
def getTemperatureFromByCoordinatesAlternate[F[_]](x: (Int, Int))(implicit me: MonadError[F, String]): F[Int] = {
  if (x._1 < 0 || x._2 < 0) me.raiseError("Invalid Coordinates")
  for { c <- getCityClosestToCoordinate[F](x)
        t <- getTemperatureByCity[F](c) } yield t
}
```
Discussed briefly on gitter. Its just missing an else statement. Will create a PR referencing this issue

PR #4046, opened 2021-11-17
4045 Add missing else keyword to example

steve-e:
Minor correction to an example in documentation for ApplicativeError. Fixes [issue 4045](https://github.com/typelevel/cats/issues/4045)


LukaJCB (approved):

satorg (approved):
Agree, without `else` this example does not make any sense.

rossabaker (approved):
Thanks!

PR #4048, opened 2021-11-19
Rollback the rollback to Scala.js 1.5.1

rossabaker:
The build on #4018 was broken and merged anyway.  Scala.js compatibility has been good enough, and enough of the ecosystem has moved on, I no longer think it's worth trying to keep this one pinned.


satorg:
Yes, I can confirm: I tried running `sbt ++2.12.15 clean validateAllJS` locally before and after merging #4018.
It succeeds right before but fails on the merge commit.

satorg (approved):
GitHub actions have succeeded – seems it fixes the build indeed.

rossabaker:
Since the build is broken and affecting a couple other PRs, I'm going to merge this hastily.  If anyone doesn't like it, speak up here before the next release. :smile:

PR #4056, opened 2021-11-28
Add Semigroupal[Id] to implicit scope

joroKr21:
I missed that in #3906

I think this is the only instance that is missing.
The rest is covered by `Invariant` and `UnorderedFoldable`.


armanbilge (approved):

rossabaker (dismissed):

johnynek:
can we add a test that exercises resolving this that fails without this?

joroKr21:
In theory you'd have to test summoning all type class instances that exist for `Id` but I can add just one for `Semigroupal` 🤷

johnynek (approved):

Issue #4059, opened 2021-11-30
Missing @implicitNotFound for ApplicativeError and MonadError


satorg:
Check it out:
```
scala> import cats.syntax.all._

scala> none[Int].raiseTo[Option]
-- Error:
1 |none[Int].raiseTo[Option]
  |                         ^
  |                      Could not find an instance of Applicative for Option

```
But in fact, `Applicative[Option]` *DOES* exist in the scope.

The error message for the above should look like: "Could not find an instance of ApplicativeError[Option, Int]" instead.
The reason why it is incorrect is that `ApplicativeError` does not have `@implicitNotFound` annotation applied and compiler picks one up from its parent.

armanbilge:
These errors frequently trip me up, and not just in Cats proper. E.g. if something in the CE hierarchy is missing then it says you are missing `Monad` (and not `Concurrent` or `Async`).

djspiewak:
Scala 3 inherits these definitions while Scala 2 does not. This has generated… no small amount of confusion. :-) Basically the consequence of this is the fact that in any typeclass hierarchy which has one `@implicitNotFound` needs to have them on *all* classes.

satorg:
Seems that Scala 2 also uses the inherited annotation (notice the sentence in parentheses):
```
could not find implicit value for parameter F: cats.ApplicativeError[Option,Int] (Could not find an instance of Applicative for Option)
```
But the overall error message is a way easier to decipher.

joroKr21:
Why does `Applicative` need an `@implicitNotFound` with such a generic message? How is it different than the default?

joroKr21:
And is that an argument against inheritance of `@implicitNotFound` messages 🤔

armanbilge:
Yes, I agree, I'm pretty sure we don't need these `@implicitNotFound` if the messages are so generic.

I experimented in a scastie (2.13) and it seems the default is to use the FQCN?
https://scastie.scala-lang.org/TUftkqyJSyyOg0XkkcntRw
```scala
trait Applicative[F[_]]
trait Foo[A]
implicitly[Applicative[Foo]]
// could not find implicit value for parameter e: Playground.Applicative[Playground.Foo]
```

armanbilge:
Lol, apparently this inheritance issue was known at the time! https://github.com/typelevel/cats/issues/581#issuecomment-150054140
> (The bad part here is that I think this would require annotating every type class with some kind of message. Either that, or wiring it up to Simulacrum somehow.)

[...]

Issue #581, opened 2015-10-21
Better error message for missing implicits


non:
Lots of newcomers forget to import implicit instances (e.g. `cats.implicits._`) which causes lots of pain.

It might be nice if when `Functor[X]` is not found, the the message indicates that you may have failed to import it (and can even mention `cats.std.all._` or `cats.implicits._`).

What do you all think?

non:
(The bad part here is that I think this would require annotating every type class with some kind of message. Either that, or wiring it up to Simulacrum somehow.)

milessabin:
Big +1 for wiring up in Simulacrum.

erwan:
I just did a PR in Simulacrum (with help from @larsrh) : https://github.com/typelevel/simulacrum/pull/144

PR #4060, opened 2021-12-01
Remove `@implicitNotFound` annotations

armanbilge:
Find/replace fix for https://github.com/typelevel/cats/issues/4059.

### DIFF
--- a/alleycats-core/src/main/scala/alleycats/ConsK.scala
+++ b/alleycats-core/src/main/scala/alleycats/ConsK.scala
@@ -4,7 +4,6 @@ import cats.SemigroupK
 import simulacrum.typeclass
 import scala.annotation.implicitNotFound
 
-@implicitNotFound("Could not find an instance of ConsK for ${F}")
 @typeclass trait ConsK[F[_]] extends Serializable {
   def cons[A](hd: A, tl: F[A]): F[A]
 }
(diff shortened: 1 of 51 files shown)


satorg:
I am for it in general – such annotations should not be necessary. It is a compiler's job after all to provide us with meaningful error messages. I guess this annotation itself was invented in Scala long ago to alleviate some compiler weaknesses at that time.

But could we double check please what the error messages look like with this change in some relatively sophisticated cases for all Scala versions we support? Just to make sure we don't make things worse.

armanbilge:
I think this annotation is still useful, if you want to provide additional context (e.g., how to get that implicit). That seems to have been the original goal in https://github.com/typelevel/cats/issues/581, although it was never executed (?) and isn't relevant anymore anyway. But just saying what is missing is something the compiler does anyway, and I think (hope) has been doing for a while? :)

I tried some simple examples with a scastie in https://github.com/typelevel/cats/issues/4059#issuecomment-983600998 but not sure if that covers our concerns.

djspiewak:
Yeah I think this actually makes the error messages *better* because we get the fully-qualified names.

djspiewak (approved):

satorg (approved):
Makes sense to me. Thanks for looking into it.

PR #4063, opened 2021-12-01
Enable MiMa for Scala 3

### LINKED ISSUE #4062: cats 2.7.0 broke binary compatibility for Scala 3 (because mima wasn't run)
As discovered by @xuwei-k, see https://gist.github.com/xuwei-k/afe78856f95124c6ca5323beb7b32b5c from https://github.com/argonaut-io/argonaut/pull/476

Seems like mima isn't setup for scala 3:
```scala
> ++3.0.2 coreJVM/mimaReportBinaryIssues
[info] Setting Scala version to 3.0.2 on 32 projects.
[info] Excluded 14 projects, run ++ 3.0.2 -v for more details.
[info] Reapplying settings...
[info] set current project to cats (in build file:/home/smarter/opt/cats/)
[error] stack trace is suppressed; run last coreJVM / mimaPreviousClassfiles for the full output
[error] (coreJVM / mimaPreviousClassfiles) sbt.librarymanagement.ResolveException: Error downloading org.typelevel:cats-core_3:2.1.0
```
(it needs to use cats-core_3:2.6.1 since that's the only previously published version for Scala 3).

Note that this is absolutely not a bug in Scala 3: we intentionally use a different type erasure algorithm for intersection/compound types, because the Scala 2 algorithm was not commutative but also because it was dependent on compiler implementation details so we couldn't fully reproduce it even if we tried, cf https://github.com/lampepfl/dotty/blob/master/compiler/src/dotty/tools/dotc/core/unpickleScala2/Scala2Erasure.scala, https://github.com/lampepfl/dotty/issues/5139.

armanbilge:
Will also fix #4062. First commit should fail to demonstrate MiMa is running.

### DIFF
--- a/binCompatTest/src/main/scala/catsBC/MimaExceptions.scala
+++ b/binCompatTest/src/main/scala/catsBC/MimaExceptions.scala
@@ -19,7 +19,7 @@ object MimaExceptions {
     cats.data.Kleisli.catsDataCommutativeFlatMapForKleisli[Option, Int],
     cats.data.IRWST.catsDataStrongForIRWST[List, Int, Int, Int],
     cats.data.OptionT.catsDataMonadErrorMonadForOptionT[List],
-    FunctionK.lift(headOption),
+    // FunctionK.lift(headOption), // Doesn't work in Scala 3
     cats.data.OptionT.catsDataMonadErrorForOptionT[Either[String, *], String],
     cats.data.OptionT[Either[String, *], Int](Right(Some(17))).ensure("error")(_ => true),
     "blah".leftNec[Int],
--- a/build.sbt
+++ b/build.sbt
@@ -345,7 +345,7 @@ def mimaPrevious(moduleName: String, scalaVer: String, ver: String, includeCats1
     }
   }
   // Safety Net For Exclusions
-  lazy val excludedVersions: List[String] = List()
+  lazy val excludedVersions: List[String] = List("2.7.0") // cursed bincompat on Scala 3
 
   // Safety Net for Inclusions
   lazy val extraCats1Versions: List[String] = List("1.0.1", "1.1.0", "1.2.0", "1.3.1", "1.4.0", "1.5.0", "1.6.1")
--- a/build.sbt
+++ b/build.sbt
@@ -524,7 +524,13 @@ def mimaSettings(moduleName: String, includeCats1: Boolean = true) =
           exclude[DirectAbstractMethodProblem]("cats.free.ContravariantCoyoneda.k"),
           exclude[ReversedAbstractMethodProblem]("cats.free.ContravariantCoyoneda.k"),
           exclude[DirectAbstractMethodProblem]("cats.free.Coyoneda.k"),
-          exclude[ReversedAbstractMethodProblem]("cats.free.Coyoneda.k")
+          exclude[ReversedAbstractMethodProblem]("cats.free.Coyoneda.k"),
+          exclude[DirectMissingMethodProblem]("cats.free.ContravariantCoyoneda.unsafeApply"),
+          exclude[DirectMissingMethodProblem]("cats.free.ContravariantCoyoneda.ks"),
+          exclude[DirectMissingMethodProblem]("cats.free.ContravariantCoyoneda.unsafeApply"),
+          exclude[DirectMissingMethodProblem]("cats.free.Coyoneda.unsafeApply"),
+          exclude[DirectMissingMethodProblem]("cats.free.Coyoneda.ks"),
+          exclude[DirectMissingMethodProblem]("cats.free.Coyoneda.unsafeApply")
         )
     }
   )
--- a/build.sbt
+++ b/build.sbt
@@ -879,9 +879,16 @@ lazy val binCompatTest = project
     // see https://github.com/typelevel/cats/pull/3079#discussion_r327181584
     // see https://github.com/typelevel/cats/pull/3026#discussion_r321984342
     useCoursier := false,
-    addCompilerPlugin(("org.typelevel" %% "kind-projector" % kindProjectorVersion).cross(CrossVersion.full)),
+    libraryDependencies ++= {
+      if (isDotty.value) Nil
+      else
+        Seq(
+          compilerPlugin(("org.typelevel" %% "kind-projector" % kindProjectorVersion).cross(CrossVersion.full))
+        )
+    },
     libraryDependencies += mimaPrevious("cats-core", scalaVersion.value, version.value).last % Provided,
-    scalacOptions ++= (if (priorTo2_13(scalaVersion.value)) Seq("-Ypartial-unification") else Nil)
+    scalacOptions ++= (if (priorTo2_13(scalaVersion.value)) Seq("-Ypartial-unification") else Nil),
+    scalacOptions ++= (if (isDotty.value) Seq("-Ykind-projector") else Nil)
   )
   .settings(testingDependencies)
   .dependsOn(core.jvm % Test)
--- a/build.sbt
+++ b/build.sbt
@@ -91,7 +91,7 @@ ThisBuild / githubWorkflowBuild := Seq(
     WorkflowStep.Sbt(
       List("clean", "validateBC"), // cleaning here to avoid issues with codecov
       name = Some("Binary compatibility ${{ matrix.scala }}"),
-      cond = Some(JvmCond + " && " + Scala2Cond)
+      cond = Some(JvmCond)
     )
   )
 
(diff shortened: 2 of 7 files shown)


build.sbt
```diff
@@ -348,9 +348,12 @@ def mimaPrevious(moduleName: String, scalaVer: String, ver: String, includeCats1
   lazy val excludedVersions: List[String] = List()
 
   // Safety Net for Inclusions
-  lazy val extraVersions: List[String] = List("1.0.1", "1.1.0", "1.2.0", "1.3.1", "1.4.0", "1.5.0", "1.6.1")
+  lazy val extraCats1Versions: List[String] = List("1.0.1", "1.1.0", "1.2.0", "1.3.1", "1.4.0", "1.5.0", "1.6.1")
```
joroKr21, line 351: Btw checking so old versions doesn't cover new definitions (added since).
IMO we should check against the latest previous version.

armanbilge: We are, these are just extras :) in sbt try `show coreJVM/mimaPreviousArtifacts` to see the list.

joroKr21: Oh ok I see 👍 - why do we add the extras? To be extra sure?

armanbilge: /shrug :)

build.sbt
```diff
@@ -876,9 +879,16 @@ lazy val binCompatTest = project
...
+      if (isDotty.value) Nil
+      else
+        Seq(
+          compilerPlugin(("org.typelevel" %% "kind-projector" % kindProjectorVersion).cross(CrossVersion.full))
+        )
+    },
     libraryDependencies += mimaPrevious("cats-core", scalaVersion.value, version.value).last % Provided,
-    scalacOptions ++= (if (priorTo2_13(scalaVersion.value)) Seq("-Ypartial-unification") else Nil)
+    scalacOptions ++= (if (priorTo2_13(scalaVersion.value)) Seq("-Ypartial-unification") else Nil),
+    scalacOptions ++= (if (isDotty.value) Seq("-Ykind-projector") else Nil)
```
johnynek, line 891: is this the right order? isn't it `if (!isDotty.value)`?

armanbilge: Not sure TBH, but this works for 2.x and 3 locally.

armanbilge: See also https://docs.scala-lang.org/scala3/guides/migration/plugin-kind-projector.html

binCompatTest/src/main/scala/catsBC/MimaExceptions.scala
```diff
@@ -19,7 +19,7 @@ object MimaExceptions {
     cats.data.Kleisli.catsDataCommutativeFlatMapForKleisli[Option, Int],
     cats.data.IRWST.catsDataStrongForIRWST[List, Int, Int, Int],
     cats.data.OptionT.catsDataMonadErrorMonadForOptionT[List],
-    FunctionK.lift(headOption),
+    // FunctionK.lift(headOption), // Doesn't work in Scala 3
```
armanbilge, line 22: Anyone know anything about this?

build.sbt
```diff
@@ -339,18 +339,21 @@ def mimaPrevious(moduleName: String, scalaVer: String, ver: String, includeCats1
...
         semverBinCompatVersions(major, minor, patch)
-          .map { case (maj, min, pat) => s"$maj.$min.$pat" }
+          .collect { case (maj, min, pat) if !scalaVer.startsWith("3.") || (maj >= 2 && min >= 7) => s"$maj.$min.$pat" }
       case _ =>
         List.empty[String]
     }
   }
   // Safety Net For Exclusions
-  lazy val excludedVersions: List[String] = List()
+  lazy val excludedVersions: List[String] = List("2.7.0") // cursed bincompat on Scala 3
```
armanbilge, line 348: 2.7.0 is cursed right? Unless we only want to curse it for Scala 3.

binCompatTest/src/main/scala/catsBC/MimaExceptions.scala
joroKr21: That's a macro which is not implemented for Scala 3:
https://github.com/typelevel/cats/blob/main/core/src/main/scala-2.x/src/main/scala/cats/arrow/FunctionKMacros.scala

armanbilge: Is it okay to disable, or maybe we should split this test into Scala 2 / Scala 3?

joroKr21: I'm not sure what this is testing TBH

armanbilge:
How do we feel about static methods for Java? @johnynek you pushed for conservatism in https://github.com/typelevel/cats-effect/pull/2343#issuecomment-923469872.

```
[error] Cats core: Failed binary compatibility check against org.typelevel:cats-core_2.13:2.6.1! Found 2 potential problems
[error]  * static method catsInstancesForId()cats.Distributive in interface cats.Invariant does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("cats.Invariant.catsInstancesForId")
[error]  * static method catsInstancesForId()cats.Bimonad in class cats.package does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("cats.package.catsInstancesForId")
```

joroKr21:
That's the issue in #4062 we were discussing though - the issue is that the type signature changed.

armanbilge:
Yes, except it's now fixed for Scala. However, by marking the old method `private[cats]` dissapears the static forwarders which AFAIK are only used by Java. That's what MiMa is complaining about, not the non-static methods. Also note this is 2.13 and not 3.

core/src/main/scala/cats/package.scala
```diff
@@ -70,7 +70,12 @@ package object cats {
 
   type Endo[A] = A => A
 
-  val catsInstancesForId: Bimonad[Id] with CommutativeMonad[Id] with NonEmptyTraverse[Id] with Distributive[Id] =
+  @deprecated("retained for binary compatibility", "2.7.1")
+  private[cats] def catsInstancesForId
+    : Bimonad[Id] with CommutativeMonad[Id] with Comonad[Id] with NonEmptyTraverse[Id] with Distributive[Id] =
+    catsInstancesForIdBinCompat1
+  val catsInstancesForIdBinCompat1
+    : Bimonad[Id] with CommutativeMonad[Id] with NonEmptyTraverse[Id] with Distributive[Id] =
```
joroKr21, line 78: Why do we need this change? 🤔 - I thought only the forwarder in `Invariant` needs a change

joroKr21:
Here is another suggestion - add one more forwarder of type `Bimonad[Id] with CommutativeMonad[Id]` to `object Monad` and restore the old type of the forwarder in `object Invariant` 😄

smarter:
I wonder if bincompat with both cats-2.6.1_3 and cats-2.7.0_3 could be kept using `@targetName` to generate overloads (requiring a scala-2/ and scala-3/ version of the original file):
```scala
import scala.annotation.targetName
// ...
// 2.6.1 compat
private[cats] def catsInstancesForId: Distributive[Id] with Comonad[Id] = ...
// 2.7.0 compat
private[cats] @targetName("catsInstancesForId") def catsInstancesForIdCompat270:
     : Distributive[Id] with Bimonad[Id] with CommutativeMonad[Id] with NonEmptyTraverse[Id] = ...
// new method
implicit def catsInstancesForIdBinCompat1
    : Distributive[Id] with Bimonad[Id] with CommutativeMonad[Id] with NonEmptyTraverse[Id] = ...
```

satorg:
> I wonder if bincompat with both cats-2.6.1_3 and cats-2.7.0_3 could be kept using `@targetName` to generate overloads (requiring a scala-2/ and scala-3/ version of the original file):

But this annotation only exists in Scala3's library, right?

smarter:
Yes, that would be in a scala-3/ version of the file, the scala-2/ version only needs one version of catsInstancesForId to keep compat with 2.6.1 and 2.7.0.

armanbilge:
So, while main has moved on to `2.8.0-SNAPSHOT` I'm wondering if we should make a patch release of 2.7.1 with this bincompat fixed. Because otherwise, the 2.7 series will be a (small, Scala 3-only) break in the binary-compatibility of the 2.x series as a whole.

satorg:
@armanbilge just curious – is this PR still in progress?

armanbilge:
This PR needs feedback on the following to proceed:
- https://github.com/typelevel/cats/pull/4063#issuecomment-984096068
- https://github.com/typelevel/cats/pull/4063#issuecomment-984127262
- https://github.com/typelevel/cats/pull/4063#issuecomment-1000316919

satorg:
> So, while main has moved on to 2.8.0-SNAPSHOT I'm wondering if we should make a patch release of 2.7.1 with this bincompat fixed. ...

I would say no need to patch 2.7.x unless there're complains from people who use it in some prod-ish projects and encountered issues with. But I guess if there were such issues we would already know about it.

satorg:
> I wonder if bincompat with both cats-2.6.1_3 and cats-2.7.0_3 could be kept using @targetName to generate overloads.

From my prospective, if it is possible to maintain bin.compatibility without spitting sources into Scala2 and Scala3 versions then it would be the best bet. But I realize that in some cases it may not be feasible – then `@targetName` can come in handy of course.

armanbilge:
Superseded by https://github.com/typelevel/cats/pull/4160.

PR #4067, opened 2021-12-02
More efficient slidingN functions

### LINKED ISSUE #4066: SlidingN StackOverflow
Last year I added slidingN for usage as part of Advent of Code 2020 (#3727). Well Advent of Code 2021 has a problem involving sliding windows again. After trying `.sliding2` with a reasonably large list a stack overflow occurs 😅.

Example showing the issue in ammonite:
```scala
@ import $ivy.`org.typelevel::cats-core:2.7.0`
import $ivy.$

@ import cats.implicits._
import cats.implicits._

@ (1 to 1000).toList.sliding2
java.lang.StackOverflowError
  cats.Eval$.loop$1(Eval.scala:341)
  cats.Eval$.cats$Eval$$evaluate(Eval.scala:363)
  cats.Eval$Defer.value(Eval.scala:262)
  cats.FoldableNFunctions.$anonfun$sliding2$1(FoldableNFunctions.scala:33)
  cats.instances.ListInstances$$anon$1.loop$2(list.scala:72)
  cats.instances.ListInstances$$anon$1.$anonfun$foldRight$1(list.scala:72)
  cats.Eval$.loop$1(Eval.scala:341)
  cats.Eval$.cats$Eval$$evaluate(Eval.scala:363)
  cats.Eval$Defer.value(Eval.scala:262)
  cats.FoldableNFunctions.$anonfun$sliding2$1(FoldableNFunctions.scala:33)
  cats.instances.ListInstances$$anon$1.loop$2(list.scala:72)
  cats.instances.ListInstances$$anon$1.$anonfun$foldRight$1(list.scala:72)
  cats.Eval$.loop$1(Eval.scala:341)
  cats.Eval$.cats$Eval$$evaluate(Eval.scala:363)
  cats.Eval$Defer.value(Eval.scala:262)
```

I suspect this is due to the `Eval.Now` https://github.com/typelevel/cats/blob/main/project/Boilerplate.scala#L591 maybe this should be a `Eval.later`. It's been a while since I've done much functional programming, so I might be off here...

joroKr21:
We piggy-back on `toIterable`.
This is also an easy way to avoid stack overflow.

Fixes #4066

### DIFF
--- a/tests/src/test/scala/cats/tests/FoldableSuite.scala
+++ b/tests/src/test/scala/cats/tests/FoldableSuite.scala
@@ -465,7 +465,7 @@ class FoldableSuiteAdditional extends CatsSuite with ScalaVersionSpecificFoldabl
     val lazySum: Eval[Int] = F.foldRightDefer(large, boom[Int])((elem, acc) => acc.map(_ + elem))
   }
 
-  def checkMonadicFoldsStackSafety[F[_]](fromRange: Range => F[Int])(implicit F: Foldable[F]): Unit = {
+  def checkMonadicFoldsStackSafety[F[_]: Foldable](fromRange: Range => F[Int]): Unit = {
     def nonzero(acc: Long, x: Int): Option[Long] =
       if (x == 0) None else Some(acc + x)
 
@@ -479,15 +479,15 @@ class FoldableSuiteAdditional extends CatsSuite with ScalaVersionSpecificFoldabl
     val src = fromRange(1 to n)
 
     val foldMExpected = n.toLong * (n.toLong + 1) / 2
-    val foldMResult = F.foldM(src, 0L)(nonzero)
+    val foldMResult = src.foldM(0L)(nonzero)
     assert(foldMResult.get == foldMExpected)
 
     val existsMExpected = true
-    val existsMResult = F.existsM(src)(gte(n, _))
+    val existsMResult = src.existsM(gte(n, _))
     assert(existsMResult.get == existsMExpected)
 
     val forallMExpected = true
-    val forallMResult = F.forallM(src)(gte(0, _))
+    val forallMResult = src.forallM(gte(0, _))
     assert(forallMResult.get == forallMExpected)
 
     val findMExpected = Some(n)
@@ -497,8 +497,15 @@ class FoldableSuiteAdditional extends CatsSuite with ScalaVersionSpecificFoldabl
     val collectFirstSomeMExpected = Some(n)
     val collectFirstSomeMResult = src.collectFirstSomeM(gteSome(n, _))
     assert(collectFirstSomeMResult.get == collectFirstSomeMExpected)
+  }
 
-    ()
+  def checkSlidingNStackSafety[F[_]: Foldable](fromRange: Range => F[Int]): Unit = {
+    val n = 1000
+    val src = fromRange(1 to n)
+
+    val sliding2Expected = List.tabulate(n)(i => (i, i + 1)).tail
+    val sliding2Result = src.sliding2
+    assert(sliding2Result == sliding2Expected)
   }
 
   test(s"Foldable.iterateRight") {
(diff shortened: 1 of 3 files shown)


armanbilge:
Can we add the example from #4066 as a test?

joroKr21:
> Can we add the example from #4066 as a test?

Where should I add this? `FoldableSuite` is kinda weird

satorg:
Aside question: how come `slidingN` functions return `List[TupleN]` rather that `F[TupleN]` itself?

I mean, if `F[_]` is e.g. `LazyList` (which is still `Foldable`) then it will be enforced although it is not necessary to do for such a method.

johnynek:
`Foldable[F]` is really basically `ToList[F]` to answer your question @satorg .

There is no methods to *create* an `F[_]` with `Foldable[F]`, just take them apart.

johnynek (dismissed):

johnynek:
I approved, but I do think adding the test that caused a stack overflow would be good.

satorg:
I'm just thinking – wouldn't `Alternative` be a better bet for `sliding`-like functionality eventually?

This is a question for the future of course – not related to this PR specifically.
This PR is great on its own )

tests/src/test/scala/cats/tests/FoldableSuite.scala
```diff
@@ -497,8 +497,15 @@ class FoldableSuiteAdditional extends CatsSuite with ScalaVersionSpecificFoldabl
...
+  }
+
+  def checkSlidingNStackSafety[F[_]: Foldable](fromRange: Range => F[Int]): Unit = {
+    val n = 1000
+    val src = fromRange(1 to n)
 
-    ()
+    val sliding2Expected = List.tabulate(n)(i => (i, i + 1)).tail
+    val sliding2Result = src.sliding2
+    assert(sliding2Result == sliding2Expected)
```
armanbilge, line 508:
```suggestion
    assertEquals(sliding2Result, sliding2Expected)
```

armanbilge (dismissed):
Thanks for the quick fix :)

armanbilge (approved):

johnynek:
@satorg I don't see a function like this in cats, but I think we could add it:

```scala
trait Alternative[F[_]] ... {
  def fromFoldable[G[_], A](ga: G[A])(implicit G: Foldable[G]): F[A] = {
    @tailrec def loop(lst: List[A], left: F[A]): F[A] =
      lst match {
        case a :: as => appendK(left, a)
        case Nil => left
     }

    loop(G.toList(ga), empty)
  } 
}
```
which we could override with more efficient implementations as needed (such as going in the reverse direction for building list, or using a List.newBuilder).

satorg:
@johnynek what I really meat is that having both `Foldable` and `Alternative` for `F` it could be possible to implement something like this:
```scala
def slidingN[F[_]: Alternative : Foldable, A](fa: F[A]): F[(A, ..., A)]
```
while preserving laziness for those `F[_]` which support it (i.e. `LazyList`).

What I was thinking about – there's `combineEvalK` available via `Alternative` which pretends to support laziness.
But in fact as for now `combineEvalK` does not support laziness for lazy collections, unfortunately.
Therefore, lazy-aware implementation would also require a `FlatMap` instance to shift `combineK` into "lazy" context:
```scala
  def lazySliding3[F[_], A](fa: F[A])(implicit FA: Alternative[F], FF: Foldable[F], FM: FlatMap[F]): F[(A, A, A)] = {
    val it =
      FF.toIterable(fa).iterator.sliding(3).withPartial(false)
        .map(_.toList).map { case a1 :: a2 :: a3 :: Nil => (a1, a2, a3) }

    def go(): F[(A, A, A)] = {
      FA
        .pure(Eval.later {
          if (it.hasNext)
            FA.prependK(it.next(), go())
          else
            FA.empty[(A, A, A)]
        })
        .flatMap(_.value) // this trick shifts all prependK into "lazy" context for lazy collections
    }

    go()
  }
```

satorg:
The issue with `combineEvalK` I described here in Discord:
https://discord.com/channels/632277896739946517/632278570512678923/916869020534317076

I am not really sure it is the issue though – maybe it is an expected behavior )

Slakah:
Thanks for fixing this @joroKr21 - doing advent of code again, so picking up Scala again 😅.

This will no doubt come in handy for the inevitable sliding window solution!

PR #4075, opened 2021-12-07
Add foldF To IorT

isomarcte:
Bump version to 2.8.0-SNAPSHOT to account for new symbols.

### DIFF
--- a/version.sbt
+++ b/version.sbt
@@ -1,1 +1,1 @@
-ThisBuild / version := "2.7.1-SNAPSHOT"
+ThisBuild / version := "2.8.0-SNAPSHOT"
--- a/core/src/main/scala/cats/data/IorT.scala
+++ b/core/src/main/scala/cats/data/IorT.scala
@@ -13,7 +13,6 @@ final case class IorT[F[_], A, B](value: F[Ior[A, B]]) {
    *
    * Example:
    * {{{
-   * scala> import cats.implicits._
    * scala> import cats.data.{Ior, IorT}
    *
    * scala> val iorT: IorT[List, String, Int] = IorT[List, String, Int](List(Ior.Right(123),Ior.Left("abc"), Ior.Both("abc", 123)))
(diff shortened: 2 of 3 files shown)


johnynek (approved):

version.sbt
```diff
@@ -1 +1 @@
-ThisBuild / version := "2.7.1-SNAPSHOT"
+ThisBuild / version := "2.8.0-SNAPSHOT"
```
satorg, line 1: Is it a typo or something?

satorg: Or, sorry didn't noticed that you mentioned it in the description.

But how come we need to bump up the version here?

armanbilge: This PR introduces new APIs so the minor version should be incremented according to semantic versioning. 

> increment the MINOR version when you add functionality in a backwards compatible manner

https://semver.org/

satorg: Right... But I believed it should be done once right before the release, shouldn't it?
I mean, if someone else add some other API then with this approach the version will climb to 2.9.0-SNAPSHOT even though `2.8.0` has not been released.

But nevermind, I was just surprised that we mix code changes and version management in a single PR. But if it looks OK for you, I am also fine with it. The PR itself looks good to me othewise.

satorg (approved):

version.sbt
zarthross: @satorg if you go to change the minor version because you added APIs, you don't bump it if it ends in `0-SNAPSHOT` because you are already on a new minor version.  If you are on say `x.x.1-SNAPSHOT` then there is a `x.x.0` release and so you should bump it.

isomarcte: @satorg I'm not super familiar with the process that cats uses the manage this, but setting it to `2.8.0-SNAPSHOT` anytime prior to release (including now) should indicate that we have added at least one new symbol and if we add another one then we don't need to change the version. As opposed to `2.8.1-SNAPSHOT` which indicates the prior version was 2.8.0, thus we'd need a bump.

That said, if there is a different process I should follow here I happy to change that.

isomarcte: Thanks @zarthross, you phrased that better than I did.

satorg: @zarthross yes agree, that makes sense to me, thanks for the explanation.
@isomarcte agree, thanks for the PR!

armanbilge (approved):

PR #4076, opened 2021-12-08
Add documentation for `Bimonad`

gatear:
The first draft for `Bimonad` docs as required in #1801
Couldn't find a better example ..  i'm open to suggestions.


docs/src/main/mdoc/typeclasses/bimonad.md
```diff
@@ -0,0 +1,55 @@
+---
+layout: docs
+title:  "Bimonad"
+section: "typeclasses"
+source: "core/src/main/scala/cats/Bimonad.scala"
+scaladoc: "#cats.Bimonad"
+---
+# Bimonad
+
+The `Bimonad` trait directly extends `Monad` and `Comonad` without introducing new behaviour. `Bimonad` is
```
armanbilge, line 10: I wonder if this is confusing—even though it doesn't introduce new methods, as you note below it does have additional laws which do affect (restrict, actually) it's behavior relative to it's parent traits. Maybe prefer the word method here?

docs/src/main/mdoc/typeclasses/bimonad.md
```diff
@@ -0,0 +1,55 @@
...
+layout: docs
+title:  "Bimonad"
+section: "typeclasses"
+source: "core/src/main/scala/cats/Bimonad.scala"
+scaladoc: "#cats.Bimonad"
+---
+# Bimonad
+
+The `Bimonad` trait directly extends `Monad` and `Comonad` without introducing new behaviour. `Bimonad` is 
+different of other `Bi` typeclasses like `Bifunctor`, `Bifoldable` or `Bitraverse` where the prefix describes a
```
armanbilge, line 11:
```suggestion
different from other `Bi` typeclasses like `Bifunctor`, `Bifoldable` or `Bitraverse` where the prefix describes a
```

docs/src/main/mdoc/typeclasses/bimonad.md
```diff
@@ -0,0 +1,55 @@
...
+`F[_, _]`. The `Bimonad` is a `F[_]` and could be better seen as a dual monad i.e. something that is both a `Monad` and
```
armanbilge, line 12: "dual monad" is another one that I worry is confusing. For one, `Comonad` is not a `Monad`. Actually it is the dual of `Monad`!
https://typelevel.org/cats/typeclasses/comonad.html

docs/src/main/mdoc/typeclasses/bimonad.md
```diff
@@ -0,0 +1,55 @@
...
+different of other `Bi` typeclasses like `Bifunctor`, `Bifoldable` or `Bitraverse` where the prefix describes a
+`F[_, _]`. The `Bimonad` is a `F[_]` and could be better seen as a dual monad i.e. something that is both a `Monad` and 
+a `Comonad`.
+
+If you use `Bimonad` as a convenience type such that `def f[T[_] : Monad, Comonad, S](fa: T[S]): S` is re-written to
+`def f[T[_] : Bimonad, S](fa: T[S]): S` keep in mind `Bimonad` has its own added laws so something that is both monadic 
+and comonadic may not necessarily be a lawful `Bimonad`.
+
+###Eval as a Bimonad
+Eval is a lawful `Bimonad` so you can chain computations and `extract` the result at the end.
```
armanbilge, line 20: Just a suggestion, to show the relationship to both parents.
```suggestion
Eval is a lawful `Bimonad` so you can chain computations (like a `Monad`) and `extract` the result at the end (like a `Comonad`).
```

docs/src/main/mdoc/typeclasses/bimonad.md
```diff
@@ -0,0 +1,55 @@
...
+
+Note the equivalence:
+```scala mdoc
+import cats._
+import cats.data._
+import cats.implicits._
+
+Bimonad[Eval].pure(true).extract === Eval.now(true).value
```
armanbilge, line 28: I think what you are trying to show here is that `pure` is implemented as `now` and `extract` as `value`? One alternative would be to define `Bimonad[Eval]` here just for demonstrative purposes, WDYT?

docs/src/main/mdoc/typeclasses/bimonad.md
```diff
@@ -0,0 +1,55 @@
...
+    .flatMap(c => Bimonad[T].pure(c + " with option C"))
+    .extract
+```
+
+This will work with all types of `Eval`:
+```scala mdoc
+make(Eval.now("config"))
+make(Eval.later("config"))
+
+//String = config with option A with option B with option C
```
armanbilge, line 46: I'm not an `mdoc` expert but I'm pretty sure it will automatically render the result when it runs :) I think you can preview this by running `docs/mdoc --watch` in sbt but not 100% sure.
```suggestion
```

docs/src/main/mdoc/typeclasses/bimonad.md
```diff
@@ -0,0 +1,55 @@
...
+```
+
+Given that `Function0` or `NonEmptyList` are also lawful bimonads the following calls are also valid:
+```scala mdoc
+make(() => "config")
+make(NonEmptyList.one("config"))
+
+//String = config with option A with option B with option C
```
armanbilge, line 54:
```suggestion
```

docs/src/main/mdoc/typeclasses/bimonad.md
```diff
@@ -0,0 +1,55 @@
...
+
+This will work with all types of `Eval`:
+```scala mdoc
+make(Eval.now("config"))
+make(Eval.later("config"))
+
+//String = config with option A with option B with option C
+```
+
+Given that `Function0` or `NonEmptyList` are also lawful bimonads the following calls are also valid:
```
armanbilge, line 49:
```suggestion
`Function0` and `NonEmptyList` are also lawful bimonads so the following calls are also valid:
```

docs/src/main/mdoc/typeclasses/bimonad.md
gatear: Yep i agree. I'll use `without introducing new methods`

docs/src/main/mdoc/typeclasses/bimonad.md
gatear: The `DualMonad`\\`SelfDualMonad` thing i picked from old issues like #1297

I'll change it to:
```
The `Bimonad` is a `F[_]` thus the `Bi` prefix has a different meaning here: it's both a `Monad` and a `Comonad`.
```

docs/src/main/mdoc/typeclasses/bimonad.md
gatear: > I think what you are trying to show here is that `pure` is implemented as `now` and `extract` as `value`?
Yes

> One alternative would be to define Bimonad[Eval] here just for demonstrative purposes, WDYT?

Something like that ?
```scala
val evalBimonad = Bimonad[Eval]

evalBimonad.pure(true).extract === Eval.now(true).value
```

armanbilge: I was thinking like how they define `optionMonad` in the `Monad` docs:
http://typelevel.org/cats/typeclasses/monad.html

docs/src/main/mdoc/typeclasses/bimonad.md
gatear: Yep it was helpful. Thanks!

docs/src/main/mdoc/typeclasses/bimonad.md
```diff
@@ -0,0 +1,85 @@
...
+Eval is a lawful `Bimonad` so you can chain computations (like a `Monad`) and `extract` the result at the end (like a `Comonad`).
+
+Here is a possible implementation based on existing monad and comonad:
+```scala mdoc
+import cats._
+import cats.data._
+import cats.implicits._
+
+implicit def evalBimonad(implicit monad: Monad[Eval], comonad: Comonad[Eval]) =
+  new Bimonad[Eval] {
```
johnynek, line 36: `Eval[_]` is a bit of a bad choice here because the Comonad is not very safe. Calling `.value` on an Eval must be done with some care (it needs to be at O(1) stack depth or you lose the stack safety.  If you are calling this as part of a recursive call it often blows up).

`Id[_]` is also a Comonad, but a bit of a boring one. I wonder if `NonEmptyList[_]` would be a better example since it is non-trivial but also not somewhat unsafe in this way.

gatear: Hm ok, I was hesitating between `Eval[_]` and `NonEmptyList[_]` for the main example.
I'll just use `NonEmptyList[_]`.

docs/src/main/mdoc/typeclasses/bimonad.md
```diff
@@ -23,41 +23,41 @@ def f[T[_] : Bimonad, S](fa: T[S]): S
...
-    override def extract[A](x: Eval[A]): A =
-      x.value
+    override def extract[A](fa: NonEmptyList[A]): A =
+      fa.head
 
-    //use the coflatMap from the Eval comonad
-    override def coflatMap[A, B](fa: Eval[A])(f: Eval[A] => B): Eval[B] =
+    //use the coflatMap from the NonEmptyList comonad
+    override def coflatMap[A, B](fa: NonEmptyList[A])(f: NonEmptyList[A] => B): NonEmptyList[B] =
       comonad.coflatMap(fa)(f)
```
armanbilge, line 47: Actually we can directly call `flatMap` and `coflatMap` directly on the `NonEmptyList` :)

gatear: I was hopping this shows `flatMap` and `coflatMap` as originating from `monad` and `comonad`. 
Don't you think it's a good idea to delegate to the `monad` and `comonad` variables ?

armanbilge: Yes, I do see that now. But then isn't it confusing, because `pure` and `extract` are also defined in `monad` and `comonad`, so we should delegate those as well?

armanbilge: > I thought using one and head in the implementation will make this more obvious

I agree, I think this is the right way to do it. So, we can just delegate `flatMap` and `coflatMap` to the list without `monad`/`comonad` and maybe leave `tailRecM` as an exercise for the reader 😉

Anyway just my opinion, not really sure what the best thing to do here is. But I think separating the implementations of `pure` and `extract` makes it seem like they are inconsistent or different from the definitions in `monad` or `comonad`.

gatear: So the current `tailRecM` implementation for `NonEmptyList[_]` doesn't look trivial .. I mean I fear Its understanding detours too much and I agree it's a good idea to leave it as an exercise 👍 

> Anyway just my opinion, not really sure what the best thing to do here is. But I think separating the implementations of pure and extract makes it seem like they are inconsistent or different from the definitions in monad or comonad.

Yes, I will remove `monad` and `comonad`.

docs/src/main/mdoc/typeclasses/bimonad.md
```diff
@@ -0,0 +1,85 @@
...
+```scala
+def f[T[_] : Bimonad, S](fa: T[S]): S
+```
+keep in mind `Bimonad` has its own added laws so something that is both monadic
+and comonadic may not necessarily be a lawful `Bimonad`.
+
+### NonEmptyList as a Bimonad
+`NonEmptyList[_]` is a lawful `Bimonad` so you can chain computations (like a `Monad`) and `extract` the result at the end (like a `Comonad`).
+
+Here is a possible implementation based on existing monad and comonad:
```
armanbilge, line 29:
```suggestion
Here is a possible implementation:
```

docs/src/main/mdoc/typeclasses/bimonad.md
```diff
@@ -0,0 +1,85 @@
...
+```scala mdoc
+import cats._
+import cats.data._
+import cats.implicits._
+
+implicit def nelBimonad =
+  new Bimonad[NonEmptyList] {
+
+    //use NonEmptyList specific methods for creation and extraction
```
armanbilge, line 38: Maybe a good place to point out the `Bimonad` law about the relationship between `pure` and `extract`? :)

gatear:
```scala
    //nelBimonad needs to respect: `nelBimonad.extract(nelBimonad.pure(a)) <-> a`
``` 
Something like this ?

docs/src/main/mdoc/typeclasses/bimonad.md
```diff
@@ -0,0 +1,85 @@
...
+Here is a possible implementation:
+```scala mdoc
+import cats._
+import cats.data._
+import cats.implicits._
+
+implicit def nelBimonad =
+  new Bimonad[NonEmptyList] {
+
+    //in order to have a lawful bimonad `pure` and `extract` need to respect: `nelBimonad.extract(nelBimonad.pure(a)) <-> a`
```
armanbilge, line 38: Just a personal style nitpick but I prefer a space here between `//` and the comment 😅 

```suggestion
    // in order to have a lawful bimonad `pure` and `extract` need to respect: `nelBimonad.extract(nelBimonad.pure(a)) <-> a`
```

docs/src/main/mdoc/typeclasses/bimonad.md
```diff
@@ -0,0 +1,85 @@
...
+If you use `Bimonad` as a convenience type such that:
+```scala
+def f[T[_] : Monad, Comonad, S](fa: T[S]): S
+```
+is re-written to:
+```scala
+def f[T[_] : Bimonad, S](fa: T[S]): S
+```
+keep in mind `Bimonad` has its own added laws so something that is both monadic
+and comonadic may not necessarily be a lawful `Bimonad`.
```
armanbilge, line 24: I think this is important enough to include at the beginning ...

docs/src/main/mdoc/typeclasses/bimonad.md
```diff
@@ -0,0 +1,85 @@
...
+section: "typeclasses"
+source: "core/src/main/scala/cats/Bimonad.scala"
+scaladoc: "#cats.Bimonad"
+---
+# Bimonad
+
+The `Bimonad` trait directly extends `Monad` and `Comonad` without introducing new methods. `Bimonad` is
+different from other `Bi` typeclasses like `Bifunctor`, `Bifoldable` or `Bitraverse` where the prefix describes a
+`F[_, _]`. The `Bimonad` is a `F[_]` and the `Bi` prefix has a different meaning here: it's both a `Monad` and a `Comonad`.
+
```
armanbilge, line 13: ... like this:
```suggestion
Keep in mind `Bimonad` has its own added laws so something that is both monadic
and comonadic may not necessarily be a lawful `Bimonad`.

```

armanbilge (commented):
Thanks for iterating on all the feedback! Looking good, just a couple tiny things :)

armanbilge (approved):
Thanks for your contribution! :)

gatear:
@armanbilge is this ready to merge or do i need 2 reviews ? I want to start working on another doc for `Bifoldable` after this one is merged

armanbilge:
Another pair of eyes is always good, maybe @johnynek can take another look at the next opportunity :)

docs/src/main/mdoc/typeclasses/bimonad.md
```diff
@@ -0,0 +1,85 @@
...
+The `Bimonad` trait directly extends `Monad` and `Comonad` without introducing new methods.  `Bimonad` is
+different from other `Bi` typeclasses like `Bifunctor`, `Bifoldable` or `Bitraverse` where the prefix describes
+a `F[_, _]`. The `Bimonad` is a `F[_]` and the `Bi` prefix has a different meaning here: it's both a `Monad` and a `Comonad`.  
+Keep in mind `Bimonad` has its own added laws so something that is both monadic
+and comonadic may not necessarily be a lawful `Bimonad`.
+
+If you use `Bimonad` as a convenience type such that:
+```scala
+def f[T[_] : Monad, Comonad, S](fa: T[S]): S
```
johnynek, line 18: I think this should be `def f[T[_]: Monad: Comonad, S](fa: T[S]): S`

I think the comma here is making`Comonad` just a type parameter.

docs/src/main/mdoc/typeclasses/bimonad.md
```diff
@@ -0,0 +1,85 @@
...
+```
+is re-written to:
+```scala
+def f[T[_] : Bimonad, S](fa: T[S]): S
```
johnynek, line 22: nit: we generally don't put a space after the type, so we would change to: `f[T[_]: Bimonad, S]`

docs/src/main/mdoc/typeclasses/bimonad.md
```diff
@@ -0,0 +1,85 @@
...
+### NonEmptyList as a Bimonad
+`NonEmptyList[_]` is a lawful `Bimonad` so you can chain computations (like a `Monad`) and `extract` the result at the end (like a `Comonad`).
+
+Here is a possible implementation:
+```scala mdoc
+import cats._
+import cats.data._
+import cats.implicits._
+
+implicit def nelBimonad =
```
johnynek, line 35: I would make this a `implicit val` since it doesn't have any args.

docs/src/main/mdoc/typeclasses/bimonad.md
gatear: Yep my bad ..`:Monad :Comonad` is what I wanted

johnynek (approved):

johnynek:
Thanks!

Issue #4077, opened 2021-12-08
Most methods on `ParallelApOps` and `ParallelApplyOps` should only require a `NonEmptyParallel` instance 


jbwheatley:
Couldn't find any mention of this in the issue tracker, so apologies if it has been discussed before!

Currently the following syntax methods are defined: 

```scala
final class ParallelApOps[M[_], A](private val ma: M[A]) extends AnyVal {

  def &>[B](mb: M[B])(implicit P: Parallel[M]): M[B] =
    P.parProductR(ma)(mb)

  def <&[B](mb: M[B])(implicit P: Parallel[M]): M[A] =
    P.parProductL(ma)(mb)

  def parProductL[B](mb: M[B])(implicit P: Parallel[M]): M[A] =
    P.parProductL(ma)(mb)

  def parProductR[B](mb: M[B])(implicit P: Parallel[M]): M[B] =
    P.parProductR(ma)(mb)

  def parProduct[B](mb: M[B])(implicit P: Parallel[M]): M[(A, B)] =
    Parallel.parProduct(ma, mb)

  def parReplicateA(n: Int)(implicit P: Parallel[M]): M[List[A]] =
    Parallel.parReplicateA(n, ma)
}

final class ParallelApplyOps[M[_], A, B](private val mab: M[A => B]) extends AnyVal {
  def <&>(ma: M[A])(implicit P: Parallel[M]): M[B] =
    Parallel.parAp(mab)(ma)(P)

  def parAp(ma: M[A])(implicit P: Parallel[M]): M[B] =
    Parallel.parAp(mab)(ma)
}
```

However, `parProductR`, and `parProductL`, `parProduct` and `parAp` only require `NonEmptyParallel`, so the above should be: 

```scala
final class ParallelApOps[M[_], A](private val ma: M[A]) extends AnyVal {

  def &>[B](mb: M[B])(implicit P: NonEmptyParallel[M]): M[B] =
    P.parProductR(ma)(mb)

  def <&[B](mb: M[B])(implicit P: NonEmptyParallel[M]): M[A] =
    P.parProductL(ma)(mb)

  def parProductL[B](mb: M[B])(implicit P: NonEmptyParallel[M]): M[A] =
    P.parProductL(ma)(mb)

  def parProductR[B](mb: M[B])(implicit P: NonEmptyParallel[M]): M[B] =
    P.parProductR(ma)(mb)

[...]

PR #4078, opened 2021-12-08
weaken constraint on ParallelApOps and ParallelApplyOps methods

jbwheatley:
Addresses #4077, following the technique in #3997 to ensure bincompat

### DIFF
--- a/core/src/main/scala/cats/syntax/parallel.scala
+++ b/core/src/main/scala/cats/syntax/parallel.scala
@@ -237,33 +237,67 @@ final class ParallelUnorderedFlatSequenceOps[T[_], M[_], A](private val tmta: T[
     Parallel.parUnorderedFlatSequence(tmta)
 }
 
-final class ParallelApOps[M[_], A](private val ma: M[A]) extends AnyVal {
+final class ParallelApOps[M[_], A](protected val ma: M[A]) extends AnyVal with ParallelApOpsBinCompat0[M, A] {
 
-  def &>[B](mb: M[B])(implicit P: Parallel[M]): M[B] =
+  @deprecated("use a NonEmptyParallel-constrained version instead", "2.7.1")
+  protected def &>[B](mb: M[B])(implicit P: Parallel[M]): M[B] =
     P.parProductR(ma)(mb)
 
-  def <&[B](mb: M[B])(implicit P: Parallel[M]): M[A] =
+  @deprecated("use a NonEmptyParallel-constrained version instead", "2.7.1")
+  protected def <&[B](mb: M[B])(implicit P: Parallel[M]): M[A] =
     P.parProductL(ma)(mb)
 
-  def parProductL[B](mb: M[B])(implicit P: Parallel[M]): M[A] =
+  @deprecated("use a NonEmptyParallel-constrained version instead", "2.7.1")
+  protected def parProductL[B](mb: M[B])(implicit P: Parallel[M]): M[A] =
     P.parProductL(ma)(mb)
 
-  def parProductR[B](mb: M[B])(implicit P: Parallel[M]): M[B] =
+  @deprecated("use a NonEmptyParallel-constrained version instead", "2.7.1")
+  protected def parProductR[B](mb: M[B])(implicit P: Parallel[M]): M[B] =
     P.parProductR(ma)(mb)
 
-  def parProduct[B](mb: M[B])(implicit P: Parallel[M]): M[(A, B)] =
+  @deprecated("use a NonEmptyParallel-constrained version instead", "2.7.1")
+  protected def parProduct[B](mb: M[B])(implicit P: Parallel[M]): M[(A, B)] =
     Parallel.parProduct(ma, mb)
 
   def parReplicateA(n: Int)(implicit P: Parallel[M]): M[List[A]] =
     Parallel.parReplicateA(n, ma)
 }
 
-final class ParallelApplyOps[M[_], A, B](private val mab: M[A => B]) extends AnyVal {
-  def <&>(ma: M[A])(implicit P: Parallel[M]): M[B] =
+sealed private[syntax] trait ParallelApOpsBinCompat0[M[_], A] extends Any { self: ParallelApOps[M, A] =>
+  def &>[B](mb: M[B])(implicit P: NonEmptyParallel[M]): M[B] =
+    P.parProductR[A, B](ma)(mb)
+
+  def <&[B](mb: M[B])(implicit P: NonEmptyParallel[M]): M[A] =
+    P.parProductL[A, B](ma)(mb)
+
+  def parProductL[B](mb: M[B])(implicit P: NonEmptyParallel[M]): M[A] =
+    P.parProductL[A, B](ma)(mb)
+
+  def parProductR[B](mb: M[B])(implicit P: NonEmptyParallel[M]): M[B] =
+    P.parProductR[A, B](ma)(mb)
+
+  def parProduct[B](mb: M[B])(implicit P: NonEmptyParallel[M]): M[(A, B)] =
+    Parallel.parProduct(ma, mb)
+}
+
+final class ParallelApplyOps[M[_], A, B](protected val mab: M[A => B])
+    extends AnyVal
+    with ParallelApplyOpsBinCompat0[M, A, B] {
+  @deprecated("use a NonEmptyParallel-constrained version instead", "2.7.1")
+  protected def <&>(ma: M[A])(implicit P: Parallel[M]): M[B] =
     Parallel.parAp(mab)(ma)(P)
 
-  def parAp(ma: M[A])(implicit P: Parallel[M]): M[B] =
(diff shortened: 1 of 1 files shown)


core/src/main/scala/cats/syntax/parallel.scala
```diff
@@ -237,35 +237,69 @@ final class ParallelUnorderedFlatSequenceOps[T[_], M[_], A](private val tmta: T[
     Parallel.parUnorderedFlatSequence(tmta)
 }
 
-final class ParallelApOps[M[_], A](private val ma: M[A]) extends AnyVal {
+final class ParallelApOps[M[_], A](protected val ma: M[A]) extends AnyVal with ParallelApOpsBinCompat0[M, A] {
 
-  def &>[B](mb: M[B])(implicit P: Parallel[M]): M[B] =
+  @deprecated("use a NonEmptyParallel-constrained version instead", "2.7.1")
```
satorg, line 242: The project version is set to "2.8.0-SNAPSHOT", if I'm not mistaken. I guess it'd be better so set versions in these `@deprecated` to "2.8.0"

satorg:
It is all about syntax changes only, right? I think it may be worth to fix `SyntaxSuite` according to these changes too.

core/src/main/scala/cats/syntax/parallel.scala
```diff
@@ -32,8 +32,12 @@ trait ParallelSyntax extends TupleParallelSyntax {
   implicit final def catsSyntaxParallelSequence1[T[_]: Traverse, M[_], A](tma: T[M[A]]): ParallelSequenceOps1[T, M, A] =
     new ParallelSequenceOps1(tma)
 
+  @deprecated("Kept for binary compatibility", "2.8.0")
   implicit final def catsSyntaxParallelAp[M[_]: FlatMap, A](ma: M[A]): ParallelApOps[M, A] =
     new ParallelApOps(ma)
+
+  implicit final def catsSyntaxParallelAp1[M[_], A](ma: M[A]): ParallelApOps[M, A] =
+    new ParallelApOps(ma)
```
jbwheatley, line 40: `FlatMap` also didn't get used so have removed it as a constraint for using the `NonEmptyParallel` syntax

jbwheatley: oh wait, maybe it is..

core/src/main/scala/cats/syntax/parallel.scala
```diff
@@ -32,8 +32,12 @@ trait ParallelSyntax extends TupleParallelSyntax {
   implicit final def catsSyntaxParallelSequence1[T[_]: Traverse, M[_], A](tma: T[M[A]]): ParallelSequenceOps1[T, M, A] =
     new ParallelSequenceOps1(tma)
 
+  @deprecated("Kept for binary compatibility", "2.8.0")
   implicit final def catsSyntaxParallelAp[M[_]: FlatMap, A](ma: M[A]): ParallelApOps[M, A] =
```
armanbilge, line 36:
```suggestion
  @deprecated("Kept for binary compatibility", "2.8.0")
  private[syntax] final def catsSyntaxParallelAp[M[_]: FlatMap, A](ma: M[A]): ParallelApOps[M, A] =
```

armanbilge: I think you reverted this because MiMa was complaining about missing static methods? I think it is is safe to filter those, because they only affect Java and I think it would be unusual for Java code to be using these syntax extensions.

satorg: I think it may be okay to leave it public if `private[syntax]` breaks BC. The only thing – the `implicit` modifier should be pulled off the identifier, but as I can see, it is already done here.

core/src/main/scala/cats/syntax/parallel.scala
```diff
@@ -237,35 +241,69 @@ final class ParallelUnorderedFlatSequenceOps[T[_], M[_], A](private val tmta: T[
...
+  protected def parProduct[B](mb: M[B])(implicit P: Parallel[M]): M[(A, B)] =
     Parallel.parProduct(ma, mb)
 
   def parReplicateA(n: Int)(implicit P: Parallel[M]): M[List[A]] =
     Parallel.parReplicateA(n, ma)
 }
 
-final class ParallelApplyOps[M[_], A, B](private val mab: M[A => B]) extends AnyVal {
-  def <&>(ma: M[A])(implicit P: Parallel[M]): M[B] =
+sealed private[syntax] trait ParallelApOpsBinCompat0[M[_], A] extends Any { self: ParallelApOps[M, A] =>
```
satorg, line 270: I'm thinking about a thing... Looks like these updated methods are not designed for the `Parallel` typeclass anymore. They all target `NonEmptyParallel` instead. So perhaps it would be better to create a separate extension holder for that, e.g.
```
final class NonEmptyParallelApOps[M[_], A] private[syntax] (private val ma: M[A]) extends AnyVal {
  def &>[B](mb: M[B])(implicit P: NonEmptyParallel[M]): M[B] = /* and so on */  
}
```
And the former class would remain almost the same as before with the corresponding methods hidden and deprecated:
```
final class ParallelApOps[M[_], A](private val ma: M[A]) extends AnyVal {

  @deprecated("use a NonEmptyParallel-constrained version instead", "2.8.0")
  protected def &>[B](mb: M[B])(implicit P: Parallel[M]): M[B] = // ....
```

In that case we wouldn't need such a quite weird `ParallelApOpsBinCompat0` mixin trait anymore and the overall syntax would become a bit clearer.

And the same is applicable for `ParallelApplyOps` below.

So wdyt?

jbwheatley: sure that makes sense. I'll do that

core/src/main/scala/cats/syntax/parallel.scala
```diff
@@ -258,14 +269,45 @@ final class ParallelApOps[M[_], A](private val ma: M[A]) extends AnyVal {
     Parallel.parReplicateA(n, ma)
 }
 
-final class ParallelApplyOps[M[_], A, B](private val mab: M[A => B]) extends AnyVal {
+final class ParallelApOps1[M[_], A](protected val ma: M[A]) extends AnyVal {
+  def parReplicateA(n: Int)(implicit P: Parallel[M]): M[List[A]] =
+    Parallel.parReplicateA(n, ma)
+}
```
satorg, line 275: This is absolutely disputable, so feel free to share your thoughts)
I'm wondering, what are the advantages we get if create a separate `ParallelApOps1` rather than keep `ParallelApOps` only and hide and deprecate only those methods which should go to `NonEmptyParallelApOps`?

My concern here is that we will end up with 3 classes in result jar: `ParallelApOps`, `ParallelApOps1`,  `NonEmptyParallelApOps` comparing to just 2 classes (`ParallelApOps` and `NonEmptyParallelApOps`) otherwise. Unfortunately, we cannot reduce this amount to just 1 due to BC, but 2 seems better to me than 3 unless there are some advantages brought up.

core/src/main/scala/cats/syntax/parallel.scala
```diff
@@ -258,14 +269,45 @@ final class ParallelApOps[M[_], A](private val ma: M[A]) extends AnyVal {
...
+
+final class NonEmptyParallelApOps[M[_], A](protected val ma: M[A]) extends AnyVal {
```
satorg, line 277: `protected val ma: M[A]` – I would suggest to keep everything `private` unless it is necessary to relax the access for some reason.

core/src/main/scala/cats/syntax/parallel.scala
jbwheatley: I was really just trying to mirror the other instances in this file, e.g. `ParallelTraversableOps`, but can do it with only 2 classes instead :)

tests/src/test/scala/cats/tests/SyntaxSuite.scala
```diff
@@ -184,7 +190,6 @@ object SyntaxSuite {
     val mb4: M[B] = ma.parProductR(mb)
     val mab2: M[(A, B)] = ma.parProduct(mb)
     val mb5: M[B] = mab.parAp(ma)
```
satorg, line 192: Would you also add a test for `parProductL` here please? Looks like it didn't exist before, but since you got your hands dirty with this piece of code, it would be nice to improve it too )

jbwheatley: its already there L189

satorg (dismissed):
Looks really good to me. Thank you for investing your time into it.

jbwheatley:
any clue why the tests aren't compiling now? can't spot it and intellij is happy 🤔

satorg:
Looks like the tests are failing for Scala 2.12.15, but IntelliJ compiles with Scala 2.13.7 as this is a default Scala version.

satorg:
Try `sbt ++2.12.15 testsJVM/test` command

jbwheatley:
latest commit seems to be necessary to get the syntax to work on 2.12.15

PR #4083, opened 2021-12-11
Reduce compilation warnings in `tests`

satorg:
This is an attempt to improve code hygiene a bit by reducing amount of warnings while compiling the `testsJVM`, `testsJS` and `testsNative` modules.

No tests are removed nor disabled, no other changes besides the `tests*` modules. But some tests were tossed over different suites or suite mixins.

Caveat: because `@nowarn` annotation is not supported in Scala 3.0.x, for this cross-build configuration most of the warnings will still remain. But things should get improved once the project upgrades to Scala 3.1.x.

Note: also in all changed files the "Organize Imports" (vs-code+metals) command was applied.

### DIFF
--- a/tests/src/test/scala/cats/tests/FoldableSuite.scala
+++ b/tests/src/test/scala/cats/tests/FoldableSuite.scala
@@ -1,23 +1,17 @@
 package cats.tests
 
 import cats._
-import cats.data.{Const, EitherK, IdT, Ior, Nested, NonEmptyList, NonEmptyStream, NonEmptyVector, OneAnd, Validated}
+import cats.data._
 import cats.instances.order._
-import cats.kernel.{Eq, Monoid}
 import cats.kernel.compat.scalaVersionSpecific._
 import cats.laws.discipline.arbitrary._
-import cats.syntax.alternative._
-import cats.syntax.either._
-import cats.syntax.foldable._
-import cats.syntax.functor._
-import cats.syntax.list._
-import cats.syntax.reducible._
-import cats.syntax.semigroupk._
+import cats.syntax.all._
 import org.scalacheck.Arbitrary
-import scala.collection.immutable.{SortedMap, SortedSet}
-import scala.util.Try
-import cats.syntax.eq._
 import org.scalacheck.Prop._
+
+import scala.collection.immutable.SortedMap
+import scala.collection.immutable.SortedSet
+import scala.util.Try
 
 @suppressUnusedImportWarningForScalaVersionSpecific
 abstract class FoldableSuite[F[_]: Foldable](name: String)(implicit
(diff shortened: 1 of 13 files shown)


satorg:
This PR is still awaiting its hero who would dare to review it :)

tests/src/test/scala/cats/tests/FoldableSuite.scala
```diff
@@ -554,6 +550,20 @@ class FoldableSuiteAdditional extends CatsSuite with ScalaVersionSpecificFoldabl
     checkSlidingNStackSafety(xs => NonEmptyVector.fromVectorUnsafe(xs.toVector))
   )
 
+  test("Foldable[List] doesn't break substitution") {
+    val result = List.range(0, 10).foldM(List.empty[Int])((accum, elt) => Eval.always(elt :: accum))
+
+    assert(result.value == result.value)
```
danicheg, line 556: I see that it wasn't a goal of this PR, but could we use `assertEquals` instead of `assert`? Then we would have pretty-printed diff on errors for free.

satorg: Oh.. This test looks changed because I moved many of its neighbors into  `FoldableSuiteAdditionalStreamSpecific` below, but this one remained in the place since it is not related to `Stream`. So I just didn't touch it actually.

Regarding your question: this particular assertion looks quite weird to me because it compares `result.value` with itself. Therefore, to me it looks like as it always produces `true` and changing it to `assertEquals` wouldn't make any difference. Perhaps, I'm just not getting what it was supposed to test. Could you suggest please?

danicheg: I was talking about not only the current assertion, rather about every `assert` as a whole.

satorg: I agree with you in general – using `assertEquals(a, b)` instead of `assert(a == b)` or even `assert(a === b)` should be a better choice in most cases. I really wonder why Cats tests prefer using the latter assertion style.

If you believe it would be a good idea, I am up to replace `assert` with `assertEquals` in those test cases which are directly affected by the PR's primary goal – I'd like to have `assertEquals` there too.

But I would hesitate to do it for all the Cats tests in general. It would blow up the scope of this PR and would mimic its primary changes. I think it is worth a separate PR – wdyt?

danicheg: Probably you're right, it's better to do it separately. By and large, I can tackle that issue.

satorg:
Hi folks! May I draw a fraction of your attention to this PR please? There's nothing dangerous with it – it just reduces compilation warnings from tests which can help to alleviate developer experience tremendously while they're working on any tests as well as coding in any IDE.

rossabaker (approved):
Thanks!

johnynek (approved):

satorg:
@rossabaker @johnynek thank you guys!

johnynek:
Thank you! Sorry for the latency!

PR #4084, opened 2021-12-15
SemigroupK/MonoidK methods

TimWSpence:
Ports missing methods from `Semigroup`/`Monoid` to `SemigroupK`/`MonoidK`

### DIFF
--- a/core/src/main/scala/cats/Alternative.scala
+++ b/core/src/main/scala/cats/Alternative.scala
@@ -104,7 +104,7 @@ import cats.kernel.compat.scalaVersionSpecific._
   def fromIterableOnce[A](as: IterableOnce[A]): F[A] =
     combineAllK(as.iterator.map(pure(_)))
 
-  def fromFoldable[G[_]: Foldable, A](as: G[A]): F[A] =
+  final def fromFoldable[G[_]: Foldable, A](as: G[A]): F[A] =
     fromIterableOnce(Foldable[G].toIterable(as))
 }
 
--- a/core/src/main/scala/cats/Alternative.scala
+++ b/core/src/main/scala/cats/Alternative.scala
@@ -1,6 +1,7 @@
 package cats
 
 import simulacrum.typeclass
+import cats.kernel.compat.scalaVersionSpecific._
 
 @typeclass trait Alternative[F[_]] extends NonEmptyAlternative[F] with MonoidK[F] { self =>
 
@@ -99,8 +100,15 @@ import simulacrum.typeclass
       val F = self
       val G = Applicative[G]
     }
+
+  def fromIterableOnce[A](as: IterableOnce[A]): F[A] =
+    combineAllK(as.iterator.map(pure(_)))
+
+  def fromFoldable[G[_]: Foldable, A](as: G[A]): F[A] =
+    fromIterableOnce(Foldable[G].toIterable(as))
 }
 
+@suppressUnusedImportWarningForScalaVersionSpecific
 object Alternative {
 
   /* ======================================================================== */
--- a/core/src/main/scala/cats/MonoidK.scala
+++ b/core/src/main/scala/cats/MonoidK.scala
@@ -41,12 +41,12 @@ import simulacrum.typeclass
    *
    * Example:
    * {{{
-   * scala> import cats.kernel.instances.string._
+   * scala> import cats.kernel.instances.list._
    *
-   * scala> Monoid[String].isEmpty("")
+   * scala> MonoidK[List].isEmpty(List.empty[String])
    * res0: Boolean = true
    *
-   * scala> Monoid[String].isEmpty("something")
+   * scala> MonoidK[List].isEmpty(List("something"))
    * res1: Boolean = false
    * }}}
    */
--- a/core/src/main/scala/cats/MonoidK.scala
+++ b/core/src/main/scala/cats/MonoidK.scala
@@ -99,7 +99,7 @@ import simulacrum.typeclass
    * }}}
    */
   override def combineNK[A](a: F[A], n: Int): F[A] =
-    if (n < 0) throw new IllegalArgumentException("Repeated combining for monoids must have n >= 0")
+    if (n < 0) throw new IllegalArgumentException("Repeated combining for monoidKs must have n >= 0")
     else if (n == 0) empty[A]
     else repeatedCombineNK(a, n)
 
--- a/core/src/main/scala/cats/MonoidK.scala
+++ b/core/src/main/scala/cats/MonoidK.scala
@@ -116,10 +116,10 @@ import simulacrum.typeclass
    * }}}
    */
   def combineAllK[A](as: IterableOnce[F[A]]): F[A] =
-    as.iterator.foldLeft(empty[A])(combineK[A])
-
-  override def combineAllOptionK[A](as: IterableOnce[F[A]]): Option[F[A]] =
-    if (as.iterator.isEmpty) None else Some(combineAllK(as))
+    combineAllOptionK(as) match {
+      case Some(fa) => fa
+      case None => empty[A]
+    }
 
   override def reverse: MonoidK[F] =
     new MonoidK[F] {
--- a/core/src/main/scala/cats/MonoidK.scala
+++ b/core/src/main/scala/cats/MonoidK.scala
@@ -118,7 +118,7 @@ import simulacrum.typeclass
   def combineAllK[A](as: IterableOnce[F[A]]): F[A] =
     combineAllOptionK(as) match {
       case Some(fa) => fa
-      case None => empty[A]
+      case None     => empty[A]
     }
 
   override def reverse: MonoidK[F] =
--- a/core/src/main/scala/cats/MonoidK.scala
+++ b/core/src/main/scala/cats/MonoidK.scala
@@ -97,7 +97,7 @@ import simulacrum.typeclass
    * scala> SemigroupK[List].combineNK(List(1), 5)
    * res0: List[Int] = List(1,1,1,1,1)
 
-   * scala> MonoidK[List].combineN(List("ha"), 0)
+   * scala> MonoidK[List].combineNK(List("ha"), 0)
    * res1: List[String] = List()
    *
    * }}}
@@ -117,7 +117,7 @@ import simulacrum.typeclass
    * scala> MonoidK[List].combineAllK(List(List("One"), List("Two"), List("Three")))
    * res0: List[String] = List("One", List("Two"), List("Three"))
    *
-   * scala> MonoidK[List].combineAll[String](List.empty)
+   * scala> MonoidK[List].combineAllK[String](List.empty)
    * res1: List[String] = List()
    * }}}
    */
--- a/core/src/main/scala/cats/MonoidK.scala
+++ b/core/src/main/scala/cats/MonoidK.scala
@@ -37,6 +37,23 @@ import simulacrum.typeclass
   def empty[A]: F[A]
 
   /**
+   * Tests if `a` is the identity.
+   *
+   * Example:
+   * {{{
+   * scala> import cats.kernel.instances.string._
+   *
+   * scala> Monoid[String].isEmpty("")
+   * res0: Boolean = true
+   *
+   * scala> Monoid[String].isEmpty("something")
+   * res1: Boolean = false
+   * }}}
+   */
+  def isEmpty[A](a: F[A])(implicit ev: Eq[F[A]]): Boolean =
+    ev.eqv(a, empty)
+
+  /**
    * Given a type A, create a concrete Monoid[F[A]].
    *
    * Example:
@@ -66,6 +83,54 @@ import simulacrum.typeclass
   override def compose[G[_]]: MonoidK[λ[α => F[G[α]]]] =
     new ComposedMonoidK[F, G] {
       val F: MonoidK[F] = self
+    }
+
+  /**
+   * Return `a` appended to itself `n` times.
+   *
+   * Example:
+   * {{{
+   * scala> import cats.kernel.instances.string._
+   *
+   * scala> Monoid[String].combineN("ha", 3)
+   * res0: String = hahaha
+   *
+   * scala> Monoid[String].combineN("ha", 0)
+   * res1: String = ""
+   * }}}
+   */
+  override def combineNK[A](a: F[A], n: Int): F[A] =
+    if (n < 0) throw new IllegalArgumentException("Repeated combining for monoids must have n >= 0")
+    else if (n == 0) empty[A]
+    else repeatedCombineNK(a, n)
+
+  /**
+   * Given a sequence of `as`, sum them using the monoid and return the total.
+   *
+   * Example:
+   * {{{
+   * scala> import cats.kernel.instances.string._
+   *
+   * scala> Monoid[String].combineAll(List("One ", "Two ", "Three"))
+   * res0: String = One Two Three
+   *
+   * scala> Monoid[String].combineAll(List.empty)
+   * res1: String = ""
+   * }}}
+   */
+  def combineAllK[A](as: IterableOnce[F[A]]): F[A] =
+    as.iterator.foldLeft(empty[A])(combineK[A])
+
+  override def combineAllOptionK[A](as: IterableOnce[F[A]]): Option[F[A]] =
+    if (as.iterator.isEmpty) None else Some(combineAllK(as))
+
+  override def reverse: MonoidK[F] =
+    new MonoidK[F] {
+      def empty[A] = self.empty
+      def combineK[A](a: F[A], b: F[A]) = self.combineK(b, a)
+      // a + a + a + ... is the same when reversed
+      override def combineNK[A](a: F[A], n: Int): F[A] = self.combineNK(a, n)
+      override def reverse = self
     }
 }
 
--- a/core/src/main/scala/cats/MonoidK.scala
+++ b/core/src/main/scala/cats/MonoidK.scala
@@ -43,8 +43,6 @@ import simulacrum.typeclass
    *
    * Example:
    * {{{
-   * scala> import cats.kernel.instances.list._
-   *
    * scala> MonoidK[List].isEmpty(List.empty[String])
    * res0: Boolean = true
    *
@@ -92,10 +90,8 @@ import simulacrum.typeclass
    *
    * Example:
    * {{{
-   * scala> import cats.kernel.instances.list._
-   *
    * scala> SemigroupK[List].combineNK(List(1), 5)
-   * res0: List[Int] = List(1,1,1,1,1)
+   * res0: List[Int] = List(1, 1, 1, 1, 1)
 
    * scala> MonoidK[List].combineNK(List("ha"), 0)
    * res1: List[String] = List()
@@ -112,10 +108,8 @@ import simulacrum.typeclass
    *
    * Example:
    * {{{
-   * scala> import cats.kernel.instances.list._
-   *
    * scala> MonoidK[List].combineAllK(List(List("One"), List("Two"), List("Three")))
-   * res0: List[String] = List("One", List("Two"), List("Three"))
+   * res0: List[String] = List(One, Two, Three)
    *
    * scala> MonoidK[List].combineAllK[String](List.empty)
    * res1: List[String] = List()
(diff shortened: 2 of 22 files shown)


armanbilge (commented):
Thanks for the PR! Just a quick question, do we need to add anything to laws to cover these new methods? E.g. to check the consistency of these implementations.

TimWSpence:
> Thanks for the PR! Just a quick question, do we need to add anything to laws to cover these new methods? E.g. to check the consistency of these implementations.

Good question! I was waiting to see if the tests passed before commenting 😅  I had two observations to make:
1. `SemigroupK` and `MonoidK` don't have `XXXFunctions` traits to add forwarders to the companion objects. Should I add those as well?
2. I _was_ going to comment that these functions for `Monoid` and `Semigroup` don't seem to have any tests but now I see that they do in fact have laws so yes, I'll add corresponding laws for these 👍

core/src/main/scala/cats/MonoidK.scala
```diff
@@ -67,8 +84,54 @@ import simulacrum.typeclass
...
+   * scala> SemigroupK[List].combineNK(List(1), 5)
+   * res0: List[Int] = List(1, 1, 1, 1, 1)
+
+   * scala> MonoidK[List].combineNK(List("ha"), 0)
+   * res1: List[String] = List()
+   *
+   * }}}
+   */
+  override def combineNK[A](a: F[A], n: Int): F[A] =
+    if (n < 0) throw new IllegalArgumentException("Repeated combining for monoids must have n >= 0")
```
johnynek, line 102: I think we should say `MonoidK` not `monoid` here.

johnynek (commented):
+1 to update the laws.

I think basically we should have laws that the implementations match the naive implementations.

BTW: I think you can override `combineAllK` for sequence types and get a significant win (list becomes O(N) vs O(N^2)) also Option can short circuit at the first non-empty item.

core/src/main/scala/cats/MonoidK.scala
TimWSpence: Sorry! Good spot!

johnynek:
note: https://github.com/typelevel/cats/pull/4084/files#diff-d189c08189989813f9b402a6993d2c49a1b7302601bc5de3dfe287eca63be6beR69

you can override the algebra there to override combineAll and similarly for SemigroupK.

Another idea, with this in hand, I think we can add:

```scala
def fromIterableOnce[A](as: IterableOnce[A]): F[A] =
  combineAllK(as.map(pure(_)))

def fromFoldable[G[_]: Foldable, A](as: G[A]): F[A] =
  fromIterableOnce(Foldable[F].toIterable(as))
```
to Alternative.

TimWSpence:
@johnynek just a heads up that I might not get to this till early in the new year but that it's at the top of my Todo list. Just in case you thought I was ignoring your feedback 😅

core/src/main/scala/cats/MonoidK.scala
johnynek: This hasn't been fixed.

core/src/main/scala/cats/MonoidK.scala
```diff
@@ -67,8 +84,54 @@ import simulacrum.typeclass
...
+   * res0: List[String] = List(One, Two, Three)
+   *
+   * scala> MonoidK[List].combineAllK[String](List.empty)
+   * res1: List[String] = List()
+   * }}}
+   */
+  def combineAllK[A](as: IterableOnce[F[A]]): F[A] =
+    as.iterator.foldLeft(empty[A])(combineK[A])
+
+  override def combineAllOptionK[A](as: IterableOnce[F[A]]): Option[F[A]] =
```
johnynek, line 121: I think it is better to do:
```scala
def combineAllK(as: IterableOnce[F[A]]): F[A] =
  combineAllOptionK(as) match {
    case Some(fa) => fa
    case None => empty[A]
  }
```

and not override `combineAllOptionK`.

The reason is, you want to only have to override one thing when you implement MonoidK and not have to override both and since combineAllOptionK is on both MonoidK and SemigroupK I think it is the easier one to remember to override.

I have sadly seen many cases of people picking the wrong one to override and the code paths not hitting the optimized method.

Alternatively you could do it this way, but I make this override final (that's allowed right)? so no one accidentally makes the mistake.

core/src/main/scala/cats/instances/list.scala
```diff
@@ -19,6 +20,14 @@ trait ListInstances extends cats.kernel.instances.ListInstances {
 
       def combineK[A](x: List[A], y: List[A]): List[A] = x ::: y
 
+      override def combineAllOptionK[A](as: IterableOnce[List[A]]): Option[List[A]] = {
+        val iter = as.iterator
+        if (iter.isEmpty) None else Some(iter.flatMap(_.iterator).toList)
+      }
+
+      override def combineAllK[A](as: IterableOnce[List[A]]): List[A] =
+        as.iterator.flatMap(_.iterator).toList
```
johnynek, line 29: this duplication is unfortunate IMO. I think in MonoidK we should be able to only override one.

In library code I think it is worth the verbosity for the possibility of slightly better performance:
```scala
val bldr = List.newBuilder[A]
val it = as.iterator
while(it.hasNext) {
  bldr ++= it.next()
}
bldr.result()
```

core/src/main/scala/cats/instances/option.scala
```diff
@@ -25,6 +26,12 @@ trait OptionInstances extends cats.kernel.instances.OptionInstances {
 
       def combineK[A](x: Option[A], y: Option[A]): Option[A] = if (x.isDefined) x else y
 
+      override def combineAllOptionK[A](as: IterableOnce[Option[A]]): Option[Option[A]] =
+        as.iterator.find(_.isDefined)
```
johnynek, line 30: I think we should avoid a closure in library code if we can:
```scala
val it = as.iterator
if (it.hasNext) {
  while (true) {
    val o = it.next()
    if (o.isDefined) return Some(o)
    if (!it.hasNext) return Some(None)
  }
  sys.error("unreachable")
}
else {
  return None
}
```

core/src/main/scala/cats/instances/queue.scala
```diff
@@ -17,6 +18,14 @@ trait QueueInstances extends cats.kernel.instances.QueueInstances {
 
       def combineK[A](x: Queue[A], y: Queue[A]): Queue[A] = x ++ y
 
+      override def combineAllOptionK[A](as: IterableOnce[Queue[A]]): Option[Queue[A]] = {
+        val iter = as.iterator
+        if (iter.isEmpty) None else Some(iter.flatMap(_.iterator).to(Queue))
```
johnynek, line 23: I think we could write a function to share like:
```scala
def appendAll[F <: Iterable[A], A](as: IterableOnce[F], bldr: Builder[F, A]): bldr.type = {
  val it = as.iterator
  while(it.hasNext) {
    bldr ++= it.next()
  }
  bldr
}
```
then we could use it for all the collection types with `appendAll(as, Queue.newBuilder[A]).result()`

core/src/main/scala/cats/instances/seq.scala
```diff
@@ -17,6 +19,14 @@ trait SeqInstances extends cats.kernel.instances.SeqInstances {
 
       def combineK[A](x: Seq[A], y: Seq[A]): Seq[A] = x ++ y
 
+      override def combineAllOptionK[A](as: IterableOnce[Seq[A]]): Option[Seq[A]] = {
+        val iter = as.iterator
+        if (iter.isEmpty) None else Some(iter.flatMap(_.iterator).toSeq)
+      }
+
+      override def combineAllK[A](as: IterableOnce[Seq[A]]): Seq[A] =
+        as.iterator.flatMap(_.iterator).toSeq
```
johnynek, line 28: same comments as above.

core/src/main/scala/cats/instances/vector.scala
```diff
@@ -18,6 +19,14 @@ trait VectorInstances extends cats.kernel.instances.VectorInstances {
 
       def combineK[A](x: Vector[A], y: Vector[A]): Vector[A] = x ++ y
 
+      override def combineAllOptionK[A](as: IterableOnce[Vector[A]]): Option[Vector[A]] = {
+        val iter = as.iterator
+        if (iter.isEmpty) None else Some(iter.flatMap(_.iterator).toVector)
+      }
+
+      override def combineAllK[A](as: IterableOnce[Vector[A]]): Vector[A] =
+        as.iterator.flatMap(_.iterator).toVector
```
johnynek, line 28: comments above.

TimWSpence:
Sorry @johnynek I was going to check this over before tagging you but you were too quick for me 😂 (hence the error message not being fixed)

core/src/main/scala/cats/instances/option.scala
TimWSpence: Just for my own edification, why do you prefer to avoid a closure? Just avoiding unnecessary allocations or something deeper?

core/src/main/scala/cats/MonoidK.scala
TimWSpence: Thanks 👍

TimWSpence:
@johnynek I _think_ I've addressed all your feedback (many thanks!) but do let me know if I've missed something. I also left a few questions for my own edification, hope that's ok!

johnynek (approved):
This is great! Thanks

armanbilge (commented):
Does `Alternative` also need a couple more laws to cover `fromIterableOnce` and `fromFoldable`? And also can't our instances offer a more optimized implementation of those? E.g. it should be possible to go from an `Iterable` to a collection without `pure` on every element.

johnynek:
Yeah. Good point. We could override using appendK/prependK and possibly skip the pure but it has to be per instance because it isn't obvious which is faster.

TimWSpence:
> Does `Alternative` also need a couple more laws to cover `fromIterableOnce` and `fromFoldable`? And also can't our instances offer a more optimized implementation of those? E.g. it should be possible to go from an `Iterable` to a collection without `pure` on every element.

Thanks! Yeah I thought as I wrote those methods that they should have laws and then didn't get round to it 😅

TimWSpence:
@armanbilge I've added a law for `fromIterableOnce` that asserts that the implementation matches its default behaviour. I wasn't sure what to do about `fromFoldable` though - if we added a similar law than it would require a binary incompatible addition of another type parameter on `AlternativeLaws` for the applicative

armanbilge:
Oh, that's annoying. I wonder if we should make that method `final` then. Since we can't test other implementations' correctness, seems like we should declare our `fromFoldable` as the One True Implementation.

TimWSpence:
> Oh, that's annoying. I wonder if we should make that method `final` then. Since we can't test other implementations' correctness, seems like we should declare our `fromFoldable` as the One True Implementation.

Yep, good call 👍 If an instance has override `fromIterableOnce` then it should probably be fine anyway

core/src/main/scala/cats/instances/queue.scala
```diff
@@ -17,6 +19,20 @@ trait QueueInstances extends cats.kernel.instances.QueueInstances {
...
       def combineK[A](x: Queue[A], y: Queue[A]): Queue[A] = x ++ y
 
+      override def combineAllOptionK[A](as: IterableOnce[Queue[A]]): Option[Queue[A]] = {
+        val iter = as.iterator
+        if (iter.isEmpty) None else Some(appendAll(iter, Queue.newBuilder[A]).result())
+      }
+
+      override def fromIterableOnce[A](as: IterableOnce[A]): Queue[A] = {
+        val iter = as.iterator
+        val builder = Queue.newBuilder[A]
```
johnynek, line 29: Won't `builder ++= as` work here?

core/src/main/scala/cats/Alternative.scala
```diff
@@ -99,8 +100,15 @@ import simulacrum.typeclass
       val F = self
       val G = Applicative[G]
     }
+
+  def fromIterableOnce[A](as: IterableOnce[A]): F[A] =
+    combineAllK(as.iterator.map(pure(_)))
+
+  final def fromFoldable[G[_]: Foldable, A](as: G[A]): F[A] =
```
johnynek, line 107: I may have suggested making this final, but on second thought I think it should not be.

It could be that F is very efficient to build from a foldRight or a foldLeft or a foldMap. Since we should assume that the foldable for G knows the most efficient way to do those, the `F` implementation here may want to call one of those rather than converting to an `Iterable[A]` and going from there.

This is probably minor, and maybe never a big deal, but the `final` will block this (I may have suggested or previously agreed with final, if so, apologies for the flip-flop).

core/src/main/scala/cats/instances/queue.scala
johnynek: I wonder if this is worth the micro optimization of `if (as.isEmpty) Queue.empty else { ... }` since in that case we don't allocate the builder?

core/src/main/scala/cats/instances/seq.scala
```diff
@@ -17,6 +20,17 @@ trait SeqInstances extends cats.kernel.instances.SeqInstances {
 
       def combineK[A](x: Seq[A], y: Seq[A]): Seq[A] = x ++ y
 
+      override def combineAllOptionK[A](as: IterableOnce[Seq[A]]): Option[Seq[A]] = {
+        val iter = as.iterator
+        if (iter.isEmpty) None else Some(appendAll(iter, Seq.newBuilder[A]).result())
+      }
+
+      override def fromIterableOnce[A](as: IterableOnce[A]): Seq[A] = {
```
johnynek, line 28: same pondering about `as.isEmpty`

core/src/main/scala/cats/instances/queue.scala
```diff
@@ -17,6 +19,17 @@ trait QueueInstances extends cats.kernel.instances.QueueInstances {
...
       def combineK[A](x: Queue[A], y: Queue[A]): Queue[A] = x ++ y
 
+      override def combineAllOptionK[A](as: IterableOnce[Queue[A]]): Option[Queue[A]] = {
+        val iter = as.iterator
+        if (iter.isEmpty) None else Some(appendAll(iter, Queue.newBuilder[A]).result())
+      }
+
+      override def fromIterableOnce[A](as: IterableOnce[A]): Queue[A] = {
+        val builder = Queue.newBuilder[A]
+        builder ++= as.iterator
```
johnynek, line 29: do we need `.iterator`? won't `builder ++= as` do?

core/src/main/scala/cats/instances/seq.scala
```diff
@@ -17,6 +20,17 @@ trait SeqInstances extends cats.kernel.instances.SeqInstances {
 
       def combineK[A](x: Seq[A], y: Seq[A]): Seq[A] = x ++ y
 
+      override def combineAllOptionK[A](as: IterableOnce[Seq[A]]): Option[Seq[A]] = {
+        val iter = as.iterator
+        if (iter.isEmpty) None else Some(appendAll(iter, Seq.newBuilder[A]).result())
```
johnynek, line 25: actually, in this pattern, won't `if (as.isEmpty) None else Some(appendAll(as.iterator, ...)` work also that way we don't need to convert to iterator in the empty case?

johnynek (dismissed):
added a few minor nits if you want to take them. If you don't want to take them we can possibly address them later.

core/src/main/scala/cats/instances/seq.scala
armanbilge: Is it possible to check `as.isEmpty` on an `IterableOnce` without consuming it?

In fact, in 2.13 `IterableOnce#isEmpty` is deprecated: "(Since version 2.13.0) Use `.iterator.isEmpty` instead".

I think what we could do here is use `IterableOnce#knownSize`.

core/src/main/scala/cats/Alternative.scala
armanbilge: Oh yeah, that's a good point.

FTR this was my suggestion in https://github.com/typelevel/cats/pull/4084#issuecomment-1024384440.

The problem is we can't currently check this implementation in the laws; see https://github.com/typelevel/cats/pull/4084#issuecomment-1024375090.

So there's a bit of performance vs yolo factor here 😉 

> This is probably minor, and maybe never a big deal, but the `final` will block this

We can always take away a `final` modifier, but we can't add it back. So I'd be fine to leave this until a practical need to change this comes up.

johnynek: why can't we check the laws? why can't we check that `fromFoldable(g) == fromIterable(g.toIterable)` in the laws?

core/src/main/scala/cats/instances/seq.scala
johnynek: okay. Bummer we have to allocate in order to check emptiness.

TimWSpence: Yeah that was what I wanted to write originally but ran into the deprecation

core/src/main/scala/cats/instances/seq.scala
TimWSpence: AFAICT `isEmpty` is [deprecated](https://www.scala-lang.org/api/2.13.7/scala/collection/IterableOnce.html#isEmpty:Boolean) in favour of `iterator.isEmpty`

TimWSpence: otherwise this would be a nice optimization

core/src/main/scala/cats/instances/queue.scala
TimWSpence: Thanks! Think I got lost in a see of different versions with different APIs. I've fixed this

core/src/main/scala/cats/Alternative.scala
TimWSpence: Yeah AFAICT we would have to add extra type parameters to `AlternativeTests#alternative` to do this. Unless I've missed something?

TimWSpence:
> added a few minor nits if you want to take them. If you don't want to take them we can possibly address them later.

Many thanks as always for the thorough review! I think I've addressed or replied to all your comments so let me know what you think or if I've missed anything!

core/src/main/scala/cats/instances/seq.scala
TimWSpence: or is allocating an `Iterator` to maybe not have to allocate a builder preferable?

core/src/main/scala/cats/data/Chain.scala
```diff
@@ -1029,6 +1046,7 @@ sealed abstract private[data] class ChainInstances extends ChainInstances1 {
 
       def empty[A]: Chain[A] = Chain.nil
       def combineK[A](c: Chain[A], c2: Chain[A]): Chain[A] = Chain.concat(c, c2)
+      override def fromIterableOnce[A](xs: IterableOnce[A]): Chain[A] = Chain.fromSeq(xs.iterator.toSeq)
```
johnynek, line 1049: nit...

god... I will never stop finding micro issues:

```scala
override def fromIterableOnce[A](xs: IterableOnce[A]): Chain[A]  =
  xs match {
    case s: Seq[A @unchecked] =>
      // Seq is a subclass of IterableOnce, so the type has to be compatible
     Chain.fromSeq(s) // pay O(1) not O(N) cost
   case notSeq =>
     Chain.fromSeq(notSeq.iterator.toSeq)
  }
```

johnynek (approved):
noticed one minor issue you can take or not (I think it is worth it in theory, but in practice maybe the method never gets called on planet earth).

core/src/main/scala/cats/data/Chain.scala
armanbilge: Could it make sense to add a `Chain.fromIterableOnce` that does this instead? I don't really know what to make of these `IterableOnce` possibly side-effecty things.

johnynek: sure, we can put it in Chain.

I don't view this as a side effect so much as we are lacking linear or affine types. What we want is to say this method takes ownership of `xs`. If we had that, then iterating it would be safe (assuming it did no side effect, which we make that assumption all the time in every function and method call).

johnynek: note: It would be really cool if scala added ownership types but I think it is a bit complex to get right and ergonomic.

TimWSpence:
> noticed one minor issue you can take or not (I think it is worth it in theory, but in practice maybe the method never gets called on planet earth).

Thanks! Yeah definitely worth making it `O(1)` for the more common case of `Seq` 👍

johnynek (approved):

PR #4090, opened 2021-12-20
cats and category theory in readme

SethTisue:
Throwing this out there — I'm not sure what the best structure or wording is, exactly.

There seems to be a persistent misconception that you're supposed to have some kind of basic grounding in category theory in order to learn or use Cats; maybe this, or something like it, would help dispel that a little.


tpolecat (approved):
:+1:

mpilquist (approved):

PR #4092, opened 2021-12-21
Add Order and PartialOrder For SortedMap

isomarcte:


johnynek:
I'm a bit nervous here.

1. `SortedMap[K, V]` has `.ordering` which returns `Ordering[K]` but we are taking the `Order[K]` in any case.
2. if we can't trust the `Ordering[K]` then actually we will be iterating in that order, but not correctly for the `Order[K]`.

So, a few solutions to my objection:

1. assume we can always use the left or right side ordering and not accept an implicit order in the caller
2. we could `require` that every `K` comparison compares the same using the `Ordering` from the right or left.

In any case, I don't see how the PartialOrder should take only a `PartialOrder[K]` since we have to have an `Order` to make the map.

kernel/src/main/scala/cats/kernel/instances/SortedMapInstances.scala
```diff
@@ -35,6 +35,43 @@ private[instances] trait SortedMapInstances2 extends SortedMapInstances1 {
...
+    if (x eq y) {
+      0
+    } else {
+      x.size.compare(y.size) match {
+        case 0 =>
+          StaticMethods.iteratorCompare(x.iterator, y.iterator)
+        case otherwise =>
+          otherwise
+      }
+    }
```
satorg, line 54: I would suppose that this approach may produce quite unexpected results. Check it out:
```
val sm1 = SortedMap(1 -> "", 3 -> "", 5 -> "")
val sm2 = SortedMap(2 -> "", 4 -> "")
```
So what is the expected result of the comparison here? According to lexicographical ordering it should be `sm1 < sm2`, but seems that your code will make it `sm1 > sm2` – is it correct? Moreover, adding one more item to the end of `sm2` will sharply switch the ordering result.

isomarcte: @satorg I removed this. This was an attempt at a micro optimization here. `.size` on `SortedMap` will, in almost all cases, be O(1) so it's a super fast way to short circuit. 

That said, I agree, it makes the ordering somewhat confusing.

I'm still no the fence myself. When I think of `Order` or `Ordering` I think about consistency of result and utility for comparison based data structures, e.g. trees. In any case, I've removed it.

isomarcte:
@johnynek 

> I'm a bit nervous here.
>
>    SortedMap[K, V] has .ordering which returns Ordering[K] but we are taking the Order[K] in any case.
>    if we can't trust the Ordering[K] then actually we will be iterating in that order, but not correctly for the Order[K].

I see what you're saying. The interaction between `Order` and `Ordering` with `SortedMap` and `SortedSet` is a bit strange. 

They way I wrote it is consistent with what we already do for `SortedSet` (https://github.com/typelevel/cats/blob/e0784fc0d95b9c4eed0bb1ddce6c69e3fd9a4a4f/kernel/src/main/scala/cats/kernel/instances/SortedSetInstances.scala#L16).

Would you prefer I change it to use the `Ordering` on the instance or stay consistent with what we do for `SortedSet`? The other option would be to rebuild the `SortedMap` values using our `Order` to ensure consistency, at a performance cost.

I'm torn because in the Scala stdlib `Ordering` has this note.

> This trait and scala.math.Ordered both provide this same functionality, but in different ways. A type T can be given a single way to order itself by extending Ordered. Using Ordering, this same type may be sorted in many other ways. Ordered and Ordering both provide implicits allowing them to be used interchangeably.

Which is the opposite of how we canonically think of `cats.Order` in terms of coherence. In other words, we can't necessarily rely on the `.ordering` to be the canonical instance of it for `K`. 

All that said, I'm fine with which ever strategy is preferred.

> In any case, I don't see how the PartialOrder should take only a PartialOrder[K] since we have to have an Order to make the map.

I was attempting to be consistent with this: https://github.com/typelevel/cats/blob/e0784fc0d95b9c4eed0bb1ddce6c69e3fd9a4a4f/kernel/src/main/scala/cats/kernel/instances/SortedMapInstances.scala#L10

We used to have a `Hash` instance for `SortedMap` which required an `Order` for the `K`, but _technically_ you didn't need that `Order` to write the `Hash` instance. You do need an `Ordering` from somewhere obviously to even created a `SortedMap`, so again it is this strange interaction between `Ordering` and `Order` and their idiomatic usage.

isomarcte:
@johnynek I noticed a few issues around SortedMap/SortedSet when looking into this deeper. I've opened this issue to discuss and we'll probably be in a holding pattern here until that is resolved.

https://github.com/typelevel/cats/issues/4103

isomarcte:
@johnynek I've removed the `PartialOrder` and `Order` constraints on the instances as you requested. Per the discussion on #4105, this PR assumes that the ordering between two instances of a `SortedMap[A, B]` is consistent.

Let me know if you want to proceed here or on #4105.

kernel/src/main/scala/cats/kernel/instances/SortedMapInstances.scala
```diff
@@ -35,6 +35,38 @@ private[instances] trait SortedMapInstances2 extends SortedMapInstances1 {
 
   implicit def catsKernelStdMonoidForSortedMap[K: Order, V: Semigroup]: Monoid[SortedMap[K, V]] =
     new SortedMapMonoid[K, V]
+
+  implicit def catsKernelStdPartialOrderForSortedMap[K: PartialOrder, V: PartialOrder]: PartialOrder[SortedMap[K, V]] =
```
johnynek, line 39: I think we don't need `K: PartialOrder` here

kernel/src/main/scala/cats/kernel/Eq.scala
```diff
@@ -242,6 +247,11 @@ private[kernel] trait PartialOrderInstances0 extends HashInstances {
     cats.kernel.instances.seq.catsKernelStdPartialOrderForSeq[A]
 }
 
+private[kernel] trait PartialOrderInstances1 extends PartialOrderInstances0 {
+  implicit def catsKernelPartialOrderForSortedMap[K: PartialOrder, V: PartialOrder]: PartialOrder[SortedMap[K, V]] =
```
johnynek, line 251: I don't think we need `K: PartialOrder` here.

kernel/src/main/scala/cats/kernel/Eq.scala
```diff
@@ -224,7 +224,12 @@ private[kernel] trait OrderInstances0 extends PartialOrderInstances {
     cats.kernel.instances.seq.catsKernelStdOrderForSeq[A]
 }
 
-private[kernel] trait PartialOrderInstances extends PartialOrderInstances0 {
+private[kernel] trait OrderInstances1 extends OrderInstances0 {
+  implicit def catsKernelOrderForSortedMap[K: Order, V: Order]: Order[SortedMap[K, V]] =
```
johnynek, line 228: I think we don't need `K: Order` here

johnynek (commented):
minor comment.

I think this is the way to go. Yes it is somewhat unsafe to have multiple typeclasses in play at the same time, but I think that risk already exists in many places and it is not something we can easily fix in a small PR.

It would require a real significant design change at least to structures that are built using typeclasses (Sets, Maps, Heaps come to mind).

kernel/src/main/scala/cats/kernel/Eq.scala
isomarcte: Fixed

johnynek (approved):
Thanks for all the effort on this.

isomarcte:
@satorg would you be able to take another look here?

satorg (approved):
Looks good to me, thanks!

PR #4093, opened 2021-12-21
Return stack-safe `Function0` and `Function1` from `Semigroup#combine`

### LINKED ISSUE #4089: `Semigroup[Function1[I, O]]` produces a function that's not stack safe
Following up on [this conversation in discord](https://discord.com/channels/632277896739946517/632278570512678923/922496169907716176). With a stack size of 512k (`-Xss512k`) I get a `StackOverflowError` with this code:

```scala
val f: Unit => Unit = _ => ()
val g = 1.to(5000).foldLeft(f)((acc, _) => Semigroup[Unit => Unit].combine(acc, f))
// StackOverflowError when I call `g`
g(())
```

I wasn't able to resolve the error using either `Eval` or `AndThen`, which were both suggested in discord, but it's entirely possible I was doing something wrong. Code snippets with my attempts in case they're helpful:

<details><summary>Eval attempt</summary>

```scala
val f: Unit => Unit = _ => ()
val g = 1.to(5000).foldLeft(Eval.now(f))((acc, _) => acc.map(Semigroup[Unit => Unit].combine(_, f)))
// StackOverflowError when I call `g`
g.value(())
```

</details>

<details><summary>AndThen attempt</summary>

```scala
// needed to define a `Semigroup[AndThen[I, O]]` as there's not one built in
implicit def andThenSemigroup[I, O](implicit O: Semigroup[O]): Semigroup[AndThen[I, O]] = new Semigroup[AndThen[I, O]] {
  def combine(x: AndThen[I, O], y: AndThen[I, O]): AndThen[I, O] =
    Apply[AndThen[I, *]].map2(x, y)(O.combine)
}

val f: AndThen[Unit, Unit] = AndThen(_ => ())
val g = 1.to(5000).foldLeft(f)((acc, _) => Semigroup[AndThen[Unit, Unit]].combine(acc, f))
// StackOverflowError when I call `g`
g(())
```

</details>

mrdziuban:
Fixes #4089.

This adds `CombineFunction0` and `CombineFunction1` classes as suggested by @johnynek in [this comment](https://github.com/typelevel/cats/issues/4089#issuecomment-998166820) to call and combine the results of two functions in a stack-safe way. Two things to note:

- This won't magically make any function passed to `Semigroup#combine` stack-safe, it just ensures that the combined function itself won't be the cause of a `StackOverflowError`
- I haven't added anything special to address the other idea of tracking call depth. If others feel this is important, I can look into doing so but may need some guidance
  > Lastly, you could track the depth of CombineFn when you are building it so you do the expensive apply when the depth gets greater than something like 500 or something, and do the naive thing when depth is < than that (something like AndThen's hack to use a constant amount of stack to do applications).

### DIFF
--- a/kernel/src/main/scala/cats/kernel/instances/FunctionInstances.scala
+++ b/kernel/src/main/scala/cats/kernel/instances/FunctionInstances.scala
@@ -1,8 +1,10 @@
 package cats.kernel
 package instances
 
+import cats.kernel.compat.scalaVersionSpecific._
 import scala.util.control.TailCalls.{done, tailcall, TailRec}
 
+@suppressUnusedImportWarningForScalaVersionSpecific
 trait FunctionInstances extends FunctionInstances0 {
 
   implicit def catsKernelOrderForFunction0[A](implicit ev: Order[A]): Order[() => A] =
@@ -128,6 +130,13 @@ trait Function1Semigroup[A, B] extends Semigroup[A => B] {
 
   override def combine(x: A => B, y: A => B): A => B =
     CombineFunction1(x, y, B)
+
+  override def combineAllOption(fns: IterableOnce[A => B]): Option[A => B] =
+    if (fns.iterator.isEmpty) None
+    else
+      Some { (a: A) =>
+        B.combineAllOption(fns.iterator.map(_.apply(a))).get
+      }
 }
 
 trait Function1Monoid[A, B] extends Function1Semigroup[A, B] with Monoid[A => B] {
@@ -164,6 +173,13 @@ trait Function0Semigroup[A] extends Semigroup[() => A] {
 
   override def combine(x: () => A, y: () => A): () => A =
     CombineFunction0(x, y, A)
+
+  override def combineAllOption(fns: IterableOnce[() => A]): Option[() => A] =
+    if (fns.iterator.isEmpty) None
+    else
+      Some { () =>
+        A.combineAllOption(fns.iterator.map(_.apply())).get
+      }
 }
 
 trait Function0Monoid[A] extends Function0Semigroup[A] with Monoid[() => A] {
--- a/kernel/src/main/scala/cats/kernel/instances/FunctionInstances.scala
+++ b/kernel/src/main/scala/cats/kernel/instances/FunctionInstances.scala
@@ -114,11 +114,11 @@ final private[instances] case class CombineFunction1[A, B](left: A => B, right: 
     extends (A => B) {
   private def call(fn: A => B, a: A): TailRec[B] =
     fn match {
-      case CombineFunction1(l, r, sg) =>
+      case ref: CombineFunction1[A, B] @unchecked =>
         for {
-          lb <- tailcall(call(l, a))
-          rb <- tailcall(call(r, a))
-        } yield sg.combine(lb, rb)
+          lb <- tailcall(call(ref.left, a))
+          rb <- tailcall(call(ref.right, a))
+        } yield ref.semiB.combine(lb, rb)
       case _ => done(fn(a))
     }
 
@@ -157,11 +157,11 @@ final private[instances] case class CombineFunction0[A](left: () => A, right: ()
     extends (() => A) {
   private def call(fn: () => A): TailRec[A] =
     fn match {
-      case CombineFunction0(l, r, sg) =>
+      case ref: CombineFunction0[A] @unchecked =>
         for {
-          la <- tailcall(call(l))
-          ra <- tailcall(call(r))
-        } yield sg.combine(la, ra)
+          la <- tailcall(call(ref.left))
+          ra <- tailcall(call(ref.right))
+        } yield ref.semiA.combine(la, ra)
       case _ => done(fn())
     }
 
--- a/kernel/src/main/scala/cats/kernel/instances/FunctionInstances.scala
+++ b/kernel/src/main/scala/cats/kernel/instances/FunctionInstances.scala
@@ -1,6 +1,8 @@
 package cats.kernel
 package instances
 
+import scala.util.control.TailCalls.{done, tailcall, TailRec}
+
 trait FunctionInstances extends FunctionInstances0 {
 
   implicit def catsKernelOrderForFunction0[A](implicit ev: Order[A]): Order[() => A] =
@@ -106,11 +108,26 @@ private[instances] trait FunctionInstances4 {
     new Function1Semigroup[A, B] { def B: Semigroup[B] = S }
 }
 
+final private[instances] case class CombineFunction1[A, B](left: A => B, right: A => B, semiB: Semigroup[B])
+    extends (A => B) {
+  private def call(fn: A => B, a: A): TailRec[B] =
+    fn match {
+      case CombineFunction1(l, r, sg) =>
+        for {
+          lb <- tailcall(call(l, a))
+          rb <- tailcall(call(r, a))
+        } yield sg.combine(lb, rb)
+      case _ => done(fn(a))
+    }
+
+  final override def apply(a: A): B = call(this, a).result
+}
+
 trait Function1Semigroup[A, B] extends Semigroup[A => B] {
   implicit def B: Semigroup[B]
 
   override def combine(x: A => B, y: A => B): A => B =
-    (a: A) => B.combine(x(a), y(a))
+    CombineFunction1(x, y, B)
 }
 
 trait Function1Monoid[A, B] extends Function1Semigroup[A, B] with Monoid[A => B] {
@@ -127,11 +144,26 @@ trait Function1Group[A, B] extends Function1Monoid[A, B] with Group[A => B] {
     (a: A) => B.inverse(x(a))
 }
 
+final private[instances] case class CombineFunction0[A](left: () => A, right: () => A, semiA: Semigroup[A])
+    extends (() => A) {
+  private def call(fn: () => A): TailRec[A] =
+    fn match {
+      case CombineFunction0(l, r, sg) =>
+        for {
+          la <- tailcall(call(l))
+          ra <- tailcall(call(r))
+        } yield sg.combine(la, ra)
+      case _ => done(fn())
+    }
+
+  final override def apply(): A = call(this).result
+}
+
 trait Function0Semigroup[A] extends Semigroup[() => A] {
   implicit def A: Semigroup[A]
 
   override def combine(x: () => A, y: () => A): () => A =
-    () => A.combine(x(), y())
+    CombineFunction0(x, y, A)
 }
 
 trait Function0Monoid[A] extends Function0Semigroup[A] with Monoid[() => A] {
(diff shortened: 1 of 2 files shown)


kernel/src/main/scala/cats/kernel/instances/FunctionInstances.scala
```diff
@@ -106,11 +108,26 @@ private[instances] trait FunctionInstances4 {
...
+  final override def apply(a: A): B = call(this, a).result
+}
+
 trait Function1Semigroup[A, B] extends Semigroup[A => B] {
   implicit def B: Semigroup[B]
 
   override def combine(x: A => B, y: A => B): A => B =
-    (a: A) => B.combine(x(a), y(a))
+    CombineFunction1(x, y, B)
 }
```
johnynek, line 131: can we override `combineAllOption` which would actually be faster I think:
```scala
override def combineAllOption(fns: IterableOnce[A => B]): Option[A => B] =
  if (fns.isEmpty) None
  else Some { a: A =>
    B.combineAllOption(fns.map(_.apply(a))).get
  }
```

kernel/src/main/scala/cats/kernel/instances/FunctionInstances.scala
```diff
@@ -127,11 +144,26 @@ trait Function1Group[A, B] extends Function1Monoid[A, B] with Group[A => B] {
...
+  final override def apply(): A = call(this).result
+}
+
 trait Function0Semigroup[A] extends Semigroup[() => A] {
   implicit def A: Semigroup[A]
 
   override def combine(x: () => A, y: () => A): () => A =
-    () => A.combine(x(), y())
+    CombineFunction0(x, y, A)
 }
```
johnynek, line 167: can we also add `combineAllOption` here?

johnynek (commented):
Thanks for taking this on!

kernel/src/main/scala/cats/kernel/instances/FunctionInstances.scala
```diff
@@ -106,11 +108,26 @@ private[instances] trait FunctionInstances4 {
...
 
+final private[instances] case class CombineFunction1[A, B](left: A => B, right: A => B, semiB: Semigroup[B])
+    extends (A => B) {
+  private def call(fn: A => B, a: A): TailRec[B] =
+    fn match {
+      case CombineFunction1(l, r, sg) =>
+        for {
+          lb <- tailcall(call(l, a))
+          rb <- tailcall(call(r, a))
+        } yield sg.combine(lb, rb)
```
johnynek, line 119: looks like scala 3 doesn't like this line. I think we could possibly fix it by doing:
```scala
private def call[C](fn: A => C, a: A): TailRec[C] =
```
and not tie the result type to `B`. Since `Semigroup` is invariant in `B` but `A => B` is covariant in B, actually the pattern match could give a `B1 <: B`. I think `TailRec` is covariant, so this should be fine... maybe scala 3 has a bug here, or I'm missing something...

mrdziuban: That fixes it in scala 3 but breaks it in scala 2 😢

```
[error] FunctionInstances.scala:119:28: type mismatch;
[error]  found   : lc.type (with underlying type C)
[error]  required: ?C1 where type ?C1 <: C (this is a GADT skolem)
[error]         } yield sg.combine(lc, rc)
```

What I found does work is using a type ascription in the match along with an `@unchecked` annotation, similar to [what `AndThen` does here](https://github.com/typelevel/cats/blob/main/core/src/main/scala/cats/data/AndThen.scala#L143).

I went ahead a committed that but would be happy to revert it if there's a better solution.

kernel/src/main/scala/cats/kernel/instances/FunctionInstances.scala
mrdziuban: Addressed

kernel/src/main/scala/cats/kernel/instances/FunctionInstances.scala
mrdziuban: Addressed

johnynek (dismissed):

johnynek (approved):

joroKr21:
What if we had simply:
```scala
class CombineFunctions[A, B](fns: NonEmptyChain[A => B], sg: Semigroup[B]) extends (A => B) {
  def apply(a: A) = fns.reduceMap(_.apply(a))
}
```

johnynek:
1. There is no non-empty list in kernel
2. You would have O(N^2) performance on repeated combine.

This with NonEmptyChain would work if that was accessible in kernel, although I'm not sure NonEmptyChain's reduceMap would be any faster than what we have here. I think we would want benchmarks.

joroKr21:
> This with NonEmptyChain would work if that was accessible in kernel, although I'm not sure NonEmptyChain's reduceMap would be any faster than what we have here. I think we would want benchmarks.

Hmm we could use `Vector` then or `LazyList/Stream`. The main point is building a collection of functions on `combine` (by checking both sides if they are already `CombineFunctions`) and then the evaluation is `reduce` - no need to trampoline.

> Also, passing the Semigroup as an implicit means it would not be used for equality which I think is a negative personally (since I've often leveraged AST equality in caching of interpreters).

Doesn't the current implementation suffer from the same problem? - i.e. if we have `CombineFunciton` on both sides we should also check that the `Semigroup` is the same.

johnynek:
repeat concatenation of Vector can result in O(N^2) performance. LazyList/Stream concatenations can result in stack overflow.

> Doesn't the current implementation suffer from the same problem? - i.e. if we have CombineFunciton on both sides we should also check that the Semigroup is the same.

Note the current implementation only uses that semigroup for the outer most pair to combine. If an inner pair is also a CombineFunction then we use the semigroup associated with that. We should add a test for that, but if we don't have that we will definitely introduce some confusing bugs. So, I think the current implementation does correctly implement equals structurally.

Also, now that I think about it, I don't see how you can use the collection approach without storing the semigroups since in principle, they can all be different.

I'm not that attached to this problem. If you don't care about the O(N^2), I guess that's fine, since I don't think I've ever used this monoid, but consider the problem: if you are overflowing the stack, you have a long list, if you get N^2 performance there it is going to be slow.

If you really want it written another way, I'd recommend 1. making a benchmark, 2. making another PR. @mrdziuban has solved the issue, and spent the time to PR.

joroKr21:
I'm not very attached to this problem either - just had a different idea.

> repeat concatenation of Vector can result in O(N^2) performance. 

That doesn't sound right. Why do you think it would be O(N^2)?

> LazyList/Stream concatenations can result in stack overflow.

Yeah, I guess if you append a lot you get the same problem.

> Note the current implementation only uses that semigroup for the outer most pair to combine. If an inner pair is also a CombineFunction then we use the semigroup associated with that. We should add a test for that, but if we don't have that we will definitely introduce some confusing bugs. So, I think the current implementation does correctly implement equals structurally.

I'm not sure what you're saying but after looking a bit closer I managed to convince myself that the it would work correctly. On the other hand we don't use the structural equality of case classes here.

> Also, now that I think about it, I don't see how you can use the collection approach without storing the semigroups since in principle, they can all be different.

Yes, you would store the semigroup either way.

> If you really want it written another way, I'd recommend 1. making a benchmark, 2. making another PR. @mrdziuban has solved the issue, and spent the time to PR.

Yes, of course, kudos to @mrdziuban for tackling this tricky problem! I just wanted to poke a bit more at the problem to see if we can find a solution which doesn't have to pay the cost of trampolining every time we invoke the resulting function but it doesn't seem to be the case.

kernel/src/main/scala/cats/kernel/instances/FunctionInstances.scala
joroKr21: This seems to work for me:
```scala
final case class CombineFunction1[-A, B](left: A => B, right: A => B, semiB: Semigroup[B]) extends (A => B) {
    private[this] def call[C](fn: A => C, a: A): TailRec[C] =
      fn match {
        case ref: CombineFunction1[A, _] =>
          for {
            lb <- tailcall(call(ref.left, a))
            rb <- tailcall(call(ref.right, a))
          } yield ref.semiB.combine(lb, rb)
        case _ => done(fn(a))
      }

    override def apply(a: A): B = call(this, a).result
  }
```

joroKr21:
Another approach I can think of would be to write a `balance` function that converts a chain of functions to a tree of functions `(((((f |+| g) |+| h) |+| i) |+| j) => (((f |+| g) |+| (h |+| i)) |+| j)` so that we need O(log2(N)) stack instead of O(N) stack but that may prove too complicated for this use case.

johnynek:
> I'm not very attached to this problem either - just had a different idea.
> 
> > repeat concatenation of Vector can result in O(N^2) performance.
> 
> That doesn't sound right. Why do you think it would be O(N^2)?

Why do you think it doesn't? 😀 I thought, e.g. prepending an item to a vector of length N did work order N, not O(1). So, if that's true, you have the same worst case as any repeated sequence operation (which motivated Chain in the first place). building a List in the wrong order (say appending on the right) of length N is O(N^2). Similarly building a string by repeatedly adding to head or tail. I think Vector has this same problem if you append to the head (the reverse order as List).
 
> Yes, of course, kudos to @mrdziuban for tackling this tricky problem! I just wanted to poke a bit more at the problem to see if we can find a solution which doesn't have to pay the cost of trampolining every time we invoke the resulting function but it doesn't seem to be the case.

Your suggestion of using associativity to maintain a tree would be a solution IMO. I think log depth is fine (there is no way we are going to blow the stack in a runtime we can wait for, if log N = 5000).

I think that is trickier and can actually be implemented later without breaking binary compatibility since the trampoline is private in the current implementation.

What about merging this, and then someone can make a follow up with the tree approach?

joroKr21:
`Vector` has "effectively constant" append and prepend. Think of it as a tree with branching factor 32, so more precisely it's O(log32(N)) amortised complexity. But it's not great in terms of memory overhead for a small number of elements (less than 32) for that same reason.

> What about merging this, and then someone can make a follow up with the tree approach?

Sounds good to me 👍

armanbilge (approved):

johnynek:
> `Vector` has "effectively constant" append and prepend. Think of it as a tree with branching factor 32, so more precisely it's O(log32(N)) amortised complexity. But it's not great in terms of memory overhead for a small number of elements (less than 32) for that same reason. It also has good concatenation performance with other vectors.

I know how vector works, but I thought that at least the version prior to 2.13 didn't have efficient concat. Looking this code, I don't see it: https://github.com/scala/scala/blob/2.13.x/src/library/scala/collection/immutable/Vector.scala

I don't see a specialized path that offers O(log N) vector contact. It looks to me like you have to append in a loop the smaller of the two items. Am I missing it?

joroKr21:
It's here: https://github.com/scala/scala/blob/2.13.x/src/library/scala/collection/immutable/Vector.scala#L1546-L1558
The loop kicks in when one of the vectors is significantly smaller.

PR #4101, opened 2022-01-02
Update scalafmt-core to 3.3.0

satorg:
After #4097 has been closed, Scala Steward is not going to resume updating `scalafmt-core` until we restart it manually.
Therefore this PR restarts these updates.

Note: there's `scalafmt-core` v3.3.1 available already and I tried to jump to that version directly, but seems that the most recent version introduces even more weird formatting changes which need to be discussed separately.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,4 +1,4 @@
-version=3.2.1
+version=3.3.0
 align.openParenCallSite = true
 align.openParenDefnSite = true
 maxColumn = 120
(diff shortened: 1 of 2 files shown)


armanbilge:
Yes, the 3.3.1 update in https://github.com/typelevel/spire/pull/1135 seems a bit strange.

satorg:
@armanbilge when applied to Cats sources v3.3.1 is even more strange – it changes formatting of boilerplate code and I doubt that we like it. But v3.3.0 looks ok to me.

rossabaker (approved):
This one seems small enough.

Issue #4103, opened 2022-01-03
SortedSet, SortedMap, NonEmptySet, and NonEmptyMap Unexpected Behaviors


isomarcte:
During the implementation of https://github.com/typelevel/cats/pull/4092 I noticed several unexpected properties of the SortedSet, SortedMap, NonEmptySet, and NonEmptyMap and the related code.

I'm not exactly sure what, if anything, should be done so I wanted to open up this issue to talk through the various topics.

# Background #

## Related Typeclass Instances ##

The topics relate to the following currently implemented typeclass instances. `NonEmptySet` and `NonEmptyMap` are mostly involved here just because they are implemented in terms of `SortedSet` and `SortedMap`, so I'll primarily focus on the `SortedSet` and `SortedMap`.

* `Order[SortedSet[A]]`
* `Eq[SortedMap[K, V]]`
* `CommutativeMonoid[SortedMap[K, V]]`

Of note, cats does not have an implementation currently of `Order[SortedMap[K, V]]`. One is proposed here, https://github.com/typelevel/cats/pull/4092.

## Interaction With Order/Ordering ##

All of these topics relate in some way or another to the fact that it is generally assumed that `cats.kernel.Order` will be coherent (only one instance exists for a given type `A`), `scala.math.Ordering` is not necessarily expected to be coherent. In fact, the Scaladocs for it suggest that one would use `Ordering` idiomatically for non-coherent alternative orderings, using `scala.math.Ordered` subtypes for the canonical ordering of a given type.

# Monoid for SortedSet Not Lawful #

This is the big issue. Specifically, if the given `SortedSet` is not using the same `Ordering` as `Monoid[SortedSet[A]].empty`, then the identity property doesn't hold.

```scala
scala> val reverseOrdering: Ordering[Int] = Ordering[Int].reverse
val reverseOrdering: Ordering[Int] = $anonfun$1@2c6a4b4d

scala> val x = SortedSet(1, 2)(reverseOrdering)
val x: scala.collection.immutable.SortedSet[Int] = TreeSet(2, 1)

scala> Eq[SortedSet[Int]].eqv(Monoid.empty[SortedSet[Int]] |+| x, x)
val res0: Boolean = false
```

This is because the `Eq` instance for `SortedSet` relies on iteration order. When we combine `Monoid.empty` with a `SortedSet` which is using a valid, but different, order then the result will not compare equal to the original value.

# Surprising `Eq` Behavior #

[...]

PR #4104, opened 2022-01-03
Make Monoid SortedSet Lawful

isomarcte:
This makes `SortedSet[A]` lawful in terms of the identity law for Monoids. Previously it was not so if the two `SortedSet[A]` values were not using a logically equivalent `Ordering` instance, e.g. `Ordering[A].reverse`.

See: https://github.com/typelevel/cats/issues/4103

### DIFF
--- a/kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
+++ b/kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
@@ -66,15 +66,17 @@ object KernelCheck {
   }
 
   // Copied from cats-laws.
-  implicit def arbitrarySortedSet[A: Arbitrary: Order]: Arbitrary[SortedSet[A]] = {
+  implicit def arbitrarySortedSet[A: Arbitrary: Order: Cogen]: Arbitrary[SortedSet[A]] = {
     // We create an arbitrary Ordering[A] which is either the same as Order[A]
     // or the reverse of it. This is important because the Ordering in use by
     // a given SortedSet[A] may not be derived from our Order and
     // scala.math.Ordering does not imply coherence.
     val orderingGen: Gen[Ordering[A]] =
-      Gen.oneOf(
-        Order[A].toOrdering,
-        Order[A].toOrdering.reverse
+      Arbitrary.arbitrary[A => Int].map(order =>
+        new Ordering[A] {
+          override def compare(x: A, y: A): Int =
+            order(x).compare(order(y))
+        }
       )
     Arbitrary(
       orderingGen.flatMap(ordering => arbitrary[Set[A]].map(s => SortedSet.empty[A](ordering) ++ s))
--- a/kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
+++ b/kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
@@ -66,14 +66,17 @@ object KernelCheck {
   }
 
   // Copied from cats-laws.
-  implicit def arbitrarySortedSet[A: Arbitrary: Order]: Arbitrary[SortedSet[A]] = Arbitrary {
+  implicit def arbitrarySortedSet[A: Arbitrary: Order: Cogen]: Arbitrary[SortedSet[A]] = Arbitrary {
     // We create an arbitrary Ordering[A] which is either the same as Order[A]
     // or the reverse of it. This is important because the Ordering in use by
     // a given SortedSet[A] may not be derived from our Order and
     // scala.math.Ordering does not imply coherence.
     val ordering: Ordering[A] = Order[A].toOrdering
     val orderingGen: Gen[Ordering[A]] =
-      Gen.oneOf(ordering, ordering.reverse)
+      Gen.oneOf(Gen.const(ordering),
+                Gen.const(ordering.reverse),
+                Arbitrary.arbitrary[A => Long].map(f => Order.by(f).toOrdering)
+      )
     for {
       ordering <- orderingGen
       sortedSet <- arbitrary[Set[A]].map(s => SortedSet.empty[A](ordering) ++ s)
--- a/kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
+++ b/kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
@@ -66,21 +66,18 @@ object KernelCheck {
   }
 
   // Copied from cats-laws.
-  implicit def arbitrarySortedSet[A: Arbitrary: Order: Cogen]: Arbitrary[SortedSet[A]] = {
+  implicit def arbitrarySortedSet[A: Arbitrary: Order]: Arbitrary[SortedSet[A]] = Arbitrary{
     // We create an arbitrary Ordering[A] which is either the same as Order[A]
     // or the reverse of it. This is important because the Ordering in use by
     // a given SortedSet[A] may not be derived from our Order and
     // scala.math.Ordering does not imply coherence.
+    val ordering: Ordering[A] = Order[A].toOrdering
     val orderingGen: Gen[Ordering[A]] =
-      Arbitrary.arbitrary[A => Int].map(order =>
-        new Ordering[A] {
-          override def compare(x: A, y: A): Int =
-            order(x).compare(order(y))
-        }
-      )
-    Arbitrary(
-      orderingGen.flatMap(ordering => arbitrary[Set[A]].map(s => SortedSet.empty[A](ordering) ++ s))
-    )
+      Gen.oneOf(ordering, ordering.reverse)
+    for {
+      ordering <- orderingGen
+      sortedSet <- arbitrary[Set[A]].map(s => SortedSet.empty[A](ordering) ++ s)
+    } yield sortedSet
   }
 
   // Copied from cats-laws.
--- a/kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
+++ b/kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
@@ -66,7 +66,7 @@ object KernelCheck {
   }
 
   // Copied from cats-laws.
-  implicit def arbitrarySortedSet[A: Arbitrary: Order]: Arbitrary[SortedSet[A]] = Arbitrary{
+  implicit def arbitrarySortedSet[A: Arbitrary: Order]: Arbitrary[SortedSet[A]] = Arbitrary {
     // We create an arbitrary Ordering[A] which is either the same as Order[A]
     // or the reverse of it. This is important because the Ordering in use by
     // a given SortedSet[A] may not be derived from our Order and
--- a/kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
+++ b/kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
@@ -72,15 +72,10 @@ object KernelCheck {
     // a given SortedSet[A] may not be derived from our Order and
     // scala.math.Ordering does not imply coherence.
     val orderingGen: Gen[Ordering[A]] =
-      Arbitrary
-        .arbitrary[Boolean]
-        .map(invert =>
-          if (invert) {
-            Order[A].toOrdering.reverse
-          } else {
-            Order[A].toOrdering
-          }
-        )
+      Gen.oneOf(
+        Order[A].toOrdering,
+        Order[A].toOrdering.reverse
+      )
     Arbitrary(
       orderingGen.flatMap(ordering => arbitrary[Set[A]].map(s => SortedSet.empty[A](ordering) ++ s))
     )
--- a/kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
+++ b/kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
@@ -66,17 +66,14 @@ object KernelCheck {
   }
 
   // Copied from cats-laws.
-  implicit def arbitrarySortedSet[A: Arbitrary: Order: Cogen]: Arbitrary[SortedSet[A]] = Arbitrary {
+  implicit def arbitrarySortedSet[A: Arbitrary: Order]: Arbitrary[SortedSet[A]] = Arbitrary {
     // We create an arbitrary Ordering[A] which is either the same as Order[A]
     // or the reverse of it. This is important because the Ordering in use by
     // a given SortedSet[A] may not be derived from our Order and
     // scala.math.Ordering does not imply coherence.
     val ordering: Ordering[A] = Order[A].toOrdering
     val orderingGen: Gen[Ordering[A]] =
-      Gen.oneOf(Gen.const(ordering),
-                Gen.const(ordering.reverse),
-                Arbitrary.arbitrary[A => Long].map(f => Order.by(f).toOrdering)
-      )
+      Gen.oneOf(ordering, ordering.reverse)
     for {
       ordering <- orderingGen
       sortedSet <- arbitrary[Set[A]].map(s => SortedSet.empty[A](ordering) ++ s)
--- a/kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
+++ b/kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
@@ -66,8 +66,25 @@ object KernelCheck {
   }
 
   // Copied from cats-laws.
-  implicit def arbitrarySortedSet[A: Arbitrary: Order]: Arbitrary[SortedSet[A]] =
-    Arbitrary(arbitrary[Set[A]].map(s => SortedSet.empty[A](implicitly[Order[A]].toOrdering) ++ s))
+  implicit def arbitrarySortedSet[A: Arbitrary: Order]: Arbitrary[SortedSet[A]] = {
+    // We create an arbitrary Ordering[A] which is either the same as Order[A]
+    // or the reverse of it. This is important because the Ordering in use by
+    // a given SortedSet[A] may not be derived from our Order and
+    // scala.math.Ordering does not imply coherence.
+    val orderingGen: Gen[Ordering[A]] =
+      Arbitrary
+        .arbitrary[Boolean]
+        .map(invert =>
+          if (invert) {
+            Order[A].toOrdering.reverse
+          } else {
+            Order[A].toOrdering
+          }
+        )
+    Arbitrary(
+      orderingGen.flatMap(ordering => arbitrary[Set[A]].map(s => SortedSet.empty[A](ordering) ++ s))
+    )
+  }
 
   // Copied from cats-laws.
   implicit def cogenSortedSet[A: Order: Cogen]: Cogen[SortedSet[A]] = {
--- a/kernel/src/main/scala/cats/kernel/Order.scala
+++ b/kernel/src/main/scala/cats/kernel/Order.scala
@@ -94,9 +94,7 @@ trait Order[@sp A] extends Any with PartialOrder[A] { self =>
    * instance.
    */
   def toOrdering: Ordering[A] =
-    new Ordering[A] {
-      def compare(x: A, y: A): Int = self.compare(x, y)
-    }
+    OrderingFromOrder[A](this)
 }
 
 abstract class OrderFunctions[O[T] <: Order[T]] extends PartialOrderFunctions[O] {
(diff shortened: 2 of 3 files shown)


kernel/src/main/scala/cats/kernel/instances/SortedSetInstances.scala
```diff
@@ -23,12 +23,20 @@ private[instances] trait SortedSetInstances1 {
 class SortedSetOrder[A: Order] extends Order[SortedSet[A]] {
   def compare(a1: SortedSet[A], a2: SortedSet[A]): Int =
     cats.kernel.instances.int.catsKernelStdOrderForInt.compare(a1.size, a2.size) match {
-      case 0 => StaticMethods.iteratorCompare(a1.iterator, a2.iterator)
+      case 0 =>
+        // In the event that the ordering instances are not _exactly_ the
+        // same, we have to sort/rebuild the sets. If we don't we risk
+        // violating the Monoid identity law.
+        if (a1.ordering eq a2.ordering) {
```
joroKr21, line 30: `eq` is too restrictive - `Ordering` defines structural equality for the builtin cases
```suggestion
        if (a1.ordering == a2.ordering) {
```

isomarcte: :+1: I'll update that.

kernel/src/main/scala/cats/kernel/instances/SortedSetInstances.scala
```diff
@@ -23,12 +23,24 @@ private[instances] trait SortedSetInstances1 {
...
+          StaticMethods.iteratorCompare(a1.toSeq.sorted(ordering).iterator, a2.toSeq.sorted(ordering).iterator)
+        }
       case x => x
     }
 
-  override def eqv(s1: SortedSet[A], s2: SortedSet[A]): Boolean =
-    StaticMethods.iteratorEq(s1.iterator, s2.iterator)
+  // Could be removed, but MiMa complains
+  override def eqv(a1: SortedSet[A], a2: SortedSet[A]): Boolean =
+    compare(a1, a2) == 0
```
armanbilge, line 43: This is equivalent to calling `super` right?

kernel/src/main/scala/cats/kernel/instances/SortedSetInstances.scala
joroKr21: Scala 2.13 also defines `Ordering.isReverseOf` so if we're willing to use version-specific sources we could add another special case to reverse one of the sequences (or e.g. do `toIndexedSeq.reverseIterator`) instead of sorting both 😄

kernel/src/main/scala/cats/kernel/instances/SortedSetInstances.scala
isomarcte: lol, yes. I'll fix that.

kernel/src/main/scala/cats/kernel/instances/SortedSetInstances.scala
durban: Wait, this compares `Ordering[A]` _instances_... I think `eq` makes perfect sense.

joroKr21: `eq` works only for primitive types but not for `List`, `Option`, tuple, etc. - what's wrong with using `==`?

kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
```diff
@@ -66,8 +66,20 @@ object KernelCheck {
...
+  implicit def arbitrarySortedSet[A: Arbitrary: Order]: Arbitrary[SortedSet[A]] = {
+    // We create an arbitrary Ordering[A] which is either the same as Order[A]
+    // or the reverse of it. This is important because the Ordering in use by
+    // a given SortedSet[A] may not be derived from our Order and
+    // scala.math.Ordering does not imply coherence.
+    val orderingGen: Gen[Ordering[A]] =
+      Gen.oneOf(
+        Order[A].toOrdering,
+        Order[A].toOrdering.reverse
+      )
```
armanbilge, line 78: These seem reasonable 👍 to get a truly arbitrary ordering, I suppose we could use an arbitrary `A => Int` or similar?

joroKr21: Not every `A => Int` is a valid ordering though 😄

kernel/src/main/scala/cats/kernel/instances/SortedSetInstances.scala
isomarcte: @durban `==` may still yield false for instances which are logically equivalent, e.g. `new Ordering[A]{ override def compare(a: A, b: B): Int = Ordering[A].compare(a, b)}`, but that's okay. The danger would be if it yielded `true` for logically equivalent instances.

As @joroKr21 points out, for _some_ cases we can use structural equality from the stdlib to avoid the expensive branch, e.g. https://github.com/scala/scala/blob/v2.13.6/src/library/scala/math/Ordering.scala#L259

kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
armanbilge: Oh? 😳 how is that?

isomarcte: Yeah, the only two globally valid orderings I think we can make arbitrarily are the canonical one and the reverse of it.

armanbilge: > you can use if if you add Cogen[A]

Yes, that's what I was thinking :)

isomarcte: Oh...I guess it is....

kernel/src/main/scala/cats/kernel/instances/SortedSetInstances.scala
durban: There is nothing wrong with `==`; I just wasn't aware that there are `Ordering` instances which use non-reference equality. Apparently they exist, so `==` makes even more sense than `eq` :-)

(Btw. it is possible, for `==` to return true for logically not equivalent `Ordering` instances, but that is a bug in the `Ordering` instance.)

kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
isomarcte: We have to add a Cogen constraint, but that seems fine to me.

kernel/src/main/scala/cats/kernel/instances/SortedSetInstances.scala
joroKr21: Basically the standard library had the same problem and came up with this solution.
Ironically `Ordering` instances created from `Order` are neither `==` nor `eq` 😞

kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
isomarcte: Updated

joroKr21: But now we're not testing the fast path. Which is tricky because
> Ironically Ordering instances created from Order are neither == nor eq 😞

I think you can achieve it if you cache the `Ordering` from `Order` and use it in `Gen.oneOf` like:
```scala
val canonical = Order[A].toOrdering
Gen.oneOf(canonical, canonical.reverse, Arbitrary.arbitrary[A => Int].map(Ordering.by))
```

joroKr21: Probably missed some `Gen.const` there but you get the idea

kernel/src/main/scala/cats/kernel/instances/SortedSetInstances.scala
johnynek: We could fix our `Ordering` from `Order` somewhat, but yeah, cats has not tried to make equality work for typeclass values.

BTW: having written several compilers and compiler-like things in scala, making equivalent things equal can be very useful for term rewriting that often happens there (i.e. taking some expressions written in scala and seeing if they are equal to others).

kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
```diff
@@ -66,8 +66,22 @@ object KernelCheck {
   }
 
   // Copied from cats-laws.
-  implicit def arbitrarySortedSet[A: Arbitrary: Order]: Arbitrary[SortedSet[A]] =
-    Arbitrary(arbitrary[Set[A]].map(s => SortedSet.empty[A](implicitly[Order[A]].toOrdering) ++ s))
+  implicit def arbitrarySortedSet[A: Arbitrary: Order: Cogen]: Arbitrary[SortedSet[A]] = {
```
johnynek, line 69: you aren't using `Order[A]` any more are you?

kernel/src/main/scala/cats/kernel/instances/SortedSetInstances.scala
```diff
@@ -23,12 +23,24 @@ private[instances] trait SortedSetInstances1 {
 class SortedSetOrder[A: Order] extends Order[SortedSet[A]] {
   def compare(a1: SortedSet[A], a2: SortedSet[A]): Int =
     cats.kernel.instances.int.catsKernelStdOrderForInt.compare(a1.size, a2.size) match {
-      case 0 => StaticMethods.iteratorCompare(a1.iterator, a2.iterator)
+      case 0 =>
+        // In the event that the ordering instances are not _exactly_ the
+        // same, we have to sort/rebuild the sets. If we don't we risk
+        // violating the Monoid identity law.
+        if (a1.ordering == a2.ordering) {
```
johnynek, line 30: don't you need to compare to `Order[A]` as well? they could be the same, but not the same as `Order[A]` right?

kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
```diff
@@ -66,8 +66,22 @@ object KernelCheck {
...
+    // scala.math.Ordering does not imply coherence.
+    val orderingGen: Gen[Ordering[A]] =
+      Arbitrary.arbitrary[A => Int].map(order =>
+        new Ordering[A] {
+          override def compare(x: A, y: A): Int =
+            order(x).compare(order(y))
+        }
+      )
+    Arbitrary(
+      orderingGen.flatMap(ordering => arbitrary[Set[A]].map(s => SortedSet.empty[A](ordering) ++ s))
```
johnynek, line 82: nit: I think `for` would be prettier here:
```scala
Arbitrary(for {
  ordering <- orderingGen
  setA <- arbitrary[Set[A]]
} yield SortedSet.empty(ordering) ++ setA)
```

kernel/src/main/scala/cats/kernel/instances/SortedSetInstances.scala
isomarcte: I don't _think_ so we _need_ to check that, but I could be wrong. This gets really tricky here, but I _think_ the laws all hold as long as the ordering used in both sets is consistent with each other, even if they are inconsistent with `Order` used to actually compare the elements. We only need _consistent_ iteration order for traversing the sets here, using our `Order` for comparing the elements.

We can't meaningfully compare `Order[A]` to the `Ordering[A]` values even if we wanted to, at least not without some additional machinery. 

FWIW, I don't really like this. As I discussed in #4103, I think we should consider adding a newtype which always and only uses the coherent `Order` for the underlying `SortedSet`, but I'm leaving that for a different PR.

kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
isomarcte: I'll update that

kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
isomarcte: @armanbilge I don't think we can use `Arbitrary[A => Int]` after all. Unless I'm missing something, the arbitrary functions generated by Scalacheck do not guarantee that each `A` maps to a `Int` value that is not shared by any other `A`, e.g. we can get a function like this.

```scala
scala> val f = Arbitrary.arbitrary[Short => Int].sample.get
val f: Short => Int = org.scalacheck.GenArities$$Lambda$9321/0x00000008017d9e58@58794be4

scala> f(-32763)
val res7: Int = 1

scala> f(-32764)
val res8: Int = 1
```

Which means they would compare equal by `Ordering`, but not by `.equals`, which breaks the `SortedSet`. I think we need a bijective function here, but we can't get that from Scalacheck unless I'm mistaken.

Am I missing something here?

kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
isomarcte: It looks like I will need it per @joroKr21's comment above to hit the fast path branch in testing.

https://github.com/typelevel/cats/pull/4104#discussion_r777611387

kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
armanbilge: Huh. If the `Ordering` and `.equals` are inconsistent, is that not also a source of unlawfulness? Since the notion of equivalence would affect the size of the set.

kernel/src/main/scala/cats/kernel/instances/SortedSetInstances.scala
isomarcte: @johnynek I'm wrong. Unless all three order/orderings agree we can break transitivity. 

https://github.com/typelevel/cats/runs/4694302475?check_suite_focus=true#step:15:598

I'll take a look here.

kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
joroKr21: Wait, the `SortedSet` should use `Ordering.compare`, not `equals`, right?

kernel/src/main/scala/cats/kernel/instances/SortedSetInstances.scala
joroKr21: Well if we only need the same iteration order we could just pick the ordering of the larger set and resort only the other one.

isomarcte:
@johnynek @armanbilge @joroKr21 My initial solution broken Order transitivity laws on the fast branch. Please take a look at https://github.com/typelevel/cats/pull/4104/commits/3f2c9277e11ef560492169e9747cee74c68438ca. I was hoping to avoid resorting to this, but it's the only way I can see that we can adhere to all the laws and still have a hope of not _always_ having to resort the sets.

armanbilge:
@isomarcte I'm also a bit confused how this interacts with equality, see https://github.com/typelevel/cats/pull/4104#discussion_r777643959. I think we might need to check the orderings for consistency before doing _any_ comparison.

kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
```diff
@@ -66,8 +66,22 @@ object KernelCheck {
...
+    // scala.math.Ordering does not imply coherence.
+    val ordering: Ordering[A] = Order[A].toOrdering
+    val orderingGen: Gen[Ordering[A]] =
+      Gen.oneOf(Gen.const(ordering),
+                Gen.const(ordering.reverse),
+                Arbitrary.arbitrary[A => Long].map(f => Order.by(f).toOrdering)
+      )
+    for {
+      ordering <- orderingGen
+      sortedSet <- arbitrary[Set[A]].map(s => SortedSet.empty[A](ordering) ++ s)
```
johnynek, line 82: it's kind of weird to mix map with for, what about:

```scala
for {
  ordering <- orderingGen
  set <- arbitrary[Set[A]]
} yield (SortedSet.empty(ordering) ++ set)
```
or
```scala
for {
  ordering <- orderingGen
  set <- arbitrary[Set[A]]
  sortedSet = SortedSet.empty(ordering) ++ set
} yield sortedSet
```

kernel/src/main/scala/cats/kernel/Order.scala
```diff
@@ -220,15 +218,73 @@ object Order extends OrderFunctions[Order] with OrderToOrderingConversion {
...
         x.compareTo(y)
     }
 }
+
+/** Wrapper class used to track when we create an `Order` instance from an
+  * `Ordering` instance. This allows us to know that given `Order` instance is
+  * consistent with the given `Ordering` instance. This becomes important when
+  * dealing with structures which rely on `Ordering`, e.g. `SortedSet`.
+  */
+final private[kernel] class OrderFromOrdering[A] private (val value: Ordering[A]) extends Order[A] {
```
johnynek, line 235: any reason not to just make it a case class and get the equals/hashCode for free?

kernel/src/main/scala/cats/kernel/Order.scala
```diff
@@ -220,15 +218,73 @@ object Order extends OrderFunctions[Order] with OrderToOrderingConversion {
...
+        new OrderFromOrdering(ev)
+    }
+}
+
+/** Wrapper class used to track when we create an `Ordering` instance from an
+  * `Order` instance. This allows us to know that given `Order` instance is
+  * consistent with the given `Ordering` instance. This becomes important when
+  * dealing with structures which rely on `Ordering`, e.g. `SortedSet`.
+  */
+final private[kernel] class OrderingFromOrder[A] private (val value: Order[A]) extends Ordering[A] {
```
johnynek, line 267: same question about why not case class here?

kernel/src/main/scala/cats/kernel/Order.scala
isomarcte: We can do that. I usually just think case classes for data, classes for typeclasses and F-Algebras, but in this case we are blurring those lines.

kernel/src/main/scala/cats/kernel/Order.scala
satorg: Could it be `AnyVal` as well?

kernel-laws/shared/src/test/scala/cats/kernel/laws/LawTests.scala
satorg: From my prospective (which can be wrong), `Gen.zip` could be a bit more suitable here:
```
Gen.zip(orderingGen, arbitrary[Set[A]]).map { case (ordering, set) => SortedSet.empty(ordering) ++ set }
```

kernel/src/main/scala/cats/kernel/Order.scala
satorg: Oh, disregard please. I see that it does not make a lot of sense to make it `AnyVal` according to the way it is used.

isomarcte:
@armanbilge @joroKr21 @johnynek @satorg @durban I'd like to propose we go a different direction here.

@armanbilge I did some more digging into why the `Order[A]` derived from `Arbitrary.arbitrary[A => Int]` was failing and the situation seems to be more difficult than I'd originally thought. It appears that the [BoundedSemilattice Law Tests](https://github.com/typelevel/cats/blob/v2.7.0/tests/src/test/scala/cats/tests/SortedSetSuite.scala#L35 "SortedSet BoundedSemilattice Law Tests") for SortedSet were failing because they require an `Eq[SortedSet[A]]`. However, the first `Eq[SortedSet[A]]` that we get is [the one for Hash[SortedSet], by implicit priority](https://github.com/typelevel/cats/blob/v2.7.0/kernel/src/main/scala/cats/kernel/instances/SortedSetInstances.scala#L59). This isn't currently sorting the two sets as a fallback, and worse, I'm not certain that would even work here, because we don't have a coherent `Order` constraint in scope for our `Hash[SortedSet[A]]`. We might be able to resolve all of these issues, though at this point I'm not 100% sure, but this feels like we are going in an increasingly complex and confusing direction, with some performance cost, all because we cann't be sure of the coherency of the `Ordering` instances.

I'd like to propose an alternative approach. What if we added a newtype over `SortedSet` called `OrderedSet`, which can only be constructed? I have a draft PR discussing the merits and pitfalls here: #4105

Until we decide which path we are going to take, if either, I'm going to move this to draft.

isomarcte:
Closed in favor of #4092

Issue #4070, opened 2021-12-03
Missing NonEmpty Collection Helper methods


zarthross:
In the Cats library there are several aliases for various NonEmpty types and methods, for example `EitherNec` and `1.leftNec`.  
However, there are a lot of inconsistencies with when a `*Nec` or `*Nel` or other non-empty collection version of a method exists. 

We have the following non-empty collection types.

- `NonEmptyChain`
- `NonEmptyLazyList`
- `NonEmptyList`
- `NonEmptyMap` (Doesn't seem to be used for any 'left' sides so not included below)
- `NonEmptySeq`
- `NonEmptySet`
- `NonEmptyVector`

See also [PR: 3998](https://github.com/typelevel/cats/pull/3998) For a WIP to add a few more methods not listed here.

Below is an accounting of the missing methods.  I would be happy to do a PR to add the below methods, but before I did,
I wanted to make sure there isn't some underlying principle, or typeclass, or something we could use to optimize this to a degree
so less of these helper methods will be added, while still granting their utility.

Some Ideas besides just adding the methods:

- Adding `toNe*` methods to the `Bifunctor` syntax/ops.
- Add a `OneOf` typeclass (sort of thing) of the form `trait OneOf[F[_], A]{ def one(a: A): F[A]}` and make a add a generic `leftLift` method on `Bifunctor`.
  - `OneOf` differs from `Pure` in AlleyCats, because it allows you to say:
    - `implicit def oneOfNes[A: Order]: OneOf[NonEmptySet, A] = a => NonEmptySet[F, A].one(a)`
    - `implicit def oneOfApplicative[F[_]: Applicative, A]: OneOf[F, A] = a => Applicative[F].pure(a)`

  - `either.leftLift[NonEmptyLazyList]: EitherNell[A, B]`
  - Added bonus, this works with Non-NonEmpty lists as well!
    - `either.leftLift[List]: Either[List[A], B]`

### Type Aliases

```scala
// Exists:
type EitherNel[+E, +A] = Either[NonEmptyList[E], A]
type EitherNec[+E, +A] = Either[NonEmptyChain[E], A]
type EitherNes[E, +A] = Either[NonEmptySet[E], A]

// Missing:
type EitherNeSeq[+E, +A] = Either[NonEmptySeq[E], A] // NeSeq isn't a great name, happy to hear ideas for other aliases.
type EitherNev[+E, +A] = Either[NonEmptyVector[E], A]
[...]

PR #4105, opened 2022-01-05
OrderedSet/OrderedMap

isomarcte:
`OrderedSet` is an newtype over `SortedSet`, but it enforces that all construction use an `Order`. The intent here is to guarantee that we are working with the coherent `Order` instance for some type `A`, allowing us to trivially construct lawful instances for `OrderedSet` which become non-trivial for `SortedSet` where we might be dealing with a non-coherent `Ordering`.

This is a proposed alternative to #4104. By taking this route, we just assume that any instance of `OrderedSet` is using the coherent `Order`, and for such instances all the typeclass implementations are trivial and lawful.

If this approach is generally accepted, I intend to make the following changes.

- [ ] Add an similar `OrderedMap` newtype
- [ ] Rewrite `NonEmptySet` in terms of `OrderedSet`. This should be doable in a binary compatible manner, as our newtypes _are_ actually `SortedSet` values.
- [ ] Rewrite `NonEmptyMap` in terms of `OrderedMap` and `OrdredSet`.
- [ ] Add all the missing ScalaDoc.

We also might consider,

* Deprecating the `SortedSet` instances.
* Moving the `SortedSet` instances to alleycats?

### DIFF
--- a/core/src/main/scala/cats/data/OrderedSet.scala
+++ b/core/src/main/scala/cats/data/OrderedSet.scala
@@ -48,14 +48,8 @@ private[data] object OrderedSetImpl extends OrderedSetInstances with Newtype {
 @suppressUnusedImportWarningForScalaVersionSpecific
 final class OrderedSetOps[A](override val set: OrderedSet[A]) extends SetOpsForOrderedSets[OrderedSet, OrderedSet, A] {
 
-  def toSortedSet: SortedSet[A] =
+  override def toSortedSet: SortedSet[A] =
     OrderedSetImpl.unwrap(set)
-
-  def nonEmpty: Boolean =
-    toSortedSet.nonEmpty
-
-  def isEmpty: Boolean =
-    !nonEmpty
 
   def headOption: Option[A] =
     toSortedSet.headOption
@@ -101,29 +95,8 @@ final class OrderedSetOps[A](override val set: OrderedSet[A]) extends SetOpsForO
   override def union(xs: OrderedSet[A]): OrderedSet[A] =
     OrderedSetImpl.create(set.toSortedSet ++ xs.toSortedSet)
 
-  override def apply(a: A): Boolean =
-    toSortedSet(a)
-
   override def diff(xs: OrderedSet[A]): OrderedSet[A] =
     OrderedSetImpl.create(toSortedSet -- xs.toSortedSet)
-
-  override def length: Int =
-    toSortedSet.size
-
-  override def foldLeft[B](b: B)(f: (B, A) => B): B =
-    toSortedSet.foldLeft(b)(f)
-
-  override def foldRight[B](lb: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B] =
-    Foldable.iterateRight(toSortedSet, lb)(f)
-
-  override def exists(f: A => Boolean): Boolean =
-    toSortedSet.exists(f)
-
-  override def forall(f: A => Boolean): Boolean =
-    toSortedSet.forall(f)
-
-  override def find(f: A => Boolean): Option[A] =
-    toSortedSet.find(f)
 
   override def filter(p: A => Boolean): OrderedSet[A] =
     OrderedSetImpl.create(toSortedSet.filter(p))
--- a/core/src/main/scala/cats/data/OrderedSet.scala
+++ b/core/src/main/scala/cats/data/OrderedSet.scala
@@ -2,6 +2,7 @@ package cats
 package data
 
 import cats.kernel._
+import kernel.compat.scalaVersionSpecific._
 import scala.annotation.tailrec
 import scala.collection.immutable.SortedSet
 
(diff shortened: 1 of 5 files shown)


core/src/main/scala/cats/data/OrderedSet.scala
```diff
@@ -0,0 +1,231 @@
...
+    } else {
+      implicit val ordering: Ordering[B] = Order[B].toOrdering
+      Some(toSortedSet.maxBy(f))
+    }
+
+  override def show(implicit A: Show[A]): String =
+    toSortedSet.iterator.map(A.show).mkString("OrderedSet(", ", ", ")")
+
+  override def ===(that: OrderedSet[A])(implicit A: Eq[A]): Boolean =
+    Eq[Int].eqv(length, that.length) && cats.kernel.instances.StaticMethods
```
johnynek, line 86: We don't need (and shouldn't use) `Eq[Int].eqv(a, b)` instead of `a == b`. It is just needlessly slower.

core/src/main/scala/cats/data/SetOps.scala
```diff
@@ -0,0 +1,125 @@
...
+  * @tparam F This is the input type we are operating on. It is the result
+  *           type for operations which can never yield an empty set.
+  * @tparam G This is the the result type for operations which ''may'' yield
+  *           an empty set.
+  *
+  * The use of the types `F` and `G` allow us to share this trait between set
+  * types which may be have an empty state, and those which many not,
+  * e.g. `OrderedSet` and `NonEmptySet`.
+  */
+private[data] trait SetOps[F[_], G[_], A] {
```
johnynek, line 21: I'd like a longer discussion on this. I'm not sure I see the need for a layer of indirection here. Why not just implement the ops directly?

core/src/main/scala/cats/data/SetOps.scala
```diff
@@ -0,0 +1,125 @@
...
+
+  def foldRight[B](lb: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B]
+
+  def exists(f: A => Boolean): Boolean
+
+  def forall(f: A => Boolean): Boolean
+
+  def show(implicit A: Show[A]): String
+
+  def ===(that: F[A])(implicit A: Eq[A]): Boolean
```
johnynek, line 49: this signature is problematic for sorted sets. What if you pass an `Eq` that is in conflict with the `Order`?

I think if we keep this, `def ===(that: F[A]): Boolean` should be the signature (since all sets have either internal Order or Hash that gives them Eq).

johnynek (changes requested):
I really don't follow how this is more sound than `SortedSet`.

`Order` has combinator methods on it. People can treat `Order` as a value and pass it around just like `Ordering`. I don't see that this buys us anything in terms of any soundness. It seems just as unsound as what we have now, but with the added cost of having a new type and confusing matters.

My bias is -1 on this and I am leaning towards just adding comments to the `SortedSet` and `SortedMap` typeclasses and warn that they are only safe with coherent `Ordering`.

btw: I imagine we can have other instances of this problem where we assume two implicits for a given type accepted at two different points are the same, but there is nothing that requires that. I think we should be circumspect here and probably inspect the library more closely before we start working on ad-hoc solutions.

I suspect this is not solvable except by the means I suggested (lifting the singleton type of the Ordering into the type of the set/map either as a type parameter or equivalently as a path dependent type).

isomarcte:
@johnynek 

> People can treat Order as a value and pass it around just like Ordering.
> 
> btw: I imagine we can have other instances of this problem where we assume two implicits for a given type accepted at two different points are the same, but there is nothing that requires that. I think we should be circumspect here and probably inspect the library more closely before we start working on ad-hoc solutions.

If we can't assume semantic coherence between two typeclass instances, then we certainly are in a difficult situation. Many of the behaviors we have laws for become suspect. Some languages enforce coherence at the language level, Rust or Haskell for example, but since we don't have that in Scala (for better or worse) if we desire to have lawful behavior we have to gain it by idiom. I am operating off of the idea that while it is not technically impossible for someone to use cats classes incoherently, we assume that they are used coherently. I am also operating off of the idea that it is not unheard of for `Ordering` to be used incoherently. This is based off feedback on #4092 as well as the Scaladoc for Ordering itself.

From the Ordering scaladoc,

> This trait and scala.math.Ordered both provide this same functionality, but in different ways. A type T can be given a single way to order itself by extending Ordered. Using Ordering, this same type may be sorted in many other ways. Ordered and Ordering both provide implicits allowing them to be used interchangeably.

That said, I honestly don't think usage of that sort is common in practice.

1. We can just treat `Ordering` as coherent. If that is the case, then we can just assume that any given `SortedSet` and `SortedMap` has the same Ordering for the purposes of typeclass implementation. In that case we don't need to be concerned about the iteration order over the `SortedMap` in #4092. (this is how I feel down this rabbit hole ;))
2. We can say `Ordering` is not trusted to be coherent, but `Order` is, and use an approach similar to what I've started here.

I think either of those are fine options.

core/src/main/scala/cats/data/OrderedSet.scala
isomarcte: If we proceed here, I'll update.

core/src/main/scala/cats/data/SetOps.scala
isomarcte: We don't have a great track record of have consistency of methods on our data classes. See: #4070. 

I was trying to avoid that pitfall again here.

I'm honestly on the fence though too, and if we were to proceed here we'd still have to have some specialized methods for `OrderedSet` and `NonEmptySet`. 

If we proceed and people don't like this, I'd be happy to just make them normal ops.

core/src/main/scala/cats/data/SetOps.scala
isomarcte: Good point. I'll update this if we proceed with this PR.

isomarcte:
@johnynek @armanbilge @joroKr21 I've updated #4092 as an alternative to this. It assumes that for any given `SortedMap` it is safe to trust that the ordering is coherent.

Let me know if you'd like to me to proceed on #4092, this PR, or another path. Thanks for all the feedback.

PR #4108, opened 2022-01-11
Add toIorT On EitherT

isomarcte:
Similar to `toEither` on `IorT`.

### DIFF
--- a/core/src/main/scala/cats/data/EitherT.scala
+++ b/core/src/main/scala/cats/data/EitherT.scala
@@ -650,7 +650,7 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
 
   /** Convert this `EitherT[F, A, B]` into an `IorT[F, A, B]`.
    */
-  def toIorT: IorT[A, B] =
+  def toIorT: IorT[F, A, B] =
     IorT.fromEitherF(value)
 }
 
--- a/core/src/main/scala/cats/data/EitherT.scala
+++ b/core/src/main/scala/cats/data/EitherT.scala
@@ -647,6 +647,11 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
         case Left(a)  => Validated.invalidNec(a)
       }
     )
+
+  /** Convert this `EitherT[F, A, B]` into an `IorT[F, A, B]`.
+   */
+  def toIorT: IorT[A, B] =
+    IorT.fromEitherF(value)
 }
 
 object EitherT extends EitherTInstances {
--- a/core/src/main/scala/cats/data/EitherT.scala
+++ b/core/src/main/scala/cats/data/EitherT.scala
@@ -650,7 +650,7 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
 
   /** Convert this `EitherT[F, A, B]` into an `IorT[F, A, B]`.
    */
-  def toIorT: IorT[F, A, B] =
+  def toIorT(implicit F: Functor[F]): IorT[F, A, B] =
     IorT.fromEitherF(value)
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -647,6 +647,11 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
         case Left(a)  => Validated.invalidNec(a)
       }
     )
+
+  /** Convert this `EitherT[F, A, B]` into an `IorT[F, A, B]`.
+   */
+  def toIorT: IorT[A, B] =
```
rossabaker, line 653: I think...

```suggestion
  def toIorT: IorT[F, A, B] =
```

rossabaker (commented):
:+1: conceptually, but I don't think it's going to compile.

rossabaker (approved):
:+1: if this one's green

satorg:
I'd suggest to add a test for this new method into `EitherTSuite`.

core/src/main/scala/cats/data/EitherT.scala
```diff
@@ -647,6 +647,11 @@ final case class EitherT[F[_], A, B](value: F[Either[A, B]]) {
         case Left(a)  => Validated.invalidNec(a)
       }
     )
+
+  /** Convert this `EitherT[F, A, B]` into an `IorT[F, A, B]`.
+   */
+  def toIorT(implicit F: Functor[F]): IorT[F, A, B] =
```
satorg, line 653: I am not sure about the name. For example, conversion to `OptionT` has the name  just `toOption` (no `T` at the end). Should we follow the same convention here and name it as just `toIor` instead?

isomarcte: @satorg hmm. I had noticed something similar on `IorT` with `toEither: EitherT[F, A, B]`. I had initially assumed this was a typo, but maybe it is an intentional convention. It seems strange to me, as there are two distinct ways view an `EitherT[F, A, B]` as an optional value, `OptionT[F, B]` and `F[Option[B]]` and I could see having both `toOptionT` and `toOptionF`. It seems we are a bit inconsistent right now.

For example, in `IorT.scala` we have [`fromEither`, which operates on `Either[A, B]` as opposed to `EitherT[F, A, B]`](https://github.com/typelevel/cats/blob/main/core/src/main/scala/cats/data/IorT.scala#L312) and [`fromEitherF` which operates on `F[Either[A, B]]`](https://github.com/typelevel/cats/blob/main/core/src/main/scala/cats/data/IorT.scala#L324).

It seems like if we want to standardize on a convention which can work with `A`, `F[A]`, and `G[F, A]` (where `G` is some transformer type), then we have to go with the `F` and `T` suffixes.

Thoughts?

satorg: @isomarcte I'm not that certain about the reasoning behind that convention, but I _guess_ that it just does not make a lot of sense to have both `toOptionT` returning `OptionT` and `toOptionF` returning `F[Option[...]]`, because `OptionT` implies `F[Option[...]]` already via its `value` field. And seems the same is applicable for `EitherT` and `IorT`.

isomarcte: @satorg I'll update it. It does seem to be the convention.

isomarcte: @satorg updated.

isomarcte:
@satorg test added

rossabaker (approved):
Good catch, @satorg.  I don't like the convention, but it's good to follow it.

satorg (approved):

PR #4113, opened 2022-01-20
Remove unused type parameter in Choice typeclass doc

b-vennes:
The `choice` function in the `Choice` typeclass documentation has an unused/unnecessary type parameter `D`.


johnynek (approved):
*edit*

Whoops, this looks great. I previous incorrectly said the below but this is just docs.

-----

I approve but this will be binary compatible but source incompatible.

The compiler will catch it for people and it is very easy to fix. Also I bet Choice is *very* rarely implemented. So given all this, I'm on the side of merge.

armanbilge:
@johnynek This is just docs :)

armanbilge (approved):

PR #4114, opened 2022-01-21
Add F[Boolean] syntax

BalmungSan:
I know that, in general, we prefer to avoid `F[Boolean]` but sometimes we have those values and I personally think this syntax can be very useful.

Let me know what you think about this change and the **Scaladoc** _(I couldn't come up with something better)_.
Also, for testing this change I could only think on a simple truth table but I am not sure if that is a good test.

### DIFF
--- a/core/src/main/scala/cats/syntax/monad.scala
+++ b/core/src/main/scala/cats/syntax/monad.scala
@@ -2,10 +2,14 @@ package cats
 package syntax
 
 trait MonadSyntax {
-  implicit final def catsSyntaxMonad[F[_], A](fa: F[A]): MonadOps[F, A] = new MonadOps(fa)
+  implicit final def catsSyntaxMonad[F[_], A](fa: F[A]): MonadOps[F, A] =
+    new MonadOps(fa)
 
   implicit final def catsSyntaxMonadIdOps[A](a: A): MonadIdOps[A] =
     new MonadIdOps[A](a)
+
+  implicit final def catsSyntaxMonadBooleanOps[F[_]](fa: F[Boolean]): MonadBooleanOps[F] =
+    new MonadBooleanOps[F](fa)
 }
 
 final class MonadIdOps[A](private val a: A) extends AnyVal {
@@ -19,4 +23,29 @@ final class MonadIdOps[A](private val a: A) extends AnyVal {
    * Iterative application of `f` until `p` holds.
    */
   def iterateUntilM[F[_]](f: A => F[A])(p: A => Boolean)(implicit M: Monad[F]): F[A] = M.iterateUntilM(a)(f)(p)
+}
+
+final class MonadBooleanOps[F[_]](private val fa: F[Boolean]) extends AnyVal {
+
+  /**
+   * Behaves like `&&` but inside the `F` context.
+   *
+   * Wont evaluate `other` unless this evaluates to `true`
+   */
+  def andM(other: => F[Boolean])(implicit F: Monad[F]): F[Boolean] =
+    F.flatMap(fa) {
+      case true  => other
+      case false => F.pure(false)
+    }
+
+  /**
+   * Behaves like `||` but inside the `F` context.
+   *
+   * Wont evaluate `other` unless this evaluates to `false`
+   */
+  def orM(other: => F[Boolean])(implicit F: Monad[F]): F[Boolean] =
+    F.flatMap(fa) {
+      case true  => F.pure(false)
+      case false => other
+    }
 }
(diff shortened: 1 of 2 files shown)


rossabaker:
There's no clear line between the two, but a lot of syntactic convenience goes in mouse instead of cats.  Might this fit better there?

core/src/main/scala/cats/syntax/monad.scala
```diff
@@ -20,3 +24,28 @@ final class MonadIdOps[A](private val a: A) extends AnyVal {
...
+   * Behaves like `||` but inside the `F` context.
+   *
+   * Wont evaluate `other` unless this evaluates to `false`
+   */
+  def orM(other: => F[Boolean])(implicit F: Monad[F]): F[Boolean] =
+    F.flatMap(fa) {
+      case true  => F.pure(false)
+      case false => other
+    }
+}
```
johnynek, line 51: This only needs Apply I think, not Monad.

BalmungSan:
I renamed `and` & `or` to `andM` & `orM` per recommendation from Rob in the **discord** server.

I also removed `xor` since I think the main selling point of this addition is preserving the laziness that most folks expect of `and` & `or` which you lose with the typical advice of `(fa, fb).mapN(_ && _)`

core/src/main/scala/cats/syntax/monad.scala
```diff
@@ -20,3 +24,28 @@ final class MonadIdOps[A](private val a: A) extends AnyVal {
...
+  def andM(other: => F[Boolean])(implicit F: Monad[F]): F[Boolean] =
+    F.flatMap(fa) {
+      case true  => other
+      case false => F.pure(false)
+    }
+
+  /**
+   * Behaves like `||` but inside the `F` context.
+   *
+   * Wont evaluate `other` unless this evaluates to `false`
```
johnynek, line 44: This is wrong. Also underscores the need for tests.

core/src/main/scala/cats/syntax/monad.scala
```diff
@@ -20,3 +24,28 @@ final class MonadIdOps[A](private val a: A) extends AnyVal {
    */
   def iterateUntilM[F[_]](f: A => F[A])(p: A => Boolean)(implicit M: Monad[F]): F[A] = M.iterateUntilM(a)(f)(p)
 }
+
+final class MonadBooleanOps[F[_]](private val fa: F[Boolean]) extends AnyVal {
+
```
johnynek, line 29: She we also add nand, nor, any, all?

johnynek (changes requested):

core/src/main/scala/cats/syntax/monad.scala
BalmungSan: Well, not. That is exactly the point of this addition.

Right now people can just do `(fa, fb).mapN(_ && _)` but that will evaluate `fb` even when it is not needed, if you need to preserve that _"laziness"_ then you to do this exactly `flatMap` which, IMHO, hides the intention.

core/src/main/scala/cats/syntax/monad.scala
BalmungSan: Care to elaborate why that is wrong?

core/src/main/scala/cats/syntax/monad.scala
Daenyth: That makes it no longer short-circuit which is the same kind of footgun that ifA has

core/src/main/scala/cats/syntax/monad.scala
BalmungSan: Since it seems that `and` and `or` are already niche enough that most folks seem to be against this addition, I doubt there is value on those others.

IMHO, these two are useful enough to be on **cats** but that is up to the consensus of everyone.

core/src/main/scala/cats/syntax/monad.scala
johnynek: My comment got moved. This comment was on xor but my review and your push crossed paths.

XOR only needs Apply. I understand flatMap on the others.

core/src/main/scala/cats/syntax/monad.scala
johnynek: Or returns true if either are true. This returns false if the first is true.

core/src/main/scala/cats/syntax/monad.scala
BalmungSan: @johnynek ah right, my bad, sorry.
Yeah, I removed `xor` exactly for that.

core/src/main/scala/cats/syntax/monad.scala
BalmungSan: ups... that happens for copypasting code 😅
I am sorry for that.

Yeah I agree tests are good, but then again how should I test them? The truth tables?

johnynek: I would say that for all pairs of inputs `Id[Boolean]` ops should behave like the normal ops on Boolean, which yes is equivalent to testing the truth table.

danicheg (commented):
I agree with @rossabaker that https://github.com/typelevel/mouse is a probably better place for that syntax. `Mouse` already has some `F[Foo]` syntax, so it would be natural to have `F[Boolean]` syntax too.

BalmungSan:
Moved to **mouse** based on feedback.
https://github.com/typelevel/mouse/pull/297

PR #4119, opened 2022-01-28
Rename `isDotty` -> `isScala3`

armanbilge:
This is necessary so that Cats can be updated in the Scala 3 community build.

See https://github.com/lampepfl/dotty/pull/14071#issuecomment-988528745. **tl;dr** `isDotty` conflicts with a setting of the same name injected by the community build plugin.


satorg (approved):

armanbilge:
Upon further research I'm not sure if this is actually necessary, but it's otherwise harmless.

PR #4125, opened 2022-02-04
Add `F[TupleN]` syntax

danicheg:
This adds convenient `_1F`...`_22F` methods for `F[Tuple3..22]`. At the moment, Cats contains those methods only for `F[Tuple2]`, so it would be great to have it for all Tuple families. 

I'm open to any suggestions/points about this.

### DIFF
--- a/core/src/main/scala/cats/syntax/functor.scala
+++ b/core/src/main/scala/cats/syntax/functor.scala
@@ -1,7 +1,7 @@
 package cats
 package syntax
 
-trait FunctorSyntax extends Functor.ToFunctorOps {
+trait FunctorSyntax extends Functor.ToFunctorOps with FunctorTupleNSyntax {
   implicit final def catsSyntaxFunctorTuple2Ops[F[_], A, B](fab: F[(A, B)]): FunctorTuple2Ops[F, A, B] =
     new FunctorTuple2Ops[F, A, B](fab)
 }
--- a/project/Boilerplate.scala
+++ b/project/Boilerplate.scala
@@ -36,7 +36,8 @@ object Boilerplate {
     GenTupleMonadInstances,
     GenTupleBifunctorInstances,
     GenTupleBitraverseInstances,
-    GenTupleUnorderedFoldableInstances
+    GenTupleUnorderedFoldableInstances,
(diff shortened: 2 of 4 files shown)


satorg:
As I understand, the worst concern regarding changes like this one is a potential increase of the result jar size.
Did you happen to estimate by chance how significantly the sizes got increased?

Link to a similar PR: #4009

johnynek:
this seems like a PR for mouse if you ask me:

https://github.com/typelevel/mouse

which, BTW, I would live to see merged into this repo since I believe cats is the only dependency. Then it would just be a submodule here.

danicheg:
@johnynek yeah, I was thinking about this, but it just confusing a bit to have `F[Tuple2]` syntax in the Cats and syntax for `F[TupleN]` in the Mouse (or have intersected on the same methods in both libraries).

danicheg:
Can we take yet another look at this? I have a concern about having this in the Mouse. It's just odd to have `_1F`, `_2F` for Tuple2 in Cats, but for TupleN in Mouse. But if it's common for all new syntaxes within `cats.syntax`, I doubtless will migrate this to Mouse.

armanbilge:
> It's just odd to have `_1F`, `_2F` for Tuple2 in Cats, but for TupleN in Mouse.

Yep, I agree with this. If this is the primary reason to target this PR to cats, maybe the real question is do `_1F` and `_2F` belong in Cats? IDK the history but at the time they were added there may not have been mouse.

> As I understand, the worst concern regarding changes like this one is a potential increase of the result jar size.
Did you happen to estimate by chance how significantly the sizes got increased?

Possibly off-topic, but I haven't been able to appreciate why this is such a big deal. See my comments in https://github.com/typelevel/cats/pull/3871#issuecomment-1030643684. But I would like to know if there's something I'm not getting :)

danicheg:
> IDK the history but maybe at the time they were added there was no mouse yet.

Mouse is about five years this May :) And `_1F` and `_2F` were added a year ago.

armanbilge:
I see, those methods were added in https://github.com/typelevel/cats/pull/3318.

Issue #2334, opened 2018-07-24
Instances for Future


ChristopherDavenport:
`Future` is largely known to not be referentially transparent. As it starts execution immediately and caches the result.

While it may have been largely necessary for these instances to be in cats before cats-effect, with that solution coming in maturity it may be time for Future instances to be pushed to alleycats.

Imports
```scala
import scala.concurrent.Future
import scala.concurrent.ExecutionContext.Implicits.global
import cats.implicits._ 
```

Example 1 - Prints Once on the first line
```scala
val future1 : Future[Unit] = Future(println("I printed"))
val futureResult1 : Future[Unit] = (future1, future1).mapN(_ => ())
```

Example 2 - Prints twice on the second line
```scala
def future2: Future[Unit] = Future(println("I printed"))
val futureResult2 : Future[Unit] = (future2, future2).mapN(_ => ())
```

As you can see we have lost equational reasoning.

What are your thoughts on the future of `Future`?

djspiewak:
Instances like `Monad[Future]` and `Applicative[Future]` are really really dangerous. They've caused problems in scalaz for years. They should definitely be in alleycats, if we have them at all.

johnynek:
My view is that any instance that is not lawful should not be in cats proper.

In this case, you have treated Future as though it were a Sync instance, which it is not. You put in side effects, print, and had some idea in mind. I think you can come up with other similar examples with other types with call by name arguments. If users see a call by name parameter as an invitation to put a side effect we have a big problem since we often use them just for short circuiting. I argue that is the fault of the user.

I advocate that we stick to the laws and argue for instances based on that. I don’t want cats to pick on some data types, whose instances are lawful, because they have other properties we don’t like.

E.g. I don’t like that scala uses universal equality and hashing for Map by default. However, I still think Map should have a Functor instance, which is lawful.

tpolecat:
I agree in principle that it's "user error", but the error is committed when using the data types *as intended*. `Future` instances are fine if you never side-effect; `Try` instances are fine if you never throw; cigarettes are fine if you never smoke them. I think we can identify a qualitative difference here that justifies warning labels.

As a practical matter I think this hits beginners hardest, because experienced cats users are unlikely to use these data types at all. Conversations about `Future` on Gitter invariably lead to "that's not going to work, sorry, have a look at IO" which makes me worries about those who *don't* ask. My guess is these instances are used almost always by beginners, and almost always incorrectly. 

[...]

PR #4126, opened 2022-02-04
StateT Documentation

benkio:
Add `StateT` documentation Page to the microsite
Update the latest `cats` version in `README.md`
Minor source path fix

Linked to #1801


armanbilge (commented):
Thanks for the PR and fixes!

The in-depth example for `StateT` is fantastic but not sure about using `Eval` as a hack to suspend side-effects without `IO`. Instead could we use a monad like `Writer` to "print" to a `List[String]`? WDYT?

benkio:
> Thanks for the PR and fixes!
> 
> The in-depth example for `StateT` is fantastic but not sure about using `Eval` as a hack to suspend side-effects without `IO`. Instead, could we use a monad like `Writer` to "print" to a `List[String]`? WDYT?

Hi There, 
I'm glad you like the example 😄 
I choose `Eval` over other options because:
- `IO` is not in scope
- `Eval` is used in `State` as well and is fairly simple to understand

`Writer` can be used as an alternative, but then the reader needs to know it (and maybe `WriterT`) in order to fully understand the example. I found `Eval` more straightforward compared to it and since `StateT` is closely related to `State`, there's a high chance that the reader knows it already. However, if more people likes `Writer` more, I'm happy to do the change. It was a stylistic choice 😅 
Finally, I could have used `Future` as others had done throughout the docs as effectful type, I think it should be considered as an option as well.
I'll wait for more input on the matter 👍

armanbilge:
Thanks for the detailed explanation :)

> It was a stylistic choice 😅

I appreciate that :) unfortunately, it's a stylistic choice with practical implications. Using `Eval` for `println` or indeed to suspend any side-effect is not generally accepted and is something I think we should avoid demonstrating in docs. See https://github.com/typelevel/cats-effect/issues/78#issuecomment-340711285 for discussion.

`Future` is similarly controversial, see https://github.com/typelevel/cats/issues/2334.

So my vote is for `Writer` or similar, although for sure let's give others a chance to chime in.

docs/src/main/mdoc/datatypes/statet.md
```diff
@@ -0,0 +1,177 @@
...
+    word <- Eval.later(readLine())
+  } yield GameState(word)
+
+  def step2(): StateT[Eval, GameState, Unit] = for {
+    state <- StateT.get[Eval, GameState]
+    _ <- StateT.liftF(Eval.later(println(GameState.showWordHidden(state))))
+  } yield ()
+
+  def step3(): StateT[Eval, GameState, String] = for {
+    _ <- StateT.liftF(Eval.later(println("Insert the guess word: ")))
```
johnynek, line 128: I also am -1 on using Eval as a poor man's IO monad.

We should not use a pattern we don't recommend in our documentation.

I would instead suggest using `StateT[Either[String, *], GameState, Unit]` in order to model steps that can somehow fail or something along those lines.

armanbilge: `Either` seems good. Easier to grok than `Writer`.

docs/src/main/mdoc/datatypes/statet.md
```diff
@@ -109,52 +126,56 @@ val gameState2 = GameState.attemptGuess(gameState1, "bats")
...
 object StateMachine {
 
-  def step1(): Eval[GameState] = for {
-    _ <- Eval.later(println("Insert the target word: "))
-    word <- Eval.later(readLine())
+  import EffectType._
+
+  def step1(): EffectType[GameState] = for {
+    _ <- performEffect[Unit](println("Insert the target word: "))
+    word <- performEffect[String](readLine())
```
armanbilge, line 138: Sorry, I think there was a miscommunication :) since we don't have `IO`, there's really no way we can use `readLine()` or `printLine()` or any side-effects at all in this example unfortunately.

One way forward is to use effects that we can model with datatypes available in cats. For example, `Either` can be used to raise errors and short circuit, or `Writer` can be used for logging (this would be the closest to what you currently have).

benkio: Ah ok, 
Now everything is clearer. 
Then, I'll proceed in this way:
- Refactor the code for the hangman example using `IO`, but not embedding the code into the docs. Just put an external link to the gist. Otherwise, I can just add the code, but ask `mdoc` to not compile/run it, even if it might expose it to go out of sync.
- Add another example with `Either` and some validation process that require a State as well.

Let me know your thoughts about this plan, if you like it, then I'll proceed 😉
Thanks for the feedback 🙏

armanbilge: 👍 to the table reservation example, I like it! :)

I think we can keep your current example (refactored to use `IO`) as a second, "advanced" example like you suggest. Instead of a gist I wonder if we could put it in a scastie.

docs/src/main/mdoc/datatypes/statet.md
```diff
@@ -0,0 +1,244 @@
...
+layout: docs
+title:  "StateT"
+section: "data"
+source: "core/src/main/scala/cats/data/package.scala"
+scaladoc: "#cats.data.StateT"
+---
+# StateT
+
+`StateT[F[_], S, A]` is a data type used to define state machine
+effectful computations that can be easily composed into a larger one.
```
armanbilge, line 11: Can we directly reference the `State` docs here? e.g.
> `StateT` is a monad transformer for `State` that enables you to combine the state machine effects of `State` with the effects of another monad `F[_]`

In fact, it's notable that `State[A] = StateT[Eval, A]` 😁 

http://typelevel.org/cats/datatypes/state.html

docs/src/main/mdoc/datatypes/statet.md
```diff
@@ -0,0 +1,244 @@
...
+
+`StateT` is a monad transformer for
+[State](https://typelevel.org/cats/datatypes/state.html), in
+particular [State](https://typelevel.org/cats/datatypes/state.html) is
+defined as a `StateT` with
+[Eval](https://typelevel.org/cats/datatypes/eval.html) as the effect
+`F`.  Therefore, `StateT` exposes the same methods of
+[State](https://typelevel.org/cats/datatypes/state.html), such as:
+`modify`, `get` and `set`.  Plus additional methods, that handles
+effectful computations, eg: `modifyF`, `setF` and `liftF`.
```
armanbilge, line 35: Oh, I see you already have this here :) 👍 IMO we should have this at the beginning. I think it's difficult to understand `StateT` without understanding `State`.

docs/src/main/mdoc/datatypes/statet.md
```diff
@@ -0,0 +1,244 @@
...
+Let's start with defining the type synonyms for the state of the
+program and the effect type:
+
+```scala mdoc:silent
+import cats.data.{StateT, NonEmptyList}
+import cats.implicits._
+import java.time.LocalTime
+
+type TRSState = Map[(Int, LocalTime), String]
+type EffectType[A] = Either[Throwable, A]
```
armanbilge, line 58: I'm not sure about these type aliases. Can we use a `case class` for `TRSState` instead? Maybe call it `Reservations`?

For `EffectType`, I wonder if we use `Either[Throwable, *]` directly in the code. Less sure about this one.

docs/src/main/mdoc/datatypes/statet.md
```diff
@@ -0,0 +1,244 @@
...
+`NonEmptyList` of reservations, inserting them one by one.
+
+```scala mdoc:silent
+object TableReservationSystem {
+
+  final case class AlreadyReservedTable(
+      tableNumber: Int,
+      time: LocalTime,
+      reservation: String
+  ) extends Throwable(
```
armanbilge, line 87:
```suggestion
  ) extends RuntimeException(
```

docs/src/main/mdoc/datatypes/statet.md
```diff
@@ -0,0 +1,244 @@
...
+  (2, LocalTime.parse("10:00:00"), "Lebron James"),
+  (1, LocalTime.parse("12:00:00"), "Tiger Woods"),
+  (2, LocalTime.parse("12:00:00"), "Roger Federer"),
+  (3, LocalTime.parse("13:00:00"), "Elizabeth Alexandra Mary"),
+  (1, LocalTime.parse("16:00:00"), "Chuck Norris"),
+  (2, LocalTime.parse("16:00:00"), "Kobe Bryant"),
+  (2, LocalTime.parse("18:00:00"), "Steven Seagal")
+)
+
+println(TableReservationSystem.evalBookings(bookings))
```
armanbilge, line 133: We don't need `println` in mdoc :) if you write `TableReservationSystem.evalBookings(bookings)` it will automatically put it in the output

docs/src/main/mdoc/datatypes/statet.md
```diff
@@ -0,0 +1,244 @@
...
+effectful computations that can be easily composed into a larger one.
+
+The type parameters are:
+- `F[_]` represents the effect in which the computation is performed.
+- `S` represents the underlying state, shared between each step of the
+  state machine.
+- `A` represents the return value.
+
+It can be seen as a way to add the capability to manipulate a shared
+state, to an existing computation in the context of `F`.
```
armanbilge, line 20:
```suggestion
state to an existing computation in the context of `F`.
```

docs/src/main/mdoc/datatypes/statet.md
```diff
@@ -0,0 +1,244 @@
...
+```
+
+
+`TRSState` stands for *Table Reservation System State* and it will
+organize the reservations in a `Map` with:
+- `(Int, LocalTime)` as key, representing the table number and hour
+respectively.
+- `String` as value, representing the reservation name.
+
+As effect type, we will just use an `Either[Throwable, A]` because each
```
armanbilge, line 68:
```suggestion
For the underlying effect, we will use an `Either[Throwable, A]` because each
```

docs/src/main/mdoc/datatypes/statet.md
```diff
@@ -0,0 +1,244 @@
...
+- The logic for the booking insertion. We can take advantage of the
+method `modifyF`.
+
+In addition, we can implement a simple function that will evaluate a
+`NonEmptyList` of reservations, inserting them one by one.
+
+```scala mdoc:silent
+object TableReservationSystem {
+
+  final case class AlreadyReservedTable(
```
armanbilge, line 83:
```suggestion
  final case class TableAlreadyReservedException(
```

docs/src/main/mdoc/datatypes/statet.md
```diff
@@ -0,0 +1,244 @@
...
+    )
+
+  def evalBookings(
+      bookings: NonEmptyList[(Int, LocalTime, String)]
+  ): EffectType[TRSState] =
+    bookings
+      .map { case (tn, t, n) =>
+        TableReservationSystem.insertBooking(tn, t, n)
+      }
+      .reduceLeft(_ *> _)
```
armanbilge, line 113: I worry if this line is confusing because of all the fancy syntax.

docs/src/main/mdoc/datatypes/statet.md
```diff
@@ -0,0 +1,244 @@
...
+    StateT.modifyF[EffectType, TRSState]((currentReservations: TRSState) =>
+      currentReservations
+        .get((tableNumber, time))
+        .fold[EffectType[TRSState]](Right(currentReservations + ((tableNumber, time) -> name)))(_ =>
+          Left(AlreadyReservedTable(tableNumber, time, name))
+        )
+    )
+
+  def evalBookings(
+      bookings: NonEmptyList[(Int, LocalTime, String)]
```
armanbilge, line 107: Maybe nice to have a `ReservationRequest` case class for `(Int, LocalTime, String)`

docs/src/main/mdoc/datatypes/statet.md
```diff
@@ -0,0 +1,244 @@
...
+
+```scala mdoc
+val gameState1 = GameState("cats")
+println(GameState.showWordHidden(gameState1))
+val gameState2 = GameState.attemptGuess(gameState1, "bats")
+println(GameState.showWordHidden(gameState2.getOrElse(???)))
+```
+
+Unfortunately, the rest of the example can't be shown here due to the
+required `cats-effect` dependency. We recommend to check out the rest
```
armanbilge, line 237: Would you consider contributing this example to the Cats Effect docs? :) I think it would be cool!

armanbilge (commented):
Thanks for the revisions!

docs/src/main/mdoc/datatypes/statet.md
benkio: I agree about the fact that it's hard to understand `StateT` without knowing `State`. My idea was to give an initial brief general definition of the `StateT` datatype and then specify the relationship with `State`. At the end of the day, it's just <20 lines below 😅 

I can definitely add the type synonym for `State` in here for clarity.

docs/src/main/mdoc/datatypes/statet.md
benkio: In addition to the comment (https://github.com/typelevel/cats/pull/4126/files#r810610617), if you like I can add to this first phrase a reference to `State` like:

```suggestion
`StateT[F[_], S, A]` is a data type that extends the `State` one further: it is used to define state machine computations as well, but it also allows to perform effects at each step.
`StateT` is also a `Monad`, therefore we can easily compose its values into a larger one.
```

docs/src/main/mdoc/datatypes/statet.md
benkio: I agree. I looked quickly at the `cats-effects` doc site, but it's not structured in the same way as this one. I struggle to identify where we can put general `IO` examples like the hangman one. 

What I would like more is to extend the `example` folder in the repo(or have a separate repo) with more use cases like this one, so newcomers can go, grab the code or just check it out easily. This also applies to `cats` repo actually, but this is more of a _Megathread_ 😅

docs/src/main/mdoc/datatypes/statet.md
benkio: I was thinking the same when coding it 😄 
What I can do is to create a `Reservation` case class this way:
```
case class ReservationId(tableNumber: Int, hour: LocalTime)
case class Reservation(id: ReservationId, name: String)
```
and then create a `Reservations` class extending `Map[ReservationId, Reservation]` that can be constructed from `NonEmptyList[Reservation]`. I would like to keep the `Map` capabilities due to the fact that the comparison needs to happen on the key only in order to identify clashes later on, wdyt?

Regarding the `EffectType` to `Either[Throwable, *]`. It can definitely be done, but since it's used multiple times below, I find it more concise to have a type synonym. It's still a matter of taste, if you insist one more time I'll do the change 😛

docs/src/main/mdoc/datatypes/statet.md
benkio: related to this comment above https://github.com/typelevel/cats/pull/4126/files#r810615136

docs/src/main/mdoc/datatypes/statet.md
benkio: This is actually the line for composing the `StateT`s to a single one. I'm open to suggestions to make it less cryptic 🙌

docs/src/main/mdoc/datatypes/statet.md
armanbilge: I agree, a brief general definition is always very good :) "state machine effectful computations" is something that's really hard to know what that means though, IMO. Even "state machiner" and "effect" by themselves are kind of complex!

docs/src/main/mdoc/datatypes/statet.md
armanbilge: Btw I had another idea for this one, maybe `type ThrowableOr[A] = Either[Throwable, A]`. Actually I wonder why we don't have this alias in cats.

armanbilge: > create a `Reservations` class extending `Map[ReservationId, Reservation]`

Hmm, I think this would be rather unusual, like using `Eval` to suspend side-effects. Using a `case class` (and defining extra methods on it) is more idiomatic than extending a `Map` (especially the Scala collections APIs are very complicated, even the collections in cats do not extend them).

I like your case classes 👍

docs/src/main/mdoc/datatypes/statet.md
armanbilge: Sure :) maybe something like this.
```scala
.reduceLeft((acc, bookNext) => acc.productL(bookNext))
```

docs/src/main/mdoc/datatypes/statet.md
armanbilge: That's true, I guess we'd have to add an "examples" section to the cats-effect docs. Hum.

Yes, this is totally a mega-thread issue :) Probably we need a unified Typelevel docs site + examples repo, that can show examples using multiple projects together, like cats and cats-effect. There are already a few projects for this purpose, like https://github.com/djspiewak/kitteh-redis.

I know the typelevel.org website is currently being redesigned and I had some ideas in https://github.com/typelevel/governance/discussions/28 that maybe we can raise some funding for improving documentation as well. Feel free to comment there or raise an issue about this topic since you seem passionate about it :)

docs/src/main/mdoc/datatypes/statet.md
benkio: I see, let's iterate once more then :) 
```suggestion
StateT[F[_], S, A]` is a data type that extends the `State` one further: it is used to define state machine computations as well, but it also allows to perform computations returning a value enclosed in `F` at each step. Such capability is often called `lifting`.
In addition, `StateT` values share the same composition properties as `F` values: if `F` is a `Functor` then `StateT[F,A,B]` will also be a `Functor`. However, beware that different `StateT` operations might required different typeclass instances, refer to the [API](https://typelevel.org/cats/api/cats/data/package$$StateT$.html) for the details.
```

docs/src/main/mdoc/datatypes/statet.md
benkio: Fair, I'll apply such changes ;)

docs/src/main/mdoc/datatypes/statet.md
benkio: I'll definitely have a look there 😉 

Passionate...I'm just a lazy coder who likes to copy-paste code 😝 That's why I like the `examples` repo idea

docs/src/main/mdoc/datatypes/statet.md
benkio: True, the problem is that some stuff is not at ground level by definition. 
What we can do is to put reference links where the reader can go and grasp the "prerequisite concepts", eg a Wikipedia link to [state machine](https://en.wikipedia.org/wiki/Finite-state_machine) or a youtube link to what an effect system is (I just attended an amazing one as keynote in ScalaLove 2022 <3 )

docs/src/main/mdoc/datatypes/statet.md
```diff
@@ -0,0 +1,281 @@
...
+      )
+
+  val emptyReservationSystem: Reservations = Reservations(List.empty)
+
+  def insertBooking(
+      reservation: Reservation
+  ): StateT[ThrowableOr, Reservations, Unit] =
+    StateT.modifyF[ThrowableOr, Reservations](_.insert(reservation))
+
+  def evalBookings(
```
armanbilge, line 108: `eval` is confusing with `Eval`. Maybe `processBookings`?

docs/src/main/mdoc/datatypes/statet.md
armanbilge: hmm 🤔 how about

> `StateT[F[_], S, A]` is a data type that generalizes `State` with the ability to compose with effects in `F[_]`. Because `StateT` is defined in terms of `F`, it is a monad only if `F` is a monad. Additionally, `StateT` may acquire new capabilities via `F`: for example, if `F` is capable of error handling via `MonadThrow[F]`, then Cats derives an instance of `MonadThrow[StateT[F, S, *]]`.

docs/src/main/mdoc/datatypes/statet.md
```diff
@@ -0,0 +1,285 @@
...
+do so, we need to keep track of the tables and the current
+reservations for each of them. The end-user can then try to insert a
+booking for a specific table and time. If such a table is available,
+then the booking is placed and the state is updated, otherwise, an
+error is returned.
+
+To simplify the logic, for each reservation we will just consider a
+single `LocalTime` starting at the beginning of the hour.
+
+Let's start with defining the type synonym for the effect type:
```
armanbilge, line 62:
```suggestion
Let's start with defining the type alias for the effect type:
```

docs/src/main/mdoc/datatypes/statet.md
```diff
@@ -0,0 +1,285 @@
...
+
+```scala mdoc:silent:reset
+import cats.data.{StateT, NonEmptyList}
+import cats.implicits._
```
armanbilge, line 66:
```suggestion
import cats.syntax.all._
```

`cats.implicits._` is deprecated

docs/src/main/mdoc/datatypes/statet.md
```diff
@@ -0,0 +1,285 @@
...
+import java.time.LocalTime
+
+type ThrowableOr[A] = Either[Throwable, A]
+```
+
+We will use an `Either[Throwable, A]` because each
+booking insertion can fail.
```
armanbilge, line 73:
```suggestion
We will use an `Either[Throwable, A]` to model either success `Right(a)` or failure `Left(ex)`.
```

docs/src/main/mdoc/datatypes/statet.md
```diff
@@ -0,0 +1,285 @@
...
+  final case class ReservationId(tableNumber: Int, hour: LocalTime)
+  final case class Reservation(id: ReservationId, name: String)
+  final case class Reservations(reservations: List[Reservation]) {
+    def insert(reservation: Reservation): ThrowableOr[Reservations] =
+      if (reservations.exists(r => r.id == reservation.id))
+        Left(TableAlreadyReserved(reservation))
+      else Right(Reservations(reservations :+ reservation))
+  }
+
+  final case class TableAlreadyReserved(
```
armanbilge, line 99:
```suggestion
  final class TableAlreadyReservedException(
```

A personal thing but I think using `case class` for exceptions is an anti-pattern.

docs/src/main/mdoc/datatypes/statet.md
```diff
@@ -0,0 +1,285 @@
...
+```
+
+We will use an `Either[Throwable, A]` because each
+booking insertion can fail.
+
+Now, we need to implement/define:
+- The type representing the reservation.
+- The type representing the state of the Table Reservation System. It
+  will wrap around a collection of Reservations.
+- An initial state, that will be just an empty state type.
```
armanbilge, line 79:
```suggestion
- An initial state, that will be just empty (no reservations).
```

docs/src/main/mdoc/datatypes/statet.md
```diff
@@ -0,0 +1,285 @@
...
+
+  def processBookings(
+      bookings: NonEmptyList[Reservation]
+  ): ThrowableOr[Reservations] =
+    bookings
+      .map(insertBooking)
+      .reduceLeft[StateT[ThrowableOr, Reservations, Unit]] {
+        case (acc, bookNext) => acc.productL(bookNext)
+      }
+      .runS(emptyReservationSystem)
```
armanbilge, line 120: I'm not sure if it's good for an example but this is `traverse` :)

docs/src/main/mdoc/datatypes/statet.md
```diff
@@ -0,0 +1,285 @@
...
+    name = "Bruce Lee"
+  )
+)
+```
+
+The full source code of this example can be found at this
+[gist](https://gist.github.com/benkio/baa4fe1d50751cd602c4175f1bb39f4d)
+or [scastie](https://scastie.scala-lang.org/MFHZpU9bRg2e73MLQY1dNQ)
+
+## Example: Hangman Game
```
armanbilge, line 186: > Unfortunately, the rest of the example can't be shown here due to the

I think it's a bit strange to have 1.5 example :) maybe you can host this one in a personal website/gist, and we can link to from the cats site.

docs/src/main/mdoc/datatypes/statet.md
```diff
@@ -0,0 +1,285 @@
...
+
+Unfortunately, the rest of the example can't be shown here due to the
+required `cats-effect` dependency. We recommend to check out the rest
+of the code at the following
+[gist](https://gist.github.com/benkio/46f5aea4f15ec059f02d6bfe9bd25e99)
+or [scastie](https://scastie.scala-lang.org/4Ab7xspkRJ2q9UKQ9OHrUQ).
+
+We hope these examples help to clarify how `StateT` can be used in
+designing a computation based on state machine steps that may require
+side effects or other capabilities, eg `Option`, `List`, `Fiber` etc.
```
armanbilge, line 285: `Fiber` is not like the others :) it's not a monad, and it's not part of Cats anyway. Maybe `Validated` could be a good addition to this list instead.

armanbilge (commented):
Thanks, just a few comments, I think it's in pretty good shape! :)

docs/src/main/mdoc/datatypes/statet.md
benkio: My bad about `Fiber`, I'll add `Validated` to the list

docs/src/main/mdoc/datatypes/statet.md
benkio: There's a `gist` and a `scastie` at the end of the example as a "continuation". So you are suggesting to remove the "possible" explanation and just leave those links at the end of the documentation hinting something like:
"for a more advanced example check..."

armanbilge: Thanks, yes, exactly. Imo it seems odd to start an example we can actually finish :)

armanbilge (changes requested):
Can we change up the reservations? Children's characters are always nice. Madeline, Elmo, Winnie-the-Pooh, etc. :)

armanbilge (approved):
Thanks for all your hard work on this! This is a very nice contribution :)

armanbilge (approved):

PR #4142, opened 2022-03-05
remove unused `Apply#tupleN` type param

xuwei-k:


johnynek (approved):
good catch. It would be nice if this were easier to review (like a system that showed us the generated code).

rossabaker (approved):

Issue #4144, opened 2022-03-08
`ifF` not available on `Option[Boolean]`


atais:
maybe I am missing something, but it should be?

example code:

```
import cats.implicits._
val a = Some(true)
a.ifM(Some("a"), Some("b"))
```

works fine, but cannot resolve for:

```
import cats.implicits._
val a = Some(true)
a.ifF("a", "b")
```

```
value ifF is not a member of Option[Boolean]
```

scala 2.12
cats 2.7

PR #4146, opened 2022-03-08
added ifF to Functor

atais:
Resolving #4144 

This is my first contribution, sorry if it's missing anything.

I am also not sure if this is "the correct way" to add this functionality to Functor. 
It seemed clean, but on the other hand - I have copied docs for the function, is that ok?

### DIFF
--- a/core/src/main/scala/cats/syntax/functor.scala
+++ b/core/src/main/scala/cats/syntax/functor.scala
@@ -4,6 +4,9 @@ package syntax
 trait FunctorSyntax extends Functor.ToFunctorOps {
   implicit final def catsSyntaxFunctorTuple2Ops[F[_], A, B](fab: F[(A, B)]): FunctorTuple2Ops[F, A, B] =
     new FunctorTuple2Ops[F, A, B](fab)
+
+  implicit final def catsSyntaxIfF[F[_]](fa: F[Boolean]): IfFOps[F] =
+    new IfFOps[F](fa)
 }
 
 final class FunctorTuple2Ops[F[_], A, B](private val fab: F[(A, B)]) extends AnyVal {
@@ -61,4 +64,20 @@ final class FunctorTuple2Ops[F[_], A, B](private val fab: F[(A, B)]) extends Any
    * }}}
    */
   def unzip(implicit F: Functor[F]): (F[A], F[B]) = F.unzip(fab)
+}
+
+final class IfFOps[F[_]](private val fa: F[Boolean]) extends AnyVal {
+
+  /**
+   * Lifts `if` to Functor
+   *
+   * Example:
+   * {{{
+   * scala> import cats.syntax.all._
+   *
+   * scala> List(true, false, false).ifF(1, 0)
+   * res0: List[Int] = List(1, 0, 0)
+   * }}}
+   */
+  def ifF[B](ifTrue: => B, ifFalse: => B)(implicit F: Functor[F]): F[B] = F.ifF(fa)(ifTrue, ifFalse)
 }
--- a/tests/src/test/scala/cats/tests/SyntaxSuite.scala
+++ b/tests/src/test/scala/cats/tests/SyntaxSuite.scala
@@ -284,6 +284,9 @@ object SyntaxSuite {
 
     val b = mock[B]
     val fb1: F[B] = fa.as(b)
+
+    val c = mock[F[Boolean]]
+    c.ifF(1, 0)
   }
 
   def testApply[F[_]: Apply: Semigroupal,


core/src/main/scala/cats/syntax/functor.scala
```diff
@@ -62,3 +65,20 @@ final class FunctorTuple2Ops[F[_], A, B](private val fab: F[(A, B)]) extends Any
...
+
+final class IfFOps[F[_]](private val fa: F[Boolean]) extends AnyVal {
+
+  /**
+   * Lifts `if` to Functor
+   *
+   * Example:
+   * {{{
+   * scala> import cats.Functor
+   * scala> import cats.implicits.catsStdInstancesForList
```
armanbilge, line 77: Side note, I think we may not need the instances import anymore in modern Cats. But we can leave modernization of scaladocs to another issue/PR :)

armanbilge (dismissed):
👍 Looks good, nice work! :)

core/src/main/scala/cats/syntax/functor.scala
satorg: If I understand it correctly, it should be possible simply to remove `import cats.implicits.catsStdInstancesForList`.  `import cats.implicits._` is obsolete starting from v2.2.0. IMO I'd be better not to encourage newcomers to make use of old-fashioned syntax in any way.

core/src/main/scala/cats/syntax/functor.scala
```diff
@@ -62,3 +65,20 @@ final class FunctorTuple2Ops[F[_], A, B](private val fab: F[(A, B)]) extends Any
...
+   *
+   * scala> Functor[List].ifF(List(true, false, false))(1, 0)
```
satorg, line 79: It is a doc for `Functor`, not for the syntax. The syntax usage should be documented something like:
```suggestion
   * scala> import cats.syntax.all._
   *
   * scala> List(true, false, false).ifF(1, 0)
```

satorg:
I think it may make sense to add a (very simple) test for this change here:
[SyntaxSuite.scala](https://github.com/typelevel/cats/blob/main/tests/src/test/scala/cats/tests/SyntaxSuite.scala#L277)

satorg:
(❗) Before we merge this PR please consider the following:

Perhaps we don't have to do it in this way. This job can be performed by Simulacrum automatically.
Currently `Functor#iff` is marked as `@noop` to make Simulacrum skipping this method when generating the syntax.
If we remove that `@noop` and run Simulacrum against `Functor.scala` we should get a valid syntax for that method.

Note: the method was marked with `@noop` in this PR: #3424
Perhaps, it was made intentionally.

armanbilge:
> Perhaps we don't have to do it in this way. This job can be performed by Simulacrum automatically.

Last I heard in https://github.com/typelevel/cats/issues/4059#issuecomment-983753124, Simulacrum isn't being run regularly anymore. I certainly don't know how to do it 😆

satorg:
>  I certainly don't know how to do it 😆

It should be as simple as
```
sbt:cats> coreJVM/scalafix TypeClassSupport --files ./core/src/main/scala/cats/Functor.scala
```
😁

But yeah – it is not run regularly. I think it could be one of the checks for the build – run scalafix against all the files and make sure there're no changes amidst them.

atais:
> (❗) Before we merge this PR please consider the following:
> 
> Perhaps we don't have to do it in this way. This job can be performed by Simulacrum automatically. Currently `Functor#iff` is marked as `@noop` to make Simulacrum skipping this method when generating the syntax. If we remove that `@noop` and run Simulacrum against `Functor.scala` we should get a valid syntax for that method.
> 
> Note: the method was marked with `@noop` in this PR: #3424 Perhaps, it was made intentionally.

it does not work with `Simulacrum`

it generates in `Functor.scala`

```
  trait Ops[F[_], A] extends Serializable {
    ...
    def ifF[B](ifTrue: => B, ifFalse: => B): F[B] = typeClassInstance.ifF[B](self)(ifTrue, ifFalse)
  }
```

but it does not compile, with an error:

```
[error] /Users/atais/Documents/cats/core/src/main/scala/cats/Functor.scala:239:78: type mismatch;
[error]  found   : F[A]
[error]  required: F[Boolean]
[error]     def ifF[B](ifTrue: => B, ifFalse: => B): F[B] = typeClassInstance.ifF[B](self)(ifTrue, ifFalse)
[error]                                                                              ^
[error] one error found
[error] one error found
```

So probably this was the reason behind it :)

armanbilge (approved):

tests/src/test/scala/cats/tests/SyntaxSuite.scala
```diff
@@ -284,6 +284,9 @@ object SyntaxSuite {
 
     val b = mock[B]
     val fb1: F[B] = fa.as(b)
+
+    val c = mock[F[Boolean]]
+    c.ifF(1,0)
```
armanbilge, line 289: I bet this is what the formater wants!

```suggestion
    c.ifF(1, 0)
```

atais: wow thanks

satorg:
> but it does not compile, with an error:
> So probably this was the reason behind #3424

Ah, indeed: `ifF` requires `F[Boolean]` not just `F[A]`. Sorry, I missed it. But thank you that you gave it a try!

core/src/main/scala/cats/syntax/functor.scala
```diff
@@ -4,6 +4,9 @@ package syntax
 trait FunctorSyntax extends Functor.ToFunctorOps {
   implicit final def catsSyntaxFunctorTuple2Ops[F[_], A, B](fab: F[(A, B)]): FunctorTuple2Ops[F, A, B] =
     new FunctorTuple2Ops[F, A, B](fab)
+
+  implicit final def catsSyntaxIfF[F[_]: Functor](fa: F[Boolean]): IfFOps[F] =
+    new IfFOps[F](fa)
```
satorg, line 9: Remove `: Functor` constraint from this method please: it is passed to the `ifF` method itself, not the `IfOps` class constructor. Therefore for this method it is completely redundant.
```suggestion
  implicit final def catsSyntaxIfF[F[_]](fa: F[Boolean]): IfFOps[F] =
    new IfFOps[F](fa)
```

satorg: I'm thinking... Wouldn't it be better to rename `IfFOps` and `catsSyntaxIfF` into something like `FunctorBooleanOps` and `catsSyntaxFunctorBooleanOps`? Seems it would closer follow the naming convention used in Cats (see the above `FunctorTuple2Ops` as an example).

@armanbilge @atais wdyt?

armanbilge: I wondered the same thing, but this follows the naming convention used by `ifM`.

satorg: Whatever... I'm ok with both options :)

satorg: I would say `FunctorBooleanOps` would look more typical to me since potentially the class can cover all syntax for `F[Boolean]`, not only `ifF` method. But on the other hand, there's no other method in `Functor` that requires `F[Boolean]` besides `ifF`. And I wonder if it is even possible to come up with another one. So at this moment both names `IfFOps` and `FunctorBooleanOps` look pretty legit.

armanbilge: Actually, I'm not sure we want to make this change. Esp. if `ifM` is implemented in the same way. The constraint may help with implicit search for syntax.

satorg: Yeah, now I am not sure either. It seems there's a lack of consistency among such methods is Cats in general, unfortunately. Some of the methods do require an typeclass constraint (even though it is not used) but some others do not:

* `MonadSyntax#catsSyntaxMonad` do not require `Monad`
* `AlternativeSyntax` both methods do not require `Alternative`
* `ApplicativeSyntax` do not require BUT
* `ApplicativeErrorSyntax#catsSyntaxApplicativeError` DO require `ApplicativeError` along with
* `FlatMapSyntax` and `FoldableSyntax` methods which also DO require their corresponding constraints.
* and so on – I haven't checked everything )

I guess it depends mostly on a contributor's personal preference rather than on a specific convention.

armanbilge (approved):
This became a rather complicated PR for such a simple change, sorry about that! But thanks for it :)

satorg (approved):

core/src/main/scala/cats/syntax/functor.scala
satorg: I raised a question regarding the convention [here](https://discord.com/channels/632277896739946517/633329569402978313/951211557231140894).

atais: soo... do guys want to do anything about this issue in this PR?
if not, btw, I think I cannot merge myself.

atais:
@armanbilge @satorg I cannot merge myself, but it seems good to go.

core/src/main/scala/cats/syntax/functor.scala
satorg: Well, as @Daenyth suggested in the Discord channel, the difference will affect compiler error messages. Check this out:
```scala
import cats.syntax.all._

case class Foo[A](a: A)

Foo(true).ifM(1, 2)
```
This snippet won't compile since there's no `Functor[Foo]` defined apparently. But if we leave the implicit method with the `Functor` constraint:
```
implicit final def catsSyntaxIfF[F[_]: Functor](fa: F[Boolean]): IfFOps[F] = ???
```
then the error message will be the following:
```
Foo(true).ifF(1, 2)
          ^
error: value ifF is not a member of Foo[Boolean]
```
but if we remove the constraint in the way I suggested initially:
```
implicit final def catsSyntaxIfF[F[_]](fa: F[Boolean]): IfFOps[F] = ???
```
then the error will look like
```
Foo(true).ifF(1, 2)
             ^
error: could not find implicit value for parameter F: cats.Functor[Foo]
```
Therefore we can justify whether to leave or remove the constraint basing on which error message looks clearer and easier to comprehend:)

satorg: @atais @armanbilge what would you prefer?

My personal preference: I would say the latter looks better for me. But I'm not going to insist on it – feel free to merge if you think otherwise.

atais: I like the second one more too.

Let me fix that.

Sorry i am not the member of the discord channel and in such situation the link does not open the discussion.

What happened to Gitter, is it not good anymore ;)? At least it was indexed in Google :)

atais: squashed my changes and pushed. thanks guys!

satorg: Well, there was some discussion some time ago and a decision to step back from Gitter in favor of Discord eventually. I wasn't following all the details but yeah Discord is a better option for now.

satorg (approved):
Cannot find a subject for nitpicking anymore )

armanbilge (approved):

danicheg (commented):
A PR to `cats` containing new syntax and here is no one mention about `mouse` within 30 comments?! 🤯  Let's fix it. FWIW, https://github.com/typelevel/mouse is a dedicated project with a bunch of syntaxes. You might want to take a look at it. Tho, I don't mind against this new syntax within cats.

PR #4152, opened 2022-03-27
Alleycats: upgrade Iterable instance from Foldable to Traverse, add TraverseFilter

bplommer:


alleycats-core/src/main/scala/alleycats/std/iterable.scala
```diff
@@ -1,13 +1,18 @@
...
 
 object iterable extends IterableInstances
 
 trait IterableInstances {
-  implicit val alleycatsStdIterableFoldable: Foldable[Iterable] =
-    new Foldable[Iterable] {
+  @deprecated("use alleycatsStdIterableTraverse", "2.7.1")
+  val alleycatsStdIterableFoldable: Foldable[Iterable] = alleycatsStdIterableTraverse
+
+  implicit val alleycatsStdIterableTraverse: Traverse[Iterable] =
```
armanbilge, line 14: Annoying, but this may need to be a `def` for bincompat.

```suggestion
  implicit def alleycatsStdIterableTraverse: Traverse[Iterable] =
```

bplommer: Oh I missed this - will that allow avoiding a BinCompat trait?

armanbilge: Yup, I think so. TIL about the BinCompat trait trick though :)

bplommer: I think your way is cleaner 😀

johnynek (approved):

rossabaker (approved):

PR #4156, opened 2022-04-02
Add missing Chain#distinctBy method

satorg:
Core:
* Adds the missing `Chain#distinctBy` method.
* Slightly optimizes its implementation (replaces two accesses to `TreeSet` via `contains` followed by `+=` with just one access via `add`).

Tests:
* Introduces `trait cats.tests.compat.ScalaVersionSpecificSyntax` that implements an extension method `distinctBy` for any `Seq`-derived collection in Scala 2.12 (and does nothing for other Scala versions).
* Adds a test for `Chain#distinctBy` to `ChainSuite`.

### DIFF
--- a/core/src/main/scala/cats/data/Chain.scala
+++ b/core/src/main/scala/cats/data/Chain.scala
@@ -27,8 +27,8 @@ import cats.kernel.instances.StaticMethods
 
 import scala.annotation.tailrec
 import scala.collection.immutable.SortedMap
-import scala.collection.immutable.TreeSet
 import scala.collection.immutable.{IndexedSeq => ImIndexedSeq}
+import scala.collection.mutable
 import scala.collection.mutable.ListBuffer
 
 import Chain.{
@@ -150,6 +150,10 @@ sealed abstract class Chain[+A] extends ChainCompat[A] {
    * Returns false if there are no elements in this collection.
    */
   final def nonEmpty: Boolean = !isEmpty
+
+  // Quick check whether the chain is either empty or contains one element only.
+  @inline private def isEmptyOrSingleton: Boolean =
+    isEmpty || this.isInstanceOf[Chain.Singleton[_]]
 
   /**
    * Concatenates this with `c` in O(1) runtime.
@@ -760,19 +764,59 @@ sealed abstract class Chain[+A] extends ChainCompat[A] {
 
   /**
    * Remove duplicates. Duplicates are checked using `Order[_]` instance.
+   * 
+   * Example:
+   * {{{
+   * scala> import cats.data.Chain
+   * scala> val chain = Chain(1, 2, 2, 3)
+   * scala> chain.distinct
+   * res0: cats.data.Chain[Int] = Chain(1, 2, 3)
+   * }}}
    */
   def distinct[AA >: A](implicit O: Order[AA]): Chain[AA] = {
-    implicit val ord: Ordering[AA] = O.toOrdering
+    if (isEmptyOrSingleton) this
+    else {
+      implicit val ord: Ordering[AA] = O.toOrdering
 
-    var alreadyIn = TreeSet.empty[AA]
-
-    foldLeft(Chain.empty[AA]) { (elementsSoFar, b) =>
-      if (alreadyIn.contains(b)) {
-        elementsSoFar
-      } else {
-        alreadyIn += b
-        elementsSoFar :+ b
+      val bldr = Vector.newBuilder[AA]
+      val seen = mutable.TreeSet.empty[AA]
+      val it = iterator
+      while (it.hasNext) {
+        val next = it.next()
+        if (seen.add(next))
+          bldr += next
       }
+      // Result can contain a single element only.
+      Chain.fromSeq(bldr.result())
+    }
+  }
+
+  /**
+   * Remove duplicates by a predicate. Duplicates are checked using `Order[_]` instance.
+   * 
+   * Example:
+   * {{{
+   * scala> import cats.data.Chain
+   * scala> val chain = Chain(1, 2, 3, 4)
+   * scala> chain.distinctBy(_ / 2)
+   * res0: cats.data.Chain[Int] = Chain(1, 2, 4)
+   * }}}
+   */
+  def distinctBy[B](f: A => B)(implicit O: Order[B]): Chain[A] = {
+    if (isEmptyOrSingleton) this
+    else {
+      implicit val ord: Ordering[B] = O.toOrdering
+
+      val bldr = Vector.newBuilder[A]
+      val seen = mutable.TreeSet.empty[B]
+      val it = iterator
+      while (it.hasNext) {
+        val next = it.next()
+        if (seen.add(f(next)))
+          bldr += next
+      }
+      // Result can contain a single element only.
+      Chain.fromSeq(bldr.result())
     }
   }
 
--- a/tests/shared/src/test/scala-2.12/cats/tests/compat/ScalaVersionSpecificSyntax.scala
+++ b/tests/shared/src/test/scala-2.12/cats/tests/compat/ScalaVersionSpecificSyntax.scala
@@ -1,0 +1,28 @@
+/*
+ * Copyright (c) 2015 Typelevel
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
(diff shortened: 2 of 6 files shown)


satorg:
Note: this `distinctBy` can be further optimized by returning `this` in cases when `Chain` is empty or contains a single element only. Apparently such a collection cannot contain duplicates. Something like this:
```
  def distinctBy[B](f: A => B)(implicit O: Order[B]): Chain[A] = {
    // Cannot just use `size` (or `length`) method here since currently they enumerate the entire collection.
    if (isEmpty || this.isInstanceOf[Chain.Singleton[_]]) this else {
      implicit val ord: Ordering[B] = O.toOrdering

      val alreadyIn = mutable.TreeSet.empty[B]
      // All the other code is the same
    }
  }
```
Not sure if it's really worth it though. Wdyt?

core/src/main/scala/cats/data/Chain.scala
```diff
@@ -617,18 +620,38 @@ sealed abstract class Chain[+A] {
...
+   */
+  def distinctBy[B](f: A => B)(implicit O: Order[B]): Chain[A] = {
+    implicit val ord: Ordering[B] = O.toOrdering
 
-    foldLeft(Chain.empty[AA]) { (elementsSoFar, b) =>
-      if (alreadyIn.contains(b)) {
-        elementsSoFar
+    val alreadyIn = mutable.TreeSet.empty[B]
+
+    foldLeft(Chain.empty[A]) { (elementsSoFar, a) =>
```
johnynek, line 649: If we are using a mutable set here why not also use a var to the Chain and a while loop to avoid the closure?

core/src/main/scala/cats/data/Chain.scala
```diff
@@ -617,18 +620,38 @@ sealed abstract class Chain[+A] {
...
+   * scala> import cats.data.Chain
+   * scala> val chain = Chain(1, 2, 2, 3)
+   * scala> chain.distinct
+   * res0: cats.data.Chain[Int] = Chain(1, 2, 3)
+   * }}}
    */
-  def distinct[AA >: A](implicit O: Order[AA]): Chain[AA] = {
-    implicit val ord: Ordering[AA] = O.toOrdering
+  def distinct[AA >: A](implicit O: Order[AA]): Chain[AA] =
+    distinctBy(identity[AA])
```
johnynek, line 632: I think it may be worth it to duplicate the code to avoid pushing everything through an extra function call.

satorg: Well, this is the code clarity VS performance dichotomy :)
Seems that Scala Library folks decided not to sweat about it too much – they simply call `distinctBy` from `distinct`.
https://github.com/scala/scala/blob/v2.13.8/src/library/scala/collection/Seq.scala#L206

But we can do the opposite, it you think it's worth it.

core/src/main/scala/cats/data/Chain.scala
satorg: Yes agree. Fixed.

core/src/main/scala/cats/data/Chain.scala
```diff
@@ -120,13 +123,19 @@ sealed abstract class Chain[+A] {
   /**
    * Returns true if there are no elements in this collection.
    */
-  def isEmpty: Boolean = !this.isInstanceOf[Chain.NonEmpty[_]]
+  def isEmpty: Boolean = this eq Chain.Empty
```
satorg, line 126: I have got a feeling that `eq` should work faster than `isInstanceOf`...

satorg: @johnynek I realized that you defined this method via `isInstanceOf` in #3533.
Not I am in doubt... Is `isIntsanceOf` more performant than just `eq`, could you clarify please?

satorg: Oh.. Seems I've caught it (perhaps). I've checked the bytecode and realized that the
`!this.isInstanceOf[Chain.NonEmpty[_]]` trick lets to avoid a `getstatic` instruction compared to `eq` which still requires it. 

Is that what you meant here?

core/src/main/scala/cats/data/Chain.scala
```diff
@@ -617,19 +626,57 @@ sealed abstract class Chain[+A] {
...
+    else {
+      implicit val ord: Ordering[AA] = O.toOrdering
+
+      var result = Chain.empty[A]
+      val seen = mutable.TreeSet.empty[AA]
+      val it = iterator
+      while (it.hasNext) {
+        val next = it.next()
+        if (seen.add(next))
+          result ++= one(next)
```
johnynek, line 649: why `result ++= one(next)` vs `result = result :+ next`

Why not use the method we already have?

core/src/main/scala/cats/data/Chain.scala
```diff
@@ -617,19 +626,57 @@ sealed abstract class Chain[+A] {
...
+    else {
+      implicit val ord: Ordering[B] = O.toOrdering
+
+      var result = Chain.empty[A]
+      val seen = mutable.TreeSet.empty[B]
+      val it = iterator
+      while (it.hasNext) {
+        val next = it.next()
+        if (seen.add(f(next)))
+          result ++= one(next)
```
johnynek, line 677: same question about `result = result :+ next`

core/src/main/scala/cats/data/Chain.scala
satorg: Fixed.

core/src/main/scala/cats/data/Chain.scala
satorg: Rolled it back.

core/src/main/scala/cats/data/Chain.scala
satorg: Well, `chain :+ a` does the same thing as `chain ++ one(a)`.
I.e. we cannot avoid creating a singleton here anyway.
But .. I agree – let's switch to more explicit method `:+` here.

satorg: fixed

core/src/main/scala/cats/data/Chain.scala
satorg: fixed

johnynek (approved):
Thanks!

core/src/main/scala/cats/data/Chain.scala
```diff
@@ -760,19 +764,57 @@ sealed abstract class Chain[+A] extends ChainCompat[A] {
...
+   */
+  def distinctBy[B](f: A => B)(implicit O: Order[B]): Chain[A] = {
+    if (isEmptyOrSingleton) this
+    else {
+      implicit val ord: Ordering[B] = O.toOrdering
+
+      var result = Chain.empty[A]
+      val seen = mutable.TreeSet.empty[B]
+      val it = iterator
+      while (it.hasNext) {
```
johnynek, line 812: one issue: this is going to create a `Chain` the costs `O(N)` to uncons.

I wonder if the better approach would build a Vector and `Wrap` that. Vector seems to be the default collection we use in other places

e.g.
```scala
val bldr = Vector.newBuilder[A]
...

Wrap(bldr.result())
```

satorg: Yeah, I wonder too. I didn't really think about it, perhaps you're right. Let's try this option.

satorg: fixed

core/src/main/scala/cats/data/Chain.scala
```diff
@@ -760,19 +764,59 @@ sealed abstract class Chain[+A] extends ChainCompat[A] {
...
+      val bldr = Vector.newBuilder[AA]
+      val seen = mutable.TreeSet.empty[AA]
+      val it = iterator
+      while (it.hasNext) {
+        val next = it.next()
+        if (seen.add(next))
+          bldr += next
+      }
+      // Result can contain a single element only.
+      Chain.fromSeq(bldr.result())
```
satorg, line 790: Potentially this line (and the similar one below) could be expanded into something more efficient like
```scala
bldr.result() match {
  case Vector(a) => Singleton(a)
  case seq       => Wrap(seq)
}
```
but... I'm not really sure it is worth it – I mean, it is unlikely that such a single call can lead to substantial (or even noticeable) performance decrease, IMO.

johnynek (approved):
This looks great to me!

PR #4158, opened 2022-04-03
Add `INothing` and `Pure` types

bplommer:
Prompted by discussion on https://github.com/http4s/http4s/pull/5712#discussion_r841249712. Because these types are defined as inequalities, separate identical definitions of them aren't unified by the compiler - therefore, http4s needs to refer to fs2's definitions of them in order to interoperate with fs2, even though they're not intrinsically specific to fs2. Moving the definition up to cats would avoid this.

### DIFF
--- a/core/src/main/scala/cats/package.scala
+++ b/core/src/main/scala/cats/package.scala
@@ -81,7 +81,10 @@ package object cats {
    *
    * {{{
    * scala> type Alg[F[_], A] = Either[F[A], A]
+   * scala> def extract[A](p: Alg[Pure, A]): A = p.merge
    * scala> val pureAlg: Alg[Pure, Int] = Right(3)
+   * scala> extract(pureAlg)
+   * res0: Int = 3
    * scala> val optAlg: Alg[Option, Int] = pureAlg
    * scala> val listAlg: Alg[List, Int] = pureAlg
    * }}}
--- a/core/src/main/scala/cats/package.scala
+++ b/core/src/main/scala/cats/package.scala
@@ -180,11 +180,3 @@ package object cats {
     def apply[F[_]](implicit ev: MonadThrow[F]): MonadThrow[F] = ev
   }
 }
-
-object Test {
-  type PureOrF[F[_], A] = Either[F[A], A]
-
-  val v: PureOrF[Pure, Int] = Right(10)
-  val v1: PureOrF[List, Int] = v
-
-}
--- a/core/src/main/scala/cats/package.scala
+++ b/core/src/main/scala/cats/package.scala
@@ -1,5 +1,3 @@
-import cats.Pure
-
 import scala.annotation.tailrec
 
 /**
--- a/core/src/main/scala/cats/package.scala
+++ b/core/src/main/scala/cats/package.scala
@@ -68,6 +68,16 @@ package object cats {
     def apply[A](a: A): Id[A] = a
   }
 
+  /**
+   * Like `Nothing` but works better with type inference.
+   */
+  type INothing <: Nothing
+
+  /**
+   * Like `INothing` but as a type constructor.
+   */
+  type Pure[A] <: Nothing
+
   type Endo[A] = A => A
 
   val catsInstancesForId: Bimonad[Id] with CommutativeMonad[Id] with NonEmptyTraverse[Id] with Distributive[Id] =
--- a/core/src/main/scala/cats/package.scala
+++ b/core/src/main/scala/cats/package.scala
@@ -1,3 +1,5 @@
+import cats.Pure
+
 import scala.annotation.tailrec
 
 /**
@@ -74,7 +76,17 @@ package object cats {
   type INothing <: Nothing
 
   /**
-   * Like `INothing` but as a type constructor.
+   * Like `INothing` but as a type constructor. When used as a parameter to a coproduct type
+   * with effectful and non-effectful cases (such as [[https://github.com/typelevel/fs2 fs2 Stream]],
+   * where it originates), prevents the effectful cases from being inhabited and allows covariance up to
+   * arbitrary effect types.
+   *
+   * {{{
+   * scala> type Alg[F[_], A] = Either[F[A], A]
+   * scala> val pureAlg: Alg[Pure, Int] = Right(3)
+   * scala> val optAlg: Alg[Option, Int] = pureAlg
+   * scala> val listAlg: Alg[List, Int] = pureAlg
+   * }}}
    */
   type Pure[A] <: Nothing
 
@@ -167,4 +179,12 @@ package object cats {
   object MonadThrow {
     def apply[F[_]](implicit ev: MonadThrow[F]): MonadThrow[F] = ev
   }
+}
+
+object Test {
+  type PureOrF[F[_], A] = Either[F[A], A]
+
+  val v: PureOrF[Pure, Int] = Right(10)
+  val v1: PureOrF[List, Int] = v
+
 }


armanbilge (dismissed):
Names subject to bikeshed, but I think these fit in well with the aliases `Id`, `Endo`, etc.

For this to be effective, fs2 will have to alias its types to these new ones.

bplommer:
> For this to be effective, fs2 will have to alias its types to these new ones.

Good point - I've opened an fs2 issue at https://github.com/typelevel/fs2/issues/2868

core/src/main/scala/cats/package.scala
```diff
@@ -68,6 +68,16 @@ package object cats {
...
 
+  /**
+   * Like `Nothing` but works better with type inference.
+   */
+  type INothing <: Nothing
+
+  /**
+   * Like `INothing` but as a type constructor.
+   */
+  type Pure[A] <: Nothing
```
johnynek, line 79: can we add some doctest example here? I think we can assume people know what Nothing is useful for, but I don't know of uses for `Pure[A]` looking at this.

SystemFw:
For what is worth, it's not all all clear to me that these belong in cats.

>  - therefore, http4s needs to refer to fs2's definitions of them in order to interoperate with fs2

http4s refers directly to `fs2.Stream` all over the place, so I don't see what the problem is with referring to `Pure` and `INothing` as well

> even though they're not intrinsically specific to fs2

are they not though? The reason they exist and the reason they look the way they do is closely tied to fs2's compilation machinery, shouldn't the motivation for moving them further downstream than fs2 include other use cases that aren't just fs2 itself?

armanbilge:
> The reason they exist and the reason they look the way they do is closely tied to fs2's compilation machinery, shouldn't the motivation for moving them further downstream than fs2 include other use cases that aren't just fs2 itself?

I think that's the situation here :) in http4s we now have `Request`, `Response`, `Entity`, etc. that are covariant in effect type. AFAICT the _covariance_ is what makes `Pure` interesting and useful. So that we can have things like this:

```scala
final case class Strict(chunk: Chunk[Byte]) extends Entity[Pure]
```
https://github.com/http4s/http4s/blob/17c1653db2568325d2db41dfff773ad4867f8aba/core/shared/src/main/scala/org/http4s/Entity.scala#L59

Besides the `Chunk` sitting in the middle of that :) it's not immediately obvious to me how this directly ties into fs2's compilation machinery.

Of course, I can't deny that fs2 is a big part of http4s and It may turn out that the only non-`Pure` entity that's useful in practice is based on `Stream` and thus that this is really all about fs2 in the end.

Maybe the thing I'm missing is: is covariance of effects truly only relevant in the context of `Stream`?

core/src/main/scala/cats/package.scala
bplommer: I've added a doctest example - it's fairly trivial, but hopefully goes some way to address @SystemFw's concern that `Pure` is in principle specific to fs2.

bplommer:
> http4s refers directly to `fs2.Stream` all over the place, so I don't see what the problem is with referring to `Pure` and `INothing` as well

`fs2.Stream` is a runtime dependency though - when http4s refers to `fs2.Stream` that's because it's working with `Stream` instances provided by fs2. When it refers to `fs2.INothing` or `fs2.Pure` it clearly isn't working with `INothing` or `Pure` instances provided by fs2. That seems a significant difference.

I do also remember you mentioning, when we discussed exposing `fs2.Compiler.Target` in the http4s api, that seeing fs2 types makes users nervous.
 
> I think that's the situation here :) in http4s we now have `Request`, `Response`, `Entity`, etc. that are covariant in effect type. AFAICT the _covariance_ is what makes `Pure` interesting and useful.

Agreed. The motivation for exposing them in the http4s api is completely independent from (though obviously inspired by) the fact that http4s is built on top if fs2. If http4s weren't built on top of fs2 we would still want to define those types, for reasons parallel to fs2's reasons. But in that hypothetical situation, just to smoothly interoperate with fs2 we would still need to define the types as aliases to fs2's ones (and therefore force an fs2 dependency at compile time) - all as an artifact of a trick for getting the compiler to play nicely with variance.

SystemFw:
>  in http4s we now have Request, Response, Entity, etc. that are covariant in effect type. [..]Besides the Chunk sitting in the middle of that :) it's not immediately obvious to me how this directly ties into fs2's compilation machinery.

If you replace that `Chunk` with the isomorphic `Stream[Pure, A]`, I think all we have here are newtypes over Stream

> that the only non-Pure entity that's useful in practice is based on Stream and thus that this is really all about fs2 in the end.
> The motivation for exposing them in the http4s api is completely independent from (though obviously inspired by) the fact that http4s is built on top of fs2. If http4s weren't built on top of fs2 we would still want to define those types, for reasons parallel to fs2's reasons

Tbh that does change my mind a bit, although I don't find it a realistic scenario in pratice, i.e. it would only happen if the streaming versions of http entities switched to a streaming abstraction which wasn't Stream.

 >  do also remember you mentioning, when we discussed exposing fs2.Compiler.Target in the http4s api, that seeing fs2 types makes users nervous.

that's fair, although the main concern around `Target` wasn't the `fs2.` prefix per se (again, they see `Stream` all over the place), but the complexity of `Target` which is aimed at solving a problem they didn't have (the discussion was in the context of Encoders/Decoders)

> Maybe the thing I'm missing is: is covariance of effects truly only relevant in the context of Stream?
> o address @SystemFw's concern that Pure is in principle specific to fs2.

That's not quite what I think, it's more that the specific way these are shaped is a direct result of constraints in the Stream api and encoding, and have changed in the past. What if `Stream` changes its encoding of effect covariance again, do we change cats?

core/src/main/scala/cats/package.scala
SystemFw: I'm not sure it does, since this compiles:
```scala
  type Alg[F[_], A] = Either[F[A], A]
  val pureAlg: Alg[Nothing, Int] = Right(3)
  val optAlg: Alg[Option, Int] = pureAlg
  val listAlg: Alg[List, Int] = pureAlg
```
so why is it named `Pure`? why is it an inequality with `Nothing`? If all you need is covariance (also address @armanbilge point about covariance being the important aspect here), `Nothing` is enough, the specific shape we came to with `Pure` is dictated by other constraints in fs2's api specifically

bplommer: Yeah, this occurred to me afterwards. I'm trying to come up with a reasonably generic example that won't compile with `Nothing` but so far without success - this inclines me towards agreeing with you, but also makes me more sad that we have to expose `fs2.Pure` rather than `Nothing` in the http4s api.

mpilquist:
> This was a slight overstatement - if fs2 weren't in the mix, we could get by with plain Nothing. That would still break interop with fs2 though, so the remainder of the point stands.

Can you provide an example of this?

core/src/main/scala/cats/package.scala
armanbilge: Yes, thanks, this answered a lot of my questions/confusion. Now I see what you are getting at when you say
> the specific shape we came to with `Pure` is dictated by other constraints in fs2's api specifically

armanbilge:
Thanks all for the discussion 🙏 I didn't realize how capable `Nothing` was of its own accord so I see now how `Pure` is specifically tied to fs2. Meanwhile, @bplommer has been able to put that to effect in https://github.com/http4s/http4s/pull/5712. So it's no longer clear to me we need this in Cats.

mpilquist:
FYI: https://github.com/typelevel/fs2/pull/2870

armanbilge:
And also:
- https://github.com/typelevel/fs2/pull/2876

Seems to me these types are on their way out.

bplommer:
Agreed - I was just summoning up the courage to admit the problem with unification of identical definitions seems to be entirely a figment of my imagination before closing this 😅

PR #4159, opened 2022-04-03
Add Chain#knownSize and Chain#lengthCompare (sizeCompare) methods

satorg:
Adds quite useful `knownSize`, `lengthCompare` (and `sizeCompare` as an alias) which can be found in Scala Library.

### DIFF
--- a/core/src/main/scala/cats/data/Chain.scala
+++ b/core/src/main/scala/cats/data/Chain.scala
@@ -586,6 +586,51 @@ sealed abstract class Chain[+A] {
     }
 
   /**
+   * Compares the length of this chain to a test value.
+   * 
+   * The method does not call `length` directly; its running time
+   * is `O(length min len)` instead of `O(length)`.
+   * 
+   * @param  len the test value that gets compared with the length.
+   * @return a negative value if `this.length < len`,
+   *         zero if `this.length == len` or
+   *         a positive value if `this.length > len`.
+   * @note   This is an adapted version of
+             [[https://github.com/scala/scala/blob/v2.13.8/src/library/scala/collection/Iterable.scala#L272-L288 Iterable#sizeCompare]]
+             from Scala Library v2.13.8
+   * 
+   * {{{
+   * scala> import cats.data.Chain
+   * scala> val chain = Chain(1, 2, 3)
+   * scala> val isLessThan4 = chain.lengthCompare(4) < 0
+   * scala> val isEqualTo3 = chain.lengthCompare(3) == 0
+   * scala> val isGreaterThan2 = chain.lengthCompare(2) > 0
+   * scala> isLessThan4 && isEqualTo3 && isGreaterThan2
+   * res0: Boolean = true
+   * }}}
+   */
+  final def lengthCompare(len: Long): Int =
+    if (len < 0) 1
+    else {
+      var sz = knownSize
+      if (sz < 0) {
+        sz = 0L
+        val it = iterator
+        while (it.hasNext) {
+          if (sz == len) return 1
+          it.next()
+          sz += 1L
+        }
+      }
+      sz.compareTo(len)
+    }
+
+  /**
+   * Alias for lengthCompare
+   */
+  final def sizeCompare(size: Long): Int = lengthCompare(size)
+
+  /**
    * Converts to a list.
    */
   final def toList: List[A] =
--- a/core/src/main/scala/cats/data/Chain.scala
+++ b/core/src/main/scala/cats/data/Chain.scala
@@ -1,6 +1,15 @@
 package cats
 package data
 
+import cats.kernel.compat.scalaVersionSpecific._
+import cats.kernel.instances.StaticMethods
+
+import scala.annotation.tailrec
+import scala.collection.immutable.SortedMap
+import scala.collection.immutable.TreeSet
+import scala.collection.immutable.{IndexedSeq => ImIndexedSeq}
+import scala.collection.mutable.ListBuffer
+
 import Chain.{
   empty,
   fromSeq,
@@ -17,12 +26,6 @@ import Chain.{
   Singleton,
   Wrap
 }
-import cats.kernel.instances.StaticMethods
-import cats.kernel.compat.scalaVersionSpecific._
-
-import scala.annotation.tailrec
-import scala.collection.immutable.{IndexedSeq => ImIndexedSeq, SortedMap, TreeSet}
-import scala.collection.mutable.ListBuffer
 
 /**
  * Trivial catenable sequence. Supports O(1) append, and (amortized)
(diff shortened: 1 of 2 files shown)


tests/src/test/scala/cats/tests/ChainSuite.scala
```diff
@@ -249,6 +294,8 @@ class ChainSuite extends CatsSuite {
 
   test("a.isEmpty == (a eq Chain.nil)") {
     assert(Chain.fromSeq(Nil) eq Chain.nil)
+    assert(Chain.fromOption(None) eq Chain.nil)
+    assert(Chain.fromIterableOnce(Iterator.empty) eq Chain.nil)
```
satorg, line 298: I think these are quite important invariants.

satorg:
Potentially #4156 can benefit from this one.

tests/src/test/scala/cats/tests/ChainSuite.scala
```diff
@@ -119,6 +119,32 @@ class ChainSuite extends CatsSuite {
...
+        case size @ (0L | 1L) => size
+        case _                => -1L
+      }
+
+      assertEquals(cu.knownSize, expected)
+    }
+  }
+
+  test("lengthCompare and sizeCompare should be consistent with length and size") {
+    forAll { (cu: Chain[Unit], diff: Byte) =>
```
johnynek, line 134: I think a more direct law is:
```scala
signum(cu.lengthCompare(x)) == signum((cu.length.compareTo(x)))
```

Can we test that one?

tests/src/test/scala/cats/tests/ChainSuite.scala
```diff
@@ -155,6 +181,34 @@ class ChainSuite extends CatsSuite {
     }
   }
 
+  test("fromOption should be consistent with one") {
+    forAll { (a: Int) =>
```
johnynek, line 185: I don't see how the value of `a` matters here. What about a non-property test where you use `()` as the value.

same comment for below.

tests/src/test/scala/cats/tests/ChainSuite.scala
satorg: fixed

tests/src/test/scala/cats/tests/ChainSuite.scala
satorg: fixed

core/src/main/scala/cats/data/Chain.scala
```diff
@@ -574,6 +574,62 @@ sealed abstract class Chain[+A] {
...
 
+  /**
+   * The number of elements in this chain, if it can be cheaply computed, -1 otherwise.
+   * Cheaply usually means: Not requiring a collection traversal.
+   */
+  final def knownSize: Long =
+    this match {
+      case _ if isEmpty       => 0
+      case Chain.Singleton(_) => 1
+      case _                  => -1
```
johnynek, line 585: could we add `case Chain.Wrap(seq) => seq.knownSize` here?

core/src/main/scala/cats/data/Chain.scala
```diff
@@ -574,6 +574,62 @@ sealed abstract class Chain[+A] {
...
+   * scala> import cats.data.Chain
+   * scala> val chain = Chain(1, 2, 3)
+   * scala> val isLessThan4 = chain.lengthCompare(4) < 0
+   * scala> val isEqualTo3 = chain.lengthCompare(3) == 0
+   * scala> val isGreaterThan2 = chain.lengthCompare(2) > 0
+   * scala> isLessThan4 && isEqualTo3 && isGreaterThan2
+   * res0: Boolean = true
+   * }}}
+   */
+  final def lengthCompare(len: Long): Int =
```
johnynek, line 612: I wonder about an alternate approach:
```scala
// invariant, sz >= 0
final def lengthCompare(len: Int): Int =
  this match {
    case Empty => Integer.compare(0, len)
    case Singleton(_) => Integer.compare(1, len)
    case Wrap(seq) => seq.lengthCompare(len)
    case _ =>
      // length 2 or more... iterate through
}
```

Since this would handle a some common cases without allocation: empty, singletons and chains that are wrapping seqs (which are returned by several operations because they are more efficient).

core/src/main/scala/cats/data/Chain.scala
satorg: Yeah, I tried it in the first place, but turned out that 2.12 does not have `knownSize` implemented. So if we really need it then we'll have to deal with the cross-build compatibility here somehow. Not sure what is the best way to approach that. Any ideas?

core/src/main/scala/cats/data/Chain.scala
satorg: Yes sure – we can add more optimizations here. The only note: `Chain` uses `Long` for length/size measurements. Do you know why, by chance?

core/src/main/scala/cats/data/Chain.scala
satorg: I mean, I can think about two approaches:
1. Implement a compat version of `knownSize` for `Seq` for 2.12.
2. Handle `case Wrap(seq)` in 2.13+ only.

core/src/main/scala/cats/data/Chain.scala
satorg: Added some optimized special cases. I think I may need to add more stress on those cases in tests.

core/src/main/scala/cats/data/Chain.scala
```diff
@@ -574,6 +574,71 @@ sealed abstract class Chain[+A] {
...
+   * scala> isLessThan4 && isEqualTo3 && isGreaterThan2
+   * res0: Boolean = true
+   * }}}
+   */
+  final def lengthCompare(len: Long): Int = {
+    this match {
+      // `isEmpty` check should be faster than `== Chain.Empty`,
+      // but the compiler fails to prove that the match is still exhaustive.
+      case _ if isEmpty       => 0L.compareTo(len)
+      case Chain.Singleton(_) => 1L.compareTo(len)
```
johnynek, line 617: I think this call boxes. I think writing java.lang.Long.compare is safer since we can be 100% sure there is no boxing.

johnynek (approved):
Minor nit about boxing but I think this is in good shape.

Thanks!

core/src/main/scala/cats/data/Chain.scala
satorg: Yes, it does. Really good catch, thank you. Fixed.

satorg:
@johnynek I added even more optimizations to the `lengthCompare` implementation. Could you take a look one more time please?

johnynek (approved):
Thanks for working so much on this!

satorg:
Thank you for your keen eye – it helped me to learn some new stuff, I really appreciate it.

Cool, I'll let this PR to marinade here for a couple of days before merging in case if someone else would like to take a look at it.

Issue #4149, opened 2022-03-19
`coreJVM/packageSrc` fails with "duplicate entry" error


SethTisue:
this came up in the Scala 2 community build

```
[error] java.util.zip.ZipException: duplicate entry: cats/instances/package.scala
[error] 	at java.base/java.util.zip.ZipOutputStream.putNextEntry(ZipOutputStream.java:241)
[error] 	at java.base/java.util.jar.JarOutputStream.putNextEntry(JarOutputStream.java:115)
[error] 	at sbt.io.IO$.addFileEntry$1(IO.scala:709)
[error] 	at sbt.io.IO$.$anonfun$writeZip$6(IO.scala:718)
[error] 	at sbt.io.IO$.$anonfun$writeZip$6$adapted(IO.scala:718)
```

armanbilge:
Woops, thanks for the report. If nobody gets to it I'll fix it while working on https://github.com/typelevel/cats/issues/4143.

SethTisue:
scala/community-build@4dfa437efaae687f985fbb9a6eb02d92e4bfeade 👍

Issue #4143, opened 2022-03-07
Migrate build to sbt-typelevel


armanbilge:
@djspiewak asked me about this on Discord and I've had this on my mind. Cats is the last flagship project without a PR on https://github.com/typelevel/sbt-typelevel/issues/83. Here's the plan that's been brewing.

1. This is going to be a big and careful effort, but my intention is to completely overhaul and modernize the cats build. If there's no objections, I'd like to enable CI snapshots and CI release.

2. I'd like to target this at a series/2.7 branch. Because sbt-typelevel enables MiMa for Scala 3, we will have to deal with https://github.com/typelevel/cats/issues/4062.

    IMHO the best way forward is to release 2.7.1 which patches that one binary-break, and declare 2.7.0 "cursed". Other options I considered:
    - Fix it in 2.8.0 and add a MiMa filter for the method. But this MiMa filter will chase us to the end of series/2.x and we will no longer get MiMa support to catch this mistake, which is bad.
    - Fix it in 2.8.0 and don't check bincompat at all against 2.7.0. But then who knows what other binary issues could fall through the cracks.
    - Fix it in 2.8.0 by splitting/duplicate the entirety of `cats/package.scala` across Scala 2/3 so we can apply new Scala 3-only techniques to fix it while also retaining bincompat to the cursed 2.7.0. As proposed by @smarter in:
      https://github.com/typelevel/cats/pull/4063#issuecomment-984127262

    So would love to get ideas here, but to me a 2.7.1 patch release fixing this one thing is the safest and most convenient way forward.

3. It doesn't have to be/won't be part of the initial migration PR, but I'd like to move the Cats site onto the sbt-typelevel-site plugin which uses [Laika](https://github.com/planet42/laika). Comments:
    - The visible changes will be purely aesthetic. Content, overall organization, and URLs will not change. I know this because I did similar migrations for e.g. the spire website which previously used the same setup the Cats site uses.
    - Laika is pure-JVM, built on Cats+CE+fs2+http4s, with helpful and responsive maintainer @jenshalm. No more fussing with rubies and gems.
    - All the website CI deploy infrastructure is in sbt-typelevel-site. So it will clean up the Cats build considerably.
    - sbt-typelevel-based Laika sites are becoming widely adopted across the Typelevel ecosystem. So we're all in this together: any improvements to the core plugin are immediately rolled out to everyone.
      - https://typelevel.org/sbt-typelevel
      - https://typelevel.org/spire
      - https://http4s.org
[...]

PR #4160, opened 2022-04-04
Migrate to sbt-typelevel

armanbilge:
Closes https://github.com/typelevel/cats/issues/4143. Fixes https://github.com/typelevel/cats/issues/4062. Fixes https://github.com/typelevel/cats/issues/4149.

Things seem to be working locally. But the battle with CI is just beginning.

Note this PR doesn't deal with the documentation/site at all (they are currently not being checked in CI). I'll prepare that in a follow-up for easier review, although they should probably merge together.


project/plugins.sbt
```diff
@@ -1,27 +1,13 @@
+val sbtTypelevelVersion = "0.4.7"
+addSbtPlugin("org.typelevel" % "sbt-typelevel-ci-release" % sbtTypelevelVersion)
+addSbtPlugin("org.typelevel" % "sbt-typelevel-settings" % sbtTypelevelVersion)
+addSbtPlugin("org.typelevel" % "sbt-typelevel-site" % sbtTypelevelVersion)
```
armanbilge, line 4: I forgot to mention: in the end, I didn't actually migrate to sbt-typelevel since that would require adding copyright headers to all the files. Instead we use a combination of sbt-typelevel-ci-release and sbt-typelevel-settings. So the build still has a little more boilerplate than it could, but it's okay.

core/src/main/scala/cats/Invariant.scala
```diff
@@ -105,6 +105,10 @@ object Invariant extends ScalaVersionSpecificInvariantInstances with InvariantIn
   implicit def catsInstancesForId
     : Distributive[Id] with Bimonad[Id] with CommutativeMonad[Id] with NonEmptyTraverse[Id] =
     cats.catsInstancesForId
+  @deprecated("Added for bincompat", "2.8.0")
+  @cats.compat.targetName("catsInstancesForId")
+  def catsInstancesForIdCompat2_6_1: Comonad[Id] =
+    cats.catsInstancesForId
```
armanbilge, line 111: This is the fix for https://github.com/typelevel/cats/issues/4062. Added `@targetName` as a dummy annotation in Scala 2 so we don't have to split this file (I tried compat `trait`s first and it didn't work).

core/src/main/scala/cats/instances/StaticMethods.scala
```diff
@@ -2,7 +2,7 @@ package cats.instances
 
 import scala.collection.mutable.Builder
 
-private[instances] object instances {
+private[instances] object StaticMethods {
```
armanbilge, line 5: Fixes https://github.com/typelevel/cats/issues/4149.

mima.sbt
```diff
@@ -0,0 +1,166 @@
+ThisBuild / mimaBinaryIssueFilters ++= {
```
armanbilge, line 1: I moved these to their own file so they don't clutter up the `build.sbt`.

build.sbt
```diff
@@ -134,101 +66,70 @@ def scalaVersionSpecificFolders(srcName: String, srcBaseDir: java.io.File, scala
...
   Test / unmanagedSourceDirectories ++= scalaVersionSpecificFolders("test", baseDirectory.value, scalaVersion.value),
-  resolvers ++= Seq(Resolver.sonatypeRepo("releases"), Resolver.sonatypeRepo("snapshots")),
-  Test / parallelExecution := false,
-  testFrameworks += new TestFramework("munit.Framework"),
+  Compile / packageSrc / mappings ++= {
+    val base = (Compile / sourceManaged).value
+    (Compile / managedSources).value.map { file =>
+      file -> file.relativeTo(base).get.getPath
+    }
+  },
```
armanbilge, line 83: I'd like to upstream this in https://github.com/typelevel/sbt-typelevel/issues/239.

build.sbt
```diff
@@ -134,101 +66,70 @@ def scalaVersionSpecificFolders(srcName: String, srcBaseDir: java.io.File, scala
...
+  scalacOptions ~= { _.filterNot(x => x.startsWith("-Wunused:")) },
```
armanbilge, line 84: This was the most disappointing thing. Seems that Cats has got lots and lots of unused warnings. So we have to disable them if we want fatal warnings for everything else. It would truly be a great thing if someone could chase down and at least `@nowarn` them if they are legit or necessary for bincompat. Then at least we can avoid adding more unused stuff.

build.sbt
```diff
@@ -804,60 +270,71 @@ lazy val testkit = crossProject(JSPlatform, JVMPlatform, NativePlatform)
...
+    ScalaUnidoc / unidoc / unidocProjectFilter := inProjects(kernel.jvm,
+                                                             core.jvm,
+                                                             free.jvm,
+                                                             algebra.jvm,
+                                                             alleycatsCore.jvm
+    ),
+    scalacOptions ~= { _.filterNot(_.startsWith("-W")) }, // weird nsc bug
+    ScalaUnidoc / unidoc / scalacOptions ++= Seq("-groups", "-diagrams")
+  )
+  .settings(commonSettings)
```
armanbilge, line 326: This introduces a new API-docs only `cats-docs` module that lets us publish unidocs to Sonatype/Maven. Then they can be browsed at javadoc.io.

Example:
https://www.javadoc.io/doc/org.http4s/http4s-docs_2.13/0.23.11/org/http4s/index.html

mima.sbt
```diff
@@ -0,0 +1,181 @@
...
+    Seq(
+      exclude[DirectAbstractMethodProblem]("cats.free.ContravariantCoyoneda.k"),
+      exclude[ReversedAbstractMethodProblem]("cats.free.ContravariantCoyoneda.k"),
+      exclude[DirectAbstractMethodProblem]("cats.free.Coyoneda.k"),
+      exclude[ReversedAbstractMethodProblem]("cats.free.Coyoneda.k")
+    )
+}
+
+// TODO: remove these filters after MiMa 1.1.0 is released with improved support for Scala 3
+ThisBuild / mimaBinaryIssueFilters ++= {
```
armanbilge, line 170: @dwijnand has very kindly fixed this in https://github.com/lightbend/mima/pull/683 :)

build.sbt
```diff
@@ -117,7 +117,8 @@ ThisBuild / tlVersionIntroduced := Map("3" -> "2.6.1")
 
 lazy val commonJvmSettings = Seq(
   Test / fork := true,
-  Test / javaOptions := Seq("-Xmx3G")
+  Test / javaOptions := Seq("-Xmx3G"),
+  doctestGenTests := { if (tlIsScala3.value) Nil else doctestGenTests.value }
```
armanbilge, line 121: Doctests are supported on Scala 3, but our examples simply don't cross-compile e.g. kind projector syntax for `FunctionK`.

build.sbt
satorg: Seems that Scala 3.0.2 ignores any `@nowarn` annotations, unfortunately. I found it out while working on #4083.

armanbilge:
Ok, looks like we're in business :)

build.sbt
armanbilge: Yes, that's right. But`@nowarn` is supported in Scala 3.1.0 and I believe we should bump for that and other reasons. See:
- https://github.com/typelevel/scalacheck/issues/878

satorg: We won't be getting any "unused" warnings anymore after this change, is that correct?

armanbilge: That's right, no more unused warnings :(

satorg: Hmm... Wouldn't it better to disable the `-Werror` option instead (along with `-Xfatal-?whatever?` used in Scala 2.12)? In this case we will still be seeing warning and hopefully one day will have them fixed :)
Wdyt?

armanbilge: It's a good question. So IIUC, you mean enable all warnings, but disable fatal warnings?

The only problem is that there are so many unused warnings they drown out everything else maybe.

satorg: Yes, I also hate seeing warning:) This is why I attempted to make some relief in #4083. And I would be glad to continue working on that.

Answering your question – yes, I think seeing warnings is still better than suppressing them completely.

armanbilge: Yes, now that I came face-to-face with the problem myself, I truly appreciate your heroic efforts in https://github.com/typelevel/cats/pull/4083 :)

Ok, that's fair. I'll restore all the default compiler flags from sbt-typelevel, but disable fatal warnings.

satorg: Btw: there're also loads of deprecation warnings in some modules (which mostly don't make any real sense). And there're so many of them that it is quite easy to miss something really important.

armanbilge: Right, that's exactly my concern. The only way to be sure you don't miss warnings, is to make them fatal. But that means we have to disable the warnings that we can't solve today.

mima.sbt
dwijnand: Sorry I've not cut a release yet - life things shuffled my priorities the last 3 weeks.

armanbilge: No worries, just means we'll appreciate it even more when it arrives! :)

armanbilge:
Ok, for ease of review I put up a PR against this branch that sets up the new site in https://github.com/armanbilge/cats/pull/2. Not much interesting there: adds ~10 lines of build config and then reshuffles files and tweaks links to make Laika happy.

You can preview the new site at: https://armanbilge.github.io/cats

You can preview locally with `sbt docs/tlSitePreview` which starts a server on port 4242. No ruby/jekyll required.

Of course, there is plenty to bikeshed and improve. For example I didn't try very hard to incorporate the cats logos/art in the new site. All ideas welcome, but let's save them for follow up work please? 🙏 😅 The goal is to make it easy to develop the site without having to install external stuff, and easy to deploy it in CI so our changes actually go live :)

armanbilge:
Requesting review from a few folks, whenever you have a chance. Probably the most important decision is what to do about the warnings, see https://github.com/typelevel/cats/pull/4160#discussion_r843162789. Thanks!

satorg:
> Probably the most important decision is what to do about the warnings

Well, from my prospective it is better to keep warnings running. But it is IMO of course. I believe it is a safer way to get rid of them eventually. I already started addressing warnings and managed to complete quite a chunk of job in #4083. And I'd be glad to keep it up. So once we'll have wiped the remaining warnings out we can safely turn the "fatal" option back.

armanbilge:
I don't want to bring things to standstill, but FYI merging other PRs is a bit hard on this one :)

project/plugins.sbt
rossabaker: I don't see a particular reason not to bite the bullet and add the headers, though it would be nice to get this substantial work to a landing.

armanbilge: Yep, super easy to make that change. Mostly it just creates a massive diff :)

Also, I wasn't completely certain what to make of Cats' [COPYING](https://github.com/typelevel/cats/blob/main/COPYING) which includes a Scalaz notice as well. E.g. whether it's appropriate to put a blanket license over every file.

rossabaker: In http4s, we generate our copyright and also preserve the original copyright on derived files.  This is a fairly common practice, and one that doesn't irritate the tooling.

armanbilge: Right, http4s has been very diligent about that which helped. But AFAICT not a single file in Cats has a copyright header, whether its derived or not.

rossabaker: Oh, I see.  I don't know how we'd go back and add that now.  I say we made a good faith effort with the blanket COPYING file, and I'd be happy to add a second header if anyone cares enough to point out the derived works.

rossabaker (dismissed):
This is wonderful.

rossabaker (approved):
I hope I didn't piss off everyone welcoming headers, but they're almost everywhere else in Typelevel...

armanbilge:
Yep, rolled out 16k lines of headers 😬 Also applied Sergei's suggestion to keep all the warnings enabled, non-fatally. Merged in my separate site PR.

Oh, I need to set a proper start year for the headers ...

satorg:
> rolled out 16k lines of headers
915 files changed versus ~140 before that.. Seems that Git has lost tracks of small files after they received the headers..

What are the key files for the review? I doubt if I can make all 915 :)

armanbilge:
`build.sbt` is really the key file. But honestly sbt is hard to understand for a good review :) best way to review IMO is to interact with the sbt console. Check that settings are configured the way you expect. Check that MiMa previous versions includes what it should (make sure tags on your local are updated). Try previewing the website with `docs/tlSitePreview`. Stuff like that :)

armanbilge:
Any thoughts on landing this one? 🙏 I need to update this PR to account for the just-merged https://github.com/typelevel/cats/pull/4166 ...

satorg:
> Any thoughts on landing this one? 

Agree. I would vote to hold off other merges until we get this landed.

Regarding the merging of this one.... I found it tremendously difficult to reason about the safety of the changes made. They look good to me, but I realize that I cannot follow everything here.

So... What about scheduling an "RC" release after merging this PR? E.g. "2.8.0-RC1" or something. Just to let everyone (who would like to participate) test the result artifacts for potential BinCompat issues first. Would it make sense?

armanbilge:
Sure, I'm 👍 to publishing Ms, RCs, etc. :) actually, as soon as this merges it should immediately publish a snapshot if all goes well.

> Just to let everyone (who would like to participate) test the result artifacts for potential BinCompat issues first. Would it make sense?

Depends on what you want to test :) if you want to test the fix for https://github.com/typelevel/cats/issues/4062, for sure, that makes sense.

If you want to test whether the build will prevent the publishing of binary-breaking changes, then we'd have to try and introduce one first :) because besides the fix for https://github.com/typelevel/cats/issues/4062 this doesn't really change any code.

satorg (approved):
I think we should go ahead and merge it.

armanbilge:
https://s01.oss.sonatype.org/content/repositories/snapshots/org/typelevel/cats-core_3/

satorg:
@armanbilge now when I run `sbt +prePR` on main I receive a lot of errors like these below:
```
[error] /Users/satorg/projects/typelevel/cats/algebra-laws/shared/src/test/scala/algebra/laws/LawTests.scala:74:23: object DeMorganLaws in package laws is deprecated (since 2.7.0): Laws moved to LogicLaws
[error]   checkAll("Boolean", DeMorganLaws[Boolean].deMorgan(DeMorgan.fromBool(Bool[Boolean])))
[error]                       ^
[error] /Users/satorg/projects/typelevel/cats/algebra-laws/shared/src/test/scala/algebra/laws/LawTests.scala:189:26: object GroupLaws in package laws is deprecated (since 2.7.0): Provided by cats.kernel.laws
[error]     checkAll("Order[N]", GroupLaws[Order[N]].monoid)
[error]                          ^
[error] /Users/satorg/projects/typelevel/cats/algebra-laws/shared/src/test/scala/algebra/laws/LawTests.scala:193:25: object GroupLaws in package laws is deprecated (since 2.7.0): Provided by cats.kernel.laws
[error]       checkAll("Eq[N]", GroupLaws[Eq[N]].boundedSemilattice)
[error]                         ^
[error] /Users/satorg/projects/typelevel/cats/algebra-laws/shared/src/test/scala/algebra/laws/LawTests.scala:197:25: object GroupLaws in package laws is deprecated (since 2.7.0): Provided by cats.kernel.laws
[error]       checkAll("Eq[N]", GroupLaws[Eq[N]].semilattice)
[error]                         ^
```
Just to clarify: is it supposed to run this way?

armanbilge:
@satorg hmm, no, that's a bug 😕 try setting `ThisBuild / tlFatalWarningsInCi := false`.

satorg:
Unfortunately, changing values of either `tlFatalWarnings` or `tlFatalWarningsInCi` does not work since `prePR` reloads the project before starting doing everything else....

armanbilge:
Sorry, I meant we should permanently add it to `build.sbt`. It's my fault for not setting this up right to begin with :)

satorg:
No worries. I can go ahead and file a PR for that.

satorg:
@armanbilge although it is merged now... But seems I missed it while was reviewing.
https://github.com/typelevel/cats/blob/main/build.sbt#L26

It sets the default Scala version to 2.12, which is different from how it was before (2.13).
I wonder was it made for some specific reason? Because to me it seems less convenient since requires switching versions manually more often while compiling from SBT command line. Moreover, Metals will use 2.12 by default to generate Bloop files.

armanbilge:
@satorg I agree, 2.13 makes more sense. I changed it to 2.12 in ce873ff9f74d78b4eab975e78e65b06c0d580fbd because mdoc wasn't working in 2.13, since the sbt-typelevel-site plugin uses the default Scala for building/deploying documentation. In principle, we can hack around this in the build ... but just like fatal warnings, best option is to fix the root problem. I.e. get the docs building on Scala 2.13. We can make an issue for it and a contributor may be interested to work on it.

satorg:
Ah I see – it was for the reason.
Yeah it is definitely worth the fixing. I'm not sure if it is going to be a "good first issue" though... :)

PR #4169, opened 2022-04-08
Prevent `Chain` instances being backed by mutable or lazy `Seq`

bplommer:
This addresses two issues:

- Changes `Chain.Wrap` to wrap `immutable.Seq` instead of `Seq`, which until 2.13 could be mutable.
- When converting from `IterableOnce`, explicitly convert to `Vector`. Previous use of `.iterator.toSeq` can return a lazy Seq, creating race conditions - in 2.12 it returns `Stream`. In 2.13 it returns `List`, but this isn't guaranteed by the return type.

I haven't attempted to prevent `Chain` from wrapping a `LazyList` or `Stream` passed directly to the constructor.


satorg:
Just to clarify: what are the race conditions you mentioned?

core/src/main/scala-2.12/cats/data/ChainCompanionCompat.scala
```diff
@@ -0,0 +1,42 @@
...
+
+  /**
+   * Creates a Chain from the specified IterableOnce.
+   */
+  def fromIterableOnce[A](xs: IterableOnce[A]): Chain[A] =
+    xs match {
+      case s: ImSeq[A]  => fromImmutableSeq(s) // pay O(1) not O(N) cost
+      case s: MutSeq[A] => fromMutableSeq(s)
+      case notSeq =>
+        fromImmutableSeq(notSeq.toVector) // toSeq could return a Stream, creating potential race conditions
```
johnynek, line 40: let's add a more complete comment here. I don't know the race conditions you are referring to, and future comment readers might also be confused.

I think `toVector` is fine even if there is no race conditions because Vector is very memory efficient (due to internally using Array).

johnynek (approved):
Thanks for taking this!

Nice safety fix.

core/src/main/scala-2.12/cats/data/ChainCompanionCompat.scala
```diff
@@ -0,0 +1,42 @@
...
+private[data] trait ChainCompanionCompat {
+
+  /**
+   * Creates a Chain from the specified sequence.
+   */
+  def fromSeq[A](s: Seq[A]): Chain[A] =
+    s match {
+      case mut: MutSeq[A] => fromMutableSeq(mut)
+      case imm: ImSeq[A]  => fromImmutableSeq(imm)
+    }
```
satorg, line 18: Not an objection but rather just a suggestion about the style.

I personally don't like such renames on imports, because they tend to introduce new not-well known names. Moreover, Scala Docs suggests a convention about it:
https://docs.scala-lang.org/overviews/collections/overview.html

I.e. you could import it in this way:
```scala
import scala.collection.immutable
import scala.collection.mutable
```
Then use it in the code:
```scala
s match {
  case seq: mutable.Seq[A] => fromMutableSeq(mut)
  case seq: immutable.Seq[A]  => fromImmutableSeq(imm)
}
```
That looks way clearer and easier to read, doesn't it?

core/src/main/scala-2.12/cats/data/ChainCompanionCompat.scala
```diff
@@ -0,0 +1,42 @@
...
+
+  private def fromImmutableSeq[A](s: ImSeq[A]): Chain[A] = {
+    if (s.isEmpty) nil
+    else if (s.lengthCompare(1) == 0) one(s.head)
+    else Wrap(s)
+  }
```
satorg, line 24: This implementation is identical in both Scala 2.12 and 2.13+ versions. I wonder – does it make sense to make it shared to avoid unnecessary duplication?

core/src/main/scala-2.12/cats/data/ChainCompanionCompat.scala
```diff
@@ -0,0 +1,42 @@
...
+
+private[data] trait ChainCompanionCompat {
+
+  /**
+   * Creates a Chain from the specified sequence.
+   */
+  def fromSeq[A](s: Seq[A]): Chain[A] =
+    s match {
+      case mut: MutSeq[A] => fromMutableSeq(mut)
+      case imm: ImSeq[A]  => fromImmutableSeq(imm)
```
johnynek, line 17: actually, `Seq` isn't sealed, so in principle there could be other subclasses.

Can we instead do:
```scala
s match {
  case imm: ImSeq[A] => fromImmutableSeq(imm)
  case _ =>
    if (s.isEmpty) Empty
    else if (s.lengthCompare(1) == 0) Singleton(s.head)
    else fromImmutableSeq(s.toVector)
}
```

johnynek (changes requested):

armanbilge:
@bplommer polite bump, seems just a few fixes and then this is mergeable :) thanks!

bplommer:
> @bplommer polite bump, seems just a few fixes and then this is mergeable :) thanks!

Completely forgot about this! Will do soon :)

core/src/main/scala-2.12/cats/data/ChainCompanionCompat.scala
armanbilge: Yup, I very much feel the same way about this.

core/src/main/scala-2.12/cats/data/ChainCompanionCompat.scala
armanbilge: Eh, I don't think it's worth the mental indirection when reading the code. It's just a few lines, it's fine.

armanbilge (approved):

johnynek (approved):

Issue #4147, opened 2022-03-09
typeclass-based collections implementations


armanbilge:
This is a recurring idea, most recently by @posco in https://github.com/typelevel/cats/pull/4118#issuecomment-1022714696.

> i.e. we should work towards having a typeclass based implementation of HashSet, SortedSet, HashMap, SortedMap that use only typeclasses, not the instance methods.

Although the issue comes up in Cats since this is where we house instances for Scala stdlib, this might be a better PR for cats-collections.

djspiewak:
Definitely better for cats-collections. 🙂 This has been a dream for a while, but it's meaningfully hard to do correctly, and it's not clear to me that people would actually _use_ them over the standard library.

armanbilge:
> it's meaningfully hard to do correctly

I thought somewhere someone suggested we could basically copy the implementations out of Scala itself. But I am very afraid of the collections hierarchy 😆 

> it's not clear to me that people would actually use them over the standard library.

yeah, fair ...

rossabaker:
Precedent: `scalaz.Maybe` and `scalaz.IList` got used successfully where people didn't care about interop with stdlib, and were cursed mightily wherever people did.

johnynek:
People definitely won't use what we don't have.

I have definitely sunk more time into code with zero users. I think it is worth it, even if it is niche. Someone could also learn a lot about HAMT by doing it, so it is a great learning experience.

One last note: cats-collections has two immutable heaps. The standard library has zero. No one generally uses them, but I have and I bet a few others. Safe immutable types are a big part of our mission. Let's not give into "no one will use this" defeatism. They may not, but that is not how we measure success.

armanbilge:
A comment by @durban in https://github.com/typelevel/cats-effect/pull/2464 reminded me of this issue:
> What I find very strange is that it uses universal equality/hash for the keys, while there is a `Hash` typeclass in Cats.
- https://github.com/typelevel/cats-effect/pull/2464#issuecomment-991976013

Understandably, it is difficult to provide even a `Ref`-based `MapRef` implementation using `Hash` when we are lacking a `Map` based on `Hash`!

So even if `Map` is something "nobody" would use, `MapRef` certainly is.

johnynek:
well, if you don't worry to much about performance, here is how:

```scala
class Hashed[K](implicit hash: Hash[K]) {

  class Key(val unwrap: K) {
    override val hashCode: Int = hash.hashOf(unwrap)
      that match {
[...]

PR #4183, opened 2022-04-12
Move contain_ to UnorderedFoldable

BalmungSan:


core/src/main/scala/cats/syntax/unorderedFoldable.scala
```diff
@@ -29,6 +29,23 @@ trait UnorderedFoldableSyntax extends UnorderedFoldable.ToUnorderedFoldableOps {
 
 final class UnorderedFoldableOps[F[_], A](private val fa: F[A]) extends AnyVal {
 
+  /**
+   * test if `F[A]` contains an `A`, named contains_ to avoid conflict with existing contains which uses universal equality
+   *
+   * Example:
+   * {{{
+   * scala> import cats.implicits._
```
armanbilge, line 37: While we're here ...
```suggestion
   * scala> import cats.syntax.all._
```

armanbilge (commented):
Note that this will be source-breaking for anyone relying on something like `import cats.syntax.foldable._` instead of a catch-all. I think this is probably ok?

armanbilge (approved):

satorg:
I'd suppose, it might be reasonable to add `contains_` to the `UnorderedFoldable` type class itself. Because potentially it could be overridden with more efficient implementations when applicable.

armanbilge:
Good point, that may very well be the case when https://github.com/typelevel/cats/issues/4147 materializes :)

BalmungSan:
@satorg @armanbilge I move it to `UnorderedFoldable` as `contains` and the `constains_` extension now forwards to the instance method.

I assume I should test it, but not sure what would be a good test?
Something similar to `exists`?

satorg:
I'd suggest two ways to test it: as you said, consistency with `exists`, and also consistency with `Set#contains` might be useful as well. I assume, `Set` implements `UnorderedFoldable` – is that correct?

BalmungSan:
@satorg I just pushed the tests, let me know what you think.

laws/src/main/scala/cats/laws/UnorderedFoldableLaws.scala
```diff
@@ -69,6 +69,8 @@ trait UnorderedFoldableLaws[F[_]] {
   def nonEmptyRef[A](fa: F[A]): IsEq[Boolean] =
     F.nonEmpty(fa) <-> !F.isEmpty(fa)
 
+  def containsConsistentWithExists[A](fa: F[A], v: A)(implicit eq: Eq[A]): IsEq[Boolean] =
+    F.contains(fa)(v) <-> F.forall(fa)(a => eq.eqv(a, v))
```
armanbilge, line 73: Nice!

core/src/main/scala/cats/UnorderedFoldable.scala
```diff
@@ -36,6 +36,12 @@ import simulacrum.{noop, typeclass}
   def unorderedFold[A: CommutativeMonoid](fa: F[A]): A =
     unorderedFoldMap(fa)(identity)
 
+  /**
+   * Tests if `F[A]` contains an `A` using the `Eq` instance for `A`
+   */
+  def contains[A](fa: F[A])(v: A)(implicit ev: Eq[A]): Boolean =
```
armanbilge, line 42: Bikeshed: to curry or not to curry? It makes sense when the second argument is a lambda, less sure in this case.

core/src/main/scala/cats/syntax/foldable.scala
```diff
@@ -70,29 +70,15 @@ final class FoldableOps[F[_], A](private val fa: F[A]) extends AnyVal {
...
-   * scala> val l: List[Int] = List(1, 2, 3, 4)
-   * scala> l.contains_(1)
-   * res0: Boolean = true
-   * scala> l.contains_(5)
-   * res1: Boolean = false
-   * }}}
-   */
-  def contains_(v: A)(implicit ev: Eq[A], F: Foldable[F]): Boolean =
+  private[syntax] def contains_(v: A, ev: Eq[A], F: Foldable[F]): Boolean =
     F.exists(fa)(ev.eqv(_, v))
```
armanbilge, line 74:
```suggestion
    F.contains(fa)(a)
```

laws/src/main/scala/cats/laws/UnorderedFoldableLaws.scala
armanbilge: Oh, should that `forall` be `exists`? 🤔

core/src/main/scala/cats/UnorderedFoldable.scala
satorg: Yeah... That's the dilemma! :)
The concomitant  `exists` and `forall` methods are declared curried.. But they both takes labmdas (in contrast to `contains`).
I personally don't like currying unless it adds a value (for lambdas it usually does). So if Simulacrum is able to generate a correct syntax for uncurried `contains` I'd vote for making it uncurried.

BTW, are we determined to keep using Simulacrum-scalafix rules at all?

satorg:
> Note that this will be source-breaking for anyone relying on something like import cats.syntax.foldable._ instead of a catch-all. I think this is probably ok?

Well, seems that `import cats.syntax.foldable._` makes available the syntax for `UnorderedFoldable` as well:
```scala
trait FoldableSyntax extends Foldable.ToFoldableOps with UnorderedFoldable.ToUnorderedFoldableOps {
   ???
}
```
So we should be safe there (not sure if it makes sense to write a test for that though)...

laws/src/main/scala/cats/laws/UnorderedFoldableLaws.scala
BalmungSan: 🙈

core/src/main/scala/cats/UnorderedFoldable.scala
BalmungSan: @armanbilge @satorg since the name clashes with the stdlib one then I don't think it is too important or not if simulacrum can generate the right syntax.

Also, most people will be using the extension method instead of the method directly on the typeclass; and those who would use the typeclass directly probably expect the curried version; is there an example of a non-curried method? I can't recall any.

satorg: That's fair... But if you prefer to keep the typeclass method corresponding to the syntax one, then you'd better to name it `contains_` (with underscore) rather than just `contains`, wouldn't you?

satorg: I personally have no strict preference here. I prefer to avoid curried method _in general_ but realize it may not be applicable in cases like this one.

BalmungSan: > then you'd better to name it contains_ (with underscore) rather than just contains, wouldn't you?

Right, I actually just realized that was done with other methods like `filter_` I thought the `_` was just in the syntax.
I am fixing that right now.

laws/src/main/scala/cats/laws/UnorderedFoldableLaws.scala
armanbilge: Yes, I think the tests were passing because in none of the cases `v` belongs to `fa`. Then, both sides of the equivalence are always false. It may be possible to fix it with some creative manipulation of the arbitraries in the discipline tests.

core/src/main/scala/cats/UnorderedFoldable.scala
armanbilge: FWIW I don't think we should curry this. For example `as` is not curried.
https://github.com/typelevel/cats/blob/23d3c6c89bb0a1f77efd009804dd0b1d3ffb9d24/core/src/main/scala/cats/Functor.scala#L150

BalmungSan: Yeah, also most methods on `Traverse` are also not curried when not needed, `get` is the only exception. I will uncurry it.

satorg:
You may need to add the new laws created to the `cats.laws.discipline.UnorderedFoldableTests`. Otherwise these new laws will have no effect on real laws-checking tests.

laws/src/main/scala/cats/laws/UnorderedFoldableLaws.scala
satorg: I'd suggest to add more laws then:
```scala
def containsConsistentWithForall[A](fa: F[A], v: A)(implicit eq: Eq[A]): IsEq[Boolean] =
  !F.contains(fa)(v) <-> F.forall(fa)(a => eq.neqv(a, v))
```

satorg: Also I think it might be useful to add a synthetic law like this one:
```scala
  def YOU_NAME_IT[A](fa: F[A]): Boolean = {
    F.unorderedFoldMap(fa)(F.contains(fa)(_))(new CommutativeMonoid[Boolean] {
      override def combine(x: Boolean, y: Boolean): Boolean = x && y
      override def empty: Boolean = true
    })
  }
```
I.e. it should check that `contains_` returns `true` for each `A` from `F[A]`.

Perhaps, that law is a controversial one, but let's discuss it :)

BalmungSan:
> You may need to add the new laws created to the cats.laws.discipline.UnorderedFoldableTests. Otherwise these new laws will have no effect on real laws-checking tests.

Oh, right, that explains why the original buggy law did not fail the tests.
Sorry about that, I was not familiar with how the laws are organized.

But I just pushed that change plus the two extra laws you proposed!

laws/src/main/scala/cats/laws/UnorderedFoldableLaws.scala
BalmungSan: @satorg I like that law, but I think it was easier to implement using `forall` :)

satorg: Sure it was :)

tests/shared/src/test/scala/cats/tests/UnorderedFoldableSuite.scala
```diff
@@ -68,6 +68,14 @@ sealed abstract class UnorderedFoldableSuite[F[_]](name: String)(implicit
       assert(fa.count(Function.const(true)) === (fa.size))
     }
   }
+
+  test(s"UnorderedFoldable[$name].contains") {
+    forAll { (fa: F[String], v: String) =>
+      implicit val F: UnorderedFoldable[F] = instance
+      assert(fa.contains_(v) === (iterator(fa).toSet.contains(v)))
```
satorg, line 75: A very minor one.. I think, since there's `iterator` available and handy, then we can even avoid `toSet`, but rather just call `iterator(fa).contains(v)` instead.

Despite you decide to adjust it or not, the PR looks really good to me now, thanks!

satorg (approved):

satorg (approved):

laws/src/main/scala/cats/laws/discipline/UnorderedFoldableTests.scala
```diff
@@ -49,7 +49,10 @@ trait UnorderedFoldableTests[F[_]] extends Laws {
       "forall true if empty" -> forAll(laws.forallEmpty[A] _),
       "nonEmpty reference" -> forAll(laws.nonEmptyRef[A] _),
       "exists is lazy" -> forAll(laws.existsLazy[A] _),
-      "forall is lazy" -> forAll(laws.forallLazy[A] _)
+      "forall is lazy" -> forAll(laws.forallLazy[A] _),
+      "contains consitent with exists" -> forAll(laws.containsConsistentWithExists[A] _),
```
armanbilge, line 53:
```suggestion
      "contains consistent with exists" -> forAll(laws.containsConsistentWithExists[A] _),
```

laws/src/main/scala/cats/laws/UnorderedFoldableLaws.scala
```diff
@@ -69,6 +69,14 @@ trait UnorderedFoldableLaws[F[_]] {
   def nonEmptyRef[A](fa: F[A]): IsEq[Boolean] =
     F.nonEmpty(fa) <-> !F.isEmpty(fa)
 
+  def containsConsistentWithExists[A](fa: F[A], v: A)(implicit eq: Eq[A]): IsEq[Boolean] =
+    F.contains_(fa, v) <-> F.exists(fa)(a => eq.eqv(a, v))
+
+  def containsConsistentWithForall[A](fa: F[A], v: A)(implicit eq: Eq[A]): IsEq[Boolean] =
+    !F.contains_(fa, v) <-> F.forall(fa)(a => eq.neqv(a, v))
+
+  def mustContainAllElementsFromItself[A](fa: F[A])(implicit eq: Eq[A]): Boolean =
```
armanbilge, line 78:
```suggestion
  def containsAllElementsFromItself[A](fa: F[A])(implicit eq: Eq[A]): Boolean =
```

core/src/main/scala/cats/UnorderedFoldable.scala
```diff
@@ -36,6 +36,12 @@ import simulacrum.{noop, typeclass}
   def unorderedFold[A: CommutativeMonoid](fa: F[A]): A =
     unorderedFoldMap(fa)(identity)
 
+  /**
+   * Tests if `F[A]` contains an `A` using the `Eq` instance for `A`
```
armanbilge, line 40:
```suggestion
   * Tests if `F[A]` contains `v` using the `Eq` instance for `A`
```

core/src/main/scala/cats/syntax/unorderedFoldable.scala
```diff
@@ -29,12 +29,30 @@ trait UnorderedFoldableSyntax extends UnorderedFoldable.ToUnorderedFoldableOps {
 
 final class UnorderedFoldableOps[F[_], A](private val fa: F[A]) extends AnyVal {
 
+  /**
+   * Tests if `F[A]` contains an `A` using the `Eq` instance for `A`,
```
armanbilge, line 33:
```suggestion
   * Tests if `F[A]` contains `v` using the `Eq` instance for `A`,
```

armanbilge (commented):
This looks great! Just some tiny nits :)

laws/src/main/scala/cats/laws/discipline/UnorderedFoldableTests.scala
BalmungSan: 🙈

armanbilge (approved):
🎉

satorg (approved):

PR #4185, opened 2022-04-13
cats.kernel.Hash port for Scala CHAMP HashSet

DavidGregory084:
This PR implements an immutable hash set using the CHAMP encoding and `cats.kernel.Hash` for hashing of elements, as per #4147.

I've written up some basics about CHAMP [here](https://gist.github.com/DavidGregory084/1fc4fced87fc64feec17c485ff871c05) in case anyone wants to take a look.

Big questions remain:

* Does this belong in cats at all or in cats-collections?
* What is the minimal interface that we should support?
* Should the `Hash` constraint appear at the method level or at the constructor level in `HashSet`?


core/src/main/scala/cats/data/HashSet.scala
```diff
@@ -0,0 +1,840 @@
...
+   * @param that the [[cats.data.HashSet]] to check for equality with this set.
+   * @return `true` if this set and `that` are equal, `false` otherwise.
+   */
+  def ===(that: HashSet[A]): Boolean = {
+    (this eq that) || {
+      val iterThis = iterator
+      val iterThat = that.iterator
+      this == that
+
+      while (iterThis.hasNext && iterThat.hasNext) {
```
johnynek, line 145: is this safe? This implies the iteration order only depends on the `Hash[A]` value and not the order in which they are added. Can we add a comment explaining? I guess specifically about what happens on hash collisions to get a deterministic order here.

johnynek (commented):
This is very exciting!

I personally don't see a strong reason for cats-collections to exist in a separate repo, but it does. I imagine this will see much more use if added here, but then again, use will probably be small no matter what.

It is very cool. I will give a more detailed review.

core/src/main/scala/cats/data/HashSet.scala
DavidGregory084: Yes that is a good point - I have not thought about what happens here when we have a hash collision. We're just using a `Vector` for those.

FWIW, this is a bit of a stopgap while I think about how to implement equality for `BitMapNode` and `CollisionNode`, since that is how `equals` for `HashSet` implemented in the standard library.

Also I'm not sure what that stray `this == that` is doing there 😆

DavidGregory084: @johnynek I have implemented `===` and `equals` for the `Node`s now which should hopefully avoid this issue

tests/shared/src/test/scala/cats/tests/HashSetSuite.scala
```diff
@@ -0,0 +1,245 @@
...
+  property("foreach consistent with contains") {
+    forAll { (ints: List[Int]) =>
+      val hashSet = HashSet.fromSeq(ints)
+
+      val foreached = mutable.ListBuffer[Int]()
+      hashSet.foreach { foreached += _ }
+
+      assert(ints.forall(foreached.contains))
+      assert(foreached.forall(ints.contains))
+      assert(foreached.distinct.toList === foreached.toList)
```
DavidGregory084, line 205: This should probably be `.toList.distinct`

djspiewak:
I'm strongly of the opinion that cats-collections should remain separate until stabilized. The bar is very high for breaking Cats. Too high for a library where things are just getting bootstrapped.

smarter:
Since this is a port, does this contain code adapted from the Scala standard library? Cats is MIT licensed whereas the Scala standard library is Apache 2 licensed, if you're adapting code from the standard library you probably need to change your license to Apache 2 (or separately license some files as Apache 2? But that seems confusing).

smarter:
You'd also have to include it in https://github.com/typelevel/cats/blob/fe40bc2b4961386d829c0dd594af9bd80941947d/build.sbt#L321

satorg:
Just following up this discussion and realized that Cats has some compatibility-related code in the tests which is derived from Scala library.
This code is not a part of any Cats distribution though, but I wonder if we still have to draw it up in some special way?

One of such piece of code which I am personally aware of:
https://github.com/typelevel/cats/blob/main/tests/shared/src/test/scala-2.12/cats/tests/compat/SeqOps.scala#L30-L46

@smarter @DavidGregory084 – wdyt?

DavidGregory084:
Yes you are right @satorg I don't think this is a new problem in cats - see e.g. the file [HashCompat](https://github.com/typelevel/cats/blob/main/kernel/src/main/scala-2.13%2B/cats/kernel/compat/HashCompat.scala) in the main branch, which already contains code from the Scala standard library.

core/src/main/scala/cats/data/HashSet.scala
```diff
@@ -0,0 +1,937 @@
...
+  * An immutable hash set using [[cats.kernel.Hash]] for hashing.
+  *
+  * Implemented using the CHAMP encoding.
+  *
+  * @see [[https://michael.steindorfer.name/publications/phd-thesis-efficient-immutable-collections.pdf Efficient Immutable Collections]]
+  *
+  * @tparam A the type of the elements contained in this hash set.
+  * @param hash the [[cats.kernel.Hash]] instance used for hashing values.
+  */
+final class HashSet[A](
```
durban, line 42: The constructor probably should be private.

core/src/main/scala/cats/data/HashSet.scala
```diff
@@ -0,0 +1,937 @@
...
+  * @return a new [[cats.data.HashSet]] which contains all elements of `seq`.
+  */
+  def fromSeq[A](seq: Seq[A])(implicit hash: Hash[A]): HashSet[A] = {
+    val rootNode = seq.foldLeft(Node.empty[A]) { case (node, a) =>
+      node.add(a, hash.hash(a), 0)
+    }
+    new HashSet(rootNode)
+  }
+
+  sealed abstract class Node[A] {
```
durban, line 225: This class probably should be private. Also its subclasses below.

core/src/main/scala/cats/data/HashSet.scala
```diff
@@ -0,0 +1,937 @@
...
+      * Creates a new empty bitmap node.
+      *
+      * @param hash the [[cats.kernel.Hash]] instance to use to hash elements.
+      * @return a new empty bitmap node.
+      */
+    def empty[A](implicit hash: Hash[A]): Node[A] =
+      new BitMapNode(0, 0, Array.empty[Any], 0)
+  }
+
+  private[data] class Iterator[A] extends scala.Iterator[A] {
```
durban, line 772: It probably doesn't matter much, but there is a `scala.collection.AbstractIterator`, which is an abstract class (instead of a trait). (Also for the reverse iterator below.)

core/src/main/scala/cats/data/HashSet.scala
```diff
@@ -0,0 +1,937 @@
...
+    rootNode.foreach(f)
+
+  /**
+    * Test whether the set contains `value`.
+    *
+    * @param value the element to check for set membership.
+    * @return `true` if the set contains `value`, `false` otherwise.
+    */
+  final def contains(value: A): Boolean =
+    rootNode.contains(value, hash.hash(value), 0)
```
durban, line 98: Some hash tries don't use the hash code directly, but "improve" it (in case the hash function is not very good). There is `scala.collection.Hashing.improve` which is not public, but there is also `scala.util.hashing.byteswap32`, which _is_ public.

core/src/main/scala/cats/data/HashSet.scala
DavidGregory084: Yes good point, I haven't thought much about visibility - I left these public initially just for exploration in the REPL 😛

core/src/main/scala/cats/data/HashSet.scala
DavidGregory084: Ahhhh thanks that's cool, should reduce bytecode size

core/src/main/scala/cats/data/HashSet.scala
DavidGregory084: This is interesting: I guess my question is, how do you "improve" an existing hash without adding any new information to it? If the existing hash function isn't very good, won't the results of "improve" be just as bad since it is a pure function? Or is it about how the existing information in the hash is distributed across the bits of the hash?

I can't find a lot of information about the `improve` function in the Scala collections - it looks like it has been around since Scala 2.3!

https://github.com/scala/scala/commit/fee21b7e701bbb4a3f311cf06ecd668cb5d0bc94#diff-86c85e543a3eed2597cca1172039d2036015f7d471e106e94f3f0e9785009b42R58=

DavidGregory084: Ah there is more explanation of the `improve` function here: https://github.com/scala/scala/blob/2.13.x/src/library/scala/collection/mutable/HashTable.scala#L403=
It looks like `byteswap32` is preferred over the previous implementation

DavidGregory084:
I have made many things private and final. I'm not exactly sure how such modifiers are typically used in cats - is usage of `final` preferred or does it make things difficult later when deprecating things?

core/src/main/scala/cats/data/HashSet.scala
DavidGregory084: So I guess my next question is, should we assume that the input `hash` is not very well distributed if it is produced by the `Hash` instance for that type?

DavidGregory084: Resolved in https://github.com/typelevel/cats/pull/4185/commits/b497ab100e8f1b384ab1faf1da0a9be5ea016936

durban: Cool. Yes, I think it's worth assuming that `Hash` might not be very good. It's certainly true for `Int` and `Long`. And even if the hash is good, I think it's unlikely that running it through `byteswap32` would make it _worse_.

DavidGregory084:
@smarter I have added copyright notices to both this PR and #4193, and opened #4196 to address other instances in cats. In the end we decided not to update the licences key in sbt for the reasons discussed in #4189.

laws/src/main/scala/cats/laws/discipline/arbitrary.scala
```diff
@@ -415,6 +415,12 @@ object arbitrary extends ArbitraryInstances0 with ScalaVersionSpecific.Arbitrary
 
   implicit val catsLawsArbitraryForMiniInt: Arbitrary[MiniInt] =
     Arbitrary(Gen.oneOf(MiniInt.allValues))
+
+  implicit def catsLawsArbitraryForHashSet[A](implicit A: Arbitrary[A], hash: Hash[A]): Arbitrary[HashSet[A]] =
+    Arbitrary(getArbitrary[List[A]].map(HashSet.fromSeq(_)(hash)))
```
johnynek, line 420: I would rather see a different way to build a hashset. My opinion is basically this should have all the fundamental operations:

1. empty
2. fromSeq
3. `recurse + a`
4. `recurse - a`
5. `recurse0 | recurse1`
6. `recurse0 & recurse1`
7. `recurse0 -- recurse1` 

something like this (but you will need to use `Gen.frequency` to make sure the probability we branch into two is low enough that this doesn't go on forever.

The idea here is to simulate all the paths of creating a HashSet to make sure there aren't some buggy paths.

DavidGregory084:
OK, as per discussion on #4193 I have reopened this as typelevel/cats-collections#533

PR #4187, opened 2022-04-16
Suppress Compiler Warnings (kernel)

satorg:
This is an ongoing effort for suppressing compiler warnings that keep plaguing the project.

Suppresses the warnings in kernel\* modules for all Scala versions.

### DIFF
--- a/kernel/src/main/scala-2.12/cats/kernel/compat/unused.scala
+++ b/kernel/src/main/scala-2.12/cats/kernel/compat/unused.scala
@@ -21,11 +21,10 @@
 
 package cats.kernel.compat
 
+import scala.annotation.nowarn
+
 /** Simulates the behaviour of `@unused` annotation originally implemented
   * in Scala 2.13 and above.
-  * 
-  * @note copied from
-  *       [[https://github.com/scala/scala-collection-compat/blob/e35e330fca28c78c39fa2d4812fa33ce892f19e3/compat/src/main/scala-2.11_2.12/scala/annotation/unused.scala scala-collection-compat]]
-  *       source code.
   */
-final private[cats] class unused extends deprecated("unused", "unused")
+@nowarn("msg=subclassing ClassfileAnnotation does not\nmake your annotation visible at runtime.")
+final private[cats] class unused extends nowarn("cat=unused")
--- a/kernel/src/main/scala-2.12/cats/kernel/compat/unused.scala
+++ b/kernel/src/main/scala-2.12/cats/kernel/compat/unused.scala
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) 2015 Typelevel
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package cats.kernel.compat
+
+/** Simulates the behaviour of `@unused` annotation originally implemented
+  * in Scala 2.13 and above.
+  * 
+  * @note copied from
+  *       [[https://github.com/scala/scala-collection-compat/blob/e35e330fca28c78c39fa2d4812fa33ce892f19e3/compat/src/main/scala-2.11_2.12/scala/annotation/unused.scala scala-collection-compat]]
+  *       source code.
+  */
+final private[cats] class unused extends deprecated("unused", "unused")
--- a/kernel/src/main/scala/cats/kernel/instances/MapInstances.scala
+++ b/kernel/src/main/scala/cats/kernel/instances/MapInstances.scala
@@ -42,12 +42,14 @@
 package cats.kernel
 package instances
 
+import cats.kernel.compat.scalaVersionSpecific._
+import cats.kernel.compat.unused
+
 import scala.collection.mutable
-import compat.scalaVersionSpecific._
 
 @suppressUnusedImportWarningForScalaVersionSpecific
 trait MapInstances extends MapInstances1 {
-  implicit def catsKernelStdHashForMap[K: Hash, V: Hash]: Hash[Map[K, V]] =
+  implicit def catsKernelStdHashForMap[K, V](implicit @unused K: Hash[K], V: Hash[V]): Hash[Map[K, V]] =
     new MapHash[K, V]
 
   implicit def catsKernelStdCommutativeMonoidForMap[K, V: CommutativeSemigroup]: CommutativeMonoid[Map[K, V]] =
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,11 +15,6 @@ on:
     tags: [v*]
 
 env:
-  PGP_PASSPHRASE: ${{ secrets.PGP_PASSPHRASE }}
-  SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
-  SONATYPE_CREDENTIAL_HOST: ${{ secrets.SONATYPE_CREDENTIAL_HOST }}
-  SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
-  PGP_SECRET: ${{ secrets.PGP_SECRET }}
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
 jobs:
@@ -51,21 +46,21 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout current branch (full)
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           fetch-depth: 0
 
       - name: Download Java (temurin@8)
         id: download-java-temurin-8
         if: matrix.java == 'temurin@8'
-        uses: typelevel/download-java@v1
+        uses: typelevel/download-java@v2
         with:
           distribution: temurin
           java-version: 8
 
       - name: Setup Java (temurin@8)
         if: matrix.java == 'temurin@8'
-        uses: actions/setup-java@v2
+        uses: actions/setup-java@v3
         with:
           distribution: jdkfile
           java-version: 8
@@ -74,14 +69,14 @@ jobs:
       - name: Download Java (temurin@17)
         id: download-java-temurin-17
         if: matrix.java == 'temurin@17'
-        uses: typelevel/download-java@v1
+        uses: typelevel/download-java@v2
         with:
           distribution: temurin
           java-version: 17
 
       - name: Setup Java (temurin@17)
         if: matrix.java == 'temurin@17'
-        uses: actions/setup-java@v2
+        uses: actions/setup-java@v3
         with:
           distribution: jdkfile
           java-version: 17
@@ -90,21 +85,21 @@ jobs:
       - name: Download Java (graalvm@11)
         id: download-java-graalvm-11
         if: matrix.java == 'graalvm@11'
-        uses: typelevel/download-java@v1
+        uses: typelevel/download-java@v2
         with:
           distribution: graalvm
           java-version: 11
 
       - name: Setup Java (graalvm@11)
         if: matrix.java == 'graalvm@11'
-        uses: actions/setup-java@v2
+        uses: actions/setup-java@v3
         with:
           distribution: jdkfile
           java-version: 11
           jdkFile: ${{ steps.download-java-graalvm-11.outputs.jdkFile }}
 
       - name: Cache sbt
-        uses: actions/cache@v2
+        uses: actions/cache@v3
         with:
           path: |
             ~/.sbt
@@ -116,7 +111,7 @@ jobs:
           key: ${{ runner.os }}-sbt-cache-v2-${{ hashFiles('**/*.sbt') }}-${{ hashFiles('project/build.properties') }}
 
       - name: Check that workflows are up to date
-        run: sbt githubWorkflowCheck
+        run: sbt 'project ${{ matrix.project }}' '++${{ matrix.scala }}' 'project /' githubWorkflowCheck
 
       - name: Check headers and formatting
         if: matrix.java == 'temurin@8'
@@ -151,7 +146,7 @@ jobs:
 
       - name: Upload target directories
         if: github.event_name != 'pull_request' && (startsWith(github.ref, 'refs/tags/v') || github.ref == 'refs/heads/main')
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v3
         with:
           name: target-${{ matrix.os }}-${{ matrix.java }}-${{ matrix.scala }}-${{ matrix.project }}
           path: targets.tar
@@ -168,21 +163,21 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout current branch (full)
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           fetch-depth: 0
 
       - name: Download Java (temurin@8)
         id: download-java-temurin-8
         if: matrix.java == 'temurin@8'
-        uses: typelevel/download-java@v1
+        uses: typelevel/download-java@v2
         with:
           distribution: temurin
           java-version: 8
 
       - name: Setup Java (temurin@8)
         if: matrix.java == 'temurin@8'
-        uses: actions/setup-java@v2
+        uses: actions/setup-java@v3
         with:
           distribution: jdkfile
           java-version: 8
@@ -191,14 +186,14 @@ jobs:
       - name: Download Java (temurin@17)
         id: download-java-temurin-17
         if: matrix.java == 'temurin@17'
-        uses: typelevel/download-java@v1
+        uses: typelevel/download-java@v2
         with:
           distribution: temurin
           java-version: 17
 
       - name: Setup Java (temurin@17)
         if: matrix.java == 'temurin@17'
-        uses: actions/setup-java@v2
+        uses: actions/setup-java@v3
         with:
           distribution: jdkfile
           java-version: 17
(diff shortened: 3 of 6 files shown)


core/src/main/scala/cats/TraverseFilter.scala
```diff
@@ -124,7 +124,12 @@ trait TraverseFilter[F[_]] extends FunctorFilter[F] {
...
-   * This is usually faster than ordDistinct, especially for things that have a slow comparion (like String).
+   * This is usually faster than ordDistinct, especially for things that have a slow comparison (like String).
+   * 
+   * @note the passed `Hash` typeclass is not currently in use.
+   * @todo replace `HashSet` from Scala library with one that can make use of the `Hash` typeclass
+   *       instead of the Java's `hashCode` method. Consider a candidate implementation in
+   *       [[https://github.com/typelevel/cats/pull/4185 PR#4185]].
    */
   def hashDistinct[A](fa: F[A])(implicit H: Hash[A]): F[A] =
     traverseFilter[State[HashSet[A], *], A, A](fa)(a =>
```
satorg, line 135: Raises the "parameter `H: Hash[A]` not used" warning because it is really so.

I'd prefer to keep this warning in place until Cats receive a suitable `HashSet` implementation (perhaps by merging #4185).

johnynek (approved):
Thanks for the cleanup!

satorg:
Seems there can be some issues with migrating to Scala 3.1.x versions (see discussion in https://github.com/http4s/http4s/pull/6300#pullrequestreview-943891503)

Perhaps, it may make sense to make use of `val _ = someUnusedParam` pattern in this PR as well...

armanbilge:
The other option is we continue to keep fatal warnings disabled on Scala 3 😕

satorg:
I don't think it's going to help... Scala 3.0 silently ignores unused parameters along with any kind of `@nowarn` annotation. Whereas Scala 3.1 will be failing on `@nowarn("cat=unused-params")` anyway regardless of the fatal warnings option.

satorg:
Added more thoughts on how to overcome the compatibility issue: https://github.com/http4s/http4s/pull/6300#discussion_r851824642

I'm considering to take that approach, but sincerely open to discussion :)

core/src/main/scala/cats/data/NonEmptySet.scala
```diff
@@ -428,7 +428,7 @@ sealed abstract private[data] class NonEmptySetInstances extends NonEmptySetInst
 }
 
 sealed abstract private[data] class NonEmptySetInstances0 extends NonEmptySetInstances1 {
-  implicit def catsDataHashForNonEmptySet[A: Order: Hash]: Hash[NonEmptySet[A]] =
+  implicit def catsDataHashForNonEmptySet[A: Hash]: Hash[NonEmptySet[A]] =
     Hash[SortedSet[A]].asInstanceOf[Hash[NonEmptySet[A]]]
```
satorg, line 432: Can someone explain (or guess at least) why does this change not concern Mima? I removed the unused implicit `Order` parameter but seems that Mima is completely fine about that.

core/src/main/scala/cats/data/NonEmptySet.scala
```diff
@@ -428,7 +428,7 @@ sealed abstract private[data] class NonEmptySetInstances extends NonEmptySetInst
 }
 
 sealed abstract private[data] class NonEmptySetInstances0 extends NonEmptySetInstances1 {
-  implicit def catsDataHashForNonEmptySet[A: Order: Hash]: Hash[NonEmptySet[A]] =
+  implicit def catsDataHashForNonEmptySet[A: Hash]: Hash[NonEmptySet[A]] =
```
satorg, line 431: Can someone explain (or guess at least) why does this change not concern Mima? I removed the unused implicit `Order` parameter but seems that Mima is completely fine about that.

satorg, line 431: Can someone explain (or guess at least) why does this change not concern Mima? I removed the unused implicit `Order` parameter but seems that Mima is completely fine about that.

kernel/src/main/scala-2.12/cats/kernel/compat/unused.scala
```diff
@@ -0,0 +1,31 @@
...
+package cats.kernel.compat
+
+/** Simulates the behaviour of `@unused` annotation originally implemented
+  * in Scala 2.13 and above.
+  * 
+  * @note copied from
+  *       [[https://github.com/scala/scala-collection-compat/blob/e35e330fca28c78c39fa2d4812fa33ce892f19e3/compat/src/main/scala-2.11_2.12/scala/annotation/unused.scala scala-collection-compat]]
+  *       source code.
+  */
+final private[cats] class unused extends deprecated("unused", "unused")
```
satorg, line 31: Just to draw more attention to this Scala version compatibility trick.

core/src/main/scala/cats/data/NonEmptySet.scala
armanbilge: Um, that seems suspect. Let me look into it, hopefully there's a good explanation ...

armanbilge: Ok, I'm really confused ... but is this all private??

`NonEmptySetInstances0` is private, and extended by `NonEmptySetInstances`, which is private, which is implemented by `NonEmptySetImpl`, also private.

satorg: Right.. But the implicit method itself is public and should be publicly accessible as a part of the `NonEmptySet` object API.

armanbilge: That's the thing .. AFAICT there _is_ no (public) `object NonEmptySet` anywhere in Cats 🙃

satorg: That's correct... It is just a val: `val NonEmptySet = NonEmptySetImpt`...

satorg: ..and the same trick is used for `NonEmptyMap` and `NonEmptyChain`

armanbilge: Ah, thanks, I'm sorry. Didn't realize that's how it's implemented.
https://github.com/typelevel/cats/blob/708cbaba2e6ac07a5ac1819b6c18b0c83a2faaba/core/src/main/scala/cats/data/package.scala#L39

The bad news is, we won't get any MiMa support on this :( The reason is that the type of `NonEmptySet` is `NonEmptySetImpl.type` aka `NonEmptySetImpl$`, which is package-private, so MiMa ignores it. Similarly, you won't find any of these APIs in the scaladocs.

In fact, I'm pretty surprised this works at all. This smells a lot to me like https://github.com/scala/bug/issues/6794.

armanbilge: If we want to get MiMa support for these, we'll have to expose them as a public `class` or `trait` somewhere, which should match the type of that `val NonEmptySet`.

satorg: Oh, I see... Is it even possible to configure Mima to force it looking into some specific objects even if they're package private? We would need to add those 3 objects then at least...

armanbilge: You can open an issue on MiMa, but not sure what can be done there.

https://github.com/typelevel/cats/blob/708cbaba2e6ac07a5ac1819b6c18b0c83a2faaba/core/src/main/scala/cats/data/NonEmptySet.scala#L30

Instead I propose this solution: 
```scala
sealed class NonEmptySetCompanion extends NonEmptySetInstances with Newtype`
private[data] object NonEmptySetImpl extends NonEmptySetCompanion

val NonEmptySet: NonEmptySetCompanion = NonEmptySetImpl
```

I think that should get us what we want.

armanbilge: Ah, but note, even if we do my proposal and publish it as 2.8.0, it won't check bincompat for us against 2.7.0 or earlier. For that we would really need changes in MiMa.

satorg: Does Mima check against previous snapshots created out of the current `main` branch? What if I first create a PR with public sealed class as you proposed, then we merge it and then – I wonder – will Mima be checking against the already merged code, wdyt?

satorg: And another question... Should it be really a class? Or will a public sealed trait be enough there as well?

armanbilge: > Does Mima check against previous snapshots created out of the current main branch?

No, it's not currently configured to check snapshots, but in theory we could try and configure that. Technically, we don't guarantee binary-compatibility against snapshots, and it might be annoying if we add something but then remove it or change it before the release.

> And another question... Should it be really a class? Or will a public sealed trait be enough there as well?

Yeah, `class`, `trait`, whatever :) so long as its public.

armanbilge: Btw, there is one other option. Cats has a manual bincompat test here:
https://github.com/typelevel/cats/blob/main/binCompatTest/src/main/scala-2/catsBC/MimaExceptions.scala

It actually compiles against an old Cats artifact, and then attempts to run it with the current branch. So that is the true test for bincompat (unlike MiMa, which is just doing an analysis). So in theory we could add use all of these methods in there, and it will be tested as part of CI even if MiMa can't check it for us.

satorg: Oh, cool, I didn't know about it. So do you mean – if I add an appropriate check for that implicit there, then it will be checked against the previous published artifact and we can see if really BinCompat or not – is that correct?

armanbilge: Yes, that's right. If you do something which uses `catsDataHashForNonEmptySet` it _should_ fail since you removed the `Order`.

DavidGregory084: You could possibly try a bytecode level compatibility checker like [Revapi](https://revapi.org/revapi-site/main/getting-started.html) or [SigTest](https://wiki.openjdk.java.net/display/CodeTools/SigTest) as a one off, although I don't know how well they will work on Scala projects

satorg: Hopefully handled in #4219.

satorg:
Hi @armanbilge. I realized that warnings for Scala 2.12 for unused .. well, everything .. were suppressed entirely in [sbt-typelevel](https://github.com/typelevel/sbt-typelevel/blame/v0.4.12/settings/src/main/scala/org/typelevel/sbt/TypelevelSettingsPlugin.scala#L88):
```
      val warnings212 = Seq("-Xlint:-unused,_")
```
So.. I wonder – was it made intentionally and can we assume now that the warnings regarding unused stuff will never show up for Scala 2.12?

armanbilge:
Oh, interesting 😅 sbt-typelevel inherited those settings from sbt-spiewak which derived them from sbt-tpolecat. So probably intentional. In any case, it's what all the builds are using now, and many were using before too, so it probably works just fine.

satorg:
That's fine... Just curious though – is it just because 2.12 is sunsetting and no one likes messing around with it? Or is it something more special...

armanbilge:
Here's the original commit. https://github.com/djspiewak/sbt-spiewak/commit/95a45d0ea91a015c1e38e9f56d5533bb3acd8f28

satorg:
Oh that's interesting... It was introduces even before Scala 2.13.0 was released. Seems like Daniel just didn't like that sort of warnings at all 😄

armanbilge:
Btw, remember sbt-typelevel hosts itself and sbt uses Scala 2.12. I'll have to check later but I'm pretty sure I have unused warnings when I develop it. So maybe it's enabled by something else 🤔

satorg:
Well, for Cats it definitely disabled because if I enable it temporarily and re-compile for 2.12, then the "unused" warnings begin coming in numbers.

armanbilge:
You're right. I was dreaming about sbt-typelevel, seems I don't have unused warnings there at all 🙃

armanbilge:
@satorg I've bumped sbt-typelevel to a snapshot with your PR (thanks again for doing that!). That should unblock your work, and we'll make sure to get on a stable release before we merge :)

satorg:
@armanbilge Initially I intended to fix compiler warnings in a bulk. And I completed with "kernel" modules quite a long ago. But turned out that "core" modules have quite a lot of warnings like these:
```
[warn] /Users/.../typelevel/cats/core/src/main/scala/cats/syntax/bitraverse.scala:26:22: parameter value evidence$1 in method catsSyntaxBitraverse is never used
[warn]   implicit final def catsSyntaxBitraverse[F[_, _]: Bitraverse, A, B](fab: F[A, B]): BitraverseOps[F, A, B] =
[warn]                      ^
[warn] /Users/.../typelevel/cats/core/src/main/scala/cats/syntax/bitraverse.scala:31:22: parameter value evidence$2 in method catsSyntaxNestedBitraverse is never used
[warn]   implicit final def catsSyntaxNestedBitraverse[F[_, _]: Bitraverse, G[_], A, B](
[warn]                      ^
[warn] /Users/.../typelevel/cats/core/src/main/scala/cats/syntax/bitraverse.scala:48:22: parameter value evidence$4 in method catsSyntaxBitraverseBinCompat0 is never use
[warn]   implicit final def catsSyntaxBitraverseBinCompat0[F[_, _]: Bitraverse, A, B](
[warn]                      ^
```
which require quite gentle and precise handling I think.

But since the "kernel" modules seem to be ready, I'd like to separate the work into multiple PRs and thus keep the kernel-related part within this PR to let it being reviewed and hopefully merged. Meanwhile I can continue working on the "core" part.

Would it work, wdyt?

satorg:
Ok nice, thank you. I think this PR is pretty much ready then.

kernel/src/main/scala/cats/kernel/instances/MapInstances.scala
```diff
@@ -47,11 +47,15 @@ import compat.scalaVersionSpecific._
 
 @suppressUnusedImportWarningForScalaVersionSpecific
 trait MapInstances extends MapInstances1 {
-  implicit def catsKernelStdHashForMap[K: Hash, V: Hash]: Hash[Map[K, V]] =
+  implicit def catsKernelStdHashForMapBinCompat[K, V: Hash]: Hash[Map[K, V]] =
```
armanbilge, line 50: I'm actually not sure about this change. Even though the `Hash[K[` instance is not used, arguably it does serve as evidence that `K` _is_ hashable (e.g., it's not a `Function`).

See Oscar's interesting take on this in https://github.com/typelevel/cats/pull/4193#issuecomment-1119933018.

satorg: Well, I'm ok to roll it back, but then another question is popping up actually: why the `MapHash` class uses the universal `.hashCode` on keys rather than `Hash[K]`? I am totally aware that Scala's `Map` itself uses universal hashes – but so what? The purpose of `MapHash` is to calculate `Hash[Map[K, V]]` out of the passed `Map` but it does not treat it as the "map", actually. All it does – just enumerates all its keys and values and produces a resulting hash value.

To me it seems that it would be better to make `MapHash` relying `Hash[K]` upon the calculation instead of the universal hash for `K`. Or are there come other considerations which I am just not aware of?

armanbilge: > but so what?

😀 yeah, ok, great point. I think it dates back to when `Hash` was introduced in https://github.com/typelevel/cats/pull/1712 and I don't see any discussion on that specifically. Probably a separate PR, but I think it makes sense to change it (anyway we currently have a law claiming that `Hash` is consistent with `hashCode` so in theory it should not change anything).

armanbilge: Ah, well, in theory you could end up with something inconsistent with `MapEq`? Since Scala forces our hand there to use universal equality.

https://github.com/typelevel/cats/blob/f8f6eff94872cba8bead3ae967d5b4f60f1b068d/kernel/src/main/scala/cats/kernel/instances/MapInstances.scala#L87-L97

But if we can't use the `Hash[K]`, I do see an argument we shouldn't ask for it. Ugh.

armanbilge: So still maybe the best strategy for now is to keep the `Hash[K]` and suppress the warning. It still has some use, by proving `K` is hashable, and it leaves the door open for us to decide what we want to do.

Anyway, we have a real `HashMap` now so we shouldn't stress over this too much ... we should figure out how to land the other PR instead.

kernel/src/main/scala-2.12/cats/kernel/compat/unused.scala
satorg: I'm on a fence with this implementation of `unused` for Scala 2.12.
Turns out that it does not always work out everywhere (due to 2.12 limitations I guess). For example, in this case (let's name it "A")
https://github.com/typelevel/cats/blob/f8f6eff94872cba8bead3ae967d5b4f60f1b068d/core/src/main/scala/cats/data/Const.scala#L42-L43
it works just fine.

But on the other hand for this case "B":
https://github.com/typelevel/cats/blob/f8f6eff94872cba8bead3ae967d5b4f60f1b068d/core/src/main/scala/cats/ApplicativeError.scala#L356-L362

it produces another quite weird warning (instead of "parameter is never used"):
```
[warn] /Users/.../typelevel/cats/core/src/main/scala/cats/ApplicativeError.scala:359:7: value NT is deprecated
[warn]       try {
[warn]       ^
```
And there are a few more similar cases in `core`. I figured out that for the latter case ("B") the following implementation of `@unused` works better:
```scala
final private[cats] class unused extends scala.annotation.nowarn("msg=is never used")
```
But it does not work at all for the former case ("A") – Scala 2.12 warns the `@nowarn` annotation does not suppress any warning.

So seems that to  handle differences between 2.12 and 2.13 properly we need to have 2 different compatibility implementations for `@unused`: (1) `@deprecated`-based and (2) `@nowarn`-based. And pick one of them depending on a particular case.

@armanbilge any thoughts on this?

armanbilge:
@satorg I'm sorry, I haven't forgotten about this PR.

The truth is I've been hesitating, because I'm weighing what I think are fairly straightforward changes here with the fact it requires putting Cats on an sbt-typelevel milestone. There are some non-trivial changes coming to sbt-typelevel and I'd rather not have to worry about accidentally breaking the Cats build (the fact that Cats Effect is stuck on the milestone is stress enough!).

- upgrade to [laika 0.19](https://github.com/typelevel/sbt-typelevel/issues/228)
- rework to use [scalac-options](https://github.com/typelevel/scalac-options)
- replacing the [sonatype client](https://github.com/davenverse/sonatype/)

In theory we could backport your scalac warning fixes to sbt-tl 0.4.x. However if you're eager to work on that what I'd actually love help with is integrating with the scalac-options library for 0.5.x.

satorg:
@armanbilge yeah, no worries. I also feel a bit hesitant about pushing this work out. Because before continuing it would be better to stabilize the build system itself.

> However if you're eager to work on that what I'd actually love help with is integrating with the scalac-options library for 0.5.x.

Ah ok, I think I can help with that. Thanks.

satorg:
Converted to "Draft" to put on hold until work on scalac compiler options is stabilized.

som-snytt:
I arrived via good first issues [sic]. I know about options on Scala 2 and somewhat on Scala 3, and I generally endorse using `-Xlint`, also it's not "fatal warnings" but `-Werror` which I pronounce "worry". I'm interested in learning how cats can be supported by the options cross-compilation, but maybe you're already working out a solution.

armanbilge:
@satorg @som-snytt if there are folks motivated to work on this, I'm thinking we should just update Cats to the sbt-typelevel milestone to unblock this work.

In https://github.com/typelevel/cats/pull/4187#issuecomment-1243184576 I was worried about too many changes happening in sbt-typelevel, but besides the Laika upgrade the other objectives have been dormant, and I don't think that they are going to land in 0.5 after all. That's okay 🙁

satorg:
@armanbilge @som-snytt if I remember correctly, there are quite a few warnings that are handled differently in different Scala versions (i.e. 2.12, 2.13, 3). And that is what we tried to address with  [scalac-compat](https://github.com/typelevel/scalac-compat). I think, it could come handy for Cats too. Are we fine with bringing it in there?

armanbilge:
Yes, scalac-compat has been wonderful! We are using it in Cats Effect now as well.

satorg:
scala/scala3#20536 (for reference)

satorg:
Superseded by #4614.

nifledcar7-blip:
> This is an ongoing effort for suppressing compiler warnings that keep plaguing the project.
> 
> Suppresses the warnings in kernel* modules for all Scala versions.

http4s/http4s#6300

PR #4191, opened 2022-04-19
Make `catsInstancesForIdCompat2_6_1` package-private

armanbilge:
I initially kept this public, because usually static forwarders go away when a method is package-private, and we need them for complete binary-compatibility. However, I learned in https://github.com/lampepfl/dotty/issues/10842 that this is not the case on Scala 3. So, there is no reason to leak this method in the public API.

### DIFF
--- a/core/src/main/scala/cats/Invariant.scala
+++ b/core/src/main/scala/cats/Invariant.scala
@@ -128,7 +128,7 @@ object Invariant extends ScalaVersionSpecificInvariantInstances with InvariantIn
     cats.catsInstancesForId
   @deprecated("Added for bincompat", "2.8.0")
   @cats.compat.targetName("catsInstancesForId")
-  def catsInstancesForIdCompat2_6_1: Comonad[Id] =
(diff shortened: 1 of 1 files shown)


core/src/main/scala/cats/Invariant.scala
```diff
@@ -128,7 +128,7 @@ object Invariant extends ScalaVersionSpecificInvariantInstances with InvariantIn
     cats.catsInstancesForId
   @deprecated("Added for bincompat", "2.8.0")
   @cats.compat.targetName("catsInstancesForId")
```
satorg, line 130: Could you provide some details on how this guy is supposed to work please? I think it might help me with my work in #4187.

armanbilge: Definitely, do you mean about the compat thing or the semantics of `targetName` which is a Scala 3 feature? You can see how it works in https://github.com/typelevel/cats/issues/4143#issuecomment-1085761689.

rossabaker (approved):

PR #4193, opened 2022-04-21
cats.kernel.Hash port for Scala CHAMP HashMap

DavidGregory084:
This PR implements an immutable hash map using the CHAMP encoding and `cats.kernel.Hash` for hashing of elements, as per #4147.

This accompanies PR #4185 which implements a hash set.


core/src/main/scala/cats/data/HashMap.scala
```diff
@@ -0,0 +1,1196 @@
...
+
+  /**
+    * Creates a new map with an additional key-value pair, unless the key is already present,
+    * in which case the value for `key` is replaced by `value`.
+    *
+    * @param key the key to be added.
+    * @param value the value to be added.
+    * @return a new map that contains all key-value pairs of this map and that also contains a mapping from `key` to `value`.
+    */
+  final def add[VV >: V](key: K, value: VV): HashMap[K, VV] = {
```
johnynek, line 156: doesn't scala use `updated` here? Can we use that here?

core/src/main/scala/cats/data/HashMap.scala
```diff
@@ -0,0 +1,1196 @@
...
+    fromSeq(kvs)
+
+  /**
+  * Creates a new [[cats.data.HashMap]] which contains all elements of `seq`.
+  *
+  * @param seq the sequence of elements to add to the [[cats.data.HashMap]].
+  * @param hashKey the [[cats.kernel.Hash]] instance used for hashing values.
+  * @return a new [[cats.data.HashMap]] which contains all elements of `seq`.
+  */
+  final def fromSeq[K, V](seq: Seq[(K, V)])(implicit hashKey: Hash[K]): HashMap[K, V] = {
```
johnynek, line 268: maybe this should be
```scala
fromFoldable[F[_], K, V](fa: F[(K, V)])(implicit hashKey: Hash[K], F: Foldable[F]): HashMap[K, V]
```

core/src/main/scala/cats/data/HashMap.scala
```diff
@@ -0,0 +1,1196 @@
...
+    }
+    new HashMap(rootNode)
+  }
+
+  sealed abstract private[data] class Node[K, +V] {
+
+    /**
+      * @return The number of value and node elements in the contents array of this trie node.
+      */
+    def allElements: Int
```
johnynek, line 280: nit: I would call this `sizeAllElements` similarly below. `allElements` sounds like you are getting a collection not an integer back.

core/src/main/scala/cats/data/HashMap.scala
```diff
@@ -0,0 +1,1196 @@
...
+        this
+      else
+        new CollisionNode(newKeyHash, contents :+ (newKey -> newValue))
+
+    final override def remove(key: K, keyHash: Int, depth: Int): Node[K, V] =
+      if (!contains(key, keyHash, depth))
+        this
+      else {
+        val newContents = contents.filterNot { case (k, _) => hashKey.eqv(key, k) }
+        if (newContents.size > 1)
```
johnynek, line 479: nit: what about `newContents.lengthCompare(1) > 0`

core/src/main/scala/cats/data/HashMap.scala
```diff
@@ -0,0 +1,1196 @@
...
+    * this node type is used to collect all of the colliding elements and implement the [[HashMap.Node]]
+    * interface at a performance cost compared with a [[HashMap.BitMapNode]].
+    *
+    * @tparam A the type of the elements contained in this node.
+    * @param collisionHash the hash value at which all of the contents of this node collide.
+    * @param contents the value elements whose hashes collide.
+    */
+  final private[HashMap] class CollisionNode[K, +V](
+    val collisionHash: Int,
+    val contents: Vector[(K, V)]
```
johnynek, line 430: should we use `NonEmptyVector[(K, V)]` here?

core/src/main/scala/cats/data/HashMap.scala
```diff
@@ -0,0 +1,1196 @@
...
+          val newContentsArray = new Array[Any](Node.StrideLength * newContents.length)
+          var i = 0
+          while (i < newContents.length) {
+            val (k, v) = newContents(i)
+            val keyIndex = Node.StrideLength * i
+            newContentsArray(keyIndex) = k
+            newContentsArray(keyIndex + 1) = v
+            i += 1
+          }
+          new BitMapNode[K, V](bitPos, 0, newContentsArray, newContents.size)
```
johnynek, line 495: nit: you are using `newContents.size` here but `.length` above. Can we use the same one?

core/src/main/scala/cats/data/HashMap.scala
```diff
@@ -0,0 +1,1196 @@
...
+        if (newContents.size > 1)
+          new CollisionNode(collisionHash, newContents)
+        else {
+          // This is a singleton node so the depth doesn't matter;
+          // we only need to index into it to inline the value in our parent node
+          val mask = Node.maskFrom(collisionHash, 0)
+          val bitPos = Node.bitPosFrom(mask)
+          val newContentsArray = new Array[Any](Node.StrideLength * newContents.length)
+          var i = 0
+          while (i < newContents.length) {
```
johnynek, line 488: I think `foreach` may be better here since repeatedly accessing a vector is actually `N log N`, but foreach can be truly `O(N)`.

core/src/main/scala/cats/data/HashMap.scala
```diff
@@ -0,0 +1,1196 @@
...
+    final private def hasValueAt(bitPos: Int): Boolean =
+      (valueMap & bitPos) != 0
+
+    final def getKey(index: Int): K =
+      contents(Node.StrideLength * index).asInstanceOf[K]
+
+    final def getValue(index: Int): V =
+      contents(Node.StrideLength * index + 1).asInstanceOf[V]
+
+    final def getMapping(index: Int): (K, V) =
```
johnynek, line 573: what about:
```scala
final def getMapping(index: Int): (K, V) = {
  val offset = Node.StrideLength * index
  (contents(offset).asInstanceOf[K],
    contents(offset + 1).asInstanceOf[V])
}
```

core/src/main/scala/cats/data/HashMap.scala
```diff
@@ -0,0 +1,1196 @@
...
+    final override def toString(): String = {
+      s"""CollisionNode(hash=${collisionHash}, values=${contents.mkString("[", ",", "]")})"""
+    }
+  }
+
+  /**
+    * A CHAMP bitmap node. Stores value element and node element positions in the `contents` array
+    * in the `valueMap` and `nodeMap` integer bitmaps.
+    *
+    * @tparam A the type of the elements contained in this node.
```
johnynek, line 532: there is no `A` type parameter. Can you reword with `K` and `V`?

core/src/main/scala/cats/data/HashMap.scala
```diff
@@ -0,0 +1,1196 @@
...
+    * @param valueMap integer bitmap indicating the notional positions of value elements in the `contents` array.
+    * @param nodeMap integer bitmap indicating the notional positions of node elements in the `contents` array.
+    * @param contents an array of `A` value elements and `Node[A]` sub-node elements.
```
johnynek, line 535: can you write the invariant here? Which values have keys, values or nodes?

core/src/main/scala/cats/data/HashMap.scala
```diff
@@ -0,0 +1,1196 @@
...
+
+    final def getValue(index: Int): V =
+      contents(Node.StrideLength * index + 1).asInstanceOf[V]
+
+    final def getMapping(index: Int): (K, V) =
+      contents(Node.StrideLength * index).asInstanceOf[K] ->
+        contents(Node.StrideLength * index + 1).asInstanceOf[V]
+
+    final def getNode(index: Int): Node[K, V] =
+      contents(contents.length - 1 - index).asInstanceOf[Node[K, V]]
```
johnynek, line 578: this is some tricky stuff! `contents` has a very complex type: some offsets have `K`, some `V` and some `Node[K, V]`.

core/src/main/scala/cats/data/HashMap.scala
```diff
@@ -0,0 +1,1196 @@
...
+
+    final def foreach[U](f: (K, V) => U): Unit = {
+      var i = 0
```
johnynek, line 581: I think:
```scala
val fnTup = fn.tupled
```
will allocate a function once and use it for the whole foreach.

core/src/main/scala/cats/data/HashMap.scala
```diff
@@ -0,0 +1,1196 @@
...
+}
+
+class HashMapMonoid[K: Hash, V](implicit V: Semigroup[V]) extends Monoid[HashMap[K, V]] {
+
+  def empty: HashMap[K, V] = HashMap.empty[K, V]
+
+  def combine(xs: HashMap[K, V], ys: HashMap[K, V]): HashMap[K, V] = {
+    if (xs.size <= ys.size) {
+      xs.iterator.foldLeft(ys) { case (my, (k, x)) =>
+        my.add(k, Semigroup.maybeCombine(x, my.get(k)))
```
johnynek, line 1188: nit:
```scala
my.get(k) match {
  case Some(y) => my.updated(k, Semigroup.combine(x, y))
  case None => my
}
```
similarly below

kernel/src/main/scala-2.12/cats/kernel/compat/HashCompat.scala
```diff
@@ -80,6 +80,27 @@ private[kernel] class HashCompat {
...
+    import scala.util.hashing.MurmurHash3._
+    var a = 0
+    var b = 0
+    var c = 1
+    var n = 0
+    xs.foreach { x =>
+      val h = A.hash(x)
+      a += h
+      b ^= h
+      if (h != 0) c *= h
```
johnynek, line 94: can we use the branch free version below?

tests/shared/src/test/scala/cats/tests/HashMapSuite.scala
```diff
@@ -0,0 +1,292 @@
...
+    }
+  }
+
+  property("Empty HashMap add consistent with contains") {
+    forAll { (i: Int, s: String) =>
+      assert(HashMap.empty[Int, String].add(i, s).contains(i))
+    }
+  }
+
+  property("Empty HashMap add and remove consistent with contains") {
```
johnynek, line 146: I think these tests would be better by generating a random HashMap from a closed set of operations:

```scala
def genHashMap[K, V](genK: Gen[K], genV: Gen[V])(implicit hashK: Hash[K]): Gen[HashMap[K, V]] =
  ...
  either:
  HashMap.empty,
  HashMap.fromFoldable,
  recurse.updated(k, v)
  recurse.remove(k)
```
So, basically randomly generate a hashmap by randomly doing all of the base operations we expose.

tests/shared/src/test/scala/cats/tests/HashMapSuite.scala
```diff
@@ -0,0 +1,292 @@
...
+  }
+
+  property("remove consistent with contains") {
+    forAll { (kvs: List[(Int, String)]) =>
+      val distinctKvs = distinctBindings(kvs)
+      val hashMap = HashMap.fromSeq(kvs)
+
+      distinctKvs.foldLeft(hashMap) { case (hm, (i, _)) =>
+        if (!hm.contains(i)) {
+          println(hm)
```
johnynek, line 169: can we remove `println`? when shrinking this will be really annoying.

tests/shared/src/test/scala/cats/tests/HashMapSuite.scala
```diff
@@ -0,0 +1,292 @@
...
+    forAll { (kvs: List[(Int, String)]) =>
+      val hashMap = HashMap.fromSeq(kvs)
+
+      kvs.foreach { case (k, _) =>
+        assert(hashMap.contains(k))
+      }
+    }
+  }
+
+  property("remove consistent with contains") {
```
johnynek, line 162: I think these tests may not be needed if we actually generate as I propose above, then we just do something like:

```scala
forAll { (h: HashMap[Int, String], key: Int) =>
  val h1 = h.remove(key)
  assert(!h1.contains(key))
  if (h.contains(key)) {
    assert(h1.size == (h.size - 1))
  }
  else {
    assert(h1 === h)
  }
}
```

tests/shared/src/test/scala/cats/tests/HashMapSuite.scala
```diff
@@ -0,0 +1,292 @@
...
+        assert(hm.contains(i))
+        assert(!hm.remove(i).contains(i))
+        hm.remove(i)
+      }
+
+      ()
+    }
+  }
+
+  property("add and remove with collisions consistent with contains") {
```
johnynek, line 182: you can actually nest forAll, so we could do this:

```scala
def genKey[K](map: HashMap[K, Any])(implicit k: Gen[K]): Gen[K] =
  if (map.isEmpty) k
  else Gen.oneOf(
    Gen.oneOf(map.keys.toSeq),
    k
  )

forAll { (h: HashMap[Int, String]) =>
  forAll(genKey(h)) { (key: Int) =>
    // now key is a collision half the time so we really exercise both the collision and non collision cases.
  }
}
```

tests/shared/src/test/scala/cats/tests/HashMapSuite.scala
```diff
@@ -0,0 +1,292 @@
...
+  }
+
+  property("iterator consistent with reverseIterator") {
+    forAll { (kvs: List[(Int, String)]) =>
+      val hashMap = HashMap.fromSeq(kvs)
+
+      val iterated = mutable.ListBuffer[(Int, String)]()
+      val reverseIterated = mutable.ListBuffer[(Int, String)]()
+      hashMap.iterator.foreach { iterated += _ }
+      hashMap.reverseIterator.foreach { reverseIterated += _ }
```
johnynek, line 236: I wonder why we have `reverseIterator` when the order is not a contract we can give. I wonder if we should just remove this.

johnynek (changes requested):

core/src/main/scala/cats/data/HashMap.scala
johnynek: I guess the front half is K, V pairs and the back half are the nodes, and the are reflected about the middle?

johnynek:
I think if we beef up the tests as I sketched, I am a champion of merging this.

I think Map APIs are pretty much settled in scala and we can use the existing ones with the natural generalization of SortedMap, Ordering to HashMap, Hash (which is mostly what was done here modulo some add renaming to updated).

I think having a HashMap to work with Hash would be a really nice addition to make Hash actually safe to use, and without it, I Hash is pretty uninteresting (except that other users might internally implement their own hashmaps based on it).

johnynek:
Can you post the benchmark results as a comment and discuss a comparison to the universal hash equivalents (which I assume is very hard to beat, but hopefully we could nearly match).

armanbilge:
> I think having a HashMap to work with Hash would be a really nice addition to make Hash actually safe to use, and without it, I Hash is pretty uninteresting

I think we will also need to revoke this law:

https://github.com/typelevel/cats/blob/3314acd0e921140d939aafffd745a98d55556814/kernel-laws/shared/src/main/scala/cats/kernel/laws/HashLaws.scala#L33-L35

johnynek:
I tend to agree. We can possibly make two sets of laws: `UniversalLikeHash` and `GeneralHash` or something.... Sometimes we want to test that we have just done the same as universal hash would do (e.g. on primitives or built in collections).

But something is nagging in the back of my mind: `Hash` and `Show` are kind of special since the JVM does basically promise `Any` can implement them. So, one can argue the existence of them as typeclasses in cats isn't the same utility as in a language without that.

Moreover, you could argue that what `Hash[K]` is really doing is promising that the type `K` is a "data" value recursively, and it never is hitting `System.identityHashCode`. In that way, every implementation of `Hash[K]` is just `.hashCode` BUT we promise NOT to implement `Hash[K]` for a type without a reproducible hash. In other words, there is no `Hash[Int => Int]`. So, we can only hash value types all the way down.

If we take that view, we can keep the law, but still Hash is useful because it proves you have a reproducible hashcode all the way down, no hiding of general `new AnyRef` or function, or typeclass values, etc...

armanbilge:
> If we take that view, we can keep the law, but still Hash is useful because it proves you have a reproducible hashcode all the way down, no hiding of general `new AnyRef` or function, or typeclass values, etc...

Putting aside the fact this wouldn't allow for alternative `Hash` implementations for `opaque` newtypes, it's a reasonable perspective. But also then, does it obviate the need for our own hashmap implementation? `Hash` would provide the evidence you can create a hashmap, but we could continue to use the scala collections implementation.

johnynek:
I think `opaque type` is maybe the best counter example to this, but still, I think it has `hashCode` since it extends `Any` right?

The utility of `Hash[K]` in this case is simply the promise that `System.identityHashCode` will not be called. You can imagine a name like: `NotIdentityHashCode[K]` as the typeclass. So, you could use `Map[K, V]` then, but you can't prove to someone else that inside aren't things with identity hashCode. By using `HashMap` you know you have hashmap that is pure (and not a function of memory location on the process you are running in).

Like, put another way: when would you not want to match `.hashCode` and why wouldn't you?

I think we may have gotten tripped up and imagined the law means it is 1:1, but it doesn't mean that: all `Hash[K]` is == `.hashCode` but not all things with `.hashCode` has a `Hash[K]`.

E.g. there would not be a `Hash[Any]` because that may wind up calling `identityHashCode`.

armanbilge:
> I think `opaque type` is maybe the best counter example to this, but still, I think it has `hashCode` since it extends `Any` right?

Yes, it certainly does have `hashCode`. The problem is suppose you want to implement a different `Eq` and `HashCode` than the underlying type e.g. `opaque type CIString = String` for case-insensitive `String`s.

> So, you could use `Map[K, V]` then, but you can't prove to someone else that inside aren't things with identity hashCode.

Sorry, I'm not entirely sure I follow 😕 I guess my point is, if we know that `Hash[A}.hash(a) == a.hashCode` then we can delegate the implementation to a Scala `Map`. For sure, we should wrap this up in our own safe newtype (such as we for `NonEmptyList` and friends) that requires evidence in the form of `Hash[K]`, but there's no need to actually use that for the implementation.

johnynek:
You seem to be arguing two contradictory points:

1. CIString is an excellent example, if you have that and want to not have to normalize the case, then you can't have `Hash[CIString]` since the `Eq[CIString]` needs to not care about case. So, the law is broken here, or you simply cannot implement `Hash[CIString]`. This tends to argue to remove the law.
2. If we do not remove the law, we don't need to copy `Map[K, V]` we could just make a newtype on `(Map[K, V], Hash[K])` if we have access to `Hash[K]` and it would be identical to the current implementation.

So, going down route 2. seems to limit `Hash[K]` especially in the world of newtypes (not just opaque types, also path dependent types). Thus, I would tend to agree we need to remove the law for matching universal equality (although in practice it will probably only impact corner cases, since when possible it is probably a good idea to simply make them match).

joroKr21:
Note that Scala has `.##` and the `sameAsUniversalHash` is already broken when `Long` is involved: #2878
See also the commend from @ctongfei about the original purpose of that law.
On another hand sometimes I do want to use `System.identityHashCode` so it seems a weird thing to guarantee that it wouldn't be called. I see it more as a guide - the typeclass nudges you to think about what hash code you want to use.

kernel/src/main/scala-2.13+/cats/kernel/compat/HashCompat.scala
```diff
@@ -100,6 +100,28 @@ private[kernel] class HashCompat {
...
+    var a, b, n = 0
+    var c = 1
+    val iterator = xs.iterator
+    while (iterator.hasNext) {
+      val x = iterator.next()
+      val h = A.hash(x)
+      a += h
+      b ^= h
+      c *= h | 1
+      n += 1
```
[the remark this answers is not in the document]
DavidGregory084: The code was copied almost verbatim from the 2.13.x standard library and the `TraversableOnce` variant from 2.12.x, as with the other snippets in these two files, but we can absolutely retrofit the improvements from 2.13.x if we're convinced they behave the same way.

tests/shared/src/test/scala/cats/tests/HashMapSuite.scala
DavidGregory084: I didn't realise I left this there. I sometimes wonder if there should be a toggleable compiler warning for `println`

tests/shared/src/test/scala/cats/tests/HashMapSuite.scala
DavidGregory084: In the standard library the `reverseIterator` is used to implement operations like `init` and `last` that don't naturally make a lot of sense for a HashMap, so there may not be any point in keeping it in cats.

core/src/main/scala/cats/data/HashMap.scala
DavidGregory084: I took the lead here from `Chain` which uses the verb `append` instead of the stdlib symbolic `+` or past tense `appended`. Definitely happy to change it 👍

DavidGregory084: I renamed this to `updated` and also `remove` to `removed` for consistency with the 2.13.x standard library maps

core/src/main/scala/cats/data/HashMap.scala
DavidGregory084: Renamed this to `allElementsCount`, and I renamed the other size fields to `keyValueCount` and `nodeCount` as well

tests/shared/src/test/scala/cats/tests/HashMapSuite.scala
DavidGregory084: I did something slightly different in the end by using the test structure that you describe, but permuting an existing key to produce a colliding string instead of an exact match. Let me know what you think!

core/src/main/scala/cats/data/HashMap.scala
DavidGregory084: I have added both `fromFoldable` and `fromIterableOnce` in addition to `fromSeq` for consistency with `Chain`

DavidGregory084:
I think I have addressed your comments @johnynek. I am just rerunning the benchmarks after today's changes and then I will post the results.

DavidGregory084:
I'm utterly bamboozled by the test failure of the "concat" test - it seems to intermittently fail, but only if the Scalacheck test parameters specify more than one worker. Has anyone seen anything like that before?

johnynek:
@DavidGregory084 are you saying that if you use these literal values:
```
  > ARG_0: HashMap(1261272920 -> 岿ⷩ㔿蔘釙ᔴ㵍䊉䕅뺷꠴)
  > ARG_1: HashMap(-2052148147 -> 潳趆冃쩑绗孇)
```
or at least the same keys, the law passes? but somehow there is something with concurrency that is causing the issue?

I've never seen that without some kind of mutable state hiding in an `object` somewhere.

DavidGregory084:
@johnynek yes that's right
```scala
  test("concat") {
    forAll { (l: HashMap[Int, String], r: HashMap[Int, String]) =>
      // val left = l
      // val right = r
      val left = HashMap(1261272920 -> Array(61738, 23743, 11753, 13631, 34072, 37337, 5428, 15693, 59995, 17033, 17733, 48823, 43060).map(_.toChar).mkString)
      val right = HashMap(-2052148147 -> Array(28531, 36230, 20867, 58606, 51793, 32471, 23367).map(_.toChar).mkString)
      val leftKvs = left.iterator.toList
      val rightKvs = right.iterator.toList
      val distinctKvs = distinctBindings(leftKvs ++ rightKvs)
      val both = left.concat(right)
      assert(both === HashMap.fromSeq(leftKvs ++ rightKvs))
      assert(both.size === distinctKvs.size)
      distinctKvs.foreach { case (k, v) =>
        assert(both.contains(k))
        assert(both.get(k) === Some(v))
        if (right.contains(k))
          assert(both.get(k) === right.get(k))
        else
          assert(both.get(k) === left.get(k))
      }
    }
  }
```
I cannot seem to cause `testsJVM/testOnly cats.tests.HashMapSuite` to fail here unless I use the parameters from `forAll`. It fails intermittently if `l` or `r` is used instead of the hardcoded values and even more frequently if both `l` and `r` are used.

johnynek:
well, the first thing I'd do is probably change as many asserts I could to `assertEquals` so we can see what is actually different. That might give us some clue.

There is internal mutability in this instances. I assume that we have a real bug and that munit is helping us find one. In fact, I would look at turning up the number of instances and we might find the problem faster.

It could be that sometimes we generate a hashmap with a broken invariant, and writing it literally doesn't have that issue. That is my guess.

core/src/main/scala/cats/data/HashMap.scala
```diff
@@ -0,0 +1,1200 @@
...
+
+import cats.kernel.compat.scalaVersionSpecific._
+
+import cats.Always
+import cats.CommutativeApplicative
+import cats.Eval
+import cats.Foldable
+import cats.Semigroup
+import cats.Show
+import cats.UnorderedTraverse
```
johnynek, line 52: nit: can we use:

```scala
import cats.{Always, CommutativeApplicative, Eval, Foldable, Semigroup, Show, UnorderedTraverse}
```

core/src/main/scala/cats/data/HashMap.scala
```diff
@@ -0,0 +1,1200 @@
...
+import cats.kernel.CommutativeMonoid
+import cats.kernel.CommutativeSemigroup
+import cats.kernel.Monoid
+import cats.kernel.Eq
+import cats.kernel.Hash
```
johnynek, line 57: same comment about using `cats.kernel.{ ...`

core/src/main/scala/cats/data/HashMap.scala
```diff
@@ -0,0 +1,1200 @@
...
+    fromSeq(kvs)
+
+  /**
+  * Creates a new [[cats.data.HashMap]] which contains all elements of `seq`.
+  *
+  * @param seq the sequence of elements to add to the [[cats.data.HashMap]].
+  * @param hashKey the [[cats.kernel.Hash]] instance used for hashing values.
+  * @return a new [[cats.data.HashMap]] which contains all elements of `seq`.
+  */
+  final def fromSeq[K, V](seq: Seq[(K, V)])(implicit hashKey: Hash[K]): HashMap[K, V] = {
```
johnynek, line 270: why not call `fromFoldable` here? There is a `Foldable[Seq]` no?

core/src/main/scala-2.12/cats/compat/HashMapCompat.scala
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  /**
+    * Creates a new map with all key-value pairs of this map, and all key-value pairs of `map`.
+    *
+    * @param traversable the collection of key-value pairs to be added.
+    * @return a new map that contains all key-value pairs of this map and `traversable`.
+    */
+  final def concat[VV >: V](hm: HashMap[K, VV]): HashMap[K, VV] = {
+    val newRootNode = hm.iterator.foldLeft(self.rootNode: HashMap.Node[K, VV]) { case (node, (k, v)) =>
```
johnynek, line 52: can't we call `concat(hm.iterator)` here and remove the duplication?

core/src/main/scala/cats/data/NonEmptyCollection.scala
```diff
@@ -40,6 +40,7 @@ private[cats] trait NonEmptyCollection[+A, U[+_], NE[+_]] extends Any {
   def filter(f: A => Boolean): U[A]
   def filterNot(p: A => Boolean): U[A]
   def collect[B](pf: PartialFunction[A, B]): U[B]
+  def collectFirst[B](pf: PartialFunction[A, B]): Option[B]
```
johnynek, line 43: did we add new tests for this? I don't see it.

core/src/main/scala/cats/data/HashMap.scala
```diff
@@ -0,0 +1,1200 @@
...
+    * @param key the key to be added.
+    * @param value the value to be added.
+    * @return a new map that contains all key-value pairs of this map and that also contains a mapping from `key` to `value`.
+    */
+  final def updated[VV >: V](key: K, value: VV): HashMap[K, VV] = {
+    val keyHash = improve(hashKey.hash(key))
+    val newRootNode = rootNode.updated(key, keyHash, value, 0)
+
+    if (newRootNode eq rootNode)
+      this
```
johnynek, line 163: how could this branch ever happen? I don't think it can. We aren't taking an `Eq[VV]` so we can't tell if this value is already inside, so this branch must (or should) never be taken.

Can we remove it?

tests/shared/src/test/scala/cats/tests/HashMapSuite.scala
```diff
@@ -0,0 +1,253 @@
...
+    }
+  }
+
+  property("size consistent with foreach") {
+    forAll { (hashMap: HashMap[Int, String]) =>
+      var size = 0
+      hashMap.foreach { case _ => size += 1 }
+      assert(hashMap.size === size)
+    }
+  }
```
johnynek, line 252: another test
```scala
forAll { (h1: HashMap[Int, String], h2: HashMap[Int, String]) =>
  if (h1.toString == h2.toString) assertEquals(h1, h2)
  if (h1.show === h2.show) assert(h1 === h2)
}
```
this could help us see if there is something weird going on where two maps print the same, but are actually different.

core/src/main/scala/cats/data/HashMap.scala
DavidGregory084: Yes, you're right - in one version of this code I did experiment with a `Eq[VV]` parameter but we always return a new instance now

DavidGregory084:
> It could be that sometimes we generate a hashmap with a broken invariant, and writing it literally doesn't have that issue. That is my guess.

Yes, I think you are right - the thing that is surprising to me is that it's intermittent, even for the same Scalacheck initial seed. Am I misunderstanding or are generators less deterministic than I thought?

Here's the core of the problem I think:

```scala
left = BitMapNode(keyValueMap=00000000000000000000000000000100, nodeMap=00000000000000000000000000000000, size=1, contents=[1261272920, 岿ⷩ㔿蔘釙ᔴ㵍䊉䕅뺷꠴])
right = BitMapNode(keyValueMap=00000000000000100000000000000000, nodeMap=00000000000000000000000000000000, size=1, contents=[-2052148147, 潳趆冃쩑绗孇])
both = BitMapNode(keyValueMap=00001000000000000000000000000100, nodeMap=00000000000000000000000000000000, size=2, contents=[1261272920, 岿ⷩ㔿蔘釙ᔴ㵍䊉䕅뺷꠴, -2052148147, 潳趆冃쩑绗孇])
```

Either the `right` or `both` `keyValueMap` bitmap is wrong here. Will dig into this some more tomorrow!

core/src/main/scala/cats/data/NonEmptyCollection.scala
DavidGregory084: Yep you're right, I missed this one - added in 5c15c530a69d2c718ffdcb9850f88f7e1789445d

core/src/main/scala-2.12/cats/compat/HashMapCompat.scala
johnynek: actually, thinking about this more, it seems a shame if this is the fastest way to go.

For instance:

1. if there are no keys in common we could iterate over the smaller of the two.
2. it seems potentially there could be ways of merging the internal structure (e.g. looking at the bitmaps..., I don't have a concrete suggestion here, but just that it seems like it could be possible for a much faster merge operation).

DavidGregory084: >  2. it seems potentially there could be ways of merging the internal structure (e.g. looking at the bitmaps..., I don't have a concrete suggestion here, but just that it seems like it could be possible for a much faster merge operation).

The standard library HashMap definitely has a [much more involved concat operation](https://github.com/scala/scala/blob/7952071b88a85e2502e9ecd4e8bd57040d9448f5/src/library/scala/collection/immutable/HashMap.scala#L1309) than I have attempted here. I'm not sure whether it makes sense to try that now so that we don't have to worry about how to make binary compatible changes to `Node` later or whether it's better to ship a definitely correct but slow version first.

core/src/main/scala/cats/data/HashMap.scala
DavidGregory084: Perhaps we should set up https://github.com/liancheng/scalafix-organize-imports in this repo to enforce a consistent style

tests/shared/src/test/scala/cats/tests/HashMapSuite.scala
DavidGregory084: In fact this would probably have caught that deletion bug even more easily

core/src/main/scala/cats/data/HashMap.scala
DavidGregory084: This is a good observation actually - this currently accepts a (potentially mutable) `scala.collection.Seq` on 2.12, which has no `Foldable` instance.

Do we keep the implementation of `fromSeq` here or restrict it to `collection.immutable.Seq`?

core/src/main/scala-2.12/cats/compat/HashMapCompat.scala
DavidGregory084: I went with your initial suggestion for now - I need to spend some time to understand the standard library `concat` implementation.

johnynek:
pending running the benchmarks, I'm in support of a merge!

bench/src/main/scala-2.12/cats/bench/HashMapBench.scala
```diff
@@ -0,0 +1,192 @@
...
+    bh.consume(hs)
+  }
+
+  @Benchmark
+  @OperationsPerInvocation(1000)
+  def scalaMapUpdated(bh: Blackhole): Unit = {
+    var ss = scalaMap
+    var i = 0
+    while (i < 1000) {
+      ss += (-i.toLong -> i)
```
johnynek, line 79: I think this may be unfair to scala. They have to allocate a tuple here they may not be able to reuse.

Can we do `ss = ss.updated(-i.toLong, i)`

core/src/main/scala/cats/data/HashMap.scala
```diff
@@ -0,0 +1,1199 @@
...
+    extends HashMapCompat[K, V] {
+
+  /**
+    * An iterator for this map that can be used only once.
+    *
+    * @return an iterator that iterates through the key-value pairs of this map.
+    */
+  final def iterator: Iterator[(K, V)] =
+    new HashMap.Iterator(rootNode)
+
```
johnynek, line 84: could we add:
```scala
final def toMap: Map[K, V] =
  // add a *wrapper* not a copy of the `HashMap`
```
I think you only need to implement like 4 methods to implement a `Map` and I think you have almost all of them already.

core/src/main/scala/cats/data/HashMap.scala
```diff
@@ -267,7 +269,41 @@ object HashMap extends HashMapInstances {
...
+  /**
+  * Creates a new [[cats.data.HashMap]] which contains all elements of `fkv`.
+  *
+  * @param fkv the [[cats.Foldable]] structure of elements to add to the [[cats.data.HashMap]].
+  * @param F the [[cats.Foldable]] instance used for folding the structure.
+  * @param hashKey the [[cats.kernel.Hash]] instance used for hashing values.
+  * @return a new [[cats.data.HashMap]] which contains all elements of `seq`.
+  */
+  final def fromFoldable[F[_], K, V](fkv: F[(K, V)])(implicit F: Foldable[F], hashKey: Hash[K]): HashMap[K, V] = {
```
DavidGregory084, line 300: I guess we could weaken this constraint to `UnorderedFoldable`

johnynek: yeah, the order matters because you have to pick one of the values in the case of key collision.

core/src/main/scala/cats/data/HashMap.scala
DavidGregory084: I had a crack at this in https://github.com/typelevel/cats/pull/4193/commits/0f663332f848ae4720513bbf57a6c4ea481ae2b0, let me know what you think!

DavidGregory084:
OK, finally I have some benchmark results for 2.12:
https://gist.github.com/DavidGregory084/597fb4504cd4d3a9bb37ada997767c44
And 2.13:
https://gist.github.com/DavidGregory084/ad835ea3c08f3421e638713a2a967587
I'm afraid that the benchmarks for 3.x are still running.

Please bear in mind that these benchmarks ran against 5c15c530a69d2c718ffdcb9850f88f7e1789445d and there have been a few changes since then.

DavidGregory084:
Looking at 2.12 we are in the ballpark of the standard library Map with a few exceptions: 
concat - The standard library has better performance especially as size increases, which we would expect given the standard library Map has a more efficient implementation of concat.
fromSeq - The standard library has slightly better performance especially as size increases, which we would expect since the standard library has an efficient builder for maps.
updated - The standard library has better performance here too - I'm not exactly sure why at the moment but I have seen some code in the standard library that does updates using shallow mutation so it's possibly related to that.

DavidGregory084:
Looking at 2.13 - much better concat, fromSeq, performance in the stdlib as before. The standard library iterator seems to exhibit better performance here too for reasons I haven't yet dug into.
Updating entries is a similar story to 2.12, stdlib is slightly better all across the range of map sizes but especially for very large maps.
Removal performance seems similar except that the standard library exhibits better performance for very large maps. It's possible that the standard library performance is being underestimated here due to the benchmark running before https://github.com/typelevel/cats/pull/4193/commits/ea5c7de0ba82e52724d93b34ba6295e9dcfe032f was applied.

DavidGregory084:
Results for 3.0.x: https://gist.github.com/DavidGregory084/564192f04381fcb9d34769d3782465d9

johnynek:
I know this may be a lot to ask, but could you interleave the results so that we can compare them test by test. Having to scroll through to find the matching one is very hard. Or maybe we can find some graphing tool that will help us?

DavidGregory084:
@johnynek yep sure I will have a crack at it!

DavidGregory084:
What I should have done: reordered the benchmark data in Excel
What I actually did: meddled with a million different Scala charting libraries 🤯 

I've reordered the data in the gists now 😆

DavidGregory084:
Benchmark charts for 2.12:
hashmap-benchmark-2 12-concat
hashmap-benchmark-2 12-contains
hashmap-benchmark-2 12-eqEquals
hashmap-benchmark-2 12-foreach
hashmap-benchmark-2 12-fromSeq
hashmap-benchmark-2 12-get
hashmap-benchmark-2 12-iterator
hashmap-benchmark-2 12-removed
hashmap-benchmark-2 12-universalEquals
hashmap-benchmark-2 12-updated

DavidGregory084:
For 2.13:
hashmap-benchmark-2 13-concat
hashmap-benchmark-2 13-contains
hashmap-benchmark-2 13-eqEquals
hashmap-benchmark-2 13-foreach
hashmap-benchmark-2 13-fromSeq
hashmap-benchmark-2 13-get
hashmap-benchmark-2 13-iterator
hashmap-benchmark-2 13-removed
hashmap-benchmark-2 13-universalEquals
hashmap-benchmark-2 13-updated

DavidGregory084:
For 3.0:
hashmap-benchmark-3-concat
hashmap-benchmark-3-contains
hashmap-benchmark-3-eqEquals
hashmap-benchmark-3-foreach
hashmap-benchmark-3-fromSeq
hashmap-benchmark-3-get
hashmap-benchmark-3-iterator
hashmap-benchmark-3-remove
hashmap-benchmark-3-removed
hashmap-benchmark-3-universalEquals
hashmap-benchmark-3-updated

DavidGregory084:
Plus if anyone would like the scala-cli script I used to generate these in order to meddle with the chart rendering, it's here 🙈  https://gist.github.com/DavidGregory084/03d617d978a8db6a3efd0d478a44e657

johnynek:
Very nice! Seems like concat and updated are the current weakest spots.

core/src/main/scala/cats/data/HashMap.scala
```diff
@@ -0,0 +1,1211 @@
...
+    final private def hasKeyValueAt(bitPos: Int): Boolean =
+      (keyValueMap & bitPos) != 0
+
+    final def getKey(index: Int): K =
+      contents(Node.StrideLength * index).asInstanceOf[K]
+
+    final def getValue(index: Int): V =
+      contents(Node.StrideLength * index + 1).asInstanceOf[V]
+
+    final def getMapping(index: Int): (K, V) = {
```
johnynek, line 636: I fear this method is hurting perf.

Can we remove it? We can avoid the allocation in the cases we don't need it if we use `getKey` and `getValue`.

core/src/main/scala/cats/data/HashMap.scala
```diff
@@ -0,0 +1,1211 @@
...
+    }
+
+    final def getNode(index: Int): Node[K, V] =
+      contents(contents.length - 1 - index).asInstanceOf[Node[K, V]]
+
+    final def foreach[U](f: (K, V) => U): Unit = {
+      var i = 0
+      val fnTupled = f.tupled
+      while (i < keyValueCount) {
+        fnTupled(getMapping(i))
```
johnynek, line 649: how about instead: 

```scala
while (i < keyValueCount) {
  f(getKey(i), getValue(i))
}
```
to avoid tupling entirely.

core/src/main/scala/cats/data/HashMap.scala
```diff
@@ -0,0 +1,1211 @@
...
+
+    final private def updateKeyValue[VV >: V](
+      bitPos: Int,
+      newKey: K,
+      newKeyHash: Int,
+      newValue: VV,
+      depth: Int
+    ): Node[K, VV] = {
+      val index = Node.indexFrom(keyValueMap, bitPos)
+      val (existingKey, existingValue) = getMapping(index)
```
johnynek, line 795: how about:

```scala
val existingKey = getKey(index)
val existingValue = getValue(index)
```

core/src/main/scala/cats/data/HashMap.scala
```diff
@@ -0,0 +1,1211 @@
...
+
+        new BitMapNode[K, V](newKeyValueMap, nodeMap, newContents, size - 1)
+      }
+    }
+
+    final private def inlineSubNodeKeyValue[VV >: V](bitPos: Int, newSubNode: Node[K, VV]): Node[K, VV] = {
+      val nodeIndex = contents.length - 1 - Node.indexFrom(nodeMap, bitPos)
+      val keyIndex = Node.StrideLength * Node.indexFrom(keyValueMap, bitPos)
+      val newContents = new Array[Any](contents.length + 1)
+      val (key, value) = newSubNode.getMapping(0)
```
johnynek, line 875: how about:
```scala
val key = newSubNode.getKey(0)
val value = newSubNode.getValue(0)
```

core/src/main/scala/cats/data/HashMap.scala
```diff
@@ -0,0 +1,1211 @@
...
+
+    final override def ===[VV >: V](that: Node[K, VV])(implicit eqValue: Eq[VV]): Boolean = {
+      (this eq that) || {
+        that match {
+          case node: BitMapNode[_, _] =>
+            (this.keyValueMap === node.keyValueMap) &&
+            (this.nodeMap === node.nodeMap) &&
+            (this.size === node.size) && {
+              var i = 0
+              while (i < keyValueCount) {
```
johnynek, line 941: how about:
```scala
if (hashKey.neqv(getKey(i), node.getKey(i)) || eqValue.neqv(getValue(i), node.getValue(i))) return false
```

core/src/main/scala/cats/data/HashMap.scala
```diff
@@ -0,0 +1,1211 @@
...
+
+      foundMoreValues
+    }
+
+    final override def hasNext: Boolean =
+      (currentValuesIndex < currentValuesLength) || getMoreValues()
+
+    final override def next(): (K, V) = {
+      if (!hasNext) throw new NoSuchElementException
+      val value = currentNode.getMapping(currentValuesIndex)
```
johnynek, line 1124: I think this is literally the only place where we don't destructure after getting the key and value.

Probably better to just allocate the tuple here where it actually needed.

core/src/main/scala/cats/data/HashMap.scala
```diff
@@ -0,0 +1,1211 @@
...
+  implicit def catsDataMonoidForHashMap[K: Hash, V: Semigroup]: Monoid[HashMap[K, V]] =
+    new HashMapMonoid[K, V]
+}
+
+class HashMapMonoid[K: Hash, V](implicit V: Semigroup[V]) extends Monoid[HashMap[K, V]] {
+
+  def empty: HashMap[K, V] = HashMap.empty[K, V]
+
+  def combine(xs: HashMap[K, V], ys: HashMap[K, V]): HashMap[K, V] = {
+    if (xs.size <= ys.size) {
```
johnynek, line 1195: this should be the same code as `concat` with `V: Semigroup` being `(x, y) => y`

Can we inline this with `concat` to get the benefit of only iterating the smaller item (which we don't do in concat).

johnynek (commented):
seeing the benchmarks made me want to stab at removing allocations and see if that could speed things up.

I have a few suggestions I think could work (I think the duplication in updated, i.e. avoiding calling contains, is probably the most promising).

core/src/main/scala/cats/data/HashMap.scala
DavidGregory084: It feels like we could extract this pattern as something like:

```scala
  def combineWith(xs: HashMap[K, V], ys: HashMap[K, V])(f: (HashMap.Node[K, V], (K, V)) => HashMap.Node[K, V]): HashMap[K, V] = {
    val newRootNode = if (xs.size <= ys.size) {
      xs.iterator.foldLeft(ys.rootNode) { case (node, kv @ (k, x)) =>
        f(node, kv)
      }
    } else {
      ys.iterator.foldLeft(xs.rootNode) { case (node, kv @ (k, y)) =>
        f(node, kv)
      }
    }

    if (newRootNode eq xs.rootNode)
      xs
    else if (newRootNode eq ys.rootNode)
      ys
    else
      new HashMap(newRootNode)
  }
```
No idea whether it's small enough to get JITted or how much the extra branch would hurt performance though

DavidGregory084: There's one other subtlety to this change - it breaks the behaviour of `concat` a bit because it changes which value is kept when the key exists in both maps.
I think that the scala collections work around this by adding a boolean flag to `updated` which determines whether the existing value should be replaced.

DavidGregory084: I have added a boolean parameter `replaceExisting` to `Node#updated`. We can use this to keep the existing key when iterating over the "rightmost" HashMap in the various combining operations.

DavidGregory084:
Benchmarks are running again now 👀

core/src/main/scala/cats/data/HashMap.scala
```diff
@@ -0,0 +1,1211 @@
...
+      Integer.bitCount(keyValueMap)
+
+    final def nodeCount: Int =
+      Integer.bitCount(nodeMap)
+
+    final private def hasNodeAt(bitPos: Int): Boolean =
+      (nodeMap & bitPos) != 0
+
+    final private def hasKeyValueAt(bitPos: Int): Boolean =
+      (keyValueMap & bitPos) != 0
```
DavidGregory084, line 649: I wonder if it might be worth adding an inline annotation to some of these methods

DavidGregory084:
@johnynek how about if we go for the most obvious approach to implementing `concat` for `BitMapNode`, something like this?

```scala
    final def concat[VV >: V](that: Node[K, VV], depth: Int): Node[K, VV] = that match {
      case bitmap: BitMapNode[_, _] =>
        var index = 0
        var thisNodeMap = this.nodeMap
        var thatNodeMap = bitmap.nodeMap
        while (index < Integer.SIZE) {
          val thisHasNode = (thisNodeMap & 1) == 1
          val thatHasNode = (thatNodeMap & 1) == 1
          if (thisHasNode && thatHasNode) {
            // concat the subnodes at depth + 1 and update this
          } else if (thatHasNode) {
            // add subnode to this
          }

          thisNodeMap >>= 1
          thatNodeMap >>= 1
        }

        var thisKeyValueMap = this.keyValueMap
        var thatKeyValueMap = bitmap.keyValueMap
        index = 0
        while (index < Integer.SIZE) {
          val thisHasKeyValue = (thisKeyValueMap & 1) == 1
          val thatHasKeyValue = (thatKeyValueMap & 1) == 1
          if (thisHasKeyValue && thatHasKeyValue) {
            // collision, create a subnode using mergeValuesIntoNode
          } else if (thatHasKeyValue) {
            // add subnode to this node
          }

          thisKeyValueMap >>= 1
          thatKeyValueMap >>= 1
        }

        this

      case _: CollisionNode[_, _] =>
        throw new IllegalStateException("Attempting to merge a bitmap node with a collision node") // this means we have broken the invariant that a CollisionNode must only exist at Node.MaxDepth + 1
    }
```

DavidGregory084:
I'm sure there is something clever we can do by XORing the bitmaps or something like that but it eludes me 😛

DavidGregory084:
I've implemented something like what I described above, very fiddly, would love some feedback and extra eyes!

DavidGregory084:
[Benchmark results](https://gist.github.com/DavidGregory084/5eb4de9940f4cd188587cdb535540680) for 2.13 at ea31a5d:
hashmap-benchmark-2 13-concat
hashmap-benchmark-2 13-contains
hashmap-benchmark-2 13-eqEquals
hashmap-benchmark-2 13-foreach
hashmap-benchmark-2 13-fromSeq
hashmap-benchmark-2 13-get
hashmap-benchmark-2 13-iterator
hashmap-benchmark-2 13-removed
hashmap-benchmark-2 13-universalEquals
hashmap-benchmark-2 13-updated

DavidGregory084:
Those results for `concat` look super encouraging even at ea31a5d, although it is making me wonder whether we should have some more varied `concat` benchmarks that test concatenating differently-sized maps.

DavidGregory084:
I wonder if the performance of the scala collections `Map` is better for `updated` because the collections can check for equality of the value and decide to do nothing if the new value is equal to the existing one.

armanbilge:
> I wonder if the performance of the scala collections `Map` is better for `updated` because the collections can check for equality of the value and decide to do nothing if the new value is equal to the existing one.

Dumb question, why can't we do this?

DavidGregory084:
Oh actually I remember why I didn't do that originally - there's no easy way to thread the `Eq` constraint to the new value type in `UnorderedTraverse` (which is `HashMap[K, U] => (U => G[V]) => G[HashMap[K, V]]`), so we would have to give that up and supply only `UnorderedFoldable`.

DavidGregory084:
Haha OK, [e7343e8](https://github.com/typelevel/cats/pull/4193/commits/e7343e841a34917e74332cacc3e0e313e1387f33) is a total bust unfortunately 😆 

[2.12 results](https://gist.github.com/DavidGregory084/eba911d0ab9968ab274e243ddfd0d5b1):
hashmap-benchmark-2 12-concat
hashmap-benchmark-2 12-contains
hashmap-benchmark-2 12-eqEquals
hashmap-benchmark-2 12-foreach
hashmap-benchmark-2 12-fromSeq
hashmap-benchmark-2 12-get
hashmap-benchmark-2 12-iterator
hashmap-benchmark-2 12-removed
hashmap-benchmark-2 12-universalEquals
hashmap-benchmark-2 12-updated

DavidGregory084:
[2.13 results](https://gist.github.com/DavidGregory084/5d2f68685e3f683fe1b6a6245e4241e2):
hashmap-benchmark-2 13-concat
hashmap-benchmark-2 13-contains
hashmap-benchmark-2 13-eqEquals
hashmap-benchmark-2 13-foreach
hashmap-benchmark-2 13-fromSeq
hashmap-benchmark-2 13-get
hashmap-benchmark-2 13-iterator
hashmap-benchmark-2 13-removed
hashmap-benchmark-2 13-universalEquals
hashmap-benchmark-2 13-updated

DavidGregory084:
I think in order for that approach to be viable we would have to figure out how to skip a bunch more work.
At the moment the benchmark is a pathological scenario for that concat implementation as it is concatenating two hashmaps with exactly the same elements which results in exactly the same structure.
This means that every node must be merged recursively and every element copied over into `this` hashmap.

armanbilge:
So, status check on this one :) firstly, round of applause to David for this wonderful piece of work and Oscar for thorough reviews. 

I optimistically added this one to the [v2.8.0 milestone](https://github.com/typelevel/cats/milestone/41) :) besides this PR, I think we're approaching a good place to wrap up 2.8.0 and ship it in the next week or two. Besides the [accumulated changes](https://gist.github.com/armanbilge/f66fdfc371ad933edbf5e3f81e7f4116) there's been requests to release with the Scala 3 bincompat fix and Scala 3 Native.

Release planning: can this PR realistically land in 2.8.0, and should it?

Even after this lands, I think there's some follow-up work that could do well to accompany it. Off the top of my head:
- the issue with the `Hash` laws requiring consistency with the universal hash method
- there's a lot of `SortedMap`-based methods in Cats, maybe would be nice to have `HashMap` versions of them too. https://github.com/typelevel/cats/pull/4215 reminded me of this.
- a `NonEmptyHashMap`?
- the `HashSet` in https://github.com/typelevel/cats/pull/4185

For the record, none of these are blockers, but they do go nicely together.

Also, some folks were less enthusiastic about adding this to cats-core versus cats-collections. Now that it's taken shape, I'd like to hear if there are (strong) objections (personally I feel 👍 ).

Thoughts? Thanks again everyone!

DavidGregory084:
> Release planning: can this PR realistically land in 2.8.0, and should it?

Yes I think it can - the most recent commit was an attempt to improve the performance of concat but I think we can revert that commit and perhaps I can revive it as a follow-up PR.

> * there's a lot of SortedMap-based methods in Cats, maybe would be nice to have HashMap versions of them too.
https://github.com/typelevel/cats/pull/4215 reminded me of this.
> * a NonEmptyHashMap?

I think these are good suggestions but could be done as follow-up work.

> the HashSet in https://github.com/typelevel/cats/pull/4185

That PR needs some work to add a bunch more of the basic operations - at the moment it's very sparse with only incl / + and excl / -. However it would really be nice to land that as then we could add operations like `keySet` for this `HashMap`.

DavidGregory084:
@johnynek do you have any thoughts on to proceed with this one? Would you be happy to revert the most recent commit experimenting with the `concat` implementation and merge this?

@armanbilge I guess the main blocker to getting this into release is the Hash consistency law? What does deprecating a law look like?

armanbilge:
> I guess the main blocker to getting this into release is the Hash consistency law? What does deprecating a law look like?

Actually I'd say the main blocker is reviews: I am very very grateful to all the time and attention to detail you and Oscar put into this PR, since it is an immense, non-trivial contribution. It would be good to hear from a couple other Cats maintainers, especially those who were hesitant about introducing this in cats-core vs cats-collections. I don't think I'll have time to do an in-depth review, but I would like to make a pass-through the public APIs :)

Everything else I mentioned including the `Hash` laws are non-blocking nice-to-haves that can definitely be follow-ups. Although personally I believe they are stronger when shipped together, and I think a 2.9 release dedicated exactly to this would be very cool.

I am very excited about this contribution :) and I do think the sooner we can release it the better, because it will need additional ecosystem support (e.g. Circe encoders/decoders) to help with adoption.

DavidGregory084:
OK, I have reverted that commit now @armanbilge and it seems like CI is happy 🤞

armanbilge:
Thanks! I will try and carve out time for a review and follow-up with Daniel about his concerns for including this in `cats-core` instead of `cats-collections`. Would appreciate a final review(s) from Oscar. And if there's anyone else that wants chance to review this, please say so, even if you won't be able to do it right away :)

armanbilge:
So following up on this: I did ask [on Discord](https://discord.com/channels/632277896739946517/633329569402978313/990702359296475136) and Daniel and Michael chimed in that this is a much better candidate for cats-collections. My own thoughts/summary from that dialog follow. See also https://github.com/typelevel/cats/pull/4185#issuecomment-1098685816.

The main two issues seem to be:
1. The high bar for maintaining compatibility in Cats.
2. Whether collections like this fundamentally belong in `cats-core` or not.

I actually feel okay about (1) since I think this is a straightforward API and I am generally optimistic about the possibility for binary-compatible evolution.

(2) is more subjective of course. `Map`s are useful for various Cats' APIs, such as "groupBy" methods and there are already a handful of collections in `cats.data.*`.

I was also hoping this will land in `cats-core` to facilitate downstream adoption e.g. Cats Effect `MapRef`, Circe `Encoder`/`Decoder`, etc. But that in of itself is not a good reason to put something in `cats-core`.

Another strategy could be to work towards stable 1.0 of `cats-collections` with these additions, that downstreams can depend on without fear of breakage. See https://github.com/typelevel/cats-collections/issues/277. /cc @typelevel/cats-collections

Thoughts?

johnynek:
so... if @armanbilge and I (who actually have spent the time on the code) feel like it *should* be here, and folks who haven't been reviewing the code and citing specific examples of their concerns in the PR think it shouldn't I don't see why they get the veto.

gemelen:
> I think the best strategy is to make cats-collections an appealing enough dependency that can get on the classpath of e.g. circe and cats-effect-std

I'm here just to support this as much as I can, since `cats-collection` is far from being widespread (last time I researched, it has only one public project dependend on it).

DavidGregory084:
Superseded by typelevel/cats-collections#534

PR #4196, opened 2022-04-25
Add copyright notices and Scala license for code derived from the standard library

### LINKED ISSUE #4189: Ensure attribution of code adapted from Scala standard library
On #4185 @smarter raised concerns about importing code into cats that is based on code from the Scala standard library.

However as @satorg pointed out there is already code in cats which is adapted from the Scala standard library, e.g.

https://github.com/typelevel/cats/blob/fe40bc2b4961386d829c0dd594af9bd80941947d/tests/shared/src/test/scala-2.12/cats/tests/compat/SeqOps.scala

https://github.com/typelevel/cats/blob/fe40bc2b4961386d829c0dd594af9bd80941947d/kernel/src/main/scala-2.13+/cats/kernel/compat/HashCompat.scala

We should ensure that the correct attribution notices are added to such code and that the redistribution requirements for that code are met.

On #4185 it was discussed that we should include the Apache license at the root of the cats repo, and add attribution notices to each file based on the scala/scala NOTICE file, as well as ensuring that the `licences` key of `build.sbt` is updated to include the Apache 2.0 license.

I don't know if that is enough to ensure that we are doing the right thing so hopefully someone else will chime in. 😛

DavidGregory084:
Fixes #4189 by adding the Scala license to the COPYING file, and adding the suggested copyright notice to code adapted from the standard library.

### DIFF
--- a/core/src/main/scala/cats/data/Chain.scala
+++ b/core/src/main/scala/cats/data/Chain.scala
@@ -19,6 +19,26 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
+/* This file is derived in part from https://github.com/scala/scala/blob/v2.13.8/src/library/scala/collection/Iterable.scala
+ * Modified by Typelevel for redistribution in Cats.
+ *
+ * Copyright EPFL and Lightbend, Inc.
+ * Scala
+ * Copyright (c) 2002-2022 EPFL
+ * Copyright (c) 2011-2022 Lightbend, Inc.
+ *
+ * Scala includes software developed at
+ * LAMP/EPFL (https://lamp.epfl.ch/) and
+ * Lightbend, Inc. (https://www.lightbend.com/).
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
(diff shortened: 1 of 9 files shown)


rossabaker (approved):

johnynek (approved):

core/src/main/scala/cats/data/Chain.scala
```diff
@@ -659,18 +679,18 @@ sealed abstract class Chain[+A] extends ChainCompat[A] {
...
+   *
    * @param  len the test value that gets compared with the length.
    * @return a negative value if `this.length < len`,
    *         zero if `this.length == len` or
    *         a positive value if `this.length > len`.
    * @note   an adapted version of
              [[https://github.com/scala/scala/blob/v2.13.8/src/library/scala/collection/Iterable.scala#L272-L288 Iterable#sizeCompare]]
              from Scala Library v2.13.8 is used in a part of the implementation.
-   * 
+   *
```
satorg, line 693: Sorry, I completely missed this part. Seems that it is not derived from the Scala library anymore. It was so initially indeed, but then it was re-worked for the sake of optimization. Moreover, Cats has never been released with the original Scala-derived version. cc @DavidGregory084 @johnynek

Issue #4199, opened 2022-05-04
`imap(f)(g)` produces an invalid Semigroup in some cases


s5bug:
If we take the `imap` defined on `Semigroup`, associativity can be proved with `f andThen g` ≡ `identity`. There may be even more lenient cases, but I'm not aware of them (pseudo-Agda):
```agda
Invariant[Semigroup].imap (f : A → B) (g : B → A) (x : Semigroup) = record
  { combine = λ a → λ b → f (x.combine (g a) (g b))
  ; combine-assoc =
      f (x.combine (g (f (x.combine (g a) (g b)))) (g c)) ≡⟨ ? ⟩
      -- we must have a proof of `g . f ≡ id` to get to this step
      f (x.combine (x.combine (g a) (g b)) (g c))         ≡⟨ ap f x.combine-assoc ⟩
      f (x.combine (g a) (x.combine (g a) (g b)))         ≡⟨ ? ⟩
      -- we reuse that proof here
      f (x.combine (g a) (g (f (x.combine (g b) (g c))))) ∎
  }
```

When this is not the case, it is possible to produce an unlawful `Semigroup` via `imap`:

![image](https://user-images.githubusercontent.com/7751296/166626726-93c4ed4f-976c-4889-8f76-9590f0597473.png)

---

<details>
<summary>Discord conversation in Typelevel #theory channel, 2022-05-03T23:35:28.439Z to 2022-05-04T05:20:19.215Z</summary>

<ins>**armanbilge** [23:35]</ins>
question: is this a bijection?
```scala
def f[A](some: Some[A]): Option[A] = some
```
seems so, but you can't write the inverse `g: Option[A] => Some[A]` 🤔

<ins>**Adam** [23:50]</ins>
On the value level, yes. But not on the type-level.
(it's not surjective on the type level)

<ins>**armanbilge** [23:57]</ins>
thanks. I have a GADT that semantically should only be transformed via bijections. invariant seemed good, until I realized I lose `.widen` which makes it less flexible
I guess the burden is now on consuming APIs to be sufficiently generic to handle `Foo[Some[A]]` when they really only need `Foo[Option[A]]`, since it's impossible to widen your way there

<ins>**armanbilge** [00:13]</ins>
hmm, but `InvariantMonoidal#point` is defined like:
```scala
def point[A](a: A): F[A] = imap(unit)(_ => a)(_ => ())
```
maybe `imap` isn't really about bijections
or maybe that signature should return `F[a.type]`

[...]

PR #4200, opened 2022-05-04
`imap` requires an isomorphism

armanbilge:
One possible fix for https://github.com/typelevel/cats/issues/4199. It seems that at least some of our implementations of `Invariant` have been relying on this precondition anyway, so this is a way forward without retracting those instances.

### DIFF
--- a/core/src/main/scala/cats/Invariant.scala
+++ b/core/src/main/scala/cats/Invariant.scala
@@ -36,8 +36,8 @@ import scala.util.control.TailCalls.TailRec
 @typeclass trait Invariant[F[_]] extends Serializable { self =>
 
   /**
-   * Transform an `F[A]` into an `F[B]` by providing a transformation from `A`
-   * to `B` and one from `B` to `A`.
+   * Transform an `F[A]` into an `F[B]` by providing a isomorphism between `A`
+   * and `B`.
    *
    * Example:
    * {{{


johnynek (dismissed):

johnynek (commented):
Actually, can we also be explicit: we require that `f.andThen(g) == identity`.

armanbilge:
Hmm, that's a good point. IIUC requiring `f.andThen(g) == identity` is weaker than requiring an isomorphism. @s5bug's proof in https://github.com/typelevel/cats/issues/4199 actually doesn't require an isomorphism, at least for `Semigroup`.

s5bug:
`imap` generally doesn't require an isomorphism one way (see the general proof in the Discord logs).

I would keep in mind that there are more lenient cases where `imap` on `Semigroup` will still behave nicely:

image

Specifically on `Semigroup`, I believe it will behave nicely whenever `g` is a Semigroup Homomorphism (more pseudo-Agda. This time it hasn't actually been checked by the type-checker, so I could be wrong):

```agda
imap-on-semigroup-hom : {A B : Type} (x : Semigroup A) (f : A → B) (g : B → A)
                      → (fSemigroupHom : ∀ {a b : A} → (imap f g x).combine (f a) (f b) ≡ f (x.combine a b))
                      → (gSemigroupHom : ∀ {a b : B} → x.combine (g a) (g b) ≡ g ((imap f g x).combine a b))
                      → ∀ {a b c : B} (imap f g x).combine ((imap f g x).combine a b) c ≡ (imap f g x).combine a ((imap f g x).combine b c)
imap-on-semigroup-hom =
  f (x.combine (g (f (x.combine (g a) (g b)))) (g c)) ≡⟨ ap (λ l → f (x.combine l (g c))) (sym gSemigroupHom) ⟩
  f (x.combine (x.combine (g a) (g b)) (g c))         ≡⟨ ap f x.combine-assoc ⟩
  f (x.combine (g a) (x.combine (g b) (g c)))         ≡⟨ ap (λ r → f (x.combine (g a) r)) gSemigroupHom ⟩
  f (x.combine (g a) (g (f (x.combine (g b) (g c))))) ∎
```

armanbilge:
Cats also has `Invariant` instances for `Monoid` and `Group` and lots of stuff.
https://github.com/typelevel/cats/blob/5a3e17512d48979e766de56ef947b34095817656/core/src/main/scala/cats/Invariant.scala#L125

I'm not sure what the minimum requirement on `imap` is for those instances to be lawful.

s5bug:
I would expect Monoid and Group Homomorphisms likewise.

s5bug:
Actually, `* 2` isn't even a semigroup homomorphism on the product semigroup.

It is not true that `(a * 2) * (b * 2) = (a * b) * 2`, and yet `imap(identity)(_ * 2)` works anyways. So it's even more lenient than that.

armanbilge:
Right, it's sort of a balancing act :) the question here is for `Invariant` in general is what should the API contract be to permit interesting instances without being too restrictive. Basically, we're trying to make a safe abstraction.

Of course, if you _know_ you are `imap`ping a `Semigroup` or whatever specifically, then you can take advantage of whatever leniencies apply to that specific situation.

core/src/main/scala/cats/Invariant.scala
```diff
@@ -37,7 +37,7 @@ import scala.util.control.TailCalls.TailRec
 
   /**
    * Transform an `F[A]` into an `F[B]` by providing a transformation from `A`
-   * to `B` and one from `B` to `A`.
+   * to `B` and its inverse from `B` to `A` i.e. `f.andThen(g) == identity`.
```
armanbilge, line 40: @jmgimeno informs me that the category-theoretic term here is "retraction".
https://en.wikipedia.org/wiki/Section_(category_theory)

armanbilge:
Any more thoughts on this one? :)

johnynek:
The comment isn't correct as it is now.

I bet there are some instances that require the `f.andThen(g) == identity` some that require some kind of homomorphism and some uses of imap that don't require anything other than the types lining up.

I actually don't know if we know very much.

I would like to see some tests that fail to help us learn more. E.g. we know we can make some of the tests fail with semigroup + poor choices. Are there are implementations of Invariant other than the algebra cases?

s5bug:
Personally, my vote is for `imap` itself to not explicitly have a hard requirement, but explain that
1. `f >>> g ≡ identity` implies `imap(f)(g) >>> imap(g)(f) ≡ identity` (and likewise for `<<<`)
2. When used on other law-having things, incorrect usage will produce something invalid

Although, with (2), isn't the standard to put instances like that in alleycats?

armanbilge:
> some uses of imap that don't require anything other than the types lining up.

Absolutely, any `Functor` or `ContravariantFunctor` should fall into this category.

---

> Are there are implementations of Invariant other than the algebra cases?

If we look at the instances of `Invariant{,Semigroupal,Monoidal}` in its companion object, then I believe they are all `Semigroup`s: `Semigroup`, `CommutativeSemigroup`, `Monoid`, `CommutativeMonoid`, `Numeric`, `Integral`, `Fractional`, `Band`, `Semilattice`, `BoundedSemilattice`, `Group`, `CommutativeGroup`.

I stumbled on this question because I wanted to implement an `InvariantMonoidal` instance for a discrete probability measure, although actually I think this requires a bijection be passed to `imap` 🤔 

---

> Personally, my vote is for imap itself to not explicitly have a hard requirement

Do we have any lawful instances then, that don't also implement `Functor` or `ContravariantFunctor`?

armanbilge:
Relevant Twitter thread, h/t @s5bug
https://twitter.com/kmett/status/1530946205316009987

s5bug:
([Archive](https://web.archive.org/web/20220530013038/https://twitter.com/kmett/status/1530948314543407107))

PR #4209, opened 2022-05-21
Add mapAccumulate to Traverse

BalmungSan:
As promised in **discord** :)

### DIFF
--- a/core/src/main/scala/cats/Traverse.scala
+++ b/core/src/main/scala/cats/Traverse.scala
@@ -134,11 +134,18 @@ trait Traverse[F[_]] extends Functor[F] with Foldable[F] with UnorderedTraverse[
     traverse[Id, A, B](fa)(f)
 
   /**
+   * Akin to [[map]], but allows to keep track of a state value
+   * when calling the function.
+   */
+  def mapAccumulate[S, A, B](init: S, fa: F[A])(f: (S, A) => (S, B)): (S, F[B]) =
+    traverse(fa)(a => State(s => f(s, a))).run(init).value
+
+  /**
    * Akin to [[map]], but also provides the value's index in structure
    * F when calling the function.
    */
   def mapWithIndex[A, B](fa: F[A])(f: (A, Int) => B): F[B] =
-    traverse(fa)(a => State((s: Int) => (s + 1, f(a, s)))).runA(0).value
+    mapAccumulate(0, fa)((i, a) => (i + 1) -> f(a, i))._2
 
   /**
    * Akin to [[traverse]], but also provides the value's index in
@@ -200,10 +207,14 @@ object Traverse {
       typeClassInstance.sequence[G, B](self.asInstanceOf[F[G[B]]])
     def flatSequence[G[_], B](implicit ev$1: A <:< G[F[B]], G: Applicative[G], F: FlatMap[F]): G[F[B]] =
       typeClassInstance.flatSequence[G, B](self.asInstanceOf[F[G[F[B]]]])(G, F)
-    def mapWithIndex[B](f: (A, Int) => B): F[B] = typeClassInstance.mapWithIndex[A, B](self)(f)
+    def mapAccumulate[S, B](init: S)(f: (S, A) => (S, B)): (S, F[B]) =
+      typeClassInstance.mapAccumulate[S, A, B](init, self)(f)
+    def mapWithIndex[B](f: (A, Int) => B): F[B] =
+      typeClassInstance.mapWithIndex[A, B](self)(f)
     def traverseWithIndexM[G[_], B](f: (A, Int) => G[B])(implicit G: Monad[G]): G[F[B]] =
       typeClassInstance.traverseWithIndexM[G, A, B](self)(f)(G)
-    def zipWithIndex: F[(A, Int)] = typeClassInstance.zipWithIndex[A](self)
+    def zipWithIndex: F[(A, Int)] =
+      typeClassInstance.zipWithIndex[A](self)
   }
   trait AllOps[F[_], A]
       extends Ops[F, A]
--- a/core/src/main/scala/cats/data/NonEmptyList.scala
+++ b/core/src/main/scala/cats/data/NonEmptyList.scala
@@ -23,6 +23,7 @@ package cats
 package data
 
 import cats.data.NonEmptyList.ZipNonEmptyList
+import cats.instances.StaticMethods
 
 import scala.annotation.tailrec
 import scala.collection.immutable.{SortedMap, TreeMap, TreeSet}
@@ -786,6 +787,15 @@ sealed abstract private[data] class NonEmptyListInstances extends NonEmptyListIn
         fa: NonEmptyList[A]
       )(f: A => G[B])(implicit G: Applicative[G]): G[NonEmptyList[B]] =
         fa.traverse(f)
+
+      override def mapAccumulate[S, A, B](init: S, fa: NonEmptyList[A])(f: (S, A) => (S, B)): (S, NonEmptyList[B]) =
+        StaticMethods.mapAccumulateFromStrictFunctor(init, fa, f)(this)
+
+      override def mapWithIndex[A, B](fa: NonEmptyList[A])(f: (A, Int) => B): NonEmptyList[B] =
+        StaticMethods.mapWithIndexFromStrictFunctor(fa, f)(this)
+
+      override def zipWithIndex[A](fa: NonEmptyList[A]): NonEmptyList[(A, Int)] =
+        fa.zipWithIndex
 
       override def foldLeft[A, B](fa: NonEmptyList[A], b: B)(f: (B, A) => B): B =
         fa.foldLeft(b)(f)
--- a/core/src/main/scala/cats/data/NonEmptySeq.scala
+++ b/core/src/main/scala/cats/data/NonEmptySeq.scala
@@ -23,11 +23,11 @@ package cats
 package data
 
 import cats.data.NonEmptySeq.ZipNonEmptySeq
+import cats.kernel.compat.scalaVersionSpecific._
 
 import scala.annotation.tailrec
 import scala.collection.mutable
 import scala.collection.immutable.{Seq, SortedMap, TreeMap, TreeSet}
-import kernel.compat.scalaVersionSpecific._
 
 /**
  * A data type which represents a `Seq` guaranteed to contain at least one element.
@@ -260,6 +260,9 @@ final class NonEmptySeq[+A] private (val toSeq: Seq[A]) extends AnyVal with NonE
   def reverse: NonEmptySeq[A] =
     new NonEmptySeq(toSeq.reverse)
 
+  def mapWithIndex[B](f: (A, Int) => B): NonEmptySeq[B] =
+    new NonEmptySeq(toSeq.zipWithIndex.map(ai => f(ai._1, ai._2)))
+
   def zipWithIndex: NonEmptySeq[(A, Int)] =
     new NonEmptySeq(toSeq.zipWithIndex)
 
@@ -433,6 +436,9 @@ sealed abstract private[data] class NonEmptySeqInstances {
         fa: NonEmptySeq[A]
       )(f: (A) => G[B])(implicit G: Applicative[G]): G[NonEmptySeq[B]] =
         G.map2Eval(f(fa.head), Always(Traverse[Seq].traverse(fa.tail)(f)))(NonEmptySeq(_, _)).value
+
+      override def mapWithIndex[A, B](fa: NonEmptySeq[A])(f: (A, Int) => B): NonEmptySeq[B] =
+        fa.mapWithIndex(f)
 
       override def zipWithIndex[A](fa: NonEmptySeq[A]): NonEmptySeq[(A, Int)] =
         fa.zipWithIndex
--- a/core/src/main/scala/cats/instances/seq.scala
+++ b/core/src/main/scala/cats/instances/seq.scala
@@ -22,14 +22,13 @@
 package cats
 package instances
 
-import cats.data.{Chain, ZipSeq}
+import cats.data.{Chain, Ior, ZipSeq}
 import cats.instances.StaticMethods.appendAll
 import cats.kernel.compat.scalaVersionSpecific._
 
 import scala.annotation.tailrec
 import scala.collection.{+:, mutable}
 import scala.collection.immutable.Seq
-import cats.data.Ior
 
 @suppressUnusedImportWarningForScalaVersionSpecific
 trait SeqInstances extends cats.kernel.instances.SeqInstances {
@@ -134,7 +133,7 @@ trait SeqInstances extends cats.kernel.instances.SeqInstances {
         G.map(Chain.traverseViaChain(fa.toIndexedSeq)(f))(_.toVector)
 
       override def mapWithIndex[A, B](fa: Seq[A])(f: (A, Int) => B): Seq[B] =
-        fa.iterator.zipWithIndex.map(ai => f(ai._1, ai._2)).toIndexedSeq
+        fa.zipWithIndex.map(ai => f(ai._1, ai._2))
 
       override def zipWithIndex[A](fa: Seq[A]): Seq[(A, Int)] =
         fa.zipWithIndex
--- a/tests/shared/src/test/scala/cats/tests/TraverseSuite.scala
+++ b/tests/shared/src/test/scala/cats/tests/TraverseSuite.scala
@@ -36,6 +36,19 @@ abstract class TraverseSuite[F[_]: Traverse](name: String)(implicit ArbFInt: Arb
     }
   }
 
+  test(s"Traverse[$name].mapAccumulate") {
+    forAll { (init: Int, fa: F[Int], fn: ((Int, Int)) => (Int, Int)) =>
+      val lhs = fa.mapAccumulate(init)((s, a) => fn((s, a)))
+
+      val rhs = fa.foldLeft((init, List.empty[Int])) { case ((s1, acc), a) =>
+        val (s2, b) = fn((s1, a))
+        (s2, b :: acc)
+      }
+
+      assert(lhs.map(_.toList) === rhs.map(_.reverse))
+    }
+  }
+
   test(s"Traverse[$name].mapWithIndex") {
     forAll { (fa: F[Int], fn: ((Int, Int)) => Int) =>
       assert(fa.mapWithIndex((a, i) => fn((a, i))).toList === (fa.toList.zipWithIndex.map(fn)))
--- a/core/src/main/scala/cats/Composed.scala
+++ b/core/src/main/scala/cats/Composed.scala
@@ -124,6 +124,9 @@ private[cats] trait ComposedTraverse[F[_], G[_]]
 
   override def traverse[H[_]: Applicative, A, B](fga: F[G[A]])(f: A => H[B]): H[F[G[B]]] =
     F.traverse(fga)(ga => G.traverse(ga)(f))
+
+  override def mapAccumulate[S, A, B](init: S, fga: F[G[A]])(f: (S, A) => (S, B)): (S, F[G[B]]) =
+    F.mapAccumulate(init, fga)((s, ga) => G.mapAccumulate(s, ga)(f))
 }
 
 private[cats] trait ComposedNonEmptyTraverse[F[_], G[_]]
--- a/core/src/main/scala/cats/data/Chain.scala
+++ b/core/src/main/scala/cats/data/Chain.scala
@@ -42,8 +42,9 @@
 package cats
 package data
 
+import cats.instances.StaticMethods
 import cats.kernel.compat.scalaVersionSpecific._
-import cats.kernel.instances.StaticMethods
+import cats.kernel.instances.{StaticMethods => KernelStaticMethods}
 
 import scala.annotation.tailrec
 import scala.collection.immutable.SortedMap
@@ -854,7 +855,8 @@ sealed abstract class Chain[+A] extends ChainCompat[A] {
     builder.result()
   }
 
-  def hash[AA >: A](implicit hashA: Hash[AA]): Int = StaticMethods.orderedHash((this: Chain[AA]).iterator)
+  def hash[AA >: A](implicit hashA: Hash[AA]): Int =
+    KernelStaticMethods.orderedHash((this: Chain[AA]).iterator)
 
   override def toString: String = show(Show.show[A](_.toString))
 
@@ -1265,8 +1267,17 @@ sealed abstract private[data] class ChainInstances extends ChainInstances1 {
           traverseViaChain {
             val as = collection.mutable.ArrayBuffer[A]()
             as ++= fa.iterator
-            StaticMethods.wrapMutableIndexedSeq(as)
+            KernelStaticMethods.wrapMutableIndexedSeq(as)
           }(f)
+
+      override def mapAccumulate[S, A, B](init: S, fa: Chain[A])(f: (S, A) => (S, B)): (S, Chain[B]) =
+        StaticMethods.mapAccumulateFromStrictFunctor(init, fa, f)(this)
+
+      override def mapWithIndex[A, B](fa: Chain[A])(f: (A, Int) => B): Chain[B] =
+        StaticMethods.mapWithIndexFromStrictFunctor(fa, f)(this)
+
+      override def zipWithIndex[A](fa: Chain[A]): Chain[(A, Int)] =
+        fa.zipWithIndex
 
       def empty[A]: Chain[A] = Chain.nil
       def combineK[A](c: Chain[A], c2: Chain[A]): Chain[A] = Chain.concat(c, c2)
@@ -1367,7 +1378,7 @@ sealed abstract private[data] class ChainInstances extends ChainInstances1 {
         traverseFilterViaChain {
           val as = collection.mutable.ArrayBuffer[A]()
           as ++= fa.iterator
-          StaticMethods.wrapMutableIndexedSeq(as)
+          KernelStaticMethods.wrapMutableIndexedSeq(as)
         }(f)
 
     override def filterA[G[_], A](fa: Chain[A])(f: A => G[Boolean])(implicit G: Applicative[G]): G[Chain[A]] =
--- a/core/src/main/scala/cats/data/Nested.scala
+++ b/core/src/main/scala/cats/data/Nested.scala
@@ -347,6 +347,15 @@ private[data] trait NestedTraverse[F[_], G[_]]
 
   override def traverse[H[_]: Applicative, A, B](fga: Nested[F, G, A])(f: A => H[B]): H[Nested[F, G, B]] =
     Applicative[H].map(FG.traverse(fga.value)(f))(Nested(_))
+
+  override def mapAccumulate[S, A, B](init: S, fga: Nested[F, G, A])(f: (S, A) => (S, B)): (S, Nested[F, G, B]) = {
+    val (finalState, fgb) = FG.mapAccumulate(init, fga.value)(f)
+    (finalState, Nested(fgb))
+  }
+
+  override def mapWithIndex[A, B](fga: Nested[F, G, A])(f: (A, Int) => B): Nested[F, G, B] = {
+    Nested(FG.mapWithIndex(fga.value)(f))
+  }
 }
 
 private[data] trait NestedDistributive[F[_], G[_]] extends Distributive[Nested[F, G, *]] with NestedFunctor[F, G] {
--- a/core/src/main/scala/cats/data/NonEmptyChain.scala
+++ b/core/src/main/scala/cats/data/NonEmptyChain.scala
@@ -23,7 +23,10 @@ package cats
 package data
 
 import NonEmptyChainImpl.create
+
+import cats.instances.StaticMethods
 import cats.kernel._
+
 import scala.collection.immutable.SortedMap
 
 private[data] object NonEmptyChainImpl extends NonEmptyChainInstances with ScalaVersionSpecificNonEmptyChainImpl {
@@ -622,6 +625,15 @@ sealed abstract private[data] class NonEmptyChainInstances extends NonEmptyChain
 
         loop(fa.head, fa.tail).value
       }
+
+      override def mapAccumulate[S, A, B](init: S, fa: NonEmptyChain[A])(f: (S, A) => (S, B)): (S, NonEmptyChain[B]) =
+        StaticMethods.mapAccumulateFromStrictFunctor(init, fa, f)(this)
+
+      override def mapWithIndex[A, B](fa: NonEmptyChain[A])(f: (A, Int) => B): NonEmptyChain[B] =
+        StaticMethods.mapWithIndexFromStrictFunctor(fa, f)(this)
+
+      override def zipWithIndex[A](fa: NonEmptyChain[A]): NonEmptyChain[(A, Int)] =
+        fa.zipWithIndex
 
       override def size[A](fa: NonEmptyChain[A]): Long = fa.length
 
--- a/core/src/main/scala/cats/data/NonEmptyVector.scala
+++ b/core/src/main/scala/cats/data/NonEmptyVector.scala
@@ -23,11 +23,12 @@ package cats
 package data
 
 import cats.data.NonEmptyVector.ZipNonEmptyVector
+import cats.instances.StaticMethods
+import cats.kernel.compat.scalaVersionSpecific._
 
 import scala.annotation.tailrec
 import scala.collection.mutable
 import scala.collection.immutable.{SortedMap, TreeMap, TreeSet, VectorBuilder}
-import kernel.compat.scalaVersionSpecific._
 
 /**
  * A data type which represents a `Vector` guaranteed to contain at least one element.
@@ -435,6 +436,14 @@ sealed abstract private[data] class NonEmptyVectorInstances {
         fa: NonEmptyVector[A]
       )(f: (A) => G[B])(implicit G: Applicative[G]): G[NonEmptyVector[B]] =
         G.map2Eval(f(fa.head), Always(Traverse[Vector].traverse(fa.tail)(f)))(NonEmptyVector(_, _)).value
+
+      override def mapAccumulate[S, A, B](init: S, fa: NonEmptyVector[A])(
+        f: (S, A) => (S, B)
+      ): (S, NonEmptyVector[B]) =
+        StaticMethods.mapAccumulateFromStrictFunctor(init, fa, f)(this)
+
+      override def mapWithIndex[A, B](fa: NonEmptyVector[A])(f: (A, Int) => B): NonEmptyVector[B] =
+        StaticMethods.mapWithIndexFromStrictFunctor(fa, f)(this)
 
       override def zipWithIndex[A](fa: NonEmptyVector[A]): NonEmptyVector[(A, Int)] =
         fa.zipWithIndex
--- a/core/src/main/scala/cats/instances/StaticMethods.scala
+++ b/core/src/main/scala/cats/instances/StaticMethods.scala
@@ -21,15 +21,43 @@
 
 package cats.instances
 
+import cats.Functor
+
 import scala.collection.mutable.Builder
 
-private[instances] object StaticMethods {
+private[cats] object StaticMethods {
 
   def appendAll[F <: Iterable[A], A](it: Iterator[F], bldr: Builder[A, F]): bldr.type = {
     while (it.hasNext) {
       bldr ++= it.next()
     }
     bldr
+  }
+
+  def mapAccumulateFromStrictFunctor[S, F[_], A, B](init: S, fa: F[A], f: (S, A) => (S, B))(implicit
+    ev: Functor[F]
+  ): (S, F[B]) = {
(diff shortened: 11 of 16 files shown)


core/src/main/scala/cats/Traverse.scala
```diff
@@ -37,6 +38,7 @@ import simulacrum.typeclass
  *
  * See: [[https://www.cs.ox.ac.uk/jeremy.gibbons/publications/iterator.pdf The Essence of the Iterator Pattern]]
  */
+@implicitNotFound("Could not find an instance of Traverse for ${F}")
```
armanbilge, line 41: We removed these in https://github.com/typelevel/cats/pull/4060. Michael is right, we should probably stop using simulacrum scalafix :(

core/src/main/scala/cats/Traverse.scala
```diff
@@ -185,12 +199,11 @@ object Traverse {
...
-      new AllOps[F, A] {
-        type TypeClassType = Traverse[F]
-        val self: F[A] = target
-        val typeClassInstance: TypeClassType = tc
-      }
+    } = new AllOps[F, A] {
+      type TypeClassType = Traverse[F]
+      val self: F[A] = target
+      val typeClassInstance: TypeClassType = tc
+    }
```
armanbilge, line 206: Sorry to nit, but probably we can revert these formatting changes 😅

BalmungSan: Sure, all that was **Scalafix**.

core/src/main/scala/cats/Traverse.scala
```diff
@@ -135,12 +135,24 @@ import simulacrum.typeclass
...
+   * Akin to [[map]], but allows to keep track of a state value
+   * when calling the function.
+   */
+  def mapAccumulate[A, S, B](fa: F[A])(init: S)(f: (A, S) => (B, S)): (F[B], S) =
+    traverse(fa) { a =>
+      State.get[S].flatMap { s1 =>
+        val (b, s2) = f(a, s1)
+        State.set(s2).map(_ => b)
+      }
+    }.run(init).value.swap
```
satorg, line 148: I wonder if this can be simplified:
```suggestion
    traverse(fa) { a =>
      State { (s: S) => f(a, s).swap }
    }.run(init).value.swap
```

core/src/main/scala/cats/Traverse.scala
```diff
@@ -135,12 +135,24 @@ import simulacrum.typeclass
   override def map[A, B](fa: F[A])(f: A => B): F[B] =
     traverse[Id, A, B](fa)(f)
 
+  /**
+   * Akin to [[map]], but allows to keep track of a state value
+   * when calling the function.
+   */
+  def mapAccumulate[A, S, B](fa: F[A])(init: S)(f: (A, S) => (B, S)): (F[B], S) =
```
satorg, line 142: Actually, I think that parameters and results of the `f` function should be flipped.
```suggestion
  def mapAccumulate[A, S, B](fa: F[A])(init: S)(f: (S, A) => (S, B)): (S, F[B]) =
```
This order is used in [FS2](https://github.com/typelevel/fs2/blob/ff9d758ae61cb19ad4d6f5cde4d0b236efab8157/core/shared/src/main/scala/fs2/Stream.scala#L1770)
and it is usually used with states – S usually goes first in a pair.

Also I am not certain if the overall result should really be `(S, F[B])` rather than `F[(S, B)]`.

armanbilge: Agree on `S` coming first.

I believe `(S, F[B])` is correct, I guess `F[(S, B)]` would also be storing the state for each iteration?

satorg: Correct, and that would reflect the idea of multiple state transitions, I think: so we have a function that makes a single transition `(S, A) => (S, B)`, where `A` is an input event, and `B` is an output one associated with a particular state transition. Therefore, when we return `F[(S, B)]`, we get all states and outputs linked to each other (in context of `F[_]` of course).

But it is just IMO. And that is how it is defined in FS2 after all :)

satorg: ...on the second thought, the result could be just `F[B]`, to keep `mapAccumulate` consistent (to some extent) with `mapWithIndex` (which simply returns `F[B]`, not `(Int, F[B])`).

satorg: Also I'm thinking about a possible next step: once we get `mapAccumulate` then someone may ask for `mapAccumulateA`:
```scala
def mapAccumulateA[A, S, B](fa: F[A])(init: S)(f: (A, S) => F[(B, S)]): RESULT_TYPE
```
So what would be `RESULT_TYPE` in that case?

BalmungSan: @satorg about the order of parameters, I was curious about what you all think since I couldn't think of a good argument for either order and I keep flipping them when asking if it existed and when proposing it on **scala-library-next**. I was sure I based this one on **fs2** but it seems I confused myself at the final moment; again 😅 
However, I also used this order because `mapWithIndex` has `(A, Int)` instead of `(Int, A)`, thus it seemed natural that the `A` was always the first parameter.
Nevertheless, I am okay with any order.

About returning `F[(S, B)]` _(or `F[(B, S)]`)_, I think it makes sense, I originally only care about the final `F[B]` then, I added the final state as suggested in the issue I opened in **scala-library-next**.

satorg: I actually don't know why `mapWithIndex` uses this parameter order. Perhaps (I'm guessing) just to be in par with `zipWithIndex` which also has index as a second item in the tuple it returns. But it is fine to me since the index is not truly a state (although it can be modeled with the latter). But AFAI can see, in most other methods that work with states or accumulators the state parameter usually goes first (that said, let's assume that `foldLeft` is an example of such behavior while `foldRight` is an exception from that :) )

Regarding the result type. Yes, good point – having `F[B]` we can always make `F[(S, B)]` out of it. And having `F[(S, B)]` we can always get the last `S`. So seems that returning `F[B]` would be the most generic approach. And regarding the `(S, F[B])` the only "pros" I can see here – well, basically we already have the final `S` at the end of the calculations, why not to return it, right? On the other hand, there are a few of "cons" as well:
1. If some one still needs to get _all_ states, not just the final one, they will end up with something like `(S, F[(S, B)])` which is way too clumsy (to me at least).
2. Receiving some final value out of `Traverse` – it is a "folding" job, not a "mapping" one. I.e. since `Traverse` extends `Foldable` it is always possible to get just a final `S` by calling some kind of `fold` methods.
3. Returning `S` directly from a map-like methods looks like escaping of `F` context which seems a bit suspicious to me.

So yeah, as a summary I would say that my objections against `(S, F[B])` are more aesthetic rather than practical.

johnynek: I am very -1 on `F[(S, B)]`. That gratuitously puts S inside potentially and effect when it does not need to be. If F is IO that means we can no longer access S without evaluating the IO when in fact S is a function of the length of the traversable and not dependent on anything about the effect.

satorg: But `Traverse` cannot be implemented for any effect, I guess. Or can it? I assume it is impossible just because it is also a `Foldable` which would mean unsafe direct access to the contained value.

satorg: I mean, folding an effect is only feasible by triggering its evaluation.

BalmungSan: @satorg 

> But AFAI can see, in most other methods that work with states or accumulators the state parameter usually goes first

It makes sense. And, since I don't have any preference and I don't see any argument for the current order I will change it. Unless either @armanbilge or @johnynek may have something to say about it.

> we can always make F[(S, B)] out of it. 

Also `F[(B, S)]` or `F[(S, B, S)]` _(input and output state)_, thus yeah I don't think it is worth it to have that there. Also, if you don't care about the states after the `map` you would need to discard them; and I would think is more common to not care about them than care... but, of course, we lack any data for either affirmation.

> And having F[(S, B)] we can always get the last S

Yes, but it may require some _(inefficient)_ traversal of the collection.

> And regarding the (S, F[B]) the only "pros" I can see here – well, basically we already have the final S at the end of the calculations, why not to return it, right?

I mean, as I said, my original idea was also just only `F[B]` it was a comment on **scala-library-next** that proposed returning the last state, which makes sense.
However, being honest, I think it hurts operation chaining, sure you can call `_._1` or `_._2` to get the `F[B]`, but I personally never liked the tuple accessors they are too opaque.

Also, if someone wants the final state they can just reimplement the logic, is not a big deal anyway.
It may be considered again if many folks ask about a way to get the final state.

@johnynek 

> That gratuitously puts S inside potentially and effect when it does not need to be.
> If F is IO that means we can no longer access S

If `F` is `IO` this won't work because there is no `Traverse[IO]` :)
Also, this function only makes sense for `Fs` that actually contain values _(note the plural)_ inside of them; e.g. `List`, `Stream`. Rather than `Fs` that are programs descriptions; e.g. `IO`, `Option`, `Either`

johnynek: yeah, sorry, I got confused there (since this is a Traverse and the generally the interesting traversals are ones where you have both the `F[_]` of the container type a `G[_]` of Applicative type, but I digress, I was incorrect).

I still feel not putting it inside is right. Suppose I don't want it inside. Forcing it inside makes me have to map it away after. Suppose you do want it inside, then just use a `B1 = (B, S)` and put it inside as you like.

So, I think the better API is to keep it as it was originally `(F[B], S)` (or reversed order, I have less strong feeling about the order).

satorg: But do we really need that `S` in the result eventually? Again, comparing `mapAccumulate` to `mapWithIndex` – the latter does not return `(F[B], Int)`, but rather just `F[B]`.

I'm a bit reluctant just because I don't think there will be a lot of use cases where both `S` and `F[B]` will be requested for. I would suppose that it will be either of them but not both. And if only `S` is necessary, then `foldLeft` will be a better method to call.

The only reason to return both I can really see – these values both are already in place. So we just take everything we have and return it in a tuple. And I admit – it is not necessarily the illicit reason though...

johnynek: Let me try one more argument:

If we change the order of the arguments and return a function:
```scala
def mapAccumulate[A, S, B](fn: (A, S) => (B, S)): (F[A], S) => (F[B], S) = ...
```
so we see the symmetry here: map-accumulate is lifting a function of type `(A, S) => (B, S)` to a function of type `(F[A], S) => (F[B], S)`.

This can be compared to others:
```scala
def map[A, B](fn: A => B): F[A] => F[B] = ..
def flatMap[A, B](fn: A => F[B]): F[A] => F[B] = ..
def map2[A, B, C](fn: (A, B) => C): (F[A], F[B]) => F[C] = ...
def tailRecM[A, B](fn:  A => F[Either[A, B]): A => F[B]

def foldLeft[A, B](fn: (A, B) => A): (A, F[B]) => A = ...
def traverse[A, B, G[_]](fn: A => G[B]): F[A] => G[F[B]]  = ...

// and now:
def mapAccumulate[A, S, B](fn: (A, S) => (B, S)): (F[A], S) => (F[B], S) = ...
```

it's really helpful, I think, to think of these functions convert on function to a very similar function with type constructors moved around.

satorg: Oh, this is a cool one, I really like that point of view.

johnynek, line 142: Why is init in a separate parameter group. Do we usually do that? Why not group with fa?

armanbilge: Probably this was to appease simulacrum, which was used to bootstrap this change?

BalmungSan: Sorry, I am just kind of used to currying by default... I made the same mistake on the `contains_` PR 😅

BalmungSan:
Okay, I think I made all the requested changes; please let me know if I forgot something or if there is something else.

tests/shared/src/test/scala/cats/tests/TraverseSuite.scala
```diff
@@ -36,6 +36,19 @@ abstract class TraverseSuite[F[_]: Traverse](name: String)(implicit ArbFInt: Arb
...
+  test(s"Traverse[$name].mapAccumulate") {
+    forAll { (fa: F[Int], init: Int, fn: ((Int, Int)) => (Int, Int)) =>
+      val lhs = fa.mapAccumulate(init)((s, a) => fn((s, a)))
+
+      val rhs = fa.foldLeft((init, List.empty[Int])) { case ((s1, acc), a) =>
+        val (s2, b) = fn((s1, a))
+        (s2, b :: acc)
+      }
+
+      assert(lhs._2.toList === rhs._2.reverse)
```
satorg, line 48: Since `mapAccumulate` returns both `S` and `F[B]` then it makes sense to add a check for `S` here as well, IMO.

core/src/main/scala/cats/Traverse.scala
```diff
@@ -135,12 +135,19 @@ import simulacrum.typeclass
   override def map[A, B](fa: F[A])(f: A => B): F[B] =
     traverse[Id, A, B](fa)(f)
 
+  /**
+   * Akin to [[map]], but allows to keep track of a state value
+   * when calling the function.
+   */
+  def mapAccumulate[A, S, B](fa: F[A], init: S)(f: (S, A) => (S, B)): (S, F[B]) =
```
johnynek, line 142: Last nit: I think the order of the parameters should be the same order as the function args: so either `F[A], S, (A, S) => (B, S)` or `S, F[A], (S, A) => (S, B)`

At least IMO.

BalmungSan: > so either F[A], S, (A, S) => (B, S)

Well, that was argued before and @satorg mentioned that it was common for the state to be on the left like in `foldLeft` _(argument)_ or `State.apply` _(return)_

> or S, F[A], (S, A) => (S, B)

Uhm, that feels weird and would not be consistent with the extension syntax.
Also, this signature is pretty close to `foldLeft` which is `F[A], B, (S, A) => B` which has this _mixed_ order.

johnynek:
BTW: also I think you can override this for list, vector, lazy list and stream that doesn't use state and is likely far faster.

johnynek:
One more thing: I think this is a very nice function.

core/src/main/scala/cats/Traverse.scala
satorg: Also I'm wondering whether should the type parameter order be
   1. `[S, A, B]`
   2. `[A, S, B]`
   3. `[A, B, S]`

I realize it might be annoying (sorry for that), but since it is a sort of design decision, I would like to think twice now rather than stuffing the library with bin-compat hacks later just due to a hasty solution.

satorg: > Also, I believe, the F[A] is always the first argument in all methods of all typeclases...

Actually, no. I know at least one example where it is not held: [NonEmptyAlternative#prependK](https://github.com/typelevel/cats/blob/3cb763920f177f045e2253be51e52ea2554770ad/core/src/main/scala/cats/NonEmptyAlternative.scala#L37). The syntax for that is defined as usual, though: `xs.prependK(x)`. So such an order is not something absolutely unbearable in Cats.

PS. Actually, it was me who defined `prependK` in that way. But I did it at the request of other reviewers :)

BalmungSan:
While trying to override the current implementation for `LazyList` I realized that the current interface is broken for lazy collections, it forces the evaluation of them, which is not required at all. And the problem is for trying to return the final state... such value may not be computed after the call to `mapAccumulate`; actually such value may never exist at all; since the collection could be infinite.
Thus, I think the best is to just return `F[B]`

What do you all think @satorg @armanbilge @johnynek?

satorg:
> BTW: also I think you can override this for list, vector, lazy list and stream that doesn't use state and is likely far faster.

Don't forget `Seq` please. As well as non-empty counterparts of all from the above: `NonEmptyList`, `NonEmptyVector`, etc. :)

But it is going to be quite a big chunk of additional work. I would suggest to do it in a separate PR, if possible.

johnynek:
Traverse already forces a LazyList doesn't it? Otherwise we would need fs2 as much. Am I mistaken?

satorg:
> Traverse already forces a LazyList doesn't it?

That's correct. But in fact, `traverse` itself cannot be processed lazily, e.g.:
```
val lzy: LazyList[Int] = ???

val traversedLzy: Option[LazyList[Int]] = lzy.traverse(a => if (someCondition(a)) Some(a) else None)
```
So we cannot emit the outer `Option` as a result without forcing the `LazyList`.

But seems that it would not be the case for `mapAccumulate` if we did not have to return the accumulator value itself...
So yeah, I am kind of hesitant here again...

satorg:
Just a note: seems that `traverse` for `LazyList` fully forces it even in this case:
```
val lzy: LazyList[Int] = ???

val traversedLzy: Option[LazyList[Int]] = lzy.traverse(_ #:: LazyList.empty)
```
Not sure that in such a case it should really be enforced...

satorg:
I can think about several options on how we could address it.

1. Just do nothing (besides we have in this PR already). Perhaps having `mapAccumulate` that behaves truly lazily for `LazyList` is not that important.
2. Make `mapAccumulate` returning `(Eval[S], F[B])` instead. In that case the `LazyList` could be optimized in the way that it is enforced on either fetching items from `F[B]` or attempting to get `S` out of `Eval` only. Could be quite difficult to implement, though.
3. Create 2 functions instead of just one `mapAccumulate`: the first one returning just `F[B]` but allowing lazy traversal where applicable and the second one returning `(S, F[B])` but always eager.

BalmungSan:
> Create 2 functions instead of just one mapAccumulate: the first one returning just F[B] but allowing lazy traversal where applicable and the second one returning (S, F[B]) but always eager.

The more I think about this the less I like it, I think the safest is just to not leak the `S` in the final signature.
As we said, if someone needs it they can just make it part of the `B`
Another argument for not doing that is that the current `mapAccumulate` of **fs2** doesn't return that final `S` _(for obvious reasons)_, thus I think it would be best to be consistent.

johnynek:
But there is no Traverse instance for fs2.Stream right? There can't be for the same reason you can't have one for IO.

I'm disappointed that the more theoretic argument about what the signature should be (the one that composes and does lifting of the `F[_]`) takes a back seat to the ideas that:

1. Even though for Traverse forces LazyList and Stream, we want to change a type signature so those two instances could not force for this one method (even though all other traverse derived functions will force).

2. We want to be consistent with another method in fs2 despite that type not being Traverse.

johnynek:
Regarding composition I mentioned in my last comment. Recall the functions here: https://github.com/typelevel/cats/pull/4209#discussion_r878777483

So if we have:
```scala
val mapList: [X, Y] => (List[X], X => Y) => List[Y] = ...
val mapOption: [X, Y] => (Option[X], X => Y) => Option[X] = ...

val mapListOption: [X, Y] => (List[Option[X]], X => Y) => List[Option[X]] =
  [X, Y] => (lo: List[Option[X]], fn: X => Y) => mapList(lo, mapOption(_: Option[X], fn)) 
```
so this is what we mean when we say Functor composes.

If you look at the shapes of the functions in https://github.com/typelevel/cats/pull/4209#discussion_r878777483 , you can see which ones are kind of "shape preserving" meaning the results are the same shape as the original with an `F[_]` added or removed. We can see map, foldLeft, traverse all have that property, which is due to the fact that those type classes all compose: if you have `Functor[F]` and `Functor[G]` you can make `Functor[[X] =>> F[G[X]]]` and so on with `Foldable` and `Traverse`.

Now, look at the case we have here:
```scala
val mapAccList: [X, S, Y] => ((X, S) => (Y, S)) => ((List[X], S) => (List[Y], S) ...
val mapAccOption: [X, S, Y] => ((X, S) => (Y, S)) => ((Option[X], S) => (Option[Y], S)) = ...

val mapAccListOption: [X, S, Y] => ((X, S) => (Y, S)) => ((List[Option[X]], S) => (List[Option[Y], S)) =
  [X, Y] => (fn: (X, S) => (Y, S)) => mapAccList(mapAccOption(fn))
```

So, in this way, if we have a `mapAccumulate` optimized for `Vector` and one optimized for `List` we can *keep* our optimizations for `[X] =>> Vector[List[X]]` since the function composes.

We can't keep our optimizations under compositions if we discard the final S.

I can see the motivations to discard the S in many cases, but I think if we really think we want that we could add a second function that just drops the S off.

But to me, the theoretical reasons should be the highest priority here since they tend to guide us to universal principles, not just the aesthetics of one person or another.

johnynek:
Lastly, I would note, this is really a generalization of:
```scala
val foldLeftList: [S, X] => ((S, X) => S) => ((S, List[X]) => S)
val mapAccList: [S, X, Y] => ((S, X) => (S, Y)) => ((S, List[X]) => (S, List[Y]))
```
Where while folding, you also do a transformation. This symmetry breaks if we discard the final S.

BalmungSan:
@johnynek 

The compose argument is appealing and I am very close to buying it; but:

1. You can also compose without returning the final state, you just use the `init` again for each nested `mapAccumulate` - Actually, I just did a quick experiment with `Nested` and that is what it does.

> (even though all other traverse derived functions will force).

2.  But those methods are overridden so they don't force evaluation, only this one would require forcing the evaluation.

Thus, I want to ask you and everyone:

Should we also override the `Nested` implementation so it reuses the last state of each nested call as the `init` of the next one? I actually like this idea; but, not sure which implementation is actually better / more usable.
Still, I would buy this point because if someone needs to reuse `init` for each nested call then they just can, whereas not providing the last state means that if someone wants the other behavior is now more difficult to get it.
So, I personally would say yes to this is the solution that allows for maximum flexibility.

Should we not override the implementation for `LazyList` and `Stream` and just let it force.
And, maybe, if you all agree, may I open a new issue about if `Traverse` makes sense at all for lazy collections? I understand that at this point **cats** won't remove instances for multiple reasons, but maybe the issue will serve as a reference point for when someone hits an issue or has a question about some behavior.

johnynek:
> 1. You can also compose without returning the final state, you just use the `init` again for each nested `mapAccumulate` - Actually, I just did a quick experiment with `Nested` and that is what it does.

I can't see how that can be true. Can you show explicitly the examples? For instance, we can implement mapWithIndex using mapAccumulate. If you reset each inner `init` with `0` in that case nesting mapWithIndex will reset the index on each of the outers, and will be the same as `map + mapWithIndex` I can't see how that can be correct.

Does it satisfy the laws when the outer has size > 1?

> > (even though all other traverse derived functions will force).
> 
> 2. But those methods are overridden so they don't force evaluation, only this one would require forcing the evaluation.

That's not true. `Traverse` extends `Foldable` and many if not all of `Foldable` force the LazyList/Stream (e.g. size, foldLeft, foldMap, etc...)

> Thus, I want to ask you and everyone:
> 
> Should we also override the `Nested` implementation so it reuses the last state of each nested call as the `init` of the next one? I actually like this idea; but, not sure which implementation is actually better / more usable. Still, I would buy this point because if someone needs to reuse `init` for each nested call then they just can, whereas not providing the last state means that if someone wants the other behavior is now more difficult to get it. So, I personally would say yes to this is the solution that allows for maximum flexibility.

I don't see how providing the original `init` on each can be lawful. It would have to forget between steps. Consider `List[Vector[_]]`. With mapWithIndex, I assume what is happening is that we are indexing so that:
```scala
List(Vector('a', 'b'), Vector('c', 'd')) -> List(Vector(0, 1), Vector(2, 3))
```
> Should we not override the implementation for `LazyList` and `Stream` and just let it force. And, maybe, if you all agree, may I open a new issue about if `Traverse` makes sense at all for lazy collections? I understand that at this point **cats** won't remove instances for multiple reasons, but maybe the issue will serve as a reference point for when someone hits an issue or has a question about some behavior.

Well, traverse is definitely bad dangerous for infinite lazy lists. So is foldable. If you call `toList, size, foldLeft` on a an infinite lazy list it will either OOM or never terminate (depending on which case).

satorg:
I would say, if we declared `mapAccumulate` with this shape:
```
((S, A) => (S, B)) => ((Eval[S], F[A]) => (Eval[S], F[B]))
```
then we could make it both composable and lazily evaluable where applicable.

But it looks like a Haskell-ish way (where everything is lazy by default), and I'm not sure if the additional complexity that `Eval` would introduce here is really worth it.

BalmungSan:
> Edit: Point 1 was invalidated by an update to comment: https://github.com/typelevel/cats/pull/4209#issuecomment-1133942359

Sorry for that.

> That's not true. Traverse extends Foldable and many if not all of Foldable force the LazyList/Stream (e.g. size, foldLeft, foldMap, etc...)

I meant `mapWithIndex` and `zipWithIndex`, but yeah I forgot that the `Foldable` ones were also derived from `traverse`

Still, I think it is clear that thinking about folding / traversing on a possibly infinite lazy collection is weird.
Thus, I would want that we focus the last three points of my last comment:

> 1. Finish the parameters and type parameters order discussion that I haven't answered because I want to focus on this first.

I will resume this after waiting for a while to see if there are no more follow up comments about returning the final `S`
But, I think this case was already closed based on @johnynek comments.

> 2. See if we all agree that we should not override the implementation of LazyList and Stream

I Think Oscar implied this... but I would rather not assume.
Do we all agree on this?

> 3. Confirm if you all think it's worth having a separate discussion about Traverse + lazy collections.

What do you all think about opening a new issue for that?
Do you think is a valuable discussion to record?

johnynek:
your latest comment all looks good to me.

I agree we shouldn't bother to override on Stream/LazyList (at least not now) I do think an issue about Traverse + lazy types would be interesting. It could possibly be there could be some kind of `LazyTraverse[F[_]]` that could exist. In haskell some of these things are hidden due to the whole language being non-strict. But with the right use of `Defer[G]` along with `Applicate[G]` for lazy effects, and possibly using `Eval[_]` in some locations it may potentially be possible to find a lazy traverse (but I think it is way out of scope for this issue, and may well fail and the real answer may be: use fs2.Stream).

johnynek:
> I would say, if we declared `mapAccumulate` with this shape:
> 
> ```
> ((S, A) => (S, B)) => ((Eval[S], F[A]) => (Eval[S], F[B]))
> ```
> 
> then we could make it both composable and lazily evaluable where applicable.
> 
> But it looks like a Haskell-ish way (where everything is lazy by default), and I'm not sure if the additional complexity that `Eval` would introduce here is really worth it.

I think for this to work with composition you actually need: `((Eval[S], A) => (Eval[S], B)) => ((Eval[S], F[A]) => (Eval[S], F[B]))` (but there may be some trick I'm not seeing).

I also don't see if that would necessarily work, since evaluating the resulting `Eval[S]` will have to have the side effect of forcing the LazyList in our example, which may be possible (since forcing LazyList isn't an effect we track) but seems a bit weird.

BalmungSan:
@johnynek @satorg @armanbilge

Okay, thank you all for the feedback, I think the definition of the method is ready to go.
I also override the implementation of `mapAccumulate`, `mapWithIndex`, and `zipWithIndex` for `List`, `Vector`, `Seq`, `Chain` and their non-empty counterparts; I made all that in a second commit thus if you prefer I can move that to another PR.

Let me know what you all think and if there is any other detail to discuss :)

core/src/main/scala/cats/data/NonEmptyList.scala
```diff
@@ -787,6 +787,33 @@ sealed abstract private[data] class NonEmptyListInstances extends NonEmptyListIn
...
+        var state = init
+
+        val fb = fa.map { a =>
+          val (newState, b) = f(state, a)
+          state = newState
+          b
+        }
+
+        (state, fb)
+      }
```
satorg, line 800: I'm wondering if we could simply re-use implementations from regular collections for their non-empty counterparts. Something like
```scala
// Assume there's `listTraverse: Traverse[List]` accessible somehow:
override def mapAccumulate[S, A, B](init: S, fa: NonEmptyList[A])(f: (S, A) => (S, B)): (S, NonEmptyList[B]) = {
  val (last, fb) = listTraverse.mapAccumulate[A, A, B](init, fa.toList)(f)
  (last, NonEmptyList.fromListUnsafe(fb)) // should be safe, actually
}
```
And the same for all other NEs – would it make sense, wdyt?

johnynek:
can we also override mapAccumulate here:

https://github.com/typelevel/cats/blob/708cbaba2e6ac07a5ac1819b6c18b0c83a2faaba/core/src/main/scala/cats/Composed.scala#L118

and:

https://github.com/typelevel/cats/blob/708cbaba2e6ac07a5ac1819b6c18b0c83a2faaba/core/src/main/scala/cats/data/Nested.scala#L348

core/src/main/scala/cats/data/NonEmptyList.scala
BalmungSan: I thought about this but decided to go the dumb route first and see what you guys thought about it.
I think I prefer Oscar's idea of having a unified implementation that just requires `Functor`.

satorg: Yes, agree. That one looks even better to be.

core/src/main/scala/cats/instances/seq.scala
```diff
@@ -133,8 +133,11 @@ trait SeqInstances extends cats.kernel.instances.SeqInstances {
       final override def traverse[G[_], A, B](fa: Seq[A])(f: A => G[B])(implicit G: Applicative[G]): G[Seq[B]] =
         G.map(Chain.traverseViaChain(fa.toIndexedSeq)(f))(_.toVector)
 
+      override def mapAccumulate[S, A, B](init: S, fa: Seq[A])(f: (S, A) => (S, B)): (S, Seq[B]) =
+        StaticMethods.mapAccumulateFromStrictFunctor(init, fa, f)(this)
```
johnynek, line 137: I don't think is actually safe. `Seq` is an interface and I don't think it makes any promise about the order that `map` is applied. I think something like:
```scala
val bldr: List[B] = List.newBuilder[B]
val it = fa.iterator
var s = init
while (it.hasNext) {
  val a = it.next()
  val (nextS, b) = f(s, a)
  bldr += b
  s = nextS
}
(s, bldr.result())
```

johnynek (commented):
Looking good. Two minor issues.

core/src/main/scala/cats/instances/seq.scala
satorg: for Scala 2.13+ the following might be even better:
```
val bldr = fs.iterableFactory.newBuilder[B]
```
but I don't remember if it works for Scala 2.12.

satorg:
Not saying we should do it here, but I wonder if there should be something like
```
trait StrictTraverse[F[_]] extends Traverse[F] {
  def mapAccumulate... // put strict-only implementation directly to here
}
```
Then let strict collections only implement this typeclass.

This would allow us to represent the idea of "strictness" it a typefull way (not just by hiding it somewhere inside Cats).

core/src/main/scala/cats/instances/seq.scala
BalmungSan: > I don't think it makes any promise about the order that map is applied

AFAIK, the only thing `Seq` promises is the order of the elements.
However, I just realized that it doesn't guarantee that the collection will be strict, mainly because both `Stream` and `LazyList` are `Seqs`

I will just remove this, `Seq` should have never had instances IMHO.

johnynek:
@satorg but traverse itself is strict isn't it? How can we lazily implement `traverse` without any constraints on `G[_]: Applicative`?

johnynek (dismissed):
I think this is pretty great.

Very nice generalization of mapWithIndex.

johnynek (approved):

satorg:
> but traverse itself is strict isn't it? How can we lazily implement `traverse` without any constraints on `G[_]: Applicative`?

Actually, this is not what I meant. Yes, you're right – it is strict itself.
(_Although such function as `mapWithIndex` still does not have to be strict, does it?_)

What I really meant: seems we could define an abstract internal mix-in trait `StrictTraverse` with pre-defined implementations for `mapAccumulate` and `mapWithIndex` (and perhaps something else) and then mix it into the implementations of `Traverse` for `List`, `Vector`, maybe `Chain`. It is just a matter of the way how the code is organized within Cats. But I think it is a really minor thing, not worth long discussion.

core/src/main/scala/cats/data/NonEmptySeq.scala
```diff
@@ -260,6 +260,23 @@ final class NonEmptySeq[+A] private (val toSeq: Seq[A]) extends AnyVal with NonE
...
+    val fb = iterator.map { a =>
+      val (newState, b) = f(state, a)
+
+      state = newState
+
+      b
+    }.toIndexedSeq
+
+    (state, new NonEmptySeq(fb))
+  }
```
satorg, line 275: Is that `.toIndexedSeq` still 100% safe here? I mean, `IndexedSeq` is an interface. Although all `IndexedSeq`-s from the Scala library might be strict collections, but we cannot completely exclude a case when someone can try to create a `NonEmptySeq` out of their own `IndexedSeq` implementation that have `iterator.toIndexedSeq` returning a non-strict collection. E.g. imagine some storage-efficient sparse-array implementation. If it happens then `state` won't be properly calculated.

I'd suggest to iterate the iterator directly via `hasNext`/`next` methods – that will definitely guarantee the enforced evaluation. Also it will probably make the implementation even more efficient since there will be no closure involved anymore.

core/src/main/scala/cats/instances/seq.scala
```diff
@@ -133,6 +132,20 @@ trait SeqInstances extends cats.kernel.instances.SeqInstances {
...
+      override def mapAccumulate[S, A, B](init: S, fa: Seq[A])(f: (S, A) => (S, B)): (S, Seq[B]) = {
+        var state = init
+
+        val fb = fa.iterator.map { a =>
+          val (newState, b) = f(state, a)
+
+          state = newState
+
+          b
+        }.toIndexedSeq
```
satorg, line 144: The same comment as for the `NonEmptySeq` above.

Actually, I think that the only implementation should be put here and then re-used from those for `Traverse[NonEmptySeq]`.

satorg: Also I still think that it would be better to preserve the original Seq's concrete type. The only thing, it should be a bit different for Scala 2.12 and 2.13+:
```
val builder = fa.companion.newBuilder[B] // 2.12
OR
val builder = fa.iterableFactory.newBuilder[B] // 2.13+
```
[Seq#companion](https://www.scala-lang.org/api/2.12.15/scala/collection/immutable/Seq.html#companion:scala.collection.generic.GenericCompanion[scala.collection.immutable.Seq])
[Seq#iterableFactory](https://www.scala-lang.org/api/current/scala/collection/immutable/Seq.html#iterableFactory:scala.collection.SeqFactory[Seq])

core/src/main/scala/cats/data/NonEmptySeq.scala
BalmungSan: Actually, there is one in the stdlib `IndexedSeqView` which means yeah this is unsafe.

core/src/main/scala/cats/instances/seq.scala
```diff
@@ -134,7 +133,7 @@ trait SeqInstances extends cats.kernel.instances.SeqInstances {
         G.map(Chain.traverseViaChain(fa.toIndexedSeq)(f))(_.toVector)
 
       override def mapWithIndex[A, B](fa: Seq[A])(f: (A, Int) => B): Seq[B] =
-        fa.iterator.zipWithIndex.map(ai => f(ai._1, ai._2)).toIndexedSeq
+        fa.zipWithIndex.map(ai => f(ai._1, ai._2))
```
armanbilge, line 136: Apologies if I missed it, why the change here?

armanbilge (commented):
One question, otherwise LGTM! Thanks for all your work :)

core/src/main/scala/cats/instances/seq.scala
BalmungSan: Check my last comment:

> replaced a previous usage of toIndexedSeq on mapWithIndex with a version that does two traversals but preserves the underlying class

Motivated by

> Also I still think that it would be better to preserve the original Seq's concrete type

armanbilge: FWIW I think we can use .view here to get the best of both worlds, but anyway.

armanbilge (approved):

core/src/main/scala/cats/instances/seq.scala
BalmungSan: @armanbilge how would we get back the original collection using `view`?

BalmungSan: > (Since version 2.13.0) Views no longer know about their underlying collection type; .force always returns an IndexedSeq

https://www.scala-lang.org/api/current/scala/collection/View.html#force:scala.collection.IndexedSeq[A]

armanbilge: Aha, casualty of the new collections, my bad :)

In 2.13 though you can do this:
```scala
fa.view.zipWithIndex.map(ai => f(ai._1, ai._2)).to(fa.factory)
```

But, too much shenanigans. I'm also skeptical about instances for `Seq`.

PR #4214, opened 2022-05-28
override more mapAccumulate methods in Traverse

johnynek:
follow up to #4209 to override I think all the remaining values for mapAccumulate.

Also did a few minor changes I noticed along the way.

### DIFF
--- a/alleycats-core/src/main/scala/alleycats/std/iterable.scala
+++ b/alleycats-core/src/main/scala/alleycats/std/iterable.scala
@@ -77,6 +77,12 @@ trait IterableInstances {
         }
         (s, vec.result())
       }
+
+      override def zipWithIndex[A](fa: Iterable[A]): Iterable[(A, Int)] =
+        fa.zipWithIndex
+
+      override def mapWithIndex[A, B](fa: Iterable[A])(f: (A, Int) => B): Iterable[B] =
+        fa.zipWithIndex.map { case (a, i) => f(a, i) }
     }
 
   implicit def alleycatsStdIterableTraverseFilter: TraverseFilter[Iterable] = new TraverseFilter[Iterable] {
--- a/alleycats-core/src/main/scala/alleycats/std/iterable.scala
+++ b/alleycats-core/src/main/scala/alleycats/std/iterable.scala
@@ -65,6 +65,18 @@ trait IterableInstances {
       override def traverse[G[_], A, B](fa: Iterable[A])(f: A => G[B])(implicit G: Applicative[G]): G[Iterable[B]] =
         if (fa.isEmpty) G.pure(Iterable.empty)
         else G.map(Chain.traverseViaChain(toImIndexedSeq(fa))(f))(_.toVector)
+
+      override def mapAccumulate[S, A, B](init: S, fa: Iterable[A])(f: (S, A) => (S, B)): (S, Iterable[B]) = {
+        val iter = fa.iterator
+        var s = init
+        val vec = Vector.newBuilder[B]
+        while (iter.hasNext) {
+          val (snext, b) = f(s, iter.next())
+          vec += b
+          s = snext
+        }
+        (s, vec.result())
+      }
     }
 
   implicit def alleycatsStdIterableTraverseFilter: TraverseFilter[Iterable] = new TraverseFilter[Iterable] {
--- a/alleycats-core/src/main/scala/alleycats/std/map.scala
+++ b/alleycats-core/src/main/scala/alleycats/std/map.scala
@@ -45,11 +45,24 @@ trait MapInstances {
             G.map(f(a))((k, _))
           }) { chain => chain.foldLeft(Map.empty[K, B]) { case (m, (k, b)) => m.updated(k, b) } }
 
+      override def mapAccumulate[S, A, B](init: S, fa: Map[K, A])(f: (S, A) => (S, B)): (S, Map[K, B]) = {
+        val iter = fa.iterator
+        var s = init
+        val m = Map.newBuilder[K, B]
+        while (iter.hasNext) {
+          val (k, a) = iter.next()
+          val (snext, b) = f(s, a)
+          m += k -> b
+          s = snext
+        }
+        (s, m.result())
+      }
+
       override def map[A, B](fa: Map[K, A])(f: A => B): Map[K, B] =
         fa.map { case (k, a) => (k, f(a)) }
 
       def foldLeft[A, B](fa: Map[K, A], b: B)(f: (B, A) => B): B =
-        fa.foldLeft(b) { case (x, (k, a)) => f(x, a) }
+        fa.foldLeft(b) { case (x, (_, a)) => f(x, a) }
 
       def foldRight[A, B](fa: Map[K, A], lb: Eval[B])(f: (A, Eval[B]) => Eval[B]): Eval[B] =
         Foldable.iterateRight(fa.values, lb)(f)
--- a/alleycats-core/src/main/scala/alleycats/std/map.scala
+++ b/alleycats-core/src/main/scala/alleycats/std/map.scala
@@ -49,6 +49,7 @@ trait MapInstances {
         val iter = fa.iterator
         var s = init
         val m = Map.newBuilder[K, B]
+        m.sizeHint(fa.size)
         while (iter.hasNext) {
           val (k, a) = iter.next()
           val (snext, b) = f(s, a)
--- a/alleycats-core/src/main/scala/alleycats/std/set.scala
+++ b/alleycats-core/src/main/scala/alleycats/std/set.scala
@@ -120,6 +120,18 @@ trait SetInstances {
         }.value
       }
 
+      override def mapAccumulate[S, A, B](init: S, fa: Set[A])(f: (S, A) => (S, B)): (S, Set[B]) = {
+        val iter = fa.iterator
+        var s = init
+        val set = Set.newBuilder[B]
+        while (iter.hasNext) {
+          val (snext, b) = f(s, iter.next())
+          set += b
+          s = snext
+        }
+        (s, set.result())
+      }
+
       override def get[A](fa: Set[A])(idx: Long): Option[A] = {
         @tailrec
         def go(idx: Int, it: Iterator[A]): Option[A] =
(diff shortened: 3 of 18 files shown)


johnynek:
cc @BalmungSan

alleycats-core/src/main/scala/alleycats/std/iterable.scala
```diff
@@ -65,6 +65,18 @@ trait IterableInstances {
       override def traverse[G[_], A, B](fa: Iterable[A])(f: A => G[B])(implicit G: Applicative[G]): G[Iterable[B]] =
         if (fa.isEmpty) G.pure(Iterable.empty)
         else G.map(Chain.traverseViaChain(toImIndexedSeq(fa))(f))(_.toVector)
+
+      override def mapAccumulate[S, A, B](init: S, fa: Iterable[A])(f: (S, A) => (S, B)): (S, Iterable[B]) = {
```
johnynek, line 69: we can't use the StaticMethod implementation here because we aren't in the cats package.

alleycats-core/src/main/scala/alleycats/std/map.scala
```diff
@@ -45,11 +45,24 @@ trait MapInstances {
...
+        }
+        (s, m.result())
+      }
+
       override def map[A, B](fa: Map[K, A])(f: A => B): Map[K, B] =
         fa.map { case (k, a) => (k, f(a)) }
 
       def foldLeft[A, B](fa: Map[K, A], b: B)(f: (B, A) => B): B =
-        fa.foldLeft(b) { case (x, (k, a)) => f(x, a) }
+        fa.foldLeft(b) { case (x, (_, a)) => f(x, a) }
```
johnynek, line 65: clear a warning I saw.

alleycats-core/src/main/scala/alleycats/std/set.scala
```diff
@@ -120,6 +120,18 @@ trait SetInstances {
         }.value
       }
 
+      override def mapAccumulate[S, A, B](init: S, fa: Set[A])(f: (S, A) => (S, B)): (S, Set[B]) = {
```
johnynek, line 123: can't use StaticMethods version outside of cats package.

core/src/main/scala-2.13+/cats/data/NonEmptyLazyList.scala
```diff
@@ -213,7 +213,7 @@ class NonEmptyLazyListOps[A](private val value: NonEmptyLazyList[A])
    * Tests if some element is contained in this NonEmptyLazyList
    */
   final def contains(a: A)(implicit A: Eq[A]): Boolean =
-    toLazyList.contains(a)
+    toLazyList.exists(A.eqv(_, a))
```
johnynek, line 216: this looks like a straight bug. The contains method was ignoring the A. This is technically changing behavior, but I would argue it is a bug fix.

core/src/main/scala-2.13+/cats/instances/arraySeq.scala
```diff
@@ -103,12 +103,11 @@ private[cats] object ArraySeqInstances {
...
         B.combineAll(fa.iterator.map(f))
 
-      def traverse[G[_], A, B](fa: ArraySeq[A])(f: A => G[B])(implicit G: Applicative[G]): G[ArraySeq[B]] = {
-        def loop(i: Int): Eval[G[ArraySeq[B]]] =
-          if (i < fa.length) G.map2Eval(f(fa(i)), Eval.defer(loop(i + 1)))(_ +: _)
-          else Eval.now(G.pure(ArraySeq.untagged.empty))
-        loop(0).value
-      }
+      def traverse[G[_], A, B](fa: ArraySeq[A])(f: A => G[B])(implicit G: Applicative[G]): G[ArraySeq[B]] =
+        G.map(Chain.traverseViaChain(fa)(f))(_.iterator.to(ArraySeq.untagged))
```
johnynek, line 107: make traverse stack safe on ArraySeq. We missed this before.

core/src/main/scala/cats/data/Ior.scala
```diff
@@ -225,7 +225,12 @@ sealed abstract class Ior[+A, +B] extends Product with Serializable {
    * res2: Option[Int] = Some(123)
    * }}}
    */
-  final def right: Option[B] = fold(_ => None, b => Some(b), (_, b) => Some(b))
+  final def right: Option[B] =
+    this match {
```
johnynek, line 229: optimization to avoid lambdas when pattern matching will do.

core/src/main/scala/cats/data/Ior.scala
```diff
@@ -947,12 +952,27 @@ sealed abstract private[data] class IorInstances0 {
...
+            val (snext, c) = f(init, b)
+            (snext, Ior.Both(a, c))
+        }
       def foldLeft[B, C](fa: A Ior B, b: C)(f: (C, B) => C): C =
         fa.foldLeft(b)(f)
       def foldRight[B, C](fa: A Ior B, lc: Eval[C])(f: (B, Eval[C]) => Eval[C]): Eval[C] =
         fa.foldRight(lc)(f)
 
-      override def size[B](fa: A Ior B): Long = fa.fold(_ => 0L, _ => 1L, (_, _) => 1L)
+      override def size[B](fa: A Ior B): Long =
```
johnynek, line 971: optimization to avoid lambdas when pattern matching will do.

alleycats-core/src/main/scala/alleycats/std/iterable.scala
BalmungSan, line 69: What about also adding `mapWithIndex` and `zipWithIndex` here?

```scala
override def mapWithIndex[A, B](fa: Iterable[A])(f: (A, Int) => B): Iterable[B] =
  fa.zipWithIndex.map(ai => f(ai._1, ai._2))

override def zipWithIndex[A](fa: Iterable[A]): Iterable[(A, Int)] =
  fa.zipWithIndex
```

The motivation being they would preserve their underling class.

BalmungSan (dismissed):

alleycats-core/src/main/scala/alleycats/std/iterable.scala
BalmungSan: There is also no guarantee that the `Iterable` would be strict.

johnynek: good point.

johnynek: I made this change but I'm suspect about the trying to add invariants we don't really promise (returning the same underlying type).

1. this is already in alleycats where there are reasons to believe things are suspect. In this case Iterable is so weak it doesn't necessarily iterate in any meaningful order (otherwise it would be Seq).
2. when people start to assume we offer some stronger contracts, they often wind up getting into trouble and file issues and then we are in a jam. For instance, mapWithIndex is exercising a different code path from mapAccumulate here, and if they want some other law we don't promise (like maybe they do a cast at the end because they assume the Iterable is of the same type), then we get into weird issues of explaining why we went out of our way to maintain the class, but sometimes fail to.


anyway, it's alleycats, so I added it. But I hope people won't rely on it.

core/src/main/scala-2.13+/cats/instances/arraySeq.scala
satorg: This one is eluding me a bit... How come the previous implementation was not stack safe if each its call to `loop` is suspended in `Eval`? Just curious, because I even tried this snippet:
```scala
  val arr1 = ArraySeq.iterate(1, 100)(_ + 1)
  val arr2 = arr1.traverse { a =>
    println(s"\n>>> a = $a")
    new Exception().printStackTrace(Console.out)
    a.some
  }
```
And seems that stack depth is kept as O(1) for the entire `traverse` evaluation.
Or am I missing something?

johnynek: Does that work when you make 100 something like 100000?

The fn call you are printing is at constant depth but there is more to it, there is also the mapping function on the map2.

That said, we have hit many issues with traverse stack safety in the past and I think the approach used here was the same there (but maybe I am wrong).

satorg: > Does that work when you make 100 something like 100000?

Yes, it does even for 1,000,000 :) 

> The fn call you are printing is at constant depth but there is more to it, there is also the mapping function on the map2.
> 
> That said, we have hit many issues with traverse stack safety in the past and I think the approach used here was the same there (but maybe I am wrong).

I mean I'm not against the new implementation (perhaps it's supposed to be more performant). Just curious, because I tried out and didn't manage to screw up the old one :)

johnynek:
@armanbilge can you review this?

alleycats-core/src/main/scala/alleycats/std/map.scala
```diff
@@ -45,11 +45,24 @@ trait MapInstances {
             G.map(f(a))((k, _))
           }) { chain => chain.foldLeft(Map.empty[K, B]) { case (m, (k, b)) => m.updated(k, b) } }
 
+      override def mapAccumulate[S, A, B](init: S, fa: Map[K, A])(f: (S, A) => (S, B)): (S, Map[K, B]) = {
+        val iter = fa.iterator
+        var s = init
+        val m = Map.newBuilder[K, B]
```
danicheg, line 51: What do you think about adding the call of `m.sizeHint`? I haven't a strong opinion it's required, considering extra computing of `Map` length, but someone used it in cats.

danicheg (approved):
LGTM.

alleycats-core/src/main/scala/alleycats/std/map.scala
johnynek: good call. I'll update that.

core/src/main/scala/cats/data/Ior.scala
armanbilge: How about simply `if (fa.isLeft) 0L else 1L`.

armanbilge: Oh, yikes, `isLeft` and friends need to be optimized too (maybe by dynamic dispatch which is how `Option` is implemented?). Regardless, I still think this method would be better written in terms of `isLeft`.

alleycats-core/src/main/scala/alleycats/std/map.scala
johnynek: done.

core/src/main/scala/cats/data/Ior.scala
johnynek: done.

armanbilge (approved):
Thanks, LGTM!

